PR #1067, opened 2015-02-11
Fix an infinite loop bug inside template rendering

Kami:
Previously we ended up in an infinite loop if rendering a param failed.

### DIFF
--- a/st2actions/st2actions/utils/param_utils.py
+++ b/st2actions/st2actions/utils/param_utils.py
@@ -216,7 +216,6 @@ def _do_render_params(renderable_params, context):
     rendered_params = {}
     rendered_params.update(context)
 
-    keys_to_delete = []
     while len(renderable_params) != 0:
         for k, v in six.iteritems(renderable_params):
             template = env.from_string(v)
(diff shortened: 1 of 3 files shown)


st2actions/st2actions/utils/param_utils.py
```diff
@@ -215,15 +215,25 @@ def _do_render_params(renderable_params, context):
...
         for k, v in six.iteritems(renderable_params):
             template = env.from_string(v)
+
             try:
                 rendered = template.render(rendered_params)
                 rendered_params[k] = rendered
-            except:
+                keys_to_delete.append(k)
+            except Exception:
+                # TODO: Maybe we should throw here and abort action execution?
```
Kami: I actually believe throwing is a right course of action - we shouldn't just skip and ignore a problem if rendering a parameter fails.

@manasdk et al. wdyt?

Kami: I went ahead and made the throw change.

Previously, I just wanted to follow the existing flow and not change the logic, but the existing logic was broken anyway (we ended up in an infinite loop).

Now if rendering a param fail, we throw an exception which causes action to fail and the error is reported to the user.

Kami:
@lakshmi-kannan ^^ you can put your phd to a good use, verify that our solution is correct and provide a formal proof on termination condition :P

lakshmi-kannan:
I don't know about formal proof but I did tell manas one day we'll find an infinite loop in this code :P. I am lakstodamus now.

PR #1069, opened 2015-02-12
[RFR] Try literal_eval if json.loads fails.

manasdk:
* Often a user friendly json string may not be somthing json.loads like because it isn't properly escaped etc. For those cases ast.literal_eval can work.

### DIFF
--- a/st2client/st2client/commands/action.py
+++ b/st2client/st2client/commands/action.py
@@ -194,7 +194,11 @@ class ActionRunCommand(resource.ResourceCommand):
             # Also support simple key1=val1,key2=val2 syntax
             if value.startswith('{'):
                 # Assume it's JSON
-                result = value = json.loads(value)
+                result = None
+                try:
+                    result = value = json.loads(value)
+                except:
+                    result = value = ast.literal_eval(value)
             else:
                 pairs = value.split(',')
 


manasdk:
On one hand I think relaxing how json is input via the CLI is a good thing on the other it could lead to unexpected behavior.

Any suggestions?

lakshmi-kannan:
We have to decide high level if we are going to support nested structures as an argument from CLI. I propose we build our own syntax for complex structures. We already have comma separated list for arrays and k1=v1,k2=v2,.. for dicts. Let's not allow deeply nested input types on command like. We can support them using a -input option that takes in a file. Thoughts?

Kami:
I do see some merit in only allowing users to specify complex objects by passing a file path as the parameter value (that's what we did in the pyrax client).

This way we avoid annoying escaping, shell interpolation and other issues. In fact, I've recently pushed a change so now you should be able to read a value for every parameter from file if you use the following syntax: `@parameter_name=<file path>` - http://docs.stackstorm.com/latest/cli.html#reading-parameter-value-from-a-file

manasdk:
Yeah both those idea make sense. Although file seems like a great option as it is simpler to implement. 

Anything to not not have to deal with this shell escaping madness.

manasdk:
So no go on this one in that case.

PR #1076, opened 2015-02-13
Add another example of mistral workflow

enykeev:
The workflow exposes several flaws in mistral integration:
 - internal workflow call doesn't create another execution. First `task1` should be a child of `call_internal_workflow` and the second `task1` and `task2` should be childs of `call_internal_workflow_multiple_tasks`.
 - mistral workflow called from another mistral workflow (`call_external_workflow`) doesn't have task name in its context. For comparison, `call_action_chain` execution has it.
screen shot 2015-02-13 at 12 03 49


Kami:
I assume #1083 is attempt to fix the parent association problem?

Besides that, LGTM.

PR #1083, opened 2015-02-16
RFR: Hack to get task name for mistral nested workflows

lakshmi-kannan:
Before this PR, nested mistral workflows kicked off via st2 (we call it external workflows) had a problem with task name. 

Example workflow:
https://gist.github.com/lakshmi-kannan/492ca36919de99e6e1bf

LiveAction DB objects: (Pay attention to the 'context' fields)
https://gist.github.com/lakshmi-kannan/d65b09700d26d74f71dd

LiveActionDB objects after fix:
https://gist.github.com/lakshmi-kannan/3d5d859e1cacc7670788

### DIFF
--- a/st2actions/st2actions/runners/mistral/v2.py
+++ b/st2actions/st2actions/runners/mistral/v2.py
@@ -183,13 +183,38 @@ class MistralRunner(AsyncActionRunner):
 
         status = LIVEACTION_STATUS_RUNNING
         partial_results = {'tasks': []}
-        context = {
-            'mistral': {
-                'execution_id': str(execution.id),
-                'workflow_name': execution.workflow_name
-            }
+
+        current_context = {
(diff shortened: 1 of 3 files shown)


st2actions/st2actions/runners/mistral/v2.py
```diff
@@ -183,13 +184,38 @@ def run(self, action_parameters):
...
+        exec_context = copy.deepcopy(self.context)
+        exec_context = self._build_mistral_context(exec_context, current_context)
+        LOG.info('Mistral query context is %s' % exec_context)
+
+        return (status, partial_results, exec_context)
 
-        return (status, partial_results, context)
+    @staticmethod
+    def _build_mistral_context(parent, current):
+        context = dict()
```
Kami: Would be good to add some comment why we need to do that.

PR #1084, opened 2015-02-16
Fix pep8 failures causing build failure

Kami:


lakshmi-kannan:
Lint me baby one more time!

PR #1085, opened 2015-02-16
Store value of ComplexDateTimeField as number of microseconds since the epoch in the db

Kami:
With this change we now store values of ``ComplexDateTimeField`` fields as number of microseconds since the epoch internally in MongoDB.

MongoEngine's version of ``ComplexDateTimeField`` stores it as a comma delimited string (e.g. ``2011,06,08,20,26,24,192284``) which results in a wrong sort order when lleading zeros are present.

This should fix an issue with action execution tests intermediate failures.

Storing is as a number should also be more efficient in terms of storage space and retrieval / querying speed.

This should fix the bug with action execution tests failing.

TODO:

- [x] Tests

### DIFF
--- a/st2common/st2common/fields.py
+++ b/st2common/st2common/fields.py
@@ -56,7 +56,7 @@ class ComplexDateTimeField(LongField):
         :param data: Number of microseconds since the epoch.
         :type data: ``int``
         """
-        result = datetime.datetime.fromtimestamp(data//SECOND_TO_MICROSECONDS)
+        result = datetime.datetime.fromtimestamp(data // SECOND_TO_MICROSECONDS)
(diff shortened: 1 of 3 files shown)


manasdk:
Change looks good. Pushed lint fix to unblock travis.

+1

lakshmi-kannan:
This will probably fix STORM-1028. Can you close the task if that's the case?

Kami:
@lakshmi-kannan Yeah, should fix that issue.

I will resolve it once this pr is finished and merged :)

Kami:
Alright, pushed tests and a bug fix.

All the tests pass now so I'm merging this into master.

PR #1086, opened 2015-02-17
Handle mistral connection errors; don't delete state

lakshmi-kannan:
I'll cherry pick this change into hotfix branch for CITIES off sha 3376d9bc0a9c1377d6daf36381e57125a3654716

### DIFF
--- a/st2actions/st2actions/query/mistral/v2.py
+++ b/st2actions/st2actions/query/mistral/v2.py
@@ -42,13 +42,25 @@ class MistralResultsQuerier(Querier):
             status, output = self._get_workflow_result(exec_id)
             if output and 'tasks' in output:
                 LOG.warn('Key conflict with tasks in the workflow output.')
+        except requests.exceptions.ConnectionError:
+            msg = 'Unable to connect to mistral.'
+            LOG.exception(msg)
+            return (LIVEACTION_STATUS_RUNNING, {'error': msg})
         except:
             LOG.exception('Exception trying to get workflow status and output for '
                           'query context: %s. Will skip query.', query_context)
             raise
 
         result = output or {}
-        result['tasks'] = self._get_workflow_tasks(exec_id)
+        try:
+            result['tasks'] = self._get_workflow_tasks(exec_id)
+        except requests.exceptions.ConnectionError:
+            msg = 'Unable to connect to mistral.'
+            LOG.exception(msg)
+            return (LIVEACTION_STATUS_RUNNING, {'error': msg})
+        except:
+            LOG.exception('Unable to get workflow results for '
+                          'query_context: %s. Will skip query.', query_context)
 
         LOG.debug('Mistral query results: %s' % result)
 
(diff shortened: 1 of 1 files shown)


st2actions/st2actions/query/mistral/v2.py
```diff
@@ -42,13 +42,25 @@ def query(self, execution_id, query_context):
             status, output = self._get_workflow_result(exec_id)
             if output and 'tasks' in output:
                 LOG.warn('Key conflict with tasks in the workflow output.')
+        except requests.exceptions.ConnectionError:
+            msg = 'Unable to connect to mistral.'
+            LOG.exception(msg)
+            return (LIVEACTION_STATUS_RUNNING, {'error': msg})
```
Kami: We update the whole results objects every time `query` method returns, right?

If so, this means that during failure we will store error in the `result` attribute of a live action, right? Is this desirable?

st2actions/st2actions/query/mistral/v2.py
```diff
@@ -42,13 +42,25 @@ def query(self, execution_id, query_context):
             status, output = self._get_workflow_result(exec_id)
             if output and 'tasks' in output:
                 LOG.warn('Key conflict with tasks in the workflow output.')
+        except requests.exceptions.ConnectionError:
+            msg = 'Unable to connect to mistral.'
```
Kami: We should probably also include actual exception message and potentially also a traceback in the `msg` attribute since that's what we will store in the database.

st2actions/st2actions/query/mistral/v2.py
lakshmi-kannan: This is not a good solution at all. This will just let us know if there was a mistral connection error. If the connection reestablishes in a subsequent run, we would get the results back.

st2actions/st2actions/query/mistral/v2.py
lakshmi-kannan: I can add the traceback but it is pretty useless since we anyways know it's a connection error.

Kami: That's true, but it could make it easier for user to debug and spot issues due to invalid configuration (e.g. maybe mistral is down or mistral url specified in the config is invalid).

lakshmi-kannan: k, good point. let me add traceback as a field.

dzimine:
After chatting with @manasdk and @lakshmi-kannan: 
this is production-worthy change. The previous impl is not resilient. Perhaps should add some retry-giveup logic and merge to master. 

Almost make me want to add "WARNING" status to action execution :)

lakshmi-kannan:
>Perhaps should add some retry-giveup logic and merge to master.

That is a more involved change. I'll add it to the backlog.

Kami:
Do we also plan to get this in the v0.8 release (master)?

lakshmi-kannan:
>Do we also plan to get this in the v0.8 release (master)?

I guess we can merge this to master. Let me do that.

dzimine:
IMO we shall merge.

PR #1087, opened 2015-02-17
Liveaction execution api changes

manasdk:
- [x] Verify that st2client tests are properly written
- [x] Verify the display attributes of st2 execution list and st2 execution get
- [ ] Pick up changes in st2web and st2client.js
- [x] more tests
      - [x] tests for get_descendants
      - [x] test for new API URL


st2common/st2common/services/executions.py
```diff
@@ -42,6 +44,22 @@
...
+    """
+    Splits the liveaction into an ActionExecution compatible dict.
+    """
+    decomposed = {'liveaction': {}}
+    liveaction_api = vars(LiveActionAPI.from_model(liveaction_db))
+    for k in liveaction_api.keys():
+        if k in SKIPPED:
+            decomposed['liveaction'][k] = liveaction_api[k]
+        else:
+            decomposed[k] = getattr(liveaction_db, k)
```
lakshmi-kannan: I hate this. Since LiveAction is a DictField in execution db, we need not have db representation of values. For other fields like start_timestamp, end_timestamp, we should use db representations. This code is error prone. I think we need to pause here and think this through. Otherwise, we are gonna have problems maintaining this.

manasdk: I see your point. Luckily not much of LiveAction is stored any longer but the same would apply to other fields like Action, Rule, Trigger etc in an ActionExecution.

What do you suggest we do?

manasdk:
@Kami : We have passibg test. Now we can add the remaining action in CLI. Also, look at +get_children and corresponding URL.

PR #1088, opened 2015-02-17
New CLI command for re-running an action

Kami:
This pull request adds a new CLI command for re-running an existing action (new action is ran with the same set of the input parameters as the existing one).

As discussed in the first iteration it's implemented fully as a client side feature. If more people think it's a good idea to also have this functionality in the UI, we can change it and implement it on the server side.

Implementing it as a new endpoint on the server is fully backward compatible. CLI syntax will stay the same, only thing which will change is the internal implementation of the command.


dzimine:
LGTM

PR #1090, opened 2015-02-18
Update docs to include proper Sensor sample.

manasdk:


Kami:
LGTM :panda_face:

As discussed before, I would prefer to put full code examples such as a sensors and actions in a file and include that file in the docs. Ideally same file would also be used by a test - this way we can be sure that the examples work.

PR #1093, opened 2015-02-19
Add retry logic to mistral API calls

m4dcoder:
Add retry logic to mistral API calls to address occasional network errors.

### DIFF
--- a/st2actions/st2actions/config.py
+++ b/st2actions/st2actions/config.py
@@ -52,14 +52,11 @@ ssh_runner_opts = [
 ]
 CONF.register_opts(ssh_runner_opts, group='ssh_runner')
 
-workflow_opts = [
-    cfg.StrOpt('url', default='http://localhost:8989', help='Mistral API server root endpoint.')
-]
-CONF.register_opts(workflow_opts, group='workflow')
-
 mistral_opts = [
-    cfg.StrOpt('v2_base_url', default='http://localhost:8989/v2/',
+    cfg.StrOpt('v2_base_url', default='http://localhost:8989/v2',
                help='Mistral v2 API server root endpoint.'),
+    cfg.IntOpt('max_attempts', default=180),
+    cfg.IntOpt('retry_wait', default=5)
 ]
 CONF.register_opts(mistral_opts, group='mistral')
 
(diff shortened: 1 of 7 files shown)


st2actions/st2actions/config.py
```diff
@@ -52,14 +52,11 @@
...
-workflow_opts = [
-    cfg.StrOpt('url', default='http://localhost:8989', help='Mistral API server root endpoint.')
-]
-CONF.register_opts(workflow_opts, group='workflow')
-
 mistral_opts = [
-    cfg.StrOpt('v2_base_url', default='http://localhost:8989/v2/',
+    cfg.StrOpt('v2_base_url', default='http://localhost:8989/v2',
                help='Mistral v2 API server root endpoint.'),
+    cfg.IntOpt('max_attempts', default=180),
```
Kami: Hm, 180 attempts ((180-1)*5)?

This seems way too much to me - or is there a good reason for such a high number of attempts? If we do need such a higher number of attempts / retry for so long we should maybe instead try with exponentiation retry delay which is reset on success.

st2actions/st2actions/query/mistral/v2.py
```diff
@@ -42,23 +43,37 @@ def query(self, execution_id, query_context):
...
+        except:
+            LOG.exception('Unable to get workflow results for '
+                          'query_context: %s. Will skip query.', query_context)
 
         LOG.debug('Mistral query results: %s' % result)
 
         return (status, result)
 
     def _get_execution_tasks_url(self, exec_id):
-        return self._base_url + 'executions/' + exec_id + '/tasks'
```
Kami: We should use some kind of util function to join the base url and a path - this double trailing slash has bitten me before.

Or some kind of util "get_base_url" function which makes sure the returned url never contains a trailing slash.

Kami: #1095

st2actions/st2actions/config.py
m4dcoder: That equates to 15 min.  We can dial that down.  It's configurable anyway.  5 min?

st2actions/st2actions/query/mistral/v2.py
m4dcoder: Thanks for the patch.  It's cool.  This makes thing more consistent.  Otherwise, we have URL config that has trailing slash and some don't.  So messy...

PR #1096, opened 2015-02-19
Update HTTP runner to parse JSON response body if Content-Type is application/json

Kami:
With this change HTTP runner now parses a body if a response contains a known Content-Type.

For now, we only parse JSON responses with Content-Type of `application/json`.

To make it easier for the user, I've also added `parsed` flag to the result. This flag indicates if the response body has been parsed or not.

In the future, we can also add support for parsing other content-types, e.g. XML and HTML (e.g. with xmltodict and beautifulsoup library).

XML and HTML don't really directly map to a dictionary, so this feature would probably be an opt-in via a runner parameter.

On a semi related note - unless I've failed at grepping, we don't have any tests for the HTTP runner.


manasdk:
+1

PR #1098, opened 2015-02-20
[RFR] Bootstrap rabbit

manasdk:
* Command to bootstrap rabbitmq. For now only exchanges are created.
* Command is called from all start/launch scripts.

Note : Since it is a bootstrap step decided to keep it outside. Similar to the registercontent step just that it is configuring an external component.

Fixes : https://stackstorm.atlassian.net/browse/STORM-1020

### DIFF
--- a/st2common/bin/bootstraprmq.py
+++ b/st2common/bin/bootstraprmq.py
@@ -1,0 +1,7 @@
+#!/usr/bin/env python2.7
+
+import sys
+import st2common.transport.bootstrap as transport_bootstrap
+
+if __name__ == '__main__':
+    sys.exit(transport_bootstrap.main())
--- a/st2actions/st2actions/cmd/actionrunner.py
+++ b/st2actions/st2actions/cmd/actionrunner.py
@@ -8,6 +8,7 @@ from st2common import log as logging
 from st2common.models.db import db_setup
 from st2common.models.db import db_teardown
 from st2common.constants.logging import DEFAULT_LOGGING_CONF_PATH
+from st2common.transport.utils import register_exchanges
 from st2actions import config
 from st2actions import worker
 
@@ -40,7 +41,7 @@ def _setup():
     password = cfg.CONF.database.password if hasattr(cfg.CONF.database, 'password') else None
     db_setup(cfg.CONF.database.db_name, cfg.CONF.database.host, cfg.CONF.database.port,
              username=username, password=password)
-
+    register_exchanges()
     # 4. Register internal triggers
     # Note: We need to do import here because of a messed up configuration
     # situation (this module depends on configuration being parsed)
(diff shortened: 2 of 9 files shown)


st2common/bin/bootstraprmq.py
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env python2.7
+
+import sys
+import st2common.transport.bootstrap as transport_bootstrap
+
+if __name__ == '__main__':
```
Kami: Couldn't we do this as part of the regular app bootstrap process - the same way we ensure MongoDB indexes on startup?

manasdk: Well I dislike that as well.

This way all processes don't end up bootstrapping RMQ. Also, there is a degree of control and can be done from outside.

Kami: Yeah, but the operation is idempotent, so multiple calls are not a problem, right?

And this way, we can be sure the exchanges will always be present - no surprises if user doesn't use launchdev / st2_deploy, etc.

And we can still leave the script, in case, for some reason user wants to also run it manually.

manasdk: Yes. Ops are idempotent but that does not mean we repeat them knowingly. 

Not wanting to use launchdev or st2_deploy is a very narrow usecase. Besides by this approach it needs to have run once before is all. 

You seem to have strong opinions against this approach so I shall change it around to your preference. I actually prefer this explicit way to doing things.

Kami: Among other things, I also thing it's good to do it this way so it's consistent with MongoDB index creation process which happens during bootstrap phase.

manasdk: Not all consistency is a good idea :P

I really think that give our architecture each process ensuring the same indexes is actually a poor implementation choice. We really should have more control over some of these bootstrap steps.

But like I said I am making that change you ask for.

Kami: Well, otherwise, in a distributed system we would need a coordinator if we wanted to ensure this piece of bootrstrap code was executed and this makes implementation and everything else more complex (and potentially error prone).

Relying on the idempotent nature of operation (where possible) and potentially calling it multiple times, is imo a better and cleaner approach. It might have a small performance penalty / increase startup time, but services are long running so shouldn't be a big deal in this case :)

manasdk:
rewrote the commit history.

Kami:
LGTM, +1

st2common/bin/bootstraprmq.py
manasdk: > Well, otherwise, in a distributed system we would need a coordinator
We definitely need something of this sort at some point.

> performance penalty / increase startup time
That does not bother me too much either.

> potentially calling it multiple times, is imo a better and cleaner approach.
Clean is too generic a term. Can you explain why it is a cleaner approach? If you said each process was independent then to some extent I would buy the argument. However, I do find it a complete abstraction leak when st2action is ensuring indexes for reactor objects. 

We can both make points in either direction and eventually it ends up at preference. I am overwhelmingly against doing work which need not be performed along a specific code-path. This really leads to confusing behavior in terms of questions like -
1. What the heck happened why did this get setup?
2. Whose responsibiilty is it really to perform this task? Everyones is not really a good answer.

manasdk:
Travis error - failed to connect to pypi.

Kami:
Yeah, happens quite often recently. I usually just manually restart the build in travis website.

PR #1099, opened 2015-02-20
Add mistral WF example for rollback and retry

m4dcoder:


contrib/examples/actions/mistral-handle-retry.yaml
```diff
@@ -0,0 +1,30 @@
...
+                action: core.local
+                input:
+                    cmd: "if [ ! -e \"/tmp/done\" ]; then exit 1; fi"
+                on-error:
+                    - undo
+            undo:
+                action: core.local
+                input:
+                    cmd: "echo \"File does not exist.\""
+                on-success:
```
dzimine: `on-complete`, I think.

contrib/examples/actions/mistral-handle-retry.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+    main:
+        type: direct
+        tasks:
+            task1:
+                workflow: work
+                policies:
+                    retry:
+                        count: 20
+                        delay: 3
```
dzimine: would be good to use break-on, but I can't come up with idea how.

contrib/examples/actions/mistral-handle-retry.yaml
```diff
@@ -0,0 +1,30 @@
+version: '2.0'
+name: "examples.mistral-handle-retry"
+
+workflows:
+
+    main:
+        type: direct
+        tasks:
+            task1:
```
dzimine: To make it fully automated, consider a parallel task with delay, which creates the file there.

contrib/examples/actions/mistral-handle-retry.yaml
m4dcoder: yeah, the thing is what if the undo failed...  right now, the whole thing retries regardless.   so the undo WF must be pretty reliable.

contrib/examples/actions/mistral-handle-retry.yaml
```diff
@@ -0,0 +1,30 @@
+version: '2.0'
+name: "examples.mistral-handle-retry"
+
+workflows:
+
```
dzimine: I would use description to tell the story of what the workflow is doing.

contrib/examples/actions/mistral-handle-retry.yaml
dzimine: task1 -> test_retry?

contrib/examples/actions/mistral-handle-retry.yaml
m4dcoder: i tried break-on.  the thing is the retry policy don't have an option to run an undo.

dzimine:
Looks good. few comments, up to you.

contrib/examples/actions/mistral-handle-retry.yaml
m4dcoder: good idea

contrib/examples/actions/mistral-handle-retry.yaml
m4dcoder: maybe i'll add a break-on if the undo fails.

m4dcoder: a bit hackish though...

PR #1100, opened 2015-02-20
Add mistral WF examples on use of with-items

m4dcoder:


m4dcoder:
Yeah, I don't have a problem with that.  I can do another commit.  It's not a problem.

manasdk:
Ok. Then I will leave it to you.

Thanks!

PR #1108, opened 2015-02-23
Sensor wrapper global datastore access

Kami:
TODO:

- [x] tests

/cc @jfryman

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -61,6 +61,8 @@ Docs: http://docks.stackstorm.com/latest
 * Allow sensors to manage global datastore items via sensor_service by passing ``local=False``
   argument to the ``get_value``, ``set_value`` and ``delete_value`` methods. (new-feature)
 * Allow sensors to list datastore items using ``list_values`` sensor_service method. (new-feature)
+* Allow users to filter datastore items by name prefix by passing ``?prefix=<value>`` query
+  parameter to the /keys endpoint. (new-feature)
 
 v0.7 - January 16, 2015
 -----------------------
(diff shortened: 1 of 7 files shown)


jfryman:
`list_value` only? Is it possible to also have a query? I'd like to avoid having responsibility of the sensor to filter.

jfryman:
Also, :metal: and thanks! :heart:

Kami:
@jfryman What kind of filtering do you have in mind?

If you mean just prefix filtering, then, yes, I can add that - I also planned to add this in the future (just wanted to reduce the scope for now) since we already perform this filtering on the client side in the CLI now.

jfryman:
I'd like to be able to do some basic regex searches, or 'contains' searches.

My exact use case: I may have N sets of attributes that are namespaced with `asg.{ASGNAME}.{ASGVALUE}`. So, I'd like to be able to look for maybe all of the keys that are `asg.XXX.alert_status`, and then dig in from there.

I mean, I can easily filter in the sensor itself with some additional code, just feels wrong. :) I don't mind punting, but I fear that this will be a single example might propagate into an anti-pattern very quickly.

Kami:
Alright, in this case, I would really like to avoid users being able to specify regex - this can result in pathological cases, ddos and many other issues.

It sounds like prefix by filter should be a good compromise though - you still need to do late filtering yourself but should be less data :D

jfryman:
> It sounds like prefix by filter should be a good compromise though - you still need to do late filtering yourself but should be less data :D

WFM! :+1:

Kami:
@jfryman Alright, now inside the sensor you can do:

```python
asg_kvps = self._sensor_service.list_values(local=False, prefix='asg.')
```

Kami:
This should also hopefully be helpful now - https://github.com/StackStorm/st2/pull/1122/files#diff-3131793d811778116d7d19aa2e34ee91R62

st2stanley:
Yesssssssss!

-James

PR #1111, opened 2015-02-23
adding tool to clean/setup dbs

jfryman:
Simple command to just handle DB cleanup. Taken from `st2ctl` and `st2_express.sh`.

### DIFF
--- a/tools/db_cleanup.sh
+++ b/tools/db_cleanup.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+# Simple app that runs DB cleanup commands, and repopulates from disk
+
+echo "Cleaning MongoDB Database..."
+mongo st2 --eval "db.dropDatabase();"
+
+echo "Cleaning Mistral Database..."
+mysql -uroot -pStackStorm -e "DROP DATABASE IF EXISTS mistral"
+mysql -uroot -pStackStorm -e "CREATE DATABASE mistral"
+mysql -uroot -pStackStorm -e "GRANT ALL PRIVILEGES ON mistral.* TO 'mistral'@'localhost' IDENTIFIED BY 'StackStorm'"
+mysql -uroot -pStackStorm -e "FLUSH PRIVILEGES"
+python ./tools/sync_db.py --config-file /etc/mistral/mistral.conf
+
+echo "Re-initializing DBs..."
+st2ctl reload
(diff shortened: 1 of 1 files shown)


tools/db_cleanup.sh
```diff
@@ -0,0 +1,19 @@
...
+mongo st2 --eval "db.dropDatabase();"
+
+echo "Cleaning Mistral Database..."
+mysql -uroot -p$ROOT_PASSWORD -e "DROP DATABASE IF EXISTS mistral"
+mysql -uroot -p$ROOT_PASSWORD -e "CREATE DATABASE mistral"
+mysql -uroot -p$ROOT_PASSWORD -e "GRANT ALL PRIVILEGES ON mistral.* TO 'mistral'@'localhost' IDENTIFIED BY '$ROOT_PASSWORD'"
+mysql -uroot -p$ROOT_PASSWORD -e "FLUSH PRIVILEGES"
+
+cd /opt/openstack/mistral
+python ./tools/sync_db.py --config-file /etc/mistral/mistral.conf
```
lakshmi-kannan: I think one needs to run this tool only when one registers external mistral actions. Check with @dzimine.

lakshmi-kannan:
<3.

tools/db_cleanup.sh
jfryman: Yeah, I'm pretty sure this has to do only with the `core` mistral actions, as I've been bitten by this bug while testing out workflows. However, tough to assume that they won't be using 100% StackStorm, no, so I thought better to just include.

@dzimine comments here appreciated! :grinning:

dzimine: std mistral actions like "std.echo" are there too; and may be even our proxy action requires this. So keep it :)

tools/db_cleanup.sh
```diff
@@ -0,0 +1,19 @@
...
+# Simple app that runs DB cleanup commands, and repopulates from disk
+ROOT_PASSWORD=${MYSQL_PASSWORD:StackStorm}
+
+
+echo "Cleaning MongoDB Database..."
+mongo st2 --eval "db.dropDatabase();"
+
+echo "Cleaning Mistral Database..."
+mysql -uroot -p$ROOT_PASSWORD -e "DROP DATABASE IF EXISTS mistral"
+mysql -uroot -p$ROOT_PASSWORD -e "CREATE DATABASE mistral"
```
dzimine: when mysql is set without password this will ask password 4 times, that's why I prefer to fire a single SQL statement with ; in between: `"DROP DATABASE IF EXISTS mistral;CREATE DATABASE mistra;.....l"`

tools/db_cleanup.sh
jfryman: :+1:

PR #1112, opened 2015-02-24
Fix bug where rerunning pack install creates a recursive dir inside pack

lakshmi-kannan:
See: https://stackstorm.atlassian.net/browse/STORM-1065


Kami:
Nice.

I wonder though if maybe we want to add "force" flag to download / install (e.g. packs.install force=true ...)

lakshmi-kannan:
>I wonder though if maybe we want to add "force" flag to download / install (e.g. packs.install force=true ...)

I thought about the same too. Yeah, let's do this. I'll work on it.

PR #1117, opened 2015-02-24
RFR: Add support for exists and nexists operators

lakshmi-kannan:
Not too happy with the changes. Things work but the op_func always assume there is a value and there is criteria_pattern. exists and nexists don't fit into this category. We'll need a better interface. I am still thinking about this.

### DIFF
--- a/st2common/st2common/operators.py
+++ b/st2common/st2common/operators.py
@@ -38,7 +38,7 @@ def get_operator(op):
 
 
 def equals(value, criteria_pattern):
-    if not criteria_pattern:
+    if criteria_pattern is None:
         return False
     return value == criteria_pattern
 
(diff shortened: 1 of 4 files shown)


st2common/st2common/operators.py
```diff
@@ -22,10 +22,24 @@
...
+    if op in operators:
+        return operators[op]
+    else:
+        raise Exception('Invalid operator: ' + op)
+
 # Operation implementations
 
 
 def equals(value, criteria_pattern):
+    if not criteria_pattern:
```
manasdk: If criteria_pattern = "" and value = "" this would return false?

manasdk:
>  Things work but the op_func always assume there is a value and there is criteria_pattern.

Seems like the right idea - no?

st2common/st2common/operators.py
```diff
@@ -22,10 +22,24 @@
     'get_allowed_operators'
 ]
 
+
+def get_allowed_operators():
+    return operators
+
+
+def get_operator(op):
+    op = op.lower()
```
manasdk: Why? 

equals, EQUALS, eQualS are all ok?

lakshmi-kannan: I just moved this code.

manasdk:
I don't get the `if not criteria_pattern:` part of the change. Why is it required to fix this issue?

lakshmi-kannan:
>Seems like the right idea - no?

exists and nexists don't have a criteria_value.

st2common/st2common/operators.py
lakshmi-kannan: fml. python. bah.

manasdk:
> exists and nexists don't have a criteria_value.

Do we need those operators right now?

lakshmi-kannan:
>Do we need those operators right now?

That's the intent for writing this code. Look at https://stackstorm.atlassian.net/browse/STORM-1083.

manasdk:
+1

Kami:
Need to update the changelog and document this behavior in the rules section of the documentation.

st2common/st2common/operators.py
```diff
@@ -22,10 +22,24 @@
...
+    if op in operators:
+        return operators[op]
+    else:
+        raise Exception('Invalid operator: ' + op)
+
 # Operation implementations
 
 
 def equals(value, criteria_pattern):
+    if criteria_pattern is None:
```
Kami: I was against like doing something like that in the past (I prefer being explicit and have strong types where possible - if it won't break in the rule not, it might break somewhere down the pipeline though), but oh well, lets say I'm OK with it now.

PR #1123, opened 2015-02-25
[WIP] WebUI deploy and static file serving changes

Kami:
TODO:

* Documentation
* Allow user to disable webui static file serving in the config

Blocked by:

* https://github.com/StackStorm/st2incubator/pull/65

### DIFF
--- a/st2api/st2api/app.py
+++ b/st2api/st2api/app.py
@@ -13,12 +13,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from oslo.config import cfg
 import pecan
+from oslo.config import cfg
 
 from st2common import hooks
 from st2common import log as logging
 from st2common.constants.system import VERSION_STRING
+
+__all__ = [
+    'get_api_app',
+    'get_webui_app'
+]
 
 
 LOG = logging.getLogger(__name__)
@@ -30,7 +35,6 @@ def __get_pecan_config():
     cfg_dict = {
         'app': {
             'root': opts.root,
-            'static_root': opts.static_root,
             'template_path': opts.template_path,
             'modules': opts.modules,
             'debug': False,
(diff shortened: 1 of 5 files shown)


st2api/st2api/cmd/api.py
```diff
@@ -64,10 +65,25 @@ def _run_server():
...
+    webui_host = api_host
+    webui_port = int(api_port) + 1
+
+    LOG.info('(PID=%s) ST2 API is serving on http://%s:%s.', os.getpid(), api_host, api_port)
+    LOG.info('(PID=%s) ST2 WebUI is serving on http://%s:%s.', os.getpid(), webui_host, webui_port)
+
+    def _run_api_server():
+        wsgi.server(eventlet.listen((api_host, api_port)), get_api_app())
+
+    def _run_webui_server():
```
Kami: This approach is not ideal, but pecan doesn't allow multiple static_roots, so that's a compromise I had to make. I could make it to work by re-using the same app, but this means I would need to set `debug=True` for the actual API server.

To make it easier to deploy and run, I just run it as part of the st2api process on port api_port + 1. Additional process and configuration file just makes deployment more complicated which is unnecessary.

And to make it clear (we should point this out in the docs) this server is there only for convenience - for actual production deployments, people should serve static files using nginx / apache.

st2api/st2api/app.py
```diff
@@ -30,7 +35,6 @@ def __get_pecan_config():
     cfg_dict = {
         'app': {
             'root': opts.root,
-            'static_root': opts.static_root,
```
Kami: static_root is ignored if debug is not True.

st2api/st2api/cmd/api.py
Kami: One of the problems with serving it using a separate server is that users will still need to configure CORS headers :/

Kami: To clarify it - if we want to serve it as part of the api server, we need to set debug=True which is less than ideal to put it lightly.

Kami:
Will open a PR with approach we have agreed upon on Slack.

PR #1124, opened 2015-02-25
Update st2api server to also server webui static files

Kami:
WebUi files are only served if `serve_webui` option is enabled. In documentation we will encourage users to disable this option in production and use something like nginx or Apache to serve static webui files there.

TODO:

* [x] Docs

Blocked by:

* https://github.com/StackStorm/st2incubator/pull/65

### DIFF
--- a/st2api/st2api/controllers/root.py
+++ b/st2api/st2api/controllers/root.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 from pecan import expose
+from pecan.core import redirect
 
 from st2common import __version__
 from st2common import log as logging
@@ -22,12 +23,22 @@ import st2api.controllers.v1.root as v1_root
 LOG = logging.getLogger(__name__)
 
 
+class WebUIRootController(object):
+    @expose(generic=True)
+    def index(self):
+        return redirect(location='/webui/index.html')
+
+
 class RootController(object):
 
     def __init__(self):
-        v1 = v1_root.RootController()
-        self.controllers = {'v1': v1}
-        self.default_controller = v1
+        v1_controller = v1_root.RootController()
+        webui_controller = WebUIRootController()
+        self.controllers = {
+            'v1': v1_controller,
+            'webui': webui_controller
+        }
+        self.default_controller = v1_controller
 
     @expose(generic=True, template='index.html')
     def index(self):
@@ -47,6 +58,10 @@ class RootController(object):
         version = ''
         if len(remainder) > 0:
             version = remainder[0]
+            if version == 'webui':
+                # Special case for handling webui requests
+                return self.controllers['webui'], remainder[1:]
+
             if remainder[len(remainder) - 1] == '':
                 # If the url has a trailing '/' remainder will contain an empty string.
                 # In order for further pecan routing to work this method needs to remove
--- a/st2api/st2api/controllers/root.py
+++ b/st2api/st2api/controllers/root.py
@@ -21,6 +21,11 @@ from st2common import __version__
 from st2common import log as logging
 import st2api.controllers.v1.root as v1_root
 
+__all__ = [
+    'WebUIRootController',
+    'RootController'
+]
+
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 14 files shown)


dzimine:
While with that, @enykeev suggested we stop serving static errors from the API services.

manasdk:
> While with that, @enykeev suggested we stop serving static errors from the API services.

We still serve static errors from API? I thought we removed all of that crap already.

Kami:
I have no idea what "static errors" are supposed to be so that probably means that, yeah, this has already been removed :)

enykeev:
> I have no idea what "static errors" are supposed to be so that probably means that, yeah, this has already been removed :)

Sadly, this is not how it works.

```# enykeev at mechanica-velocity.local in ~ [9:38:34]
$ http 172.168.50.50:9101/non-existent/
HTTP/1.1 404 Not Found
Access-Control-Allow-Headers: Content-Type,Authorization,X-Auth-Token
Access-Control-Allow-Origin: http://104.236.26.111:3000
Access-Control-Expose-Headers: Content-Type,X-Limit,X-Total-Count
Content-Length: 154
Content-Type: text/html; charset=UTF-8
Date: Thu, 26 Feb 2015 03:28:07 GMT

<html>
 <head>
  <title>404 Not Found</title>
 </head>
 <body>
  <h1>404 Not Found</h1>
  The resource could not be found.<br /><br />



 </body>
</html>
```

st2common/st2common/hooks.py
```diff
@@ -37,14 +37,38 @@ def after(self, state):
...
+
+        if serve_webui_files:
+            # Local API URL
+            default_allowed_origins.append('http://localhost:%s' % (api_port))
+            default_allowed_origins.append('http://127.0.0.1:%s' % (api_port))
+
+        if serve_webui_files and public_api_url:
+            # Public API URL
+            default_allowed_origins.append(public_api_url)
+
```
enykeev: Strictly speaking, you don't need this two IF's, the only thing you need to make sure is that url in your webui config exactly matches the url you are opening in your browser.

So, say, we have default devenv running in vm. Browser won't ask server for CORS if you open http://172.168.50.50:9101/webui in your browser and then pick Devenv from drop down (which also corresponds to http://172.168.50.50:9101). But if you for some reason would manage to run the browser on a VM itself and try to access web ui as http://localhost:9101/webui while still picking Devenv as an API server, you'll get an error because your Origin `http://localhost:9101` doesn't match API address `http://172.168.50.50:9101`. And it wont be a problem if you create another webui config record that points to http://localhost:9101 instead.

Yeah, CORS is kind of awful this way...

Kami:
Ah, you mean not always returning JSON. I agree, that's nasty.

Kami:
Pushed a fix for response body not always being JSON in 4c2400f53be193925142ed11dfeb2d14003aa425

st2common/st2common/hooks.py
Kami: True - so what do you think, should we keep it or remove it?

enykeev: On one hand, it might be confusing for someone who knows how it works. On the other hand, it might same some time for those who don't. I don't see any security risk here, so I say we should keep and carefully document that. Like, "we are adding origins A,B and C for your convenience, but we still suggest you to add every origin you are going to use manually".

st2api/st2api/controllers/root.py
```diff
@@ -14,20 +14,40 @@
...
+    'RootController'
+]
+
 LOG = logging.getLogger(__name__)
 
 
+class WebUIRootController(object):
+    @expose(generic=True)
+    def index(self):
+        return redirect(location='/webui/index.html')
```
Kami: To clarify it - redirect is there because of the way static middleware works - there is no support for index files and directory listings, so without a redirecting, HTTP server would previously return 404 not found.

So this is basically just a workaround (we also can't and dont want to use "template" functionality here - templates are in a different directory) :)

PR #1131, opened 2015-02-27
Mistral v0.8 Workflow Upgrade Tool

jfryman:
This is an upgrade tool that parses 0.7 workflows and spits out 0.8 compatible workflows. This is for the users out there that we may not know about (HI! :wave:) that may be using Mistral, giving them a painless way to upgrade when the time comes.

### DIFF
--- a/tools/upgrade-mistral-wf-0.8
+++ b/tools/upgrade-mistral-wf-0.8
@@ -18,7 +18,7 @@
 #######################################################################
 
 function usage {
-  echo "Usage: `basename $0` <workflow file>"
+  echo "Usage: `basename $0` <mistral file>"
 }
 
 # Exit Codes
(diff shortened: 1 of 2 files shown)


tools/upgrade-mistral-wf-0.8
```diff
@@ -0,0 +1,78 @@
...
+## Mistral YAQL Migration tool
+#
+# This tool inspects a given Mistral workflow, parses it using a few
+# regex filters, and attempts to rewrite the file to adhere to the
+# updated YAQL syntax introduced in https://review.openstack.org/#/c/157741/
+#
+# There appear to be two different ways that the old style YAQL is
+# represented in workflows. This includes:
+#   * bare         - (node: $.node)
+#   * interpolated - (node: 'Node: {$.node}')
```
dzimine: pls be sure to catch { 1 + $.node } expressions.

tools/upgrade-mistral-wf-0.8
```diff
@@ -0,0 +1,78 @@
...
+#   * functions    - (node: range(0, $.nodes).list() )
+#
+# This command runs sed with two passes attempting to clean up a given
+# workflow. This command will also ensure a backup exists in the event
+# that things do not go as expected to prevent data loss.
+#######################################################################
+
+function usage {
+  echo "Usage: `basename $0` <workflow file>"
```
dzimine: Technically it's 'mistral file', workflows, workbooks, actions...
say workflow/workbook file?

tools/upgrade-mistral-wf-0.8
jfryman: Added https://github.com/StackStorm/st2/pull/1131/files#diff-57b2224ac653d88d3d86728c4be4c945R70

jfryman:
Alright, this should be GTG for release. /cc @dzimine @lakshmi-kannan @DoriftoShoes

PR #1132, opened 2015-02-27
Cli help

manasdk:
* Fix wording
* Add samples

### DIFF
--- a/st2client/st2client/commands/action.py
+++ b/st2client/st2client/commands/action.py
@@ -582,15 +582,19 @@ class ActionExecutionListCommand(resource.ResourceCommand):
 
         # Filter options
         self.group.add_argument('--action', help='Action reference to filter the list.')
-        self.group.add_argument('--status', help='Only return executions with the provided status.')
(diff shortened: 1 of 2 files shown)


st2client/st2client/commands/action.py
```diff
@@ -582,15 +582,19 @@ def __init__(self, resource, *args, **kwargs):
 
         # Filter options
         self.group.add_argument('--action', help='Action reference to filter the list.')
-        self.group.add_argument('--status', help='Only return executions with the provided status.')
+        self.group.add_argument('--status', help=('Only return executions with the provided status.'
+                                                  ' Possible values are \'succeded\', \'running\','
```
Kami: Typo - should be `succeed` :P

But yeah, sadly, since it's a CLI, we can't use constants from st2common. Anyway, might still be better to add `POSSIBLE_ACTION_STATUS_VALUES = []` module level variable and use that. Easier to update if needed.

manasdk: https://github.com/StackStorm/st2/pull/1135

PR #1134, opened 2015-02-27
Promote linux pack to st2

dzimine:
Please double-check the packaging changes.


DoriftoShoes:
+1

manasdk:
IMO the rule and sensor should be removed.

Kami:
Agreed about the rule, sensor just needs some work to use trigger parameters instead of the config.

PR #1136, opened 2015-02-27
Remove Sensu sample from linux pack.

dzimine:


dzimine:
moved to Sensu pack in https://github.com/StackStorm/st2contrib/pull/144

PR #1138, opened 2015-02-27
Fixes in st2_deploy

dzimine:
- [x] Set up config.js for st2web
- [x] Quiet pip to reduce output verbosity, pronounced sections
- [x] Fix current stable version (was 0.6.0 for some reason)
- [x] add "latest" key to install TOT latest

### DIFF
--- a/tools/st2_deploy.sh
+++ b/tools/st2_deploy.sh
@@ -4,11 +4,15 @@ function version_ge() { test "$(echo "$@" | tr " " "\n" | sort -V | tail -n 1)" 
 
 if [ -z $1 ]
 then
-  VER='0.6.0'
+  VER='0.7'
+elif [[ "$1" == "latest" ]]; then
+   VER='0.8dev'
 else
   VER=$1
 fi
 
+echo "Installing version ${VER}"
+ 
 INSTALL_ST2CLIENT=${INSTALL_ST2CLIENT:-1}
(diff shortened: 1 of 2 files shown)


tools/st2_deploy.sh
```diff
@@ -4,11 +4,15 @@ function version_ge() { test "$(echo "$@" | tr " " "\n" | sort -V | tail -n 1)"
 
 if [ -z $1 ]
 then
-  VER='0.6.0'
```
Kami: I always thought this was replaced somewhere in the build pipeline, but I'm probably wrong and someone just manually updated scripts on ops.stackstorm.net :P

tools/st2_deploy.sh
```diff
@@ -374,6 +395,16 @@ install_webui() {
   mkdir -p /opt/stackstorm/static/webui
   cp -R ${temp_dir}/* /opt/stackstorm/static/webui
 
+  # Replace config.js
+  echo -e "'use strict';
```
Kami: That's fine for now, but if needed or wanted, we can also move this to the build step.

Kami:
Just pushed infinite loop redirect fix if ui files don't exist here - f63e08f49a2ed09694d02347e90e66c99705d1e1

Kami:
LGTM, +1

PR #1139, opened 2015-02-27
Fix action chain runner not including result for tasks / actions which failed during param rendering phase

Kami:
This pull request fixes a bug in the action chain runner - if the action failed during parameter rendering phase (before we actually scheduled and ran it), `LiveActionDB` and `ActionExecutionDB` object were never created and the result object didn't contain item in the `tasks` attribute for the failed action.


st2actions/st2actions/runners/actionchainrunner.py
```diff
@@ -282,7 +313,7 @@ def _format_action_exec_result(self, action_node, liveaction_db, created_at, upd
         result['updated_at'] = isotime.format(dt=updated_at)
 
         if error or not liveaction_db:
-            result['state'] = LIVEACTION_STATUS_SUCCEEDED
+            result['state'] = LIVEACTION_STATUS_FAILED
```
Kami: lulz, oops.

manasdk:
Parameter rendering = poor input by user. It is a failure at the highest level and not at any specific execution. Heck infact an execution was not even created.

I am -1 to this change.

st2actions/st2actions/runners/actionchainrunner.py
```diff
@@ -152,23 +154,51 @@ def run(self, action_parameters):
...
                     action_node=action_node, parent_execution_id=self.liveaction_id,
                     params=resolved_params)
             except Exception:
-                # Save the traceback and error message.
+                # Save the traceback and error message
                 LOG.exception('Failure in running action %s.', action_node.name)
                 error = {'error': traceback.format_exc(10)}
                 context_result[action_node.name] = error
+
+                if resolved_params is None:
```
manasdk: I disagree with this approach. Creating a LiveActionDB in a runner is an abstraction leak.

manasdk:
I think the error should be on the ActionChain main execution and the top level workflow should fail.

st2actions/st2actions/runners/actionchainrunner.py
manasdk: Yeah just read the original code and was very confused ... lol

Kami:
Discussed this offline with @manasdk - we decided we should go with a different approach where we treat parameter rendering as a top level action chain failure.

Now if the parameter rendering fails, we abort and fail the whole chain and propagate the error to the user in the top-level `error` result attribute.

PR #1142, opened 2015-02-27
Add documentation about error reporting.

manasdk:


docs/source/actionchain.rst
```diff
@@ -87,3 +87,64 @@ Note:
...
+::
+
+   "result": {
+        "tasks": [
+            {
+                "created_at": "2015-02-27T19:29:02.057885+00:00",
+                "execution_id": "54f0c57e0640fd177f278052",
+                "id": "c1",
+                "name": "c1",
+                "result": {
```
Kami: There could also be a case when `result` attribute contains an `error` key (e.g. running an action for some reason fails).

PR #1144, opened 2015-02-27
RFR: Allow rendering of non-string parameters in rule

lakshmi-kannan:
Refactored render_values used in action_chain to st2common to avoid code dupes. Improved tests.

### DIFF
--- a/st2actions/st2actions/runners/actionchainrunner.py
+++ b/st2actions/st2actions/runners/actionchainrunner.py
@@ -14,9 +14,6 @@
 # limitations under the License.
 
 import eventlet
-import jinja2
-import json
-import six
 import traceback
 import uuid
 import datetime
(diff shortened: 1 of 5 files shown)


st2common/st2common/util/jinja.py
```diff
@@ -0,0 +1,43 @@
+import json
```
Kami: Missing license header.

st2actions/st2actions/runners/actionchainrunner.py
```diff
@@ -35,35 +32,13 @@
...
 from st2common.util import action_db as action_db_util
 from st2common.util import isotime
+from st2common.util import jinja as jinja_utils
 
 
 LOG = logging.getLogger(__name__)
 RESULTS_KEY = '__results'
 
 
-def render_values(values, context):
```
Kami: Yay, much better for this to live in the utils.

st2common/st2common/util/jinja.py
lakshmi-kannan: Oopsy dawg!

lakshmi-kannan: Also missed changelog :/

PR #1148, opened 2015-02-27
Linux pack improvements part I 

lakshmi-kannan:


contrib/linux/actions/dig.py
```diff
@@ -1,25 +1,35 @@
...
+import subprocess
+import random
+import re
 from st2actions.runners.pythonrunner import Action
 
+
 class DigAction(Action):
 
-    def run(self,rand,count,nameserver,hostname,queryopts):
+    def run(self, rand, count, nameserver, hostname, queryopts):
```
Kami: Should also make arguments which are now optional default to `None`.

contrib/linux/actions/traceroute.yaml
```diff
@@ -19,3 +19,5 @@ parameters:
     description:  No. of quereies to each hop
     default: 3
     position: 3
+  hosts:
```
Kami: Wait, there are two arguments now - host and hosts?

And other arguments are positional and this one is not - don' think this will work well :P

PR #1150, opened 2015-02-27
Update docs for pack install

manasdk:


docs/source/packs.rst
```diff
@@ -70,8 +70,8 @@ Let's install the Docker pack:
     # Check out README.md and if necessary, adjust configuration for your environment
     less /opt/stackstorm/packs/docker/README.md
 
-    # Reloads the content
-    st2 run packs.load
+    # Reload all the content
+    st2 run packs.load register=all
```
Kami: In 99% cases you want to skip loading of the rules. I know `packs.load` doesn't support that use case now, but we should make it possible to only load sensors and actions.

dzimine: I am adding some more comments to clarify, PR is coming.

PR #1152, opened 2015-02-28
Fix log config

manasdk:
* general cleanup of history from configuration and logging


Kami:
Nice. The less code the better :+1:

PR #1153, opened 2015-02-28
Update sensor docs and examples

m4dcoder:


contrib/hello-st2/sensors/sensor1.py
```diff
@@ -0,0 +1,33 @@
+import eventlet
+
+from st2reactor.sensor.base import Sensor
+
+
+class HelloSensor(Sensor):
```
manasdk: Could use a PollingSensor also.

dzimine:
How are we going to use it? 
Just pointing to st2 is suboptimal: to get the example, I need to clone the whole repo. 
Shall we deploy it under /usr/share/doc/st2, along with examples? 
Or even place it into examples (I am -0.5 on it as the examples now are like a whole pack themselves, convinient to just copy and load the whole set

PR #1156, opened 2015-02-28
Linux pack improvements part deux

lakshmi-kannan:


contrib/linux/actions/netstat_grep.yaml
```diff
@@ -11,4 +11,4 @@
       position: 0
     pids:
       description: "List of pids"
```
Kami: Would be nice to say comma delimited list of pids

PR #1158, opened 2015-02-28
st2 key create user stanley -> st2 key set user stanley

kirilstrax:
The command:
st2 key create user stanley
does not work; the command
st2 key set user stanley
does what the description describes (creates ``user=stanley`` key-value pair)


dzimine:
+1
Thanks!

PR #1162, opened 2015-02-28
Log path fix

kirilstrax:
/var/logs/st2/logs/st2actionrunner.*.log -> /var/log/st2/logs/st2actionrunner.*.log


Kami:
Thanks, LGTM.

PR #940, opened 2015-01-05
Allow content pack resources to be disabled via the CLI commands

Kami:
This pull request adds CLI command for disabling and enabling CLI resources (sensors, actions, rules).

Currently it has a known limitation (same as the changes done with via UI) - those changes will be overwritten with the values from disk next time st2-register-content is ran. This will be fixed once the content distribution and syncing is implemented.

PR #1165, opened 2015-03-01
Testing the page - changes

kirilstrax:
1. Format and text - minor changes
2. Incorrect example rule path fixed
3. Since the same rule is used by Quick Start, user may get an error creating it. Added a small piece that handles this scenario
4. Moved commands that list rules into separate section (separate from creation)
5. Ordered operators by type (more or less)
6. In "long description" for a few operators was missing
(Side note: why do we need to describe operators twice? Isn't summary enough? Or is it to have static link to them?)
7. Removed reference to Getting Started (we are repeating the steps here + likely someone gets to this page after getting started
8. Rule creation removed from Rule Location section and moved rule location to be under Managing Rules
9. Moved what's next to the bottom

### DIFF
--- a/docs/source/rules.rst
+++ b/docs/source/rules.rst
@@ -79,12 +79,16 @@ If the rule with the same name already exists, the above command will return an 
     ERROR: 409 Client Error: Conflict
     MESSAGE: Tried to save duplicate unique keys (E11000 duplicate key error index: st2.rule_d_b.$name_1  dup key: { : "examples.webhook_file" })
         
-Existing rule can be updated as follows:
+To update the rule, edit the rule definition file and run ``st2 rule update`` command, as in the following example:
     
 .. code-block:: bash
     
     st2 rule update examples.webhook_file /usr/share/doc/st2/examples/rules/sample_rule_with_webhook.yaml
-    
+
+.. note::
+
+    **Hint:** It is a good practice to always edit the original rule file, so that keep your infrastructure in code. You still can get the rule definition from the system by ``st2 rule get <rule name> -j``, update it, and load it back.
+
 To see all the rules, or to get an individual rule, use commands below:
 
 .. code-block:: bash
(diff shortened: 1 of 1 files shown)


docs/source/rules.rst
```diff
@@ -62,16 +63,39 @@ contain such values e.g.
...
+
+If the rule with the same name already exists, the above command will return an error:
 
-To deploy a rule, use CLI:
+.. code-block:: bash
+    
+    ERROR: 409 Client Error: Conflict
+    MESSAGE: Tried to save duplicate unique keys (E11000 duplicate key error index: st2.rule_d_b.$name_1  dup key: { : "examples.webhook_file" })
+        
+Existing rule can be updated as follows:
```
dzimine: Say something like 
> "To update the rule, edit the rule definition file and run `st2 rule update`:"

Also, add something like this: 
>
Hint: It is a good practice to always edit the original rule file, so that keep your infrastructure in code. You still can get the rule definition from the system by `st2 rule get <rule name> -j `, update it, and load it back. 
>

docs/source/rules.rst
```diff
@@ -62,18 +63,52 @@ contain such values e.g.
...
+    st2 rule create /usr/share/doc/st2/examples/rules/sample_rule_with_webhook.yaml
+
+If the rule with the same name already exists, the above command will return an error:
+
+.. code-block:: bash
+    
+    ERROR: 409 Client Error: Conflict
+    MESSAGE: Tried to save duplicate unique keys (E11000 duplicate key error index: st2.rule_d_b.$name_1  dup key: { : "examples.webhook_file" })
+        
+To update the rule, edit the rule definition file and run ``st2 rule update`` command, as in the following example:
```
Kami: I believe when doing update via file, file also needs to contain object id (please correct me if I'm wrong).

If that's indeed the case, we should point this out here.

docs/source/rules.rst
```diff
@@ -62,18 +63,52 @@ contain such values e.g.
...
+To see all the rules, or to get an individual rule, use commands below:
 
 .. code-block:: bash
 
-    st2 rule create /opt/stackstorm/packs/examples/rules/sample_rule_with_webhook.yaml
-    st2 rule create /opt/stackstorm/examples/rules/sample_rule_with_webhook.yaml
     st2 rule list
     st2 rule get examples.webhook_file
 
+To undeploy a rule, run ``st2 rule delete ${RULE_NAME_OR_ID}``. For example, to undeploy the examples.webhook_file rule we deployed previously, run:
```
Kami: Just a reminder for myself that I need to get #940 merged so user can also do `st2 rule disable <id>`.

Kami:
Besides the comment about definition file needing to include id on update, LGTM.

PR #1166, opened 2015-03-01
Fixes for ActionChain docs and examples

dzimine:
- [x] Fix broken examples, update docs, add "action registration" instructions.
- [x] Move workflows under `actions/workflows` 
- [ ] Describe and add example for `vars` and `publish`.

### DIFF
--- a/docs/source/actionchain.rst
+++ b/docs/source/actionchain.rst
@@ -12,8 +12,7 @@ ActionChain's are described in YAML (JSON supported for backward compatibiltiy) 
 ActionChain definition
 ~~~~~~~~~~~~~~~~~~~~~~
 
-Following is sample ActionChain workflow definition named :github_st2:`echochain.yaml
-<contrib/examples/actions/chains/echochain.yaml>`:
+Following is sample ActionChain workflow definition named :github_st2:`echochain.yaml <contrib/examples/actions/chains/echochain.yaml>`:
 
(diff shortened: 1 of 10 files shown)


contrib/examples/actions/workflows/echochain_param.yaml
```diff
@@ -11,14 +11,14 @@
             name: "c2"
             ref: "core.local"
             params:
-                cmd: "echo c2 {{c1.localhost.stdout}}" # refers upsteram action parameter
```
Kami: Yep, localhost key was true for pre v0.7.

docs/source/actionchain.rst
```diff
@@ -9,39 +9,55 @@ Authoring an ActionChain
...
-Note:
+Details:
 
 * `chain` is the array property that contains action elements.
 * Action elements are named action execution specifications. The name is scoped to an ActionChain and is used as a reference to an action element.
 * `ref` property of an action element points to an Action registered in |st2|.
 * `on-success` is the link to action element to invoke next on a successful execution. If not provided the ActionChain will terminate with status set to success.
 * `on-failure` is the link to action element to invoke next on a failed execution. If not provided the ActionChain will terminate with the status set to error.
-* `default` is the top level property that specifies start of an ActionChain.
+* `default` is an optional top level property that specifies the start of an ActionChain.
```
Kami: Should probably make it clear and say something along the lines of "if default is not specified, first action in the chain is considered as default and executed first".

PR #1168, opened 2015-03-01
Fix CORS for st2auth and st2api

enykeev:


Kami:
LGTM, +1

st2auth/st2auth/config.py
```diff
@@ -59,6 +59,8 @@ def _register_app_opts():
     api_opts = [
         cfg.ListOpt('allow_origin', default=['http://localhost:3000'],
             help='List of origins allowed'),
+        cfg.BoolOpt('serve_webui_files', default=True,
```
Kami: Would if we should actually move it to st2common.config since we already have some common API options there...

enykeev: Lets do it in another PR. It can wait.

PR #1173, opened 2015-03-01
Add verbose flag to register content in st2ctl

m4dcoder:
By default, register content will only output logs that has level ERROR or above. If "-v" or "--verbose" is included, the log level will change to DEBUG.

```
Examples:
st2ctl reload -v
st2ctl reload --verbose
st2ctl clean -v
st2ctl clean --verbose
```

### DIFF
--- a/tools/st2ctl
+++ b/tools/st2ctl
@@ -122,12 +122,18 @@ function register_content() {
   echo "Registering content..."
 
   if [ ! ${1} ]; then
-      REGISTER_FLAGS="--register-sensors --register-actions"
+    REGISTER_FLAGS="--register-sensors --register-actions"
+  elif [ ${1} == "-v" -o ${1} == "--verbose" ]; then
+    if [ ! ${2} ]; then
+      REGISTER_FLAGS="--register-sensors --register-actions --verbose"
+    else
+      REGISTER_FLAGS="${2} --verbose"
+    fi
   else
-      # Note: Scripts already call reload with "--register-<content>"
-      # TODO: Update packs. actions to only pass in a resource name excluding
-      # --register prefix
-      REGISTER_FLAGS="${1}"
+    # Note: Scripts already call reload with "--register-<content>"
+    # TODO: Update packs. actions to only pass in a resource name excluding
+    # --register prefix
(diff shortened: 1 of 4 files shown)


manasdk:
+1

tools/st2ctl
```diff
@@ -122,12 +122,18 @@ function register_content() {
   echo "Registering content..."
 
   if [ ! ${1} ]; then
```
dzimine: we need to fortify this. Discussed with @manask. 

https://stackstorm.atlassian.net/browse/STORM-1104
As you're with this, you can take it over: 

agreed to fortify by only allowing to pass through the following keys, and if not, print out the list of the keys:
--register-sensors, register-actions, --register-rules, --register-all.  

So please add code to check if the ${1} in the list above, and if not, print the "help" message. 

Even if it's a throw-away, the feature is central to content management, and mistyping the keys brings hell of confusion.

tools/st2ctl
```diff
@@ -122,12 +122,18 @@ function register_content() {
...
-      REGISTER_FLAGS="--register-sensors --register-actions"
+    REGISTER_FLAGS="--register-sensors --register-actions"
+  elif [ ${1} == "-v" -o ${1} == "--verbose" ]; then
+    if [ ! ${2} ]; then
+      REGISTER_FLAGS="--register-sensors --register-actions --verbose"
```
dzimine: I think default should NOT be verbose, it's exactly the problem we are trying to solve?

dzimine:
Also, can you please validate that it returns non-zero code if it fails to register one or more elements of the content? If it's easy, just do it, if not, file a bug. Thanks!

tools/st2ctl
manasdk: Yeah this makes sense. Also, the --help passing into register-content was just weird.

tools/st2ctl
m4dcoder: Verbose is not default.  See line 126.

tools/st2ctl
m4dcoder: https://github.com/StackStorm/st2/pull/1175

PR #1174, opened 2015-03-02
Action chain fixes

manasdk:
* Report failure in looking up next nodes
* Preserve order in default node lookup

### DIFF
--- a/st2actions/st2actions/runners/actionchainrunner.py
+++ b/st2actions/st2actions/runners/actionchainrunner.py
@@ -66,13 +66,18 @@ class ChainHolder(object):
         # 1. There are no loops in the chain. Even if there are loops there is
         #    at least 1 node which does not end up in this loop.
         # 2. There are no fragments in the chain.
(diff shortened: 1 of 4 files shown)


st2actions/st2actions/runners/actionchainrunner.py
```diff
@@ -83,10 +88,15 @@ def _get_rendered_vars(vars):
...
 
-    def get_node(self, node_name=None):
+    def get_node(self, node_name=None, raise_on_failure=False):
+        if not node_name:
+            return None
         for node in self.actionchain.chain:
             if node.name == node_name:
                 return node
+        if raise_on_failure:
+            raise runnerexceptions.ActionRunnerException('Unable to find node with name %s.',
```
Kami: That's something we could also do in the pre-processing phase (before running the chain), right?

Kami: I guess we only can't do it if user uses a jinja expression for on-success / on-failure which is evaluated during run-time.

In any case, nice change.

PR #1175, opened 2015-03-02
Add check args and print usage to st2ctl

m4dcoder:
Add basic check args to st2ctl. If check args fail, st2ctl will print usage information.

### DIFF
--- a/tools/st2ctl
+++ b/tools/st2ctl
@@ -1,12 +1,12 @@
 #!/bin/bash
+
+set -e
+
 LOGFILE="/tmp/st2_startup.log"
 COMPONENTS="actionrunner st2api sensor_container rules_engine mistral st2resultstracker"
 STANCONF="/etc/st2/st2.conf"
 PYTHON=`which python`
-if [ -z "$AR" ];
- then
-   AR=10
-fi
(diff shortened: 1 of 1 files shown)


tools/st2ctl
```diff
@@ -1,12 +1,12 @@
 #!/bin/bash
+
+set -e
+
 LOGFILE="/tmp/st2_startup.log"
 COMPONENTS="actionrunner st2api sensor_container rules_engine mistral st2resultstracker"
 STANCONF="/etc/st2/st2.conf"
 PYTHON=`which python`
-if [ -z "$AR" ];
```
DoriftoShoes: Did we add in the ability to set the number of action runners somewhere else?  The current configuration allows us to use envvars to do it.

m4dcoder: No.  I was not clear from this script what that block was doing because $AR didn't seem to come from anywhere.

dzimine:
LGTM, solves the problem (assuming it works :) :+1:

m4dcoder:
Addressed @DoriftoShoes comment.  Yeah, I hand tested this on st2express.

PR #1176, opened 2015-03-02
Avoid showing filters for workflow records

enykeev:
STORM-1115 #resolved


tools/launchdev.sh
```diff
@@ -1,8 +1,5 @@
 #!/bin/bash
 
-echo $AR
```
Kami: arr, arr!

PR #1178, opened 2015-03-02
More specific grep to avoid uninstalling non StackStorm packages

DoriftoShoes:
Fedora 21 installation was failing on the uninstall step for the packages.  This makes the grep more specific to help avoid this situation.


Kami:
LGTM, +1

PR #1182, opened 2015-03-02
Fix action chain examples

m4dcoder:
Separate the meta from the action chain definition files. Fix variable references in the action chain.


manasdk:
I like the .meta sub(?)fix. It help identify that this is meta and not the chain.

It is cool though I don't mind either way.

manasdk:
+1

m4dcoder:
Sure.  I can add the meta back in.

PR #1185, opened 2015-03-02
Don't use colors in the CLI when running in the terminal which doesn't support it

Kami:


st2client/st2client/commands/action.py
```diff
@@ -665,7 +665,7 @@ def __init__(self, resource, *args, **kwargs):
         task_list_arg_grp.add_argument('--depth', type=int, default=-1,
                                        help='Depth to which to show sub-tasks. \
                                              By default all are shown.')
-        task_list_arg_grp.add_argument('-w', '--width', nargs='+', type=int, default=[28],
+        task_list_arg_grp.add_argument('-w', '--width', nargs='+', type=int, default=None,
```
Kami: We want to default to the terminal window size.

PR #1187, opened 2015-03-02
WIP: Create virtualenv for linux pack in deb postinst

lakshmi-kannan:


Kami:
We need to do the same for RPM packages, right?

Sounds like we could split this common functionality into a bash file with shared functions and use this function in both places (not actually sure if you can source bash files in postinst and rpm spec files though).

st2common/packaging/debian/postinst
```diff
@@ -6,3 +6,31 @@ if [ -d /etc/st2 ]; then
...
 chmod 755 /usr/bin/st2ctl
+
+PYTHON_BIN=`grep "python_binary" /etc/st2/st2.conf | head -n 1 | awk '{print $3}'`
+if [ -z "$PYTHON_BIN" ]
+then
+    PYTHON_BIN=`which python`
+fi
+
+PIP=`which pip`
+if [ -z $PIP ]
```
Kami: Not consistent quoting (in the upper if you used double quotes but not here) :P

lakshmi-kannan:
@Kami we discussed all this on slack. I have the same doubts on sourcing. @DoriftoShoes offered to look.

DoriftoShoes:
I dislike this.  I feel like we are shipping packs with the system but doing all sorts of hackery within the installer try and treat them like they are not external packs.  We probably need to include this in order to get this release out but I think we should look towards shipping the system bare (no packs at all) with a basic command to bootstrap some packs from an external repo.  That way we can include the command in st2_deploy or just point users to it in the docs.  Spinning the venv from the package install feels like it will break.

lakshmi-kannan:
@DoriftoShoes  This PR needs to die in a fire. I wanted to open it during last release cos we decided to ship linux pack with st2 in a ghetto way. I'll close this.

DoriftoShoes:
+1 to dying in a fire.

jfryman:
:fire:

PR #1188, opened 2015-03-02
Report failure when the path in actionchain is broken

manasdk:


Kami:
LGTM

st2actions/tests/unit/test_actionchain.py
```diff
@@ -207,7 +207,8 @@ def test_chain_runner_broken_success_path(self, schedule):
 
     @mock.patch.object(action_db_util, 'get_action_by_ref',
                        mock.MagicMock(return_value=ACTION_1))
-    @mock.patch.object(action_service, 'schedule', side_effect=RuntimeError('Test Failure.'))
+    @mock.patch.object(action_service, 'schedule',
+                       return_value=(DummyActionExecution(status=LIVEACTION_STATUS_FAILED), None))
```
Kami: Would be also good to add assert on the result here to make sure it contains the correct `error` and `traceback` attribute.

Kami: #1211

PR #1193, opened 2015-03-03
[RFR] Avoid unnecessary timestamp

manasdk:
* By assigning datetime.now in case of None we end up assigning a value to end_timestamp


Kami:
I just copied the behavior over from the original mongoengine field type, but this change looks good and correct to me.

Kami:
Would also be good to add a test case for it though.

Kami:
Added tests and merged this into 0.8.1 and master.

PR #1208, opened 2015-03-03
Add the single quote

DoriftoShoes:


docs/source/conf.py
```diff
@@ -228,7 +228,7 @@
     'source_suffix': source_suffix,
     'versions': [
         ('latest', 'http://docs.stackstorm.com/latest'),
-        (version, 'http://docs.stackstorm.com/),
+        (version, 'http://docs.stackstorm.com/'),
```
dzimine: It is on branch 0.8.0. Version number is 0.8. So you get two 0.8 versions. Get rid of explicit 0.8 and you'll be fine.

PR #1209, opened 2015-03-03
Fix docs versions, forever.

dzimine:
See #1210 against master for explanations.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -14,6 +14,7 @@
 
 import sys
 import os
+import itertools
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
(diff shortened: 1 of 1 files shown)


docs/source/conf.py
```diff
@@ -219,6 +220,16 @@
 # Output file base name for HTML help builder.
 htmlhelp_basename = 'StackStormDoc'
 
+
+def previous_version(ver):
+    # XXX: on incrementing major version, minor version counter is lost!
+    major, minor = ver.split('.')
```
DoriftoShoes: Does this also need `ver.split(',',2)` to drop the patch level?

dzimine: It's dropped by next line. Else in case of 0.999dev we don't have a second separator. Or how do yuo mean?

Kami: Unpacking the values into `major` and `minor` will fail though if `ver` contains 2 dots (split will return 3 values if you don't do `ver.split(',', 1)).

PR #1210, opened 2015-03-03
Fix doc version on master.

dzimine:
Let's cross the t's on how we deploy and reference docs. 

## What we need:
* one docs/source/conf.py [1]
* Two build jobs.

## How it looks:
Picture worth 1000 words. 
image

## On Release:
* <stable_version> branch cut, __versioin__ updated to new <stable_version>, e.g. from 0.9dev to 0.9.
* On master, __version__ incremented to next dev version, e.g. from 0.9dev to 0.10dev.
* Create a new s3./<stable_version> bucket and 's3/<next_version>.dev buckets
* Cut a new <stable_version> branch, switch stable build to it.
* "Build Stable" keeps building s3/root and s3/<stable_version>. s3/root now contains new stable version.
* Increment version on master to <next_version>.dev, switch master build to it. 
* "Build Master" keeps building s3/latest and s3/<next_version>.dev s3/latest now contains new latest version.
* All previous docs are intact, and referencing each other properly. User can navigate to anything follow 'latest' and than down the version number. 

[1] The docs/source/conf.py will need to be manually updated on major version increment as there's no way to guess a minor number at the time.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -14,6 +14,7 @@
 
 import sys
 import os
+import itertools
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
@@ -219,6 +220,16 @@ html_static_path = ['_static']
 # Output file base name for HTML help builder.
 htmlhelp_basename = 'StackStormDoc'
 
+
+def previous_version(ver):
+    # XXX: on incrementing major version, minor version counter is lost!
+    major, minor = ver.split('.')
+    minor = int("".join(itertools.takewhile(str.isdigit, minor)))
+    return ".".join([major, str(minor - 1)])
+
+version_minus_1 = previous_version(version)
+version_minus_2 = previous_version(version_minus_1)
+
 # Variables to be used by templates
 html_context = {
     'github_repo': 'StackStorm/st2',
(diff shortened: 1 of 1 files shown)


dzimine:
Exactly same conf.py to go to 0.8.

dzimine:
same against v0.8 https://github.com/StackStorm/st2/pull/1209

docs/source/conf.py
```diff
@@ -219,6 +220,16 @@
 # Output file base name for HTML help builder.
 htmlhelp_basename = 'StackStormDoc'
 
+
+def previous_version(ver):
+    # XXX: on incrementing major version, minor version counter is lost!
+    major, minor = ver.split('.')
```
Kami: This will only work if we stick with major.minor approach (not major.minor.patch), right?

DoriftoShoes: True, we need to drop the patch level.  
`ver.split(',',2)`  would work I assume.

dzimine: it works for both `0.11.22` and `0.11dev`

enykeev: I'm sorry to be importunate, but that kind of problems were solved long ago... https://github.com/rbarrois/python-semanticversion

PR #1211, opened 2015-03-03
Action chain test improvements

Kami:
* Fix exception error message
* Add some additional asserts which make sure we are testing the right case and that the tests return expected error message


manasdk:
+1

Thanks!

PR #1019, opened 2015-01-27
[RFR/No merge]Include the raw result of stdout and stderr in stdout_raw and stderr_raw.

manasdk:
* output of `core.local cmd="date"`
```
 st2 execution get 54c72f6b0640fd24cfc343f1 -d -j
{
    "action": "core.local",
    "callback": {},
    "context": {
        "user": "stanley"
    },
    "end_timestamp": "2015-01-27T06:25:47.670000Z",
    "id": "54c72f6b0640fd24cfc343f1",
    "parameters": {
        "cmd": "date"
    },
    "result": {
        "failed": false,
        "return_code": 0,
        "stderr": "",
        "stderr_raw": "",
        "stdout": "Tue Jan 27 06:25:47 UTC 2015\n",
        "stdout_raw": "Tue Jan 27 06:25:47 UTC 2015\n",
        "succeeded": true
    },
    "start_timestamp": "2015-01-27T06:25:47.619000Z",
    "status": "succeeded"
}
```

* output of `core.local cmd="echo 1"`
```
$ st2 execution get 54c7301d0640fd24cfc343f3 -dj
{
    "action": "core.local",
    "callback": {},
    "context": {
        "user": "stanley"
    },
    "end_timestamp": "2015-01-27T06:28:45.374000Z",
    "id": "54c7301d0640fd24cfc343f3",
    "parameters": {
        "cmd": "echo 1"
    },
    "result": {
        "failed": false,
        "return_code": 0,
        "stderr": "",
        "stderr_raw": "",
        "stdout": 1,
        "stdout_raw": "1\n",
        "succeeded": true
    },
    "start_timestamp": "2015-01-27T06:28:45.320000Z",
    "status": "succeeded"
}
```

PR #1212, opened 2015-03-03
Allow user to exclude result from /actionexecutions, new endpoints for retrieving execution result, stdout and stderr

Kami:
## Changes

This pull request includes the following changes.

* Allow user to retrieve executions without the `result` attribute by passing `?exclude_result=1` query parameter to the `/actionexecutions/` endpoint.
* New endpoints:
  * `GET /actionexecutions/<id>/result` - retrieve execution result object
  * `GET /actionexecutions/<id>/stdout` - retrieve raw value of the execution result `stdout` attribute
  * `GET /actionexecutions/<id>/stderr` - retrieve raw value of the execution result `stderr` attribute

Other changes:

* Technical debt cleanup 
  * Removed awful pecan controller hackery and replaced it with nested controllers.
  * Tried to remove as much as the kwargs abuse and magical code as I could.

## Example responses

TBW

## Reasoning / Context

Currently, if an action generates a lot of output (`stdout`, `stderr`), retrieving all the actions executions will be very expensive. On top of that, JSON parsing of the response might fail due to the memory exhaustion if the object is too big (this happened in the ui when running in the browser).

This pull request solves this problem by allowing user to only retrieve action executions without the `result` attribute. In addition to that, it adds new endpoints for retrieving result, stdout and stderr attributes. This way, we can, for example inside the UI load the actions executions first and only request stdout and stderr when user hovers / clicks on the action execution (we can also pre-fetch some of that, etc., but that's a ui performance optimization).

This also goes along well with "result attribute needs to contain a result in a structured format" and the value of "stdout" and "stderr" is just an opaque (binary) data to us. I know this is currently not 100% true, but imo, we need to change and fix this in the near future (see https://github.com/StackStorm/st2/pull/1019#issuecomment-71622276).

This implementation is also very similar to the one I proposed for streaming (https://github.com/StackStorm/discussions/issues/27) and in the future, we can build streaming on top of those new endpoints.

Another thing related to the "stdout" and "stderr" being just an opaque binary data to us - currently we more or less use MongoDB for storing different meta data in it. This means storing potentially large "binary" / log structured data in it, in addition to the existing metadata is not the best fit (different access patterns, we don't need to query based on the value of this binary data, etc.)

In the future, we should (and probably will need to because of the performance reasons) look into storing this "binary" / log structured data in a different way / data store (e.g. even storing it in flat files might be better in some cases, but something like Cassandra is of course better than that).

### DIFF
--- a/st2common/st2common/models/db/__init__.py
+++ b/st2common/st2common/models/db/__init__.py
@@ -142,7 +142,6 @@ class MongoDBAccess(object):
             dt1 = isotime.parse(values[0])
             dt2 = isotime.parse(values[1])
 
-
             k__gte = '%s__gte' % k
             k__lte = '%s__lte' % k
             if dt1 < dt2:
--- a/st2common/st2common/models/db/__init__.py
+++ b/st2common/st2common/models/db/__init__.py
@@ -126,11 +126,22 @@ class MongoDBAccess(object):
     @process_null_filter
     @process_datetime_ranges
     def query(self, *args, **kwargs):
+        # TODO: Fix kwargs abuse
         offset = int(kwargs.pop('offset', 0))
         limit = kwargs.pop('limit', None)
-        eop = offset + int(limit) if limit else None
         order_by = kwargs.pop('order_by', [])
-        return self.model.objects(**kwargs).order_by(*order_by)[offset:eop]
+        exclude_fields = kwargs.pop('exclude_fields', [])
+        eop = offset + int(limit) if limit else None
+
+        result = self.model.objects(**kwargs)
+
+        if exclude_fields:
+            result = result.exclude(*exclude_fields)
+
+        result = result.order_by(*order_by)
+        result = result[offset:eop]
+
+        return result
 
     def distinct(self, *args, **kwargs):
         field = kwargs.pop('field')
--- a/st2common/st2common/models/db/__init__.py
+++ b/st2common/st2common/models/db/__init__.py
@@ -57,46 +57,6 @@ def db_teardown():
     mongoengine.connection.disconnect()
 
 
-def process_null_filter(func):
-    def decorate(*args, **kwargs):
-        filters = {k: v for k, v in kwargs.iteritems()
-                   if v is None or (type(v) in [str, unicode] and str(v.lower()) == 'null')}
-        for k, v in filters.iteritems():
-            kwargs['%s__exists' % k] = False
-            del kwargs[k]
-        return func(*args, **kwargs)
-    return decorate
-
-
-def process_datetime_ranges(func):
-    def decorate(*args, **kwargs):
-        ranges = {k: v for k, v in kwargs.iteritems()
-                  if type(v) in [str, unicode] and '..' in v}
-        for k, v in ranges.iteritems():
-            values = v.split('..')
-            dt1 = isotime.parse(values[0])
-            dt2 = isotime.parse(values[1])
-            order_by_list = kwargs.get('order_by', [])
-            k__gte = '%s__gte' % k
-            k__lte = '%s__lte' % k
-            if dt1 < dt2:
-                query = {k__gte: dt1, k__lte: dt2}
-                sort_key, reverse_sort_key = k, '-' + k
-            else:
-                query = {k__gte: dt2, k__lte: dt1}
-                sort_key, reverse_sort_key = '-' + k, k
-            del kwargs[k]
-            kwargs.update(query)
-            if reverse_sort_key in order_by_list:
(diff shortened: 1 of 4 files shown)


st2common/st2common/models/db/__init__.py
```diff
@@ -123,14 +83,24 @@ def get_all(self, *args, **kwargs):
     def count(self, *args, **kwargs):
         return self.model.objects(**kwargs).count()
 
-    @process_null_filter
-    @process_datetime_ranges
-    def query(self, *args, **kwargs):
-        offset = int(kwargs.pop('offset', 0))
-        limit = kwargs.pop('limit', None)
+    def query(self, offset=0, limit=None, order_by=None, exclude_fields=None,
```
Kami: This kwargs abuse and decorator magic was really awful. It's really hard to debug and follow what's going on. We should avoid kwargs abuse and such magic - remember, we are writing code for people and not for computers.

enykeev:
Just want to add we also have a problem with `core.st2.generic.actiontrigger` exposing the whole execution including the results.

Kami:
@enykeev You are referring to the trigger_instance attribute on the execution object, right?

enykeev:
yeah, the initial execution gets exposed through `trigger_instance.payload`

Kami:
@enykeev What do you think about handling it in a similar manner?

E.g. `?exclude_trigger_payload=1` and `/actionexecution/<id>/trigger_payload`?

manasdk:
I dont agree with these endpoints -

GET /actionexecutions/<id>/stdout - retrieve raw value of the execution result stdout attribute
GET /actionexecutions/<id>/stderr - retrieve raw value of the execution result stderr attribute

stdout and stderr are properties specific to runners. Exposing them here feels broken. If the results endpoint was able to handle retrieval of a specific property then it makes sense.

Basically, we are looking to support projections of an object. Can it be done generically rather than these specific endpoints?

Kami:
@manasdk Yeah, I was thinking about that (being specific to a runner). In the streaming endpoint, I proposed an endpoint called "output" which would combine stdout and stderr, but we still want potentially retrieve it separately in this case.

Something like this could work:

```
GET /actionexecutions/<id>/result/<attribute name>
```

To be on the safe side, we should still limit the attributes to something like stdout, stderr, trigger.payload.

enykeev:
what are we going to do with runners like `run-remote` that don't have stdout in the root of the results instead having more complex structure?

Kami:
@enykeev That's a good point - one option is to have a per-runner specific processing, but I'm not a big fan of what.

Other option is to have per-runner specific endpoints, e.g. for remote runner, something like `GET /actionexecutions/<id>/result/<hostname>/<attribute name>/`. Not a huge fan of that either.

Kami:
Alright, so we had a pretty heated (:smiling_imp:)  debate about this on Slack.

We decided to go with the following approach / changes:

* `GET /actionexecutions/?exclude_attributes=result,trigger.instance` - allow user to exclude particular attributes from the response when listing all the executions
* `GET /actionexecutions/<id>?exclude_attributes=result,trigger_instance` - allow user to exclude particular attributes from the response when retrieving details for a single execution
* `GET /actionexecutions/<id>/attribute/<{result,trigger_instance}>` - allow user to retrieve a value of a particular attribute for a particular execution

Just first change should hopefully get us far, but second and third change are an additional "bonus".

I'm working on those changes as we speak. @enykeev I'll ping you when it's ready for you to hook it up in the UI.

Kami:
Also, to clarify it, when I get a green light, I will merge those changes into `master` and `0.8.1` branch.

Kami:
@enykeev Alright, changes are ready for you. Please let me know when you have tested the changes and hooked them up in the UI and I will get them merged into 0.8.1 and master.

enykeev:
@Kami LGTM. Changes in st2web are made, although it would be great if we could test them in an actual problematic environment before pushing 0.8.1. WDYT, is there any way for us to deploy them on master002 or its exact copy to make sure it actually solves the problem?

Feel free to merge StackStorm/st2web#108 as soon as this PR lands.

Kami:
@enykeev Thanks - I'll try to coordinate this with @DoriftoShoes.

In any case, merging into master is fine in both case (we should just wait with merging it into 0.8 branch :))

Kami:
I also checked out the st2web branch and tested it against this branch and it looks good.

In the future we could also look in exclude result and trigger instance being a default behavior - maybe we can do this change in 0.9 and in 0.8 we keep it backward compatible and as it is now.

Kami:
Merged changes into `0.8.1` and `master` branch.

PR #1217, opened 2015-03-04
Remove example sensu rule

Kami:
TODO:

- [x] Update docs (`docs/source/resources/monitoring.rst`)

As an alternative, we can modify packaging installation steps to not copy this rule over to /opt/stackstorm/packs


dzimine:
Thanks! 
Let's just update the dosc by writing down the rule into the .rst instead of linking to the file;
and point out to the rule samples in the sensu pack (just mention "rule samples available in the pack /opt/stackstorm/packs/sensu/rules (link github/st2contrib/sensu/packs/sensu/rules)

Kami:
K, sounds good.

PR #1218, opened 2015-03-04
Execution get improvements - automatically infer action type from the execution and format user-friendly output accordingly

Kami:
## Changes

1. We now automatically infer action type (workflow / non-workflow) when using `execution get` command. If execution refers to a workflow, `--tasks` option is implied and the output formatted accordingly. If user wants a raw (non formatted output), they can use the existing `-d` flag or a new `--raw` flag. For backward compatibility reasons, `--tasks` option wasn't removed and is still there.

2. Print an error if user uses `--tasks` option with a non-workflow action (tasks doesn't make sense there).

Example outputs - https://gist.github.com/Kami/6702083a1a05e0b47f19

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,8 @@ v0.8.1 - in development
   and ``/actionexecutions/<execution id>/`` endpoint (new-feature)
 * Add new ``/actionexecutions/<id>/attribute/<attribute name>`` endpoint which allows user to
   retrieve a value of a particular action execution attribute. (new-feature)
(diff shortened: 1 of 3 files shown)


dzimine:
LGTM.
Can you add the same to `st2 run my.workflow_action` defaults?

Kami:
@dzimine I can do it, but it requires some refactoring.

Maybe merge this first and work on run change in a separate branch? This way, run change doesn't block 0.8.1 if we want to get it out asap.

dzimine:
Sure!

PR #1219, opened 2015-03-05
Rename attributes in mistral task entry

m4dcoder:
Mistral renamed wf_name to workflow_name and execution_id to workflow_execution_id in the task object.

### DIFF
--- a/st2actions/st2actions/query/mistral/v2.py
+++ b/st2actions/st2actions/query/mistral/v2.py
@@ -129,8 +129,8 @@ class MistralResultsQuerier(Querier):
 
         result['id'] = task['id']
         result['name'] = task['name']
-        result['execution_id'] = task.get('execution_id', None)
-        result['workflow'] = task['wf_name']
+        result['workflow_execution_id'] = task.get('workflow_execution_id', None)
+        result['workflow_name'] = task['workflow_name']
         result['created_at'] = task.get('created_at', None)
         result['updated_at'] = task.get('updated_at', None)
         result['state'] = task.get('state', None)


st2actions/st2actions/query/mistral/v2.py
```diff
@@ -129,8 +129,8 @@ def _format_task_result(self, task):
 
         result['id'] = task['id']
         result['name'] = task['name']
-        result['execution_id'] = task.get('execution_id', None)
-        result['workflow'] = task['wf_name']
+        result['workflow_execution_id'] = task.get('workflow_execution_id', None)
```
Kami: For backward compatibility / consistency reasons, we could leave the attributes on our side named the same, right?

Kami: To be honest though, I like the new attribute names more since they are more explicit - so this time, we should probably change the names on our side as well (but in the future, if needed, we should consider the other option) :)

m4dcoder: LOL.  Yeah.  I was back and forth with this too.

m4dcoder: @Kami FYI, I'm going to have to push this to 0.8.1 as well.  There's a bug fix we need in mistral.  But this backward compatibility change superseded the bug fix in mistral.

Kami: @m4dcoder Sounds good - it sounds like we will also need to update v0.8.0 mistral branch, right (if this hasn't been done yet)?

PR #1221, opened 2015-03-05
Fix http runner json body parsing bug

Kami:
We didn't parse the body if the JSON content-type header also contained a charset.


Kami:
Closing this one - wanted to open it against 0.8.1 and not master. See https://github.com/StackStorm/st2/pull/1222

PR #1226, opened 2015-03-05
Fix mistral task wf exec

m4dcoder:


manasdk:
LGTM. Not a fan of the mistral branch selection stuff but I guess it is necessary.

m4dcoder:
Yep.  When@lakshmi-kannan revisit the mistral packaging story, hopefully this will be cleaned up as well.

Kami:
LGTM

PR #1230, opened 2015-03-06
URL corrections

kirilstrax:
1. Pip: https://ops.stackstorm.net/releases/st2/0.9dev/requirements.txt (404) -> https://raw.githubusercontent.com/StackStorm/st2/master/requirements.txt (found in st2_deploy.sh)

2. RabbitMQ: http://www.rabbitmq.com/install-rpm.html -> http://www.rabbitmq.com/install-debian.html

### DIFF
--- a/docs/source/install/deb.rst
+++ b/docs/source/install/deb.rst
@@ -23,7 +23,7 @@ The easiest way to install these is to use the requirements.txt file from the |s
 
 ::
 
-    curl -q -k -O https://ops.stackstorm.net/releases/st2/0.9dev/requirements.txt
+    curl -q -k -O https://raw.githubusercontent.com/StackStorm/st2/master/requirements.txt
     pip install -r requirements.txt
 
 RabbitMQ
@@ -33,7 +33,7 @@ In order to get the latest version of RabbitMQ, you will want to follow the dire
 
 ::
 
-    http://www.rabbitmq.com/install-rpm.html
+    http://www.rabbitmq.com/install-debian.html
 
 Once you have RabbitMQ installed, you will need to run the following commands to enable certain plugins.
 


docs/source/install/deb.rst
```diff
@@ -23,7 +23,7 @@ The easiest way to install these is to use the requirements.txt file from the |s
 
 ::
 
-    curl -q -k -O https://ops.stackstorm.net/releases/st2/0.9dev/requirements.txt
+    curl -q -k -O https://raw.githubusercontent.com/StackStorm/st2/master/requirements.txt
```
dzimine: This can't be right - it'll get requirements for the top of the tree, not for a given version. And don't we install from packages? Suggest Patrick or James review this whole section and fix it to match realtiy. We do have a bug on this.

docs/source/install/deb.rst
```diff
@@ -33,7 +33,7 @@ In order to get the latest version of RabbitMQ, you will want to follow the dire
 
 ::
 
-    http://www.rabbitmq.com/install-rpm.html
+    http://www.rabbitmq.com/install-debian.html
```
dzimine: right.

docs/source/install/deb.rst
Kami: And yeah, we really should link to our server (ops.) and not github.

lakshmi-kannan: I think now the URL is downloads.stackstorm.net.

DoriftoShoes: I think we should get the requirements.txt file from GitHub.  No real reason to be copying it around all over the place when we have branches for each release.
https://raw.githubusercontent.com/StackStorm/st2/v0.8/requirements.txt
https://raw.githubusercontent.com/StackStorm/st2/master/requirements.txt

Kami: This way we depend on the availability on Github, which I guess, is not that big of a deal since we already indirectly depend on that when installing Python dependencies (some if them are installed from a Github git repo).

Ideally, down the road we would move to self hosted wheels and in the short term at least our mirror of PyPi packages we use...

manasdk: > This way we depend on the availability on Github

As opposed to pypi or downloads.stackstorm.net?

Kami:
Backported the docs fix and closing this PR. Thanks.

PR #1234, opened 2015-03-08
[RFR] Update action chain runner - allow users to access action parameters inside "publish" scope

Kami:
Previously, user was unable to access action parameters inside the "publish" scope. This limitation made any things harder.

Here is an example of a work-flow where I use this functionality:

Metadata:

```yaml
 ...
  parameters:
    ip_address:
      type: "string"
      description: "IP address of a remote server to collect the informatiom from"
      required: true
```

Workflow:

```yaml
---
  chain:
    -
      name: "create_temporary_directory"
      ref: "core.remote_sudo"
      params:
        hosts: "{{ ip_address }}"
        cmd: "mktemp -d"
      publish:
        temp_dir: "{{ create_temporary_directory[parameters.ip_address].stdout }}"
      on-success: "retrieve_logged_users"

...
```

TODO:

- [x] Tests
- [x] Update changelog
- [x] Update docs

### DIFF
--- a/st2actions/st2actions/runners/actionchainrunner.py
+++ b/st2actions/st2actions/runners/actionchainrunner.py
@@ -136,17 +136,13 @@ class ActionChainRunner(ActionRunner):
         fail = True
         action_node = None
 
-        # Add action parameters to the context result so they can be accessed
-        # inside "publish"
-        context_result['parameters'] = action_parameters
-
         try:
             action_node = self.chain_holder.get_next_node()
         except Exception as e:
             LOG.exception('Failed to get starting node "%s".', action_node.name)
 
             error = ('Failed to get starting node "%s". Lookup failed: %s' %
-                    (action_node.name, str(e)))
+                     (action_node.name, str(e)))
             trace = traceback.format_exc(10)
             top_level_error = {
                 'error': error,
@@ -199,8 +195,9 @@ class ActionChainRunner(ActionRunner):
 
                 # Render and publish variables
                 rendered_publish_vars = ActionChainRunner._render_publish_vars(
-                    action_node=action_node, execution_result=liveaction.result,
-                    previous_execution_results=context_result, chain_vars=self.chain_holder.vars)
+                    action_node=action_node, action_parameters=action_parameters,
+                    execution_result=liveaction.result, previous_execution_results=context_result,
+                    chain_vars=self.chain_holder.vars)
 
                 if rendered_publish_vars:
                     self.chain_holder.vars.update(rendered_publish_vars)
@@ -252,8 +249,8 @@ class ActionChainRunner(ActionRunner):
         return (status, result, None)
 
     @staticmethod
-    def _render_publish_vars(action_node, execution_result, previous_execution_results,
-                             chain_vars):
+    def _render_publish_vars(action_node, action_parameters, execution_result,
+                             previous_execution_results, chain_vars):
         """
         If no output is specified on the action_node the output is the entire execution_result.
         If any output is specified then only those variables are published as output of an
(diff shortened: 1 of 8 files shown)


st2actions/st2actions/runners/actionchainrunner.py
```diff
@@ -130,13 +130,16 @@ def pre_run(self):
...
-
         result = {'tasks': []}  # holds final result we store
         context_result = {}  # holds result which is used for the template context purposes
         top_level_error = None  # stores a reference to a top level error
         fail = True
         action_node = None
 
+        # Add action parameters to the context result so they can be accessed
+        # inside "publish"
+        context_result['parameters'] = action_parameters
```
Kami: Since this is a "special" attribute maybe we should prefix it with `_` or similar. I'm open to suggestions.

manasdk: similar to https://github.com/StackStorm/st2/blob/action_chain_runner_publish_access_to_action_params/st2actions/st2actions/runners/actionchainrunner.py#L39 ?

Kami: Yep, I saw the results and I was wondering how and if we should prefix it.

I'm personally not a big fan of double underscore since `__` in Python means something internal which you shouldn't really touch. In the action parameters case, I want to prefix it to avoid potential name clashes and not because it's an "internal" / "secret" parameter.

Kami: To clarify it - inside the `params" section, action parameters are accessible in a root scope (there is no prefix), so we could also drop the prefix all together in the `publish` section.

E.g.

```yaml
---
  chain:
    -
      name: "create_temporary_directory"
      ref: "core.remote_sudo"
      params:
        hosts: "{{ ip_address }}"  -> no prefix here
        cmd: "mktemp -d"
      publish:
        temp_dir: "{{ create_temporary_directory[parameters.ip_address].stdout }}" -> we probably shouldn't have a prefix here for consistency reasons
      on-success: "retrieve_logged_users"
```

Kami:
As mentioned in https://github.com/StackStorm/st2/pull/1234#discussion_r26008026, I've decided to drop the attribute name prefix.

This way it's consistent with the existing code. Existing code already has a potential name clashes issue and solving it in just one case is worse than being consistent. We should consider this when designing an improved action chain runner and make sure we prefix the special attributes there.

dzimine:
Nice!

manasdk:
+1

jfryman:
Oh man, and this just merged as I was writing comments...

Adding `parameters` is not consistent with how we operate in Mistral. The diverging behaviors could lead to confusion...

Kami:
@jfryman Action parameters are already accessible everywhere else in the action chain (params section, vars section, etc.) except the `publish` section - this pr fixed that.

Not exactly sure how this affects the mistral runner though - probably I just don't have enough context on it.

jfryman:
@Kami Certainly following the PR... let me try to explain again.

Everywhere else in ActionChain, I can address parameters without the prefix `parameters`.

But, you have this... 

> Existing code already has a potential name clashes issue and solving it in just one case is worse than being consistent

Is there no way around this in the short term? There is nothing with Mistral that is directly affected, but the behavior is just odd. I have to remember the `parameters` prefix only in a specific part of the ActionChain workflow (publish). That's ugly.

Kami:
@jfryman I just had a look at the existing mistral workflow definition - I assume you are referencing how publish works there - you need to explicitly define which action parameter you want to be available using `input`?

If so, I see what you mean, but I don't think this is an issue - there is no such thing as input in the action chain runner right now so the behaviors can't be exactly the same.

Kami:
@jfryman Sounds like we are on the same page - I dropped the "parameters" prefix in the last commit. It's probably just my wording which made it confusing :)

jfryman:
@Kami AH! Great! Ok. Yeah, I was trying to parse the comments and code, and got confused. That was really my only concern. :heart:

Kami:
Heh, np - and yeah, I was kinda talking and replying to myself so it was slightly confusing :)

Kami:
Merged into 0.8.1 and master.

PR #1241, opened 2015-03-08
Add a script for visualizing action chain workflows

Kami:
I had this script laying around locally for a while now. I decided to clean it up a bit and add it to the repo.

I believe james created a similar thing so it might make sense to combine those scripts together.

Generated visualization file looks like this ([here](https://github.com/Kami/stackstorm-bsides-demo/blob/master/packs/bsides/actions/workflows/information_collection.yaml) is a work-flow used as an input):

information_collection


jfryman:
:+1:

lakshmi-kannan:
Yumm!

PR #1252, opened 2015-03-10
Fix a bug in the run-remote-script runner

Kami:
The runner ignored environment variables and authentication settings supplied as parameters (it always just used default global authentication settings).

We should also get this into the 0.8.2 release.


jfryman:
Woot!!!

Kami:
Merged into master and cherry picked changes into 0.8.2.

PR #1265, opened 2015-03-14
RFR: Tool to purge executions older than some timestamp

lakshmi-kannan:
```
(virtualenv)~/st2 git:STORM-1146/purge_execs_ops_tool ❯❯❯ tools/purge_executions.py --timestamp "2015-03-16T14:31:59.702134Z"
Purging executions older than timestamp: 2015-03-16T14:31:59.702134Z
#### Total number of executions to delete: 1
#### Total execution models deleted: 1
(virtualenv)~/st2 git:STORM-1146/purge_execs_ops_tool ❯❯❯ tools/purge_executions.py --timestamp "2015-03-16T14:31:59.702134Z"
Purging executions older than timestamp: 2015-03-16T14:31:59.702134Z
#### Total number of executions to delete: 0
#### Total execution models deleted: 0
(virtualenv)~/st2 git:STORM-1146/purge_execs_ops_tool ❯❯❯
```

### DIFF
--- a/tools/purge_executions.py
+++ b/tools/purge_executions.py
@@ -33,8 +33,11 @@ from st2common.models.db import db_setup
 from st2common.models.db import db_teardown
 from st2common.persistence.action import LiveAction
 from st2common.persistence.execution import ActionExecution
+from st2common.util import isotime
+
 
 DEFAULT_TIMEDELTA_DAYS = 2  # in days
+DELETED_COUNT = 0
 
 
 def _monkey_patch():
@@ -57,13 +60,19 @@ def do_register_cli_opts(opts, ignore_errors=False):
 
 def _purge_action_models(execution_db):
     liveaction_id = execution_db.liveaction['id']
+
     if not liveaction_id:
         print('Invalid LiveAction id. Skipping delete: %s', execution_db)
+
     liveaction_db = None
     try:
         liveaction_db = LiveAction.get_by_id(liveaction_id)
     except:
         print('LiveAction with id: %s not found. Skipping delete.', liveaction_id)
+    else:
+        global DELETED_COUNT
+        DELETED_COUNT += 1
+
     try:
         ActionExecution.delete(execution_db)
     except Exception as e:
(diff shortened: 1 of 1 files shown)


tools/purge_executions.py
```diff
@@ -0,0 +1,95 @@
...
+               help='Will delete data older than this timestamp. (default 48 hours).'),
+    cfg.StrOpt('action-ref', default='', help='action-ref to delete executions for.')
+]
+do_register_cli_opts(cli_opts)
+
+
+def _purge_executions(timestamp, action_ref):
+    print('Purging executions older than timestamp: %s',
+          timestamp.strftime('%Y-%m-%d %H:%M:%S.%f'))
+    executions = ActionExecution.get_all()
```
manasdk: I think you can use time_stamp lessthan query here.

tools/purge_executions.py
```diff
@@ -0,0 +1,95 @@
...
+    for execution in executions:
+        if execution.start_timestamp < timestamp:
+            if action_ref != '':
+                if execution.liveaction['action'] == action_ref:
+                    print('Deleting Execution: %s' % execution)
```
manasdk: don't forget to remove action execution and liveaction.

tools/purge_executions.py
lakshmi-kannan: Oh fancy pantsy. So I just got to convert to micros since epoch and add the query op.

manasdk:
Should we introduce a REST API to perform delete?

tools/purge_executions.py
Kami: Yeah, what manas said - you should be able to just convert it to a datetime object and perform timestamp__lt query.

tools/purge_executions.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def _purge_executions(timestamp, action_ref):
+    print('Purging executions older than timestamp: %s',
+          timestamp.strftime('%Y-%m-%d %H:%M:%S.%f'))
+    executions = ActionExecution.get_all()
+    for execution in executions:
+        if execution.start_timestamp < timestamp:
+            if action_ref != '':
+                if execution.liveaction['action'] == action_ref:
```
Kami: Oh my - 4 levels of nesting - cyclomatic complexity is off the roof! :trollface:

tools/purge_executions.py
Kami: But yeah, this is a good idea from performance reasons, especially in this case since action execution objects can be huge and retrieving a lot of very big objects can be slow.

An extra optimization would be to only retrieve fields we need (live_action, timestamp), but that's not a biggie.

tools/purge_executions.py
lakshmi-kannan: I mean I would write some functional mumbo jumbo.

lakshmi-kannan:
I am -1 to exposing an API to delete execution objects without introducing RBAC first. The tool at least requires a db username and password. We could imagine only admins having access to those.

PR #1268, opened 2015-03-16
[RFC] [WIP] Structured logging with contexual information

Kami:
This pull request introduces support for structured log messages with additional context information.

Structured logs by itself are really not all that useful if you just wrap `message` attribute in a JSON object. They become more useful when you include additional contextual information (you can then query, filter, etc. based on the presence and value of those additional attributes / contextual information).

I've considered using `structlog` and discussed it with @lakshmi-kannan, but I believe the custom formatters + extra attribute approach I went with is better in this case. We already use logging module from Python stdlib and this approach just builds on top of that.

All of the existing logging code and infrastructure (filters, formatters, handlers, etc.) will still work. If we want to use new structured format, we simply need to configure formatter in the logging configs as shown bellow.

And this approach also allows us to incrementaly build the context - same as structlog.

## Configuration

```ini
...
[formatter_consoleFormatter]
class=st2common.logging.formatters.ConsoleLogFormatter
format=%(asctime)s %(levelname)s [-] %(message)s

[formatter_gelfFormatter]
class=st2common.logging.formatters.GelfLogFormatter
format=%(message)s
...
```

## Example log message

```python
extra = {'_host': db_host, '_port': db_port, '_username': username, '_foo': [1,2,3], '_aa': {'abc': "d"}}
LOG.info('Connecting to database "%s" @ "%s:%s" as user "%s".' % (db_name, db_host, db_port, str(username)), extra=extra)
```

`extra` argument is a dictionary which includes additional context information which is added to the log message. All the items are prefixed with an underscore to avoid clashes with standard log record attributes (https://docs.python.org/2/library/logging.html#logrecord-attributes).

If user includes a model or some other class instance in `extra`, we call `to_dict()` method if one exists, otherwise we fall back to `repr()`. This allows us to eventually implement custom logging representation for model, api and other objects.

### ConsoleLogFormatter

This is a default "user-friendly" console log formatter we should use when logging to console and files.

```
2015-03-16 15:27:12,186 INFO [-] Connecting to database "st2" @ "0.0.0.0:27017" as user "None". (host='0.0.0.0',foo=[1, 2, 3],username='None',aa={'abc': 'd'},port=27017)
```

This formatter includes extra attributes at the end of the message as key=value pairs. This is nice because it allows users to grep for stuff (e.g. `grep id=foo`).

We should update all of our services to use this format by default when logging to console and file.

### GELFLogFormatter

This formatter formats messages as GELF-2 (https://www.graylog.org/resources/gelf-2/). Gelf format is basically structured JSON with some common attributes.

```json
{"_username": "None", "version": "1.1", "level": 20, "timestamp": 1426519705, "_python": {"process": 4880, "funcName": "db_setup", "module": "__init__", "filename": "__init__.py"}, "_port": 27017, "_foo": [1, 2, 3], "host": "vagrant-ubuntu-trusty-64", "full_message": "Connecting to database \"st2\" @ \"0.0.0.0:27017\" as user \"None\".", "_aa": {"abc": "d"}, "_host": "0.0.0.0", "short_message": "Connecting to database \"st2\" @ \"0.0.0.0:27017\" as user \"None\"."}
```

This formatter can be used when user wants structured log messages (e.g. when shipping logs to Graylog / loggly / splunk / logstash / etc.).

## TODO

In short, the goal is to get the foundation and basic changes in first. After those changes have been merged, we can start populating `extra` and adding contextual information to log lines.

- [x] Update affected code in st2api/
- [x] Update affected code in st2auth/
- [ ] Update affected code in st2common/
- [ ] Update affected code in st2debug/
- [ ] Update affected code in st2actions/
- [ ] Update affected code in st2reactor/
- [ ] Update default logging configs
- [x] Documentation section on logging

### DIFF
--- a/st2common/st2common/logging/formatters.py
+++ b/st2common/st2common/logging/formatters.py
@@ -31,11 +31,11 @@ __all__ = [
 HOSTNAME = socket.gethostname()
 COMMON_ATTRIBUTE_NAMES = [
     'process',
+    'processName',
     'module',
     'filename',
     'funcName',
     'lineno'
-    'processName',
 ]
 
 
@@ -149,7 +149,7 @@ class GelfLogFormatter(BaseExtraLogFormatter):
             data['_exception'] = str(exc_value)
             data['_traceback'] = tb_str
 
-        # Include common Python process attributes
+        # Include common Python log record attributes
         data['_python'] = common_attributes
 
         # Include user extra attributes
--- a/st2common/st2common/logging/formatters.py
+++ b/st2common/st2common/logging/formatters.py
@@ -114,7 +114,10 @@ class ConsoleLogFormatter(BaseExtraLogFormatter):
         # Call the parent format method so the final message is formed based on the "format"
         # attribute in the config
         msg = super(ConsoleLogFormatter, self).format(record)
-        msg = '%s (%s)' % (msg, attributes)
+
+        if attributes:
+            msg = '%s (%s)' % (msg, attributes)
+
         return msg
 
     def _dict_to_str(self, attributes):
--- a/st2common/st2common/logging/formatters.py
+++ b/st2common/st2common/logging/formatters.py
@@ -19,6 +19,7 @@ import logging
 import socket
 import time
 import json
+import traceback
 
 import six
 
@@ -113,19 +114,32 @@ class GelfLogFormatter(BaseExtraLogFormatter):
         attributes = self._format_extra_attributes(attributes=attributes)
 
         msg = record.msg
+        exc_info = record.exc_info
         now = int(time.time())
 
         common_attributes = self._get_common_extra_attributes(record=record)
+        full_msg = super(GelfLogFormatter, self).format(record)
 
         data = {
             'version': '1.1',
             'host': HOSTNAME,
             'short_message': msg,
-            'full_message': msg,
+            'full_message': full_msg,
             'timestamp': now,
             'level': record.levelno
         }
+
+        if exc_info:
+            # Include exception information
+            exc_type, exc_value, exc_tb = exc_info
+            tb_str = ''.join(traceback.format_tb(exc_tb))
+            data['_exception'] = str(exc_value)
+            data['_traceback'] = tb_str
+
+        # Include common Python process attributes
         data['_python'] = common_attributes
+
+        # Include user extra attributes
         data.update(attributes)
 
         msg = json.dumps(data)
--- a/st2common/st2common/logging/formatters.py
+++ b/st2common/st2common/logging/formatters.py
@@ -109,6 +109,18 @@ class GelfLogFormatter(BaseExtraLogFormatter):
     """
     Formatter which formats messages as GELF 2 - https://www.graylog.org/resources/gelf-2/
     """
+
+    # Maps python log level to syslog / gelf log level
+    PYTHON_TO_GELF_LEVEL_MAP = {
+        50: 2,  # critical -> critical
+        40: 3,  # error -> error
+        30: 4,  # warning -> warning
+        20: 6,  # info -> informational
+        10: 7,  # debug -> debug
+        0: 6,  # notset -> information
+    }
+    DEFAULT_LOG_LEVEL = 6  # info
+
     def format(self, record):
         attributes = self._get_extra_attributes(record=record)
         attributes = self._format_extra_attributes(attributes=attributes)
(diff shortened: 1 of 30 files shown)


st2common/st2common/models/api/base.py
```diff
@@ -109,6 +111,16 @@ def decorate(f):
...
 
+            # Common request information included in the log context
+            request_info = {'method': pecan.request.method, 'path': pecan.request.path,
+                            'remote_addr': pecan.request.remote_addr}
+
+            # Log the incoming request
+            values = copy.copy(request_info)
+            values['filters'] = kwargs
+            extra = prefix_with_underscore(values)
+            LOG.info('%(method)s %(path)s with filters=%(filters)s' % values, extra=extra)
```
Kami: This way we can get rid of all the duplicated log statements in the API controllers. In addition to that, all those log messages will now include additional context such as request method, path, client ip address, etc.

Kami:
@DoriftoShoes Are you OK with me changing the log formats in the following way:

* console - Use the `ConsoleLogFormatter` formatter described above.
* file - Use `GelfLogFormatter` formatter described above. Or do we want to stick to the more console friendly `ConsoleLogFormatter` for now and only use gelf (structured JSON) for syslog?
* syslog - Use `GelfLogFormatter` formatter described above.

Kami:
I've pushed some changes so now you don't need to prefix `extra` keys with an underscore anymore.

There is some "magic" involved and I really hate magic, but I think in this case it's fine since we always obtain a reference to the logger using `st2common.log.getLogger` and we don't do any nasty stuff with stdlib.

st2common/st2common/logging/formatters.py
```diff
@@ -0,0 +1,209 @@
...
+        msg = record.msg
+        exc_info = record.exc_info
+        now = int(time.time())
+        level = self.PYTHON_TO_GELF_LEVEL_MAP.get(record.levelno, self.DEFAULT_LOG_LEVEL)
+
+        common_attributes = self._get_common_extra_attributes(record=record)
+        full_msg = super(GelfLogFormatter, self).format(record)
+
+        data = {
+            'version': '1.1',
```
lakshmi-kannan: Nit: Probably better to declare in constants.

st2common/st2common/logging/formatters.py
```diff
@@ -0,0 +1,209 @@
...
+
+        return msg
+
+    def _dict_to_str(self, attributes):
+        result = []
+        for key, value in six.iteritems(attributes):
+            item = '%s=%s' % (key[1:], repr(value))
+            result.append(item)
+
+        result = ','.join(result)
```
lakshmi-kannan: ', ' would be better to read IMO.

Kami: Slightly confused - what do you mean with that? It does use a comma right now. Or you mean comma + whitespace?

Kami: We could also wrap each value in quotes to make value boundaries more explicit, but this is a default user-friendly "somewhat structured" format and we don't use JSON which means it's a trade off between making it user-friendly and at least semi-structured so user can easily grep for stuff.

lakshmi-kannan: umm. not sure how that whitespace kills grep but i'll leave this to you. it's minor.

st2common/st2common/logging/formatters.py
Kami: I was thinking of declaring a module level constant with a base dictionary template, but then we would really need to use copy.copy inside the format calls which is less than ideal.

st2common/st2common/logging/formatters.py
DoriftoShoes: whitespace makes awk foo harder.

Kami: @lakshmi-kannan whitespace killed the radio star.

(bad reference to the 80's song :P)

DoriftoShoes:
Overall I am +1 to this.  I am curious how this will look with some of the really hairy execution results.  Logstash has a gelf input so it should be just as easy to consume this.

Kami:
Alright, pushed the configs change:

* Use ConsoleLogFormatter for console and file handlers
* Use GelfLogFormatter for audit handler
* Leave syslog configs and handlers untouched
* Also leave action runner audit log runner untouched co it continues to use json formatter. Eventually we should switch to the gelf one though.

DoriftoShoes:
+1

Kami:
Alright, I've merged my changes into master. We now have a foundation for structured logging in place.

We still need to update most of the existing log calls to include context to make them actually more useful, but that's something we can do incrementally and along the way. We should of course also do this for all the newly added code.

Good thing is that now that we have a foundation in place (and once all of the log statements include a context) we (or the user) can easily update log configs to change the log format for a particular handler (console / file / syslog / etc.).

PR #1269, opened 2015-03-17
Important Doc Fixes

jfryman:
This is a bit of cleanup from a review @kirilstrax did for us. I'd like to be able to use something like `|version|` to dynamically populate areas like this in code, and will look at this shortly. In the meantime, this fixes documentation to be functional for debian.

:cherries:-picked a similar fix at https://github.com/StackStorm/st2/commit/545b34fb40068a70a070fab8ed42abd44993c3ed

### DIFF
--- a/docs/source/install/deb.rst
+++ b/docs/source/install/deb.rst
@@ -19,11 +19,10 @@ APT
 Pip
 '''
 
-The easiest way to install these is to use the requirements.txt file from the |st2| downloads server.  This is kept up to date for the version specified in the path.
+The easiest way to install these is to use the requirements.txt file from the |st2| repository.  This is kept up to date for the version specified in the path.
 
 ::
-
-    curl -q -k -O https://ops.stackstorm.net/releases/st2/0.9dev/requirements.txt
+    curl -q -k -O https://raw.githubusercontent.com/StackStorm/st2/master/requirements.txt
     pip install -r requirements.txt
 
 RabbitMQ
@@ -33,7 +32,7 @@ In order to get the latest version of RabbitMQ, you will want to follow the dire
 
 ::
 
-    http://www.rabbitmq.com/install-rpm.html
+    http://www.rabbitmq.com/install-debian.html
 
 Once you have RabbitMQ installed, you will need to run the following commands to enable certain plugins.
 


dzimine:
|release| and |version| are already available in the docs, with one unfortunate exception - I dind't figure how to use them in the formatted text or code block (specifically, [here](https://github.com/StackStorm/st2/blame/master/docs/source/install/index.rst#L9))

docs/source/install/deb.rst
```diff
@@ -19,11 +19,10 @@ APT
 Pip
 '''
 
-The easiest way to install these is to use the requirements.txt file from the |st2| downloads server.  This is kept up to date for the version specified in the path.
+The easiest way to install these is to use the requirements.txt file from the |st2| repository.  This is kept up to date for the version specified in the path.
 
 ::
-
-    curl -q -k -O https://ops.stackstorm.net/releases/st2/0.9dev/requirements.txt
+    curl -q -k -O https://raw.githubusercontent.com/StackStorm/st2/master/requirements.txt
```
Kami: Afaik, we want to link to the current version here, not to master, right?

jfryman: 0.9dev does not exist as a branch, and I assumed (incorrectly?) that TOT is 0.9dev.

Kami: Your assumption is correct, but requirements should live on ops.stackstorm.net and there should be a 0.9dev directory there.

Edit: I just had a look - looks like we need to update / fix our build workflow to also copy the requirements file to the dist directory - new versions are missing it.

st2stanley: Right. I believe the missing link is in fact why Kiril brought it up in the first place. 
But, is there a reason why we wouldn't point to the github repo? Is this to try and track downloads?

-James 




On Tue, Mar 17, 2015 at 9:20 AM -0700, "Tomaz Muraus" <notifications@github.com> wrote:












In docs/source/install/deb.rst:
>  
>  ::
> -
> -    curl -q -k -O https://ops.stackstorm.net/releases/st2/0.9dev/requirements.txt
> +    curl -q -k -O https://raw.githubusercontent.com/StackStorm/st2/master/requirements.txt



Your assumption is correct, but requirements should live on ops.stackstorm.net and there should be a 0.9dev directory there.



Edit: I just had a look - looks like we need to update / fix our build workflow to also copy the requirements file to the dist directory - new versions are missing it.

manasdk:
Presumably not very important since it is stuck around for more than a month :P

PR #1270, opened 2015-03-17
Add ability to get verbose logs for st2 pip rqmnts installation

lakshmi-kannan:


Makefile
```diff
@@ -127,7 +133,7 @@ requirements: virtualenv $(REQUIREMENTS)
 	@echo
 	for req in $(REQUIREMENTS); do \
 		echo "Installing $$req..." ; \
-		. $(VIRTUALENV_DIR)/bin/activate && pip install -U -q -r $$req ; \
+		. $(VIRTUALENV_DIR)/bin/activate && pip install $(PIP_OPTIONS) $$req ; \
```
Kami: We always want to specify at least `-r` since we are installing from a file, right?

lakshmi-kannan: https://github.com/StackStorm/st2/pull/1270/files#diff-b67911656ef5d18c4ae36cb6741b7965R32 -r is specified by default.

Kami: Yeah, I meant if user explicitly specifies the arguments - if they don't specify -r it would blow up.

lakshmi-kannan: Yeah. They better know what they are doing if they are overriding it :P. 

Alternatively, we can make variable changes to just get 'quiet' or not from env variable. Seems less flexible.

Kami: I would just always put -r in front of the file since without it, it doesn't really make sense.

Anyway :P

PR #1275, opened 2015-03-18
Flow/Grammar changes for Rules

jfryman:
Here is a first pass at changes for the Rules page. My goal here is to ensure that there is a basic flow to each of the pages:

* Detailed description of each section
* Deep dive into context specific sections

This adds a beginning section that describes each portion of a rule in greater depth, hopefully providing additional context to users.

I also intend to rip out the part in this document with built-in sensors, and create a cross-link. However, leaving for now to limit scope. 

:eyes: and :book: appreciated.


lakshmi-kannan:
<3 it.

PR #1278, opened 2015-03-19
Enable authentication by default for production (package) installations

Kami:
This pull request enables auth by default for production (package) based installations.

I've explicitly didn't enable it by default in the source tree. If a user checks out the code from the source tree, this usually means they want to play with StackStorm or do the development.

In this case it's nice if we can avoid the overhead of enabling and configuring the authentication (user would need to configure certificates, apache, etc.) which slows things down.


st2common/packaging/debian/postinst
```diff
@@ -3,6 +3,9 @@
 if [ -d /etc/st2 ]; then
   sed -i -r "s~(st2.*)/conf~/etc/\1~g" /etc/st2/st2.conf
   sed -i "s~vagrant~/home/stanley~g" /etc/st2/st2.conf
+
+  # Enable auth by default
+  sed -i "s~enable = False~enable = True~g" /etc/st2/st2.conf
```
Kami: I'm not a fan of this approach - I prefer to be explicit. Doing that inside postinst step makes it kinda invisible...

I was thinking about having 2 configs - `conf/st2.dev.conf` and `conf/st2.prod.conf` and we use prod one for packages. @DoriftoShoes wdyt?

This way it's more explicit and visible

DoriftoShoes: I agree.  Two configs.  The we can offer a script based solution for initial setup if people want it.  Like `mysql_secure_install`

lakshmi-kannan: +1 to explicit configurations for dev, prod.

On a related note: https://github.com/StackStorm/st2/blob/master/st2tests/conf/st2.conf We should rename that.

Kami: Moved to two configs approach in 99ca964fd8dee5eb73389ef8b625921b472f5eba

I can also move tests config to `conf/st2.tests.conf` if ppl think this is a better place.

PR #1280, opened 2015-03-19
Fix packs.setup_virtualenv so it works correctly if user specified multiple packs search paths

Kami:
Previously packs.setup_virtualenv command didn't work correctly if user used multiple packs search paths (`content.packs_base_paths` setting).

### DIFF
--- a/contrib/packs/actions/pack_mgmt/unload.py
+++ b/contrib/packs/actions/pack_mgmt/unload.py
@@ -1,6 +1,5 @@
 from oslo.config import cfg
 
-import st2common.config as config
 from st2common.models.db import db_setup
 from st2actions.runners.pythonrunner import Action
 from st2common.persistence import reactor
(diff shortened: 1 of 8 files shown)


contrib/packs/actions/pack_mgmt/unload.py
```diff
@@ -16,13 +15,7 @@ def __init__(self, config=None):
         self.initialize()
 
     def initialize(self):
```
Kami: This was basically broken all along - we didn't correctly inherit parent arguments which means those actions never had access to the right config values (they always just access to the default ones).

Pack management API, here we come.

st2common/st2common/content/utils.py
```diff
@@ -138,6 +139,32 @@ def get_pack_base_path(pack_name):
...
+
+    :param pack_name: Pack name.
+    :type pack_name: ``str``
+
+    :return: Pack to the pack directory.
+    :rtype: ``str`` or ``None``
+    """
+    packs_base_paths = get_packs_base_paths()
+    for packs_base_path in packs_base_paths:
+        pack_base_path = os.path.join(packs_base_path, pipes.quote(pack_name))
```
lakshmi-kannan: this is not pack_base_path anymore. can you use a new variable?

Kami: Wdym?

lakshmi-kannan:
LGTM.

st2common/st2common/content/utils.py
lakshmi-kannan: I meant os.path.join(packs_base_path, pipes.quote(pack_name) would return pack_path and not packs_base_path :P. Sorry.

Kami:
Tests passed locally, for some reasons travis is stuck again.

Merged into 0.8.3 and master.

PR #1283, opened 2015-03-20
RFR: Fix st2_deploy to use st2-register-content

lakshmi-kannan:
Just noticed we don't add the extension for other tools. So fixed that to be consistent across.


Kami:
:+1:

(better late than never)

PR #1284, opened 2015-03-20
[0.8.3] Fix api url used by sensors

Kami:
Previously, sensors used api listen host and port setting to obtain a public URL to the API.

Obviously this wasn't a good idea and it wouldn't work if sensor container ran on a different host or similar (well, you could get it to work, but you would need to use different values for `api.host` setting on different servers which is nasty). This pull request fixes that and makes it use `auth.api_url` setting.

This way it's also consistent with other code and we only have one placed where a public URL to the API is configured.

Installation documentation already mentions that this setting needs to be configured correctly because it's also used in other places so no changes are needed there.

### DIFF
--- a/st2common/tests/unit/test_util_api.py
+++ b/st2common/tests/unit/test_util_api.py
@@ -1,0 +1,68 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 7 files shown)


Kami:
I need to make the same change against master, but I will open a separate PR for that since master has diverged substantially from 0.8.3...

st2common/tests/unit/test_util_api.py
```diff
@@ -0,0 +1,68 @@
...
+from st2common.util.api import get_base_public_api_url
+from st2common.util.api import get_full_public_api_url
+from st2tests.config import parse_args
+parse_args()
+
+
+class APIUtilsTestCase(unittest2.TestCase):
+    def test_get_base_public_api_url(self):
+        values = [
+            'http://foo.bar.com',
```
lakshmi-kannan: You should always use https </troll>

Kami: :facepunch:

lakshmi-kannan:
LGTM.

PR #1288, opened 2015-03-23
First selenium test

kirilstrax:
This is a very simple, but working selenium test case

What it does
------------------
- opens WebUI, 
- navigates to Actions tab, 
- selects "core.local" and specifies the command 'uname'
- navigates to History tab
- waits for command to complete 
- checks that the output is as expected ('linux')

Limitations
-----------------
- Not using any libraries yet
- Not in pack or workflow format yet
- Although I've made it reasonably readable, it's far from ideal
- Some hardcoded values/assumptions made in terms of page structure

### DIFF
--- a/st2tests/tests/selenium/selenium_sanity.py
+++ b/st2tests/tests/selenium/selenium_sanity.py
@@ -1,0 +1,96 @@
+import datetime
+
+from selenium import webdriver
+from selenium.common.exceptions import TimeoutException
+from selenium.webdriver.support.ui import WebDriverWait
+from selenium.webdriver.support import expected_conditions as EC
+from selenium.webdriver.common.by import By
+
+def navigate_to_page(page_url, page_signature_xpath):
+    print "Navigating to %s" % page_url
+    driver = webdriver.Firefox()
+    driver.get(page_url)
+    condition = EC.presence_of_element_located((By.XPATH, page_signature_xpath))
+    WebDriverWait(driver, 10).until(condition)
+    return driver
+
+def wait_for_element_by_xpath(driver, xpath):
+    condition = EC.presence_of_element_located((By.XPATH, xpath))
+    element = WebDriverWait(driver, 10).until(condition)
+    return element
+
+def wait_for_view_to_load(driver, view_name):
+    print "Wait for %s to load" % view_name
+    wait_for_element_by_xpath(driver, '//div[contains(@class,"panel__toolbar-title") and contains(text(), "%s")]' % view_name)
+    return
+
+def login(driver):
+    print "Login"
+    condition = EC.element_to_be_clickable((By.XPATH,'//input[contains(@class,"login__button")]'))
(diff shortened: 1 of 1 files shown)


dzimine:
@kirilstrax we need to find a better place for it, st2 is not a good one, and I don't know the answer yet, let's discuss on slack. Obvious options are 1) with st2web and 2) separate repo for e2e tests cc @DoriftoShoes

kirilstrax:
@dzimine I submitted it here as a trial. I think separate repo is more practical, but don't have permissions to create one.
How about st2e2e? Or whatever you prefer really

st2tests/tests/selenium/selenium_sanity.py
```diff
@@ -0,0 +1,96 @@
+import datetime
```
lakshmi-kannan: Needs LICENSE info.

lakshmi-kannan:
If it's in st2web, it should be all javascript right? These tests should be changed to run as part of a test suite/test runner. Our testing story cannot be here run these scripts in order. This definitely needs a discussion.

st2tests/tests/selenium/selenium_sanity.py
kirilstrax: Do you mean for Selenium?
It's Apache 2.0 (http://www.seleniumhq.org/about/license.jsp)

kirilstrax:
@lakshmi-kannan This could be another argument for putting selenium tests on a separate repository. Although moving them to JS driver (Node based) is possible as well.

dzimine:
Between two Kirills, please decide what's the best. 
If you decide that selenium is WebUI tests, it may be reasonalbe to keep it in st2web with the web.
Making selenium for complete e2e tests - more reasons to move it out to a separate repo.
On languages, it would be good if we stay with JavaScript and Python.

enykeev:
I'm pretty sure we would need several types of selenium tests to cover anything that should be covered. This one seems like a basic e2e covering st2web, st2 and potentially st2contrib. The tests may fail due to changes in any of the projects so it seems logical to me
- keep them separate (+1 for st2e2e, https://github.com/StackStorm/st2e2e here you go)
- write them in python (most of our codebase written in Python, anyone in the team can write in it so anyone would be able to fix them or spot the problem. We can formulate some kind of guidelines st2web should follow to avoid breaking tests too often)
- expect them to fail sometime (having multiple repos means there's always a window of opportunity for code and tests end up being out of sync. We definitely should not run them on repo push\merge)

We would also need selenium units and e2e specifically for st2web covering every feature we have there. Those should be in JS, take into account angular specific, be a part of st2web repository and run on every st2web build (we'll see whether it would be possible to run them for every commit in PR as the ones Travis runs for us right now.).

Kami:
Closing since this is not relevant anymore. Tests have moved to the st2tests repo - https://github.com/StackStorm/st2tests.

PR #1292, opened 2015-03-24
Update st2_deploy to configure authentication service with test credentials

Kami:
TODO:

- [x] Add back auth backend specific docs.
- [x] This depends on the st2client.js change - client needs to be updated to inherit the protocol and not default to https for auth (@enykeev is on it)


Kami:
Alright, @enykeev merged his changes and bumped st2client version in st2web.

PR #1295, opened 2015-03-26
Update out of date datastore docs, fix some broken client datastore operations

Kami:


st2client/st2client/models/datastore.py
```diff
@@ -27,3 +27,8 @@ class KeyValuePair(core.Resource):
     _plural = 'Keys'
     _plural_display_name = 'Key Value Pairs'
     _repr_attributes = ['name', 'value']
+
+    @property
+    def id(self):
```
Kami: This is far from ideal, but client needs a lot of refactoring love anyway.

Non "id" primary keys should definitely be supported by the client (CLI already has some workarounds to support non "id" PKs - there is a concept of a "pk" which is not necessary an id, etc.).

PR #1297, opened 2015-03-26
Moved download server to variable and pointed to downloads server

DoriftoShoes:
This PR creates a variable for the downloads server and sets it to the new public downloads server.

### DIFF
--- a/tools/st2_deploy.sh
+++ b/tools/st2_deploy.sh
@@ -15,6 +15,7 @@ EOM
 WARNING_SLEEP_DELAY=5
 
 # Common variables
+DOWNLOAD_SERVER="https://downloads.stackstorm.net"
 RABBIT_PUBLIC_KEY="rabbitmq-signing-key-public.asc"
 PACKAGES="st2common st2reactor st2actions st2api st2auth st2debug"
 CLI_PACKAGE="st2client"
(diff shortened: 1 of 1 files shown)


Kami:
There appears to be some differences between this new server and https://ops.stackstorm.net/releases/st2/.

* https://downloads.stackstorm.net/releases/st2/0.8.3/ - misses st2_deploy.sh file.
* https://downloads.stackstorm.net/releases/st2/0.8.9dev/ - misses st2_deploy.sh file
* https://downloads.stackstorm.net/releases/st2/ - no 0.8dev directory
* ...

DoriftoShoes:
The new wrapper looks in the {TYPE} subdirectory for the st2_deploy.sh file.  I did not copy over the 0.8dev packages since people should be pulling either 0.8.X or 0.9dev.  If we feel the 0.8dev packages are necessary I can put them there.  Do we want to leave 0.8dev out there?

Kami:
I'm fine with not including 0.8dev - just wanted to make sure it was a conscious decision :)

DoriftoShoes:
Yup :) I did it on purpose.  Also the st2_deploy.sh location doesn't matter since it is hardcoded in to the new wrapper script that is pulled from the scripts/ directory.

DoriftoShoes:
For consistency's sake I copied the appropriate st2_deploy.sh scripts into the root of the version dir.

PR #1298, opened 2015-03-26
Fabric capture stderr stdout in logs.

lakshmi-kannan:
rm -f the script. This might fix the problem we are seeing in st2build001.

### DIFF
--- a/st2common/st2common/models/system/action.py
+++ b/st2common/st2common/models/system/action.py
@@ -512,7 +512,7 @@ class FabricRemoteScriptAction(RemoteScriptAction, FabricRemoteAction):
             result = action_method()
 
             # Cleanup.
-            cmd1 = self._get_command_string(cmd='rm', args=[self.remote_script])
+            cmd1 = self._get_command_string(cmd='rm -f', args=[self.remote_script])
             cmd2 = self._get_command_string(cmd='rm -rf', args=[self.remote_dir])
             self._execute_remote_command(cmd1)
             self._execute_remote_command(cmd2)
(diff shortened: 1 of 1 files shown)


st2common/st2common/models/system/action.py
```diff
@@ -544,6 +544,15 @@ def _execute_remote_command(self, command):
...
+            # unicode and action fails. For now, just log stdout and stderr so we can debug
+            # from logs.
+            # Fabric will show an exception traceback like:
+            # 'ascii' codec can't encode character u'\u2018' in position 93:
+            # ordinal not in range(128)
+            #
+            LOG.error('stderr: %s', output.stderr)
+            LOG.error('stdout: %s', output.stdout)
             LOG.error(msg)
             raise Exception(msg)
```
Kami: Ah, I see.

We throw and that's why stdout and stderr is not included in the output. Imo it would be good and useful to include it. We could throw a special exception and handle this case in the fabric runner.

lakshmi-kannan: No. I tried this approach but fabric task has failed in this case. So env.abort throws a FabricFailureException anyways. So you see that crazy ascii decode error.

Kami: Sorry, I might not understand you fully / correctly.

Even though the task has failed, if the command actually ran and it's not a different error, stdout and stderr should still be included to aid with debugging. If nothing else, maybe in the exception message.

lakshmi-kannan: Yeah, it's hard to explain here. If a task fails, none of the results there is collected cos this throws.
https://github.com/StackStorm/st2/blob/master/st2actions/st2actions/runners/fabricrunner.py#L144

That line throws because rm was setting the fabric state to fail even though the original fabric action to run the script succeeded. Am I making sense?

PR #1299, opened 2015-03-26
Catch exception around criteria evaluation

lakshmi-kannan:


st2reactor/st2reactor/rules/filter.py
```diff
@@ -92,7 +92,11 @@ def _check_criterion(self, criterion_k, criterion_v, payload_lookup):
 
         op_func = criteria_operators.get_operator(criteria_operator)
 
-        return op_func(value=payload_value, criteria_pattern=criteria_pattern)
+        try:
+            return op_func(value=payload_value, criteria_pattern=criteria_pattern)
+        except:
+            LOG.exception('There might be a problem with critera in rule %s.', self.rule)
+            False
```
manasdk: return False?

lakshmi-kannan: heh. :Face Palm:

st2reactor/st2reactor/rules/filter.py
```diff
@@ -92,7 +92,11 @@ def _check_criterion(self, criterion_k, criterion_v, payload_lookup):
 
         op_func = criteria_operators.get_operator(criteria_operator)
 
-        return op_func(value=payload_value, criteria_pattern=criteria_pattern)
+        try:
+            return op_func(value=payload_value, criteria_pattern=criteria_pattern)
+        except:
+            LOG.exception('There might be a problem with critera in rule %s.', self.rule)
```
Kami: I guess that's fine since we don't really have any other fail to fail loudly and "propagate" this error to the user in this case.

lakshmi-kannan: Correct.

PR #1300, opened 2015-03-26
Added auth token to deploy script

DoriftoShoes:
St2 now ships with auth on by default.  We need to get a token in order for the test command to work.

### DIFF
--- a/tools/st2_deploy.sh
+++ b/tools/st2_deploy.sh
@@ -498,7 +498,8 @@ echo "# Starting St2 Services"
 st2ctl restart
 sleep 20
 ##This is a hack around a weird issue with actions getting stuck in scheduled state
-st2 run core.local date &> /dev/null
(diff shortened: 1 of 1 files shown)


tools/st2_deploy.sh
```diff
@@ -498,7 +498,8 @@ echo "# Starting St2 Services"
 st2ctl restart
 sleep 20
 ##This is a hack around a weird issue with actions getting stuck in scheduled state
-st2 run core.local date &> /dev/null
+TOKEN=`st2 auth ${TEST_ACCOUNT_USERNAME} -p ${TEST_ACCOUNT_PASSWORD} | grep token | awk '{print $4}'a`
```
Kami: Is "a" at the end intentional?

Kami:
Good catch.

tools/st2_deploy.sh
```diff
@@ -498,7 +498,8 @@ echo "# Starting St2 Services"
...
+ST2_AUTH_TOKEN=${TOKEN} st2 run core.local date &> /dev/null
```
Kami: Could also use -t option here instead of env variable. Anyway, either works :)

PR #1306, opened 2015-03-28
Filename change, based on the name referenced in metadata file

kirilstrax:
Metadata file defines this file as
...
name: "test_quickstart"
...
entry_point: "chains/test_quickstart.yaml"
...

So I am changing the name of this file to match that reference


Kami:
Merged. Thanks.

PR #1308, opened 2015-03-29
Added test_quickstart_rules workflow (tests rules section of quick start)

kirilstrax:


contrib/tests/actions/chains/test_quickstart_rules.yaml
```diff
@@ -0,0 +1,104 @@
...
+vars:
+        tested_rule: examples.webhook_file
+
+chain:
+    -
+        # Check if tested rule is already registered
+        name: setup_sample_rule_with_webhook_exists
+        ref: core.local
+        params:
+            cmd: "st2 rule get {{ tested_rule }}"
```
DoriftoShoes: We should add the rule we want to test as a parameter.

contrib/tests/actions/chains/test_quickstart_rules.yaml
```diff
@@ -0,0 +1,104 @@
...
+        ref: core.local
+        params:
+            cmd: "st2 rule delete {{ tested_rule }} "
+        on-success: test_create_rule
+        on-failure: error_handler
+    -
+        name: test_create_rule
+        ref: core.local
+        params:
+            cmd: "st2 rule create /usr/share/doc/st2/examples/rules/sample_rule_with_webhook.yaml"
```
DoriftoShoes: We can also parameterize the path to the the samples.

contrib/tests/actions/chains/test_quickstart_rules.yaml
```diff
@@ -0,0 +1,104 @@
...
+        ref: core.local
+        params:
+            cmd: "st2 rule get {{ tested_rule }}"
+        on-success: setup_post_to_web_hook_via_curl
+        on-failure: error_handler
+    -
+        name: setup_post_to_web_hook_via_curl
+        ref: core.local
+        params:
+            cmd: "curl http://localhost:9101/v1/webhooks/sample -d '{\"foo\": \"bar\", \"name\": \"st2\"}' -H 'Content-Type: application/json'"
```
DoriftoShoes: base url as a parameter also.  Possibly even the sample data.

kirilstrax:
@DoriftoShoes indeed, originally I added those 2 things (the name and path of the rule) as vars. But then I realized that there are 4 steps that are not quite as flexible: setup_post_to_web_hook_via_curl, test_check_output_after_post_via_wget, setup_post_to_web_hook_via_st2 and test_check_output_after_post_via_st2

So how should we handle those steps? We can of course make the entire cmd string in those steps a parameter, but then it won't be very readable. 
What do you think?

PR #1309, opened 2015-03-29
Added test_quickstart_key workflow (tests 3 key commands from quick start)

kirilstrax:


contrib/tests/actions/chains/test_quickstart_key.yaml
```diff
@@ -0,0 +1,43 @@
+---
+chain:
+    -
+        name: setup_uid
+        ref: core.local
+        params:
+            cmd: "uuidgen | sed 's/-//g' | tr -d ' \n'"
```
dzimine: note for @jfryman : good candidate for Linux/utils pack.

contrib/tests/actions/chains/test_quickstart_key.yaml
```diff
@@ -0,0 +1,43 @@
+---
```
lakshmi-kannan: I generally recommend double quoting all the string values (as in key values) in YAML for safety.

lakshmi-kannan:
LGTM

PR #1312, opened 2015-03-29
Allow users to reference datastore items inside the rule criteria pattern

Kami:
This pull request allows users to access datastore items inside a rule criteria pattern by utilizing a jinja variable expression.

Before, you could only reference a datastore item inside the criteria key, but not inside the pattern value - imo, value makes a lot more sense than criterion key.

@DoriftoShoes had a good use case for this one and I also have one.

For example:

```yaml
...

    criteria:
        trigger.parameters.cmd:
            type: "equals"
            pattern: "{{ system.action_cmd }}"
...
```

TODO:

- [x] Tests
- [x] Documentation


DoriftoShoes:
woohoo!  Thanks buddy.

st2reactor/st2reactor/rules/filter.py
```diff
@@ -85,19 +109,33 @@ def _check_criterion(self, criterion_k, criterion_v, payload_lookup):
...
             return False
 
+        return result
+
+    def _render_criteria_pattern(self, criteria_pattern):
+        if not criteria_pattern:
+            return None
+
+        if not isinstance(criteria_pattern, six.string_types):
+            # We only perform rendering if value is a string - rendering a string value makes no
```
lakshmi-kannan: This comment also makes no sense :P

Kami: Yeah, there is something missing :P

st2common/st2common/util/templating.py
```diff
@@ -0,0 +1,63 @@
...
+
+    :param context: Template context.
+    :type context: ``dict``
+    """
+    assert isinstance(value, six.string_types)
+    context = context or {}
+
+    env = Environment(undefined=StrictUndefined)
+    template = env.from_string(value)
+    rendered = template.render(context)
```
lakshmi-kannan: We might have to wrap this within try catch.

Kami: We do this inside the calling function (inside RuleFilter).

Our goal should always be to propagate exceptions at the low level and handle them higher up where it's most appropriate.

lakshmi-kannan: Got it.

st2reactor/st2reactor/rules/filter.py
```diff
@@ -41,8 +43,22 @@ def __init__(self, trigger_instance, trigger, rule):
...
+
     def filter(self):
-        LOG.info('Validating rule %s for %s.', self.rule.id, self.trigger['name'])
+        """
+        Return true if the rule is applicable to the provided trigger instance.
+
+        :rtype: ``bool``
+        """
+        LOG.info('Validating rule %s for %s.', self.rule.id, self.trigger['name'],
+                 extra=self._base_logger_context)
```
manasdk: putting the entire rule, trigger and triggerinstance object will be excessive.

st2reactor/st2reactor/rules/filter.py
```diff
@@ -54,7 +70,9 @@ def filter(self):
 
         payload_lookup = PayloadLookup(self.trigger_instance.payload)
 
-        LOG.debug('Trigger payload: %s', self.trigger_instance.payload)
+        LOG.debug('Trigger payload: %s', self.trigger_instance.payload,
```
manasdk: This is pointless if you have the previous line at info

st2reactor/st2reactor/rules/filter.py
```diff
@@ -62,20 +80,26 @@ def filter(self):
...
-        if 'pattern' not in criterion_v:
-            criterion_v['pattern'] = None
+        criteria_operator = criterion_v['type']
+        criteria_pattern = criterion_v.get('pattern', None)
+
+        # Render the pattern (it can contain a jinja expressions)
+        try:
+            criteria_pattern = self._render_criteria_pattern(criteria_pattern=criteria_pattern)
+        except Exception:
+            LOG.exception('Failed to render pattern value for key %s' % (criterion_k),
```
manasdk: include the pattern as well. Easier for debugging.

st2reactor/st2reactor/rules/filter.py
```diff
@@ -85,19 +109,33 @@ def _check_criterion(self, criterion_k, criterion_v, payload_lookup):
...
+                          extra=self._base_logger_context)
             return False
 
-        criteria_pattern = criterion_v['pattern']
-
         op_func = criteria_operators.get_operator(criteria_operator)
 
         try:
-            return op_func(value=payload_value, criteria_pattern=criteria_pattern)
+            result = op_func(value=payload_value, criteria_pattern=criteria_pattern)
```
manasdk: is there any difference with this and previous impl?

manasdk:
+1

st2reactor/st2reactor/rules/filter.py
Kami: True, I didn't really touch the log level though, I just added additional info to the context.

st2reactor/st2reactor/rules/filter.py
Kami: Nope, just a style thing - same end result / outcome.

st2reactor/st2reactor/rules/filter.py
Kami: Agreed. Done in ea3f6e4b507494730c7c1be65ac66b8576441703.

PR #1313, opened 2015-03-30
Fix default on quickstart test and add bootstrap script for examples and tests

DoriftoShoes:
This adds a script to bootstrap the examples or tests packs that are included in contrib.  This will be used by the e2e deployment workflows.


tools/st2-setup-samples.sh
```diff
@@ -0,0 +1,20 @@
...
+
+PACK=$1
+if [[ "$PACK" == "examples" ]]; then
+    PATH="/usr/share/doc/st2/"
+elif [[ "$PACK" == "tests" ]]; then
+    PATH="/usr/share/stackstorm/"
+fi
+
+echo -e "Adding packs_base_paths for ${PATH}/${PACK}..."
+grep packs_base_path /etc/st2/st2.conf || echo -e "\n[content]\npacks_base_paths = ${PATH}/${PACK}" >> /etc/st2/st2.conf'
```
Kami: I know this is not an issue in this case, but there is a double slash between path and pack.

tools/st2-setup-samples.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+PACK=$1
+if [[ "$PACK" == "examples" ]]; then
+    PATH="/usr/share/doc/st2/"
+elif [[ "$PACK" == "tests" ]]; then
+    PATH="/usr/share/stackstorm/"
+fi
```
Kami: Would be good to add else and throw / exit if uses passes in an invalid / unsupported pack name.

```bash
...
else:
  echo "Invalid pack: ${PACK}
  exit 2
fi
...
```

tools/st2-setup-samples.sh
```diff
@@ -0,0 +1,20 @@
...
+
+echo -e "Adding packs_base_paths for ${PATH}/${PACK}..."
+grep packs_base_path /etc/st2/st2.conf || echo -e "\n[content]\npacks_base_paths = ${PATH}/${PACK}" >> /etc/st2/st2.conf'
+
+if [[ $? ne 0 ]]; then
```
Kami: Missing dash (`-ne`)

tools/st2-setup-samples.sh
Kami: Extra single quote at the end of the line :P

DoriftoShoes:
It was late...stop making fun of me :P.  I'm fixing this stuff now and adding virtualenv creation in case we add any python actions.

tools/st2-setup-samples.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
```
dzimine: Suggestion: only install tests pack, and leave installing examples up to the tests, as needed.

Kami:
@DoriftoShoes I know, sorry, nothing personal :P

Kami:
LGTM, +1

PR #1317, opened 2015-03-31
Add Makefile target for cleaning the MonoDB and RabbitMQ instance

Kami:
This comes handy in a local development environment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -130,6 +130,16 @@ clean: .cleanpycs .cleandocs
 	@mongo --eval "rs.initiate()"
 	@sleep 5
 
+.PHONY: .cleanmysql
+.cleanmysql:
+	@echo "==================== cleanmysql ===================="
+	@echo "----- Dropping all Mistral MYSQL databases -----"
+	@mysql -uroot -pStackStorm -e "DROP DATABASE IF EXISTS mistral"
+	@mysql -uroot -pStackStorm -e "CREATE DATABASE mistral"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -119,6 +119,25 @@ clean: .cleanpycs .cleandocs
 	@echo "Removing generated documentation"
 	rm -rf $(DOC_BUILD_DIR)
 
+.PHONY: .cleanmongodb
+.cleanmongodb:
+	@echo "==================== cleanmongodb ===================="
+	@echo "----- Dropping all MongoDB databases -----"
+	@sudo service mongodb force-stop
+	@sudo rm -rf /var/lib/mongodb/*
```
Kami: That's obviously only meant to be used in a local development environment :)

I remove the data dir, because just removing a database, doesn't reclaim the disk space (you need to run a compaction, etc.).

Makefile
```diff
@@ -119,6 +119,25 @@ clean: .cleanpycs .cleandocs
...
+	@sudo chown -R mongodb:mongodb /var/lib/mongodb/
+	@sudo service mongodb start
+	@mongo --eval "rs.initiate()"
```
lakshmi-kannan: We probably don't want this. I don't think we have replication turned on.

lakshmi-kannan:
LGTM

Makefile
Kami: You still need to run initiate after you nuke the db folder, even though we don't use replication (otherwise the server thinks it's a slave).

PR #1321, opened 2015-04-02
Make model validation optional

enykeev:
Model validation is rather expensive operation that is only needed in a limited number of cases (usually, when we add new documents to the system).

Right now, we are revalidating all the actions and executions we return during GET requests. This drastically affects response time (up to several seconds of TTFB) harming usability in a major way.

This patch would require us to explicitly validate every model manually the moment it is created.

The patch resulting in 2x reduction of response time for `limit=50` over the collection of 1214 action execution records and 10x reduction for the whole collection of 363 actions.

### DIFF
--- a/st2actions/st2actions/bootstrap/actionsregistrar.py
+++ b/st2actions/st2actions/bootstrap/actionsregistrar.py
@@ -111,6 +111,7 @@ class ActionsRegistrar(ResourceRegistrar):
                             (pack, pack_field))
 
         action_api = ActionAPI(**content)
+        action_api.validate()
         action_validator.validate_action(action_api)
         model = ActionAPI.to_model(action_api)
 
--- a/st2actions/st2actions/bootstrap/runnersregistrar.py
+++ b/st2actions/st2actions/bootstrap/runnersregistrar.py
@@ -375,6 +375,7 @@ def register_runner_types():
             update = False
 
         runnertype_api = RunnerTypeAPI(**runnertype)
+        runnertype_api.validate()
         runner_type_model = RunnerTypeAPI.to_model(runnertype_api)
 
         if runnertype_db:
--- a/st2common/st2common/models/api/action.py
+++ b/st2common/st2common/models/api/action.py
@@ -13,8 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import jsonschema
-
 from st2common.util import isotime
 from st2common.util import schema as util_schema
 from st2common import log as logging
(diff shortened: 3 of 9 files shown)


manasdk:
I am ok with this change. I do dislike having to remember to call validate immediately after construction but thats ok.

manasdk:
+1

Kami:
Thanks and sorry for the delay - totally forgot about this PR.

I'm general, I'm OK with this change, but I am slightly worried that someone will forget to call validate when we are taking data in. I would be more in favor of defaulting to validate=True and requiring programmer to explicitly specify validate=False in places where we don't want the validation to happen. This way it's safer and harder to make a mistake.

enykeev:
@Kami it seems to me like we need to figure out some general patterns of using this API models to understand how it should work. For now, I found three use cases for them:
- we're getting JSON from user using REST API endpoint. The Model gets created and validated during jsexpose.
- we're loading JSON from file during content registering. The validation currently should be triggered manually.
- we're getting data from DB and converting it (using from_model) to API model to give it to the user in form of JSON. This is the most problematic place since we may end up with huge amounts of data we have already validated once. There should be no validation as we're trusting our DB to contain valid data... or at least that's my line of reasoning.

Am I missed something?

I don't like the idea of implicit validation as it would be always a source of this excessive and unnecessary computations which would be rather hard to spot (partially because they need a big volume of data to become noticeable and it is hard to gather such a volume during routine tests), while an absence of validation could be easily spotted by extensive unit\integration testing (just pass invalid json and check for error) along with shared understanding when API models should and should not been used (we should fight an urge to use API models to just convert db document to json to send it though queue. At least for purely semantical reasons).

Kami:
I've already mentioned this a bit in the past and we have already discussed it a bit, but in short, the main problem really is that we are (ab)using API classes stuff for both, data coming in and data going out.

We should have separate code paths for handling user provided input (validation) and serializing the output. Input is mostly straight-forward, but for output you usually need more flexibility since a lot of times you want to modify or add additional data and not directly return a raw database object - e.g. have different "views" for a particular object / resource (that's also where the current the approach breaks).

We created and used node-swiz (https://github.com/racker/node-swiz) in Cloud Monitoring which worked along those lines.

That would solve our problem (we would always just implicitly perform the validation on input - I do think this needs to be opt-out and not opt-in so it's harder to make a mistake) and it's something we should aim to do long term.

Kami:
To recap - for now, I'm OK with this, but in the near future, I think we should move to the approach proposed above (I basically agree with all your points and approach suggested above should solve all those issues).

PR #1323, opened 2015-04-03
[RFR] [Chatops] ActionAlias and AliasExecution

manasdk:
ActionAlias and AliasExecution are the basis of chatops integration. The models provide a chatops friendly abstraction to Action and ActionExecution. 

* ActionAlias is a way to alias an Action in a chatops friendly way. ActionAlias is stored in the DB and picked up from the file system. It is managed separately from an Action to better work with the user model that an action/pack author may not have the best control over the chatops application and availability of an action. 
* For now the paradigm around ActionAlias are very similar to content delivered via packs.

Full CRUD
```
$ http :9101/exp/actionalias

[
    {
        "action_ref": "core.local",
        "formats": [
            "{{cmd}}"
        ],
        "id": "551cbf0d32ed35294e692119",
        "name": "runlocal"
    }
]
``` 

* AliasExecution is a shim around Actionexecution and is only a way to provide access to all implementation to the back-end of st2 as opposed to performing that in a specific bot etc. 

With the current impl all a client has to do is perform the equivalent of -

```
 echo '{"command": "runlocal date"}' | http POST :9101/exp/aliasexecution
HTTP/1.1 200 OK
Access-Control-Allow-Headers: Content-Type,Authorization,X-Auth-Token
Access-Control-Allow-Origin: http://localhost:3000
Access-Control-Expose-Headers: Content-Type,X-Limit,X-Total-Count
Connection: keep-alive
Content-Length: 26
Date: Thu, 02 Apr 2015 04:02:31 GMT

"551cbf5732ed35292c302c4c"
```

* For now all API end-points are placed under exp. Perhaps these could be under `v2` once we decide that the API is ready for public consumption.

Improvements -
- [ ] Better alias definitions. Current aliasing is too similar to the DB layout.
- [ ] Support to provide aliases in a pack.
- [ ] All aliases can be specified in a single file etc.
- [ ] Flexible alias format to get closer to natural languages. Perhaps even explore some advanced techniques.
- [ ] Availability of an action for chatops implies that an alias exists. We have to refine this idea to make an action available by default etc.

Note : This is more of an opinionated view of chatops implementation by trying to come up with a general concept which is not chatops specific. Implementation-wise I guess where the impl lives is still open for debate. Perhaps it is better for some of these pieces to live in a specific bot etc.


dzimine:
Just one comment re /exp vs /v2.
We have not officially shipped a final, supported, documented API. 
Thus still have quite some lee-way here.

manasdk:
Put it under exp as a temporary measure. I guess your point is v1 is as good as exp.

lakshmi-kannan:
I am not sure why we needed separate models for ActionAlias and ActionAliasExecution. Why can't alias be a field in action model itself? Can we talk about the high level plan here?

manasdk:
Alias need to be outside of Action as they are defined by the administration of the environment and not the Action author. If we go with that use case then alias cannot be a property of an action as we would not want an administrator change Action model. 

Again I created a model since that was easy rather than a file based storage approach.  I do t have to worry about efficient query,  duplicates etc

jfryman:
3085324c-79f8-11e4-9bfe-d860a2bed056

PR #1324, opened 2015-04-03
RFR: Move service tests to unit

lakshmi-kannan:
Before this PR, these tests weren't even run. I moved them to unit folder but they were failing because of a bug. Validation was using default value even if a value was specified and was validating types for the default value and not using the type of the specified value.


Kami:
Good catch.

PR #1325, opened 2015-04-05
Added test_packs_pack

kirilstrax:
Tests almost all actions on packs page, except for:
- st2 run packs.install register=all repo_url=https://github.com/StackStorm/st2incubator.git - installing all packs as part of sanity seems too much
- st2 run packs.restart_component servicename=st2sensorcontainer - I didn't find the way to reliably test that this action does something
- st2 action list --pack=docker, st2 sensor list --pack=docker, st2 trigger list --pack=docker - since they are not parts of packs pack


DoriftoShoes:
+1

PR #1326, opened 2015-04-07
Simplify jsexpose

enykeev:


st2api/tests/unit/controllers/v1/test_runnertypes.py
```diff
@@ -35,7 +35,7 @@ def test_get_all(self):
         self.assertTrue(len(resp.json) > 0, '/v1/runnertypes did not return correct runnertypes.')
 
     def test_get_one_fail(self):
-        resp = self.app.get('/v1/runnertype/1', expect_errors=True)
```
manasdk: lol ... glad this was checking for failure :P

st2common/st2common/hooks.py
```diff
@@ -30,6 +31,18 @@
 
 LOG = logging.getLogger(__name__)
 
+# A list of method names for which we don't want to log the result / response
+RESPONSE_LOGGING_METHOD_NAME_BLACKLIST = [
+    'get_all'
+]
+
+# A list of controller classes for which we don't want to log the result / response
+RESPONSE_LOGGING_CONTROLLER_NAME_BLACKLIST = [
```
manasdk: This is weird. A blacklist of controllers that exist in st2api having literals used in st2api.

Kami:
Hm, looks like my commit comments disappeared. Did you force push?

In any case, as noted in my previous comments, I like this approach, it makes code a lot more readable.

LGTM, +1

manasdk:
How about an opt-in logging approach where the method is explicitly annotated with a `@logme` annotation?

Extra work in the controllers but atleast we avoid weird strings being maintained in a file and control of what to log or not is obvious in the controller itself.

manasdk:
I like the jsexpose simplification and think this is absolutely step in the right direction. So as such am +1.

st2common/st2common/hooks.py
Kami: To clarify it - he just moved this code around. This code already existed before. I added it during the logging changes.

We now log HTTP response for all the endpoints except "list" and those 3 blacklisted endpoints. Those endpoints are blacklisted because they could potentially contain a massive response and we don't want to log that.

manasdk: I know. I did add the blacklist last time either :P. In that case I noted it much later :)

lakshmi-kannan:
This is much better than what was there already :). Me likey <3.

enykeev:
> Hm, looks like my commit comments disappeared. Did you force push?

I did, sorry. Had to rebase to remove validation commit from this PR.

> How about an opt-in logging approach where the method is explicitly annotated with a @logme annotation?

I think we should log every request, even the one we don't have a controller for (currently, we only log the fact this requests have failed, but not the fact they were requested in the first place).

I believe you are mostly worried about having to blacklist some of them in an entirely different part of application. For that, yeah, we can add an additional decorator that would say to the logging hook whether it should or should not log request body\result. So, log meta for everything, log content for selected few. Any strong points on whether it should be opt-in and not opt-out?

manasdk:
I think you should get this change in now.there I'd a good deal of progress here and opt-in opt-out ain't changing things practically right now

enykeev:
Going to merge it today. Just want to check first that it doesn't break st2web for some vague reason.

st2common/st2common/hooks.py
```diff
@@ -137,32 +153,83 @@ def _validate_token(request):
...
         if cfg.CONF.api.serve_webui_files and request_path.startswith('/webui'):
             # We want to return regular error response for requests to /webui
             return
 
-        status_code = state.response.status
-        if status_code == httplib.NOT_FOUND:
-            message = 'The resource could not be found'
-        elif status_code == httplib.INTERNAL_SERVER_ERROR:
-            message = 'Internal Server Error'
+        LOG.debug('API call failed: %s' % (str(e)))
```
manasdk: Without a traceback turns out this log message isn't very useful. I will figure out a fix and create a PR

lakshmi-kannan: Heh.

PR #1327, opened 2015-04-07
[COMMENT] Logrotate Rules

jfryman:
This PR introduces a logrotate file that will eventually be placed on systems to encourage good health and hygiene of logs on a system.

Things I don't love:
* Restarting components: Can't `HUP` currently and let go of file descriptors for old log files. As such, a component restart is in order. But, I don't necessarily want to restart the entire component just to release a log file.
* ActionRunner PID files: This is interesting... old log files from runners will just collect here because two new log files are created for each runner. I'm doing some ops trickery there to cleanup old files, but also ugly.

### DIFF
--- a/conf/logrotate.conf
+++ b/conf/logrotate.conf
@@ -1,0 +1,65 @@
+## Global
+compress
+copytruncate
+notifyempty
+
+## Rules engine
+/var/log/st2/st2rulesengine.audit.log {
+  rotate 5
+  size 10M
+  postrotate
+    st2ctl restart-component rules_engine
+  endscript
+}
+
+/var/log/st2/st2rulesengine.log {
+  size 100M
+  rotate 5
+  postrotate
+    st2ctl restart-component rules_engine
+  endscript
+}
+
+## API
+/var/log/st2/st2api.log {
+  daily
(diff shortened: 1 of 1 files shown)


manasdk:
Whenever I have done this in the past I have had some application library handle log rotation. I like the handle from outside approach but that leads to this file descriptor issue and requires a HUP - a limitation with python logging?

PIDs : not an issue if using syslog. 

I think we should use logrotate only with syslog and have application deal with rotation if no syslog is present.

lakshmi-kannan:
I am pretty sure python logging library allows log rotation. Let me check.

manasdk:
@jfryman @lakshmi-kannan https://github.com/StackStorm/st2/pull/1336

Kami:
There is an issue with action runner ending up in "infinite log rotation and restart loop" I've observed on st2master001:

```
kami [12:02]
i guess i see the issue
kami [12:02]
we restart it
kami [12:02]
after removing old files in post rotate
kami [12:02]
and it's set to keep maximum if 5 old files
kami [12:02]
so if number of actions runners > 5
kami [12:02]
we would end in an infinite rotation loop
```

PR #1330, opened 2015-04-08
Added auth token

DoriftoShoes:
This PR adds the st2 auth token as a parameter to be passed in to the quickstart workflows.


lakshmi-kannan:
LGTM

PR #1336, opened 2015-04-13
[COMMENT] Rotating file handler

manasdk:
Apart from logrotate.conf use RotatingFileHandler.

* logrotate can be used with syslog
* RotatingFileHandler based rotation when no syslog is setup


Kami:
> logrotate can be used with syslog

Just to clarify that - log rotate can be used with any application which supports writing logs to a file and re-opens a log file on SIGUSR1 :)

It would be ideal if we could also use log rotate for our Python services (one place / config to configure all the log rotation settings), but it might be too big of a PITA to do that so I'm fine with this approach.

Kami:
Edit: Looks like something like this should work - https://bitbucket.org/cherrypy/cherrypy/issue/679/log-file-handling-with-standard-tools

manasdk:
Actually the issue I want to work out is with the multiple python processes for actionrunner. Also, for dev setup logrotate might be a better approach anyway.

Kami:
@manasdk We already "manage" multiple action runner processes in st2ctl so sending SIGUSR1 to all the action runner processes should be no different than restarting all the action runner processes.

As far as the dev environment goes - I don't think we really need log rotation there.

manasdk:
@kami When logging via syslog we can use logrotate. When no syslog we can use RotatingFileHandler. This avoid all that restart business entirely.

Anyway, if you guys see this as solved I am ok with abandoning this PR.

Kami:
@manasdk With the SIGUSR1 approach described in the link above we can avoid restarts and also use logrorate for our Python services.

jfryman:
It'd be awesome to have something where I could send a soft kill (HUP/USR1) to release the lock. The `copytruncate` in the logrotate is a direct response to not being able to release the lock on the file, and that could very easily land us in some problems down the road (imagine a full log, and lack of disk. In order to cleanup, I need log-size x2 it make the operation happen before compression can occur, and could end up in a bad spot.

As far as solving this... I'm not sure we have. Maybe closer. :)

Kami:
To clarify it - yes, this does require us adding some code (see the link above) and installing SIGUSR1 handler, but I like this approach since this way our Python services also play nicely with all other tools (user can use arbitrary external tool to handler log rotation) which assume reopening of the log files on SIGUSR1.

manasdk:
About SIGUSR1 approach -

All I care about is no unbounded growth of logfiles. 

I am closing this PR. I do not plan to work on any logrotation stuff at this point i.e. assume that someone else will be taking up SIGUSR1

Kami:
Alright, I've opened an issue so we don't forget to do it - #1342.

Kami:
:+1:

manasdk:
Will work on https://github.com/StackStorm/st2/issues/1342 as part of this PR.

manasdk:
@jfryman This includes logrotate.conf. Can that be merged in as well?

jfryman:
@manasdk yes sir. we'll need to wire it up to the packaging, so getting it merged is :+1:

manasdk:
right I was just wondering about packaging. Ignoring that for now.

PR #1341, opened 2015-04-14
[RFR] Full sample config.

manasdk:
Tool to generate complete config and generate config file.


conf/st2.full.conf
```diff
@@ -0,0 +1,182 @@
...
+serve_webui_files = True
+# StackStorm API server host
+host = 0.0.0.0
+# Send empty message every N seconds to keep connection open
+heartbeat = 25
+# StackStorm API server port
+port = 9101
+
+[api_pecan]
+# None
```
Kami: Hm, seems about to add description for all the arguments which are missing it :)

dzimine:
Good. suggest we call it st2.conf.sample

conf/st2.full.conf
manasdk: I am going to be lazy and skip the entire api_pecan section. Not sure if there is much value in advertising that group.

f78b3836ab7e001618f01d0fae48f8236ca8fd8e

manasdk:
> suggest we call it st2.conf.sample

Fixed in 09b867bae0a5f7e6f1418fa8e7b6690b1c89b914

conf/st2.full.conf
Kami: I'm fine with that.

Kami:
LGTM, +1

Kami:
Perhaps also reference this sample config with all the options somewhere in the docs?

manasdk:
docs updated : 25d736560fa013707e7f8744490af22d50d2882f

Kami:
:+1:

jfryman:
👍

PR #1343, opened 2015-04-14
Re-open log files on SIGUSR1 signal

Kami:
With this pull request all the Python service processes now re-open log files when receiving `SIGUSR1` signal.

To test it start the st2api service and run the commands listed bellow.

Screen 1:

```bash
tail -F logs/st2api.log
```

Screen 2:

```bash
python ./st2client/st2client/shell.py runner list - you should see new messages in the log file
rm logs/st2api.log
python ./st2client/st2client/shell.py runner list - you should see no new messages in the log file since the process is still referencing the old file handle
kill -USR1 <st2api pid>
python ./st2client/st2client/shell.py runner list - you should see new messages in the log file because SIGUSR1 causes process to reopen the log files
```

TODO:

- [x] Do the same for all the Python services
- [x] Update log rotate configs
- [x] Remove rotating file handler code

### DIFF
--- a/st2common/st2common/logging/misc.py
+++ b/st2common/st2common/logging/misc.py
@@ -1,0 +1,44 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import absolute_import
+
+import logging
+
+LOG = logging.getLogger(__name__)
+
+
+def reopen_log_files(handlers):
+    """
+    This method iterates through all of the providers handlers looking for the FileHandler types.
+
+    A lock is acquired, the underlying stream closed, reopened, then the lock is released.
+
+
+    This method should be called when logs are to be rotated by an external process. The simplest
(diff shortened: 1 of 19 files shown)


st2common/st2common/logging/misc.py
```diff
@@ -0,0 +1,44 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
```
Kami: I'm open to suggestions for a better module name.

manasdk:
Do we need to remove RotatingFileHandler?

Kami:
@manasdk The question is if we need it for the dev environment or not. If we do, we then need two different set of logging configs (dev, prod).

Or we could just leave it there, but update the default values so no rotation happens. This way user needs to explicitly enable it if they want to use that instead of logrotate.

manasdk:
@kami - I guess I don't understand why logrotate is better than RotatingFileHandler.

Kami:
@manasdk Alright, let me try to clarify it again.

> e.g. who calls kill -USR1 <st2api pid>

This would be called by an external program which is responsible for rotating logs. In our case, this is logrotate - we just need to update logrotate.conf to send this signal to the service instead of restarting it.

manasdk:
Oh ok. in postrotate - got it.

So still these are 2 parallel choices - yes/no?

Kami:
@manasdk Yeah.

@jfryman added a log rotate config and we will install this config in the package based deployment which means log rotation will already be taken care of in the packaged based deployments.

Kami:
I've pushed a change which updates log rotate config - f19f0580249683915a31de4741bb0b44e02ee995, 07404817b838e67447249edfec7d79c310ee020b

I also pushed a logging config change which disables log rotation by default. I didn't touch the handler so if user for some reasons wants to use Python and not logrotate to rotate the logs for them, they can do that by updating the logging configs (need to add some docs for the whole log rotation stuff).

Kami:
Added some docs about configuring log rotation - 423d998418ac25c87d9d7310e4915c6283eb72c0.

manasdk:
@Kami - Seems like logrotate wins by default. No use arguing over this any longer.

+1

Kami:
Note / reminder for us - we still need to update packaging manifest files to copy over logrotate config on installation.

PR #1345, opened 2015-04-15
Install logrotate config file during package based installations

Kami:


st2common/packaging/debian/postinst
```diff
@@ -2,3 +2,4 @@
 
 chmod 755 /usr/bin/st2ctl
 chmod 755 /usr/lib/python2.7/dist-packages/st2common/bin/st2-setup-tests
+mv /etc/logrotate.d/logrotate.conf /etc/logrotate.d/st2.conf
```
Kami: This is a bit nasty, but you can't rename files when using dh_install.

I'm open to suggestions. One simply solution would be to do something like "mkdir -p logrotate.d && cp ../conf/logrotate.conf logrotate.d/st2.conf" inside the Makefile and then copy the file directly without renaming it in postinst.

DoriftoShoes: I like adding it to the makefile...keeps it consistent with the other conf files.

Kami: Alright, pushed a change - 2ccc14dbbbd71f4197dc1139ec9b624afa738e4d

PR #1346, opened 2015-04-16
[WIP] Add file_uri to models and some model adjustments

manasdk:


st2common/st2common/models/db/action.py
```diff
@@ -215,6 +209,9 @@ class ActionAliasDB(stormbase.StormBaseDB):
     formats = me.ListField(
         field=me.StringField(),
         help_text='Possible parameter formats that an alias supports.')
+    file_uri = me.StringField(
```
Kami: Hm, this field with the same constructor arguments is already declared on `stormbase.StormContentDB` so it shouldn't be needed here, right?

st2common/st2common/models/db/stormbase.py
```diff
@@ -78,6 +78,23 @@ class StormBaseDB(StormFoundationDB):
...
 
 
+class StormContentDB(StormBaseDB):
+    """Abstraction for a user content model from a pack."""
+    name = me.StringField(required=True)
+    pack = pack = me.StringField(
+        required=False,
+        help_text='Name of the content pack.',
+        unique_with='name')
+    file_uri = me.StringField(
```
Kami: Can we use `artifact_uri` instead of `file_uri` to be consistent with the sensor model?

st2common/st2common/models/db/action.py
manasdk: ActionAlias does not have a pack yet. Not sure if it should etc. so left that out for now and not inheriting from `stormbase.StormContentDB`

st2common/st2common/models/db/stormbase.py
manasdk: Per https://github.com/StackStorm/st2/blob/master/st2reactor/st2reactor/container/utils.py#L212 `artifact_uri` points to the sensor code file on disk and not location of the metadata file. Therefore, using the same name would not work. Even for senortypedb 2 different properties are required.

manasdk: On a separate note `entry_point` and `atrifact_uri` are both in SensorTypeDB. AFAICT `entry_point` here is the same as `entry_point` in ActionDB - so why do we need 2 properties?

manasdk: I read through some more code and see the benefit of `entry_point` https://github.com/StackStorm/st2/blob/master/st2reactor/st2reactor/container/utils.py#L203. 

I guess we will have entry_point, artifact_uri and file_uri for SensorTypeDB. Other will gave file_uri.

I am open to renaming of file_uri to anything else more suitable.

st2common/st2common/models/db/action.py
Kami: Ah, that's actually an action alias and not action model. 

On a related note, `RuleDB` also doesn't have a pack attribute, but we still inherit from `StormContentDB` it looks like :)

manasdk: Oh thats coz we are adding a pack attribute to RuleDB. Should've been clearer. Need to finish up remaining changes ofcourse.

st2common/st2common/models/db/stormbase.py
lakshmi-kannan: Looking at this action models need to use artifact_uri for the python script or the yaml in workflows. 

meta_uri seems better than file_uri

lakshmi-kannan:
https://github.com/StackStorm/st2/pull/1347 depends on this PR. We need to improve the content models. It's now very inconsistent. DB/API models are not consistent with file models. I can pick up tasks here. Let's get a TODO in the description ;).

st2common/st2common/models/db/stormbase.py
manasdk: meta_uri works for me. Don't want to go around renaming entry_point to artifact_uri :P

manasdk:
How the heck did we get to this point? We have issues with our models already!!

Kami:
> #1347 depends on this PR. We need to improve the content models. It's now very inconsistent. DB/API models are not consistent with file models. I can pick up tasks here. Let's get a TODO in the description ;).

That's also related to something @manasdk has mentioned on Slack. I don't agree that the ModelAPI object need to look exactly the same as the metadata file on disk. That's also related to the current ModelAPI (ab)use.

If we are talking about this, we should do it right (we already discussed this many times). We really need three things:

1. Database representation of an object (model)
2. A way to validate the user input and convert it to the model (validation & conversion)
3. A way to format and serialize the model for user consumption (output serialization and formatting)

Currently, we kinda have 1 & 2, but we also need 3 (and a better way to handle 2). Input and output don't always need to (or should) look the same. There are many cases where we want to present / format the object differently (e.g. depending on the media format, API endpoint, requested serialization format, etc.) and currently we handle this in a hacky manner.

We should do a similar thing we did with node-swiz in cloud monitoring - http://www.rackspace.com/blog/node-swiz-node-js-library-for-serializing-deserializing-and-validating-objects-in-rest-apis/

Ideally, we shouldn't reinvent the wheel. I'm sure for handling that already exists in Python.

manasdk:
I am almost convinced we should abandon this change. Since the chances we will tackle Content management soonish are high I suspect this or some other way will be needed to map between model and file system artifact. Without something leveraging this property it is anyway half baked so no point in continuing in this direction.

PR #1348, opened 2015-04-16
uwsgi support

enykeev:
You can still run it as before by using cmd.py or you can serve it using uWSGI:
```
uwsgi --http :9101 --module 'st2api.wsgi:setup()' --pyargv '--config-file conf/st2.dev.conf'
```

In case of uWSGI, stream won't work due to:
a) uWSGI doesn't support eventlet, only gevent
b) uWSGI expects `application(environ, start_response)` to be a generator itself rather than return generator as Pecan does it. I've spent the whole day digging though Pecan code to figure out how to change it behavior accordingly and wasn't able to find another way except for making stream a middleware (which I also haven't tested yet).
c) every doc in uwsgi at least slightly related to streaming or sse insist on using uWSGI off-loading to avoid resources (processes and threads) being fully consumed by long-living sse connections. Common sense agrees, I think we need to remove streams from API and make them a separate service. Whether we're going to off-load connections to it or make it a separate eventlet\gevent app on its own port, we'll see, but it is clear now that streams have very different performance requirements and mixing short-living API calls and long-living Stream connections is not we want to do.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -23,3 +23,4 @@ passlib>=1.6.2,<1.7
 lockfile>=0.10.2,<0.11
 python-gnupg>=0.3.7,<0.4
 jsonpath-rw>=1.3.0
+uwsgi>=2.0
--- a/st2api/st2api/cmd/api.py
+++ b/st2api/st2api/cmd/api.py
@@ -21,14 +21,9 @@ from oslo.config import cfg
 from eventlet import wsgi
 
 from st2common import log as logging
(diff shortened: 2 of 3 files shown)


lakshmi-kannan:
@enykeev Is this still valid PR?

enykeev:
We still need to do it eventually, though, I'm not working on it at the moment. Also, is has no real use until we extract the stream as a separate process. Closing.

PR #1349, opened 2015-04-17
update st2ctl and prod config to get st2notifier to run st2notifier

manasdk:
* Validated on a deployed system that this st2ctl actually works.


lakshmi-kannan:
LTMG.

PR #1350, opened 2015-04-17
Removing incorrect comments from config.js

DoriftoShoes:


Kami:
@DoriftoShoes sorry if I missed the discussion - how did we determine that those are incorrect / obsolete? Did the webui code change or my original comment was wrong?

I added those comments when I talked with @enykeev about how the config stuff works. I added the comments because the way config parsing works is actually quite confusing (empty string is not the same as null, etc.).

Kami:
Nvm - this was a failure on my part - comment should have started with "//" and not "#".

PR #1353, opened 2015-04-19
Added a note about MySQL -> MariaDB replacement on Fedora

kirilstrax:
See discussion on STORM-1131. Since @dzimine considers "manually install mysql" a proper solution, added note that tells user to do so.


kirilstrax:
Not a good change apparently (does not work).

PR #1360, opened 2015-04-21
Implement action execution scheduler

m4dcoder:
Create a scheduling service that takes the action execution request, performs scheduling logic (placeholder), and dispatch to the action runner(s).  Move the execute_action from the action runner worker to the action service.  Create request, schedule, and execute methods in the action service. Introduced new "requested" status.


st2common/st2common/services/action.py
```diff
@@ -97,11 +102,110 @@ def schedule(liveaction):
...
+
+    # Launch action
+    extra = {'action_execution_db': action_execution_db, 'liveaction_db': liveaction_db}
+    LOG.audit('Launching action execution.', extra=extra)
+
+    # the extra field will not be shown in non-audit logs so temporarily log at info.
+    LOG.info('Dispatched {~}action_execution: %s / {~}live_action: %s with "%s" status.',
+             action_execution_db.id, liveaction_db.id, liveaction.status)
+    try:
+        result = container.dispatch(liveaction_db)
```
manasdk: I somehow get a feeling that this code belongs to st2actions. My line of thought here is that container is an st2actions owned interface. If this were not a dynamic language we would need some sort of explicit import from st2actions and then effectively break the dependency direction which is st2action depends on st2common.

I am curious to know why you choose to move this into st2common.

st2common/st2common/services/action.py
```diff
@@ -97,11 +102,110 @@ def schedule(liveaction):
...
+
+    try:
+        liveaction_db = action_utils.get_liveaction_by_id(liveaction.id)
+    except StackStormDBObjectNotFoundError:
+        LOG.exception('Failed to find liveaction %s in the database.',
+                      liveaction.id)
+        raise
+
+    # stamp liveaction with process_info
+    runner_info = system_info.get_process_info()
```
manasdk: Should move this to the runner that actually executes.

manasdk:
I am a little confused by the change in the publish for liveaction. 

Currently all objects either publish on create, update or delete. For liveaction there is now publish_schedule. schedule is only a state change therefore it should fall under update at the most rather than its own routing key.

In general I am ok to completely changing the publish structure for LiveAction. I would strive to make that internally consistent though,

e.g. we could go with publish on all state transition of LiveAction so
`publish_schedule`, `publish_requested`,`publish_execute` and `publish_finished`. In this scheme we entirely drop support for `publish_{create/update/delete}`. 

I still believe that update can be used as per the original scheme as that keeps it consistent with other object types and lends to some design predictability.

manasdk:
Another slightly orthogonal comment is that I do not see a clear separation between the scheduler and worker forming. Perhaps as you work on the details that will naturally emerge.

We should be able to simply rip out the scheduling and seamlessly move it into a standalone process. This way the actionrunner process is entirely independent. Going forward we will have platform specific actionrunner and at that point mixing scheduling and action running concerns would be an issue. In my view actionrunner should only work on actions that are ready to be executed.

Anyway, I am not saying this does not look like your plan I am simply pointing out that I hope the design ends up progressing in that direction.

st2common/st2common/services/action.py
m4dcoder: I want to move the request, schedule, and execute functions into the action service.  The service is located in st2common.

st2common/st2common/services/action.py
m4dcoder: Sure.

st2common/st2common/services/action.py
manasdk: I disagree but we should take finer details. My sense is that whatever this code does is st2action specific.

m4dcoder:
So on the publish_schedule, I initially have the worker for the action runner listen on "update" instead of "schedule".  This means the kombu consumer for the action runner needs filter on the update routing key.  This creates a lot of unnecessary traffic to the action runner because whenever the record is updated the action runner will receive it and then it has to check the status to see if it should launch the action or not.  It will end up having more noise than work.

m4dcoder:
On the concern whether the scheduler can be separated, the short answer is yes.  The scheduling function is currently placed in the action service.  The only code in the worker right now is setting up the consumer.  There's no scheduling logic there.

Kami:
Just to confirm, this is targeted for 0.10 and not 0.9, right?

manasdk:
> It will end up having more noise than work.

Yes. I can see that which is why I think you should strongly consider a different and independent publishing topology for liveactions. The current design mixes the 2  and makes me nervous. I am in fact ok with a single change leads to multiple publish appraoch as well.

st2common/st2common/services/action.py
m4dcoder: I can move these functions to another module under st2actions. My initial concern is that I do not want these logic in the worker process.  I rather have them in a separate module where it's cleaner, more testable and reusable.

m4dcoder:
Sure. I'll revisit the publishing topology after my 0.9 release tasks. I'm trying to understand what you mean by different publishing topology.  Do you mean don't use the existing publishing strategy in the persistent models?

st2common/st2common/services/action.py
manasdk: container.dispatch is very st2actions specific. I am not sure how/why you would want to encourage reuse.

lakshmi-kannan:
>Just to confirm, this is targeted for 0.10 and not 0.9, right?

Let's do this for 0.10.

manasdk:
>  I'll revisit the publishing topology after my 0.9 release tasks.

Yeah. It was more of a design discussion that I wanted to spark. I have absolutely no expectation that any code is written before we ship 0.9. Heck we can even stop discussing until then if it helps.

m4dcoder:
I don't mind the design discussions. Please continue.

manasdk:
Some concrete suggestions -

* Move `action_service.execute` back to worker. This is code is st2actions specific and I feel it is suited in st2actions. Putting in st2common brings about an expectation that it shareable and it is most likely not. It is sort of like putting sensor_container code in st2common.
* For scheduling purposes instead of using both publish_create and publish_schedule via the same exchange I believe we should split the publishing into 2 separate exchanges. 
    * `st2.liveaction` : This exchange continues to see all create and update
    * `st2.liveaction_schedule` : This exchange sees all the state transitions. We could use proper routing keys to signify state. e.g. `liveaction.requested`, `liveaction.scheduled` ... so on. (note : name of exchange is of course debatable, that is the first name that popped in my head.)

Note that a state change is reflected as an update so there should be a message on the `st2.liveaction` exchange as well which may or may be handled. I think we will use this for audit purposes or client notifications etc.  The `st2.liveaction_schedule` exchange is a purely for scheduling purpose and I expect this will evolve. With this scheme st2actions will mostly only care about the  `st2.liveaction_schedule` exchange.

I pulled the branch and looked closer at the code. As of now the state transitions are not clear, I suspect this is because the names are not intuitive. I do not have better names for states but lets start thinking what the transitions might be. What truly threw me off was `worker.schedule_request` and `worker.execute_request` - in context these names make sense but when looked at holistically I was definitely :confused:.

p.s. : These are suggestions and not required :)

lakshmi-kannan:
I just had a look at the code and I think we need to structure code in a different way. I looked at manas's comments and mostly agree to what he is saying. I was expecting this entire thing to contain two classes (and other minor things). An ActionRequester class in st2common (currently a module) that exposes request method and a ActionMonitor class that handles the state transitions from requested -> pending or requested -> scheduled things like that. I also think we can do better with rabbit exchanges to be less confusing. Let's chat in person. I might be missing some nuances.

m4dcoder:
Still WIP, please feel free to review, but it's not done yet. I still have to move pieces of code back into st2action worker.  Also, I will move the publish_status to the update_liveaction_status method in util.

m4dcoder:
RFR on the scheduler.  I took some feedbacks from the last review.  The st2actions worker, scheduler, notifier, and results tracker can be refactored to share common code (i.e. QueueConsumer). I want to do a checkpoint here though and get some more feedback first.  I didn't take the feedback to implement ActionRequester because currently there's a module in st2common/services for action (previously a schedule method and now is renamed to request). Unless we do away with how we use st2common/services overall, I don't want to handle that differently.  On the ActionMonitor, I recommend to revisit this in another iteration when it's more clear and concrete how we need to monitor and act on various statuses.  Also, the concept of the ActionMonitor seems to overlap with some other roles today (i.e. scheduler acts on requested status and worker acts on scheduled status, etc.).

st2actions/st2actions/cmd/actionrunner.py
```diff
@@ -51,10 +51,18 @@ def _setup():
...
     LOG.info('(PID=%s) Worker started.', os.getpid())
+
+    components = [
+        scheduler.get_scheduler(),
+        worker.get_worker()
+    ]
+
     try:
-        worker.work()
+        [component.start() for component in components]
```
lakshmi-kannan: Not sure you want to use list comprehensions here.

m4dcoder: It works though, at least on my dev box. What's your concern with it?

st2common/st2common/transport/consumers.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def _process_message(self, body):
+        try:
+            self._handler.process(body)
+        except:
+            LOG.exception('%s failed to process message: %s', self.__class__.__name__, body)
+
+
+@six.add_metaclass(abc.ABCMeta)
+class MessageHandler(object):
```
manasdk: Would be useful to have a description of the start + wait protocol. Code is clear so this is a nice to have.

m4dcoder: Absolutely

st2actions/st2actions/cmd/actionrunner.py
manasdk: It create an array for now reason. But more importantly list comprehension is not the right sematic - the goal is to run a for-loop so causes confusion. Likely that some linter might while as there is no assignment.

st2common/st2common/transport/consumers.py
```diff
@@ -0,0 +1,80 @@
...
+from kombu.mixins import ConsumerMixin
+
+from st2common import log as logging
+from st2common.util.greenpooldispatch import BufferedDispatcher
+
+
+LOG = logging.getLogger(__name__)
+
+
+class QueueConsumer(ConsumerMixin):
```
manasdk: +1 to the refactoring.

st2actions/st2actions/scheduler.py
```diff
@@ -0,0 +1,75 @@
...
+
+LOG = logging.getLogger(__name__)
+
+ACTIONRUNNER_REQUEST_Q = liveaction.get_status_management_queue(
+    'st2.actionrunner.req', routing_key=action_constants.LIVEACTION_STATUS_REQUESTED)
+
+
+class ActionExecutionScheduler(consumers.MessageHandler):
+    def process(self, request):
+        if request.status != action_constants.LIVEACTION_STATUS_REQUESTED:
```
lakshmi-kannan: One thing we want to do (in another PR) is to have these classes define the `type` of request object and do validation.

st2common/st2common/transport/consumers.py
m4dcoder: whew.  :dancers:

st2actions/st2actions/cmd/actionrunner.py
m4dcoder: Ok.  I'll redo that in simply for loop.

st2actions/st2actions/cmd/actionrunner.py
```diff
@@ -51,10 +51,18 @@ def _setup():
...
+        [component.wait() for component in components]
     except (KeyboardInterrupt, SystemExit):
         LOG.info('(PID=%s) Worker stopped.', os.getpid())
+        [component.shutdown() for component in components]
```
manasdk: In case a one of the wait explodes we may not call shutdown on other components. Sort of another risk with the comprehension.

m4dcoder: Good catch.  I'll add for loop + try catch on the shutdown.

lakshmi-kannan: Exactly.

st2actions/st2actions/cmd/st2notifier.py
```diff
@@ -49,6 +49,7 @@ def _run_worker():
     actions_notifier = notifier.get_notifier()
     try:
         actions_notifier.start()
```
manasdk: suggestion : can add a single start_and_wait() function in MessageHandler. Simplifies this code a little bit.

st2common/st2common/transport/consumers.py
lakshmi-kannan: I like the refactoring myself but looking at the entire PR, this could have been a PR by itself. It's hard to focus on the requestor and scheduler changes. It's hard to be diligent about this but something to keep in the back of the mind.

m4dcoder: Yeah.  It was obvious from the changes so while waiting for review, I refactored it anyway...

manasdk: I like the refactor. But it is messing with reviewing. I would have done it as a separate PR.

manasdk:
Wish I could review without https://github.com/StackStorm/st2/commit/f74b2318a2fce47f1da4c0dc65c53db8e32ae9d1 commit. it is a good commit but so much noise :page_with_curl:

st2actions/st2actions/scheduler.py
```diff
@@ -0,0 +1,75 @@
...
+            return
+
+        try:
+            liveaction_db = action_utils.get_liveaction_by_id(request.id)
+        except StackStormDBObjectNotFoundError:
+            LOG.exception('Failed to find liveaction %s in the database.', request.id)
+            raise
+
+        # stamp liveaction with process_info
+        runner_info = system_info.get_process_info()
```
manasdk: this cannot be here ... the runner that actually handles the execution i.e. scheduled -> running should add this value.

m4dcoder: Alright.  Thanks for clarifying.  I'll remove this.

st2actions/st2actions/scheduler.py
```diff
@@ -0,0 +1,75 @@
...
+
+        # Update liveaction status to "scheduled"
+        liveaction_db = action_utils.update_liveaction_status(
```
manasdk: There is almost no 'real work' between requested -> scheduled. That is a separate PR?

m4dcoder: Yes. There will be real work in the next set of PR(s) when we evaluation of policies here.

manasdk: Perfect. Works for me.

st2common/st2common/util/action_db.py
```diff
@@ -163,6 +165,10 @@ def update_liveaction_status(status=None, result=None, context=None, end_timesta
 
     LOG.debug('Updated status for LiveAction object: %s', liveaction_db)
 
+    if publish and status != old_status:
+        LiveAction.publish_status(liveaction_db)
```
manasdk: usually when publishing changes to properties. It is better to publish both old and new.

I think there is room for generalization here i.e. instead of only dealing with state change you can have a property_change exchange. And publish property changes .. At this point you need to publish ..

{
'objectid': 'skdfhd'
'property_name': 'foo'
'old': x
'new': y
}

That would be a nice generalization. We don't have to tackle it now - just something to think about.

manasdk:
Enough said. This works for me although I have not looked at the tests. This is a good base to build on in my view.

+1

lakshmi-kannan:
Yeah, I think we can make small incremental changes from here. If tests pass and things work on devenv, go for the merge.

m4dcoder:
Thanks guys for the quick response. I'll make those minor changes from the comments before merging. I know you guys are all busy. Thanks for making the time here.

m4dcoder:
I addressed most of the comments in the PR.

st2actions/st2actions/cmd/actionrunner.py
```diff
@@ -51,12 +51,35 @@ def _setup():
...
+            try:
+                component.shutdown()
+            except:
+                LOG.exception('Unable to shutdown %s.', component.__class__.__name__)
+                errors = True
+
+        if errors:
+            return 1
     except:
         return 1
```
Kami: Should also log an exception here.

st2actions/st2actions/worker.py
```diff
@@ -14,93 +14,62 @@
...
-        # Note: We only want to execute actions which haven't completed yet
-        if liveaction.status == LIVEACTION_STATUS_CANCELED:
-            LOG.info('Not executing liveaction %s. User canceled execution.', liveaction.id)
-            if not liveaction.result:
-                update_liveaction_status(status=LIVEACTION_STATUS_CANCELED,
-                                         result={'message': 'Action execution canceled by user.'},
-                                         liveaction_id=liveaction.id)
-            return
+    def process(self, liveaction):
+        if liveaction.status != action_constants.LIVEACTION_STATUS_SCHEDULED:
```
Kami: Would be good to add a doctring to the class which explain what it does - processes actions which haven't finished yet and for canceled actions without a result it sets a result object...

st2actions/st2actions/worker.py
```diff
@@ -14,93 +14,62 @@
...
+            LOG.info('%s is not executing %s (id=%s) with "%s" status.',
+                     self.__class__.__name__, type(liveaction), liveaction.id, liveaction.status)
+
+            if (not liveaction.result and
```
Kami: Personally, I would find this easier to read and understand if we had 2 "top level" ifs:

1. One which updates result for canceled actions without a result and returns
2. One which returns for all the actions which have already finished

st2actions/tests/unit/test_mistral_v2.py
```diff
@@ -146,14 +145,30 @@
 
 
 def process_create(payload):
```
Kami: Looks like the same set of functions is used in multiple test files.

Maybe we could refactor it in some common mock class - e.g. `MockCUDPublic` which implements those methods...

st2actions/tests/unit/test_notifier.py
```diff
@@ -79,6 +79,5 @@ def test_notify_triggers(self):
         liveaction.start_timestamp = datetime.datetime.utcnow()
 
         dispatcher = NotifierTestCase.MockDispatcher(self)
-        notifier = Notifier(q_connection=None,
-                            trigger_dispatcher=dispatcher)
-        notifier.handle_action_complete(liveaction)
+        notifier = Notifier(None, [], trigger_dispatcher=dispatcher)
```
Kami: Minor thing - please use keyword arguments where possible (`Notifier(connection=None, queues=[], ...`). It makes code easier to read and bugs easier to spot.

st2client/st2client/commands/action.py
```diff
@@ -37,6 +37,8 @@
 
 LOG = logging.getLogger(__name__)
 
+LIVEACTION_STATUS_REQUESTED = 'requested'
+LIVEACTION_STATUS_DELAYED = 'delayed'
```
Kami: Is this status used anywhere? I quickly grepped for it and couldn't find it.

st2common/st2common/persistence/base.py
```diff
@@ -144,3 +144,12 @@ def _get_by_object(cls, object):
         name = getattr(object, 'name', '')
         pack = getattr(object, 'pack', '')
         return cls.get_by_ref(ResourceReference.to_string_reference(pack=pack, name=name))
+
+
+class StatusBasedResource(Access):
+
```
Kami: Should class level docstring which explains what class does would be nice.

st2reactor/st2reactor/cmd/rulesengine.py
```diff
@@ -49,26 +49,40 @@ def _teardown():
...
+
+    return 0
+
+
+def main():
+    try:
+        _setup()
+        return _run_worker()
+    except SystemExit as exit_code:
+        sys.exit(exit_code)
```
Kami: We will only exit with the provided status code on `SystemExit` exception.

Ideally, we would follow the same convention for all the service entry points - return exit code from the `main()` function and update `bin/script` to do something like:

```python
if __name__ == '__main__':
    status_code = rulesengine.main()
    sys.exit(status_code)
```

m4dcoder: I'll leave this for another PR as I plan to refactor the launch scripts.

Kami: Sounds good.

m4dcoder:
Alright, addressed all comments from @Kami.

m4dcoder:
I will be merging this after travis completes.

PR #1362, opened 2015-04-22
Start serving st2web with SimpleHTTPServer instead of st2api

enykeev:


enykeev:
I've missed a few places in docs where I should have mentioned we now serve it from the root at 80 port. Also, stream is not working once again. I'm going to fix that tomorrow.

tools/st2ctl
```diff
@@ -122,6 +122,7 @@ function st2start(){
   else
     service mistral start
   fi
+  (cd /opt/stackstorm/static/webui && nohup python -m SimpleHTTPServer 80) &>> ${LOGFILE} &
```
enykeev: @DoriftoShoes I'm pretty sure you have some strong opinions on this part... It looks to me like a best place to put this kind of stuff, though you might have better idea. Also, we need to figure out a way to disable this thing on production so it doesn't reserve 80 port.

conf/st2.dev.conf
```diff
@@ -5,7 +5,6 @@
 host = 0.0.0.0
 port = 9101
 logging = st2api/conf/logging.conf
-serve_webui_files = True
 # allow_origin is required for handling CORS in st2 web UI.
 # allow_origin = http://myhost1.example.com:3000,http://myhost2.example.com:3000
```
enykeev: CORS is another problem I haven't solved yet. Seems like the only way is to set it to * in st2_deploy and call it a day. On production, this thing would be handled by nginx anyway and I'm starting to have some doubts it's useful in the first place (we need to consult with security. For now, I can picture a couple of ways how to neutralize CORS during phishing attempt).

DoriftoShoes: I think we just set allow_origin = * during st2_deploy install and call it a day.  st2_deploy shouldn't be used for production deployments (they should use apache or nginx anyway, as you mentioned) so it should be fine, I think.

tools/st2ctl
DoriftoShoes: This is fine for now.  Honestly, we just need to deprecate this damn script and move to upstart/init/etc

Kami: In any case, I think it would be better to use some other port than 80 to avoid potential conflict with an existing web server running on that port.

conf/st2.dev.conf
Kami: Yeah, I agree with @DoriftoShoes on this one.

As an extra precaution, we could print a "don't use this in prod" warning during st2api startup if `allow_origin = *`.

tools/st2ctl
enykeev: And yet we still run st2ctl with sudo every time (probably because of mistral service, not sure).

I would very much prefer us to run it on 80 port by default, but we can always provide env variable to change it. Deal?

conf/st2.dev.conf
enykeev: In production, we probably want nginx to handle CORS. At least he knows what our hostname is beforehand. Ok, I'll add awk replace in st2_deploy, we'll figure out the rest later.

enykeev:
I've dealt with the rest, but stream is still broken and at the moment, I don't see another way except to specify publicly accessible hostname in config.js. The behavior for relative urls with only port specified is not defined in RFC, it works for http client, but doesn't work for Eventsource and I don't see what I can do about it.

Kami:
`tools/launchdev.sh` also needs to be updated.

On a related note - I still think we should run it on non-privileged port, at least in the launchdev script, since none of the services there require root.

In this case, it's probably better if we also run it on a non-privileged port in st2ctl, since running it on a different port in st2ctl and launchdev.sh would be confusing and would require additional dev environment specific documentation.

tools/st2_deploy.sh
```diff
@@ -521,9 +523,6 @@ else
   echo "  st2 is installed and ready to use."
 fi
 
-if [ ${INSTALL_WEBUI} == "1" ]; then
-  echo "  WebUI at http://`hostname`:9101/webui/"
-fi
 echo "=========================================="
```
dzimine: Why is this gone?

dzimine:
> In this case, it's probably better if we also run it on a non-privileged port in st2ctl, 
Let's run on privileged port. Consider this DECIDED

Also, @DoriftoShoes what is the other place where we show how to connect to web ui when we run in st2_express and/or docker? Needs to be updated, too.

enykeev:
> In this case, it's probably better if we also run it on a non-privileged port in st2ctl,
> Let's run on privileged port. Consider this DECIDED

So what exactly have we decided once again?

tools/st2_deploy.sh
enykeev: Because now it is the part of st2ctl. See https://github.com/StackStorm/st2/pull/1362/files#diff-0c83a67f0358e42d5815eef73dad98adR319

enykeev:
> tools/launchdev.sh also needs to be updated.
Nope, I don't think so. We're using launchdev on devenv type of environment and we do not install webui on it by default. My view is that if you want to use UI on devenv, you need to manually clone st2web and use gulp.

enykeev:
Btw, I've hacked around stream issue that prevented us from defining port without host in config (it will be available as soon as StackStorm/st2web#137 will be merged). No more changes on my part besides the ongoing discussions.

dzimine:
> Let's run on *NON*-privileged port. Consider this DECIDED
what a typo! 
Pick a port other than 80 and run webui there (9191)?  provide way to change it (env variable would do).

enykeev:
>  Pick a port other than 80 and run webui there (9191)? provide way to change it (env variable would do).

there is already an env variable for that and you can change its default value here: https://github.com/StackStorm/st2/pull/1362/files#diff-0c83a67f0358e42d5815eef73dad98adR18

Kami:
We discussed this and decided to run it on a non-privileged port (8080) so it works without any changes when running as a non-root user (e.g. with a dev environment).

We are bit in a rush so I will go ahead, make the necessary changes and merge this into master. Thanks!

Kami:
Related PRs:

*https://github.com/StackStorm/st2workroom/pull/22
* https://github.com/StackStorm/st2express/pull/42

PR #1363, opened 2015-04-22
Chatops alias updates

manasdk:
- No need to change the Action metadata to inclued channel and message.
- User is also pass through the notifications system.

- [x] pass user and channel from bot to notification
- [x] fix alias location to be inside a pack
- [x] fix tests
- [x] update rules in st2incubator/hubot [https://github.com/StackStorm/st2incubator/pull/190]
- [x] update chatops.md/some other instructions [https://github.com/StackStorm/blogs/pull/4]

### DIFF
--- a/st2api/st2api/controllers/exp/aliasexecution.py
+++ b/st2api/st2api/controllers/exp/aliasexecution.py
@@ -22,7 +22,7 @@ from pecan import rest
 from st2common import log as logging
 from st2common.models.api.base import jsexpose
 from st2common.models.api.action import AliasExecutionAPI
-from st2common.models.db.action import LiveActionDB
+from st2common.models.db.action import LiveActionDB, NotificationSchema, NotificationSubSchema
 from st2common.models.utils import action_alias_utils, action_param_utils
 from st2common.persistence.action import ActionAlias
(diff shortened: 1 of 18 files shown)


st2api/st2api/controllers/exp/aliasexecution.py
```diff
@@ -70,13 +71,21 @@ def _extract_parameters(self, action_alias_db, param_stream):
                                                             param_stream=param_stream)
         return parser.get_extracted_param_value()
 
-    def _schedule_execution(self, action_alias_db, params):
+    def _get_notify_field(self, payload):
```
lakshmi-kannan: Something like this may be needed in executions controller too.

lakshmi-kannan: Having the user in payload may not work with executions. There we might tap into x-requester-user or some other x header to get this info. Just calling it out so we think about that.

manasdk:
@jfryman See 7a08fd2e00bcca2c72945439278c7a505bffc018

Open to alternate names ofcourse. I just did not want to introduce a chatops-ism in AliasExecution.

PR #1367, opened 2015-04-22
0.10.0dev - RFR - Rule should be part of packs (similar to actions and sensors) 

lakshmi-kannan:
## What?

Previously, rules did not belong in a pack. So we were getting rules by name. 

Now, rules belong to a pack and therefore we need to use 'ref'. 

## TODO

- [X] - rulesregistrar fixes
- [x] - CHANGELOG
- [x] - Docs
- [X] - CLI
- [X] - WebUI - https://github.com/StackStorm/st2web/pull/138/files cc: @enykeev 

## CLI

```
(virtualenv)~/st2 git:rule_pack ❯❯❯ st2 rule list                                                                                                                                                                                                        ✭ ◼
+----------------------------------+----------+-------------------------+-------------+
| ref                              | pack     | name                    | description |
+----------------------------------+----------+-------------------------+-------------+
| examples.sample_notify_slack     | examples | sample_notify_slack     |             |
| examples.sample_on_actiontrigger | examples | sample_on_actiontrigger |             |
| examples.sample_with_timer       | examples | sample_with_timer       |             |
| examples.webhook_file            | examples | webhook_file            |             |
+----------------------------------+----------+-------------------------+-------------+
(virtualenv)~/st2 git:rule_pack ❯❯❯ st2 rule get examples.sample_notify_slack                                                                                                                                                                            ✭ ◼
+----------+------------------------------------------------+
| Property | Value                                          |
+----------+------------------------------------------------+
| id       | 55395c0b6d0aff6b8ed87fe5                       |
| ref      | examples.sample_notify_slack                   |
| pack     | examples                                       |
| name     | sample_notify_slack                            |
| enabled  | True                                           |
| action   | {                                              |
|          |     "ref": "slack.post_message",               |
|          |     "parameters": {                            |
|          |         "message": "{{trigger.message}}"       |
|          |     }                                          |
|          | }                                              |
| criteria | {                                              |
|          |     "trigger.channel": {                       |
|          |         "pattern": "slack",                    |
|          |         "type": "equals"                       |
|          |     }                                          |
|          | }                                              |
| tags     |                                                |
| trigger  | {                                              |
|          |     "type": "core.st2.generic.notifytrigger",  |
|          |     "parameters": {},                          |
|          |     "pack": "core"                             |
|          | }                                              |
+----------+------------------------------------------------+
(virtualenv)~/st2 git:rule_pack ❯❯❯
```


st2api/st2api/controllers/v1/rules.py
```diff
@@ -69,6 +60,8 @@ def post(self, rule):
                 POST /rules/
         """
         try:
+            if not hasattr(rule, 'pack'):
+                setattr(rule, 'pack', DEFAULT_PACK_NAME)
```
Kami: Just thinking out loud and wondering if that's the right approach.

Kami:
I assume, this also required related webui changes, right?

Kami:
And probably also need to update CLI to also display ref.

lakshmi-kannan:
Good points about the webui and CLI. Will note it in the TODO.

st2api/st2api/controllers/v1/rules.py
lakshmi-kannan: We do this for actions as well. We can revisit this though.

manasdk:
so this is gonna break backwards compatibility?

lakshmi-kannan:
>so this is gonna break backwards compatibility?

Unfortunately, yes. We can provide a migration script that deletes all rule models and reloads them from packs with the pack info.

lakshmi-kannan:
@Kami @manasdk Let's decide about the backward compatibility part. This PR has been open for too long :P.

manasdk:
Can we some how preserve backwards compatiblity? Say assume pack is `default` if not provided.

manasdk:
also existing rules in DB can get new property filled in as a migrations script?

lakshmi-kannan:
>Can we some how preserve backwards compatiblity? Say assume pack is default if not provided.

Since rules would switch to using ResourceReference, backward compatibility would mean do a get_by_name (Which requires a data model that has name set to unique) and if that fails do a get by ref (Data model needs to be unique on ref). So to support both, we'd need to have both the data models somehow. It will add lot more complication to code.

manasdk:
All we need is to make sure that a default value is assumed? Why woudl we need to keep both model. I am thinking about making sure we go in an update all existing rule in an st2 install at upgrade.

Also, a rule which does not have pack can be assigned a value automatically.

lakshmi-kannan:
We decided we'll run a pre-upgrade migration script that by passes mongoengine validation and writes to mongo directly with pack: default for all pre-existing rules, upgrade the bits and things should be fine.

manasdk:
+1

PR #1369, opened 2015-04-23
Allow user to skip mistral installation when using st2_deploy

Kami:
The change is fully backward compatible since we default to installing mistral by default, but user now has a chance to opt-out.

### DIFF
--- a/tools/st2_deploy.sh
+++ b/tools/st2_deploy.sh
@@ -60,6 +60,7 @@ echo "Installing version ${VER}"
 
 INSTALL_ST2CLIENT=${INSTALL_ST2CLIENT:-1}
 INSTALL_WEBUI=${INSTALL_WEBUI:-1}
+INSTALL_MISTRAL=${INSTALL_MISTRAL:-1}
 
 # Determine which mistral version to use
 if version_ge $VER "0.9"; then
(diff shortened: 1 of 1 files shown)


tools/st2_deploy.sh
```diff
@@ -14,6 +14,11 @@ EOM
 
 WARNING_SLEEP_DELAY=5
 
+# Options which can be provied by the user via env variables
+INSTALL_ST2CLIENT=${INSTALL_ST2CLIENT:-1}
```
lakshmi-kannan: I think the default value should be 1 for st2client.

lakshmi-kannan:
LGTM. @DoriftoShoes ?

tools/st2_deploy.sh
DoriftoShoes: That is 1...you silly, no bash understanding person.

jfryman: http://www.network-theory.co.uk/docs/bashref/ShellParameterExpansion.html

lakshmi-kannan: LOL. Really? WTH.

DoriftoShoes:
LGTM

PR #1372, opened 2015-04-24
Added st2_bootstrap

DoriftoShoes:
This PR adds the script that lives on the download servers for st2_deploy.sh deployments.

### DIFF
--- a/tools/st2_bootstrap.sh
+++ b/tools/st2_bootstrap.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+BASE_URL="https://downloads.stackstorm.net/releases/st2"
+BOOTSTRAP_FILE="/tmp/st2_boostrap.sh"
+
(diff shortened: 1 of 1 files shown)


jfryman:
I don't quite understand this script, and the relation to `st2_deploy.sh`. Why the extra layer?

DoriftoShoes:
It let's us have one entry point for every version of the script.  Dz
wanted st2_deploy to live in /scripts/.  In order to not break backwards
compatibility and allow us to have a single entry point I wrote this
wrapper.  It actually sits on the download server as st2_deploy.sh

dzimine:
> Dz wanted st2_deploy to live in /scripts/.
don't say that :) This is 'how', not 'what'.

I am asking for a non confused way for a user to get 1) current stable 2) current latest and 3) an given version. And for us to change deployment scripts on any version without updating others (practically, stable and latest).

manasdk:
+1. I see the point of the wrapper layer, still cannot be sure this is the right way forward. Lets go with this for now.

PR #1375, opened 2015-04-25
Added auth_usage fragment to be used with various pages

kirilstrax:
Since almost each installation page will require similar explanation, I am moving this fragment into separate page, to include into multiple pages


dzimine:
LGTM

PR #1376, opened 2015-04-25
Changed the page to use auth_usage.rst

kirilstrax:
The paragraph about authentication usage is useful for multiple pages; thus moved it to the separate page and including here


dzimine:
LGTM.

PR #1381, opened 2015-04-27
Mariadb

DoriftoShoes:
This PR adds support for MariaDB for Fedora.  The hostname for the rabbitmqadmin curl command was changed to 127.0.0.1 due to an /etc/hosts issue with older versions of Vagrant and chkconfig was used instead of systemctl to avoid errors during provisioning.


Kami:
Nice, :+1:

PR #1382, opened 2015-04-27
Examples installer & self-check v1

DoriftoShoes:
This adds an installs script for the examples pack and fixes a path issue on the tests install script.  A self-check script was also added.  The self-check script runs all of the tests in the tests pack.

Sample output of self-check:

```
root@st2stage201:/home/phool# ./self-check 
Attempting Test tests.test_packs_pack...OK!
Attempting Test tests.test_quickstart...OK!
Attempting Test tests.test_quickstart_key...OK!
Attempting Test tests.test_quickstart_rules...OK!
SUCCESS: All tests passed!
```

### DIFF
--- a/tools/self-check
+++ b/tools/self-check
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+## Script to install samples, tests, and run all workflows in tests pack.
+
 USER='testu'
 PASS='testp'
 
--- a/tools/self-check
+++ b/tools/self-check
@@ -2,9 +2,9 @@
 
 ## Script to install samples, tests, and run all workflows in tests pack.
 
-USER='testu'
-PASS='testp'
+PACKS="tests examples"
 
+#Determine Distro
 DEBTEST=`lsb_release -a 2> /dev/null | grep Distributor | awk '{print $3}'`
 
 if [[ "$DEBTEST" == "Ubuntu" ]]; then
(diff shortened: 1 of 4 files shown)


dzimine:
LGTM! 
Style hint - one liner of what script doing in the first line - just a notch better than a name.

tools/self-check
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+
+USER='testu'
+PASS='testp'
```
dzimine: pass as args? with defaults?

tools/self-check
```diff
@@ -0,0 +1,43 @@
...
+
+INSTALL=`${PYTHONPACK}/st2common/bin/st2-setup-examples`
+
+EXITCODE=$?
+echo $INSTALL
+
+ST2_AUTH_TOKEN=`st2 auth ${USER} -p ${PASS} | grep token | awk '{ print $4 }'`
+export ST2_AUTH_TOKEN=${ST2_AUTH_TOKEN}
+
+TEST_ACTION_LIST=`st2 action list --pack=tests | awk '{ print $2 }' | grep -v "|" | grep -v "ref"`
```
dzimine: I would bring this out to a separate script. This way I can install tests and examples once, and run them many times, when needed (assume already authenticated shell).

DoriftoShoes: I don't mind assuming the token exists in the env already.  That way I can just rip out the whole bit with the user/pass.  I'll keep the install bits in but wrap them in some logic to see if they are already installed.

dzimine:
Isn't it lovely? LGTM!

PR #1386, opened 2015-04-27
Install Window runner dependencies in st2_deploy

Kami:
TODO:

- [x] Update st2_deploy to add StackStorm repos


Kami:
Tested it on a fresh Ubuntu 14.04 and Fedora 20 host. It works fine on both.

There is an issue on Fedora 21 (#1390), but that's unrelated to this change and the problem already existed before.

PR #1391, opened 2015-04-27
RFR: Add linux pack requirements to st2 requirements

lakshmi-kannan:
Since `linux` is now a system pack, we need it's requirements to be part of system requirements. This is definitely a work around for now. We need to come up with a better way to ship packs with st2.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -23,3 +23,9 @@ passlib>=1.6.2,<1.7
 lockfile>=0.10.2,<0.11
 python-gnupg>=0.3.7,<0.4
 jsonpath-rw>=1.3.0
+# Requirements for linux pack
+# used by file watcher sensor
+pyinotify>=0.9.5,<=0.10
+-e git+https://github.com/Kami/logshipper.git@stackstorm_patched#egg=logshipper
+# used by nmap actions
+python-nmap>=0.3.4,<0.4


DoriftoShoes:
+1

requirements.txt
```diff
@@ -23,3 +23,9 @@ passlib>=1.6.2,<1.7
 lockfile>=0.10.2,<0.11
 python-gnupg>=0.3.7,<0.4
 jsonpath-rw>=1.3.0
+# Requirements for linux pack
+# used by file watcher sensor
+pyinotify>=0.9.5,<=0.10
+-e git+https://github.com/Kami/logshipper.git@stackstorm_patched#egg=logshipper
```
lakshmi-kannan: @Kami's job is so secure.

Kami:
LGTM, +1

requirements.txt
```diff
@@ -23,3 +23,9 @@ passlib>=1.6.2,<1.7
 lockfile>=0.10.2,<0.11
 python-gnupg>=0.3.7,<0.4
 jsonpath-rw>=1.3.0
+# Requirements for linux pack
```
Kami: Just a FYI - comments are fine here, but right now we can't use them in st2client requirements.txt since we directly use requirements.txt in the st2client setup.py without ignoring the comments.

PR #1392, opened 2015-04-28
Ignore comments in the st2client requirements file inside setup.py

Kami:
The title says it all.


lakshmi-kannan:
One step at a time. We should use pip's rqmnts file parser code. LGTM.

PR #1393, opened 2015-04-28
!requiretty

manasdk:


DoriftoShoes:
+1

tools/st2_deploy.sh
```diff
@@ -147,6 +147,9 @@ create_user() {
     then
       echo "${SYSTEMUSER}    ALL=(ALL)       NOPASSWD: ALL" >> /etc/sudoers.d/st2
     fi
+    # make sure requiretty is disabled.
+    # unable to get ^Defaults[\s]+requiretty to work.
+    sed -i "s/^Defaults    requiretty/# Defaults    requiretty/g" /etc/sudoers
```
lakshmi-kannan: Ghetto!

Kami: Probably safer to use a regex for whitespace (\s+) in case there are more or less whitespace characters.

manasdk: I tried 

`^Defaults[\s]+requiretty`, `^Defaults(\s)+requiretty` and `^Defaults(\s+)requiretty`

Neither worked so I gave up.

Kami: `sed -i "s/^Defaults\s\+requiretty/# Defaults    requiretty/g" /etc/sudoers`

Should do it I believe.

Kami: I tested the one I posted above locally with GNU sed and it works fine.

PR #1395, opened 2015-04-28
Fix mistral test case where task status is not updated yet

m4dcoder:
If multiple tasks are specified for on-error task defaults and fail is in the list, the WF will fail immediately without waiting for the other tasks to finish.

### DIFF
--- a/st2tests/integration/mistral/base.py
+++ b/st2tests/integration/mistral/base.py
@@ -42,8 +42,12 @@ class TestWorkflowExecution(unittest2.TestCase):
         for i in range(wait):
             eventlet.sleep(3)
             execution = self.st2client.liveactions.get_by_id(execution.id)
+
             if execution.status in ['succeeded', 'failed']:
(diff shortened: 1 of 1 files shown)


st2tests/integration/mistral/base.py
```diff
@@ -42,8 +42,12 @@ def _wait_for_completion(self, execution, wait=300):
         for i in range(wait):
             eventlet.sleep(3)
             execution = self.st2client.liveactions.get_by_id(execution.id)
+
             if execution.status in ['succeeded', 'failed']:
-                break
+                if hasattr(execution, 'result') and 'tasks' in execution.result:
+                    if ([task for task in execution.result['tasks']
```
Kami: Minor style thing - it would be a bit more readable if you moved the whole list comprehension out of the if statement condition:

```python
finished_tasks = [list comprehension]
if finished_tasks:
   break
```

PR #1397, opened 2015-04-28
switch st2 apt repo to http

DoriftoShoes:
Also:
use DOWNLOAD_SERVER variable


Kami:
Hm, we now retrieve key over a non-secure connection which makes us vulnerable to mitm, was this intentional?

I also tested it on a clean Ubuntu server box and https worked just fine.

Did you encounter any issues or?

DoriftoShoes:
I ran into issues while testing out the st2express Docker deployment.  I switched them both to http just for simplicity sake but I can switch the key back to https.

Kami:
Yeah, getting the key over https would be good :)

Apt repo itself is not a big deal since packages are signed.

PR #1398, opened 2015-04-28
Delay invocation of runner generated auth token for async runner

m4dcoder:
Deletion of the runner generated auth token is delayed until post run
when the async action completes. Async actions such as Mistral workflows
require the auth token to launch actions in the workflow.


st2actions/st2actions/container/base.py
```diff
@@ -122,10 +122,12 @@ def _do_run(self, runner, runnertype_db, action_db, liveaction_db):
                                                                 result, context)
             executions.update_execution(updated_liveaction_db)
             LOG.debug('Updated liveaction after run: %s', updated_liveaction_db)
-            try:
-                self._delete_auth_token(runner.auth_token)
-            except:
-                LOG.warn('Unable to clean-up auth_token.')
+
+            if not isinstance(runner, AsyncActionRunner):
```
Kami: Add a docstring / comment on why we do that would be good.

PR #1400, opened 2015-04-28
add_self_check_deb

DoriftoShoes:
Add install of self-check script in debian catalog.

### DIFF
--- a/st2common/packaging/debian/install
+++ b/st2common/packaging/debian/install
@@ -10,3 +10,4 @@ st2/st2.conf etc/st2
 logrotate.d/st2.conf etc/logrotate.d/
 tools/st2ctl usr/bin/
 tools/st2-setup-tests usr/lib/python2.7/dist-packages/st2common/bin/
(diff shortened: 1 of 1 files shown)


st2common/packaging/debian/install
```diff
@@ -10,3 +10,4 @@ st2/st2.conf etc/st2
 logrotate.d/st2.conf etc/logrotate.d/
 tools/st2ctl usr/bin/
 tools/st2-setup-tests usr/lib/python2.7/dist-packages/st2common/bin/
+tools/self-check usr/lib/python2.7/dist-packages/st2common/bin/
```
Kami: Just curious why this bad boy doens't have `st2-` prefix? :)

DoriftoShoes: I felt it was redundant.  It makes sense for the globally installed binaries but it just seemed excessive if the script was already living in st2common/bin.  If we want them all to have the st2 prefix I can update it though.

PR #1407, opened 2015-04-28
Update ActionChain Example with new HTTPServer Port

jfryman:
This PR changes the default port in one of the example ActionChain workflows from `:9101`, being served via `st2api` to `:8080`, which is the listener port for `st2web`.

/cc https://github.com/StackStorm/st2/pull/1362


Kami:
LGTM, +1

PR #1411, opened 2015-04-28
[WIP] st2rc

manasdk:
* rc file for client (sortof)
* contains all the exported variables etc
* will setup an auth token

- [ ] remove jq rdependency or make sure jq is installed
- [ ] simple file for user to modify to pick up all variables? Ideally, no need to touch st2rc.sh on a user box

### DIFF
--- a/tools/st2rc.sh
+++ b/tools/st2rc.sh
@@ -1,0 +1,15 @@
+echo -n "Please enter your StackStorm username: "
+read ST2_USERNAME
+export ST2_USERNAME=$ST2_USERNAME
+
+echo -n "Please enter your StackStrom password: "
+read -s ST2_PASSWORD
+echo
+export ST2_PASSWORD=$ST2_PASSWORD
+
+export ST2_BASE_URL=http://127.0.0.1
+export ST2_AUTH_URL=http://127.0.0.1:9100
+export ST2_API_URL=http://127.0.0.1:9101/v1
+
+TOKEN=`st2 auth $ST2_USERNAME -p $ST2_PASSWORD -j | jq '.token'`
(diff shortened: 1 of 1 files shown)


tools/st2rc.sh
```diff
@@ -0,0 +1,15 @@
...
+echo -n "Please enter your StackStrom password: "
+read -s ST2_PASSWORD
+echo
+export ST2_PASSWORD=$ST2_PASSWORD
+
+export ST2_BASE_URL=http://127.0.0.1
+export ST2_AUTH_URL=http://127.0.0.1:9100
+export ST2_API_URL=http://127.0.0.1:9101/v1
+
+TOKEN=`st2 auth $ST2_USERNAME -p $ST2_PASSWORD -j | jq '.token'`
```
Kami: This requires jq to be installed and available.

auth command in master now includes -t/ --only-token flag so you can use that here and then we don't need jq.

(re-posting comment from #1410)

tools/st2rc.sh
```diff
@@ -0,0 +1,15 @@
+echo -n "Please enter your StackStorm username: "
+read ST2_USERNAME
+export ST2_USERNAME=$ST2_USERNAME
```
Kami: I believe CLI doesn't support st2_username and st2_password environment variable so we probably shouldn't export it (also for the security reasons).

(re-posting comment from #1410)

tools/st2rc.sh
manasdk: Yeah I have an item in todo to remove jq.

tools/st2rc.sh
manasdk: yeah st2_password is definietly a bad idea. Will remove.

manasdk:
Closing. I think we are going a differnt route.

PR #1416, opened 2015-04-29
Add st2-setup-examples and change permissions on st2-self-check

DoriftoShoes:


dzimine:
Can you add the missing TESTPATH here, too? 
https://github.com/StackStorm/st2/blob/master/tools/st2-setup-examples#L11

dzimine:
1) tests installed under a different path, why bother, what difference it makes? 
https://github.com/StackStorm/st2/blob/master/tools/st2-setup-tests#L17-L18

2) st2-set-examples - won't run standalone because we don't add the extra path pack to the config...

Were I you I'd simplify and won't touch st2.conf, install all under usual /opt/stackstorm/packs, and leave them there. May be there's a use case I don't see. But simpler = better?

PR #1417, opened 2015-04-29
Make url in st2client trailing slash agnostic

manasdk:
* Specifying ST2_AUTH_URL and ST2_API_URL with or without trailing slash has no impact
  on the behavior. Previously Specifying a trailing slash would cause st2client to
  fail.


lakshmi-kannan:
LGTM

PR #1418, opened 2015-04-29
Update docs for setup from source

m4dcoder:


docs/source/install/sources.rst
```diff
@@ -52,63 +55,48 @@ and install required dependencies, and run tests.
...
-need to perform to create and setup one that works with fabric\_runner. For more information, see
-:ref:`config-configure-ssh`.
-
 Running
 ~~~~~~~
+Run the following to start |st2|. The script will start |st2| components in screen sessions. ::
 
-To run |st2| from source, it's assumed that python virtual environment
-is activated and in use.
+    sudo ./st2/tools/launchdev.sh start
```
Kami: sudo isn't required and I believe it actually won't work if you use sudo while following this guide.

`make requirements` creates a virtual environment, activates it and installs the dependencies there. Using sudo means virtual environment which is currently activated won't be preserved so starting stuff would fail.

Kami: On a related note, we should also include a paragraph which describes how user can manually activate and use the virtual environment created during `make requirements` (`source virtualenv/bin/activate`).

m4dcoder:
The instructions don't ask user to activate virtualenv.  ./tools/launchdev.sh runs the st2 processes in virtualenv already.  Anyway, I removed the sudo.  But if you guys have time, please try this on a vanilla ubuntu server.

Kami:
You are right about activating the virtualenv in launchdev.

LGTM, +1

PR #1428, opened 2015-04-29
Update deb.rst

rgaertner:
fix download link for requirements.txt


jfryman:
Thanks for the fix! We actually are right now in the process of releasing v0.9, and it should be uploaded shortly. So, I'm going to close this, but make sure it's fixed in the upcoming branch.

Thanks for the PR!

manasdk:
https://github.com/StackStorm/st2/blob/v0.9/docs/source/install/deb.rst#pip

PR #1432, opened 2015-04-30
Allow option to bypass SSL Certificate Check

jfryman:
:cherries:-pick of https://github.com/StackStorm/st2/pull/1431 for v0.9 branch

The recommended behavior is to have all certificates checked for
validity before performing a clone. However, there are times when
this is outright not possible (development with a self-signed cert,
for example).

This commit updates the `packs.download` command, introducing a new
attribute `verifyssl`. By default, the behavior is to continue checking
SSL certificate validity, but now it is possible to clone with an
invalid certificate and proceed along without a care in the world.

Be careful using this... especially in the context of Infra as Code.

### DIFF
--- a/contrib/packs/actions/download.yaml
+++ b/contrib/packs/actions/download.yaml
@@ -19,3 +19,6 @@
       type: "string"
       default: "/opt/stackstorm/packs/"
       immutable: true
+    verifyssl:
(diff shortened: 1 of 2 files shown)


jfryman:
Merging into proper branch

PR #1439, opened 2015-05-03
Fixed alignment of error handling

kirilstrax:


Kami:
Just to clarify it - is this change there just for consistency reasons or is it needed because the workflow is not working is intended?

I'm just curious, because AFAIK, it should work fine as long as all the items on the same level are indented using the same number of whitespace characters (e.g. it doesn't matter if there are two spaces or four).

In any case, LGTM and merged into trunk,

PR #1440, opened 2015-05-03
Separated more visibly a Datastore section of quick start

kirilstrax:
Datastore was embedded in the Define a rule section, although it seems quite independent. Instead, I separated it to be a separate subsection (also referencing it in the https://github.com/StackStorm/st2/blob/master/contrib/tests/README.md)

### DIFF
--- a/docs/source/start.rst
+++ b/docs/source/start.rst
@@ -166,6 +166,7 @@ as ``{{system.my_parameter}}``. This creates ``user=stanley`` key-value pair:
     st2 key set user stanley
     st2 key list
 
+For more information on datastore, check :doc:`datastore`
 
 Deploy a Rule
 -------------------------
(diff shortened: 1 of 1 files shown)


docs/source/start.rst
```diff
@@ -163,6 +153,20 @@ how to configure it, and what is it’s payload structure:
...
+While the most data are retrieved as needed by |st2|, you may need to
+store and share some common variables. Use |st2| datastore service to store
+the values and reference them in rules and workflows
+as ``{{system.my_parameter}}``. This creates ``user=stanley`` key-value pair:
+
+.. code-block:: bash
+
+    st2 key set user stanley
+    st2 key list
+
```
Kami: LGTM.

We could also add a link which says for more info see cli.rst.

PR #1441, opened 2015-05-03
Added info on the tests pack and how to run it manually.

kirilstrax:


contrib/tests/README.md
```diff
@@ -1,9 +1,40 @@
...
+.. code-block:: bash
+
+     sudo su
+     export ST2_AUTH_TOKEN=`st2 auth testu -p testp -t`
+
+2. Check if ``tests`` pack is installed; if it's not installed, install it using usual pack installation routine. For example:
+
+.. code-block:: bash
+
+     sudo cp -r /usr/share/stackstorm/tests /opt/stackstorm/packs/
```
Kami: It's probably better and safer if the user runs `st2-setup-tests` script which also creates a virtual environment and installs the requirements (if any).

To clarify it - your approach should work fine for now, but it will break if we ever add a python runner actions which requires a virtual environment to exist.

kirilstrax: @Kami The usage does start from self-check script description and the link (which includes st2-setup-tests), but since someone using this readme will possibly be using only tests pack, I was trying to also provide a stand-alone procedure.

Kami:
Merged, thanks.

PR #1442, opened 2015-05-03
Proposing to add st2-self-check into this section

kirilstrax:


Kami:
IIRC, self-check script won't work out of the box since it depends on `examples` and `tests` packs which are not installed by default anymore (on CI server we run setup-examples and setup-test scripts).

Kami:
If that's indeed the case, we should consider adding a new section called "Extended Validation" or similar where we explain that the user first needs to set up test packs using setup-* scripts...

st2stanley:
Self-check installs those packs if they aren't already there.

Pwh

Kami:
@DoriftoShoes Thanks. I did actually check the self check script, but I missed that part, lol.

In this case, LGTM and ignore my comments :)

dzimine:
-1: Good content but move this to "troubleshooting" section: we shall keep the 'green path' documentation short and right to the point.

kirilstrax:
Based on @dzimine 's comment moved to troubleshooting, closing this pull request

PR #1443, opened 2015-05-04
Update apache config sample

enykeev:
Based on IRC report from @chrisssss.

In previous version, st2common.logging and st2auth.cmd had more priority than python's own cmd.py and logging.py which resulted in problems with Pecan.core and PDB modules.

Also, to avoid authentication being applied to preflight requests, we need to make an exception for OPTIONS requests.

Other common problems include an absence of `stanley` group and lack of privileges to write into log folder, though it seems they should be addressed on a case-by-case basis.


Kami:
LGTM, +1

PR #1450, opened 2015-05-06
Fix broken packs.download action

Kami:
We didn't propagate `verifyssl` argument to the _clone_repo method.

### DIFF
--- a/contrib/packs/actions/pack_mgmt/download.py
+++ b/contrib/packs/actions/pack_mgmt/download.py
@@ -20,7 +20,8 @@ class InstallGitRepoAction(Action):
         lock_name = hashlib.md5(repo_name).hexdigest() + '.lock'
 
         with LockFile('/tmp/%s' % (lock_name)):
-            abs_local_path = self._clone_repo(repo_url, branch=branch)
+            abs_local_path = self._clone_repo(repo_url=repo_url, verifyssl=verifyssl,
+                                              branch=branch)
             try:
                 # st2-contrib repo has a top-level packs folder that actually contains the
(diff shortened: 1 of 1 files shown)


Kami:
On a related note - I will work up on hooking up `pylint` in our lint process. This error would get caught by pytlint.

Kami:
@kirilstrax Are we already working on some end to end tests for packs actions?

lakshmi-kannan:
LGTM

kirilstrax:
@Kami : it's already done, available in "tests" pack and afaik works on build (https://github.com/StackStorm/st2/tree/master/contrib/tests > test_packs_pack.*)
If I understand correctly (and clone repo is part of packs.install from repo), then it should be tested by "test_packs_install_multiple_packs_from_repo" step of that workflow

DoriftoShoes:
The test needs to be added back  I yanked it out when the whitespace issue was causing failures.  Either way, this would not have caught the verifyssl bug since we don't use it.

Kami:
@DoriftoShoes Sorry, can you please clarify it - we don't use what?

DoriftoShoes:
Maybe I missed the context of this PR.  Was the action completely broken?

Kami:
@DoriftoShoes Yeah, "packs.download" was completely broken which means each packs.install and packs.download action would fail.

Kami:
@kirilstrax Thanks. I had a look at those tests and yeah, you are right, those tests should have caught that bug.

I'll bug @DoriftoShoes and make sure those tests hooked up again :P

PR #1451, opened 2015-05-06
Run pylint during build step, fix bugs and violations in the existing code

Kami:
We should have done this since the day one. While pep8 and pyflakes mostly catch style violations, pylint also catches a lot of actual bugs in the code.

I'm positive this will save us tons of time in the future which we would otherwise spend debugging things and fixing subset of bugs which are common in the dynamic languages (not to mention that pylint already discovered actual bugs and issues in the code).


st2api/st2api/controllers/exp/actionalias.py
```diff
@@ -85,7 +85,7 @@ def post(self, action_alias):
 
     @jsexpose(arg_types=[str], body_cls=ActionAliasAPI)
     def put(self, action_alias_id, action_alias):
-        action_alias_db = self.__get_by_id(action_alias_id)
```
Kami: This also indicates there are no tests for PUT code path - we should work on fixing that.

st2actions/st2actions/runners/mistral/v2.py
```diff
@@ -81,6 +81,7 @@ def _save_workbook(self, name, def_yaml):
             wb = self._client.workbooks.create(def_yaml)
 
         # Update the workbook definition.
+        # pylint: disable=no-member
```
Kami: I need to look into the Mistral client code for that one.

I imagine it does some "dynamic code magic". It might be possible to fix that upstream and make it possible for pylint to detect which attributes are available on `Workbook` and `Execution` class.

st2actions/st2actions/runners/fabricrunner.py
```diff
@@ -227,54 +227,3 @@ def _get_result_status(result, allow_partial_failure):
                 if not success:
                     return LIVEACTION_STATUS_FAILED
         return LIVEACTION_STATUS_SUCCEEDED if success else LIVEACTION_STATUS_FAILED
-
```
manasdk: Finally!!

:100:

lakshmi-kannan:
:ship:

st2actions/st2actions/runners/fabricrunner.py
Kami: Lol, yeah, I wanted to remove this many times before already but never done it :P

st2common/st2common/models/db/stormbase.py
```diff
@@ -13,10 +13,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import bson
 import datetime
```
manasdk: pylint requires a specific order?

Kami: It doesn't, I just sorted it so it follows order from other modules (stdlib, 3rdparty, local).

lakshmi-kannan: Our code base is horrible with this.

st2actions/st2actions/runners/fabricrunner.py
lakshmi-kannan: :phool_fu:

manasdk:
+1

PR #1455, opened 2015-05-07
Throw a better error when enforcing a rule if referenced action doesn't exist

Kami:
Before this change:

```bash
015-05-07 16:56:17,262 140034745491280 ERROR engine [-] Exception enforcing rule RuleDB(action=ActionExecutionSpecDB@140034739839824(ref="sdlcm", parameters="{u'change': u'{trigger.body.change}'}"), criteria={u'trigger.body.change': {u'pattern': u'', u'type': u'exists'}}, description="firing the CR submit case from /v1/webhooks/Sdlcm/sdlcm", enabled=True, id=554a7b0b0ba1603bd4571511, name="wh_sdlcm", tags=[], trigger="core.7bfd8485-8c52-4778-ab9e-ae247c1217dd"): 'NoneType' object has no attribute 'parameters'
  File "/usr/lib/python2.7/dist-packages/st2reactor/rules/engine.py", line 59, in enforce_rules
    enforcer.enforce()  # Should this happen in an eventlet pool?
  File "/usr/lib/python2.7/dist-packages/st2reactor/rules/enforcer.py", line 48, in enforce
    liveaction = RuleEnforcer._invoke_action(self.rule.action, data, context)
  File "/usr/lib/python2.7/dist-packages/st2reactor/rules/enforcer.py", line 65, in _invoke_action
    params = action_param_utils.cast_params(action_ref, params)
  File "/usr/lib/python2.7/dist-packages/st2common/models/utils/action_param_utils.py", line 86, in cast_params
    action_parameters_schema = action_db.parameters
AttributeError: 'NoneType' object has no attribute 'parameters'
```

After this change:

```bash
2015-05-07 19:32:40,388 ERROR [-] Exception enforcing rule RuleDB(action=ActionExecutionSpecDB@140211180910928(ref="core.localaaa", parameters="{u'cmd': u'echo "{{trigger.executed_at}}"'}"), criteria={}, description="Sample rule using an Interval Timer.", enabled=True, id=554bb2400640fd04aae15ec0, name="sample.with_timer", tags=[], trigger="core.f62572f1-b04e-4583-8495-c562e09d6dd3"): Action "core.localaaa" doesn't exist
  File "/data/stanley/st2reactor/st2reactor/rules/engine.py", line 59, in enforce_rules
    enforcer.enforce()  # Should this happen in an eventlet pool?
  File "/data/stanley/st2reactor/st2reactor/rules/enforcer.py", line 44, in enforce
    raise ValueError('Action "%s" doesn\'t exist' % (action_ref))
ValueError: Action "core.localaaa" doesn't exist
```

Spotted by @luisgtz on IRC.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,9 @@ Changelog
 in development
 --------------
 
+* Throw a more-user friendly exception when enforcing a rule if an action referenced inside
+  the rule definition doesn't exist. (improvement)
+
 v0.9.0 - April 29, 2015
 -----------------------
 
--- a/st2reactor/st2reactor/rules/enforcer.py
+++ b/st2reactor/st2reactor/rules/enforcer.py
@@ -17,6 +17,7 @@ import json
 
 from st2common import log as logging
 from st2common.util import reference
+from st2common.util import action_db as action_db_util
 from st2reactor.rules.datatransform import get_transformer
 from st2common.services import action as action_service
 from st2common.models.db.action import LiveActionDB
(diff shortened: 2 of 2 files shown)


Kami:
Note: We could (and probably should?) do this check during the rule creation as well. Even if we do that, we still need to keep this check since there are many possible "races" and cases which could result in action not being available here (e.g action gets removed after rule creation, etc.).

manasdk:
> We could (and probably should?) do this check during the rule creation as well.

No point. Need to check later and also causes a weird dependency between rule and action loading.

Having said taht it is inconsistent as we do require the trigger to be present.

Kami:
Yeah, that's why I wanted to hear comments from other people about that.

Triggers are a bit different though since they are needed earlier in the rule processing. In any case, I'm fine with keeping it as it is right now and not changing the semantics.

lakshmi-kannan:
>Triggers are a bit different though since they are needed earlier in the rule processing. In any case, I'm fine with keeping it as it is right now and not changing the semantics.

I am +1 to this.

manasdk:
wow we all agree :)

+1

PR #1456, opened 2015-05-08
[WIP] Installation Integration Tests

jfryman:
This PR introduces the first of integration tests for the installation of StackStorm. This is needed for a plethora of reasons at this point:

* We have many install methods (Chef/Puppet/`st2_deploy.sh`), and they are already starting to drift
* Asserting success on installation is a bit on a hope and a prayer today (Happy Path). This actually gets us some confidence in what's making it to end-machine
* Needed to start a build pipeline. In order to start packaging up Vagrant into a package that we can distribute a bit more shrink-wrapped, need to ensure that we're wrapping the right stuff
* Confidence migrating to `st2workroom` for `0.10` release.
* Docker Build. In order to properly build out docker containers, need to have some baseline.

This PR models `st2_deploy.sh`, with a passing grade. This does not pass Puppet installation (some defaults, and auth not being enabled), but this is good... there is now a Red/Green Refactor cycle.

## Next Steps

I'd like to get :eyes:, approval in general for the tool and what we have, and then merge. Will then start using for build pipeline work.

Also needed: `pip` requirements set as tests. I played a bit with this tonight, and will play some more after some :zzz:.

Downsides:
* Ruby. Yes, more of it. I did in fact look for a similar tool I could leverage in Python, and while I found one  (agamotto), it looked poor.


DoriftoShoes:
+1 Me likey.

tools/st2_deploy.sh
```diff
@@ -298,7 +298,7 @@ start on runlevel [2345]
 stop on runlevel [016]
 respawn
 
-exec /opt/openstack/mistral/.venv/bin/python /opt/openstack/mistral/mistral/cmd/launch.py --config-file /etc/mistral/mistral.conf --log-config-append /etc/mistral/wf_trace_logging.conf
+exec /opt/openstack/mistral/.venv/bin/python /opt/openstack/mistral/mistral/cmd/launch.py --config-file /etc/mistral/mistral.conf --log-file /var/log/mistral.log --log-config-append /etc/mistral/wf_trace_logging.conf
```
dzimine: isn't the log file specified  inside /etc/mistral/wf_trace_logging.conf ?

tools/st2_deploy.sh
```diff
@@ -266,11 +266,11 @@ fi
 touch $config
 cat <<mistral_config >$config
 [database]
-connection=mysql://mistral:StackStorm@localhost/mistral
-max_pool_size=50
+connection = mysql://mistral:StackStorm@localhost/mistral
+max_pool_size = 100
```
dzimine: Please double-check with @lakshmi-kannan and @m4dcoder they did some experiments on these numbers.

dzimine:
Kudos for more tests. 

Let's review how it fits overall test picture.
Main questions are 1) when/how do we run it in build/test process [1] 2) what are we testing and 3) what is the most efficient way to test it. 

Other smaller questions: 
- why integration tests in st2
- when/how do we run it in build process
- what does it test (st2_deploy.sh)
- do you guys think that indirect testing not sufficient (e.g. if api call to 9101 retrieves the data can we assume that server is running and listening on port and config file is in a right location) 
   - test maintenance: e.g. change config file name/location - now change the test
   - Ruby 
- how vagrant is relevant to st2 repo?
- `/spec` as new top level in st2 repo, IMO not nice.

I don't oppose the approach, just let's not take it lightly. We have a bunch of tests now but only unit tests really work. The rest of the tests needs to come to order.

>  In order to start packaging up Vagrant into a package that we can distribute a bit more shrink-wrapped, 

There are some assumptions about deployment here (Vagrant?), if I am the only one missed the memo - fill me in.

[1] I am behind on that is the build/test process, too; waiting for @DoriftoShoes to finish up OpenStack preps to catch up.

jfryman:
@dzimine Great Questions! Let me see if I can get them all.

## Main Questions

> when/how do we run it in build/test process

This one is a bit fuzzy to me as well. But, in a grander sense, it would be run in two spots:

* after CI passed, which then would kick off asset generation (docker images, vagrant images) or;
* after an update to our install tools (updates on Chef Cookbook, Puppet Module, `st2_deploy.sh`, future Ansible/Salt installers). 

> what are we testing

We're testing that the instructions we are providing to end-users does the needful. At the end of installation of a Docker Container, Vagrant `st2workroom` for testing, or installation via Chef/Puppet/`st2_deploy.sh`... after installing via any of these methods, is the system configured and in the state we desire? Right now, we test code validity, not the integration of our install methods on real systems.

> what is the most efficient way to test it.

Right now, I don't know. The build pipeline is still being finalized, but I also think these are good places to inject this:

* Ad-hoc test installation with workflow (provision node, install, test, repackage). Kicked off by ChatOps
* Automated hooks on update from CM tools, or when `st2_deploy.sh` is updated in the https://github.com/Stackstorm/st2 repo.
* `st2workroom` also has a factory method, which will also be used for rapid local development testing. More on this as it becomes real.

## Other smaller questions:

> why integration tests in st2

Where else might they go? In my reasoning, the app running is what I care about, and keeping these tests as close to the code makes sense to me. Packaging an artifact is going to be different based on the packaging method, so splaying tests to some other repository will give us drift as it's not associated with the code it's meant to be validating.

> when/how do we run it in build process

See above (the most efficient way to test it).

> what does it test (st2_deploy.sh)

See above (what are we testing)

> do you guys think that indirect testing not sufficient (e.g. if api call to 9101 retrieves the data can we assume that server is running and listening on port and config file is in a right location)

You could assume this, but it doesn't allow me to take into account the environment differences between docker/vagrant/rhel/ubuntu/etc. Additional granularity is needed to ensure the *how* the thing is bootstrapped is just as important as answering True/False of if it's bootstrapped at all.

Likewise, these tests allow us to expose an actual problem - that we're doing way too much outside of packaging. Our packages should be smart enough to do a lot of this logic... but because we don't have testing for this earlier in the pipeline, it's needed here. The fact that we now have a test will provide additional confidence while we evaluate where in the build pipeline some of these tests might be better served and earlier in the process as a whole.

> how vagrant is relevant to st2 repo?

It's not so much, and I'd ask you to ignore the layout for just a small bit. This large spec file, now defined, will allow me to do some refactoring to organize a bit more. In the end, I expect that tests will actually test the components (st2api, st2reactor, common requirements, etc), and be organized as such. But, because this is monolithically modeled after `st2_deploy.sh`, this is the first step. The important thing though isn't where the test lies, but that we have the test. The rest after this point is refactoring, with actual signals (red/green) telling me when things are ok or not.

> /spec as new top level in st2 repo, IMO not nice.

Yes, I know. I even knew this would cause angst. I spent about almost an hour and a half trying to start 'correctly', but it was not working. I spent time debugging an issue where `rake` was not loading up a moved directory (I originally had this in `tests/integration/serverspec`), and then basically :rage:d and started working on the tests themselves.Again, goal here is to get the tests, and then figure out how and where things need to live. I do not intend to keep them where they are for the long duration.

spec/vagrant/core_spec.rb
```diff
@@ -0,0 +1,341 @@
+require 'spec_helper'
+
+# Install Packages onto Operating System
```
Kami: Nice, :+1:

tools/st2_deploy.sh
```diff
@@ -176,7 +176,7 @@ install_apt() {
   fi
 
   # Add StackStorm APT repo
-  echo "deb http://downloads.stackstorm.net/deb/ trusty_unstable main" > /etc/apt/sources.list.d/stackstorm.list
+  echo "deb https://downloads.stackstorm.net/deb/ trusty_unstable main" > /etc/apt/sources.list.d/stackstorm.list
```
Kami: I think @DoriftoShoes explicitly reverted it back to http a while ago because of some compatibility issues.

Using https would be good, but it's a not a huge deal breaker here since the packages are already signed.

tools/st2_deploy.sh
jfryman: Yeah, what is the # here? It is configurable in Puppet, but the default is `100`

This is a prime example of mismatch, and why this is needed btw.. It used to be `100`, and that's what was coded into the Puppet Module as the default. But, now it's `50`.

spec/vagrant/core_spec.rb
```diff
@@ -0,0 +1,341 @@
...
+AUTH_FILE_PATH='/etc/st2/htpasswd'
+
+describe file(WEBUI_CONFIG_PATH) do
+  it { should be_file }
+  it { should contain "name: 'StackStorm'" }
+  it { should contain "url: '//:9101'" }
+  it { should contain "auth: '//:9100'" }
+end
+
+describe port(8080) do
```
Kami: For service names, ports and other constants like that it would probably be better to define it at the top of the module.

This way, it's only defined in one place and it's easier to see what's going on and keep it up to date.

tools/st2_deploy.sh
jfryman: Could we not just add `apt-transport-https` to this file? It'll ensure we can support HTTPS, as it seems to be the big deal breaker when HTTPS with `apt` does come up.

I'll revert it, but again, another mismatch with the Puppet module, where HTTPS is default. I'd prefer HTTPS, and we encourage the transition to HTTPS everywhere versus keep legacy if we can avoid it. I'll hold judgement because the bug may be different.

spec/vagrant/core_spec.rb
jfryman: These are purposely split like this, because I intend to have them in their own files after a refactor round. That'll be necessary with docker testing. So, this makes it easy to copy/paste, which literally is next.

Kami: Or maybe even better, would could have a JSON / YAML file which describes different important information for each service (expected name, listen port, log file locations, binary path, etc.).

I know this requires some work and we would need to write some code which converts this metadata into those tests, but imo, it's worth it since it's not specific to those tests and other tests and tools can also leverage this information.

On top of that, it's easier to keep things in sync and up to date since we only need to update one file.

jfryman: I totally would dig that. Can I call that a thing for next round?

tools/st2_deploy.sh
jfryman: Is it? There are three ways this is represented then. One with flags, different between RHEL/Debian, and then a config file. The test should be canonical. What do we want?

spec/vagrant/core_spec.rb
Kami: Yeah, I'm fine with skipping this for now since some tests are better than no tests at this point but we should come back and get that done ASAP.

enykeev:
Yeah, Ruby, nope...

Ok, jokes aside, I'm also not certain what exactly are we testing here and this goes multiple ways:
- for one, I agree that st2_deploy is confusing enough it would benefit from this kind of tests. May be this tests could help us refactor it, though, it might be easier to just throw it away along with st2ctl.

- for another, I do believe that proper indirect integration tests should be enough. With this tests for example, I can't tell whether st2 actually works with mistral and not just running side by side. And this is, for me personally, one of the main indicators that deployment was successful.

- finally, this is Ruby and although I like the simplicity in code, this is not something I would want in st2 repo. Although, I would love all st2web tests to be written in js and running in zombie.js, we're open source project and the idea that you need to know three languages to make a single change would make a number of external contributions negative. That's why when discussing it with @kirilstrax we've decided to put minimally viable number of e2e tests in python and in its own repo and then make sure that changes in st2web won't break them.

@jfryman if you believe this tests are needed, I suggest we put them somewhere else for now, puppet repo or a new one, let them grow (along with our understanding) and see what happens next...

jfryman:
> for another, I do believe that proper indirect integration tests should be enough. With this tests for example, I can't tell whether st2 actually works with mistral and not just running side by side. And this is, for me personally, one of the main indicators that deployment was successful.

This may be ok from a development perspective, but I don't think it's going to fly. To be clear, this test is not nearly meant to be as comprehensive as the end-to-end tests being proposed. It's really only meant to ensure that the methods we use to install the software are behaving as we expected. Namely: is the configuration what we asked for, and are all the services up and running (and running after a reboot).

We need a way to ensure that Docker images are good, that installs via packages are good, etc. E2E tests what go in the box, these test what comes out of the box after shipping. Two sides of the same coin.

> finally, this is Ruby and although I like the simplicity in code, this is not something I would want in st2 repo.

I'm not going to lie, I did struggle with this. Originally, this was going to be in the `st2workroom` project, but I feared that these tests would just fall to the wayside and not get the :eyes: needed. Likewise, where the configuration is currently stored, as well as the install scripts, these are all in this repo as well. So, if I abstract these tests somewhere else, will they get the same visibility? I think the `e2e` tests suffer from a similar problem... because they are not close to the code, they don't get the same consideration.

So, if not here, where do these tests go? The things I am testing are here... so let me turn the question around. Should config come (sans dev environments) come out of this repo entirely? If that is true, then PRs like this:

* https://github.com/StackStorm/st2/pull/1462
* https://github.com/StackStorm/st2/pull/1386
* https://github.com/StackStorm/st2/pull/1372
* https://github.com/StackStorm/st2/pull/1369
and so many more...

They all become part of somewhere else. I can certainly put this somewhere else, but it needs to be visible enough so that when we do add things like this, there is a test somewhere to keep all of our deployment methods in sync. Right now, there is huge latency and discrepancy between them all.

>  I would love all st2web tests to be written in js and running in zombie.js, but we're an open source project and the premise that you need to know three languages to make a single change would make a number of external contributions negative.

I disagree with this assessment.

First, these tests are really only applicable to post-installation. There are only a few parts of code that are being tested with this. `st2_deploy.sh`, `Makefile`s for package installation, and Docker/Vagrant images. Folks caring about these tests are going to be: probably @DoriftoShoes and myself. So, for the core contributor, I still only need to know Python to contribute to the core language. 

Second, I also think your comparison to `st2web` is actually incorrect. I actually would argue that you *should* use `zombie.js` to test your web code? Why? Because the target language is Javascript. Why not write the tests in Javascript. It may seem I'm proving your point, but consider this: There is not an equivalent tool in Python. Shipping this same thing to keep in harmony with the language would require additional work to build a framework, or hack around the test language to keep things in order. I didn't want to choose a tool that 

--------

All this being said, I am open to moving these tests somewhere... but I stand behind the ruby choice. I still think there needs to be actual consideration to where these tests live in relation to the code it's testing. If these tests don't belong in this repo, then neither do the configs they are testing.

PR #1457, opened 2015-05-08
Optimize and speed up Travis builds

Kami:
With this pull request we now also build documentation on Travis. In addition to that, I split things out so we now run 4 builds in parallel (flake8, pylint, docs, tests) instead of running everything sequentially in a single build.

Well, in reality we now actually run 8 builds because we run every task under Python 2.6 and 2.7 (2 x 4), but I will optimize this once we switch to tox (the way Travis works it's not possible to do that without something like tox).

This means the builds should now complete faster (fingers crossed).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,13 @@ sudo: false
 python:
   - "2.6"
   - "2.7"
-script: make tests-travis
+env:
+  - "TASK=flake8"
+  - "TASK=pylint"
+  - "TASK=docs"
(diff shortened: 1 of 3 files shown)


Kami:
Actually, for now (until we switch to tox) we could just run tests under Python 2.6 on Travis.

Python 2.6 is a subset of 2.7 and if we use six library correctly, everything which passes on 2.6 should also pass on 2.7 (vice versa is not true if six library is not used correctly).

PR #1459, opened 2015-05-08
Add missing "_repr_attributes" class variable

Kami:
This makes the whole repl st2client experience more pleasant.


lakshmi-kannan:
LGTM.

PR #1460, opened 2015-05-08
Add "to_dict" method to the Resource class

Kami:
I plan to use this functionality inside actions inside the ``st2`` pack. Returning actual Python objects from the ``st2`` actions makes little sense since common currency in StackStorm (workflows, etc). are simple types + dict and you can't do anything with those custom objects before converting them to a dictionary.

If you think this should really only leave in st2 pack, I can also move it there.

### DIFF
--- a/st2client/st2client/models/core.py
+++ b/st2client/st2client/models/core.py
@@ -57,6 +57,22 @@ class Resource(object):
         for k, v in six.iteritems(kwargs):
             setattr(self, k, v)
 
+    def to_dict(self):
+        """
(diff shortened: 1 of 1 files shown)


lakshmi-kannan:
What the client is missing is it's own model objects for each of they types from st2. This to_dict() belongs in those model classes. Anyways, that's a much bigger change and we should do it when we rewrite CLI.

lakshmi-kannan:
I am ok with this being merged.

PR #1461, opened 2015-05-09
Cache pip packages in Travis

Kami:
Builds should now complete even faster.

If there is ever an issue with cache some how getting corrupted, pip should automatically take care of it, but if it doesn't, you can manually nuke it on this page  - https://travis-ci.org/StackStorm/st2 (Settings -> Cache).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,16 @@ python:
   - "2.6"
   - "2.7"
 env:
-  - "TASK=flake8"
-  - "TASK=pylint"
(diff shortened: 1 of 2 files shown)


Kami:
I've also pushed a change so now we only run tests (and not other tasks) under multiple Python versions.

With all those changes, Travis builds complete a lot faster now (there might still be a delay sometimes when a build queue is overloaded, but we now also use container based builds which should help with the queue).

PR #1462, opened 2015-05-11
Add config for complete Apache deployment

enykeev:
Also, modified st2_deploy and st2ctl to run all the services as `$SYSTEMUSER` since the whole intention of proxy mode auth was that we don't have to run st2auth as root.

### DIFF
--- a/st2auth/conf/apache.full.conf
+++ b/st2auth/conf/apache.full.conf
@@ -1,0 +1,49 @@
+<VirtualHost *:443>
+
+    ServerName 172.168.50.50:443
+
+    DocumentRoot /opt/stackstorm/static/webui
+
+    ProxyPass /api http://localhost:9101
+
+    WSGIScriptAlias /auth /usr/lib/python2.7/dist-packages/st2auth/wsgi.py
+    WSGIDaemonProcess st2auth user=stanley group=stanley processes=2 threads=25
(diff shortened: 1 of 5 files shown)


st2auth/conf/apache.full.conf
```diff
@@ -0,0 +1,49 @@
+<VirtualHost *:443>
+
+    ServerName 172.168.50.50:443
```
lakshmi-kannan: This apache conf needs to be inside vagrant, right? Why can't we use loopback IP here?

DoriftoShoes: We should just leave ServerName out.  If we want it to only listen on that IP, then it should be up in the VirtualHost definition instead of *

jfryman:
Also to throw this out there... this is exactly where https://github.com/StackStorm/st2/pull/1456 tests would come into play.

st2auth/conf/apache.full.conf
lakshmi-kannan: +1 to ^. I missed the `*`.

dzimine:
should we ditch this pr now?

manasdk:
Abandoning!

We have moved to nginx and are going to fully deprecate st2_deploy.sh. Reopen if you disagree.

PR #1464, opened 2015-05-11
Fix default parameter parsing in the aliases parser

Kami:
This pull request fixes alias parser so it correctly parses the default values and works if param stream is empty.

Previously alias_stream was depended on the alias_stream, which means that if param_stream was empty, we would end up in an infinite loop and never parse the alias_stream. That's not the desired behavior, since we want to parse and use the default param values even if the param_stream is empty.


manasdk:
so much voodoo ...

+1

Issue #1315, opened 2015-03-30
st2 run should show the output of the failed task


tobijb:
- st2 commands should show the error of the failed task.  The task failed but I have to go searching for why.  It definitely slows down the development workflow.  It would be awesome if we could include the failure of the last task in the st2 run output.  Something like below.

```
vagrant@st2express:/opt/stackstorm/packs$ st2 run deployer.environment.create domain=BLAH name=master owner=jb
...
id: 551986fd9c99384380121844
action.ref: deployer.environment.create
status: failed
start_timestamp: 2015-03-30T17:25:17.218963Z
end_timestamp: None
+--------------------------+-----------+-------------------------------+-------------------------------+-------------------------------+
| id                       | status    | task                          | action                        | start_timestamp               |
+--------------------------+-----------+-------------------------------+-------------------------------+-------------------------------+
| 551986fd9c99384380121846 | succeeded | get_current_epoch             | deployer.epoch                | Mon, 30 Mar 2015 17:25:17 UTC |
| 551986fe9c99384380121848 | succeeded | set_deployer_environment_crea | st2.kv.set                    | Mon, 30 Mar 2015 17:25:18 UTC |
|                          |           | ted_key                       |                               |                               |
| 551986fe9c9938438012184b | succeeded | set_deployer_environment_owne | st2.kv.set                    | Mon, 30 Mar 2015 17:25:18 UTC |
|                          |           | r_key                         |                               |                               |
| 551986ff9c9938438012184e | failed    | create_environment            | deployer.create_staging_envir | Mon, 30 Mar 2015 17:25:19 UTC |
|                          |           |                               | onment                        |                               |
+--------------------------+-----------+-------------------------------+-------------------------------+-------------------------------+

Failure:
   <FAILURE OUTPUT>
```

Kami:
@tobijb I've merged #1466 which should resolve this.

If it doesn't or you think more can be done, please re-open this ticket or create a new (more specific) one.

PR #1466, opened 2015-05-12
Display error message when running an execution when a workflow action fails

Kami:
The title says it all.

In short this saves everyone an additional command (previously you needed to run `execution get` to retrieve the error message and the error and traceback attributes were also "hidden" among other data).

Before:

screenshot from 2015-05-12 17 00 25

After (top level workflow failure):

top_level_failure

After (task level workflow failure):

task_level_failure

This partly solves #1315, but not fully. We still need to account for a case when there is no top level error but a task fails. I will make that change shortly.

### DIFF
--- a/st2client/st2client/commands/action.py
+++ b/st2client/st2client/commands/action.py
@@ -281,6 +281,13 @@ class ActionRunCommandMixin(object):
             # No child error, there might be a global error, include result in the output
             options['attributes'].append('result')
 
+        # On failure we also want to include error message and traceback at the top level view
+        if instance.status == 'failed':
+            # TODO: If there is no top level error, retrieve error from the last task
+            status_index = options['attributes'].index('status')
+            options['attributes'].insert(status_index + 1, 'result.error')
+            options['attributes'].insert(status_index + 2, 'result.traceback')
+
         # print root task
         self.print_output(instance, formatter, **options)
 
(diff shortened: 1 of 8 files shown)


lakshmi-kannan:
result.error and result.traceback doesn't sound right at top level. It is confusing as to what results you are talking about. Can we change that to failed_on: task name failure_reason: error msg and failure_traceback: tb? Even better would be to have a dict view of failure: task_name: error: tb:

Kami:
There are two types of failures:

1. Top level failures
2. Task level failures (which also cause workflow / cation as a whole to fail)

Currently, this PR only handles first one. When I add support for the second one then yeah, I will need to change the formatting a bit and indicate on which task it failed.

lakshmi-kannan:
If you are solving (1), it still makes less sense to use `result.`. Why can't they simply be error and traceback?

Kami:
I pushed a change so we now also display task-level workflow failure at the top level. I have also updated the PR description so you can see how we now display a top-level and task-level workflow failure.

st2client/st2client/commands/action.py
```diff
@@ -316,6 +339,34 @@ def _get_execution_result(self, execution, action_exec_mgr, args, **kwargs):
...
+    def _get_top_level_error(self, live_action):
+        """
+        Retrieve a top level workflow error.
+
+        :return: (error, traceback)
+        """
+        error = live_action.result.get('error', None)
+        traceback = live_action.result.get('traceback', None)
+
+        return error, traceback
```
lakshmi-kannan: Nit: Paranthesis please :P

lakshmi-kannan:
:+1:

PR #1471, opened 2015-05-12
Fix a bug with alias parser

jfryman:
:cherries:-pick of https://github.com/StackStorm/st2/commit/92063f8347b356c483fdb48310d34387e0767a76.

> Fix a bug with alias parser - make sure it works correctly with the default values and also works if param_stream is none or not provided (we should still parse format stream in that case).


manasdk:
+1

PR #1475, opened 2015-05-13
Include runner_ref in the notification trigger payload.

manasdk:
* Including the payload means client can use this info to better understand
  the results.

### DIFF
--- a/st2actions/st2actions/notifier.py
+++ b/st2actions/st2actions/notifier.py
@@ -121,7 +121,9 @@ class Notifier(consumers.MessageHandler):
         payload = {'execution_id': str(liveaction.id),
                    'status': liveaction.status,
                    'start_timestamp': str(liveaction.start_timestamp),
+                   # deprecate 'action_name' at some point and switch to 'action_ref'
                    'action_name': liveaction.action,
+                   'action_ref': liveaction.action,
                    'runner_ref': self._get_runner(liveaction.action),
                    'parameters': liveaction.parameters,
                    'result': liveaction.result}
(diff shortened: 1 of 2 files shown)


Kami:
LGTM, +1

st2actions/st2actions/notifier.py
```diff
@@ -120,12 +122,17 @@ def _post_generic_trigger(self, liveaction):
                    'status': liveaction.status,
                    'start_timestamp': str(liveaction.start_timestamp),
                    'action_name': liveaction.action,
```
Kami: Looks like there is an inconsistency - above we call it `action_ref` and we call it `action_name` here.

We should also change it to `action_ref` here (yeah, I know that's now your change but since you are already changing the code...).

lakshmi-kannan: Action trigger uses action_name. When I wrote the notifier, I didn't want to break the behavior for action trigger. For notification trigger, I went with action_ref because that's the right thing to do :P. We could have both action_ref and action_name for action trigger and deprecate action_name at some point?

manasdk: It will break https://github.com/StackStorm/st2incubator/blob/master/packs/st2cd/rules/st2_docs_prod.yaml#L8 and many more.

action_name is from before ref and now we have backward compatibilty to deal with.

I can add a new property `action_ref` and slowly deprecate `action_name`.

lakshmi-kannan: Isn't that what I said Capt. obvious? @manasdk

manasdk: Ah. I didn't see the update as I typed my response.

manasdk: Done bae99c5c19384e52662de19f0b8b9ac6a4868123

PR #1476, opened 2015-05-13
Notify & action triggers should contain Execution id as opposed to LiveAction id.

lakshmi-kannan:
We don't want to expose LiveAction id to users. LiveAction is an internal data model. This PR fixes that. 

I went with the approach to consume Execution updates from queue but decided against it because we want the system to deal with LiveAction objects which are there for this purpose. This minimizes the chance of us messing up the execution object. 

We can port this to 0.9.1.

### DIFF
--- a/st2actions/st2actions/notifier.py
+++ b/st2actions/st2actions/notifier.py
@@ -59,9 +59,9 @@ class Notifier(consumers.MessageHandler):
         execution_id = self._get_execution_id(liveaction)
 
         if liveaction.notify is not None:
-            self._post_notify_triggers(liveaction, execution_id)
+            self._post_notify_triggers(liveaction=liveaction, execution_id=execution_id)
 
-        self._post_generic_trigger(liveaction, execution_id)
+        self._post_generic_trigger(liveaction=liveaction, execution_id=execution_id)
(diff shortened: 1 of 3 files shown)


st2actions/st2actions/notifier.py
```diff
@@ -69,15 +80,15 @@ def _post_notify_triggers(self, liveaction):
 
         if notify.on_complete:
             self._post_notify_subsection_triggers(
-                liveaction, notify.on_complete, default_message_suffix='completed.')
+                liveaction, execution_id, notify.on_complete, default_message_suffix='completed.')
```
Kami: Using keyword args would make it more readable :P

Kami:
LGTM, although again, some tests would be nice :P

lakshmi-kannan:
I could have added a test to inject an execution object corresponding to liveaction into the db. I'll work on it.

st2actions/st2actions/notifier.py
lakshmi-kannan: Yeah, I saw this code and I had the same thought. I'll add a commit. Wanted to get this reviewed first.

PR #1477, opened 2015-05-13
`local-shell-cmd` documentation fix

armab:
According to [RUNNER_TYPES](https://github.com/StackStorm/st2/blob/master/st2actions/st2actions/bootstrap/runnersregistrar.py#L62) there are no such Action Runners like [`local-script-cmd`](https://github.com/StackStorm/st2/blob/master/docs/source/runners.rst#local-command-runner-local-script-cmd)
and [`local-script`](https://github.com/StackStorm/st2/blob/master/docs/source/actions.rst#available-runners)

Seems documentation typo.

### DIFF
--- a/docs/source/actions.rst
+++ b/docs/source/actions.rst
@@ -84,17 +84,19 @@ Currently the system provides the following runners:
 2. ``local-shell-script`` - This is the local runner. Actions are implemented as
    scripts. They are executed on the same hosts where |st2| components are
    running.
-3. ``remote-script`` - This is a remote runner. Actions are implemented as scripts.
+3. ``remote-shell-cmd`` - This is a remote runner. This runner executes
+   a Linux command on one or more remote hosts provided by the user.
+4. ``remote-shell-script`` - This is a remote runner. Actions are implemented as scripts.
    They run on one or more remote hosts provided by the user.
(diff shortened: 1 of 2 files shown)


Kami:
Good catch, thanks.

Looks like we missed some spots when renaming the runners for consistency reasons.

docs/source/actions.rst
```diff
@@ -79,20 +79,22 @@ Available runners
...
 Currently the system provides the following runners:
 
-1. ``local-script`` - This is the local runner. Actions are implemented as
+1. ``local-shell-cmd`` - This is the local runner. This runner executes
+   a Linux command on the same host where |st2| components are running.
+2. ``local-shell-script`` - This is the local runner. Actions are implemented as
    scripts. They are executed on the same hosts where |st2| components are
    running.
-2. ``remote-script`` - This is a remote runner. Actions are implemented as scripts.
+3. ``remote-script`` - This is a remote runner. Actions are implemented as scripts.
```
Kami: This should also be remote-shell-script and looks like there is also a missing entry for remote-shell-cmd.

armab:
@Kami it's updated

Kami:
Merged into master. Thanks.

PR #1479, opened 2015-05-14
Add PolicyType and Policy models

m4dcoder:
Add appropriate DB, persistence, and API models, REST controllers, and CLI operations for PolicyType and Policy.

### DIFF
--- a/st2api/st2api/controllers/resource.py
+++ b/st2api/st2api/controllers/resource.py
@@ -188,7 +188,7 @@ class ResourceController(rest.RestController):
         return resource_db
 
 
-class ContentPackResourceControler(ResourceController):
+class ContentPackResourceController(ResourceController):
     include_reference = False
 
     @jsexpose(arg_types=[str])
@@ -221,7 +221,7 @@ class ContentPackResourceControler(ResourceController):
         return result
 
     def _get_all(self, **kwargs):
-        result = super(ContentPackResourceControler, self)._get_all(**kwargs)
+        result = super(ContentPackResourceController, self)._get_all(**kwargs)
 
         if self.include_reference:
             for item in result:
--- a/st2api/st2api/controllers/v1/policies.py
+++ b/st2api/st2api/controllers/v1/policies.py
@@ -1,0 +1,263 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import copy
+
+from mongoengine import ValidationError
+import pecan
+from pecan import abort
+from six.moves import http_client
+
+from st2api.controllers import resource
+from st2common import log as logging
+from st2common.constants.pack import DEFAULT_PACK_NAME
+from st2common.exceptions.apivalidation import ValueValidationException
+from st2common.exceptions.db import StackStormDBObjectConflictError
+from st2common.models.api.base import jsexpose
+from st2common.models.api.policy import PolicyTypeAPI, PolicyAPI
+from st2common.models.db.policy import PolicyTypeReference
+from st2common.models.system.common import InvalidReferenceError
+from st2common.persistence.policy import PolicyType, Policy
+from st2common.validators.api.misc import validate_not_part_of_system_pack
+
+
+LOG = logging.getLogger(__name__)
+
+
+class PolicyTypeController(resource.ResourceController):
+    model = PolicyTypeAPI
+    access = PolicyType
+
+    supported_filters = {
+        'resource_type': 'resource_type'
+    }
+
+    query_options = {
+        'sort': ['resource_type', 'name']
+    }
+
+    include_reference = False
+
+    @jsexpose(arg_types=[str])
+    def get_one(self, ref_or_id):
+        return self._get_one(ref_or_id)
+
+    @jsexpose()
+    def get_all(self, **kwargs):
+        return self._get_all(**kwargs)
+
+    def _get_one(self, ref_or_id):
+        LOG.info('GET %s with ref_or_id=%s', pecan.request.path, ref_or_id)
+
+        try:
+            instance = self._get_by_ref_or_id(ref_or_id=ref_or_id)
+        except Exception as e:
+            LOG.exception(e.message)
+            abort(http_client.NOT_FOUND, e.message)
+            return
+
+        result = self.model.from_model(instance)
+        if result and self.include_reference:
+            resource_type = getattr(result, 'resource_type', None)
+            name = getattr(result, 'name', None)
+            result.ref = PolicyTypeReference(resource_type=resource_type, name=name).ref
+
+        LOG.debug('GET %s with ref_or_id=%s, client_result=%s',
+                  pecan.request.path, ref_or_id, result)
+
+        return result
+
+    def _get_all(self, **kwargs):
+        result = super(PolicyTypeController, self)._get_all(**kwargs)
+
+        if self.include_reference:
+            for item in result:
+                resource_type = getattr(item, 'resource_type', None)
+                name = getattr(item, 'name', None)
+                item.ref = PolicyTypeReference(resource_type=resource_type, name=name).ref
+
+        return result
+
+    def _get_by_ref_or_id(self, ref_or_id):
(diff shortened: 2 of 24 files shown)


st2api/st2api/controllers/v1/policies.py
```diff
@@ -0,0 +1,263 @@
...
+    @jsexpose(body_cls=PolicyAPI, status_code=http_client.CREATED)
+    def post(self, instance):
+        """
+            Create a new policy.
+            Handles requests:
+                POST /policies/
+        """
+        op = 'POST /policies/'
+
+        if not hasattr(instance, 'enabled'):
```
manasdk: Could be handled in DB?

st2api/st2api/controllers/v1/policies.py
```diff
@@ -0,0 +1,263 @@
...
+            LOG.debug('%s found object: %s', op, db_model)
+        except Exception as e:
+            LOG.exception('%s unable to find object.', op)
+            abort(http_client.NOT_FOUND, e.message)
+            return
+
+        db_model_id = db_model.id
+
+        try:
+            validate_not_part_of_system_pack(db_model)
```
manasdk: Why not?

st2api/st2api/controllers/v1/policies.py
```diff
@@ -0,0 +1,263 @@
...
+
+        return filters
+
+
+class PolicyController(resource.ContentPackResourceController):
+    model = PolicyAPI
+    access = Policy
+
+    supported_filters = {
+        'pack': 'pack'
```
manasdk: `resource_ref ` could also be a useful filter

m4dcoder: Good point!

st2client/st2client/client.py
```diff
@@ -73,55 +73,79 @@ def __init__(self, base_url=None, auth_url=None, api_url=None, api_version=None,
 
         # Instantiate resource managers and assign appropriate API endpoint.
         self.managers = dict()
-        self.managers['Token'] = models.ResourceManager(
-            models.Token, self.endpoints['auth'], cacert=self.cacert, debug=self.debug)
-        self.managers['RunnerType'] = models.ResourceManager(
-            models.RunnerType, self.endpoints['api'], cacert=self.cacert, debug=self.debug)
+
```
manasdk: Would be nice to avoid stylistic changes if the affected files > 5.

st2api/st2api/controllers/v1/policies.py
m4dcoder: I copied this from the ContentPack controller.  I guess I can use cases where users may want to either delete and change the policies that are deployed as part of the core pack.  But it is going to be inconsistent with how resources (i.e. actions) are managed under the core pack.  Are we ok with that?

st2api/st2api/controllers/v1/policies.py
m4dcoder: Yes.  I just copied this from https://github.com/StackStorm/st2/blob/master/st2api/st2api/controllers/v1/actions.py#L80.

st2client/st2client/client.py
m4dcoder: Ok.  I'll keep that in mind.  It is too tempting.  The stuffs are not sorted and it was hard to read.... :P

st2common/st2common/models/db/policy.py
```diff
@@ -0,0 +1,180 @@
...
+        help_text='A flag indicating whether the runner for this type is enabled.')
+    module = me.StringField(
+        required=True,
+        help_text='The python module that implements the policy for this type.')
+    parameters = me.DictField(
+        help_text='The specification for parameters for the policy type.')
+
+    def __init__(self, *args, **kwargs):
+        super(PolicyTypeDB, self).__init__(*args, **kwargs)
+        self.ref = PolicyTypeReference.to_string_reference(resource_type=self.resource_type,
```
manasdk: The fact that this ref is diffrent from other ref in the system bothers me. This ref does not follows the standard `pack.name` convention either.

I see the value in scope by name of the type but couldn't that be simply captured in the name? To deviate from the usual form of ref which is basically the way to uniquely identify, in a human readable form, objects in our system we need absolute conviction. I am not convinced.

manasdk:
I am +1 to the overall change.

I am not in favor of `PolicyTypeReference` as meaning anything other than ref i.e. `pack.name`.

st2common/st2common/models/db/policy.py
m4dcoder: I thought about that and started initially to name the policy like "action.concurrency".  But how is it going to be enforced?  Just checking the name and then throw an exception that user needs to prefix it with "action" (or other type)?

st2api/st2api/controllers/v1/policies.py
manasdk: Policies are fundamentally deployment specific. Heck even the fact that `system` actions cannot be changed is broken in my view.

st2client/st2client/client.py
manasdk: I understand the temptation but if you believe that you would like the code to be reviewed then keeping in mind that this adds noise and make reviewing harder is also important.

st2common/st2common/models/db/policy.py
manasdk: Why enforce?

st2api/st2api/controllers/resource.py
```diff
@@ -188,7 +188,7 @@ def _get_by_name(self, resource_name):
         return resource_db
 
 
-class ContentPackResourceControler(ResourceController):
+class ContentPackResourceController(ResourceController):
```
lakshmi-kannan: These changes should definitely be in its own PR.

st2client/st2client/client.py
m4dcoder: Got it.

st2api/st2api/controllers/v1/policies.py
```diff
@@ -0,0 +1,263 @@
...
+            abort(http_client.BAD_REQUEST, str(e))
+            return
+
+        LOG.debug('%s updated object: %s', op, db_model)
+        LOG.audit('Policy updated. Policy.id=%s' % (db_model.id), extra={'policy_db': db_model})
+
+        return self.model.from_model(db_model)
+
+    @jsexpose(arg_types=[str], status_code=http_client.NO_CONTENT)
+    def delete(self, ref_or_id):
```
lakshmi-kannan: I am not sure if we should support delete this early. Don't we have to delete all references to the policy before deleting the policy itself?

st2client/st2client/client.py
lakshmi-kannan: Plus you are gonna see merge conflict with https://github.com/StackStorm/st2/pull/1480 :P.

st2common/st2common/models/db/policy.py
m4dcoder: > I see the value in scope by name of the type
I prefer organization and cleanliness.  Enforcing some naming standard makes the UI cleaner.  But if this is causing a lot of heartache, I can remove it.

st2api/st2api/controllers/v1/policies.py
m4dcoder: Do you mean the other way around?  Policy overlays on top of a resource (i.e. action).  Deleting a policy should be independent from the action.  On the other hand, if a user deletes an action for example, should all the existing policies that reference the action be deleted as well?  At this point (experimental), I don't think it matters.  If the action is deleted, the action would never be executed, and any policies that reference the action won't be touched.

st2client/st2client/client.py
m4dcoder: Hm, maybe I should merge now. :p

st2api/st2api/controllers/v1/policies.py
lakshmi-kannan: I guess I don't how the association is going to established between policies and content at this point. So yeah, deleting a policy cannot be simple as just deleting the model from db. It probably has more logic to it is all I am saying. I am fine with leaving the code as such.

st2common/st2common/models/db/policy.py
manasdk: > organization and cleanliness

organization and cleanliness are subjective for the most part. Carrot and no stick would be my approach. Same property different meaning depending on the object leads to confusion. 

Anyway we are discussing 2 separate matters -
1. ref property meaning something different PolicyType. Should not.
2. Enforcing a specific structure for any naming property - that is arguable but I am not against.

st2api/st2api/controllers/v1/policies.py
m4dcoder: Another question, what should the behavior be for a st2 upgrade?  If users modified/deleted the policies in core, during an upgrade, should we overlay an existing installation?

st2api/st2api/controllers/v1/policies.py
m4dcoder: Sure. When we can articulate what that association is, we can add logic to the delete.

manasdk:
I guess we all have decided not to block on PolicyTypeReference.

m4dcoder:
Except for validate_not_part_of_system_pack(db_model), I addressed comments above. If we want to allow PUT/DELETE for policies in system pack, I can remove validate_not_part_of_system_pack(db_model) in a separate commit.

PR #1480, opened 2015-05-14
New API endpoint for re-running an execution

Kami:
This pull request adds a new API endpoint for re-running an action execution.

```
POST /v1/actionexecutions/<execution id>/re_run/
```

Previously, this functionality was implemented client side in the CLI, but now this functionality is also needed by ChatOps so it finally makes sense to move it to the server. The server side logic is exactly the same was the client side logic was so the existing "execution re-run" command works in  exactly the same way as well.

Having this functionality on the server also means WebUI can now easily implement re-run as well.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,8 @@ in development
 * Update CLI so it displays the error at the top level when using ``run``, ``execution run`` or
   ``execution get`` when executed workflow fails. (improvement)
 * Action trigger now contains execution id as opposed to liveaction id. (bug-fix)
+* Add new API endpoint for re-running an execution (``POST /actionexecutions/<id>/re_run/``).
+  (new-feature)
 
 v0.9.0 - April 29, 2015
 -----------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,7 +18,7 @@ in development
 * Update CLI so it displays the error at the top level when using ``run``, ``execution run`` or
   ``execution get`` when executed workflow fails. (improvement)
 * Action trigger now contains execution id as opposed to liveaction id. (bug-fix)
-* Add new API endpoint for re-running an execution (``POST /actionexecutions/<id>/re_run/``).
+* Add new API endpoint for re-running an execution (``POST /executions/<id>/re_run/``).
   (new-feature)
 
 v0.9.0 - April 29, 2015
--- a/st2api/st2api/controllers/v1/actionexecutions.py
+++ b/st2api/st2api/controllers/v1/actionexecutions.py
@@ -192,6 +192,13 @@ class ActionExecutionReRunController(ActionExecutionsControllerMixin, ResourceCo
 
     @jsexpose(body_cls=ExecutionParameters, status_code=http_client.CREATED)
     def post(self, execution_parameters, execution_id):
+        """
+        Re-run the provided action execution optionally specifying override parameters.
+
+        Handles requests:
+
+            POST /executions/<id>/re_run
+        """
         parameters = execution_parameters.parameters
 
         # Note: We only really need parameters here
--- a/st2api/st2api/controllers/v1/actionexecutions.py
+++ b/st2api/st2api/controllers/v1/actionexecutions.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import copy
 import datetime
 
 import jsonschema
@@ -31,6 +32,7 @@ from st2common.constants.action import CANCELABLE_STATES
 from st2common.models.api.action import LiveActionAPI
 from st2common.models.api.base import jsexpose
 from st2common.models.api.execution import ActionExecutionAPI
+from st2common.models.db.action import LiveActionDB
 from st2common.persistence.action import LiveAction
 from st2common.persistence.execution import ActionExecution
 from st2common.services import action as action_service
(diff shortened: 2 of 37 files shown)


st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -130,6 +173,45 @@ def get(self, id, attribute, **kwargs):
...
 
 
+class ActionExecutionReRunController(ActionExecutionsControllerMixin, ResourceController):
+    supported_filters = {}
+    exclude_fields = [
+        'result',
+        'trigger_instance'
+    ]
+
+    class ExecutionParameters(object):
```
Kami: This is a "limitation" of our existing `jsexpose` decorator - in most other places we can just use DB models, but in this case API only takes dictionary with "parameters" key so we need to create a "dummy" class for that.

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -141,7 +223,9 @@ class ActionExecutionsController(ActionExecutionsControllerMixin, ResourceContro
 
     children = ActionExecutionChildrenController()
     attribute = ActionExecutionAttributeController()
+    re_run = ActionExecutionReRunController()
```
lakshmi-kannan: Should we also add rerun for convenience?

st2api/tests/unit/controllers/v1/test_executions_simple.py
```diff
@@ -281,6 +281,47 @@ def test_post_with_st2_context_in_headers_failed(self):
...
 
+    def test_re_run_success(self):
+        # Create a new execution
+        post_resp = self._do_post(LIVE_ACTION_1)
+        self.assertEqual(post_resp.status_int, 201)
+        execution_id = self._get_actionexecution_id(post_resp)
+
+        # Re-run created execution (no parameters overrides)
+        data = {}
+        re_run_resp = self.app.post_json('/v1/actionexecutions/%s/re_run' %
```
lakshmi-kannan: Can you use executions instead of actionexecutions? We should deprecate actionexecutions.

st2api/tests/unit/controllers/v1/test_executions_simple.py
```diff
@@ -281,6 +281,47 @@ def test_post_with_st2_context_in_headers_failed(self):
...
+                (execution_id), data)
+        self.assertEqual(re_run_resp.status_int, 201)
+
+        # Re-run created execution (with parameters overrides)
+        data = {'parameters': {'a': 'val1'}}
```
lakshmi-kannan: I personally like this.

CHANGELOG.rst
```diff
@@ -18,6 +18,8 @@ in development
 * Update CLI so it displays the error at the top level when using ``run``, ``execution run`` or
   ``execution get`` when executed workflow fails. (improvement)
 * Action trigger now contains execution id as opposed to liveaction id. (bug-fix)
+* Add new API endpoint for re-running an execution (``POST /actionexecutions/<id>/re_run/``).
```
lakshmi-kannan: s/actionexecutions/executions.

lakshmi-kannan:
Overall LGTM.

manasdk:
`POST /v1/actionexecutions/<execution id>/re_run/`

re_run is not a property of an execution. This is encoding a verb in a URL - REStful still?

`POST /v1/actionexecutions/`

with 

BODY : {'re_run': '<execution_id>'}

seems more in line - no?

st2api/tests/unit/controllers/v1/test_executions_simple.py
manasdk: Yup. Parameter overrides are awesome.

st2client/st2client/client.py
```diff
@@ -73,25 +75,25 @@ def __init__(self, base_url=None, auth_url=None, api_url=None, api_version=None,
 
         # Instantiate resource managers and assign appropriate API endpoint.
         self.managers = dict()
-        self.managers['Token'] = models.ResourceManager(
+        self.managers['Token'] = ResourceManager(
```
manasdk: even you are making unnecessary changes ... you monster!

lakshmi-kannan:
@manasdk and I had a discussion on slack. I have an alternate proposal.

Maybe we could drop the re-run and say POST /executions/id creates that execution again to be more RESTish?

st2api/st2api/controllers/v1/actionexecutions.py
Kami: I think we should have one or another and not both ("There should be one-- and preferably only one --obvious way to do it." :P).

st2api/tests/unit/controllers/v1/test_executions_simple.py
Kami: Yeah, I personally never liked "actionsexecutions", but that's a better change.

I'm fine with changing just this one controller and change other in a separate PR.

Kami:
@manasdk The problem with REST is that there is no one way / one standard way to do it so everyone does it a bit differently.

In this case "re-run" is an operation on a resource and usually that goes to the URI.

I'm -1 for "Maybe we could drop the re-run and say POST /executions/id creates that execution" - it's not explicit and you have no idea what that operation does if you don't check the docs.

st2api/tests/unit/controllers/v1/test_executions_simple.py
lakshmi-kannan: I already handled this. You can just replace this and it should work. Let's deprecate actionexecutions in another PR. 
https://github.com/StackStorm/st2/blob/master/st2api/st2api/controllers/v1/root.py#L31

manasdk:
> In this case "re-run" is an operation on a resource and usually that goes to the URI.

What? I don't understand REST. Verbs should not be in URI but whatever makes you happy.

st2api/tests/unit/controllers/v1/test_executions_simple.py
Kami: Ah, I see. Alright, cool, will make that change (will also update client to use just "executions" for other operations as well).

Kami: 86fc3e478e96732e410a554c66ea17ea6d9b307a

lakshmi-kannan: 86fc3e4 looks good.

PR #1481, opened 2015-05-14
Update mongoengine version in requirements.txt

m4dcoder:


requirements.txt
```diff
@@ -6,11 +6,11 @@ gitpython==0.3.2.1
 jinja2
 jsonschema>=2.3.0
 kombu
-mongoengine>=0.8.7,<0.9
+pymongo<3.0
+mongoengine>=0.9
```
manasdk: Reason for this change?

Also, seems strange that pymongo is pegged but mongoengine has no upper bound.

m4dcoder: Yes. I need the "modify" function in 0.9.0 that supports atomic update in mongoengine document. There's an issue with pymongo >= 3.0 and will throw an exception "False is not a read preference".    Reference: https://github.com/MongoEngine/mongoengine/issues/935  pip installs mongoengine 0.9.0.  The fix seems to be in 0.9.1.

st2common/st2common/services/executions.py
```diff
@@ -53,7 +53,7 @@
 
 LOG = logging.getLogger(__name__)
 
-SKIPPED = ['id', 'callback', 'action', 'runner_info']
+SKIPPED = ['id', 'callback', 'action', 'runner_info', 'notify']
```
lakshmi-kannan: This seems like an related change. Can you revert this?

requirements.txt
m4dcoder: I'll change the requirements to mongoengine==0.9. I think that should ease the concern here on unbound version of mongoengine with a bound version of pymongo.

st2common/st2common/services/executions.py
m4dcoder: This is actually causing an error when mongoengine is upgraded.  I tested this before and after the upgrade.  Basically, the ActionExecutionDB don't have a notify field and the code is trying to assign notify to it.  Mongoengine is complaining about it.

requirements.txt
Kami: @m4dcoder Yeah, we should use more strict version pinning otherwise we will encounter problems again in the future -e.g. `>=0.9.1,<0.10`

Kami:
On a related note - we should also look at upgrading pymongo - last time I had a look it looked like it only requires some small changes to make it work.

m4dcoder:
The problem with pymongo 3.0 is more involved.  I got pass the connection error on read_preference in pymongo but now I'm getting other errors.  So I'm going to leave it at pymongo<3.0 for now.

ERROR: test_chain_runner_no_default_multiple_options (tests.unit.test_actionchain.TestActionChainRunner)
  File "/home/vagrant/code/stackstorm/st2/virtualenv/local/lib/python2.7/site-packages/mock.py", line 1201, in patched
    return func(*args, **keywargs)
  File "/home/vagrant/code/stackstorm/st2/st2actions/tests/unit/test_actionchain.py", line 161, in test_chain_runner_no_default_multiple_options
    chain_runner.run({})
  File "/home/vagrant/code/stackstorm/st2/st2actions/st2actions/runners/actionchainrunner.py", line 252, in run
    task_result = self._format_action_exec_result(**format_kwargs)
  File "/home/vagrant/code/stackstorm/st2/st2actions/st2actions/runners/actionchainrunner.py", line 383, in _format_action_exec_result
    execution_db = ActionExecution.get(liveaction__id=str(liveaction_db.id))
  File "/home/vagrant/code/stackstorm/st2/st2common/st2common/persistence/base.py", line 57, in get
    return cls._get_impl().get(*args, **kwargs)
  File "/home/vagrant/code/stackstorm/st2/st2common/st2common/models/db/__init__.py", line 84, in get
    instance = instances[0] if instances else None
  File "/home/vagrant/code/stackstorm/st2/virtualenv/local/lib/python2.7/site-packages/mongoengine/queryset/base.py", line 177, in __nonzero__
    return self._has_data()
  File "/home/vagrant/code/stackstorm/st2/virtualenv/local/lib/python2.7/site-packages/mongoengine/queryset/base.py", line 172, in _has_data
    return False if queryset.first() is None else True
  File "/home/vagrant/code/stackstorm/st2/virtualenv/local/lib/python2.7/site-packages/mongoengine/queryset/base.py", line 309, in first
    result = queryset[0]
  File "/home/vagrant/code/stackstorm/st2/virtualenv/local/lib/python2.7/site-packages/mongoengine/queryset/base.py", line 160, in __getitem__
    return queryset._document._from_son(queryset._cursor[key],
  File "/home/vagrant/code/stackstorm/st2/virtualenv/local/lib/python2.7/site-packages/mongoengine/queryset/base.py", line 1410, in _cursor
    **self._cursor_args)
  File "/home/vagrant/code/stackstorm/st2/virtualenv/local/lib/python2.7/site-packages/pymongo/collection.py", line 929, in find
    return Cursor(self, *args, **kwargs)
TypeError: __init__() got an unexpected keyword argument 'snapshot'

m4dcoder:
Here's the mongoengine issue https://github.com/MongoEngine/mongoengine/issues/939.  It's closed but the new version is not out on pip yet.

lakshmi-kannan:
LGTM. Let's get this merged. We need a timer action to check on github issue status.

PR #1484, opened 2015-05-16
Fix doc link in st2api

dzimine:
Setting up my st2workroom. Wile with this, fixing broken link in api root.
Doc version is two-digit (0.9).

IMAO, the root should not return html, but a proper json with the links to the next level resources. `GET https://api.github.com/` to see what I mean.


lakshmi-kannan:
LGTM. Good catch.

PR #1485, opened 2015-05-17
Changes to test_packs_pack

kirilstrax:
1. changed verification step to use json format for output (since one of the periodic failures is too narrow field format in the output; instead of fighting it,it's more convenient to use json)
2. changed test_packs_setup_virtualenv success criteria (text changed in the message)
3. changed test_packs_load_register_all success criteria - now it reports 4 items ("aliases" was not previously there)

### DIFF
--- a/contrib/tests/actions/chains/test_packs_pack.yaml
+++ b/contrib/tests/actions/chains/test_packs_pack.yaml
@@ -57,7 +57,7 @@ chain:
         params:
             env:
               ST2_AUTH_TOKEN: "{{token}}"
-            cmd: "st2 action list --pack {{ pack_to_install_1 }} | grep '{{ pack_to_install_1 }}'"
+            cmd: "st2 action list -j --pack {{ pack_to_install_1 }} | grep '{{ pack_to_install_1 }}'"
         on-success: test_check_pack_to_install_2_installed
         on-failure: error_handler
     -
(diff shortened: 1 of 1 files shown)


contrib/tests/actions/chains/test_packs_pack.yaml
```diff
@@ -120,7 +120,7 @@ chain:
         params:
             env:
               ST2_AUTH_TOKEN: "{{token}}"
-            cmd: "st2 run packs.load register=all | grep -c 'Registered' | grep 3"
+            cmd: "st2 run packs.load register=all | grep -c 'Registered' | grep 4"
```
lakshmi-kannan: This `4` could be an input variable defined in the workflow. So we run the validate workflow with this number passed in.

lakshmi-kannan:
LGTM.

Kami:
We should also figure out how to hook up those tests on TravisCI (probably, it would be quite a PITA).

Or at least, make sure our StackStorm CI system updates pull request status when the tests finished. There is a Github action for that so it should be possible to do that without a lot of work.

contrib/tests/actions/chains/test_packs_pack.yaml
kirilstrax: @lakshmi-kannan: yes, a few things currently are hard-coded and can be improved. But I wanted first of all to fix current flow, to get it back to the build process, as this is important. So I tried to do minimum changes.

PR #1487, opened 2015-05-17
Fixing STORM-1193

kirilstrax:
Resolves "Class "SimplePollingSensor" doesn't implement required "update_trigger" method from the base class" error when running sensor using "st2sensorcontainer --config-file=/etc/st2/st2.conf --sensor-name=SimplePollingSensor"


Kami:
Merged, thanks.

PR #1488, opened 2015-05-17
Fixing STORM-1193 - part 2

kirilstrax:
Resolves "Class "SimpleSensor" doesn't implement required "update_trigger" method from the base class" error when running sensor using "st2sensorcontainer --config-file=/etc/st2/st2.conf --sensor-name=SimpleSensor"


Kami:
Merged, thanks.

PR #1494, opened 2015-05-18
Refactor remote runners into a base class + "class per runner"

### LINKED ISSUE #1365: Refactor remote runners, use two classes instead of one
Currently, remote command and remote script runners both point to the same `FabricRunner` class and inside that class we do nasty branching depending on the value of the `runner_name` inside the action metadata file.

Instead of doing that, we should have two classes `RemoteCommandRunner` and `RemoteScriptRunner` and reference those classes when registering the runners.

This way, the classes itself don't need to be aware of the actual runner names which also decreases the coupling.

Also see: #1364

Kami:
This is a minor technical debt cleanup, part of #1365.

This pull request refactors remote runners into two classes and reduces coupling between the implementation detail (fabric) and the actual remote runner interface.

This should also make switching backend implementation (e.g. moving away from fabric) slightly easier.

Resolves #1365.


e1.yaml
```diff
@@ -0,0 +1,18 @@
+    name: "examples.webhook_file1"
```
lakshmi-kannan: Intentional that you checked this in?

Kami: Totally not intentional :P

lakshmi-kannan:
LGTM. Thanks for doing this. +1 to merge if tests pass.

PR #1495, opened 2015-05-18
Resolving "No connection adapters were found for..." error

kirilstrax:
stormbot_say returns error: "No connection adapters were found for '{'action': 'http://localhost:9101', 'datastore': 'http://localhost:9101', 'reactor': 'http://localhost:9101'}:9101/v1/keys/?name=bot_host'
I think client has its own defaults now, so I think setting up endpoints from script is no longer needed.

### DIFF
--- a/contrib/core/actions/stormbot_say/stormbot_say
+++ b/contrib/core/actions/stormbot_say/stormbot_say
@@ -21,12 +21,6 @@ import requests
 
 from st2client.client import Client
 
-st2_endpoints = {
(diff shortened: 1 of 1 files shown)


manasdk:
Oh, this still exists! We should remove this action. The new chatops stuff is very different.

https://github.com/StackStorm/st2/blob/master/instructables/chatops.md

kirilstrax:
@manasdk I will close pull request then and will assign you the issue that asks to remove this action. I have no permissions

PR #1496, opened 2015-05-18
Implicit "key=value" command format support for action aliases

Kami:
This pull request adds support for implicit "key=value" command format for action aliases.

This means that by default, if you want to make action available via ChatOps, you just need to create an alias file and you don't need to define any format strings since the "key=value" format is available by default.

For example (alias definition):

```yaml
---
name: "st2-sensors-list"
action_ref: "st2.sensors.list"
description: "List available StackStorm sensors."
```

And lets say this action takes "pack" and "limit" parameter. With this change, the following command now works out of the box:

`!st2-sensors-list pack=examples limit=2`

They way it's implemented, we merge "key=value" parameters with parameters parsed from a custom format. 

This allows users to mix and match parameters and use commands like this:

Format:

```yaml
formats:
  - "actions from {{pack}}"
```

Command:

`list actions from examples limit=2`

I also plan to open a workroom pull request which documents this default format. Eventually we should also move some of those docs into this repo.

Related PRs:

* https://github.com/StackStorm/hubot-stanley/pull/10

screenshot from 2015-05-18 21 17 50

### DIFF
--- a/st2api/st2api/controllers/exp/aliasexecution.py
+++ b/st2api/st2api/controllers/exp/aliasexecution.py
@@ -66,7 +66,10 @@ class ActionAliasExecutionController(rest.RestController):
         return (tokens[0], tokens[1] if len(tokens) > 1 else None)
 
     def _extract_parameters(self, action_alias_db, param_stream):
-        alias_format = action_alias_db.formats[0]
+        if action_alias_db.formats:
+            alias_format = action_alias_db.formats[0]
+        else:
+            alias_format = None
         parser = action_alias_utils.ActionAliasFormatParser(alias_format=alias_format,
                                                             param_stream=param_stream)
         return parser.get_extracted_param_value()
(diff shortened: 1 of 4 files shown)


manasdk:
Looks like we will not support

format = `somestuff {{a}} more stuff {{b}}`
chat = `somestuff a=foobar more stuff coobar`

param = {'a':'foobar', 'b': 'coobar'}

i.e. no mixing?

manasdk:
Ah looks like for the k,v approach the format is pointless. Hmm, I am little unsure as it will lead to some opaqueness in the help string since the user will find it hard to figure out the input.

Kami:
@manasdk I believe this should work since we merge the parameters together (key=value params + format params).

In any case, that's a good point - we don't have any test cases for mixing right now. I will add them.

Kami:
@manasdk That's something we discussed on Slack.

In general I'm on favor of being explicit, but @jfryman convinced me it's fine to use this approach here. I personally also like it since commands now work out of the box and user doesn't need to specify any format (and it's also consistent with the existing hubot approach).

In any case, I do agree that this needs to be documented and obvious to the user.

manasdk:
@Kami 

Yeah the missing test threw me off and since the token don't seem to be consume I suspect the default parsers might just use up the param stream and mess things up. So tests would be nice.

On the explicit approach I am ok with what we have here - honestly IDK what really works so more features and broader scope seems good :P.

PR #1497, opened 2015-05-19
Implement synchronization lock for distributed workers

m4dcoder:


m4dcoder:
This is pre-requisite for implementing concurrency between multiple schedulers.  The current design is to have the scheduler acquire a lock with appropriate naming for the action (and/or target) prior to scheduling. With the lock, the scheduler queries the DB to determine whether to schedule or not.  Once the action is scheduled or delayed, the lock will be released.  The lock has an expiration. So if the scheduler dies, the lock will auto expire.  I also decided to use the same backend DB to implement the lock so to avoid adding more dependencies.

st2common/st2common/models/db/__init__.py
```diff
@@ -59,6 +60,13 @@ def db_teardown():
     mongoengine.connection.disconnect()
 
 
+def convert_to_python(instance):
```
manasdk: docs please.

st2common/st2common/models/db/__init__.py
```diff
@@ -121,12 +129,15 @@ def aggregate(self, *args, **kwargs):
         return self.model.objects(**kwargs)._collection.aggregate(*args, **kwargs)
 
     @staticmethod
+    def add(instance):
+        instance.save(force_insert=True)
```
manasdk: why force_insert=True

st2common/st2common/exceptions/db.py
```diff
@@ -20,6 +20,10 @@ class StackStormDBObjectNotFoundError(StackStormBaseException):
     pass
 
 
+class StackStormDBObjectFoundError(StackStormBaseException):
```
lakshmi-kannan: The name doesn't really suggest it should an error. Maybe something like StackStormDBObjectAlreadyExistsError?

st2common/st2common/models/db/__init__.py
lakshmi-kannan: Can be _ too.

st2common/st2common/util/synchronization.py
```diff
@@ -0,0 +1,66 @@
...
+from st2common.exceptions import synchronization as sync_exc
+from st2common.models.db import synchronization as models
+from st2common.persistence import synchronization as persistence
+from st2common.util import system_info
+
+
+LOG = logging.getLogger(__name__)
+
+
+class Lock(object):
```
lakshmi-kannan: Perhaps this should be called DBLock or something like that to be less confusing?

st2common/st2common/util/synchronization.py
```diff
@@ -0,0 +1,66 @@
...
+
+            eventlet.sleep(1)
+
+        LOG.debug('Acquired lock for %s.', name)
+
+    def release(self):
+        try:
+            persistence.Lock.delete(self._lock)
+        except db_exc.StackStormDBObjectFoundError:
+            raise sync_exc.LockReleaseError('Lock still exists.')
```
manasdk: Lock exists?

st2common/st2common/util/synchronization.py
```diff
@@ -0,0 +1,66 @@
...
+LOG = logging.getLogger(__name__)
+
+
+class Lock(object):
+
+    def __init__(self, lock):
+        self._lock = lock
+
+    @classmethod
+    def acquire(cls, name, expires=60, timeout=180):
```
manasdk: this impl could lead to starvation. It will be hard to get this impl right - have you looked at some distributed lock implementations? 

They are notoriously hard to get right.

manasdk:
Could/Did you look at https://github.com/vaidik/sherlock?

Can you tell us why something like that does or does not work for us?

st2common/st2common/util/synchronization.py
```diff
@@ -0,0 +1,66 @@
...
+        lock = models.LockDB(name=name,
+                             owner=uuid.uuid4().hex,
+                             proc_info=system_info.get_process_info(),
+                             expiry=(datetime.datetime.utcnow() +
+                                     datetime.timedelta(seconds=expires)))
+
+        dt_timeout = datetime.datetime.utcnow() + datetime.timedelta(seconds=timeout)
+
+        while True:
```
lakshmi-kannan: I am not feeling positive about this while True loop. If you think about acquiring 10s or 100s of locks this while loop will kill performance.

st2common/st2common/util/synchronization.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def __init__(self, lock):
+        self._lock = lock
+
+    @classmethod
+    def acquire(cls, name, expires=60, timeout=180):
+        lock = models.LockDB(name=name,
+                             owner=uuid.uuid4().hex,
+                             proc_info=system_info.get_process_info(),
+                             expiry=(datetime.datetime.utcnow() +
```
lakshmi-kannan: Is this expiry used anywhere? I don't see any garbage collection for expired objects?

st2common/st2common/util/synchronization.py
```diff
@@ -0,0 +1,66 @@
...
+                LOG.warn('Unable to acquire lock for %s because it is in use.', name)
+
+            if dt_timeout <= datetime.datetime.utcnow():
+                raise sync_exc.LockTimeoutError('Timed out waiting to acquire lock for %s.', name)
+
+            eventlet.sleep(1)
+
+        LOG.debug('Acquired lock for %s.', name)
+
+    def release(self):
```
lakshmi-kannan: I checked at python lock implementation and their APIs. acquire() is not a class method. It would be good to be consistent with the python Lock APIs.

lakshmi-kannan:
Other than my mostly cosmetic comments, I do think the lack of transactions will cause false locks and therefore make this unreliable. I don't believe NoSQL datastores are good solutions for locking. Perhaps you have found a way out?

Kami:
I agree with @lakshmi-kannan. It doesn't look like MongoDB exposes necessary primitives we need to implement actual reliable distributed locking on top.

If we want to implement actual reliable and working distributed locking we should use something like ZooKeeper. There are also other "simpler" (to run and integrate with) options out there, but they have different and usually less available HA properties.

Redis and memcached is one of those "simpler" options (you can implement safe distributed locking on top), but if we are introducing a new tool / technology it probably makes sense to introduce something with better scaling and HA guarantees (e.g. ZooKeeper, or perhaps etcd).

In any case, distributed locking is something we will need in the future in other places as well so, imo, it makes sense to invest some time and do it right.

P.S. Cassandra has had "lightweight transactions" for a while now so you can actually implement safe distributed locking on top of Cassandra now.

st2common/st2common/models/db/__init__.py
m4dcoder: This is to ensure mongodb document is inserted and not updated. The idea here is that if the doc exists then the lock is in use.  We don't want an accidental update here.

st2common/st2common/exceptions/db.py
m4dcoder: Sure

st2common/st2common/models/db/__init__.py
m4dcoder: Uh sure.  If didn't have docs to begin with.  I just consolidated the few lines of code here.

st2common/st2common/util/synchronization.py
m4dcoder: I disagree here. The confusion is that the mechanism is implemented with the backend DB.

st2common/st2common/util/synchronization.py
m4dcoder: Maybe this is confusing due to the naming of the exception object?  But if the mongoDB doc cannot be deleted, this means we are not able to release the lock.

st2common/st2common/util/synchronization.py
m4dcoder: That's why the lock is expiring.  I will gladly take a look at recommendations.

m4dcoder:
In this design, acquiring a lock = creating a doc (with a specific name). The assumption here is that insert (with unique key) is an atomic operation and that if trying to insert a doc with the same key, then exception will be thrown.  I'm open to something like zookeeper/redis/memcache. But this means increasing the footprint of st2.  I'm good with cassandra but only if replacing our DB backend otherwise this makes no sense.

st2common/st2common/util/synchronization.py
m4dcoder: Expiry is done automatically by MongoDB.  See the LockDB definition in the index section.

st2common/st2common/util/synchronization.py
lakshmi-kannan: The confusion happens because the name Lock is in python stdlib too. The stdlib implementation is an in memory lock which is cheap to acquire. Anyways, not a big deal.

m4dcoder:
@manasdk: I reviewed sherlock. it works but it requires dependencies like redis/memcache currently. and my premise here is to avoid increasing st2 footprint.  so if i were to implement a backend using the same DB backend as st2 (mongoDB), i would be implementing a plugin in sherlock and that would not be far off from this currently implementation. but if we are ok with adding dependencies to st2 install and let's just use redis/memcache/zookeeper, then i have no problem using sherlock if the backend is supported.

st2common/st2common/util/synchronization.py
lakshmi-kannan: Ahh got it. It would still appear that you need to check on a lock object get that it is not expired for safety. The indexer works async and there is no way it deletes the object on exact expiry time.

lakshmi-kannan:
>The assumption here is that insert (with unique key) is an atomic operation 

That is true only if the db provided transactions support. Hence the question of reliability.

m4dcoder:
@lakshmi-kannan: Please review http://docs.mongodb.org/manual/core/write-operations-atomicity. If I'm reading this right, a recommendation on concurrency control is to use unique index which is what I'm using here.

st2common/st2common/util/synchronization.py
m4dcoder: The python Lock API https://docs.python.org/2/library/threading.html#lock-objects is referring to a single lock (maybe a singleton within the process).  I'm not sure how I can follow that in this implementation with class methods only (and not args).  But I'm using the acquire and release method names here.

st2common/st2common/util/synchronization.py
m4dcoder: Ok. I'll think of another name.

st2common/st2common/util/synchronization.py
m4dcoder: Yeah.  The expiry here is just a safe guard in case the worker dies and to minimize deadlocks with other workers.

m4dcoder:
The other option in terms of library is https://github.com/openstack/tooz. I probably prefer this as there are more contributors.

PR #1498, opened 2015-05-19
Document how to pass data between action-chain workflows

Kami:
The title says it all.

### DIFF
--- a/docs/source/actionchain.rst
+++ b/docs/source/actionchain.rst
@@ -91,19 +91,6 @@ The input parameter ``input1`` can now be referenced in the parameters field of 
 by `Jinja templating <http://jinja.pocoo.org/docs/dev/templates/>`__.
 Similar constructs are also used in :doc:`Rule </rules>` criteria and action fields.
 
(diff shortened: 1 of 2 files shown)


docs/source/actionchain.rst
```diff
@@ -127,23 +113,111 @@ All variables are referred with Jinja syntax.
...
+use data from one workflow in another, the most common approach to that is using
+built-in key value datastore.
+
+Inside the first workflow you store data in the datastore and inside the second
+workflow you retrieve this data from a datastore.
+
+This approach creates more tight coupling between two workflows and makes them
+less re-usable and harder to run independently of each other. Because of that,
+you are encouraged to (where possible) design the workflow in a way so you can
+pass data using action parameters.
```
lakshmi-kannan: It might be worth to mention that these action parameters can use values from key value store. Or do you think it will be too confusing?

lakshmi-kannan:
LGTM.

manasdk:
+1

PR #1500, opened 2015-05-19
Add support for whitespaces in the action alias command names

Kami:
The title says it all.

In addition to that, tokenzation (splitting command string into command name and arguments) now also happens on the client and this data gets sent to the API endpoint.

See related PR for more information.

Related PRs:

* https://github.com/StackStorm/hubot-stanley/pull/11

### DIFF
--- a/st2common/st2common/models/api/action.py
+++ b/st2common/st2common/models/api/action.py
@@ -452,6 +452,11 @@ class AliasExecutionAPI(BaseAPI):
                 "description": "Name of the action alias.",
                 "required": True
             },
+            "arguments":{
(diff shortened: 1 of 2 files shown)


st2common/st2common/models/api/action.py
```diff
@@ -452,6 +452,11 @@ class AliasExecutionAPI(BaseAPI):
                 "description": "Name of the action alias.",
                 "required": True
             },
+            "arguments":{
```
Kami: Maybe `parameters` is a better name? Open to suggestions.

PR #1502, opened 2015-05-20
Implement coordination service for distributed workers

m4dcoder:


m4dcoder:
This is an alternate implementation of https://github.com/StackStorm/st2/pull/1497 that uses tooz coordination lib. (https://github.com/openstack/tooz).  Redis server is used for testing here since the service is supported in travis-ci.

manasdk:
I prefer this for sure. I guess we better also make sure that people don't have concerns with redis. I am perfectly happy with this solution.

jfryman:
Neat. 

I do understand the need for this, but it introduces another element into the stack. Two questions:

1) Can this have a feature flag associated with it? I know we're under the gun to get something to CITIES, so I don't want to block, but I do want to discuss the architecture change being proposed here. In that sense, could this be a toggable feature, disabled by default, and then enabled where we need it.

2) Is the idea here to use this lib so we have some options with implementation? Looks like we could interchange Redis for ZK in environments where we need additional stability/resiliency. Is this a library that would allow us to eventually collapse to a known element like psql? The lib says things are only partially implemented for psql. Could the calls you do today already integrate with PSQL right now, such that this is something we could 'dark ship' for CITIES with Redis, and then

jfryman:
To be clear, I like this. I just sort of want to trade a bit of debt to add this into the stack so the balance of the world remains somewhat stable.

m4dcoder:
Right now, the distributed lock is only used when a concurrency policy is defined for an action and there are multiple schedulers. I also think we can create a core action for this feature and we can replaces something like the mistral lock in our cicd process. To answer your question on feature flag, it's possible.  But then what does this means? Not installing the dependencies? Or installing them but keeping them dormant?

Here's the feature compatibility list for the different tooz drivers.  http://docs.openstack.org/developer/tooz/compatibility.html.  Maybe I should switch to memcached for unit testing since it supports all three features (grouping, leaders, and locking). So on psql, it only supports locking so if our directions is to consolidate to a single backing, this means we will have to contribute to the psql drivers for the features we need.

dzimine:
Are we using anything but get_lock()? If so, can we use file_driver, and not get extra? 

DZ.

manasdk:
>  file_driver, and not get extra?

file is limited to a single machine. I assume we design it to be used with scheduler/worker on separate machines.

dzimine:
I am echoing James question with a proposal: 
We have 2 use cases: 
1) CITIES and the like - distributed environment, throttling needed, complexity acceptable
- solution - likely Zookeeper
2) All in one installation - avoid introducing another element into the stack, while still be able to run the tests - solution - here where I think file-driver may be better than Reddis or memcache.

m4dcoder:
By default for unit testing, the "zake" driver is used. It's a fake client w/ a fake storage for zookeeper and it doesn't require an actual zookeeper server. For production deployment, users still have the configurable options to choose one of the available backends supported by tooz.

manasdk:
I feel longer term the right answer is zookeeper - it is horrible to maintain/operate but it is battle tested for these purposes.

Kami:
I also prefer this approach.

As noted on Slack the other day, to begin with, we can ship with a Redis backend by default (redis is a relatively small and simply dependency and works fine in this case).

If someone wants more resiliency and better HA properties, they can switch to an other backend such as ZooKeeper.

As far as file backend goes - I'm pretty sure it only works if all the services which require coordination (use locking) run on the single server. This won't work in distributed deployments and in our case where users can and want to scale and run action runner on multiple servers.

jfryman:
Just so I am clear. How will this work once merged? Will existing users need to get Redis installed as soon as they upgrade, or is it opt in?

m4dcoder:
I like to hear from @jfryman and @DoriftoShoes on prod deployment. We can add a flag to disable the install of zookeeper/redis.  If disabled and if users want it post install, users will have to install one of the backend that tooz supports (i.e. zookeeper/redis) and then they have to update the st2 configuration file and modify the url to the appropriate endpoint. But if users do nothing, for the throttling policy implementation specially, I'm planning to throw a LOG warning stating that the concurrency is best effort since the coordination service is not configured.

st2stanley:
I really like that plan. Specifically: * The feature degrades gracefully. We are not increasing the operational burden unless a user explicitly   

* A flag exists to optionally enable/disable the code path. This is a great practice to ship code that is early in feature, get it into the code base to continue iterating, and provide a fail safe in the event the feature does not degrade gracefully.  I would make three suggestions on top of this.  1) Make it opt-in. Even though it degrades gracefully, I would like to make it something that is explicit. This will help when we do troubleshooting early on, because we can quickly reason if the feature is enabled or not. This also notes that users take a moment to learn about the feature via docs to get started. After we have this battle tested a bit, we can re-evaluate the default config.   Also a good habit to get into as we introduce feature flags to our architecture.    2) Add some documentation on how to configure both ZK and Redis configuration options in the config file.   3) even with Opt-in, I still like the idea of throwing LOG for best-effort if it is not configured. Add a link to the docs from suggestion 2, And it'd be solid.    -James  

On May 21, 2015 at 10:04:20 AM PDT, W Chan <notifications@github.com> wrote:I like to hear from @jfryman and @DoriftoShoes on prod deployment. We can add a flag to disable the install of zookeeper/redis. If disabled and if users want it post install, users will have to install one of the backend that tooz supports (i.e. zookeeper/redis) and then they have to update the st2 configuration file and modify the url to the appropriate endpoint. But if users do nothing, for the throttling policy implementation specially, I'm planning to throw a LOG warning stating that the concurrency is best effort since the coordination service is not configured.  —Reply to this email directly or view it on GitHub.

st2stanley:
Does merging mean you agree? :)

-James

On May 21, 2015 at 10:44:45 AM PDT, W Chan <notifications@github.com> wrote:Merged #1502.  —Reply to this email directly or view it on GitHub.

m4dcoder:
Yeah, I agree with what you laid out. I didn't really see any objections here so I merged.  I will incrementally implement those items (flag, zk/redis install, docs, etc.).

st2stanley:
:metal: thanks!

-James

On May 21, 2015 at 10:50:39 AM PDT, W Chan <notifications@github.com> wrote:Yeah, I agree with what you laid out. I didn't really see any objections here so I merged. I will incrementally implement those items (flag, zk/redis install, docs, etc.).  —Reply to this email directly or view it on GitHub.

PR #1507, opened 2015-05-21
RFR: Move sensor type creation code to st2common

lakshmi-kannan:
Unfortunately all the code to convert a sensor metadata file to a sensor db model lived in st2reactor. This is because we didn't hookup SensorType controller completely. 

With this PR, I am creating a to_model in SensorTypeAPI and moving all the code required to create the sensor db model to s2common. 

You'll see changes in sensorsregistrar which now creates a SensorTypeAPI from metadata which is then converted to db model and then written to disk. This is consistent with rules and actions and aliases. 

## Note

I split the reactor api models into sensor api models and trigger api models. I had to do this because there turned out to be a cyclic dependency. Anyways, I think this is how it should always have been. 

We should also split db models and persistence models as well but it maybe beyond the scope of this PR. 

See https://github.com/StackStorm/st2/pull/1512. 

## TODO
* [X] - Move tests
* [X] - Tests should pass
* [X]  - Add tests for sensor_type_utils
* [X] - Validate sensor model in master and sensor model with this branch are identical. 
         (See https://gist.github.com/lakshmi-kannan/f763450b473db343c6a6)
* [ ] - Merge https://github.com/StackStorm/st2/pull/1511 to this branch.


manasdk:
Did not dig through details.

LGTM.

PR #1508, opened 2015-05-21
Support for improved ChatOps command format notation

Kami:
This pull request adds support for a more flexible and future-proof format for specifying ChatOps command.

`name` attribute is now only used as a primary key for a particular alias and the command format is specified using `formats` attribute.

This notation is more flexible and allows users to group related formats together in a single metadata file. Previously, if a user wanted to use formats which are specified bellow, they would need to create multiple alias metadata files.

Note: Getting this to work required moving more logic into the client which resulted in more coupling between the client and the server. Sadly there is no way around it if we want to support this notation with the current hubot approach.

Lets say we have the following alias definition:

```yaml
---
name: "st2_sensors_list"
action_ref: "st2.sensors.list"
description: "List available StackStorm sensors."
formats:
    - "list sensors"
    - "list sensors from {{ pack }}"
    - "sensors list"
```

With the definition above, the following commands now work:

```
!sensors list
!list sensors
!sensors list pack=examples
!list sensors from examples
!list sensors from examples limit=2
...
```

Related PRs:

* https://github.com/StackStorm/hubot-stanley/pull/14
* https://github.com/StackStorm/st2contrib/pull/212

### DIFF
--- a/st2api/st2api/controllers/exp/aliasexecution.py
+++ b/st2api/st2api/controllers/exp/aliasexecution.py
@@ -17,7 +17,6 @@ import jsonschema
 import pecan
 import six
 
-from mongoengine import ValidationError
 from pecan import rest
 from st2common import log as logging
 from st2common.models.api.base import jsexpose
@@ -26,7 +25,6 @@ from st2common.models.db.action import LiveActionDB, NotificationSchema, Notific
 from st2common.models.utils import action_alias_utils, action_param_utils
 from st2common.persistence.action import ActionAlias
 from st2common.services import action as action_service
-from st2common.exceptions.db import StackStormDBObjectNotFoundError
 
 
 http_client = six.moves.http_client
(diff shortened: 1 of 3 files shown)


manasdk:
This was an approach I initially stayed away from. Sort of tried to make hubot dumb i.e. it just see a text string that seems like st2 would be interested in and then pushes it for st2 to fully process.

The downside of doing it this way is that any text based interface like twitter, sms etc cannot push an aliasexecution down to st2. The argument could be made that those should/can only be integrated via sensors and the sensor could handle the mappings.

Long story short I see why this needs to be done this way and we loose some capabilities but I guess it is better for chatops integration.

+1

Kami:
Yeah, I'm also all for making the client (hubot) dumb and handle all / most of the logic on the server.

In any case, this approach works right now and we now have an action alias format which should be more future proof. Moving to the server model means that the external API (action alias metadata file and how the user communicates with hubot) will stay the same and we just need to change the internals aka implementation details.

I will open a new ticket to make sure we eventually do this, but it's not a big priority right now.

PR #1512, opened 2015-05-22
RFR - Move db, api, persistence models to individual files

lakshmi-kannan:
I couldn't control my OCD and renamed

access -> auth
datastore -> keyvalue [not keyvaluestore because models.db.keyvalue]


Kami:
LGTM so far.

PR #1514, opened 2015-05-24
[cli] Skip authentication if auth server is not available, improve workflow error reporting formatting

Kami:
This pull request includes 3 fixes / improvements:

1. If auth server is unavailable when we try to authenticate we now assume auth is disabled and still try to proceed with the request to the API service without a token. If auth is indeed enabled, request to the API service will fail. Previously, if credentials were specified in `~/.st2/config` and the auth service was unavailable (disabled) CLI would explode.

2. Improve workflow error formatting - there were some edge cases with Mistral errors where CLI would explode and not report the error correctly.2

3. Update tests so they don't throw if the CLI config file is present.


st2client/tests/base.py
```diff
@@ -86,6 +86,8 @@ def setUp(self):
             if var in os.environ:
                 del os.environ[var]
 
+        os.environ['ST2_SKIP_CLI_CONFIG'] = '1'
```
lakshmi-kannan: Is this needed only for tests? I am not seeing a genuine use case outside tests?

lakshmi-kannan:
LGTM.

st2client/st2client/shell.py
```diff
@@ -29,6 +29,7 @@
 import traceback
 
 import six
+import requests
```
lakshmi-kannan: Nit: alpha order.

st2client/tests/base.py
Kami: Heh, that's actually kinda related to the auth change we talked about.

In short - there are scenarios where you want don't want to use the CLI config (e.g. you only want to use values provided as the command line arguments).

Without this option, you need to "temporary" remove or move the config.

lakshmi-kannan: >Without this option, you need to "temporary" remove or move the config.
Yeah, that's more explicit and will be the thing with least surprises. People will forget to reset env variables.

Kami: Environment variables can also be specified on per command invocation basis. In fact, that's how I envisioned people would use it.

An alternative would be adding ``--skip-config`` flag. I'm fine with that.

lakshmi-kannan: I'd leave the env variable thing undocumented for now. I am pretty sure people will abuse it by using `export` and will complain on IRC. I do like --skip-config.

Kami: fd2130b6257740844867db3ea5939b705847079b

PR #1516, opened 2015-05-24
 Support for notifications in Mistral

Kami:
This pull request adds support for notifications in Mistral workflows. It works in exactly the same way it works in the action runner.

I also noticed `mistral-v1` runner is still registered even though we removed the Python module quite a while ago. I removed the runnersregistrar entry and all the references to it.

Note: This PR is based on #1514.

## Related PRs

* https://github.com/StackStorm/st2mistral/pull/8


st2client/st2client/shell.py
```diff
@@ -241,6 +242,10 @@ def get_client(self, args, debug=False):
             try:
                 token = self._get_auth_token(client=client, username=username, password=password,
                                              cache_token=cache_token)
+            except requests.exceptions.ConnectionError as e:
+                LOG.warn('API server is not available, skipping authentication.')
+                LOG.exception(e)
+                return client
```
lakshmi-kannan: Is this intentional? Why does this return a client?

Kami: Yes, we still want to continue.

I was thinking about it, but bubbling the exception up instead of continuing makes it harder to test and switch between servers which have auth enabled and disabled.

lakshmi-kannan: I do see it. As a client, we have no idea if server is auth enabled or not. It looks like there needs to a handshake at the start to discover server capabilities. I am not proposing we do it now but it looks like we might soon need it.

lakshmi-kannan: That said, I am curious how would client.* operations would work if auth server died but auth is enabled in server. We'd keep getting 401s back?

Kami: It doesn't matter as long as you have a valid token. If the auth is down, but you have a valid token, all the operations will still work until the token expires.

lakshmi-kannan:
LGTM

PR #1522, opened 2015-05-26
[STORM-1339] Alias docs

manasdk:
Doc for aliases. This PR is general and not ChatOps specific but does liberally callout to ChatOps for example purposes.

This PR starts us off with having some docs for Action Alias. As we notice more behaviors and details or answer user question will add more content to the docs.


jfryman:
Nice

Kami:
:+1:

PR #1523, opened 2015-05-27
[STORM-1344] remove stormbot_say

manasdk:
* stormbot_say is no longer used as all that functionality has been moved
  over to the hubot pack in st2incubator.


Kami:
LGTM, +1

PR #1525, opened 2015-05-27
Ability to refire a trigger

manasdk:
* Need to have access to id of the trigger instance.
* script needs access to mongodb and rabbitmq

Usage
```
$ st2reactor/bin/st2-trigger-refire -t 5566588332ed3528eccc8a1d
'Trigger re-fired'
```

```
$ st2reactor/bin/st2-trigger-refire -h
usage: st2-trigger-refire [-h] [--config-dir DIR] [--config-file PATH]
                          [--nouse-debugger]
                          [--trigger-instance-id TRIGGER_INSTANCE_ID]
                          [--use-debugger] [--verbose] [--version]
                          [--noverbose]

optional arguments:
  --config-dir DIR      Path to a config directory to pull *.conf files from.
                        This file set is sorted, so as to provide a
                        predictable parse order if individual options are
                        over-ridden. The set is parsed after the file(s)
                        specified via previous --config-file, arguments hence
                        over-ridden options in the directory take precedence.
  --config-file PATH    Path to a config file to use. Multiple config files
                        can be specified, with values in later files taking
                        precedence. The default files used are: None.
  --nouse-debugger      The inverse of --use-debugger
  --trigger-instance-id TRIGGER_INSTANCE_ID, -t TRIGGER_INSTANCE_ID
                        Id of trigger instance
  --use-debugger        Enables debugger. Note that using this option changes
                        how the eventlet library is used to support async IO.
                        This could result in failures that do not occur under
                        normal operation.
  --verbose, -v         Print more verbose output
  --version             show program's version number and exit
  --noverbose           The inverse of --verbose
```

### DIFF
--- a/st2reactor/bin/st2-trigger-refire
+++ b/st2reactor/bin/st2-trigger-refire
@@ -1,0 +1,6 @@
+#!/usr/bin/env python2.7
+
+from st2reactor.cmd import trigger_re_fire
+
+if __name__ == '__main__':
+    trigger_re_fire.main()
--- a/st2reactor/st2reactor/cmd/trigger_re_fire.py
+++ b/st2reactor/st2reactor/cmd/trigger_re_fire.py
@@ -1,0 +1,102 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+import logging as std_logging
+import pprint
+
+from oslo.config import cfg
+
+from st2common import config as st2cfg
+from st2common import log as logging
+from st2common.models.db import db_setup
+from st2common.models.db import db_teardown
+from st2common.persistence.trigger import TriggerInstance
+from st2common.transport.reactor import TriggerDispatcher
+
+__all__ = [
+    'main'
+]
+
+CONF = cfg.CONF
+
+
+def _parse_config():
+    cli_opts = [
+        cfg.BoolOpt('verbose',
+                    short='v',
+                    default=False,
+                    help='Print more verbose output'),
+        cfg.StrOpt('trigger-instance-id',
+                   short='t',
+                   required=True,
+                   dest='trigger_instance_id',
+                   help='Id of trigger instance'),
+    ]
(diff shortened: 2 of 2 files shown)


DoriftoShoes:
Nice.  Me likey.

Kami:
:+1:

Let's play also add some documentation on how to use it.

Kami:
(thinking out loud / style nit) - maybe "re-emit" is better? :P

Kami:
Are we also going to update packaging manifest files to install it into `/usr/local/bin` (or similar)? We do that for rules tester.

jfryman:
Why is this a different behavior from `st2 execution re-run`? The difference here seems weird. I expect that I should be able to `st2 trigger re-emit`.

Also, I like `re-emit`. :smile:

DoriftoShoes:
Two use cases:
1) triggers don't always come from actions.  It could be from a sensor.
2) In the actiontrigger case, you don't always want to re-run the action to debug why a rule isn't firing.  The specific use case this was made for was debugging why the prod pkg rules didn't fire after the st2_deploy_test.  The deploy test workflow takes 15+ minutes so we really just needed the trigger.

jfryman:
:cool:. Good reasons. Nice to add to :scroll:.

Does that preclude the case for when we might need to `trigger re-emit` and have the action re-run as well? I still think that might come in handy as well. I'll log/track unless someone has comments otherwise.

DoriftoShoes:
`trigger re-emit` and have the action run would just be an `re-run` of the original execution, no?  It really depends on where you need to restart from.  I think a combo of re-run and this (or re-emit) should give us a good entry point anywhere in the chain.

manasdk:
> `re-emit`

Yes. I prefer this as well so will change name in a separate PR.

> docs + packaging changes.

I was lazy. Will do both in same new PR and actually make this operational. Still have an open question about getting trigger_instance_id from DB. No good way to get ID just yet since there is no CLI support but we do have some API.

manasdk:
> Why is this a different behavior from `st2 execution re-run`? The difference here seems weird. I expect that I should be able to `st2 trigger re-emit`.

* `st2 trigger *` - This is a space for TriggerTypes and not TriggerInstance so I would prefer `st2 triggerinstance re-emit` if at all.

* There is no proper API support to re-emit triggerinstances - not saying this cannot be done or anything but that is primarily the reason why it is not in the CLI and is an ad-hoc tool like `st2-rule-tester`.

If we decide to invest some effort in `st2 triggerinstance *` then I would vote to move this feature there. I would even make the case that it would be quite useful while debugging/operating the system.

Also, I hacked this up to help debug a particularly annoying bug so this is what you get with a 30min effort :)

manasdk:
@jfryman I can create the story around triggerinstances for CLI and track the entire feature. We can prioritize this effort separately.

jfryman:
> Also, I hacked this up to help debug a particularly annoying bug so this is what you get with a 30min effort :)

30 minute hacks ftw. :hand: 

> If we decide to invest some effort in st2 triggerinstance * then I would vote to move this feature there. I would even make the case that it would be quite useful while debugging/operating the system.

I agree we'll need this. I think we can make a good argument this will be needed pre 1.0. 

> I can create the story around triggerinstances for CLI and track the entire feature. We can prioritize this effort separately.

Agreed. Glad we have your hack, and it moves us forward. This should make it better after field testing. :)

Issue #1493, opened 2015-05-18
Fix a bug with TriggerInstance payload not able to contain special characters such as "{{"


Kami:
See #1492 for context and details.

manasdk:
Fixed on master.

PR #1527, opened 2015-05-28
Temporarily silencing rule create and get tasks

DoriftoShoes:
There is a bug with actiontrigger parsing when the nested object has Jinja templating.  This PR redirects output to /dev/null for the two tasks that have output that would trigger this bug.  This is a work-around until the platform bug is fixed.


Kami:
I believe this is the same bug as the one described in #1493.

PR #1528, opened 2015-05-28
Skip payload render

manasdk:
Payload render in cases where the payload itself contains tokens from a rule etc is incorrect and also causes failure.

Apart from that there is no real reason to render these token. If a value is to be rendered eventually that will be handled at actionexecution time anyway.

### DIFF
--- a/st2reactor/st2reactor/rules/datatransform.py
+++ b/st2reactor/st2reactor/rules/datatransform.py
@@ -13,9 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import json
 import copy
-import jinja2
 
 from st2common.constants.rules import TRIGGER_PAYLOAD_PREFIX
 from st2common.constants.system import SYSTEM_KV_PREFIX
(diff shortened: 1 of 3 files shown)


Kami:
Does this (fully) fix #1493?

lakshmi-kannan:
I am very confused about this PR. It looks like what you are saying is action can anyways render params so why do it in the rule? Why is it incorrect to render params when the rule is being applied?

manasdk:
> Does this (fully) fix #1493?
I believe so. It could still get weird if you try to use in an action execution.

manasdk:
> I am very confused about this PR. It looks like what you are saying is action can anyways render params so why do it in the rule? Why is it incorrect to render params when the rule is being applied?

Well we render execution params in rule. The execution params can reference system variables so the right context is being setup in the JinjaBasedTransformer.

Here I skip rendering of the payload not the params. The trigger payload which should be context for param rendering is not processed and etc.

Kami:
I was actually thinking about doing the same / similar change when I
initially discovered the bug, but I thought it's a backward incompatible
change which might break stuff so I haven't touched the code.

(sharing "concerns" with lakshmi here)

But if you guys think it's fine, I'm fine with it as well :)

manasdk:
> but I thought it's a backward incompatible change which might break stuff

Yes. This is backward incompatible. IMO still think we should go ahead with this as this is the safer behavior. I will update the Changelog and I guess also take the hit on IRC if any.

DoriftoShoes:
I would bet dollars to donuts the support impact of this will be minimal.  I feel that our internal use case with st2cd are edge cases.  I would be surprised if anyone else ever hit this.

Kami:
Yeah, I agree and I'm +1 because of that.

PR #1530, opened 2015-05-31
Adding console based logger configs

jfryman:
This commit adds console logger configs to each of the components
within this project. These configs are meant to be used with
containerized configurations, where log messages are piped only
to STDOUT as general practice. (See http://12factor.net/logs)

### DIFF
--- a/st2actions/conf/console.conf
+++ b/st2actions/conf/console.conf
@@ -1,0 +1,23 @@
+[loggers]
+keys=root
+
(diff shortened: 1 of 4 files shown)


manasdk:
I am ok with the change.

How does the actions console logging work? Does each action runner run in a single container?

jfryman:
@manasdk yes, each runner runs in a separate container.

PR #1536, opened 2015-06-01
Windows script args

jamiees2:
We needed to be able to pass arguments from StackStorm to PowerShell scripts, for the windows backend. This attempts(to my best understanding) to resolve that, by providing the same functionality for `windows-script` as `remote-script`.

I tried to write some automated tests, to the best of my understanding, but they might be inadequate :).

Through some testing on our server, this seems to work, and work pretty well!

P.S: Alot of the argument output is based on this http://stackoverflow.com/questions/2157554/how-to-handle-command-line-arguments-in-powershell


st2actions/st2actions/runners/windows_script_runner.py
```diff
@@ -151,6 +162,37 @@ def _run_script(self, script_path):
...
+
+        :rtype: ``str``
+        """
+        cmd_parts = []
+        if positional_args:
+            cmd_parts.append(positional_args)
+        if named_args:
+            for (arg, value) in six.iteritems(named_args):
+                arg = quote_windows(arg)
+                if value is None or (isinstance(value, (str, unicode)) and len(value) < 1):
```
manasdk: can use six.string_types

manasdk:
LGTM. Should wait for @kami to respond since he understands this code better.

jamiees2:
Neat, I forgot to add that I essentially copied the syntax for parsing arguments from `remote-shell-script`, so it is the same syntax.

st2actions/st2actions/runners/windows_script_runner.py
```diff
@@ -151,6 +162,37 @@ def _run_script(self, script_path):
...
+            cmd_parts.append(positional_args)
+        if named_args:
+            for (arg, value) in six.iteritems(named_args):
+                arg = quote_windows(arg)
+                if value is None or (isinstance(value, six.string_types) and len(value) < 1):
+                    LOG.debug('Ignoring arg %s as its value is %s.', arg, value)
+                    continue
+                if isinstance(value, bool):
+                    if value:
+                        cmd_parts.append("-%s" % (arg))
```
Kami: It's probably a good idea to make `-` prefix configurable in the same way we do it in the remote and local script runners (`kwarg_op` parameter), but in this case we should default it to `-`.

Kami:
Ideally, eventually we would refactor get argument methods and some other functionality into `WindowsShellScriptAction` class (same as we do for Linux shell commands and scripts), but I'm fine with not doing this right now in this PR.

st2actions/st2actions/runners/windows_script_runner.py
```diff
@@ -125,12 +133,15 @@ def run(self, action_parameters):
         self._log_action_completion(logger=LOG, result=output, status=status, exit_code=exit_code)
         return (status, output, None)
 
-    def _run_script(self, script_path):
+    def _run_script(self, script_path, arguments=""):
```
Kami: I would personally prefer to default this to `None` and below do something like:

```python
if arguments:
    command = 'powershell.exe %s %s' % (quote_windows(script_path), arguments)
else:
   command = 'powershell.exe %s' % (quote_windows(script_path))
```

st2actions/st2actions/runners/windows_script_runner.py
```diff
@@ -151,6 +162,37 @@ def _run_script(self, script_path):
...
+            for (arg, value) in six.iteritems(named_args):
+                arg = quote_windows(arg)
+                if value is None or (isinstance(value, six.string_types) and len(value) < 1):
+                    LOG.debug('Ignoring arg %s as its value is %s.', arg, value)
+                    continue
+                if isinstance(value, bool):
+                    if value:
+                        cmd_parts.append("-%s" % (arg))
+                    else:
+                        cmd_parts.append("-%s:$false" % (arg))
```
Kami: Minor style thing - we use single quotes to quote strings in our Python code.

Kami:
Besides the couple of comments I added, the PR looks good to me. Nice work.

On a related note - it would also be good to add some documentation to the Windows script runner section on how to use the positional on named arguments (I believe we already have some docs on how to use them with the local and remote runners).

st2actions/st2actions/runners/windows_script_runner.py
jamiees2: I'm not sure about that, as powershell only parses the `-` notation, so I'm not sure of the use cases.

On a related note, I was also thinking of titleizing the key, to better conform with powershell's style of keyword arguments, but felt that had to be discussed first.

Kami: I saw some examples online using two dashes, but if that's the case, I'm fine with not adding that.

Kami: As far as "titlelizing" parameter names go - it sounds like this could be a boolean windows script runner parameter (e.g. "titlelize_named_parameters" or something better :P).

jamiees2: I agree, that makes sense :) I'm not even sure if it is a neccessary feature, as it could always just be titleized in the YAML definition though.

jamiees2:
I'm not sure why the setup on python 2.6 fails, even though that part was never modified? Everything passes on 2.7 though.

I added array support, as we needed it, and cleaned up the code per your suggestions.

Kami:
I fixed a typo, added test case for array values (5f81e0bd0c3c2a861f6020615a01454f20c32a5f, b94a71d726fb408bd99e6033f0f98fd5a843a2a4) and merged changes into master.

Thanks.

jamiees2:
Awesome, thanks! I think this PR can be closed then, right?

Kami:
Yep, thanks.

PR #1541, opened 2015-06-03
Make outgoing email address reflect the hostname in sendmail action

jtopjian:


DoriftoShoes:
+1

Kami:
We could even make `email` parameter, but yeah, that's fine for now.

PR #1542, opened 2015-06-03
send_empty_body parameter in sendmail action

jtopjian:
Although "body" is a required parameter, results of actions can be empty
or contain all whitespace characters. Sometimes it's only desireable to
receive an email if the action actually had output (similar to cron) and
this can now be done by setting this parameter to False.

### DIFF
--- a/contrib/core/actions/send_mail/send_mail
+++ b/contrib/core/actions/send_mail/send_mail
@@ -8,9 +8,12 @@ TO=$1
 shift
 SUBJECT=$1
 shift
+SEND_EMPTY_BODY=$1
+shift
 BODY="$@"
 
-${MAIL} <<EOF
+if [[ -z "${BODY// }" && $SEND_EMPTY_BODY == 'True' ]] || [[ -n $BODY ]]; then
+  ${MAIL} <<EOF
 TO: ${TO}
 FROM: ${FROM}
 SUBJECT: ${SUBJECT}
@@ -21,6 +24,8 @@ ${BODY}
 ${FOOTER}
 
 EOF
+
+fi
 
 if [ $? -ne 0 ]; then
   echo "Failed to send mail to ${TO}"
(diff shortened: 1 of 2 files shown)


jtopjian:
I chose to have the default value "True" so it doesn't break anything already in place.

I'm not sure if changing the parameter order would mess anything up.

manasdk:
Seems reasonable to me. Should wait for @DoriftoShoes to confirm.

contrib/core/actions/send_mail/send_mail
```diff
@@ -8,9 +8,12 @@ TO=$1
 shift
 SUBJECT=$1
 shift
+SEND_EMPTY_BODY=$1
+shift
 BODY="$@"
 
-${MAIL} <<EOF
+if [[ -z "${BODY// }" && $SEND_EMPTY_BODY == 'True' ]] || [[ -n $BODY ]]; then
```
jfryman: My only thought right away ... what is the literal `true` being sent? `True`, `true` and a ton other ways to express `true` as a bareword are valid in YAML. Will this catch at least `true` and `True`?

jtopjian: That's a good question. I assumed that when a parameter is of type "boolean", there are some other internal checks of the myriad of "true" and "false" values and ultimately set the parameter to True or False, being the Python bool values. That's all a huge assumption and I was hoping someone would flag it, like has just been done :smile:

manasdk: Since st2 translates to a python boolean True/true in yaml will always turn into `True` - I am pretty sure pyyaml deals with this for us. Plus we do perform validation of values at execution time. From the REPL -

```
>>> str(True)
'True'
>>> str(False)
'False'
```

TLDR; should only end up with the literal `True` or `False`

jfryman: Neat. :+1: then

PR #1543, opened 2015-06-03
[DONT MERGE] Fix st2_deploy.sh script for Debian based systems

Kami:
This pull request fixes st2_deploy issue spotted by a person on the IRC.

In v0.9.2, the deploy script is broken for Debian based systems since downloads server contains package named `python-st2client...deb`, but the script expects package to just be called `st2client...deb`.

See:

* http://downloads.stackstorm.net/releases/st2/0.9.2/debs/current/
* http://downloads.stackstorm.net/releases/st2/0.9.1/debs/current/
* http://downloads.stackstorm.net/releases/st2/0.9.2/rpms/current/
* http://downloads.stackstorm.net/releases/st2/0.9.1/rpms/current/

(rpm package doesn't have `python-` prefix)

I believe this issue is related to our new apt repository since the package name in the apt repository also contains `python-` prefix.

This is a temporary fix for now to get v0.9.2 working. As far as long term solution and approach goes, I will wait for @DoriftoShoes input (we should also try to figure this out asap and get a final "solution" into v0.9.3).

### DIFF
--- a/tools/st2_deploy.sh
+++ b/tools/st2_deploy.sh
@@ -24,7 +24,7 @@ INSTALL_WINDOWS_RUNNER_DEPENDENCIES=${INSTALL_WINDOWS_RUNNER_DEPENDENCIES:-1}
 DOWNLOAD_SERVER="https://downloads.stackstorm.net"
 RABBIT_PUBLIC_KEY="rabbitmq-signing-key-public.asc"
 PACKAGES="st2common st2reactor st2actions st2api st2auth st2debug"
-CLI_PACKAGE="st2client"
+CLI_PACKAGE="st2client"  # TODO: Use same package name on Debian and RHEL based systems
 PYTHON=`which python`
 BUILD="current"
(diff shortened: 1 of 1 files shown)


Kami:
Actually, looks like package is also corrupted (permissions are correct) since gdebi refuses to install it:

```bash
root@test-install:~# gdebi --n /home/stanley/debs/python-st2client_0.9.2-19_amd64.deb
Reading package lists... Done
Building dependency tree        
Reading state information... Done
Building data structures... Done 
Failed to open the software package
The package might be corrupted or you are not allowed to open the file. Check the permissions of the file.
```

I will wait on @DoriftoShoes.

Kami:
On a related note - this means we should also get v0.9.3 out ASAP since 0.9.1 is broken as well and only working version in v0.9 series now is v0.9.0.

DoriftoShoes:
Something went wrong during the 0.9.2 release and the package didn't get renamed properly.  I fixed it on the downloads server and created an issue to track down the root cause.  We can close this PR as this is no longer an issue.

Kami:
Yep, thanks.

PR #1547, opened 2015-06-03
Validate parameters for system triggers during rule creation

Kami:
This pull request adds support for validating parameters for system triggers (webhook, timers) during rule creation.

Previously, rule creation would proceed normally even if user didn't provide all or provided invalid parameters for a system trigger.

For example:

```bash
(virtualenv)vagrant@vagrant-ubuntu-trusty-64:/data/stanley$ python st2client/st2client/shell.py rule create rule.yaml 
MESSAGE: Additional properties are not allowed (u'until' was unexpected)

Failed validating 'additionalProperties' in schema:
     'properties': {'delta': {'type': 'integer'},
                    'timezone': {'type': 'string'},
                    'unit': {'enum': ['weeks',
                                      'days',
                                      'hours',
                                      'minutes',
                                      'seconds']}},
     'required': ['unit', 'delta'],
     'type': 'object'}

On instance:
    {u'delta': 4, u'until': u'hours'}
```

In the future, we could and should also turn this on for user created triggers (trigger types) which specify parameters schema in the metadata file. Currently, most of the triggers in st2contrib/ already specify it, but a lot of it is out of date or using wrong syntax since we don't actually really validate or use it.

### DIFF
--- a/st2common/st2common/models/api/rule.py
+++ b/st2common/st2common/models/api/rule.py
@@ -164,7 +164,7 @@ class RuleAPI(BaseAPI):
         # Validate criteria
         validator.validate_criteria(model.criteria)
 
-        # Validate triggerd
+        # Validate trigger parameters
         validator.validate_trigger_parameters(trigger_db=trigger_db)
 
         model.action = ActionExecutionSpecDB()
--- a/st2common/st2common/models/api/rule.py
+++ b/st2common/st2common/models/api/rule.py
@@ -160,7 +160,13 @@ class RuleAPI(BaseAPI):
         model.criteria = dict(getattr(rule, 'criteria', {}))
         model.pack = str(rule.pack)
         model.ref = ResourceReference.to_string_reference(pack=model.pack, name=model.name)
+
+        # Validate criteria
         validator.validate_criteria(model.criteria)
+
+        # Validate triggerd
+        validator.validate_trigger_parameters(trigger_db=trigger_db)
+
         model.action = ActionExecutionSpecDB()
         model.action.ref = rule.action['ref']
         model.action.parameters = rule.action['parameters']
--- a/st2common/st2common/validators/api/reactor.py
+++ b/st2common/st2common/validators/api/reactor.py
@@ -16,7 +16,7 @@
 import six
 
 from st2common.exceptions.apivalidation import ValueValidationException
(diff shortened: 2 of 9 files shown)


st2common/st2common/models/api/rule.py
```diff
@@ -160,7 +160,13 @@ def to_model(cls, rule):
         model.criteria = dict(getattr(rule, 'criteria', {}))
         model.pack = str(rule.pack)
         model.ref = ResourceReference.to_string_reference(pack=model.pack, name=model.name)
+
+        # Validate criteria
         validator.validate_criteria(model.criteria)
+
+        # Validate triggerd
```
manasdk: validate trigger parameters?

Kami: Yeah, that was definitely some fat fingers there :P

st2common/st2common/validators/api/reactor.py
```diff
@@ -36,3 +45,29 @@ def validate_criteria(criteria):
...
+
+    trigger_type_ref = trigger_db.type
+    parameters = trigger_db.parameters
+
+    if trigger_type_ref not in SYSTEM_TRIGGER_TYPES:
+        # Not a system trigger, skip validation for now
+        return None
+
+    parameters_schema = SYSTEM_TRIGGER_TYPES[trigger_type_ref]['parameters_schema']
+    VALIDATOR(parameters_schema).validate(parameters)
```
manasdk: programming with exception!

works for me :)

manasdk:
+1

st2common/st2common/validators/api/reactor.py
```diff
@@ -36,3 +45,29 @@ def validate_criteria(criteria):
...
+    :param trigger_db: Trigger DB object.
+    :type trigger_db: :class:`TriggerDB`
+    """
+    if not trigger_db:
+        return None
+
+    trigger_type_ref = trigger_db.type
+    parameters = trigger_db.parameters
+
+    if trigger_type_ref not in SYSTEM_TRIGGER_TYPES:
```
manasdk: We could support non system triggers also if they do have parameters.

Kami: Yep - see my note in PR.

We would need to go over trigger definitions in st2contrib since I believe a lot of them are out of date or wrong.

PR #1552, opened 2015-06-04
Also start WebUI HTTP server in launchdev

Kami:
Now we also start WebUI HTTP server in launchdev script if the WebUI directory exists.

### DIFF
--- a/tools/launchdev.sh
+++ b/tools/launchdev.sh
@@ -1,8 +1,12 @@
 #!/bin/bash
 
-runner_count=1
+RUNNER_COUNT=1
+WEBUI_PORT=8080
+WEBUI_LOG_FILE="/var/log/st2/st2web.log"
+WEBUI_DIR="/opt/stackstorm/static/webui"
+
 if [ "$#" -gt 1 ]; then
-    runner_count=${2}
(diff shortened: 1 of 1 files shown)


lakshmi-kannan:
GoGoGo

manasdk:
This is very strange. I am not sure I agree with starting up components that do not come from this repo. 

We would not want to start up a hubot in launchdev would we?

Kami:
I think it's reasonable to do it with the WebUI since it's a "core" component (unlike hubot which is pluggable and not core).

It was started before when the WebUI was server as part of the API service.

enykeev:
Well, we need to make it clear somehow that webui in the folder is production version. It is read only and is here to make it easier to develop st2. Only for that. For any change in st2web, you need to have a clone of st2web repo and start it with `gulp`.

Kami:
Yeah, launchdev is only meant to be used for development anyway.

I personally handle that via symlinks.

manasdk:
webui is part of API service?

Anyway, I don't like this approach but #yolo. I will not -1 or +1. If you guys want to merge go for it.

Kami:
(it was)

Closing this. @manasdk if you change your mind, let me know :P

PR #1558, opened 2015-06-04
Use background scheduler in the notifier

m4dcoder:
Use a background scheduler instead of blocking scheduler so it yields to the notifier to do work.


st2actions/st2actions/notifier.py
```diff
@@ -56,6 +56,8 @@ def __init__(self, connection, queues, trigger_dispatcher=None):
             name=ACTION_TRIGGER_TYPE['name'])
 
     def process(self, liveaction):
+        LOG.debug('Processing liveaction. %s', liveaction)
```
Kami: :+1:

PR #1562, opened 2015-06-05
Notification consistency

manasdk:
* actionchain already uses 'on-success' and similar therefore making
  sure that notification which can also exist in an actionchain task use
  'on-success'. All instances like Action & LiveAction can contain
  notfication schema have been updated for sake of consistency.
* The preferred value is 'on_success' however mistral DSL has already
  taken the approach to use '-' and not '_'. To avoid confusion and
  inconsistency to user better to stick with '-'.
* All the internal proeprties though use '_' since '-' being the 'minus'
  operator should not be used in a property name.

### DIFF
--- a/contrib/examples/actions/echochain.meta.yaml
+++ b/contrib/examples/actions/echochain.meta.yaml
@@ -15,7 +15,7 @@ parameters:
     default:
       - c2
 notify:
-  on_complete:
+  on-complete:
     message: "\"@channel: Action succeeded.\""
     channels:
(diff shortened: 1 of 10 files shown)


manasdk:
Either way something is going to end up looking weird -

https://github.com/StackStorm/st2/pull/1562/files#diff-228e7fa1c9555813e603c5931ce21797R7

lakshmi-kannan:
LGTM.

PR #1554, opened 2015-06-04
RFR: Docs and changelog for st2 triggerinstance re_emit

lakshmi-kannan:

PR #1572, opened 2015-06-06
Trigger instance typo and command name consistency fix

Kami:
See:

* https://github.com/StackStorm/st2/pull/1554#discussion_r31784141
* https://github.com/StackStorm/st2/pull/1554#discussion_r31784041


Kami:
Looks like there is some kind of race causing the tests to fail (https://travis-ci.org/StackStorm/st2/jobs/65668421) or is this because we are using a dummy locking backend? If the latter, we should still figure something out (e.g. maybe use file based backend? this should work fine for the tests which run on a single host) to make sure tests are reliable. @m4dcoder can you please have a look?

PR #1574, opened 2015-06-06
Adding ability in API and CLI to filter executions based on the trigger instance id

skarmark:
This will let users track the sequence of what happened after a particular trigger was dispatched.

### DIFF
--- a/st2api/st2api/controllers/v1/executionviews.py
+++ b/st2api/st2api/controllers/v1/executionviews.py
@@ -33,6 +33,7 @@ SUPPORTED_FILTERS = {
     'timestamp': 'start_timestamp',
     'trigger': 'trigger.name',
     'trigger_type': 'trigger_type.name',
+    'trigger_instance': 'trigger_instance.id',
     'user': 'liveaction.context.user'
 }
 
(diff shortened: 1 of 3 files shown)


Kami:
Thanks.

The CLI change looks good, but I believe we also need to update API controller (st2api/st2api/controllers/v1/executionsviews.py) to also support filtering on trigger instance id.

With regarding to the test failure (https://travis-ci.org/StackStorm/st2/jobs/65717211#L985) it seems to be the same issue as described here -https://github.com/StackStorm/st2/pull/1572#issuecomment-109541195 (/cc @m4dcoder).

skarmark:
Kami, above commit does contain change to st2api/st2api/controllers/v1/executionsviews.py as well. Did you check that? I was able to run the api and provide the trigger instance id as one of the filters when testing.

Kami:
@skarmark Ah, sorry, it looks like I missed that line :)

Kami:
I've made minor tests change (previously we weren't correctly testing filters) and merged change into trunk. Thanks.

Since it's a minor change, I will also cherry-pick it on top of v0.11.1.

skarmark:
Great. Thanks!

PR #1570, opened 2015-06-05
Pack deployment chatops

jfryman:
This commit introduces several new fixes for the existing Pack deployment
actions. Namely:

* Do not overwrite config.txt on existing pack installations
* Allow short repo_url for GitHub shortcuts (repo_url=StackStorm/st2contrib)
* Deploy from monorepo repositories and single-pack repositories
  - Introduces subtree flag, for used when packs are nested within top-level packs directory
* Tag each download with Git information
  - Used to introspect versions of deployed packs

PR #1575, opened 2015-06-06
Repair failing pack installation

jfryman:

* Double `.git` extension
* SSH protocol downloads via Git

In certain circumstances, the `repo_url` passed into the `packs.download`
action is returns a double `.git.git` extension. Contains fix to only
append `.git` suffix if doesn't already exist.


contrib/packs/actions/pack_mgmt/download.py
```diff
@@ -141,10 +141,12 @@ def _eval_subtree(repo_url, subtree):
     @staticmethod
     def _eval_repo_url(repo_url):
         """Allow passing short GitHub style URLs"""
+        has_git_extension = re.compile('.git$')
```
Kami: You can also use endswith - `has_git_extension = repo_url.endswith('.git')`

contrib/packs/actions/pack_mgmt/download.py
```diff
@@ -141,10 +141,12 @@ def _eval_subtree(repo_url, subtree):
...
         if len(repo_url.split('/')) == 2 and not "git@" in repo_url:
-            return "https://github.com/{}.git".format(repo_url)
```
Kami: I'm not that big of a fan of this implicit approach, but I know why it was done (to make chatops command more user-friendly).

Also, it looks like that right now we default to http (non ssh) URL if user doesn't provide a full url, is this correct?

contrib/packs/actions/pack_mgmt/download.py
jfryman: oh! very nice.

contrib/packs/actions/pack_mgmt/download.py
jfryman: So, it's sort of all over the place.

I saw that some of the action metadata has `https` defined, while the `download.py` had `git` as the protocol defined. In my original PR, I went toward HTTPS to avoid having git keys splayed all over the place, but I kept running into issues with prompting for HTTPS logins on clean installs.

I think we should be on HTTPS, but the change broke things. I'm looking into why I'm being prompted for auth on public repos this AM, and may end up changing this back ultimately.

contrib/packs/actions/workflows/install.yaml
```diff
@@ -25,6 +25,12 @@
       ref: "packs.load"
       params:
         register: "{{register}}"
+      on-success: "reload-aliases"
```
jfryman: I want to explicitly call this out. This enables direct deploys of ChatOps from ChatOps with the single command. 

The `packs.load` command won`t allow me to take an array of things to register, and registering `all` is not the behavior we want (because of potentially rogue rules).

This works, and is a bit of a hack. If this makes you feel dirty, and it probably should, I'd ask for some cycles thinking about why I cannot just `all` here and be done with it. Having values that may or may not be desired to be loaded in an `all` sort of defeats the purpose. Maybe change the meaning of `all` (sensors, actions, aliases, etc). I don't know. Thoughts appreciated.

manasdk: Done in c2d22599ba2789c5a87febe841eb52424df454e8.

If we are comfortable doing this in packs.install might as well handle in st2ctl.

contrib/packs/actions/pack_mgmt/download.py
```diff
@@ -141,10 +142,12 @@ def _eval_subtree(repo_url, subtree):
...
     def _eval_repo_url(repo_url):
         """Allow passing short GitHub style URLs"""
+        has_git_extension = repo_url.endswith('.git')
         if len(repo_url.split('/')) == 2 and not "git@" in repo_url:
-            return "https://github.com/{}.git".format(repo_url)
+            url = "https://github.com/{}".format(repo_url)
         else:
-            return "{}.git".format(repo_url)
+            url = repo_url
+        return url if has_git_extension.search(url) else "{}.git".format(url)
```
manasdk: `has_git_extension` will be `True` or `False` and at that point has_git_extension.search(url) works?

I have my doubts that this works. I will test and if broken will push changes.

manasdk: Fixed in 317eb560ca1179672bfcd8d4e9bdfad44c35aadd

Kami:
I've added some tests for `_eval_subtree` and `_eval_repo_url `(d3a168b90ebf3235514f356c78e7e085eb8a2a4f, 2bf8e55b182b192d6a90221a533c6dca4647db8d). I've added a note to the code (see https://github.com/StackStorm/st2/commit/3d3cb08af8de3e4642839022547d79d238388e17#diff-7aa2154ce55d5016cfbc18ae47e1cec8R149), to me it seems like we should use OR and not XOR since we also want to return True if both evaluate to True.

It's also possible that I'm missing some context on how `subtree` argument / parameter should work since there is no description and docstring for that parameter / argument.

contrib/packs/actions/pack_mgmt/download.py
```diff
@@ -155,17 +161,23 @@ def _validate_result(result, packs, repo_url):
...
-        st2_repos = re.compile("st2(contrib|incubator)")
-        match = True if st2_repos.search(repo_url) else False
-        return subtree ^ match
+        match = False
+        for stackstorm_repo_name in STACKSTORM_CONTRIB_REPOS:
+            if stackstorm_repo_name in repo_url:
+                match = True
+                break
+
+        return subtree ^ match  # TODO: Shouldn't this be | (OR) ?
```
Kami: @manasdk I believe james is offline this week, but you might know the answer.

manasdk: OR makes sense to me as well. XOR is almost never right :P

Kami: Alright, will change it to OR and update the tests.

manasdk:
+1

PR #1576, opened 2015-06-06
Repair failing pack installation

jfryman:

* Double `.git` extension
* SSH protocol downloads via Git

In certain circumstances, the `repo_url` passed into the `packs.download`
action is returns a double `.git.git` extension. Contains fix to only
append `.git` suffix if doesn't already exist.


manasdk:
NO merge!


Need to go through https://github.com/StackStorm/st2/tree/v0.11.1

PR #1578, opened 2015-06-07
Repair packs download

manasdk:

Double .git extension
SSH protocol downloads via Git
In certain circumstances, the repo_url passed into the packs.download
action is returns a double .git.git extension. Contains fix to only
append .git suffix if doesn't already exist.

Recreating https://github.com/StackStorm/st2/pull/1576

### DIFF
--- a/contrib/packs/actions/pack_mgmt/download.py
+++ b/contrib/packs/actions/pack_mgmt/download.py
@@ -144,7 +144,7 @@ class InstallGitRepoAction(Action):
         """Allow passing short GitHub style URLs"""
         has_git_extension = repo_url.endswith('.git')
         if len(repo_url.split('/')) == 2 and not "git@" in repo_url:
-            url = "git@github.com:{}".format(repo_url)
+            url = "https://github.com/{}".format(repo_url)
         else:
             url = repo_url
         return url if has_git_extension.search(url) else "{}.git".format(url)
--- a/contrib/packs/actions/pack_mgmt/download.py
+++ b/contrib/packs/actions/pack_mgmt/download.py
@@ -142,7 +142,7 @@ class InstallGitRepoAction(Action):
     @staticmethod
     def _eval_repo_url(repo_url):
         """Allow passing short GitHub style URLs"""
-        has_git_extension = re.compile('.git$')
+        has_git_extension = repo_url.endswith('.git')
         if len(repo_url.split('/')) == 2 and not "git@" in repo_url:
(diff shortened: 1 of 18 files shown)


manasdk:
Ugh. No good, will need to cherry pick commits. Closing.

st2common/st2common/models/utils/action_param_utils.py
```diff
@@ -79,7 +79,7 @@ def is_immutable(param_meta):
     return (required_params, optional_params, immutable_params)
 
 
-def cast_params(action_ref, params):
+def cast_params(action_ref, params, cast_overrides=None):
```
Kami: It would be good to add a docstring which describes what `cast_overrides` parameter does.

contrib/packs/actions/pack_mgmt/info.py
```diff
@@ -0,0 +1,16 @@
...
+
+class PackInfo(Action):
+    def run(self, pack, pack_dir="/opt/stackstorm/packs"):
+        gitinfo = os.path.join(pack_dir, pack, GITINFO_FILE)
+        try:
+            with open(gitinfo) as data_file:
+                details = json.load(data_file)
+                return details
+        except:
+            print "Unable to load git info for {}".format(pack)
```
Kami: Hm, why don't we let exception bubble up or throw a custom one with a more friendly error message (Unable to load pack info for foo: %s)?

Not we don't let it bubble up and we don't exit with non-zero which means action will be considered as succeeded even on failure. Is this intentional?

contrib/packs/actions/pack_mgmt/download.py
```diff
@@ -115,3 +132,40 @@ def _validate_result(result, packs, repo_url):
...
         return sanitized_result
+
+    @staticmethod
+    def _eval_subtree(repo_url, subtree):
+        st2_repos = re.compile("st2(contrib|incubator)")
+        match = True if st2_repos.search(repo_url) else False
+        return subtree ^ match
+
+    @staticmethod
+    def _eval_repo_url(repo_url):
```
Kami: Some test cases for this function would be nice.

st2common/st2common/models/utils/action_param_utils.py
manasdk: Agreed. Happened during a major push to get the release out so slipped.

Next time I touch this file will get it done. 

Btw, this PR is closed so no real point in conversations here.

PR #1588, opened 2015-06-08
update index.rst

epowell101:
Added a couple of sentences because right now we go from "Start Now" to direction on how to run us on two particular Linux versions.  We THEN mention - after this - that there are other ways to start with us.

If a reader is busy and does not scan below they will not know there are although alternatives.


manasdk:
+1

dzimine:
I bet @epowell101 wants it on 0.11 branch, too? Else it won't see the light till next release...

PR #1593, opened 2015-06-08
Fix docs version selector.

dzimine:
* skip versions ending with 0 as our CI/CD don't grok them yet.

This is a hack but does the trick. I can have bug to remind us to remove it once CI/CD fixed.

### DIFF
--- a/conf/st2.dev.conf
+++ b/conf/st2.dev.conf
@@ -46,7 +46,7 @@ excludes = requests,paramiko
 redirect_stderr = False
 
 [system_user]
-user = stanley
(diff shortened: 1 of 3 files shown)


manasdk:
Would be better to just skip v0.10.

I think we know how to fix future version properly.

dzimine:
I though `prev = 9 if prev == 10 else prev` would look uglier. But ok, can do this.

manasdk:
Well either way is ugly but at least this way it is limited to 0.10 only.

PR #1597, opened 2015-06-09
RFR - Add st2 exporter - Dump executions to JSON files on disk.

lakshmi-kannan:


st2exporter/README.md
```diff
@@ -0,0 +1,3 @@
+## St2 exporter
+
+Listens to exeuction updates from RabbitMQ and creates JSON files out of them to be exported to cloud.
```
Kami: Fat fingers :P

st2exporter/st2exporter/exporter/dumper.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+class Dumper(object):
+
+    def __init__(self, queue=None, files_dir=None, file_format='json',
+                 file_prefix='st2-executions-',
+                 batch_size=1000, sleep_interval=60,
+                 max_files_per_sleep=5,
+                 file_writer=None):
+        if not queue:
```
Kami: If `queue` argument is required you can just remove a default `None` value.

Kami:
Looks like a good start :+1: 

Definitely needs tests though :P

Kami:
As far as overall design goes - currently, this only listens for new executions on the queue and we use a topic exchange, right?

This means that the existing executions which are already in the database will be ignored. Is this a feature or a known limitation right now?

lakshmi-kannan:
>This means that the existing executions which are already in the database will be ignored. Is this a feature or a known limitation right now?

What are you talking about? https://github.com/StackStorm/st2/pull/1597/files#diff-1647271472cf1ca6e536771caafb9902R78

>Definitely needs tests though :P

yes, capt. obvious. I am definitely not doing TDD :P.

PR #1599, opened 2015-06-09
Make common st2 environment variables available to the actions

Kami:
This pull request updates local, remote and python runner so the action environment now includes common StackStorm environment variables:

* `ST2_ACTION_API_URL` - Full URL to the public API endpoint.
* `ST2_ACTION_AUTH_TOKEN` - Auth token which is available during the life-time of the action.

We have already kinda did this before for remote actions (we just included an auth token), but this pull request unifies it and makes it available to all the runners.

This is something @jfryman has suggested before and we have also discussed it in the past. Having those environment variables available make many things easier - e.g. now user can easy hit the API using curl from a local or a remote action.

Note: I have explicitly added "ST2_ACTION_" prefix to the environment variable names so they don't collide with environment variables used by the client. The environment variable abuse which happens inside the client is already messy (I plan to remove it once the refactoring takes place) and I prefer people to be explicit instead of things happening magically.

For example, if a user wants to use token environment variable in the local shell script action they can do something like this:

```bash
st2 action list --token $ST2_ACTION_AUTH_TOKEN
```

or

```bash
export ST2_AUTH_TOKEN=$ST2_ACTION_AUTH_TOKEN
st2 action list
```

TODO:

* [x] Docs


manasdk:
This makes sense.

+1

lakshmi-kannan:
LGTM.

PR #1600, opened 2015-06-09
st2_deploy: Update pip

DoriftoShoes:


Kami:
LGTM

Kami:
(although some users might not like that you are messing with their system wide installation version, but I think that's fine for st2_deploy use case)

DoriftoShoes:
I think the use case for st2_deploy.sh is pretty solid.  Stand alone VMs used for demo/poc of StackStorm.  If users are super worried about pip upgrades or similar I would push them towards a more advanced installation method.

lakshmi-kannan:
@DoriftoShoes: Dude, why not easy_install pip? This pip installing pip scares me.

DoriftoShoes:
I'm following the directions given to me by pip.

PR #1604, opened 2015-06-09
Jinja filters

manasdk:
* jinja filter for regex and working with semver


Kami:
Nice, I believe that's something I've asked for in the past :)

Kami:
It would also be good to add some docs on custom filters which are available in the templates.

Kami:
To prevent regular expression ddos attacks and pathological use case we should eventually consider using something like re2 - https://pypi.python.org/pypi/re2/

manasdk:
> It would also be good to add some docs on custom filters which are available in the templates.
I got lazy after writing all those tests :package: 

But, I will add docs.

manasdk:
Docs added at 22f19deb2557c118f555612c1f9aa5ecd5451c59. Not the best but atleast filters are enumerated

PR #1605, opened 2015-06-10
Refactor Mistral to use PostgreSQL instead of MySQL

m4dcoder:


lakshmi-kannan:
This also needs a corresponding change in workroom.

tools/st2_deploy.sh
```diff
@@ -97,6 +99,22 @@ else
     MISTRAL_STABLE_BRANCH="st2-0.5.1"
 fi
 
+update_hosts_file()
+{
+  if [ -z "`cat /etc/hosts | grep -e '127.0.0.1\slocalhost'`" ]; then
```
lakshmi-kannan: Just noting in workroom, we use 127.0.1.1.

manasdk:
Are we sure about this change? I don't think we should be changing anyones /etc/hosts without notifying them - if I have some terrible reason for this configuration I would hate it if some deploy script went in an changed things!

tools/st2_deploy.sh
m4dcoder: > Are we sure about this change? I don't think we should be changing anyones /etc/hosts without notifying them - if I have some terrible reason for this configuration I would hate it if some deploy script went in an changed things!

I'm open to suggestions. I don't know how else to fix this on vagrant w/ chef/fedora-20.

lakshmi-kannan: IIUC, this basically sets localhost to 127.0.0.1. Why can't you edit to configuration for the service (postgres in this) on the box and replace localhost with whatever IP we want?

manasdk: Yeah that seems like a better approach.

m4dcoder: st2api and st2auth are the one causing problem. actually this doesn't have anything to do with the postgresql change.  i can happily remove this and let the f20 fail for now.

manasdk: so st2api and st2auth have always failed on fedora and no one has complained?

m4dcoder: Looks like it.

DoriftoShoes: It wasn't always this way.  I don't know when it changed but in the beginning it wasn't happening.

DoriftoShoes:
st2_deploy.sh is designed for standalone deployments for testing/POC purposes.  I don't have a problem with us ensuring that the localhost entry we are looking for exists on these boxes.  Now if we were doing this as part of a package install it would be a different story...but this all in one installer isn't designed to be dropped in to a production environment.

tools/st2_deploy.sh
m4dcoder: Anyway, I found a fix. The problem is in the st2client default base URL.  I'm going to set it to http://0.0.0.0 (currently it's http://localhost).  That fixed the problem.

m4dcoder: The fix is done separately @ https://github.com/StackStorm/st2/pull/1609.

PR #1606, opened 2015-06-10
Fix a bug with secrets masking

Kami:
This pull request fixes a bug which would cause the masking could to modify the original dictionary in place instead of modifying a copy of it and using a copy to interpolate the log message.


manasdk:
Nice. 

+1

PR #1608, opened 2015-06-10
Some fixes to the st2 action-alias * commands

manasdk:
* `list` -  fixed it and now works
* `get` - added attribute display order

### DIFF
--- a/st2client/st2client/commands/action_alias.py
+++ b/st2client/st2client/commands/action_alias.py
@@ -15,8 +15,6 @@
 
 from st2client.models.action_alias import ActionAlias
 from st2client.commands import resource
-from st2client.commands.resource import add_auth_token_to_kwargs_from_cli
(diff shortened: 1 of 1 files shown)


m4dcoder:
LGTM

Kami:
Both list and get worked fine for me before, wonder what broke it?

In any case, less code, LGTM.

manasdk:
> Both list and get worked fine for me before, wonder what broke it?

List was broken, it would throw an exception at
https://github.com/StackStorm/st2/blob/master/st2client/st2client/models/core.py#L234

Also, get was fine. I added the command to include to reorder the display attributes.

manasdk:
Not sure what changed for list but I did not spend much time investigating either since the fix was simple.

Kami:
Weird. Anyway, good catch and change.

PR #1609, opened 2015-06-10
Explicitly set st2client base_url in st2_deploy.sh

m4dcoder:


DoriftoShoes:
+1 this is definitely cleaner than my /etc/hosts hack.

lakshmi-kannan:
s/0.0.0.0/127.0.0.1

m4dcoder:
Ok. Fixed to 127.0.0.1.

lakshmi-kannan:
LGTM

manasdk:
+1

tools/st2_deploy.sh
```diff
@@ -515,6 +515,13 @@ install_st2client() {
   fi
   popd
 
+  # Write ST2_BASE_URL to env
+  if [[ "$TYPE" == "rpms" ]]; then
```
Kami: Hm, a bit confused about this part - do we still need it?

CLI config should also work on Fedora just fine (but yeah, unlike global bashrc we are using here, CLI config is specific to a system user).

PR #1615, opened 2015-06-11
Add support for "secret" action parameters, mask them from the logs

Kami:
This pull request adds support for a `secret` attribute to the action parameter metadata. This attribute indicates that the parameter in question contains sensitive data and should be treated as such.

Right now, that means we mask the values of the secret parameters in the logs but we also have a framework in place now so we can extend and improve on that later.

Also, keep in mind that this pull request != secret management. That's a different feature and something which will be done later / when needed.

## TODO

- [ ] Decide on secret masking in API response
- [x] Update st2exporter code to mask values (this depends on the decision above - we either need to update it to use `to_serilizable_dict` or update to `APIModel.from_model` to support masking)

## Open questions

* What else should we do with the "secret" parameters?

I imagine eventually we will also want to mask them when interacting with the API, but this makes no sense right now without RBAC / roles.


st2common/st2common/logging/formatters.py
```diff
@@ -73,7 +74,7 @@ def serialize_object(obj):
     if getattr(obj, 'to_dict', None):
         value = obj.to_dict()
     elif getattr(obj, 'to_serializable_dict', None):
-        value = obj.to_serializable_dict()
+        value = obj.to_serializable_dict(mask_secrets=True)
```
Kami: Note: I tried really hard and wanted to avoid this and keep all the masking code and related functionality inside the log formatter since this should ideally only be responsibility of the log formatter, but sadly that's not possible so some of the masking code has leaked / spilled into other places as well :/

This makes me quite unhappy, but there is really no way around that and I still did my best to keep leakage to the minium.

Kami:
As noted in one of the comments, this PR / change is making me quite unhappy.

It's not hard to get it done, but it's hard to get it done right which means making it maintainable and safe (e.g. all the masking happens in the centralized place). It's obviously impossible to make it 100% safe, but we should make it as easy as possible to do it right and as hard as possible to mess it up for the developers.

Even though (as much as possible) masking happens in a centralized place (in this case that's the log formatter), this won't help if the programmers don't use the logger correctly - this means including context in the `extra` attribute which gets automatically processed and cleaned.

Sadly that's something which is really hard to enforce in an automatic fashion so everyone will need to keep this in mind and leak for potential sensitive data leakage when reviewing PRs.

st2actions/st2actions/container/base.py
```diff
@@ -133,7 +136,7 @@ def _do_run(self, runner, runnertype_db, action_db, liveaction_db):
                 try:
                     self._delete_auth_token(runner.auth_token)
                 except:
-                    LOG.warn('Unable to clean-up auth_token.')
+                    LOG.execution('Unable to clean-up auth_token.')
```
lakshmi-kannan: Doesn't look right :P

manasdk: Lol.

st2actions/st2actions/bootstrap/runnersregistrar.py
```diff
@@ -382,7 +384,8 @@
             'password': {
```
lakshmi-kannan: I guess the evolution step would be to `detect` some keywords as secret params by default? Like password, key, etc...

lakshmi-kannan: We also do some `detection` on user supplied values. If it looks like API key, maybe we can hide it automatically.

st2actions/st2actions/container/base.py
```diff
@@ -93,10 +95,25 @@ def _do_run(self, runner, runnertype_db, action_db, liveaction_db):
...
+
+            LOG.debug('Performing pre-run for runner: %s', runner.runner_id)
             runner.pre_run()
 
-            LOG.debug('Performing run for runner: %s', runner)
+            # Mask secret parameters in the log context
+            runner_parameters_specs = runnertype_db.runner_parameters
+            action_parameters_sepcs = action_db.parameters
+
+            secret_runner_parameters = get_secret_parameters(parameters=runner_parameters_specs)
```
lakshmi-kannan: I am not too happy with the explicit calls. Maybe what we could do is create things like log_execution(), log_action() apis and these explicit calls would be wrapped inside those. Thoughts?

st2common/st2common/constants/secrets.py
```diff
@@ -0,0 +1,31 @@
...
+# Note: If an attribute is an object or a dict, we try to recursively process it and mask the
+# values.
+MASKED_ATTRIBUTES = [
+    'password',
+    'auth_token',
+    'token'
+]
+
+# Value with which the masked attribute values are replaced
+MASKED_ATTRIBUTE_VALUE = '********'
```
lakshmi-kannan: Is it worth to create as many `*s` as the number of characters in the secret? I can think this is a security problem by itself but it will be unintuitive to have different number of *s than the number of characters in the secret.

dzimine:
> I imagine eventually we will also want to mask them when interacting with the API, but this makes no sense right now without RBAC / roles.

This is orthogonal: the secrets don't appear in API regardless of roles/permissions. The def of secret is "something that goes in but never goes out". See @manasdk capture here https://github.com/StackStorm/discussions/issues/65

st2common/st2common/constants/secrets.py
dzimine: knowing the length of the secret is a hint to break it. Imagine `***` :)

st2common/st2common/models/db/liveaction.py
```diff
@@ -69,13 +70,23 @@ class LiveActionDB(stormbase.StormFoundationDB):
...
     meta = {
         'indexes': ['-start_timestamp', 'action']
     }
 
+    def to_serializable_dict(self, mask_secrets=False):
+        result = super(LiveActionDB, self).to_serializable_dict(mask_secrets=mask_secrets)
+
+        if mask_secrets:
+            # TODO: This sucks, but it's only non slow approach
+            del result['parameters']
```
lakshmi-kannan: I am confused about this. Won't this delete the `parameters` section completely?

st2common/st2common/logging/formatters.py
lakshmi-kannan: I am fine with this. I don't see a way out either. People have to tell us what params are secrets. Sure, we can add in some auto-detection but that is down the lane.

st2common/st2common/constants/secrets.py
lakshmi-kannan: Yeah, I noted that in my comment but still it's counter-intuitive. And if people really have 3 character secrets, they deserve to be hacked for sure :P.

lakshmi-kannan:
In general, I am fine with the `secrets` field approach. I think we need high level logging APIs so there is a single code point for this and we won't have to think twice when using it. We can debate this and see if we can do anything better.

lakshmi-kannan:
>This is orthogonal: the secrets don't appear in API regardless of roles/permissions. The def of secret is "something that goes in but never goes out". See @manasdk capture here StackStorm/discussions#65

I am not sure this is good user experience. I would like to know why we decided on this?

dzimine:
@Kami may be you do the part 2, "mask from API", than look back and see if this can be done differently, when you have the two cases to reason about. 

> this won't help if the programmers don't use the logger correctly 

There's a limit to what we can do. If someone print secret to the console, etc. Our goal is to give facility to hide secrets and explain how to use it. Many ways to leak them still. As a user I can do a workflow that publishes secret parameter, without applying 'secret' attribute and it will be exposed... We won't deal with this cases for now.

lakshmi-kannan:
If we have high level APIs for logging st2 first class models viz. execution, action, sensor, triggers, rules we are mostly covered. st2 contributors need some discipline in using these high level apis for logging. Perhaps we could add a lint step to detect raw logging. Kinda hard I guess. But yeah, generally, we explain with good documentation and wherever possible, we'll make sure people don't shoot themselves in the foot. Other than that, there isn't much we could do.

Kami:
@lakshmi-kannan 

> In general, I am fine with the secrets field approach. I think we need high level logging APIs so there is a single code point for this and we won't have to think twice when using it. We can debate this and see if we can do anything better.

That was the whole idea with the logger approach approach - user should include log context inside the `extra` attribute. This attribute is then processed in the formatter and we can mask the secret values.

Sadly as noted above, we can't fully contain this logic in the log formatter so it spills in other places as well.

Kami:
@dzimine @lakshmi-kannan 

> This is orthogonal: the secrets don't appear in API regardless of roles/permissions. The def of secret is "something that goes in but never goes out". See @manasdk capture here StackStorm/discussions#65

I'm slightly confused about that as well. I never seen an API which does that and I'm not too big of a fan of that approach.

This would make sense with RBAC, but without it it's confusing imo. In addition to that, it also breaks "piping API result to a file, updating the file and sending it back to the API" approach if it's used with models which can be manipulated via the API.

In any case, if we think that's the way to go, I can make the relevant changes.

lakshmi-kannan:
>In any case, if we think that's the way to go, I can make the relevant changes.

I'd definitely hold off on making those API changes. Let's discuss first. I am not clear about that requirement.

st2common/st2common/constants/secrets.py
Kami: @lakshmi-kannan Yeah, using a static length string is better, faster and safer imo.

I'm pretty sure the code right now is not 100% safe either, but when using a variable length, we could also increase the attack surface area and potentially make timing attacks easier.

Kami:
Yeah, I agree. I would definitely not continue with that until we are all on the same page and agree that's the way to go.

Overall, it does look like we are on the same page and I also think code is more or less there. The "API" we are talking about is `extra` keyword argument which is passed to the logger method.

Creating a new / additional logger method wrapper makes little sense imo - we already encourage developers to include context in "extra" so a new wrapper method wouldn't really get us much here.

lakshmi-kannan:
>Creating a new / additional logger method wrapper makes little sense imo - we already encourage developers to include context in "extra" so a new wrapper method wouldn't really get us much here.

I guess you could argue that :P. It's just I have to make 3 different calls before even logging the object seems like a hassle to me.

Kami:
> I guess you could argue that :P. It's just I have to make 3 different calls before even logging the object seems like a hassle to me.

You don't need to make any additional calls if you are logging DB model instances (e.g. `ExecutionDB`). Default implementation of `to_serializable_dict` will take care of that.

It's a different story if you are logging a "random" dict / value - in this case we have no context except the global variable name blacklist. Having no context means there is nothing we can really do and it's up to the programmer to tell us which variables are secret - having additional logger methods wouldn't change that. Even if we had something like `LOG.foo('message', extra=extra, secrets=dict)`, user would still need to figure out which variables are secrets and which are not in the place where they are calling the logger method.

st2actions/st2actions/bootstrap/runnersregistrar.py
Kami: Yeah, we already kinda do that - we have a static global blacklist of attribute / variable names which are masked - https://github.com/StackStorm/st2/pull/1615/files#diff-32ea10d4a187e2830fe2faf165473ca6R24

st2actions/st2actions/container/base.py
Kami: Yeah, I agree.

This whole method would need some refactoring love, but I will just refactor the logging stuff for now.

Kami: Just FYI - I've refactored this and made the code cleaner - cb7c3154cd4fda10cb7272b2771b7646321c9ff8.

Now all the masking code is encapsulated inside the `mask_secrets` method and it uses existing masking abstraction / framework.

PR #1617, opened 2015-06-11
Use issubclass instead of isinstance check

manasdk:
* isinstance is obviously the wrong check in this case since we are comparing
  types therefore using issubclass.


Kami:
Good catch.

I believe there are some tests in st2tests for that, but they obviously didn't catch that issue :D

manasdk:
> I believe there are some tests in st2tests for that

I looked in st2tests and st2reactor. Don't see a relevant test. I guess I can just write one.

Kami:
Yeah, looks like it didn't include tests for the polling sensors (https://github.com/StackStorm/st2tests/pull/11).

manasdk:
I will try and add a unit test and something in st2tests.

PR #1620, opened 2015-06-12
Fix a bug in ComplexDateTimeField

Kami:
We were not handling the values correctly (sometimes we used local instead of UTC time) which means that sometimes the result would be a datetime object in a local instead of a UTC timezone.


Kami:
@manasdk @lakshmi-kannan :dancers:

lakshmi-kannan:
Update CHANGELOG please.

Kami:
I've merged branch into master and cherry-picked changes into 0.11.2 branch (see #1621).

PR #1622, opened 2015-06-12
New CloudSlang runner

Kami:
This pull request builds on top of #1491 - it includes some improvements and style changes.

1. Allow user to specify inputs as action parameters. This way it's consistent with other runners - 857a829d661801996ac301ba9f2a4a538a0238d7
2. Use standard `entry_point` attribute to point to the flow file instead of using custom `flow_path` runner attribute. This way it's consistent with other workflow runners (action chain, mistral) and using relative path now also works since entry_point is resolved inside the runner container service. - ca3704e68b0845dea37bde6e7d9b25e85e3d2f59
3. Add a note to the docs that this runner is currently still in an experimental phase.


orius123:
Since it is experimental, should I add the experimental flag in the runerrsregistrar?

Kami:
@orius123 I think it's fine to skip that since if you add experimental attribute this means the runner is not registered by default when running register-content script and user needs to explicitly run register-content script with "--experimental" flag.

Kami:
Alright, I've merged initial support into trunk - thanks!

Feel free to open new pr(s) for documentation and other updates if needed.

orius123:
Thanks! :fireworks: 
We will.

Issue #1619, opened 2015-06-11
st2 run packs.install should register all pack components by default


armab:
It seems related to recent: `Action aliases are registered by default. (improvement)` change.

The problem I'm experiencing in chatops is that not everything loaded/unloaded when running `install`/`uninstall` for packs. It's just nice to have `install`/`uninstall` process complete, so user shouldn't care about low-level details.

My behavior: 

* `st2 run packs.install packs=hubot`
* install, configure and run hubot with `hubot-stackstorm`
* expect that I get command output from st2 in chat `!pack info hubot `
* because `chatops.notify_hubot` rule shipped with `hubot` pack wasn't registered - I don't get any output in chat

I think `install` could do all the things by default, not only [register `actions`](https://github.com/StackStorm/st2/blob/68607959047321031d7a3214c75e8eb7fb94dc40/contrib/packs/actions/install.meta.yaml#L24). Maybe I'm missing some use cases here when only `actions` should be registered?

------
Related problem for uninstall. It could be better if that command unload all pack contents as well.

```sh
st2 run packs.uninstall packs=hubot
...
```
After uninstall there is still rule registered:
```sh
st2 rule list -p hubot
+----------------------------+-------+----------------------+------------------------------------------+
| ref                        | pack  | name                 | description                              |
+----------------------------+-------+----------------------+------------------------------------------+
| hubot.chatops.notify_hubot | hubot | chatops.notify_hubot | Notification rule to send messages to    |
|                            |       |                      | Hubot                                    |
+----------------------------+-------+----------------------+------------------------------------------+

```

-----

Third problem, 
seems action-aliases are still not registered:
```sh
st2 run packs.install packs=st2-google repo_url=jfryman/st2-google
...
st2 action-alias list | grep google
# no action aliases registered
```
These commits might be related: [[ 1 ]](https://github.com/StackStorm/st2/commit/a22b3e9c058569aebeb3236fcae257cb52f5c94a) [[ 2 ]](https://github.com/StackStorm/st2/commit/4ceca01035cc86f8a399acabc47b10a27222fdc3)

-----
^^ Tested on `latest` version.
```sh
st2 --version
st2 ('0.12dev.31',)
```

If that makes sense I can submit PR and verify & test everything again. My intent is to install/uninstall st2 packs in chatops way without any additional steps.

manasdk:
[...]

Issue #1231, opened 2015-03-06
Preserve type in ActionChain publish variables


manasdk:
Since ActionChain publishing uses Jinja the types for published variables are not preserved and everything gets a string representation.

Fix either by switch all this to YAQL or some other post-jinja magic.

armab:
The same is for actions, right?

Wanted to edit [load.yml](https://github.com/StackStorm/st2/blob/master/contrib/packs/actions/load.yaml):
```yaml
---
  name: "load"
  runner_type: "local-shell-cmd"
  description: "Action that reloads all st2 content."
  enabled: true
  entry_point: ""
  parameters:
    sudo:
      immutable: true
    cmd:
      immutable: true
      # this line :(
      # it's a string, man: ["actions", "aliases", "sensors"]
      default: "st2 reload{% for item in register %} --register-{{item}}{% endfor %}"
    register:
      type: "array"
      items:
        type: "string"
      default:
        - "actions"
        - "aliases"
        - "sensors"
      description: "Possible options are all, sensors, actions, rules, aliases."
    kwarg_op:
      immutable: true
```


And cried here:
```yaml
    cmd:
      immutable: true
      default: "echo '{{ register|join('-') }}'"
    register:
      type: "array"
      items:
        type: "string"
      default:
        - "actions"
        - "aliases"
        - "sensors"
```
->
`[-"-a-c-t-i-o-n-s-"-,- -"-a-l-i-a-s-e-s-"-,- -"-s-e-n-s-o-r-s-"-]`

Or I'm doing something wrong?

lakshmi-kannan:
@armab I am not sure what you want the cmd as in simple text?

armab:
@lakshmi-kannan I expected I can work with `{{ register }}` variable as a sequence.

lakshmi-kannan:
{% for item in register %} --register-{{item}}{% endfor %} like in original load.yaml is a sequence, right?

manasdk:
@armab Not sure why this borked when you tried to use an array. Most likely the jinja template.

enykeev:
Confirming still a problem. Potential solution might be to use [`callable_expression`](https://github.com/pallets/jinja/blob/master/jinja2/environment.py#L568) jinja method when template consist of only a single jinja variable. The potential consequences for existing chains are going to be severe.

PR #1624, opened 2015-06-12
Pass many 'register' components for 'packs.install', 'packs.load', 'st2ctl reload' #1619

armab:
#### Problem

One of the problems described in #1619 is that `packs.install` reloads only `actions` by default, bringing problems (especially in ChatOps) when installed pack contains aliases or sensors:
```
st2 run packs.install packs=st2-google repo_url=jfryman/st2-google
```

#### Solution
This PR brings possibility to pass several `register` components.
If not provided `register=actions,aliases,sensors` is default value.

```sh
# before (1 component only):
st2 run packs.load register=actions

# now:
st2 run packs.install packs=hubot register=aliases,rules
st2 run packs.load register=rules,sensors
```
which would trigger reload with many arguments (only 1 was accepted before):
```sh
st2ctl reload --register-aliases --register-rules
```

### DIFF
--- a/contrib/packs/actions/install.meta.yaml
+++ b/contrib/packs/actions/install.meta.yaml
@@ -21,5 +21,5 @@
       default: "master"
     register:
       type: "string"
-      default: "actions"
+      default: "actions,aliases,sensors"
       description: "Possible options are all, sensors, actions, rules, aliases."
--- a/tools/st2ctl
+++ b/tools/st2ctl
@@ -47,12 +47,12 @@ function print_usage() {
     echo
     echo "usage: st2ctl {reload, clean}"
     echo "optional arguments:"
-    echo "  --register-all      Register all sensors, rules, and actions."
-    echo "  --register-sensors  Register all sensors only."
-    echo "  --register-rules    Register all rules only."
-    echo "  --register-actions  Register all actions only."
-    echo "  --register-aliases  Register all aliases only."
-    echo "  --register-policies Register all policies only."
+    echo "  --register-all      Register all."
+    echo "  --register-sensors  Register all sensors."
+    echo "  --register-rules    Register all rules."
+    echo "  --register-actions  Register all actions."
+    echo "  --register-aliases  Register all aliases."
+    echo "  --register-policies Register all policies."
     echo "  --verbose           Output additional debug and informational messages."
 }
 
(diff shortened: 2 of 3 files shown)


tools/st2ctl
```diff
@@ -93,19 +93,26 @@ then
...
+            if [[ " ${ALLOWED_REGISTER_FLAGS[*]} " != *" $arg "* ]]; then # argument not allowed
+                echo -e "\e[31mError: Invalid argument provided: ${arg} \e[0m\n"
+                print_usage
+                exit 1
+            fi
+        done
+    fi
+
     if [ -z ${2} ]; then
-        REGISTER_FLAGS="--register-sensors --register-actions --register-aliases"
```
Kami: Yeah, this is a lot better. Hated this monstrosity :P

Kami: I'm just slightly worried, since I don't remember if we have any end to end tests for st2ctl.

If we don't, now it would be a good time to add some to st2tests repo.

contrib/packs/actions/install.meta.yaml
```diff
@@ -21,5 +21,5 @@
       default: "master"
     register:
       type: "string"
-      default: "actions"
+      default: "actions,aliases,sensors"
```
Kami: This is misusing string parameter type (it should really be `array`), but I assume you have done it because you can't really specify arrays via chatops now, right?

If so, that's fine for now. As a future improvement, we could look into re-using CLI logic for specifying arrays and objects as strings, but this might require more work since aliases are not aware of the parameter types.

manasdk: Aliases work with array like the Cli. Also there is some type awareness

Kami: Oh, they do? Cool, I didn't know that.

(I do see the `CAST_OVERRIDES` code now...)

armab: I've tried with arrays first, but it's converted into string. See:
https://github.com/StackStorm/st2/issues/1231#issuecomment-111285096

manasdk: > I tried with array first, but it's converted into string. See: #1231 (comment)

Most likely due to the specific jinja template being used. Have not tried it out myself so can't say what works and what does not.

> Also, as I understand array type won't be preserved here in install workflow

It will if load.yaml say it expect input as an array. The output type is not preserved since we lack type information. However, input to an action is cast, as much as possible, into the correct types.

I will agree that all this is very confusing and how the translations happens and types are preserved/not preserved during transitions is quite unclear.

manasdk: For now if you have figured out a way we should go with that since we know that it works.

manasdk:
This works for me. We can sort out the array pieces laters.

The only drawback with string are the type is that `actions, sensors, aliases` is not the same as `actions,sensors,aliases`. You could write a trim filter for jinja if motivated to support both

armab:
> `actions, sensors, aliases`

Good point,
I've checked `st2 run packs.install packs="hubot, ansible"` is parsed well, so I would like to include trim logic here for consistency.

armab:
@manasdk when you'll have time take a look at that https://github.com/StackStorm/st2/issues/1231#issuecomment-111285096 if it's possible to do it in right way.

Splitting that string into array is absolutely dirty. If it's possible, I would be happy to use an array there.

tools/st2ctl
armab: st2tests looks like some kind of acceptance tests, pretty cool thing. I'll do it a bit later.

@Kami https://github.com/StackStorm/st2tests/issues/16

armab:
@Kami if this PR looks ok for you, can you double-check and merge it?
Or should I include st2tests first?

Kami:
Sorry for the delay. I will go ahead, test it and if everything looks OK, merge it.

Tests are not a blocker right now, but having them asap would be great :)

Kami:
Merged into master - changes look and work fine.

Thanks.

armab:
Thanks!

PR #1629, opened 2015-06-12
RFR - Add exporter to launchdev

lakshmi-kannan:


Kami:
Hm, to me it seems exporter is something which shouldn't really be running in dev (similar to #1552).

PR #1630, opened 2015-06-13
Make sure all the datetime objects we are working with are in UTC

Kami:
Currently, `isotime.parse` returns datetime object with a timezone which is defined in the string which is parsed.

This means that if a users post ISO8601 timestamp with a non UTC timezone to the API things will go bad since we never convert this to UTC before we store in the database (among other things, this also breaks `ComplexDateTimeField`).

With this pull request, all the `datetime` objects which are returned by `isotime.parse` are now converted to UTC by default. In addition to that, I also included some other timezone related fixes and improvements.

P.S. I noted this as a blocker for v0.11.2 in https://github.com/StackStorm/st2/commit/5df21b168c34bdf82bf66cb3ce08e27a474eb057#commitcomment-11651986, but my comment was either missed or ignored :P


st2common/st2common/util/isotime.py
```diff
@@ -22,6 +22,7 @@
 __all__ = [
```
Kami: Note: This is just a first pass - I wanted to get all the code updated and tests passing first.

As a second pass I will probably move general datetime related utility functions into `st2common.util.date` so we have a better separation and we can differentiate between ISO time and other date related utility functions.

Kami:
Pushed `st2common.util.date` refactor changes in 3e7c6b586d6c7fe71fa82aa42abcf4cb58faf9fe.

Kami:
As noted in the PR description, this should also go into v0.11.2, but I didn't so I think we should spin up v0.11.3 with those changes.

What do others think?

lakshmi-kannan:
This is great! Thanks for doing this.

st2common/st2common/fields.py
```diff
@@ -18,7 +18,7 @@
 
 from mongoengine import LongField
 
-from st2common.util import isotime
+from st2common.util import date
```
m4dcoder: import date as date_utils to be consistent?

m4dcoder:
What's the general guideline on when to use isotime utils vs. date utils?

lakshmi-kannan:
I think date_utils should house things that deal with datetime objects. isotime should contain anything to do with generation of iso timstamp strings or parsing one.

st2common/st2common/fields.py
Kami: Yeah, good catch - will push a consistency fix.

Kami:
Yeah, `date` module should host general date related utility functions and `isotime` one should host ISO8601 time related utility functions.

Now both modules have `parse` function - the one in isotime module only parses timestamps in ISO8601 format, but the one in `date` module parsed many different date string formats.

On a related note - I also updated date related rule comparison operators to use `date.parse` instead of `isotime.parse` - now users can use those operators on many string date formats and not just ISO8601 ones.

PR #1631, opened 2015-06-15
Update windows_cmd with all the parameters it consumes

kirilstrax:


Kami:
Thanks, but those are runner parameters which are already defined at the runner level (st2actions/st2actions/bootstrap/runnersregistrar.py) which means they don't (and shouldn't) need to be defined again in the action. Parameters which can be passed to an action are a union of runner and action parameters.

You would only define (override) them in the action if you want to specify a default value and / or make it immutable.

PR #1632, opened 2015-06-15
Update windows_cmd with all the parameters it consumes

kirilstrax:


kirilstrax:
Duplicate pull request (received timeout on previous one, so thought it never came through)

Issue #1646, opened 2015-06-17
Environment variables are confusing in python actions and shell actions. 


lakshmi-kannan:
Shell actions like core.local and core.remote run as system_user whereas python actions run as user as which st2 processes run. So setting environment variables needs to happen for different users for different actions. We need a less confusing way. Perhaps some documentation around this. 

Reported by:  @skarmark on freenode.

Kami:
To clarify some things (things also need to be clarified in the docs):

By default, st2actionrunner should also run under system_user (`stanley`) and as such, the result should be the same, unless user wants to run a local runner action under a different user or use sudo to run it as a root.

In this case, we will use `sudo` and the result will depend on the sudo settings (e.g. should environment variables be preserved or not).

Kami:
Edit: Actually, I just had a look - we use `sudo -E`, but user still needs to put `setenv` option in the sudoers file otherwise environment won't be preserved and -E wont have an affect.

Kami:
To clarify my comment and question on IRC - that's how it works right now as long as you correctly run st2actionrunner under stanley (system) user.

If you want to use a custom (non stanley / system) user, you need give him paswordless sudo access if you want to run local runner actions as a different user and as a root. In addition to that, you also need to make sure `SETENV` sudoers option is set for that user - otherwise environment variables would be preserved when running action as a different / root user.

Kami:
@skarmark Please let me know if my comment above and documentation updates make it more clear and resolve this issue.

lakshmi-kannan:
@skarmark ping! can you answer kami's question?

skarmark:
Yes, it is clear. Sorry about the delay!  Somehow missed the update on this thread.

Kami:
Thanks for letting us know.

Closing this issue.

PR #1648, opened 2015-06-18
Document SETENV sudo requirement for system (stanley) user

Kami:
If `SETENV` option is not set and user runs local runner action under a different user, environment variables which are specified as parameter or available to the system user won't be available to the action since sudo will clean the environment variables.

This pull request "fixes" that by documenting that `SETENV` sudoers option needs to be set system user.

Related to #1646


Kami:
I've added integration test case for it (2ab3283f497ed5e656c4036b35319ce9a6690544), but I also need to make sure `SETENV` option is set in the sudoers file on the build server otherwise the tests will fail.

Note (@lakshmi-kannan @manasdk @m4dcoder et al): Same goes for everyone running integration tests locally) - you need `setenv` option in the sudoers file otherwise integration tests will fail. Fee changes for an example - also make sure to use `visudo` if you are editing the sudoers file manually so you don't bork it.

PR #1649, opened 2015-06-18
Graceful API WSGI server shutdown

Kami:
This pull request implements graceful API WSGI server shutdown. On shutdown (or CTRL-C / SIGINT signal), we stop receiving new requests and give existing active requests some time to shut down (2 seconds by default) before we kill requests which still didn't finish.

This is needed since we support long-running requests (stream endpoint). Previously, if there was an active long-running request, it would block API server from gracefully shutting down and we would need to send `SIGTERM` to the server to kill it.

This is also important and will be utilized when we support requests draining and zero-impact and zero-downtime (rolling) upgrades / restarts.

Note: This is something @enykeev were passively trying to figure out in the past.


dzimine:
Nice.

PR #1650, opened 2015-06-18
Support for secret masking in the API responses, admin role

Kami:
This pull request includes support for masking secrets (action parameters with attribute `secret` set to `True`) in the API responses. Secrets are masked by default, but this setting can be tweaked (for debugging purposes) inside a config.

In addition to that, it also adds support for an administrator ("admin") role. User which has that role can see real (unmasked values) of the secret parameter by passing `?mask_secrets=false` query parameter to the API endpoint.

Currently, administrator users can be defined inside the config (a list of usernames), but once we implement RBAC, this mapping will be stored elsewhere (in the database). On a related note - this pull request also includes some basic framework / functionality which will be re-used / built upon in the RBAC change.

## Other changes

* Allow users to disable secrets masking in the logs by setting the configuration option

## TODO

* [x] Mask parameters in the actiontrigger


Kami:
I haven't added any documentations on how the admin role works and how to configure it yet, because if I understood it correctly, we won't ship next release before RBAC (not just secret masking) is done.

@dzimine Can you please confirm that? If we want to do release earlier (before full RBAC is done), I'm fine with that. The code is fully functioning, and we can also add a migration for admin users (read the config, write it to the DB, ...).

dzimine:
> @dzimine Can you please confirm that?
Confirmed, fine to delay docs.

st2api/st2api/controllers/resource.py
```diff
@@ -129,7 +132,14 @@ def _get_all(self, exclude_fields=None, **kwargs):
             pecan.response.headers['X-Limit'] = str(limit)
         pecan.response.headers['X-Total-Count'] = str(instances.count())
 
-        return [self.model.from_model(instance) for instance in instances[offset:eop]]
+        from_model_kwargs = self._get_from_model_kwargs_for_request(request=pecan.request)
+
+        result = []
+        for instance in instances[offset:eop]:
+            item = self.model.from_model(instance, **from_model_kwargs)
```
lakshmi-kannan: You probably need a try catch around this.

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -71,6 +73,20 @@ class ActionExecutionsControllerMixin(RestController):
...
     ]
 
+    def _get_from_model_kwargs_for_request(self, request):
+        from_model_kwargs = {'mask_secrets': cfg.CONF.api.mask_secrets}
+        query_string = request.query_string
+        query_params = dict(urlparse.parse_qsl(query_string))
+
+        show_secrets = query_params.get('show_secrets', 'false')
+        show_secrets = show_secrets.lower()
+        show_secrets = show_secrets == 'true'
```
lakshmi-kannan: You could define true as a constant in this file. :P

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -71,6 +73,20 @@ class ActionExecutionsControllerMixin(RestController):
...
+
+        if show_secrets and request_user_is_admin(request=request):
+            from_model_kwargs['mask_secrets'] = False
```
lakshmi-kannan: So if a non-admin user asks for unmasked secrets, should be actually fail the request with BAD_REQUEST?

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -129,11 +146,14 @@ def _get_children(self, id_, depth=-1, result_fmt=None):
         # make sure depth is int. Url encoding will make it a string and needs to
         # be converted back in that case.
         depth = int(depth)
+        from_model_kwargs = self._get_from_model_kwargs_for_request(request=pecan.request)
         LOG.debug('retrieving children for id: %s with depth: %s', id_, depth)
         descendants = execution_service.get_descendants(actionexecution_id=id_,
                                                         descendant_depth=depth,
                                                         result_fmt=result_fmt)
-        return [self.model.from_model(descendant) for descendant in descendants]
```
lakshmi-kannan: Why were none of these calls wrapped in a try catch? from_model can throw, right?

st2api/st2api/listener.py
```diff
@@ -78,10 +78,12 @@ def get_consumers(self, consumer, channel):
...
     def processor(self, model):
         def process(body, message):
+            from_model_kwargs = {'mask_secrets': cfg.CONF.api.mask_secrets}
             meta = message.delivery_info
-            event_name = "%s__%s" % (meta.get('exchange'), meta.get('routing_key'))
+            event_name = '%s__%s' % (meta.get('exchange'), meta.get('routing_key'))
+
             try:
-                self.emit(event_name, model.from_model(body))
+                self.emit(event_name, model.from_model(body, **from_model_kwargs))
```
lakshmi-kannan: Why can't you just pass mask_secrets=cfg.CONF.api.mask_secrets directly?

st2common/st2common/constants/rbac.py
```diff
@@ -0,0 +1,25 @@
...
+# limitations under the License.
+
+__all__ = [
+    'SystemRole'
+]
+
+
+class SystemRole(object):
+    """
+    Enum class representing names of the system roled which can't be changed or deleted.
```
lakshmi-kannan: I did not know this is how python does enums :P.

st2common/st2common/hooks.py
```diff
@@ -100,7 +101,22 @@ def before(self, state):
...
             return
 
-        state.request.context['token'] = self._validate_token(request=state.request)
+        token_db = self._validate_token(request=state.request)
+
+        try:
+            user_db = User.get(token_db.user)
+        except ValueError:
+            # User doesn't exist - we should probably also invalidate token if
+            # this happen
```
lakshmi-kannan: Nit: happens

lakshmi-kannan:
You could add mask_secrets=False options in dev.conf so if we are testing this feature, it's easy.

st2api/st2api/controllers/resource.py
lakshmi-kannan: EDIT: I see the original code didn't do this. So you can do this in another PR.

st2api/st2api/controllers/resource.py
```diff
@@ -187,6 +198,16 @@ def _get_by_name(self, resource_name):
...
 
+    def _get_from_model_kwargs_for_request(self, request):
+        """
+        Retrieve kwargs which are passed to "LiveActionAPI.model" method.
+
+        :param request: Pecan request object.
+
+        :rtype: ``dict``
+        """
+        return self.from_model_kwargs
```
lakshmi-kannan: I am actually confused with this method. It takes in `request` argument but does nothing with it :P. Can you use abc and declare it an abstract method and have a pass implementation here? I also see that you set the from_model_kwargs in executions controller but it's kind of confusing. What you want is a default kwargs defined for the controller and then some kv pairs you get from requests. This method should return a fully formed dict from these two dicts.

st2common/st2common/config.py
```diff
@@ -62,7 +62,9 @@ def register_opts(ignore_errors=False):
 
     system_opts = [
         cfg.StrOpt('base_path', default='/opt/stackstorm',
-                   help='Base path to all st2 artifacts.')
+                   help='Base path to all st2 artifacts.'),
+        cfg.ListOpt('admin_users', default=None,
```
lakshmi-kannan: Should the default value be None?

st2api/st2api/controllers/resource.py
Kami: This is a base implementation so it doesn't necessary need to do anything with the `request` argument.

First I went with the approach you have described in the executions controller, but later on, I moved `from_model_kwargs` from class attribute to inline so it's a bit more efficient.

st2api/st2api/controllers/v1/actionexecutions.py
Kami: We could, but I decided to ignore it right now since we don't really have established consistent patterns for handling API errors just yet.

In a lot of places we just silently consume additional arguments and don't do anything with / about them (it mostly happens because some controllers abuse `**kwargs`).

st2api/st2api/controllers/resource.py
Kami: Yeah, it's really just a question of semantics - which layer should throw, which one should consume exception, which one should format them for the response...

st2api/st2api/controllers/v1/actionexecutions.py
Kami: It can, a lot of methods we call in the API handlers can (and sometimes do throw) and we do nothing about it.

I'm not saying that's good, I'm just saying we first need to decide on a consistent way of handling exception in the API.

st2common/st2common/config.py
Kami: Good catch - defaulting to an empty list here would actually be fine since we don't mutate it later.

st2api/st2api/controllers/v1/actionexecutions.py
Kami: True.

st2api/st2api/listener.py
Kami: Consistency with other API controller code, but no other reason beyond that :P

Kami:
> You could add mask_secrets=False options in dev.conf so if we are testing this feature, it's easy.

Done in a9919ba195d2248840f868be6409949d49d510f4.

If you want to test it, you can turn the auth on, turn the secret masking on and use the `testu` user which is specified as an admin in the dev environment.

st2common/st2common/config.py
Kami: Done - 16af8c31ac05fb8a7f4eab317a8d775c93da12cf.

PR #1656, opened 2015-06-22
Add "--debug" CLI flag to all the services, refactor duplicated common service setup / bootstrap code, refactor bad service coupling

Kami:
This pull requests adds a new "--debug" CLI flag to all the service entry points. When this flag is specified, log level for all the loggers is set to `DEBUG`.

This makes debugging easier since it means you can just start service with `--debug` flag and you don't need to make temporary edits to logging config. In the future, we can also enable / expose additional debugging functionality when this flag is present.

In addition to that, this pull request also gets rid of all the nasty and duplicated common service setup code (long awaited). It moves common setup code in a shared module and updates all the affected service. This centralized approach also makes it easier to add other common setup steps (e.g. register service in the service registry, etc.).


st2api/st2api/config.py
```diff
@@ -28,10 +28,23 @@
...
+def register_opts():
+    _register_common_opts()
+    _register_app_opts()
+
+
 def _register_common_opts():
     common_config.register_opts()
 
 
+def get_logging_config_path(service='api'):
```
lakshmi-kannan: The argument is not used. Why do we need it passed in? Same comment applies to other services.

lakshmi-kannan:
Nice PR! Changes look good to me.

st2api/st2api/config.py
Kami: The problem is that we don't have one `config.py` per service.

First I went with adding `get_logging_config_path` method to `config.py` module approach, but it broke down with the services in `st2actions/` module since those service doesn't have separate config (see https://github.com/StackStorm/st2/pull/1656/files#diff-32a22336e395c4f129c6b33e1c91545eR96 for details).

I would actually be OK with refactoring those services and also creating a config for resulttracker and notifier (move both in a package under st2action - st2actions/{notifier,result)tracker}/config.py. This approach is probably better anyway since it would make notifier and result tracker follow established and consistent patterns from other services.

WDYT?

lakshmi-kannan: Ahh. I understand now. Perhaps a better refactoring would be to move the notifier out of st2actions at some point. But yeah, having packages inside packages is fine. Not sure how this would affect packaging though. Gotta be careful not to break stuff. I'd do this in another PR just to be safe.

Kami: I decided to do the refactoring and move `resultstracker` and `notifier` in a package under `st2actions/`.

Eventually, we will probably want to do a bigger refactoring and move those to packages to the top level outside of `st2actions/`.

On a related note - during this refactoring, I also noticed some bad coupling between scheduler / re-scheduler - currently both actionrunner and notifier depend on it and use it.

When we are doing a bigger refactoring we should also fix this coupling.

Kami: Looking at the code, it should be easy to move rescheduler part from `st2actions/scheduler.py` to `st2actions/notifier/scheduler.py` where it belongs and fix the coupling.

Kami: Pushed scheduler and re-scheduler decoupling refactor in 602a8c2afa3e15499eb83553ca585560b4b522da.

Re-scheduler had no place to live in `st2actions/scheduler.py` since it was only used by the notifier code.

Code is now a lot more self-sustaining which means it should also be easier to do the bigger refactoring later on.

manasdk:
Great!

PR #1657, opened 2015-06-23
Provide sensor_ref instead of sensor_name to run a single sensor

manasdk:
* We only guarantee uniqueness of ref and not of name therefore to
      target a single sensor it is better to use sensor_ref.

### DIFF
--- a/st2reactor/st2reactor/cmd/sensormanager.py
+++ b/st2reactor/st2reactor/cmd/sensormanager.py
@@ -45,14 +45,14 @@ def main():
     try:
         _setup()
         container_manager = SensorContainerManager()
-        sensors = _get_all_sensors()
+        sensors = None
 
-        if cfg.CONF.sensor_name:
-            # Only run a single sensor
-            sensors = [sensor for sensor in sensors if
-                       sensor.name == cfg.CONF.sensor_name]
+        if cfg.CONF.sensor_ref:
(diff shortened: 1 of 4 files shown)


st2reactor/st2reactor/cmd/sensormanager.py
```diff
@@ -45,14 +45,14 @@ def main():
...
-        sensors = _get_all_sensors()
+        sensors = None
 
-        if cfg.CONF.sensor_name:
-            # Only run a single sensor
-            sensors = [sensor for sensor in sensors if
-                       sensor.name == cfg.CONF.sensor_name]
+        if cfg.CONF.sensor_ref:
+            sensors = [SensorType.get_by_ref(cfg.CONF.sensor_ref)]
             if not sensors:
```
Kami: If `SensorType.get_by_ref` returns `None` and doesn't throw on in-existent DB object, we will never get into the if clause bellow since `bool([None]) is True`.

In short, we need to do something like:

```python
sensor = SensorType.get_by_ref(cfg.CONF.sensor_ref)
if not sensor:
    ...

sensors = [sensor]
```

Kami:
Besides the comment above, LGTM.

st2reactor/st2reactor/cmd/sensormanager.py
manasdk: Good catch. Fixed 60ba1f95abac424df9bb147fe4074ae91e516a62

PR #1659, opened 2015-06-24
Only run enabled sensors

manasdk:


st2reactor/st2reactor/cmd/sensormanager.py
```diff
@@ -36,7 +36,8 @@ def _teardown():
 
 
 def _get_all_sensors():
```
Kami: Minor thing - maybe the method should be called `_get_all_enabled_sensors` :)

Anyway, LGTM.

manasdk: Done https://github.com/StackStorm/st2/pull/1662

PR #1661, opened 2015-06-24
Dispatch internal trigger on datastore CUD action, add locking to avoid race conditions

Kami:
This pull request includes two changes:

### 1. Add locking using the new coordination service to avoid race conditions in the datastore API operations.

The title is self explanatory. This was a known issue / limitation since when we were making that change we were fine with that it until we implement locking. We didn't want to implement locking / coordination at that time, but not we have the necessary primitives / framework in places so we should use it.

### 2. Dispatch internal trigger when a datastore item is updated, deleted and when the value changes

This is something we have discussed a while back and we should ideally really do for all the content pack / system resources. The only difference is that for other resources we would only publish `create`, `update` and `delete` event, but no `value_change` event since this one is specific to the key value pair model. In theory, we could also publish `object_change` (or similar) event for other resources, but it has little value outside of the key value pair resource.

### TODO

- [x] Agree on the backend implementation and do the refactor (see https://github.com/StackStorm/st2/pull/1661#issuecomment-114850772 for details)
- [x] Add docs for the internal trigger events
- [x] Refactor the internal trigger registrar code, move it to the API process bootstrap phase (see https://github.com/StackStorm/st2/pull/1661#issuecomment-114859722 for details)


Kami:
As far as the "public API" for internal datastore events goes (how the trigger name and payload look like) - I think this should stay more or less the same, but the backend implementation is up for a discussion.

I've put event dispatch in the API layer right now since this is already an established pattern which we do in some other places (e.g. webhook), but we might want to move it to the persistence layer. Moving it to the persistence layer would be nice, especially if we want to do the same for other resources since it would result in a nicer programming API and reduce code duplication, but there are some caveats:

1. We would need to implement addition (per resource) logic which knows when a resource has changed. This shouldn't be a big issue and we would only do it for KeyValuePair resource.

2. We already have some event publishing code in place, but the events published by `CUDPublisher` & friends are system events which are only consumed by StackStorm. On the other hand, internal triggers are trigger events which can be consumed by users and also the StackStorm platform. To make this clear in the code we should use explicit (verbose) naming (e.g. `get_publisher` vs `get_trigger_publisher`, `publish_update` vs `publish_trigger_update_event`, etc.)

If you guys are OK with moving it to the persistence layer, I can make that change. To begin with, I would only enable publishing of those events for the KeyValuePair resource.

Kami:
On a related note - with this change, internal triggers will now also be published by the API process which means we can get rid of the internal trigger registrar and simply register all the internal triggers inside the API process during the bootstrap / setup phase.

Honestly, I don't know / remember why we didn't just do that in the past - internal trigger registrar needs access to the API and for the API service to be online to register the triggers so it's not really any different then registering triggers directly in the API process (in fact, doing it outside of the API by hitting the API involves all kinds of issues and limitations of which we are already aware of).

manasdk:
> If you guys are OK with moving it to the persistence layer, I can make that change.

Yes. This makes sense. Will reduce code and can be an easy opt-in mechanism for various resources.

> We already have some event publishing code in place, but the events published by CUDPublisher & friends are system events which are only consumed by StackStorm. On the other hand, internal triggers are trigger events which can be consumed by users and also the StackStorm platform.

CUDPublisher does not publish triggers those are internal messaging constructs. In this case we will always publish to the TriggerInstance exchange in RMQ and can be used in Rules. I am comfortable with the distinction and this should not lead to any confusion (i hope). so +1

manasdk:
> On a related note - with this change, internal triggers will now also be published by the API process which means we can get rid of the internal trigger registrar and simply register all the internal triggers inside the API process during the bootstrap / setup phase.

I don't see this code. What am I missing?

Kami:
@manasdk I didn't make that change yet - first wanted to make sure we are all on the same page and agree on the approach :)

manasdk:
> first wanted to make sure we are all on the same page and agree on the approach

No specific issue with the API. Registering from any component multiple instances bothers me but only because it is extra work. Having said that API is better than almost any other component.

Kami:
@manasdk Yes, currently we register it from multiple components, but with the new approach we will only register it inside the API service.

The end result will still be the same since previously, other services hit API to register the triggers, but the whole thing will be less hacky and more efficient and robust (e.g. triggers not getting registered when API is offline, etc.).

If user ran multiple API services this means we would still register it multiple times, but the operation is idempotent and eventually, we can also add locking if needed.

lakshmi-kannan:
Ideally, I'd have liked this PR to be two PRs - one for introducing locking and one for trigger events. 

I have problems with introducing locking. Looking at the coordinator code, if coordinator is disabled in configuration, like so https://github.com/StackStorm/st2/blob/master/st2common/st2common/services/coordination.py#L28

i have no idea how these locks would work. See 
https://github.com/StackStorm/st2/pull/1661/files#diff-b3d716a438dc579e526acc19503da088R108
So maybe we need to pay attention that case. I just want to make sure if I don't want locking enabled (st2express), things still work fine (even if race happens).

jfryman:
Help me understand why there is no `value` or `object` change event emitted? This is pretty much the thing that I'd need to really refactor the autoscale case to get rid of the sensor hack I created, because I'm detecting on a state change for growing/decreasing the case.

lakshmi-kannan:
After discussing more on slack, I think I am -1 to introducing locking for key value controller. We have similar issues for other models too. So we we need a common path for addressing all these. Having some models use locking and some not will cause confusions. Also, the deployment aspects of locking system haven't been nailed down.

Kami:
> Help me understand why there is no value or object change event emitted? This is pretty much the thing that I'd need to really refactor the autoscale case to get rid of the sensor hack I created, because I'm detecting on a state change for growing/decreasing the case.

@jfryman we do actually emit value change trigger right now (https://github.com/StackStorm/st2/pull/1661/files#diff-b3d716a438dc579e526acc19503da088R200) and I also plan to emit if after the refactor :)

When talking about not emitting it, I was referring to other resources, excluding datastore items. I think there is little value in emitting "object change" triggers for other resources right now, but I could also be convinced otherwise.

Kami:
> After discussing more on slack, I think I am -1 to introducing locking for key value controller. We have similar issues for other models too. So we we need a common path for addressing all these. Having some models use locking and some not will cause confusions. Also, the deployment aspects of locking system haven't been nailed down.

@lakshmi-kannan We do, but in other places issues usually aren't that big since we don't do read before write which is what we do in KeyValuePair update controller.

And I'm aware that we haven't decided on the deployment scenario for locking service just yet, but this code works in exactly the same manner as other code which uses locking right now - if user doesn't deploy and configure a coordination / locking service, this code won't do anything.

manasdk:
> So in short - I'm following an existing established approach. If you don't like that, then you should also voice your opinion earlier when we are introducing initial code for coordination service :P

The way m4dcoder put it locking is probabilistic. I do not fully understand what that means but it is not an advertised feature so I am comfortable with locking not truly working since I don't expect anyone to use that outside of a very controlled scenario. Here though it is different and once we set the expectation that there is some synchronization then then there better be synchronization!

manasdk:
> I don't expect anyone to use that outside of a very controlled scenario

In the controlled scenario a real co-ordination backend would exist.

Kami:
> The way m4dcoder put it locking is probabilistic. I do not fully understand what that means but it is not an advertised feature so I am comfortable with locking not truly working since I don't expect anyone to use that outside of a very controlled scenario. Here though it is different and once we set the expectation that there is some synchronization then then there better be synchronization!

We set expectations ourselves in the documentation - we should make it clear that if the user doesn't configure a coordination / locking backend there potential race conditions can occur in some code paths.

Also, as far as I understood it from our past discussions, we wanted to put code infrastructure in place first and eventually we will turn it on once we decided what locking backend we will include in a default installation.

manasdk:
> We set expectations ourselves in the documentation - we should make it clear that if the user doesn't configure a coordination / locking backend there potential race conditions can occur in some code paths.

What docs? This was dark shipped - are there docs?

> we wanted to put code infrastructure in place first and eventually we will turn it on once we decided what locking backend we will include in a default installation.

Yup. Limited to policy really. Again you can add locking and that will be a no-op if there is no co-ordination backend. Policy is opt-in which this impl is not opt-in ... I guess you would then make the case that it is harmless so why not keep it?

Kami:
I've pushed a change which introduces a "framework" for dispatching triggers for resource CUD operations - 606649b4505a2902e315b9d952b2ec7f4d07029c.

For now it's only enabled for KeyValuePair resource. In addition to CUD triggers, we also dispatch a special "high level" `value_change` trigger for this resource.

To distinguish between internal event publishing and trigger dispatching I've prefixed all the methods / variables with "dispatch" in contrast to "publish". This way it's also consistent with the existing code - internally, we always refer to it as dispatching and not publishing a trigger.

Kami:
Also, internal trigger registration refactor / change is in 70305cb5824a6bb5ad46bc51381fa1b7b9128172.

Kami:
Just FYI - I also pushed some documentation changes and improvements.

We have a couple of scripts which generate .rst files which are included in the documentation. Those files are now placed in `docs/source/_includes/` and those script are ran during `docs` Make target.

Also keep in mind that all the files in that directory are automatically generated from the code and shouldn't be edited manually.

scripts/utils.py
```diff
@@ -0,0 +1,87 @@
+# Note: This file is licensed under MIT license
```
manasdk: What is this file?

PR #1663, opened 2015-06-25
Parameterize Content-Type for SMTP action

jfryman:
This PR updates the `core.smtp` action to take `Content-Type` parameters of both `text/html` and `text/plain`. This repairs issues when attempting to send plain text, and having linefeeds not being respected.

Reported from IRC via hikertechy

/cc #1664


contrib/core/actions/sendmail.yaml
```diff
@@ -21,6 +21,14 @@ parameters:
     position: 0
     required: true
     type: string
+  content_type:
+    type: string
+    description: Content type of message to be sent
+    enum:
+      - "text/plain"
+      - "text/html"
+    default: "text/plain"
```
manasdk: Changing the default here - should the default be preserved?

manasdk:
+1

contrib/core/actions/send_mail/send_mail
```diff
@@ -18,7 +20,7 @@ if [[ -z "${BODY// }" && $SEND_EMPTY_BODY == 'True' ]] || [[ -n $BODY ]]; then
 TO: ${TO}
 FROM: ${FROM}
 SUBJECT: ${SUBJECT}
-Content-Type: text/html
+Content-Type: ${CONTENT_TYPE}
 
 ${BODY}
 <br><br>
```
Kami: Should probably also replace this with "\n" if content type is text/plain.

Besides that, LGTM.

PR #1667, opened 2015-06-27
CLI, Mistral "end_timestamp" bug fix

Kami:
This pull request fixes two bugs:

1. CLI throwing an exception when displaying a result for a failed Mistral workflow run.
2. CLI not displaying `end_timestamp` attribute for Mistral workflows. In this case the problem lied in the querier module - we never set `end_timestamp` attribute for a LiveActionDB object which has completed.


Kami:
To clarify the `end_timestamp` issue - we did actually assign `end_timestamp`on the ActionExecutionDB object, but we didn't assign it on the `LiveActionDB` and CLI happens to use `LiveActionDB` in this case.

One lesson we also get out of this is that we should get rid of LiveActionDB stuff.

PR #1668, opened 2015-06-29
Setuptools packaging improvements - make each st2 component packagable and installable via  package

Kami:
This pull requests builds on top of of #1647.

It includes the following changes and improvements:

TODO:

- [ ] Verify it doesn't break our current release workflows
- [ ] Add end to end tests
- [ ] Add new workflow steps for generating Python package distributions and publishing to to PyPi (not needed right now)


Kami:
@dennybaa Can you please work on end to end tests in the `st2tests` repo?

Kami:
I've added some docs which explains how everything works and how to manage the requirements - c0abf833582df556d4743c0596f8394e29d6bd65

Eventually I will move this to the "Developer" section in our docs.

PR #1672, opened 2015-06-30
Create shadow tigger on internal trigger registration

manasdk:
* Previously internal trigger registration was via the API which automatically
  handled shadow trigger registration. Since that was skipped to registering
  in st2api the shadow trigger registration was also dropped leading to regressions.
* Added registration back in and also some basic test to validate shadow trigger
  creation.

Testing -

Validated that creation of `notify_hubot` from the hubot pack works.

```
$ st2 rule create notify_hubot.yaml
+-------------+--------------------------------------------------------+
| Property    | Value                                                  |
+-------------+--------------------------------------------------------+
| id          | 5592174132ed355332f81a8e                               |
| name        | chatops.notify_hubot                                   |
| pack        | default                                                |
| description | Notification rule to send messages to Hubot            |
| action      | {                                                      |
|             |     "ref": "hubot.post_result",                        |
|             |     "parameters": {                                    |
|             |         "user": "{{trigger.data.user}}",               |
|             |         "channel": "{{trigger.data.source_channel}}",  |
|             |         "result": "{{trigger}}"                        |
|             |     }                                                  |
|             | }                                                      |
| criteria    | {                                                      |
|             |     "trigger.channel": {                               |
|             |         "pattern": "hubot",                            |
|             |         "type": "equals"                               |
|             |     }                                                  |
|             | }                                                      |
| enabled     | True                                                   |
| ref         | default.chatops.notify_hubot                           |
| tags        |                                                        |
| trigger     | {                                                      |
|             |     "type": "core.st2.generic.notifytrigger",          |
|             |     "parameters": {},                                  |
|             |     "pack": "core"                                     |
|             | }                                                      |
+-------------+--------------------------------------------------------+
```

### DIFF
--- a/st2common/st2common/triggers.py
+++ b/st2common/st2common/triggers.py
@@ -13,17 +13,45 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import six
+
+from mongoengine import ValidationError
 from oslo_config import cfg
 
 from st2common import log as logging
 from st2common.constants.triggers import (INTERNAL_TRIGGER_TYPES, ACTION_SENSOR_TRIGGER)
-from st2common.services.triggers import create_trigger_type_db
+from st2common.exceptions.db import StackStormDBObjectConflictError
+from st2common.services.triggers import create_trigger_type_db, create_shadow_trigger
 
 __all__ = [
     'register_internal_trigger_types'
 ]
 
 LOG = logging.getLogger(__name__)
+
+
+def _register_internal_trigger_type(trigger_definition):
+    trigger_type_db = create_trigger_type_db(trigger_type=trigger_definition)
+
+    # trigger types with parameters do no require a shadow trigger.
+    if not trigger_type_db.parameters_schema:
+        LOG.info('Registered trigger: %s.', trigger_definition['name'])
+        try:
+            trigger_db = create_shadow_trigger(trigger_type_db)
+
+            extra = {'trigger_db': trigger_db}
+            LOG.audit('Trigger created for parameter-less TriggerType. Trigger.id=%s' %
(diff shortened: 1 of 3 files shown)


Kami:
Good catch :+1: 

A future improvement would be to add a new method to `services/triggers.py` module which creates a `TriggerTypeDB` and shadow `TriggerDB` if needed (e.g. `create_trigger_type_db_with_shadow_trigger`).

Currently we have two different ways / code paths for doing that (one in the API and one in the internal trigger registrar code).

manasdk:
Yes. I know about the multiple paths and I will collapse into a single path. Needed to get the change in without more regression so took the simpler approach. In fact I pretty much copied the code over from the API controller.

All this needs some rethink especially how the exceptions will get thrown etc so prior to changing more behavior I decided to keep the code separate. Once I get some time to think about the right solution I shall rework everything into a shared code path.

Kami:
Yeah, that's fine.

I also always go with "tests" + "simple approach" since once the tests are in place it's easier to refactor the code.

st2common/st2common/triggers.py
```diff
@@ -26,6 +30,30 @@
 LOG = logging.getLogger(__name__)
 
 
+def _register_internal_trigger_type(trigger_definition):
+    trigger_type_db = create_trigger_type_db(trigger_type=trigger_definition)
+
+    # trigger types with parameters do no require a shadow trigger.
+    if not trigger_type_db.parameters_schema:
+        LOG.info('Registered trigger: %s.', trigger_definition['name'])
```
Kami: Minor thing - this log message should be outside of the if statement.

PR #1677, opened 2015-07-01
[WIP] Setuptools packaging improvements - make each st2 component packagable and installable via package

Kami:
This pull requests builds on top of of #1647 and #1668.

I spent a lot of time finishing, polishing and testing all the changes from #1668 locally. Everything worked and looked fine so I decided to merge it into master. Sadly this turned out to be a mistake - there were many, many issues with packaging and other workflows used by our CI.

After 3-4 hours of digging in and fixing issues, I still haven't managed to fix all of them so I decided to revert changes in the master and unbreak it. I believe that's better than trying to rush to whole thing. Next step will be to address all the issues and run all the packaging and related workflows on this branch and only merge it into master when everything passes.

I will work with / ask @DoriftoShoes for help with running everything on a particular branch (not sure if we are there yet, might require some changes).

The biggest offenders currently are:

- [x] Making sure we use pip >= 6.0.0 everywhere. I believe I managed to fix this by using virtualenv everywhere, but there might still be issues somewhere.
- [x] Getting rid of the "recursive-include" abuse in the `MANIFEST.in` and replacing it with collecting all the code packages inside the `setup.py `files.

This branch includes the following changes and improvements:

TO FIX:

- [ ] mistral_itests workflow fails (times out) on CI. I dug into it, but I couldn't reproduce the problem. It works fine locally and it also works fine if I manually run commands on the CI server.
- [ ] st2_deploy_tests workflow fails on CI. I believe I've fixed the RHEL issue, but I don't know what's wrong with the Ubuntu one.


Makefile
```diff
@@ -205,6 +209,12 @@ $(VIRTUALENV_DIR)/bin/activate:
 	echo '  functions -e old_deactivate' >> $(VIRTUALENV_DIR)/bin/activate.fish
 	echo 'end' >> $(VIRTUALENV_DIR)/bin/activate.fish
 	touch $(VIRTUALENV_DIR)/bin/activate.fish
+	
+	# Make sure we use latest version of pip
+	$(VIRTUALENV_DIR)/bin/pip install --upgrade pip
+
+	# Install stdeb st2client build dependency
```
dennybaa: @Kami  what's stdeb?

Kami: It's a Python package we use in combination with setup.py to build a Debian package for `st2client`.

I wasn't aware of it before yesterday either. I added it here to make "make deb" self sustaining and now it also works in a local environment. Previously, this package would only magically get installed inside one of our CI workflows...

Kami:
This is not relevant anymore since https://github.com/StackStorm/st2/pull/1917 was merged.

PR #1679, opened 2015-07-01
setup.py reverted (Setuptools packaging improvments)

dennybaa:
Since we can't rush and change all the code right away, @Kami suggested to do it incrementally.

Here comes the work from https://github.com/StackStorm/st2/pull/1677 with all `setup.py` files reverted to the original look.


Kami:
This is not relevant anymore since https://github.com/StackStorm/st2/pull/1917 was merged. Closing.

PR #1680, opened 2015-07-01
Mistral fixes for CLI, end_timestamp, and capturing published variables

m4dcoder:


st2actions/st2actions/query/base.py
```diff
@@ -119,8 +122,14 @@ def _update_action_results(self, execution_id, status, results):
         liveaction_db = LiveAction.get_by_id(execution_id)
         if not liveaction_db:
             raise Exception('No DB model for liveaction_id: %s' % execution_id)
+
         liveaction_db.result = results
         liveaction_db.status = status
+
+        done = status in DONE_STATES
+        if done and not liveaction_db.end_timestamp:
+            liveaction_db.end_timestamp = isotime.add_utc_tz(datetime.datetime.utcnow())
```
manasdk: Can you use get_datetime_utc_now?

manasdk:
+1

st2actions/st2actions/query/base.py
m4dcoder: I would end up having to cherry pick several commits that are over two weeks ago.  That's why I went with this route since the date utils is not introduced here yet.

Kami: Yeah, that's unfortunate, but not much we can do this time.

:+1:

PR #1682, opened 2015-07-01
Sensor partition

manasdk:
Ability to define partition maps for a sensor nodes. In this PR 3 partition schemes are supported

* `default` - Just as before. Will run all sensors.

* `kvstore` - Partition map can be specified in the key-value store.

* `file` - Partition map can be specific in a static file.

* `hash` - Partition based on a hash range.

The current approach is to make this work in a static manner. However the "pluggable"  mode of partition provider means we could add other schemes (hash-based, consul-based) pretty easily. These would be somewhat more dynamic and could better adjust to changing needs.


docs/source/sensors.rst
```diff
@@ -192,6 +192,90 @@ API Docs
 .. autoclass:: st2reactor.container.sensor_wrapper.SensorService
     :members:
 
+Partitioning Sensors
```
dzimine: It's great to have it documented and docs coming with the feature. 

I would just prefer to somehow separate basic and advanced topics. Else, sensors.rst may become a monster doc and difficult to read through. Think about it (minor).

conf/st2.dev.conf
```diff
@@ -11,6 +11,8 @@ mask_secrets = False
 
 [sensorcontainer]
 logging = st2reactor/conf/logging.sensorcontainer.conf
+sensor_node_name = sensornode1
+partition_provider = name:default
```
dzimine: should we call it "all"? Default sounds more like a property which can be assigned to any provider.

docs/source/sensors.rst
manasdk: Reasonable. I thought about it as well but could not work out where to add this relatively advanced topic. I am happy to change.

conf/st2.dev.conf
manasdk: It is a reference to the provider or name of the provider. I prefer default as `all` does not mean anything. Besides most often no one even needs to include this config as the value default is well default.

docs/source/sensors.rst
manasdk: Done 40790c52b09006c02d9bbda25b0d7588258d2066

conf/st2.dev.conf
manasdk: I guess if we can come up with a better name for the default provider I am happy to change.

st2reactor/st2reactor/container/partitioner.py
```diff
@@ -0,0 +1,165 @@
...
+        return None
+
+
+class KVStorePartitioner(DefaultPartitioner):
+
+    def __init__(self, sensor_node_name):
+        super(KVStorePartitioner, self).__init__(sensor_node_name=sensor_node_name)
+        self._supported_sensor_refs = None
+
+    def is_sensor_supported(self, sensor_db):
```
lakshmi-kannan: Nit: is_sensor_owner() ?

docs/source/reference/sensor_partitioning.rst
```diff
@@ -0,0 +1,84 @@
...
+default scheme which ships out of the box. It is useful in cases where there is a single
+sensor node.
+
+No change required in the config file but for complete-ness the config would be as follows:
+
+::
+
+    [sensorcontainer]
+    ...
+    sensor_node_name = sensornode1
```
lakshmi-kannan: I don't know. Maybe this should hostname-uuid?

st2common/st2common/constants/sensors.py
```diff
@@ -15,3 +15,8 @@
 
 # Minimum poll interval for a sensor
 MINIMUM_POLL_INTERVAL = 4
+
+# keys for PARTITION loaders
+DEFAULT_PARTITION_LOADER = 'default'
```
lakshmi-kannan: These could use URI structures like default:///. Wdyt?

st2reactor/st2reactor/container/partitioner.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+PROVIDERS = {
+    DEFAULT_PARTITION_LOADER: DefaultPartitioner,
+    KVSTORE_PARTITION_LOADER: KVStorePartitioner,
+    FILE_PARTITION_LOADER: FileBasedPartitioner
+}
+
+
+def get_sensors_provider():
```
lakshmi-kannan: sensor_provider doesn't sound right. Should be senor_partitioner, right?

lakshmi-kannan:
Many naming nits. Seems like a reasonable start though.

docs/source/reference/sensor_partitioning.rst
manasdk: The value has no semantic meaning for us right now. So it could be `hostname-uuid` if the user chooses but no support from us really.

st2reactor/st2reactor/container/partitioner.py
manasdk: It provides sensors by partitioning them so a provider. I guess I named it per the client expected interface as opposed to its own nature but I don't have a strong allegiance to either.

st2common/st2common/constants/sensors.py
manasdk: Sure. It looks a lot smarter although not sure what I would do with that URI structure.

st2reactor/st2reactor/container/partitioner.py
manasdk: makes sense.

st2reactor/st2reactor/container/partitioner.py
manasdk: Fixed a97868a7f49b9c0bbf7e6abcd350f1459ba8bb2f

st2reactor/st2reactor/container/partitioner.py
manasdk: Doen a97868a7f49b9c0bbf7e6abcd350f1459ba8bb2f

docs/source/reference/sensor_partitioning.rst
manasdk: Done 1f72c2450d6b0f5c3d8e75d802a39cdef766a601

using `hostname_uuid`

PR #1690, opened 2015-07-02
Fix duplicate in error and result for execution in CLI

m4dcoder:


dzimine:
OK with this. Thanks!

PR #1693, opened 2015-07-05
Added short explanation of exit codes.

kirilstrax:
Most importantly exit code -9 - which unless you know that it's a TIMEOUT_EXIT_CODE = -9 constant defined for runners, is very mysterious exit code.


Kami:
LGTM.

I thought we also set `timed_out` attribute in the result object to `True` if the command times out, but that's actually not the case right now (we just set the error message to indicate the timeout and set status code to -9).

manasdk:
+1

Kami:
Merged, thanks.

PR #1695, opened 2015-07-05
Making powershell run in definitely non-interactive mode

kirilstrax:
By default powershell would run a script and will NOT quit, waiting for user's commands. Which means powershell process will never quit.
Adding "powershell.exe -InputFormat None" resolves that issue
Also added it on top, so that it's more visible

### DIFF
--- a/st2actions/st2actions/runners/windows_script_runner.py
+++ b/st2actions/st2actions/runners/windows_script_runner.py
@@ -54,6 +54,8 @@ CREATE_DIRECTORY_TIMEOUT = 10
 DELETE_FILE_TIMEOUT = 10
 DELETE_DIRECTORY_TIMEOUT = 10
 
+POWERSHELL_COMMAND = 'powershell.exe -InputFormat None'
+
 
 def get_runner():
     return WindowsScriptRunner(str(uuid.uuid4()))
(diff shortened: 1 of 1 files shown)


Kami:
I never encountered this issue when testing it with Rackspace Cloud Windows servers.

Is it possible this can be some how set on a global level and that's a default value on Rackspace Cloud Windows servers?

Edit: I just checked the docs about this switch and it says:

> -InputFormat {Text | XML}
> Describes the format of data sent to Windows PowerShell. Valid values are "Text" (text strings) or  "XML" (serialized CLIXML format).

It doesn't seem None is even a valid value.

kirilstrax:
I had this hanging issue all the time with PowerShell 2 on Windows 2008. The solution to use 
"-InputFormat None" is from this solution on MS site: https://connect.microsoft.com/PowerShell/feedback/details/572313/powershell-exe-can-hang-if-stdin-is-redirected
I tested it and it indeed worked.

Kami:
Sorry for the delay.

Looks like this is an un-documented feature and haven't had the time to test it yet.

On a (maybe) related note - @dzimine originally encountered some issue with winexe hanging which was resolved by also also passing handle to stdin to the winexe process, but this looks like a different issue.

kirilstrax:
Tried -NonInteractive - it doesn't help

Kami:
I didn't encounter this issue myself and I couldn't reproduce it, but I trist @kirilstrax's testing so I will merge it.

Kami:
Merged, thanks.

PR #1698, opened 2015-07-07
Add pack model and API endpoint for listing installed packs

Kami:
This pull request adds a new model for representing a Pack and an API endpoint for listing installed packs.

In addition to that, we now also lazy / late register packs when loading content from disk.

---

Note 1: I wanted to avoid doing that before we work on the pack managment API since a lot of things will change then anyway and adding it basically means we are just adding more debt, but @manasdk managed to convinced me to add some initial code before it.

Inside the RBAC branch, users will be able to assign permissions to a pack and he argued that having a model will make things easier. I personally only agree with the discovery part - now users will be able to list installed packs via API / CLI, but besides that it shouldn't matter since all the permission resolving code will live elsewhere and  the model and related code have nothing do with it (decoupling) besides providing "get_by_uid" functionality.

The model itself should stay the same when we add a pack management API, but the whole registration flow (implementation detail / non user visible change) will be moved elsewhere.

Note 2: I talked with @manasdk and he agreed that right now it's better to register a pack when we are registering other content from disk. Registering it inside ``packs.install`` workflow includes a tons of edge cases and it's not ideal since packs which are not installed via ``packs.install`` (e.g. symlinked / dev packs) would not be visible in the system, etc.

In short, if you don't like any of that, blame @manasdk


st2common/st2common/bootstrap/base.py
```diff
@@ -13,15 +13,30 @@
...
     'ResourceRegistrar'
 ]
 
+LOG = logging.getLogger(__name__)
+
+# Note: We use the cache to avoid manipulating the DB object for the same pack multiple times
+# during the same register-content run.
+# This works fine since those classes are only uses from register-content which is a script and not
+# a long running process.
+REGISTERED_PACKS_CACHE = {}
```
manasdk: set?

st2common/st2common/bootstrap/base.py
```diff
@@ -45,3 +60,64 @@ def get_resources_from_pack(self, resources_dir):
...
+        manifest_path = os.path.join(pack_dir, MANIFEST_FILE_NAME)
+
+        if not os.path.isfile(manifest_path):
+            raise ValueError('Pack "%s" is missing %s file' % (pack_name, MANIFEST_FILE_NAME))
+
+        content = self._meta_loader.load(manifest_path)
+        if not content:
+            raise ValueError('Pack "%s" metadata file is empty' % (pack_name))
+
+        content['ref'] = pack_name
```
manasdk: I guess pack_name works for now but the usual ResourceReference would fail - right?

st2common/st2common/content/aliasesregistrar.py
```diff
@@ -43,8 +43,10 @@ def register_aliases_from_packs(self, base_dirs):
         :return: Number of aliases registered.
         :rtype: ``int``
         """
-        registered_count = 0
+        # Register packs first
+        self.register_packs(base_dirs=base_dirs)
```
manasdk: I guess having to do this in every registrar is annoying but works for now.

manasdk:
Great!

Thanks for picking up the change :)

st2api/st2api/controllers/v1/packs.py
```diff
@@ -0,0 +1,35 @@
...
+from st2api.controllers.resource import ResourceController
+from st2common.models.api.pack import PackAPI
+from st2common.persistence.pack import Pack
+
+__all__ = [
+    'PacksController'
+]
+
+
+class PacksController(ResourceController):
```
manasdk: Ah good ol' inheritance :P

This class has literally no method and everything is from the base class. Something very strange about that one ... also I am sure the linter must be unhappy seeing no public methods?

st2common/st2common/bootstrap/base.py
Kami: Yep.

st2common/st2common/content/aliasesregistrar.py
Kami: Yeah, sadly I couldn't come up with an approach which is much better than this.

We do have a base class with some shared code for resource registrars now, but sadly each class still duplicates a lot of code.

st2api/st2api/controllers/v1/packs.py
Kami: Linter is fine with it - the methods are not abstract, just the properties are.

st2common/st2common/bootstrap/base.py
Kami: On a related note, in the RBAC PR I will amend (and eventually replace) this with the UID.

I couldn't make that change here already because other UID changes are in a different branch and we have a chicken and the egg problem.

PR #1702, opened 2015-07-08
Fix a bug with actions update API endpoint throwing on missing "enabled" attribute

Kami:
If a user didn't specify a value for `enabled` attribute when updating an action, an exception would be throw and 500 internal server error returned to the user.

This pull request fixes that by specifying a default value inside the API model and not inside the controller (previously, the controller would only set it on create, but not on update).


dzimine:
Thanks!

PR #1703, opened 2015-07-08
Fix late-assignment of the variables on the model instances

Kami:
This branch updates all the `to_model` methods on the API models and other related code to pass all the model arguments to the constructor instead of performing a late variable assignment on the model instance.

Late variable assignment shouldn't be used unless there is a good and valid use case for it (same as with the decorators) since it breaks many assumptions and "static" analysis. We don't know if and when variable will be set which makes constructors useless.

For example, if I want to create a new attribute based on the value of other attributes and assign it inside the constructor, I can't do that when using late variable assignment approach. Only way to get this to kinda consistently work with a late variable assignment is to instead implement `__getattribute__`, but that's a hack on top of another hack.

P.S. 1: I noticed @m4dcoder encounter a similar issue when working Policy models. He assigns `ref` attribute inside the constructor and to get this to work, he couldn't do late variable assignment - https://github.com/StackStorm/st2/blob/master/st2common/st2common/models/api/policy.py#L135

P.S. 2: We have a similar issue with the runners, but we can fix that later. I couldn't wait with fixing this though since I have some related changes in the RBAC branch which wouldn't be possible without this change (well, I could write a hack on top of a hack, but that's awful and we already have enough debt in other places).

### DIFF
--- a/st2common/st2common/models/api/auth.py
+++ b/st2common/st2common/models/api/auth.py
@@ -41,6 +41,12 @@ class UserAPI(BaseAPI):
         "additionalProperties": False
     }
 
+    @classmethod
+    def to_model(cls, user):
+        name = user.name
+        model = cls.model(name=name)
+        return model
+
 
 class TokenAPI(BaseAPI):
     model = TokenDB
@@ -79,10 +85,10 @@ class TokenAPI(BaseAPI):
         return cls(**doc)
 
     @classmethod
-    def to_model(cls, token):
-        model = super(cls, cls).to_model(token)
-        model.user = str(token.user) if token.user else None
-        model.token = str(token.token) if token.token else None
-        model.ttl = getattr(token, 'ttl', None)
-        model.expiry = isotime.parse(token.expiry) if token.expiry else None
+    def to_model(cls, instance):
+        user = str(instance.user) if instance.user else None
+        token = str(instance.token) if instance.token else None
+        expiry = isotime.parse(instance.expiry) if instance.expiry else None
+
+        model = cls.model(user=user, token=token, expiry=expiry)
         return model
--- a/st2common/st2common/models/api/execution.py
+++ b/st2common/st2common/models/api/execution.py
@@ -125,7 +125,7 @@ class ActionExecutionAPI(BaseAPI):
 
     @classmethod
     def to_model(cls, instance):
-        model = cls.model()
+        values = {}
         for attr, meta in six.iteritems(cls.schema.get('properties', dict())):
             default = copy.deepcopy(meta.get('default', None))
             value = getattr(instance, attr, default)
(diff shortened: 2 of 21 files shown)


st2common/st2common/models/api/auth.py
```diff
@@ -79,10 +85,10 @@ def from_model(cls, model, mask_secrets=False):
         return cls(**doc)
 
     @classmethod
-    def to_model(cls, token):
-        model = super(cls, cls).to_model(token)
-        model.user = str(token.user) if token.user else None
-        model.token = str(token.token) if token.token else None
-        model.ttl = getattr(token, 'ttl', None)
```
Kami: `TokenDB` class doesn't even have ttl attribute / field.

st2common/st2common/models/api/execution.py
```diff
@@ -136,7 +136,10 @@ def to_model(cls, instance):
...
             if attr not in ActionExecutionAPI.SKIP:
-                setattr(model, attr, value)
-        model.start_timestamp = isotime.parse(instance.start_timestamp)
-        model.end_timestamp = isotime.parse(instance.end_timestamp)
+                values[attr] = value
+
+        values['start_timestamp'] = isotime.parse(instance.start_timestamp)
+        values['end_timestamp'] = isotime.parse(instance.end_timestamp)
+
+        model = cls.model(**values)
```
manasdk: I really don't get why this is a better pattern. This clearly looks like an expensive approach to construction given you create an extra object.

Is there some benefit to this approach that I am missing?

Kami: @manasdk I posted an explanation on Slack two days ago and there is some in 553c57f1c9de9934867a34fdd2069d644b1d2b1a.

In short, previously we had no idea when (and if) the variables were available and this made constructor useless. If we wanted to, for example, generate and assign a derived field based on values of other field values, we couldn't do that consistently. We could, for example, do that in `save()`, but that's mongoengine specific and that field would only be available after save is called.

As far as "expensivness" goes, if you are referring to a temporary dict here and temporary variables in other places, then this is a non-issue. We have 999 other places where performance might eventually become an issue (logging, queries, etc.), but this is not one of them.

manasdk:
> As far as "expensivness" goes

My point there was it does not seem better to me and looks more expensive so what the heck is the value.

> In short, previously we had no idea when (and if) the variables were available 

We still have no guarantee. e.g https://github.com/StackStorm/st2/blob/master/st2common/st2common/models/db/action.py#L43

the constructor is in the base class and will assign any property as is passed in via kwargs. So still no guarantee apart from a convention. I am ok with adopting a convention but practically this offers no better guarantees than before. Knowing if or when a value is assigned is still best practice.

>and this made constructor useless.

Object construction has many paradigms. Useless constructor is a fairly reasonable pattern (factory, builder etc) so I do not buy this argument.

I am not against the pattern you employ here, I am fine with the choice. I am trying to understand the choice and why it was made e.g. if this were java I would almost always reject a constructor pattern and push to the builder pattern or beans style. The guarantees you are looking for don't seem to be offered by your approach - unless we come up with some sort of contract which explicitly calls out when an object is ready or we start defining __init__ for every DB object (however mongoengine would most likely not work).

PR #1705, opened 2015-07-09
RFR: Coverage reports html

lakshmi-kannan:
HTML coverage reports will be available in $repo_root/cover/*.html

XML coverage report will be in $repo_root/coverage.xml


Kami:
Sweet.

PR #1708, opened 2015-07-10
adds st2auth to the list of components that can be restarted

Itxaka:
Looks like st2auth is missing from the restart-component checks, so you are unable to restart only that component.

@Kami ?


Kami:
The change looks good to me, I just need to test it.

Kami:
Seems to be working fine.

Merged into trunk. Thanks.

PR #1712, opened 2015-07-10
Fixes #1654, adds no timeout for action runners

### LINKED ISSUE #1654: Infinite timeout in action parameters
The problem is that parameter `timeout: 0` in action is considered as 60 seconds and there is no option to make it "infinite", eg. no timeout.

Here is example action `actions/test_timeout.yaml`:
```yaml
---
name: test_timeout
runner_type: local-shell-cmd
entry_point: ""
enabled: true
parameters:
  timeout:
    # works as default 60 seconds
    default: 0
  cmd:
    description: "Command to run"
    immutable: true
    default: "sleep 90 && echo 'done!'"
```

Running it: 
```sh
# st2 run st2-pack-bug.test_timeout
..............................
id: 5586ca9b0a84b40a4a08a168
status: failed
result:
{
    "succeeded": false,
    "stdout": "",
    "return_code": -9,
    "failed": true,
    "stderr": "",
    "error": "Action failed to complete in 60 seconds"
}
```

Not sure if it's bug or not, but proposal is to consider `timeout: 0` as "no timeout", instead of default 60 seconds. Lots of cases when action doesn't needs any timeouts.

Because things like this in action:
```yaml
parameters:
  timeout:
    default: 99999999
```
definitely looks like there should be a better way.

armab:
Adds an option to run actions with no timeout.

Also fixes an issue #1654 , when runner parameters were not overridden with "falsey" action parameters (0, False, etc), so this meta was buggy:
```yaml
---
  name: "test"
  description: "Action that tests timeout."
  enabled: true
  entry_point: ""
  parameters:
    timeout:
      # this worked as default 60 seconds
      #default: 0

      # this worked as 0 seconds
      # default: "0"

      # now 0 means no timeout
      default: 0
    cmd:
      default: "echo 'TIMEOUT: {{timeout}}'"
```

### DIFF
--- a/st2common/st2common/util/green/shell.py
+++ b/st2common/st2common/util/green/shell.py
@@ -57,7 +57,7 @@ def run_command(cmd, stdin=None, stdout=subprocess.PIPE, stderr=subprocess.PIPE,
                 environment from the current process is inherited.
     :type env: ``dict``
 
-    :param timeout: How long to wait before timing out.
+    :param timeout: How long to wait before timing out. 0 means no timeout.
     :type timeout: ``float``
 
     :param preexec_func: Optional pre-exec function.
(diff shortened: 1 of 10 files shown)


st2common/st2common/util/green/shell.py
```diff
@@ -80,29 +80,30 @@ def run_command(cmd, stdin=None, stdout=subprocess.PIPE, stderr=subprocess.PIPE,
...
-
-    timeout_thread = eventlet.spawn(on_timeout_expired, timeout)
-    stdout, stderr = process.communicate()
-    timeout_thread.cancel()
-    exit_code = process.returncode
-
-    if exit_code == TIMEOUT_EXIT_CODE:
-        timed_out = True
+    if not timeout:
+        stdout, stderr = process.communicate()
```
lakshmi-kannan: Isn't this dangerous? We should pretty much not support unlimited timeout. Wouldn't this block the worker?

armab: I might not understand all the low-level kitchen.
Logically is there any difference between this and big timeout like: `timeout: 999999999` ?

lakshmi-kannan: Yeah, I was also thinking about a case when timeout is given a large value. We might need to implement some sort of a max timeout. I'd like to hear from others about this as well.

armab: Yeah, @Kami commented here, same idea: https://github.com/StackStorm/st2/issues/1654#issuecomment-113996779

My understanding from user's side, there should be an option like: 
> Run my action until it's finished, never block it.

This is standard behavior for example for configuration management tools.

Aren't action runners multi-threaded by default, eg. I can run many actions at the same time in parallel, no?

lakshmi-kannan: We achieve true parallelism by asking users to run multiple workers at the same time. Each worker takes a message from the queue and does use a green thread to process the message (kicking off the execution). So yeah, you are right that it should just block a thread with infinite timeout. But if you run multiple such actions, you'd pretty much start blocking the pool.

I also think there's a bug in our current implementation that we ack early. See https://github.com/StackStorm/st2/blob/master/st2common/st2common/transport/consumers.py#L52. I am afraid this might mean we tell st2 that action is done or kicked off while it might still be in the dispatcher queue waiting for an eventlet to work on. We are probably not hitting the case because pool size is 50. This is a different issue but your question is tangentially related.

armab:
@Kami @lakshmi-kannan 
As I understand allowing shell action runner to have no timeout is kind of vulnerable/arguable topic :cactus: 

What can I do to make it productive is to extract 2 commits from this PR that fixes small bug with default parameters (+ unit tests), see:
https://github.com/armab/st2/commit/f3d925a700bf8885a5c1317300d62695634046d2
https://github.com/armab/st2/commit/d13795ff520f632a59ac948f6181de707a574f99

and leave "infinite timeout" for better times.
Would that work?

Kami:
@armab Yeah that works and would be great.

Please open a separate PR with a corresponding changelog entry and we will get it merged.

lakshmi-kannan:
Agree with @Kami.

PR #1713, opened 2015-07-11
Fix Coveralls & Optimize Travis builds

armab:
I started to dig into Coveralls.io problem we recently discussed before with @lakshmi-kannan, when only files from one st2package were included in coverage report, but had more fun and curiosity and optimized travis builds also.
Surprising that some bits of code were incompatible with Python2.6, despite 2.6 is included in Travis build matrix.

- [x] [Fix Coveralls coverage, all files are included now](https://coveralls.io/builds/3034692) [Coverage Status](https://coveralls.io/github/armab/st2?branch=coveralls)
- [x] Add compile .py files task, useful as syntax check
- [x] Optimize Travis build matrix, use exclude
- [x] Merge `flake8`, `pylint`, `docs` tasks into single build (`make requirements` is expensive)

##### Travis builds before:
travis

##### Travis builds after:
travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,24 +1,26 @@
 language: python
 sudo: false
 python:
-  - "2.6"
-  - "2.7"
-install:
-  - pip install python-coveralls
+  - 2.7
+
 env:
-  global:
-    - "PIP_DOWNLOAD_CACHE=$HOME/.pip-cache"
-  matrix:
-    - "TASK=flake8"
-    - "TASK=pylint"
-    - "TASK=docs"
-    - "TASK=tests"
-cache:
-  directories:
-    - $HOME/.pip-cache/
-script:
-  - ./scripts/travis.sh
+  - TASK=checks
+  - TASK=unit
+
 services:
   - mongodb
   - rabbitmq
-after_success: coveralls
+
+cache:
+  directories:
+    - $HOME/.cache/pip/
+
+install:
+ - if [[ ${TASK} == 'unit' ]]; then pip install python-coveralls; fi
+ - make requirements
+
+script:
+  - ./scripts/travis.sh
+
+after_success:
+  - if [[ ${TASK} == 'unit' ]]; then coveralls; fi
--- a/scripts/travis.sh
+++ b/scripts/travis.sh
@@ -1,27 +1,18 @@
 #!/usr/bin/env bash
 
-function version_ge() { test "$(echo "$@" | tr " " "\n" | sort -V | tail -n 1)" == "$1"; }
-
 if [ -z ${TASK} ]; then
   echo "No task provided"
   exit 2
-  fi
-
-# Note: We only want to run tests under multiple Python versions
-
-if version_ge ${TRAVIS_PYTHON_VERSION} "2.7" && [ ${TASK} != "tests" ]; then
-    echo "Skipping task ${TASK} on ${TRAVIS_PYTHON_VERSION}"
-    exit 0
 fi
 
-if [ ${TASK} == "flake8" ]; then
-  make flake8
-elif [ ${TASK} == "pylint" ]; then
-  make pylint
-elif [ ${TASK} == "docs" ]; then
-  make docs
-elif [ ${TASK} == "tests" ]; then
-  make tests-travis
+if [ ${TASK} == 'checks' ]; then
+  make .flake8
+  make .pylint
+  make .docs
+elif [ ${TASK} == 'unit' ]; then
+  # compile .py files, useful as compatibility syntax check
+  make compile
+  make unit-tests-coverage-xml
 else
   echo "Invalid task: ${TASK}"
   exit 2
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,4 @@
+; http://nedbatchelder.com/code/coverage/config.html
+
+[run]
+omit = *tests/*
--- a/Makefile
+++ b/Makefile
@@ -122,6 +122,12 @@ lint: requirements .lint
 .PHONY: clean
 clean: .cleanpycs .cleandocs
 
+.PHONY: compile
+compile:
+	@echo "======================= compile ========================"
+	@echo "------- Compile all .py files (syntax check test) ------"
+	@if python -c 'import compileall,re; compileall.compile_dir(".", rx=re.compile(r"/virtualenv"), quiet=True)' | grep .; then exit 1; else exit 0; fi
+
 .PHONY: .cleanpycs
 .cleanpycs:
 	@echo "Removing all .pyc files"
@@ -218,10 +224,10 @@ tests: pytests
 tests-travis: requirements unit-tests-coverage-xml
 
 .PHONY: pytests
-pytests: requirements .flake8 .pylint .pytests-coverage
+pytests: compile requirements .flake8 .pylint .pytests-coverage
 
 .PHONY: .pytests
-.pytests: unit-tests itests clean
+.pytests: compile unit-tests itests clean
 
 .PHONY: .pytests-coverage
 .pytests-coverage: unit-tests-coverage-html itests clean
--- a/Makefile
+++ b/Makefile
@@ -252,7 +252,7 @@ unit-tests-coverage-xml:
 		echo "Running tests in" $$component; \
 		echo "==========================================================="; \
 		. $(VIRTUALENV_DIR)/bin/activate; nosetests -sv --with-coverage \
-			--cover-inclusive --cover-erase --cover-xml \
+			--cover-inclusive --cover-xml \
 			--cover-package=$$component $$component/tests/unit &&  \
 			mv coverage.xml coverage-$$component.xml || exit 1; \
 	done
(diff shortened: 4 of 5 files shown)


lakshmi-kannan:
I dig this PR. What was the issue with coverage not ending up in coveralls.io? Was it --cover-erase option?

manasdk:
Do we need py2.6 compatibilty? I doubt we really care and I sorta like the `{}` dictionary syntax.

Kami:
> Merge flake8, pylint, docs tasks into single build (make requirements is expensive)

@armab The whole idea of splitting it into multiple builders was to parallelize and speed up the build. We use container based build infrastructure and dependency caching (see #1457 and #1461) so container spin up times and installing dependencies should be fast.

.travis.yml
```diff
@@ -1,24 +1,35 @@
...
-    - "PIP_DOWNLOAD_CACHE=$HOME/.pip-cache"
-  matrix:
-    - "TASK=flake8"
-    - "TASK=pylint"
-    - "TASK=docs"
-    - "TASK=tests"
+  - TASK=checks
+  - TASK=unit
+
+matrix:
```
Kami: Nice.

scripts/travis.sh
```diff
@@ -1,26 +1,18 @@
...
-  make docs
-elif [ ${TASK} == "tests" ]; then
+if [ ${TASK} == 'checks' ]; then
+  make requirements
+  make .flake8
+  make .pylint
+  make .docs
+elif [ ${TASK} == 'unit' ]; then
+  # compile .py files, useful as py2.6 compatibility syntax check
+  python -c "import compileall; compileall.compile_dir('.', quiet=True)" | grep . && exit 1
```
Kami: Ideally, we could add this check to the Makefile so it can also be ran locally.

Kami:
Eventually, I would like us to migrate to something like `tox` and run the same targets / checks everywhere (locally, on Travis CI and on our CI server).

Using Travis CI is nice, but usually when you start doing more complex stuff and want to speed up the builds, the local and Travis targets start to divert significantly which makes debugging and reproducing issues and test failures hard.

.travis.yml
```diff
@@ -1,24 +1,35 @@
...
-  - "2.6"
-  - "2.7"
-install:
-  - pip install python-coveralls
+  - 2.6
+  - 2.7
+
 env:
-  global:
-    - "PIP_DOWNLOAD_CACHE=$HOME/.pip-cache"
```
Kami: I assume you removed it because that has been a default while in pip for a while now?

Kami:
Re Python 2.6 compatibility - yeah, we need to decide what we need and want.

I thought some of our current users run it on older versions of RHEL which still ship 2.6 by default, but I could be wrong (IIRC, @DoriftoShoes built custom 2.7 RHEL packages).

Probably also worth including @DoriftoShoes and @jfryman here. Is running it under Python 2.7 an issue for some of the people you talked to? Or can we just says 2.7 is the minimum supported version and be done with that?

Kami:
Also, Python 2.6 compatibility related changes in this branch are not a blocker. I'm fine with merging that.

It's just something we should decide on soon and document it. We can update the build targets later.

armab:
> I dig this PR. What was the issue with coverage not ending up in coveralls.io? Was it --cover-erase option? 

Yep. Every run of `nosetests --with-coverage` generates `.coverage` file, which is significant. When you run several tests, - they are combined automatically into `.coverage` file. `--cover-erase` deletes that file before every run. Was surprised to find that it works by default, without need to merge anything manually.


> @armab The whole idea of splitting it into multiple builders was to parallelize and speed up the build. We use container based build infrastructure and dependency caching (see #1457 and #1461) so container spin up times and installing dependencies should be fast.

Yeah, I understand.
But in fact there wasn't speed up, see the screenshots. For every `flake8, pylint, docs` execution there was very long-running and expensive `make requirements`. Just flake or pylint or docs build runs in several seconds, while requirements takes 90% of the time. I investigated this case.
Also it's important to remember that with Travis we have 4 parallel builds. Eg. when first 4 tasks are in progress - other tasks are waiting.
So combining them was improvement in this case.

[Old build](https://camo.githubusercontent.com/42a1daf0beba93748c7f385058f070fc0d3b2b66/687474703a2f2f692e696d6775722e636f6d2f443154393436502e706e67) runs for `~4+ mins` (calculate time, considering 4 parallel builds and 8 total tasks).
[New build](https://camo.githubusercontent.com/b9085edf7497d432b8e7e32dd0372716a971bee5/687474703a2f2f692e696d6775722e636f6d2f6a62537a5a48592e706e67) runs for `2 min 27sec` (max time, considering we have 3 tasks and 4 parallel builds).

----

About Python 2.6. Coveralls fails under python2.6 build unless there are syntax errors in code. If we won't support it - then it's even better. I'll revert those syntax changes and we could remove 2.6 from Travis matrix and have more free slots for future Travis builds, for example for integration or end-to-end tests.

armab:
Also yeah. RHEL6/CentOS6 is shipped with Python 2.6 version by default. It feels like RHEL6/CentOS6 is still very commonly used.
```
# cat /etc/*release*
CentOS release 6.6 (Final)
# python --version
Python 2.6.6
```

> Just related info, [Docker drops support for RHEL6/CentOS6](https://github.com/docker/docker/issues/14365).

.travis.yml
armab: Yes.

In Travis logs:
```
  DEPRECATION: --download-cache has been deprecated and will be removed in the future. Pip now automatically uses and configures its cache.
```

scripts/travis.sh
armab: True.
Included in Makefile, thanks for the notice.

lakshmi-kannan:
I am +1 to merging this. I am pretty meh about supporting 2.6 but I want @DoriftoShoes to weigh in. He has more experience supporting RHEL customers.

jfryman:
I think this is a good PR, and we should merge. If anything it gets us faster builds and a nice coverage box.

As far as 2.6 support... there is more to do iirc before we can :heavy_check_mark: the box with full support. But, it's a good step. WRT support, we'll be taking a the installer approach that is currently under development, so less of a concern IMO.

manasdk:
Why do we want to support python 2.6. It will impact libraries we choose and there will be a constant struggle with code. If there is anything we should care about is python 3 support.

jfryman:
@manasdk I don't want to support 2.6. If anything, it's best effort. My point being is that as long as it's not a blocker, then we press forward. Our strategy for RHEL/CentOS 6.x doesn't need to be native 2.6 support.

st2actions/st2actions/policies/concurrency_by_attr.py
```diff
@@ -46,9 +46,9 @@ def _get_lock_uid(self, target):
         return json.dumps(meta)
 
     def _get_filters(self, target):
-        filters = {('parameters__%s' % k): v
-                   for k, v in six.iteritems(target.parameters)
-                   if k in self.attributes}
+        filters = dict((('parameters__%s' % k), v)
```
manasdk: Why is this change needed?

jfryman:
To the point of this PR. I'm :+1: on the coverage and speed increases. :-1: on the Python 2.6 support.

manasdk:
@armab This PR tackles atleast 3 separate issues -
1. Fix coverage reporting
2. Improve travis build efficiency
3. Python 2.6 support

I guess too much is happening here and could've been separate PRs.

IMO we should not make any code changes for python 2.6 support or at least even if we fix the code now  no guarantee anyone will write python 2.6 compatible code. Also, the question is why bother? I suspect truly supporting python 2.6 will also require all our dependencies to support python 2.6. We can figure out the RHEL 6 answer via installer.

armab:
I'm personally +1 to not support py 2.6. I just included syntax fixes, because Python 2.6 already existed in Travis build matrix.
Btw I have unit and integration tests running under py2.6.

But from the other side OpenSource guys who use st2 + CentOS6 will have problems. From yesterday IRC:
> 12:25: eMptywee	doesn't really work well on RHEL 6.4
12:26: eMptywee	it's okay, I just had to fix it slightly
...
12:27: eMptywee	err, it's dirty. Mostly because default python on RHEL6 is 2.6
12:27: eMptywee	e.g. changing shebang to python2.7
12:29: eMptywee	well, I don't really need python 2.6 compatibility. It's just when st2 scripts use #!/usr/bin/python, it doesn't work on RHEL6 :)
12:29: eMptywee	even under virtualenv for some reason

manasdk:
> But from the other side OpenSource guys who use st2 + CentOS6 will have problems.

We will have to work out options for those folks and be somewhat prescriptive.

I have no issues with the changes going in except that there is no guarantee we continue to write compatible code.

armab:
Removed python 2.6 from Travis build matrix, as well as code syntax changes.

Also, how about `python 2.7` badge, to clarify better supported version:
python 2.7 badge

lakshmi-kannan:
+1 to explicit python badge. I am still unsure about the plan for supporting python 2.6. I'd really like to hear from @DoriftoShoes about this.

manasdk:
+1

DoriftoShoes:
I don't think we need to worry about 2.6 support.  Even the customers that we have deployed on EL6 are using custom installations on Python 2.7.

lakshmi-kannan:
+1 then.

armab:
merge it!

armab:
Thanks everyone for discussion!

PR #1714, opened 2015-07-12
Add verification before deleting LOGDIR contents

kirilstrax:
I had no files in the ${LOGDIR}, which caused rm ${LOGDIR}/* to fail
Added a verification: only attempt deletion, if there are files in the folder

### DIFF
--- a/tools/launchdev.sh
+++ b/tools/launchdev.sh
@@ -183,7 +183,9 @@ function st2clean(){
     mongo st2 --eval "db.dropDatabase();"
     # start with clean logs
     LOGDIR=$(dirname $0)/../logs
-    rm ${LOGDIR}/*
+    if [ "$(ls -A ${LOGDIR} | wc -l)" != "0" ]; then
+        rm ${LOGDIR}/*
+    fi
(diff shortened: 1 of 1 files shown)


tools/launchdev.sh
```diff
@@ -183,7 +183,9 @@ function st2clean(){
     mongo st2 --eval "db.dropDatabase();"
     # start with clean logs
     LOGDIR=$(dirname $0)/../logs
-    rm ${LOGDIR}/*
+    if [ "$(ls -A ${LOGDIR} | wc -l)" != "0" ]; then
```
lakshmi-kannan: if [ -d $LOGDIR ] should test directory exists. Wondering why the ls -A | wc -l?

lakshmi-kannan: or you could simply use rm -rf.

Kami: Yeah `-d` should be sufficient.

lakshmi-kannan:
Closing in favor of https://github.com/StackStorm/st2/pull/1874

PR #1716, opened 2015-07-13
RFR - Initial commit to pass context when rendering params

lakshmi-kannan:


st2actions/st2actions/container/base.py
```diff
@@ -67,9 +67,6 @@ def dispatch(self, liveaction_db):
         return liveaction_db.result
 
     def _do_run(self, runner, runnertype_db, action_db, liveaction_db):
-        # Finalized parameters are resolved and then rendered.
```
manasdk: Seems correct.

This change was done https://github.com/StackStorm/st2/commit/8b1d04b08b43d15cccb1671b013de91660c56568#diff-ef7b3e3b028fe99ecaed2faaad32fd11

as well but somehow got dropped on the next commit.

manasdk:
I suspect that if the right context is missing the render will fail and the execution will fail. I am not sure we want that behavior - this would mean that an action which assumes execution for chat cannot ever work in a different context.

Could you confirm the behavior and add a test-case?

lakshmi-kannan:
>this would mean that an action which assumes execution for chat cannot ever work in a different context.

According to the action, it's a parameter that is needed for the action to run. If they want to run from a non-chat interface, they need to supply a value for the param. I don't see a problem with this behavior. Anything else would be confusing IMO.

manasdk:
Hmm ... yeah I am ok with the behavior. Could you capture that in a test-case. So we don't inadvertently make a change that reverses this choice.

lakshmi-kannan:
>Hmm ... yeah I am ok with the behavior. Could you capture that in a test-case. So we don't inadvertently make a change that reverses this choice.

Yes, good point. I'll add one.

st2actions/st2actions/container/base.py
lakshmi-kannan: heh. For everything unknown, it's always safe to blame @Kami. </troll>

lakshmi-kannan:
@manasdk Added the test. I'll merge. As always leave comments if I need to fix things.

st2api/st2api/controllers/exp/aliasexecution.py
```diff
@@ -101,16 +110,17 @@ def _get_notify_field(self, payload):
...
 
-    def _schedule_execution(self, action_alias_db, params, notify):
+    def _schedule_execution(self, action_alias_db, params, notify, context):
         try:
             # prior to shipping off the params cast them to the right type.
             params = action_param_utils.cast_params(action_ref=action_alias_db.action_ref,
                                                     params=params,
                                                     cast_overrides=CAST_OVERRIDES)
-            context = {
-                'action_alias_ref': reference.get_ref_from_model(action_alias_db),
```
Kami: It would probably be better to always include those two attributes here.

Otherwise we need to add a docstring to the function which says that the `context` which is passed to this function must contain those two attributes.

Kami:
Some docs would also be nice.

Kami:
Also, I still haven't decided if I like the "alias execution context" change yet.

Main problem I have with it is that now some of those attributes are only available when the action is executed inside chat which makes actions less portable. I know we already did some of that before, but in a different manner (wrapper action for actions exposed via ChatOps), but I wish we could come up with a better pattern which doesn't reduce actions "portability".

PR #1718, opened 2015-07-14
Add NG_INIT flag for migration to init scripts

jfryman:
This PR adds a new flag to our old trusty `st2ctl` command, which is bound to be with us for some time longer.

In lieu of this, we're about to add real, honest to goodness init scripts for all of our services to the native packages. However, it's going to take a bit of :frog: jumping around to get to the final state.

This first commit adds an internal-only flag to the init script. giA bit ugly, but it's a way to allow testing to happen in parallel while not breaking existing behavior.

### DIFF
--- a/tools/st2ctl
+++ b/tools/st2ctl
@@ -7,8 +7,6 @@ WEBUI_LOG_FILE="/var/log/st2/st2web.log"
 COMPONENTS="st2actionrunner st2api st2auth st2sensorcontainer st2rulesengine mistral st2web st2resultstracker st2notifier"
 STANCONF="/etc/st2/st2.conf"
 PYTHON=`which python`
-NG_INIT=false
-
 
 # AR is assumed to be an environment variable.
 if [ -z "$AR" ];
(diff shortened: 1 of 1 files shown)


lakshmi-kannan:
I am +1. Where are the init scripts for components themselves?

jfryman:
@lakshmi-kannan they are in two places. The first (and temporarily, you'll see them in the puppet-st2 PR that is referenced in this PR). Ultimately, however, they'll land in whatever version of https://github.com/dennybaa/st2-packages ends up as. We can talk about the right final resting ground, but everything is composed right now.

jfryman:
I added another flag `ST2_DISABLE_HTTPSERVER` so that I can control the loading of the simple HTTP. Tested and works fine.

manasdk:
ng = next gen?

jfryman:
@manasdk yeah. I'm getting fancy, but also trying to just avoid namespace problems.

manasdk:
you know we used next gen in a past project. Then it became current gen but somehow the next gen name stuck. So whenever someone uses next gen I always wonder what happens to the name once it becomes current gen. The problem really shows up if next gen gets stuck in the lexicon.

Anyway, just a data point name as you wish :)

manasdk:
oh +1

PR #1720, opened 2015-07-15
Additional detail on configuring trusted SSL CA

jfryman:
As reported in https://github.com/StackStorm/st2contrib/issues/240,
not all environments are treated equally. Many environment fall into a
scenario where a custom CA must be trusted by the OS and Applications,
or else validation will fail and connections will be rejected. To that
end, this commit adds detail on how to configure the underlying Python
`requests` library to load a custom SSL CA bundle into the OpenSSL trust
chain.

h/t @amaline :tophat:


lakshmi-kannan:
LGTM.

PR #1726, opened 2015-07-16
Fix Travis build script

armab:
Exit immediately if any command in `scripts/travis.sh` exits with a non-zero status.

My fail with recent changes, see: 
https://travis-ci.org/StackStorm/st2/jobs/71310549#L156

### DIFF
--- a/scripts/travis.sh
+++ b/scripts/travis.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -e
 
 if [ -z ${TASK} ]; then
   echo "No task provided"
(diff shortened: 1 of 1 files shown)


armab:
Also @lakshmi-kannan coveralls fails when coverage decreased (-0.01%).

Seems increasing threshold in Coveralls settings just a bit higher should help: https://github.com/lemurheavy/coveralls-public/issues/565#issuecomment-120924694

lakshmi-kannan:
Changed decrease threshold to 2%

lakshmi-kannan:
@armab Can you add a dummy commit so we can validate?

armab:
I'll merge, since it's hotfix :hotsprings:

PR #1730, opened 2015-07-17
WIP: Python client library for st2. 

lakshmi-kannan:
I am actually establishing overall how the new client library is going to look. I just started hacking things to satisfy my urge. Mostly not ready for review.

### DIFF
--- a/st2pyclient/st2pyclient/http_client/authed.py
+++ b/st2pyclient/st2pyclient/http_client/authed.py
@@ -1,0 +1,64 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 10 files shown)


st2pyclient/st2pyclient/http_client/authed.py
```diff
@@ -0,0 +1,64 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from st2pyclient.http_client.base import BaseClient
+
+
+class AuthedClient(BaseClient):
```
manasdk: Might be better to compose or even use a mixin for auth.

Kami:
I would prefer to have two packages - `st2client` (Python client library) and `st2cli` (CLI which depends on the client library).

I assume you used `st2pyclient` name to avoid renaming and conflicts with the existing `st2client` package. I'm fine with doing that right, but before any code starts depending on it we should sort out the names.

lakshmi-kannan:
>I would prefer to have two packages - st2client (Python client library) and st2cli (CLI which depends on the client library).

Yes, that's the plan. 

>I assume you used st2pyclient name to avoid renaming and conflicts with the existing st2client package. I'm fine with doing that right, but before any code starts depending on it we should sort out the names.

Aye aye.

Development in my fork. Closing this PR. Will open a new PR in my fork.

PR #1731, opened 2015-07-17
Display friendly message for error in action parameters validation

m4dcoder:
Display only the human friendly message from the jsonschema validation error. Remove the unicode prefix in the string.

```
~/st2$ st2 run core.local
MESSAGE: 'cmd' is a required property

~/st2$ st2 run default.ping count=x
MESSAGE: Additional properties are not allowed ('count' was unexpected)
```

### DIFF
--- a/st2api/st2api/controllers/v1/actionexecutions.py
+++ b/st2api/st2api/controllers/v1/actionexecutions.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import copy
+import re
 
 import jsonschema
 from oslo_config import cfg
@@ -100,7 +101,7 @@ class ActionExecutionsControllerMixin(BaseRestControllerMixin):
             abort(http_client.BAD_REQUEST, str(e))
         except jsonschema.ValidationError as e:
             LOG.exception('Unable to execute action. Parameter validation failed.')
-            abort(http_client.BAD_REQUEST, str(e))
+            abort(http_client.BAD_REQUEST, re.sub("u'([^']*)'", r"'\1'", e.message))
         except Exception as e:
             LOG.exception('Unable to execute action. Unexpected error encountered.')
             abort(http_client.INTERNAL_SERVER_ERROR, str(e))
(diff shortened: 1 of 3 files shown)


st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -100,7 +101,7 @@ def _handle_schedule_execution(self, execution):
             abort(http_client.BAD_REQUEST, str(e))
         except jsonschema.ValidationError as e:
             LOG.exception('Unable to execute action. Parameter validation failed.')
-            abort(http_client.BAD_REQUEST, str(e))
+            abort(http_client.BAD_REQUEST, re.sub("u'([^']*)'", r"'\1'", e.message))
```
lakshmi-kannan: heh.

lakshmi-kannan:
This is def better than what there is. So let's merge this now. I'd really like us to throw validation exceptions with better messages so we don't have to do reg exp foo at that layer. Another PR.

I am +1.

lakshmi-kannan:
Also, CHANGELOG needs an entry.

manasdk:
So, it used to be

```
(virtualenv)manas@st2dev:~$ st2 run core.local
MESSAGE: u'cmd' is a required property

Failed validating 'required' in schema['properties'][u'cmd']:
     'description': u'Action that executes an arbitrary Linux command on the localhost.',
     'properties': {u'cmd': {u'description': u'Arbitrary Linux command to be executed on the remote host(s).',
                             u'required': True,
                             u'type': u'string'},
                    u'cwd': {u'description': u'Working directory where the command will be executed in',
                             u'type': u'string'},
                    u'env': {u'description': u'Environment variables which will be available to the command(e.g. key1=val1,key2=val2)',
                             u'type': u'object'},
                    u'kwarg_op': {u'default': u'--',
                                  u'description': u'Operator to use in front of keyword args i.e. "--" or "-".',
                                  u'type': u'string'},
                    u'sudo': {u'immutable': True},
                    u'timeout': {u'default': 60,
                                 u'description': u"Action timeout in seconds. Action will get killed if it doesn't finish in timeout seconds.",
                                 u'type': u'integer'}},
     'title': u'local',
     'type': 'object'}

On instance[u'cmd']:
    {u'kwarg_op': u'--', u'timeout': 60}
```

manasdk:
+1

Kami:
On validation error it might also make sense to display all the available parameters for a particular action (same as we do if you use --help flag). This way we potentially save user 1 additional command.

Basically, same as we did before, but instead of displaying unfriendly json schema, we should use a human-friendly format (we already have a function for generating user-friendly format in the CLI).

PR #1732, opened 2015-07-20
add keystone auth backend

Itxaka:
Not sure how I would go about doing the tests as it needs a keystone endpoint. Should I just mock it?

### DIFF
--- a/st2auth/st2auth/backends/keystone.py
+++ b/st2auth/st2auth/backends/keystone.py
@@ -22,8 +22,8 @@ try:
     from urlparse import urlparse
     from urlparse import urljoin
 except ImportError:
-    from urllib.parse import urlparse
-    from urllib.parse import urljoin
+    from urllib.parse import urlparse  # pylint: disable=E0611
+    from urllib.parse import urljoin  # pylint: disable=E0611
 
 __all__ = [
     'KeystoneAuthenticationBackend'
--- a/st2auth/st2auth/backends/keystone.py
+++ b/st2auth/st2auth/backends/keystone.py
@@ -17,6 +17,12 @@
 from st2common import log as logging
 from st2auth.backends.base import BaseAuthenticationBackend
 import requests
+try:
+    from urlparse import urlparse
+    from urlparse import urljoin
+except ImportError:
+    from urllib.parse import urlparse
+    from urllib.parse import urljoin
 
 __all__ = [
     'KeystoneAuthenticationBackend'
@@ -39,21 +45,49 @@ class KeystoneAuthenticationBackend(BaseAuthenticationBackend):
         :param keystone_version: Keystone version to authenticate against (default to 2).
         :type keystone_version: ``int``
         """
+        url = urlparse(keystone_url)
+        if url.path != '' or url.query != '' or url.fragment != '':
+            raise Exception("The Keystone url {} does not seem to be correct.\n"
+                            "Please only set the scheme+url+port (e.x.: http://example.com:5000)".format(keystone_url))
         self._keystone_url = keystone_url
         self._keystone_version = keystone_version
 
     def authenticate(self, username, password):
-        creds = {
-            "auth": {
-                "passwordCredentials": {
-                    "username": username,
-                    "password": password
+        if self._keystone_version == 2:
+            creds = {
+                "auth": {
+                    "passwordCredentials": {
+                        "username": username,
+                        "password": password
+                    }
                 }
             }
-        }
-        login = requests.post(self._keystone_url, json=creds)
+            login = requests.post(urljoin(self._keystone_url, 'v2.0/tokens'), json=creds)
 
-        if login.status_code == 200:
+        elif self._keystone_version == 3:
+            creds = {
+                "auth": {
+                    "identity": {
+                        "methods": [
+                            "password"
+                        ],
+                        "password": {
+                            "domain": {
+                                "id": "default"
+                            },
+                            "user": {
+                                "name": username,
+                                "password": password
+                                }
+                            }
+                        }
+                    }
+                }
+            login = requests.post(urljoin(self._keystone_url, 'v3/auth/tokens'), json=creds)
+        else:
+            raise Exception("Keystone version {} not supported".format(self._keystone_version))
+
+        if login.status_code in [200, 201]:
             LOG.debug('Authentication for user "{}" successful'.format(username))
             return True
         else:
--- a/st2auth/st2auth/backends/keystone.py
+++ b/st2auth/st2auth/backends/keystone.py
@@ -19,7 +19,7 @@ from st2auth.backends.base import BaseAuthenticationBackend
 import requests
 import httplib
 
-from six.moves.urllib.parse import parse as urlparse
+from six.moves.urllib.parse import urlparse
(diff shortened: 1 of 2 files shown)


Kami:
Nice.

And yeah, as far as testing goes, we would need a mock endpoint. We have some mocks in Libcloud and in some of the Rackspace projects I've worked on, we started mock Keystone server as part of out integration test suite (https://github.com/racker/node-rproxy/blob/master/tests/mock-auth-server.js).

I believe there is now also an OpenStack project which provides some of the required mocks.

st2auth/st2auth/backends/keystone.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def __init__(self, keystone_url, keystone_version=2):
+        """
+        :param keystone_url: Url of the Keystone server to authenticate against.
+        :type keystone_url: ``str``
+        :param keystone_version: Keystone version to authenticate against (default to 2).
+        :type keystone_version: ``int``
+        """
+        self._keystone_url = keystone_url
+        self._keystone_version = keystone_version
```
Kami: Looks like this argument is unused and we only support v2 right now?

If so, we should make that clear and throw if an unsupported version is provided.

Kami: On a related note, it might be better to just use python-keystoneclient.

Ideally, eventually we would also switch to using setup.py entry_points for plugins and provide plugins as separate Python packages with it's own requirements.txt file, but for now I'm OK with putting keystone-client dependency to the shared requirements.txt file.

Itxaka: Actually, Im doing the v2/v3 split now and some other checks. I wanted the PR to go up quick due to the testing question.

Re: keystone client. Totally against that tbh. It turns a simple authentication plugin with one dependency into dependency hell and also ties into a big external projects, while requests is one package with minimal dependencies and probably already installed on every decent python env which needs to do a network call.

And this auth backend just needs to do one simple call. To have keystoneclient as dependency is overkill IMO.

st2auth/st2auth/backends/keystone.py
```diff
@@ -0,0 +1,100 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from st2common import log as logging
+from st2auth.backends.base import BaseAuthenticationBackend
+import requests
+
+try:
+    from urlparse import urlparse
```
Kami: We depend on six so you can instead use:

```python
from six.moves.urllib import parse as urlparse
from six.moves.urllib.parse import urljoin
```

st2auth/st2auth/backends/keystone.py
```diff
@@ -0,0 +1,100 @@
...
+                            }
+                        }
+                    }
+                }
+            }
+            login = requests.post(urljoin(self._keystone_url, 'v3/auth/tokens'), json=creds)
+        else:
+            raise Exception("Keystone version {} not supported".format(self._keystone_version))
+
+        if login.status_code in [200, 201]:
```
Kami: Minor style thing - `httlib.OK, httplib.CREATED` :)

st2auth/st2auth/backends/keystone.py
Itxaka: IDD! Much easier!

Itxaka:
@Kami Finally all passing with all changes requested :D

manasdk:
+1

Will wait for @kami to take a look and merge.

st2auth/st2auth/backends/keystone.py
```diff
@@ -0,0 +1,97 @@
...
+                            },
+                            "user": {
+                                "name": username,
+                                "password": password
+                            }
+                        }
+                    }
+                }
+            }
+            login = requests.post(urljoin(self._keystone_url, 'v3/auth/tokens'), json=creds)
```
lakshmi-kannan: Just curious, is it v3.0 or just v3? I see v2.0 for the version 2 APIs :). <3 to openstack.

st2auth/st2auth/backends/keystone.py
```diff
@@ -0,0 +1,97 @@
...
+        if url.path != '' or url.query != '' or url.fragment != '':
+            raise Exception("The Keystone url {} does not seem to be correct.\n"
+                            "Please only set the scheme+url+port "
+                            "(e.x.: http://example.com:5000)".format(keystone_url))
+        self._keystone_url = keystone_url
+        self._keystone_version = keystone_version
+
+    def authenticate(self, username, password):
+        if self._keystone_version == 2:
+            creds = {
```
lakshmi-kannan: Heh. _get_v2_creds() and _get_v3_creds() would have been nicer ;)

lakshmi-kannan:
Thanks for the contribution!

st2auth/st2auth/backends/keystone.py
Kami: Agreed :)

Kami:
@Itxaka I made the "payload method" change and merge it into master.

Thanks!

st2auth/st2auth/backends/keystone.py
Itxaka: Just v3. Its the openstack way (tm) of never agreeing to use the same thing from release to release :D

Itxaka:
@Kami @lakshmi-kannan One last question. How are dependencies managed for backends? Are you gonna use the  extras_require keyword in the setup.py? How does that translates to a debian/rpm pakcage?
The greater number of backends, the greater dependencies that will end up in there rigth? 
What if the next backend is a pg/mysql/sqlite/microsoft access(just joking) backend? Then for installing stauth you are bringing a lot of packages for things that you may never use.

Just a thougth :)

Kami:
@Itxaka as noted in my previous comment (https://github.com/StackStorm/st2/pull/1732#discussion_r34984166), right now we put backend dependencies in the global requirements.txt file. It's not ideal, but it works.

Eventually we will want to move to "repository & Python package" per auth backend model, but to get there we also need to improve out current packaging situation. There were some attempts to improve it (e.g. https://github.com/StackStorm/st2/pull/1677), but there were a lot of issues with our current CI system and I simply didn't have the time to finish it yet (it was very slow and time consuming debugging process).

I believe @dennybaa is also currently working on some packaging improvements. In short, eventually we will get there :)

Itxaka:
@Kami Thanks for the info :+1:

dennybaa:
@Itxaka you are talking about saving space by not bringing extra dependencies. Yeah it's generally okay, as soon as st2 is distributed bi pypi it can be easily done. We can manage that behavior...

But if we are talking about os packages, it will bring unnecessary complexity in native packaging. The current chosen approach is to provide os package with a built-in virtualenv which contains all need dependencies (including extra). In this case packages like **st2api** or **st2auth** and all others don't depend on each other, so you can easily install/upgrade/remove any of st2* packages independently.

@ltxaka, @Kami currently st2 sources are not fully prepared to be merged with new packaging support, the code of st2 is updated with modifications from here https://github.com/dennybaa/st2-packages/tree/master/sources. The code in https://github.com/dennybaa/st2-packages actually builds st2 packages and then performs integration testing. (PS. it's almost finished for debian).

PR #1738, opened 2015-07-21
Minor ContentPackResourceController refactoring and improvements

Kami:
This pull request refactors `ContentPackResourceController._get_by_ref_or_id` method to throw a correct `StackStormDBObjectNotFoundError` exception.

This way we can get rid of now unnecessary try / except blocks when calling `_get_by_ref_or_id` and we can simply translate this exception into 404 inside the JSON error response hooks.

@enykeev made the JSONErrorResponseHook refactor ago. This hook is now the place where all the exceptions should be translated to a correct status code and error message. This is good since we previously didn't have a centralized way to translate and handle API exceptions and means we can get rid of a lot of duplicated code.

In the future, when working on new controllers we should make sure we throw correct exceptions there and translate it into an error inside the hook. If you see yourself calling `abort()` function inside the controller this means you are probably doing something wrong and should throw an exception instead.

This is more refactoring to be done with regards to translating the errors inside the hook, but I only need this change right now for the RBAC stuff.

### DIFF
--- a/st2api/st2api/controllers/v1/actionalias.py
+++ b/st2api/st2api/controllers/v1/actionalias.py
@@ -20,7 +20,6 @@ from mongoengine import ValidationError
 from st2api.controllers import resource
 from st2common import log as logging
 from st2common.exceptions.apivalidation import ValueValidationException
-from st2common.exceptions.db import StackStormDBObjectConflictError
 from st2common.models.api.action import ActionAliasAPI
 from st2common.persistence.actionalias import ActionAlias
 from st2common.models.api.base import jsexpose
@@ -61,11 +60,6 @@ class ActionAliasController(resource.ContentPackResourceController):
         except (ValidationError, ValueError, ValueValidationException) as e:
             LOG.exception('Validation failed for action alias data=%s.', action_alias)
             pecan.abort(http_client.BAD_REQUEST, str(e))
-            return
-        except StackStormDBObjectConflictError as e:
-            LOG.warn('ActionAlias creation of %s failed with uniqueness conflict.', action_alias,
-                     exc_info=True)
-            pecan.abort(http_client.CONFLICT, str(e), body={'conflict-id': e.conflict_id})
             return
 
         extra = {'action_alias_db': action_alias_db}
--- a/st2common/st2common/hooks.py
+++ b/st2common/st2common/hooks.py
@@ -25,6 +25,7 @@ from webob import exc
 from st2common import log as logging
 from st2common.persistence.auth import User
 from st2common.exceptions import auth as exceptions
+from st2common.exceptions import db as db_exceptions
 from st2common.util.jsonify import json_encode
 from st2common.util.auth import validate_token
 from st2common.constants.auth import HEADER_ATTRIBUTE_NAME
@@ -176,6 +177,9 @@ class JSONErrorResponseHook(PecanHook):
 
         if isinstance(e, exc.HTTPException):
             status_code = state.response.status
+            message = str(e)
+        elif isinstance(e, db_exceptions.StackStormDBObjectNotFoundError):
+            status_code = httplib.NOT_FOUND
             message = str(e)
         elif isinstance(e, ValueError):
             status_code = httplib.BAD_REQUEST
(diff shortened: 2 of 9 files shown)


st2api/st2api/controllers/v1/actionalias.py
```diff
@@ -62,11 +61,6 @@ def post(self, action_alias):
             LOG.exception('Validation failed for action alias data=%s.', action_alias)
             pecan.abort(http_client.BAD_REQUEST, str(e))
             return
-        except StackStormDBObjectConflictError as e:
-            LOG.warn('ActionAlias creation of %s failed with uniqueness conflict.', action_alias,
```
Kami: Yes, this means we don't have "custom per controller method" log message anymore, but we still log the error inside the JSONErrorResponseHook and those messages are now unified.

On top of that, we now include additional context there (https://github.com/StackStorm/st2/pull/1738/files#diff-545d9d9f5fecdd988de6c2d684fe7a89R170) which should make it easier to find relevant errors, especially if you are used structured logging and something like Graylog2 / loggly (which everyone should be using in production deployments anyway).

st2common/st2common/hooks.py
```diff
@@ -166,8 +167,16 @@ class JSONErrorResponseHook(PecanHook):
...
-        LOG.debug(traceback.format_exc())
+        extra = {
+            'exception_class': e.__class__.__name__,
+            'exception_message': str(e),
+            'exception_data': e.__dict__
+        }
+
+        LOG.debug('API call failed: %s', error_msg, extra=extra)
+
+        if cfg.CONF.debug:
```
manasdk: We actually use this flag?

st2common/st2common/hooks.py
```diff
@@ -166,8 +167,16 @@ class JSONErrorResponseHook(PecanHook):
...
+            LOG.debug(traceback.format_exc())
```
manasdk: Are we sure this is called in a stack frame handling an exception?

manasdk: Answer : Yes.

https://github.com/stackforge/pecan/blob/master/pecan/core.py#L703

Not sure if it is safe to assume.

st2common/st2common/hooks.py
```diff
@@ -166,8 +167,16 @@ class JSONErrorResponseHook(PecanHook):
...
         error_msg = getattr(e, 'comment', str(e))
-        LOG.debug('API call failed: %s', error_msg)
-        LOG.debug(traceback.format_exc())
+        extra = {
+            'exception_class': e.__class__.__name__,
+            'exception_message': str(e),
+            'exception_data': e.__dict__
+        }
+
+        LOG.debug('API call failed: %s', error_msg, extra=extra)
```
manasdk: Without Greylog what will we actually see in the logs?

Kami: Something like that when using console formatter:

> 2015-07-21 08:11:47,829 DEBUG [-] API call failed: Resource with a reference or id "dasda" not found (exception_message='Resource with a reference or id "dasda" not found',exception_data={},exception_class='StackStormDBObjectNotFoundError')

st2common/st2common/hooks.py
Kami: Logging a full stack trace on every exception is a bit excessive since not all the exceptions are fatal and some are expected.

We should probably just always log it on unrecognized exception which results in internal server error.

st2common/st2common/hooks.py
manasdk: Knowing the type would definitely be useful while debugging. Anyway, if needed we can add it back.

st2common/st2common/hooks.py
manasdk: Actually I was just surprised to see this flag being used. Not seen it in the code base anywhere else so this is a little bit surprising. 

Btw, logging exceptions at debug is perfectly fine - even if they are expected exceptions. Typical deployment scenarios would be logs at info level at least that is what we ship by default so should not expect anything else.

Kami: I pushed a change so we now log exceptions which result in internal server error using ERROR log level (log.exceptions) - 839d8f16a184cfb5ce571e4f20862aa59a1971d8

Those are unexpected and usually means something is wrong / broken so it should indeed be logged under ERROR.

PR #1739, opened 2015-07-21
Ability to run multiple of the same sensor type with different config.

manasdk:
* SensorInstance is a parameterized and deployment specific instantiation
  of a SensorType
* SensorExecution indicates the execution of a SensorInstane on a specific
  Sensor node. There a 1:N relation between SensorInstance:SensorExecution
  depending on the sensor partition plan.

Steps:
- [x] DB and API models for SensorInstance and SensorExecution
- [x] API to retrieve SensorInstance and SensorExecution objects.
- [x] SensorInstancesRegistrar to load from sensorinstances from pack
- [ ] Auto creation of SensorInstance in case of parameter-less SensorType
- [ ] Use SensorInstance rather than SensorType in SensorContainer
- [ ] Ability to start SensorInstance on specific node from CLI


Kami:
I some how totally missed this PR :)

armab:
Does this PR means that future Sensors will be more atomic/separated?

Let's say: It's better (?) to have 3 independent sensor processes with unique configs (including auth credentials), instead of specifying all details in one config of single Sensor process.

For example GitHub sensor: https://github.com/StackStorm/st2contrib/blob/ff33de98b2928ec1927feb9d3d6bb794f5a7cf3d/packs/github/config.yaml#L7 which listens to events for many repositories.

Just writing Trello sensor which listens for new events in different boards/lists (same approach as GitHub sensor).
Also asking, because this PR is not updated for a long time, to know the plans and have better understanding, what's future idea about the Sensors?

manasdk:
> Let's say: It's better (?) to have 3 independent sensor processes with unique configs (including auth credentials), instead of specifying all details in one config of single Sensor process.

I would think so. However, this is a choice we would present to the Sensor author after the PR is merged.

> Also asking, because this PR is not updated for a long time, to know the plans and have better understanding, what's the future idea about the Sensors?

Yes this PR got pushed back due to shifting priorities. I would like to restart this work after 1.0 and should take me a week or so to complete. This is definitely the future direction of sensors.

armab:
Thanks @manasdk for info!
I'm :+1: for this approach after implementing Sensor with many independent "targets" to process.

bri365:
We're implementing sensors with multiple independent target configurations as well; +1 for SensorInstance.

haad:
Would be nice to get this one merged if it's finished if not I would like to get it in to mergable state does anyone knows what's missing ?

manasdk:
> Would be nice to get this one merged if it's finished if not I would like to get it in to mergable state does anyone knows what's missing ?

I would have to dig in to this and refresh my memory to get this work going again. Perhaps I can in a comment post about the design and also talk about missing pieces so that some progress can be made. I suspect I will be able to get time mid next week to think about this with a clear mind.

lakshmi-kannan:

PR #1742, opened 2015-07-23
Rename execution variable to liveacton in action execution controller

lakshmi-kannan:
It was utterly confusing when I was reading the code to call liveaction as execution. So here's something.


manasdk:
+1

Kami:
Agreed.

I've fixed some in the past, but I believe there are some more places like that left.

PR #1744, opened 2015-07-23
Adding required description to ActionAlias example

jfryman:
Found while working today at a client site. Copy-paste of this block fails, because `description` is a required field.


manasdk:
+1

PR #1745, opened 2015-07-23
RFR: `parent` should be a dict containing more info about parent execution. 

lakshmi-kannan:
## What? 

This PR turned out to be two parts. I am going to split it now. The major focus of this PR is as follows: 

* We need to be able to add more contextual information
about the parent to tasks in the workflow. So instead
of parent being a simple string field, I am changing
this to a dict.

* Also, liveaction_id is an internal only representation
which should not be exposed to outside. I am also
changing that behavior. `execution_id` will be used now.

* With execution_id being in context, we need to get
liveaction corresponding to the execution. This will
subsequently vanish once all contextual info is in the dict. 
We get `user` and other info from parent liveaction context object.

## TODO
* - [ ] - Changelog


manasdk:
I am cool with direction but I am worried about the regression. Hopefully we actually have sufficient tests.

m4dcoder:
What additional parent context do you plan to add? Are these context info not queryable?

lakshmi-kannan:
@m4dcoder: https://github.com/StackStorm/st2/pull/1745/files#diff-21bb35b73032d803366de9b6ff330a9fR54 So with the execution_id change, we have to get the execution object first, then the liveaction object and then the context. This leads to two database queries. Making a database query for just getting context seems less optimal than building a context dict that's passed around. This will also help in debugging requests between st2 and other engines such as cloudslang/mistral I think.

st2actions/st2actions/runners/actionchainrunner.py
```diff
@@ -171,6 +172,12 @@ def run(self, action_parameters):
                 'traceback': trace
             }
 
+        parent_context = {
+            'execution_id': self.execution_id
+        }
+        if getattr(self.liveaction, 'context', None):
+            parent_context.update(self.liveaction.context)
```
m4dcoder: How about change to the following? Less lines.
```
parent_context = getattr(self.liveaction, 'context', {})
parent_context['execution_id'] = self.execution_id
```

st2actions/st2actions/runners/mistral/v2.py
```diff
@@ -142,9 +142,15 @@ def try_run(self, action_parameters):
         endpoint = 'http://%s:%s/v1/actionexecutions' % (cfg.CONF.api.host, cfg.CONF.api.port)
 
         # Build context with additional information
+        parent_context = {
+            'execution_id': self.execution_id
+        }
+        if getattr(self.liveaction, 'context', None):
+            parent_context.update(self.liveaction.context)
```
m4dcoder: Same comment as above.

st2actions/st2actions/runners/actionchainrunner.py
lakshmi-kannan: heh. sure.

st2common/st2common/services/action.py
```diff
@@ -138,3 +139,7 @@ def update_status(liveaction, new_status, publish=True):
 def is_action_canceled(liveaction_id):
     liveaction_db = action_utils.get_liveaction_by_id(liveaction_id)
     return liveaction_db.status == action_constants.LIVEACTION_STATUS_CANCELED
+
+
+def _get_parent_liveaction(parent_execution_id):
+    return liveaction_services.get_liveaction_for_execution(parent_execution_id)
```
m4dcoder: Redundant IMO unless there's plan to do more here...

lakshmi-kannan: This will all go away once I get all the contextual info in the dict. For now, I stubbed it out for unit testing so I can mock the method.

st2actions/st2actions/runners/actionchainrunner.py
Kami: This would actually modify `self.liveaction.context` attribute in place if `self.liveaction.context` is defined.

I don't think that's desired (we would need to make a copy) :)

Kami:
> Also, liveaction_id is an internal only representation
> which should not be exposed to outside. I am also
> changing that behavior. execution_id will be used now.

Good. Live action is a mess right now :P

st2actions/st2actions/runners/actionchainrunner.py
```diff
@@ -343,7 +352,7 @@ def _run_action(self, action_node, parent_execution_id, params, wait_for_complet
             LOG.debug('%s: Task notify set to: %s', action_node.name, liveaction.notify)
 
         liveaction.context = {
-            'parent': str(parent_execution_id),
+            'parent': parent_context,
```
Kami: What's our migration path for this change?

If I ignore the read path (listing, etc.), I think it only affects the actions which get started (requested / scheduled) with an old version and executed by a new version. Is this correct?

st2actions/st2actions/runners/actionchainrunner.py
lakshmi-kannan: Yeah, this is true which is why I had what I had :P.

st2actions/st2actions/runners/actionchainrunner.py
lakshmi-kannan: Correct. That seemed fine to me. Wdyt?

Kami: I think so. But it would be good to at least test it (manually). Who knows what other thing might break as well :P

Kami:
LGTM

PR #1746, opened 2015-07-23
Enhanced chain vars

manasdk:
* Ability to reference parameters of the action chain itself in the vars of an actionchain. This provides us an ability to parse parameters and apply some transformation which can be later used in the actionchain.


st2actions/st2actions/runners/actionchainrunner.py
```diff
@@ -53,8 +53,12 @@ def __init__(self, chainspec, chainname):
         LOG.debug('Using %s as default for %s.', self.actionchain.default, self.chainname)
         if not self.actionchain.default:
             raise Exception('Failed to find default node in %s.' % (self.chainname))
-        # finalize the vars and save them around to be used at execution time.
-        self.vars = self._get_rendered_vars(self.actionchain.vars) if self.actionchain.vars else {}
+        self.vars = {}
+
+    def init_vars(self, action_parameters):
+        if self.actionchain.vars:
+            self.vars = self._get_rendered_vars(self.actionchain.vars,
```
lakshmi-kannan: This change will work but I am wondering why we aren't letting this code https://github.com/StackStorm/st2/blob/enhanced_chain_vars/st2actions/st2actions/runners/actionchainrunner.py#L330 render vars before evaluating action_params?

lakshmi-kannan: Never mind. Then we have to render this for every task in the action.

lakshmi-kannan:
LGTM. Using action_params in vars is a little awkward but sure.

st2actions/st2actions/runners/actionchainrunner.py
manasdk: :)

manasdk:
Ignoring scrutinizer. Cannot work out why that fails.

st2actions/st2actions/runners/actionchainrunner.py
```diff
@@ -53,8 +53,12 @@ def __init__(self, chainspec, chainname):
         LOG.debug('Using %s as default for %s.', self.actionchain.default, self.chainname)
         if not self.actionchain.default:
             raise Exception('Failed to find default node in %s.' % (self.chainname))
-        # finalize the vars and save them around to be used at execution time.
-        self.vars = self._get_rendered_vars(self.actionchain.vars) if self.actionchain.vars else {}
+        self.vars = {}
+
+    def init_vars(self, action_parameters):
+        if self.actionchain.vars:
```
Kami: What happens if `self.actionchain.vars` is not defined? It looks like we just silently ignore that condition.

Is that desirable / a good idea?

Kami:
We need some docs on the whole context and variables availability - which variables are available in the context where and when.

PR #1747, opened 2015-07-24
Fix escaped liveaction result

m4dcoder:
The DB model permanently changed the dict result when escaping the characters for persistence and then cascade the changes to other components. The fix makes a copy of the dict before escaping.

### DIFF
--- a/st2common/st2common/models/db/stormbase.py
+++ b/st2common/st2common/models/db/stormbase.py
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import copy
 import datetime
 
 import bson
@@ -111,11 +110,11 @@ class StormBaseDB(StormFoundationDB):
 class EscapedDictField(me.DictField):
 
     def to_mongo(self, value):
-        value = mongoescape.escape_chars(copy.deepcopy(value))
+        value = mongoescape.escape_chars(value)
         return super(EscapedDictField, self).to_mongo(value)
 
     def to_python(self, value):
-        value = super(EscapedDictField, self).to_python(copy.deepcopy(value))
+        value = super(EscapedDictField, self).to_python(value)
         return mongoescape.unescape_chars(value)
 
     def validate(self, value):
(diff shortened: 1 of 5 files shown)


st2common/st2common/models/db/stormbase.py
```diff
@@ -110,11 +111,11 @@ class StormBaseDB(StormFoundationDB):
 class EscapedDictField(me.DictField):
 
     def to_mongo(self, value):
-        value = mongoescape.escape_chars(value)
+        value = mongoescape.escape_chars(copy.deepcopy(value))
         return super(EscapedDictField, self).to_mongo(value)
 
     def to_python(self, value):
-        value = super(EscapedDictField, self).to_python(value)
+        value = super(EscapedDictField, self).to_python(copy.deepcopy(value))
```
manasdk: This may not be required. Thoughts?

manasdk:
I am surprised we have not seen an issue with this before.

st2common/st2common/models/db/stormbase.py
m4dcoder: I did it to be symmetric with to_mongo. Maybe leave it alone until a bug is identified?

m4dcoder:
I think most of the tasks in Mistral WF we've been working are mostly python. And if shell, rarely do we use the output elsewhere.  But while working on the Mistral CI, I had to use the output and that's when I saw the problem.

st2common/st2common/models/db/stormbase.py
lakshmi-kannan: I'd add an actual unit test case to make sure copy was done if we have unit tests for EscapedDictField. 

My preference is to make the deepcopy both places.

lakshmi-kannan:
Good catch!

st2common/tests/unit/test_action_db_utils.py
```diff
@@ -132,6 +132,22 @@ def test_update_liveaction_status(self):
         self.assertDictEqual(newliveaction_db.context, context)
         self.assertEqual(newliveaction_db.end_timestamp, now)
 
+        # Update result in dict.
```
manasdk: Could we have a separate testcase. It is useful when reading why this test matters etc.

Kami:
Nice catch, indeed.

Kami:
I don't remember if I updated all the code to use the `EscapedDictField`, so there might some other places left which still use `mongoescape.{es,un}escape_chars`

st2common/st2common/models/db/stormbase.py
Kami: It seems like the best place to handle that would actually be `escape_chars` and `unescape_chars` function which is called by this field.

manasdk:
+1

PR #1748, opened 2015-07-24
Use python2.7 explicitly

DoriftoShoes:
RHEL/CentOS6 uses Python 2.6 as the system Python. We need to explicitly point to the python2.7 binary in order for this to work out of the box on those hosts.

### DIFF
--- a/tools/st2ctl
+++ b/tools/st2ctl
@@ -6,7 +6,7 @@ LOGFILE="/tmp/st2_startup.log"
 WEBUI_LOG_FILE="/var/log/st2/st2web.log"
 COMPONENTS="st2actionrunner st2api st2auth st2sensorcontainer st2rulesengine mistral st2web st2resultstracker st2notifier"
 STANCONF="/etc/st2/st2.conf"
-PYTHON="which python2.7"
+PYTHON=`which python2.7`
 
 # AR is assumed to be an environment variable.
 if [ -z "$AR" ];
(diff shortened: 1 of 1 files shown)


Kami:
Does this work on all the supported distros? Also in places where Python 2.7 is already a default?

DoriftoShoes:
Yup.  From what I have seen each version of python ships with a binary like this.

```
[phool@fedora-staging201 ~]$ which python2.7
/usr/bin/python2.7
[phool@fedora-staging201 ~]$ cat /etc/redhat-release 
Fedora release 20 (Heisenbug)

##########

phool@ubuntu-staging201:~$ which python2.7
/usr/bin/python2.7
phool@ubuntu-staging201:~$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 14.04.1 LTS
Release:	14.04
Codename:	trusty
```

tools/st2ctl
```diff
@@ -6,7 +6,7 @@ LOGFILE="/tmp/st2_startup.log"
 WEBUI_LOG_FILE="/var/log/st2/st2web.log"
 COMPONENTS="st2actionrunner st2api st2auth st2sensorcontainer st2rulesengine mistral st2web st2resultstracker st2notifier"
 STANCONF="/etc/st2/st2.conf"
-PYTHON=`which python`
```
manasdk: You could always check exit code != 0 and either bail out or switch to `which python`.

manasdk:
+1

tools/st2ctl
Kami: Yeah, I posted on the IRC the other day if it might make sense to ship our own wrapper script which detects the right path to Python 2.7 interpreter and use this as the shebang entry point.

This would be more robust, but I'm not sure if users / system admins would like this approach since it adds some indirection and it's not "unixy".

Kami: On a related note, we should probably use `#!/usr/bin/env python{2.7}` so it's easier for user to use a different Python interpreter without needing to modify the entry point script or use a symlink.

PR #1749, opened 2015-07-24
Adds proper Debian test in st2_deploy.sh

jfryman:
This commit adds a test that matches the test in `st2_bootstrap.sh`, allowing the installer to run on Debian systems.


jfryman:
Currently fails on Debian Jesse.

manasdk:
silly bash!

+1

PR #1754, opened 2015-07-27
Action alias requires a description

amaline:
Prior to 0.12 it seems like a description was not required, but it is now.


manasdk:
Thanks!

PR #1755, opened 2015-07-27
RFR - Action context params chain

lakshmi-kannan:
I wrote the documentation and I am unhappy with the solution because of following reasons:

* The list of variables you can access via action_context depends on whether it is a simple action or a workflow. Also, it depends on whether the action was kicked off via chatops or not.

* I think execution_id and correlation id (when available) should make it to action_context. 

* We need a better way to programatically list variables that are going to be in action_context. (See @Kami's comment: https://github.com/lakshmi-kannan/st2/pull/1/files#diff-157d9a0f26f16d59cb8468550a11f495R185)

### DIFF
--- a/docs/source/actions.rst
+++ b/docs/source/actions.rst
@@ -182,6 +182,51 @@ In the example above, ``to_number`` parameter contains attribute ``secret``
 which value is ``true``. If an attribute is marked as a secret, value of that
 attribute will be masked in the |st2| service logs.
 
+Parameters in actions
+~~~~~~~~~~~~~~~~~~~~~
+
+In the previous example, you probably noticed how you can access parameters from
+from key value store by using the ``system`` prefix in the template. You can also
+get access variables from the context of the execution. For example:
+
+.. code-block:: yaml
+
+    parameters:
+      user:
+        type: "string"
+        description: "User of this action."
+        required: true
+        default: "{{action_context.api_user}}"
+
+
+The prefix ``action_context`` is used to refer to variables in action context. Depending on how
+the execution is executed and nature of action (simple vs workflow), variables in action_context changes.
+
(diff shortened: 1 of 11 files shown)


docs/source/actions.rst
```diff
@@ -182,6 +182,51 @@ In the example above, ``to_number`` parameter contains attribute ``secret``
 which value is ``true``. If an attribute is marked as a secret, value of that
 attribute will be masked in the |st2| service logs.
 
+Parameters in actions
```
lakshmi-kannan: @Kami's comment on my fork:
```
It would be great if we could generate a list of the available base variables for each "execution type" (API call and ChatOps) from code (e.g. something like this - https://github.com/StackStorm/st2/commit/6fb274ccd272c5f4cc31f1ecff0841b89455ca72).

I know this requires some refactoring work since right now context is created inside the method and is not a constant like in the PermissionType case.
```

lakshmi-kannan: Yeah, this is not possible now since we don't have DB schema for what goes into context. It's simply a dict which means you can have arbitrary KV pairs. This is both good and bad. I do think though that this is beyond the scope of this PR.

lakshmi-kannan:
I tried adding execution_id to action_context. This leads to problems with action chain because action_chain renders parameters before calling an action in the action chain. At this point, there is no valid execution object for the item in the chain and therefore chain blows up. One potential solution is to delay rendering of params of item in the chain to when the item actually gets executed. This would mean that we pass all the context (results of items executed so far, chain vars...) as context to the liveaction object. We discussed this offline and we are conflicted on the right design. So I am reverting the commits to add execution_id to context.

manasdk:
LGTM

Kami:
LGTM.

And yeah, sounds like we should get on the same page first wrt the execution_id changes.

PR #1756, opened 2015-07-28
RFR: Disable parallel execution of SSH actions

lakshmi-kannan:
* Fabric doesn't behave well with eventlets especially when parallel is
set to True. Parallel basically executes a command on a bunch of hosts.
Parallel uses multiprocessing and calls fork(). We don't understand the
root cause yet. Fabric has other issues as well. So until we use
Paramiko straight and replace our SSH runner, let's disable parallel
SSH execution.

### DIFF
--- a/st2actions/st2actions/bootstrap/runnersregistrar.py
+++ b/st2actions/st2actions/bootstrap/runnersregistrar.py
@@ -156,7 +156,7 @@ RUNNER_TYPES = [
             'parallel': {
                 'description': 'Default to parallel execution.',
                 'type': 'boolean',
-                'default': True,
+                'default': False,
                 'immutable': True
             },
             'sudo': {
@@ -220,7 +220,7 @@ RUNNER_TYPES = [
             'parallel': {
                 'description': 'Default to parallel execution.',
                 'type': 'boolean',
-                'default': True,
+                'default': False,
                 'immutable': True
             },
             'cwd': {


manasdk:
Go for it!

Kami:
Keep in mind that we enabled parallel execution as a work around for another bug.

Kami:
The real solution is to switch away from fabric. This doesn't solve anything, it will simply move from one bug to another.

DoriftoShoes:
We are definitely working on a different ssh runner but I have to ask...what is the other bug?  We've been running with parallel off in the build pipeline since 0.8 and I don't think we have hit it.

manasdk:
Lol ... https://github.com/StackStorm/st2/commit/4e8bdc97a3ae3073057791c873b074dd62128637

manasdk:
> The real solution is to switch away from fabric. 
+1

@lakshmi-kannan is on this already. This was an interim step to hold us for a week or so.

Kami:
@DoriftoShoes It was some kind of IO error IIRC (we used to hit it very often).

If we have indeed been running with parallel disabled on build server, then I'm fine with this is a temporary work around. Although, the question is - are we going to do a minor release with this change or not? 

If we are not going to do it, then it makes little sense to commit this change - we should just to the full runner in a major release :)

DoriftoShoes:
We are going to do a patch release on 0.12, yes.  On Friday I believe.  So this should be cherry picked and applied there also.

lakshmi-kannan:
Ok, I am going to merge this and cherry pick this for 0.12.1 patch release.

Kami:
@DoriftoShoes @lakshmi-kannan Sounds good.

PR #1759, opened 2015-07-28
Unregister rules and aliases if the pack is uninstalled

manasdk:
* Leaving behind rules and action-aliases of a pack when the pack is incorrect.
  Removing all the content is the correct behavior.


Kami:
:+1:

PR #1760, opened 2015-07-29
Rmq cluster support

manasdk:
* Config varible messaging.cluster_urls can be used to specify a list of host
  for rabbitmq cluster configuration.
* messaging.urls is effectively deprecated but presevered to support existing
  configuration.
* utility added to hide away precendence order of cluster_urls and urls.

Testing
* Non clustered RMQ
   - [X] With `messaging.url` 
   - [X] With `messaging.cluster_urls`
* clustered RMQ
   - [X] With `messaging.url` pointing to a single node in the cluster
   - [x] With `messaging.cluster_urls` pointing to a single node in the cluster
   - [x] With `messaging.cluster_urls` pointing to all nodes in the cluster
* Failover
   - [x] pull the plug on connected node which stackstorm is running
   - [x] first in the list of nodes is offline i.e. starts by trying to connect to offline node

### DIFF
--- a/st2common/st2common/transport/bootstrap_utils.py
+++ b/st2common/st2common/transport/bootstrap_utils.py
@@ -1,0 +1,49 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from kombu import Connection
+from st2common import log as logging
+from st2common.transport import utils as transport_utils
+from st2common.transport.execution import EXECUTION_XCHG
+from st2common.transport.liveaction import LIVEACTION_XCHG
+from st2common.transport.reactor import TRIGGER_CUD_XCHG, TRIGGER_INSTANCE_XCHG
+from st2common.transport.reactor import SENSOR_CUD_XCHG
+
+LOG = logging.getLogger('st2common.transport.bootstrap')
+
+__all__ = [
+    'register_exchanges'
+]
+
+EXCHANGES = [EXECUTION_XCHG, LIVEACTION_XCHG, TRIGGER_CUD_XCHG, TRIGGER_INSTANCE_XCHG,
+             SENSOR_CUD_XCHG]
+
+
+def _do_register_exchange(exchange, channel):
+    try:
+        channel.exchange_declare(exchange=exchange.name, type=exchange.type,
+                                 durable=exchange.durable, auto_delete=exchange.auto_delete,
+                                 arguments=exchange.arguments, nowait=False, passive=None)
+        LOG.debug('registered exchange %s.', exchange.name)
+    except Exception:
+        LOG.exception('Failed to register exchange : %s.', exchange.name)
+
+
+def register_exchanges():
+    LOG.debug('Registering exchanges...')
+    with Connection(transport_utils.get_messaging_urls()) as conn:
+        channel = conn.default_channel
+        for exchange in EXCHANGES:
+            _do_register_exchange(exchange, channel)
--- a/st2common/st2common/transport/bootstrap_utils.py
+++ b/st2common/st2common/transport/bootstrap_utils.py
@@ -16,6 +16,7 @@
 from kombu import Connection
 from st2common import log as logging
 from st2common.transport import utils as transport_utils
+from st2common.transport.connection_retry_wrapper import ConnectionRetryWrapper
 from st2common.transport.execution import EXECUTION_XCHG
 from st2common.transport.liveaction import LIVEACTION_XCHG
 from st2common.transport.reactor import TRIGGER_CUD_XCHG, TRIGGER_INSTANCE_XCHG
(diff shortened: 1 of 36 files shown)


Kami:
Re failover: Looks like kombu already supports that using `failover-strategy` Connection kwarg. Or did you mean something else?

manasdk:
> Re failover: Looks like kombu already supports that using failover-strategy Connection kwarg. Or did you mean something else?

Well. I think the issue is with our publishers which sort of manage their own connection and do not seem to be setting up failover correctly. 

Also, hate the kombu docs - can never easily find what I am looking for.

manasdk:
Per http://kombu.readthedocs.org/en/latest/reference/kombu.html#kombu.Connection `round-robin` is default but in the publishers case I do not see this actually failing over. The consumers(`ConsumerMixin`) seem to failover ok.

st2common/st2common/transport/connection_retry_wrapper.py
```diff
@@ -0,0 +1,164 @@
...
+            return should_stop, -1
+        wait = 0
+        should_stop = False
+        if self._nodes_attempted % self.cluster_size == 0:
+            wait = self.wait_between_cluster
+        self._nodes_attempted += 1
+        return should_stop, wait
+
+
+class ConnectionRetryWrapper(object):
```
Kami: I've asked a similar thing before - can't we just use the existing failover and retry support from kombu (http://kombu.readthedocs.org/en/latest/reference/kombu.html#connection, http://kombu.readthedocs.org/en/latest/reference/kombu.html#kombu.Connection.failover_strategy, http://kombu.readthedocs.org/en/latest/reference/kombu.html#kombu.Connection.autoretry, http://kombu.readthedocs.org/en/latest/reference/kombu.html#kombu.Connection.ensure_connection)?

It looks like it already implements all the things we need (including retrying of a particular function), or am I missing something? :)

And the whole failover and retrying happening inside kombu seems better anyway since there it's closer to the place where the actual connection is established, etc. (not to mention the fact their retry code is more battle tested than ours).

manasdk: Yes I wanted to believe so too but guess what those are all lies that kombu is spreading!

So there are 4 error see https://github.com/celery/kombu/blob/master/kombu/connection.py#L781-L811.

Of these there are some which are characterized as recoverable errors. 

If you look at the source of http://kombu.readthedocs.org/en/latest/reference/kombu.html#kombu.Connection.autoretry and consequently http://kombu.readthedocs.org/en/latest/reference/kombu.html#kombu.Connection.ensure you will see that it only handles the recoverable errors i.e `recoverable_connection_errors` and `recoverable_channel_errors`. 

We already use that function for dealing with the recoverable errors (thank to @lakshmi-kannan) and we still have the code https://github.com/StackStorm/st2/pull/1760/files#diff-cd01bec138927aee7af667f279ff7358R163.

For the "unrecoverable" errors which are thrown when a node in a cluster goes down we are on the hook to deal with the error. `ConsumerMixin` actually deals with this part https://github.com/celery/kombu/blob/master/kombu/mixins.py#L172 but I could not find any Producer code which handled this in any reasonable manner.

If you see the impl https://github.com/StackStorm/st2/pull/1760/files#diff-cd01bec138927aee7af667f279ff7358R113 you will see that net is resembles the `ConsumerMixin` code which is spread across multiple methods.

## Some wishes -
* I wish I could use some kombu primitives to handle node failures properly.
* I wish kombu APIs were properly documented and set exact expectations so I did not have to read their horrible and convoluted code. (its not that bad I was probably just annoyed)
* I wish someone updates kombu to write their producers nicely and deal with ConnectionPools etc like any self respecting remote service connection library should.
* I wish I can burn all this code I have written once some library handles this mess.

Kami: Alright, thanks for the clarification.

I will dig in and look better into the code changes later.

st2common/st2common/transport/bootstrap_utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+__all__ = [
+    'register_exchanges'
+]
+
+EXCHANGES = [EXECUTION_XCHG, LIVEACTION_XCHG, TRIGGER_CUD_XCHG, TRIGGER_INSTANCE_XCHG,
+             SENSOR_CUD_XCHG]
+
+
+def _do_register_exchange(exchange, connection, channel, retry_wrapper):
```
lakshmi-kannan: This could have been an ExchangeRegitrar class :P

st2common/st2common/transport/bootstrap_utils.py
```diff
@@ -0,0 +1,68 @@
...
+    with Connection(connection_urls) as conn:
+        # Use ConnectionRetryWrapper to deal with rmq clustering etc.
+        retry_wrapper = ConnectionRetryWrapper(cluster_size=len(connection_urls), logger=LOG)
+
+        def wrapped_register_exchanges(connection, channel):
+            for exchange in EXCHANGES:
+                _do_register_exchange(exchange=exchange, connection=connection, channel=channel,
+                                      retry_wrapper=retry_wrapper)
+
+        retry_wrapper.run(connection=conn, wrapped_callback=wrapped_register_exchanges)
```
lakshmi-kannan: Somehow this API is very confusing because of back and forth. As a programmer, I want to make sure channel.exchange_declare with its arguments gets sent to rabbit with x attempts. 

RabbitPublisher.run(func, func_kwargs, retry = 3) would have been a better high level API IMO. So I am saying there is one abstraction level missing. A RabbitPublisher (not sure about the name and publish doesn't sound right) that uses a ConnectionRetryWrapper would have been easier to read. Thoughts?

st2common/st2common/transport/connection_retry_wrapper.py
lakshmi-kannan: Yeah, these kombu APIs are pretty sketch. Unfortunately, it does look like we got to handle node down issues in a cluster setup.

st2common/st2common/transport/bootstrap_utils.py
manasdk: Well I suspect we will add more junk here later. I don't care either way what this is named. Can make it a class etc as well.

st2common/st2common/transport/bootstrap_utils.py
lakshmi-kannan: Had an offline convo with @manasdk with why we cannot do what I suggested. The problem stems from the fact that the functions we are trying to invoke are part of channel object. A channel is not guaranteed to exist in a clustered rabbitmq when a node goes down. This makes it hard to do what I suggest.

FWIW, kombu APIs suck really badly :/.

lakshmi-kannan:
LGTM

PR #1762, opened 2015-07-29
RFR: Disable parallel execution of SSH actions

lakshmi-kannan:
* Fabric doesn't behave well with eventlets especially when parallel is
set to True. Parallel basically executes a command on a bunch of hosts.
Parallel uses multiprocessing and calls fork(). We don't understand the
root cause yet. Fabric has other issues as well. So until we use
Paramiko straight and replace our SSH runner, let's disable parallel
SSH execution.


lakshmi-kannan:
@Kami @DoriftoShoes Cherry picked for 0.12.1 release.

Kami:
:cherries:

PR #1765, opened 2015-07-29
[WIP] Initial RBAC implementation

Kami:
This pull request includes initial work in progress changes needed for RBAC.

I plan to finish and polish it up and get those changed merged into master soon. Permission loading from files is not there yet, but I plan to implement it soon (I will gist some examples in the upcoming days).

Keep in mind that those changes are behind a feature flag (config option, disabled by default) so they can be merged into master without any user impact. 

P.S. Sorry for all the (merge) commits. I tried to be a good citizen and make general cleanup, refactoring and improvements changes in a branch based off master, get it merged into master first and then merge it back into this branch.

## TODO (merge blockers)

- [x] More tests
- [x] Finish and polish up the code
- [x] More and better docs
- [x] Loading roles and permission grants from YAML files (incremental change, can be done in a separate PR as well)
- [x] Migration for assigning "uid" attribute to the existing resources in the database

## TODO (later)

- [ ] Lint script which checks that all the controllers perform permission checking
- [ ] Performance optimizations
- [ ] Log "access denied" errors in a special audit log
- [ ] Make sure all the operations currently carry over the user who performed the operation in the context
- [ ] Add new "enable" and "disable" permission type for each content pack resource
- [ ] Update all the list (get_all) views so the resulting set if filtered based on the user resource permissions
- [ ] Implement executions permission resolver - executions should inherit view permissions from the action they belong to

### What to do with other controllers where we currently don't perform any permission checking?

Things such as runner types, actions executions, history, stream, etc. We could add a new set of permission for each of those resources, but this might be an over kill right now. We could also add a single "SYSTEM_RESOURCE" permission type which applies to all the other (non content pack) resources.

### How to handle action execution permissions

Should we introduce new permissions for "action executions" endpoint? We could also inherit actions permissions, but I'm not too big of a fan of that.

### How to handle temporary tokens which are passed to sensors and actions

I propose to tie those tokens to a "system" user (this user can't be manipulated) which has an "admin" role.

We could (and maybe should) also tie the token to the user which invoked the action, but this gets more tricky, especially for "event" actions which are not directly triggered by the API.

Maybe we could also perform do a mix:

1. If an action is triggered by the API tie token to the user who triggered the action
2. If it's a sensor or "event" triggered actions tie it to the system user

### Blacklist for globally disabled runner type

Per @lakshmi-kannan that was a feature request. Imo that seems a reasonable feature. I think we should handle it via a special "blacklisted_runner_types" (or similar) configuration option since it doesn't really fit that much into the RBAC model and the config option is imo simpler and better.

### Permission which allows user to disable / enable a resource

@dzimine said someone asked for that. It would be great to get some more context into it since I don't really understand how it's useful.

In any case, this should be easily to implement via new "<resource>_modify_state" (or similar) permission type.


Kami:
Alright, lets discuss RBAC definition files format.

A while ago @lakshmi-kannan and I had a short discussion about the potential RBAC definition file(s) format.

We both agreed that spliting it into multiple files might be better, since it makes it easier to write.

On the other hand, splitting it into multiple files makes it a bit harder to see / visualize the end result.

## RBAC definition files format

Location: ``/opt/stackstorm/rbac/``
Files:

* ``/opt/stackstorm/rbac/<role_name>_role.yaml - role definition
* ``/opt/stackstorm/rbac/<username>_roles.yaml - user role assignments

Definition type (role definition vs role assignment) is inferred from the file suffix.

As an alternative, we could put files into subdirectories. For example:

* ``/opt/stackstorm/rbac/roles/<role_name1>.yaml``
* ``/opt/stackstorm/rbac/roles/<role_name2>.yaml``
* ``/opt/stackstorm/rbac/role_assignments/<username1>.yaml``
* ``/opt/stackstorm/rbac/role_assignments/<username2>.yaml``

### Example role definition format

``my_super_role_role.yaml``

```yaml
---
    name: "my_super_role"
    description: "Super cool role"
    permission_grants:
        -
            resource_ref: "mypack.myaction"
            permission_types:
               - "action_delete"
               - "action_modify"
               - "action_execute"
        -
            resource_ref: "mypack.myrule"
            permission_types:
               - "rule_modify"
               - "rule_delete"
        -
            resource_ref: "mypack.mywfaction"
            permission_types:
               - "action_execute"
        -
            resource_ref: "mypack"
            permission_types:
               - "action_execute"
```

Note 1: ``permission_grants`` could also be a dict where the key is ``resource_ref``.

Note: 2: We don't need ``resource_type`` or ``resource_uid`` attribute since we can (mostly) infer resource type from the permission type. I would personally still prefer to use ``resource_uid`` which includes the type since user could potentially try to grant invalid permission type for the resource and the loading of the permissions would fail (and explicit is usually also better than implicit).

### Example user role assignments format

Note: Username is parsed from the file name.

``user_1_roles.yaml``

```yaml
---
    roles:
        - my_custom_role_1
        - my_custom_role_1
```

## Implementation

For v1, the implementation could load role definitions and assignments when the register content script is called. 

A better solution would probably be to add a separate script for loading rbac related info (e.g. ``st2 load-rbac-info` or similar).

The difference between loading those definitions and pack content from disk is that this script would perform a full sync between disk and DB content - new roles would get added, existing ones updated, and removed one would be removed. That's the same model I believe we should also eventually move to for the content pack resources.

Performing only add + merge is not a good idea in this case.

Definitions would be loaded in the following order:

1. Role definitions in alphabetic order
2. User role assignments

jfryman:
Instead of separate files, can it be the same file with different top-level keys or something?

Splitting up the roles feels weird to me. Conceptually, I get the reasoning - they are different concerns. But, to your point on visualization, I can see cases where something like default permissions for ChatOps would want to be shipped as a single unit.

Are groups covered here? 1:1 Role/User seems tedious, and also makes it difficult to bundle ACLs

Kami:
@jfryman 

> Instead of separate files, can it be the same file with different top-level keys or something?
> 
> Splitting up the roles feels weird to me. Conceptually, I get the reasoning - they are different concerns. But, to your point on visualization, I can see cases where something like default permissions for ChatOps would want to be shipped as a single unit.

It can, that was the initial idea, but @lakshmi-kannan and I had some concerns. In a typical scenario, this file will also be very big.

> Are groups covered here? 1:1 Role/User seems tedious, and also makes it difficult to bundle ACLs

Not sure I understand what you mean with groups. Role == group? Or you mean something else - e.g. grouping multiple roles in a group?

jfryman:
> Example user role assignments format

> Note: Username is parsed from the file name.

> user_1_roles.yaml

How do I add a user to a group, and then apply groups to the role?

Kami:
@jfryman There is no concept of a group right now.

It's flat so there are just roles. I assume that with groups user could group together related permissions and then apply groups to the roles.

If so, the same can be achieved with roles - user can have multiple roles assigned with it and the idea is to group related permission grants in a role (there are some w.i.p. docs available at https://github.com/StackStorm/st2/pull/1765/files#diff-0116a6a629a2ce95463ea20aace3c760R24).

I do see some value in the groups concept (if I understand what you means with groups correctly), but it makes the whole think a bit more complicated so not fully sold on it yet.

jfryman:
Groups are necessary when I'm shipping features, but I don't know what the end-state of my target StackStorm system is going to be. I can ship ACLs all pinned to a group, and then apply local users to those groups. Otherwise, we'd have to ship something that would allow me as an author to quickly and easily drop users into roles.

Seems like not so big a deal? Maybe with <50 users, but it will get nuts quickly. Groups should be part of this construct, imo.

dzimine:
> Things such as runner types, actions executions, history...
Protecting history is quite important. A lot of info I would rather not others see go into input and output params of my actions. Including the very fact that by looking into "everyone's" history, I can see what actions are in the system even if I don't have permissions to view the actions. Let's think it over.

> If it's a sensor or "event" triggered actions tie it to the system user.
We should inherit the user from the rule that resulted in action. Trigger can trip multiple rules. Rules may be created by different users. Those who create a rule are MORE responsible for this action than those who created the triggers/launched sensors. 

> How do I add a user to a group, and then apply groups to the role?
I assume user groups belong to identity provider (e.g., linux group or LDAP group), and we can map to roles both individual users and roles. Yes?

> Should we introduce new permissions for "action executions" endpoint? We could also inherit actions permissions, but I'm not too big of a fan of that.
First, executions don't have good way to reference them up, other but than by properties. 
For POST(create), it's reasonable that it interits action.execute.
For GET, I don't know; the most typical use cases are to see all, to see "all actions run by user=ME", or by tags, once we begin to use them more, and inherit tags by executions from actions.
Good question.

### PermissionTypes
PermissionType - should they really have to be static tuples? like `ACTION_VIEW`? Isn't it permission assigments that mix permission type with entities? In [design](https://docs.google.com/document/d/1wkdfeZ41I5hyc4NZFg7pXOZwTQVa6hLng0oMsdu7q-A/edit#heading=h.c4xawg70ooos), permission types were VIEW/MODIFY/EXECUTE/ALL. Why combining them statically with resource types? And how to use them? I can't use ACTION_VIEW and RULE_VIEW interchably in code, or configuration. 
```
@request_user_has_permission(permission_type=PermissionType.ACTION_VIEW)
```
This needs clarification.

manasdk:
> PermissionType

I actually like the separate permissions. It allows the precise permission to be applied on groups. e.g. 

```
---
    name: "aws_operator"
    description: "AWS operator role"
    permission_grants:
        -
            resource_ref: "aws"   # I assume this will imply the pack
            permission_types:
               - "action_view"
               - "action_execute"
               - "execution_view"
```

This way there is no ambiguity on what permissions the roles provide. In this case I rather have a tedious approach to permission definition as that provides control - also since role definition and permission assignment should be `admin` functionality I dare say that it should be ok to expose such granularity.

Kami:
@dzimine 

> Protecting history is quite important. A lot of info I would rather not others see go into input and output params of my actions.

Alright, I will introduce new permission types for each of those endpoints and wrap them with permission checks.

I believe the existing "mask_secrets" approach already solves that. After this change is in, this will still work as before, just the admin role will be assigned differently (previously admin user was specific in the config, now you will assign admin role to the user).

We could also build on that and also show those values to every user which has "view" permission on a particular action, but this is starting to get quite complicated and expensive (it would include a lot of late filtering and I'm quite worried about the overhead).

> PermissionType - should they really have to be static tuples? like ACTION_VIEW? Isn't it permission assigments that mix permission type with entities? In design, permission types were VIEW/MODIFY/EXECUTE/ALL. Why combining them statically with resource types? And how to use them? I can't use ACTION_VIEW and RULE_VIEW interchably in code, or configuration.

The change came after a discussion with @lakshmi-kannan and @manasdk a while back. They wanted a set of permissions for each resource type.

This does make some sense since not all the permission types apply to all the resource types, but we could also do this differently - store a resource_type to permission types list internally, drop resource type prefix from an external PermissionType and add a new meta endpoint which returns a list of available permission types for a particular resource type.

One of the problems we wanted to address is determining resource type without the user needing to specify an UID which encodes the resource type or additional resource_type attribute.

I personally prefer resource_uid + permission_type without the prefix approach. Something like this:

```yaml
---
    name: "aws_operator"
    description: "AWS operator role"
    permission_grants:
        -
            resource_uid: "action:aws:create_instance"
            permission_types:
               - "view"
              - "execute"
```

Anyways, @manasdk also had a good point above (aka making it explicit).

> @request_user_has_permission(permission_type=PermissionType.ACTION_VIEW)
> This needs clarification.

Which part? How the decorator works, how to use it or both?

There are some docstrings in the decorators.py module right now, but I also plan to add more / improve it.

dzimine:
> Which part? How the decorator works, how to use it or both?
the whole thing. Not this string. Never mind. We already discussing :)

dzimine:
I got the problem - applying permissions to a mixed bags of resources. I still don't like the proposed solution (define a bag, than apply permission-assigments (aka tuples of (resource_type, permission_type)) to the bag. Like in @manaask case, bag is a "pack". Let's think generically, how to express defining a bunch of mixed resources, and applying a set of permissions to it? Including selecting multiple resources, beyond pack/across packs, based on properties. 

To throw some ideas: 

1) Define a bag of resources with a mixture of 
```
---
name: 
permission_grants:
    - 
    		# Some flexible way to define a bag of mixed resources
        resources:
           # Invent defining a group of resources
           # Specify by uid, by ref, by widlchart, by properties, some can co-exist
           ref: core.* #
           parameters:
              enabled: True
              tags:
           # Filter by resource type
           resource_type: [action, rules] # default "any"
        # Apply permissions 
        permissions:
           - execute
           - view
    - 
        resources:
        	uid: 1234590393029
        	...
```

2) specify the bag and give hint to the system on which actions apply to which types of resources in the bag:
```
---
name: 
permission_grants:
    - 
    		# Some flexible way to define a bag of mixed resources
        resources:
           # HERE: invent syntax here which defines a group of resources
           ref: core.* #
           uid: 1234590393029
           parameters:
              enabled: True
        # Apply permissions, specifying to which resource types they apply. 
        permissions:
           - execute: [actions, rules]
           - view: all
```

For one thing, before we commit to any model and correspondent syntax, lets pick 3-4 specific use cases and see how they solve the problem. 

This may surface if we are not on the same page re which cases we shall support, which is good to surface now.

dzimine:
* Can we use users and groups interchangably here, so I can assign a role to LDAP user OR group.
* What are the benefits of splitting it to file/per/user-group, as opposed to a single assigment file?
* `permission_grants`: shall we offer blacklist + whitelist combination? call them `assigments`, and each array element will either "allow" or "deny" described permissions. In cases where I want to allow everything except just one action, will be easier to define.

dzimine:
Use cases (my quick crack, please @jfryman help refine and add more): 
* Assign actions, rules and sensors/triggers by pack.
* Allow individual actions or rules (VERY unlikely!)
* Deny individual actions or rules (QUITE likely)
* Allow only executions initiated by user ME (admin sees all executions, users only see THEIR executions)
* Group all relevant resources across packs by a tag, and assign them all
* Allow/restrict access to specific KV-store sections
* Allow only disable-enable, and deny other CREATE/CHANGE, for rules and actions (Netflix/@phool ask).

dzimine:
Couple of syntax pointers: 
* AWS IAM Policy structure http://docs.aws.amazon.com/IAM/latest/UserGuide/policies_overview.html#policies-introduction
* I want to place OpenStack as an example but it’s almost serves as an anti-example: http://docs.openstack.org/kilo/config-reference/content/section_keystone-policy.json.html

lakshmi-kannan:
* I do think action_view, action_modify kinda permissions are more clear. These can be applied to groups of resources or individual resources without any problems. 

* I definitely do think we need grouping constructs for specifying permission grants to resources. We can start with a list of resources called out by names and allow glob expressions. So +1 to dzimine's suggestion.

```
permission grants:
  -
  resources: [list of resources by explicit names or glob]
  permissions: [list of permissions]
```

> What are the benefits of splitting it to file/per/user-group, as opposed to a single assigment file?

IMO manageability. We'll start hating the one file structure as soon as we have serious number of content/users. I like Tomaz's current approach of convention based.

* I do agree with jfryman on group level for users. I don't think this is needed for initial prototype though. If we make our models flexible, it seems we can add this construct later. 

>permission_grants: shall we offer blacklist + whitelist combination? call them assigments, and each array element will either "allow" or "deny" described permissions

Usually, it's a best practice to deny everything and start allowing things. I don't know if that's the right approach for RBAC though. Maybe we can have a global config for deny all or allow all by default. If there is a RBAC schema, then it trumps everything.

Kami:
@dzimine 

> Couple of syntax pointers:

> * AWS IAM Policy structure http://docs.aws.amazon.com/IAM/latest/UserGuide/policies_overview.html#policies-introduction
> * I want to place OpenStack as an example but it’s almost serves as an anti-example: http://docs.openstack.org/kilo/config-reference/content/section_keystone-policy.json.html
It is generally good to base on something people are generally familiar, like AWS (all other things being equal).

I personally think they are both anti-patterns.

AWS one is powerful, but it's anything but simple and easy to understand (writing that JSON is also not fun).

I'm trying to find a middle ground between "enterprise" (1000 switches and options, hard to understand write and manage, everyone hates it including myself) and "startup" (in some cases "maybe" not flexible enough) solution.

> Use cases (my quick crack, please @jfryman help refine and add more):

* Assign actions, rules and sensors/triggers by pack.
* Allow individual actions or rules (VERY unlikely!)
* Deny individual actions or rules (QUITE likely)
* Allow only executions initiated by user ME (admin sees all executions, users only see THEIR executions)
* Group all relevant resources across packs by a tag, and assign them all
* Allow/restrict access to specific KV-store sections
* Allow only disable-enable, and deny other CREATE/CHANGE, for rules and actions (Netflix/@phool ask).

Most of those things are possible with the existing implementation (other things which are not possible right now can be implemented later as add-on features). Only thing which is not possible is "Deny individual actions or rules (QUITE likely)". To keep things simple and easy to understand (no possibility of an overlap and conflicting grants), we default to "Deny all" and allow users to grant "allow" on particular resource or set of resources.

If more people think we definitely need "deny" I can implement it, but do keep in mind (as noted above) that this will increase complexity and will make it harder to understand. In any case, it would be great if we can get on the same page on this ASAP.

Another thing which is not possible right now is granting permissions on tags. That's by choice to reduce the scope. I consider it as an add-on feature (easy to add later on) and existing design accounts for that.

Kami:
@lakshmi-kannan 

> I definitely do think we need grouping constructs for specifying permission grants to resources. We can start with a list of resources called out by names and allow glob expressions. So +1 to dzimine's suggestion.

We already have one grouping concept right now and it's called a pack. In the future we will have another one - "tags".

Glob expressions make some sense (for things such datastore item keys), but I don't think we need another concept in addition to those. It doesn't add anything, just complexity (everything can already be achieved with the existing constructors / concepts).

Issue #1717, opened 2015-07-14
Execution stuck in "Running" state


felin-arch:
I recently ran into a problem which caused the state of the execution to incorrectly report "Running". Looking at the logs, the actions clearly fails. There is no way to cancel it or remove it via the UI or the CLI. The action never seems to time out.

It happened when the result of the action contained some non-utf8 characters. I posted the action runner log at [here](https://gist.github.com/felin-arch/2c9f5b07e38e6d661e84).

Some background info: running stackstorm 0.11.6 on CentOS 7 with the latest python available on the OS.

I was able to reproduce this multiple times. The action should fail with the error message seen in the logs.

emptywee:
I had the same on RHEL 6.4, st2 0.11.6 with errors in yaml files for certain actions. Though the task timed out eventually.

manasdk:
Seen this once in a while but it was elusive enough that the RC was not obvious. The logs should help.

Btw, you can `cancel` the running execution with `st2 execution cancel <execution_id>`. It is not a proper cancel i.e. will not stop an execution which is actively performing an activity but would work well in this case where the `RUNNIG` state is incorrect.

http://docs.stackstorm.com/cli.html?highlight=cancel#cancel-an-execution

felin-arch:
@manasdk I did try to cancel it the way you described. Unfortunately it said the cancellation failed.

manasdk:
Good to know. Will fix the root-cause anyway and work out why the cancel failed as well.

lakshmi-kannan:
@felin-arch: Looking at the code, we do set the action status as "failed"
https://github.com/StackStorm/st2/blob/master/st2actions/st2actions/worker.py#L106

Are you able to repro this every time? Some more info about the script and the data will definitely help. 

Regarding cancellation failing, we decided that canceling a simple action would mean different things based on the runner. So it's impossible to guarantee anything. So on cancelation, we check if the status is 'scheduled'. Only then we cancel. We are talking about this here but wanted to throw more context at you. In workflows, we stop at the action last running before cancelation and mark the workflow as canceled.

felin-arch:
I will try to set up a minimum environment for reproduction. Then I can test a few different ideas. I'll be back once that is done.

felin-arch:
I was not successful in isolating the problem in a minimal environment. Currently I am using st2 to run scheduled jobs and after a day of running it I always end up with more than 10 jobs stuck in running state. Needless to say the jobs are not actually running.

[...]

PR #1767, opened 2015-07-29
RFR: Update execution object when liveaction is updated

lakshmi-kannan:
Fixes: https://github.com/StackStorm/st2/issues/1717

Several things were uncovered with this bug report.

1) A non UTF-8 string '\x82' in stdout causes mongo to throw an
exception when updating the results. We don't have a check
for validating if string is UTF-8 in document schema.
So this must be some mongo internal validation. We should
figure out if we can turn it off.

2) An exception thrown is caught by worker and liveaction
object is updated but the execution object is not. This
results in execution state shown as `running` even though
liveaction status is `failed`.

3) While fixing this, I also noticed that when an action is
canceled, we don't execute the execution object status to
`canceled` which also is consistent with the bug report.

## TODO: 

* [X] - Tests
* [ ] - CHANGELOG

## Example:

```
(virtualenv)/m/s/s/st2 git:master ❯❯❯ st2 run core.local cmd="python -c 'print \"\x82\"'" -a                 ✭ ✱
To get the results, execute:
 st2 execution get 55b948ca32ed3506d57fabff
(virtualenv)/m/s/s/st2 git:master ❯❯❯ st2 execution get 55b948ca32ed3506d57fabff                             ✭ ✱
id: 55b948ca32ed3506d57fabff
status: failed
result:
{
    "traceback": "  File "/mnt/src/storm/st2/st2actions/st2actions/worker.py", line 102, in process
    result = self.container.dispatch(liveaction_db)
  File "/mnt/src/storm/st2/st2actions/st2actions/container/base.py", line 59, in dispatch
    liveaction_db = self._do_run(runner, runnertype_db, action_db, liveaction_db)
  File "/mnt/src/storm/st2/st2actions/st2actions/container/base.py", line 135, in _do_run
    result, context)
  File "/mnt/src/storm/st2/st2actions/st2actions/container/base.py", line 179, in _update_live_action_db
    liveaction_db=liveaction_db)
  File "/mnt/src/storm/st2/st2common/st2common/util/action_db.py", line 195, in update_liveaction_status
    liveaction_db = LiveAction.add_or_update(liveaction_db)
  File "/mnt/src/storm/st2/st2common/st2common/persistence/base.py", line 118, in add_or_update
    model_object = cls._get_impl().add_or_update(model_object)
  File "/mnt/src/storm/st2/st2common/st2common/models/db/__init__.py", line 134, in add_or_update
    instance.save()
  File "/mnt/src/storm/st2/virtualenv/local/lib/python2.7/site-packages/mongoengine/document.py", line 267, in save
    upsert=True, **write_concern)
  File "/mnt/src/storm/st2/virtualenv/local/lib/python2.7/site-packages/pymongo/collection.py", line 572, in update
    check_keys, self.uuid_subtype), safe)
",
    "error": "strings in documents must be valid UTF-8: '\x82\n'"
}
(virtualenv)/m/s/s/st2 git:master ❯❯❯
```

### DIFF
--- a/st2actions/st2actions/container/base.py
+++ b/st2actions/st2actions/container/base.py
@@ -129,10 +129,17 @@ class RunnerContainer(object):
             LOG.debug('Action "%s" completed.' % (action_db.name), extra=extra)
 
             # Always clean-up the auth_token
-            updated_liveaction_db = self._update_live_action_db(liveaction_db.id, status,
-                                                                result, context)
-            executions.update_execution(updated_liveaction_db)
+            try:
+                LOG.debug('Setting status: %s for liveaction: %s', status, liveaction_db.id)
+                updated_liveaction_db = self._update_live_action_db(liveaction_db.id, status,
+                                                                    result, context)
+            except:
+                error = 'Cannot update LiveAction object for id: %s, status: %s, result: %s.' % (
+                    liveaction_db.id, status, result)
+                LOG.exception(error)
+                raise
 
+            executions.update_execution(updated_liveaction_db)
             extra = {'liveaction_db': updated_liveaction_db}
             LOG.debug('Updated liveaction after run', extra=extra)
 
--- a/st2actions/st2actions/worker.py
+++ b/st2actions/st2actions/worker.py
@@ -13,6 +13,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import sys
+import traceback
+
 from kombu import Connection
 from oslo_config import cfg
 
@@ -58,10 +61,11 @@ class ActionExecutionDispatcher(consumers.MessageHandler):
             LOG.info('%s is not executing %s (id=%s) with "%s" status.',
                      self.__class__.__name__, type(liveaction), liveaction.id, liveaction.status)
             if not liveaction.result:
-                action_utils.update_liveaction_status(
+                updated_liveaction = action_utils.update_liveaction_status(
                     status=liveaction.status,
                     result={'message': 'Action execution canceled by user.'},
                     liveaction_id=liveaction.id)
+                executions.update_execution(updated_liveaction)
             return
 
         if liveaction.status != action_constants.LIVEACTION_STATUS_SCHEDULED:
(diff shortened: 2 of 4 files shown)


manasdk:
@lakshmi-kannan Do you plan to tackle the UTF-8 issue? I think mongo requires UTF-8; not sure.

lakshmi-kannan:
@manasdk: Result for action is now an EscapedDynamicDictField. This helps us because result can be free form. If the value is a string and is a non UTF-8 character, mongo doesn't let you save it. This is as per mongo design. https://jira.mongodb.org/browse/PHP-82.

I would say that non utf-8 is basically invalid data. We have to think about supporting binary data as output and maybe as part of that, we could address this.

manasdk:
Works for me. I am ok with requiring utf8 encoded strings. 

Binary...  Hmm. We should figure this part out

Kami:
> 2) An exception thrown is caught by worker and liveaction
object is updated but the execution object is not. This
results in execution state shown as running even though
liveaction status is failed.

> 3) While fixing this, I also noticed that when an action is
canceled, we don't execute the execution object status to
canceled which also is consistent with the bug report.

Good catch.

st2actions/st2actions/container/base.py
```diff
@@ -129,10 +129,17 @@ def _do_run(self, runner, runnertype_db, action_db, liveaction_db):
...
-            executions.update_execution(updated_liveaction_db)
+            try:
+                LOG.debug('Setting status: %s for liveaction: %s', status, liveaction_db.id)
+                updated_liveaction_db = self._update_live_action_db(liveaction_db.id, status,
+                                                                    result, context)
+            except:
+                error = 'Cannot update LiveAction object for id: %s, status: %s, result: %s.' % (
+                    liveaction_db.id, status, result)
+                LOG.exception(error)
+                raise
```
Kami: Slightly confused - this doesn't really change the behavior since we `raise` here. If updating the live action fails, we will re-throw the exception which means execution won't be updated. The only difference right now is that we log the message.

Is this intentional?

Kami:
> Works for me. I am ok with requiring utf8 encoded strings.

> Binary... Hmm. We should figure this part out

Agreed. We should only support utf-8 strings for now.

If user wants to store / pass through binary data they should base64 encode it on the way in and base64 decode it on the way out (that's similar to what we do here - https://github.com/StackStorm/st2contrib/pull/224/files#diff-4586d63da60760f60d1a61ae2c510937R28).

st2actions/st2actions/container/base.py
lakshmi-kannan: Yes. When I was debugging, the action actually succeeded but updating the liveaction failed because of non UTF-8 characters. This log line would have been so useful. Also the LOG.debug line above. So this is intentional. No behavior change intended.

manasdk:
+1

PR #1772, opened 2015-07-30
St2client py27

DoriftoShoes:
This fixes the shebang on the st2 CLI for systems that use something other than Python 2.7 as the system Python version.


lakshmi-kannan:
consistency ftw. let's hope nothing breaks. lgtm. :ship:

PR #1775, opened 2015-07-30
Fix escaped liveaction result

m4dcoder:
The DB model permanently changed the dict result when escaping the characters for persistence and then cascade the changes to other components. The fix makes a copy of the dict before escaping.


lakshmi-kannan:
Looks same as https://github.com/StackStorm/st2/pull/1747/files. Needs a changelog entry in 0.12.1. Otherwise, LGTM.

m4dcoder:
Ok.  Keep forgetting to add CHANGELOG entry.

Kami:
:+1:

PR #1777, opened 2015-07-31
Allow user to include additional context when submitting debug info

Kami:
This change allows user to include additional context (name, email and optional email) when submitting debug info.

Keep in mind that this is opt-in and user is not required to provide this info.

### DIFF
--- a/st2debug/st2debug/cmd/submit_debug_info.py
+++ b/st2debug/st2debug/cmd/submit_debug_info.py
@@ -312,6 +312,9 @@ def create_archive(include_logs, include_configs, include_content, include_syste
             file_path = os.path.normpath(file_path)
             source_dir = file_path
 
+            if not os.path.exists(source_dir):
+                continue
+
             if '.' in file_path:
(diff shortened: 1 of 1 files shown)


st2debug/st2debug/cmd/submit_debug_info.py
```diff
@@ -447,6 +463,19 @@ def main():
             print('Aborting')
             sys.exit(1)
 
+    # Prompt user for optional additional context info
+    user_info = {}
+
+    print('If you want us to get back to you via email, you can provide additional context '
```
lakshmi-kannan: I'd just say "Please provide us your name, email and an optional comment if you want us to contact you about this report". Wdyt?

lakshmi-kannan:
lgtm

st2debug/st2debug/cmd/submit_debug_info.py
Kami: I'm fine with whatever :P

/cc @dzimine

PR #1778, opened 2015-07-31
Mistral itests in Travis

armab:
Adds existing mistral tests to Travis.

As I remember there were some mistral installation problems before (dependencies). This addition might be useful to identify and react on mistral-related issues faster and it's just cool to have more tests in Travis.

---

[`scripts/travis/setup-mistral.sh`](https://github.com/armab/st2/blob/c2a8b002b92f554d7bd967f9ab5de665bf9d1064/scripts/travis/setup-mistral.sh) uses code from [`st2_deploy.sh`](https://github.com/StackStorm/st2/blob/master/tools/st2_deploy.sh)

The thing I'm afraid of is synchronization with original `st2_deploy.sh`. I couldn't find an easy way to reuse bash functions from that script.
Not sure if it worth to rewrite a bit `st2_deploy.sh` and use Pythonish way `if __name__ == '__main__':` (you got the idea) to allow reusing functions from that file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -304,6 +304,15 @@ mistral-itests: requirements .mistral-itests
 	@echo
 	. $(VIRTUALENV_DIR)/bin/activate; nosetests -s -v st2tests/integration || exit 1;
 
+.PHONY: .mistral-itests-coverage-html
+.mistral-itests-coverage-html:
+	@echo
+	@echo "==================== MISTRAL integration tests with coverage (HTML reports) ===================="
+	@echo "The tests assume both st2 and mistral are running on localhost."
+	@echo
+	. $(VIRTUALENV_DIR)/bin/activate; nosetests -s -v --with-coverage \
+		--cover-inclusive --cover-html st2tests/integration || exit 1;
+
 .PHONY: rpms
 rpms:
 	@echo
--- a/Makefile
+++ b/Makefile
@@ -302,7 +302,7 @@ mistral-itests: requirements .mistral-itests
 	@echo "==================== MISTRAL integration tests ===================="
 	@echo "The tests assume both st2 and mistral are running on localhost."
 	@echo
-	. $(VIRTUALENV_DIR)/bin/activate; nosetests -s -v st2tests/integration || exit 1;
+	. $(VIRTUALENV_DIR)/bin/activate; nosetests -s -v st2tests/integration/mistral || exit 1;
 
 .PHONY: .mistral-itests-coverage-html
 .mistral-itests-coverage-html:
@@ -311,7 +311,7 @@ mistral-itests: requirements .mistral-itests
 	@echo "The tests assume both st2 and mistral are running on localhost."
 	@echo
 	. $(VIRTUALENV_DIR)/bin/activate; nosetests -s -v --with-coverage \
-		--cover-inclusive --cover-html st2tests/integration || exit 1;
+		--cover-inclusive --cover-html st2tests/integration/mistral || exit 1;
 
 .PHONY: rpms
 rpms:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,11 @@ python:
 env:
   - TASK=unit
   - TASK=integration
+  - TASK=mistral
 
 services:
   - mongodb
+  - postgresql
   - rabbitmq
 
 cache:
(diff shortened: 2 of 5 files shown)


Kami:
Nice.

I still have the same concerns as I had before (local, CI and Travis CI builders all use different setups and setup script which makes it hard debug and to maintain), but having all the tests running on TravisCI is good, so I won't block.

Kami:
> The thing I'm afraid of is synchronization with original st2_deploy.sh. I couldn't find an easy way to reuse bash functions from that script.
Not sure if it worth to rewrite a bit st2_deploy.sh and use Pythonish way if __name__ == '__main__': (you got the idea) to allow reusing functions from that file.

Yeah, making the script and parts more reusable would be nice (this would also make it easier to unit test), but I'm not sure how much time we want to spend on it.

We also need to decide what's the path going forward for this script. We now have this script, pre-built images and we are also working on an installer. We definitely need an easy way to set up everything on a single box, but we don't have enough bandwidth to maintain that many installation methods.

Ideally, we could also split functionality into multiple files since this would make it easier to maintain. @DoriftoShoes just opened a PR which adds tons of code to the file which makes it a bit unwieldy :P

DoriftoShoes:
I won't pollute this PR but there are things we can do for st2_deploy.sh to make it less "unwieldy" for code reading but it would make it more "unwieldy" from the support/debugging angle.

manasdk:
Can we make it optional?  I know these tests are non - voting.  Travis is slow as is this will make it worse and pretty make me ignore them anyway. 

Btw,  mistral tests run post the merge commit anyway. I would be happier if we worked towards making those tests and these tests align.

Makefile
```diff
@@ -304,6 +304,15 @@ mistral-itests: requirements .mistral-itests
...
 	. $(VIRTUALENV_DIR)/bin/activate; nosetests -s -v st2tests/integration || exit 1;
 
+.PHONY: .mistral-itests-coverage-html
+.mistral-itests-coverage-html:
+	@echo
+	@echo "==================== MISTRAL integration tests with coverage (HTML reports) ===================="
+	@echo "The tests assume both st2 and mistral are running on localhost."
+	@echo
+	. $(VIRTUALENV_DIR)/bin/activate; nosetests -s -v --with-coverage \
+		--cover-inclusive --cover-html st2tests/integration || exit 1;
```
lakshmi-kannan: integration/mistral/

lakshmi-kannan:
I agree with manas. I'd hate for mistral tests to block a PR unless the PR itself is dealing with mistral changes. We can first turn them on in non-voting mode (if Travis allows that). I am sure we'll find issues and once the tests stabilize, we can debate about making it fail the PR.

armab:
Added things you asked: 
* If mistral is failed - don't mark build as red.
* At the same time, mark build as finished without waiting for mistral test results.

StackStorm has a lot of points of failure. I'm for more flags that could signalize if something is wrong.

armab:
BTW, I played during this weekend with circleci.com:
* [Here is build example with 3 parallel tasks](https://circleci.com/gh/armab/st2/59)

Just for your information, some facts I liked about it: 

* Very fast. It runs x1.5 - x2 times faster comparing to TravisCI. Also every node has 4GB of memory.
* 4 parallel builds (nodes) are allowed with possibility to SSH between the running nodes.
* SSH debug mode for 30 minutes.
* Can work with Docker (run/build/deploy docker).

----

My thinking was:
because it's so robust and fast, we can use it for `st2tests` in some future.

My limited understanding didn't allow me to make those `st2tests` running against some dev branch. I could [build deb packages](https://github.com/armab/st2/blob/circleci/circle.yml#L25) and install them under CircleCI.
As I understand, to run `st2tests` in development environment I need to rework `st2_deploy.sh` script (to have full installation). Correct me if I'm wrong.

For now I won't dive more into it, because `st2tests` is kind of black magic box for me, with almost no documentation.

---

But in future, I believe we can use those free for OpenSource horse-power services to signalize every GitHub pull request. So any developer could write st2test when implementing new st2 features with less configuration efforts.

armab:
Let me know what do you think about it/any existing plans, to have better vision.


And this PR is RFR now.

armab:
@lakshmi-kannan @manasdk please take a look.

As requested, mistral itests won't block the build.

lakshmi-kannan:
I like it. I'll merge the PR.

>My limited understanding didn't allow me to make those st2tests running against some dev branch. 
st2tests are simply st2 actions and workflows that test st2. There is an implicit assumption that action chains work. We also use python actions and core.local action in the action chains. So in short, it's not a great testing engine but it has helped us catch some issues. Running these tests require a running st2 that we want to test against and packs in st2tests installed to that st2 installation. We can probably run these tests in a Dockerized st2 environment when that becomes ready.

PR #1779, opened 2015-07-31
Update st2client version to 0.13dev

m4dcoder:


DoriftoShoes:
Ouch.  We need to get this added to the release workflow.

m4dcoder:
Yep

PR #1780, opened 2015-07-31
Action: fix default parameters override

### LINKED ISSUE #1654: Infinite timeout in action parameters

armab:
Fixes #1654 , when runner parameters were not overridden with "falsey" action parameters (0, False, etc), so this meta was buggy:

```yaml
---
  name: "test"
  description: "Action that tests timeout."
  enabled: true
  entry_point: ""
  parameters:
    timeout:
      # this worked as default 60 seconds
      # default: 0

      # however this worked as 0 seconds
      # default: "0"
    cmd:
      default: "echo 'TIMEOUT: {{timeout}}'"
```

(extracted from #1712 )

### DIFF
--- a/st2actions/st2actions/utils/param_utils.py
+++ b/st2actions/st2actions/utils/param_utils.py
@@ -65,7 +65,7 @@ def _get_resolved_runner_params(runner_parameters, action_parameters,
         # pickup the override.
         if param_name in action_parameters:
             action_param = action_parameters[param_name]
-            if action_param.get('default', False):
+            if 'default' in action_param:
                 resolved_params[param_name] = action_param['default']
 
             # No further override (from liveaction) if param is immutable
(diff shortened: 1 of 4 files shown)


Kami:
LGTM, +1

Kami:
Merged, thanks.

PR #1787, opened 2015-08-04
Freeze tooz version

manasdk:
* tooz 1.21.0 breaks test and unleashes monsters from the pits of hell. So until
  the gods have mercy on us and fix the break freeze to last known version.
* Test test_over_threshold (tests.unit.policies.test_concurrency.ConcurrencyPolicyTest)
  woudl simply block.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,7 @@ pyyaml
 requests
 setuptools==11.1
 six==1.9.0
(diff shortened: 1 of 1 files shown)


Kami:
Good catch. :+1:

armab:
God bless you!

PR #1788, opened 2015-08-04
Validate action parameters in tasks of Mistral WF

m4dcoder:
Throw exception if required parameters are not given and if unexpected parameters are present.

### DIFF
--- a/st2actions/st2actions/runners/mistral/utils.py
+++ b/st2actions/st2actions/runners/mistral/utils.py
@@ -101,13 +101,41 @@ def _eval_inline_params(spec, action_key, input_key):
             _merge_dicts(spec[input_key], inputs)
 
 
+def _validate_action_params(spec, action_key, input_key, action):
+
+    if not action or action_key not in spec or spec.get(action_key) != 'st2.action':
+        return
+
+    action_input = spec.get(input_key)
+    action_params = set(action_input.get('parameters', {}).keys())
+
+    # Check required parameters that have no default defined.
+    required = set([name for name, meta in six.iteritems(action.parameters)
+                    if meta.get('required', False) and 'default' not in meta])
+
+    missing = sorted(required.difference(action_params))
+
+    if missing:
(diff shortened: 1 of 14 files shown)


st2actions/st2actions/runners/mistral/utils.py
```diff
@@ -101,14 +101,42 @@ def _eval_inline_params(spec, action_key, input_key):
             _merge_dicts(spec[input_key], inputs)
 
 
+def _validate_action_params(spec, action_key, input_key, action):
+
+    if not action or action_key not in spec or spec.get(action_key) != 'st2.action':
+        return
+
+    action_input = spec.get(input_key)
+    action_params = set(action_input.get('parameters', {}).keys())
```
Kami: If in the line `.get` above returns `None`, this line will throw since you don't pass a default value to get, right?

Kami:
I made a similar change to our custom json schema validator code a while ago (support for default values).

It looks like that the code in this PR follows the same validation logic as we do for action parameters so I'm curious if we can re-use the same code we use for actions (this way we have less code to maintain and we don't need to worry the validation logic will diverge)?

st2actions/st2actions/runners/mistral/utils.py
m4dcoder: If the parameters key doesn't exists, it returns an empty dict here.

m4dcoder:
@Kami: Can you point me exactly to the validator code you're referring to?

m4dcoder:
@Kami: are you referring to https://github.com/StackStorm/st2/blob/master/st2common/st2common/util/schema/__init__.py#L98?  If so, I'm not trying to assign default here.  I'm just checking if the list of required parameters (w/ no defaults) are provided or if there are unexpected parameters in the list.

Kami:
@m4dcoder Yeah, I was referring to that.

In this case that's OK. We should probably just move it to action param utils module since it's something which is can also be re-used in other parts, not just Mistral :)

PR #1790, opened 2015-08-04
Add ldap backend to st2auth

ruslantum:
Backend which reads authentication information from a ldap server

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,6 +24,7 @@ passlib>=1.6.2,<1.7
 lockfile>=0.10.2,<0.11
 python-gnupg>=0.3.7,<0.4
 jsonpath-rw>=1.3.0
+ldap
 # Requirements for linux pack
 # used by file watcher sensor
 pyinotify>=0.9.5,<=0.10
--- a/st2auth/st2auth/backends/ldap_backend.py
+++ b/st2auth/st2auth/backends/ldap_backend.py
@@ -30,7 +30,10 @@ LOG = logging.getLogger(__name__)
 
 class LdapAuthenticationBackend(BaseAuthenticationBackend):
     """
-    Backend which reads authentication information from a ldap server
+    Backend which reads authentication information from a ldap server.
+    The backend tries to bind the ldap user with given username and password.
+    If the bind was successful, it tries to find the user in the given group.
+    If the user is in the groupi, he will be authenticated.
     """
 
     def __init__(self, ldap_server, base_dn, group_dn):
@@ -39,6 +42,8 @@ class LdapAuthenticationBackend(BaseAuthenticationBackend):
         :type ldap_server: ``str``
         :param base_dn: Base DN on the LDAP Server
         :type base_dn: ``str``
+        :param group_dn: Group DN on the LDAP Server which contains the user as member
+        :type group_dn: ``str``
         """
         self._ldap_server = ldap_server
         self._base_dn = base_dn
@@ -47,19 +52,23 @@ class LdapAuthenticationBackend(BaseAuthenticationBackend):
     def authenticate(self, username, password):
         search_filter = "uniqueMember=uid=" + username + "," + self._base_dn
         try:
-            connect = ldap.open(self._ldap_server)
-            connect.bind_s("uid=" + username + "," + self._base_dn, password)
-            result = connect.search_s(self._group_dn, ldap.SCOPE_SUBTREE, search_filter)
-            connect.unbind_s()
-            if result is None:
-                LOG.debug('User "%s" doesn\'t exist' % (username))
-            elif result is False:
-                LOG.debug('Invalid password for user "%s"' % (username))
-            elif result is True:
-                LOG.debug('Authentication for user "%s" successful' % (username))
-            return bool(result)
-        except ldap.LDAPError:
-            LOG.debug('LDAP Error')
+            connect = ldap.initialize(self._ldap_server)
+            connect.simple_bind_s("uid=" + username + "," + self._base_dn, password)
+            try:
+                result = connect.search_s(self._group_dn, ldap.SCOPE_SUBTREE, search_filter)
+                if result is None:
+                    LOG.debug('User "%s" doesn\'t exist in group "%s"' % (username, self._group_dn))
+                elif result:
+                    LOG.debug('Authentication for user "%s" successful' % (username))
+                    return True
+                return False
+            except:
+                return False
+            finally:
+                connect.unbind()
+        except ldap.LDAPError as e:
+            LOG.debug('LDAP Error: %s' % (str(e)))
+            return False
 
     def get_user(self, username):
         pass
--- a/st2auth/st2auth/backends/ldap_backend.py
+++ b/st2auth/st2auth/backends/ldap_backend.py
@@ -1,0 +1,65 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# to be able to import ldap run pip install python-ldap
+import ldap
+
+from st2common import log as logging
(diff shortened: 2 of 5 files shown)


jfryman:
Nice!

Thanks for the code. Please make sure that all tests are passing. We'll take a look and provide you some comments so we can merge this in. 

Stay tuned!

Kami:
Yep, what james said - thanks for your contribution :)

I will review it and add inline comments shortly.

st2auth/st2auth/backends/ldap_backend.py
```diff
@@ -0,0 +1,65 @@
...
+        :type ldap_server: ``str``
+        :param base_dn: Base DN on the LDAP Server
+        :type base_dn: ``str``
+        """
+        self._ldap_server = ldap_server
+        self._base_dn = base_dn
+        self._group_dn = group_dn
+
+    def authenticate(self, username, password):
+        search_filter = "uniqueMember=uid=" + username + "," + self._base_dn
```
Kami: I'm not that familiar with Python `ldap` package, but it seems like this opens up a potential injection and security vulnerability if you pass in bad value for `username` argument.

Is there any safer way to do it (e.g. some way to bind the parameters in the string) or do we need to validate and sanitize the username ourselves?

st2auth/st2auth/backends/ldap_backend.py
```diff
@@ -0,0 +1,65 @@
...
+            connect.unbind_s()
+            if result is None:
+                LOG.debug('User "%s" doesn\'t exist' % (username))
+            elif result is False:
+                LOG.debug('Invalid password for user "%s"' % (username))
+            elif result is True:
+                LOG.debug('Authentication for user "%s" successful' % (username))
+            return bool(result)
+        except ldap.LDAPError:
+            LOG.debug('LDAP Error')
```
Kami: We should also explicitly return `False` on exception.

And we should also log the exception to help with the debugging (either by doing `LOG.debug('LDAP error: %s' % (str(e)` or by using `LOG.exception`).

st2auth/st2auth/backends/ldap_backend.py
```diff
@@ -0,0 +1,65 @@
...
+        self._base_dn = base_dn
+        self._group_dn = group_dn
+
+    def authenticate(self, username, password):
+        search_filter = "uniqueMember=uid=" + username + "," + self._base_dn
+        try:
+            connect = ldap.open(self._ldap_server)
+            connect.bind_s("uid=" + username + "," + self._base_dn, password)
+            result = connect.search_s(self._group_dn, ldap.SCOPE_SUBTREE, search_filter)
+            connect.unbind_s()
```
Kami: We should probably wrap the search statement in `try / except` block and call `unbind_s` inside `finally` block to make sure connection is always closed (even if exception is thrown).

st2auth/st2auth/backends/ldap_backend.py
```diff
@@ -0,0 +1,65 @@
...
+        :type base_dn: ``str``
+        """
+        self._ldap_server = ldap_server
+        self._base_dn = base_dn
+        self._group_dn = group_dn
+
+    def authenticate(self, username, password):
+        search_filter = "uniqueMember=uid=" + username + "," + self._base_dn
+        try:
+            connect = ldap.open(self._ldap_server)
```
Kami: Looking at the docs (http://python-ldap.org/doc/html/ldap.html#ldap.open), `ldap.open` is deprecated in favor of `ldap.initialize` which is also more flexible because it allows user to pass in the URI.

st2auth/st2auth/backends/ldap_backend.py
```diff
@@ -0,0 +1,65 @@
...
+            connect.bind_s("uid=" + username + "," + self._base_dn, password)
+            result = connect.search_s(self._group_dn, ldap.SCOPE_SUBTREE, search_filter)
+            connect.unbind_s()
+            if result is None:
+                LOG.debug('User "%s" doesn\'t exist' % (username))
+            elif result is False:
+                LOG.debug('Invalid password for user "%s"' % (username))
+            elif result is True:
```
Kami: To prevent potential programmer errors, it would be safer to explicitly return `True` `if result is True` and explicitly return False at the end.

This way, even if some reason code falls through to the end, we will still explicitly return False.

requirements.txt
```diff
@@ -24,6 +24,7 @@ passlib>=1.6.2,<1.7
 lockfile>=0.10.2,<0.11
 python-gnupg>=0.3.7,<0.4
 jsonpath-rw>=1.3.0
+ldap
```
Kami: Yep, that's fine for now (some more context is available here - https://github.com/StackStorm/st2/pull/1732#issuecomment-123224276).

Eventually, I need to spend a couple of days and make each auth backend a separate package so the user only installs the one they need.

Kami:
I've added some in-line comments.

The pull request looks like a good start - thanks.

After the comments are addressed, it would also be good to add some tests and documentation on how to use this backend (see `docs/source/authentication.rst` for example).

ruslantum:
I didn't find anything how to make the bind safer. And about the tests: I can write a unit test for the backend, but it will work only with my ldap server. How do I write a test which will work everywhere?

Kami:
@ruslantum I will research bind situation later on.

As far as the tests go - unit tests where we mock some of the LDAP client methods should be sufficient for now (integration / end to end tests would also be nice, but that's not a big priority right now).

requirements.txt
manasdk: Hmm ... looks like there is no package ldap on pypi. Do you want to use https://pypi.python.org/pypi/python-ldap?

This is also most likely the root cause of the travis and scrutinizer failures.

manasdk:
Btw, if we do plan on using python-ldap then the system on which the build (pip install python-ldap) will need to satisfy http://www.python-ldap.org/doc/html/installing.html#prerequisites.

This might need some work to update all our dev environments etc.

st2auth/st2auth/backends/ldap_backend.py
```diff
@@ -0,0 +1,74 @@
...
+        :type group_dn: ``str``
+        """
+        self._ldap_server = ldap_server
+        self._base_dn = base_dn
+        self._group_dn = group_dn
+
+    def authenticate(self, username, password):
+        search_filter = "uniqueMember=uid=" + username + "," + self._base_dn
+        try:
+            connect = ldap.initialize(self._ldap_server)
```
Kami: I've checked the LDAP package documentation and it would also be good if we supported TLS (we can make it a config option).

See http://www.python-ldap.org/doc/html/ldap.html#ldap.LDAPObject.start_tls_s

st2auth/st2auth/backends/ldap_backend.py
jfryman: The search filter should also allow specifying which attribute of an LDAP entry is queried against. `sAMAccountName` for Active Directory comes to mind

st2auth/st2auth/backends/ldap_backend.py
```diff
@@ -0,0 +1,74 @@
...
+            connect.simple_bind_s("uid=" + username + "," + self._base_dn, password)
+            try:
+                result = connect.search_s(self._group_dn, ldap.SCOPE_SUBTREE, search_filter)
```
jfryman: The only other thing I can think of on top of the comments already made: scope search. Right now, it's assuming `subtree` search, but I also think that should be a configurable variable as well.

Won't block on that, so if it doesn't get added here I'll configure it later.

Kami: Yeah, I've checked some other implementations and ideally we would allow user to also specify a filter query (e.g. something like that https://github.com/ContinuumIO/flask-ldap-login/blob/master/flask_ldap_login/__init__.py#L24)

Kami:
@ruslantum Sorry for the delay.

There is more work which needs to be done on my side before this can get merged - I need to add support for per-backend requirements file. This requires quite a bit of work and right now I'm busy with other things so I can't promise an ETA.

In any case, in the mean time I will also try to come up with a quick work-around so we can get this merged without breaking the build and installation without requiring me to do per-backed requirements stuff.

Kami:
Sorry for the delay, but I finally have some good news :)

I starting working on the "package per backend" model which means this pull request is now superseded by https://github.com/StackStorm/st2-auth-backend-ldap.

You will be able to install this backend when this PR is merged - https://github.com/StackStorm/st2/pull/1881 (either when a new release with this change is available or as soon as the PR is merged if you are OK with using "dev" packages).

Thanks again.

Kami:
@ruslantum When you have some time it would also be great if you can add some unit tests and open a PR against this repo - https://github.com/StackStorm/st2-auth-backend-ldap

Thanks!

st2auth/st2auth/backends/ldap_backend.py
juncas2205-debug: [Uploading Rompiendo la confianza_ Matices de crisis en una cadena de suministro de software insegura - Atlantic Council.pdf…]()

PR #1791, opened 2015-08-05
Misc API improvements and fixes

Kami:
I just cherry-picked some general API improvements and fixes from my other branch.

The biggest one is correctly returning 404 if user hits a URL which doesn't exist, but partially maps to an existing controller. Previously, an internal error was thrown.

Keep in mind that this PR only fixes it for controllers which don't abuse kwargs. Controllers which abuse kwargs will silently accept all the arguments and 404 won't be thrown (I will fix this some other time).

### DIFF
--- a/st2common/st2common/models/api/base.py
+++ b/st2common/st2common/models/api/base.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import re
 import abc
 import copy
 import functools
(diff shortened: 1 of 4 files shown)


st2common/st2common/models/api/base.py
```diff
@@ -171,7 +172,17 @@ def callfunction(*args, **kwargs):
                 pecan.response.status = status_code
                 return json_encode(None)
 
-            result = f(*args, **kwargs)
+            try:
+                result = f(*args, **kwargs)
+            except TypeError as e:
+                message = str(e)
+                # Invalid number of arguments passed to the function meaning invalid path was
```
Kami: Yes, the code is less ideal to put it nicely, but it works for now.

Eventually, when we fix kwargs abuse we should be able to solve in a slightly nicer manner by inspecting the method signature (we already kinda do this if `jsexpose` declares arg_types).

PR #1793, opened 2015-08-05
Support to allow ssh_config for fabric based runners.

manasdk:
* Setting is not enabled by default but exposed as configuration to avoid
  any odd behavior - Since it is fabric afterall.
* Fabric support for ssh_config is limited and should only really be used
  for simple things like ports (http://docs.fabfile.org/en/1.4.0/usage/execution.html#leveraging-native-ssh-config-files)


manasdk:
merging my changes since it does not have really affect default behavior.

Kami:
LGTM, but the question is if we also plan to support this in the new Paramiko based SSH runner?

PR #1798, opened 2015-08-06
Automatically restart sensors which die inside the sensor containainer

Kami:
This pull request adds support for automatically restarting sensor processes which die to the sensor container.

Sensor process is only restarted if it dies / exists with a non-zero status code and we only attempt to restart it 2 times before giving up. To prevent a thundering herd effect, we also add a small delay before restarting a sensor.

## Reasoning / Why

From our personal experience, most of the random / unexpected sensor deaths (IRC, Slack, Github, etc.) are related to errors such as  the server side connection being terminated, network issue or some other error which results into an uncaught exception. In most cases, you can simply recover from those errors by restarting the sensor container and therefore restarting the dead sensor.

Yes, in some specific cases, sensors could also be made more robust so they catch specific exceptions, but in most cases when a re-initialization is required (e.g. re-authentication) the safest and best approach is to simply restart the sensor (think erlang processes).

Previously, we have been doing this manually, but now the sensor container will do this for us.

In the future, we should also add support for restarting a particular sensor to sensor container (HTTP API or whatever). This way we can also dog food StackStorm for that (we already dispatch a trigger when a process exists).

### DIFF
--- a/st2reactor/st2reactor/container/process_container.py
+++ b/st2reactor/st2reactor/container/process_container.py
@@ -57,10 +57,7 @@ SENSOR_MAX_RESPAWN_COUNTS = 2
 SENSOR_SUCCESSFUL_START_THRESHOLD = 10
 
 # How long to wait (in seconds) before respawning a dead process
-SENSOR_RESPAWN_DELAY = 2
-
-# Backoff (in seconds) added to respawn delay before each respawn
-SENSOR_RESPAWN_BACKOFF = 1.5
+SENSOR_RESPAWN_DELAY = 2.5
 
 # TODO: Allow multiple instances of the same sensor with different configuration
 # options - we need to update sensors for that and add "get_id" or similar
@@ -360,7 +357,8 @@ class ProcessSensorContainer(object):
 
         LOG.debug('Respawning dead sensor', extra=extra)
 
-        sleep_delay = SENSOR_RESPAWN_DELAY * SENSOR_RESPAWN_BACKOFF
+        self._sensor_respawn_counts[sensor_id] += 1
+        sleep_delay = (SENSOR_RESPAWN_DELAY * self._sensor_respawn_counts[sensor_id])
         eventlet.sleep(sleep_delay)
 
         try:
@@ -370,8 +368,6 @@ class ProcessSensorContainer(object):
 
             # Disable sensor which we are unable to start
             del self._sensors[sensor_id]
-
-        self._sensor_respawn_counts[sensor_id] += 1
 
     def _should_respawn_sensor(self, sensor_id, sensor, exit_code):
         """
--- a/st2reactor/st2reactor/container/process_container.py
+++ b/st2reactor/st2reactor/container/process_container.py
@@ -19,6 +19,8 @@ import time
 import json
 import subprocess
 
+from collections import defaultdict
+
 import eventlet
 from eventlet.support import greenlets as greenlet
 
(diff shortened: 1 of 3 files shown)


jfryman:
> (think erlang processes).

yesssssss! die quickly and often.

st2reactor/st2reactor/container/process_container.py
```diff
@@ -61,65 +76,109 @@ def __init__(self, sensors, poll_interval=5):
...
+
                 self._dispatch_trigger_for_sensor_exit(sensor=sensor,
                                                        exit_code=status)
-                self._delete_sensors(sensor_id)
+
+                # Try to respawn a dead process (maybe it was a simple failure which can be
+                # resolved with a restart)
+                self._respawn_sensor(sensor_id=sensor_id, sensor=sensor, exit_code=status)
+            else:
+                sensor_start_time = self._sensor_start_times[sensor_id]
```
lakshmi-kannan: You probably want to use .get(sensor_id, None) and log an error on None. We fixed some bugs where there were KeyError

Kami: This would indicate a programmer error so I skipped using get.

manasdk:
This is gonna cause me a lot of misery with https://github.com/StackStorm/st2/pull/1739. All the APi stuff you talk about will have to come through that PR.

Have you looked at using the policy mechanism to define if a sensor should be started. Primarily we need some sort of opt-in/opt-out mechanism from this behavior.

manasdk:
Since this solves an immediate problem I am ok with the approach you take here but I do not believe this is how we truly ship the feature.

st2reactor/st2reactor/container/process_container.py
lakshmi-kannan: Bugs are usually programmer errors :P. I'd still insist on a get. It was hard to debug last time this happened.

st2reactor/st2reactor/container/process_container.py
```diff
@@ -61,65 +76,109 @@ def __init__(self, sensors, poll_interval=5):
...
+                sensor_respawn_count = self._sensor_respawn_counts[sensor_id]
+                successfuly_started = (now - sensor_start_time) >= SENSOR_SUCCESSFUL_START_THRESHOLD
+
+                if successfuly_started and sensor_respawn_count >= 1:
```
lakshmi-kannan: s/successfully_started/successfully_running

st2reactor/st2reactor/container/process_container.py
```diff
@@ -280,7 +343,49 @@ def _stop_sensor_process(self, sensor_id, exit_timeout=5):
             # Process hasn't exited yet, forcefully kill it
             process.kill()
 
-        self._delete_sensors(sensor_id)
+        self._delete_sensor(sensor_id)
+
+    def _respawn_sensor(self, sensor_id, sensor, exit_code):
```
lakshmi-kannan: This method is called inside a for loop and there is an eventlet.sleep(). Maybe the respawn should be called in a separate eventlet to prevent one sensor restart from blocking others.

st2reactor/st2reactor/container/process_container.py
Kami: Using get here would mask a bigger issue - value not correctly being set elsewhere (if a sensor is started, which is true in this case, start time should be set).

I would argue that no using get makes it easier to debug since it immediately indicates that code somewhere else is not doing something correctly.

Kami:
@manasdk 

> This is gonna cause me a lot of misery with #1739. All the APi stuff you talk about will have to come through that PR.

I actually some how missed that this PR exists and touches the same files.

> Have you looked at using the policy mechanism to define if a sensor should be started. Primarily we need some sort of opt-in/opt-out mechanism from this behavior.

Allowing to define that would be nice (`restart_on_crash` or similar), but I would actually argue that restarting a sensor which has crashed by default is a sane default behavior (we only restart it on a non-success status code so if sensor is stopped or gracefully exists, it won't be restarted).

st2reactor/st2reactor/container/process_container.py
lakshmi-kannan: It won't mask because you'd still log the sensor_id which doesn't exist and you can raise an exception :P. Anyways, you have my suggestion. I can always say "I told you" :P.

st2reactor/st2reactor/container/process_container.py
Kami: Yeah, I was thinking about that - will make that change.

Kami: Done in c20dd998e20bad4cb2157850128163009b5e2c38.

PR #1800, opened 2015-08-07
The flat_file auth backend reads the Htpasswdfile on every authentica…

manasdk:
…te call.

This enables changes made to htpasswd file dynamic i.e. picked up without
requiring restart of the auth process.


Kami:
LGTM, +1

Need to also update changelog

manasdk:
> Need to also update changelog

Thinking of cherry-picking to 0.12.2 and making Changelog fixes there and moving the 0.12.2 changelog here eventually.

lakshmi-kannan:
Good catch. Would've been nice to add some tests :P.

PR #1802, opened 2015-08-08
Add more indexes to ActionExecutionDB to speed query

m4dcoder:


st2common/st2common/models/db/execution.py
```diff
@@ -70,7 +70,10 @@ class ActionExecutionDB(stormbase.StormFoundationDB):
             {'fields': ['start_timestamp']},
             {'fields': ['end_timestamp']},
             {'fields': ['status']},
-            {'fields': ['parent']}
+            {'fields': ['parent']},
+            {'fields': ['-start_timestamp']},
```
manasdk: From mongo docs

http://docs.mongodb.org/manual/core/index-compound/#sort-order

" For single-field indexes, the sort order of keys doesn’t matter because MongoDB can traverse the index in either direction. "

m4dcoder: Ok. I'll remove the single entry one.

manasdk:
+1


Thanks!

manasdk:
At some point we should look at https://mongoengine-odm.readthedocs.org/guide/defining-documents.html#global-index-default-options and `index_background `. Not this PR but if we notice index creation becoming an issue.

m4dcoder:
Yeah. Good info.

PR #1805, opened 2015-08-10
RFR: oslo_config is the right import

lakshmi-kannan:
cc: @manasdk.


manasdk:
lol.

Still not sure how this actually works.

manasdk:
+1

Kami:
Good catch, indeed :P

But yeah, no idea how this worked on the CI (is this code path not covered? or did it use different oslo.config version?)

PR #1806, opened 2015-08-10
Adding initial all-in-one installation

jfryman:
This PR adds the initial documentation for the All-in-one installer. This adds instructions on how to get the AIO install via Vagrant or AWS.

The actual AMI instructions do not exist yet. Still waiting on upstream approval, but that shouldn't stop a cursory review of content.

More to come!


docs/source/install/all_in_one.rst
```diff
@@ -0,0 +1,136 @@
+All-in-one Installation
+=======================
+|st2| provides an all-in-one installer aimed at assisting users with the initial setup and configuration. The installer comes pre-bundled in a number of different provisioning options for convenience, or can also be manually deployed and installed manually on a server.
+
+Pre-Requsites
```
lakshmi-kannan: aspull.

lakshmi-kannan:
Some info on how to get help on errors and linking them to slack-community in the doc would be useful I think.

lakshmi-kannan:
<3. Good PR.

jfryman:
> Some info on how to get help on errors and linking them to slack-community in the doc would be useful I think.

Ok, I'll keep that in mind as I go through these updates, but not going to stop to block here.

docs/source/install/all_in_one.rst
```diff
@@ -0,0 +1,136 @@
...
+The initial setup is designed to configure itself to a freshly deployed server with no additional services running. If you do not have or do not wish to provide SSL and SSH keys during the initial setup, |st2| will generate random new keys automatically.
+
+Sizing the Server
+-----------------
+A standard all-in-one installation has all of the various components of |st2|, as well as the supporting infrastructure components. While the system can operate with less equipped servers, these are recommended for the best experience while testing or deploying |st2|.
+
+Testing
+~~~~~~~
+* Single or Dual CPU system
+* At least 4GB of RAM
```
dzimine: Really? 4Gb? 

or depenging on which "testing", all our samples and blogs will run with 2GB easily. may be "quick look/evaluating"?

docs/source/install/all_in_one.rst
```diff
@@ -0,0 +1,136 @@
...
+
+              _   ___     ____  _  __
+             | | |__ \   / __ \| |/ /
+          ___| |_   ) | | |  | | ' /
+         / __| __| / /  | |  | |  <
+         \__ \ |_ / /_  | |__| | . \
+         |___/\__|____|  \____/|_|\_\
+    st2 is installed and ready to use.
+    First time starting this machine up?
+    Visit https://172.168.50.11/setup to configure |st2|
```
dzimine: |st2| won't expand in formatted text in Sphinx. Have to type `StackStorm`.

docs/source/install/index.rst
```diff
@@ -53,6 +53,7 @@ and WebUI url:
 .. toctree::
     :maxdepth: 1
 
+    All-in-one Installation <all_in_one>
```
dzimine: I would stress on "Installer" (not installation) `All-in-one Installer`.

dzimine:
Good stuff.

Kami:
Hawt!

docs/source/install/all_in_one.rst
armab: 4Gig looks a lot for me too.

Don't really know this Installer requirements, but I successfully installed production with all st2 components, Chatops with Hubot and Ansible, custom packs on `t2.small` with 2 GiB RAM  [~20$/month](https://aws.amazon.com/ec2/pricing/) and it looks pretty well as minimum viable configuration: 
aws-st2-usage

My thinking about high requirements - is not to make it barrier. But you know better, just my 0.01c.

Kami: Yeah, it would be good to have "minimum required" and "minimum recommended" section.

MongoDB is memory hungry since it directly maps data files into memory, so more is better (especially, when the database size grows).

jfryman: Well, I'm glad this spurred the conversation I hoped it would.

The fact of the matter is that there are two different databases running (Mongo+MySQL), as well as all of the various processes. I actually started at the low bar of 2GB, but thought I would put out the 4GB to see what everyone said.

At the very least, I don't want folks starting up with >1GB, because we have seen folks try to start up in Docker containers and small ec2 instances, only to be surprised that things didn't work.

I think 2GB is reasonable, but does that create the experience we hope? Right now, we only have feelings and anecdotes here. We're coming up on 1.0, and we need to have hard data before then. I know we don't now, but that cannot be for much longer. /cc @dzimine

armab: As I understand now, the idea is: 
* `2GB` is ok enough in general
* but let's keep min `4GB` for this big prod installer

----

Related thing: I want to check (already checking) minimum amount of memory for Ansible provisioner. 
What recommended numbers should I set to keep things in sync? [ [1G](https://github.com/StackStorm/st2workroom/blob/master/stacks/st2.yaml#L5) ], [ [2G](https://github.com/StackStorm/st2workroom/blob/master/stacks/template.yaml#L5) ] or 4 ?

PR #1811, opened 2015-08-11
Fix the sort key and index for action executions

m4dcoder:


st2common/st2common/models/db/execution.py
```diff
@@ -71,9 +71,7 @@ class ActionExecutionDB(stormbase.StormFoundationDB):
             {'fields': ['end_timestamp']},
             {'fields': ['status']},
             {'fields': ['parent']},
-            {'fields': ['-start_timestamp', 'action']},
-            {'fields': ['-start_timestamp', 'parent']},
-            {'fields': ['-start_timestamp', 'action', 'parent']}
```
lakshmi-kannan: Don't we need this with action.ref?

m4dcoder: I removed them and didn't see any difference in the different `st2 execution list` queries.

lakshmi-kannan: Probably because we already index by field parent in line 73.

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -265,7 +265,7 @@ class ActionExecutionsController(ActionExecutionsControllerMixin, ResourceContro
 
     # ResourceController attributes
     query_options = {
-        'sort': ['-start_timestamp', 'action']
+        'sort': ['-start_timestamp', 'action.ref']
```
lakshmi-kannan: We probably need tests although I don't know how this worked before. Hmm.

lakshmi-kannan:
Seems fine to me. Let @manasdk also look.

st2common/st2common/models/db/execution.py
m4dcoder: So at this point, I think we should add them when there's obvious performance problem. Otherwise, it may have negative effect having too many indexes.

lakshmi-kannan: +1 to that. We need some metrics around this.

manasdk:
LGTM

st2api/st2api/controllers/v1/actionexecutions.py
m4dcoder: I'm not sure how to test function in unit test.

m4dcoder:
Thanks.

PR #1812, opened 2015-08-11
RFR: Rule update 500 no pack fix

lakshmi-kannan:


st2common/st2common/models/api/rule.py
```diff
@@ -162,7 +163,7 @@ def to_model(cls, rule):
 
         trigger = reference.get_str_resource_ref_from_model(trigger_db)
         criteria = dict(getattr(rule, 'criteria', {}))
-        pack = str(rule.pack)
+        pack = getattr(rule, 'pack', DEFAULT_PACK_NAME)
```
manasdk: Should we make `pack` required? 

OTH setting pack to default might not be the right approach either.

dzimine: Isn't action defaults to the pack which contains it? Which was handy for creating them in one pack (e.g. default) and moving to another without changing the metadata files. 
I think we should apply the same logic to all types of content.

lakshmi-kannan: +1 to what dzimine says. It should be consistent. We can either make `pack` required everywhere or optional everywhere. 

>Which was handy for creating them in one pack (e.g. default) and moving to another without changing the metadata files.

There is a problem that we don't clean up the one in default.

Kami: Yeah, we have the same logic in actions controller right now (defaulting to "default" pack if `pack` is not provided), but I would prefer us to eventually move to `pack` being required.

Kami: It's also worth clarifying, that as far as I know, this only applies to the rules which were created using the API before we added `pack` field to the `Rule` model.

When we added `pack` field, we also added a change which sets `pack` to a default value if the value is not provided in the API request payload.

lakshmi-kannan:
I am going to rage merge this.

PR #1816, opened 2015-08-12
RFR: Fix trigger-instance manager to handle kwargs passed down

lakshmi-kannan:
An extra keyword token is now passed to the managers which is causing trigger_instance re_emit to break. Gosh, abusing kwargs.

### DIFF
--- a/st2client/st2client/models/core.py
+++ b/st2client/st2client/models/core.py
@@ -333,7 +333,7 @@ class LiveActionResourceManager(ResourceManager):
 
 class TriggerInstanceResourceManager(ResourceManager):
     @add_auth_token_to_kwargs_from_env
-    def re_emit(self, trigger_instance_id):
+    def re_emit(self, trigger_instance_id, **kwargs):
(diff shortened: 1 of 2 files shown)


st2client/st2client/models/core.py
```diff
@@ -333,7 +333,7 @@ def re_run(self, execution_id, parameters=None, **kwargs):
 
 class TriggerInstanceResourceManager(ResourceManager):
     @add_auth_token_to_kwargs_from_env
-    def re_emit(self, trigger_instance_id):
+    def re_emit(self, trigger_instance_id, **kwargs):
```
lakshmi-kannan: I did what we are doing for re_run in executions. Scroll this file and you'll see. I hate this kwargs abuse but whatever.

manasdk:
+1

st2client/st2client/models/core.py
Kami: Fine :P

PR #1818, opened 2015-08-13
Fix bug in triggers emitted on key value pair changes and sensor spawn/exit.

lakshmi-kannan:
We were publishing the triggers with just name before. This failed lookup when creating trigger instances. So basically this feature never worked before. 

I added tests for sensor spawn and exit but for KVP, the code has multiple levels of abstraction.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,6 +24,7 @@ in development
 * Fix key name for error message in liveaction result. (bug-fix)
 * Fix 500 API response when rule with no pack info is supplied. (bug-fix)
 * Fix bug in trigger-instance re-emit (extra kwargs passed to manager is now handled). (bug-fix)
+* Fix bug in triggers emitted on key value pair changes and sensor spawn/exit. (bug-fix)
 
 0.12.1 - July 31, 2015
 ----------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,7 +24,8 @@ in development
 * Fix key name for error message in liveaction result. (bug-fix)
 * Fix 500 API response when rule with no pack info is supplied. (bug-fix)
 * Fix bug in trigger-instance re-emit (extra kwargs passed to manager is now handled). (bug-fix)
-* Fix bug in triggers emitted on key value pair changes and sensor spawn/exit. (bug-fix)
+* Fix bug in triggers emitted on key value pair changes and sensor spawn/exit. When
+  dispatching   those triggers, the reference used didn't contain the pack names which meant it was invalid and lookups in the rules engine would fail. (bug-fix)
 
 0.12.1 - July 31, 2015
 ----------------------
--- a/st2reactor/st2reactor/rules/worker.py
+++ b/st2reactor/st2reactor/rules/worker.py
@@ -48,6 +48,8 @@ class TriggerInstanceDispatcher(consumers.MessageHandler):
 
             if trigger_instance:
                 self.rules_engine.handle_trigger_instance(trigger_instance)
+            else:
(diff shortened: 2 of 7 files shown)


lakshmi-kannan:
@manasdk found the root cause. So praise him. I am just a donkey writing code. No brain involved.

st2reactor/st2reactor/rules/worker.py
```diff
@@ -48,6 +48,8 @@ def process(self, instance):
 
             if trigger_instance:
                 self.rules_engine.handle_trigger_instance(trigger_instance)
+            else:
+                LOG.info('No trigger found in db for instance. %s', instance)
```
lakshmi-kannan: I added a log info statement but debating whether this should be error.

manasdk: Being unable to create the trigger_instance implies that the trigger was not found? I guess if the check was in `create_trigger_instance` I would be more convinced :)

manasdk:
Overall LGTM.

Kami:
Good catch.

This one was pretty nasty since triggers would still get dispatched on the queue, but the lookup / creation of trigger instance in the rules engine would fail, right?

st2reactor/tests/unit/test_process_container.py
```diff
@@ -13,7 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import time
+
 import eventlet
+from mock import (MagicMock, Mock, patch)
```
Kami: Drop the parenthesis :P

st2reactor/tests/unit/test_process_container.py
```diff
@@ -32,3 +35,34 @@ def test_no_sensors_dont_quit(self):
...
+            'class_name': 'pack.StupidSensor'
+        }
+        process = Mock()
+        process_attrs = {'pid': 1234}
+        process.configure_mock(**process_attrs)
+        cmd = 'sensor_wrapper.py --class-name pack.StupidSensor'
+
+        process_container._dispatch_trigger_for_sensor_spawn(sensor, process, cmd)
+        mock_dispatcher.dispatch.assert_called_with(
+            'core.st2.sensor.process_spawn',
```
Kami: That's test is fine for now and definitely a lot better than nothing, but in ideal world, we could some how test the whole pipeline.

(If we previously had a similar test, it probably wouldn't catch this issue since the person writing the test would probably incorrectly just check the name).

Kami:
LGTM

CHANGELOG.rst
```diff
@@ -24,6 +24,7 @@ in development
 * Fix key name for error message in liveaction result. (bug-fix)
 * Fix 500 API response when rule with no pack info is supplied. (bug-fix)
 * Fix bug in trigger-instance re-emit (extra kwargs passed to manager is now handled). (bug-fix)
+* Fix bug in triggers emitted on key value pair changes and sensor spawn/exit. (bug-fix)
```
Kami: Might be good to add a bit more context in the second sentence. "...When dispatching those triggers, the reference used didn't contain the pack names which meant it was invalid and lookups in the rules engine would fail." or similar.

st2reactor/tests/unit/test_process_container.py
lakshmi-kannan: I definitely wrote the test with just the bug fix I am doing in mind. I agree we need end to end tests.

manasdk:
+1

st2reactor/tests/unit/test_process_container.py
lakshmi-kannan: Can I please be a hipster?

PR #1820, opened 2015-08-13
Return correct content type on an error inside the auth hook

### LINKED ISSUE #1819: API should return proper content type for errors
Currently, if you try to get the list of actions without providing a token on `auth: true` host, you'll get a json formatted error message `{"faultstring": "unauthorized"}`, but headers would say this is `Content-Type: text/plain` preventing most of the clients (incuding st2client.js) from automatically parsing it.

If we respond with json, we need to properly announce that through content-type header.

Place in code likely requiring the change: https://github.com/StackStorm/st2/blob/4f1ac919f6c1220d03e9098e4d6949013331ddc7/st2common/st2common/hooks.py#L138-L141

Kami:
The title says it all.

Resolves #1819

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,7 +24,9 @@ in development
 * Fix key name for error message in liveaction result. (bug-fix)
 * Fix 500 API response when rule with no pack info is supplied. (bug-fix)
 * Fix bug in trigger-instance re-emit (extra kwargs passed to manager is now handled). (bug-fix)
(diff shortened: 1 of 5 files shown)


lakshmi-kannan:
Changelog? No tests? :P

Kami:
@lakshmi-kannan Yeah, I was trying to decide if it's worth writing tests for it or not :P 

In any case, I will add it.

Kami:
Added tests and fixed some more issues which I discovered during testing :)

PR #1821, opened 2015-08-13
Add config options for Mistral auth with Keystone

m4dcoder:
Refactor Mistral querier and callback handler to use Mistral Python Client instead of the API. The Mistral python client already handles authentication with Keystone to obtain auth token. This minimizes additional code to talk to the Keystone API directly. Also, the Mistral runner is already using the Mistral python client so there's no impact in terms of additional dependencies. The Keystone configuration is optional. If the options are not provided, then st2 does not pass any credential to Mistral.


st2actions/tests/unit/test_mistral_v2.py
```diff
@@ -226,7 +235,7 @@ def test_launch_workflow(self):
     @mock.patch.object(
         access_service, 'create_token',
         mock.MagicMock(return_value=TOKEN_DB))
-    def test_launch_workflow_with_auth(self):
+    def test_launch_workflow_with_st2_auth(self):
```
lakshmi-kannan: st2_auth?

m4dcoder: Test when our auth is enabled and we passed our token thru.

st2actions/st2actions/resultstracker/config.py
```diff
@@ -46,15 +46,16 @@ def _register_results_tracker_opts():
...
+    # config, notably the mistral section which is registered below.
     mistral_opts = [
-        cfg.StrOpt('v2_base_url', default='http://localhost:8989/v2',
-                   help='Mistral v2 API server root endpoint.'),
-        cfg.IntOpt('max_attempts', default=180,
-                   help='Maximum no of attempts made to connect to Mistral.'),
-        cfg.IntOpt('retry_wait', default=5,
-                   help='Time in seconds to wait before retrying connection to Mistral.')
+        cfg.StrOpt('v2_base_url', default='http://localhost:8989/v2', help='v2 API root endpoint.'),
+        cfg.IntOpt('max_attempts', default=180, help='Max attempts to reconnect.'),
```
lakshmi-kannan: OMG. Dude, this is excessive :P.

st2actions/tests/unit/test_mistral_v2.py
```diff
@@ -264,6 +273,69 @@ def test_launch_workflow_with_auth(self):
...
+    @mock.patch.object(
+        workflows.WorkflowManager, 'get',
+        mock.MagicMock(return_value=WF1))
+    @mock.patch.object(
+        workflows.WorkflowManager, 'create',
+        mock.MagicMock(return_value=[WF1]))
+    @mock.patch.object(
+        executions.ExecutionManager, 'create',
+        mock.MagicMock(return_value=executions.Execution(None, WF1_EXEC)))
+    def test_launch_workflow_with_mistral_auth(self):
```
lakshmi-kannan: If you could move these tests to test_mistral_v2_auth.py, it would be great. This is > 700 lines file :P.

lakshmi-kannan:
LGTM. I don't remember what we decided on keystone auth. So please wait on jfryman or manasdk to double check.

st2actions/st2actions/resultstracker/config.py
m4dcoder: Equates to 15 minutes total retry time. I'm actually thinking of using the retrying python module and then wrap it around with an exponential backoff.  But that's for another PR.

st2actions/tests/unit/test_mistral_v2.py
m4dcoder: Sure I'll move that auth related tests to test_mistral_v2_auth.py.

m4dcoder:
We decided that Keystone is optional. Users can set up Keystone and then configure that here. But the t2_deploy and puppet modules won't include the Keystone install in the setup.

Kami:
I fixed the pull request description so it says Mistral Python client and not CLI.

I was confused for a second when I've read "CLI" since using CLI wouldn't really be a good idea :)

st2actions/st2actions/resultstracker/config.py
Kami: Yeah, using exponential backoff would be better. I'm fine with that being a separate PR.

st2actions/st2actions/handlers/mistral.py
```diff
@@ -62,22 +62,32 @@ def callback(url, context, status, result):
                 # tracker before sending the output.
                 result.pop('tasks', None)
 
+            match = re.search('(.*)/action_executions/(.*)', url)
```
Kami: Need to use `+` (1 or more) here instead of `*` (zero or more) otherwise, there will always be 2 matches if the URL contains at least the `/action_executions/` string.

It also might be worth splitting this into a method so it can be tested.

st2actions/st2actions/handlers/mistral.py
```diff
@@ -62,22 +62,32 @@ def callback(url, context, status, result):
...
+
             for i in range(cfg.CONF.mistral.max_attempts):
                 try:
                     LOG.info('Sending callback to %s with data %s.', url, data)
-                    response = requests.request(method, url, data=json.dumps(data), headers=headers)
-                    if response.status_code == 200:
-                        break
+                    manager.update(action_execution_id, **data)
+                    break
                 except requests.exceptions.ConnectionError as conn_exc:
```
Kami: Does the Mistral Python client throw the same exception if there is a connection issue (just want to confirm that)?

st2actions/st2actions/handlers/mistral.py
```diff
@@ -62,22 +62,32 @@ def callback(url, context, status, result):
...
                     LOG.exception(conn_exc)
                     if i < cfg.CONF.mistral.max_attempts:
                         eventlet.sleep(cfg.CONF.mistral.retry_wait)
-
-            if response and response.status_code != 200:
```
Kami: Hm, if I'm reading this correctly, this changes the existing behavior and means we won't throw even if we fail to reach Mistral after `max_attempts` requests.

Or am I missing something?

st2actions/st2actions/query/mistral/v2.py
```diff
@@ -134,9 +124,6 @@ def _format_task_result(self, task):
         result['created_at'] = task.get('created_at', None)
         result['updated_at'] = task.get('updated_at', None)
         result['state'] = task.get('state', None)
-        result['input'] = task.get('input', None)
```
Kami: Heh, good catch.

st2actions/st2actions/config.py
```diff
@@ -78,12 +78,13 @@ def _register_action_runner_opts():
     CONF.register_opts(ssh_runner_opts, group='ssh_runner')
 
     mistral_opts = [
-        cfg.StrOpt('v2_base_url', default='http://localhost:8989/v2',
-                   help='Mistral v2 API server root endpoint.'),
-        cfg.IntOpt('max_attempts', default=180,
-                   help='Maximum no of attempts made to connect to Mistral.'),
-        cfg.IntOpt('retry_wait', default=5,
-                   help='Time in seconds to wait before retrying connection to Mistral.')
+        cfg.StrOpt('v2_base_url', default='http://localhost:8989/v2', help='v2 API root endpoint.'),
```
Kami: Looks like there are some shared / common options for `mistral` and `resultstracker` section. Might be working refactoring this a bit.

st2actions/st2actions/handlers/mistral.py
m4dcoder: Yes. I tested that.

st2actions/st2actions/handlers/mistral.py
m4dcoder: You're right. I need to fix this. Good catch!

st2actions/st2actions/handlers/mistral.py
m4dcoder: Thanks. I'll split it to a different method.

st2actions/st2actions/config.py
m4dcoder: Yeah I saw that but there was a comment indicating known issue. I don't want to refactor this just yet as it's going to increase the scope of this PR. But I can work on it separately.

jfryman:
No major concerns from me. A comment and a question.

* Why so long on the retry. Seems like that if I have critical workflows depending on Mistral, I don't want to potentially not know for up to 15 minutes that something is wrong while waiting for the results tracker to figure itself out.

I certainly don't want to block on knowing this, but I don't quite understand how having this fits in the rest of our auth story. Will we be locking ourselves to a keystone-only view by not creating an adapter pattern here?

m4dcoder:
The connection errors seems to be recoverable as things move in Mistral. Therefore, I kept the retry on a longer interval. I'll be fixing it with exponential backoff. But that's a completely different scope from this PR. So, end of this discussion for now.

Mistral currently is coupled with Keystone for auth because it's OpenStack. There's no adapter story there. So the question is if we want to ship st2 with Mistral w/o an option to secure Mistral.  I prefer we ship with an option to secure it. This does not mean we stop shopping for a better integration with Mistral.

st2actions/st2actions/query/mistral/v2.py
```diff
@@ -107,19 +107,9 @@ def _get_workflow_tasks(self, exec_id):
         :type exec_id: ``str``
         :rtype: ``list``
         """
-        url = self._get_execution_tasks_url(exec_id)
-        resp = requests.get(url)
-        result = resp.json()
-        tasks = result.get('tasks', [])
-
-        result = []
+        wf_tasks = tasks.TaskManager(self._client).list(workflow_execution_id=exec_id)
```
lakshmi-kannan: How many HTTP calls will be made for this? One per task?

lakshmi-kannan: Looks like it's the same as what we were doing before. https://github.com/StackStorm/python-mistralclient/blob/master/mistralclient/api/v2/tasks.py#L30

m4dcoder: One HTTP call to list all the tasks for the given workflow_execution_id.

lakshmi-kannan:
LGTM

docs/source/mistral.rst
```diff
@@ -126,4 +126,48 @@ There are more workflow examples under :github_st2:`/usr/share/doc/st2/examples
 
 Check out this step-by-step tutorial on building a workflow in |st2| http://stackstorm.com/2015/07/08/automating-with-mistral-workflow/
 
-And more details on Mistral workflow syntax on https://wiki.openstack.org/wiki/Mistral/DSLv2
\ No newline at end of file
+And more details on Mistral workflow syntax on https://wiki.openstack.org/wiki/Mistral/DSLv2
+
+Configuration Options
```
Kami: This should probably go to the installation / configuration section.

Currently, this document contains end user docs so putting configuration info here might make it confusing...

st2actions/st2actions/handlers/mistral.py
```diff
@@ -49,30 +61,35 @@ def callback(url, context, status, result):
...
-                        break
+                    manager.update(action_execution_id, **data)
+                    break
                 except requests.exceptions.ConnectionError as conn_exc:
-                    LOG.exception(conn_exc)
-                    if i < cfg.CONF.mistral.max_attempts:
+                    if i == cfg.CONF.mistral.max_attempts - 1:
+                        raise conn_exc
+                    else:
+                        LOG.exception(conn_exc)
```
Kami: I know you didn't change this in this PR, but it seems it would be better to log this under warn or similar, since we haven't reached max attempts yet and this exception is not fatal yet.

st2actions/st2actions/handlers/mistral.py
```diff
@@ -40,6 +43,15 @@ def get_handler():
     return MistralCallbackHandler
 
 
+def get_action_execution_id_from_url(url):
+    match = re.search('(.+)/action_executions/(.+)', url)
+    if not match or len(match.groups()) != 2:
+        raise ValueError('Unable to extract the action execution ID '
+                         'from the callback URL (%s).' % url)
```
Kami: Minor style thing - it's safer to put parenthesis around the format string arguments (`% (url)`) in case we add more arguments later.

Kami:
I added a couple of more comments. Mostly small things.

Besides that, LGTM, +1

PR #1822, opened 2015-08-13
WIP: Refactor executions exporter to own file

lakshmi-kannan:
I am moving executions exporter to own file. This is needed for introducing exporters for other models. All model exporters will have a base class. This will probably be next PR.

### DIFF
--- a/st2exporter/st2exporter/worker.py
+++ b/st2exporter/st2exporter/worker.py
@@ -13,118 +13,65 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import Queue
-
 import eventlet
+
 from kombu import Connection
-from oslo_config import cfg
 
 from st2common import log as logging
-from st2common.constants.action import (LIVEACTION_STATUS_SUCCEEDED, LIVEACTION_STATUS_FAILED,
-                                        LIVEACTION_STATUS_CANCELED)
-from st2common.models.api.execution import ActionExecutionAPI
-from st2common.models.db.execution import ActionExecutionDB
-from st2common.persistence.execution import ActionExecution
-from st2common.persistence.marker import DumperMarker
-from st2common.transport import consumers, execution, publishers
+from st2common.transport import execution, publishers
 from st2common.transport import utils as transport_utils
-from st2common.util import isotime
-from st2exporter.exporter.dumper import Dumper
+from st2exporter.consumers.execution import ExecutionsExporter
 
-__all__ = [
-    'ExecutionsExporter'
-]
-
-COMPLETION_STATUSES = [LIVEACTION_STATUS_SUCCEEDED, LIVEACTION_STATUS_FAILED,
-                       LIVEACTION_STATUS_CANCELED]
-LOG = logging.getLogger(__name__)
 
 EXPORTER_WORK_Q = execution.get_queue(
     'st2.exporter.work', routing_key=publishers.UPDATE_RK)
+LOG = logging.getLogger(__name__)
 
 
-class ExecutionsExporter(consumers.MessageHandler):
-    message_type = ActionExecutionDB
+class Exporter(object):
+    def __init__(self):
+        self._model_exporters = []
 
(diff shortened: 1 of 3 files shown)


st2exporter/st2exporter/worker.py
```diff
@@ -13,118 +13,65 @@
...
-        LOG.info('Querying for executions with filters: %s', filters)
-        return ActionExecution.query(**filters)
-
-    def _get_all_executions_from_db(self):
-        return ActionExecution.get_all()  # XXX: Paginated call.
+    def wait(self):
+        for model_exporter in self._model_exporters:
+            model_exporter.wait()
+            LOG.info('Model exporter %s quit.', model_exporter)
+            # XXX: We should probably raising an exception at this point.
```
manasdk: > I have one question on that PR.
I guess if you do not expect an exporter to every quit then yeah throwing an execption is a good idea. But this means that the other model_exporter will blow up also. Btw, what will actually catch this exception and what would the handling be?

st2exporter/st2exporter/worker.py
```diff
@@ -13,118 +13,65 @@
...
-LOG = logging.getLogger(__name__)
 
 EXPORTER_WORK_Q = execution.get_queue(
     'st2.exporter.work', routing_key=publishers.UPDATE_RK)
+LOG = logging.getLogger(__name__)
+
 
+class Exporter(object):
+    def __init__(self):
+        self._model_exporters = []
```
Kami: First I was worried we will potentially have one worker (process) for each model which we want to export, but this looks better :)

st2exporter/st2exporter/worker.py
```diff
@@ -13,118 +13,65 @@
...
-            self._bootstrap()
-        except:
-            LOG.exception('Unable to bootstrap executions from db. Aborting.')
-            raise
-        self._consumer_thread = eventlet.spawn(super(ExecutionsExporter, self).start, wait=True)
-        self._dumper.start()
+        LOG.info('Starting model exporters...')
+        for model_exporter in self._model_exporters:
+            while self._model_exporters_pool.free() <= 0 and not self._stopped:
+                eventlet.sleep(0.1)
```
Kami: Minor thing - it would be nice if the sleep delay is configurable.

st2exporter/st2exporter/worker.py
```diff
@@ -13,118 +13,65 @@
...
-        try:
-            self._bootstrap()
-        except:
-            LOG.exception('Unable to bootstrap executions from db. Aborting.')
-            raise
-        self._consumer_thread = eventlet.spawn(super(ExecutionsExporter, self).start, wait=True)
-        self._dumper.start()
+        LOG.info('Starting model exporters...')
+        for model_exporter in self._model_exporters:
+            while self._model_exporters_pool.free() <= 0 and not self._stopped:
```
Kami: Hm, if I'm reading this correctly (which I might not be since the diff is a bit hard to read), this condition makes no sense.

Exporters are long running "processes" (right?) so in case `free()` returns `< 0` this means we simply miss-configured the pool size at the beginning since the pool should be static (we are not dynamically adding or removing exporters, size == number of exporter workers).

The pool size would only change if an existing exporter dies or is shut down, right?

lakshmi-kannan: Yeah, it is possible that this whole pool.free() check is a bit over engineered. I was trying to protect the case when pool size is not set to len(exporters)

Kami: Yeah, but as mentioned above, this would indicate a programmer error, right?

If so, we should fail fast and throw an exception.

Anyway.

lakshmi-kannan:
We had a discussion whether we want to ship entities to backend yet. Except for metrics like how many rules were created vs deleted, there isn't a lot that we'd get out of shipping entities like rules, actions. We also decided to punt on this in interest of other priorities. So closing this PR. Will open when needed.

PR #1823, opened 2015-08-14
Traces

manasdk:
Trace is a new system entity that track an execution sequence

### TriggerInstance emitted into StackStorm
If a matching `Rule` is found this `TriggerInstance` will lead to an `ActionExecution` therefore the Trace will contain

`TriggerInstance` (original) -> `Rule` -> `ActionExecution` -> `TriggerInstance` (ActionTrigger)

### ActionExecution Requested
No rule ends up being matched but there is still a Trace.

`ActionExecution` -> `TriggerInstance` (ActionTrigger)

### Linking automation on the basis of ActionTrigger
A useful pattern is to use an `ActionTrigger` to initiate next action and so on. In this case all the `Rules` , `ActionExecution` and `TriggerInstance` that form a logical chain/ automation will be part of the same `Trace`.

## trace_id
When a `TriggerInstance` is emitted into the system the Sensor can provide a `TraceContext` which will contain a `trace_id`. This `trace_id` if unique can be used to track the resulting `Trace`.

Similarly, a `TraceContext` and therefore a `trace_id` can be provided given on starting of an execution.

## Design and impl choices
* A `Trace` points to various entities like triggerinstance, rule and actionexecution. These entities do not hold a bi-directional reference to a `Trace`. It is possible to lookup a `Trace` given the id of one of these entities.
* A `Trace` can be looked up by `trace_id`. Since `trace_id` is externally provided StackStorm cannot assume uniqueness therefore StackStorm maintains its own unique id.
* All child executions  of a workflow belong to the parent execution `Trace`.

## Tasks
- [x] DB models, persistence, API models and REST endpoint
- [x] Basic unit tests for all these layers
- [x] Service methods to make interaction simple
- [x] Manual tests from API with and without a sensor

For design and requirements see https://github.com/StackStorm/discussions/issues/89


st2common/st2common/models/db/trace.py
```diff
@@ -0,0 +1,87 @@
...
+    updated_at = ComplexDateTimeField(
+        default=date_utils.get_datetime_utc_now,
+        help_text='The timestamp when the TraceComponent was included.')
+
+    def __str__(self):
+        return 'TraceComponentDB@(object_id:{}, updated_at:{})'.format(
+            self.object_id, self.updated_at)
+
+
+class TraceDB(stormbase.StormFoundationDB):
```
Kami: Trace should ideally be a graph, but looking at this model it doesn't look like we can infer which trigger triggered an action execution since we don't store pointers or similar.

Or am I missing something / not understanding something correctly?

st2common/st2common/services/trace.py
```diff
@@ -0,0 +1,258 @@
...
+    if not trace_context.id_ and not trace_context.trace_id:
+        raise ValueError('Atleast one of id_ or trace_id should be specified.')
+
+    if trace_context.id_:
+        try:
+            return Trace.get_by_id(trace_context.id_)
+        except (ValidationError, ValueError):
+            LOG.warning('Database lookup for Trace with id="%s" failed.',
+                        trace_context.id_, exc_info=True)
+            raise StackStormDBObjectNotFoundError(
```
Kami: Not really a blocker for this PR, but shouldn't `StackStormDBObjectNotFoundError` really be thrown in the lower layer? (IIRC, we also talked about making some of those changes, right?)

st2common/st2common/services/trace.py
```diff
@@ -0,0 +1,258 @@
...
+            return trace_db
+    # 3. Check if the action_execution associated with liveaction leads to a trace_db
+    execution = ActionExecution.get(liveaction__id=str(liveaction.id))
+    if execution:
+        trace_db = get_trace_db_by_action_execution(action_execution=execution)
+    # 4. No trace_db found, therefore create one. This typically happens
+    #    when execution is run by hand.
+    if not trace_db:
+        trace_db = TraceDB(trace_id='execution-%s' % str(liveaction.id))
+    return trace_db
```
Kami: It looks like it would be a good to return a tuple, something like `(created, trace_db)` (similar to how "get_or_create" methods usually work).

st2common/st2common/models/db/trace.py
manasdk: I guess it is a difference of expectation really. If you see the discussion I did not view Trace as a directional graph. However, I think that is the right perception. We have some of that information stored in the ActionExecution still a Trace would be ideal to store that connection information.

Right now I am trying to get out 2 usecases -
1. What are the activities performed when a Trigger was fired into the system using an external identifier?
2. What are the logical group of connected activities?

What I have provides for those cases and connections would be next on this list for sure. (How are these activities linked?) Having said that I think that will be the next level of feature set for Traces. 
The TraceComponentDB can easily be extended to provide this information therefore I am not too worried about making the change in a subsequent PR.

st2common/st2common/services/trace.py
manasdk: Yeah I agree it should be thrown in a lower layer. I do not want to make that change in this PR - too many regression risks.

st2common/st2common/models/db/trace.py
Kami: Alright, thanks for the clarification.

I'm totally fine with starting like that, I just wanted to understand the use case better.

manasdk: Cool. 

Also checkout https://github.com/StackStorm/discussions/issues/89

st2api/st2api/controllers/v1/traces.py
```diff
@@ -0,0 +1,37 @@
...
+    'TracesController'
+]
+
+
+class TracesController(ResourceController):
+    model = TraceAPI
+    access = Trace
+    supported_filters = {
+        'trace_id': 'trace_id',
+        'action_execution': 'action_executions.object_id',
```
lakshmi-kannan: Let's just use execution on the LHS.

st2api/tests/unit/controllers/v1/test_traces.py
```diff
@@ -0,0 +1,89 @@
...
+
+        self.assertEqual(resp.json[0]['trace_id'], self.trace1['trace_id'],
+                         'Correct trace not returned.')
+
+    def test_query_by_action_execution(self):
+        resp = self.app.get('/v1/traces/?action_execution=action_execution_2')
+
+        self.assertEqual(resp.status_int, 200)
+        self.assertEqual(len(resp.json), 1,
+                         '/v1/traces?action_execution=x did not return correct trace.')
```
lakshmi-kannan: ?execution=

st2api/tests/unit/controllers/v1/test_traces.py
```diff
@@ -0,0 +1,89 @@
...
+        self.assertEqual(len(resp.json), 3, '/v1/traces did not return all traces.')
+
+        retrieved_trace_ids = [trace['trace_id'] for trace in resp.json]
+
+        self.assertEqual(retrieved_trace_ids,
+                         [self.trace1.trace_id, self.trace2.trace_id, self.trace3.trace_id],
+                         'Incorrect traces retrieved.')
+
+    def test_query_by_trace_id(self):
+        resp = self.app.get('/v1/traces/?trace_id=test-trace-1')
```
lakshmi-kannan: This is weird. I'd prefer /v1/traces/trace_id as the API.

st2api/tests/unit/controllers/v1/test_traces.py
```diff
@@ -0,0 +1,89 @@
...
+        self.assertEqual(resp.status_int, 200)
+        self.assertEqual(len(resp.json), 1, '/v1/traces?trace_id=x did not return correct trace.')
+
+        self.assertEqual(resp.json[0]['trace_id'], self.trace1['trace_id'],
+                         'Correct trace not returned.')
+
+    def test_query_by_action_execution(self):
+        resp = self.app.get('/v1/traces/?action_execution=action_execution_2')
```
lakshmi-kannan: I am really not sure what this means? Are you looking for trace where execution id is foo? If so, let's use object id in the example so it's less confusing.

st2common/st2common/services/trace.py
```diff
@@ -0,0 +1,258 @@
...
+    return traces[0]
+
+
+def get_trace_db_by_live_action(liveaction):
+    """
+    Given a liveaction does the best attempt to return a TraceDB.
+    1. From trace_context in liveaction.context
+    2. From parent in liveaction.context
+    3. From action_execution associated with provided liveaction
+    4. Creates a new TraceDB (which calling method is on the hook to persist).
```
lakshmi-kannan: Not a fan of this. Why can't this be done by the caller? if not get_trace_db(liveaction): create_trace()

st2tests/st2tests/fixtures/traces/traces/trace_multiple_components.yaml
```diff
@@ -0,0 +1,25 @@
+---
+trace_id : test-trace-3
+action_executions :
+    - object_id: 'action_execution_1'
```
lakshmi-kannan: Can you please use real examples? This is a constant problem in our code. Fixtures don't represent real models.

st2common/st2common/services/trace.py
```diff
@@ -0,0 +1,258 @@
...
+        # since trace_db is None need to end up with a valid trace_context
+        trace_context = _get_valid_trace_context(trace_context)
+        trace_db = TraceDB(trace_id=trace_context.trace_id)
+    return add_or_update_given_trace_db(trace_db=trace_db,
+                                        action_executions=action_executions,
+                                        rules=rules,
+                                        trigger_instances=trigger_instances)
+
+
+def add_or_update_given_trace_db(trace_db, action_executions=None, rules=None,
```
lakshmi-kannan: One thing it would be good to do is sort the executions in order of start_timestamp. Same for rules I suppose when we have enforcement objects.

lakshmi-kannan:
In general, this is not what I had in my mind in terms of implementation. I thought trace information would be like a context object in each model themselves. One problem with the current solution is we might forget to add things to trace object (programmer error). There are advantages to current solution. Querying basically is looking up one db object. We'll have to evaluate how this works for us in the long run.

st2common/st2common/services/trace.py
manasdk: Simply 'cause I want caller to do less work. If you feel strongly I can change.

st2common/st2common/services/trace.py
manasdk: Yeah I have definitely left that part a little open right now. I was thinking we sort client side since it will be a merge sort anyway.

st2api/tests/unit/controllers/v1/test_traces.py
manasdk: I guess it is now trace_tag. but still a query parameter.

st2common/st2common/services/trace.py
lakshmi-kannan: I definitely feel strongly about this but don't want to block the PR on this.

st2common/st2common/services/trace.py
manasdk: Done a9913df2e27e377225f8502cab34e390049af9c1

st2api/tests/unit/controllers/v1/test_traces.py
manasdk: Why you get confused so easy :P

lakshmi-kannan: Because I am old.

st2api/st2api/controllers/v1/traces.py
manasdk: Done 59c688fbb3b786f8553b890fe909e23de6d4539b

manasdk:
>  I thought trace information would be like a context object in each model themselves

We talked through why this does not really work for our use-cases. Shadowing the trace_tag in the model will have to be done to provide other features but the links will be from TraceDB particularly to be able to query.

st2common/st2common/services/trace.py
manasdk: Alright ... i got 99 problems but this ain't one.

st2api/tests/unit/controllers/v1/test_traces.py
manasdk: Fixed 8795112472974573887252a205e8f90f34d7e074

manasdk:
Took away WIP. 

I have addressed comments other tasks will be covered in independent PRs.

Kami:
I know this has already been merged, but we still need to update changelog and add a documentation section on tracing (what it is, how to use it, etc.).

manasdk:
> but we still need to update changelog and add a documentation section

Should've been clearer when I merged the PR. I merged this into master so @lakshmi-kannan could add some extra features that I had not originally planned. Will be working on docs, examples and changelog today.

PR #1824, opened 2015-08-14
Feedback: Handle sudo in remote script action

lakshmi-kannan:


st2common/st2common/models/system/action.py
```diff
@@ -358,10 +360,6 @@ def __str__(self):
         return ', '.join(str_rep)
 
 
-class ParamikoSSHCommandAction(SSHCommandAction):
```
lakshmi-kannan: This wasn't used at all. So I removed it.

st2common/st2common/models/system/action.py
```diff
@@ -210,6 +186,36 @@ def _get_script_arguments(self, named_args=None, positional_args=None):
...
+            if user and user != LOGGED_USER_USERNAME:
+                # Need to use sudo to run as a different user
+                user = quote_unix(user)
+
+                if script_arguments:
+                    command = quote_unix('%s %s' % (script_path_abs, script_arguments))
+                else:
+                    command = quote_unix(script_path_abs)
+
+                command = 'sudo -E -u %s -- bash -c %s' % (user, command)
```
lakshmi-kannan: This is actually confusing for me. (I just moved the code). Since now we can now connect to boxes as particular user and private key/password, I wonder if this `sudo` is needed here? Unfortunately, this might be needed in Fabric case but in paramiko case, I am thinking this sudo as user isn't needed. Thoughts?

manasdk: I think this is needed for LocalRunner and not necessarily for RemoteRunner.

lakshmi-kannan:
This was a fail PR.

PR #1826, opened 2015-08-14
Fix mongoescape util to handle complex dict

m4dcoder:
Fix mongoescape util to handle escaped chars in nested list of dicts.

### DIFF
--- a/st2common/st2common/util/mongoescape.py
+++ b/st2common/st2common/util/mongoescape.py
@@ -31,25 +31,38 @@ RULE_CRITERIA_UNESCAPE_TRANSLATION = dict(zip(RULE_CRITERIA_ESCAPED,
                                               RULE_CRITERIA_UNESCAPED))
 
 
+def _prep_work_items(d):
+    return [(k, v, d) for k, v in six.iteritems(d)]
+
+
 def _translate_chars(field, translation):
     # Only translate the fields of a dict
     if not isinstance(field, dict):
         return field
-    work_items = [(k, v, field) for k, v in six.iteritems(field)]
+
+    work_items = _prep_work_items(field)
+
     while len(work_items) > 0:
         work_item = work_items.pop(0)
         oldkey = work_item[0]
         value = work_item[1]
         work_field = work_item[2]
         newkey = oldkey
+
         for t_k, t_v in six.iteritems(translation):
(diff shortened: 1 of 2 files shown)


Kami:
Good catch.

So the problem was nested dictionaries weren't escaped correctly, or?

st2common/st2common/util/mongoescape.py
```diff
@@ -31,25 +31,38 @@
...
+
         if newkey != oldkey:
             work_field[newkey] = value
             del work_field[oldkey]
+
         if isinstance(value, dict):
-            nested_work_items = [(k, v, value) for k, v in six.iteritems(value)]
-            work_items.extend(nested_work_items)
+            work_items.extend(_prep_work_items(value))
+        elif isinstance(value, list):
```
Kami: It looks like this function could be made simpler and more robust if we just used recursion.

m4dcoder: Recursion was what I would have chosen to begin with. I didn't know whether the team wants to avoid recursion doing it this way...

m4dcoder:
The problem is that if there are dicts in a list, those weren't handled.  You can login to our build server and look at any action chain result doing remote SSH command/script.

st2common/st2common/util/mongoescape.py
manasdk: I did not particularly enjoy writing this method without recursion. But avoiding recursion was a very conscious and deliberate choice. I've seen stacks blow on large object.

lakshmi-kannan: Heh. Recursion is for interviews. </tweet>

manasdk:
+1

m4dcoder:
So should I change mongoescape translate_chars to do recursion or are we comfy with the current method?

m4dcoder:
I see the comments. NVM. I'll leave the function as is.

Kami:
Just noticed it - can you please also update changelog?

st2common/st2common/util/mongoescape.py
Kami: I know this has already been merged, but to be on the safe side we should do `if isisinstance(item, (tuple, list))` so we also handle other iterable types.

PR #1828, opened 2015-08-15
Fix result for ActionChain actions in Mistral

m4dcoder:
Mistral callback handler removed tasks from the result therefore the WF is unable to access output for the ActionChain actions.


Kami:
LGTM.

I'm just curious what we removed it before aka what was the reasoning behind it.

manasdk:
Yeah. If we are removing this and no test is being impacted I am not sure we can reason very well about regressions.

m4dcoder:
I don't recall why the tasks list was popped. But rethinking about the problem, the results of the action should be preserved and sent back to Mistral.  There's no unit tests for looking into the result being sent back therefore no regression.

PR #1833, opened 2015-08-18
Make sure that the "default" pack which we create during installation has a pack.yaml file

Kami:
This pull request makes sure that the directory for the "default" pack which we create during the installation has a "pack.yaml" file.

If a pack directory doesn't contain pack.yaml file, that directory is not treated as a pack (rightly so since pack.yaml makes a pack). This breaks many things. Among others, it means that the content from this pack won't be loaded when register-content is used.

This issue was originally discovered by @enykeev while trying to use the new "data_files" functionality - he was not able to create new files in the "default" pack because we could not find a corresponding `PackDB` model in the database (as noted above, the directory didn't contain "pack.yaml" which means PackDB model wasn't not created when the register content script was ran).

P.S. Yes you read it correctly - it breaks some things, but not all. The problem is that we are not 100% consistent across the code base and we use slightly different logic in different places. If we were consistent and used the same strict and correct logic everywhere, things would explode and be noticed earlier (which is the right behavior). In any case, baby steps and we are getting better with that :)

### DIFF
--- a/contrib/default/actions/README.md
+++ b/contrib/default/actions/README.md
@@ -1,0 +1,5 @@
+# actions
+
+The actions folder contains action scripts and metadata files. See [Actions](http://docs.stackstorm.com/actions.html)
+and [Workflows](http://docs.stackstorm.com/workflows.html) for specifics on writing actions. Note that the lib sub-folder
+is always available for access for an action script.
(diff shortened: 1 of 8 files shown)


enykeev:
:+1:

manasdk:
:+1:

PR #1834, opened 2015-08-18
linux.wait_for_ssh should throw on timeout

Kami:
Previously, we returned `False` if a timeout was reached which means the action would exit with a zero status code instead of non-zero.

This means that if the action was used inside a workflow, `on-success` would be called instead of `on-failure`.

Inside one of out workflows we were lucky, because "retries * sleep delay" was >= default runner action timeout (600 seconds) so the exception was still thrown, but in this case it was because of the runner action timeout.

### DIFF
--- a/contrib/linux/actions/wait_for_ssh.py
+++ b/contrib/linux/actions/wait_for_ssh.py
@@ -1,9 +1,11 @@
 #!/usr/bin/env python
 
+import time
+
 import paramiko
-import argparse
+
 from st2actions.runners.pythonrunner import Action
-import os, yaml, json, time
+
 
 class BaseAction(Action):
     def run(self, keyfile, username, hostname, ssh_timeout, retries):
@@ -15,7 +17,7 @@ class BaseAction(Action):
             try:
                 client.connect(hostname=hostname, username=username, pkey=key)
                 return True
-            except Exception, e: 
+            except Exception, e:
                 self.logger.info(e)
                 time.sleep(ssh_timeout)
             time.sleep(20)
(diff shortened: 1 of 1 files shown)


contrib/linux/actions/wait_for_ssh.py
```diff
@@ -17,9 +17,9 @@ def run(self, keyfile, username, hostname, ssh_timeout, retries):
             try:
                 client.connect(hostname=hostname, username=username, pkey=key)
                 return True
-            except Exception, e: 
+            except Exception, e:
                 self.logger.info(e)
                 time.sleep(ssh_timeout)
             time.sleep(20)
```
manasdk: Why this double sleep?

contrib/linux/actions/wait_for_ssh.py
```diff
@@ -17,9 +17,9 @@ def run(self, keyfile, username, hostname, ssh_timeout, retries):
             try:
                 client.connect(hostname=hostname, username=username, pkey=key)
                 return True
-            except Exception, e: 
+            except Exception, e:
                 self.logger.info(e)
```
lakshmi-kannan: Can you please change the message to "Attempt: w/e." + str(e)?

manasdk: exc_info =True

Kami:
I didn't want to touch much other functionality, but fine. I will also clean up the logging and stuff.

contrib/linux/actions/wait_for_ssh.py
Kami: No clue, it looks like some old code leftover which shouldn't be there.

I'll remove it.

lakshmi-kannan:
LGTM

PR #1836, opened 2015-08-18
Ansible installer docs

armab:
Initial docs showing how to install StackStorm with Ansible, from [`ansible-st2`](https://github.com/StackStorm/ansible-st2) repo.

English syntax-check is welcome :)

----

The bad thing about `.rst` table, have no idea how to make multi-line `description` column, to avoid scrolling here:

### DIFF
--- a/docs/source/install/ansible.rst
+++ b/docs/source/install/ansible.rst
@@ -37,7 +37,7 @@ Below is the list of variables you can redefine in your playbook to customize st
 | ``st2_version``        | ``stable``      | StackStorm version to install. Latest ``stable``, ``unstable``           |
 |                        |                 | to get automatic updates or pin it to numeric version like ``0.12.1``.   |
 +------------------------+-----------------+--------------------------------------------------------------------------+
-| ``st2_revision``       | ``current``     | :: StackStorm revision to install. ``current`` to get the                |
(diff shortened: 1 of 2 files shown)


lakshmi-kannan:
So is this tested? Better question, how are we going to validate this doesn't break after every release. 

I don't mean to block your PR but we have a policy that anything that goes in the docs needs to be tested before every release and without automation it's hard to scale this.

armab:
I see.

In that case I need further work on improvements to make it more stable, provide automatic tests and fix some existing issues. Everything is part of big checklist: https://github.com/StackStorm/ansible-st2/issues/2 on which I'll work further.
Of course I've tested existing state manually many times.

But as I understand @jfryman wanted to have some initial docs ASAP.
So you guys decide is it mergeable or Ansible installer needs more tests and stability. I'm ok with any decision and anyway will continue to work on `ansible-st2`.

jfryman:
@lakshmi-kannan I did give both @dennybaa and @armab direction to get some docs out there.

Can we exclude Config Management from this requirement for the time being? I want to give visibility to these installation methods, but we have some work to do before they'll be at this higher bar of quality and it's being worked on right now.

In terms of the actual plan, @armab has it outlined in his checklist.

lakshmi-kannan:
Yeah, if you already made a call to include it, I am fine. I just wanted to make sure we are thinking about testing etc. LGTM.

PR #1838, opened 2015-08-19
RFR: Add trace_tag support to st2 run

lakshmi-kannan:
Examples:
https://gist.github.com/lakshmi-kannan/43ad4fec2cc12918571c


manasdk:
+1

PR #1842, opened 2015-08-19
Add a mandatory ST2_AUTH_TOKEN param

jfryman:
This PR adds a mandatory ST2_AUTH_TOKEN attribute to the `self-check` to ensure tests run correctly.


manasdk:
Isn't this an optional value? What if no auth is enabled - then what token?

jfryman:
@manasdk yeah, you're right. :/

Issue #1841, opened 2015-08-19
st2 api and st2 auth exits with an exception


johnugeorge:
when st2ctl start is executed, st2api and st2auth starts and then exits with an exception.

```
19 07:46:39,283 140620771120464 INFO log [-] (PID=2303) ST2 Auth API is serving on http://0.0.0.0:9100.
2015-08-19 07:46:39,314 140620771120464 ERROR log [-] (PID=2303) ST2 Auth API quit due to exception.
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/st2auth/cmd/api.py", line 86, in main
    return _run_server()
  File "/usr/lib/python2.7/dist-packages/st2auth/cmd/api.py", line 75, in _run_server
    wsgi.server(socket, app.setup_app())
  File "/usr/lib/python2.7/dist-packages/st2auth/app.py", line 51, in setup_app
    **app_conf
  File "/usr/local/lib/python2.7/dist-packages/pecan/__init__.py", line 86, in make_app
    app = Pecan(root, **kw)
  File "/usr/local/lib/python2.7/dist-packages/pecan/core.py", line 821, in __init__
    super(Pecan, self).__init__(*args, **kw)
  File "/usr/local/lib/python2.7/dist-packages/pecan/core.py", line 243, in __init__
    root = self.__translate_root__(root)
  File "/usr/local/lib/python2.7/dist-packages/pecan/core.py", line 278, in __translate_root__
    module = __import__(name, fromlist=fromlist)
  File "/usr/lib/python2.7/dist-packages/st2auth/controllers/root.py", line 16, in <module>
    from st2auth.controllers import auth
  File "/usr/lib/python2.7/dist-packages/st2auth/controllers/auth.py", line 34, in <module>
    class TokenController(rest.RestController):
  File "/usr/lib/python2.7/dist-packages/st2auth/controllers/auth.py", line 43, in TokenController
    @jsexpose(body_cls=TokenAPI, status_code=http_client.CREATED)
  File "/usr/lib/python2.7/dist-packages/st2common/models/api/base.py", line 183, in decorate
    pecan_json_decorate(callfunction)
  File "/usr/local/lib/python2.7/dist-packages/pecan/decorators.py", line 89, in decorate
    cfg['argspec'] = getargspec(f)
  File "/usr/local/lib/python2.7/dist-packages/pecan/util.py", line 39, in getargspec
    return getargspec(method)
[...]

PR #1843, opened 2015-08-19
[st2ctl] Start st2web before Mistral

Kami:
This is a simple workaround for a problem described here - https://github.com/StackStorm/st2/issues/1841#issuecomment-132733152

> I am not planning to use mistral and by default, mistral is not installed. But, st2ctl script doesn't start st2web service unless mistral is installed. https://github.com/StackStorm/st2/blob/master/tools/st2ctl#L184 . It fails when mistral service is tried to start and it exits before starting st2web.

The solution is not ideal since "st2ctl start" will still exit with non-zero if Mistral is not installed, but at least st2web will still be started.

Ideal / proper solution would be to detect if Mistral is installed / available and only start it if it is. That solution requires a lot more work though, especially if we want to do it the right way.


manasdk:
+1

lakshmi-kannan:
LGTM

PR #1848, opened 2015-08-19
Trace docs

manasdk:
* Updated docs
* Updated sensor sample
* Updated changelog


manasdk:
Merging. Happy to pick up any comments. Likely the docs are horrible but its a start :)

PR #1849, opened 2015-08-20
Icons for all the packs

emedvedev:


jfryman:
Piling on here a bit, but... @Kami difficult to inject to the README.md script?

Kami:
@jfryman Not really, just need to decide how we want to display it (e.g. do we also want to display it in the table, etc.).

On a related note - I was thinking about it the other day and I couldn't decide if it's better to put it in a sub-directory (e.g. `etc/` or `icons/`).

emedvedev:
I asked @enykeev about this and he said it’s ok to put this in the pack root. I feel the same way, don’t think there should be a folder just for the icon.

Kami:
Alright, I'm fine with that :)

I also like `icon.png` naming convention (kinda like favicon :P). On a related note - we need to add a new section on pack.yaml which also describes conventions for the icon (name, size).

manasdk:
I suspect we will need icon of different sizes therefore having an icon folder etc would help a bit.

Could also support a section in pack.yaml -
icon:
   small: icons/icon-small.png
   medium: icons/icon-medium.png
   large: icons/icon-large.png

If not provided we can switch to convention of placing icon in the root and named icon.png. If we ship now with the convention approach now and tack on the explicit icon location in pack.yaml later that works for me.

enykeev:
If we may need icons in several sizes, I strongly suggest to use SVG instead. A single icon of common size (32 or 64 pixels) are picked for convenience. If we can expect user to make such an an effort as creating multiple icons, we can also assume he'll be able to came up with the vector one.

emedvedev:
I doubt it. Some icons are easily findable in large sizes, but almost none in any of the vector formats. I think I've seen no more than ten out of almost eighty while I was searching.

Kami:
Yeah, I agree with @emedvedev on that. Ideally we could use vector format, but it's usually not that easy to find icons / logos in vector format.

Converting raster image to vector also wouldn't work since a lot of trademarks explicitly forbid "modifications" like that.

emedvedev:
Do we merge this for now? The one for st2contrib got merged already. @kami @enykeev @manasdk

PR #1852, opened 2015-08-20
Quotes parameter values that have multiple lines

m4dcoder:


st2common/st2common/util/action_db.py
```diff
@@ -213,14 +218,14 @@ def get_args(action_parameters, action_db):
     positional_args = []
     positional_args_keys = set()
     for _, arg in six.iteritems(position_args_dict):
-        positional_args.append(str(action_parameters.get(arg)))
+        positional_args.append(_quote_text_lines(str(action_parameters.get(arg))))
```
lakshmi-kannan: You may want to use quote_unix from st2common.util.shell. 

Example:

In [3]: from st2common.util.shell import quote_unix

In [4]: quote_unix('foo\nbar whatevs')
Out[4]: "'foo\nbar whatevs'"

In [5]:

Also, to make sure nothing is broken, run ssh tests from st2tests. 

https://github.com/StackStorm/st2tests/blob/master/packs/tests/actions/test_quickstart_remote_script_actions.yaml

https://github.com/StackStorm/st2tests/blob/master/packs/tests/actions/test_quickstart_local_script_actions.yaml

Kami:
Good catch and yeah what @lakshmi-kannan said.

(@lakshmi-kannan I did ask you about escaping the arguments in the other PR and you said they are escaped somewhere up the chain :P)

lakshmi-kannan:
@Kami This seems like a bug not with new runner but all runners in general. I am still not sure what's happening.

st2common/st2common/util/action_db.py
Kami: Edit: I've looked at the changes again.

I don't think this is the right place to escape the arguments. Action utils are (and should be) unaware of where the action runs and escaping is different on Windows and Linux. This mean escaping should happen in a different (right) layer - ideally this would be SSH runner or system actions used by the SSH runner.

I would personally prefer to do it in the SSH runner (closest to the place where commands are actually executed) and I tried to convince @lakshmi-kannan to do that in the other PR, but he did it in the system models...

m4dcoder:
Replaced by https://github.com/StackStorm/st2/pull/1853

m4dcoder:
Replaced by https://github.com/StackStorm/st2/pull/1853

PR #1853, opened 2015-08-21
Postional arguments need to be quoted

lakshmi-kannan:
st2tests PR: https://github.com/StackStorm/st2tests/pull/26

### DIFF
--- a/st2common/st2common/models/system/action.py
+++ b/st2common/st2common/models/system/action.py
@@ -206,7 +206,8 @@ class ShellScriptAction(ShellCommandAction):
 
         # add the positional args
         if positional_args:
-            command_parts.append(positional_args)
+            pos_args_string = ' '.join(map(quote_unix, positional_args))
+            command_parts.append(pos_args_string)
         return ' '.join(command_parts)
 
 
--- a/st2common/st2common/models/system/action.py
+++ b/st2common/st2common/models/system/action.py
@@ -206,7 +206,8 @@ class ShellScriptAction(ShellCommandAction):
 
         # add the positional args
         if positional_args:
-            pos_args_string = ' '.join(map(quote_unix, positional_args))
+            quoted_pos_args = [quote_unix(pos_arg) for pos_arg in positional_args if pos_arg]
+            pos_args_string = ' '.join(quoted_pos_args)
             command_parts.append(pos_args_string)
         return ' '.join(command_parts)
 
--- a/st2common/st2common/util/action_db.py
+++ b/st2common/st2common/util/action_db.py
@@ -215,7 +215,6 @@ def get_args(action_parameters, action_db):
     for _, arg in six.iteritems(position_args_dict):
         positional_args.append(str(action_parameters.get(arg)))
         positional_args_keys.add(arg)
-    positional_args = ' '.join(positional_args)  # convert to string.
 
     named_args = {}
     for param in action_parameters:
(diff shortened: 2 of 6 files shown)


st2common/st2common/util/action_db.py
```diff
@@ -215,7 +215,6 @@ def get_args(action_parameters, action_db):
     for _, arg in six.iteritems(position_args_dict):
         positional_args.append(str(action_parameters.get(arg)))
         positional_args_keys.add(arg)
-    positional_args = ' '.join(positional_args)  # convert to string.
```
lakshmi-kannan: We were converting pos args to a string here which is not right. We should pass it as a list and let ssh runners deal with it. I am also not happy that this method exists in this file. This is specific to script runners. Anyways, one step at a time.

st2common/tests/unit/test_shell_action_system_model.py
```diff
@@ -158,7 +158,7 @@ def test_command_construction_with_parameters(self):
         kwargs['sudo'] = False
         kwargs['user'] = LOGGED_USER_USERNAME
         kwargs['named_args'] = {'key1': 'value1', 'key2': 'value2'}
-        kwargs['positional_args'] = 'ein zwei drei'
+        kwargs['positional_args'] = ['ein', 'zwei', 'drei']
```
lakshmi-kannan: These list changes are fairly easy to understand why I did it.

st2common/tests/unit/test_paramiko_script_action_model.py
```diff
@@ -37,10 +37,11 @@ def test_get_command_string(self):
                                                    parallel=True,
                                                    sudo=False,
                                                    timeout=60)
-        expected = '/tmp/remote_script.sh song=\'b s\' "taylor swift"'
+        expected = '/tmp/remote_script.sh song=\'b s\' \'taylor swift\''
```
lakshmi-kannan: This seems fine.

st2common/tests/unit/test_paramiko_script_action_model.py
```diff
@@ -37,10 +37,11 @@ def test_get_command_string(self):
...
         self.assertEqual(script_action.get_full_command_string(), expected)
 
         # Test with sudo
         script_action.sudo = True
-        ex = 'sudo -E -- bash -c \'/tmp/remote_script.sh song=\'"\'"\'b s\'"\'"\' "taylor swift"\''
+        ex = 'sudo -E -- bash -c ' + \
+             '\'/tmp/remote_script.sh song=\'"\'"\'b s\'"\'"\' \'"\'"\'taylor swift\'"\'"\'\''
```
lakshmi-kannan: Kinda skeptic whether this is fine. We need to write a sudo action that actually takes params with spaces (both positional and named args) and validate that args are passed in correctly. I'd be happy if we added this test in st2tests. cc: m4dcoder.

st2common/st2common/util/action_db.py
Kami: Agreed.

st2common/st2common/models/system/action.py
```diff
@@ -206,7 +206,8 @@ def _get_script_arguments(self, named_args=None, positional_args=None):
 
         # add the positional args
         if positional_args:
-            command_parts.append(positional_args)
+            pos_args_string = ' '.join(map(quote_unix, positional_args))
```
Kami: Any particular reason why you used map over a list comprehension?

In any case, LGTM.

lakshmi-kannan: I should really use list comprehension. I'll make that change.

st2common/tests/unit/test_paramiko_script_action_model.py
lakshmi-kannan: Verified behavior with st2tests PR.

m4dcoder:
LGTM. Thank You!

PR #1854, opened 2015-08-21
Fix `packs/views/file` endpoint throwing error on binary data

enykeev:
Had to fork pecan and manually cherry pick the change from later version.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@ kombu
 mongoengine>=0.8.7,<0.9
 oslo.config>=1.12.1,<1.13
 paramiko
-pecan==0.7.0
+-e git+https://github.com/StackStorm/pecan.git@st2-patched#egg=pecan
 pymongo<3.0
 python-dateutil
 python-json-logger


lakshmi-kannan:
LGTM. This seems the easiest way to unblock us.

Kami:
Yeah, I'm not the biggest fan of that, but what lakshmi said. Hopefully we can soon get rid of the fork though.

On a related not (minor naming thing) - for consistency with fabric and other forks we should probably name the branch "st2-patched".

dzimine:
Bug in JIRA to track this is needed.

enykeev:
> for consistency with fabric and other forks

```...
git+https://github.com/StackStorm/fabric.git@stanley-patched
...
-e git+https://github.com/Kami/logshipper.git@stackstorm_patched#egg=logshipper
...```

Sure =)

@lakshmi-kannan can we fit it in this release? I'm not merging it because you've asked, but it would be great if you can merge it after you're done with the rest.

lakshmi-kannan:
I don't see a problem with merging it for 0.13.

PR #1855, opened 2015-08-21
Returning 404 rather than failing on a non-existent file

emedvedev:
I'll need this for the icons and overall makes sense.

### DIFF
--- a/st2api/st2api/controllers/v1/packviews.py
+++ b/st2api/st2api/controllers/v1/packviews.py
@@ -99,7 +99,7 @@ class FileController(BaseFileController):
         pack_db = self._get_by_name_or_id(name_or_id=name_or_id)
 
         if not pack_db:
-            raise StackStormDBObjectNotFoundError('Pack with name_or_id "%s" doesn't exist' % (name_or_id))
+            raise StackStormDBObjectNotFoundError('Pack with name_or_id "%s" does not exist' % (name_or_id))
 
         if not file_path_components:
             raise ValueError('Missing file path')
--- a/st2api/st2api/controllers/v1/packviews.py
+++ b/st2api/st2api/controllers/v1/packviews.py
@@ -14,9 +14,10 @@
 # limitations under the License.
 
 import os
+import mimetypes
 
 import six
-from pecan import abort
+from pecan import abort, expose, response
 from pecan.rest import RestController
 
 from st2api.controllers import resource
@@ -88,10 +89,10 @@ class FileController(BaseFileController):
     """
     Controller which allows user to retrieve content of a specific file in a pack.
     """
-    @jsexpose(content_type='text/plain', status_code=http_client.OK)
+    @expose()
     def get_one(self, name_or_id, *file_path_components):
         """
-            Outputs the content of all the files inside the pack.
+            Outputs the content of a specific file in a pack.
 
             Handles requests:
                 GET /packs/views/files/<pack_name>/<file path>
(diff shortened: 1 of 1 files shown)


st2api/st2api/controllers/v1/packviews.py
```diff
@@ -98,6 +98,9 @@ def get_one(self, name_or_id, *file_path_components):
         """
         pack_db = self._get_by_name_or_id(name_or_id=name_or_id)
 
+        if not pack_db:
+            return abort(404)
```
Kami: The pattern we tend to use in the API layer is throwing `StackStormDBObjectNotFoundError` exception with an appropriate error message ("Pack with name_or_id foo doesn't exist"). Same goes for other errors (e.g. ValueError will result in 400 response).

This exception is then correctly mapped to 404 inside the JSONErrorResponse hook. Using this approach means error translation is centralized in the hook and it's easy to change the response message or similar (we just need to update one place in the hook).

st2api/st2api/controllers/v1/packviews.py
```diff
@@ -98,6 +98,9 @@ def get_one(self, name_or_id, *file_path_components):
         """
         pack_db = self._get_by_name_or_id(name_or_id=name_or_id)
 
+        if not pack_db:
+            raise StackStormDBObjectNotFoundError('Pack with name_or_id "%s" doesn't exist' % (name_or_id))
+
```
emedvedev: @Kami Thanks for the advice. Correct now?

Kami:
Yep.

LGTM, +1.

emedvedev:
@Kami Had to add headers, too. It just sent 'text/plain' for every file (most browsers can ignore it, but it still presents problems, also it's just plain wrong). Looks good?

enykeev:
I feel that just bluntly abandoning jsexpose might break a few things in most unexpected places. I don't see anything in particular, but would very much like someone from the @StackStorm/team to give it another look

Kami:
Yeah, we do a some parameter type checking and other stuff in jsexpose, but I think we should be fine now since most of the error handling and translation has been moved to the JSONErrorResponseHook.

This means we should still return JSON in case something goes wrong.

st2api/st2api/controllers/v1/packviews.py
```diff
@@ -110,8 +115,10 @@ def get_one(self, name_or_id, *file_path_components):
             # Ignore references to files which don't exist on disk
             raise StackStormDBObjectNotFoundError('File "%s" not found' % (file_path))
 
-        content = self._get_file_content(file_path=normalized_file_path)
-        return content
+        response.headers['Cache-Control'] = 'public, max-age=86400'
```
Kami: I think directly returning new `webob.Response` object is safer than modifying a global variable. That's also what we do in other places.

Besides that, LGTM, +1.

st2api/st2api/controllers/v1/packviews.py
```diff
@@ -110,8 +115,10 @@ def get_one(self, name_or_id, *file_path_components):
...
+        response.headers['Content-Type'] = mimetypes.guess_type(normalized_file_path)[0]
```
Kami: It looks like `mimetypes.guess_type` returns `None` if a content type cannot be detected. It's probably safer if we check for that and return `application/octet-stream` or some sane default value?

emedvedev: `application/octet-stream` is a sensible default. I'll add the check.

st2api/st2api/controllers/v1/packviews.py
emedvedev: I'll look for an example, but `jsexpose` modifies `pecan.response`, too. If it's not the preferred way, maybe we should take a note and refactor it later.

Kami: Alright, in that case it's fine :)

I know most web framework (sadly) rely on thread local / global variables anyway.

PR #1857, opened 2015-08-21
Correctly set "PATH" environment variable for the sandboxed Python runner process

Kami:
Previously, we didn't add `<path to virtualenv>/bin` directory to the "$PATH" environment variable which is set for the sandboxed Python runner process (good catch @manasdk!).

This means that if the Python action tried to access a binary / script which was installed inside a virtual environment bin directory (e.g. ansible Python package installs ansible binary, etc.) or if they wanted to spawn a new python process, a default system version would be used instead of the one which is specific to the virtual environment.

Once this issue is resolved, workaround described at https://github.com/StackStorm/st2contrib/issues/273 won't be needed anymore.

@manasdk and I discussed this change and decided to skip inclusion in v0.13.0. This change needs a bit more testing so we can be sure it doesn't introduce some unexpected regression.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -47,6 +47,8 @@ in development
 * Add X-Request-ID header to all API calls for easier debugging. (improvement)
 * Add new CLI commands for disabling and enabling content pack resources
   (``{sensor,action,rule} {enable, disable} <ref or id>``) (feature)
+* Make sure that the ``$PATH`` environment variable which is set for the sandboxed Python
+  process contains "<virtualenv path>/bin" directory as the first entry. (bug fix)
 
 0.12.2 - August 11, 2015.
 -------------------------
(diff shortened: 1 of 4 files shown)


jfryman:
Very nice. This will also simplify many of the Ansible examples as well.

/cc @armab

manasdk:
+1

Will try out with openstack pack and report back today.

manasdk:
Tested - works.

manasdk:
I think we should include in 0.13

Kami:
@manasdk Cool.

I'm fine with including it in v0.13.

In case something breaks we can always spin up a patch release and fix it. It's also worth nothing that if something does break, it relied on a ninvalid / broken behavior so it's a good idea to fix it anyway.

armab:
One of those geniusly simple solutions, which solves several problems at once :cake:

PR #1859, opened 2015-08-21
Hide liveaction from everywhere as much as possible

manasdk:
* using execution_id is the right behavior


Kami:
LGTM, +1

lakshmi-kannan:
I see WIP. So you are most probably assessing where else we screwed up. LGTM so far.

manasdk:
> So you are most probably assessing where else we screwed up.

Also, some unit tests

Kami:
Yeah, sorry, I originally missed the WIP prefix so take your time.

manasdk:
Taking away the WIP. Some sloppy searches on my part suggest we are ok. There is some basic testing in place to confirm that env vars are in fact passed along, no check for actual content of values.

PR #1865, opened 2015-08-23
Fix error message formatting in paramiko SSH

lakshmi-kannan:


lakshmi-kannan:
cc: @kirilstrax

Kami:
LGTM, +1

st2actions/st2actions/runners/ssh/parallel_ssh.py
```diff
@@ -242,7 +242,7 @@ def _run_command(self, host, cmd, results, timeout=None):
             results[host] = {'stdout': stdout, 'stderr': stderr, 'return_code': exit_code,
                              'succeeded': is_succeeded, 'failed': not is_succeeded}
         except:
-            error = 'Failed executing command %s on host %s', cmd, host
+            error = 'Failed executing command %s on host %s' % (cmd, host)
```
Kami: Also, I assume that this resulted in an exception which was caught in the runner container, right?

lakshmi-kannan: It's for an individual host. You'll see error and traceback in the host's result JSON.

PR #1868, opened 2015-08-23
Docs auto-check broken cross-references.

dzimine:
* This will check for broken cross-references and fail sphinx if one found.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -22,7 +22,6 @@ import itertools
 sys.path.insert(0, os.path.abspath('../../st2common'))
 sys.path.insert(0, os.path.abspath('./_themes'))
 
-print sys.path
 from st2common import __version__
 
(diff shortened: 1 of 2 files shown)


dzimine:
Looking at what to do with this, hints welcome. 
```
Warning, treated as error:
/home/vagrant/st2/docs/source/actions.rst:435: WARNING: py:class reference target not found: st2actions.runners.pythonrunner.Action
```

dzimine:
@Kami pls take a look at ways to not generate warnings by `autodoc`; what I have there is dirty.

Kami:
I had a look at the linking issue and I believe it's a bug in Sphinx.

If i add `:autoclass:` above `:class:` call it works. This most likely indicates a bug in `:class` directive (for some reason module is not imported if autoclass is not used earlier).

Kami:
Also, good call on enabling nitpicky option :+1:

PR #1869, opened 2015-08-24
Update jinja_filters.rst

jonico:
Modified example for regex_replace to show how backreferences can be used.

So far, the documentation did not contain any backreferences, which may led users to the assumption
they cannot be used.


Kami:
Merged into master - thanks!

PR #1870, opened 2015-08-24
Filter PostgreSQL from st2ctl process list.

jfryman:
This commit removes postgresql from the list of processes queried when running `st2ctl status`.


Kami:
LGTM, +1

PR #1871, opened 2015-08-24
Print a better error message if we are unable to read / write auth token from / to the cache

Kami:
The title says it all.

### DIFF
--- a/st2client/st2client/shell.py
+++ b/st2client/st2client/shell.py
@@ -416,14 +416,19 @@ class Shell(object):
 
         if not os.access(ST2_CONFIG_DIRECTORY, os.R_OK):
             # We don't have read access to the file with a cached token
-            LOG.warn('User "%s" doesn\'t have read access to file "%s"' % (os.getlogin(),
-                                                                           CACHED_TOKEN_PATH))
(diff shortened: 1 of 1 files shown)


Kami:
/cc @dzimine 

Better?

st2client/st2client/shell.py
```diff
@@ -416,14 +416,19 @@ def _get_cached_auth_token(self, client, username, password):
 
         if not os.access(ST2_CONFIG_DIRECTORY, os.R_OK):
             # We don't have read access to the file with a cached token
-            LOG.warn('User "%s" doesn\'t have read access to file "%s"' % (os.getlogin(),
-                                                                           CACHED_TOKEN_PATH))
+            message = ('Unable to retrieve cached token from "%s" (user %s doesn\'t have read '
+                       'access to the parent directory). Subsequent requests won\'t use a '
+                       'cached token meaning they will be slower.' % (CACHED_TOKEN_PATH,
```
jfryman: "and may be slower". I mean, it will be because of the round trip, but, we're not talking huge amounts here.

Kami: yep, will make the change :)

lakshmi-kannan:
L.G.T.M

PR #1880, opened 2015-08-25
Pull requirements.txt from specific release

manasdk:
The released version requirements.txt differs from master now therefore the download is also versioned.

PR #1881, opened 2015-08-25
Move auth backends to separate packages and repos

Kami:
This pull request moves authentication backends to "repo and package per backend" model.

Benefit of this approach over the old one is that users can now easily install and enable custom authentication backends (it makes it more modular and pluggable). Previously, they would need to "hack" StackStorm core to use a custom backend, now they simple need create a custom package for their authentication backend, register it using entry_points, install it in the same virtualenv as st2auth is running and configure it in the config.

In addition to that, backend dependencies are now part of the backend package which means that only dependencies for used backend will be installed (e.g. if user doesn't install LDAP backend, ldap dependency won't be installed).

Internally, we now also use stevedore to register and load plugins (eventually we should move most of the dynamic discovery and loading code to this approach so we don't reinvent the wheel a bit differentely in every place).

For backward compatibility reasons, flat file backend will be installed by default (for now, we will install it directly from git) so this change will have no visible impact for our users.

## New backend repos

* https://github.com/StackStorm/st2-auth-backend-flat-file
* https://github.com/StackStorm/st2-auth-backend-mongodb
* https://github.com/StackStorm/st2-auth-backend-keystone
* https://github.com/StackStorm/st2-auth-backend-ldap

### TODO

* [x] Update st2_deploy.sh script for older releases to correctly use requirements from downloads server instead of the ones from git master (https://github.com/StackStorm/st2/pull/1880). If we don't do that, this change will break st2_deploy.sh for all the versions which pull requirements.txt directly from master.


jfryman:
BEST DESCRIPTION EVER

NEVER FORGET!

Kami:
@jfryman lol, yeah, that time when you accidentally copy description in the title field :trollface:

Kami:
This PR is now ready to be reviewed. Each backend repo is also hooked up to Travis CI.  If there are no blockers I will go ahead and merge it.

As noted on Slack, the split resulted in some duplicated files in each backend repo, but sadly I couldn't avoid it at this stage. At some point in the future, I will try to come up with a way of reducing duplicates files (e.g. create a new repo with shared configs or similar).

requirements.txt
```diff
@@ -32,3 +33,5 @@ pyinotify>=0.9.5,<=0.10
 # used by nmap actions
 python-nmap>=0.3.4,<0.4
 semver>=2.1.2
+# flat_file auth backend is installed and available by default
+git+https://github.com/StackStorm/st2-auth-backend-flat-file.git@master#egg=st2auth_flat_file_backend
```
manasdk: Hmm, I think we should look at packaging that and publishing to pypi. Is that something you already considered?

st2auth/st2auth/backends/__init__.py
```diff
@@ -38,16 +46,12 @@ def get_backend_instance(name):
...
-    classes = _get_classes_in_module(module=module)
-
     try:
-        cls = [klass for klass in classes if klass.__name__.endswith('AuthenticationBackend')][0]
-    except IndexError:
-        raise ValueError('"%s" backend module doesn\'t export a compatible class' % (name))
+        manager = DriverManager(namespace=BACKENDS_NAMESPACE, name=name,
+                                invoke_on_load=False)
+    except RuntimeError as e:
+        raise ValueError('Invalid authentication backend specified: %s (%s)' %
```
manasdk: Log the original exception. Likely we will loose useful information otherwise.

st2auth/st2auth/backends/__init__.py
```diff
@@ -59,4 +63,6 @@ def get_backend_instance(name):
     else:
         kwargs = {}
 
-    return cls(**kwargs)
+    cls = manager.driver
+    cls_instance = cls(**kwargs)
```
manasdk: This could throw - right?

st2auth/st2auth/backends/__init__.py
```diff
@@ -14,23 +14,31 @@
...
 ]
 
-BACKEND_MODULES = {
-    'flat_file': 'st2auth.backends.flat_file',
-    'mongodb': 'st2auth.backends.mongodb'
-}
+LOG = logging.getLogger(__name__)
 
-VALID_BACKEND_NAMES = BACKEND_MODULES.keys()
+BACKENDS_NAMESPACE = 'st2auth.backends.backend'
```
manasdk: move to some constants file?

manasdk:
+1

This is great!

Please see how the proper original author attribution for keystone backend could be made in the new repo. I think that is critical.

st2auth/st2auth/backends/__init__.py
Kami: Yeah, it's a fatal exception (e.g. you pass invalid option / keyword argument to the backend class) which means it should propagate all the way up.

The error is not the most user-friendly right now, but for more friendly error messages we would really need a (json) schema for each backend configuration which is out of scope now (it's more of a new feature).

Kami:
@manasdk re preserving original commit author - it's a bit hard (I could commit initial changes as the original patch author), but I did include attribution in the commit message (e.g. https://github.com/StackStorm/st2-auth-backend-keystone/commit/247927bf9ccee3ba135cdc2b7997164d3ad376b7).

st2auth/st2auth/backends/__init__.py
Kami: I did include it in the new exception we throw (`str(e)`), but I can also log the original exception using. LOG.exception, but this will result in exception being logged twice.

manasdk:
@Kami http://stackoverflow.com/questions/1365541/how-to-move-files-from-one-git-repo-to-another-not-a-clone-preserving-history?

I can give it a try. 

I don't think commit message is good enough. I would include the attribution in the code file at the very least. Sorry for being an ass about this one - I feel it is very important.

st2auth/st2auth/backends/__init__.py
manasdk: str(e) does not include the original traceback - this is useful to see the source of the exception etc.

I see the value in raising a wrapped exception which adds a better/user appropriate message. This is really a different exception and therefore whenever we employ that pattern I feel we ought to preserve the original traceback and exception in the logs as it helps with debugging.

Kami: That's correct and I agree about preserving the original exception and traceback :)

lakshmi-kannan:
Me likey likey.

Kami:
@manasdk The problem is that files and locations changed so it's not that straightforward.

One think I could do is create multiple commits:

1. Initial skeleton by me
2. Actual backend commit, authored by original backend author
3. More additional changes which are needed by me

st2auth/st2auth/backends/__init__.py
lakshmi-kannan: backends.backend seems less ideal. Why not just st2auth.backends?

docs/source/config/authentication.rst
```diff
@@ -216,14 +259,55 @@ Example ``auth`` section in the |st2| configuration file. ::
...
+  ``uniqueMember=uid=%(username)s,base_dn``)
+
+Example ``auth`` section in the |st2| configuration file.
+
+.. sourcecode:: ini
+
+    [auth]
+    mode = standalone
+    backend = ldap_backend
+    backend_kwargs = {"ldap_server": "ldap://ds.example.com", "base_dn": "ou=people,dc=example,dc=com", "group_dn": "cn=sysadmins,ou=groups,dc=example,dc=com", "scope": "", "use_tls": True}
```
lakshmi-kannan: I am not a big fan. We could have [ldap_auth], [some_backend_auth] sections in config and look for backend_auth section in config.

requirements.txt
Kami: Yep, to begin with we should just use tags + branch imo.

And when we have some more time we can also hook up all the workflows and publish packages to PyPi.

st2auth/st2auth/backends/__init__.py
Kami: It's just a "convention" I followed from the stevedore documentation examples. it's a backend which belongs to "st2auth.backends" package, hence st2auth.backends.backend.

In any case, it's not a big deal so either st2auth.backend or st2auth.backends works.

docs/source/config/authentication.rst
Kami: You are a bit late to the game. The way config options are specified didn't change :P

We can tackle this later, if needed.

PR #1886, opened 2015-08-26
[CLI] Update CLI to exit with non-zero if the action fails

Kami:
This pull request updates `st2 run`, `st2 execution run` and `st2 execution re-run` commands to exit with a non-zero status code when running in a non-async mode if the action failed.

I believe not exiting with non-zero code also masked the pecan regression introduced in v0.13.0. We have tests in `st2tests` for `packs.install`, but `st2 run` exited with zero so we didn't detect this issue.

@enykeev had some concerns that there might be a reason why we don't already do that. 

If there is indeed a good reason for this not being a default behavior, we can introduce a new flag (e.g. ``--exit-on-failure`` or similar), but I'm not too big of a fan of that since it's not consistent with other operations (e.g. get in-existent action).

Another concern was how to differentiate between CLI errors and API errors. We already use the same behavior in other places so I don't think this is an issue (e.g. we exit with non-zero if API returns 404 object non found) - we just need to make sure we use different / appropriate exit codes to differentaite between those error conditions.

### DIFF
--- a/st2actions/st2actions/runners/actionchainrunner.py
+++ b/st2actions/st2actions/runners/actionchainrunner.py
@@ -374,7 +374,10 @@ class ActionChainRunner(ActionRunner):
             liveaction, _ = action_service.request(liveaction)
         except:
             liveaction.status = LIVEACTION_STATUS_FAILED
-            raise Exception('Failed to schedule liveaction.')
+
+            message = 'Failed to schedule liveaction.'
+            LOG.exception(message)
+            raise Exception(message)
 
         while (wait_for_completion and
                liveaction.status != LIVEACTION_STATUS_SUCCEEDED and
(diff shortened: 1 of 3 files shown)


st2actions/st2actions/runners/actionchainrunner.py
```diff
@@ -374,7 +374,10 @@ def _run_action(self, liveaction, wait_for_completion=True):
             liveaction, _ = action_service.request(liveaction)
         except:
             liveaction.status = LIVEACTION_STATUS_FAILED
-            raise Exception('Failed to schedule liveaction.')
+
+            message = 'Failed to schedule liveaction.'
+            LOG.exception(message)
```
manasdk: would be nice if we continued raising the same exception

lakshmi-kannan:
lgtm

st2actions/st2actions/runners/actionchainrunner.py
Kami: Yeah, I was actually thinking about that, but decided not to do it since we also don't do it in other places.

In any case, if you think that's better, I can make the change.

manasdk: well this lead to 2 back to back logs. not a deal breaker. I would prefer not to do that but it is a fairly minor point.

manasdk:
+1

PR #1887, opened 2015-08-26
Update all_in_one.rst

epowell101:
Love it.  Minor nit ... Prequisites:  When I first read this it sounded like I needed a physical server which of course is not the case due to the prepackaged options....  I suggest we add something mentioning that we have at least a VMDK and AMI option now as well here.


jfryman:
:+1:

PR #1889, opened 2015-08-26
Fix mistral examples to be YAQL 1.0 compatible

m4dcoder:


contrib/examples/actions/workflows/mistral-workbook-multiple-subflows.yaml
```diff
@@ -39,38 +35,31 @@ workflows:
...
-            - cpu_cores
-            - memory_mb
-        task-defaults:
-          on-error:
-            - fail
+            - v1
+            - v2
         tasks:
-            task1:
+            t1:
```
dzimine: nit: names like t1 don't look cool in UI, I think giving longer names here will also enforce the naming convention pattern.

PR #1890, opened 2015-08-27
cwd for remote_script should use cwd runner param

lakshmi-kannan:
Noticed this issue in st2build001. To use the new runner in st2build001, we need this patch in 0.13.1.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,7 +5,6 @@ in development
 --------------
 
 * Add YAQL v1.0 support to Mistral. Earlier versions are deprecated. (improvement)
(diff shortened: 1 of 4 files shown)


manasdk:
+1

Can you CP to 0.13.1?

manasdk:
some tests would be nice.

Kami:
Good catch and what @manasdk said.

Besides that, LGTM.

lakshmi-kannan:
@manasdk @Kami Testing this is hard. We'll end up mocking everything. I could add a test in st2tests.

manasdk:
WFM.

PR #1891, opened 2015-08-27
Trace fix

manasdk:
* 	Ability to push an new TraceComponent into a TraceDB list property. Support for pushAll operator as that fits our usage better.

* Change service impl to update without upsert whenever possible

Testing -
* With 5 and 10 actionrunners started off the same workflow with nested workflows multiple time (10 or so). Passed in the same trace_id to the execution in order to guarantee writing to the same trace thus causing contention and previously seen racing behavior. All workflows passed and trace was updated properly.

### DIFF
--- a/st2actions/st2actions/notifier/notifier.py
+++ b/st2actions/st2actions/notifier/notifier.py
@@ -141,7 +141,7 @@ class Notifier(consumers.MessageHandler):
             payload['action_ref'] = liveaction.action
             payload['runner_ref'] = self._get_runner_ref(liveaction.action)
 
-            trace_context = self._get_trace_context(liveaction=liveaction)
+            trace_context = self._get_trace_context(execution_id=execution_id)
 
             failed_routes = []
             for route in routes:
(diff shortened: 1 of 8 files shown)


manasdk:
Ready for review but I still need to stress test the change. Also, change will be added to next patch

st2common/st2common/persistence/trace.py
```diff
@@ -23,3 +23,28 @@ class Trace(Access):
     @classmethod
     def _get_impl(cls):
         return cls.impl
+
+    @classmethod
+    def push_components(cls, instance, action_executions=None, rules=None, trigger_instances=None):
```
Kami: Push it!

:+1:

st2actions/st2actions/notifier/notifier.py
```diff
@@ -141,7 +141,7 @@ def _post_notify_subsection_triggers(self, liveaction=None, execution_id=None,
             payload['action_ref'] = liveaction.action
             payload['runner_ref'] = self._get_runner_ref(liveaction.action)
 
-            trace_context = self._get_trace_context(liveaction=liveaction)
+            trace_context = self._get_trace_context(execution_id=execution_id)
```
Kami: Good.

Kami:
LGTM

st2common/st2common/persistence/trace.py
lakshmi-kannan: :+1:

manasdk:
Ignoring scrutinizer

PR #1896, opened 2015-08-27
Need nay must have a make target to run test without requirements

manasdk:
* This is critical for productivity! Requirements targets are slow.

`make .pytests` - use that to run tests without requirements. 
`make pytests` - use that to run tests with requirements. This is coverage but that ain't hurting anyone.


Kami:
:+1:

PR #1897, opened 2015-08-27
Fix timer trigger regression

manasdk:
* Trigger in timer trigger is a dict so access properties as a dict.
* Update tests to reflect the internals

### DIFF
--- a/st2reactor/st2reactor/timer/base.py
+++ b/st2reactor/st2reactor/timer/base.py
@@ -13,6 +13,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import uuid
+
 from apscheduler.schedulers.background import BlockingScheduler
 from apscheduler.triggers.cron import CronTrigger
 from apscheduler.triggers.date import DateTrigger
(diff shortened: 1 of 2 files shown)


Kami:
This will also go into v0.13.1, right?

manasdk:
> This will also go into v0.13.1, right?
yes

st2reactor/st2reactor/timer/base.py
```diff
@@ -132,8 +134,9 @@ def _emit_trigger_instance(self, trigger):
             'executed_at': str(utc_now),
             'schedule': trigger['parameters'].get('time')
         }
+
         trace_context = TraceContext(trace_tag='%s-%s' % (self._get_trigger_type_name(trigger),
-                                                          trigger.name))
+                                                          trigger.get('name', uuid.uuid4().hex)))
```
Kami: Is it possible to have a trigger without name at this point?

We randomly generate one if one is not provided so it seems like this shouldn't be possible.

Kami:
LGTM

st2reactor/st2reactor/timer/base.py
manasdk: No. not possible. 

This trigger thing is such a disaster that I would hate for it to blow up for dumb things.

PR #1905, opened 2015-08-28
Update mistral branch

manasdk:
Not sure if there is a better way to do update the branch or if this is even required.


m4dcoder:
I think version_ge $VER "0.13.1" would also do.  But LGTM as is.

manasdk:
> I think version_ge $VER "0.13.1" would also do. But LGTM as is.

7c7d652a92b46aab0187c80985ec70e84bc1347d

PR #1906, opened 2015-08-28
Add missing . to fix Makefile target.

manasdk:


lakshmi-kannan:
heh. wow. i am blind.

PR #1909, opened 2015-08-28
Fix linux.diag_loadavg

dzimine:
It refers to `slack` pack not installed by default, thus failing. 

Should redo it with proper notification.


Kami:
LGTM

PR #1910, opened 2015-08-28
Don't explode if user doesn't include "enabled" attribute when creating a rule

Kami:
I personally couldn't reproduce it since we set a default value in the RuleAPI jsonschema, so the default value (`True`) should be used.

In any case I pushed a change so we now follow the same pattern as we do for actions.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,9 @@ in development
   default. (new feature, improvement)
 * New st2auth authentication backend for authenticating against LDAP servers -
   https://github.com/StackStorm/st2-auth-backend-ldap. (new feature)
+* Default to rule being disabled if the user doesn't explicitly specify ``enabled`` attribute when
+  creating a rule via the API or inside the rule metadata file when registering local content
+  (previously it defaulted to enabled).
 
 0.13.0 - August 24, 2015
 ------------------------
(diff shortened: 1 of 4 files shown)


manasdk:
default to false is safer?

Kami:
It depends on what behavior we want.

We default to True for other resources (e.g. action) and we already default to True in jsonschema so if nothing else it's consistent.

manasdk:
I feel it is safer to have rules disabled by default unless explicitly specified since rules lead to executions. Given the "active" nature of rule I am comfortable if we keep this behavior inconsistent with rest of the content.

manasdk:
+1

Kami:
I'm fine with that, but I will just note it as a minor backward incompatible changes in the changelog.

Kami:
Changed the default "enabled" behavior and added a test case for it.

dzimine:
I am late to the party but I am -2 on "disable" by default. 

This is confusing: user creates a rule (min typing, get example from somewhere), st2 rule create myrule.yaml, obviously doesn't read through the returned data, and than wonders why the rule doesn't work. Especially if actions, etc, still work. 

We already have safe-guards of not loading the rules. it may be enough. 

If we want it safe, fine, let's give up "default" and make it required parameter. And just be sure that it's 400.

Kami:
Yeah, I should have waited a bit longer for more feedback.

I personally agree with @dzimine (other safeguards like not registering rules by default should be enough).

I'm also fine with making the attribute required without a default value, but then we should be consistent and do it for all the pack resources, imo.

PR #1911, opened 2015-08-28
Remove a Mistral example that is causing deadlock in DB

m4dcoder:


manasdk:
+1

Is there a mistral bug to track the issue?

m4dcoder:
I have to pinpoint where the actual problem is before I can file a bug at mistral.

manasdk:
> I have to pinpoint where the actual problem is before I can file a bug at mistral.

We have a test i.e. a repro scenario so figuring out the issue is now part of fixing the bug.

PR #1917, opened 2015-08-31
[WIP] Updates to support st2 packaging

Kami:
Another attempt based on #1677 and #1916.

This time more incremental and hopefully less breaking and more smooth.

TODO:

- [ ] Make sure latest version of pip (7.0) is available on all the build servers


Kami:
I finally managed to track down all the CI/CD pipeline issues and everything now passes so I went ahead and merged changes into master.

We still need to be a bit careful when rolling out a new "major" release (0.14.0 / 1.0.0) since some of the stuff has changed, but it should really only affect st2client if anything (I did fix and test all the st2client related issues locally, but it's also possible that our workflows do some weird stuff with it which I didn't catch :P)

PR #1918, opened 2015-09-01
Some cases (e.g. core.st2.IntervalTimer) don't require sending parameters to the action

alfasin:
As for now, in order to create a rule that doesn't send any parameters to the action, we need to specify an empty object as the parameters:

action:
        parameters: {}

### DIFF
--- a/st2common/st2common/models/api/rule.py
+++ b/st2common/st2common/models/api/rule.py
@@ -176,7 +176,7 @@ class RuleAPI(BaseAPI):
         validator.validate_trigger_parameters(trigger_db=trigger_db)
 
         action = ActionExecutionSpecDB(ref=rule.action['ref'],
-                                       parameters=rule.action['parameters'])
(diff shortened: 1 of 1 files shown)


manasdk:
I am fine with this change. The only downside is that when the Rule is retrieved from the API it will come with a default value i.e `{}`. I do not see this as breaking anything so should be ok. 

Can you add a testcase to https://github.com/StackStorm/st2/blob/master/st2api/tests/unit/controllers/v1/test_rules.py#L100

manasdk:
Once the test is included I will merge the PR.

Thanks!

manasdk:
Will pick up tests later.

alfasin:
Sorry for replying only now...
I believe that this scenario is already tested here:
https://github.com/alfasin/st2/blob/master/st2common/tests/unit/test_db.py#L210

Thanks!

PR #1919, opened 2015-09-01
[RFR] Insert with log_error as debug flag

manasdk:
* Implemented an insert method and used that to add Roles to the system to avoid exceptions.
* The insert method is clear but the same could've been achieved by adding the log_as_debug property to add_or_update method. 

There is already an update method but that is different from add_or_update. We may still need to iterate on the 2 methods and see which way to go. I am not motivated to invest more time here as we may be moving away from mongo anyway so why bother.

### DIFF
--- a/st2common/st2common/models/db/__init__.py
+++ b/st2common/st2common/models/db/__init__.py
@@ -142,22 +142,26 @@ class MongoDBAccess(object):
     def aggregate(self, *args, **kwargs):
         return self.model.objects(**kwargs)._collection.aggregate(*args, **kwargs)
 
-    @staticmethod
-    def add_or_update(instance):
+    def insert(self, instance):
+        instance = self.model._qs.insert(instance)
+        return self._undo_dict_field_escape(instance)
+
+    def add_or_update(self, instance):
         instance.save()
+        return self._undo_dict_field_escape(instance)
+
+    def update(self, instance, **kwargs):
+        return instance.update(**kwargs)
+
+    def delete(self, instance):
+        instance.delete()
+
+    def _undo_dict_field_escape(self, instance):
         for attr, field in instance._fields.iteritems():
             if isinstance(field, stormbase.EscapedDictField):
                 value = getattr(instance, attr)
                 setattr(instance, attr, field.to_python(value))
         return instance
-
-    @staticmethod
-    def update(instance, **kwargs):
-        return instance.update(**kwargs)
-
-    @staticmethod
-    def delete(instance):
-        instance.delete()
 
     def _process_null_filters(self, filters):
         result = copy.deepcopy(filters)
--- a/st2common/st2common/models/db/__init__.py
+++ b/st2common/st2common/models/db/__init__.py
@@ -143,7 +143,7 @@ class MongoDBAccess(object):
         return self.model.objects(**kwargs)._collection.aggregate(*args, **kwargs)
 
     def insert(self, instance):
-        instance = self.model._qs.insert(instance)
+        instance = self.model.objects.insert(instance)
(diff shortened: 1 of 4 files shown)


manasdk:
@Kami @lakshmi-kannan Feel free to unleash your rage.

st2common/st2common/persistence/base.py
```diff
@@ -112,12 +112,52 @@ def aggregate(cls, *args, **kwargs):
...
     @classmethod
-    def add_or_update(cls, model_object, publish=True, dispatch_trigger=True):
+    def insert(cls, model_object, publish=True, dispatch_trigger=True,
+               log_not_unique_error_as_debug=False):
+        if model_object.id:
+            raise ValueError('id for object %s was unexpected.' % model_object)
+        try:
+            model_object = cls._get_impl().insert(model_object)
+        except NotUniqueError as e:
+            if log_not_unique_error_as_debug:
```
Kami: Honestly, we could add the same argument to `add_or_update` method (without needing to add insert and update methods) and be done with it :P

st2common/st2common/persistence/base.py
```diff
@@ -112,12 +112,52 @@ def aggregate(cls, *args, **kwargs):
...
+                LOG.debug('Conflict while trying to save in DB.', exc_info=True)
```
Kami: Imo, a better approach would be to not log anything here and only log a a message in the outer most layer. The reason for that is `NotUniqueError` representing a fatal error or not depends on the consumer (code which called this function) and the code which calls it should also be responsible for logging it appropriately.

We already discussed and bike shed this a lot yesterday, but imo, we are not solving the right problem. Solving a right problem in this case would be deciding on consistent conventions for logging exception since the same "problem" exists through the code base.

manasdk: I do agree that our logging pattern is pretty bogus through the code-base. Since we capture exceptions and raise new ones we completely loose context of the original execption and specifically the tarceback - while debugging issue this is like hitting a wall. Anyway, unless we can preserve the original traceback anc full exception content in the new exception I am reluctant to hide this exception completely.

Kami: Yeah, I agree about lack of full traceback being an issue. Having said that, IIRC, if you use traceback module to retrieve the traceback it should also include a stack trace from the previous / underlying exception (unless you explicitly limit it to the last stack frame using limit=1).

st2common/st2common/persistence/base.py
lakshmi-kannan: add_or_update() makes a lazy programmer use the API incorrectly and will get us into errors. The semantics for insert, delete and update are well defined. I'd rather have them separate. I think we have had this discussion multiple times :P.

st2common/st2common/models/db/__init__.py
```diff
@@ -142,23 +142,27 @@ def distinct(self, *args, **kwargs):
     def aggregate(self, *args, **kwargs):
         return self.model.objects(**kwargs)._collection.aggregate(*args, **kwargs)
 
-    @staticmethod
-    def add_or_update(instance):
+    def insert(self, instance):
+        instance = self.model.objects.insert(instance)
+        return self._undo_dict_field_escape(instance)
+
+    def add_or_update(self, instance):
```
lakshmi-kannan: Can we add a deprecated flag if we want people to move away from this API?

st2common/st2common/persistence/base.py
lakshmi-kannan: ^ I am +1 to this idea. Can we open a JIRA and get this done for 1.0?

PR #1920, opened 2015-09-01
RFR: String trailing '\n' char from stdout and stderr

lakshmi-kannan:
Examples:

Before the change

```
/m/s/s/st2 git:master ❯❯❯ st2 run core.remote cmd="whoami" hosts=localhost                                                                                                                                                      ⏎ ✭ ◼
.
id: 55e627b332ed35263f3df84e
status: succeeded
result:
{
    "localhost": {
        "succeeded": true,
        "failed": false,
        "return_code": 0,
        "stderr": "",
        "stdout": "lakshmi
"
    }
}
/m/s/s/st2 git:master ❯❯❯ st2 run core.local whoami                                                                                                                                                                               ✭ ◼
.
id: 55e627bb32ed35263f3df851
status: succeeded
result:
{
    "failed": false,
    "stderr": "",
    "return_code": 0,
    "succeeded": true,
    "stdout": "lakshmi
"
}

```

After the change

```
/m/s/s/st2 git:master ❯❯❯ st2 run core.local whoami                                                                                                                                                                             ✭ ✱ ◼
.
id: 55e627ea32ed3528abe9564a
status: succeeded
result:
{
    "failed": false,
    "stderr": "",
    "return_code": 0,
    "succeeded": true,
    "stdout": "lakshmi"
}
/m/s/s/st2 git:master ❯❯❯ st2 run core.remote cmd=whoami hosts=localhost                                                                                                                                                        ✭ ✱ ◼
.
id: 55e627f932ed3528abe9564d
status: succeeded
result:
{
    "localhost": {
        "succeeded": true,
        "failed": false,
        "return_code": 0,
        "stderr": "",
        "stdout": "lakshmi"
    }
}
/m/s/s/st2 git:master ❯❯❯
```


lakshmi-kannan:
I'll add this to 0.13.2 and create a changelog in that branch.

Kami:
I think we need to consistently and conclusion handle "\n" characters everywhere - we need to find a root cause of where this character is coming from and fix that.

The problem with just always stripping is that it might result in an incorrect / unexpected behavior if user does something like `cmd='echo \n'.

lakshmi-kannan:
> The problem with just always stripping is that it might result in an incorrect / unexpected behavior if user does something like `cmd='echo \n'.

Bash commands insert a \n so prompt comes to a newline. So \n is part of stdout and stderr. If someone uses print as opposed to sys.stream.write, we'd get a '\n'. I had a discussion with manas here and I also had the same concern with rstrip. I opened the PR to get feedback from everyone. One thing, we could do is just strip only one \n from output. If we don't strip \n, we won't be able to do things like {{task_name.stdout.property}} in action chains.

Kami:
@lakshmi-kannan You sure the line break is coming from bash and not echo ("echo -n 'foo' shouldn't produce a new line)?

DoriftoShoes:
I am fine with stripping the \n from stdout and stderr.  As long as we only strip one we shouldn't run in to problems with removing ones that were purposely put in there.  @Kami it depends.  echo "foo" will add a new line echo -e "foo" will not.  There is definitely \n appended to most stuff that is ran in bash though.

lakshmi-kannan:
Just FYI, Fabric was already stripping. https://gist.github.com/lakshmi-kannan/09e5e3ddb655eb8e2b4e

manasdk:
Yeah and core.local does not. Odd we did not notice this earlier.

Kami:
As discussed on Slack, I'm fine with this change if we make sure we only strip it once and we also document this behavior.

st2common/st2common/util/misc.py
```diff
@@ -65,3 +65,21 @@ def compare_path_file_name(file_path_a, file_path_b):
...
+    :param input_str: Input string to be stripped.
+    :type input_str: ``str``
+
+    :rtype: ``str``
+    """
+    if not input_str:
+        return input_str
+
+    if input_str.endswith('\n'):
+        return input_str[0:len(input_str) - 1]
```
manasdk: input_str[:-1]

manasdk:
+1

lakshmi-kannan:
This is picked up in 0.13.2 https://github.com/StackStorm/st2/pull/1921. I'll remove the private_key fix changeling entry from master in this PR.

docs/source/runners.rst
```diff
@@ -16,7 +16,12 @@ Local command runner (local-shell-cmd)
 ---------------------------------------
 
 This is the local runner. This runner executes a Linux command on the same host
-where |st2| components are running.
+where |st2| components are running. `stdout` and `stderr` fields
```
Kami: Minor thing - this is a bit hard to "parse" imo. Maybe something like:

"
`stdout` and `stderr`` attributes in the runner result object have the last line break (`\n`) character removed if present. This is done so you can more easily re-use the result of common commands such as `uptime`, `whoami`, etc. which include a trailing line break in other actions and workflows. If you have an action which requires a trailing line break character to be present, you can add it explicitly to the result, e.g. "echo -e 'test\n'" (this will result into two line break characters and only one of them will be stripped / removed from the result).
"

Kami:
Do we need a changelog entry?

Besides that, LGTM, +1

lakshmi-kannan:
>Do we need a changelog entry?

I picked this up in 0.13.2. Changeling entry will be there.

docs/source/runners.rst
lakshmi-kannan: Thanks for writing this up. I'll use this verbatim :P.

PR #1922, opened 2015-09-02
Rule types and backstop rules

manasdk:
This change introduces a notion of types to `Rules`. For now there are 2 kinds of rules

1. __standard__ : The are the plain old rules that always apply
2. __backstop__ : New type of rule which only apply if no other rule applied for a specific trigger therefore backstop.

The PR also sets up for different types of rules in the future e.g. scheduled rules etc. The general idea is to have rule applicability decided by system or factors outside of a trigger-instance i.e. time of day, whether a specific rule was applicable etc.

To implement `backstop` rule evaluation is carried out in 2 passes in the rules engine. 

Sample `backstop` rules

```
---
  name: "backstop"
  description: "Test rule using an Interval Timer."
  trigger:
    type: "fixtures.test_passive_trigger.dummy"
  criteria: {}
  type:
    ref: backstop
  action:
    ref: "core.local"
    parameters:
      cmd: "echo \"backstop\""
  enabled: true
```

Old rules do not have to change and are defaulted to `standard` rules.


Kami:
Do we plan to merge / include this in v1.0?

I know we said no new features, but this looks like a pretty big change.

manasdk:
Yeah I prepared the PR and have some misgivings about merging to 1.0. Basically, the need come from @DoriftoShoes describing auto-remidiation use cases and expressing how our current system of rules might not be sufficient. I will check with @DoriftoShoes and see if this is a must for 1.0 - depending on that will decide on merge/no-merge.

DoriftoShoes:
I would really like this now.  I feel it is critical to getting us used as the control plane for remediation tasks.  Without this people will still use multiple event handlers and have to jump between systems to determine the route an alert took.  This will be a friction point for adoption in the auto remediation use case.  When did we officially "freeze" features for 1.0?

manasdk:
> When did we officially "freeze" features for 1.0?

Realistically, I don't think that we will ever really "freeze" features for 1.0.  

Since you make a strong case I think we should include this for 1.0.

manasdk:
@Kami SO yes this will be for 1.0

Kami:
@manasdk Alright, in this case we need tests and some docs on what this feature is on how to use it :)

manasdk:
> Alright, in this case we need tests and some docs on what this feature is on how to use it :)

Yes on that today. I might add the testing to st2tests since it is really integration. I have some units here but that is not enough for the feature.

st2api/st2api/controllers/v1/ruletypes.py
```diff
@@ -0,0 +1,77 @@
...
+from st2common.models.api.base import jsexpose
+from st2common.models.api.rule import RuleTypeAPI
+from st2common.persistence.rule import RuleType
+
+http_client = six.moves.http_client
+
+LOG = logging.getLogger(__name__)
+
+
+class RuleTypesController(RestController):
```
lakshmi-kannan: Can it inherit from ContentPackResourceController?

st2common/st2common/models/api/rule.py
```diff
@@ -53,6 +69,49 @@ class ActionSpec(BaseAPI):
...
+        'title': 'RuleType',
+        'description': 'A specific type of rule.',
+        'type': 'object',
+        'properties': {
+            'id': {
+                'description': 'The unique identifier for the action runner.',
+                'type': 'string',
+                'default': None
+            },
+            'name': {
```
lakshmi-kannan: Just checked TriggerType. We have pack and ref too. Can we please use the same idea here? We migrated rule from name to pack + name and it was PITA. If we don't want pack, we should have a really good reason. I don't see one.

st2reactor/st2reactor/bootstrap/ruletypesregistrar.py
```diff
@@ -0,0 +1,80 @@
...
+RULE_TYPES = [
+    {
+        'name': RULE_TYPE_STANDARD,
+        'description': 'standard rule that is always applicable.',
+        'enabled': True,
+        'parameters': {
+        }
+    },
+    {
+        'name': RULE_TYPE_BACKSTOP,
```
lakshmi-kannan: CATCHALL might be straight forward? This sounds like backstab.

st2reactor/st2reactor/rules/filter.py
```diff
@@ -137,7 +137,39 @@ def _render_criteria_pattern(self, criteria_pattern):
         return criteria_pattern
 
 
-class PayloadLookup():
+class SecondPassRuleFilter(RuleFilter):
```
lakshmi-kannan: CatchAllRuleFilter.

st2api/st2api/controllers/v1/ruletypes.py
manasdk: no packs involed so avoided. this is more like runnertype.

st2common/st2common/models/api/rule.py
manasdk: there is no pack involved same as runnertype.

lakshmi-kannan:
Overall, this is what I had in mind. A RuleType and RuleInstance. I think this is going in the right direction but shipping this in 1.0 is a little too ambitious. I want to use this a little more as a st2 user before declaring this publicly available to consume.

st2reactor/st2reactor/bootstrap/ruletypesregistrar.py
manasdk: It is a baseball analogy so used backstop which is sort the right real work example. I am cool with catchall but find it less expressive than backstop and which is why I used this name. But you know democracy and all so I would put it to a vote and see what others prefer.

```
back·stop
ˈbakstäp/
noun
1.
a person or thing placed at the rear of or behind something as a barrier, support, or reinforcement.
"bullets volleyed into the backstop of a flood-control canal"
verb
1.
BASEBALL
act as backstop for.
```

st2reactor/st2reactor/rules/filter.py
manasdk: Nope. I expect that there will be more rules that fall into the 2nd pass set of rules so trying not to keep it sepcific.

manasdk:
> Overall, this is what I had in mind. A RuleType and RuleInstance.

Do you mean Rule -> RuleInstance? If so nope not going to rename all that and particularly since it provides almost no value.

> shipping this in 1.0 is a little too ambitious.

We can dark-ship but I think having the code in without all the docs has value so we can explore what this does for autoremidiation use cases.

lakshmi-kannan:
>Do you mean Rule -> RuleInstance? If so nope not going to rename all that and particularly since it provides almost no value.

Yeah, Rule is fine :P. I tried to sound less confusing and ended up confusing more.

dzimine:
+1 for dark-ship; we need this for internal and demo, and while using it internally will work out the sharp edges.

manasdk:
> some docs on what this feature is on how to use it 

I will be skipping this since we are moving to dark-ship.

st2reactor/st2reactor/bootstrap/ruletypesregistrar.py
DoriftoShoes: +1 for backstop.  I originally thought catchall but backstop sounds cool (and it makes sense).

manasdk:
I am gonna merge this change after travis turns green.

st2common/st2common/models/api/rule.py
Kami: Looks like you follow similar model as `PolicyType` follows.

That's good, Ideally we would eventually try to refactor some of that functionality and conventions in a base class so everyone can follow the same established pattern.

st2common/st2common/models/api/rule.py
```diff
@@ -29,6 +29,22 @@
 import st2common.validators.api.reactor as validator
 
 
+class RuleTypeSpec(BaseAPI):
```
Kami: Minor thing - missing `API` suffix :P

st2common/st2common/models/api/rule.py
```diff
@@ -53,6 +69,49 @@ class ActionSpec(BaseAPI):
...
+        'description': 'A specific type of rule.',
+        'type': 'object',
+        'properties': {
+            'id': {
+                'description': 'The unique identifier for the action runner.',
+                'type': 'string',
+                'default': None
+            },
+            'name': {
+                'description': 'The name of the action runner.',
```
Kami: Descriptions are a bit out of date with this model... :P

st2common/st2common/models/api/rule.py
```diff
@@ -105,6 +164,7 @@ class RuleAPI(BaseAPI):
             'description': {
                 'type': 'string'
             },
+            'type': RuleTypeSpec.schema,
```
Kami: Hm, looks like this `type.ref` attribute is required and we don't specify a default value in ` RuleTypeSpec`.

Shouldn't we specify a default value of `standard` there otherwise the creation will fail if this attribute is not provided?

st2reactor/st2reactor/rules/matcher.py
```diff
@@ -26,9 +27,40 @@ def __init__(self, trigger_instance, trigger, rules):
...
+        second_pass = []
+        for rule in self.rules:
+            if self._is_first_pass_rule(rule):
+                first_pass.append(rule)
+            else:
+                second_pass.append(rule)
+        return first_pass, second_pass
+
+    def _is_first_pass_rule(self, rule):
+        return rule.type['ref'] != RULE_TYPE_BACKSTOP
```
Kami: Won't this fail for all the existing rules in the DB since we only populate `type` attribute if the database object contains this attribute (https://github.com/StackStorm/st2/pull/1922/files#diff-d8db8c8c2b0e88de1bec747fd4b3d0a1R243)?

PR #1924, opened 2015-09-02
Sendmail Action Fixes

jtopjian:
In the send_mail script, BODY was grabbing everything by using $@, including the new content_type parameter. So to fix, the two parameters needed switched.

Also, it looks as though BODY was still being set, although empty, which was throwing off the send_empty_body tests. New tests have been made and verified.

### DIFF
--- a/contrib/core/actions/send_mail/send_mail
+++ b/contrib/core/actions/send_mail/send_mail
@@ -18,9 +18,9 @@ SUBJECT=$1
 shift
 SEND_EMPTY_BODY=$1
 shift
-BODY="$@"
-shift
 CONTENT_TYPE=$1
+shift
+BODY="$@"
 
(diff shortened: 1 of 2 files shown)


Kami:
Thanks.

The change looks good to me, but we should also figure out how to add some tests for it. I will see what I can do with regards to the tests.

dzimine:
I think as there's no existing tests to email anyways, we can take the patch and add tests later, the patch does make thing better. @Kami ?

Kami:
@dzimine Yeah, I'll merge it now, but we should also work on adding some tests ASAP.

Otherwise, how do we really know it works if there are no tests :)

Kami:
@jtopjian Merged - thanks!

manasdk:
Also the reason for so many bugs in this one is likely lack of tests :(. Its fine we shall figure it out.

PR #1928, opened 2015-09-02
No more st2_startup.log

DoriftoShoes:
Redirect to /dev/null instead of /tmp/st2_startup.log

### DIFF
--- a/tools/st2ctl
+++ b/tools/st2ctl
@@ -2,7 +2,7 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


manasdk:
Finally!

+1

manasdk:
Btw, although I +1 and we had a very real issue your original instinct is right and still applies. 

There are exceptions that would show up only in a startup log. Likely we need to take a closer look at the cracks through which some of these exceptions can leak out.

PR #1931, opened 2015-09-03
Fix Travis builds for Mistral

### LINKED ISSUE #1925: Update Mistral script for Travis
https://github.com/StackStorm/st2/blob/master/scripts/travis/setup-mistral.sh needs updating with recent changes.

Because of that it fails Mistral tests in Travis.

Additional notes from @m4dcoder 

> if the commit is on st2 master, then the MISTRAL_STABLE_BRANCH should be master.  otherwise, should match the st2 version number (i.e. st2-0.13.1)

armab:
* Synchronize Mistral Installation script for Travis with recent `st2_deploy` changes.
* PR for st2 [`master`](https://github.com/StackStorm/st2/tree/master) branch now should use  `MISTRAL_STABLE_BRANCH="master"`
* Otherwise use `MISTRAL_STABLE_BRANCH="st2-0.13.0"`

Fixes #1925

### DIFF
--- a/scripts/travis/setup-mistral.sh
+++ b/scripts/travis/setup-mistral.sh
@@ -6,9 +6,39 @@ if [ "$(whoami)" != 'root' ]; then
     exit 2
 fi
 
-MISTRAL_STABLE_BRANCH="master"
+function version_ge() { test "$(echo "$@" | tr " " "\n" | sort -V | tail -n 1)" == "$1"; }
+
+# Determine which mistral version to use
+if [[ "${TRAVIS_BRANCH}" == 'master' ]]; then
+    MISTRAL_STABLE_BRANCH='master'
+elif [[ "${TRAVIS_BRANCH}" =~ ^v[0-9] ]]; then
+    # remove 'v' prefix from version
+    VER=`echo ${TRAVIS_BRANCH} | cut -c2-`
+    if version_ge $VER "0.13"; then
+        MISTRAL_STABLE_BRANCH="st2-0.13.0"
+    elif version_ge $VER "0.9"; then
+        MISTRAL_STABLE_BRANCH="st2-0.9.0"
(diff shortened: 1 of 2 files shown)


scripts/travis/setup-mistral.sh
```diff
@@ -6,8 +6,39 @@ if [ "$(whoami)" != 'root' ]; then
     exit 2
 fi
 
-MISTRAL_STABLE_BRANCH="st2-0.9.0"
+function version_ge() { test "$(echo "$@" | tr " " "\n" | sort -V | tail -n 1)" == "$1"; }
+
+# Determine which mistral version to use
+if [[ "${TRAVIS_BRANCH}" == 'master' ]]; then
```
Kami: This is and will be a pain to maintain in sync with st2_deploy :/

armab: What can I do here?

Refactor `st2_deploy.sh` to re-use ssh functions as mentioned in https://github.com/StackStorm/st2/pull/1778#issue-98418711 ?

If so, isn't `st2_deploy.sh` to be deprecated soon?

armab: I think it's good that we made `mistral` builds optional. So they have just optional informational purpose.

Kami: Realistically speaking - there is not much we can do right now.

Ideally, we could make parts of st2_deploy.sh more re-usable (so we could re-use parts we need here), but I don't think we want to do this right now.

Kami:
Tests pass so I'm going to merge this for now.

Eventually, we should come up with a better and more maintainable plan.

armab:
Yeah, I'm for having at least some information/feedback from such "open for everyone" CI tests, instead of having nothing.
Especially useful for OSS when someone creates PR.

Kami:
Merged, thanks.

PR #1933, opened 2015-09-03
[RFR] View controller for rules

manasdk:
* Adds extra properties for UI display purpose and returns a RuleViewAPI object

**[API] rules/views**
```
$ http localhost:9101/rules/views/default.rule-99
{
    "action": {
        "description": "Action that executes an arbitrary Linux command on the localhost.",
        "parameters": {
            "cmd": "date"
        },
        "ref": "core.local"
    },
    "criteria": {},
    "enabled": true,
    "id": "55e93d6732ed3558db7790df",
    "name": "rule-99",
    "pack": "default",
    "ref": "default.rule-99",
    "tags": [],
    "trigger": {
        "description": "Notification trigger.",
        "parameters": {},
        "ref": "core.st2.generic.notifytrigger",
        "type": "core.st2.generic.notifytrigger"
    },
    "uid": "rule:default:rule-99"
}
```

**[API] rules**
```
$ time http localhost:9101/rules/default.rule-99

{
    "action": {
        "parameters": {
            "cmd": "date"
        },
        "ref": "core.local"
    },
    "criteria": {},
    "enabled": true,
    "id": "55e93d6732ed3558db7790df",
    "name": "rule-99",
    "pack": "default",
    "ref": "default.rule-99",
    "tags": [],
    "trigger": {
        "parameters": {},
        "ref": "core.st2.generic.notifytrigger",
        "type": "core.st2.generic.notifytrigger"
    },
    "uid": "rule:default:rule-99"
}
```

### DIFF
--- a/st2api/st2api/controllers/v1/ruleviews.py
+++ b/st2api/st2api/controllers/v1/ruleviews.py
@@ -1,0 +1,108 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import six
+
+from st2common import log as logging
+from st2api.controllers import resource
+from st2common.models.api.rule import RuleViewAPI
+from st2common.models.api.base import jsexpose
+from st2common.persistence.action import Action
+from st2common.persistence.rule import Rule
+from st2common.persistence.trigger import TriggerType, Trigger
+from st2common.rbac.types import PermissionType
+from st2common.rbac.decorators import request_user_has_permission
+from st2common.rbac.decorators import request_user_has_resource_permission
+
+http_client = six.moves.http_client
+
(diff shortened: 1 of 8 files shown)


Kami:
LGTM.

st2api/st2api/controllers/v1/ruleviews.py
```diff
@@ -0,0 +1,108 @@
...
+
+http_client = six.moves.http_client
+
+LOG = logging.getLogger(__name__)
+
+
+__all__ = ['RuleViewController']
+
+
+class RuleViewController(resource.ContentPackResourceController):
```
Kami: Would be good to add a docstring on how it works / what it returns.

st2api/st2api/controllers/v1/ruleviews.py
```diff
@@ -0,0 +1,108 @@
...
+        for rule in rules:
+            action_refs.add(rule.action['ref'])
+            trigger_refs.add(rule.trigger['ref'])
+            trigger_type_refs.add(rule.trigger['type'])
+
+        action_by_refs = {}
+        trigger_by_refs = {}
+        trigger_type_by_refs = {}
+
+        for ref in action_refs:
```
Kami: That's probably fine for now, but eventually we should optimize it so we do 3 instead of 3 * n queries (retrieve all the objects of a particular type in a single query) :P

manasdk: I could not work out how to do that with mongoengine. Do see some bulk APIs in 3.0. I will take a closer look at mongoengine docs.

st2api/st2api/controllers/v1/ruleviews.py
manasdk: 8425dc08d95c9e86b3aa2179241430caea35bf98

st2api/st2api/controllers/v1/ruleviews.py
manasdk: ed3f147e12ce38d499204874d3ae8b3977c1482e

PR #1934, opened 2015-09-04
Check if st2_deploy.sh is launched from superuser

armab:
Community fix, covers issue when users trying to run `st2_deploy.sh` from user with insufficient rights.


DoriftoShoes:
LGTM.  Merging now.

Kami:
LGTM

PR #1937, opened 2015-09-04
Implement cancellation of mistral workflows

m4dcoder:


st2actions/st2actions/container/base.py
```diff
@@ -160,8 +164,31 @@ def _do_run(self, runner, runnertype_db, action_db, liveaction_db):
...
+            extra = {'runner': runner}
+            LOG.debug('Performing cancel for runner: %s', (runner.runner_id), extra=extra)
+
+            runner.cancel()
+
+            liveaction_db.status = action_constants.LIVEACTION_STATUS_CANCELED
+            liveaction_db.end_timestamp = date_utils.get_datetime_utc_now()
+            liveaction_db.result = {'message': 'Action canceled by user.'}
+            liveaction_db = LiveAction.add_or_update(liveaction_db)
+            executions.update_execution(liveaction_db)
```
lakshmi-kannan: add_or_update might fail and raise an exception in which case execution will still be in running state. We have seen his happen when payload contains non UTF-8 characters and mongo tries to validate them as a valid string. So when that exception happens, we need to handle that and set status to canceled. So have a finally block that sets status to cancel. If that throws, we'll see the logs. You'll need to move the try catch blocks appropriately. 

Related: https://github.com/StackStorm/st2/pull/1767/files Exception you are looking out for is https://github.com/StackStorm/st2/pull/1767/files#diff-ede474c97745f257ac9dfd3414da4caaR128

lakshmi-kannan: So in this case, you are writing the message. So you might be fine. Still remember, add_or_update could throw.

st2actions/st2actions/container/base.py
```diff
@@ -160,8 +164,31 @@ def _do_run(self, runner, runnertype_db, action_db, liveaction_db):
...
+    def _do_cancel(self, runner, runnertype_db, action_db, liveaction_db):
+        try:
+            extra = {'runner': runner}
+            LOG.debug('Performing cancel for runner: %s', (runner.runner_id), extra=extra)
+
+            runner.cancel()
+
+            liveaction_db.status = action_constants.LIVEACTION_STATUS_CANCELED
+            liveaction_db.end_timestamp = date_utils.get_datetime_utc_now()
+            liveaction_db.result = {'message': 'Action canceled by user.'}
```
lakshmi-kannan: Can you append the actual user as well?

st2actions/st2actions/runners/mistral/v2.py
```diff
@@ -245,6 +245,16 @@ def run(self, action_parameters):
...
 
+    def cancel(self):
+        mistral_ctx = self.context.get('mistral', dict())
+
+        if not mistral_ctx.get('execution_id'):
+            raise Exception('Unable to cancel because mistral execution_id is missing.')
+
+        # There is no cancellation state in Mistral. Pause the workflow so
+        # actions that are still executing can gracefully reach completion.
+        self._client.executions.update(mistral_ctx.get('execution_id'), 'PAUSED')
```
lakshmi-kannan: So if this is the case, I am not sure what the value of CANCEL is? It will still run to completion, right? In action chain, the ongoing execution is allowed to complete but the next tasks are not picked up.

m4dcoder: The WF will finish whatever tasks that is currently running and then halt. No more tasks will be scheduled for this WF.

lakshmi-kannan:
Overall, lgtm. Will need docs and changelog as well.I'd like to take a look again but don't block on that.

st2actions/st2actions/query/mistral/v2.py
```diff
@@ -44,38 +48,64 @@ def query(self, execution_id, query_context):
...
-            return (LIVEACTION_STATUS_RUNNING, {'error': msg, 'traceback': trace})
+            return (action_constants.LIVEACTION_STATUS_RUNNING, {'error': msg, 'traceback': trace})
         except:
-            LOG.exception('Unable to get workflow results for '
-                          'query_context: %s. Will skip query.', query_context)
+            LOG.exception('[%s] Unable to fetch mistral workflow tasks. %s',
+                          execution_id, query_context)
+            raise
+
+        # Get the liveaction object to compare state.
```
lakshmi-kannan: heh. Please refactor this into own method for easy unit testing.

st2actions/st2actions/runners/mistral/v2.py
lakshmi-kannan: Ahh cool. Can you add that to the comment?

manasdk: Can a WF be restarted from the PAUSED state?

st2common/st2common/constants/action.py
```diff
@@ -57,6 +59,7 @@
     LIVEACTION_STATUS_RUNNING,
     LIVEACTION_STATUS_SUCCEEDED,
     LIVEACTION_STATUS_FAILED,
+    LIVEACTION_STATUS_CANCELING,
```
manasdk: since you added this state I think the is_action_canceled method should become is_action_canceled_or_canceling

st2actions/st2actions/runners/mistral/v2.py
m4dcoder: Mistral can resume a WF from PAUSED or ERROR state. It makes no difference. Ideal would be to add a CANCEL (non-resumable) state in Mistral.

docs/source/mistral.rst
```diff
@@ -127,3 +127,7 @@ There are more workflow examples under :github_st2:`/usr/share/doc/st2/examples
 Check out this step-by-step tutorial on building a workflow in |st2| http://stackstorm.com/2015/07/08/automating-with-mistral-workflow/
 
 And more details on Mistral workflow syntax on https://wiki.openstack.org/wiki/Mistral/DSLv2
+
+Canceling Workflow Execution
+++++++++++++++++++++++++++++
+An execution of a Mistral workflow can be cancelled by running ``st2 execution cancel <execution-id>``. Workflow tasks that are still running will be allowed to complete before |st2| marks the workflow execution as completely cancelled.
```
lakshmi-kannan: "that are still running will be allowed to complete" makes it sound like we have an option to kill ongoing tasks. Maybe something along the lines are "that are still running will not be canceled and will run to completion. Further tasks are not picked for execution..." or something along those lines?

docs/source/_includes/available_permission_types.rst
```diff
@@ -3,53 +3,45 @@
...
-| execution_view   | Ability to view an execution.                                              |
-+------------------+----------------------------------------------------------------------------+
-| execution_re_run | Ability to create a new action.                                            |
-+------------------+----------------------------------------------------------------------------+
-| execution_stop   | Ability to stop (cancel) a running execution.                              |
-+------------------+----------------------------------------------------------------------------+
-| execution_all    | Ability to perform all the supported operations on a particular execution. |
-+------------------+----------------------------------------------------------------------------+
++-----------------+-------------------------------------------------------------------------+
+| Permission name | Description                                                             |
```
lakshmi-kannan: Definitely looks like something happened here. Is this intentional change?

st2common/st2common/services/action.py
```diff
@@ -154,9 +154,35 @@ def update_status(liveaction, new_status, publish=True):
...
+
+    if liveaction.status not in action_constants.CANCELABLE_STATES:
+        raise Exception('Unable to cancel execution because it is already in a completed state.')
+
+    result = {
+        'message': 'Action canceled by user.',
+        'user': requester
+    }
+
+    update_status(liveaction, action_constants.LIVEACTION_STATUS_CANCELING, result=result)
```
lakshmi-kannan: Does the end_timestamp get updated? Please double check.

lakshmi-kannan:
Besides my comments, +1 from me.

docs/source/_includes/available_permission_types.rst
m4dcoder: Nope. Accidentally added this file. It's a result of running "make all".  Annoying...

docs/source/mistral.rst
m4dcoder: Ok. Will change.

st2common/st2common/services/action.py
m4dcoder: The end_timestamp is updated in _do_cancel in st2actions/st2actions/container/base.py when the status is updated to CANCELED.

PR #1953, opened 2015-09-05
Use python2.7 explicitly

DoriftoShoes:


Kami:
We actually default to the Python binary which was used to start the service (action runner) so if the service was started using Python 2.7 it should work just fine.

In any case, if you prefer to be explicit, I'm also fine with that :)

DoriftoShoes:
Hmmm.  I ended up doing this because rhel6 was picking up the system python (2.6) when creating the virtualenvs for the packs.  The shebang in the various scripts is 2.7 so I'm not sure what's going on.  I'll merge this for now and investigate later.

DoriftoShoes:
I found the issue.  It was the way the tests/examples were being installed.  It was using a script instead of the workflow and not specifying the path to the python binary in the virtualenv call.  I submitted another PR to fix it.  I'm going to let this one stay for now as I feel it is a little better to be more explicit on the platforms that don't have py2.7 as the system python.  We can discuss later if we want to undo this.

Kami:
Yeah, when I touched the code to fix the delete venv issue I was thinking about changing it to use `setup_virtualenv` instead of doing it manually, but I decided to leave it as is is because maybe someone had a reason to do it manually (although, there are no comments indicating that).

PR #1954, opened 2015-09-06
Virtualenv2.7

DoriftoShoes:
The test setup script was using the system python to create the virtualenv.  this explicitly uses python2.7.

### DIFF
--- a/tools/st2-setup-examples
+++ b/tools/st2-setup-examples
@@ -2,6 +2,7 @@
 
 echo "Installer for the examples pack..."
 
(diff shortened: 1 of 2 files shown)


Kami:
:+1: 

I assume this won't affect Ubuntu hosts since python binary is probably just a symlink to python2.7, right?

DoriftoShoes:
Correct.  We'll need to revisit this stuff if we move to Python3 but this shouldn't affect anything right now.

PR #1959, opened 2015-09-07
Updates for st2 packaging

dennybaa:
A new portion. Consolidating configs into one directory `/etc/st2`.


Kami:
`st2exporter/conf/logging.conf → st2exporter/conf/logging.exporter.conf`

Do we also need to update `st2exporter.config` to reflect this change?

dennybaa:
Actually it's so https://github.com/StackStorm/st2/blob/master/st2exporter/st2exporter/config.py#L44
it was my previous renaming, but it didn't make sense. So I reverted.

Kami:
LGTM

Kami:
Merged, thanks.

PR #1960, opened 2015-09-07
Move all the resource registrar modules to "st2common" package

### LINKED ISSUE #1958: Need to decouple binaries from st2common component
When packages are shipped separately like st2actions or st2reactor, common binaries will fail to work.
Since in one case it will fail to load sensor specific modules and in the second case actions specific..

http://drone.vos.io/github.com/dennybaa/st2-packages/master/97fd4f2a41777c7517ea201d7a9277f6741ddda4

Details:  https://gist.github.com/dennybaa/9fe10027a8eb167a7e0b

Kami:
This pull request resolves #1958 and moves all the resource registrar modules to st2common package.

This "issue" was discovered by @dennybaa while working on the new packaging changes. The problem is that with the new setup, each component is isolated and running in a separate virtual. This means that unless every server had every package installed (not just st2common + component which is needed), it wouldn't work. User would only be able to register rules on the server where st2reactor is running, etc.

This pull request fixes that - now all the registrar modules are located in st2common package which is by default installed on every server / container since every component depends on it.

The "side affect" of this change is also getting rid of the "nasty" late imports which usually indicate bad module / package organization and code smell.

Note: I have been able to fix it for every registrar except policy types. Sadly policy types are too coupled to st2actions package.


dennybaa:
@Kami thank you. Seems that the fix helped very well http://drone.vos.io/github.com/dennybaa/st2-packages/master/07ad820a9b6777cfe55d07e7f83edc4033a06c88

1. content is registered.
2. test action is fired and succeeded.

LGTM.

manasdk:
Argh. Not a fan but I see the need.

Kami:
@manasdk Care to elaborate why you are not a fan of it?

I think the split is fine since the resource registrar code is not coupled to a particular component and having it in common makes a lot of sense - it's a common functionality.

manasdk:
> it's a common functionality

Yeah likely this is a reasonable way to look at the problem. Since all models and a ton of service classes are in st2common anyway this is prehaps a logical move.

lakshmi-kannan:
Funny. We have gone back and forth on this a bunch of times. I think with our current registration mechanism, this is the way to do it. So I am +1.

manasdk:
+1.. Merge away

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1450729372016887

PR #1961, opened 2015-09-07
Makefiles update: publish st2* wheels into the wheelhouse

dennybaa:
We need local wheels for each of the components so that "st2bundle" can be created.

### DIFF
--- a/st2api/Makefile
+++ b/st2api/Makefile
@@ -59,7 +59,7 @@ all: install
 
 install: wheelhouse changelog
 
-post_install:
+post_install: bdist_wheel
 	# post_install is triggered debian/rules file, don't call it from install target!
 	# We don't want anything like below to be postinst
 	install conf/logging.conf $(DESTDIR)/etc/st2/logging.api.conf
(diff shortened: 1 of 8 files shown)


st2api/Makefile
```diff
@@ -59,7 +59,7 @@ all: install
 
 install: wheelhouse changelog
 
-post_install:
+post_install: bdist_wheel
```
Kami: Does latest version of pip ship with wheel support or we still need to explicitly install the wheel package?

dennybaa: We install pip wheels locally, not because of pip itself. But because of this
https://github.com/dennybaa/st2-packages/blob/st2_bundle/sources/st2actions/debian/rules#L35

In other words, there's bundle requiremets.txt which contain st2common, st2actions since pip can not fetch or wheels from pypi, we use local wheelhouse which is on line #35.

Kami: Sorry, I was talking about the `wheel` Python package - https://pypi.python.org/pypi/wheel.

It used to be that `bdist_wheel` command was only available if `wheel` package was installed locally.

dennybaa: Actually it seem to work good with the latest pip which we use, i.e. I haven't installed `wheel` package anywhere explicitly.
```
ls -la /tmp/wheelhouse | grep -c wheel
0
```
Yeah, It's not needed now.

Kami: Maybe it's installed by default with latest version of setuptools.

Anyways.

Kami:
Merged, thanks.

st2api/Makefile
dennybaa: yeah you are right, it's dependent somehow, just  made `pip list` >>> wheel (0.24.0)
it's there

PR #1963, opened 2015-09-08
Change `packs.install` meta action to require pack

jfryman:
This commit changes the `packs.install` action metadata from * to `None` with `required: true`. This forces the user to explicitly download packs they want/need versus accidentally getting everything.

### DIFF
--- a/contrib/packs/actions/install.meta.yaml
+++ b/contrib/packs/actions/install.meta.yaml
@@ -11,8 +11,7 @@
       type: "array"
       items:
         type: "string"
-      default:
-        - "*"
+      required: true
     repo_url:
       type: "string"
       default: "StackStorm/st2contrib"


jfryman:
I'm dropping this here for comment. I never understood why `*` was a default.

Kami:
I'm fine with that change.

At some point we probably had some kind of *maybe good) reason to default to installing all packs.

lakshmi-kannan:
`*` was a mandate from @dzimine at some point. With the number of packs we have in st2contrib, * is an insane default. Aren't there supposed to related docs changes?

manasdk:
What if a user supplied `*` as the value - is that ok?

Yeah at a point when we did not have many pack `*` was a sane default. Not anymore so I am +1 to changing behavior.

jfryman:
Oh yeah. If user wants to supply wildcard, go crazy. But, dangerous default.

Kami:
> * was a mandate from dzimine at some point. With the number of packs we have in st2contrib, * is an insane default. Aren't there supposed to related docs changes?

Heh, I thought something like that was the case :)

PR #1964, opened 2015-09-09
[RFR] Api keys

manasdk:
CLI creation etc -

* __Creation__
```
$ st2 apikey create -m '{"used_by": "webhook1"}'
+----------+--------------------------------------------------------------+
| Property | Value                                                        |
+----------+--------------------------------------------------------------+
| id       | 55f27e8232ed35674529da72                                     |
| key      | MGY0ZDE3MDdjYTMzZmUxZTU2NDI3OTE1MWNlYTc4NWZiODIxZWI4Njc1ZWEw |
|          | ODI0ODA0MGM3YTI5NTlhNmI4Nw                                   |
| metadata | {                                                            |
|          |     "used_by": "webhook1"                                    |
|          | }                                                            |
| user     | manas                                                        |
+----------+--------------------------------------------------------------+
```

* __Listing__
```
$ st2 apikey list
+--------------------------+-------+---------------------------+
| id                       | user  | metadata                  |
+--------------------------+-------+---------------------------+
| 55f27d2532ed35617d31c725 | manas | {}                        |
| 55f27e8232ed35674529da72 | manas | {u'used_by': u'webhook1'} |
+--------------------------+-------+---------------------------+
```

* __Get by ID__
```
$ st2 apikey get 55f27e8232ed35674529da72
+----------+---------------------------+
| Property | Value                     |
+----------+---------------------------+
| id       | 55f27e8232ed35674529da72  |
| user     | manas                     |
| metadata | {                         |
|          |     "used_by": "webhook1" |
|          | }                         |
| key      | ********                  |
+----------+---------------------------+
```

* __Get by API key__
```
$ st2 apikey get MGY0ZDE3MDdjYTMzZmUxZTU2NDI3OTE1MWNlYTc4NWZiODIxZWI4Njc1ZWEwODI0ODA0MGM3YTI5NTlhNmI4Nw
+----------+---------------------------+
| Property | Value                     |
+----------+---------------------------+
| id       | 55f27e8232ed35674529da72  |
| user     | manas                     |
| metadata | {                         |
|          |     "used_by": "webhook1" |
|          | }                         |
| key      | ********                  |
+----------+---------------------------+
```

* __delete__
```
$ st2 apikey delete 55f27e8232ed35674529da72
Resource with id "55f27e8232ed35674529da72" has been successfully deleted.
```

* __Using in API__
```
$ http :9101/v1/apikeys St2-Api-Key:<MASKED>
HTTP/1.1 200 OK
Access-Control-Allow-Headers: Content-Type,Authorization,X-Auth-Token,X-Request-ID
Access-Control-Allow-Origin: http://127.0.0.1:9101/
Connection: keep-alive
Content-Length: 176
Date: Fri, 11 Sep 2015 08:27:32 GMT
X-Request-Id: e28bb1a6-b07d-4532-8699-b745fcce8159

[
    {
        "id": "55f28e5532ed3569c8eea5b1",
        "key": "********",
        "metadata": {},
        "user": "testu"
    },
    {
        "id": "55f28edc32ed3569c8eea5b2",
        "key": "********",
        "metadata": {},
        "user": "testu"
    }
]
```

TODO
---------
- [x] docs
- [x] usage in API as a token
- [x] tests for API usage
- [x] CHANGELOG
- [x] rbac resolver


st2api/st2api/controllers/v1/auth.py
```diff
@@ -0,0 +1,156 @@
...
+            LOG.exception(msg)
+            abort(http_client.NOT_FOUND, msg)
+
+        try:
+            return ApiKeyAPI.from_model(api_key_db)
+        except (ValidationError, ValueError) as e:
+            LOG.exception('Failed to serialize API key.')
+            abort(http_client.INTERNAL_SERVER_ERROR, str(e))
+
+    @request_user_has_permission(permission_type=PermissionType.API_KEY_VIEW)
```
jfryman: I'm curious, what is your take on the approach of "you only get to see it once at generation" as opposed to having a permission to view the key. Decrease the surface attack area.

manasdk: The RBAC scoping for this is still not sorted so I have gone with a slightly relaxed primitive i.e. you get to view if you have permission on the object. This means an admin can view keys for every user etc.

I suppose this is "sorta" like an admin being able to see another users password.

I do like the you only get to see it once at generation idea and did not really consider the approach. Whenever I've used API keys I often tend to go back to the application to grab the old keys so it would make me slightly annoyed to generate a new one. Having said that I have no strong opinion about this behavior and am happy to experiment.

jfryman: For something that is the control plane, I'd rather err on the side of caution and not allow ourselves to leak keys if we can help it. If a user leaks keys, that's on them. I think it's also a pretty accepted pattern with other IAM providers as well, so I don't think we'll get too much user revolt.

Kami: I agree with @jfryman on this one - I think it would be good if user can only see it in plain text when we generate it.

This means we need to only store a hash in the database (salted of course) and we also need to make sure that value doesn't leak in the logs.

manasdk: My experience is with the likes of google, twitter, slack etc where I can always go back to keys. Our users will probably prefer the view once approach if others are already using that approach. I will not immediately execute on this; will figure it out in this PR but after I get the validation part working. This also gives other some time to chime-in and voice opinion on this stuff :)

btw, a related question is do you feel we need to keep a one way hash of the key in the DB? Treat it like a password. This way anyone with access to the DB also cannot really make sense of the key or use it to abuse. It may be related to your concerns.

Kami: I think those things are closely related - if we want to return plain version of the API keys back to the user at some later point in time, this means we either need to also store plain-text version or an encrypted version of it.

We don't have a code infrastructure for doing safe encryption and decryption in place right now so I think the best and safest approach to go with is only returning plain-text version on generating and only storing a hash.

manasdk: > so I think the best and safest approach to go with is only returning plain-text version on generating and only storing a hash.

Makes sense.

 Btw, how about if we ship this under experimantal for 1.0? I suspect we are not really ready to nail some of these behaviors down yet.

jfryman: +1

lakshmi-kannan: Yeah, I am +1 to whatever cloud companies do. They have possibly thought through this a lot. Also, just showing the API key one time is freaking annoying but whatever makes the backend easier :P.

dzimine: Hmm. 
* as administrator, I want to be able to list and delete all API keys (even here, list is questionable, may be users don't like admin seeing and potentially using their API keys).

* as user, I only want to see my API keys (or at least my admin may not appreciate me wanting to see ALL api keys :)

How does this work here?

manasdk: I did not find the right RBAC primitives for this support just yet. I do agree that the API key should only be visible to the user that created. Likely we can tackle this part later since this we would have to introduce some dynamic permissions in such a case. I know we can design and implement this or it might be already in place. Still feel this is out of 1.0 scope.

For now we are only going to show that a key exists as opposed to showing the value of the key therefore there will be no way to use another users key. In fact if a user loses a key then they will have to create a new one and delete the old key. We will go as far as creating a one-way hash to make it impossible to retrieve the original key to make this somewhat secure.

Admin can delete anyones key.

dzimine:
The user flow here seems like: 
1) user authenticates and requests an API key, if has permissions to do so. 
2) with the API key, user can ... (looks like: "authenticate as the "user"; and apply the same RBAC that he has? 
3) user can view or delete his keys (+1 for hash idea but don't have knowledge based opinion here, just follow what Github is doing is fine IMO)
4) admin can view all the keys, see which user they belong to, and delete them at will. 
5) there's no expiration on API keys.

With this in mind, questions: 
1) check that non-priviledged user can't see all the keys (note to @Kami: we probably should separate superuser and admin as out-of-box roles, default admin is getting dangerous).
2) need CLI :)
3) what do I do if I only want to give access to webhooks, or to a webhook? Is it much better from creating webhook_user with given username/password? @DoriftoShoes what do you think?

manasdk:
> The user flow here seems like: 

Yes. That is the flow. Still working on realizing this so it is a WIP.

>With this in mind, questions: 
>1) check that non-priviledged user can't see all the keys (note to @Kami: we probably should separate superuser and admin as out-of-box roles, default admin is getting dangerous).

RBAC permissions should apply out of the box. For now user with ability to view keys __can see all keys__

> 2) need CLI :)

In the works.

> 3) what do I do if I only want to give access to webhooks, or to a webhook? Is it much better from creating webhook_user with given username/password? @DoriftoShoes what do you think?

Create an API key and "assign" to webhook(s). 

It is certainly much better than `webhook_user` -
1. not all webhooks will support basic_auth or some form of supplemental headers.
2. spreading plain-text passwords around is pretty risky and definitely less secure. Plus we take up some of the management making it simpler for users of our system in general.

st2common/st2common/util/hash.py
```diff
@@ -0,0 +1,26 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import hashlib
+
+
+FIXED_SALT = 'saltnpeper'
```
manasdk: This constant hash is pointless. I guess we don't really need a salt and simply saving the hash would be sufficient anyway.

Kami: It's not totally pointless as long as it's sufficiently random since it still prevents rainbow table attacks.

Kami: It's also worth noting that changing the salt will "break" the existing hashes in the DB so something we need to keep in mind.

manasdk:
Code mostly done. Still working on TODOs so left in WIP but it is definitely ready for review.

st2api/st2api/controllers/v1/auth.py
manasdk: @jfryman @Kami @lakshmi-kannan 

Storing hashes of the key and masking the has from API in get calls. The API key is only returned for create.

st2api/st2api/controllers/v1/auth.py
```diff
@@ -0,0 +1,171 @@
...
+        'sort': ['user']
+    }
+
+    def __init__(self):
+        super(ApiKeyController, self).__init__()
+        self.get_one_db_method = ApiKey.get_by_key_or_id
+
+    @request_user_has_resource_permission(permission_type=PermissionType.API_KEY_VIEW)
+    @jsexpose(arg_types=[str])
+    def get_one(self, api_key_id_or_key):
```
Kami: Hm, slightly confused - we still allow keys to be retrieved using the actual key?

Shouldn't we only support id now since hashed key makes little sense?

st2api/st2api/controllers/v1/auth.py
```diff
@@ -0,0 +1,171 @@
...
+    def delete(self, api_key_id_or_key):
+        """
+            Delete the key value pair.
+
+            Handles requests:
+                DELETE /apikeys/1
+        """
+        api_key_db = None
+        try:
+            api_key_db = ApiKey.get_by_key_or_id(api_key_id_or_key)
```
Kami: You can let "object not found" exceptions to propagate.

They are caught in the response hook where they are also translated to the corresponding error message and status code.

st2api/st2api/controllers/v1/auth.py
```diff
@@ -0,0 +1,171 @@
...
+            LOG.exception(msg)
+            abort(http_client.NOT_FOUND, msg)
+
+        LOG.debug('DELETE /apikeys/ lookup with api_key_id_or_key=%s found object: %s',
+                  api_key_id_or_key, api_key_db)
+
+        try:
+            ApiKey.delete(api_key_db)
+        except Exception as e:
+            LOG.exception('Database delete encountered exception during '
```
Kami: Same here, this is also handled in the response hook.

st2api/st2api/controllers/v1/auth.py
```diff
@@ -0,0 +1,171 @@
...
+
+        try:
+            return ApiKeyAPI.from_model(api_key_db, mask_secrets=True)
+        except (ValidationError, ValueError) as e:
+            LOG.exception('Failed to serialize API key.')
+            abort(http_client.INTERNAL_SERVER_ERROR, str(e))
+
+    @request_user_has_permission(permission_type=PermissionType.API_KEY_VIEW)
+    @jsexpose(arg_types=[str])
+    def get_all(self, **kw):
```
Kami: I think that for list and get one response it makes sense to return the following attributes:

* id
* description
* created at
* enabled

I know that  some of those might not exist yet, but they would come handy.

st2common/st2common/models/db/auth.py
```diff
@@ -49,4 +53,32 @@ class TokenDB(stormbase.StormFoundationDB):
...
 
 
-MODELS = [UserDB, TokenDB]
+class ApiKeyDB(stormbase.StormFoundationDB, stormbase.UIDFieldMixin):
+    """
+    """
+    RESOURCE_TYPE = ResourceType.API_KEY
+    UID_FIELDS = ['id']
+
+    user = me.StringField(required=True)
```
Kami: Just to confirm - this is a foreign key to the user (name) the key belongs to, right?

st2common/st2common/hooks.py
```diff
@@ -108,20 +110,11 @@ def before(self, state):
...
-            # this happens
-            user_db = None
-
-        # Store token and related user object in the context
-        # Note: We also store token outside of auth dict for backward compatibility
-        state.request.context['token'] = token_db
+        # Store related user object in the context. The token is not passed
+        # along any longer as that should only be used in the auth domain.
         state.request.context['auth'] = {
-            'token': token_db,
```
Kami: :+1:

st2api/st2api/controllers/v1/auth.py
```diff
@@ -0,0 +1,171 @@
...
+            abort(http_client.INTERNAL_SERVER_ERROR, str(e))
+            return
+
+        extra = {'api_key_db': api_key_db}
+        LOG.audit('ApiKey deleted. ApiKey.id=%s' % (api_key_db.id), extra=extra)
+
+    def _get_user(self, api_key):
+        # If a user is provided in api_key try to use that value. In the future
+        # this behavior should change.
+        user = getattr(api_key, 'user', '')
```
Kami: I'm a bit confused about this part - do we simply read the `user` attribute provided by the user which is included in the API model?

If so, this definitely doesn't look right - this means any user could generate a token for any other user which is definitely not a good idea :)

st2api/st2api/controllers/v1/auth.py
```diff
@@ -0,0 +1,171 @@
...
+        except Exception as e:
+            LOG.exception('Database delete encountered exception during '
+                          'delete of api_key_id_or_key="%s". ', api_key_id_or_key)
+            abort(http_client.INTERNAL_SERVER_ERROR, str(e))
+            return
+
+        extra = {'api_key_db': api_key_db}
+        LOG.audit('ApiKey deleted. ApiKey.id=%s' % (api_key_db.id), extra=extra)
+
+    def _get_user(self, api_key):
```
Kami: api_key -> api_key_api to make it clear what's getting passed in.

st2common/st2common/constants/auth.py
```diff
@@ -30,6 +30,9 @@
 HEADER_ATTRIBUTE_NAME = 'X-Auth-Token'
 QUERY_PARAM_ATTRIBUTE_NAME = 'x-auth-token'
 
+HEADER_API_KEY_ATTRIBUTE_NAME = 'St2-Api-Key'
```
Kami: X-API-Key for consistency with token? :P

st2common/st2common/hooks.py
```diff
@@ -162,18 +166,52 @@ def _abort_other_errors():
...
-        return validate_token(token_in_headers=token_in_headers,
-                              token_in_query_params=token_in_query_params)
+
+        api_key_in_headers = headers.get(HEADER_API_KEY_ATTRIBUTE_NAME, None)
+        api_key_in_query_params = query_params.get(QUERY_PARAM_API_KEY_ATTRIBUTE_NAME, None)
+
+        if (token_in_headers or token_in_query_params) and \
+           (api_key_in_headers or api_key_in_query_params):
+            raise auth_exceptions.MultipleAuthSourcesError(
+                'Only one of Token or API key expected.')
```
Kami: :+1:

st2common/st2common/hooks.py
```diff
@@ -162,18 +166,52 @@ def _abort_other_errors():
         return webob.Response(body=body, status=status, headers=headers)
 
     @staticmethod
-    def _validate_token(request):
+    def _get_validated_user(request):
```
Kami: Minor thing - maybe "_validate_token/creds_and_get_user"?

st2common/st2common/models/api/auth.py
```diff
@@ -92,3 +94,49 @@ def to_model(cls, instance):
...
+
+        return cls(**attrs)
+
+    @classmethod
+    def to_model(cls, instance):
+        user = str(instance.user) if instance.user else None
+        key = str(instance.key) if instance.key else None
+        metadata = getattr(instance, 'metadata', {})
+
+        # The api_key is never really stored in the DB. It is assumed
```
Kami: Can we create and use multiple API models (we already follow this approach in some other places) to make this a bit less confusing:

1. APIKeyAPI - id, key_hash, etc. - used for get all and get one response
2. APIKeyCreateResponseAPI - id, key, etc. - used for create response

This way we don't need this "key_hash=key" thing which is confusing and the whole thing is a bit more explicit.

st2common/st2common/models/db/auth.py
```diff
@@ -49,4 +53,32 @@ class TokenDB(stormbase.StormFoundationDB):
...
+    key_hash = me.StringField(required=True, unique=True)
+    metadata = me.DictField(required=False,
```
Kami: I would prefer `description` over metadata but I just noticed that token also used metadata so that's fine.

On top of that, as noted above I think created_at and enabled attribute would also come handy here.

Kami: Edit: I just saw how you used metadata in the fixtures / examples below. That's fine we should just document that and encourage users to populate this field :)

st2common/st2common/util/auth.py
```diff
@@ -58,3 +66,55 @@ def validate_token(token_in_headers, token_in_query_params):
...
+        LOG.audit('API key is not found in header or query parameters.')
+        raise exceptions.ApiKeyNotProvidedError('API key is not provided.')
+
+    if api_key_in_headers:
+        LOG.audit('API key provided in headers')
+
+    if api_key_query_params:
+        LOG.audit('API key provided in query parameters')
+
+    api_key = api_key_in_headers or api_key_query_params
```
Kami: Hm, I'm slightly confused here - did I miss the part where we actually hash the "plain-text" version of the key which is passed in by the user?

Kami: I just grepped for hash_utils and I found it gets hashed in the model get method - https://github.com/StackStorm/st2/pull/1964/files#diff-a15ecbc227b545b35658d0cbb697c99dR73.

I'm personally not a big fan of that since the DB model itself doesn't and shouldn't know anything about the raw key and how to handle that. I would prefer to hash it either inside the auth hook or here.

st2api/st2api/controllers/v1/auth.py
manasdk: Well this is to enable delete. Imagine you have the `ApiKey` but now don't know which to delete from StackStorm therefore this is supported.

Kami: Ah, that's a good point.

It would also be less confusing if the argument was called `api_key_id_or_hash` :)

st2api/st2api/controllers/v1/auth.py
manasdk: I am definitely looking to add `created`. We can consider `enabled` and `description` later.

`user` is useful until we can get RBAC sorted and perhaps even later.

st2common/st2common/models/db/auth.py
manasdk: Not foreign key, it is username assuming name is unique. I could switch to ID but then it requires a join for the result of ApiKeyDB to be useful.

st2api/st2api/controllers/v1/auth.py
manasdk: Correct. It sorta supports the non-auth enabled use-case i.e. without auth enabled a user should still be able to say create and API-key scoped to my user. Likely this is pointless as RBAC anyway would not work in such a case so maybe in the auth disabled usecase I will switch over to system user. It maybe the simpler approach.

st2common/st2common/constants/auth.py
manasdk: Following https://tools.ietf.org/html/rfc6648

There is little to no value using X as the prefix as per that RFC and the entire purpose of the RFC is to get folks to deprecate X prefixed headers.

st2common/st2common/models/api/auth.py
manasdk: Assuming it is not too messy to handle in the controller I can make that change.

What is passed into POST @ /apikeys? For now that method takes ApiKeyAPI. Should it be ApiKetCreateSpec since key_hash cannot be input? I suppose I can entirely ignore key_hash for POST still it does not particularly help with consistency.

st2common/st2common/models/db/auth.py
manasdk: As you noticed I copied TokenDB. We likely should've used tags :)

Will stay away from description for now since we maybe able to support queries on metadata/tags in the future.

st2common/st2common/util/auth.py
manasdk: @kami - That is not the DB model but service on top of DB model. Its fine for that to manage such abstractions and also a very good reason for that layer to even exist.

I can move it to auth code but that mean even all the get calls in controllers have to worry about this detail. I do honestly prefer my approach.

st2api/st2api/controllers/v1/auth.py
manasdk: I copied from an existing controller. Looks like the change has not propagated to all controllers yet. I will update this one.

st2api/st2api/controllers/v1/auth.py
manasdk: Concretely,

if auth -> will use user from auth headers.
if no-auth -> token scoped as system_user

st2api/st2api/controllers/v1/auth.py
manasdk: 258671af69f011ba407bb25c05c78117d889db7c

st2api/st2api/controllers/v1/auth.py
manasdk: 258671af69f011ba407bb25c05c78117d889db7c

st2common/st2common/hooks.py
manasdk: a3c76b08c90e23bdb51a6530d8b5212f5c20ecf5

st2api/st2api/controllers/v1/auth.py
manasdk: It is not the hash it is the actual key so the name is exactly as per the expected property.

st2api/st2api/controllers/v1/auth.py
manasdk: fd13301a572b08be03ad92145920d6dc23d0e28b

st2common/st2common/models/api/auth.py
manasdk: a129342fa56788e7521912eb7fd7608566ecf2fe

st2api/st2api/controllers/v1/auth.py
manasdk: created_at and enabled - 960ac0a3a35d3fb952ab7ca6336565b9c657742d

PR #1966, opened 2015-09-09
Add example to file_watch_sensor

jfryman:
This commit adds an example of how to consume the `file_watch_sensor` in the `config.yaml` for the Pack.


Kami:
LGTM, +1

PR #1967, opened 2015-09-09
Add pydoc for BaseSensor

armab:
Very tiny cosmetic change for better IDE support when writing [new Sensors](http://docs.stackstorm.com/sensors.html#authoring-a-sensor)

autocomplete-ide-support

### DIFF
--- a/st2reactor/st2reactor/sensor/base.py
+++ b/st2reactor/st2reactor/sensor/base.py
@@ -16,6 +16,13 @@ class BaseSensor(object):
     """
 
     def __init__(self, sensor_service, config=None):
+        """
+        :param sensor_service: Sensor Service instance.
+        :type sensor_service: :class:``st2reactor.container.sensor_wrapper.SensorService``
+
+        :keyword config: Sensor config.
+        :type config: ``dict`` or None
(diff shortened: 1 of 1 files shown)


manasdk:
+1

st2reactor/st2reactor/sensor/base.py
```diff
@@ -16,6 +17,13 @@ class BaseSensor(object):
     """
 
     def __init__(self, sensor_service, config=None):
+        """
+        :param sensor_service: Sensor Service instance.
+        :type sensor_service: ``SensorService``
```
Kami: Ideally it should say ":class:`SensorService`"

armab: Nice thing, added that as well: https://libcloud.readthedocs.org/en/latest/development.html#docstring-conventions

However curious, is there any difference?

Also is there any difference between single and double quotes: 
```
``dict``
```
and
```
`dict`
```
?

armab:
Again cleaned my messy history of commits with force push

Kami:
Merged, thanks.

PR #1968, opened 2015-09-09
Sensor container fixes

Kami:
* Fix sensor container service so the ``config`` argument is correctly passed to the sensor instances in the system packs. Previously, this argument didn't get passed correctly to the FileWatchSensor from the system linux pack.
* Make sure sensor processes correctly pick up parent ``--debug`` flag. This makes debugging a lot easier since user simply needs to start sensor container with ``--debug`` flag and all the sensor logs with level debug or higher will be routed to the container log.


lakshmi-kannan:
Can you add info --debug to docs if it isn't there already?

lakshmi-kannan:
Good catch! Nice PR! LGTM.

PR #1970, opened 2015-09-10
wsgify st2api

lakshmi-kannan:
Copied from st2auth and modified. Works in st2express.

### DIFF
--- a/st2api/st2api/wsgi.py
+++ b/st2api/st2api/wsgi.py
@@ -1,0 +1,42 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from pecan import load_app
+from oslo_config import cfg
+
+from st2api import config  # noqa
+from st2common import log as logging
+from st2common.models import db
+
+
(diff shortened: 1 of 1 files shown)


st2api/st2api/wsgi.py
```diff
@@ -0,0 +1,42 @@
...
+cfg.CONF(args=['--config-file', '/etc/st2/st2.conf'])
+
+logging.setup(cfg.CONF.api.logging)
+
+username = cfg.CONF.database.username if hasattr(cfg.CONF.database, 'username') else None
+password = cfg.CONF.database.password if hasattr(cfg.CONF.database, 'password') else None
+db.db_setup(cfg.CONF.database.db_name, cfg.CONF.database.host, cfg.CONF.database.port,
+            username=username, password=password)
+
+pecan_config = {
```
manasdk: I suspect the equivalent of https://github.com/StackStorm/st2/blob/master/st2api/st2api/app.py#L58 will be needed

st2api/st2api/wsgi.py
```diff
@@ -0,0 +1,42 @@
...
+
+from pecan import load_app
+from oslo_config import cfg
+
+from st2api import config  # noqa
+from st2common import log as logging
+from st2common.models import db
+
+
+cfg.CONF(args=['--config-file', '/etc/st2/st2.conf'])
```
Kami: So this will only work with the default config path, right?

Maybe we should document that or it's not an issue right now for AIO installer?

jfryman: Not an issue for AIO

st2api/st2api/wsgi.py
manasdk: I am wrong, looking at pecan code (https://github.com/pecan/pecan/blob/cbcc8b3e511f8366aabb04b50ce5c482cfcbb6f3/pecan/core.py#L205) app.py:setup_app will be called by pecan as part of load_app.

manasdk:
+1

enykeev:
I didn't confirm it yet, but I'm pretty sure that with that change we just broke Stream API so no live updates and no chatops.

Back in the days when I've tried to move st2api to uwsgi, I wasn't able to make uwsgi to serve stream to more than one connection and that was the reason I haven't merged it.

PR for reference: https://github.com/StackStorm/st2/pull/1348

May be I'm missing something, but we really need to confirm this ASAP.

manasdk:
Oh crap. Alright hopefully we are not using wsgi.py yet.

PR #1974, opened 2015-09-10
Fix cast params bug

Kami:
Previously, `cast_params` would explode with an unexpected error if the provided action didn't exist locally.

This pull request fixes it to throw a ValueError if the action doesn't exist.

This was discovered by @alustweiss on Slack when trying to schedule an execution using hubot and the action referenced in the alias didn't exist (https://gist.github.com/indiggio/aac040b391c0f99a2ead)


manasdk:
+1

PR #1976, opened 2015-09-11
Fix trigger_instance being too broad for unique filters

enykeev:


manasdk:
I had to look at a whole bunch of code to work out whats going on here but now this makes sense to me. My intuition is that we screwed up some of the code arrangement here and the interaction between classes is actually unclear - also perhaps some comments are lacking. 

For starters please add a comment on SUPPORTED_FILTERS as to how things can fall apart if IGNORE_FILTERS is not updated.

Further, the use of executions/views/filters and its impact on UI is obviously unclear. We definitely need some tests - it is very unlikely that the next person who touches these will be able to work out the cascading effects.

manasdk:
+1

st2api/st2api/controllers/v1/executionviews.py
```diff
@@ -23,6 +23,14 @@
 
 LOG = logging.getLogger(__name__)
 
+# List of supported filters and relation between filter name and execution property it represents.
```
manasdk: Thanks for adding this message :)

Kami:
:+1:

PR #1977, opened 2015-09-11
Packs APIs fixes and improvements

Kami:
This pull request includes various fixes and improvements in the packs and packs file(s) view APIs.

1. Allow user to reference a pack using the reference or id (before it was either using name or id). In the pack case, reference is pack directory name. The problem before was that in some pack.yaml files `name` attribute doesn't match the pack directory name. In all other places we refer to the pack by pack directory name (ref) so we should do the same in the API.

2. Fix to exclude non-text (binary) files in the files API response. Binary files can't be directly included in the files JSON response (we could base64 encode it) so for now it's simple better to exclude it. You can still retrieve binary (files) using the "file" API since that API returns binary data and not JSON.

3. Add some safe-guards so users can't retrieve large files using this API. This helps with preventing DDoS and resource exhausting attacks (buffering very large files in memory). Eventually we can make another improvement and switch to reading file in smaller chunks and streaming the response incrementally in the file API.


Kami:
Added tests, merging.

PR #1978, opened 2015-09-11
gunicorn wsgi config and other gunicorn related fixes

Kami:
This pull request adds gunicorn config to st2api and st2auth and fixes issues which were preventing it to work under gunicorn.

As I mentioned it on Slack, I've tested it and after minor fix stream handler also works fine now.

Here are the commands used to star gunicorn server / workers. Those are the commands we will also need to put in service manager start files.

On top of that, we also need to make sure that PYTHONPATH is set correctly in the service manager (st2api and st2auth packages need to be in PYTHONPATH).

st2api:

```bash
gunicorn_pecan st2api/st2api/gunicorn_config.py -k eventlet
```

st2api (using non-default listen info):

```bash
gunicorn_pecan st2api/st2api/gunicorn_config.py -k eventlet -b 127.0.0.1:9101
```

st2auth:

```bash
gunicorn_pecan st2auth/st2auth/gunicorn_config.py -k eventlet
```

st2auth (using non default listen info):

```bash
gunicorn_pecan st2auth/st2auth/gunicorn_config.py -k eventlet -b 127.0.0.1:9100
```

Keep in mind that the actual listen ip and port can also be specified in the gunicorn config, but supplying them as a command line argument is more flexible.

This pull request also updates launchdev.sh script - you can now pass `-g` flag to it and st2api and st2auth will be running under gunicorn.

## Future optimizations

In the future, it would make sense to add a separate entry point for stream handler and serve it using a separate gunicorn server (workers).

Why?

Regular (non-stream) API requests are short lived, but the stream one are long-lived so serving everything using the same workers is not ideal. If we have a separate handler we can tweak / optimize timeout and other settings for the stream server.

In any case, it works fine for now and stream periodically sends a heart-beat to the client so timeout is not an issue right now.

## TODO

- [x] Add integration tests which spin up a gunicorn server using those configs
 - [ ] Remove wsgi.py after we transition to the new setup (post merge)


st2api/st2api/gunicorn_config.py
```diff
@@ -0,0 +1,47 @@
...
+
+__all__ = [
+    'app'
+]
+
+DEFAULT_ST2_CONFIG_PATH = '/etc/st2/st2.conf'
+ST2_CONFIG_PATH = os.environ.get('ST2_CONFIG_PATH', DEFAULT_ST2_CONFIG_PATH)
+
+CONFIG_ARGS = ['--config-file', ST2_CONFIG_PATH]
+common_setup(service='api', config=config, setup_db=True, register_mq_exchanges=True,
```
Kami: @manasdk @lakshmi-kannan Keep in mind that the entry point in wsgi.py was incorrect. It might have worked, but it was incomplete and incorrect since we only connected to the db, but we didn't do other bootstrap work (register exchanges, register internal trigger types, handle --debug flag, etc.).

That's also the reason why I have refactored common bootstrap code in `service_setup.setup` in the past. Before that, each service duplicated and copy and pasted this code which was error prone and a maintenance headache.

Kami: On a related note - do we still need broken wsgi.py or can I nuke it?

/cc @lakshmi-kannan @jfryman

Kami:
Re: `gunicorn_pecan` binary / script - as mentioned on slack, this small wrapper ships with pecan and sets some pecan options for gunicorn.

We can also make it works by directly using gunicorn binary it simply means we need more clutter in `gunicorn_config.py`.

st2api/st2api/gunicorn_config.py
jfryman: 🔥

Kami:
I can't get the test to work on TravisCI, so for now I will make sure it's skipped on Travis (but we will still run it locally).

The test is not comprehensive anyway, but it's better than nothing - we test that the gunicorn config works and that the server starts.

Kami:
I pushed a change so the default listen ip and port is now specified in the gunicorn config. This means you need to only provide `-b` command line argument if you want the server to listen to a non-standard port (one which is not set in the st2 config).

Kami:
Note: We decided we also want to include this in v0.13.3.

st2api/st2api/gunicorn_config.py
lakshmi-kannan: Yep :fire:

PR #1987, opened 2015-09-14
Fix "packs.info" action

### LINKED ISSUE #1986: st2 run packs.info pack=NONEXISTANT always succeeds
Existing pack:
```
st2 run packs.info pack=hubot

id: 55f6bae8e8b34d6189b1fc92
status: succeeded
result: 
{
    "stdout": "", 
    "result": {
        "ref": "bf3a689479398aa234e814964373953c94f33299", 
        "repo_url": "https://github.com/StackStorm/st2contrib.git", 
        "branch": "master", 
        "subtree": true
    }, 
    "stderr": "", 
    "exit_code": 0
}
```

Non existing pack
```
st2 run packs.info pack=whatever

id: 55f6ba93e8b34d6189b1fc8f
status: succeeded
result: 
{
    "stdout": "Unable to load git info for whatever
", 
    "result": null, 
    "stderr": "", 
    "exit_code": 0
}
```

shouldn't the exit code be different than 0 and the status Failed?
This means that a simple automated way of checking which packs are installed is not possible as it always exits 0

Kami:
This branch fixes two issues:

1. Make sure action exits with non-zero if pack doesn't exist or if it doesn't contain a valid `.gitinfo` file
2. Correctly search all the pack search paths defined in the config. Previously we only searched ``/opt/stackstorm/packs`` which was hard coded so the action also wouldn't work if the user used a custom packs dir.

This resolves #1986

### DIFF
--- a/contrib/packs/actions/pack_mgmt/info.py
+++ b/contrib/packs/actions/pack_mgmt/info.py
@@ -13,9 +13,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from st2actions.runners.pythonrunner import Action
-import json
 import os
+import sys
+import json
+
+from st2actions.runners.pythonrunner import Action
 
 GITINFO_FILE = '.gitinfo'
 
@@ -28,4 +30,5 @@ class PackInfo(Action):
                 details = json.load(data_file)
                 return details
         except:
-            print "Unable to load git info for {}".format(pack)
+            error = 'Pack %s doesn\'t exist or it doesn\'t contain .gitinfo file' % (pack)
+            raise Exception(error)
--- a/contrib/packs/actions/pack_mgmt/info.py
+++ b/contrib/packs/actions/pack_mgmt/info.py
@@ -14,21 +14,42 @@
 # limitations under the License.
 
 import os
-import sys
 import json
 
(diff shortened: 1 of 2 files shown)


contrib/packs/actions/pack_mgmt/info.py
```diff
@@ -13,19 +13,43 @@
...
+
+        pack_git_info_path = None
+        for packs_base_path in packs_base_paths:
+            git_info_path = os.path.join(packs_base_path, pack, GITINFO_FILE)
+
+            if os.path.isfile(git_info_path):
+                pack_git_info_path = git_info_path
+                break
+
+        error = ('Pack "%s" doesn\'t exist or it doesn\'t contain a valid .gitinfo file' % (pack))
```
manasdk: can move this into if block to avoid string creation

manasdk:
+1

contrib/packs/actions/pack_mgmt/info.py
Kami: Heh, yeah.

I originally re-used the same error message in both places, but later on I decided to make the second error message a bit more descriptive, but I forgot to move the variable declaration.

PR #1990, opened 2015-09-14
Add API endpoint for mistral workflow validation

m4dcoder:
Add an experimental endpoint in the API to validate mistral workbook and workflow definition.


st2api/st2api/controllers/exp/root.py
```diff
@@ -23,3 +24,6 @@ class RootController(object):
     # Deprecated. Use /v1/ instead.
     actionalias = ActionAliasController()
     aliasexecution = ActionAliasExecutionController()
+
+    # Experimental
+    validators = ValidationController()
```
manasdk: We mostly use singular

Kami: I thought we mostly use plural for API endpoints - /executions, /actions, /packs, etc. (models are sadly a different story).

manasdk: Something about the /validators/mistral REST url bothers me - I am not able to exactly wrap my head around what is being created. It is definitely not a validator.

Since this is experimental I assume we have space to change the URL structure later. I guess I could get behind creating a `validation` where a `validation` is a noun similar to creation of an `actionexecution`

manasdk:
+1

m4dcoder:
I struggled where to put the endpoint so figuring this is exp, I placed it under validators.  I will use validation for the endpoint instead.

PR #1991, opened 2015-09-14
Add support for logging MongoDB queries and profile data

Kami:
This pull request adds support for logging actual MongoDB queries which are executed and query-related profile data (to retrieve the query related profile data it uses MongoDB explain functionality).

Queries and profile data are only logged if the service is started with `--debug` or `--profile` flag.

This functionality will make debugging, optimizing and troubleshooting MongoDB index and other performance related issues easier.

Here is an example log statement which is generated when the user hits /v1/executions API endpoint:

```bash
2015-09-14 20:57:39,961 DEBUG [-] MongoDB query: db.action_execution_d_b.find({'parent': {'$exists': False}}).sort({start_timestamp: -1, action.ref: 1}); (mongo_query={'parent': {'$exists': False}},mongo_shell_query="db.action_execution_d_b.find({'parent': {'$exists': False}}).sort({start_timestamp: -1, action.ref: 1});")
2015-09-14 20:57:39,962 DEBUG [-] MongoDB explain data: {u'allPlans': [{u'cursor': u'BtreeCursor start_timestamp_-1_action.ref_1_status_1',
                u'indexBounds': {u'action.ref': [[{u'$minElement': 1},
                                                  {u'$maxElement': 1}]],
                                 u'start_timestamp': [[{u'$maxElement': 1},
                                                       {u'$minElement': 1}]],
                                 u'status': [[{u'$minElement': 1},
                                              {u'$maxElement': 1}]]},
                u'n': 64,
                u'nscanned': 70,
                u'nscannedObjects': 70},
               {u'cursor': u'BtreeCursor parent_1',
                u'indexBounds': {u'parent': [[None, None]]},
                u'n': 69,
                u'nscanned': 69,
                u'nscannedObjects': 69},
               {u'cursor': u'BasicCursor',
                u'indexBounds': {},
                u'n': 61,
                u'nscanned': 69,
                u'nscannedObjects': 69}],
 u'cursor': u'BtreeCursor parent_1',
 u'indexBounds': {u'parent': [[None, None]]},
 u'indexOnly': False,
 u'isMultiKey': False,
 u'millis': 0,
 u'n': 69,
 u'nChunkSkips': 0,
 u'nYields': 0,
 u'nscanned': 69,
 u'nscannedAllPlans': 208,
 u'nscannedObjects': 69,
 u'nscannedObjectsAllPlans': 208,
 u'scanAndOrder': True,
 u'server': u'vagrant-ubuntu-trusty-64:27017'}
```


st2common/st2common/models/db/__init__.py
```diff
@@ -113,7 +115,9 @@ def get_all(self, *args, **kwargs):
         return self.query(*args, **kwargs)
 
     def count(self, *args, **kwargs):
-        return self.model.objects(**kwargs).count()
+        result = self.model.objects(**kwargs).count()
+        log_query_and_profile_data_for_queryset(queryset=result)
```
Kami: My original plan was to monkey patch mongoengine `QuerySet.__call__` method, but it turns out this doesn't work too well with the chained methods approach (e.g. `foo.query().order_by().limit()...`).

In any case, luckily we have an additional abstraction layer for talking to the database so we can just add this instrumentation here. Some people might also argue that this approach is also better since it's more explicit.

manasdk: I would definitely argue that this explicit approach is clearer.

st2common/st2common/models/db/profiling.py
```diff
@@ -0,0 +1,105 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
```
manasdk: since this is not a model can you move it to utils?

`st2common.models.utils` is likely a better home for profiling.py

manasdk:
Great.

+1

st2common/st2common/models/db/profiling.py
Kami: Yeah, I was thinking about that, but because some other non-model related functionality already lives in this package (db_setup, mongo access, etc.) I decided to put it there.

Ideal split would imo be, moving all the non model related db stuff to `st2common.db`. In any case, I can move it to utils for now.

Kami:
Added some tests, verified that all the tests pass with profiling mode enabled (this uncovered some bugs), refactored the code and added some docs.

Will go ahead and merge it.

lakshmi-kannan:
Super cool!

PR #1993, opened 2015-09-15
Add missing pack.yaml to examples pack

Kami:
Caught by @enykeev.

(As mentioned in the other pr which adds a missing pack.yaml to default pack, not having a pack.yaml breaks all kinds of things)


enykeev:
[the other pr](1833), just for reference

lakshmi-kannan:
Cool.

PR #1995, opened 2015-09-17
ETag and Last-Modified headers for files

manasdk:
Fixed as per https://www.keycdn.com/blog/a-guide-to-http-cache-headers/

* ETag and Cache-Control are the HTTP 1.1 way of handling intermediate caches
* Reduce the max-age to 90s from 24Hrs. This is ok since the browser will request the value but will not necessarily end up retrieving content if content has not changed. However, a backend request will still be made.

### DIFF
--- a/st2api/st2api/controllers/v1/packviews.py
+++ b/st2api/st2api/controllers/v1/packviews.py
@@ -13,13 +13,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
 import codecs
 import mimetypes
+import os
 
 import six
-from pecan import abort, expose, response
+from pecan import abort, expose, request, response
 from pecan.rest import RestController
+from wsgiref.handlers import format_date_time
 
 from st2api.controllers.v1.packs import BasePacksController
 from st2common.exceptions.db import StackStormDBObjectNotFoundError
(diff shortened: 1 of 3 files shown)


enykeev:
Right now, if you try to get the workflow definition before 90s pass from the last attempt, you'll get previous version no matter how many changes you've made in between. We should probably remove `Cache-Control` completely and solely rely on last-modified. We simply don't have lifetime requirements on this type of resources to say "this response would be valid for 90s no matter what". We also don't fully understand how this thing would affect proxy servers on the route to our user.

enykeev:
Also, as far as I know, ETag is supposed to be content-based, not time-based, so if file was just `touch`ed, it would still return 304. Not that it is in any way critical to us, though if you know some easy way to get file hash (preferably without loading the file from disk)...

manasdk:
> Also, as far as I know, ETag is supposed to be content-based, not time-based, so if file was just touched, it would still return 304.

You are right it should be content based. However, no way to get to the hash without loading the file so I decided to use modified time.  I guess we can take the hit for touch.

manasdk:
@enykeev cache-control removed 0749f435e69a4e7cd715ab32784244ec18b1f46f

enykeev:
:+1:

Kami:
I share @enykeev's original "concerns" (which it looks that they are already not relevant anymore) , but overall I like the change :+1: 

Also, now that we have nginx in the middle in the production deployments we could also leverage nginx as a caching proxy for static files...

enykeev:
I wonder, is there a way to detect frontend webserver to avoid sending `x-sendfile`/`x-accel-redirect` headers directly to the user?

PR #1997, opened 2015-09-17
fixing in-requirements.txt for previously add mistralclient into st2api

dennybaa:
Introducing dependencies without adding them into appropriate `in-requirements.txt` can cause inconsistency of packages in the future https://github.com/StackStorm/st2/commit/0ae8fd08bda9ff40d14bb6a79fb92f187e542a9d#diff-8d5ef5953c18fe21054d0958e6e79160R20. @m4dcoder 

New packaging has separate components which are *st2api*, *st2common*, *st2actions* etc.
These components are bundled with their own separate virtual environment, thus missing including dependency into specific `in-requirements.txt` will result in "missing module", even if the global (toplevel) `requirements.txt` contain it...

### DIFF
--- a/st2api/in-requirements.txt
+++ b/st2api/in-requirements.txt
@@ -8,3 +8,4 @@ oslo.config
 pymongo
 six
 git+https://github.com/StackStorm/pecan.git@st2-patched#egg=pecan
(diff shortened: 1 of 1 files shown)


manasdk:
+1

Kami:
Yep, +1

PR #1998, opened 2015-09-18
Retry db connection

manasdk:
Initial connection to the database are a bit more resilient with some retry attempts. In some scenarios as the services are coming up it is likely that the DB has not started yet and providing some retry capabilities makes services startup a little more resilient.

Note : This changes does nothing for cases where DB goes down while services are running. That is still likely to require a restart and manifest strangely.

### DIFF
--- a/st2api/st2api/wsgi.py
+++ b/st2api/st2api/wsgi.py
@@ -27,8 +27,8 @@ logging.setup(cfg.CONF.api.logging)
 
 username = cfg.CONF.database.username if hasattr(cfg.CONF.database, 'username') else None
 password = cfg.CONF.database.password if hasattr(cfg.CONF.database, 'password') else None
-db.db_setup(cfg.CONF.database.db_name, cfg.CONF.database.host, cfg.CONF.database.port,
-            username=username, password=password)
+db.db_setup_with_retry(cfg.CONF.database.db_name, cfg.CONF.database.host, cfg.CONF.database.port,
+                       username=username, password=password)
 
 pecan_config = {
     'app': {
--- a/st2api/st2api/wsgi.py
+++ b/st2api/st2api/wsgi.py
@@ -18,7 +18,7 @@ from oslo_config import cfg
 
 from st2api import config  # noqa
 from st2common import log as logging
-from st2common.models import db
+from st2common.persistence import db_init
 
 
 cfg.CONF(args=['--config-file', '/etc/st2/st2.conf'])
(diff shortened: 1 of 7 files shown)


st2api/st2api/wsgi.py
```diff
@@ -18,7 +18,7 @@
 
 from st2api import config  # noqa
 from st2common import log as logging
-from st2common.models import db
+from st2common.persistence import db_init
```
lakshmi-kannan: This is going to be confusing. persistence seems like the right place though. sigh.

st2common/st2common/config.py
```diff
@@ -94,6 +94,12 @@ def register_opts(ignore_errors=False):
         cfg.StrOpt('db_name', default='st2', help='name of database'),
         cfg.StrOpt('username', help='username for db login'),
         cfg.StrOpt('password', help='password for db login'),
+        cfg.IntOpt('connection_retry_max_delay_m', help='Connection retry total time (minutes).',
```
lakshmi-kannan: please use seconds consistently.

lakshmi-kannan:
lgtm

st2common/st2common/config.py
manasdk: minutes makes a nicer API.

st2api/st2api/wsgi.py
manasdk: models.db was my first attempt and failed miserably. This is the next best option at this point.

st2common/st2common/config.py
Kami: Yeah, I wish we would use the same consistent setting names and API everywhere where we implement retries (message bus, querier, mistral, ...).

Otherwise it makes for a inconsistent and confusing interface...

st2common/st2common/persistence/db_init.py
```diff
@@ -0,0 +1,60 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
```
Kami: Ideally, we would eventually also move db_setup and db_teardown here.

st2common/st2common/config.py
Kami: On a related note - we should update existing code to also use the same `retrying` library in other places (Mistral, querier, etc.). Right now the same logic is implemented a bit differently in at least two other places which is, at the minimum, a maintenance headache.

manasdk: Sure. Next time someone mucks around that area or someone has free cycles we can go ahead and change.

PR #1999, opened 2015-09-19
[RFR] Reference count trigger when created via rule

manasdk:
* Trigger objects are reference counted and as the rules are added
  and removed the reference count is updated. This only applies to
  Triggers with parameters while triggers without parameters don't
  need any reference counting.
* Once the reference count drops to 0 the trigger_db is removed.

Testing
* Added some dedicated tests since it is quirky enough.
* Validate with multiple timer rules and multiple webhook rules.


Kami:
Woo, finally!

st2common/st2common/services/triggers.py
```diff
@@ -175,9 +175,26 @@ def create_trigger_db_from_rule(rule):
...
 
     if not existing_trigger_db:
-        return create_or_update_trigger_db(trigger_dict)
+        trigger_db = create_or_update_trigger_db(trigger_dict)
+    else:
+        trigger_db = existing_trigger_db
+
+    # Special reference counting for trigger with parameters.
+    if trigger_dict.get('parameters', None):
+        Trigger.update(trigger_db, inc__ref_count=1)
```
Kami: This uses MongoDB's atomic increment operation, right?

st2common/st2common/models/api/trigger.py
```diff
@@ -125,6 +125,8 @@ class TriggerAPI(BaseAPI):
     @classmethod
     def from_model(cls, model, mask_secrets=False):
         trigger = cls._from_model(model, mask_secrets=mask_secrets)
+        # Hide ref count from API.
+        trigger.pop('ref_count', None)
```
Kami: :laughing:

Kami:
The change looks good to me, but some tests would be good.

st2common/st2common/persistence/trigger.py
```diff
@@ -42,6 +45,28 @@ def _get_publisher(cls):
                 urls=transport_utils.get_messaging_urls())
         return cls.publisher
 
+    @classmethod
+    def delete_if_unreferenced(cls, model_object, publish=True, dispatch_trigger=True):
+        # Found in the innards of mongoengine
+        delete_query = model_object._object_key
```
Kami: Curious, what does this attribute contain?

st2common/st2common/services/triggers.py
manasdk: Yes

st2common/st2common/persistence/trigger.py
manasdk: Some weird magic :P 

I will test and document this a little better

st2common/st2common/models/db/__init__.py
```diff
@@ -163,7 +163,10 @@ def update(self, instance, **kwargs):
         return instance.update(**kwargs)
 
     def delete(self, instance):
-        instance.delete()
+        return instance.delete()
+
+    def delete_by_query(self, **query):
+        return self.model.objects.filter(**query).delete()
```
Kami: You can modify this method a bit so we also log the query and profile data:

```python
result = self.model.objects.filter(**query).delete()
log_query_and_profile_data_for_queryset(result)
result = result.delete()
return result
```

It's slightly annoying that you can't simply pass result of `delete` to this method, but mongoengine API is not ideal and not all the methods return an instance of `QuerySet` object. That's also one of the reasons why right now we only log "find()" queries.

manasdk: Done d93266a5c4e08ebdeec19489c06153e12e217081

st2common/st2common/persistence/trigger.py
manasdk: Documented 40059d547f3fa11accaa3206db8064cfc3fb43b9

st2api/st2api/controllers/v1/rules.py
```diff
@@ -90,6 +91,9 @@ def post(self, rule):
                                                                        rule_api=rule)
 
             rule_db = Rule.add_or_update(rule_db)
+            # After the rule has been added modify the ref_count. This way a failure to add
+            # the rule due to violated constraints will have no impact on ref_count.
+            increment_trigger_ref_count(rule_api=rule)
```
Kami: Not related or a blocker for this PR (also not your fault) and something which has already bothered me in the past - wish we had a single code path / abstractions which was executed in the API and during the resource registrar run...

Needing to copy and paste it across multiple files and places is a disaster waiting to happen...

st2common/st2common/services/triggers.py
```diff
@@ -175,9 +175,39 @@ def create_trigger_db_from_rule(rule):
...
+    '''
+    trigger_dict = _get_trigger_dict_given_rule(rule_api)
+
+    # Special reference counting for trigger with parameters.
+    if trigger_dict.get('parameters', None):
+        trigger_db = _get_trigger_db(trigger_dict)
+        Trigger.update(trigger_db, inc__ref_count=1)
+
+
+def cleanup_trigger_db_for_rule(rule_db):
```
Kami: Would also be good to add a function docstring there.

Kami:
Nice, LGTM :+1:

Kami:
Minor thing - also need to update CHANGELOG.

Kami:
Another thing - how will this affect existing timer triggers in the database?

st2common/st2common/models/api/trigger.py
lakshmi-kannan: LOL

lakshmi-kannan:
>Another thing - how will this affect existing timer triggers in the database?

This is going to be tricky. Also, st2ctl reload --register-rules should be modified to add ref count I think since it doesn't use the API.

Kami:
@lakshmi-kannan 

> Also, st2ctl reload --register-rules should be modified to add ref count I think since it doesn't use the API.

I believe it's already there (rules registrar change).

lakshmi-kannan:
>I believe it's already there (rules registrar change).

Search fail with Safari. Search for `registrar` didn't show `rulesregistrar`.

st2api/st2api/controllers/v1/rules.py
manasdk: Yup. Bothered me a lot yesterday but I decided to minimize entropy.

manasdk:
> Another thing - how will this affect existing timer triggers in the database?

Thing will be very unhappy so let me see what I can do.

st2api/st2api/controllers/v1/rules.py
manasdk: Btw, could be totally my fault ... I may have written a bunch of this code myself.

manasdk:
Existing triggers are update via this hacked up migration approach 79c5e86548b4afc524936415bf931aaaa10ae639

jfryman:
@manasdk Yes, you can do this.

What I would do is to add an `exec` block to https://github.com/StackStorm/puppet-st2/blob/master/manifests/profile/server.pp at the bottom with a `$version` conditional as you have in the shell script. Once you're good, here are the docs on pushing a module to the forge. https://github.com/StackStorm/puppet-st2/blob/master/docs/release_to_forge.md

I know we just need to get this done, but let me throw out this disclaimer. Having migrations like this at the installer level may not be the best place, given that we have multiple ways of installation. Our lowest common denominator are Packages... maybe something there as part of `post-install`.

manasdk:
> Having migrations like this at the installer level may not be the best place, given that we have multiple ways of installation.

Yeah. Agreed this is really a hack since no one has spent time in working out what a real migration story would look like. In this case also the only reason to even write this script was to make sure that following an upgrade things continue to work.

I hope we can invest in consistent migration story post 1.0

Kami:
Agreed on migrations.

Also, far as as migrations go, we really need two types of migrations:

1. Data and schema migrations
2. Execute "random" code migrations (what's now in installer - e.g. delete virtualenv, run some script, etc.)

As far as 1) goes - this belongs in st2 and something I have already talked about in the past (happy to work on it post v1.0)

As far as 2) goes - ideally we would only have one installation method (installer) so those migrations could live in that repo.

PR #2001, opened 2015-09-20
st2 packaging: release fix in version string

dennybaa:
minor fix of the release of debian packages

### DIFF
--- a/st2actions/Makefile
+++ b/st2actions/Makefile
@@ -77,7 +77,7 @@ requirements:
 
 changelog: populate_version
 ifeq ($(DEBIAN),1)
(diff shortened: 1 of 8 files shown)


Kami:
Just curious, will this counter (ST2PKG_RELEASE) get auto incremented on each release or how is this supposed to work?

Kami:
In any case, LGTM & merged. Thanks.

dennybaa:
np! noop, it's not a counter. It should be implemented externally, this is just an input variable. So if there's a proper number somewhere it can be passed as this env var.

Kami:
Ah, thought it might be something like that.

Thanks for the clarification.

PR #2002, opened 2015-09-20
Add support in mistral runner for encrypted st2 API endpoint

m4dcoder:


DoriftoShoes:
+1

st2actions/st2actions/runners/mistral/v2.py
```diff
@@ -144,7 +144,10 @@ def try_run(self, action_parameters):
         inputs = self.runner_parameters.get('context', dict())
         inputs.update(action_parameters)
 
-        endpoint = 'http://%s:%s/v1/actionexecutions' % (cfg.CONF.api.host, cfg.CONF.api.port)
+        # If host value does not start with http, then default to http://.
+        protocol = 'http' if not cfg.CONF.api.use_ssl else 'https'
```
Kami: I know I'm a bit late to the game, but `auth.api_url` config option was added in the past for this purpose. This setting contains a base public URL to the API endpoint which should be uses by all the clients (Mistral is also just another client).

Kami: Another problem (lack of consistency) with calling this option `use_ssl` is that `auth`service / section also exposes an option with the same name, but it serves a different purpose.

If this option is enabled, auth API service listens on https when running in a standalone mode. So this lack of consistency could cause some confusion.

Kami: If we decide to use this option we also need to update `st2common.util.api.get_full_public_api_url` function.

This function is used in many other places and if we also don't update / fix it there, those other places will still be broken.

I will open a PR for that and also update this section to use this function.

Kami: To clarify it - if we update this place to use `get_full_public_api_url` we can get rid of this new (and potentially confusing) option all together.

We just need to make sure `cfg.CONF.auth.api_url` option is configured correctly, but this should already be the case otherwise other things wouldn't work.

PR #2003, opened 2015-09-21
Update mistral handler to use get_full_public_api_url() function

Kami:
This PR builds on #2002 and updates Mistral runner to use `get_full_public_api_url` function to retrieve a base URL to the API endpoint.

This function and underlying `auth.api_url` config option was added quite a while back. It's purpose was and is to serve as a single source of truth for the public base API URL (think static URL registry). All the clients which need to talk to the API endpoint on the public interface should use this value. Like other clients, Mistral is also just another API client so it should use the same value. If for some reason Mistral needs to talk on a private / different interface we should introduce a new option instead of "abusing" api.host and api.port option (one of the reasons for introducing another option in the past was that listen settings don't always directly map to the actual public URL which client uses - think API behind load balancer, etc.).

As noted in #2002, this setting needs to be configured correctly, otherwise things won't work as expected (might require st2enterprise puppet changes).

/cc @DoriftoShoes @jfryman re puppet setting

### DIFF
--- a/st2actions/st2actions/runners/mistral/v2.py
+++ b/st2actions/st2actions/runners/mistral/v2.py
@@ -30,6 +30,7 @@ from st2common import log as logging
 from st2common.models.api.notification import NotificationsHelper
 from st2common.util.workflow import mistral as utils
 from st2common.util.url import get_url_without_trailing_slash
+from st2common.util.api import get_full_public_api_url
 
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


m4dcoder:
LGTM. Is this consistent with how we configure https thru out st2?

Kami:
@m4dcoder Not exactly sure what you mean with "Is this consistent with how we configure https thru out st2", but the value of this setting is, among other places, also used in st2auth when user successfully authenticates - we return public URL to the API endpoint (st2auth also acts as a "service catalog / registry").

Kami:
Corresponding st2enterprise PR - https://github.com/StackStorm/st2enterprise/pull/36

jfryman:
:+1:.

I prefer the URI encapsulated format. Should be able to infer protocol from there.

PR #2006, opened 2015-09-21
permission 775 on all default pack resources

manasdk:
* minimum 660 is required since for some reason reading metadata i.e. os.stats
  of some files required `w` privileges. This more open than ideal but atleast
  icons render nicely in the UI.


Kami:
You mean 755, not 775, since the difference is probably made by "write" on "other", right?

PR #2009, opened 2015-09-21
Assign expected permissions at install time.

manasdk:
* The right permissions for installed packs are assigned at package
  install time. Trying to achieve this via git was a non-starter and
  likely the wrong approach anyway.

### DIFF
--- a/contrib/packs/actions/pack_mgmt/download.py
+++ b/contrib/packs/actions/pack_mgmt/download.py
@@ -17,6 +17,7 @@ import os
 import shutil
 import hashlib
 import json
+import stat
 
 import six
 from git.repo import Repo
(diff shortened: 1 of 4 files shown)


Kami:
Shouldn't we also set "correct" permissions on `packs.install`?

manasdk:
It somehow magically does the right thing and packs.install already ends up with 770 - I could not figure out :)

manasdk:
Some investigation -

`git clone` permissions are different depending either on git version or platform. e.g. `git clone` in ubuntu is clone files with permission 770 and 750 on my local osx box. This means we are just getting lucky on linux :).

DoriftoShoes:
LGTM.  Ship it!

lakshmi-kannan:
lgtm

PR #2011, opened 2015-09-22
Run st2ctl reload as sudo

manasdk:
* opt/stackstorm/packs permission is set to 770 therefore st2ctl needs to run as root


manasdk:
The rationale is that `775` on /opt/stackstorm/packs means config is visible to all users.

jfryman:
:+1:

PR #2012, opened 2015-09-22
Switch packs to 775

manasdk:
* 770 is restrictive and causes local scripts to fail. Switching to 775 should unblock those features. We can work out how to properly support this so that packs can be more secure.


manasdk:
Feel free to add your comments for the preferred behavior. For now I just want to unblock packages.

PR #2013, opened 2015-09-22
fix permissions for examples

manasdk:
* Looks like AMI does have examples deployed by default


Kami:
OK, so the issue was indeed related to a non-fresh install.

Solution which would fix it fox existing installations would be to add `chmod -Rf 775 %{buildroot}/opt/stackstorm/packs/examples` to install script. We use `-f` flag so it would simply be a no-op if the directory doesn't exist.

Issue #1923, opened 2015-09-02
st2 command displays requests warning


powellchristoph:
When running the st2 command, you are spammed with warnings from the requests library. It would be nice to disable these as it make reading/parsing the output difficult.

This was installed on Ubuntu 14.04.03 LTS with the All-In-One-Installer.

```
root@stackstorm-test:~# st2 action list --pack=core
/usr/local/lib/python2.7/dist-packages/requests/packages/urllib3/util/ssl_.py:90: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
/usr/local/lib/python2.7/dist-packages/requests/packages/urllib3/connectionpool.py:768: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
/usr/local/lib/python2.7/dist-packages/requests/packages/urllib3/util/ssl_.py:90: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
/usr/local/lib/python2.7/dist-packages/requests/packages/urllib3/connectionpool.py:768: InsecureRequestWarning: Unverified HTTPS request is being made. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.org/en/latest/security.html
  InsecureRequestWarning)
+------------------+------+-------------+----------------------------------------------+
| ref              | pack | name        | description                                  |
+------------------+------+-------------+----------------------------------------------+
| core.http        | core | http        | Action that performs an http request.        |
| core.local       | core | local       | Action that executes an arbitrary Linux      |
|                  |      |             | command on the localhost.                    |
| core.local_sudo  | core | local_sudo  | Action that executes an arbitrary Linux      |
|                  |      |             | command on the localhost.                    |
| core.remote      | core | remote      | Action to execute arbitrary linux command    |
|                  |      |             | remotely.                                    |
| core.remote_sudo | core | remote_sudo | Action to execute arbitrary linux command    |
|                  |      |             | remotely.                                    |
[...]

PR #2014, opened 2015-09-22
Add "silence_ssl_warnings" CLI config option

Kami:
The title says it all.


DoriftoShoes:
+1

armab:
Closes https://github.com/StackStorm/st2/issues/1923 ?

Sad there is no silence by default, since warnings just spam everywhere. Especially annoying is to get them in `stderr`, what can affect rules, workflows and so on.

Maybe intercept and show custom message is an option? From what I've seen in Ansible codebase, - they intercept almost everything they could catch to show own error messaging.

PR #2021, opened 2015-09-28
Paramiko STDOUT/STDERR UTF-8 decode bug fix. 

lakshmi-kannan:
From the commit "When consuming stdout and stderr byte streams from paramiko
channel, consume all the raw bytes and then decode. Decoding
chunk as they come won't work because utf-8 is variable
length encoding. When we want to support streaming, this
is going to be a pain."

### DIFF
--- a/st2actions/st2actions/runners/ssh/paramiko_ssh.py
+++ b/st2actions/st2actions/runners/ssh/paramiko_ssh.py
@@ -421,19 +421,22 @@ class ParamikoSSHClient(object):
         Try to consume stdout data from chan if it's receive ready.
         """
 
+        out = bytearray()
         stdout = StringIO()
         if chan.recv_ready():
             data = chan.recv(self.CHUNK_SIZE)
+            out += data
 
             while data:
-                stdout.write(self._get_decoded_data(data))
                 ready = chan.recv_ready()
 
                 if not ready:
                     break
 
                 data = chan.recv(self.CHUNK_SIZE)
+                out += data
 
+        stdout.write(self._get_decoded_data(out))
         return stdout
 
     def _consume_stderr(self, chan):
@@ -441,28 +444,30 @@ class ParamikoSSHClient(object):
         Try to consume stderr data from chan if it's receive ready.
         """
 
+        out = bytearray()
         stderr = StringIO()
         if chan.recv_stderr_ready():
             data = chan.recv_stderr(self.CHUNK_SIZE)
+            out += data
 
             while data:
-                stderr.write(self._get_decoded_data(data))
                 ready = chan.recv_stderr_ready()
 
                 if not ready:
                     break
 
                 data = chan.recv_stderr(self.CHUNK_SIZE)
+                out += data
 
+        stderr.write(self._get_decoded_data(out))
         return stderr
 
     def _get_decoded_data(self, data):
-        decoded_data = str(data)
         try:
-            decoded_data = decoded_data.decode('utf-8')
+            return data.decode('utf-8')
         except:
             self.logger.warning('Non UTF-8 character found in data: %s', data)
-        return decoded_data
+            return ''
 
     def _get_pkey_object(self, key):
         """
--- a/st2actions/st2actions/runners/ssh/paramiko_ssh.py
+++ b/st2actions/st2actions/runners/ssh/paramiko_ssh.py
@@ -466,8 +466,8 @@ class ParamikoSSHClient(object):
         try:
             return data.decode('utf-8')
         except:
-            self.logger.warning('Non UTF-8 character found in data: %s', data)
-            return ''
+            self.logger.exception('Non UTF-8 character found in data: %s', data)
+            raise
 
     def _get_pkey_object(self, key):
         """
--- a/st2actions/tests/unit/test_paramiko_ssh.py
+++ b/st2actions/tests/unit/test_paramiko_ssh.py
@@ -249,11 +249,9 @@ class ParamikoSSHClientTests(unittest2.TestCase):
 
     @patch('paramiko.SSHClient', Mock)
     def test_consume_stdout(self):
-        """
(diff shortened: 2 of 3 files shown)


st2actions/tests/unit/test_paramiko_ssh.py
```diff
@@ -246,3 +246,49 @@ def test_put_dir(self, *args):
...
+            '\xF0'.decode('utf-8')
+            self.fail('Test fixture is not right.')
+        except UnicodeDecodeError:
+            pass
+        stdout = mock._consume_stdout(chan)
+        self.assertEqual(u'\U00010348', stdout.getvalue())
+
+    @patch('paramiko.SSHClient', Mock)
+    def test_consume_stderr(self):
+        """
```
manasdk: This shows up in the test output instead of functioname. Could get hard to track failed test.

Kami:
Good catch.

I you can figure out if a chunk contains an end / start of the utf-8 sequence we could build our own incremental decoder (buffer until we find end of sequence, decode, repeat the process...), but anyway, yeah, that seems like an overkill for now.

Kami:
Would be great to also submit this changes upstream to Libcloud - it might just require a bit more work since we support Python 2.5 - 3.5 there...

st2actions/tests/unit/test_paramiko_ssh.py
```diff
@@ -246,3 +246,49 @@ def test_put_dir(self, *args):
...
+        conn_params = {'hostname': 'dummy.host.org',
+                       'username': 'ubuntu'}
+        mock = ParamikoSSHClient(**conn_params)
+        mock.CHUNK_SIZE = 1
+        chan = Mock()
+        chan.recv_stderr_ready.side_effect = [True, True, True, True, False]
+
+        chan.recv_stderr.side_effect = ['\xF0', '\x90', '\x8D', '\x88']
+        try:
```
manasdk: I have no idea what this block does.

st2actions/tests/unit/test_paramiko_ssh.py
```diff
@@ -246,3 +246,49 @@ def test_put_dir(self, *args):
...
+        chan.recv_ready.side_effect = [True, True, True, True, False]
+
+        chan.recv.side_effect = ['\xF0', '\x90', '\x8D', '\x88']
+        try:
+            '\xF0'.decode('utf-8')
+            self.fail('Test fixture is not right.')
+        except UnicodeDecodeError:
+            pass
+        stdout = mock._consume_stdout(chan)
+        self.assertEqual(u'\U00010348', stdout.getvalue())
```
manasdk: what is this magic string?

st2actions/st2actions/runners/ssh/paramiko_ssh.py
```diff
@@ -441,28 +444,30 @@ def _consume_stderr(self, chan):
...
 
     def _get_decoded_data(self, data):
-        decoded_data = str(data)
         try:
-            decoded_data = decoded_data.decode('utf-8')
+            return data.decode('utf-8')
         except:
             self.logger.warning('Non UTF-8 character found in data: %s', data)
-        return decoded_data
+            return ''
```
manasdk: Hmm, empty string seems like a legit response ... how about something dramatic like __ERROR__ ?

st2actions/st2actions/runners/ssh/paramiko_ssh.py
```diff
@@ -441,28 +444,30 @@ def _consume_stderr(self, chan):
...
-                stderr.write(self._get_decoded_data(data))
                 ready = chan.recv_stderr_ready()
 
                 if not ready:
                     break
 
                 data = chan.recv_stderr(self.CHUNK_SIZE)
+                out += data
 
+        stderr.write(self._get_decoded_data(out))
```
Kami: (thinking out loud)

Since we only write to StringIO object once now the only reason to keep it is for the API to be backward compatible.

manasdk:
+1

We can worry about stream decoding later. Painful but solvable.

st2actions/tests/unit/test_paramiko_ssh.py
lakshmi-kannan: I just pulled this multi-byte uff-8 character from Wikipedia :P.

st2actions/st2actions/runners/ssh/paramiko_ssh.py
lakshmi-kannan: Correct. I definitely don't like it. Maybe we can change the APIs locally but upstream we should be more careful not to break the API. But it is underscore :P. LMK and I can fix.

st2actions/tests/unit/test_paramiko_ssh.py
lakshmi-kannan: Yep. It's annoying. I was following other tests in the file. Let me make them # comments.

st2actions/tests/unit/test_paramiko_ssh.py
lakshmi-kannan: This validates that the first part of the multi-byte character isn't decodable. Otherwise, the test has no meaning.

st2actions/st2actions/runners/ssh/paramiko_ssh.py
```diff
@@ -441,28 +444,30 @@ def _consume_stderr(self, chan):
...
         return stderr
 
     def _get_decoded_data(self, data):
-        decoded_data = str(data)
         try:
-            decoded_data = decoded_data.decode('utf-8')
+            return data.decode('utf-8')
         except:
             self.logger.warning('Non UTF-8 character found in data: %s', data)
```
Kami: +1 after we change this to throw

lakshmi-kannan: Fixed 887eeec3e96b81a0a6430084a340975389176b61

PR #2022, opened 2015-09-29
RFR: Allocate a pseudo tty when executing commands over SSH

lakshmi-kannan:


st2actions/st2actions/runners/ssh/paramiko_ssh.py
```diff
@@ -349,6 +349,7 @@ def run(self, cmd, timeout=None, quote=False):
         chan = transport.open_session()
 
         start_time = time.time()
+        chan.get_pty()
```
lakshmi-kannan: I'll check tomorrow what fabric does but without this running apt-get commands fail with return code -1 and no visible error. http://serverfault.com/questions/489939/ssh-session-never-closes-when-executing-apt-get-install recommends allocating a pseudo tty. With this, puppet_bootstrap always succeeds.

Kami:
Hm, curious what is going on here.

This definitely works in Libcloud because the deploy functionality is mostly used for bootstrapping a server which usually is something along the lines of apt-get update ; apt-get install puppet-server / chef, etc.

st2actions/st2actions/runners/ssh/paramiko_ssh.py
lakshmi-kannan: https://github.com/fabric/fabric/blob/master/fabric/operations.py#L721 Fabric always sets pty=True unless you explicitly disable it. This might explain why puppet_bootstrap always succeeds with fabric.

lakshmi-kannan: So finally, I got to the bottom of this. Fabric always sets pty=True for sudo operations which is what linux_bootstrap uses. Look at ``stdout`` and ``stderr`` fields carefully in the following logs.

Fabric success: https://gist.github.com/lakshmi-kannan/55eebe65e94ee42cda09
Paramiko success: https://gist.github.com/lakshmi-kannan/ac4131f7de0ff9648650
Paramiko failure: https://gist.github.com/lakshmi-kannan/4d744b2f7938b42dc251

Also a reminder that with pseudo tty enabled, paramiko never failed. Fabric sets pseduo tty for sudo actions because sudo actions take in a password. So setting pseudo tty on sudo actions will solve the problem for us now but I cannot understand why the action still fails. The invoke-rc.d errors in failure case seems fishy. I didn't go anywhere with it. Calling @jfryman and @DoriftoShoes to look at the failure logs. More info posted on slack.

Kami: Does the "interleaved streams" issue still exist with pty on?

lakshmi-kannan: >Does the "interleaved streams" issue still exist with pty on?

There will be no stderr with pty on as you can see in fabric logs. But fabric turns it on only for sudo actions.

jfryman: The error in the failure is interesting. 

```invoke-rc.d: -----------------------------------------------------
invoke-rc.d: WARNING: 'invoke-rc.d puppet start' called
invoke-rc.d: during shutdown sequence.
invoke-rc.d: enabling safe mode: initscript policy layer disabled
invoke-rc.d: -----------------------------------------------------```

This usually means that the server is in a different run-level, so when `apt-get` attempts to start the daemon, it bombs out. 

In the `/etc/init.d/puppet` init script...

```
# Default-Start:     2 3 4 5
# Default-Stop:      0 1 6
```

So, the server is in runlevel 0, 1, or 6 here, as this is how it would manifest. So, I'm curious about what else is happening around the workflow. Is the server still starting up, and not fully in runlevel 2? Is something else shutting down the server?

Does this provide any hints?

File: st2actions/st2actions/runners/ssh/paramiko_ssh.py [hunk 2]
```diff
@@ -351,6 +351,10 @@ def run(self, cmd, timeout=None, quote=False):
         chan = transport.open_session()
 
         start_time = time.time()
+        if cmd.startswith('sudo'):
```
Kami: A bit late, but `cmd.startswith('sudo ')` is probably a bit safer (what if someone has a binary which name starts with sudo) :P

manasdk: Perhaps we should have a separate `run` and `sudo` method in the runner.

PR #2024, opened 2015-09-29
Throw more user-friendly exception if uses passes path, not content for the "private_key" parameter

Kami:
The title says it all.

We already have documentation which clarifies that this parameter needs to contain key material and not a path, and we also have docs with examples which show how to use `@` CLI notation, but users are still stumbling on this so we should do something to mitigate this and save both (ourselves and them) some time.

### DIFF
--- a/st2actions/st2actions/runners/ssh/paramiko_ssh.py
+++ b/st2actions/st2actions/runners/ssh/paramiko_ssh.py
@@ -37,6 +37,8 @@ __all__ = [
     'SSHCommandTimeoutError'
 ]
 
+PRIVATE_KEY_HEADER = 'PRIVATE KEY-----'.lower()
+
 
 class SSHCommandTimeoutError(Exception):
     """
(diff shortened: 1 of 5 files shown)


st2actions/st2actions/runners/ssh/paramiko_ssh.py
```diff
@@ -483,5 +485,14 @@ def _get_pkey_object(self, key):
             else:
                 return key
 
-        msg = 'Invalid or unsupported key type'
+        # If a user passes in something which looks like file path we throw a more friendly
+        # exception letting the user know we expect the contents a not a path.
+        # Note: We do it here and not up the stack to avoid false positives.
+        contains_header = PRIVATE_KEY_HEADER in key.lower()
+        if not contains_header and (key.count('/') >= 1 or key.count('\\') >= 1):
```
Kami: As noted in the comment - the best place to check for this would be in the runner, but doing there could result in false positives so I do it here after we try to load the key.

lakshmi-kannan:
Nice. +1.

manasdk:
+1

PR #2025, opened 2015-09-29
CLI per user token cache

Kami:
This pull request improves token caching in the CLI - it now supports caching tokens for different users.

This means you can easily use multiple CLI config via `ST2_CONFIG_FILE` option or similar without disabling token cache.

On top of that it also adds some tests.

### DIFF
--- a/st2client/st2client/shell.py
+++ b/st2client/st2client/shell.py
@@ -409,14 +409,6 @@ class Shell(object):
 
         return token
 
-    def _get_cached_token_path_for_user(self, username):
-        """
-        Retrieve cached token path for the provided username.
-        """
-        file_name = 'token-%s' % (username)
-        result = os.path.abspath(os.path.join(ST2_CONFIG_DIRECTORY, file_name))
-        return result
-
     def _get_cached_auth_token(self, client, username, password):
         """
         Retrieve cached auth token from the file in the config directory.
@@ -461,9 +453,11 @@ class Shell(object):
 
         now = int(time.time())
         if (expire_timestamp + TOKEN_EXPIRATION_GRACE_PERIOD_SECONDS) < now:
+            LOG.debug('Cached token from file "%s" has expired' % (cached_token_path))
             # Token has expired
             return None
 
+        LOG.debug('Using cached token from file "%s"' % (cached_token_path))
         return token
 
     def _cache_auth_token(self, token_obj):
@@ -514,6 +508,7 @@ class Shell(object):
         with os.fdopen(fd, 'w') as fp:
             fp.write(data)
 
+        LOG.debug('Token has been cached in "%s"' % (cached_token_path))
         return True
 
     def _authenticate_and_retrieve_auth_token(self, client, username, password):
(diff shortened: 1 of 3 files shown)


st2client/st2client/shell.py
```diff
@@ -449,14 +451,17 @@ def _get_cached_auth_token(self, client, username, password):
...
         except Exception as e:
-            msg = 'File with cached token is corrupted: %s' % (str(e))
+            msg = ('File "%s" with cached token is corrupted or invalid (%s). Please delete '
+                   ' this file' % (cached_token_path, str(e)))
             raise ValueError(msg)
 
         now = int(time.time())
-        if (expire_timestamp + TOKEN_EXPIRATION_GRACE_PERIOD_SECONDS) < now:
+        if (expire_timestamp - TOKEN_EXPIRATION_GRACE_PERIOD_SECONDS) < now:
+            LOG.debug('Cached token from file "%s" has expired' % (cached_token_path))
```
manasdk: use comma, so string replacement is delayed to when required.

lakshmi-kannan:
With this I think we should fix CLI so if I supply credentials on the command line, we don't read from the rc config file. It got horribly confusing when I was doing that. 

st2 auth admin -p <password> would throw an authentication error because CLI was using creds from ~/.st2/config. Another PR is fine but calling out that this is indeed a problem.

st2client/st2client/shell.py
Kami: Premature optimization ftw :P

Kami:
@lakshmi-kannan Yeah, that's reasonable - don't use / skip CLI config when user uses `auth` command.

manasdk:
I am very confused with this change. If a token is cached then the token is already scoped to some user so why would a different user even be provided. Also, what usecase do we solve for where on the same client box we would like to use separate users? Also, the cached tokens etc are stored in the local users home dir i.e. this should literally be the same human being correct - so what is the usecase for supporting different usernames?

Kami:
@manasdk Testing and using different user accounts and connecting to different environment (staging, production, development, etc.).

Think supernova for OpenStack (https://github.com/major/supernova), etc.

For example:

```bash
ST2_CONFIG_FILE=~/.st2/user1.ini st2 action list
ST2_CONFIG_FILE=~/.st2/user2.ini st2 action list
ST2_CONFIG_FILE=~/.st2/user3.ini st2 action list

etc.
```

That's what I did in my RBAC demo, but I needed to disable token caching to make it work.

In short - making it easy for a single user to work / test with different accounts and / or environments.

manasdk:
> In short - making it easy for a single user to work / test with different accounts and / or environments.

I suspect this is a precedence issue. Something very much like what @lakshmi-kannan alluded to his comment.

I see why you would want to prefer using a token over authenticating every request, this is likely an issue with our approach which should be fixed with API keys and not caching tokens.

Anyway, go with what you have right now but this feel like bandage rather than a solve.

manasdk:
+1

lakshmi-kannan:
API keys vs tokens. Hmm. This will be a confusion pretty soon until we define what context does one prefer one over other?

manasdk:
> API keys vs tokens. Hmm. 

Then ignore I said API keys for now :P

PR #2027, opened 2015-09-30
Move registrars

manasdk:
* Having registrars in different location is confusing so collecting them in the same module.
* info line is unnecessary so moved it to debug like other registrars.


Kami:
:+1:

PR #2029, opened 2015-09-30
[CLI] Skip automatic authentication on "auth" command

Kami:
The title says it all.


st2client/st2client/shell.py
```diff
@@ -78,6 +78,11 @@
     'debug': ['cli', 'debug']
 }
 
+# A list of command classes for which automatic authentication should be skipped.
+SKIP_AUTH_CLASSES = [
+    'TokenCreateCommand'
```
manasdk: Can you use an import so that name changes impact are immediate and obvious. 

e.g. if TokenCreateCommand is renamed we will see a very visible regression.

manasdk:
+1

st2client/st2client/shell.py
Kami: First I actually check for "auth" being first value in argv, but then I decided to go with the current approach...

I'm also working on some tests, but it's a bit of a pita.

lakshmi-kannan:
Holy jolly +1

st2client/st2client/shell.py
Kami: @manasdk So let me know what you think, i'm fine with changing it back to checking for "auth" argv.

manasdk: I am cool with this ... all I am suggesting is

```
from st2client.commands.auth import TokenCreateCommand
SKIP_AUTH_CLASSES = [TokenCreateCommand.__name__]
```

assuming the import allow

Kami: Sure, I can do that.

PR #2030, opened 2015-09-30
Support single quote in actionaliasexecution

manasdk:
* values with spaces can be represented either by single or double quotes

### DIFF
--- a/st2common/st2common/models/utils/action_alias_utils.py
+++ b/st2common/st2common/models/utils/action_alias_utils.py
@@ -62,9 +62,6 @@ class JsonValueParser(object):
 
 
 class StringValueParser(object):
-    start = '"'
-    end = '"'
-    escape = '\\'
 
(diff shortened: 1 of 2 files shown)


Kami:
Do we need to make same changes to hubot plugin?

manasdk:
> Do we need to make same changes to hubot plugin?

Not same but related. Doing that right now.

Kami:
Yeah, that's what I meant :)

To confirm (I didn't see that in tests, maybe I missed it), does this allow user to mix quotes (e.g. one string is quoted using single quotes and one is using double quotes)?

manasdk:
> does this allow user to mix quotes

yes. I am considering that as a +ve thing :)

manasdk:
Corresponding fix in hubot-stackstorm https://github.com/StackStorm/hubot-stackstorm/pull/41

PR #2031, opened 2015-10-01
RFR - Strip '\r' or '\r\n' chars from stdout and stderr

lakshmi-kannan:
Found this issue by enabling paramiko on build002. 

It appears Fabric is already stripping '\r' or '\n' or '\r\n' characters from stdout and stderr. https://github.com/fabric/fabric/blob/master/fabric/io.py#L121

70808a0 removes the duplicate stripping from our fabric interface code. 

84aec38 mimics the stripping behavior in both local runner and paramiko runner.

### DIFF
--- a/st2common/st2common/util/misc.py
+++ b/st2common/st2common/util/misc.py
@@ -56,9 +56,26 @@ def compare_path_file_name(file_path_a, file_path_b):
     return file_name_a < file_name_b
 
 
-def strip_last_newline_char(input_str):
+def strip_shell_chars(input_str):
     """
-    Strips the last char if its newline.
+    Strips the last '\r' or '\n' or '\r\n' string at the end of
+    the input string. This is typically used to strip ``stdout``
+    and ``stderr`` streams of those characters.
+
+    :param input_str: Input string to be stripped.
+    :type input_str: ``str``
+
+    :rtype: ``str``
+    """
+    stripped_str = rstrip_last_char(input_str, '\n')
+    stripped_str = rstrip_last_char(stripped_str, '\r')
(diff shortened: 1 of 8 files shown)


manasdk:
+1

st2common/st2common/util/misc.py
```diff
@@ -68,7 +85,10 @@ def strip_last_newline_char(input_str):
     if not input_str:
         return input_str
 
-    if input_str.endswith('\n'):
-        return input_str[:-1]
+    if not char_to_strip:
+        return input_str
+
+    if input_str.endswith(char_to_strip):
```
lakshmi-kannan: rstrip is dangerous because it can remove multiple instances. 

E.g. In [1]: s = 'shit\n\n'

In [2]: s.rstrip('\n')
Out[2]: 'shit'

In [3]:

manasdk: Ah. Makes sense.

st2common/st2common/util/misc.py
```diff
@@ -56,9 +56,26 @@ def compare_path_file_name(file_path_a, file_path_b):
...
+    :type input_str: ``str``
+
+    :rtype: ``str``
+    """
+    stripped_str = rstrip_last_char(input_str, '\n')
+    stripped_str = rstrip_last_char(stripped_str, '\r')
+    return stripped_str
+
+
+def rstrip_last_char(input_str, char_to_strip):
```
Kami: It looks like the second argument should be called `word/str_to_strip` since it supports stripping multiple characters and that's also what you do in some of the tests :P

PR #2033, opened 2015-10-02
RFR: Env variables support paramiko

lakshmi-kannan:
Spent several hours on this and tested various cases. Seems to work now. But would appreciate more eyes and testing.  cc: @manasdk , @Kami

### DIFF
--- a/st2common/st2common/models/system/action.py
+++ b/st2common/st2common/models/system/action.py
@@ -52,6 +52,8 @@ LOGGED_USER_USERNAME = pwd.getpwuid(os.getuid())[0]
 
 
 class ShellCommandAction(object):
+    EXPORT_CMD = 'export'
+
     def __init__(self, name, action_exec_id, command, user, env_vars=None, sudo=False,
                  timeout=None, cwd=None):
         self.name = name
@@ -85,6 +87,18 @@ class ShellCommandAction(object):
 
     def get_cwd(self):
         return self.cwd
+
+    def _get_env_vars_export_string(self):
+        if self.env_vars:
+            exports = ' '.join(
+                '%s=%s' % (quote_unix(k), quote_unix(v))
+                for k, v in self.env_vars.iteritems()
+            )
+            shell_env_str = '%s %s' % (ShellCommandAction.EXPORT_CMD, exports)
+        else:
+            shell_env_str = ''
+
+        return shell_env_str
 
     def _get_command_string(self, cmd, args):
         """
--- a/st2common/st2common/models/system/action.py
+++ b/st2common/st2common/models/system/action.py
@@ -216,7 +216,7 @@ class ShellScriptAction(ShellCommandAction):
                     if value is True:
                         command_parts.append(arg)
                 else:
-                    command_parts.append('%s=%s' % (arg, quote_unix(str(value))))
+                    command_parts.append('%s=%s' % (quote_unix(arg), quote_unix(str(value))))
 
         # add the positional args
         if positional_args:
--- a/st2common/st2common/models/system/action.py
+++ b/st2common/st2common/models/system/action.py
@@ -90,6 +90,8 @@ class ShellCommandAction(object):
 
     def _get_env_vars_export_string(self):
         if self.env_vars:
+            # Envrionment variables could contain spaces and open us to shell
+            # injection attacks. Always quote the key and the value.
             exports = ' '.join(
                 '%s=%s' % (quote_unix(k), quote_unix(v))
                 for k, v in self.env_vars.iteritems()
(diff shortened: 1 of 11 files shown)


st2common/st2common/models/system/action.py
```diff
@@ -86,6 +88,18 @@ def get_timeout(self):
     def get_cwd(self):
         return self.cwd
 
+    def _get_env_vars_export_string(self):
```
lakshmi-kannan: Note to self: needs unit test

lakshmi-kannan:
Also found one more issue where we do cd {{cwd_dir}} && sudo -E -- bash -c 'command'. This is broken if only sudo user has access to cwd. So I'll be adding a commit to fix that.

lakshmi-kannan:
This still doesn't get this command to work on a box: https://github.com/StackStorm/st2cd/blob/master/actions/workflows/st2_deploy_test.yaml#L42

I am debugging that.

Kami:
(thinking out loud)

1. Does paramiko indeed use `export` or does it just use local variables and `FOO=bar ./script.sh` syntax? export is probably better in this case, especially if user / we run multiple commands...
2. export only works in some shells (sh, bash) but not others (csh), but I don't think that's a big deal since we pretty much assume bash.

Kami:
> Also found one more issue where we do cd {{cwd_dir}} && sudo -E -- bash -c 'command'. This is broken if only sudo user has access to cwd. So I'll be adding a commit to fix that.

Good catch. We should indeed cwd in the sudo block.

lakshmi-kannan:
>Does paramiko indeed use export or does it just use local variables and FOO=bar ./script.sh syntax? export is probably better in this case, especially if user / we run multiple commands

I manually create an ``export`` string and add it to the command. E.g. ``export FOO=BAR && /tmp/exec_id/script.py --arg1=foo``

>export only works in some shells (sh, bash) but not others (csh), but I don't think that's a big deal since we pretty much assume bash.
Yep. This is what fabric does too. Too bad if they use csh :P.

lakshmi-kannan:
I'd definitely appreciate :eyes: on the tests. I think now we have better coverage than before. There are still more tests I could write. If there is anything in particular you want me to add, let me know. This PR was a PITA.

st2common/st2common/models/system/action.py
```diff
@@ -86,6 +88,18 @@ def get_timeout(self):
...
+        if self.env_vars:
+            exports = ' '.join(
+                '%s=%s' % (quote_unix(k), quote_unix(v))
```
manasdk: AFAICT quote_unix will run on exports later when the complete command is prepared. Is it still necessary to quote here?

lakshmi-kannan: Trust me, it's needed :P.

st2common/tests/unit/test_paramiko_script_action_model.py
```diff
@@ -20,7 +20,107 @@
...
 
-    def test_get_command_string(self):
+    def test_get_command_string_no_env_vars(self):
+        script_action = ParamikoRemoteScriptActionTests._get_test_script_action()
+        ex = 'cd /tmp && /tmp/remote_script.sh song=\'b s\' \'taylor swift\''
+        self.assertEqual(script_action.get_full_command_string(), ex)
+
+        # Test with sudo
+        script_action.sudo = True
+        ex = 'sudo -E -- bash -c ' + \
```
manasdk: lol ... I look at this an accept it as fact

st2common/st2common/models/system/action.py
manasdk: Then please add a comment saying why it is needed. Else, it could easily be dropped in some refactoring/clean-up effort.

manasdk:
Seems like a liberal sprinkling of quote_unix  fixes our woes. For me very hard to rationalize this PR.

So +1

Kami:
> Seems like a liberal sprinkling of quote_unix fixes our woes. For me very hard to rationalize this PR.

The problem with "liberally sprinkling" quote_unix everywhere is that things will break if values are escaped multiple times.

As mentioned before, imo, escaping is same as handling unicode / bytes in Python 3 - as long as you do it explicitly and you know in every step you are dealing with (bytes / unicode, unescaped / escape) value, there will be no problems.

Anyway, yeah, the biggest problem is it's kinda hard to follow through those changes right now and know where the value is escaped.

Maybe in the future we should come up with some kind of class abstraction for escaped values which behaves like `str`. It might be a bit heavy and unpythonic, but the idea behind is that we can do "static" analysis / lint and know what we are dealing with in each line.

Kami:
Also, I know that I'm late, but LGTM :+1:

PR #2034, opened 2015-10-02
Add RBAC migration for deleting a bad (mistyped) role

Kami:
That's about as ghetto as it gets, but until we invest in a proper schema / data migration framework that's what we are left with.


manasdk:
lol ... what happened that this bad role got through?

Kami:
fat fingers

PR #2035, opened 2015-10-02
Remove st2common logging dependency from st2client

m4dcoder:


Kami:
LGTM, +1

(and also sorry, I think I made change)

Kami:
One thing we could eventually do to get rid of copy and paste is copy the file over during build in Makefile, similar as we do with `dist_utils.py`.

m4dcoder:
@Kami: Don't worry about it.

m4dcoder:
That's an interesting idea. Do you mean still reference st2common but copy over on make?  But how will it impact unit testing?

PR #2037, opened 2015-10-02
Use a different username than stanley to save the build

lakshmi-kannan:


st2common/tests/unit/test_paramiko_command_action_model.py
```diff
@@ -22,7 +22,7 @@ class ParamikoRemoteComamndActionTests(unittest2.TestCase):
 
     def test_get_command_string_no_env_vars(self):
         cmd_action = ParamikoRemoteComamndActionTests._get_test_command_action('echo boo bah baz')
-        ex = 'sudo -E -u stanley -- bash -c \'cd /tmp && echo boo bah baz\''
```
Kami: I'm slightly confused how this caused the build to fail since it looks like we only retrieve the command string and commands are not actually executed, but anyway...

PR #2038, opened 2015-10-03
Transform JSON output in stdout & stderr streams in Paramiko runner

lakshmi-kannan:


st2actions/st2actions/runners/ssh/parallel_ssh.py
```diff
@@ -254,8 +254,9 @@ def _run_command(self, host, cmd, results, timeout=None):
             client = self._hosts_client[host]
             (stdout, stderr, exit_code) = client.run(cmd, timeout=timeout)
             is_succeeded = (exit_code == 0)
-            results[host] = {'stdout': stdout, 'stderr': stderr, 'return_code': exit_code,
-                             'succeeded': is_succeeded, 'failed': not is_succeeded}
+            result_dict = {'stdout': stdout, 'stderr': stderr, 'return_code': exit_code,
```
Kami: So much magic everywhere :P

Kami:
Hm, looking at the existing code, it looks like Fabric runner does it in the Action system model.

In any case, I'm personally not a big fan of the whole Action system model abstraction so hopefully we can get rid of that once we deprecate and remove Fabric runner (I hope we will be able to do this for v1.0 once we fix all the issues in the new runner).

PR #2041, opened 2015-10-05
Resolve type properly

manasdk:
* Allow type to be comprised of underscores


Kami:
Thanks - can you please also add a test case for it?

manasdk:
Yes. Test in the works.

st2common/st2common/rbac/types.py
```diff
@@ -38,8 +38,8 @@ class PermissionType(Enum):
     PACK_ALL = 'pack_all'
 
     # Note: Right now we only have read endpoints for sensors types
-    SENSOR_VIEW = 'sensor_view'
-    SENSOR_ALL = 'sensor_all'
+    SENSOR_VIEW = 'sensor_type_view'
```
Kami: I explicitly decided to use sensor instead of sensor type to hide the type abstraction here since I don't think it's necessary.

Kami: Besides that, LGTM.

Thanks

st2common/st2common/rbac/types.py
```diff
@@ -50,7 +50,7 @@ class PermissionType(Enum):
 
     # Note: Execution create is granted with "action_execute"
     EXECUTION_VIEW = 'execution_view'
-    EXECUTION_RE_RUN = 'execution_re_run'
+    EXECUTION_RE_RUN = 'execution_rerun'
```
Kami: Any particular reason for this change? :P

st2common/st2common/rbac/types.py
manasdk: `def get_resource_type(cls, permission_type):` Will not work with the way it is right now if you use `sensor_*` instead of `sensor_type_*`.

I am not sure how you would want me to resolve the discrepancy.

st2common/st2common/rbac/types.py
manasdk: See the implementation of `get_resource_type`

`get_resource_type('execution_re_run') == execution_re` and what we want is get_resource_type('excution_rerun') == execution`

`_` can eutehr be in type name or permission name.

Kami: Good catch.

I'm fine with this change for now.

st2common/st2common/rbac/types.py
Kami: I'm fine with this change for now, I can figure something else out in the future PRs if needed.

PR #2042, opened 2015-10-06
RBAC Tutorial

manasdk:
This is tutorial style docs as opposed to a reference which is already present http://docs.stackstorm.com/latest/rbac.html

### DIFF
--- a/instructables/configuring_rbac.rst
+++ b/instructables/configuring_rbac.rst
@@ -4,14 +4,14 @@ Configuring RBAC
 StackStorm enterprise edition brings support for Role Based Access Control(RBAC). RBAC allows a StackStorm administrator
 control over what functions are accessible to a user within StackStorm.
 
-Possible scenarios :
+**Possible scenarios :**
 
-- A user owns a pack i.e is able to view, create, delete, modify and where applicable execute various resources like
+1. A user owns a pack i.e is able to view, create, delete, modify and where applicable execute various resources like
   actions, rules, sensors.
-- A user can create rules, execute actions and view a handful of actions.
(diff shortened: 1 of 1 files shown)


dzimine:
early feedback: +1 for format and structure.

manasdk:
I did a 2nd pass over the tutorial. I am happy to include more changes but for the time being will merge this in. Can pick up any suggestions in a separate PR.

instructables/configuring_rbac.rst
```diff
@@ -0,0 +1,148 @@
...
+-------------
+A newly created user has no assigned permissions. Each permission must be explicitly assigned to a user. To assign
+permission grants StackStorm requires creation of a role and then associating this role with a user. In this case we are trying to create a pack owner role.
+
+Lets first make sure there is a pack `x` we can use to experiment.
+
+.. sourcecode:: bash
+
+    $ cd /opt/stackstorm/packs/
+    $ mkdir x
```
Kami: I think using "real" names (e.g. let's just use libcloud or some other real pack name) in the examples is better.

manasdk: Yeah x is a placeholder. On the other hand using a real name is hard as I want to showcase action create and rule create - to do that with something like libcloud gets messy since I am moving actions and rule into the pack etc.

Kami:
I just noticed this is not in the docs/.

Are you OK with moving this to a new "Tutorials" section in the docs?

manasdk:
> Are you OK with moving this to a new "Tutorials" section in the docs?

Yeah. I can move it to tutorials.

Kami:
Done it in #2067

PR #2044, opened 2015-10-06
Paramiko sudo handling

manasdk:
* Local user and remote user do not have to match since anyway the command  will be excuting over ssh.


lakshmi-kannan:
I recommend a CHANGELOG entry. Otherwise LGTM.

Kami:
Lol, I was confused by the other PR a bit, but this clears it up.

PR #2045, opened 2015-10-06
Include security options in Requests library

jfryman:
Adds SSL support and removes annoying insecure platform warnings.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -26,7 +26,7 @@ python-gnupg<0.4,>=0.3.7
 python-json-logger
 python-nmap<0.4,>=0.3.4
 pyyaml<4.0,>=3.11
-requests<3.0,>=2.7.0 [security]
+requests<3.0,>=2.7.0
 retrying
 semver>=2.1.2
 six==1.9.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -26,7 +26,7 @@ python-gnupg<0.4,>=0.3.7
 python-json-logger
 python-nmap<0.4,>=0.3.4
 pyyaml<4.0,>=3.11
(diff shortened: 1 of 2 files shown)


Kami:
Requirements are managed a bit differently and fixed-requirements.txt needs to be updated (requirements.txt is generated during make step). I can take care of those changes, if needed.

requirements.txt
```diff
@@ -26,7 +26,7 @@ python-gnupg<0.4,>=0.3.7
 python-json-logger
 python-nmap<0.4,>=0.3.4
 pyyaml<4.0,>=3.11
-requests<3.0,>=2.7.0
+requests<3.0,>=2.7.0 [security]
```
Kami: Does this notation works (I always assumed it requires another entry)? :)

jfryman:
ok. one sec.

requirements.txt
jfryman: https://pip.readthedocs.org/en/1.1/requirements.html#the-requirements-file-format

Kami:
LGTM, +1

Lets merge this for now. We'll fix fixate-requirements.py and related code shortly.

requirements.txt
dennybaa: Thanks for the link. However i've started to dig into. 
Opposed to what is written there pip 7.1.2 **can not** parse `requests<3.0,>=2.7.0 [security]`
/cc @jfryman @Kami 
I'm looking to the issue.

dennybaa: This is the right link:
https://pip.readthedocs.org/en/stable/reference/pip_install/#requirements-file-format

Should be `requests[security]<3.0,>=2.7.0` :)

PR #2046, opened 2015-10-07
Revert "Include security options in Requests library"

jfryman:
Reverts StackStorm/st2#2045


jfryman:
So, this didn't work... and it caused issues with AIO installer. We need a better way to get these deps.

/cc @dennybaa @Kami

Kami:
Yeah, I assume it's related to an old(er) version of pip and / or setuptools.

I will investigate tomorrow.

PR #2048, opened 2015-10-07
Fix token priority and precedence in the CLI commands

Kami:
This bug fixes the issue with credentials in the CLI config having precedence over the token specified as en environment variable or as a command line argument.


manasdk:
+1

PR #2049, opened 2015-10-07
No more fail on error for st2ctl

jfryman:
Removes `set -e` on error, which allows services to start/stop cleanly when NG_INIT=true


DoriftoShoes:
+1

lakshmi-kannan:
merge

armab:
Same! :fireworks: :tada: 

Was annoying if any service down.

PR #2054, opened 2015-10-08
New "<resource>_list" permission type

Kami:
This pull request implements new <resource>_list permission type which is used to protect "get all" API endpoints.

Keep in mind that this is stop gap for v1.0. Post v1.0, we will implement actual list filtering.


st2common/st2common/rbac/resolvers.py
```diff
@@ -176,8 +217,37 @@ class SensorPermissionsResolver(PermissionsResolver):
     """
 
     def user_has_permission(self, user_db, permission_type):
-        # TODO
-        return True
+        assert permission_type in [PermissionType.SENSOR_LIST]
```
lakshmi-kannan: There is a lot of code duplication here. Can you make a note in JIRA to refactor this?

Kami: Sure, it was actually on my TODO, will just do it today :P

manasdk:
Post 1.0 will we completely drop `*_list` permission? If we do keep it how will it work with `*_get`. Just want to make sure we do not create a backwards compat concern for us.

manasdk:
I am overall +1 to the change.

Kami:
> Post 1.0 will we completely drop *_list permission? If we do keep it how will it work with *_get. Just want to make sure we do not create a backwards compat concern for us.

We can either decide to keep it (it will simple have precedence over implicit "view" grants on a particular resources) or remove it. If we decide to remove it we can decide to provide a migration tool.

manasdk:
Hmm, I am not convinced about keep it post 1.0 and nor I am unable to follow a reasonable migration path. But we shall see - hopefully we can entirely drop it without much of a penalty.

Kami:
Pushed the refactor.

PR #2057, opened 2015-10-08
Switch to using index based lookup instead of first

manasdk:
* Oddly .first() did not return the value that [0] did eventhough
  there is a value. It happened for a specific type so likely something
  to do with the query and perhaps a bug in mongoengine.

### DIFF
--- a/st2common/st2common/services/triggers.py
+++ b/st2common/st2common/services/triggers.py
@@ -40,9 +40,10 @@ LOG = logging.getLogger(__name__)
 def get_trigger_db_given_type_and_params(type=None, parameters=None):
     try:
         parameters = parameters or {}
+        trigger_dbs = Trigger.query(type=type,
+                                    parameters=parameters)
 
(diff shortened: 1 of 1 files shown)


manasdk:
We use first() is a lot of palce through our code-base. Not know why this failed makes me very uneasy.

lakshmi-kannan:
So weird! I am skeptical.

manasdk:
> So weird! I am skeptical.

Try it out locally. You shall see.

Kami:
Do we have some tests which catch / reproduce this issue?

manasdk:
> Do we have some tests which catch / reproduce this issue?

No.

manasdk:
@Kami @lakshmi-kannan Not able to repro in a test. This is really weird; I am not sure why mongoengine hates me.

lakshmi-kannan:
@manasdk noob :P

PR #2058, opened 2015-10-08
Fix pip path ubuntu post python-pip installation

lakshmi-kannan:
Initially we set PIP=`which pip`. This is empty for ubuntu. After installation, we don't set PIP variable again. So this script was failing. 

For RHEL/Fedora etc, we set PIP=pip2.7 explicitly. So for those, we should be good.

### DIFF
--- a/tools/st2_deploy.sh
+++ b/tools/st2_deploy.sh
@@ -465,6 +465,8 @@ install_apt() {
   # Install packages
   echo "Installing ${APT_PACKAGE_LIST}"
   apt-get install -y ${APT_PACKAGE_LIST}
+  # Now that pip is installed set PIP=`which pip` again.
+  PIP=`which pip`
   setup_rabbitmq
   install_pip
 }


manasdk:
+1

tools/st2_deploy.sh
```diff
@@ -465,6 +465,8 @@ install_apt() {
   # Install packages
   echo "Installing ${APT_PACKAGE_LIST}"
   apt-get install -y ${APT_PACKAGE_LIST}
+  # Now that pip is installed set PIP=`which pip` again.
+  PIP=`which pip`
```
Kami: I believe hash -r should also do it.

Kami: Also, ideally you would put it immediately after the line where we upgrade pip otherwise it's not really clear why it's here / needed.

lakshmi-kannan: Yeah, I am actually debating that. Let me do that.

lakshmi-kannan: Turns out this is where it gets installed. lol. merging.

PR #2059, opened 2015-10-09
Speed up filters aggregation

enykeev:
Filter request times out on our staging server with half a million of execution records (more precisely, it takes him 220 seconds to complete while default http timeout is 30).

This PR reverts the changes done in 8307cf7 since we don't have compound keys anymore.


Kami:
I'm a bit confused - we still have compound fields AFAIK. Or are you talking about live action and not action execution?

PR #2061, opened 2015-10-09
Control the list of filters from frontend

enykeev:
Since #2059 was not enough to bring filter view performance up to an expected level, we can additionally specify filters we want st2web to query.

Currently, UI only shows four types while server queries and sends nine.

### DIFF
--- a/st2api/st2api/controllers/v1/executionviews.py
+++ b/st2api/st2api/controllers/v1/executionviews.py
@@ -49,19 +49,26 @@ SUPPORTED_FILTERS = {
 IGNORE_FILTERS = ['parent', 'timestamp', 'liveaction', 'trigger_instance']
 
 
+def csv(s):
+    return s.split(',')
+
+
 class FiltersController(RestController):
-    @jsexpose()
-    def get_all(self):
+    @jsexpose(arg_types=[csv])
+    def get_all(self, types=None):
         """
             List all distinct filters.
 
             Handles requests:
-                GET /executions/views/filters
+                GET /executions/views/filters[?types=action,rule]
+
+            :param types: Comma delimited string of filter types to output.
+            :type types: ``str``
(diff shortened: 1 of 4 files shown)


st2api/tests/unit/controllers/v1/test_executions_filters.py
```diff
@@ -303,5 +303,12 @@ def test_filters_view(self):
...
         self.assertEqual(response.status_int, 200)
         self.assertIsInstance(response.json, dict)
-        for key, value in six.iteritems(history_views.ARTIFACTS['filters']):
+        for key, value in six.iteritems(history_views.ARTIFACTS['filters']['default']):
+            self.assertEqual(set(response.json[key]), set(value))
+
+    def test_filters_view_specific_types(self):
+        response = self.app.get('/v1/executions/views/filters?types=action,user,nonexistent')
+        self.assertEqual(response.status_int, 200)
+        self.assertIsInstance(response.json, dict)
```
Kami: To make sure that the controller indeed works as expected it would also be good to check the length of the dict (should be 2 here).

st2api/st2api/controllers/v1/executionviews.py
```diff
@@ -49,9 +49,13 @@
...
 
+def csv(s):
+    return s.split(',')
+
+
 class FiltersController(RestController):
-    @jsexpose()
-    def get_all(self):
+    @jsexpose(arg_types=[csv])
+    def get_all(self, types=None):
```
Kami: Would be to document `types` argument in the docstring.

Also, I didn't actually know you can reference query parameters in the same way you reference path arguments, I thought you needed to do `request.query` or something along those lines.

Besides that, LGTM.

Kami:
LGTM, +1

PR #2064, opened 2015-10-09
Change http runner to not verify ssl cert for https requests

m4dcoder:
Add an immutable runner parameter named verify_ssl_cert for the http runner and set it to False so the user knows that for HTTPS requests SSL cert is not verified. Pass the value to requests.reqest in the HTTPClient.

### DIFF
--- a/st2common/st2common/bootstrap/runnersregistrar.py
+++ b/st2common/st2common/bootstrap/runnersregistrar.py
@@ -305,6 +305,12 @@ RUNNER_TYPES = [
                 'type': 'boolean',
                 'default': False
             },
+            'verify_ssl_cert': {
+                'description': 'SSL verification is not supported.',
+                'type': 'boolean',
+                'default': False,
+                'immutable': True
+            }
         },
         'runner_module': 'st2actions.runners.httprunner'
     },
--- a/st2common/st2common/bootstrap/runnersregistrar.py
+++ b/st2common/st2common/bootstrap/runnersregistrar.py
@@ -306,10 +306,10 @@ RUNNER_TYPES = [
                 'default': False
             },
             'verify_ssl_cert': {
-                'description': 'SSL verification is not supported.',
+                'description': 'Set to True to verify SSL certificate. Verification '
+                               'using a CA cert bundle is not yet supported.',
                 'type': 'boolean',
-                'default': False,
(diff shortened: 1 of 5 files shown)


st2common/st2common/bootstrap/runnersregistrar.py
```diff
@@ -305,6 +305,12 @@
                 'type': 'boolean',
                 'default': False
             },
+            'verify_ssl_cert': {
+                'description': 'SSL verification is not supported.',
+                'type': 'boolean',
+                'default': False,
+                'immutable': True
```
manasdk: Why immutable?

We should make default=True and allow it to be turned of.

m4dcoder: So does this mean we are giving an option to verify SSL cert but not an option to pass a cacert bundle?

lakshmi-kannan: I like default = False and allow it to be mutable. 

You could also add cacert but it has to be content and not path. I'd think this is not a frequent use case. We can add it if we later need it.

Kami: Agreed with others - we should default to verifying the cert.

By default we should use system ca bundle and bundle shipped with requests which included most common CAs so this should cover the common use case.

Supporting custom CA bundle would also be nice (e.g. for internal infrastructure), but for now users can just disable SSL verification in case they use a custom CA.

st2common/st2common/bootstrap/runnersregistrar.py
```diff
@@ -305,6 +305,12 @@
                 'type': 'boolean',
                 'default': False
             },
+            'verify_ssl_cert': {
+                'description': 'Set to True to verify SSL certificate. Verification '
```
Kami: Should say "using a custom CA bundle is not yet supported."

Kami: Besides that, LGTM, +1.

Kami: On a related note - it would be good if we document SSL cert validation behavior in the runner docs - mention that we default to verifying the cert and we use requests CA bundle which comes from Mozilla...

PR #2068, opened 2015-10-12
Remote runner improvements

Kami:
* Set `return_code` to `-9` on command timeout - this way it's consistent with local runner and old fabric based runner
* If the timeout happens include stdout and stderr which was consumed so far
* Include original exception error message in the `error` attribute of the result object
* Strip environment variable which contains the action specific auth token for the error returned to the user

### DIFF
--- a/st2actions/st2actions/runners/ssh/parallel_ssh.py
+++ b/st2actions/st2actions/runners/ssh/parallel_ssh.py
@@ -14,12 +14,14 @@
 # limitations under the License.
 
 import json
+import re
 import os
 import sys
 import traceback
 
 import eventlet
 
+from st2common.constants.secrets import MASKED_ATTRIBUTE_VALUE
 from st2actions.runners.ssh.paramiko_ssh import ParamikoSSHClient
 from st2actions.runners.ssh.paramiko_ssh import SSHCommandTimeoutError
 from st2common import log as logging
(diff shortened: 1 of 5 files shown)


st2actions/st2actions/runners/ssh/paramiko_ssh.py
```diff
@@ -44,9 +44,19 @@ class SSHCommandTimeoutError(Exception):
     """
     Exception which is raised when an SSH command times out.
     """
-    def __init__(self, cmd, timeout):
```
Kami: I will port those and other changes upstream to Libcloud since they would also come handy there. On top of that it also makes me unhappy that we need to copy those changes here instead of relying on upstream changes.

st2actions/st2actions/runners/ssh/parallel_ssh.py
```diff
@@ -319,12 +320,25 @@ def _get_host_port_info(self, host_str):
...
+        stderr = getattr(ex, 'stderr', None) or ''
+
         error_dict = {
-            'error': error_msg,
-            'traceback': ''.join(traceback.format_tb(tb, 20)) if tb else '',
             'failed': True,
             'succeeded': False,
-            'return_code': 255
+            'return_code': return_code,
+            'stdout': stdout,
```
lakshmi-kannan: For connect errors, having stdout and stderr won't make sense, right? So if they aren't empty, then let's show the fields?

lakshmi-kannan:
Needs CHANGELOG.

lakshmi-kannan:
LGTM

st2actions/st2actions/runners/ssh/parallel_ssh.py
Kami: Yeah, I can exclude them if they are not present (aka they are `None`).

Kami: Actually, I changed my mind :P

It would be weird if we only included one, but not the other (e.g. there is no stderr), so it's probably better to just always include it.

PR #2073, opened 2015-10-13
Add missing dependencies for requests[security]

tgermain:
Those two deps ( `libffi-dev` and `libssl-dev` ) are required for cryptography and pyOpenSSL.

The problem occurs when installing from source inside of a docker container (ubuntu:14.04).


jfryman:
Excellent! Thanks for the doc update!

PR #2075, opened 2015-10-13
Modify the mistral_examples chain to run only selected examples

m4dcoder:
The mistral_examples WF is mostly used for health check and doesn't need to include all the mistral examples. Only selected examples are included here to test that mistral is installed and configured correctly.


Kami:
I personally think the more tests we run the better, but if those tests already run elsewhere and / or take a really long time, then I'm fine with disabling it here :)

m4dcoder:
It's already run elsewhere in st2ci.mistral on a daily basis.

PR #2078, opened 2015-10-15
Fix self check

m4dcoder:


tools/st2-self-check
```diff
@@ -47,31 +97,42 @@ TEST_ACTION_LIST=`st2 action list --pack=tests -w 90 | awk '{ print $2 }' | grep
...
+    else
+        echo "OK!"
+    fi
+done
 
 echo -n "Attempting Example examples.mistral_examples..."
 MISTRAL_EXAMPLES=`st2 run examples.mistral_examples -j`
-CHECK_STATUS=`echo ${MISTRAL_EXAMPLES} | grep '"status": "succeeded"'`
-if [ $? -ne 0 ]; then
+CHECK_STATUS=`echo ${MISTRAL_EXAMPLES} | grep '"status": "failed"'`
```
DoriftoShoes: Why did we invert the logic here?

tools/st2-self-check
```diff
@@ -18,6 +49,25 @@ else
   exit 2
 fi
 
+#Determine API HTTP Protocol
+PROTOCOL="http"
```
manasdk: Its sorta crazy you need to do all this. Anyway, lets roll with this for now - we can redesign later if needed.

Clearly something is not right with our current test design.

manasdk:
Assuming travis passes +1

tools/st2-self-check
m4dcoder: Because during some runs, tests failed but since action chain and mistral lists all status, some of the task status gets in the way.  I just find out failed doesn't work either because we expect some tasks to fail...

tools/st2-self-check
m4dcoder: LOL.  I couldn't agree more...

tools/st2-self-check
DoriftoShoes: Cool.  Lgtm!

PR #2081, opened 2015-10-16
(RFR) Reduce watcher sleep

manasdk:
* Sleeping for 4s between message retrievals causes unnecessary latency.
  Reducing to 0 since this is for co-oprative yields anyway so length
  should not matter. If it does I promise you we have a much larger problem.
* An ability to define Qs as exclusive to a connection.

* Processing old messages leads to unpredicatable behavior. With the
  TriggerWatcher consumers the idea is to 1st read from DB and then
  start consuming messages. However, non exclusive Qs cause old messages
  to be consumed leading to bugs and unpredicatbale behavior. e.g. and
  old create/update/delete might be processed.
* SensorWrapper should not share Qs. This would mean that the messages
  end up being distributed across sensor_wrappers. Each sensor_wrapper
  should get all messages and choose to process them or not.

Testing -
1. Validate passive sensor
2. Validated polling sensor

### DIFF
--- a/st2reactor/st2reactor/container/sensor_wrapper.py
+++ b/st2reactor/st2reactor/container/sensor_wrapper.py
@@ -328,7 +328,9 @@ class SensorWrapper(object):
                                                update_handler=self._handle_update_trigger,
                                                delete_handler=self._handle_delete_trigger,
                                                trigger_types=self._trigger_types,
-                                               queue_suffix='sensorwrapper')
+                                               queue_suffix='sensorwrapper_%s_%s' %
+                                               (self._pack, self._class_name),
+                                               exclusive=True)
 
         # 4. Set up logging
         self._logger = logging.getLogger('SensorWrapper.%s' %
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -69,6 +69,9 @@ in development
 * Update remote runner to include stdout and stderr which was consumed so far when a timeout
   occurs. (improvement)
 * Fix st2-self-check script to check whether to use http/https when connecting to st2, to disable Windows test by default, and to check test status correctly. (bug-fix)
+* Reduce the wait time between message consumption by TriggerWatcher to avoid latency (improvement)
+* Use exclusive messaging Qs for TriggerWatcher to avoid having to deal with old messages
+  and related migration scripts. (bug-fix)
 
 0.13.2 - September 09, 2015
 ---------------------------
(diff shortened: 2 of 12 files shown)


Kami:
Yeah, in the past the issue was TriggerWatcher not yielding and preventing sensors from running.

I believe this should work fine as well, but yeah, we definitely need to test the sensors during the same time where a lot of trigger events are being dispatched to the queue.

Also, another thing we could do if it turns out this is still causing an issue for sensors is allowing user to specify sleep interval per class instance - I can see that for webhooks and other places we really want to use a smaller delay.

st2reactor/st2reactor/container/sensor_wrapper.py
```diff
@@ -328,7 +328,9 @@ def __init__(self, pack, file_path, class_name, trigger_types,
                                                update_handler=self._handle_update_trigger,
                                                delete_handler=self._handle_delete_trigger,
                                                trigger_types=self._trigger_types,
-                                               queue_suffix='sensorwrapper')
+                                               queue_suffix='sensorwrapper_%s_%s' %
+                                               (self._pack, self._class_name),
+                                               exclusive=True)
```
Kami: Good catch :+1:

Kami: On a related note - eventually we also need to improve TriggerWatcher to utilize routing key since sensors are really only interested in the triggers which belong to that sensor. Right now we do late filtering in the sensor class which is not really efficient if there are many different sensors and triggers running.

manasdk:
> I believe this should work fine as well, but yeah, we definitely need to test the sensors during the same time where a lot of trigger events are being dispatched to the queue.

Tested.

manasdk:
> Also, another thing we could do if it turns out this is still causing an issue for sensors is allowing user to specify sleep interval per class instance - I can see that for webhooks and other places we really want to use a smaller delay.

I am not sure there is much value in sleeping if a co-operative yield works fine. I can of course add this in but not sure to what effect this would be used. Making this configurable is trivial so I would like to wait for the use-case to emerge.

st2reactor/st2reactor/container/sensor_wrapper.py
manasdk: Yes I agree with using separate routing keys. I almost wanted to go with that right but too many changes - we will have to accept another messaging setup migration at a later point.

Issue #2083, opened 2015-10-19
Digital Ocean support for All in One installer


kevinSuttle:
Just a heads up: none of the installers work on Digital Ocean. Ubuntu 14.04 droplet.

dzimine:
Thanks for the report Kevin,
were you trying 0.13 or latest aka 0.14dev? 

once we stabilize 0.14dev installer, we will test on digital ocean (or ask you to try again).

kevinSuttle:
The "All in One"  
`curl -sSL https://raw.githubusercontent.com/StackStorm/st2workroom/master/script/bootstrap-st2express | sh` 

Comes back with `sh: 1: Not: not found`.

kevinSuttle:
I should note that 0.13 works.

dzimine:
Kevin you need an elevated user, either do `sudo su`, then run the command above, or do this (note sudo): 

```
curl -sSL https://stackstorm.com/install.sh | sudo sh
```
this is shortcut to the same:
```
curl -sSL https://raw.githubusercontent.com/StackStorm/st2workroom/master/script/bootstrap-st2express | sudo sh
```

Kami:
From the error you posted it looks like `sh` shell is not available (that's possible, but not likely).

In any case, try using bash instead:

```bash
curl -sSL https://raw.githubusercontent.com/StackStorm/st2workroom/master/script/bootstrap-st2express | sudo bash
```

kevinSuttle:
Thanks, I'll give that a shot.

PR #2084, opened 2015-10-19
[docs] Change sh to bash for AIO installer instructions

armab:
Tiny doc update.

As @Kami noted here: https://github.com/StackStorm/st2/issues/2083#issuecomment-149188086 AIO installer script
https://raw.githubusercontent.com/StackStorm/st2workroom/master/script/bootstrap-st2express requires `bash`.

### DIFF
--- a/docs/source/install/all_in_one.rst
+++ b/docs/source/install/all_in_one.rst
@@ -54,7 +54,7 @@ Bring Your Own Box
 
 ::
 
-    curl -sSL https://raw.githubusercontent.com/StackStorm/st2workroom/master/script/bootstrap-st2express | sudo sh
+    curl -sSL https://raw.githubusercontent.com/StackStorm/st2workroom/master/script/bootstrap-st2express | sudo bash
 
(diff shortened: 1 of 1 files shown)


jfryman:
The 'fix' is better to change the #! In the bootstrap script to sh.

POSIX and everything. I think it's just an oversight that the bootstrap requires bash. All scripts should be Bourne Shell compat

armab:
Good to know :+1:

Kami:
Honestly, I think that requiring bash is totally fine.

It's unlikely that we will ever support a distro / OS where bash is not available by default and even if we do (e.g. FreeBSD) we should just ask user to install bash.

Kami:
Actually, I just remembered that local and remote runners both require bash so if bash is not available, user will have other issues as well :)

jfryman:
@Kami It's another :chicken: / :egg: problem.

You can always assert `sh` or some POSIX compat shell (like, god forbid, `dash`), but you cannot always assert `bash` on a system.

So, we can put Bash there, but, this is bootstrap and we have to assume lowest common denominator to get the system in a working state. That's `sh`.

jfryman:
Fixed in https://github.com/StackStorm/st2workroom/commit/3e3a45ad21c53406d1afbf691feb93e949190e2f, will be merging shortly.

PR #2087, opened 2015-10-20
Do not raise delete trigger event on an update of a rule

manasdk:
Turns out the check to validate if delete_by_reference actually deleted a TriggerDB was missing. Check brought in therefore should now fix the regression.

### DIFF
--- a/st2common/st2common/persistence/trigger.py
+++ b/st2common/st2common/persistence/trigger.py
@@ -52,15 +52,24 @@ class Trigger(ContentPackResource):
         delete_query = model_object._object_key
         delete_query['ref_count__lte'] = 0
         cls._get_impl().delete_by_query(**delete_query)
+
+        # Since delete_by_query cannot tell if teh delete actually happened check with a get call
+        # if the trigger was deleted. Unfortuantely, this opens up to races on delete.
+        confirmed_delete = False
+        try:
+            cls.get_by_id(model_object.id)
+        except ValueError:
+            confirmed_delete = True
+
         # Publish internal event on the message bus
-        if publish:
+        if confirmed_delete and publish:
             try:
                 cls.publish_delete(model_object)
             except Exception:
                 LOG.exception('Publish failed.')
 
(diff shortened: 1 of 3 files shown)


m4dcoder:
+1

st2common/st2common/persistence/trigger.py
```diff
@@ -52,15 +52,24 @@ def delete_if_unreferenced(cls, model_object, publish=True, dispatch_trigger=Tru
         delete_query = model_object._object_key
         delete_query['ref_count__lte'] = 0
         cls._get_impl().delete_by_query(**delete_query)
+
+        # Since delete_by_query cannot tell if teh delete actually happened check with a get call
```
Kami: That sucks.

Anyways, I was curious and looks like this was "improved" in newer versions of MongoDB (https://docs.mongodb.org/manual/reference/method/db.collection.remove/):

"Changed in version 2.6: The remove() returns an object that contains the status of the operation."

MongoDB 2.4 which we use is quite old so we should really consider upgrading in the near future...

st2common/st2common/persistence/trigger.py
```diff
@@ -52,15 +52,24 @@ def delete_if_unreferenced(cls, model_object, publish=True, dispatch_trigger=Tru
...
+        # if the trigger was deleted. Unfortuantely, this opens up to races on delete.
+        confirmed_delete = False
+        try:
+            cls.get_by_id(model_object.id)
+        except ValueError:
```
Kami: Is there a more specific exception we can catch?

IIRC, ValueError is also returned on invalid id and some other scenarios in which object wouldn't be deleted.

manasdk: This is the raised exception https://github.com/StackStorm/st2/blob/master/st2common/st2common/models/db/__init__.py#L126

Cannot handle anything else at this level. If I were to pick an exception to re-raise it would be StackstormDBObjectNotFound or something.

PR #2088, opened 2015-10-20
Fix launchdev script

m4dcoder:


tools/launchdev.sh
```diff
@@ -18,8 +18,8 @@ while getopts ":r:gsc" o; do
         g)
             use_gunicorn=true
             ;;
-        s)
-            skip_examples=true
+        x)
+            copy_examples=true
```
Kami: What was the problem, `-g` flag collided with `s` or something else?

Just curious because I tested `-g` and it worked for sure.

manasdk:
I guess we are inverting whether or not examples should be copied is all. Works for me :)

m4dcoder:
1) the evaluation of g when launch st2-api and st2-auth don't work so i fixed that.
2) I want to invert whether or not to copy the examples. I use launchdev most of the time and copying the examples is annoying for me since I usually symlink to ./contrib/examples that way I can make and test changes to examples.

m4dcoder:
@manasdk, do you know where we copy examples in st2cd for testing?  i think this may break st2cd.

Kami:
> I guess we are inverting whether or not examples should be copied is all. Works for me :)

Yeah, I'm fine with that as well.

Kami:
@m4dcoder Probably somewhere in st2tests repo.

LGTM, +1

DoriftoShoes:
We run this to install examples:
https://github.com/StackStorm/st2/blob/master/tools/st2-setup-examples

m4dcoder:
Thanks!

PR #2090, opened 2015-10-20
Update webui docs

enykeev:


docs/source/config/webui.rst
```diff
@@ -6,21 +6,21 @@ st2web is an Angular-based HTML5 web application. It allows you to control the w
...
 
-It can also be installed by extracting the latest tar-ball from :ops_latest:`webui` into ``/opt/stackstorm/static/webui``.
+It can also be installed by extracting the latest tar-ball from :ops_latest:`webui` into ``/opt/stackstorm/static/webui`` and serving the folder with web server of choice.
 
-st2web is a pure html5 application and consist only of js scripts, html templates, css styles and a number of static files including custom fonts and svg images. For application to work correctly, all this files should be served to the browser.
-By default they are served by python SimpleHTTPServer. For your custom deployments,
-you are encouraged to host it off nginx, Apache or a similar dedicated web server to serve those static files.
+st2web is a pure html5 application and consist only of js scripts, html templates, css styles and a number of static files including custom fonts and svg images. For application to work correctly, all this files should be served to the browser. By default they are served by nginx alongside st2installer and api endpoints. For your custom deployments, you can also use Apache or a similar dedicated web server to serve those static files.
```
dzimine: I'd drop "alongside st2installer and api endpoints"

docs/source/config/webui.rst
```diff
@@ -47,7 +47,7 @@ On a web client side, a file ``config.js`` in a root of the project which contai
 
 Multiple servers could be configured for user to pick from. To disconnect from the current server and return to login screen, pick 'Disconnect' from the drop down at the top right corner of the UI.
 
-On an |st2| side, `CORS <https://en.wikipedia.org/wiki/Cross-origin_resource_sharing>`__ should also be properly configured. In st2.conf, ``allow_origin`` property of the [api] section should contain the Origin header browser sends with every request. For example, if you have deployed UI on its own server and accessing it using url `http://webui.example.com:8000`, your config should look like that:
+On an |st2| side, `CORS <https://en.wikipedia.org/wiki/Cross-origin_resource_sharing>`__ should also be properly configured. In st2.conf, ``allow_origin`` property of the [api] section should contain the Origin header browser sends with every request. For example, if you have deployed UI on its own server and accessing it using url `http://webui.example.com`, your config should look like that:
```
dzimine: Is it so? Even when we serve via nginx, does the allow_origin still apply?

docs/source/config/webui.rst
```diff
@@ -58,7 +58,7 @@ On an |st2| side, `CORS <https://en.wikipedia.org/wiki/Cross-origin_resource_sha
    logging = st2api/conf/logging.conf
    # List of allowed origins for CORS, use when deploying st2web client
    # The URL must match the one the browser uses to access the st2web
-   allow_origin = http://webui.example.com:8000
+   allow_origin = http://webui.example.com
```
dzimine: why not https here?

dzimine:
couple of comments, other than that looks perfect.

docs/source/config/webui.rst
enykeev: We need to check this one with @jfryman as he struggled with it lately and I'm not sure where he end up. In general, having proper allow_origin would not hurt.

docs/source/config/webui.rst
enykeev: >  For example, if you have deployed UI on its own server and accessing it using url `http://webui.example.com`, your config should look like that:

enykeev: So it depends on a URL you are using to access the UI. In case of AIO, you don't need to edit it at all.

docs/source/config/webui.rst
Kami: It's on my TODO for testing - we manually injected CORS headers in nginx, but since this nginx module for injecting headers on non successful responses (e.g. 401 on auth failure) is not available in standard RHEL repos we decided to remove that.

To clarify - this only applies to non successful responses / status codes, we still correctly inject headers on all non 400 and 500 responses in Nginx.

Kami: FYI - I just tested standalone auth mode using pam backend and it looks like headers are now correctly propagated / proxied through also on non-success status codes since the correct error is displayed on invalid credentials.

jfryman: :camel:

jfryman: @emedvedev: `allow_origin` in that context will only apply if user is using the standalone daemon. Otherwise, we set `*` at https://github.com/StackStorm/st2workroom/blob/0a96e6f811c1cb3577888af71721ae9ea887b977/modules/profile/manifests/st2server.pp#L157. 

If we find we need to make this configurable for users, we can re-address. But, the value in the config is all but ignored when served by something else.

docs/source/config/webui.rst
dzimine: right!

docs/source/config/webui.rst
dzimine: Guys, a note on language: `standalone` is used all over to separate two 'auth' modes. We also need a word to separate running under WSGI app server (with nginx) from running "standalone" (pardon my language) == with simple http server embedded in pecan.

This becomes especially obvious when you look at auth docs and config files.
Please come up with something. I did a try in the upcoming PR>

manasdk:
Merging. If we need some other changes to this then PR. I don't want to miss this change due to some weird quirk of releases.

PR #2092, opened 2015-10-20
Write webhook file to home

manasdk:
* On RHEL7 writing to /tmp does not work. Not sure why but it fails causing
  test failures. Therefore writing to ~ which should always be allowed.


DoriftoShoes:
+1

PR #2093, opened 2015-10-20
Change default mode for authentication to standalone

m4dcoder:


Kami:
LGTM

It's also worth noting that this doesn't affect launchdev.sh (or pretty much anything right now) since launchdev using conf/st2.dev.conf which uses standalone auth mode.

PR #2094, opened 2015-10-21
Add support for virtual environment update

jsjeannotte:
Add support for virtual environment update (instead of full removal/creation) using setup_virtualenv.py. This reduces the likelihood of a running action failing during a pack upgrade.


contrib/packs/actions/setup_virtualenv.yaml
```diff
@@ -9,3 +9,7 @@
       type: "array"
       items:
         type: "string"
+    update:
+      type: "boolean"
+      default: false
+      description: "Check this option if the virtual environment already exists and if you only want to perform an update and installation of new dependencies. If you don't check this option, the virtual environment will be destroyed then re-created."
```
manasdk: nit : new line at end of file pls.

contrib/packs/actions/pack_mgmt/setup_virtualenv.py
```diff
@@ -44,19 +48,19 @@ def __init__(self, config=None):
         self._base_virtualenvs_path = os.path.join(cfg.CONF.system.base_path,
                                                    'virtualenvs/')
 
-    def run(self, packs):
+    def run(self, packs, update):
```
manasdk: I would assign a default value `update=False` in this code also.

manasdk:
Looks good!

Once you pick up some of my minor comments I will merge the PR.

ps : scrutinizer failed on some upstream issue - safe to ignore that failure.

Kami:
It might be better and more explicit to have a separate command to update virtualenv (e.g. packs.update_virtualenv) what do others think?

jsjeannotte:
@Kami Yes, that was my original idea, but since so much of the logic is the same, maintenance will be  easier if it's in the same action. Also, I'm thinking of adding logic do that if you call update and the virtualenv doesn't exist, it will perform the creation as well (instead of failing).

Kami:
@jsjeannotte It's a Python runner action which means you should be able to reduce duplication by refactoring common functionality in a common class in a different file (e.g. `pack_mgmy/base.py`) :)

I would still prefer for it to be a separate action (more explicit what action is doing), but if others are fine with this approach, I'm fine with it as well so I can go ahead and merge it.

/cc @manasdk @lakshmi-kannan

manasdk:
We can also have a separate action by simply using a 2nd metadata file. No python code change needed in that case.

Kami:
@manasdk Good point. Lets do this.

We can get this merged and then I can make that changes.

manasdk:
@kami +1

Kami:
Merged, thanks.

Kami:
@jsjeannotte 

Added new ``packs.update_virtualenv`` action in 1fc282fccfb64445401d9266cc6e45ec29eba3ea.

Thanks again.

jsjeannotte:
Thanks @Kami and @manasdk.

PR #2097, opened 2015-10-21
Integrate "FROM" parameter into send_mail action

pixelrebel:
If a single username is provided, the local hostname is appended.  This allows for default 'stanley' to pass a proper email address. 

Tested in st2 version 0.13.2.5.


Kami:
Thanks for the PR.

The changes look good to me.

On a related note - if you want you can also tackle footer "issue" in this PR - one option is to also add a footer argument and if not provided, we default to the same old value.

pixelrebel:
>one option is to also add a footer argument and if not provided, we default to the same old value.

This is a bit trickier because the user may want to access the action name and host name in the custom footer.  I was simply suggesting replacing "Stanley" with "StackStorm."  

Then again, I suppose it doesn't hurt to have a custom footer option with no script/host information.  What do you think?

Kami:
I'm fine with replacing Stanley with StackStorm for now :)

Kami:
Thanks, merged into master.

PR #2102, opened 2015-10-22
Move auth enable check to the hook itself

enykeev:
There's an inconsistency in handling the `x-auth-token` query param when sending it to the server with disabled auth makes /stream endpoint return 400 due to different controller signature.

### DIFF
--- a/st2api/st2api/app.py
+++ b/st2api/st2api/app.py
@@ -82,9 +82,7 @@ def setup_app(config=None):
     active_hooks = [hooks.RequestIDHook(), hooks.JSONErrorResponseHook(),
                     hooks.LoggingHook()]
 
-    if cfg.CONF.auth.enable:
-        active_hooks.append(hooks.AuthHook())
-
+    active_hooks.append(hooks.AuthHook())
     active_hooks.append(hooks.CorsHook())
 
     app = pecan.make_app(app_conf.pop('root'),
--- a/st2common/st2common/hooks.py
+++ b/st2common/st2common/hooks.py
@@ -123,13 +123,14 @@ class AuthHook(PecanHook):
                 re.search(AUTH_TOKENS_URL_REGEX, state.request.path)):
             return
 
-        user_db = self._validate_creds_and_get_user(request=state.request)
+        if cfg.CONF.auth.enable:
+            user_db = self._validate_creds_and_get_user(request=state.request)
 
-        # Store related user object in the context. The token is not passed
-        # along any longer as that should only be used in the auth domain.
-        state.request.context['auth'] = {
-            'user': user_db
-        }
+            # Store related user object in the context. The token is not passed
+            # along any longer as that should only be used in the auth domain.
+            state.request.context['auth'] = {
+                'user': user_db
+            }
 
         if QUERY_PARAM_ATTRIBUTE_NAME in state.arguments.keywords:
             del state.arguments.keywords[QUERY_PARAM_ATTRIBUTE_NAME]
--- a/st2stream/st2stream/app.py
+++ b/st2stream/st2stream/app.py
@@ -86,9 +86,7 @@ def setup_app(config=None):
     active_hooks = [hooks.RequestIDHook(), hooks.JSONErrorResponseHook(),
                     hooks.LoggingHook()]
 
-    if cfg.CONF.auth.enable:
-        active_hooks.append(hooks.AuthHook())
-
+    active_hooks.append(hooks.AuthHook())
     active_hooks.append(hooks.CorsHook())
 
     app = pecan.make_app(app_conf.pop('root'),


Kami:
I personally like the previous approach more.

Is it possible / would it be easy to change the client to simply not send this argument if auth is disabled? If auth is disabled, the token should be empty / null, so this should be possible, right?

enykeev:
This is mostly a consistency issue. We end up with a different sets of query parameters so stuff would stop working in most unexpected places. And client doesn't know whether auth is enabled or disabled, it just sends the token when it has it. It probably won't be able to obtain it, but if you provide it manually like we do in st2client, there's really nothing we can do on client side.

I really don't see any benefit in having it the way it was before this PR.

manasdk:
> I really don't see any benefit in having it the way it was before this PR.

That way the hook is never add in and we don't have unnecessary code. I am surprise that app.setup_app is not actually called and it does not remove the hook.

enykeev:
The code is necessary though. It removes query parameter and allows controller to be found during routing.

manasdk:
> The code is necessary though. It removes query parameter and allows controller to be found during routing.

What AuthHook!

Kami:
> What AuthHook!

Pretty sure if we are routing in AuthHook that is the wrong place.

Yeah, that was basically my main concern.

If this code is needed unrelated of the auth code, it should live elsewhere and not inside an auth hook (separate hook is probably fine).

manasdk:
I looked at AutHook code - nothing spooky there. (Yes I am now making halloween references)

Kami:
So what's are our plan for that? Are we going to move it in a separate hook or modify / fix the client to not send the token if the token / credentials are not provided (I prefer the later)?

enykeev:
Not sending the token making it work on clients discretion and (given, on some rare occasions) will result in code throwing errors while it might have been successfully executed.

Say, we had a server with `auth: enabled`, client authenticated against this server, got himself a token and is fully certain the server supports authentication. Now, we disable authentication on the server and it suddenly starts throwing errors that either `controller is not found` or `token is given when it should not`. The problem gets bigger since at the same time, server is perfectly fine to receive token as a header even with auth disabled (because routing is not depending on headers) so only come of controllers would fail and client still have no way to figure out what's happened because server do not expose in any way auth is disabled.

See, you may not like this few lines of code simply because they do not fit the picture you've drawn in your head, but I'm telling you that the picture is actually much bigger and scarier. You are welcome to find a better way to solve the problem, just please, please, don't shove it down to client and look away hoping the problem disappear. Because it won't.

manasdk:
Here is how I see this -

1. Auth disabled is not a real usecase we need to stop worrying in the long term and even short term to some extent.
2. Switching auth from enabled to disabled which a webui is running is not something we should support. Refreshing the browser of whatever at this point is perfectly reasonable. Also, this should almost never happen except for debugging and weird behavior when that happens is something we can take heat for - my would claim there would not be much heat.
3. What we have right now is explicit i.e. pass in token when auth is enabled and no token when auth is disabled. Passing in a token when auth is disabled and getting no feedback from the server is pretty much bad anyway. I would vote that we should in fact be rejecting such a request to keep the client-server interface clear and no surprises.

Does my line of thinking make sense in context of this PR?

manasdk:
> The problem gets bigger since at the same time, server is perfectly fine to receive token as a header even with auth disabled (because routing is not depending on headers) so only some of controllers would fail and client still have no way to figure out what's happened because server do not expose in any way auth is disabled.

This is bad. We should simply reject requests that supply and auth token when auth is disabled. This is likely cleaner and more obvious from the client standpoint. This way we know things like RBAC etc will not be working and there should be no surprises.

enykeev:
I've stubble upon it again and I'm starting to get furious. I understand that it might not be a problem for real user, but it's for sure a problem for me since I have to constantly switch between branches of st2 on my dev machine and it always overrides config with `auth=false` on st2, but still preserves `auth: true` on st2web.

I may agree with your general line of thinking, but now I'm positive that it won't solve my own problem. Instead, while rejecting tokens for `auth=false` being clean and right, it will make me do like 10 additional actions every time I switch the branch. I personally do not accept that price.

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1450199724013776

jfryman:
Just a note... I ran into this today. See above breadcrumb

manasdk:
Seems like we are still in a holding pattern here.

lakshmi-kannan:

enykeev:
The problem resurfaced on #community. I appreciate your feedback, guys, but I'm going to rebase and merge it anyway. This thing is just silly.

emedvedev:
+1. This is a problem with quite a few users, actually, and might be the main reason people don't get responses from the bot. "Nobody disables auth" is an exaggerated statement, apparently, although I was surprised to learn that as well

enykeev:
Ninja-merging while the rest of the team is asleep 😛

Kami:
Congrats on merging more than one year old PR :P

Michaeljczhang:
When I am in a datacenter environment, I don't need auth and I disabled it.
I experience a problem where all things are functioning except that hubot is not able to pick up the result from the steam and nor post to the channel.
By looking at the st2chatops.log, there is an error log. 

> [Thu Nov 17 2016 06:10:07 GMT+0000 (UTC)] ERROR Stream error: Event { type: 'error', status: 400 }

You can view the log file at https://stackstorm-community.slack.com/files/jczhang/F332KSVHP/-.pl
Please look at the [https://stackstorm-community.slack.com/archives/community/p1479362121003041](chat history)

In order to make it work, I modify /opt/stackstorm/packs/chatops/actions/workflows/post_result.yaml so that it call slack.chat.postMessage directly.

PR #2105, opened 2015-10-22
Pack permissions

manasdk:
* use chgrp to switch pack ownership to configure pack_owner group
* config to allow specification of pack owner group

### DIFF
--- a/contrib/packs/actions/pack_mgmt/download.py
+++ b/contrib/packs/actions/pack_mgmt/download.py
@@ -13,8 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import grp
 import os
 import shutil
+import subprocess
 import hashlib
 import json
 import stat
@@ -54,6 +56,9 @@ STACKSTORM_CONTRIB_REPOS = [
 # Only if you can emphatically answer 'YES' to allow about questions you should
 # touch this file. Else, be warned you might loose a part of you soul or sanity.
 #####
+
+
+PACK_GROUP_CFG_KEY = 'pack_group'
 
 
 class DownloadGitRepoAction(Action):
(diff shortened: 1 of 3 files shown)


contrib/packs/actions/pack_mgmt/download.py
```diff
@@ -145,18 +150,26 @@ def _move_packs(self, abs_repo_base, packs, abs_local_path, subtree):
...
-    @staticmethod
-    def _apply_pack_permissions(pack_path):
+    def _apply_pack_permissions(self, pack_path):
         """
         Will recursively apply permission 770 to pack and its contents.
         """
+        # 1. switch owner group to configuered group
+        pack_group = self.config.get(PACK_GROUP_CFG_KEY, None)
+        if pack_group:
+            subprocess.call(['sudo', 'chgrp', '-R', pack_group, pack_path])
```
Kami: Won't this explode if the group doesn't exist?

DoriftoShoes:
+1

contrib/packs/actions/pack_mgmt/download.py
Kami: If it will, we should leave the config empty.

And even if it's defined, we should probably throw and fail fast if an invalid / in-existent group is referenced.

Kami: Also, probably better to use `st2common.util.(green).shell.run`

manasdk: Using green utils.

Not throwing exception if group permission cannot be changed. Somehow doing all this so icons can be properly rendered seems very like a bad idea to fail an install.

manasdk: I made an attempt to deal with errors in this case but this entire reporting of this action is very very messy. I cannot raise an exception. I will re-work it later. 

It might be better to break out the permission assignment part to some other action.

PR #2107, opened 2015-10-23
Add enterprise section to auth doc

m4dcoder:


docs/source/authentication.rst
```diff
@@ -78,17 +77,60 @@ repo. The following is a sample auth section in the config file for the PAM back
...
     debug = False
 
-StackStorm developed auth backends such as LDAP are only available in the enterprise edition. For 
+The following is a list of auth backends for the community edition to help get things started.
+
+* `PAM <https://github.com/StackStorm/st2-auth-backend-pam>`_
+* `Flat File <https://github.com/StackStorm/st2-auth-backend-flat-file>`_
+* `OpenStack Keystone <https://github.com/StackStorm/st2-auth-backend-keystone>`_
+
+Auth Backends (Enterprise)
```
dzimine: I'd really love to see LDAP here in TOC. Can we make it something like LDAP (Enterprise)?

m4dcoder: This will make the auth doc structure inconsistent... When we add another enterprise backend, do we keep adding subsection to TOC?  It's probably better to have a 3rd level in the TOC...

PR #2108, opened 2015-10-23
Hard code versioning for docs

manasdk:
* Opening PR against master but will only marge to 1.1 and then to master once 1.1 docs are out. This way doc links will not explode.


manasdk:
Btw, this is pretty bad. I am not sold on how this versioning is done but I am solving the issue in the scope of what we got rather than any holisitic update.

dzimine:
ok by me as it does the job.

PR #2111, opened 2015-10-24
Fix runner type parameters override

m4dcoder:


dzimine:
lovely!

manasdk:
Surprising. Good catch

Kami:
Indeed, good catch :+1:

PR #2116, opened 2015-10-26
Disable SSL warnings as early as possible

manasdk:
* Disable SSL warnings before any code path allow the method
  to quite the moment it is established that config can be read.

Should be a fairly harmless fix and help with the Ux but by no means is this a stopper. Can possibly pick up after release.


dzimine:
Thanks @manasdk

PR #2118, opened 2015-10-26
Sensor contain should kill children processes on forceful termination

Kami:
This pull request updates sensor container so it makes sure all the children processes (sensor instances / sensor wrappers) are killed if the parent process is forcefully killed (e.g. SIGKILL, we already clean up correctly on SIGTERM and SIGINT).

This functionality is not cross-platform and Unix only, but we use best effort approach and simply don't do anything if the functionality is not available (e.g. Windows).

TODO:

- [x] Integration tests


Kami:
I've added some integration tests, but they fail on Travis (they work fine locally).

I couldn't track down the root cause (looked like a permission issue, but it's not). For now, I will make sure we skip those tests on Travis.

PR #2122, opened 2015-10-27
Use 127.0.0.1 instead of localhost

manasdk:
* Usually mistral binds to 127.0.0.1 and localhost may not
  re-direct to 127.0.0.1 therefore safer to use 127.0.0.1.


DoriftoShoes:
+1

PR #2125, opened 2015-10-27
Adjusted structure for good Nav on the left.

dzimine:
* moved things around to make key articles show in the left TOC.
* add link from config section.

@jfryman take a look. Sphinx doesn't like 2nd level headers in index, if it finds them there, than the toctree-included files go one level under, and disappear from the nav on the left. I changed the structure to show key points on the left nav with least possible harm to the story. Adjust as needed.


manasdk:
Not sure what exactly is going on here so made this gif -

toc_magic

dzimine:
@manasdk AFAIK it's a partial update, shall not happen with full build, can you `rm -rf docs/build/` and rebuild the whole docs?

jfryman:
It's fine.

PR #2128, opened 2015-10-27
Refine Installation/Vagrant

dzimine:
* Clarify separation between st2workroom and all-in-one on Vagrant box
* Take Manas' Vagrant file as example, and keep it with source.

@jfryman pls take a look. Except links fixes, these are optional adjustments. Also, while I like having a real Vagrantfile, if you people think it's a wrong place and we don't have a good place, ok to copy it literally, or revert to simplistic one.

### DIFF
--- a/docs/source/install/vagrant.rst
+++ b/docs/source/install/vagrant.rst
@@ -1,7 +1,7 @@
 Vagrant
 =======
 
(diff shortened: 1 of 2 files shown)


emedvedev:
My two cents: one possible drawback to @manasdk's Vagrantfile is that by doing `vagrant up` it tries to set up 3 machines, all 3 having the same hostname. If a user does `vagrant up` without examining the file first, this may come as an unpleasant surprise. Maybe we should strip it before giving out to the users and just have ubuntu there?

jfryman:
@dzimine I do not want to include this. `st2workroom` is our supported Vagrantfile, and the small blurb should be only for illustrative purposes.

PR #2130, opened 2015-10-27
Correct typo in CentOS 7.0 box name

lattwood:


dzimine:
Thanks for the catch!

PR #2134, opened 2015-10-27
New versions for everything

manasdk:


Kami:
New versions fur alle!

PR #2138, opened 2015-10-28
Refactor retries in mistral runner to exponential backoff

m4dcoder:


Kami:
Since this is backward incompatible change and requires "manual" update to the config, please document this in upgrade notes section of the docs.

m4dcoder:
I already added notes to the upgrade notes.

PR #2139, opened 2015-10-28
Add more mistral examples to integration test

m4dcoder:


contrib/examples/actions/workflows/mistral-env-var.yaml
```diff
@@ -0,0 +1,14 @@
...
+examples.mistral-env-var:
+    description: A basic workflow that illustrates how to get the workflow's env vars.
+    type: direct
+    output:
+        stdout: <% $.stdout %>
+    tasks:
+        task1:
+            action: core.local
+            input:
+                cmd: echo <% env().st2_execution_id %>
```
dzimine: I suggest we show the example with the URL. It's a real use case people are asking and we use broadly. Hardcoding IP is fine, they'll get it. Like `https://172.168.50.11/#/history/<% env().st2_execution_id %>/general`

This gives an idea, and answers a bunch of know-how.

contrib/examples/actions/workflows/mistral-env-var.yaml
```diff
@@ -0,0 +1,14 @@
...
+            publish:
+                stdout: <% $.task1.stdout %>
```
dzimine: I would also publish env to make it ready for an exploring user to see what else good inside env. 
`env: <% env() %> `

dzimine:
not directly relevant, but is there a reason to turn "run all mistral examples" action chain to "run some mistral examples" ?

contrib/examples/actions/workflows/mistral-env-var.yaml
m4dcoder: Sure.

contrib/examples/actions/workflows/mistral-env-var.yaml
m4dcoder: Ok.

m4dcoder:
Yes, the AC you are referring to is there to test basic mistral functionality.  All the examples are tested in itests for CICD purposes.

PR #2140, opened 2015-10-28
Make runners.sh actually work (on Jessie)

lattwood:
Debian keeps `systemctl` in `/bin`, CentOS/RHEL keeps it in `/usr/bin`. CentOS/RHEL symlinks `/bin` to `/usr/bin`, so this change enables Jessie compatibility.


jfryman:
:cool:

PR #2141, opened 2015-10-29
Make aliasexecution endpoint return 201

enykeev:
NO MERGE

### DIFF
--- a/st2api/st2api/controllers/v1/aliasexecution.py
+++ b/st2api/st2api/controllers/v1/aliasexecution.py
@@ -23,6 +23,7 @@ from st2common.models.api.base import jsexpose
 from st2common.exceptions.db import StackStormDBObjectNotFoundError
 from st2common.models.api.action import AliasExecutionAPI
 from st2common.models.api.auth import get_system_username
+from st2common.models.api.execution import ActionExecutionAPI
 from st2common.models.db.liveaction import LiveActionDB
 from st2common.models.db.notification import NotificationSchema, NotificationSubSchema
 from st2common.models.utils import action_alias_utils, action_param_utils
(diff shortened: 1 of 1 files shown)


st2api/st2api/controllers/v1/aliasexecution.py
```diff
@@ -45,7 +46,7 @@
 
 class ActionAliasExecutionController(rest.RestController):
 
-    @jsexpose(body_cls=AliasExecutionAPI, status_code=http_client.OK)
+    @jsexpose(body_cls=AliasExecutionAPI, status_code=http_client.CREATED)
```
lakshmi-kannan: CREATED seems like the right code.

Kami: Yeah POST /executions endpoint also returns 201.

st2api/st2api/controllers/v1/aliasexecution.py
```diff
@@ -87,7 +88,7 @@ def post(self, payload):
                                              notify=notify,
                                              context=context)
 
-        return str(execution.id)
+        return execution
```
Kami: Yeah, that's not consistent with /executions endpoint as well.

Thanks.

Kami:
Will we also need to update / make related changes to the hubot plugin?

Kami:
Thanks.

So far the changes look good to me (since this is backward incompatible change we will also need to update upgrade notes, maybe also hubot-stackstorm plugin and add tests, but I assume you are already aware of this :))

st2api/st2api/controllers/v1/aliasexecution.py
manasdk: I am happy to see this changed but what is the usecase? e.g. why do we need more than an execution-id at this point?

The need for consistency with /executions is not obvious to me - can you guys explain why it is desired?

st2api/st2api/controllers/v1/aliasexecution.py
manasdk: +1

enykeev:
@Kami yeah, actually, https://github.com/StackStorm/hubot-stackstorm/pull/49 already supports both ways.

emedvedev:
Included in #2229 (along with updated tests for 201), closing.

PR #2142, opened 2015-10-29
Add bindir to st2actions for debian packages

lattwood:
Same change as in StackStorm/st2#2077


jfryman:
:+1: 

Let's get another set of :eyes:, but shouldn't be an issue.

lattwood:
Whoops, I'll move that commit to it's own branch and reopen @jfryman .

PR #2148, opened 2015-10-30
Add notification route to action context

enykeev:
This would allow us to catch relevant executions on hubot side through Stream API without waiting for `hubot.post_results`.

### DIFF
--- a/st2api/st2api/controllers/v1/aliasexecution.py
+++ b/st2api/st2api/controllers/v1/aliasexecution.py
@@ -82,6 +82,9 @@ class ActionAliasExecutionController(rest.RestController):
             'source_channel': payload.source_channel
         }
 
+        if hasattr(payload, 'notification_route'):
+            context['route'] = getattr(payload, 'notification_route')
+
         execution = self._schedule_execution(action_alias_db=action_alias_db,
                                              params=execution_parameters,
                                              notify=notify,


st2api/st2api/controllers/v1/aliasexecution.py
```diff
@@ -82,6 +82,9 @@ def post(self, payload):
             'source_channel': payload.source_channel
         }
 
+        if hasattr(payload, 'notification_route'):
+            context['route'] = getattr(payload, 'notification_route')
```
Kami: Would probably be better to just always include this attribute and just default it to `None` if no route is specified (`context['route'] = getattr(payload, 'notification_route', None)`).

This way it's more explicit and easier to do Jinja comparisons since the attribute is always present.

Kami: It would also be nice to update docs for that (we already have a small section which describes ChatOps specific context attributes).

Besides that, LGTM

lakshmi-kannan: +1 to kami's comment.

lakshmi-kannan:
@enykeev can you make the docs change and merge this?

enykeev:
Actually, I don't need that anymore. It was the way to enable chatops before announcements and I don't see any other use for that. I can indeed update the docs if we find it a new purpose.

lakshmi-kannan:
Closing then.

PR #2149, opened 2015-10-30
Add bindir to st2actions for debian packages

lattwood:
Same change as in StackStorm/st2#2077.

Accidentally added a commit I shouldn't have to #2142 , this rectifies that.


DoriftoShoes:
LGTM.  Thanks!

Issue #2144, opened 2015-10-29
Better support for SSH Bastion Hosts


jfryman:
In scenarios where a user is required to have a bastion host to access private resources (for example, AWS VPCs), then it can be cumbersome to manage the security of these nodes. Where do the keys live? Is there a good way to actually keep a wall of security versus having to find awkward ways to punch holes in security design.

Something to consider, as a request from @lattwood in community.

lakshmi-kannan:
Closing! Looks like this was addressed by @lattwood

PR #2150, opened 2015-10-30
Rough SSH Bastion Host Support (relates to #2144)

lattwood:
- Added `ParamikoSSHBastionClient` as a subclass of `ParamikoSSHClient`, with overridden `connect` and `close` methods for connection bouncing.
- Use `ParamikoSSHBastionClient` if a `!` exists in the host. The left side of the `!` is the bastion host, right side is the target host.
- All connection parameters for the target host are used for connecting to the bastion host. See https://github.com/lattwood/st2/blob/7cf7ba0938d23ddba29ce6c55d39ac6e49f7aaea/st2actions/st2actions/runners/ssh/paramiko_ssh_bastion.py#L43-L51

Rudimentary testing was done with https://gist.github.com/lattwood/4c52ba7b8583ae4565eb, I'll be doing more extensive testing using `st2workroom` today. This touches on #2144 

This is the most python I've written so I apologize in advance if anything isn't done the python way.

Is a CLA required or would Apache licensing suffice?


st2actions/st2actions/runners/ssh/paramiko_ssh_bastion.py
```diff
@@ -0,0 +1,132 @@
...
+                         '_timeout': self.timeout}
+
+        self.logger.debug('Connecting to bastion host', extra=bastion_extra)
+
+        self.bastion_client.connect(**bastion_conninfo)
+
+        transport = self.bastion_client.get_transport()
+        real_addr = (self.hostname, self.port)
+        # no attempt is made to bind local_addr, so something very "wrong" is used
+        local_addr = ('256.256.256.256', 65536)
```
lattwood: I wasn't 100% sure on what values to use here. Originally I had 127.0.0.1:1234, but since it doesn't actually bind the socket, I thought it made sense to use something that was completely outside of spec.

st2actions/st2actions/runners/ssh/paramiko_ssh_bastion.py
```diff
@@ -0,0 +1,132 @@
...
+        channel = transport.open_channel("direct-tcpip", real_addr, local_addr)
+
+        conninfo = {'hostname': '256.256.256.256',
+                    'port': 65536,
+                    'sock': channel,
```
lattwood: By passing a socket to ParamikoSSH, it doesn't go through the motions of opening the socket. An alternative could have been to add a `sock` parameter to `ParamikoSSHClient`, and make opening a socket via a bastion host the responsibility of the caller.

st2actions/packaging/debian/install
```diff
@@ -1,4 +1,5 @@
 st2actions usr/lib/python2.7/dist-packages/
 st2actions usr/lib/python2.7/dist-packages/
 conf/* etc/st2actions/
+bin usr/lib/python2.7/dist-packages/st2actions/
```
manasdk: I guess this change is part of https://github.com/StackStorm/st2/pull/2149

lattwood: I screwed up, it is. Forgot to switch branches back to master. I'll rebase my fork's branch and remove it.

manasdk: Np.

manasdk:
I am taking a look at this PR but we should wait for @lakshmi-kannan and @Kami.

st2actions/st2actions/runners/ssh/parallel_ssh.py
```diff
@@ -216,6 +218,12 @@ def _execute_in_pool(self, execute_method, **kwargs):
         return results
 
     def _connect(self, host, results, raise_on_any_error=False):
+        bastion_hostname, bastion_port = None, None
+
+        if "!" in host:
```
manasdk: Typically breaking out magic string as a constant helps. 

Probably place it somewhere https://github.com/StackStorm/st2/blob/master/st2common/st2common/constants/runners.py

st2actions/st2actions/runners/ssh/paramiko_ssh_bastion.py
```diff
@@ -0,0 +1,132 @@
...
+
+        self.bastion_client.connect(**bastion_conninfo)
+
+        transport = self.bastion_client.get_transport()
+        real_addr = (self.hostname, self.port)
+        # no attempt is made to bind local_addr, so something very "wrong" is used
+        local_addr = ('256.256.256.256', 65536)
+        channel = transport.open_channel("direct-tcpip", real_addr, local_addr)
+
+        conninfo = {'hostname': '256.256.256.256',
```
manasdk: can use local_addr[0] and local_addr[1]

lattwood: That makes sense from a logging perspective.

This made me realize the `AutoAddPolicy` the ParamikoSSH client uses will end up erroring if you have two hosts with the same IP behind the bastion host- I'll change the policy to `WarningPolicy`.

lattwood: Update: nevermind, it appears as though it's not saved on disk as no host_keys_file is specified on the clients.

manasdk:
High level this is great. most of my comments are really cosmetic things.

Few thoughts -
1. Instead of overriding the hosts property perhaps a `bastion_host` property would make sense on the runner?
2. Perhaps a special runner make more sense and keep things explicit for users. This could be a consumption hinderance so maybe this is not such a good idea.

manasdk:
Also, perhaps a good idea to add some units -

https://github.com/StackStorm/st2/blob/master/st2actions/tests/unit/test_parallel_ssh.py

lattwood:
I was thinking about a bastion_host property, but my understanding is that it would require the modification of any actions using it to add the bastion_host(s) parameter. I would like to make it explicit, but that's a large drawback.

If there is a way to implicitly add the bastion_host(s) parameter to all actions supporting the `remote-shell-cmd`, how do you run a single command on mixed groups of hosts? (ie, how do you indicate hosts a,b,c are behind bastion1, d,e,f are behind bastion2, and g,h are directly connectable)

manasdk:
> but my understanding is that it would require the modification of any actions using it to add the bastion_host(s) parameter.

Actually if we make bastion_host(s) an optional runner_parameter then all actions get it for free.

> how do you run a single command on mixed groups of hosts?

This is a tricky one. Unless we come up with a `host_group` concept this will not work very nicely. For the near term I would suggest the for each execution we only allow 1 bastion host i.e. all hosts for that execution would be assumed to be behind the same bastion. 

To run against multiple groups would require multiple executions. If practically it turns out to be the common case then we can look at options as the next step.

lattwood:
@manasdk much cleaner this way, just need to figure out unit tests

st2actions/st2actions/runners/ssh/paramiko_ssh.py
```diff
@@ -111,6 +111,11 @@ def __init__(self, hostname, port=22, username=None, password=None,
         self.client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
         self.logger = logging.getLogger(__name__)
         self.sftp = None
+        self.bastion_host = bastion_host
```
lakshmi-kannan: self._bastion_host. Can you also use the _ convention everywhere consistently?

st2actions/st2actions/runners/ssh/paramiko_ssh.py
```diff
@@ -144,6 +149,9 @@ def connect(self):
                  '_username': self.username, '_timeout': self.timeout}
         self.logger.debug('Connecting to server', extra=extra)
 
+        if self.bastion_host and self._bastion_connect():
```
lakshmi-kannan: self._bastion_host

st2actions/st2actions/runners/ssh/paramiko_ssh.py
```diff
@@ -514,6 +524,51 @@ def _get_pkey_object(self, key_material):
 
         raise paramiko.ssh_exception.SSHException(msg)
 
+    def _bastion_connect(self):
```
lakshmi-kannan: Once you get this to work, please refactor as there is duplication in code with connect() method.

st2actions/st2actions/runners/ssh/paramiko_ssh.py
```diff
@@ -144,6 +149,9 @@ def connect(self):
                  '_username': self.username, '_timeout': self.timeout}
         self.logger.debug('Connecting to server', extra=extra)
 
+        if self.bastion_host and self._bastion_connect():
+            conninfo['sock'] = self.bastion_socket
```
lakshmi-kannan: I saw your comment but I am not sure how this really works. We should be connecting to the end host and running the command on the end host. This means proxying needs to happen on the bastion host. I saw your test cases and it looks like your tests cover only localhost. Maybe you can test this with hosts != localhost as end hosts? I'll look at paramiko too to see if they support bastion directly.

lakshmi-kannan: It looks like fabric does the same thing. https://github.com/fabric/fabric/commit/c2a9bbfd50f560df6c6f9675603fb405c4071cad

In any case, I'd still like some testing with non localhosts as end hosts. Also, since we set environment variables like ST2_AUTH_TOKEN and user provided ones, it would be good to validate those are available in the end host with the bastion setup.

st2actions/st2actions/runners/ssh/paramiko_ssh.py
```diff
@@ -514,6 +524,51 @@ def _get_pkey_object(self, key_material):
...
+                         '_timeout': self.timeout}
+
+        self.logger.debug('Connecting to bastion host', extra=bastion_extra)
+
+        self.bastion_client.connect(**bastion_conninfo)
+
+        transport = self.bastion_client.get_transport()
+        real_addr = (self.hostname, self.port)
+        # no attempt is made to bind local_addr, so something very "wrong" is used
+        local_addr = ('256.256.256.256', 65536)
```
lakshmi-kannan: https://github.com/fabric/fabric/commit/c2a9bbfd50f560df6c6f9675603fb405c4071cad you can probably get away with '' and port 0 like here. I'd use the same thing so we can always refer back to fabric. Wdyt?

st2actions/st2actions/runners/ssh/paramiko_ssh.py
lattwood: I didn't use `_bastion_host` as the other properties on `ParamikoSSHClient` aren't prefixed.

st2actions/st2actions/runners/ssh/paramiko_ssh.py
lattwood: See: https://github.com/StackStorm/st2/pull/2150#discussion_r43634831

st2actions/st2actions/runners/ssh/paramiko_ssh.py
lattwood: `ParamikoSSH` attempts to open a socket to the target SSH server if a `sock` parameter isn't given. If it's given a `sock` parameter, it won't attempt to connect and will use that socket blindly.

https://github.com/paramiko/paramiko/blob/master/paramiko/client.py#L279

The intent behind the code is to connect to the bastion host, open a `Channel` to the host behind the bastion host, and pass that to the 'real' ssh client getting used.

http://docs.paramiko.org/en/1.15/api/client.html#paramiko.client.SSHClient.connect

It's the equivalent of
```
ssh -L 1234:realhost:22 stanley@bastionhost
ssh stanley@localhost -p 1234
```
Instead of
```
ssh stanley@realhost
```
Because the original connection is occurring via a tunnelled ssh connection, environment variables are being preserved

Any pointers on how those integration tests would be setup/how I can run them locally?

st2actions/st2actions/runners/ssh/paramiko_ssh.py
lattwood: I had thought about this, but the solution I came up with wasn't very elegant- it involved adding boolean parameters to a shared `_connect` method that `connect` would call twice, depending on whether a bounced connection was required. I'll take a stab at it but I don't think it would help with code readability.

st2actions/st2actions/runners/ssh/paramiko_ssh.py
lattwood: I see Paramiko is the default client going forward, are there plans for deprecating the use of fabric?

st2actions/st2actions/runners/ssh/paramiko_ssh.py
```diff
@@ -120,31 +122,15 @@ def connect(self):
...
-        self.client.connect(**conninfo)
+        if self.bastion_host:
+            self.logger.debug('Bastion host specified, connecting')
+            self.bastion_client = self._connect(self.bastion_host)
+            transport = self.bastion_client.get_transport()
+            real_addr = (self.hostname, self.port)
+            local_addr = ('', 0)
+            self.bastion_socket = transport.open_channel('direct-tcpip', real_addr, local_addr)
+
+        self.client = self._connect(self.hostname, self.bastion_socket)
```
lakshmi-kannan: self._connect(host=self.hostname, socket=self.bastion_socket) would be better.

st2actions/st2actions/runners/ssh/paramiko_ssh.py
lakshmi-kannan: This method seems not needed now?

st2actions/st2actions/runners/ssh/paramiko_ssh.py
lakshmi-kannan: Fabric will deprecated soon. We still have it to make sure we didn't miss any feature in fabric that hasn't been ported over to.

Kami: Yep, I think Paramiko client is stable enough now (we have been using it ourselves for a while now and we fixed a bunch of bugs) so we can completely remove Fabric based runner in the next release.

st2actions/st2actions/runners/ssh/paramiko_ssh.py
```diff
@@ -120,31 +122,15 @@ def connect(self):
...
-                 '_username': self.username, '_timeout': self.timeout}
-        self.logger.debug('Connecting to server', extra=extra)
-
-        self.client.connect(**conninfo)
+        if self.bastion_host:
+            self.logger.debug('Bastion host specified, connecting')
+            self.bastion_client = self._connect(self.bastion_host)
+            transport = self.bastion_client.get_transport()
+            real_addr = (self.hostname, self.port)
+            local_addr = ('', 0)
```
lakshmi-kannan: If it's not too much hassle, could you please include the link to the fabric commit as a link? This might throw off other people when they read the code. Some helpful hinting would be nice IMO.

lakshmi-kannan:
It's looking much better now in terms of code and tests. 

* Docs - Please add some info about bastion host in ssh docs. https://github.com/StackStorm/st2/blob/master/docs/source/troubleshooting/ssh.rst

* Changelog - Add an entry to CHANGELOG with your name in the "ind evelopment " section. https://github.com/StackStorm/st2/blob/master/CHANGELOG.rst

One other thing I thought was if bastion host had a non standard SSH port, we'll need to add ability to handle that. But we'll make the person requiring the feature to add that ;). 

Thanks for being patient. We'll merge this soon!

st2actions/st2actions/runners/ssh/paramiko_ssh.py
lattwood: I'll do the same on line 127 as well.

lattwood:
@lakshmi-kannan something something security something obscurity :)

lakshmi-kannan:
LGTM. Merging now.

saisandeepachanta:
@lakshmi-kannan or @lattwood  I has started using stack storm recently. My environment is accessed by only one jump box. My stack storm is installed on a new box where i have passwordless connectivity to jump server. i have to use my jump server as bastion host and then trigger my scripts on target hosts. can you please post me some examples how to use or configure the bastion host details in stackstorm. i desperately need your help.

PR #2154, opened 2015-10-30
Adding info on action_context variables set by action aliases

jonico:
This info is part of the action documentation but I could not find it here and only learned about it after asking in the community slack channel.

### DIFF
--- a/docs/source/chatops/aliases.rst
+++ b/docs/source/chatops/aliases.rst
@@ -133,6 +133,13 @@ In this the action google.get_search_results would be called with the parameters
        query: StackStorm
        limit: 10
 
+Additional ChatOps parameters passed to the command
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+An execution triggered via chatops will contain variables such as ``action_context.api_user``, ``action_context.user`` and ``action_context.source_channel``. ``api_user`` is the user who's kicking off the chatops command from
+client and ``user`` is the |st2| user configured in hubot. ``source_channel`` is the channel
+in which the chatops command was kicked off.
+
 Multiple formats in single alias
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


docs/source/chatops/aliases.rst
```diff
@@ -165,3 +172,5 @@ ChatOps
 To see how to use aliases with your favorite Chat client and implement ChatOps in your infrastructure
 go `here <https://github.com/StackStorm/st2/blob/master/instructables/chatops.md>`_.
 
+
```
manasdk: nit : extra couple of lines

manasdk:
Great. Thanks!

docs/source/chatops/aliases.rst
```diff
@@ -133,6 +133,13 @@ In this the action google.get_search_results would be called with the parameters
        query: StackStorm
        limit: 10
 
+Additional ChatOps parameters passed to the command
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+An execution triggered via chatops will contain variables such as ``action_context.api_user``, ``action_context.user`` and ``action_context.source_channel``. ``api_user`` is the user who's kicking off the chatops command from
```
Kami: Since this is already documented in docs/source/actions.rst it might be better to just link there or maybe refactor it in a single file and use an include directive :)

I'm all for making it user-friendly and repeating it as much as needed, but if it's repeated "manually" (e.g. not auto generating it and / or using include directive) it's easy for such information to get out of sync with the code (and documentation which is incorrect is usually worse then no documentation).

One improvement we could do in the future is to add some documentation for each attribute in the code and then auto-generate a table with each attribute description and include it in the documentation (we already do this for runner parameters, RBAC permission types, etc.).

Kami: Opened a ticket for a future improvement https://github.com/StackStorm/st2/issues/2156

For now, that works and what manas has said - thanks for the contribution!

PR #2155, opened 2015-10-30
Add mistral WF state and state_info to the query result

m4dcoder:


st2actions/st2actions/query/mistral/v2.py
```diff
@@ -115,9 +113,13 @@ def _get_workflow_result(self, exec_id):
         """
         execution = executions.ExecutionManager(self._client).get(exec_id)
 
-        output = jsonify.try_loads(execution.output) if execution.state in DONE_STATES else None
+        result = jsonify.try_loads(execution.output) if execution.state in DONE_STATES else {}
+        result['workflow'] = {
```
Kami: This attribute is now Mistral workflow specific, right?

I'm fine with that change, maybe we should also make this a convention and put all the Mistral workflow specific stuff / data in the `workflow` attribute (in the past, we did our best to unify the result output for action chain and Mistral workflows, but there will always be cases where we can't do that).

st2actions/st2actions/query/mistral/v2.py
```diff
@@ -134,15 +136,16 @@ def _format_task_result(self, task):
...
-        result['state'] = task.get('state', None)
+        result = {
+            'id': task['id'],
+            'name': task['name'],
+            'workflow_execution_id': task.get('workflow_execution_id', None),
+            'workflow_name': task['workflow_name'],
+            'created_at': task.get('created_at', None),
+            'updated_at': task.get('updated_at', None),
+            'state': task.get('state', None),
+            'state_info': task.get('state_info', None)
```
Kami: Hm, as mentioned above and to make it a bit more obvious which attributes are Mistral specific maybe we should put `state_info` in `extra` dictionary (result['extra']['state_info']) or similar?

st2actions/st2actions/query/mistral/v2.py
```diff
@@ -115,9 +113,13 @@ def _get_workflow_result(self, exec_id):
         """
         execution = executions.ExecutionManager(self._client).get(exec_id)
 
-        output = jsonify.try_loads(execution.output) if execution.state in DONE_STATES else None
+        result = jsonify.try_loads(execution.output) if execution.state in DONE_STATES else {}
+        result['workflow'] = {
+            'state': execution.state,
```
lakshmi-kannan: Is it added to be more explicit when debugging what's the mistral state?

m4dcoder: Yes, sometimes when Mistral fails at the WF level (not task level), it'll put additional state information in state_info.  We should pull that info over for the user so when the WF fails, the info is available.

st2actions/st2actions/query/mistral/v2.py
m4dcoder: I'm trying not to break compatibility. Right now, the WF output parameters are directly under the WF result.  I don't want to move those output parameters to another nested level.  I also don't want to have name conflicts either so I move the WF state and state_info under result['workflow'].  How about if we do something like

result = {
    'output1': 'value1',
    'output2': 'value2',
    ...
    'extra': {
        'state': 'SUCCESS',
        'state_info': None,
        'tasks': [
            {
                'name': 'task1',
                'state': 'SUCCESS',
                'state_info': None,
                ...
            },
            ...
        ]
    }
}

st2actions/st2actions/query/mistral/v2.py
```diff
@@ -115,9 +109,14 @@ def _get_workflow_result(self, exec_id):
         """
         execution = executions.ExecutionManager(self._client).get(exec_id)
 
-        output = jsonify.try_loads(execution.output) if execution.state in DONE_STATES else None
+        result = jsonify.try_loads(execution.output) if execution.state in DONE_STATES else {}
 
-        return (execution.state, output)
+        result['extra'] = {
```
Kami: Yeah, that looks better.

st2actions/st2actions/query/mistral/v2.py
```diff
@@ -74,7 +74,7 @@ def query(self, execution_id, query_context):
             raise
 
         try:
-            tasks = self._get_workflow_tasks(mistral_exec_id)
+            result['extra']['tasks'] = self._get_workflow_tasks(mistral_exec_id)
```
Kami: Slightly confused - why do we also put this in extra?

I believe this will break existing "unified" tasks format.

Kami: Besides that LGTM, but I would like to get this sorted out.

m4dcoder: I want to keep the WF output separate from all the noises. This includes the task lists/details.

Kami: Can we do this part of a separate PR / later?

Since this is backward incompatible and I believe it will also affect the current CLI output...

PR #2159, opened 2015-11-02
WIP: Add ability to specify glob patterns for skip_notify

lakshmi-kannan:


st2common/st2common/util/glob_match.py
```diff
@@ -0,0 +1,31 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import re
+
+
+def matches(string, pattern):
```
manasdk: This method is stupid if it basically only calls the same util method in the re package with the same argument semantics.

lakshmi-kannan: I agree. @Kami wanted something like this. I am not really clear what he wants.

Kami: Yeah, I didn't want that.

As far as globing support goes, I think we should keep it simple and avoid regexes (simple `*` support for a substring should be sufficient imo).

Kami: To clarify what I wanted - I think we should support same globing notation everywhere (this means we should be able to re-use the same approach when we add support for RBAC datastore item permissions + globs).

PR #2161, opened 2015-11-02
Remove newline characters in sendmail

pixelrebel:
In text/plain version, the newline characters show literally in the body of the email.  This solution achieves the desired "double-spaced" formatting.

### DIFF
--- a/contrib/core/actions/send_mail/send_mail
+++ b/contrib/core/actions/send_mail/send_mail
@@ -30,7 +30,7 @@ BODY="$@"
 if [[ "${CONTENT_TYPE}" = "text/html" ]]; then
   LINE_BREAK="<br><br>"
 else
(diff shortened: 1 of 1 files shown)


lakshmi-kannan:
It seems like just the \n characters were removed and nothing else was done. Does this really fix the problem in text version?

pixelrebel:
Honestly I don't know why it works this way.  It seems that the newline characters should convert properly into line spaces, but they don't.  Instead, the line spaces pass from the newlines inside the `EOF` stanza.

Before the fix:
```
Email body.
\n\n
Footer
```
After the fix
 ```
Email body.

Footer
```

lakshmi-kannan:
Bizarre!

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1450472897016277

PR #2164, opened 2015-11-03
Introduce Announcements

enykeev:
An attempt to generalize the idea of st2 posting a message to stream consumer.

It is intended to be drop-in replacement for `hubot.post_results` that will allow us to switch to using streams in hubot without redesigning the whole pipeline top to bottom.

It is also intended to be used as a general way of sending notifications to the public as opposed to `Notification` mechanic for reacting on events within st2 internally. Currently internal Notifications becoming external Announcements by means of rules. Announcements could also be called directly as a part of workflow, same way `hubot.post_message` and `slack.post_message` work right now.

### DIFF
--- a/st2api/st2api/listener.py
+++ b/st2api/st2api/listener.py
@@ -21,7 +21,7 @@ from oslo_config import cfg
 
 from st2common.models.api.action import LiveActionAPI
 from st2common.models.api.execution import ActionExecutionAPI
-from st2common.transport import liveaction, execution, publishers
+from st2common.transport import announcement, liveaction, execution, publishers
 from st2common.transport import utils as transport_utils
 from st2common import log as logging
 
@@ -31,11 +31,6 @@ __all__ = [
 ]
 
 LOG = logging.getLogger(__name__)
-
-QUEUE = Queue(None,
-              liveaction.LIVEACTION_XCHG,
-              routing_key=publishers.ANY_RK,
-              exclusive=True)
 
 _listener = None
 
@@ -49,6 +44,11 @@ class Listener(ConsumerMixin):
 
     def get_consumers(self, consumer, channel):
         return [
+            consumer(queues=[announcement.get_queue(routing_key=publishers.ANY_RK,
+                                                    exclusive=True)],
+                     accept=['pickle'],
+                     callbacks=[self.processor()]),
+
             consumer(queues=[execution.get_queue(routing_key=publishers.ANY_RK,
                                                  exclusive=True)],
                      accept=['pickle'],
@@ -62,14 +62,17 @@ class Listener(ConsumerMixin):
                      callbacks=[self.processor(LiveActionAPI)])
         ]
 
-    def processor(self, model):
+    def processor(self, model=None):
         def process(body, message):
             from_model_kwargs = {'mask_secrets': cfg.CONF.api.mask_secrets}
             meta = message.delivery_info
             event_name = '%s__%s' % (meta.get('exchange'), meta.get('routing_key'))
 
             try:
-                self.emit(event_name, model.from_model(body, **from_model_kwargs))
+                if model:
+                    body = model.from_model(body, **from_model_kwargs)
+
+                self.emit(event_name, body)
             finally:
                 message.ack()
 
--- a/st2common/st2common/bootstrap/runnersregistrar.py
+++ b/st2common/st2common/bootstrap/runnersregistrar.py
@@ -405,6 +405,20 @@ RUNNER_TYPES = [
         'description': 'A runner for emitting an announcement event on the stream.',
         'enabled': True,
         'runner_parameters': {
+            'experimental': {
+                'description': 'Flag to indicate acknowledment of using experimental runner',
+                'type': 'boolean',
+                'required': True,
+                'default': False
+            },
+            'route': {
+                'description': ('The routing_key used to route the message to consumers. '
+                                'Might be a list of words, delimited by dots.'),
+                'type': 'string',
+                'default': 'general',
+                'minLength': 1,
+                'maxLength': 255
+            }
         },
         'runner_module': 'st2actions.runners.announcementrunner'
     },
--- a/st2common/st2common/bootstrap/runnersregistrar.py
+++ b/st2common/st2common/bootstrap/runnersregistrar.py
@@ -400,6 +400,15 @@ RUNNER_TYPES = [
 
     # Experimental runners below
     {
+        'name': 'announcement',
+        'aliases': [],
+        'description': 'A runner for emitting an announcement event on the stream.',
+        'enabled': True,
+        'runner_parameters': {
+        },
+        'runner_module': 'st2actions.runners.announcementrunner'
+    },
+    {
         'name': 'windows-cmd',
         'aliases': [],
         'description': 'A remote execution runner that executes commands'
--- a/st2common/st2common/transport/announcement.py
+++ b/st2common/st2common/transport/announcement.py
@@ -37,19 +37,19 @@ class AnnouncementPublisher(object):
 
 class AnnouncementDispatcher(object):
     """
-    This trigger dispatcher dispatches trigger instances to a message queue (RabbitMQ).
+    This announcement dispatcher dispatches announcements to a message queue (RabbitMQ).
     """
 
     def __init__(self, logger=LOG):
         self._publisher = AnnouncementPublisher(urls=transport_utils.get_messaging_urls())
         self._logger = logger
 
-    def dispatch(self, routing_key, payload=None, trace_context=None):
+    def dispatch(self, routing_key, payload, trace_context=None):
         """
         Method which dispatches the announcement.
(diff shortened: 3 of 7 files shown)


st2common/st2common/transport/announcement.py
```diff
@@ -0,0 +1,74 @@
...
+    def __init__(self, urls):
+        self._publisher = publishers.PoolPublisher(urls=urls)
+
+    def publish(self, payload=None, routing_key=None):
+        self._publisher.publish(payload, ANNOUNCEMENT_XCHG, routing_key)
+
+
+class AnnouncementDispatcher(object):
+    """
+    This trigger dispatcher dispatches trigger instances to a message queue (RabbitMQ).
```
lakshmi-kannan: Copy pasta

Kami:
To recap my Slack feedback - I see the need for this functionality (dispatching a message which is then fanned out to all the consumers on the /stream endpoint), but I'm personally not a fan of action runner approach.

I think we should expose this as an API (e.g. POST to /v1/stream). If we go with the API approach we can still create a wrapper Python runner action which simply sends data to that endpoint so the end result is the same.

enykeev:
For API approach, I really wonder why do you think this is a good idea. I remember the times when some of st2 internal components used REST for communication and some other were starting to use MQ. You've guys actually spend a noticeable amount of effort to make sure we now have a single channel of communication between internal components. Why do you think is a good idea for one internal component (action runner in this case) to talk to another internal component (st2api where stream resides) by means of REST calls instead of MQ? We then would have to protect this publicly accessible endpoint from actors willing to impersonate action runner and send a message to our users and this would require us to acquire an API token for every action runner. Do we want to go that far?

Kami:
I'm not talking about the internal communication method, but how should an external, user visible API look like.

In this case, I think API is better and fits this model better than the runner approach. And we can still have Python runner action so as mentioned above, end result will be the same.

It's true that REST API approach adds some overhead, but it presents a better API, imo.

As far as RBAC and permissions go - if we go with the API approach we could simply add a new permission type (send_stream_message or similar).

As far as token goes - we already generate a temporary auth token which is available for duration of action execution and is passed to every runner action. Right now this token belongs to a system user, but eventually, we will tie it to the user which triggered the action (and as such, inherit those user permissions).

enykeev:
@Kami I'm sorry, but I still fail to see how it better fits this model when action runner needs to call st2 api directly instead of routing the message through the proper channels (which we by the way already using for live updates and simply can't stop using). I would have made sense if we were expecting external services to post an announcement in st2 stream, but I don't see why we may want to do that outside the loop "webhook-rule enforcement-announcement".

We are probably having very different models in mind.

st2common/st2common/transport/announcement.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+
+    def __init__(self, logger=LOG):
```
manasdk: any reason for passing in logger

st2common/st2common/transport/announcement.py
```diff
@@ -0,0 +1,74 @@
...
+
+# Exchange for Announcements
+ANNOUNCEMENT_XCHG = Exchange('st2.announcement', type='topic')
+
+
+class AnnouncementPublisher(object):
+    def __init__(self, urls):
+        self._publisher = publishers.PoolPublisher(urls=urls)
+
+    def publish(self, payload=None, routing_key=None):
```
manasdk: no payload and no routing key seems problematic.

enykeev: No routing key is problematic (and I'm pretty sure this is not the only place where it happens in our code since I basically copypasted it before I realized messages just doesn't hit the queue with `None` as a routing key.

Empty payload might be considered a ping, but yeah, no reason to sacrifice the clarity with this one.

st2common/st2common/transport/announcement.py
enykeev: once again, copypasta, but I'll check and remove it if there's no use.

st2common/st2common/bootstrap/runnersregistrar.py
```diff
@@ -400,6 +400,30 @@
 
     # Experimental runners below
     {
+        'name': 'announcement',
+        'aliases': [],
+        'description': 'A runner for emitting an announcement event on the stream.',
+        'enabled': True,
+        'runner_parameters': {
+            'experimental': {
```
manasdk: Any reason you believe this ack is important? I guess we could simply not document this and by obfuscation it would be experimental.

st2api/st2api/listener.py
```diff
@@ -49,6 +44,11 @@ def __init__(self, connection):
 
     def get_consumers(self, consumer, channel):
         return [
+            consumer(queues=[announcement.get_queue(routing_key=publishers.ANY_RK,
```
manasdk: an exclusive Q without a unique name is asking for trouble. Anyway this is not a problem for your change specifically so I will take care of it later.

st2common/st2common/transport/announcement.py
```diff
@@ -0,0 +1,74 @@
...
+            'payload': payload,
+            TRACE_CONTEXT: trace_context
+        }
+
+        self._logger.debug('Dispatching announcement (routing_key=%s,payload=%s)',
+                           routing_key, payload)
+        self._publisher.publish(payload=payload, routing_key=routing_key)
+
+
+def get_queue(name=None, routing_key=None, exclusive=False):
```
manasdk: lets make name mandatory pls. I might have caused the first example of such a thing but this make debugging harder.

manasdk:
Some thoughts - 
1. I think we all need to agree on external APIs. I see announcements as a consume only API and not an API that receives any POSTs at this point. 
2. The announcement concept is still opaque from an external perspective i.e. what is announced, what is the structure of an announcement and how does one control what gets announced - we can iterate.
3. If you compare and contrast a `notification` and `announcement` I suspect we will see some confusion here simply on the basis of the terminology so likely we think through that part as well. e.g. what is the difference between an `broadcast notification` and `announcement`. Also, if we do collapse this the terms does it make sense to have such divergent implementations?
4. The good part is that this builds on notifications and it is a fairly clever use of a runner. It may not be a natural use of the runner but then everything is eventually code anyway.

Although my comments seem to highlight concerns in the design/no-design here I am actually +ve of the notion of announcements.

st2common/st2common/bootstrap/runnersregistrar.py
enykeev: @lakshmi-kannan was worried users may start using it before it is ready. I myself agree that it might not be such a big problem based on our previous experience, but decided to do it just in case.

st2api/st2api/listener.py
enykeev: actually, no. The exclusive means it would be deleted as soon as consumer stop listening it. Since we want every consumer to listen to its own queue instead of sharing the same queue with another consumer, we absolutely can not give it a name.

The problem with sharing is that if one consumer will consume the message, another one will never get it so if one st2api instance get an update on execution, the users from another instance of st2api will never get it.

st2common/st2common/transport/announcement.py
enykeev: As I said, name should be unique and as far as I know you can't make rabbitmq to assign it a readable and at the same time unique name unless you're willing to manage uniqueness yourself.

enykeev:
> The announcement concept is still opaque from an external perspective i.e. what is announced, what is the structure of an announcement and how does one control what gets announced - we can iterate.

The way the runner is implemented right now, you can actually send any kind of data on the channel providing you're willing to handle it yourself the other end. Any action on top of this runner would start to make it opinionated one way or the other.

For example, `core.announcement` expect a single parameter `message` of type object and will send it to `general` route key by default so on the consumer end, there will be an event `st2.announcement__general` with content: 
```js
{
  payload: {
    message: {...}
  }
}
```

On the other hand, new `hubot.post_message` also expects `message`, but this time is a type of string. It also expects three more parameters and defines a route to be `chatops`. The resulting event would be `st2.announcement__chatops` with content: 
```js
{
  payload: {
    message: '...',
    user: '...',
    channel: ...
  }
}
```

> If you compare and contrast a notification and announcement I suspect we will see some confusion here simply on the basis of the terminology so likely we think through that part as well. e.g. what is the difference between an broadcast notification and announcement. Also, if we do collapse this the terms does it make sense to have such divergent implementations?

The confusion is of course to be there as the terms are pretty close to each other, although I don't see how we can collapse the terms.

Notifications are internal mechanic to let one **internal** component know something had happened with the other component. Like right now we're letting st2notifier know the execution has finished so it can dispatch a new trigger.

Announcements is a way for st2 to inform **external** component on something that happened inside stackstorm that consumer didn't specifically asked for. The closest concept is push notifications.

Notifications may be converted into Announcements by way of rules. Not all of them should, though. Notification may inform the system that one of actionrunners is mysteriously failed, but system may recover without human intervention.

PR #2168, opened 2015-11-03
Improve speed of "execution list" and "execution get" commands

Kami:
A while back, I was debugging and checking why the output of `st2 execution list` command on our build servers is so slow.

It turns out, most of the time is spend reading data (notably the result attribute which is really large in our case since a lot of workflow actions produce tons of data) from the DB and sending all of this data over wire.

``st2 execution list`` command doesn't even display ``result`` and ``trigger_instance`` attribute by default so it makes no sense to request it.

Similar story for ``st2 execution get`` - we don't use ``trigger_instance`` by default and if user only requests one attribute using --attr option it makes no sense to also request and download other data.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,7 +10,10 @@ in development
   the targeted box. (new feature, improvement) #2144, #2150 [Logan Attwood]
 * Improve speed of ``st2 execution list`` command by not requesting ``result`` and
   ``trigger_instance`` attributes. The effect of this change will be especially pronounced for
-  installations with a lot of large executions.
+  installations with a lot of large executions (large execution for this purpose is an execution
+  with a large result).
+* Improve speed of ``st2 execution get`` command by not requesting ``result`` and
+  ``trigger_instance`` attributes.
 
 1.1.0 - October 27, 2015
 ------------------------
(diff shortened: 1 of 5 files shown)


manasdk:
+1

Kami : Here is a :star: for you!

manasdk:
350f16ac-7a2a-11e4-8f9f-002e636534f9

Kami:
Tested it on the build server (yeah, I totally scped it :P)

Before this change:

```bash
tomaz@st2build002:~$ time st2 execution list > /dev/null

real	0m8.809s
user	0m0.721s
sys	0m0.323s
```

After:

```bash
tomaz@st2build002:~$ time st2 execution list > /dev/null

real	0m0.504s
user	0m0.209s
sys	0m0.026s
```

Now the command basically returns instantly.

lakshmi-kannan:
Nice! LGTM.

Kami:
Another possible improvement would be to use "get execution attribute" endpoint when user only requests a single attribute when using "execution get" command, but we already exclude "result" and "trigger_instance" attribute if not requested, so in practice it would make little difference since other attributes are quite small.

Kami:
Cherry picked those changes onto v1.1.1 branch.

PR #2174, opened 2015-11-05
Modifying the ActionAlias schema

emedvedev:
Extending the ChatOps schema, as discussed: 
1. Support custom ack and result parameters.
2. Support aliases.

This PR is complete, but I'm putting it on hold until the changes in hubot-stackstorm are ready to merge as well.

### DIFF
--- a/st2common/st2common/models/api/action.py
+++ b/st2common/st2common/models/api/action.py
@@ -537,7 +537,7 @@ class ActionAliasAPI(BaseAPI, APIUIDMixin):
         result = alias.result
 
         model = cls.model(name=name, description=description, pack=pack, ref=ref, enabled=enabled,
-                          action_ref=action_ref, formats=formats, aliases=aliases, ack=ack, 
+                          action_ref=action_ref, formats=formats, aliases=aliases, ack=ack,
                           result=result)
         return model
 
--- a/st2common/st2common/models/api/action.py
+++ b/st2common/st2common/models/api/action.py
@@ -496,7 +496,7 @@ class ActionAliasAPI(BaseAPI, APIUIDMixin):
             "formats": {
                 "type": "array",
                 "items": {
-                    "anyOf": [ 
+                    "anyOf": [
                         {"type": "string"},
                         {
                             "type": "object",
--- a/st2common/st2common/models/api/action.py
+++ b/st2common/st2common/models/api/action.py
@@ -495,13 +495,22 @@ class ActionAliasAPI(BaseAPI, APIUIDMixin):
             },
             "formats": {
                 "type": "array",
-                "items": {"type": "string"},
+                "items": {
+                    "anyOf": [ 
+                        {"type": "string"},
+                        {
+                            "type": "object",
+                            "properties": {
+                                "display": {"type": "string"},
+                                "representation": {
+                                    "type": "array",
+                                    "items": {"type": "string"}
+                                }
+                            }
+                        }
+                    ]
+                },
                 "description": "Possible parameter format."
-            },
-            "aliases": {
-                "type": "array",
-                "items": {"type": "string"},
-                "description": "Parameter format hidden from public list."
             },
             "ack": {
                 "type": "object",
@@ -532,13 +541,11 @@ class ActionAliasAPI(BaseAPI, APIUIDMixin):
         enabled = getattr(alias, 'enabled', True)
         action_ref = alias.action_ref
         formats = alias.formats
-        aliases = getattr(alias, 'aliases', None)
         ack = getattr(alias, 'ack', None)
         result = getattr(alias, 'result', None)
 
         model = cls.model(name=name, description=description, pack=pack, ref=ref, enabled=enabled,
-                          action_ref=action_ref, formats=formats, aliases=aliases, ack=ack,
-                          result=result)
+                          action_ref=action_ref, formats=formats, ack=ack, result=result)
         return model
 
 
--- a/st2common/st2common/models/api/action.py
+++ b/st2common/st2common/models/api/action.py
@@ -532,9 +532,9 @@ class ActionAliasAPI(BaseAPI, APIUIDMixin):
         enabled = getattr(alias, 'enabled', True)
         action_ref = alias.action_ref
         formats = alias.formats
-        aliases = alias.aliases if hasattr(alias, 'aliases') else None
-        ack = alias.ack
-        result = alias.result
+        aliases = getattr(alias, 'aliases', None)
+        ack = getattr(alias, 'ack', None)
+        result = getattr(alias, 'result', None)
 
         model = cls.model(name=name, description=description, pack=pack, ref=ref, enabled=enabled,
                           action_ref=action_ref, formats=formats, aliases=aliases, ack=ack,
--- a/st2common/st2common/models/api/action.py
+++ b/st2common/st2common/models/api/action.py
@@ -506,7 +506,7 @@ class ActionAliasAPI(BaseAPI, APIUIDMixin):
             "ack": {
                 "type": "object",
                 "properties": {
-                    "disabled": {"type": "boolean"},
+                    "enabled": {"type": "boolean"},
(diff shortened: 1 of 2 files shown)


st2common/st2common/models/api/action.py
```diff
@@ -497,6 +497,27 @@ class ActionAliasAPI(BaseAPI, APIUIDMixin):
                 "type": "array",
                 "items": {"type": "string"},
                 "description": "Possible parameter format."
+            },
+            "aliases": {
+                "type": "array",
+                "items": {"type": "string"},
+                "description": "Parameter format hidden from public list."
+            },
+            "ack": {
```
Kami: Do we ever want to support any other ack related attributes? If not, it's probably better to just use a simple "boolean" attribute (e.g. `require_ack`).

st2common/st2common/models/api/action.py
```diff
@@ -497,6 +497,27 @@ class ActionAliasAPI(BaseAPI, APIUIDMixin):
...
+                "type": "object",
+                "properties": {
+                    "disabled": {"type": "boolean"},
```
Kami: We use `enabled` attribute everywhere else (resource definitions, rbac definitions, etc.) so if we are going with the object approach we should also use `enabled` here for consistency.

Kami:
Hm, that's weird, I don't think failure is related to your change (https://travis-ci.org/StackStorm/st2/jobs/89456490#L265), I will dig in.

st2common/st2common/models/api/action.py
emedvedev: We're supporting `format` as well (at least I want to, it's still up for discussion), so I'd go with object.

st2common/st2common/models/api/action.py
emedvedev: Will fix.

st2common/st2common/models/api/action.py
```diff
@@ -497,6 +497,27 @@ class ActionAliasAPI(BaseAPI, APIUIDMixin):
                 "type": "array",
                 "items": {"type": "string"},
                 "description": "Possible parameter format."
+            },
+            "aliases": {
```
Kami: Btw, what is this attribute going to contain?

st2common/st2common/models/api/action.py
```diff
@@ -497,6 +497,27 @@ class ActionAliasAPI(BaseAPI, APIUIDMixin):
...
+                "properties": {
+                    "enabled": {"type": "boolean"},
+                    "format": {"type": "string"}
+                },
+                "description": "Acknowledgement message format."
+            },
+            "result": {
+                "type": "object",
+                "properties": {
+                    "enabled": {"type": "boolean"},
```
Kami: Yeah, I like that.

Eventually when and if we are going to support per adapter / route formatting we can also add another attribute (route / adapter) here.

st2common/st2common/models/api/action.py
Kami: I might have missed a discussion or something so I'm slightly confused about this attribute.

Besides that, LGTM

emedvedev: We discussed it in the chatops meeting on Mon. It's pretty much same as formats (matching user strings), with the notable exception of not being exposed in Hubot's `!help`. It's part of my effort to "humanize" the bot: `formats` contain well-defined string commans, and `aliases` could contain regexes for the same command, or alternative strings, catching common typos or even adding a tiny bit of NLP. 

This came from experience in exposing new people to chatops: if you have a lot of aliases, people will forget exact commands, or misspell them, or try to use a different syntax, so it's nice to actually have some flexibility on that without flooding `!help` with 10 variations on a command or a huge regex.

So the name pretty much sums it up: it's a list of aliases for format.

st2common/st2common/models/api/action.py
```diff
@@ -497,6 +497,27 @@ class ActionAliasAPI(BaseAPI, APIUIDMixin):
...
+                    "format": {"type": "string"}
```
lakshmi-kannan: String? I am confused. How is this going to look? Can you write out an example?

emedvedev:
```
format: "your execution id is {{ execution.id }} and the result is {{ execution.result }}"
```
Something like that. @enykeev is working on the back-end changes for that, and `ack` parameters (both `enable` and `format`) are (mostly) implemented in my recent `hubot-stackstorm` PR.

st2common/st2common/models/api/action.py
Kami: I see, thanks for the clarification. I'm fine with that, I just found the name a bit confusing - alias for an alias! lol.

Anyways, LGTM, +1

manasdk: Yeah name is very confusing. It certainly does not capture the intent. Can you provide an example of an actual alias?

emedvedev: The one example I have (not very characteristic, but still) is a command getting a nearest event from the calendar: "what's next?"

I've seen people writing it as "what is next?" and "what's next" very often, which is one concern, and the other is the single quote in "what's", which Slack changes to proper apostrophe, while IRC, for example, will not, so I need to match both.

At the same time, I want `!help` to provide a clear and meaningful list, not a list of the same command with typos and variations. So that's what I do:
```
formats:
    - "what's next?"
    - "when is {{ query }}?"
aliases:
    - "what(.s| is) next\\??"
    - "when is {{ query }}\\??"
```
In `formats` I have commands that are also help entries, in `aliases` I do a wider match with a regex, covering all possible variations—and putting it in `!help` will not be user-friendly in the slightest. So, I really can't think of a name other than `aliases` because that's exactly what it is. Still, aliases for aliases is kind of weird, and I'm open to suggestions on this one.

manasdk: Hmm, we need to think this through. I am not convinced. I can see these being combined into formats after all that is the entire point of a format. I see being able to support regex in formats as useful.

st2common/st2common/models/api/action.py
manasdk: Also the slack special stuff is already handled

st2common/st2common/models/api/action.py
manasdk: https://github.com/StackStorm/hubot-stackstorm/blob/master/scripts/stackstorm.js#L211

emedvedev: We do support regex in formats. The whole point is not exposing this in `!help` and (in the future) other public lists. Specifying long regexes in formats,—or five formats for the same command,—is cluttering the list, it's not user-friendly and I'm pretty sure it hurts adoption. We could somehow flag the formats to not show up in help, but I couldn't come up with a better idea. 

I do share your concern though, I just think it's the best option in our current state: it's fully reusing the current code, it doesn't require you to learn anything new, and it's completely backwards compatible (just don't use it if you don't have to).

emedvedev: Kudos for `normalizeCommand`, btw. Didn't have it when I first encountered that apostrophe issue.

manasdk: Have you considered something along the lines of -
```
formats:
    - display:  "what's next?"
      representation: 
          - "what(.s| is) next\\??"
    - display: "when is {{ query }}?"
      representation: 
           - "when is {{ query }}\\??"
```
This would be horrible to write and also backwards incompatible(we do need to worry about such things soon)

The 2 things that does is -
1. provides a concrete relation between formats and representation.
2. clarifies that is the displayable version and the actual matched version.

Kami: Yeah, now that I read the original explanation again I was how is this different then existing formats attribute.

Anyway, I do like this new proposed formats change though.

emedvedev: @manasdk That's exactly what I considered, but I really don't like breaking backwards compatibility without a major release. What you're proposing is a great idea, but are we really fine with introducing something this drastic?

PR #2176, opened 2015-11-05
[CLI] Display elapsed time for running actions / workflows

Kami:
With this change we now display elapsed time in seconds when using "execution list" and "execution get" commands for all the actions in the "running" state.

This was big annoyance for me every time I was checking things on our build server where we have a lot of long running actions - previously I would need to check current time and start timestamp, but now you can see elapsed time in a glance.

selection_287

selection_288

### DIFF
--- a/st2client/st2client/formatters/table.py
+++ b/st2client/st2client/formatters/table.py
@@ -32,6 +32,14 @@ LOG = logging.getLogger(__name__)
 MIN_ID_COL_WIDTH = 26
 DEFAULT_ATTRIBUTE_DISPLAY_ORDER = ['id', 'name', 'pack', 'description']
 
+# Attributes which contain bash escape sequences - we can't split those across multiple lines
+# since things would break
+COLORIZED_ATTRIBUTES = {
+    'status': {
+        'col_width': 40
+    }
+}
+
 
 class MultiColumnTable(formatters.Formatter):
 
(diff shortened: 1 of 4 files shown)


st2client/st2client/formatters/table.py
```diff
@@ -59,10 +67,23 @@ def format(cls, entries, *args, **kwargs):
...
 
             widths = []
+            subtract = None
             for index in range(0, len(attributes)):
+                attribute_name = attributes[index]
+
                 if index == 0:
                     widths.append(first_col_width)
                 else:
+                    if attribute_name in COLORIZED_ATTRIBUTES:
```
Kami: We need to do that since if we break / wrap line incorrectly (e.g. before / in the middle of escape character), things will be messed up.

Previously we were just lucky that we never experienced this issue since "status" attribute never contained too much text (well, if a user had a really narrow terminal window, they would still experience this issue).

st2client/st2client/utils/color.py
```diff
@@ -67,5 +67,19 @@ def colorize(value, color=''):
...
+        status = split[0].strip()
+        reminder = '(' + split[1]
+    else:
+        status = value
+        reminder = ''
+
+    color = STATUS_LOOKUP.get(status, DisplayColors.YELLOW)
+    result = DisplayColors.colorize(status, color)
+
+    if reminder:
```
manasdk: remainder :P

st2client/st2client/utils/color.py
```diff
@@ -67,5 +67,19 @@ def colorize(value, color=''):
 
 
 def format_status(value):
-    color = STATUS_LOOKUP.get(value, DisplayColors.YELLOW)
-    return DisplayColors.colorize(value, color)
+    # Support status values with elapsed info
+    split = value.split('(', 1)
```
manasdk: This is a little hokey but I can live with it :)

manasdk:
+1

Good change :)

st2client/st2client/utils/color.py
Kami: lol, good catch, will fix it

Kami:
Merged into master, will also cherry pick onto v1.1.1.

PR #2177, opened 2015-11-06
Change protocol in AIO oneliner

enykeev:
Despite the fact we're doing temporary redirect to https source, we still receiving this redirect through insecure channel. And from what I can tell, `curl` won't upgrade the connection automatically.


emedvedev:
I'm sorry, Dave. I'm afraid I can't do that.

The reason we had to switch to plain HTTP is that RHEL6 is stupid: its version of curl doesn't support any of the ciphers CloudFlare uses, and we can't really configure that list, CloudFlare gives no control over it.

In the end we decided to use http instead of having a separate docs entry for RHEL6. After all, the request does end up with an https connection to github.

enykeev:
> After all, the request does end up with an https connection to github.

Unless it will get intercepted before and then we'll just running an arbitrary script with sudo.

emedvedev:
I feel your pain, honestly, so I don't mind leaving it up for discussion. When everyone's back next week, we can decide what to do, but simply changing the protocol to `https` will not work: we still need to support RHEL6 somehow.

lakshmi-kannan:
Gosh. Old operating systems :/

manasdk:
publish the MD5 and use something like http://jbenet.github.io/hashpipe/ in our command. This way atleast there is some security about what is being run is what StackStorm expects to run.

Btw, we should publish the checksum for sure.

enykeev:
I guess we should*, though serving the installer from another host supporting RHEL6 ciphers is probably a better way (providing this ciphers are actually more secure than md5 :stuck_out_tongue_winking_eye:) 

* I advice on SHA since it is relatively easy to forge the collision with md5

manasdk:
can we close this yet?

enykeev:
Nothing have changed really, we still have a potential for MITM right there. We need to find a way to enable proper ciphers on el6 or move redirect outside of CloudFlare on a separate host.

emedvedev:
Can we just post the hashes as previously agreed? Both moving outside of CloudFlare and changing RHEL cipherlist aren't really worth the effort, tbh.

Let me prove my point here. If I'm able to change someone's incoming traffic during stackstorm's installation, I can:
1. Change the link in the docs to whatever I need, including plain http or my own gist.
2. Change some of the OS/pip dependency downloads.
3. Make StackStorm instance into a phishing login form, because hardly anybody even gets post-install root CA messages now, let alone actually uses one, people ignore warnings anyway.
And that's just off the top of my head.
4. Try to intercept any other insecure communication to/from server. Come on, it's MITM, there's no way _everything_ is secure. 

I'm in no way saying that we should abandon security here, just that cost of moving from CloudFlare or trying to bless an outdated OS with a new cipherlist is too much, given there are many other possible attack vectors if you can do a MITM attack on someone's machine.

And there's always good old thermorectal cryptanalysis.

emedvedev:
And if someone cares about security a lot, they will notice hashes and will use them. It's really enough.

emedvedev:
After some raging (mostly from my side) and a discussion with @enykeev, I think a good solution would be moving the link to a separate domain like `https://install.stackstorm.com`, which will not be controlled by CloudFlare. This way we can maintain our own cipherlist compatible with RHEL6 (and any other outdated systems should we decide to support them in the future), address the security concerns and still keep operational overhead to a minimum. Wdyt?

jfryman:
@emedvedev Let's do it!

emedvedev:
Installer endpoint is now `https://install.stackstorm.com/`. The redirect is managed by cPanel at our hosting, and I modified the cipherlist for compatibility, so RHEL6 is working. Safe to merge now.

The old redirect is also working, of course.

jfryman:
@emedvedev Would you type up some very small docs on what is setup and where? I'd like to ensure all this is documented so it's not SPOF on you.

PR #2178, opened 2015-11-06
Action chain validation improvements

Kami:
This pull request includes some action chain validation improvements.

1. Perform on-success and on-failure task name validation during run time. Previously validation happened during run time which means develop-feedback loop was very slow if user was testing workflow with long running tasks and there was, for en example, typo in the take name referenced in on-success / on-failure. This should speed up the whole develop - test loop since this issues are now detected during "pre_run" phase.
2. Throw more friendly exception when action parameter or "publish" parameter resolving fails. Now we also include task where the exception occurred, parameter name and the actual template string value.

### DIFF
--- a/st2actions/st2actions/runners/actionchainrunner.py
+++ b/st2actions/st2actions/runners/actionchainrunner.py
@@ -408,7 +408,14 @@ class ActionChainRunner(ActionRunner):
         context.update(chain_vars)
         context.update({RESULTS_KEY: previous_execution_results})
         context.update({SYSTEM_KV_PREFIX: KeyValueLookup()})
-        rendered_result = jinja_utils.render_values(mapping=action_node.publish, context=context)
+
+        try:
+            rendered_result = jinja_utils.render_values(mapping=action_node.publish, context=context)
+        except Exception as e:
+            msg = ('Failed rendering value for publish parameter "%s" in task "%s" (template string=%s): %s' %
+                   (e.key, action_node.name, e.value, str(e)))
+            raise ParameterRenderingFailedException(msg)
+
         return rendered_result
 
     @staticmethod
@@ -425,8 +432,11 @@ class ActionChainRunner(ActionRunner):
             rendered_params = jinja_utils.render_values(mapping=action_node.params,
                                                         context=context)
         except Exception as e:
-            LOG.exception('Jinja rendering failed.')
-            raise ParameterRenderingFailedException(e)
+            LOG.exception('Jinja rendering for parameter "%s" failed.' % (e.key))
+
+            msg = ('Failed rendering value for parameter "%s" in task "%s" (template string=%s): %s' %
+                   (e.key, action_node.name, e.value, str(e)))
+            raise ParameterRenderingFailedException(msg)
         LOG.debug('Rendered params: %s: Type: %s', rendered_params, type(rendered_params))
         return rendered_params
 
--- a/st2actions/st2actions/runners/actionchainrunner.py
+++ b/st2actions/st2actions/runners/actionchainrunner.py
@@ -75,23 +75,25 @@ class ChainHolder(object):
         perform only simple validation during compile / create time.
         """
         all_nodes = self._get_all_nodes(action_chain=self.actionchain)
-        on_success_nodes = self._get_all_on_success_nodes(action_chain=self.actionchain)
-        on_failure_nodes = self._get_all_on_failure_nodes(action_chain=self.actionchain)
 
-        # 1. Check all the "on-success" paths
-        for node_name in on_success_nodes:
-            valid_name = self._is_valid_node_name(all_node_names=all_nodes, node_name=node_name)
+        for node in self.actionchain.chain:
+            on_success_node_name = node.on_success
+            on_failure_node_name = node.on_failure
 
+            # Check "on-success" path
+            valid_name = self._is_valid_node_name(all_node_names=all_nodes,
+                                                  node_name=on_success_node_name)
             if not valid_name:
-                msg = 'Unable to find node with name "%s".' % (node_name)
+                msg = ('Unable to find node with name "%s" referenced in "on-success" in '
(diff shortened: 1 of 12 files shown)


st2actions/st2actions/runners/actionchainrunner.py
```diff
@@ -92,6 +125,50 @@ def _get_default(_actionchain):
         return _actionchain.chain[0].name
 
     @staticmethod
+    def _get_all_nodes(action_chain):
```
Kami: I follow an existing style approach, but honestly, I'm not a big fan of it.

I would rather move those utility functions on the actual `ActionChain` system model class. What do others think?

st2actions/st2actions/runners/actionchainrunner.py
```diff
@@ -63,6 +67,35 @@ def init_vars(self, action_parameters):
             self.vars = self._get_rendered_vars(self.actionchain.vars,
                                                 action_parameters=action_parameters)
 
+    def validate(self):
```
Kami: Similar as below - another option is to move this to the ActionChain class (it doesn't require any run time info) and validate the workflow and throw during resource loading phase (meaning when user runs register-actions).

What do other people think?

Kami: To add to this - that's something we should agree on and consistently follow for all the resources and use cases.

Imo, ideally we would always try to perform as much validation as possible as early as possible (this means during create / load time). This provides a better user experience since user doesn't need to wait for the run time to get back feedback.

manasdk: Yes. Early validation is good for the all the meta stuff. In workflow cases we only can really validate at runtime since the workflow could be changed without StackStorm even knowing. I do that myself all the time.

Therefore, we necessarily have to perform validations at runtime but can additionally do so at create/update time - these might have limited value IMO.

st2actions/st2actions/runners/actionchainrunner.py
manasdk: Models should have no notion of how they are consumed. Utilities of these sort belong out side of model classes since the meaning and interaction should necessarily be handled outside of the model class. This way arguably different services could interpret the same model in a different way - unlikely but possible.

e.g. 
https://github.com/StackStorm/st2/blob/master/st2common/st2common/models/db/stormbase.py#L79
and
https://github.com/StackStorm/st2/blob/master/st2common/st2common/models/db/stormbase.py#L90

Do not have to belong to this type. They are operations that are closely related but not a feature of the type itself.

ONTH https://github.com/StackStorm/st2/blob/master/st2common/st2common/models/db/stormbase.py#L223

the mixin approach is somewhat different pattern since it is the only purpose of the mixin is to augment with specific functionality. It is arguable if mixins should be used in the first place with models as they are better suited with service classes.

Model classes hold data and service classes perform business logic.

You could have a pragmatist argument about this with and say how else is one supposed to interpret the ActionChain model and I would counter with well what else is supposed to interpret the model? Also, if there is a need to perform such a validation then it can easily be done in a standalone actionchain validator.

manasdk: Theoretical argument aka feel free to ignore -

The fact that an ActionChain is valid i.e. the `on-success` and `on-failure` links are broken or how to interpret the broken-ness is a function of the chainrunner. It is possible for a different implementation of the ActionChainRunner to interpret this condition differently. Therefore, validity of an ActionChain in this context can only be decided by this runner.

In the context of StackStorm this argument is mostly invalid since we would not be writing another runner anyway therefore theoretical. However, since we build software with certain recognizable techniques (aka patterns) and hope that the next new person who shows up can semi-intuitively understand the recognizable patterns and follow along I feel it is worthwhile to be consistent about such things.

As such I rest my case and if you do prefer to move it out to ActionChain model itself go for it :)

st2actions/st2actions/runners/actionchainrunner.py
Kami: > In workflow cases we only can really validate at runtime since the workflow could be changed without StackStorm even knowing. I do that myself all the time.

Yeah, that's a good point. I also utilize this "feature" a lot myself as well (I think this was probably more of a by product of an implementation an not an actual feature, but anyway, people like it :)).

So yeah, for now, I will leave it as it is.

Kami:
I've added some more fixes / improvements.

Now we return a more friendly error if a action parameter or publish parameter resolving fails. Previously only the actual Jinja exception was bubbled up which was not really friendly and made debugging harder.

Now the error message includes task where the exception happened, parameter name and the actual template string.

Before:

```
ParameterRenderingFailedException: UndefinedError: 'foo_bar' is undefined
```

After:

```
ParameterRenderingFailedException: Failed rendering value for parameter "p1" in task "c2" (template string={{ foo_bar }}): 'foo_bar' is undefined
```

st2actions/st2actions/runners/actionchainrunner.py
```diff
@@ -63,23 +67,52 @@ def init_vars(self, action_parameters):
...
+        all_nodes = self._get_all_nodes(action_chain=self.actionchain)
+        on_success_nodes = self._get_all_on_success_nodes(action_chain=self.actionchain)
+        on_failure_nodes = self._get_all_on_failure_nodes(action_chain=self.actionchain)
+
+        # 1. Check all the "on-success" paths
+        for node_name in on_success_nodes:
+            valid_name = self._is_valid_node_name(all_node_names=all_nodes, node_name=node_name)
+
+            if not valid_name:
+                msg = 'Unable to find node with name "%s".' % (node_name)
```
lakshmi-kannan: Maintaining the original task name and/or line number seems important. Maybe next steps?

Kami: Getting the task name is possible, but getting the line number is a bit harder right now and would be more of a hack.

Kami: Pushed the improvement in  f1995fe.

PR #2180, opened 2015-11-08
Fixate reqs update

dennybaa:
st2 debug builds with latest pecan. But expect it should use ours.


Kami:
As noted on slack - only st2api and st2debug depend on pecan.

st2debug contains a standalone CLI utility which doesn't depend and need pecan.

dennybaa:
ok. sure I removed st2debug pecan addition. It's not here'
Just merge the makefiles. I've cleand up this PR. Thank you.

Kami:
Merged, thanks.

dennybaa:
gracias!

Issue #1761, opened 2015-07-29
Move from pickle to JSON serializer for message bug messages


Kami:
(I've mentioned this multiple times before, but now I'm creating ticket so we can track / prioritize it).

We should move from pickle which we use right now to the JSON serializer for the messages which are dispatched on the RabbitMQ message bus.

Using pickle is fine if all data which goes through the message bus is controlled and generated by us (and if you don't mind about cross Python version and / or cross programming language compatibility), but this is not the case here (triggers can contain arbitrary data since we poll 3rd systems, etc.). This caries a big risk since if data is maliciously constructed.

Switching away from pickle means we won't be able to directly dispatch class instances to the messages bus anymore and we will need to use something like Thrift / Protobus and define structure for all the classes and data we want to pass through the message bus.

That's actually a good thing since those classes will now be decoupled from Python and we will be able to use Thrift / Protobuf definitions to work with them in other languages (e.g. JavaScript).

lakshmi-kannan:
+100.

manasdk:
msgpack might work also? 

So it is primarily a security issue or a correctness issue?

Kami:
It's primarily a security issue and secondary a "portability" issue (using thrift / protobuffs / msgpack /etc.) makes it easier to re-use the models and other classes in other languages.

manasdk:
@Kami Ok. Portability is not a concern in the short-term - right? 

Security is always a concern and maybe that is something which raises priority. Again, the attack vector and impact would be important in working out the relative priority. 

The reason I picked `pickle` was to avoid a short term dealing with serialization. The assumption which I made were -
1. The ingress and egress points at StackStorm and we know it will only be python in the medium term (there is no requirement yet that violates this assumption)
2. We have control over messages and can exert some control on the content so if some special encoding is required that can be provided to plug the security hole.

I do feel we ought to move over to json/msgpack/protbuf i.e. some portable serialization structure. This serialization most likely involves very strict model definitions etc and formalizing the content schema which could be somewhat tedious. Therefore my concern is this change is not simple so how far can we push it out?

Kami:
Yes, I agree that portability is not a big issue or a priority right now, but once we will be addressing that, we will also need to address serialization issue and if we use something like Thrift we won't be reinventing the wheel and we will solve two problems at once.

[...]

PR #2181, opened 2015-11-09
Add werkzeug to not break on MultiData

dzimine:
Problem: 
rule engine doesn't decode MultiDict payload
Problem: sensor passes a trigger payload as MultiDict object. We can likely deal with it. But rule engine breaks with the following error:
``
ERROR 140498111651920 mixins [-] Can't decode message body: DecodeError(ImportError('No module named werkzeug.datastructures',),) (type:u'application/x-python-serialize' encoding:u'binary' raw:'\'\\x80\\x02}q\\x01(U\\rtrace_contextq\\x02NU\\x07triggerq\\x03U\\x16activecampaign.webhookq\\x04U\\x07payloadq\\x05}q\\x06(U\\x04bodyq\\x07cwerkzeug.datastructures\\nImmutableMultiDict\\nq\\x08]q\\t(U\\rcontact[tags]q\\nX\\x00\\x00\\x00\\x00\\x86q\\x0bU\\tdate_timeq\\x0cX\\x13\\x00\\x00....
```
Adding werkzeug to python dependencies makes it work.

### DIFF
--- a/st2reactor/in-requirements.txt
+++ b/st2reactor/in-requirements.txt
@@ -7,3 +7,4 @@ jsonschema
 kombu
 oslo.config
 six
+werkzeug


lakshmi-kannan:
Good catch! Can you add an entry in CHANGELOG?

Kami:
Ah, that's the same nasty issue we had in the past with the email sensor.

I'm -1 on this change - we should fix the sensor to only dispatch simple types (e.g. probably dict in this case). Allowing user to dispatch arbitrary types is a bad thing from security and portability perspective (when we finally switch away from pickle to json this won't work anymore...).

Where does the trigger come from? It looks similar to the email sensor one - I've fixed one of those issues in the past, but it's possible similar issue exists somewhere else in the code.

dzimine:
I fixed the sensor, but there's nothing enforcing it on the sensor container type as it takes any Python object and passes it down.

Kami:
Yes, ideally we would add some checking to the trigger dispatch.

I believe it was discussed before, but we didn't add it since it would be too expensive - we would need to recursively check all the attributes which would get expensive for nested objects.

Real solution is #1761 - after we switch to JSON, serialization will throw an error if the object contains non-serializable fields and we don't need to add any additional validation ourselves.

Kami:
@dzimine Where is the sensor fix? I can't find it in st2contrib.

dzimine:
This is the sensor, and the fix is to_dict on MultiDict: https://github.com/StackStorm/st2incubator/blob/master/packs/activecampaign/sensors/ac_webhook.py#L41

Kami:
Alright, thanks, I'll close this PR then.

PR #2182, opened 2015-11-09
actionrunner invocation script improvments

dennybaa:


Kami:
This script is only used on Ubuntu inside docker image, right?

Kami:
LGTM

dennybaa:
inside packages. but yeah this a stub for debian docker containers... Not to use systemd

Kami:
Merged, thanks.

PR #2185, opened 2015-11-09
Support for default values on the API models

Kami:
This pull request introduces support for default values on API model (we already support this for action parameters).

Most of the API models already defined "default" attributes in the JSON schema, but those attributes were actually ignored and not used. Not using those attributes means we need to assign default value in multiple places - you can see a bunch of `getattr` in to_model, etc. This also resulted in a lot of bugs in the past ("rule" has no property enabled, etc.).

That's obviously bad since default values needs to spread and set all around the code base. This pull request resolves this and introduces a consistent interface - default attribute will be set on the model API instance after "model.validate()" is called if the JSON schema defined a default value. This means no more nasty getattr stuff and now we know exactly when attributes are set and available.

### DIFF
--- a/st2common/st2common/models/api/action.py
+++ b/st2common/st2common/models/api/action.py
@@ -87,7 +87,7 @@ class RunnerTypeAPI(BaseAPI):
                 "description": "Input parameters for the action runner.",
                 "type": "object",
                 "patternProperties": {
-                    "^\w+$": util_schema.get_draft_schema(version='action_params')
+                    "^\w+$": util_schema.get_action_parameters_schema()
                 }
             }
         },
@@ -178,7 +178,7 @@ class ActionAPI(BaseAPI, APIUIDMixin):
                 "description": "Input parameters for the action.",
                 "type": "object",
                 "patternProperties": {
-                    "^\w+$": util_schema.get_draft_schema(version='action_params')
+                    "^\w+$": util_schema.get_action_parameters_schema()
                 },
                 "default": {}
             },
--- a/st2common/st2common/models/api/action.py
+++ b/st2common/st2common/models/api/action.py
@@ -87,7 +87,7 @@ class RunnerTypeAPI(BaseAPI):
                 "description": "Input parameters for the action runner.",
                 "type": "object",
                 "patternProperties": {
-                    "^\w+$": util_schema.get_draft_schema()
+                    "^\w+$": util_schema.get_draft_schema(version='action_params')
                 }
             }
         },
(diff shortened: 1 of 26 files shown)


st2common/st2common/models/api/base.py
```diff
@@ -71,7 +70,22 @@ def __json__(self):
...
+        """
+        Perform validation and return cleaned object on success.
+        """
+        schema = getattr(self, 'schema', {})
+        attributes = vars(self)
+
+        cleaned = util_schema.validate(instance=attributes, schema=schema,
+                                       cls=util_schema.CustomValidator, use_default=True)
+
+        # Note: For backward compatibility reasons, we re-assign attributes one the current
```
Kami: I would personally prefer a different interface where `validate()` doesn't mutate object in place but returns a new "cleaned" object?

That's similar to how most other existing validation frameworks (in Django, mongonegine, etc.) work.

Changing it would obviously require more work since all the existing code needs to be updating, but I think it's worth it. What do others think?

st2common/st2common/models/api/action.py
```diff
@@ -88,7 +87,7 @@ class RunnerTypeAPI(BaseAPI):
                 "description": "Input parameters for the action runner.",
                 "type": "object",
                 "patternProperties": {
-                    "^\w+$": util_schema.get_draft_schema()
+                    "^\w+$": util_schema.get_draft_schema(version='action_params')
```
manasdk: Can you make `action_params` a constant?

st2common/st2common/models/api/base.py
manasdk: Yes. destructive interfaces like this are problematic and lead to surprising behaviors. It would be great if we move away to something safer for sure.

Kami: Alright, I will make that change. Hopefully it won't be too bad :)

Kami: I made the "return new cleaned instance on validate" change in 37f3f8e7e1cbf26983a53ea5ce2aead53e8d86d7.

All the tests pass and I grepped for "validate" in the code base and I didn't find any more usages of it. In any case, we should still keep an eye on it in case I missed something and a regression pops up in the future.

st2common/st2common/models/api/action.py
Kami: Yeah, sure - I will actually just add a new method - "get_actions_parameters_schema".

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1450759295017313

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1450759291017312

PR #2186, opened 2015-11-09
Handle sensor being updated with an enabled status of False

vCabbage:
* Previous behavior was to reload the sensor on any update.
* This change handles the sensor being disabled by only unloading if enabled is set to False.


manasdk:
LGTM

+1

PR #2187, opened 2015-11-09
Delete sensor from process container before terminating

vCabbage:
* Eliminates a race condition where the process could be respawned before `_delete_sensor` is called.


manasdk:
LGTM 

+1

Kami:
Thanks good catch - the change looks good to me.

Kami:
@manasdk Since you merged the PR, can you please also update changelog?

Kami:
Updated the changelog and also cherry-picked this fix onto v1.1.1 branch.

Thanks again.

PR #2188, opened 2015-11-09
Change headers and params core.http action parameters type to object

Kami:
This changes the type of headers and params ``core.http`` action parameters type from ``string`` to ``object``.

I assume the action was defined before CLI supported complex type so action defined type as string instead of object. This is invalid and incorrect use of the types - the type should indeed be object (dict) since the runner shouldn't be responsible of figuring out and parsing the parameters.

This change is backward incompatible, but I think it's fine. I wanted to make it backward compatible since we support list for parameter `type` attribute but things broke, horrible. I got the CLI to work, but everything else broke down.

On a related note - I think we should also make a change so user can't specify a list for action parameter `type` attribute since things (parameter casting, etc.) break horribly and we don't support it.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,6 +21,8 @@ in development
 * Update action chain runner so it performs on-success and on-error task name validation during
   pre_run time. This way common errors such as typos in the task names can be spotted early on
   since there is no need to wait for the run time.
+* Change ``headers`` and ``params`` ``core.http`` action paramer type from ``string`` to
+  ``object``.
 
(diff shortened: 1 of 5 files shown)


manasdk:
+1

Kami:
> On a related note - I think we should also make a change so user can't specify a list for action parameter type attribute since things (parameter casting, etc.) break horribly and we don't support it.

Made that change in #2185 which already deals with validation - b4a56ea5ba998a919ef95c02a13c4dec10231c61.

PR #2189, opened 2015-11-10
Feedback: Jinja support for notifications

lakshmi-kannan:
## What?

This PR adds Jinja support for message and data in notifications. 

## TODO

* - [X] Examples
* - [X] Docs
* - [X] Changelog


docs/source/chatops/notifications.rst
```diff
@@ -90,11 +89,33 @@ with when you setup chatops. An example is below:
...
+  on-success:
+    routes:
+      - email
+    message: '"@channel: Woohoo!". Action run by user {{action_context.user}} succeeded.'
+    data:
+      cmd: "{{cmd}}"
+      stdout: "{{stdout}}"
 
 How do I setup notifications in action chain?
------------st2api/st2api/controllers/v1/ruleviews.py----------------------------------
```
manasdk: lol. No clue what happened here and how I managed to check-in that change.

st2actions/st2actions/notifier/notifier.py
```diff
@@ -159,6 +181,22 @@ def _post_notify_subsection_triggers(self, liveaction=None, execution_id=None,
             if len(failed_routes) > 0:
                 raise Exception('Failed notifications to routes: %s' % ', '.join(failed_routes))
 
+    def _build_jinja_context(self, liveaction, execution):
+        context = {SYSTEM_KV_PREFIX: KeyValueLookup()}
+        context.update(liveaction.parameters)
```
manasdk: no prefix here? This could lead to overrides from the result.

st2actions/st2actions/notifier/notifier.py
```diff
@@ -159,6 +181,22 @@ def _post_notify_subsection_triggers(self, liveaction=None, execution_id=None,
...
+        context.update({ACTION_CONTEXT_KV_PREFIX: liveaction.context})
```
manasdk: Hmm, the content of this value is opaque. I suspect it also depends on the source of the action etc.

manasdk:
Looks like tasks in an ActionChain do not get an special handling therefore may not get access to task name etc - right?

st2actions/st2actions/notifier/notifier.py
lakshmi-kannan: I brought this up on slack. This is the same story in action chain too. Prefixng here but not there would lead to inconsistency.

st2actions/st2actions/notifier/notifier.py
lakshmi-kannan: Yep. We've documented what goes into action_context. Other than that, I don't have a better suggestion.

lakshmi-kannan:
>Looks like tasks in an ActionChain do not get an special handling therefore may not get access to task name etc - right?

Correct. We don't have that visibility in the liveaction or execution model.

lakshmi-kannan:
@manasdk Can we talk about this PR in your morning so I can make progress on this?

st2actions/st2actions/notifier/notifier.py
manasdk: I supposed we could add a prefix in ActionChain as well. It sounds like a mistake due to inexperience in ActionChain. I do not think we should continue with the same pattern here as well.

st2actions/st2actions/notifier/notifier.py
manasdk: Works for me.

manasdk:
@lakshmi-kannan I think you should go ahead and merge the changes.

+1

st2actions/st2actions/notifier/notifier.py
lakshmi-kannan: I'll address this before merging.

PR #2190, opened 2015-11-10
[WIP] [RFC] Update local runner to use login shell when running as non-root / non-current user

Kami:
Before this change when a local runner action resulted in using sudo (e.g. action runner is running as root, but the action is ran as stanley or when user requests action to run as a different non-system / root user), the command would not run using login shell (bash_profile is not sourced, etc.) which would result in an unexpected behavior.

This means that $HOME variable would reflect the value of the user which ran sudo and not the value of the actual user under which the command ran, etc. This behavior also already resulted in some confusion for our users in the past.

This change is backward incompatible since it changes the local runner behavior so it should be included in v1.2 and not v1.1.x.

This also fixes an issue with some of our end to end and deploy tests (stanley not being able to write to /root/, etc. - for details, see Slack).

See:

```bash
root@st2stage201:/home/tomaz# sudo -u stanley -- bash -c 'echo $HOME;whoami'
/root
stanley
root@st2stage201:/home/tomaz# sudo -i -u stanley -- bash -c 'echo $HOME;whoami'
/home/stanley
stanley
root@st2stage201:/home/tomaz# sudo -H -u stanley -- bash -c 'echo $HOME;whoami'
/home/stanley
stanley
```

Keep in mind that using login shells also affects other things, not just $HOME.

TODO:

- [ ] Tests
- [ ] Update docs
- [ ] Update upgrade notes


st2reactor/st2reactor/sensor/compat.py
```diff
@@ -1,73 +0,0 @@
-# Licensed to the StackStorm, Inc ('StackStorm') under one or more
```
Kami: That's unrelated to this PR, I just noticed this file is not actually used and obsolete.

Kami:
It turns out this will require a bit more work - now that using -i, variables which are passed to the subprocess are not preserved (st2_action_api_url, etc.) so I need to fix / change that.

Kami:
Another option is to use `-H` instead of `-i`. The end result is not the same since `-H` just sets $HOME, but at least it resolves some sudo related issues.

```
      -H, --set-home
                 Request that the security policy set the HOME environment variable to the home directory specified
                 by the target user's password database entry.  Depending on the policy, this may be the default
                 behavior.
```

What do other people think? Another option to work around -E and -i is to explicitly set all the environment variables using export inside the sudo command, but I'm not big fan of that, even though we do a similar thing in the remote runner (there are less variables to be set and inherited in the remote runner)...

Kami:
I thought about it some more and I think we should just go with `-H` for now.

It's not ideal, but for the most part it's a small and not breaking change and solves the most pressing problem with home directory not being set to the selected user.

Using -i without preserving existing variables (-E which doesn't work directly with -i) would also probably break some existing deployments - I know some users who set environment variables for stanley user and with this change those variables wouldn't be available when running as non stanley user.

Kami:
Alright, discussed this with @manasdk on Slack - we will proceed with `-i` and I will make the change to add all the environment variables before the command, similar as we do in the new remote runner.

In short - yes, change is breaking in some scenarios, but we believe is the right path forward.

manasdk:
+1 for `-i` change.

lakshmi-kannan:

PR #2193, opened 2015-11-10
[CLI] Table improvements

Kami:
Fixes some regressions introduced when adding "elapsed" feature.


manasdk:
This entire calculation part is horrible. I don't understand the code much anymore.

Kami:
I also cherry picked it into v1.1.1.

Kami:
> This entire calculation part is horrible. I don't understand the code much anymore.

True, it's less than ideal, but I just wanted to get it to work for the common use case. We definitely should clean it up and improve it when we have more time.

manasdk:
Actually I was hoping some library would deal with this mess. But yeah problem for our future us.

PR #2199, opened 2015-11-12
Escape action parameters before saving to db

### LINKED ISSUE #2198: Special characters in object keys
Trying to run an action with `type:object` parameter with the key containing `.` like
```
{
  "a.b": "c"
}
```
returns 500 and creates a record in logs:
```
2015-11-11 23:55:43,183 ERROR [-] Unable to execute action. Unexpected error encountered.
  File "/mnt/st2/st2api/st2api/controllers/v1/actionexecutions.py", line 119, in _handle_schedule_execution
    return self._schedule_execution(liveaction=liveaction)
  File "/mnt/st2/st2api/st2api/controllers/v1/actionexecutions.py", line 149, in _schedule_execution
    _, actionexecutiondb = action_service.request(liveactiondb)
  File "/mnt/st2/st2common/st2common/services/action.py", line 98, in request
    liveaction = LiveAction.add_or_update(liveaction, publish=False)
  File "/mnt/st2/st2common/st2common/persistence/base.py", line 155, in add_or_update
    model_object = cls._get_impl().add_or_update(model_object)
  File "/mnt/st2/st2common/st2common/models/db/__init__.py", line 174, in add_or_update
    instance.save()
  File "/mnt/st2/virtualenv/local/lib/python2.7/site-packages/mongoengine/document.py", line 224, in save
    self.validate(clean=clean)
  File "/mnt/st2/virtualenv/local/lib/python2.7/site-packages/mongoengine/base/document.py", line 323, in validate
    raise ValidationError(message, errors=errors)
ValidationError: ValidationError (LiveActionDB:None) (Invalid dictionary key name - keys may not contain "." or "$" characters: ['parameters'])
2015-11-11 23:55:43,192 DEBUG [-] API call failed: None (exception_message='ValidationError (LiveActionDB:None) (Invalid dictionary key name - keys may not contain "." or "$" characters: [\'parameters\'])',exception_data={'comment': None, '_headers': None, 'conditional_response': False, 'detail': 'ValidationError (LiveActionDB:None) (Invalid dictionary key name - keys may not contain "." or "$" characters: [\'parameters\'])', '_app_iter': [''], '_status': '500 Internal Server Error', '_headerlist': [('Content-Type', 'text/html; charset=UTF-8'), ('Content-Length', '0')]},exception_class='HTTPInternalServerError')
2015-11-11 23:55:43,193 INFO [-] 2e24e698-6aa7-40d6-8122-b598b978bb44 - POST /v1/executions (remote_addr='172.168.60.1',method='POST',status_code='500 Internal Server Error',request_id='2e24e698-6aa7-40d6-8122-b598b978bb44',path='/v1/executions')
172.168.60.1 - - [11/Nov/2015 23:55:43] "POST /v1/executions HTTP/1.1" 500 646 0.021786
```

As much as I would like to just say "there's the list of characters you can't use in your object keys", we have a `core.remote` action which keys do indeed contain dots.

enykeev:
resolves #2198

### DIFF
--- a/st2common/st2common/models/db/action.py
+++ b/st2common/st2common/models/db/action.py
@@ -73,7 +73,7 @@ class ActionDB(stormbase.StormFoundationDB, stormbase.TagsMixin,
     runner_type = me.DictField(
         required=True, default={},
         help_text='The action runner to use for executing the action.')
-    parameters = me.DictField(
+    parameters = stormbase.EscapedDynamicField(
         help_text='The specification for parameters for the action.')
     notify = me.EmbeddedDocumentField(NotificationSchema)
 
--- a/st2actions/tests/unit/test_notifier.py
+++ b/st2actions/tests/unit/test_notifier.py
@@ -80,7 +80,8 @@ class NotifierTestCase(unittest2.TestCase):
                 self.tester.fail('Test failed')
 
     @mock.patch('st2common.util.action_db.get_action_by_ref', mock.MagicMock(
-        return_value=ActionDB(pack='core', name='local', runner_type={'name': 'run-local-cmd'})))
+        return_value=ActionDB(pack='core', name='local', runner_type={'name': 'run-local-cmd'},
+                              parameters={})))
     @mock.patch('st2common.util.action_db.get_runnertype_by_name', mock.MagicMock(
         return_value=RunnerTypeDB(runner_parameters={})))
     @mock.patch.object(Action, 'get_by_ref', mock.MagicMock(
--- a/st2common/st2common/models/db/execution.py
+++ b/st2common/st2common/models/db/execution.py
@@ -54,7 +54,7 @@ class ActionExecutionDB(stormbase.StormFoundationDB):
         help_text='The timestamp when the liveaction was created.')
     end_timestamp = ComplexDateTimeField(
         help_text='The timestamp when the liveaction has finished.')
-    parameters = me.DictField(
+    parameters = stormbase.EscapedDynamicField(
         default={},
         help_text='The key-value pairs passed as to the action runner & action.')
     result = stormbase.EscapedDynamicField(
--- a/st2common/st2common/models/db/liveaction.py
+++ b/st2common/st2common/models/db/liveaction.py
@@ -59,7 +59,7 @@ class LiveActionDB(stormbase.StormFoundationDB):
     action = me.StringField(
         required=True,
         help_text='Reference to the action that has to be executed.')
-    parameters = me.DictField(
+    parameters = stormbase.EscapedDynamicField(
         default={},
         help_text='The key-value pairs passed as to the action runner & execution.')
     result = stormbase.EscapedDynamicField(


Kami:
LGTM - good catch.

I'm wondering why we didn't do that before (probably an oversight :)).

Kami:
On a related note - if we don't do that yet we should probably come up with more strict validation rules for parameter name - e.g. something similar as we do for pack names.

lakshmi-kannan:
@Kami I don't know if one should EscapedDynamicField or EscapedDictField here?

Kami:
I believe the only difference is that `EscapedDictField` performs some extra validation to make sure that the value is dict and dict keys are strings, right?

If so and if the tests pass, we should probably use `EscapedDictField`.

lakshmi-kannan:
I am leaning towards DictField as well. About tests, they don't pass for me.

Kami:
If so we need to dig in and see if it's a legit failure or if some tests use bad fixtures or similar.

Kami:
Also, we should cherry-pick this onto v1.1.1 once it's ready.

manasdk:
As I think about it why is providing support for `.` a good idea? I am not sure this will work when we try to reference parameters in criteria or workflow I.e. and jinja context. 

Please use this parameter in a rule or actionchain and see it work properly. E.g. `-` are not allowed in jinja templates.

Kami:
True, that's a good point - so if anything, we should make the validation change :)

st2common/st2common/models/db/action.py
```diff
@@ -73,7 +73,7 @@ class ActionDB(stormbase.StormFoundationDB, stormbase.TagsMixin,
     runner_type = me.DictField(
         required=True, default={},
         help_text='The action runner to use for executing the action.')
-    parameters = me.DictField(
+    parameters = stormbase.EscapedDynamicField(
```
manasdk: DictField to EscapedDictField.

manasdk:
-1 to this change. I am not sure I see the value.

enykeev:
The cause of this change is that I've tried to pass execution (not execution id, but whole execution) as parameter and end up with an error on remote runner where execution results contain `.` in key (host cmd is run against). If we decide to disallow `.` inside params, we should do the same for the rest of our dicts.

enykeev:
The idea is that user may not have complete control over data he passes to the action so in some cases, he end up in the same situation I am now when I can't possibly use an action with the data I have.

enykeev:
As for Jinja, pretty sure it allows you to get an item of the mapping no matter what kind of characters its key contain providing you do that using square bracket notation.

Kami:
So, yeah, the problem is also the attribute value, not just the parameter name, right?

If so, I do see the problem. We should allow users to pass in a parameter value which is a dict and contains dots or other special characters in the keys (as @enykeev this could be arbitrary data which is not controlled by the user).

Having said that - it would still be a good idea to add some restrictions on parameter names (not values).

manasdk:
Thanks for the explanation. I see the point now. Without context these changes are very hard to reason.

Ok. I can see how you want to use this; my personal sense is that this border-lining abuse especially that part that user has no control over data being passed into the action seems like a red-herring but we use this pattern ourselves very often so I can't call this as incorrect usage.

Alright, you will need an `EscapedDynamicField` and cannot use an `EscapedDictField`. Although in both cases the escaping happens only at the top-level i.e. the mongodb addressable property in the dynamicfield anything inside the document will be ignored. We can get away with it for now but this may not actually work in a non-mongo setup but that is a problem for future us :)

FYI `.` and `$` are both restricted.

Also, you are right it is possible to use in jinja with the dictionary notation. I wonder how yaql would fare - maybe that would also work.

Kami:
LGTM, +1.

Let's get this merge and run it on the build server for a while before picking onto v1.1.1.

Kami:
Cherry picked onto v1.1.1.

PR #2205, opened 2015-11-13
Allow Mistral to talk to the StackStorm API using a non public URL

### LINKED ISSUE #2203: WebUI or Mistral - pick which won't work, when external IP not reachable from inside the box
Reported by @tonybaloney: 

he got internal and external IP address, and NAS. If mistral connects to internal api address as set up by . If so, WebUI can't connect to auth and API. And vice versa: if the API url is set to external IP address, Mistral can't connect to it because it's behind the NAT and not routable from within the box. 

In short, mistral needs to connect via internal ip, UI needs to connect to external IP/address, and we only have one URL parameter, the `api_url` in the `[auth]` section of `/etc/st2/st2.conf`

Temp workaround: via local /etc/hosts. 

@tonybaloney please add details if I missed smth. 

[stackstorm-community.slack.com link](https://stackstorm-community.slack.com/archives/community/p1447383434000310)

Kami:
This change allows operator to configure which URL Mistral uses to talk back to the StackStorm API.

For backward compatibility and by default, it uses full public API url. If needed, user can configure it to use a different URL (e.g. internal URL or local URL on single server deployments).

This should resolve #2203

### DIFF
--- a/st2common/st2common/util/api.py
+++ b/st2common/st2common/util/api.py
@@ -21,7 +21,9 @@ from st2common.util.url import get_url_without_trailing_slash
 
 __all__ = [
     'get_base_public_api_url',
-    'get_full_public_api_url'
+    'get_full_public_api_url',
+
+    'get_mistral_api_url'
 ]
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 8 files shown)


st2common/st2common/util/api.py
```diff
@@ -53,3 +55,19 @@ def get_full_public_api_url(api_version=DEFAULT_API_VERSION):
...
+    """
+    Return a URL which Mistral uses to talk back to the StackStorm API.
+
+    Note: If not provided it defaults to the public API url.
+    """
+    if cfg.CONF.mistral.api_url:
+        api_url = get_url_without_trailing_slash(cfg.CONF.mistral.api_url)
+        api_url = '%s/%s' % (api_url, api_version)
+    else:
+        LOG.warn('"mistral.api_url" not set, using auth.api_url')
```
lakshmi-kannan: auth.api_url? typo?

Kami: Nope, not a typo.

That's what it's using now and this value acts as a single source of truth for public API URL.

manasdk:
One more URL to add to the confusion :)

+1

Kami:
Cherry picked changes into v1.1.1.

PR #2209, opened 2015-11-13
spelt out what rbac means

jonico:
figured that not everybody knows what rbac stands for

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -25,7 +25,7 @@ Contents:
     traces
     cli
     authentication
-    rbac
+    rbac (role based access control)
     client_libraries
     reference <reference/index>
     troubleshooting/index


docs/source/index.rst
```diff
@@ -25,7 +25,7 @@ Contents:
     traces
     cli
     authentication
-    rbac
+    rbac (role based access control)
```
manasdk: Will check if it renders properly in the left nav.

Kami: Since it's a link you need to use `title <link syntax>`

Kami:
Thanks, but it actually already renders correctly on the website - http://docs.stackstorm.com/latest/ (see navigation bar on the left).

In this case "rbac" is a link to a file and the title (Role Based Access Control) is read from the referenced / link file (docs/source/rbac.rst).

manasdk:
I think we are ok here as Kami pointed out i.e. RBAC is already spelled out.

PR #2210, opened 2015-11-13
Use log.exception inside rabbit retry code

Kami:
This way we should also get the traceback back.


manasdk:
This could generate a lot of traceback but still safer to have more data in logs

+1

Kami:
Yeah, maybe we should only log it as error / exception on should_stop? Otherwise use debug with exc_info?

manasdk:
> Yeah, maybe we should only log it as error / exception on should_stop?
Yes that is ideal. However, I have not seen the final log state in the logs anyway so I would leave what you have. I suspect we will have to muck around in this area again.

PR #2213, opened 2015-11-17
Remove st2web from the st2ctl components list

enykeev:
The fact `st2web` shows up as `not running` during `st2ctl status` confuses the living hell out of people on a daily basis.


manasdk:
I think we should distinguish between NG and old style

enykeev:
Do we still support old style deployments with standalone st2web at all?

manasdk:
> Do we still support old style deployments with standalone st2web at all?

We still have http://docs.stackstorm.com/install/st2_deploy.html#scripted-installer. As long as that is the backup plan the answer to your question is yes.

enykeev:
Ok, then we need to figure out where we're going because clearly people get confused **a lot**. @dzimine your call

manasdk:
How about a separate list for NG and non-NG. I suspect we still use the variable `NG_INIT` and on the basis of the we can switch the component list.

enykeev:
I'm worried that it would bring more confusion. How do we ensure proper env variable is here in every case? If we don't, we won't solve the problem since people only really bother to read `st2ctl status` only when things go sour and in this situations checking if `NG_INIT` is set is just another hassle.

emedvedev:
Can this make it into release somehow? I agree with @enykeev: it's really confusing when you ask for status and see that something is not running, even I was puzzled the first couple of times.

I also think `NG_INIT` is fine: if we can't drop st2web status from scripted install, we should at least provide _some_ degree of compatibility. Yeah, `NG_INIT` can get screwed up, but that won't lead to anything horrible, anyway, so that's probably an acceptable risk.

dzimine:
Time is now. With 1.3.1 I am ditching st2deploy and all the support. So OK with this going in, on master for sure.

manasdk:
re-sync and merge away.

+1

PR #2214, opened 2015-11-17
Fix the value of "user" attribute in the context of execution created by alias execution API endpoint

Kami:
This pull request fixes the value of "user" attribute in the "context" of live action created by the alias execution API.

Previously, ``user`` was incorrectly set to system user (``stanley``) instead of the authenticated user which triggered the execution. Not setting the user correctly means RBAC and some other things won't work correctly - token which gets created by the action runner and is available for the duration of the action execution will be tied to the system user (which has no permissions by default) instead of the authenticated user (in workroom case, we create a special use which is used for ChatOps by hubot and we grant that user admin role).

This issue was spotted by @enykeev.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,6 +19,9 @@ in development
   validate the parameters before creating a TriggerDB object. (bug fix)
 * Update local runner so all the commands which are executed as a different user and result in
   using sudo set $HOME variable to the home directory of the target user. (improvement)
+* Fix a bug with a user inside the context of the live action which was created using alias
+  execution endpoint incorrectly being set to the system user (``stanley``) instead of the
+  authenticated user which triggered the execution. (bug fix)
 
 1.1.1 - November 13, 2015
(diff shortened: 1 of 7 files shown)


emedvedev:
LGTM. This issue is a blocker for most default st2 aliases, should we release it as a hotfix or backport it?

Kami:
Yeah, I think it's reasonable to backport it to v1.1.2 :)

enykeev:
Thanks, Kami. I'll wait for the build then make a testing pass and start publishing stuff one by one.

Kami:
Travis failure in master merge is related to mistral auth test - https://github.com/StackStorm/st2/commit/37f4daaf61a8b755ff0ab25c479898c403bf63cb#commitcomment-14442182.

I reverted it for now and will merge this PR.

Kami:
@enykeev I believe packages should be built now. Please let me know how end to end testing goes.

PR #2215, opened 2015-11-17
Modifying ActionAliasFormatParser to support more flexible chatops

emedvedev:
We have several problems with the way ActionAliasFormatParser is implemented now:
1. In hubot we have a regex pattern matching for chatops commands, which in some edge cases isn't caught by the back-end parser. This results in hubot sending a legitimate command which is not recognized or executed. 
2. The parser doesn't match multi-word commands if they aren't quoted or multi-line commands ever. It limits user's flexibility in passing long arguments or making a bot more responsive and, again, "human-like".
3. It would be nice to have full support for regexes in aliases in general: we'll be able to create more flexible commands.
4. @enykeev's catch: there's no support for default params in the middle of the string (his example was `st2 list {{ limit=999 }} actions`).

This PR solves all of the above by trying a different take on the parser: instead of iterating over a string catching values, we're transforming our format string into a regex (which is not new, because it's exactly the approach we take in `hubot-stackstorm`) and build a parameter list. Since the format becomes a regex, it also allows much greater flexibility and degree of control. It also simplifies parser greatly, because there's no need for specialized value parsers or an iterator anymore (this is why this PR has 10x more deletions than it has additions).

There's one thing on which I'll need input from @manasdk: in our parser we support matching of arbitrary key-value pairs without a format string, which is something I don't see the use case for, not even in theory. With the current approach it will not work, because we'll need a format string to match any parameters, but then again, I don't see why it existed in the first place. @manasdk, @Kami: let's discuss this.

To be clear, I'm talking about something like this case I found in tests:
```
alias_format = ''
param_stream = 'mixed a=foobar1 some more b=boobar2 text c=coobar3 yeah'
```
Why is it here? Do we use it? Do we plan to? If we do, I'd like to know where and, in all honesty, why. If not, it should be safe to merge this PR.

cc @enykeev @dzimine

### DIFF
--- a/st2common/st2common/models/utils/action_alias_utils.py
+++ b/st2common/st2common/models/utils/action_alias_utils.py
@@ -34,7 +34,21 @@ class ActionAliasFormatParser(object):
         # As there's a lot of questions about using regular expressions,
         # I'll try to be thorough when documenting this code.
 
-        # Firstly, we'll match parameters with default values in form of
+        # We're parsing the arbitrary key-value pairs at the end of the stream
+        # to support passing of parameters not specified in the format string,
+        # and cutting them from the stream as they're no longer needed.
+        # Possible values are quoted strings, a word, or anything inside "{}".
+        pairs_match = r'(?:^|\s+)(\S+)=("(.*?)"|\'(.*?)\'|({.*?})|(\S+))'
+        extra = re.match(r'.*?((' + pairs_match + r'\s*)*)$',
+                         self._param_stream, re.DOTALL)
+        if extra:
+            kv_pairs = re.findall(pairs_match,
+                                  extra.group(1), re.DOTALL)
+            for pair in kv_pairs:
+                result[pair[0]] = ''.join(pair[2:])
+            self._param_stream = self._param_stream.replace(extra.group(1), '')
+
+        # Now we'll match parameters with default values in form of
         # {{ value = parameter }} (and all possible permutations of spaces),
         # compiling them into a list.
         # "test {{ url = http://google.com }} {{ extra = Test }}" will become
--- a/st2common/st2common/models/utils/action_alias_utils.py
+++ b/st2common/st2common/models/utils/action_alias_utils.py
@@ -53,13 +53,13 @@ class ActionAliasFormatParser(object):
         # compiling them into a list.
         # "test {{ url = http://google.com }} {{ extra = Test }}" will become
         # [ ["url", "http://google.com"], ["extra", "Test"] ]
-        params = re.findall(r'{{\s*(.+?)\s*(?:=\s*[\'"]?(.+?)[\'"]?)?\s*}}',
+        params = re.findall(r'{{\s*(.+?)\s*(?:=\s*[\'"]?({.+?}|.+?)[\'"]?)?\s*}}',
                             self._format, re.DOTALL)
 
         # Now we're transforming our format string into a regular expression,
         # substituting {{ ... }} with regex named groups, so that param_stream
         # matched against this expression yields a dict of params with values.
-        reg = re.sub(r'\s+{{\s*(\S+)\s*=.+?}}',
+        reg = re.sub(r'\s+{{\s*(\S+)\s*=(?:{.+?}|.+?)}}',
                      r'(?:\s+[\'"]?(?P<\1>.+?)[\'"]?)?',
                      self._format)
         reg = re.sub(r'{{\s*(.+?)\s*}}',
--- a/st2common/st2common/models/utils/action_alias_utils.py
+++ b/st2common/st2common/models/utils/action_alias_utils.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import re
+from st2common.exceptions import content
 
 __all__ = [
     'ActionAliasFormatParser'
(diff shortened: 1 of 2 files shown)


emedvedev:
While we do allow to pass parameters in chatops commands without specifying them in a format string, I think this:
— should be limited to the end of the string, not just injecting params wherever you please;
— is not transparent to the end user;
— is a potential security risk if you want to "shield" some of the action parameters from modifying via alias.

Kami:
> 3. It would be nice to have full support for regexes in aliases in general: we'll be able to create more flexible commands.

That's true, but it also brings new potential security risks (e.g. pathological regexes). Luckily, only operators can add new aliases so that's probably fine as long as we make it clear to the people writing aliases on what kind of precautions they need to take.

Eventually, we should also switch to a regular expression library which doesn't support backtracking and supports constant time matching in all the places where a regex can be provided by a user (rule criteria value, and now aliases).

st2common/tests/unit/test_action_alias_utils.py
```diff
@@ -14,293 +14,64 @@
...
-        # Multiple params provided
-        alias_format = ''
-        param_stream = 'a=foobar1 b=boobar2 c=coobar3'
-        parser = ActionAliasFormatParser(alias_format, param_stream)
-        extracted_values = parser.get_extracted_param_value()
-        self.assertEqual(extracted_values, {'a': 'foobar1', 'b': 'boobar2', 'c': 'coobar3'})
-
-        # Multiple params provided
-        alias_format = ''
-        param_stream = 'a=foobar4 b=boobar5 c=coobar6'
```
Kami: I see a lot of tests good removed - does this means we don't fully support the old notation anymore?

Kami:
On a related note - we should really make the client dumb and move all the parsing to the backend so we don't need to maintain regexes in two places.

Or at least we could begin with constructing regexes on the server and sending it down to the client when retrieving aliases. This should be reasonable. @emedvedev wdyt?

st2common/tests/unit/test_action_alias_utils.py
emedvedev: As agreed, we do not support arbitrary pairs mixed with normal strings, i.e. `mixed a=foobar1 some more b=boobar2 text c=coobar3 yeah`. We discussed this with @jfryman and @manasdk, and agreed this is a rare case which we don't really need. Continuing to support them would result in unexpected (and, honestly, illogical) behavior in edge cases.

Rest of the tests are still there, and I've added a couple new ones, too.

emedvedev:
I'd like to do that eventually, yeah. I don't think we can include it in this release, but refactoring parts of hubot-stackstorm is on my list, and getting compiled regexes from the server would only make sense.
Overall, it's exactly the direction we're going: making the client thinner and dumber.

Kami:
On a related string parsing note, I saw this tweet the other day - https://twitter.com/mitsuhiko/status/666994266758057984

Not something we need to use, but still interesting and could come handy some time.

st2common/st2common/models/utils/action_alias_utils.py
```diff
@@ -13,222 +13,71 @@
...
+                            self._format, re.DOTALL)
+
+        # Now we're transforming our format string into a regular expression,
+        # substituting {{ ... }} with regex named groups, so that param_stream
+        # matched against this expression yields a dict of params with values.
+        reg = re.sub(r'\s+{{\s*(\S+)\s*=.+?}}',
+                     r'(?:\s+[\'"]?(?P<\1>.+?)[\'"]?)?',
+                     self._format)
+        reg = re.sub(r'{{\s*(.+?)\s*}}',
+                     r'[\'"]?(?P<\1>.+?)[\'"]?',
```
Kami: Ah, so we use backtracking here, this could become problematic with regard to pathological regex.

emedvedev: I couldn't come up with a good example to encounter catastrophic behavior unexpectedly or even consciously exploit it. There is backtracking happening, but we're matching pretty long non-repeating patterns (in case with the template params) or repeating patterns constrained by the end of string (in case with pairs). I see how backtracking is a problem theoretically, but not in this case.

emedvedev: That said, we should make sure it doesn't block or introduce a timeout.

Kami:
I"m not too happy with the backtracking regex stuff which could cause issues in the future.

For now though, I would feel a lot more confident if we add more tests (especially for edge cases and more complicated input strings).

emedvedev:
@Kami I've extended the test suite a little to check for defaults at the middle of the string, as well as different combinations of format string parameters + default parameters + arbitrary pairs. Should be comprehensive enough.

Kami:
LGTM, +1, but I would also wait on another +1

emedvedev:
Makes sense. No rush here anyway.

st2common/st2common/models/utils/action_alias_utils.py
manasdk: Will it handle default params of the kind properly?

`deploy instance {{name}} {{region = {{system.default_region}} }}`

emedvedev: No. Do we support them now or is it a feature request? I don't recall a test case like this. I suppose it wouldn't be too hard to add though, so let me try.

manasdk: I suspect we sorta just managed to support them and/or adding support was possible anyway. I am trying to see if we shut the door on something like this with the current approach.

Will be nice if we support but I would not sweat it much.

st2common/tests/unit/test_action_alias_utils.py
```diff
@@ -14,293 +14,64 @@
...
-            self.assertTrue(False, 'Parsing failure expected.')
-        except content.ParseException:
-            self.assertTrue(True)
+from st2common.models.utils.action_alias_utils import ActionAliasFormatParser
 
 
 class TestActionAliasParser(TestCase):
-    def test_default_key_value_param_parsing(self):
-        # Empty string
+    def testEmptyString(self):
```
manasdk: while at it can you also use `test_empty_string` naming convention.

st2common/tests/unit/test_action_alias_utils.py
```diff
@@ -14,293 +14,64 @@
...
+                                            'b': 'boobar2 3 4',
+                                            'c': 'coobar3 4',
+                                            'd': '{"a": "b"}'})
 
-        # Param with quotes, make sure they are stripped
-        alias_format = ''
-        param_stream = 'mixed a="foobar1"'
+        # Params along with a "normal" alias format
+        alias_format = '{{ captain }} is my captain'
+        param_stream = 'Malcolm Reynolds is my captain weirdo="River Tam"'
```
manasdk: We need more firefly references!

manasdk:
Alright I powered through this one. I literally do no understand some of the pathological regex concern so no input from me in this case.

+1

I like the change it makes sense. Even though I am still not enamored to regex it seems like the right tool for the job :)

st2common/tests/unit/test_action_alias_utils.py
emedvedev: Sure. The original testing suite had both camelCase and underscores, so I just went with the former. Will change. :)

jfryman:
@manasdk regex is the only way. :) Embrace it.

st2common/tests/unit/test_action_alias_utils.py
jfryman: :metal:

Extra points to add Firefly quotes to Hubot fun. :)

Kami:
@manasdk https://en.wikipedia.org/wiki/ReDoS

emedvedev:
The latest commit includes important things @manasdk requested:
1. Support for templated defaults (`deploy instance {{name}} {{region = {{system.default_region}} }}`) and a test case for it.
2. Underscores in the test suite.
3. One more Firefly reference.

By the power vested in me (in my dreams) by captain Malcolm Reynolds, I now merge this PR.

PR #2217, opened 2015-11-17
New "timeout" action execution state and retry on failure/timeout policy

Kami:
This pull request introduces a new "timeout" action execution state. This state is a special type of a failure and represents an execution timeout (previously you couldn't directly identify a timeout - you would need to manually parse the error attribute to figure that out).

In addition to that, it also adds a new "retry" policy which allows executions to be automatically retried in case of a failure or a timeout.

TODO:

- [x] Tests
- [x] Examples
- [x] Documentation
- [x] Add more debug logging to st2notifier for easier debugging
- [x] Update paramiko runner so it uses new timeout status constant


st2actions/st2actions/policies/retry.py
```diff
@@ -0,0 +1,141 @@
...
+
+        return target
+
+    def _get_live_action_retry_count(self, live_action_db):
+        """
+        Retrieve current retry count for the provided live action.
+
+        :rtype: ``int``
+        """
+        # TODO: Ideally we would store retry_count in zookeeper or similar and use locking so we
```
Kami: Another option would be to use a new MongoDB collection and atomic increment and decrement operations on a counter type (this should work fine for now).

What do others think?

st2actions/st2actions/policies/retry.py
```diff
@@ -0,0 +1,141 @@
...
+class ExecutionRetryPolicyApplicator(ResourcePolicyApplicator):
+    def __init__(self, policy_ref, policy_type, retry_on, max_retry_count=2):
+        """
+        :param retry_on: Condition to retry the execution on.
+        :type retry_on: ``str``
+
+        :param max_retry_count: Maximum number of times to try to retry an action.
+        :type max_retry_count: ``int``
+        """
+        # TODO: Also support sleep delay before retrying
```
Kami: We could implement simple non-fail safe sleep / delay using `eventlet.sleep`. This would work fine for small delays (< 5 seconds), but it's obviously not fail safe (e.g. if notifier is restarted / dies). A more robust and safe approach would be to use DELAYED state, but this would mean we need to update existing concurrency policies which already use that state (we would need to differentiate between different type of delays).

Feedback is welcome.

st2actions/st2actions/policies/retry.py
Kami: Actually, now that I think of it - this should work fine as long as we use round robin and not fan out message routing.

If we don't do that, other things would be broken as well in st2notifier if we run multiple instances.

st2actions/st2actions/policies/retry.py
manasdk: I guess difference between time based delay and concurrency based delay.  I feel that the current `DELAYED` state should probably be renamed to `BLOCKED` then there will be no confusion.

st2actions/st2actions/policies/retry.py
Kami: Yeah, ignore my original comment. I just verified and we use topic exchange so we are fine here.

We don't need additional code, it should already work fine even with multiple instances. Will remove the invalid code comment.

manasdk: My initial reaction here is that the retry_context does not belong to a single execution. Keeping it separate does have the extra benefits of being able to manage this information better. However, clients that care to demonstrate this information will then have to make multiple requests which can be bandaged over by using the views endpoint.

st2actions/st2actions/policies/retry.py
```diff
@@ -0,0 +1,141 @@
...
+        return retry_count
+
+    def _re_run_live_action(self, live_action_db):
+        retry_count = self._get_live_action_retry_count(live_action_db=live_action_db)
+
+        # Add additional policy specific info to the context
+        context = getattr(live_action_db, 'context', {})
+        new_context = copy.deepcopy(context)
+        new_context['policies'] = {
+            'applied_policy': self._policy_ref,
```
manasdk: Taking up `policies` might be generic since other policies could also try to write to the context.

st2common/st2common/policies/base.py
```diff
@@ -53,8 +62,29 @@ def apply_after(self, target):
...
+        Return a safe string which can be used as a lock name.
+
+        :param values: Dictionary with values to use in the lock name.
+        :type values: ``dict``
+
+        :rtype: ``st``
+        """
+        lock_uid = []
+
+        for key, value in six.iteritems(values):
```
manasdk: Does the iteritems always perform a sorted iteration?

manasdk:
I really like the idea of this policy but a couple of questions that are in my mind are -

1. UI will show separate executions - is this ok for the user?
2. How can we preserve the ChatOps context?
2. What happens when the retry is for an Action that runs on a task in an ActionChain.

st2common/st2common/policies/base.py
Kami: Good point - need to fix this.

st2actions/st2actions/policies/retry.py
Kami: Yeah, I was thinking about it, but I couldn't decide on what to do.

Another option is `['policies']['retry']`, but this might be too nested?

Kami:
@manasdk 

> UI will show separate executions - is this ok for the user?

Correct, it's a similar behavior to the "re-run". It is indeed a new execution. I'm open for future UI improvements though. We have data we need to track it back to the execution which was retried so we can do something useful with it in the future. Maybe an easy way to find all the executions which belong to the original execution which was retried and vice versa.

> How can we preserve the ChatOps context?

Unlike in re-run case, here we preserve the original execution context - we just add some additional retry policy info to the context.

> What happens when the retry is for an Action that runs on a task in an ActionChain.

Same thing if you use re-run to re-run a particular task in the workflow (I've done this before). There are cases where it makes sense and it's useful and there are cases where it's not.

Having "retry_parent_on_failure" or something similar might be another option, but this gets more complicated and it's probably something which should be done inside the workflow. I'm open to discussion though.

PR #2218, opened 2015-11-17
Minor fix to retry in mistral querier

m4dcoder:
Minor code refactoring and unit tests added.

### DIFF
--- a/st2actions/tests/unit/test_mistral_querier_v2.py
+++ b/st2actions/tests/unit/test_mistral_querier_v2.py
@@ -16,18 +16,28 @@
 import copy
 import datetime
 import json
+import requests
 import uuid
 
 import mock
+from mock import call
 
 from mistralclient.api.v2 import executions
 from mistralclient.api.v2 import tasks
+from oslo_config import cfg
+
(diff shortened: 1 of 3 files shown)


manasdk:
The goal is to retry even on a ConnectionError - correct?

m4dcoder:
Yes. The goal is to exponential retry on ConnectionError. This patch though is just refactoring to be consistent with using they retrying decorator.

st2actions/tests/unit/test_mistral_querier_v2.py
```diff
@@ -230,3 +240,78 @@ def test_query(self):
...
+                'state_info': MOCK_WF_EX.state_info
+            }
+        }
+
+        for task in expected['tasks']:
+            task['input'] = json.loads(task['input'])
+            task['result'] = json.loads(task['result'])
+            task['published'] = json.loads(task['published'])
+
+        self.assertEqual(action_constants.LIVEACTION_STATUS_SUCCEEDED, status)
```
Kami: Do we have any test case where a maximum number of retries is reached and it still fails?

PR #2219, opened 2015-11-18
Move chatops pack to st2 contrib/

enykeev:
Per @jfryman suggestion, moving it here to have more control over where and when it gets deployed


Kami:
I'm fine with this pack living here.

I'm still a bit mind blown by the whole Jinja thing (in general, I'm OK with the template approach, it just seems there is a lot / too much of logic in the template), but anyway.

LGTM for now.

enykeev:
Later, we may want to split it somehow. Really, with this thing I wanted to see whether we can actually use Jinja all the way through.

My previous experience with it in Github Pages wasn't really successful, but ^^ proved me wrong.

PR #2229, opened 2015-11-20
Ack message formatting in AliasExecution

emedvedev:
1. Sending HTTP 201 instead of 200.
2. Formatting ack with jinja if the format string is specified.
3. Sending more context along with the execution id and formatted string for a client to use if desired.

Tests are TBD; I’m submitting this to review and to provide context for the hubot-stackstorm PR that will follow.

### DIFF
--- a/st2api/st2api/controllers/v1/aliasexecution.py
+++ b/st2api/st2api/controllers/v1/aliasexecution.py
@@ -97,8 +97,7 @@ class ActionAliasExecutionController(rest.RestController):
 
         if action_alias_db.ack and 'format' in action_alias_db.ack:
             result.update({
-                'message': render({'alias': action_alias_db.ack['format']},
-                                  {'alias': result})['alias']
+                'message': render({'alias': action_alias_db.ack['format']}, result)['alias']
             })
 
         return result
--- a/st2api/st2api/controllers/v1/aliasexecution.py
+++ b/st2api/st2api/controllers/v1/aliasexecution.py
@@ -16,9 +16,7 @@
 import jsonschema
 import pecan
 import six
-import jinja2
 from pecan import rest
-
 from st2common import log as logging
 from st2common.models.api.base import jsexpose
 from st2common.exceptions.db import StackStormDBObjectNotFoundError
@@ -34,6 +32,7 @@ from st2common.services import action as action_service
 from st2common.util import action_db as action_utils
 from st2common.util import reference
 from st2common.util.api import get_requester
+from st2common.util.jinja import render_values as render
 from st2common.rbac.types import PermissionType
 from st2common.rbac.utils import assert_request_user_has_resource_db_permission
 
@@ -97,10 +96,9 @@ class ActionAliasExecutionController(rest.RestController):
         }
 
         if action_alias_db.ack and 'format' in action_alias_db.ack:
-            jinja = jinja2.Environment(trim_blocks=True, lstrip_blocks=True)
-            jinja.tests['in'] = lambda item, list: item in list
             result.update({
-                'message': jinja.from_string(action_alias_db.ack['format']).render(result)
+                'message': render({'alias': action_alias_db.ack['format']},
+                                  {'alias': result})['alias']
             })
 
         return result
--- a/st2api/st2api/controllers/v1/aliasexecution.py
+++ b/st2api/st2api/controllers/v1/aliasexecution.py
@@ -16,13 +16,16 @@
 import jsonschema
 import pecan
 import six
+import jinja2
 from pecan import rest
 
 from st2common import log as logging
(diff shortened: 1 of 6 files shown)


st2api/st2api/controllers/v1/aliasexecution.py
```diff
@@ -88,7 +91,19 @@ def post(self, payload):
                                              notify=notify,
                                              context=context)
 
-        return str(execution.id)
+        result = {
```
Kami: That's backward incompatible, right?

emedvedev: Yes, and we'll be shipping the new version of `hubot-stackstorm` at the same time this PR is released. Hubot, on the other hand, will actually be compatible with both new and old versions (they're pretty easy to separate because they send different HTTP status codes).

manasdk: But this itself is not backwards compatible i.e. an old hubot will not work with this change.I guess this is ok?

emedvedev: I think it's fine. We discussed this with @enykeev, and I can't see a realistic scenario where hubot-stackstorm will be older than st2. Even if that's the case, i.e. someone did an incomplete upgrade, we can just ask to do an `npm update`.

That's exactly why I'm asking to review this PR though: maybe you guys will see a case we overlooked where it can become a problem.

st2api/st2api/controllers/v1/aliasexecution.py
```diff
@@ -88,7 +91,19 @@ def post(self, payload):
...
+            'execution': execution,
+            'actionalias': ActionAliasAPI.from_model(action_alias_db)
+        }
+
+        if action_alias_db.ack and 'format' in action_alias_db.ack:
+            jinja = jinja2.Environment(trim_blocks=True, lstrip_blocks=True)
```
manasdk: Any reason this 2 lines of jinja config are applicable only in this context i.e. any reason other place where we use jinja this may not be required? 

Please see if https://github.com/StackStorm/st2/blob/master/st2common/st2common/util/jinja.py#L127 can be used here so we have a consistent jinja experience though stackstorm.

also, check if those 2 lines should be included.

emedvedev: These two lines, as well as the "in" comparison, is mostly a matter of convenience and personal preference. If you're ok with transferring them into util (trimming just makes sense to get rid of excessive whitespace, and "in" is just useful), I can use the method. In fact, using util is probably better anyway, for the sake of consistency.

manasdk: Ok. Then please move the changes to util and use utils.

Thanks!

st2api/st2api/controllers/v1/aliasexecution.py
manasdk: Since we do not prevent an old hubot-stackstrom from connection to a new stackstorm I wonder what the error will even look like. If you can KB/document this i.e. `If you see X it is time to upgrade hubot-stackstorm` then it would help us.

Ideally, within some reasonable window old client should be supported by new server. However, this is not a concern for this endpoint alone.

emedvedev: Since we change the status code and hubot-stackstorm catches `200` in the old version, I think it'll yield a "error: status 201" error or something like that. It's not very informative, but at least it's a well-defined error with the only probable cause being the need to update, because nothing else will ever give 201 status at this endpoint. :) As soon as we get a KB, I'll put a message there, of course.

I tried to find a good way to make this backwards-compatible without introducing extra flags or parameters, and, honestly, couldn't, because old hubot expects a string.

emedvedev:
@manasdk I'm using Jinja from utils now, and I just remembered @enykeev's ChatOps pack uses it the same way I did. We should probably use utils there, too.

st2api/st2api/controllers/v1/aliasexecution.py
enykeev: Right now, when st2client.js receives 200 code on POST, it throws an error (since it expects 201). We're catching the error in hubot-stackstorm (https://github.com/StackStorm/hubot-stackstorm/blob/master/scripts/stackstorm.js#L209-L217), checking status and constructing an object resembling expected response.

If you run a new hubot-stackstorm against an old st2, it would act as if no additional parameters are set. If you run an old hubot against a new st2, it would just output ypu a status code and a message body into the chat (https://github.com/StackStorm/hubot-stackstorm/blob/v0.2.5/scripts/stackstorm.js#L180)

emedvedev:
@manasdk @Kami: this PR is good to merge; in addition I hand-tested with our hubot branch supporting ack. Do you have any additional questions?

PR #2232, opened 2015-11-21
Update "packs.install" and "packs.setup_virtualenv" action so it supports "env" parameter

Kami:
This change allows user to pass ``env`` parameter to ``packs.setup_virtualenv`` and ``packs.install`` action.

This comes handy if user wants pip to use an HTTP(s) proxy (HTTP_PROXY and HTTPS_PROXY 
environment variable) when installing pack dependencies.

Keep in mind that this is not the whole solution for many proxy related issues - this is just a stop gap for packs.install.

In the mean time we need to decide if we are going to support proxies or not and if we will, we need to spend time to make the whole system work with proxies. Keep in mind that there are many moving pieces and many libraries and tools we depend on so supporting proxies everywhere might be a huge pain and / or impossible.


Kami:
Hm, I totally thought this PR has already been merged.

PR #2233, opened 2015-11-23
Retry policy improvements

Kami:
* Update action chain runner so it reports chain status as timed out if the last task times out (previously, it was reported as failed).
* Fix infinite loop in the action chain runner which happened if user canceled a task (action) in the chain while the chain was running
* Update notifier so it correctly checks for all the completed and failed states
* Reduce code duplication by utilizing constants from the actions_constants module


st2actions/st2actions/runners/actionchainrunner.py
```diff
@@ -366,36 +371,42 @@ def run(self, action_parameters):
...
-                        trace = traceback.format_exc(10)
-                        top_level_error = {
-                            'error': error,
-                            'traceback': trace
-                        }
-                        # reset action_node here so that chain breaks on failure.
-                        action_node = None
-                        break
-                else:
+                if self._stopped:
```
manasdk: +1 to the early return

st2actions/st2actions/runners/actionchainrunner.py
```diff
@@ -494,10 +509,8 @@ def _run_action(self, liveaction, wait_for_completion=True):
             LOG.exception('Failed to schedule liveaction.')
             raise e
 
-        while (wait_for_completion and
-               liveaction.status != LIVEACTION_STATUS_SUCCEEDED and
-               liveaction.status != LIVEACTION_STATUS_FAILED):
-            eventlet.sleep(1)
+        while (wait_for_completion and liveaction.status not in COMPLETED_STATES):
```
manasdk: Oh man such a terrible bug. How did you notice?

Kami: Randomly when checking the code :P

lakshmi-kannan: I am not sure there is a bug here. The new code reads better though.

Kami: Previously, if `wait_for_completion` was `True` (it is by default and we never explicitly pass `False`) and if user canceled the action (task) while it was still running this while loop would never finish since termination condition would never be met (liveaction status would be set to canceled for which we didn't check).

PR #2236, opened 2015-11-25
RFR: Purge models 

lakshmi-kannan:
TODO:

* - [X] Tests for purge executions
* - [X] Purge trigger instances
* - [X] Docs
* - [X] Changelog

EXAMPLES:

https://gist.github.com/lakshmi-kannan/c7181d4bf6f5ab910bc3
https://gist.github.com/lakshmi-kannan/af62e7872fbc61be4b1f


tools/purge_executions.py
```diff
@@ -58,101 +55,100 @@ def do_register_cli_opts(opts, ignore_errors=False):
...
+                   help='Will delete data older than ' +
+                   'this timestamp. (default 48 hours). ' +
+                   'Example value: 2015-03-13T19:01:27.255542Z'),
+        cfg.StrOpt('action-ref', default='',
+                   help='action-ref to delete executions for.')
+    ]
+    _do_register_cli_opts(cli_opts)
+
+
 def _purge_action_models(execution_db):
```
Kami: action -> live_action :P

tools/purge_executions.py
```diff
@@ -58,101 +55,104 @@ def do_register_cli_opts(opts, ignore_errors=False):
...
+            try:
+                LiveAction.delete(liveaction_db)
+            except:
+                LOG.exception('Zombie LiveAction left in db: %s. Exception: %s',
+                              liveaction_db)
+
 
+def _should_delete(execution_db, action_ref, timestamp):
+    if execution_db.status in IN_PROGRESS_STATES:
+        return False
```
lakshmi-kannan: @manasdk @Kami Need some opinions here. I made a choice that if executions older than cut off timestamp are not in DONE states (succeeded, failed, timedout, canceled), I leave them in the DB. What do you guys think? I have no strong opinions here.

DoriftoShoes: I think we should give them the option.  Leaving running by default is fine but an arg to remove all of them would be good.

lakshmi-kannan: @DoriftoShoes Fixed in 4ae86f6

DoriftoShoes: w00t.

DoriftoShoes:
+1

tools/purge_executions.py
manasdk: Having executions stuck in running was actually a pain-point so the option is good. This is especially annoying in an environment with policies where a stale running execution prevents new executions from spawning.

docs/source/troubleshooting/purging_old_data.rst
```diff
@@ -0,0 +1,56 @@
...
+a lot of executions happening/triggers coming in, database fills up. If you are
+looking for a way to purge old data in bulk for performance reasons or cleaning up the db,
+the following tools will help you do so.
+
+Purging executions older than some timestamp
+--------------------------------------------
+
+::
+
+    tools/purge_executions.py --timestamp="2015-11-25T21:45:00.000000Z"
```
manasdk: I wish we could say `tools/purge_executions.py --older_than="1week"`

st2tests/st2tests/fixtures/generic/liveactions/liveaction4.yaml
```diff
@@ -0,0 +1,17 @@
+---
```
manasdk: damn generic pack!

tools/purge_executions.py
```diff
@@ -58,101 +56,110 @@ def do_register_cli_opts(opts, ignore_errors=False):
...
-                    execution_db.start_timestamp < timestamp)
-        else:
-            return execution_db.start_timestamp < timestamp
+    LOG.info('Purging executions older than timestamp: %s' %
+             timestamp.strftime('%Y-%m-%dT%H:%M:%S.%fZ'))
 
     # XXX: Think about paginating this call.
-    filters = {'start_timestamp__lt': isotime.parse(timestamp)}
+    filters = {'end_timestamp__lt': isotime.parse(timestamp)}
     executions = ActionExecution.query(**filters)
```
manasdk: Likely a richer filter i.e. taking into account 'action-ref' and 'purge-incomplete' would be better.

Wonder if there is a way to only retrieve ids instead of full objects.

Btw, I think we can even leverage https://github.com/StackStorm/st2/blob/master/st2common/st2common/models/db/__init__.py#L183. This would avoid unnecessary reads.

tools/purge_executions.py
```diff
@@ -58,101 +56,110 @@ def do_register_cli_opts(opts, ignore_errors=False):
...
-    print('#### Total number of executions to delete: %d' % len(executions_to_delete))
+    executions_to_delete = [execution for execution in executions
+                            if _should_delete(
+                                execution_db=execution, action_ref=action_ref,
+                                timestamp=timestamp,
+                                purge_incomplete=purge_incomplete)]
+    LOG.info('#### Total number of executions to delete: %d' % len(executions_to_delete))
 
     # Purge execution and liveaction models now
     for execution_db in executions_to_delete:
```
manasdk: We have to assume this will be a large set. Do you think a bulk delete if you choose not to use delete_by_query?

tools/purge_executions.py
```diff
@@ -58,101 +56,110 @@ def do_register_cli_opts(opts, ignore_errors=False):
...
-                   help='Will delete data older than ' +
-                   'this timestamp. (default 48 hours). ' +
-                   'Example value: 2015-03-13T19:01:27.255542Z'),
-        cfg.StrOpt('action-ref', default='',
-                   help='action-ref to delete executions for.')
-    ]
-    do_register_cli_opts(cli_opts)
-    config.parse_args()
+    _register_cli_opts()
+    common_setup(config=config, setup_db=True, register_mq_exchanges=True)
```
manasdk: Hmm, I wonder if we should register exchanges.

tools/purge_executions.py
```diff
@@ -58,101 +56,110 @@ def do_register_cli_opts(opts, ignore_errors=False):
...
-    # Connect to db.
-    db_setup(cfg.CONF.database.db_name, cfg.CONF.database.host, cfg.CONF.database.port,
-             username=username, password=password)
+    purge_incomplete = cfg.CONF.purge_incomplete
 
     if not timestamp:
-        now = datetime.now()
-        timestamp = now - timedelta(days=DEFAULT_TIMEDELTA_DAYS)
+        LOG.error('Please supply a timestamp for purging models. Aborting.')
+        return 1
```
manasdk: +1

Explicit better than implicit.

tools/purge_executions.py
```diff
@@ -58,101 +56,110 @@ def do_register_cli_opts(opts, ignore_errors=False):
...
     else:
         timestamp = datetime.strptime(timestamp, '%Y-%m-%dT%H:%M:%S.%fZ')
+        timestamp = timestamp.replace(tzinfo=pytz.UTC)
```
manasdk: argh! Stupid TZs :P

tools/purge_executions.py
lakshmi-kannan: Good point. I'll make this change.

tools/purge_executions.py
lakshmi-kannan: Copy Pasta.

tools/purge_executions.py
```diff
@@ -58,101 +56,110 @@ def do_register_cli_opts(opts, ignore_errors=False):
...
-    _purge_executions(timestamp=timestamp, action_ref=action_ref)
+    purge_executions(timestamp=timestamp, action_ref=action_ref, purge_incomplete=purge_incomplete)
 
-    # Disconnect from db.
-    db_teardown()
+    common_teardown()
 
 if __name__ == '__main__':
-    main()
+    sys.exit(main())
```
manasdk: sys.exit(None) is exit code 0. I guess that make sense :P

tools/purge_executions.py
```diff
@@ -58,101 +56,110 @@ def do_register_cli_opts(opts, ignore_errors=False):
...
-            return (execution_db.liveaction['action'] == action_ref and
-                    execution_db.start_timestamp < timestamp)
-        else:
-            return execution_db.start_timestamp < timestamp
+    LOG.info('Purging executions older than timestamp: %s' %
+             timestamp.strftime('%Y-%m-%dT%H:%M:%S.%fZ'))
 
     # XXX: Think about paginating this call.
-    filters = {'start_timestamp__lt': isotime.parse(timestamp)}
+    filters = {'end_timestamp__lt': isotime.parse(timestamp)}
```
manasdk: I don't think this filter will return the older running ones since those would not have an end_timestamp

tools/purge_trigger_instances.py
```diff
@@ -0,0 +1,113 @@
...
+    if not timestamp:
+        LOG.error('Specify a valid timestamp to purge.')
+        return
+
+    LOG.info('Purging trigger instances older than timestamp: %s' %
+             timestamp.strftime('%Y-%m-%dT%H:%M:%S.%fZ'))
+
+    # XXX: Think about paginating this call.
+    filters = {'occurrence_time__lt': isotime.parse(timestamp)}
+    instances = TriggerInstance.query(**filters)
```
manasdk: I think you should use delete_by_query here; with 10K+ objects making as many DB calls seems a bit much.

tools/purge_trigger_instances.py
```diff
@@ -0,0 +1,113 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""
+A utility script that purges trigger instances older than certain
+timestamp.
+
+*** RISK RISK RISK. You will lose data. Run at your own risk. ***
```
manasdk: lol

It is called purge not sure how that is not self evident.

manasdk:
I suspect there is 1 real bug and some optimizations which we can work on later.

tools/purge_executions.py
Kami: > @manasdk @Kami Need some opinions here. I made a choice that if executions older than cut off timestamp are not in DONE states (succeeded, failed, timedout, canceled), I leave them in the DB. What do you guys think? I have no strong opinions here.

I think that is a reasonable default, but also need to take into consideration what phool and manasdk say (looks like it's already done).

tools/purge_trigger_instances.py
lakshmi-kannan: Yeah, we had this discussion last time. Some profiling here is in order. I'll take this up in a subsequent PR.

lakshmi-kannan:
@manasdk I got some optimizations and fixed the bug related to in progress executions. Have a look now. I am still looking through delete_by_query.

manasdk:
I am ok with this PR. I doubt we will revisit this script anytime soon so whatever you do is likely going to be picked up now. If the delete_by_query is proving to be too much trouble feel free to ignore for now since the functionality is coded up and that is more important.

Is it fair to assume the run time for 10K execution is not atrocious i.e. less than 5min?

manasdk:
+1

lakshmi-kannan:
>Is it fair to assume the run time for 10K execution is not atrocious i.e. less than 5min

I have no idea. Like I said, some profiling is due.

PR #2238, opened 2015-11-25
Support "representation+value" objects in format strings

emedvedev:
As previously discussed with the extended schema.

### DIFF
--- a/st2api/st2api/controllers/v1/aliasexecution.py
+++ b/st2api/st2api/controllers/v1/aliasexecution.py
@@ -107,7 +107,13 @@ class ActionAliasExecutionController(rest.RestController):
         return (tokens[0], tokens[1] if len(tokens) > 1 else None)
 
     def _extract_parameters(self, action_alias_db, format_str, param_stream):
-        if action_alias_db.formats and format_str in action_alias_db.formats:
+        formats = []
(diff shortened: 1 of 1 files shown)


st2api/st2api/controllers/v1/aliasexecution.py
```diff
@@ -107,7 +107,13 @@ def _tokenize_alias_execution(self, alias_execution):
         return (tokens[0], tokens[1] if len(tokens) > 1 else None)
 
     def _extract_parameters(self, action_alias_db, format_str, param_stream):
-        if action_alias_db.formats and format_str in action_alias_db.formats:
+        formats = []
+        for formatstring in action_alias_db.formats:
+            if 'representation' in formatstring:
```
manasdk: >>> x = 'arepresentationb'
>>> 'representation' in x
True
>>> x = {'representation': ''}
>>> 'representation' in x
True

manasdk:
Tests please :P

st2api/st2api/controllers/v1/aliasexecution.py
emedvedev: Good point. Format string can actually be a string with "representation" in it. :) Will fix.

PR #2240, opened 2015-11-26
ChatOps: Skipping post_message when result output is disabled

emedvedev:
As discussed with @manasdk, adding a dummy action to execute at `format_result`'s failure, thus avoiding the failure of the whole action-chain. Not an ideal solution, but better than any other option we currently have.


emedvedev:
@enykeev This PR is a blocker. We need to either merge it or have a trumping PR with what you proposed (the whole chain failing). Let's clear our heads, have a cup of coffee and agree on a solution. :) I don't mind either, but would prefer this PR's way (branching the chain).

emedvedev:
As discussed, reverted to failing loudly.

PR #2243, opened 2015-11-28
in-requirements fixes: osslo.config depends on stevedore, thus we mus…

dennybaa:
…t provide it explicitly


Kami:
I'm slightly confused about this change.

Why do we need to list it in in-requirements.txt if it's a transitive dependency and not a direct dependency of our stuff outside of st2auth?

We depend on many libraries and those libraries depend on many other dependencies (transitive dependencies) and we obviously don't and shouldn't list all those dependencies in in-requirements.txt.

Either I'm missing something or this PR is not fixing the right thing :)

dennybaa:
Yeah you are apparently right) However I need you guidance since I don't now how to work it around,
Oh man hold on this, I just figured out how to work this around probably)

But yeah, it indeed fails, due to my current implementation. However it should not, the problem is that I didn't use top-level `requirements.txt`... When I generate and install requirements for each component separately in parallel (these operations conflict). Basically, 

**st2common** common has transitive *stevedore* dependency as well as **st2auth**, say it if *"install requirements process handled in parallel"*, we run into races.
Namely `st2auth` can uninstall newer version of *stevadore* (since it requires an outdated one), however **st2common** might be not built by that time and that is what I see:



So as it's clear, I guess I need to convert to generating one global `requirements.txt` from all of the st2*/in-requirements.txt then I won't have these kind of races.

Kami:
To clarify - st2auth is different since it also has a direct dependency on stevedore.

As far as the race goes - it seems like it's an issue lies in the build process and that's where it should also be fixed. Not fixing the root cause will just cause more issues in the future.

dennybaa:
Yeah, I totally agree) Just opened the ticket and suddenly come and found some path... Apparently I will close the PR soon :) Thank you!

Kami:
Great and np :)

dennybaa:
Thank you. Fixed this.

PR #2248, opened 2015-11-30
Representation testing and "in" fix

emedvedev:
Addressing requests from @manasdk at #2238


Kami:
LGTM, +1

Although, `isinstance(foo, dict)` would probably be sufficient in this case since we load this info from YAML so it can't really be any special type which behaves like a dict / implements dict interface.

st2api/st2api/controllers/v1/aliasexecution.py
```diff
@@ -109,7 +110,8 @@ def _tokenize_alias_execution(self, alias_execution):
     def _extract_parameters(self, action_alias_db, format_str, param_stream):
         formats = []
         for formatstring in action_alias_db.formats:
-            if 'representation' in formatstring:
+            if (isinstance(formatstring, collections.Mapping) and
+               'representation' in formatstring and formatstring['representation']):
```
Kami: `and formatstring('representation', None)` is slightly shorter and has the same effect :)

emedvedev:
@Kami thanks! Fixed. :)

manasdk:
+1

PR #2251, opened 2015-11-30
Add documentation for using YAQL in Mistral

m4dcoder:
Add a quick start guide to using YAQL in Mistral.

### DIFF
--- a/docs/source/mistral_yaql.rst
+++ b/docs/source/mistral_yaql.rst
@@ -1,0 +1,201 @@
+Mistral + YAQL
+==============
+YAQL is typically used for simple conditional evaulation and data transformation in Mistral 
+workflows. There will be many cases where you did not author the actions but there's a need to 
+decide from the result of the action whether to continue or there's a need to transform the 
+result to another value or structure for the next action in the workflow.
+
+Here are some examples of usages.
+
+* Select values for a key from a list of dictionary.
+* Filter the list where one or more fields match condition(s).
+* Transform a list to dictionary or vice versa.
+* Simple arithmetics.
+* Evaluation of boolean logic.
+* Any combination of select, filter, transform, and evaluate.
+
+.. note::
+
+    Please refer to offical OpenStack documentation for Mistral and YAQL. The documentation here
+    is meant to help |st2| users get a quick start. 
+    `YAQL unit tests <https://github.com/openstack/yaql/tree/master/yaql/tests>`_ are also a great
+    source of reference on how to use and what features are supported in YAQL especially to cover
+    gaps in OpenStack YAQL documentation.
+
+Basics
+++++++
+The following are statements in the workflow and task definition that accepts YAQL. 
+
+* task action input
+* task concurrency
+* task on-complete
+* task on-error
+* task on-success
+* task pause-before
+* task publish
+* task retry break-on
+* task retry continue-on
+* task retry count
+* task retry delay
+* task timeout
+* task wait-before
+* task wait-after
+* task with-items
+* workflow output
+
+Each of the statement can take a string with one or more YAQL expressions. Each expression in the 
+string should be encapsulated with ``<% %>``. When evaluating a YAQL expression, Mistral also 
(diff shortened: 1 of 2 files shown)


docs/source/mistral_yaql.rst
```diff
@@ -0,0 +1,201 @@
...
+workflows. There will be many cases where you did not author the actions but there's a need to 
+decide from the result of the action whether to continue or there's a need to transform the 
+result to another value or structure for the next action in the workflow.
+
+Here are some examples of usages.
+
+* Select values for a key from a list of dictionary.
+* Filter the list where one or more fields match condition(s).
+* Transform a list to dictionary or vice versa.
+* Simple arithmetics.
```
manasdk: nit arithmetic

docs/source/mistral_yaql.rst
```diff
@@ -0,0 +1,201 @@
...
+            },
+            {
+                'name': 'vmdb2',
+                'region': 'us-west',
+                'role': 'db'
+            }
+        ]
+    }
+
+The YAQL expression ``<% $.vms.select($.name) %>`` returns only the list of VM names.
```
manasdk: If possible can you paste the full output for each of the expressions? Having the exact output would help comprehension. Not sure if that would be too tedious.

manasdk:
This PR is good - it definitely helps move the need on making YAQL accessible. 

I am unable to grasp who the target audience would be - without a doubt a stackstorm + mistral user but what would be the level of this user. I personally would look to this document to answer -

* How do I use an expression to publish variables?
* How do I pass values?
* How do I access from the environment and the tasks?
* What is the context at various stages (outside YAQL scope but then again as a user I sort of don't really care)?

Likely you do not agree with the intent of the document and perceive its usage differently.

docs/source/mistral_yaql.rst
```diff
@@ -0,0 +1,201 @@
+Mistral + YAQL
+==============
+YAQL is typically used for simple conditional evaulation and data transformation in Mistral 
+workflows. There will be many cases where you did not author the actions but there's a need to 
+decide from the result of the action whether to continue or there's a need to transform the 
+result to another value or structure for the next action in the workflow.
+
+Here are some examples of usages.
+
+* Select values for a key from a list of dictionary.
```
dzimine: I think the usage is "what for",  like "define conditions on transitions between tasks", "pass input from upward task results" and "form values for publish data or workflow output". This list is good highlight of YAQL power features. "It can do cool stuff like". 
Make somethign right here what YAQL can do and Jinga can or suck at, will be good, too.

dzimine:
Great start, few general suggestions:

Give some practical examples of YAQL used in 1) on-complete/on-error 2) with-items (including passing params into action here) and 3) env() - just give this example when we publish the url to execution. May be another most typical case. It may be not full examples but enough to stress that it can be used a hint on how.

m4dcoder:
@dzimine, there's already an on-success example included from one of our example WF.  I'll skip on the on-complete/on-error as the logic is the same. I'll include the with-items example here as well and add a short example on usage for the env() function.

dzimine:
+1 to what @manasdk comment: 
it will help to have "if you want to do X, here is how" recipes (double-value: what user CAN do, which they  may not even think, and how exactly); as for the structure you please decide, one way may be a "HOW-TO" section in the doc.

m4dcoder:
@manasdk, @dzimine: the target audiences of the doc here are StackStorm + Mistral users. It is going to be useful for almost any Mistral users. I'm not trying to duplicate what the official Mistral doc is doing on how to pass values, publish variables, access context, etc. Users can find updated docs at the Mistral doc site. But since YAQL docs is missing from both YAQL and Mistral official, what I'm trying to do is point out where YAQL can be used in Mistral open up wonderful possibilities when used with Mistral.  It's not meant to be exhaustive either.

m4dcoder:
@dzimine: I'm trying to find the balance between teaching folks how to fish (giving them the right level of info) and then try to document every single recipes what user CAN do...

Kami:
Nice, definitely a good start :+1:

PR #2253, opened 2015-11-30
RFR: Write rule enforcement objects to db

lakshmi-kannan:
## What? 

This PR will allow users to see rule enforcements. You can filter enforcements by rule (ref or id) and also filter them by timestamps. 

## Why?

To improve the visibility of rules matching/not matching, we now decided to expose the "positive" enforcements of rules. Whenever a trigger instance results in an execution, we write an enforcement object. This will allow users to see when was the last time the rule was successfully invoked. 

## API examples

```
(virtualenv)/m/s/s/st2 git:STORM-1713/rule_enforcement ❯❯❯ http GET http://localhost:9101/v1/ruleenforcements                                                                                                                                          ✭ ✱ ◼
HTTP/1.1 200 OK
Access-Control-Allow-Origin: http://127.0.0.1:9101/
Connection: keep-alive
Content-Length: 990
Date: Mon, 07 Dec 2015 23:30:35 GMT
X-Request-Id: 617564b5-067e-4ffd-a67a-280a7ffcbec7
X-Total-Count: 3

[
    {
        "enforced_at": "2015-12-07T23:30:32.537389Z",
        "execution_id": "5666169832ed3513669fa429",
        "id": "5666169832ed3513669fa42a",
        "rule": {
            "id": "5666167a32ed35137b77d06c",
            "ref": "git.st2.webhook.github.pulls.merge.sample",
            "uid": "rule:git:st2.webhook.github.pulls.merge.sample"
        },
        "trigger_instance_id": "5666169832ed3513669fa426"
    },
    {
        "enforced_at": "2015-12-07T23:30:31.130438Z",
        "execution_id": "5666169732ed3513669fa421",
        "id": "5666169732ed3513669fa422",
        "rule": {
            "id": "5666167a32ed35137b77d06c",
            "ref": "git.st2.webhook.github.pulls.merge.sample",
            "uid": "rule:git:st2.webhook.github.pulls.merge.sample"
        },
        "trigger_instance_id": "5666169732ed3513669fa41c"
    },
    {
        "enforced_at": "2015-12-07T23:30:28.608499Z",
        "execution_id": "5666169432ed3513669fa419",
        "id": "5666169532ed3513669fa41a",
        "rule": {
            "id": "5666167a32ed35137b77d06c",
            "ref": "git.st2.webhook.github.pulls.merge.sample",
            "uid": "rule:git:st2.webhook.github.pulls.merge.sample"
        },
        "trigger_instance_id": "5666169432ed3513669fa416"
    }
]

(virtualenv)/m/s/s/st2 git:STORM-1713/rule_enforcement ❯❯❯ http GET http://localhost:9101/v1/ruleenforcements/5666169832ed3513669fa42a                                                                                                                 ✭ ✱ ◼
HTTP/1.1 200 OK
Access-Control-Allow-Origin: http://127.0.0.1:9101/
Connection: keep-alive
Content-Length: 328
Date: Mon, 07 Dec 2015 23:32:02 GMT
X-Request-Id: d8121800-3735-4afc-b417-04477cbed915

{
    "enforced_at": "2015-12-07T23:30:32.537389Z",
    "execution_id": "5666169832ed3513669fa429",
    "id": "5666169832ed3513669fa42a",
    "rule": {
        "id": "5666167a32ed35137b77d06c",
        "ref": "git.st2.webhook.github.pulls.merge.sample",
        "uid": "rule:git:st2.webhook.github.pulls.merge.sample"
    },
    "trigger_instance_id": "5666169832ed3513669fa426"
}

(virtualenv)/m/s/s/st2 git:STORM-1713/rule_enforcement ❯❯❯
```

More API filtering examples: https://gist.github.com/lakshmi-kannan/a0dde0312bfda3f999e3

## CLI examples

https://gist.github.com/lakshmi-kannan/3752e2823565b86d8131

## TODO: 

* - [x] Views controller to filter by rule (ref or id), filter by timestamps
* - [X] Tests
* - [X] CLI exposure
* - [X] RBAC
* - [X] Tests for RuleEnforcement RBAC resolver 
* - [X] Docs
* - [X] Changelog


lakshmi-kannan:
@Kami @manasdk: Open questions for now:

1. Should we also expose /rules/ref_or_id/enforcements or just have rule_enforcements as a separate URL?
2. Should I call it rule_enforcements or just enforcements in URL? 
3. Are you happy with what the payload for each enforcement shows? (We decided to add execution status and other details as part of views controller. We won't save the execution object or trigger instance object in entirety in the rule enforcement object for performance reasons.)

dzimine:
Should the execution id available here, or it also goes into the views controller?

On "error during invoking the action" use case:  
- are rule enforcement failed to create execution go into view? or does it makes sense to separate successful from erroneous right in the model?
- where to keep the error info when the action execution not created?

Kami:
A couple of things.

1. Something like this definitely comes handy. Since in a sense this is similar to action executions (this is a rule execution :P) and used for audit purposes, I believe we should store the whole rule object on the Enforcement DB class and not just a reference (similar as we do with rule and action attributes on `ActionExecutionDB`). If we don't do that, things will get confusing since rules can (and do get) updated and deleted (they are not immutable).

On the other hand, storing just a reference to a trigger instance is probably fine since that object is more or less immutable right now.

2. This collection could potentially also contain a lot of data so it's something to think about. If nothing else, we should also hook it up to the TTL stuff I'm working on (that's more a note for myself).

Kami:
> Should we also expose /rules/ref_or_id/enforcements or just have rule_enforcements as a separate URL?

I would prefer to use a sub-path, but for consistency with out current stuff we should probably just add a new endpoint (/rule_enforcements).

> Should I call it rule_enforcements or just enforcements in URL?

Former, if we are going to add a new endpoint.

Kami:
Just FYI - will also need to update tests in `test_rbac_for_supported_endpoints.py` and add new ones for rule enforcements endpoint.

On a related note - we also need to decide how to handle RBAC and implement permissions resolving RBAC for rule enforcements. An option is to inherit permissions for rules (e.g. view / all on a rule grants user view on all the enforcements which belong to that rule - that's similar as we do with actions and action executions).

Kami:
P.S. I also dared to update PR description and add "RBAC" task to TODO (implementing resolving, tests, etc. is not that of a small task and without that, the PR is not ready) :P

st2reactor/st2reactor/rules/enforcer.py
```diff
@@ -16,21 +16,26 @@
...
+from st2common.services import action as action_service
+from st2common.services import trace as trace_service
+from st2common.util import reference
+from st2common.util import action_db as action_db_util
+from st2reactor.rules.datatransform import get_transformer
 
 
 LOG = logging.getLogger('st2reactor.ruleenforcement.enforce')
 
+EXEC_KICKED_OFF_STATES = [action_constants.LIVEACTION_STATUS_SCHEDULED,
```
Kami: Maybe we should move this to `constants/action/execution.py`?

st2reactor/st2reactor/rules/enforcer.py
```diff
@@ -69,21 +74,31 @@ def enforce(self):
...
-        if not liveaction_db:
-            extra = {'trigger_instance_db': self.trigger_instance, 'rule_db': self.rule}
-            LOG.audit('Rule enforcement failed. Liveaction for Action %s failed. '
+        extra = {'trigger_instance_db': self.trigger_instance, 'rule_db': self.rule}
+        enforcement_db = RuleEnforcementDB(trigger_instance_id=str(self.trigger_instance.id),
+                                           rule_ref=self.rule.ref,
+                                           rule_id=str(self.rule.id))
+        try:
+            execution_db = RuleEnforcer._invoke_action(self.rule.action, data, context)
+            execution_db.execution_id = str(execution_db.id)
```
Kami: Just curious, why do we need to do this?

lakshmi-kannan: It should be enforcement_db.execution_id.

st2reactor/st2reactor/rules/enforcer.py
lakshmi-kannan: Maybe. I don't want to expose this as the golden standard for EXEC_KICKED_OFF. In this context, this makes sense.

lakshmi-kannan:
Some good points from people overall. 

@dzimine:

>Should the execution id available here, or it also goes into the views controller?
Yes, the model contains execution_id. Turns out there is a bug in the code which I'll fix. 

>are rule enforcement failed to create execution go into view? or does it makes sense to separate successful from erroneous right in the model?
An exception will be logged today. The execution_id will be None. We could add an error field to the enforcement model but it won't make sense from a model perspective. This won't be a common case when we add rule validation i.e. rule is syntactically right. We might refer to an action with changed parameters in the rule in which case, logging is the way to expose it now. I'll think about this a little more. 

@kami:

 > I believe we should store the whole rule object on the Enforcement DB class and not just a reference (similar as we do with rule and action attributes on ActionExecutionDB). If we don't do that, things will get confusing since rules can (and do get) updated and deleted (they are not immutable). 
That's a good point. I'll save the rule object in enforcement. 

>Just FYI - will also need to update tests in test_rbac_for_supported_endpoints.py and add new ones for rule enforcements endpoint.
More work :P. But yes, I do agree about the RBAC stuff.

manasdk:
> Should we also expose /rules/ref_or_id/enforcements or just have rule_enforcements as a separate URL?
Should I call it rule_enforcements or just enforcements in URL?

`/ruleenforcements` like `/actionexecutions` for consistency. 


> Are you happy with what the payload for each enforcement shows? (We decided to add execution status and other details as part of views controller. We won't save the execution object or trigger instance object in entirety in the rule enforcement object for performance reasons.)

Yes. What we have seems reasonable. To distinguish between a failed enforcement i.e. where a match happened and yet for whatever reason failed to create an execution we will likely require a `status` and/or `message` property.

st2common/st2common/models/api/rule_enforcement.py
```diff
@@ -0,0 +1,80 @@
...
+                'description': 'Reference to the rule object.',
+                'type': 'string',
+                'required': True
+            },
+            'rule_id': {
+                'description': 'ID of the corresponding rule.',
+                'type': 'string',
+                'required': True
+            },
+            'timestamp': {
```
manasdk: `created_at` - this sorta follows existing conventions.

st2reactor/st2reactor/rules/enforcer.py
```diff
@@ -103,6 +118,13 @@ def _update_trace(self):
                                                               rules=[str(self.rule.id)])
```
manasdk: Note to self : Should save rule_enforcement_id instead of rule.id. I will take care of this separately

enykeev:
I've probably missed a part of offline discussion so please explain to me why we can't distinguish the last time rule was enforced using execution records?

Kami:
> I've probably missed a part of offline discussion so please explain to me why we can't distinguish the last time rule was enforced using execution records?

I think I also missed that discussion :P

As far as your question goes, I believe you can already kinda do that with the trace functionality (@manas please correct me if I'm wrong), but I don't think we store necessary information on the ActionExecutionDB object.

In any case, this brings up a good question - if information is already available via trace, why do we need new endpoint and everything else which goes with that? For one, I believe trace doesn't include point in time snapshot of object in question, but it's still a valid question and we need a consistent story for it.

enykeev:
Not even talking about trace. The information from the first message is easily accessible through execution object: https://gist.github.com/enykeev/a271f920c8bd5e2b609d

lakshmi-kannan:
Yes, we discussed that this info is already available in execution records. But there can be cases when rule fired and execution could not be kicked off. We are trying to address that case as well. Also, we want to be able to track this from `rules` perspective and not `executions` perspective. When everything is working perfectly, you can get the same info from executions object. When things go wrong, rule enforcements give you an extra layer of debug info. 

Trace is useful only when you want to know what happened to a trigger instance in the system. From a operator perspective, when he/she expects a rule to fire and wants to know when they were fired, there is no direct info in the system (other than executions).

lakshmi-kannan:
BTW @Kami, since we add the execution id and the execution contains the rule, it might be useless to persist the entire rule again in this object. So just persisting the ref and id should suffice IMO.

manasdk:
Trace + rule enforcements -

Trace answer the question of what are all related `trigger-instances`, `executions` and `rule-enforcements`. Rule enforcement would be a component of trace. Right now trace does not contain the enforcement object but the rule object. This is actually wrong since this is not accurate information. So infact `trace` would leverage RuleEnforcement. 

I suspect we will see that there will be more rule specific information that would not be suitable to store in an ActionExecution and we will see this object evolve into a richer object that perhaps provides more information like debugging info etc.

lakshmi-kannan:
>Also need some RBAC docs which explain how permissions work for rule enforcement objects...

yes, will add them.

lakshmi-kannan:
I think it would be better to use ``enforced_at`` instead of ``created_at`` for the timestamp. I'll make this change.

st2common/tests/unit/test_rbac_resolvers_rule_enforcement.py
```diff
@@ -0,0 +1,427 @@
...
+        user_db = self.users['custom_role_rule_all_grant']
+        resource_db = self.resources['rule_enforcement_1']
+        self.assertTrue(self._user_has_resource_db_permissions(
+            resolver=resolver,
+            user_db=user_db,
+            resource_db=resource_db,
+            permission_types=all_permission_types))
+
+        # Custom role - "rule_modify" grant on rule_1
+        # XXX: This test is failing because I think there is a bug in
```
lakshmi-kannan: @Kami @manas: So I think there is an issue that when an user has modify permissions alone on a rule, enforcement view won’t work because https://github.com/StackStorm/st2/blob/master/st2common/st2common/services/rbac.py#L169

https://github.com/StackStorm/st2/pull/2253/files#diff-972afd8e65629a580a1b3b13f3645d2dR538

So when we create a modify permission for a user, do we write a view grant as well? When I dumped the permission grant models from database for this case, I see a modify but not a view grant. I could change the permission_types in the array to include CREATE and MODIFY if the requested type is VIEW but that doesn't seem right. Maybe we do write the VIEW grant in the code but fixtures don't capture it?

Kami: It's worth noting that those are "implicit" grants meaning we don't write anything to database or file system (implicit grant resolving happens inside the resolver).

See https://github.com/StackStorm/st2/blob/master/st2common/st2common/rbac/resolvers.py#L216 for an example

st2common/st2common/models/api/rule_enforcement.py
```diff
@@ -0,0 +1,93 @@
...
+                'description': 'ID of the corresponding rule.',
+                'type': 'string',
+                'required': True
+            },
+            'rule_uid': {
+                'description': 'UID of the corresponding rule.',
+                'type': 'string',
+                'required': True
+            },
+            'rule_pack': {
```
manasdk: Why is the `rule_pack` needed?

st2common/st2common/models/api/rule_enforcement.py
```diff
@@ -0,0 +1,93 @@
...
+                               'that flipped the rule.',
+                'type': 'string',
+                'required': True
+            },
+            'execution_id': {
+                'description': 'ID of the action execution that was invoked as a response.',
+                'type': 'string',
+                'required': True
+            },
+            'rule_ref': {
```
manasdk: Would it be better to have a `Rule` property with `ref`, `uid` and `id` are child properties?

st2common/st2common/models/api/rule_enforcement.py
lakshmi-kannan: Because, rbac. User could have pack level permissions and we need to know the pack, the rule in enforcement belongs to. I could have done get_by_ref on rule in rbac resolver and figured out the pack name but this turned out cheaper.

st2common/st2common/models/db/rule_enforcement.py
```diff
@@ -0,0 +1,46 @@
...
+
+import mongoengine as me
+
+from st2common.fields import ComplexDateTimeField
+from st2common.models.db import MongoDBAccess
+from st2common.models.db import stormbase
+from st2common.util import date as date_utils
+
+
+class RuleEnforcementDB(stormbase.StormFoundationDB, stormbase.TagsMixin):
```
manasdk: Don't see indexes. Likely, you want to add indexes for `rule_ref`

st2common/st2common/models/api/rule_enforcement.py
manasdk: Can we hide from API?

st2common/st2common/models/api/rule_enforcement.py
lakshmi-kannan: I definitely thought about this but I am iffy. If we are very particular, I can make this change. Should we do a similar change for trigger_instance and execution too? Those are flat.

st2common/st2common/models/api/rule_enforcement.py
lakshmi-kannan: There is no easy way to do this AFAIK. I do want to hide it. It's not exposed in CLI though.

st2common/st2common/models/db/rule_enforcement.py
lakshmi-kannan: Good point.

st2common/st2common/models/api/rule_enforcement.py
lakshmi-kannan: Your idea of just removing the entry from the model seems to have worked. See e8b3d26.

st2common/st2common/models/db/rule_enforcement.py
lakshmi-kannan: Fixed in e8b3d26

st2common/st2common/models/api/rule_enforcement.py
lakshmi-kannan: Also, in 6438941 I got rid of the `pack` in models altogether. It's easy to get_pack from `ref`.

st2common/st2common/models/api/rule_enforcement.py
lakshmi-kannan: Done. Fixed in 6438941. Updated API examples.

PR #2256, opened 2015-12-01
Pack testing mock classes and related code

Kami:
This pull request introduces some mock classes and other related code to help with pack testing.

In addition to that, it also establishes and documents some pack testing conventions.


docs/source/development/pack_testing.rst
```diff
@@ -0,0 +1,98 @@
...
+sensor and action modules directly in your code.
+
+For example, if you have an action named ``actions/parse_xml.py`` you can do the
+following inside your test module:
+
+```python
+import parse_xml
+```
+
+Keep in mind that both sensor and action modules are not namespaced which means
```
Kami: I'm not too big of a fan of this, but it's the most pragmatic solution.

Ideally you could do `from sensors import sensor_name` and `from actions import action_name`, but that's not possible since pack root directory is not a package (it doesn't contain `__init__.py`).

Even though it's not a package it would still be possible to make it work with an import hook hack. I'm against the import hook approach though since it's magic and running tests would only work if the user uses `run-pack-tests.sh` script - this would cause all kinds of confusion.

Kami: Another option of course is adding `__init__.py` to the pack root dir, but I'm not a fan of that option either.

Kami:
Alright, this is ready for a review - feedback welcome.

docs/source/development/pack_testing.rst
```diff
@@ -0,0 +1,98 @@
...
+
+    import parse_xml
+
+Keep in mind that both sensor and action modules are not namespaced which means
+sensor and action module names need to be unique to avoid conflicts.
+
+Running Tests
+-------------
+
+To run all the tests in a particular pack you can use the ``run-pack-tests.sh``
```
jfryman: What's your thought on having this file in the `st2contrib` repo versus baking it into the actual `st2` repo?

jfryman:
Nice. Look forward to seeing a pack with tests in it as an example as well. :dancers:

docs/source/development/pack_testing.rst
```diff
@@ -0,0 +1,98 @@
...
+
+For example, if you have an action named ``actions/parse_xml.py`` you can do the
+following inside your test module:
+
+.. sourcecode:: python
+
+    import parse_xml
+
+Keep in mind that both sensor and action modules are not namespaced which means
+sensor and action module names need to be unique to avoid conflicts.
```
tim-ireland: Wondering whether adding the pack root directory `<packname>/` to PYTHONPATH instead would cause a problem?  Then you could avoid collisions, `from action import parse_xml`, and allow for more code re-use in pack development.  Perhaps at the cost of assuring a consistent pack structure though?

docs/source/development/pack_testing.rst
```diff
@@ -0,0 +1,98 @@
...
+Test File Locations and Names
+-----------------------------
+
+All the test files should go into ``<pack name>/tests/`` directory. If tests
+include any fixtures, they should be put in the ``<pack name>/tests/fixtures/``
+directory.
+
+Test files should follow the following naming conventions:
+
+* ``test_action_<action name>.py`` for action tests. For example, if the action
```
manasdk: Is this a recommendation or a requirement?

docs/source/development/pack_testing.rst
```diff
@@ -0,0 +1,98 @@
...
+------------
+
+In addition to all the |st2| and pack dependencies listed in
+``requirements.txt`` and ``requirements-tests.txt``, the following libraries are
+also available by default inside the tests:
+
+* ``unittest2``
+* ``mock``
+
+In addition those dependencies, sensors (``<pack name>/sensors/``) and actions
```
manasdk: This implies using the script [run-pack-tests.sh](https://github.com/StackStorm/st2contrib/blob/pack_testing/scripts/run-pack-tests.sh)?

docs/source/development/pack_testing.rst
manasdk: Although not entirely a blocker right now but it is a very awkward limitation. If nothing else lets try and work out some options.

docs/source/development/pack_testing.rst
```diff
@@ -0,0 +1,98 @@
...
+(``<pack name>/actions/``) directory is added to PYTHONPATH meaning you can import
+sensor and action modules directly in your code.
+
+For example, if you have an action named ``actions/parse_xml.py`` you can do the
```
manasdk: what happens if the path is `actions/code/parse_xml.py` similar to [packs_mgmt](https://github.com/StackStorm/st2/tree/master/contrib/packs/actions/pack_mgmt)?

Would we be able to write

`from pack_mgmt import info` or `import pack_mgmt.info`?

docs/source/development/pack_testing.rst
manasdk: I vote for moving to `st2`. Does make running tests in `st2contrib` awkward.

st2tests/st2tests/base.py
```diff
@@ -402,6 +403,81 @@ def assertProcessExited(self, proc):
...
+    """
+    Base class for sensor tests.
+
+    This class provides some utility methods for verifying that a trigger has
+    been dispatched, etc.
+    """
+
+    sensor_cls = None
+
+    def setUp(self):
```
manasdk: setUp and not SetUpClass?

st2tests/st2tests/base.py
```diff
@@ -402,6 +403,81 @@ def assertProcessExited(self, proc):
...
+            else:
+                payload_matches = True
+
+            if trace_context:
+                trace_context_matches = (item['trace_context'] == trace_context)
+            else:
+                trace_context_matches = True
+
+            if trigger_matches and payload_matches and trace_context_matches:
+                return True
```
manasdk: Would `self.assertTrue(True)` make more sense? This way the method asserts correctly.

st2tests/st2tests/base.py
```diff
@@ -402,6 +403,81 @@ def assertProcessExited(self, proc):
...
+
+        msg = 'Trigger "%s" hasn\'t been dispatched' % (trigger)
+        raise AssertionError(msg)
```
manasdk: how about `self.assertTrue(False, msg)`

docs/source/development/pack_testing.rst
```diff
@@ -0,0 +1,98 @@
...
+This section includes information on pack testing - where to put the tests,
+how to write the tests, mock classes which can be used to make testing
+easier, etc.
+
+Test File Locations and Names
+-----------------------------
+
+All the test files should go into ``<pack name>/tests/`` directory. If tests
+include any fixtures, they should be put in the ``<pack name>/tests/fixtures/``
+directory.
```
tim-ireland: Would it make sense to also add a `<pack name>/mocks/` directory as a convention?  These could be used to mock sessions to remote services for example.

docs/source/development/pack_testing.rst
Kami: I was thinking about that, it could live in either repo.

If we move it to st2, where would it live? tools? scripts?

docs/source/development/pack_testing.rst
Kami: Yeah, that would work.

Here is some context on that - https://github.com/StackStorm/st2/pull/2256#discussion_r46377024

docs/source/development/pack_testing.rst
lakshmi-kannan: I prefer this to be in st2contrb. Don't we want st2contrib travis to run it?

docs/source/development/pack_testing.rst
Kami: The resource part in the name is a recommendation, but the `test_*.py` part is a requirement from a test runner.

This requirement can be relaxed of course, but I think it's a good one since all the Python tests tend to follow it :)

docs/source/development/pack_testing.rst
Kami: Correct. 

I can clarify that in the docs. The idea was to use no import and other hacks so tests can still be ran manually if all the requirements are set up correctly.

So if a user manually creates a virtual environment with all the necessary dependencies, tests will still work.

Having said that, maybe I should refactor it in two script - `setup-pack-tests-environment.sh` and `run-pack-tests.sh`.

Wdyt?

docs/source/development/pack_testing.rst
Kami: Yeah, this works since pack_mgmt is a package and contains `__init__.py`.

st2tests/st2tests/base.py
Kami: Yes - it's intentional. This way it's called before every test method so all the internal state on the sensor service is cleared between each tests.

st2tests/st2tests/base.py
Kami: Good point :)

docs/source/development/pack_testing.rst
Kami: We could, but I personally prefer to put all the test related stuff under `tests/` directory (so `tests/fixtures/` and / or `tests/mocks/`).

Kami:
@jfryman Some example tests can be found here:

* https://github.com/StackStorm/st2contrib/pull/327/files#diff-71ff97ec034072a4d97521756d767c7dR1
* https://github.com/StackStorm/st2contrib/pull/327/files#diff-1a1fd9b153258a2789845e92bd24fb28R1

docs/source/development/pack_testing.rst
jfryman: @lakshmi-kannan give me a valid argument as to why. I don't accept preference as an argument. 

From my point of view, we need this close to the user experience. As it is right now, we currently have our tools splayed all over the place.

For instance: can you tell me how and where a pack generation tool is? Probably... but have you used it? It's not part of our story, It's detached from core, and so it just floats out there.

This should be right in front of the users as part of a default install, not something they need to seek out. If we want to encourage good behaviors, the less friction that exists is better.

lakshmi-kannan: Well, I think we are both saying the same thing. I want this to be in st2contrib which is where the pack developement takes place. I am a little lost.

st2tests/st2tests/base.py
Kami: That's actually a bit weird:

```python======================================================================
FAIL: test_poll (test_sensor_docker_sensor.DockerSensorTestCase)
  File "/w/stackstorm/st2contrib/packs/docker/tests/test_sensor_docker_sensor.py", line 41, in test_poll
    payload={'container_info': MOCK_CONTAINER_DATA})
  File "/w/stackstorm/st2/st2tests/st2tests/base.py", line 477, in assertTriggerDispatched
    self.assertTrue(False, 'Trigger "%s" hasn\'t been dispatched' % (trigger))
AssertionError: False is not true : Trigger "docker.container_tracker.startedx" hasn't been dispatched
-------------------- >> begin captured logging << --------------------
````

```python
FAIL: test_poll (test_sensor_docker_sensor.DockerSensorTestCase)
  File "/w/stackstorm/st2contrib/packs/docker/tests/test_sensor_docker_sensor.py", line 41, in test_poll
    payload={'container_info': MOCK_CONTAINER_DATA})
  File "/w/stackstorm/st2/st2tests/st2tests/base.py", line 478, in assertTriggerDispatched
    raise AssertionError(msg)
AssertionError: Trigger "docker.container_tracker.startedx" hasn't been dispatched
-------------------- >> begin captured logging << --------------------
```

Imo, assertion error looks better and more appropriate.

docs/source/development/pack_testing.rst
jfryman: Also, let me also throw out there. I'm not bullish on the `st2` repo in general, but it's the only story we can assert makes it to our users 100% of the time. If it's not `st2`, where is it? I admit, this statement _is_ a bit of a :bike: :house:, and probably for a different/bigger conversation, but the intent I hope we :dart: for is encouraging effective behaviors for all our users.

docs/source/development/pack_testing.rst
Kami: To clarify it further - I'm against `__init__.py` in the pack root since it kinda implies that a pack is (just) a Python package, but that's not necessary true since pack doesn't even need to contain any Python files / modules (actions can be written in an arbitrary programming language).

st2common/tests/unit/test_unit_testing_mocks.py
```diff
@@ -0,0 +1,87 @@
...
+    'BaseSensorTestCaseTestCase',
+    'MockSensorServiceTestCase'
+]
+
+
+class MockSensorClass(object):
+    pass
+
+
+class BaseSensorTestCaseTestCase(BaseSensorTestCase):
```
Kami: We could also move pack related test case classes to a different module (e.g. `st2tests.packs.BaseSensorTestCase`).

What do others think?

docs/source/development/pack_testing.rst
Kami: I do agree that most of our pack development happens inside st2contrib, but users could be developing private packs which doesn't necessary happen inside st2contrib.

Imo, ideal scenario would be for this script to live as a standalone "tool" (e.g. st2-run-pack-tests) inside st2-sdk (or similar), but I agree that we don't have a good and consistent story for delivering SDK to the users so st2 is probably the best bet for now.

For the time being, I will move it to this repo (we will still run the tests on every commit to st2contrib repo).

Kami: Alright, I moved the script in 642537116e8bd41a99d9e0e0fa9b11e6cae84d70.

Keep in mind that the move and script living inside this repo implies different context so I also need to made a bunch of changes to the script:

* Script is now installed into `/usr/local/bin/st2-run-pack-tests` by default (I can't remove that if people don't agree with it)
* Script creates a temporary virtual environment for the pack tests by default (it installs all the necessary dependencies there). That's to avoid the pollution of the global namespace if the user runs script directly on a system where StackStorm is installed. Inside st2contrib we of course skip that (we use "-x" option) to speed things up - we simply use shared virtualenv where we install all the dependencies.

Kami: Another thing worth noting is that now we only install StackStorm dependencies listed in `st2/requirements.txt` and `st2/test-requirements.txt` if `ST2_REPO_PATH` environment variable is set (e.g. when we run the tests on Travis inside `st2contrib`).

By default, when using development VM (st2express / st2workroom) a virtual environment should already be created and all the st2 dependencies installed there so we don't need to install them again.

Here is an example when I run tests inside dev VM using ``-x`` option:

```bash
trib/packs/docker/
Running tests for pack: docker
test_poll (test_sensor_docker_sensor.DockerSensorTestCase) ... ok

Ran 1 test in 0.952s

OK
```

Kami:
Now that we have some infra in place, who's going to write some tests for packs in `contrib/*`?

@lakshmi-kannan? :trollface:

Kami:
Also hooked it up to Travis so now all the tests for packs in `contrib/*

Kami:
Also hooked it up to Travis so now all the tests for packs in `contrib/*` run on Travis - https://travis-ci.org/StackStorm/st2/jobs/94581382

Now we just need some tests :P

docs/source/development/pack_testing.rst
manasdk: Yeah I agree that pack root dir should not have `__init__.py`. I guess this is our only option.

docs/source/development/pack_testing.rst
manasdk: `test_*.py` I am on board with.

Since the other naming convention is not a requirement this works for me. I guess we could tweak the language in the docs from "should follow the following ..." to "are recommended to follow ...". Since that will make the `test_*.py` part awkward to explain I do not insist.

docs/source/development/pack_testing.rst
manasdk: Yeah I was thinking that would make a great deal of sense. Perhaps a Makefile that does the right things placed in a test pack would also work.

In the short-term splitting into 2 scripts makes sense

st2tests/st2tests/base.py
manasdk: Would be nice if that were a choice but this works just as well.

st2tests/st2tests/base.py
manasdk: Yup you are right.

docs/source/development/pack_testing.rst
manasdk: I would also prefer if `mocks` and `fixtures` would be under tests.

st2common/tests/unit/test_unit_testing_mocks.py
manasdk: I personally prefer current approach.

manasdk:
+1

PR #2257, opened 2015-12-01
Append_url parameter in ack messages

emedvedev:
Discussed this with @jfryman yesterday: StackStorm execution details URL will be appended to the end of the ack message by default unless specifically disabled with `append_url: false`.

### DIFF
--- a/st2common/st2common/models/api/action.py
+++ b/st2common/st2common/models/api/action.py
@@ -516,7 +516,8 @@ class ActionAliasAPI(BaseAPI, APIUIDMixin):
                 "type": "object",
                 "properties": {
                     "enabled": {"type": "boolean"},
-                    "format": {"type": "string"}
+                    "format": {"type": "string"},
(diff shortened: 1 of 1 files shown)


st2common/st2common/models/api/action.py
```diff
@@ -516,7 +516,8 @@ class ActionAliasAPI(BaseAPI, APIUIDMixin):
                 "type": "object",
                 "properties": {
                     "enabled": {"type": "boolean"},
-                    "format": {"type": "string"}
+                    "format": {"type": "string"},
+                    "append_url": {"type": "boolean"}
```
manasdk: default value is `true`?

emedvedev: Yes, same as enabled. We discussed it with @jfryman: URL is appended by default to everything (even a custom format string) unless explicitly set to "false".

manasdk: K. Then call out default value in the schema

PR #2261, opened 2015-12-03
CircleCI Chat Notification only for major branches

armab:
See: https://circleci.com/docs/configuration#per-branch-notifications

However this can fully work only when we'll propagate `circle.yml` configs to entire `st2` repository/all branches.


lakshmi-kannan:
+1

PR #2263, opened 2015-12-03
No-Op runner

jfryman:
This PR introduces a runner that does *absolutely nothing*. Yup!

I need this as a stub and placeholder when all I need to do is send a notification, but it's inappropriate to do in the context of the task I am currently in.

:eyes: and other critiques are most certainly appreciated.

### DIFF
--- a/st2actions/st2actions/runners/nooprunner.py
+++ b/st2actions/st2actions/runners/nooprunner.py
@@ -40,8 +40,6 @@ class NoopRunner(ActionRunner):
             'failed': False,
             'succeeded': True,
             'return_code': 0,
-            'stdout': "",
-            'stderr': ""
         }
 
         status = LIVEACTION_STATUS_SUCCEEDED
--- a/st2actions/st2actions/runners/nooprunner.py
+++ b/st2actions/st2actions/runners/nooprunner.py
@@ -33,8 +33,8 @@ class NoopRunner(ActionRunner):
     def run(self, action_parameters):
 
         LOG.info('Executing action via NoopRunner: %s', self.runner_id)
-        LOG.info('[Action info] name: %s, Id: %s' %
-                 (self.action_name, str(self.liveaction_id)))
+        LOG.info('[Action info] name: %s, Id: %s' ,
+                 self.action_name, str(self.execution_id))
 
         result = {
             'failed': False,
--- a/st2actions/st2actions/runners/nooprunner.py
+++ b/st2actions/st2actions/runners/nooprunner.py
@@ -30,6 +30,9 @@ class NoopRunner(ActionRunner):
     def __init__(self, runner_id):
         super(NoopRunner, self).__init__(runner_id=runner_id)
 
+    def pre_run(self):
+        pass
+
     def run(self, action_parameters):
 
         LOG.info('Executing action via NoopRunner: %s', self.runner_id)
--- a/st2actions/st2actions/runners/nooprunner.py
+++ b/st2actions/st2actions/runners/nooprunner.py
@@ -1,0 +1,48 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from st2common import log as logging
+from st2actions.runners import ActionRunner
+from st2common.constants.action import LIVEACTION_STATUS_SUCCEEDED
+import st2common.util.jsonify as jsonify
+
+LOG = logging.getLogger(__name__)
+
+
+class NoopRunner(ActionRunner):
+    """
+    Runner which does absolutely nothing. No-op action.
+    """
+    KEYS_TO_TRANSFORM = ['stdout', 'stderr']
+
+    def __init__(self, runner_id):
+        super(NoopRunner, self).__init__(runner_id=runner_id)
+
+    def run(self, action_parameters):
+
+        LOG.info('Executing action via NoopRunner: %s', self.runner_id)
+        LOG.info('[Action info] name: %s, Id: %s' %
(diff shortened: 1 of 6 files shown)


manasdk:
> I need this as a stub and placeholder when all I need to do is send a notification, but it's inappropriate to do in the context of the task I am currently in.

Can you provide an example?

jfryman:
@manasdk sure.

```
---
chain:
  -
    name: "search_kb"
    ref: "reamaze.article_search"
    params:
      query: "{{ query }}"
    publish:
      kb_results: "{{ search_kb.result }}"
    on-success: "chatops_parse"
    on-failure: "fail_search_kb"
  -
    name: "chatops_parse"
    ref: "st2-community.parse"
    params:
      input: "{{ kb_results }"
    publish:
      seach_chatops_output: "{{ result }}"
  -
    name: "send_chatops_to_user"
    ref: "core.noop"
    params: {}
    notify:
      on-success:
        message: "Found some articles for you! Check them out! {{ kb_results }}"   
  -
    name: "fail_search_kb"
    ref: "core.noop"
    params: {}
    notify:
      on-success:
        message: "Unable to find KB article for {{query}}"
```

In action chain, it's less prevalent to see, but I see a notification coupled to the action that is performing something to be reserved for system notifications, but having a `noop` lets me keep human responses in a separate leaf in my workflow.

manasdk:
There is perhaps some nuance here that I am missing why the notification should be separate. 

In any case a `noop` action seems reasonable - it is unfortunate that it requires a special runner as well but it is what it is I suppose.

st2actions/st2actions/runners/nooprunner.py
```diff
@@ -0,0 +1,48 @@
...
+    KEYS_TO_TRANSFORM = ['stdout', 'stderr']
+
+    def __init__(self, runner_id):
+        super(NoopRunner, self).__init__(runner_id=runner_id)
+
+    def run(self, action_parameters):
+
+        LOG.info('Executing action via NoopRunner: %s', self.runner_id)
+        LOG.info('[Action info] name: %s, Id: %s' %
+                 (self.action_name, str(self.liveaction_id)))
```
manasdk: should use execution_id.

st2actions/st2actions/runners/nooprunner.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    KEYS_TO_TRANSFORM = ['stdout', 'stderr']
+
+    def __init__(self, runner_id):
+        super(NoopRunner, self).__init__(runner_id=runner_id)
+
+    def run(self, action_parameters):
+
+        LOG.info('Executing action via NoopRunner: %s', self.runner_id)
+        LOG.info('[Action info] name: %s, Id: %s' %
```
manasdk: `LOG.info('[Action info] name: %s, Id: %s', self.action_name, str(self.liveaction_id))`

is better as the evaluation of the string gets delayed.

st2actions/st2actions/runners/nooprunner.py
```diff
@@ -0,0 +1,48 @@
...
+                 (self.action_name, str(self.liveaction_id)))
+
+        result = {
+            'failed': False,
+            'succeeded': True,
+            'return_code': 0,
+            'stdout': "",
```
manasdk: Is there much value in including `stdout` and `stderr` properties? If not I would skip them for the no-op runner

st2actions/st2actions/runners/nooprunner.py
```diff
@@ -0,0 +1,48 @@
...
+        super(NoopRunner, self).__init__(runner_id=runner_id)
+
+    def run(self, action_parameters):
+
+        LOG.info('Executing action via NoopRunner: %s', self.runner_id)
+        LOG.info('[Action info] name: %s, Id: %s' %
+                 (self.action_name, str(self.liveaction_id)))
+
+        result = {
+            'failed': False,
```
manasdk: (unrelated)
It has always struck me as odd that we have 2 separate properties `failed` and `succeeded`.

st2common/st2common/bootstrap/runnersregistrar.py
```diff
@@ -521,6 +521,14 @@
             }
         },
         'runner_module': 'st2actions.runners.cloudslang.cloudslang_runner'
+    },
+    {
+        'name': 'noop',
+        'aliases': [],
+        'description': 'A runner that does absolutely nothing',
```
manasdk: Now that not true :)

It actually simply responds with a static response which does not change regardless of the input.

st2actions/st2actions/runners/nooprunner.py
jfryman: No, none at all. What I was more concerned about is if this is consumed somewhere else upstream... so I kept the payload consistent. I'll drop if that's safe

manasdk: Drop it. Runners are allowed to define the structure of the result and that is generally treated as opaque by StackStorm. 

Workflows and rules tend to be runner aware as they need to intrinsically understand these properties to consume the result. There should be no impact in this case so I would keep the response specific to the no-op runner.

st2actions/st2actions/runners/nooprunner.py
jfryman: Yeah, I'll express the same puzzlement. status: Failed/Success.

If any historians want to throw down some knowledge, always fun to know. :)

manasdk:
Please update the CHANGELOG.

jfryman:
:green_apple:

Gonna merge! Thanks for the :eyes:, @manasdk!

dzimine:
This is awesome! 
I was thinking about something like this to do a "dry-run". Just to beef this up with runner parameters "dry-result" and "dry-status", and we can create any test workflows. Further, we can add these parameters to a base runner, make other runners ignore it and have a dryrun key, so that when it's used, the dry-runner is substituted for all other runners, and actions return dry resutl and dry status. Again this is a rough idea, you guys can make something good from it, hopefully. 

But this is awesome as it is, for testing workflows.

Kami:
Yeah, `noop` action is something we have discussed before - there are definitely some valid uses in the workflows.

As far as the runner goes - couldn't we simple make it a `core.local` action which does `cmd=exit 0`"? I know it's not exactly the same as doing nothing, but it's close enough :P

st2actions/st2actions/runners/nooprunner.py
```diff
@@ -0,0 +1,49 @@
...
+    def run(self, action_parameters):
+
+        LOG.info('Executing action via NoopRunner: %s', self.runner_id)
+        LOG.info('[Action info] name: %s, Id: %s',
+                 self.action_name, str(self.execution_id))
+
+        result = {
+            'failed': False,
+            'succeeded': True,
+            'return_code': 0,
```
Kami: `return_code` is specific to local and remote runners so we can also omit it here. Same goes for `failed` attribute.

PR #2264, opened 2015-12-04
Add new entries to mistral status map

m4dcoder:
The new timeout status causes zombie mistral workflows because the callback handler did not know how to handle the new state.

### DIFF
--- a/st2actions/st2actions/handlers/mistral.py
+++ b/st2actions/st2actions/handlers/mistral.py
@@ -31,12 +31,17 @@ from st2common.util.workflow import mistral as utils
 LOG = logging.getLogger(__name__)
 
 
-STATUS_MAP = dict()
-STATUS_MAP[action_constants.LIVEACTION_STATUS_REQUESTED] = 'RUNNING'
-STATUS_MAP[action_constants.LIVEACTION_STATUS_SCHEDULED] = 'RUNNING'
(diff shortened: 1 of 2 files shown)


Kami:
Good catch :+1:

st2actions/st2actions/handlers/mistral.py
```diff
@@ -31,12 +31,17 @@
 LOG = logging.getLogger(__name__)
 
 
-STATUS_MAP = dict()
-STATUS_MAP[action_constants.LIVEACTION_STATUS_REQUESTED] = 'RUNNING'
-STATUS_MAP[action_constants.LIVEACTION_STATUS_SCHEDULED] = 'RUNNING'
-STATUS_MAP[action_constants.LIVEACTION_STATUS_RUNNING] = 'RUNNING'
-STATUS_MAP[action_constants.LIVEACTION_STATUS_SUCCEEDED] = 'SUCCESS'
-STATUS_MAP[action_constants.LIVEACTION_STATUS_FAILED] = 'ERROR'
+STATUS_MAP = {
```
Kami: Maybe we can add some code check here or in the tests to throw if not all of the statuses are mapped here.

This way issues like this will be detected earlier in the future.

Kami: Actually, ignore that.

I see you added some tests for that below - https://github.com/StackStorm/st2/pull/2264/files#diff-6153b0d1287b2dcbcc120c6e037bd783R601

PR #2266, opened 2015-12-07
Avoid deep render issue by passing execution_id

enykeev:


emedvedev:
We'll need to get back to this and revise this whole deep rendering problem, but +1 for now.

Kami:
Thanks.

To make it easier and more friendly for other contributors (including ourselves) it would be good idea to also add some context to the PR description in the future :)

Kami:
As far as the change itself goes - I missed the Slack discussion (I assume it happened there), but I think I know what is going on.

I would prefer to keep it an object since it's more generic and less tied to an execution, but I see an immediate need so I'm fine with this for now.

enykeev:
Yeah, sorry, we were in a bit of a rush with that one to bother with proper PR messages.

The problem we're having there is that we're attempting to render templates for a single action twice. First time we do that in actionchain\rule engine and the second one in action container. So if the first render resulting in liveaction.parameters having a new template (either by string concatenation or by results of previous action containing a template as a string inside of it), it tries and usually fails to evaluate this new template too.

So in this particular case, execution object very often contain some other template inside of it, so it then fails to evaluate because context missing a variable and the whole workflow ends up in error state.

enykeev:
example of the failing workflow https://gist.github.com/enykeev/96e58e778c5c04bde0f1

PR #2271, opened 2015-12-07
Tim/pack testing

tim-ireland:
Modifications to enable execution of st2 pack style tests in the `plexxiconnectst2/docker/base` docker container.

Depends on:
[StackStorm st2 PR 2256] (https://github.com/StackStorm/st2/pull/2256)

#### st2-run-pack-tests ####
* Moved `VIRTUALENV_DIR` settings within the test for creating a new virtual env.  This allows us to test using our existing virtual environments that is created when the docker image starts.

* Also remove the re-installation of requirements.txt.  Only pip installs should be to install test dependancies.

#### st2tests/st2tests/mocks/sensor.py ####
* Fix to handle traceback discovered during testing.

Note: The remaining changes are related to mac development - originally branched from Andrew's mac_dev branch.


Kami:
@tim-ireland Do you plan to re-open this PR?

As far as `st2-run-pack-tests` script goes. As noted in the other PR, I think the improvement would be to split it into two scripts:

* `st2-setup-pack-tests-environment` - Script which sets up the environment so tests can run in it (would mostly mean moving code from existing `st2-run-pack-tests` script into this file)
* `st2-run-pack-tests` - Script which only runs the tests.

This way the users who already have an existing environment or want to manage the environment themselves can only run `st2-run-pack-tests`.

PR #2273, opened 2015-12-08
Fix cancellation of actions in non-running state

m4dcoder:
This bug only manifests currently for the mistral runner. Since the action is in requested state, the workflow execution has not started. The container service invokes cancel for a workflow that doesn't exists and fails.


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452245475021534

PR #2274, opened 2015-12-08
Update runner params docs

m4dcoder:


Kami:
lol, I was wondering how long it takes before someone commits `announcemen.rst` file :)

I personally didn't commit it since I thought we don't want to document it - in any case, if we don't want to document it we should update the generation script and add it to the exclude list (or similar).

Kami:
Actually, now that I think more about it - I don't think this file is included anywhere in the docs so it's fine to just commit it.

PR #2275, opened 2015-12-08
RFR: Use delete_by_query to delete models in purge tool

lakshmi-kannan:


st2common/st2common/persistence/trigger.py
```diff
@@ -84,3 +84,7 @@ class TriggerInstance(Access):
     @classmethod
     def _get_impl(cls):
         return cls.impl
+
+    @classmethod
+    def delete_by_query(cls, **query):
```
Kami: Can't we just add this to the base class (`Access`)?

lakshmi-kannan: This is done so we have control over which models should support purge. It's a little confusing for me when I looked at the code.

Kami:
Do we actually install those tools somewhere or the user needs to clone the repo to use them?

I quickly checked st2common/bin/packaging and it doesn't seem so.

lakshmi-kannan:
@Kami yep, we want to test this out in st2build before making it available as tool. I stress tested in my VM some. build and ops box testing would be better.

Kami:
Cool, sounds good.

st2common/st2common/persistence/trigger.py
Kami: Yeah, assumed something like that was the case.

PR #2277, opened 2015-12-09
Fix docs mentioning 9100 and 9101 ports

enykeev:
There will be a lot of confusion next couple of month related to this new schema and how port and path corelate and when you should use one or the other.

The rule of thumb I used while making this commit and suggesting you should use when in doubt is rather simple:

> `https://example.org/api` is a public url and should be used for external services talking to st2 (webhooks, browsers, off-server or potentially off-server running services like hubot).

>`http://localhost:9101/` is a private url and should be used for testing or wiring between internal components (so it is ok to use it as a mock in tests, example when testing custom auth is properly set up or to route the request from nginx to st2api).

There's a ton of potentially tricky situations where it's not immediately clear whether it should be private or public url and we'll have to deal with each of them on a case by case basis.

And yeah, if you can avoid mentioning st2api url in examples that are not directly tied to wiring up the system, use some other url instead. http://httpbin.org/get is a nice general purpose example url.

### DIFF
--- a/contrib/examples/actions/workflows/mistral-env-var.yaml
+++ b/contrib/examples/actions/workflows/mistral-env-var.yaml
@@ -5,11 +5,11 @@ examples.mistral-env-var:
     type: direct
     output:
         env: <% env() %>
-        url: <% $.url %> 
+        url: <% $.url %>
     tasks:
         task1:
             action: core.local
             input:
-                cmd: echo https://127.0.0.1:9101/history/<% env().st2_execution_id %>
+                cmd: echo http://127.0.0.1:9101/executions/<% env().st2_execution_id %>
             publish:
                 url: <% $.task1.stdout %>
--- a/docs/source/actions.rst
+++ b/docs/source/actions.rst
@@ -57,10 +57,10 @@ To execute an action manually, you can use ``st2 run <action with parameters>`` 
 .. code-block:: bash
 
    # Execute action immediately and display the results
-   st2 run core.http url="http://localhost:9101"
+   st2 run core.http url="http://httpbin.org/get"
 
    # Schedule action execution
-   st2 action execute core.http url="http://localhost:9101"
+   st2 action execute core.http url="http://httpbin.org/get"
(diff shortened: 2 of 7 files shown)


pixelrebel:
@enykeev I'm a bit confused by this.  I have external hosts pushing to a public st2 webhook using 9101.  So moving forward, everything connects over 443?  I sort of liked how I only had to open up 9101 on the firewall.  This kept the Web GUI accessible only from inside the network.

enykeev:
@pixelrebel it was always the plan, but we didn't started serving API on https port from day one because at that point we neither had nginx as a part of the system nor had a good understanding how are we going to deploy it.

And we had to pay a high price for that decision by handling problems with CORS, self-signed certificates that you need to manually allow on every port or inability to use cookies as a storage for token.

We're now slowly moving back to the original plan, but this ports won't stop working overnight and we will help you transition seamlessly. Most likely we're going to keep them, but deny access in firewall by default or will provide an option in answers.json that will allow you to keep using them providing you know how to handle all the issues related. Or we may simply separate webhook into additional service outside st2 itself and you'll be able to run it on any port you want.

It's hard to say at that point, we would need to discuss that, but we hear you.

dzimine:
And surely you can just configure it to keep it open, and further constrain it to the way you like. The file that is responsible for the port is still there, /etc/nginx/sites-enabled/st2api.conf, so it's up to you to re-enable it on your infra if you want. 

Also, we often create a DMZ webhook. This comes handy when we work with external source of web hooks, but reluctant to expose st2 api internally ([here is one example of a one-off DMZ webhook](https://github.com/StackStorm/st2incubator/blob/master/packs/activecampaign/sensors/ac_webhook.py))

lakshmi-kannan:
This is defnitely going to be confusing and we need to hand hold people with the migration. But I am +1 to this change.

dzimine:
+1 for the change (merge?)

Suggestion: a KB article on how to enable serving the api/auth off 9100/9101.

PR #2281, opened 2015-12-10
Rework rule list columns

manasdk:
### Before
```
$ st2 rule list
+------------------------------+---------+-------------+---------+
| ref                          | pack    | description | enabled |
+------------------------------+---------+-------------+---------+
| default.passive_trigger_rule | default | jdlksdj     | True    |
+------------------------------+---------+-------------+---------+
```

### After
```
$ st2 rule list
+------------------------------+--------------------------------+------------+---------+
| ref                          | trigger.ref                    | action.ref | enabled |
+------------------------------+--------------------------------+------------+---------+
| default.passive_trigger_rule | fixtures.test_passive_trigger. | core.local | True    |
|                              | dummy                          |            |         |
+------------------------------+--------------------------------+------------+---------+
```

PR #2283, opened 2015-12-10
Revert "Rework rule list columns"

manasdk:
Reverts StackStorm/st2#2281


dzimine:
+1 Please revert. 

I insist on having description by default, as the rule by itself is generally not informative enough, AND showing descriptions rewards user behavior of adding meaninful descriptions (and vice versa).

If someone cares to argue to change it bring convincing arguments.

manasdk:
I decided to preserve the trigger and action view as a alternate view see https://github.com/StackStorm/st2/pull/2284.

Closing this revert.

PR #2284, opened 2015-12-10
Provide an iftt switch for rule list display

manasdk:
If trigger and action are desired in rule display use `st2 rule list --iftt`


Kami:
`iftt` or `ifttt`? (I'm personally not a fan of any of those since they are not really explicit) :P

manasdk:
> iftt or ifttt? (I'm personally not a fan of any of those since they are not really explicit) :P

don't worry about it too much, its an easter egg of sorts anyway. I don't want to use ifttt and that was a conscious choice :). If we like this then yes we can think of a better name.

Kami:
OK, works for me.

I'm personally not a big fan of `ifttt` word anyway since I always forgot to add the last `t` :)

PR #2286, opened 2015-12-11
Bug Fix: add get_logger function to MockSensorService

tim-ireland:
When testing a pack unit test for a sensor that calls `get_logger` in it's `__init__` method, I observed a traceback:

```
  File "/st2contrib/packs/plexxi/tests/test_plexxi_event_sensor.py", line 31, in test_run
    sensor = self.get_sensor_instance()
  File "/st2/st2tests/st2tests/base.py", line 440, in get_sensor_instance
    instance = self.sensor_cls(**kwargs)  # pylint: disable=not-callable
  File "/st2contrib/packs/plexxi/sensors/plexxi_event_sensor.py", line 78, in __init__
    self._logger = self._sensor_service.get_logger(name=self.__class__.__name__)
  File "/st2/st2reactor/st2reactor/container/sensor_wrapper.py", line 71, in get_logger
    logger_name = '%s.%s' % (self._sensor_wrapper._logger.name, name)
AttributeError: 'MockSensorWrapper' object has no attribute '_logger'
```

My sensor service init method looks like:
```
    super(EventSensor, self).__init__(sensor_service=sensor_service, config=config)
    self._logger = self._sensor_service.get_logger(name=self.__class__.__name__)
```

Considering that it's a mock, and I'm not interested in testing logging, I'm returning `None` for the logger instance.

### DIFF
--- a/st2tests/st2tests/mocks/sensor.py
+++ b/st2tests/st2tests/mocks/sensor.py
@@ -49,7 +49,7 @@ class MockSensorService(SensorService):
 
     def get_logger(self, name):
         return None
-        
+
     def dispatch_with_context(self, trigger, payload=None, trace_context=None):
         item = {
             'trigger': trigger,
--- a/st2tests/st2tests/mocks/sensor.py
+++ b/st2tests/st2tests/mocks/sensor.py
@@ -47,6 +47,9 @@ class MockSensorService(SensorService):
         # Holds a list of triggers which were dispatched
         self.dispatched_triggers = []
 
+    def get_logger(self, name):
+        return None
+        
     def dispatch_with_context(self, trigger, payload=None, trace_context=None):
         item = {
             'trigger': trigger,


Kami:
Sorry for the delay - I missed this one. Good catch.

One thing we should probably do is return an actual logger instance local to that sensor (we should store a reference to the logger on the class - this way users can potentially also check logger for messages). If we don't do that it will throw on sensor which actually expect logger to be present.

Kami:
Actually, now that I think more about it - Ideally we should return a `MockLogger` instance (might need to write that class ourselves, I've already done something like this in the past, but I'm not sure if  in this repo) which also stores all the `log` method calls in an internal array.

This way users can also easily check if a particular message was logged. Let me know if you are going to do that, otherwise I can do it.

manasdk:
MockLogger sounds great. Likely a separate PR?

Kami:
Alright, I will merge this and get `MockLogger` change done in a separate PR.

tim-ireland:
Hi @Kami, I'd be happy to work on the MockLogger as a separate PR.

Kami:
@tim-ireland Thanks, I actually just committed it here - https://github.com/StackStorm/st2/pull/2297 :)

tim-ireland:
@Kami ah, excellent!  :beer:

Issue #2242, opened 2015-11-27
Self-signed certificate is broken


ramelito:
$ sudo dpkg-reconfigure ca-certificates
Processing triggers for ca-certificates (20141019ubuntu0.14.04.1) ...
Updating certificates in /etc/ssl/certs... unable to load certificate
140168073868960:error:0D07207B:asn1 encoding routines:ASN1_get_object:header too long:asn1_lib.c:150:
WARNING: StackStorm_Auto-Generated_Trusted_CA.pem does not contain a certificate or CRL: skipping
WARNING: Skipping duplicate certificate st2_root_ca.pem
WARNING: Skipping duplicate certificate st2_root_ca.pem

lakshmi-kannan:
@ramelito Do you have more logs? I'll try to repro myself. If not, more logs would help.

ramelito:
Well, this output is from Ubuntu box when i tried to update ca-certificates with st2_root_ca.crt i have downloaded from WebUI install page. So i dont think i have more logs. I have deleted /etc/ssl/st2/* certificates and start update-system to generate certificates once again, they are generated ok, but i still cannot to add them to Ubuntu, same error.

lakshmi-kannan:
When you do ls -l /etc/ssl/st2, do you see a st2.key file and is > 0 bytes? We are talking about this on slack community. Feel free to signup stackstorm.com/community-signup.

jfryman:
Hi @ramelito!

Sorry you're running into issues. It looks like that we needed to push a small fix to get the proper SSL CA into the trusted cert store. That was pushed in https://github.com/StackStorm/st2workroom/pull/272.

Would you try re-running `update-system` and reporting back? Also, not a bad idea to manually delete `/usr/local/ca-certificates/StackStorm_Auto-Generated_Trusted_CA.pem`

Let us know how it goes!

ramelito:
Hi, @jfryman 
Here is the output from update-system
Error: /Stage[main]/Profile::St2server/Ca_cert::Ca[stackstorm-ca]/File[stackstorm-ca.crt]: Could not evaluate: Could not retrieve file metadata for : Cannot find file: Invalid mount ''
Notice: /Stage[main]/Ca_cert::Update/Exec[ca_cert_update]: Dependency File[stackstorm-ca.crt] has failures: true
Warning: /Stage[main]/Ca_cert::Update/Exec[ca_cert_update]: Skipping because of failed dependencies
[...]

PR #2287, opened 2015-12-11
Requests verify kwarg must be True or False.

pixelrebel:
This is the first step to addressing issue https://github.com/StackStorm/st2/issues/2242 (InsecurePlatformWarning).  

This patch along with putting the following into `~/.st2/config` will allow for a successfully verified request in the CLI.
```
[api]
# The FQDN must match the Common Name in the host cert
url = https://st2.mydomain.com:9101
[auth]
# The FQDN must match the Common Name in the host cert
url = https://st2.mydomain.com:9100
[general]
cacert  = /etc/ssl/st2/st2_ca.crt 
```
Note, even if you configure the chatops user's `.st2/config` the requests fail to verify via action_alias, chatops request.

### DIFF
--- a/st2client/st2client/utils/httpclient.py
+++ b/st2client/st2client/utils/httpclient.py
@@ -29,7 +29,7 @@ def add_ssl_verify_to_kwargs(func):
     def decorate(*args, **kwargs):
         if isinstance(args[0], HTTPClient) and 'https' in getattr(args[0], 'root', ''):
             cacert = getattr(args[0], 'cacert', None)
-            kwargs['verify'] = cacert if cacert else False
+            kwargs['verify'] = True if cacert else False
         return func(*args, **kwargs)
     return decorate
 


emedvedev:
Thanks for submitting! You'll need to update the tests as well, please take a look at the Travis log: https://travis-ci.org/StackStorm/st2/builds/96229500.
 
I have some reservations about passing `True` instead of CA path, but since we push the CA into the cert store anyway, it should be good.

pixelrebel:
>I have some reservations about passing True instead of CA path

I think you may be right actually.  Setting to `True` makes the configuring the cacert path mute.

For my GeoTrust signed cert, though I have to set `cacert = /usr/share/ca-certificates/mozilla/GeoTrust_Global_CA.crt` for the verification to work otherwise.  If no cacert is specified, the `verify` kwarg reverts to False.

But because the [requests module doesn't actually use the local ca-cert store](http://docs.python-requests.org/en/latest/user/advanced/#ca-certificates), self-signers need a way to specify a cacert. So you need to pass a path.

Furthermore, I have a separate hubot server, and when I have all my configs in order for my chatops user, they don't seem to work anyway.  My action_aliases still return the `InsecurePlatformWarning` warning in chat.  I think it may be better to move these configs to a global config.

PR #2288, opened 2015-12-11
Fix double rendering issue

enykeev:
The problem we're currently having is that in some cases, st2 attempts to render templates for a single action twice: first time it does that in context of actionchain\rule engine and the second time happens in context of action container. So if the first render resulting in parameter having a new template (either by string concatenation or by results of previous action containing a template as a string inside of it), it tries and usually fails to evaluate this new template during the second attempt.

The solution for this problem is to make sure every string only rendered once. One of the ways to achieve that is to make sure we only rendering user input ad not already processed data.

Currently, there's 5 places in st2 where jinja rendering is happening and 5 artifacts that are getting rendered there: action chain runner transform templates from action chain workflow definitions, reactor transform templates from rule definitions, notifier transforms notifications from action definitions, chatops pack transforms formats from aliases and action container handles user input on manual action invocation, mistral calls (which are at the moment basically the same thing) and, due to some previous oversight, the results of the previous four transformations.

In case problem is not obvious to you, here is an example of an action chain that throws an error which from the looks of it, it should not:

```yaml
---
chain:
    -
        name: "do"
        ref: "core.local"
        params:
            cmd: "echo -n {; echo -n { missing_variable }}"
        on-success: "fail"
    -
        name: "fail"
        ref: "core.local"
        params:
            cmd: "echo {{ do.stdout }}"
```

After the first action has been executed, it's stdout now contains a string `{{ missing_variable }}` which should be treated by the system as any other source of random data, simply as a string. Action chain transformer resolves `{{ do.stdout }}` in `cmd` parameter during an execution of a second task and sends the resulting `echo {{ missing_variable }}` to action container which also treats it as a template instead of a simple string and returns `missing dependency` error because clearly, there's no such variable in action context. Or in case the variable is here, it passes correctly and then does something very different from what author initially planned.

So, to make sure every template only rendered once I'm proposing to split action container renderer into two independent functions and wire the system in such a way, that calls coming from mistral or user input are handled with both functions when the one coming from action chain or reactor only processed by the second one. First function should only transform parameters user have entered during this particular execution (aka liveaction.parameters). Second one would be handling parameters that user haven't provided, but are still required for action to execute (defaults from runner and action parameters). We should treat parameters coming from rule engine and action chain as already transformed and only use them as values for rendering default ones.

Since this change would require us to almost completely rewrite action container renderer, I've also decided to implement a new one using `networkx` graph library to determine the dependency graph instead of iterating over parameters until they are all rendered or failed enough times.


emedvedev:
This is some really impressive work. Trying to catch up.

st2actions/st2actions/utils/param_utils.py
```diff
@@ -45,295 +43,158 @@ def pf(params, skips):
...
+        return v
 
-        # Finally pick up override from actionexec_runner_parameters
-        if param_name in actionexec_runner_parameters:
-            resolved_params[param_name] = actionexec_runner_parameters[param_name]
+    return cast(v)
 
-    return resolved_params
 
+def process(G, name, value):
```
manasdk: docs please. Docs here shoudl explain what does process intend to do?

st2actions/st2actions/utils/param_utils.py
```diff
@@ -45,295 +43,158 @@ def pf(params, skips):
...
+    env = Environment(undefined=StrictUndefined)
+    if isinstance(value, str):
```
manasdk: why?

st2actions/st2actions/utils/param_utils.py
```diff
@@ -45,295 +43,158 @@ def pf(params, skips):
...
+        value = to_unicode(value)
+    template_ast = env.parse(value)
+    dependencies = meta.find_undeclared_variables(template_ast)
+    if dependencies:
```
manasdk: comment explaining what you aim to achieve with dependencies

st2actions/st2actions/utils/param_utils.py
```diff
@@ -45,295 +43,158 @@ def pf(params, skips):
...
+    render_context = {}
+    env = Environment(undefined=StrictUndefined)
 
-def _get_resolved_action_params(runner_parameters, action_parameters,
-                                actionexec_action_parameters):
-    # Create action parameters by merging default values with dynamic values
-    resolved_params = {k: v['default'] if 'default' in v else None
-                       for k, v in six.iteritems(action_parameters)
-                       if k not in runner_parameters}
+    G = nx.DiGraph()
```
manasdk: I think you should use a more descriptive variable name rather then G.

st2actions/st2actions/utils/param_utils.py
```diff
@@ -45,295 +43,158 @@ def pf(params, skips):
...
+    G.add_node(ACTION_CONTEXT_KV_PREFIX, value=action_context)
 
-    # pick overrides from actionexec_action_parameters
-    for param_name, param_value in six.iteritems(action_parameters):
-        # No override if param is immutable
-        if param_value.get('immutable', False):
-            continue
-        if param_name in actionexec_action_parameters and param_name not in runner_parameters:
-            resolved_params[param_name] = actionexec_action_parameters[param_name]
+    for name, value in six.iteritems(liveaction_parameters):
```
manasdk: Why are `liveaction_parameters` processed before others?

st2actions/st2actions/utils/param_utils.py
```diff
@@ -45,295 +43,158 @@ def pf(params, skips):
...
-        if not _check_availability(k, v, renderable_params, context):
-            msg = 'Dependecy unsatisfied - %s: %s.' % (k, v)
-            raise actionrunner.ActionRunnerException(msg)
-        dep_chain = []
-        dep_chain.append(k)
-        if not _check_cyclic(dep_chain, dependencies):
-            msg = 'Cyclic dependecy found - %s.' % dep_chain
+    for name in G.nodes():
+        if 'value' not in G.node[name] and 'template' not in G.node[name]:
+            msg = 'Dependecy unsatisfied in %s' % name
```
manasdk: (Nice to have)
Can we point out which dependency?

st2actions/st2actions/utils/param_utils.py
```diff
@@ -45,295 +43,158 @@ def pf(params, skips):
...
             raise actionrunner.ActionRunnerException(msg)
 
+    if not nx.is_directed_acyclic_graph(G):
```
manasdk: This can be a separate method. Would lead some automatic documentation. 

Also, I personally find long method hard to read even though I have been responsible for a fair few long methods myself.

st2actions/st2actions/utils/param_utils.py
```diff
@@ -45,295 +43,158 @@ def pf(params, skips):
...
+        msg = 'Cyclic dependecy found'
+        raise actionrunner.ActionRunnerException(msg)
 
-def _do_render_params(renderable_params, context):
+    for name in nx.topological_sort(G):
```
manasdk: This can be a separate method. Would lead some automatic documentation.

Also, I personally find long method hard to read even though I have been responsible for a fair few long methods myself.

st2actions/st2actions/utils/param_utils.py
```diff
@@ -45,295 +43,158 @@ def pf(params, skips):
...
+                render_context[name] = env.from_string(node['template']).render(render_context)
+            except Exception as e:
+                LOG.debug('Failed to render %s: %s', name, e, exc_info=True)
+                msg = 'Failed to render parameter "%s": %s' % (name, str(e))
+                raise actionrunner.ActionRunnerException(msg)
+        if 'value' in node:
+            render_context[name] = node['value']
+
+    params = {}
+    for name in liveaction_parameters:
```
manasdk: Even runner_parameters would require casting. Is this handled elsewhere?

manasdk:
Overall the switch to `networkx` seems to make sense. 

Will watch for changes that fix the double rendering issues.

st2actions/st2actions/utils/param_utils.py
enykeev: Sure, I'll add them. This was a repeatable pattern across the whole implementation so I've made it a method the last moment so i don't have to fix it in multiple places. What it does is that it creates a graph node corresponding to the parameter. It will either be a value node, a root of dependency subtree or a template node - a branch to another value node.

st2actions/st2actions/utils/param_utils.py
enykeev: Because of this test https://github.com/StackStorm/st2/blob/fix/double_rendering/st2actions/tests/unit/test_param_utils.py#L329-L345. If you don't convert this bytestring to unicode, jinja will default to ascii parser won't be able to handle unicode characters. As a rule of thumb, in 2015, all strings should be unicode strings, no exceptions.

st2actions/st2actions/utils/param_utils.py
enykeev: I'll add. I need dependencies to traverse the resulting tree in a way that we only reach a node only when all its dependencies are resolved and have a way to spot cyclic and missing deps without making any transformations.

st2actions/st2actions/utils/param_utils.py
enykeev: G is what nx used in their examples and though I can be more descriptive, there's nothing special in this this variable except that it is a graph. A general purpose graph we're creating and throwing away in the span of single method. I'm even planning to rewrite `process` so it didn't go there either.

st2actions/st2actions/utils/param_utils.py
enykeev: The order we add them in the graph not important in this case. Until we start traversing it, the graph is just a pile of nodes and edges. Nodes we create manually with `add_node` would have either template or value property and the nodes that are created automatically as one of the points in `add_edge` would be empty. If we end up with an empty node, it would mean a missing dependency.

st2actions/st2actions/utils/param_utils.py
enykeev: Yeah, I guess we can try and follow the connection to see where we're using this dependency.

st2actions/st2actions/utils/param_utils.py
enykeev: I didn't do any refactoring besides separating `processing` and I only did that because this pattern was repeating quite extensively. I personally find it hard to read the code when you need to jump across multiple methods, files or even projects to understand what's going on (especially, when you're not using PyCharm). I'll try to find a middle ground as soon as I finish the functional part of this fix.

st2actions/st2actions/utils/param_utils.py
enykeev: To be honest, I didn't yet quite figured out when casting should or should not happen. It would be extremely awesome if you could provide couple more tests illustrating how casting should work in extreme situations.

st2actions/st2actions/utils/param_utils.py
manasdk: Actually I wanted you to drop a comment in the code.

enykeev: and I will =)

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1450157328013421

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1450157578013438

manasdk:
I tried out this branch to get a clear grip on the behavior -

setup a key
```
$ st2 key set timeout 60
+----------+---------+
| Property | Value   |
+----------+---------+
| name     | timeout |
| value    | 60      |
+----------+---------+
```

Now the execution
```
$ st2 run core.local cmd="echo {{system.timeout}}"
```

What is save in the DB.
```
$ st2 execution get 56709e9932ed35726cd0791d -dj
{
    "action": {
        "ref": "core.local"
    },
    "context": {
        "user": "stanley"
    },
    "end_timestamp": "2015-12-15T23:13:29.575703Z",
    "id": "56709e9932ed35726cd0791d",
    "liveaction": {
        "action": "core.local",
        "callback": {},
        "id": "56709e9932ed35726cd0791c",
        "parameters": {
            "cmd": "echo 60"
        },
        "runner_info": {
            "hostname": "st2dev",
            "pid": 29298
        }
    },
    "parameters": {
        "cmd": "echo {{system.timeout}}"
    },
    "result": {
        "failed": false,
        "return_code": 0,
        "stderr": "",
        "stdout": 60,
        "succeeded": true
    },
    "start_timestamp": "2015-12-15T23:13:29.466055Z",
    "status": "succeeded"
}
```

I see `liveaction.parameters` is as per your offline description i.e resolved and `parameters` is unresolved. Ideally, would've preferred for liveaction.parameters to be unresolved but this is perfectly acceptable. It is just worth noting for future reference that these values will differ.

Now, what happen if we change `timeout`

```
$ st2 key set timeout 65
+----------+---------+
| Property | Value   |
+----------+---------+
| name     | timeout |
| value    | 65      |
+----------+---------+
```

and on a re-run

```
$ st2 execution re-run 56709e9932ed35726cd0791d
.
id: 56709eca32ed35726cd07920
status: succeeded
parameters:
  cmd: echo {{system.timeout}}
result:
  failed: false
  return_code: 0
  stderr: ''
  stdout: 65
  succeeded: true
```

This behavior is reasonable :)

:+1:

enykeev:
:metal: then I just need to test that it's still reasonable in the other cases (chains, mistral, rules, notifications and stuff) and split the tests so they would start test functions individually instead of their cumulative effect.

Any idea how we can ensure I didn't introduce any regressions? All I can think of is to test a lot of stuff and make sure it behaves the same on both an old and a new version or throws kind of similar errors.

manasdk:
> Any idea how we can ensure I didn't introduce any regressions?

Hopefully our units are sufficient. If not the integrations should capture some basic stuff. Apart from that our only real option is to watch for regression in the field.

Kami:
Agreed with @manasdk comment.

Kami:
On a related note - I also encounter this issue while working on Github weekly activity report email.

I will test it with this branch and see if it works.

Kami:
I've confirmed that my workflow works with your change - nice work :+1:

jfryman:
> I've confirmed that my workflow works with your change - nice work :+1: 

mkms

enykeev:
Ok, since local dev env is giving me a hard time today, I'm going to merge that, test on staging and fix the problems as they appear. Apologies if it will trigger a rebase for you.

Kami:
:clap:

st2common/st2common/util/param.py
```diff
@@ -0,0 +1,212 @@
...
+        try:
+            context[name] = _render(node, context)
+        except Exception as e:
+            LOG.debug('Failed to render %s: %s', name, e, exc_info=True)
+            msg = 'Failed to render parameter "%s": %s' % (name, str(e))
+            raise ParamException(msg)
+    return context
+
+
+def _cast_params_from(params, context, schemas):
```
Kami: Sorry for a bit late comment, but it would also be nice to add some unit tests for all of those "private" functions (there are plenty).

enykeev: Yeah, that's my plan. The only thing I'm not sure about is whether I should do it now or postpone until we have a clear understanding this thing is working and doesn't require a major changes because it suddenly stopped supporting some crucial use case.

enykeev:
I've deployed this one on st2stage201 and from what I can tell, it works just fine. If you guys can kick some tires for a little bit to help us gain confidence in the change, I would really appreciate that. May be we should go as far as deploy it to ops or build. I tried running st2cd actions on stage, but I'm afraid I'll spend just too much time trying to make them work due to amount of stuff happening behind the scene.

Kami:
Yeah, I think it's a good idea to deploy it to st2build.

In case thing breaks we will need to revert to an older version, but that's not a problem since the packages are already built (there will be a little back and forth if things break, but nothing too bad - we will need to revert to an older version, fix issue, build packages for fixed version, deploy fixed versions).

manasdk:
Caused a regression -

```
$ st2 run fixtures.streamwriter-script-local stream=stdout
.
id: 56734ed932ed356478d5bcde
status: failed
parameters:
  stream: stdout
result:
  error: 'unsupported operand type(s) for +: ''NoneType'' and ''unicode'''
  traceback: "  File "/mnt/st2repos/st2/st2actions/st2actions/container/base.py", line 116, in _do_run
    (status, result, context) = runner.run(action_params)
  File "/mnt/st2repos/st2/st2actions/st2actions/runners/localrunner.py", line 101, in run
    named_args = self._transform_named_args(named_args)
  File "/mnt/st2repos/st2/st2actions/st2actions/runners/__init__.py", line 163, in _transform_named_args
    return {self._kwarg_op + k: v for (k, v) in six.iteritems(named_args)}
  File "/mnt/st2repos/st2/st2actions/st2actions/runners/__init__.py", line 163, in <dictcomp>
    return {self._kwarg_op + k: v for (k, v) in six.iteritems(named_args)}
"
```

Seen in the CI environment.

PR #2291, opened 2015-12-12
Add flag to register content script to fail on failure

Kami:
This pull request adds `--register-fail-on-failure` flag to the resource registrar script. This value defaults to `False` so it's fully backward compatible and the default behavior doesn't change.

If this flag is provided, registrar will fail and exit with non-zero if a registration of any resource fails (validation error, etc.)

This will come handy in a new st2contrib test task which I'm working on - this task will fail if any of the resource registration fails (we don't want broken definitions in the repo and we want to fail early and loudly).

P.S. Yes I know those files new more refactoring love, there is sadly still tons of duplicated code (I already did some of the refactoring in the past, but it needs a lot more love).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,6 +19,9 @@ in development
   action.ref and enabled. This aligns closer the UI and also brings important information front and
   center. (improvement)
 * Action and Trigger filters for rule list (new-feature)
+* Add ``--register-fail-on-failure`` flag to ``st2-register-content`` script. If this flag is
+  provided, the script will fail and exit with non-zero status code if registering some resource
+  fails. (new feature)
 
 1.2.0 - December 07, 2015
 -------------------------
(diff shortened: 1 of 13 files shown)


Kami:
If there are no objections I will go ahead and merge this.

manasdk:
Seeing 

```
2015-12-15 16:25:42,476 ERROR [-] Failed registering all sensors from pack "None": 'NoneType' object has no attribute 'endswith'
  File "/mnt/st2repos/st2/st2common/st2common/bootstrap/sensorsregistrar.py", line 60, in register_sensors_from_packs
    sensors = self._get_sensors_from_pack(sensors_dir)
  File "/mnt/st2repos/st2/st2common/st2common/bootstrap/sensorsregistrar.py", line 105, in _get_sensors_from_pack
    return self.get_resources_from_pack(resources_dir=sensors_dir)
  File "/mnt/st2repos/st2/st2common/st2common/bootstrap/base.py", line 69, in get_resources_from_pack
    if resources_dir.endswith('/'):
AttributeError: 'NoneType' object has no attribute 'endswith'
```

After this was merged.

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1450225641014510

PR #2294, opened 2015-12-14
Task/vki/pxpse 421 add st2auth to st2conflogrotat

vignesh-terafast:
Added st2auth.log and st2auth.audit.log to conf/logrotate.conf


Kami:
Good catch - no idea how we missed that :)

conf/logrotate.conf
```diff
@@ -20,6 +20,23 @@ notifempty
   endscript
 }
 
+## AUTH
+/var/log/st2/st2auth.log {
```
andrew-regan: Should this be "/var/log/st2/st2auth*.log" since the audit entry added below is identical to this one?

Kami: We only run one instance of st2auth process and don't include PID in the log file name (like in the action runner case) so I believe this is fine.

andrew-regan: I probably should have said use this "/var/log/st2/st2auth*.log" and then eliminate this identical section for audit:
/var/log/st2/st2auth.audit.log {
  daily
  rotate 5
  postrotate
    st2ctl reopen-log-files st2auth
  endscript
}

Kami: Usually it's better to have two entries one for normal and one for audit file - the reason for that is that users might want to keep audit log files longer than regular log files.

andrew-regan: Should we break out results tracker, notifier, and sensor container into separate entries for audit and non-audit to be consistent? They are currently using *.log. Last question. :-)

Kami: IIRC, only st2api and st2auth really log audit log messages so we should be fine.

Kami:
Merged, thanks.

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1450120535013167

PR #2297, opened 2015-12-15
Add missing MockSensorService.get_logger method

Kami:
This pull request updates `MockSensorService.get_logger` method to return an instance of `Mock` class. This allows user to assert that a particular message has been logged / logger method has been called with particular arguments (you can find an example of how to do that in the tests below).

The reason why I have decided to simply use Mock class instead of re-inventing the wheel is that it does the job and it exposes a familiar Mock interface.


manasdk:
:+1:

PR #2298, opened 2015-12-16
Ignore missing type directories

manasdk:
Fixes regression reported in https://github.com/StackStorm/st2/pull/2291#issuecomment-164944170

### DIFF
--- a/st2common/st2common/bootstrap/actionsregistrar.py
+++ b/st2common/st2common/bootstrap/actionsregistrar.py
@@ -54,6 +54,9 @@ class ActionsRegistrar(ResourceRegistrar):
                                                 content_type='actions')
 
         for pack, actions_dir in six.iteritems(content):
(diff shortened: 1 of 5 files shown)


lakshmi-kannan:
lgtm

Kami:
Good catch and thanks :+1: 

It looks like I missed this because we do skip directories which are None in `register_*_from_pack`, but not in `register_*_from_packs`.

Kami:
On a related note - now we have some integration tests for register content script (https://github.com/StackStorm/st2/pull/2291/files#diff-29a7344e3222e95345c48c7d3daba058R1) so it would also be a good idea to add a test case for this scenario.

I will go ahead and do that in a separate PR.

Kami:
Here we go - https://github.com/StackStorm/st2/pull/2300

PR #2299, opened 2015-12-16
Avoid leaving zombie executions

manasdk:
The general idea is that at the quit of an `actionrunner` a specific process will simply marking incomplete executions as abandoned. This way they are not left in a `RUNNING` state.

* This is done on SystemExit leveraging existing shutdown code-path.
* SIGTERM is handled but SIGKILL will still leave zombies. Using `kill -9 <runner-pid>` is not recommended and I guess will like be an extreme case anyway.


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1450241217014554

st2actions/st2actions/worker.py
```diff
@@ -88,6 +89,12 @@ def process(self, liveaction):
                 if liveaction.status == action_constants.LIVEACTION_STATUS_SCHEDULED
                 else self._cancel_action(liveaction_db))
 
+    def shutdown(self):
+        super(ActionExecutionDispatcher, self).shutdown()
+        # Abandon running executions if incomplete
+        for liveaction_id in self._running_liveactions:
+            executions.abandon_execution_if_incomplete(liveaction_id=liveaction_id)
+
```
lakshmi-kannan: Better to explicitly remove items from the list here.

st2common/st2common/services/executions.py
```diff
@@ -136,10 +137,29 @@ def update_execution(liveaction_db, publish=True):
...
+    execution implies that its end state is unknown and cannot anylonger
+    be determined. This method should only be called if the owning process
+    is certain it can no longer determine status of an execution.
+    """
+    liveaction_db = action_utils.get_liveaction_by_id(liveaction_id)
+    # No need to abandon and already complete action
+    if liveaction_db.status in action_constants.LIVEACTION_COMPLETED_STATES:
+        return
+    liveaction_db = action_utils.update_liveaction_status(
+                status=action_constants.LIVEACTION_STATUS_ABANDONED,
```
lakshmi-kannan: How would this work for action chain? We have to abandon the parent workflow execution and the task execution, right?

st2actions/st2actions/cmd/actionrunner.py
```diff
@@ -19,9 +20,22 @@
     time=True)
 
 
+def _setup_sigterm_handler():
```
lakshmi-kannan: I'd prefer a base class that individual classes can inherit from - BaseProcess or something like that. I haven't thought about how it would fit our existing code.

st2actions/st2actions/cmd/actionrunner.py
```diff
@@ -19,9 +20,22 @@
...
+
+        def sigterm_handler(signum=None, frame=None):
+            # This will cause SystemExit to be throw and we call sensor_container.shutdown()
```
lakshmi-kannan: Copy pasta.

st2actions/st2actions/worker.py
manasdk: I am not sure what that would achieve. It is a shutdown method.

st2common/st2common/services/executions.py
manasdk: Yeah that just work since it is eventually an execution being handled by some action-runner. If the action-runner handling the child execution is still around and processing then that execution will succeed.

st2actions/st2actions/cmd/actionrunner.py
manasdk: Its possible. It will be easier to implement a Base type if a pattern emerges. For now it will be just 1 method that registers handlers.

There is already https://github.com/StackStorm/st2/blob/master/st2common/st2common/service_setup.py

st2common/st2common/services/executions.py
lakshmi-kannan: Should we at least try to cancel the child task?

manasdk: I am honestly not sure if thats a good idea. The owing ActionRunner should either abandon that execution or that execution will complete which is a good thing.

manasdk: Further there is also https://github.com/StackStorm/st2/blob/master/st2common/st2common/signal_handlers.py

st2actions/st2actions/cmd/actionrunner.py
manasdk: Further there is also https://github.com/StackStorm/st2/blob/master/st2common/st2common/signal_handlers.py

I would like to avoid any major refactoring as part of this PR or create another interface/utility. It would be nice to try to collect some of these into a cohesive module and see what approach makes most sense. It can easily become a rabbit hole.

manasdk: Avoiding this can of worms right now. Will make a note to refactor.

st2actions/st2actions/handlers/mistral.py
```diff
@@ -39,6 +39,7 @@
     action_constants.LIVEACTION_STATUS_SUCCEEDED: 'SUCCESS',
     action_constants.LIVEACTION_STATUS_FAILED: 'ERROR',
     action_constants.LIVEACTION_STATUS_TIMED_OUT: 'ERROR',
+    action_constants.LIVEACTION_STATUS_ABANDONED: 'ERROR',
```
lakshmi-kannan: The string looks incorrect on RHS

st2actions/st2actions/worker.py
```diff
@@ -88,6 +89,16 @@ def process(self, liveaction):
...
 
+    def shutdown(self):
+        super(ActionExecutionDispatcher, self).shutdown()
+        # Abandon running executions if incomplete
+        while self._running_liveactions:
+            liveaction_id = self._running_liveactions.pop()
+            try:
+                executions.abandon_execution_if_incomplete(liveaction_id=liveaction_id)
+            except:
+                LOG.exception('Failed to abandon liveaction %s.', liveaction_id)
```
lakshmi-kannan: You still didn't delete the liveaction from the list :P

st2actions/st2actions/handlers/mistral.py
lakshmi-kannan: NVM. This is just weird but whatever. Sorry, wrong call!

docs/source/upgrade_notes.rst
```diff
@@ -3,6 +3,13 @@ Upgrade Notes
 
 |st2| In Development
 --------------------
+* New ``abandoned`` action execution status has been introduced. This is effectively a failure
```
lakshmi-kannan: An execution is set to ``abandoned`` state when an action runner running it is killed via TERM.

lakshmi-kannan:
l,g,t,m :ship:

st2actions/st2actions/worker.py
manasdk:
```
In [1]: x = set([1, 2, 3])

In [2]: x
Out[2]: {1, 2, 3}

In [3]: while x:
   ...:     print x.pop()
   ...:
1
2
3

In [4]: x
Out[4]: set()
```

CHANGELOG.rst
```diff
@@ -27,6 +27,8 @@ in development
 * Add a missing ``get_logger`` method to the `MockSensorService``. This method now returns an
   instance of ``Mock`` class which allows user to assert that a particular message has been
   logged. [Tim Ireland, Tomaz Muraus]
+* Introduce a new ``abandoned`` state that is applied to executions that we cannot guarantee as
+  completed. Typically happen when an actionrunner currently running some executions quits.
```
lakshmi-kannan: s/quits/killed via TERM.

st2actions/st2actions/worker.py
lakshmi-kannan: oh pop. cool cool.

CHANGELOG.rst
manasdk: Ignoring comment.

lakshmi-kannan: lol. Sorry, my bad. When it quits or killed via TERM.

manasdk: done. updated.

lakshmi-kannan: <3

Kami:
As far as the general approach goes - I'm fine with that solution as a stop gap (it's probably sufficient for now).

As you have pointed out, there is nothing we can do with this approach if service is killed with `SIGKILL`.

Having said that, as a long term and more robust / fail-safe solution I think it would be a good idea to check for zombies and (also) perform rectification on the service startup - ideally the code should be the same so we can call the same code in different places (signal handler, service startup, periodic garbage collection green thread or similar).

Imo, that's generally a good pattern in distributed systems - service and process can die or be killed at any time so you can't rely on a particular signal handler to always be called - yes you can still have it there for a common / average case, but for the worst case you still need something else (and distributed systems all all about worst cases anyway :P)

Anyway, this should do for now, just something to think about for a future solution (or in case people still complain a lot about zombie execution after this change is out).

Kami:
Also on a slightly related note - I'm working on garbage collection service for purging old data and eventually we could also put something like this there to make sure the cleanup runs on some periodic interval.

manasdk:
>  I'm working on garbage collection service for purging old data and eventually we could also put something like this there to make sure the cleanup runs on some periodic interval.

Original plan was to come up with a 2 step approach.
1. cleanup on service/process exit
2. write a purge tool that can be run ad-hoc or on a schedule.

Once the need for a reaper showed up I decided to hold off on step 2 and I think it will be best place to write this code. So will wait on that PR to land and then include.

manasdk:
> perform rectification on the service startup

This is a little tricky as we also need to know all the actionrunner PIDs to identify zombies. Some sort of service discovery would be useful.

PR #2300, opened 2015-12-16
Add a regression test case for register content script

Kami:
This PR adds a regression test case for this scenario - https://github.com/StackStorm/st2/pull/2298.

It makes sure the script doesn't throw if a pack directory is missing directory for some resource which is to be registered (sensors/, etc.).

### DIFF
--- a/st2common/tests/integration/test_register_content_script.py
+++ b/st2common/tests/integration/test_register_content_script.py
@@ -27,10 +27,12 @@ SCRIPT_PATH = os.path.abspath(SCRIPT_PATH)
 
 BASE_CMD_ARGS = [SCRIPT_PATH, '--config-file=conf/st2.tests.conf', '-v', '--register-actions']
 
-test_config.parse_args()
-
 
 class ContentRegisterScripTestCase(IntegrationTestCase):
+    def setUp(self):
+        super(ContentRegisterScripTestCase, self).setUp()
+        test_config.parse_args()
+
     def test_register_from_pack_success(self):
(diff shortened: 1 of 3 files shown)


st2common/tests/integration/test_register_content_script.py
```diff
@@ -66,3 +68,21 @@ def test_register_from_pack_action_metadata_fails_validation(self):
...
         self.assertTrue('object has no attribute \'get\'' in stderr)
         self.assertEqual(exit_code, 1)
+
+    def test_register_from_packs_doesnt_throw_on_missing_pack_resource_folder(self):
+        # dummy_pack_4 only has actions folder, make sure it doesn't throw when
+        # sensors and other resource folders are missing
+
+        # Note: We want to use a different config which sets fixtures/packs_1/
+        # dir as packs_base_paths
+        cmd = [SCRIPT_PATH, '--config-file=conf/st2.tests1.conf', '-v', '--register-sensors']
```
Kami: I'm not a fan of a separate config thing, but it's a workaround for now.

I already mentioned that we need to improve this in the past. Right now, a lot of tests assume `st2tests/st2tests/fixtures/` is set for `packs_base_paths`, but that's not actually a good thing since not everything under that directory represents a pack.

To fix this, we need to move actual packs (mostly dummy_pack_* to some sub-directory, set that for `packs_base_paths` in tests configs and update affected code).

I can't commit to this right now since it will probably be quite consuming to get all the tests to pass, but I will open an issue.

PR #2302, opened 2015-12-16
Add get_runner function to noop runner

DoriftoShoes:
The noop runner is missing the get_runner function that is needed during runner import.


manasdk:
Lint fixes needed

:+1:

lakshmi-kannan:
@DoriftoShoes Lint it baby one more time!

PR #2306, opened 2015-12-18
Fix param dependency override regression

enykeev:
As a result of #2288, we have a regression when action parameter without default `p1: {}` overrides runner parameter with default value `p1: {default: '...'}` resulting in a final parameter rendering as `None` instead of runner parameter default.

Kudos to @manasdk 

STORM-1924 #resolve

### DIFF
--- a/st2common/st2common/util/param.py
+++ b/st2common/st2common/util/param.py
@@ -106,8 +106,11 @@ def _process_defaults(G, schemas):
     '''
     for schema in schemas:
         for name, value in six.iteritems(schema):
-            if name not in G.node or value.get('immutable', False):
-                _process(G, name, value.get('default', None))
+            absent = name not in G.node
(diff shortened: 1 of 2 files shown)


st2common/st2common/util/param.py
```diff
@@ -106,8 +106,11 @@ def _process_defaults(G, schemas):
     '''
     for schema in schemas:
         for name, value in six.iteritems(schema):
-            if name not in G.node or value.get('immutable', False):
-                _process(G, name, value.get('default', None))
+            abscent = name not in G.node
```
manasdk: absent?

manasdk:
+1

st2common/st2common/util/param.py
enykeev: Sure. Can't spell =)

Kami:
:+1:

PR #2307, opened 2015-12-18
New garbage collection service

Kami:
This pull request introduces new garbage collection service. Right now it periodically deletes action executions and trigger instances which are older than number of days defined in the config.

Eventually we could also move other periodic garbage collection code into this service - e.g. deleting zombie executions.

For backward compatibility reasons and because this value is very installation and user specific garbage collection is disabled by default.

TODO:

- [x] Logging config
- [x] Log rotate config (conf/logrotate.conf)
- [x] Update tools/st2ctl
- [x] dev config update
- [x] Update tools/config_gen.py
- [x] Sample config update
- [x] Integration tests
- [x] Allow user to force garbage collection via SIGUSR2 signal
- [x] Fix existing purge tests
- [x] Documentation
- [x] Update <component>/setup.py (if needed)
- [x] Update conf/st2.{dev,prod,package}.conf
- [x] Packaging manifests (<component>/packaging/{debian,rpm}/*)
- [ ] New packaging stuff

Keep in mind that the PR is not ready yet, I just opened it so I can track what's still to be done. One thing we should also do is move this list (things needs to be done when adding a new service) to development section of our docs.


Kami:
@armab @dennybaa What's the process for adding a new service for the new packages?

Feel free to write it here and I will add it to the development documentation along with the other steps.

armab:
At a high-level as I understand it: including start-up scripts here and there (deb/rpm): 
https://github.com/StackStorm/st2-packages/tree/master/packages/st2actions/rpm
https://github.com/StackStorm/st2-packages/blob/master/packages/st2actions/debian/rules

And then adding Rspec tests to check if services started after builds (smoke tests): 
https://github.com/StackStorm/st2-packages/blob/master/rake/spec/default/60-st2_all-services-ok_spec.rb

You can also test it locally with `docker-compose` (including those smoke tests): https://github.com/stackstorm/st2-packages#usage

That's what I can assume by browsing `st2-packages`.
@dennybaa knows all low-level internals/details behind the packages.

st2reactor/bin/st2garbagecollector
```diff
@@ -0,0 +1,35 @@
...
+#
+
+import os
+import sys
+
+from st2reactor.cmd import garbagecollector
+
+DEVEL = True
+
+if DEVEL:
```
manasdk: what is this?

st2reactor/st2reactor/cmd/garbagecollector.py
```diff
@@ -0,0 +1,64 @@
...
+    common_teardown()
+
+
+def main():
+    try:
+        _setup()
+        garbage_collector = GarbageCollectorService()
+        exit_code = garbage_collector.run()
+    except SystemExit as exit_code:
+        return exit_code
```
manasdk: this is the exception - right?

conf/st2.conf.sample
```diff
@@ -100,6 +100,16 @@ logging = conf/logging.exporter.conf
 # Directory to dump data to.
 dump_dir = /opt/stackstorm/exports/
 
+[garbagecollector]
```
lakshmi-kannan: It would be nice if we could control starting this process based on whether a config is enabled. Are we going to offload to this init systems? By virtue of whether service is enabled or disabled, this will run/not run?

st2reactor/st2reactor/cmd/garbagecollector.py
```diff
@@ -0,0 +1,64 @@
...
+    except:
+        LOG.exception('(PID:%s) GarbageCollector quit due to exception.', os.getpid())
+        return FAILURE_EXIT_CODE
+    finally:
+        _teardown()
+
+    return exit_code
```
manasdk: Will it ever reach here? If it does exit_code may not be defined at this point.

st2reactor/st2reactor/cmd/sensormanager.py
```diff
@@ -34,7 +39,7 @@
     time=True)
 
 
-LOG = logging.getLogger('st2reactor.bin.sensors_manager')
+LOG = logging.getLogger('st2reactor.bin.sensormanager')
```
manasdk: __name__ ?

st2reactor/st2reactor/cmd/sensormanager.py
```diff
@@ -59,6 +64,6 @@ def main():
         return 1
     except:
         LOG.exception('(PID:%s) SensorContainer quit due to exception.', os.getpid())
-        return 2
+        return FAILURE_EXIT_CODE
```
manasdk: was returning 2 now it will be 1.

st2reactor/st2reactor/cmd/sensormanager.py
```diff
@@ -59,6 +64,6 @@ def main():
         return 1
```
manasdk: FAILURE_EXIT_CODE

st2reactor/st2reactor/garbage_collector/base.py
```diff
@@ -0,0 +1,158 @@
...
+
+__all__ = [
+    'GarbageCollectorService'
+]
+
+# TODO: Make config option
+DEFAULT_COLLECT_INTERVAL = 300
+
+# Minimum value for TTL. If user supplies value lower than this, we will throw.
+MINIMUM_TTL_DAYS = 7
```
manasdk: why?

conf/st2.conf.sample
Kami: I was thinking about it, but imo it's an overkill (and it makes monitoring this service harder and more complex)

If no TTLs are configured, the service is very light on resources (it only polls every 300 seconds) so it's not a big deal if it's running.

st2reactor/bin/st2garbagecollector
Kami: That's what I asked on Slack the other days.

It seems like an old code, we have it in other service. Should verify if it's still needed.

st2reactor/st2reactor/garbage_collector/base.py
```diff
@@ -0,0 +1,158 @@
...
+
+LOG = logging.getLogger(__name__)
+
+
+class GarbageCollectorService(object):
+    def __init__(self, collection_interval=60):
```
manasdk: collection_interval should be configurable as well.

st2reactor/st2reactor/cmd/garbagecollector.py
Kami: SystemExit is a bit different, but yes.

st2reactor/st2reactor/cmd/garbagecollector.py
Kami: Yes, if the stuff inside try clause doesn't throw.

st2reactor/st2reactor/cmd/sensormanager.py
Kami: Correct, doesn't look like it matter that much though.

conf/st2.conf.sample
lakshmi-kannan: I am OK with this for now.

st2reactor/st2reactor/garbage_collector/base.py
```diff
@@ -0,0 +1,158 @@
...
+        """
+        Validate that a user has supplied reasonable TTL values.
+        """
+        if self._action_executions_ttl and self._action_executions_ttl < MINIMUM_TTL_DAYS:
+            raise ValueError('Minimum possible TTL in days is %s' % (MINIMUM_TTL_DAYS))
+
+        if self._trigger_instances_ttl and self._trigger_instances_ttl < MINIMUM_TTL_DAYS:
+            raise ValueError('Minimum possible TTL in days is %s' % (MINIMUM_TTL_DAYS))
+
+    def _perform_garbage_collection(self):
```
manasdk: This needs to be somewhat more pluggable. I guess it can be future work.

It would be nice if we can simply register collectors.

st2reactor/st2reactor/garbage_collector/base.py
Kami: I think it's good to have safe-guard, but I could be convinced to lower this threshold.

In theory, we could also support very low TTLs (less than one day), but specifying TTL that low indicates a problem elsewhere and would cause all kinds of issues (MongoDB would probably have tons of issues catching up with the compactions and performance would suck - our current data and storage model is not optimized for that).

st2reactor/st2reactor/garbage_collector/base.py
```diff
@@ -0,0 +1,158 @@
...
+        LOG.info('Performing garbage collection...')
+
+        # todo sleep between queries to avoid busy waiting
+        if self._action_executions_ttl >= MINIMUM_TTL_DAYS:
```
manasdk: Any objection to running in separate eventlet threads and then join on all those threads? This way we would get some pseudo parallelism.

st2reactor/st2reactor/garbage_collector/base.py
lakshmi-kannan: I am +1 to having a bound for now. If people complain, then we can evaluate the actual issue.

st2reactor/st2reactor/garbage_collector/executions.py
```diff
@@ -0,0 +1,105 @@
...
+    'purge_executions'
+]
+
+DONE_STATES = [action_constants.LIVEACTION_STATUS_SUCCEEDED,
+               action_constants.LIVEACTION_STATUS_FAILED,
+               action_constants.LIVEACTION_STATUS_TIMED_OUT,
+               action_constants.LIVEACTION_STATUS_CANCELED]
+
+
+def purge_executions(logger, timestamp, action_ref=None, purge_incomplete=False):
```
manasdk: Why is passing in the logger a good idea. This takes away from the interface of this method and  adds very little value.

st2reactor/st2reactor/garbage_collector/base.py
Kami: No objection, but I don't think we should really be I/O bound here - with the new `delete_by_query` approach it's 3 queries total - and those queries don't block since compactions are performed later in the background so I don't think it would really gain us much, but I can profile it and be sure.

st2reactor/st2reactor/garbage_collector/base.py
manasdk: > but specifying TTL that low indicates a problem elsewhere
That this is a heavily used StackStorm instance?

st2reactor/st2reactor/garbage_collector/base.py
Kami: Yeah, I was actually thinking about something along those lines at the beginning (keeping it simple, having "collections" class attribute which contains a list of callables), but it would be an overkill for now and we can just do it once we move more work here.

st2reactor/st2reactor/garbage_collector/base.py
manasdk: I am sorta trying to push us towards a longer term approach here coz I do want to handle `abandon` work here as well. Anyway, can refactor myself if needed.

manasdk:
Highlevel comment -

probably something more generic than garbagecollector. This could be a good place some background system tasks.

1. Handle some of the `abandon` work in this process.
2. Move over https://github.com/StackStorm/st2/blob/6607b82d84263ba52c3705787af77406ca6406c6/st2actions/st2actions/cmd/st2notifier.py#L32
3. See what failed/abandoned executions can be restarted etc.

manasdk:
> I bet we will come with with more such things.

Alternatively if you can make a case for why such background processing should not even live here then that would be a way to think about this as well.

Kami:
> Alternatively if you can make a case for why such background processing should not even live here then that would be a way to think about this as well.

If we make it any more generic we will just end up calling it a "worker" (hey, we can move everything in here then! :)) and in that sense it won't be an improvement over st2notifier :)

I would personally prefer to only have garbage collection stuff in this service (this means zombie deletion could also live here, etc.).

Yes services add a lot of complexities, are hard to manage, monitor, etc. but we already have a lot of them so we should come up with a concrete plan how to make this easier (service registry, etc.) and then we should embrace that and simply add new task specific services.

manasdk:
> If we make it any more generic we will just end up calling it a "worker"

Fair point but then again this will likely happen anyway.

+1 to service registry.

st2reactor/st2reactor/cmd/garbagecollector.py
dennybaa: Looks like a mult-threaded app) Since usually I haven't seen the purpose to intercept `SystemExit`)

conf/st2.conf.sample
dennybaa: If you'd like to have it as a separate service, it could be nice to bundle it into st2common package...

Plus if you want to enabled/disabled flag it should be better controll it on os level, rather then config. `/etc/default/st2garbage..` or `/etc/sysconfig/st2garbage..`, imo.

Kami: I'm confused, why would we bundle it with st2common? This has nothing to do with st2common (besides depending on it like many other services). It's a separate standalone service,

Maybe I still don't understand new packaging / current state of the new packaging, but I thought we will have a package per service (or the very least, to start with a package per component).

st2reactor/st2reactor/cmd/sensormanager.py
Kami: We totally could :)

st2reactor/st2reactor/garbage_collector/base.py
Kami: No - us not designing the app correctly to handle a very high load :)

As noted above, in all the compaction based systems, doing a lot of deletes in a short time frame usually puts a lot of pressure on the database during the compaction phase. So while we could support a very low TTL, it would probably trash the database and make the system unusable.

If we wanted to support very low TTL, the way to go would be capped collections - MongoDB already kinda solved this problem with capped collections. Those of course have a limit of a fixed file size which is not ideal for this case, but it avoids compaction overhead...

st2reactor/st2reactor/garbage_collector/base.py
Kami: Agreed, will make the change. I will also bump it to something more reasonable by default (e.g. 300 or 600 seconds).

conf/st2.conf.sample
dennybaa: It's ok to have **a component** per package... However it requires to ship all virtualenv and co with the package, this might be okay... But doesn't seem to be so good when there are many packages. Each package - additional same bundle of same everything, which is ~15-20mb of packaged size)) So less packages is better then having many packages, imo.

st2reactor/st2reactor/garbage_collector/base.py
manasdk: OK. I do agree but we probably do not know what a good value is so suggesting it not be enforced is all. Anyway your call

st2reactor/st2reactor/garbage_collector/base.py
Kami: Yeah, I decided to leave it as it is since it doesn't really add any value right now.

Collection of action executions and trigger instances intentionally doesn't run in parallel to avoid too much pressure on the database. But as you have said, when we add more things here (especially ones which need to be collected / rectified more often) and the service gets more busy, using green threads / thread pool might make more sense.

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1450723164016782

Kami:
Alright, all the comments addressed.

I will merge it into master after all the tests pass. I will open a separate PR with changes for the new packaging stuff.

Kami:
Changes for new packages are here - https://github.com/StackStorm/st2-packages/pull/55

Kami:
Apparently there is yet another place which needs to be updated - https://github.com/StackStorm/st2-dockerfiles/pull/2

PR #2308, opened 2015-12-18
Update st2client description and classifiers

Kami:
The title says it all.


estee-tew:
https://stackstorm.slack.com/archives/general/p1450450806001066

PR #2317, opened 2015-12-21
Fix RHEL build pipeline

enykeev:


enykeev:
yolo merge

Kami:
Good catch - we always forget those, heh :)

PR #2318, opened 2015-12-21
Using certifi for SSL validation in st2client

emedvedev:
This PR adds `certifi` to st2client dependencies and changes the SSL validation behavior to True.

To merge this, we need to make sure that our CA gets added to the `certifi` bundle; I’ll do that.

### DIFF
--- a/st2client/st2client/utils/httpclient.py
+++ b/st2client/st2client/utils/httpclient.py
@@ -29,7 +29,7 @@ def add_ssl_verify_to_kwargs(func):
     def decorate(*args, **kwargs):
         if isinstance(args[0], HTTPClient) and 'https' in getattr(args[0], 'root', ''):
             cacert = getattr(args[0], 'cacert', None)
-            kwargs['verify'] = cacert if cacert else False
+            kwargs['verify'] = cacert if cacert else True
         return func(*args, **kwargs)
     return decorate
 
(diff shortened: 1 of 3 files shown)


Kami:
> To merge this, we need to make sure that our CA gets added to the certifi bundle; I’ll do that.

I assume you mean locally aka in our code, right?

Kami:
We also need to make sure that the code where we add our CA to certifi bundle is added early on so it also works in other places where requests is used (where we talk to mistral and back, etc.), not just in st2client.

emedvedev:
Yes. What I want to do is append our root CA into the `certifi` bundle as soon as we generate it. Since it's global, we'll be able to use it anywhere, just need to make sure we don't set `verify=False` in the places you want it to work.

emedvedev:
I kinda want to resurrect this PR in the context of new packages. Any chance we could install certifi as a dependency during the package install and then append our own CA to it?

st2client/st2client/utils/httpclient.py
```diff
@@ -29,7 +29,7 @@ def add_ssl_verify_to_kwargs(func):
     def decorate(*args, **kwargs):
         if isinstance(args[0], HTTPClient) and 'https' in getattr(args[0], 'root', ''):
             cacert = getattr(args[0], 'cacert', None)
-            kwargs['verify'] = cacert if cacert else False
+            kwargs['verify'] = cacert if cacert else True
```
manasdk: Hmm, seems like this should remain False if there is no cacert.

emedvedev: It defaults to `True` since it would check for `certifi` CAs—including our self-signed one, which we're supposed to push to the certifi bundle. 

On a different note, we should really resurrect our efforts to get rid of the "insecure request" warning.

lakshmi-kannan:
Closing.Reopen if you need it.

PR #2319, opened 2015-12-21
Adding extra scripts

DoriftoShoes:


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1450728208016853

Kami:
Good catch :+1:

PR #2322, opened 2015-12-22
added virtualenv dependency

dennybaa:
trying to fix virtualenv dependency https://github.com/StackStorm/st2-packages/issues/50

### DIFF
--- a/fixed-requirements.txt
+++ b/fixed-requirements.txt
@@ -22,3 +22,4 @@ stevedore>=1.7.0,<1.8
 bencode>=1.0,<1.1
 paramiko>=1.15.2,<1.16
 networkx==1.10
+virtualenv>=13.1.2,<14.0
--- a/st2actions/in-requirements.txt
+++ b/st2actions/in-requirements.txt
@@ -19,3 +19,5 @@ lockfile
 python-nmap
 pyinotify
 git+https://github.com/Kami/logshipper.git@stackstorm_patched#egg=logshipper
+# required by pack_mgmt/setup_virtualenv.py#L135
+virtualenv


Kami:
LGTM, but lets also pin it to some known working version / set of versions.

dennybaa:
just a moment... i want to add  st2-auth-backend-flat-file here as well...
I thought recent versions works fine. It'll be > 13.1, @Kami Do you have any specific suggestion?

Kami:
`>=13.1.2,<14.0` just in case :)

dennybaa:
I'm fine to add this version specification. Maybe somebody else wants to draw a word... If not I'll add the above and merge it tomorrow.

Kami:
Just FYI - this change (including virtualenv in top level requirements.txt) might some how breaking workroom.

For now I will remove it from top level requirements.txt (this shouldn't affect your changes), but we should try to dig in and find the root cause,

dennybaa:
Hmm. @Kami I actually haven't added into top level reqs. It shouldn't be there. Where I've added it is `fixed-requirements.txt`. I'd like to add that it's added automatically when you regenerate requirements.

Kami:
@dennybaa I know you haven't added it but we generate top-level `requirements.txt` from those component specific requirements inside our Makefile and include it in the repo.

dennybaa:
@Kami we can easily  remove it from st2client if it's needed. Since there's no use for it until something is decided about refactoring to support STANDALONE components. Split packages can't be used until then anyway.

Kami:
@dennybaa It's fine for now, #2332 solves it :)

But if we decide it it's not needed, then we should remove it, yes.

PR #2326, opened 2015-12-23
Make st2ctl to work with new Packages

armab:
Addresses https://github.com/StackStorm/st2-packages/issues/51

[`NG_INIT=true`](https://github.com/StackStorm/puppet-st2/blob/master/README.md#ng-init) env variable comes from `puppet-st2` which enables using init scripts for services.

### DIFF
--- a/tools/st2ctl
+++ b/tools/st2ctl
@@ -10,6 +10,11 @@ PYTHON=`which python2.7`
 # Does not happen consistently with all OSes we support.
 [ -r /etc/environment ] && . /etc/environment
 
+# In short, currently the best assumption for new packages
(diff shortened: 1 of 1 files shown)


armab:
As discussed here: https://github.com/StackStorm/st2-packages/pull/58#issuecomment-166980996

Will ship new version of `st2ctl` inside [`st2common/bin`](https://github.com/StackStorm/st2/tree/master/st2common/bin) instead of one from `tools`.

This will allow us to get rid of obsolete code like https://github.com/StackStorm/st2/blob/master/tools/st2ctl#L171-L317 and start new story for `st2ctl` (new packages are shipped with all included init services).

PR #2330, opened 2015-12-24
Fix Hubot restart command

saintaardvark:
In my ST2 installation (Ubuntu 14.04, using 'curl -sSL https://install.stackstorm.com/ | sudo sh'), the Hubot instance runs in a Docker container, and is controlled via /etc/init.d/docker-hubot.


emedvedev:
Since Docker is our default now, makes sense.

emedvedev:
@saintaardvark Thanks!

PR #2331, opened 2015-12-24
Fix st2common <-> st2actions circular dependency

Kami:
This pull request fixes st2actions <-> st2common circular dependency in the register content
script.

This means register content script now only depends on st2common and should work fine with the new packages.

P.S. We should figure out if we can create some automated tooling around that - sometimes it's easy to inadvertently create circular dependencies and because of the lazy imports this is not always detected early enough (I guess some quick grep checks could potentially do to  begin with) :).


Kami:
I added some very simple and dumb circular dependency checks here - https://github.com/StackStorm/st2/blob/a9872d5e4830deee5557d93d0062e6f913c32893/Makefile#L150

In case they cause too many false positives in the future we can improve and / or disable them.

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1450956483018976

dennybaa:
Awesome you've fixed it man! Thanks +1

Kami:
Thanks, merged :)

PR #2332, opened 2015-12-24
Skip virtualenv from top-level generated requirements.txt file

Kami:
#2322 was causing workroom to fail in a weird manner (I couldn't track down the root cause in a short time frame I spend debugging it).

For now, I just decided to not include `virtualenv` in top-level `requirements.txt`.

This won't affect the new packages packages since they use package / component specific `requirements.txt` file (and not the top-level one), but it will prevent workroom failures.

In the future, we should also dig in, track down the root cause and see why this is causing the workroom failure.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -212,7 +212,7 @@ requirements: virtualenv .sdist-requirements
 	$(VIRTUALENV_DIR)/bin/pip install --upgrade pip
 
 	# Generate all requirements to support current CI pipeline.
-	$(VIRTUALENV_DIR)/bin/python scripts/fixate-requirements.py -s st2*/in-requirements.txt -f fixed-requirements.txt -o requirements.txt
+	$(VIRTUALENV_DIR)/bin/python scripts/fixate-requirements.py --skip=virtualenv -s st2*/in-requirements.txt -f fixed-requirements.txt -o requirements.txt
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -212,7 +212,7 @@ requirements: virtualenv .sdist-requirements
 	$(VIRTUALENV_DIR)/bin/pip install --upgrade pip
 
 	# Generate all requirements to support current CI pipeline.
-	$(VIRTUALENV_DIR)/bin/python scripts/fixate-requirements.py -s st2*/in-requirements.txt -f fixed-requirements.txt -o requirements.txt
+	$(VIRTUALENV_DIR)/bin/python scripts/fixate-requirements.py --skip=virtualenv -s st2*/in-requirements.txt -f fixed-requirements.txt -o requirements.txt
```
dennybaa: haha. @Kami  elegant!

Kami:
It looks like workroom tests passed on re-run so the change was probably related to APT repos being offline and being unable to install virtualenv from there (although the error / exception totally didn't indicate that - we need to fail early...).

Kami:
I will still go ahead and merge this change though.

dennybaa:
+1 why not, anyway it's good to have such an option as `skip`

PR #2333, opened 2015-12-24
New st2ctl version used only by new Packages

armab:
> Addresses issue https://github.com/StackStorm/st2-packages/issues/51

New packages are shipped with their own [init scripts](https://github.com/StackStorm/st2-packages/tree/master/packages/st2api/debian), changed virtualenv, some PATHS and directories were changed that are incompatible with `st2ctl`.

Instead of [patching](https://github.com/StackStorm/st2/pull/2326) existing [`tools/st2ctl`](https://github.com/StackStorm/st2/blob/master/tools/st2ctl), another approach discussed here: https://github.com/StackStorm/st2/pull/2326#issuecomment-166984726 and here: https://github.com/StackStorm/st2-packages/pull/58#issuecomment-166980996 is to ship new `st2ctl` script under `st2common/bin/`, which will be used only by new packages.

It's a chance to refactor `st2ctl` and clean obsolete pieces.

There are even upholders to remove it completely with new packages.
But I think this helper script is useful at least to restart/reload everything in one command, also some core actions depends on `st2ctl`.

----

The intent is to keep it at least more minimal. If you think there are more things we can clean/wipe/remove - let me know. Anyway there is a chance to radically wipe it many times until we ship brand new packages to users.

Also I found previous change/discussion with `NG_INIT`, should be useful to mention it: https://github.com/StackStorm/st2/pull/1718

@StackStorm/team take a look (not urgent, I'll keep it open for several days here).


st2common/bin/st2ctl
```diff
@@ -0,0 +1,197 @@
+#!/usr/bin/env bash
+
+if [ $(id -u) -ne 0 ]; then
+    echo "Please run with root privileges"
```
armab: Addresses https://github.com/StackStorm/st2workroom/issues/261#issuecomment-166872541

st2common/bin/st2ctl
```diff
@@ -0,0 +1,197 @@
...
+  stop)
+    st2stop
+    ;;
+  restart)
+    st2stop
+    sleep 1
+    st2start
+    getpids
+    ;;
+  restart-component)
```
armab: Used: in [`packs.restart_component`](https://github.com/StackStorm/st2/blob/2f6465f366bc22c8adff79c031acd0a3b3837728/contrib/packs/actions/restart_component.yaml#L13)

st2common/bin/st2ctl
```diff
@@ -0,0 +1,197 @@
...
+
+function st2stop() {
+  for COM in ${COMPONENTS}; do
+    service ${COM} stop
+  done
+}
+
+# Next candidate for removal
+# after we implement native `service st2component reopen` (nginx example) for each component
+function reopen_component_log_files() {
```
armab: Used in [`logrotate.conf`](https://github.com/StackStorm/st2/blob/2f6465f366bc22c8adff79c031acd0a3b3837728/conf/logrotate.conf#L11)

See bash comments.

Kami:
We also need to decide if we are going to support upgrading from old packages to the new ones on the existing servers (so upgrading existing installations).

We kinda supported that with the new installer when we switched from init script to OS native service manager, but we didn't ship a "migration script" which kills and removed old instances running with init scripts so a lot of people ended up with stray running processes.

Kami:
Overall a lot better and more clean - old st2ctl is quite a mess which is hard to maintain :)

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452285259022046

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452286110022058

DoriftoShoes:
I would rather update the commands in the packs pack actions and logrotate to no longer use st2ctl and just completely deprecate it.  The only piece we need to add to the new packaging is an init script to start stop all services at once.

armab:
@DoriftoShoes At least I know @dennybaa will be +1 to completely deprecate `st2ctl` :)

armab:
My personal feeling about `st2ctl` - it's useful helper script in general.

Remember yourself in situation helping #community. How many times we suggested simple 2-words command:
```
st2ctl reload
```
and the new way will be: 
```sh
st2-register-content --config-file /etc/st2/st2.conf --register-actions --register-aliases --register-sensors
```
It's clear that previous one was more user-friendly.

Regarding `st2ctl restart`. How new init example will look like?
`service st2 restart`? What naming is preferred?

Also remember myself using `st2ctl clean` as well, it saved time.

----
Just ^^ info to keep in mind if you guys want to deprecate it completely. I'm ok with any decision, but don't see any harm in `st2ctl`, if we make it cleaner.

In case of deprecating completely - we should just bring all breaking changes like this in one release. I mean being in this state in several places:

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452536773022387

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452537492022410

lakshmi-kannan:
As per our slack convo, st2ctl should stay short term. So I am +1 to this PR being merged.

PR #2335, opened 2015-12-29
Fix explicit none to optional argument on action execution

m4dcoder:
JSON schema throws exception if None is passed explicitly to an optional parameter for an action execution.

### DIFF
--- a/st2common/st2common/util/schema/__init__.py
+++ b/st2common/st2common/util/schema/__init__.py
@@ -29,6 +29,13 @@ __all__ = [
     'get_action_parameters_schema',
     'get_schema_for_action_parameters',
     'get_schema_for_resource_parameters',
+    'is_property_type_single',
+    'is_property_type_list',
+    'is_property_type_anyof',
+    'is_property_type_oneof',
+    'is_property_nullable',
+    'is_attribute_type_array',
+    'is_attribute_type_object',
     'validate'
 ]
 
@@ -105,6 +112,47 @@ CustomValidator = create(
 )
 
 
+def is_property_type_single(property_schema):
+    return (isinstance(property_schema, dict) and
+            'anyOf' not in property_schema.keys() and
+            'oneOf' not in property_schema.keys() and
+            not isinstance(property_schema.get('type', 'string'), list))
+
+
+def is_property_type_list(property_schema):
+    return (isinstance(property_schema, dict) and
+            isinstance(property_schema.get('type', 'string'), list))
+
+
+def is_property_type_anyof(property_schema):
+    return isinstance(property_schema, dict) and 'anyOf' in property_schema.keys()
+
+
+def is_property_type_oneof(property_schema):
+    return isinstance(property_schema, dict) and 'oneOf' in property_schema.keys()
+
+
+def is_property_nullable(property_type_schema):
+    # For anyOf and oneOf, the property_schema is a list of types.
+    if isinstance(property_type_schema, list):
+        return len([t for t in property_type_schema
+                    if ((isinstance(t, six.string_types) and t == 'null') or
+                        (isinstance(t, dict) and t.get('type', 'string') == 'null'))]) > 0
+
+    return (isinstance(property_type_schema, dict) and
+            property_type_schema.get('type', 'string') == 'null')
+
+
+def is_attribute_type_array(attribute_type):
+    return (attribute_type == 'array' or
+            (isinstance(attribute_type, list) and 'array' in attribute_type))
+
+
+def is_attribute_type_object(attribute_type):
+    return (attribute_type == 'object' or
+            (isinstance(attribute_type, list) and 'object' in attribute_type))
+
+
 def assign_default_values(instance, schema):
     """
     Assign default values on the provided instance based on the schema default specification.
@@ -117,6 +165,7 @@ def assign_default_values(instance, schema):
         return instance
 
     properties = schema.get('properties', {})
+
     for property_name, property_data in six.iteritems(properties):
         has_default_value = 'default' in property_data
         default_value = property_data.get('default', None)
@@ -136,7 +185,8 @@ def assign_default_values(instance, schema):
         schema_items = property_data.get('items', {})
 
         # Array
-        if attribute_type == 'array' and schema_items and schema_items.get('properties', {}):
+        if (is_attribute_type_array(attribute_type) and
+                schema_items and schema_items.get('properties', {})):
             array_instance = instance.get(property_name, None)
             array_schema = schema['properties'][property_name]['items']
 
@@ -146,7 +196,7 @@ def assign_default_values(instance, schema):
                                                                 schema=array_schema)
 
         # Object
-        if attribute_type == 'object' and property_data.get('properties', {}):
+        if is_attribute_type_object(attribute_type) and property_data.get('properties', {}):
             object_instance = instance.get(property_name, None)
             object_schema = schema['properties'][property_name]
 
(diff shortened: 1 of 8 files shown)


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1451349562019634

st2common/tests/unit/services/test_action.py
```diff
@@ -115,6 +118,16 @@ def test_request_invalid_parameters(self):
...
 
+    def test_request_optional_parameter_none_value(self):
+        parameters = {'hosts': 'localhost', 'cmd': 'uname -a', 'a': None}
+        request = LiveActionDB(action=ACTION_REF, parameters=parameters)
+        request, _ = action_service.request(request)
+
+    def test_request_optional_parameter_none_value_no_default(self):
+        parameters = {'hosts': 'localhost', 'cmd': 'uname -a', 'b': None}
+        request = LiveActionDB(action=ACTION_REF, parameters=parameters)
+        request, _ = action_service.request(request)
```
lakshmi-kannan: This doesn't seem to test much. Are you just validating that request doesn't throw? Same question about prev test case.

lakshmi-kannan:
Minor things like Changelog and some documentation on allowing None and 'oneOf' in action meta parameter types would be nice. I didn't know that you could use oneOf and specify an array for types.

st2common/tests/unit/test_json_schema.py
```diff
@@ -47,6 +47,109 @@
...
+                {'type': 'string'},
+                {'type': 'boolean'}
+            ]
+        },
+        'arg_optional_no_default_oneof_none': {
+            'description': 'Foo',
+            'oneOf': [
+                {'type': 'string'},
+                {'type': 'boolean'},
+                {'type': 'null'}
```
lakshmi-kannan: Nice test cases

lakshmi-kannan:
Works for me. Please wait in @Kami to have a look as well.

st2common/tests/unit/services/test_action.py
m4dcoder: Just validating the parameters pass json schema validation. These tests fail before I patched the util so they do work.

m4dcoder:
Yep I will add line item to changelog.

st2common/st2common/util/schema/__init__.py
```diff
@@ -136,7 +137,9 @@ def assign_default_values(instance, schema):
         schema_items = property_data.get('items', {})
 
         # Array
-        if attribute_type == 'array' and schema_items and schema_items.get('properties', {}):
+        if (schema_items and schema_items.get('properties', {}) and
+                (attribute_type == 'array' or
+                    (isinstance(attribute_type, list) and 'array' in attribute_type))):
```
Kami: Good catch with `list` - we don't actually allow user to provide multiple types for action parameters (so it's never a list), but there are probably some other places we control (e.g. API models perhaps?) where the type can be a list, right?

st2common/st2common/util/schema/__init__.py
```diff
@@ -146,7 +149,9 @@ def assign_default_values(instance, schema):
                                                                 schema=array_schema)
 
         # Object
-        if attribute_type == 'object' and property_data.get('properties', {}):
+        if (property_data.get('properties', {}) and
```
Kami: Imo, it would make it easier to read it we store those in temporary variable and do something like:

```python
if has_properties and (is_object or is_list_and_has_object):
```

st2common/st2common/util/schema/__init__.py
```diff
@@ -164,33 +169,49 @@ def modify_schema_allow_default_none(schema):
...
-            # Allow "None" to be also used as a valid attribute value
-            property_type = schema['properties'][property_name]['type']
-
-            if isinstance(property_type, list) and 'null' not in property_type:
-                schema['properties'][property_name]['type'].append('null')
-            elif isinstance(property_type, six.string_types) and 'null' not in property_type:
-                schema['properties'][property_name]['type'] = [property_type, 'null']
+        if (has_default_value or is_optional) and default_value is None:
+            # If property is anyOf and oneOf then it has to be process differently.
+            if set(['anyOf', 'oneOf']) & set(schema['properties'][property_name].keys()):
```
Kami: This is a bit hard to read and follow - would it be possible to split in a separate method or similar?

Kami:
Nice work :+1: 

Overall the change looks good to me. Only thing which would be great to do is to refactor the code to make it easier to read and follow (some of the chunks could be split in standalone utility methods). 

Some of those methods were already on the larger side before this change and now they are even larger. Good news is that we have a pretty decent coverage for this method so it shouldn't be too painful to do it :)

st2common/st2common/util/schema/__init__.py
m4dcoder: The list needs to be supported for the inline schema change to allow None. For optional parameters where None is allowed, internally we create a list and put the original param type in a list along with null (i.e. ['string', 'null'], ['number', 'null'], ['array', 'null']).  This is the same with an param with type array or object. Since array and object are processed separately, we need to account for use cases where these params are optional and none is allowed.

Kami: Thought so - thanks for the clarification.

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452034559020170

Kami:
It looks like that one of the st2tests test started failing some time after this change:

```
|                 |     raise e                                                  |
|                 | ValidationError: True is not of type 'string', 'null'        |
|                 |                                                              |
|                 | Failed validating 'type' in schema['properties'][u'sudo']:   |
|                 |     {u'immutable': False, 'type': ['string', 'null']}        |
|                 |                                                              |
|                 | On instance[u'sudo']:                                        |
|                 |     True                                                     |
|                 | ",                                                           |
|                 |                                                     "error": |
|                 | "Task "test_stdout_local_script_action_sudo" failed: True is |
|                 | not of type 'string', 'null'                                 |
```

The test in question is https://github.com/StackStorm/st2tests/blob/master/packs/tests/actions/chains/test_quickstart_local_script_actions.yaml#L66

It looks like we pass `true` (boolean) for the sudo value, but the schema expects a string. Unless I am missing something, there is a bug somewhere in the schema validation (it's a local script action so `sudo` argument type should be boolean, not string).

Is it possible that there is a bug in schema validation code and we some how accidentally overwrite parameter type (or similar) when we dynamically manipulate the schema?

@m4dcoder would appreciate if you can have a look.

On a related note - this looks like something which should be easy to replicate in the unit tests, so we should also add a unit test for it (unless the bug only manifests itself under some conditions, depending on the other action parameters and not in isolation).

Kami:
Btw, thanks for refactoring the code :+1:

Kami:
I'm also debugging the issue mentioned above.

It looks like it's potentially related to parameter overwriting in the action.

This doesn't work:

```yaml
  sudo:
    immutable: false
    type: boolean
```

And this does:

```yaml
  sudo:
    immutable: false
```

So it seems like we don't correctly merge runner + action parameters when performing schema validation.

Kami:
I'm looking at the code, it seems like I found the issue (`get_schema_for_action_parameters` function):

```python
    parameters_schema = {}
    parameters_schema.update(runner_type.runner_parameters)
    parameters_schema.update(action_db.parameters)
```

`.update()` doesn't perform a merge, which is what we want and need in this case.

```python
a = {}
b = {'a': {'a': 1}}
c = {'a': {'b: 2}}

a.update(b)
a.update(c)

a
{'a': {'b': 2}}
```

This should have never worked correctly. I imagine it worked in the past because we haven't performed a strict schema type validation.

I'm looking into a fix now.

PR #2336, opened 2015-12-30
Story/mas/pxpse 422 create log file for pack actio

azamsheriff:
Completed the code refactoring to create separate log file with pack name to log their appropriate actions.

### DIFF
--- a/st2actions/conf/logging.conf
+++ b/st2actions/conf/logging.conf
@@ -21,13 +21,13 @@ args=(sys.stdout,)
 class=st2common.log.FormatNamedFileHandler
 level=INFO
 formatter=verboseConsoleFormatter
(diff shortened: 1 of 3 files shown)


Kami:
It looks like this PR was accidentally submitted to this repo instead of the fork.

I'm still curious what this change is for and if some of those changes are something which is more generally applicable and could also be useful to other users (we are happy to work with your on such changes and get them merged upstream).

Maintaining a diverging fork is sometimes necessary because of the time pressure and other requirements, but it can also be "expensive" so it should be in both of our interest (we of course also want to keep our users happy) to keep the diff between official repo and the fork as small as possible :)

It looks like you want a log file per Python runner action, is this correct? In addition to that, I'm not totally sure I understand the "pack name in the log file name" change. Do you only run single pack per single action runner instance or something else?

PR #2340, opened 2016-01-04
Updated the "Configure Logging" section

azamsheriff:
Updated the list of parameters passed in the example of  RotatingFileHandler args in the "Configure Logging" section of config.rst due to the args list had a additional empty parameter which will throw run time error.


Kami:
Good catch - thanks!

P.S. Here is just a link to the reference docs for anyone checking this issue later - https://docs.python.org/2/library/logging.handlers.html#rotatingfilehandler

Kami:
Merged, thanks!

PR #2341, opened 2016-01-04
Deprecate "params" action chain task attribute and replace it with new "parameters" attribute

### LINKED ISSUE #2337: Action metadata file uses "parameters" and action chain itself uses "params". This is really confusing!
https://stackstorm-community.slack.com/archives/community/p1451501940009873

Reported by: Fabio Brito on Slack.

Kami:
This pull request resolves #2337 and replaces `params` task attribute in the action chain definition file with the new `parameters` attribute - this way it's now consistent with the action metadata files.

For backward compatibility reasons, `params` is still supported, but we should remove support for it in a future major release.


Kami:
On a related note - we should probably create a special Github milestone or some kind of a tracking issue for deprecated features / code which should be removed in a future release.

We already accumulated some of those, but we have no central place to track them.

manasdk:
:+1:

st2actions/st2actions/runners/actionchainrunner.py
```diff
@@ -474,7 +477,7 @@ def _resolve_params(action_node, original_parameters, results, chain_vars, chain
         context.update({SYSTEM_KV_PREFIX: KeyValueLookup()})
         context.update({ACTION_CONTEXT_KV_PREFIX: chain_context})
         try:
-            rendered_params = jinja_utils.render_values(mapping=action_node.params,
+            rendered_params = jinja_utils.render_values(mapping=action_node.get_parameters(),
```
Kami: Another option would be to use `@property` / attribute getter in the `Node` class but this means we would need to change code a bit and prefix all the class variables with `_`.

st2common/st2common/models/system/actionchain.py
```diff
@@ -86,6 +92,24 @@ def __init__(self, **kw):
             prop = string.replace(prop, '-', '_')
             setattr(self, prop, value)
 
+    def validate(self):
```
Kami: I just noticed we don't actually validate Node parameters against the schema defined above (or am I missing something and we do validate it somewhere?).

If we indeed don't, we should also update this method to validate it here - I'm kind reluctant to do it in this PR thought since who knows what kind of issues it might expose.

st2actions/st2actions/runners/actionchainrunner.py
Kami: Here is an alternative approach using property getters - b8ee345af1b3243af4151d4ec9f9710cd4504b58

Advantage of this approach is that it allows us to consistently access `Node` class attributes. With the previous approach, we need to use getter for parameters (`node.get_parameters()`) and directly access instance variable for others (`node.name`).

So, dunno. What do others think?

Kami:
Another thing to note - we can't really update all the action chain definitions in `st2contrib` until we assume that all the users run latest version with this change.

st2actions/st2actions/runners/actionchainrunner.py
manasdk: Looks good. I like property accessors :)

manasdk:
+1

lakshmi-kannan:
There are some inline examples in docs/actionchain.rst which still use `params`. Can you fix those while with this? 

LGTM otherwise.

Kami:
> There are some inline examples in docs/actionchain.rst which still use params. Can you fix those while with this?

`<rant>`We should have no examples directly in the docs, all of the examples should be included files which are tested`</rant>`

Good catch, will update it.

PR #2346, opened 2016-01-04
Add wait-before to mistral example to workaround known bug

m4dcoder:
Add wait-before to one of the tasks in mistral-workbook-complex to workaround a mistral bug until it is fixed (https://bugs.launchpad.net/mistral/+bug/1530982).


Kami:
:+1:

Issue #2344, opened 2016-01-04
Action chain should allow one to print all variables that can be used in jinja templating. 


lakshmi-kannan:
This is required when we are debugging things. 

Reported by: lakstorm on slack.

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1451939867019775

Kami:
Yeah, I mentioned this multiple times already - we should also automatically generate a table or similar and include it in the docs which lists all the variable which are available in each context (triggered via ChatOps, via API, etc.).

This would of course only include "static" variables which are known to be available during run time. For the actual "live print" of all the available variables at that time, we need some more code changes, yeah.

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452026720020048

PR #2347, opened 2016-01-04
Publish action chain vars

manasdk:
* Action chain vars are published in the result of an action chain under the `published` property.
* Also, added support for a special `__context` property that can be used in ActionChain, Rules, AliasExecution and Notification sections. This property will help with debugging.

### Sample ActionChain
```
---
    chain:
        -
            name: "c1"
            ref: "core.local"
            params:
                cmd: "date"
            publish:
                date: "{{c1.stdout}}"
            on-success: "c2"
            on-failure: "c4"
        -
            name: "c2"
            ref: "core.local"
            params:
                cmd: "echo \"c2: parent exec is {{action_context.parent.execution_id}}.\""
            publish:
                context: "{{__context}}"
            on-success: "c3"
            on-failure: "c4"
        -
            name: "c3"
            ref: "core.local"
            params:
                cmd: "echo c3"
            on-failure: "c4"
        -
            name: "c4"
            ref: "core.local"
            params:
                cmd: "echo fail c4"
    default: "c1"
```
### Results of an execution
```
"result": {
        "published": {
            "context": "{'date': u'Mon Jan  4 16:06:59 PST 2016', '__results': {'c2': {u'failed': False, u'stderr': u'', u'return_code': 0, u'succeeded': True, u'stdout': u'c2: parent exec is 568b092332ed354a1d2d19ea.'}, 'c1': {u'failed': False, u'stderr': u'', u'return_code': 0, u'succeeded': True, u'stdout': u'Mon Jan  4 16:06:59 PST 2016'}}, u'i1': u'hi', 'c2': {u'failed': False, u'stderr': u'', u'return_code': 0, u'succeeded': True, u'stdout': u'c2: parent exec is 568b092332ed354a1d2d19ea.'}, 'c1': {u'failed': False, u'stderr': u'', u'return_code': 0, u'succeeded': True, u'stdout': u'Mon Jan  4 16:06:59 PST 2016'}, 'system': <st2common.services.keyvalues.KeyValueLookup object at 0x7f6bfc2c0b90>}",
            "date": "Mon Jan  4 16:06:59 PST 2016"
        },
        "tasks": [
           ...
        ]
    }
```

Fixes https://github.com/StackStorm/st2/issues/2344

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,8 +33,8 @@ in development
 * Add new ``st2garbagecollector`` service which periodically deletes old data from the database
   as configured in the config. By default, no old data is deleted unless explicitly configured in
   the config.
-* All published variables are now available in the result of ActionChain execution under the
-  ``published`` property.
+* All published variables can be available in the result of ActionChain execution under the
+  ``published`` property if ``display_published`` property is specified.
 
 1.2.0 - December 07, 2015
 -------------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,6 +33,8 @@ in development
 * Add new ``st2garbagecollector`` service which periodically deletes old data from the database
   as configured in the config. By default, no old data is deleted unless explicitly configured in
   the config.
+* All published variables are now available in the result of ActionChain execution under the
+  ``published`` property.
 
 1.2.0 - December 07, 2015
 -------------------------
--- a/docs/source/actionchain.rst
+++ b/docs/source/actionchain.rst
@@ -98,7 +98,7 @@ ActionChain offers the convinience of named variables. Global vars are set at th
 Tasks publish new variables with the ``publish`` keyword. Variables are handy when you need to mash up
 a reusable value from the input, globals, DataStore values, and results of multiple actions executions.
 All variables are referred with Jinja syntax. The cumulative published variables are also available in the result of an
-ActionChain execution under the ``published`` property.
+ActionChain execution under the ``published`` property is ``display_published`` property was supplied.
 
 .. code-block:: yaml
 
--- a/docs/source/actionchain.rst
+++ b/docs/source/actionchain.rst
@@ -95,9 +95,10 @@ Variables
 ~~~~~~~~~
 
 ActionChain offers the convinience of named variables. Global vars are set at the top of the definition with the ``var`` keyword.
-Tasks publish new variables with the ``publish`` keyword. Variables handy when you need to mash up
+Tasks publish new variables with the ``publish`` keyword. Variables are handy when you need to mash up
 a reusable value from the input, globals, DataStore values, and results of multiple actions executions.
-All variables are referred with Jinja syntax.
+All variables are referred with Jinja syntax. The cumulative published variables are also available in the result of an
+ActionChain execution under the ``published`` property.
 
 .. code-block:: yaml
 
--- a/st2actions/st2actions/runners/actionchainrunner.py
+++ b/st2actions/st2actions/runners/actionchainrunner.py
@@ -223,6 +223,7 @@ class ActionChainRunner(ActionRunner):
         self._meta_loader = MetaLoader()
         self._stopped = False
         self._skip_notify_tasks = []
+        self._display_published = False
         self._chain_notify = None
 
     def pre_run(self):
@@ -258,6 +259,7 @@ class ActionChainRunner(ActionRunner):
             self._chain_notify = getattr(self.liveaction, 'notify', None)
         if self.runner_parameters:
             self._skip_notify_tasks = self.runner_parameters.get('skip_notify', [])
+            self._display_published = self.runner_parameters.get('display_published', False)
 
         # Perform some pre-run chain validation
         try:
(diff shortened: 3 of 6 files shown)


dzimine:
nice

CHANGELOG.rst
```diff
@@ -33,6 +33,8 @@ in development
 * Add new ``st2garbagecollector`` service which periodically deletes old data from the database
   as configured in the config. By default, no old data is deleted unless explicitly configured in
   the config.
+* All published variables are now available in the result of ActionChain execution under the
```
Kami: It's worth noting that this slightly "breaks" the "unified" action chain and mistral result format.

IIRC, in Mistral result object we added Mistral specific attributes under `extra` attribute, maybe we should do a similar thing here and keep it consistent.

Kami:
Overall, I like the idea, but I wonder if it should be enabled by default. A runner parameter perhaps?

Many times you pass temporary variables around which are only used by some tasks, but not actually returned in the end result. Some of those variables can contain a lot of data and returning all this data in the result object could potentially cause issues and break things.

manasdk:
>  but I wonder if it should be enabled by default. A runner parameter perhaps?

I personally do not want to have the response controllable by a runtime property. It would be nice if we could simply make this consistent - can compare to mistral and see what happens in that case.

Kami:
On a related note - if we enable it by default it would also be a good idea to test and run it on our build server for a while.

On a related note, `?exclude_attributes=result.published` would potentially also be useful.

manasdk:
> if we enable it by default it would also be a good idea to test and run it on our build server for a while.

This is really an ask by @lakshmi-kannan  - I am mostly meh to this feature. Can put it behind a feature flag or as you pointed previously make it controlled by a runner parameter. I do feel that publishing the computed variables is a good idea but if you suspect that there will be issue I am happy to put up some guard-rails  (not sure what :P)

Kami:
Yeah, I'm fine if we leave it disabled by default and make it a runner parameter :)

dzimine:
I am +1 for adding published vars to results [by default, or always]; user has control on what to publish and can/should form it so that it doesn't pollute the result.

In mistral we went the other way around offering an option for the task to suppress keeping task results in the context. User does not always have control of what the action returns, and if the action returns tons of data, user can pick the right piece of data, publish it, and supress keeping action output in the context (so no pollution).

This is a case when feature will shape the use: once published variables are part of result, they will likely be used to form the "real, useful" part of result. To be seen.

Kami:
> I am +1 for adding published vars to results [by default, or always]; user has control on what to publish and can/should form it so that it doesn't pollute the result.

Yes, user has control over it, but in a lot of cases, publish is only used to make "temporary" aliases for variables so they can be more easily accessed in tasks further down the chain. Imo, publishing a variable so it's available in the context of the next task(s) is different than returning a result / publishing a variable from the last task in the chain.

Another reason why I'm also not a big fan of it is that it's not really explicit. I think of it as "temporary  variable assignment" in the code which is different then "returning a result" from the function.

Lack of explicitness can be somewhat be offset by good documentation, but I still always strive to keep things explicit when possible (makes it easier to work with, debug, understand, etc.).

In any case, I won't block, but at the very least I just want to make sure stuff won't break with tasks which publish temporary variables which contain a lot of data (weekly github report action chain workflow is a good example of that - temporary published variables contain tons of data and even if the things won't break, included so much data which is not relevant to the result will clutter it).

enykeev:
Have we investigated the possibility of disabling `tasks` by default too? I'm missing a lot of context at the moment, but I'm assuming this is our end goal for the change.

manasdk:
> Have we investigated the possibility of disabling tasks by default too? I'm missing a lot of context at the moment, but I'm assuming this is our end goal for the change.

Not end goal for this change. The end goal is to make @lakshmi-kannan happy. See https://github.com/StackStorm/st2/issues/2344

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452019094019933

lakshmi-kannan:
>Another reason why I'm also not a big fan of it is that it's not really explicit. I think of it as "temporary variable assignment" in the code which is different then "returning a result" from the function.

I agree but this PR doesn't address that. Having a runner parameter to enable this or not is indeed a good idea and seems straight forward to add. I am not sure if this immediately would be beneficial. Lot of people on community have issues with this jinja templating part. So maybe we just publish everything for a bit and see.

dzimine:
@Kami good points re meaning of variables vs output however proper definition of actionchain output vs local variables is a matter of ActionChain v2; let's not bloat this one but note and apply when we do v2. 

I am stressing that lack of published variables in action chain result has been a confusion already, and this is a quick way to address this confusion. So +1 for the change as is.

OK with a runner parameter to control __context, but as for published variables, it should be clear one behavior, and again, I say, "part of results".

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452024546019985

manasdk:
> OK with a runner parameter to control __context

This is super secret and internal. Lets not include it in the discussion here ;)

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452027288020084

Kami:
:+1:

PR #2349, opened 2016-01-05
Mark execution as failed on ParamExecption

manasdk:
* Previously the execution would be left in a requested state. This way the
  execution is put in a failed state and the reasoning is provided.

### DIFF
--- a/st2api/st2api/controllers/v1/actionexecutions.py
+++ b/st2api/st2api/controllers/v1/actionexecutions.py
@@ -156,7 +156,6 @@ class ActionExecutionsControllerMixin(BaseRestControllerMixin):
                 runnertype_db.runner_parameters, action_db.parameters, liveaction_db.parameters,
                 liveaction_db.context)
         except ParamException:
-            LOG.info('ParamException')
             # By this point the execution is already in the DB therefore need to mark it failed.
             _, e, tb = sys.exc_info()
             action_service.update_status(
--- a/st2api/st2api/controllers/v1/actionexecutions.py
+++ b/st2api/st2api/controllers/v1/actionexecutions.py
@@ -16,6 +16,8 @@
 import copy
 import re
 import httplib
+import sys
+import traceback
 
 import jsonschema
 from oslo_config import cfg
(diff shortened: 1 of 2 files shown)


Kami:
Good catch :+1:

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -153,7 +155,16 @@ def _schedule_execution(self, liveaction):
...
                 liveaction_db.context)
-        except ParamException as e:
+        except ParamException:
+            LOG.info('ParamException')
+            # By this point the execution is already in the DB therefore need to mark it failed.
+            _, e, tb = sys.exc_info()
+            action_service.update_status(
+                liveaction=liveaction_db,
+                new_status=LIVEACTION_STATUS_FAILED,
+                result={'error': str(e), 'traceback': ''.join(traceback.format_tb(tb, 20))})
```
Kami: Potential future code improvement - it would be nice if we had a single place / function where the result object for error scenarios is constructed.

If it's constructed in multiple places it's hard to keep it in sync.

Kami:
Changelog entry would be nice :)

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -153,7 +155,16 @@ def _schedule_execution(self, liveaction):
             liveaction_db.parameters = param_utils.render_live_params(
                 runnertype_db.runner_parameters, action_db.parameters, liveaction_db.parameters,
                 liveaction_db.context)
-        except ParamException as e:
+        except ParamException:
+            LOG.info('ParamException')
```
Kami: Hm, this log message is not really useful (and it doesn't include traceback or anything).

Should probably be LOG.warn with a better error message and traceback?

manasdk: Oh, that was for debugging. I will clean up the log line.

manasdk:
> Changelog entry would be nice :)

I considered but decided against since this is a regression in the new approach introduced by @enykeev  and we have not yet "shipped" this behavior. I maybe do not understand how CHANGELOG works - does every code change get a CHANGELOG entry? If so how is it really different from looking at the commit log or closed PRs?

lakshmi-kannan:
> I maybe do not understand how CHANGELOG works - does every code change get a CHANGELOG entry? If so how is it really different from looking at the commit log or closed PRs?

In an ideal world, our commits are clean (only merge commits to master) and we rebase to single commit in feature/bug fixes branches. As long as we have "lint me baby one more time" commits, it looks like CHANGELOG is needed. In this particular case, I find it silly to add a CHANGELOG for the reasons you pointed out. You are doing good.

st2api/st2api/controllers/v1/actionexecutions.py
manasdk: future :dancers:

PR #2351, opened 2016-01-05
Cherry-pick doc updates to v1.2.1

dzimine:


dzimine:
Manually pushed to stage, check here before I push to docs.stackstorm.com 

http://docs-staging.uswest2.stackstorm.net/1.2/
http://docs-staging.uswest2.stackstorm.net/

PR #2352, opened 2016-01-05
Fix missing 1.2 version in docs selector

dzimine:
Quick fix for missed 1.2 version. Need a better fix.


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1451964087019873

manasdk:
+1 

FYI, I hacked in a fix on S3 and therefore this actually work but forgot to update actual code.

Thanks!

PR #2354, opened 2016-01-05
Allow user to specify TTL via "--ttl" option when using "st2 key set" command

### LINKED ISSUE #2343: st2 key set in CLI doesn't allow one to set a TTL.
Reported by: psychopenguin on slack

Kami:
This pull request allows users to specify datastore item expiration (TTL) via ``--ttl`` option when using ``st2 key set`` command (e.g. ``st2 key set key1 value1 --ttl=60``).

It also updates ``key get`` and ``key set`` command to display ``expire_timestamp`` attribute.

This resolves #2343

### DIFF
--- a/st2client/st2client/commands/keyvalue.py
+++ b/st2client/st2client/commands/keyvalue.py
@@ -82,11 +82,11 @@ class KeyValuePairListCommand(resource.ResourceListCommand):
 
 class KeyValuePairGetCommand(resource.ResourceGetCommand):
     pk_argument_name = 'name'
-    display_attributes = ['name', 'value']
+    display_attributes = ['name', 'value', 'expire_timestamp']
 
 
 class KeyValuePairSetCommand(resource.ResourceCommand):
-    display_attributes = ['name', 'value']
+    display_attributes = ['name', 'value', 'expire_timestamp']
 
     def __init__(self, resource, *args, **kwargs):
         super(KeyValuePairSetCommand, self).__init__(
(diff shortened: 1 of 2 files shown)


st2client/st2client/commands/keyvalue.py
```diff
@@ -99,6 +99,8 @@ def __init__(self, resource, *args, **kwargs):
                                  metavar='name',
                                  help='Name of the key value pair.')
         self.parser.add_argument('value', help='Value paired with the key.')
+        self.parser.add_argument('--ttl', type=int, default=None,
+                                 help='TTL for this value.')
```
manasdk: Can you specify unit of TTL in help? I assume it is seconds

manasdk:
+1

st2client/st2client/commands/keyvalue.py
lakshmi-kannan: Something to think about is the consistency with st2 token create --ttl. I think that supports `-l`. Please double check and make them consistent.

lakshmi-kannan:
Definitely needs documentation. LGTM otherwise.

st2client/st2client/commands/keyvalue.py
Kami: > Can you specify unit of TTL in help? I assume it is seconds

Sure thing.

Kami: > Something to think about is the consistency with st2 token create --ttl. I think that supports -l. Please double check and make them consistent.

Added `-l` ttl option "alias" for consistency.

PR #2359, opened 2016-01-07
Fix overrides of runner parameter attributes

m4dcoder:
Validate that only a selected set of attributes can be overriden for any runner parameters. It does't make sense to let an action parameter override the type in the runner parameter. Also, this will fix a bug where a partial dict in the action parameter will completely override the runner parameter. In the case where type is not defined in the action parameter, the runner parameter will no longer have a type after the override.

### DIFF
--- a/st2api/st2api/controllers/v1/actions.py
+++ b/st2api/st2api/controllers/v1/actions.py
@@ -28,6 +28,7 @@ from st2api.controllers import resource
 from st2api.controllers.v1.actionviews import ActionViewsController
 from st2common import log as logging
 from st2common.constants.triggers import ACTION_FILE_WRITTEN_TRIGGER
+from st2common.exceptions.action import InvalidActionParameterException
 from st2common.exceptions.apivalidation import ValueValidationException
 from st2common.models.api.base import jsexpose
 from st2common.persistence.action import Action
@@ -94,9 +95,16 @@ class ActionsController(resource.ContentPackResourceController):
             Handles requests:
                 POST /actions/
         """
-        # Perform validation
-        validate_not_part_of_system_pack(action)
-        action_validator.validate_action(action)
+
+        try:
+            # Perform validation
+            validate_not_part_of_system_pack(action)
+            action_validator.validate_action(action)
+        except (ValidationError, ValueError,
+                ValueValidationException, InvalidActionParameterException) as e:
+            LOG.exception('Unable to create action data=%s', action)
+            abort(http_client.BAD_REQUEST, str(e))
+            return
 
         # Write pack data files to disk (if any are provided)
         data_files = getattr(action, 'data_files', [])
--- a/st2common/st2common/validators/api/action.py
+++ b/st2common/st2common/validators/api/action.py
@@ -15,10 +15,12 @@
 
 import six
 
+from st2common.exceptions.action import InvalidActionParameterException
 from st2common.exceptions.apivalidation import ValueValidationException
 from st2common.exceptions.db import StackStormDBObjectNotFoundError
 from st2common import log as logging
 from st2common.util.action_db import get_runnertype_by_name
+from st2common.util import schema as util_schema
 from st2common.content.utils import get_packs_base_paths
 from st2common.content.utils import check_pack_content_directory_exists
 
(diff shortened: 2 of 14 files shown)


Kami:
I think I'm missing a bit of context. It would be great if you could update PR description with some more information (and it's also good for other non-stormers).

Which parameters were causing a problem and are not supposed to be overwritten (`position`, anything else)?

st2common/st2common/validators/api/action.py
```diff
@@ -58,18 +60,27 @@ def _is_valid_pack(pack):
 
 
 def _validate_parameters(action_params=None, runner_params=None):
-    for param, action_param_meta in six.iteritems(action_params):
+    for action_param, action_param_meta in six.iteritems(action_params):
+        # Check if overridden runner parameters are permitted.
+        if action_param in runner_params:
+            for action_param_attr, value in six.iteritems(action_param_meta):
```
Kami: It looks like similar code is duplicated in two functions :)

It would be great if you could refactor it in a separate function (e.g. `is_parameter_overridable` or similar). In addition to that, it would make the code a bit easier to read / follow.

Kami:
Re backward incompatibility of this change (from slack) - for the actions which override parameters which are not supposed be overridden with the same values - is it possible for us to detect that (overridden value is the same as the parent one) and not throw (I know it might require us to refactor a bunch of code and pass a lot of state around so it might not be possible, at least with reasonable effort)?

This would allow us to make it mostly backward compatible and we can drop that code in 1.4 (or in the next major release).

m4dcoder:
@Kami: the code here is looking to see if the value differ for unpermitted overrides before throwing an exception.

m4dcoder:
@Kami: I updated commit message.

st2api/st2api/controllers/v1/actions.py
```diff
@@ -94,9 +95,16 @@ def post(self, action):
...
+
+        try:
+            # Perform validation
+            validate_not_part_of_system_pack(action)
+            action_validator.validate_action(action)
+        except (ValidationError, ValueError,
+                ValueValidationException, InvalidActionParameterException) as e:
+            LOG.exception('Unable to create action data=%s', action)
+            abort(http_client.BAD_REQUEST, str(e))
+            return
```
manasdk: unnecessary

manasdk:
I am ok with the change. Atleast the expected behavior is coded up deterministically now. 

I suspect there will be some regressions as there always are when basic things like these change. We will have to watch out and squash those asap.

st2api/st2api/controllers/v1/actions.py
m4dcoder: If I don't do this, then API will come back with 500 instead of bad request with the actual fault message.

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452198230021316

st2api/st2api/controllers/v1/actions.py
manasdk: I could've sworn that abort broke out immediately. I could be wrong.

armab:
This breaks `st2-register-content` on actions, see build:
https://circleci.com/gh/StackStorm/st2/296

jat0dwk 1

DoriftoShoes:
That's correct behavior.  Whatever action it fails on shouldn't override type on a runner param.

m4dcoder:
https://github.com/StackStorm/st2/pull/2362 is going to fix the st2-register-content error (basically fixing the override for the headers parameter in the http-request runner).

Kami:
@m4dcoder Thanks for updating the description and clarifying things.

#2361 is not a duplicate of this issue since it fixes a different problem - the problem in that case it's not type overwriting, but that we don't perform a correct deep runner and actions parameters merge which will cause validation to fail since if the action parameter which is overriding runner one doesn't explicitly provide a type, it will assume to a default type of string.

Kami:
Having said that, it looks like #2361 fixes the real underlying problem so the question is do we even need this PR?

It seems like it introduces some potential braking changes which could cause issues.

Kami:
Here is my monologue from Slack:

```
kami [08:35] 
@m4dcoder: https://github.com/StackStorm/st2/pull/2359#issuecomment-169917841 sorry for the late feedback, but I didn't fully understand the underlying issue at first. If I do understand it correctly now (maybe I still dont), it seems to me that this PR is not really needed and we only really need just https://github.com/StackStorm/st2/pull/2361

​[08:35] 
or am I missing something?

​[08:36] 
since yeah, your PR does introduce some more strict parameter overwrite handling - previously we didn't have any rules for that so everything worked, but if I understand it correctly the actual problem and bug lives elsewhere and that's what my PR is fixing

kami [08:37] 
and I think it would be good to figure out if i'm correct and that's indeed the case because your PR is introducing some potentially breaking changes which I dont think we need when we merge #2361

[08:40] 
checked your PR again, so it looks like part of our changes are the same - the deep merging issue, so yeah, that's the actual fix. but i dont think we need the "only allow to override those attributes" and this is a breaking change
```

I think it would be good to come to the bottom of this and figure out if we really do need the "only allow to override those parameters" part which is a breaking change.

In general I'm all for explicit rules and strict validation (previously we had no rules and validation on that so everything was allowed), but this could potentially cause a lot of issues to the users (we also don't allow overriding `enum` attribute now and some other attributes and this will likely cause issues to the users).

And if we decided that we indeed want this change we also need to explicitly document those override rules.

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452274858021695

m4dcoder:
@Kami: IMO, deep merge will fix the surface problem. but the deeper problem here is that action shouldn't be overriding certain attributes (type) of the runner parameters. going back to fix some of the packs is necessary as they were wrong to being with (i.e. http-request runner has the headers param type WRONG and then the core.http action corrected it.  that's just silly.)

Kami:
@m4dcoder I don't think it's a surface problem - it's the actual underlying bug.

As far as what is correct and what is wrong is to be decided since before those changes, we had no rules about that (well, it was decided in the PR which was merged).

In general, I do think it's a good idea to not allow users to overwrite "position" and "type" but besides that, I think everything else should be fine (we also don't allow overriding enum right now).

PR #2360, opened 2016-01-07
Document how to send auth token / API key in the query param

### LINKED ISSUE #2328: Webhook should support basic auth
In some cases, a user will need to expose StackStorm to the public world. Currently, we support token authorization, but not all services will allow header customization. The best available is just the URI. (for example: SNS). It would be fantastic if the same API key could ​*also*​ be used in basic-auth with the webhook sensor.

This was reported by @tt in #stackstorm-community

Kami:
This pull request documents previously undocumented feature which was only used by st2web.

We decide to document it, because there are more scenarios where you can't specify auth token / API key in the headers (e.g. third party services which don't allow you to send custom headers) and where this approach comes handy.

Resolves #2328.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -39,7 +39,7 @@ in development
   (improvement)
 * Fix validation error when None is passed explicitly to an optional argument on action
   execution. (bug fix)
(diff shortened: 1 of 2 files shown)


manasdk:
Oh no! Now I need to migrate this to st2docs :(

PR #2361, opened 2016-01-07
Fix action parameter validation when action overrides a parameter

Kami:
This pull request fixes a bug described here https://github.com/StackStorm/st2/pull/2335#issuecomment-169615849.

The problem was that we didn't perform a deep merge of runner + action parameters which means that if user would override runner parameter in an action and only specify a single parameter attribute (e.g. only immutable, but not type, etc.) things would explode.

Schema validation code changes didn't touch this part so probably the only reason it worked in the past is because we didn't actually perform full strict validation of the runner parameter types (lovely, eh? :P)

TODO:

- [x] Tests


st2common/st2common/util/misc.py
```diff
@@ -92,3 +93,18 @@ def rstrip_last_char(input_str, char_to_strip):
         return input_str[:-len(char_to_strip)]
 
     return input_str
+
+
+def deep_update(d, u):
```
Kami: I was looking for an existing library, but the one I found on PyPi actually breaks on some edge cases so I just decided to use a simple function.

m4dcoder:
@Kami: I think this is duplicate of https://github.com/StackStorm/st2/pull/2359

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452194722021258

Kami:
@m4dcoder I think those are two different issues, at least that's what I thought when I first saw your PR.

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452202169021366

Kami:
It looks like I can't really resolve conflicts and merge this PR until we decide on this - https://github.com/StackStorm/st2/pull/2359#issuecomment-169921028 so I will wait on that.

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452245563021539

Kami:
Alright, I resolved the merge conflicts as discussed with @m4dcoder.

I will go ahead, merge this PR into master and open a new PR for auto-generated documentation for parameter attributes which can be overriden.

PR #2362, opened 2016-01-07
Fix type for the headers param in the http-request runner

m4dcoder:
The type for the headers param should be object.


manasdk:
+1

PR #2363, opened 2016-01-08
Minor fixes to validation of runner parameter override

m4dcoder:


st2common/st2common/util/schema/__init__.py
```diff
@@ -320,11 +329,9 @@ def get_schema_for_action_parameters(action_db):
             parameters_schema.update({name: schema})
         else:
             for attribute, value in six.iteritems(schema):
-                if (attribute not in RUNNER_PARAM_OVERRIDABLE_ATTRS and
-                        parameters_schema[name].get(attribute) != value):
-                    raise InvalidActionParameterException(
-                        'The attribute "%s" for the runner parameter "%s" cannot'
-                        'be overridden.' % (attribute, name))
+                validate_runner_parameter_attribute_override(
```
Kami: In the future please prefer to use keyword argument when calling functions.

While working on resolving merge conflicts in my PR, I believe I encounter a minor issue - function order is `runner_param_attr_value, action_param_attr_value), but the values here are passed to the function in the reverse order.

Luckily, in this case the order doesn't matter, because those values are only used in a simple `!=` if condition, but it could matter in other scenarios.

PR #2364, opened 2016-01-08
Fix parameter override in actions of linux pack

m4dcoder:
The position attribute cannot be overridden for the cmd parameter in the run-remote runner.

### DIFF
--- a/contrib/linux/actions/rsync.yaml
+++ b/contrib/linux/actions/rsync.yaml
@@ -9,21 +9,17 @@
             type: 'string'
             description: 'List of files/directories to to be copied'
             required: true
-            position: 1
         dest_server:
             type: 'string'
             description: "Destination server for rsync'd files"
             required: true
-            position: 2
         destination:
             type: 'string'
             description: 'Destination of files/directories on target server'
             required: true
-            position: 3
         cmd:
             immutable: true
             default: 'rsync {{args}} {{source}} {{dest_server}}:{{destination}}'
-            position: 0
         connect_timeout:
             type: 'integer'
             description: 'SSH connect timeout in seconds'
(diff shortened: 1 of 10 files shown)


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452217922021466

contrib/linux/actions/rsync.yaml
```diff
@@ -9,21 +9,20 @@
             type: 'string'
             description: 'List of files/directories to to be copied'
             required: true
-            position: 1
+            position: 0
```
DoriftoShoes: The positions here should probably just be removed all together.  They get used in the cmd default anyway...seems odd that this even works.  I would think all that data would get passed through twice.

manasdk: I think the runner ignores all the other values so it does not really matter much.

m4dcoder: Is the intent to pass the parameters to the st2 action like this?  `linux.rsync <source> <dest_server> <destination>` instead of `linux.rsync source=<source> dest_server=<dest_server> destination=<destination>`?

DoriftoShoes: The intent is that the parameters are passed to the `cmd` parameter which works as a template for the positions of the parameters.  I rarely need this anymore but this was a common pattern early on for giving user friendly parameters in the meta and then building the proper linux command in the `cmd` default.

DoriftoShoes: I think the positions should just be removed.

m4dcoder: Removed!

PR #2365, opened 2016-01-08
Remove docs

manasdk:
Docs now live in https://github.com/StackStorm/st2docs therefore removing from this repo


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452224203021495

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452303203022204

Kami:
Looks like I lost this battle.

Do we still build docs on Travis? If we don't, we should. And ideally, we would also do this in this repo - we can add a script which downloads st2docs repo, etc. and builds it.

It's not 100% ideal because docs live in a different repo, but a lot of documentation itself and automatically generated files depend on the code from this repo so it's a good idea to do it - this way it's also clear that docs are tied to this repo and people making the changes should also make corresponding documentation changes.

manasdk:
> Looks like I lost this battle.

It would seem so but not exactly true. My approach to sort this particular problem was to come up with a solution that I could possibly achieve in a short enough time. I didn't see any credible option apart from separating docs so I went ahead an got it done. The converse is if there is an alternate solution we can conjure that could work for us and is understandably better I am happy to even get that done and/or enable others to achieve - so in effect there is no real battle on my part just a means to an end.

> Do we still build docs on Travis?

No.

>  If we don't, we should.

Absolutely. Need to add that asap.

> And ideally, we would also do this in this repo

Questionable. Can of course be done. I would rather have it all hooked off st2docs since other st2 repos like st2web etc also should have similar logical dependencies.

> people making the changes should also make corresponding documentation changes.

captain obvious :P

Kami:
Imo, we should build it in both repos - st2docs and here.

The reason why I think we should build it here is to let contributors know that documentation is integral part of the project and building the documentation in this repo as well, makes it harder to miss documentation changes.

On a related not, we also need to figure out our linting story - we need to add all the configs (and Makefile targets) to a shared repo and re-use this in all of our repos which contain Python code, including docs (there are some scripts with Python code).

I can take this task, because I think consistency, etc. is important.

Kami:
On a related note of decreasing barrier to contribution - we should build documentation for every branch and publish built documentation to a temporary S3 / ReadTheDocs bucket.

Installing all the dependencies needed to build the documentation locally takes ages and immensely increases barrier to contribution.

PR #2366, opened 2016-01-08
Fix runner param override

m4dcoder:
Add enum to the list of overridable attributes and update documentation.


Kami:
As far as documentation goes, I think it would be a good idea to add a script which generates a table with which parameters can be overrided based on the value of `RUNNER_PARAM_OVERRIDABLE_ATTRS` and include this file in the docs.

This way code and docs won't get out of sync (we already do this in other places).

I"m also happy to do this change myself.

Kami:
The change for auto-generating this list is here - https://github.com/StackStorm/st2docs/pull/11

PR #2367, opened 2016-01-08
Packs pack no longer uses st2ctl

DoriftoShoes:
This PR is to make the packs pack compatible with the new packaging

### DIFF
--- a/contrib/packs/actions/load.yaml
+++ b/contrib/packs/actions/load.yaml
@@ -10,7 +10,7 @@
       default: true
     cmd:
       immutable: true
-      default: "st2ctl reload{% for item in register.split(',') %} --register-{{ item|trim }}{% endfor %}"
+      default: "st2-register-content{% for item in register.split(',') %} --register-{{ item|trim }}{% endfor %}"
     register:
       type: "string"
       default: "actions,aliases,sensors"
--- a/contrib/packs/actions/load.yaml
+++ b/contrib/packs/actions/load.yaml
@@ -10,7 +10,7 @@
       default: true
     cmd:
       immutable: true
-      default: "st2-register-content{% for item in register.split(',') %} --register-{{ item|trim }}{% endfor %}"
+      default: "st2-register-content --config-file /etc/st2/st2.conf{% for item in register.split(',') %} --register-{{ item|trim }}{% endfor %}"
     register:
       type: "string"
       default: "actions,aliases,sensors"
--- a/contrib/packs/actions/load.yaml
+++ b/contrib/packs/actions/load.yaml
@@ -10,10 +10,14 @@
       default: true
     cmd:
       immutable: true
-      default: "st2-register-content --config-file /etc/st2/st2.conf{% for item in register.split(',') %} --register-{{ item|trim }}{% endfor %}"
+      default: "st2-register-content --config-file {{config_file}}{% for item in register.split(',') %} --register-{{ item|trim }}{% endfor %}"
     register:
       type: "string"
       default: "actions,aliases,sensors"
       description: "Possible options are all, sensors, actions, rules, aliases."
+    config_file:
+      type: "string"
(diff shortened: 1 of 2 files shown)


contrib/packs/actions/load.yaml
```diff
@@ -10,7 +10,7 @@
       default: true
     cmd:
       immutable: true
-      default: "st2ctl reload{% for item in register.split(',') %} --register-{{ item|trim }}{% endfor %}"
+      default: "st2-register-content --config-file /etc/st2/st2.conf{% for item in register.split(',') %} --register-{{ item|trim }}{% endfor %}"
```
manasdk: Hard coded conf file could be an issue for the future?

DoriftoShoes: Yeah I should probably expose it as a parameter.

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452293699022162

lakshmi-kannan:
sweet!

armab:
This will break `st2_deploy.sh`, which currently still works well (surprisingly).

Just a reminder, if we start bringing Breaking Changes including https://github.com/StackStorm/st2/pull/2333#issuecomment-170120950 and there will be probably others related to packages, it makes sense to ship them in one big release, so everything will be in sync.

contrib/packs/actions/restart_component.yaml
```diff
@@ -10,7 +10,7 @@
       immutable: true
     cmd:
       immutable: true
-      default: "st2ctl restart-component {{servicename}}"
+      default: "service {{servicename}} restart"
```
armab: Probably description:
```yaml
description: "Action that restarts st2 service."
```
needs updating too, since something like this is legit now:
```sh
st2 run restart_component servicename=cron
```

DoriftoShoes:
That's because as much as everyone ragged on st2_deploy.sh, it worked well and was pretty reasonable to read and edit (Despite being massive).

contrib/packs/actions/restart_component.yaml
dennybaa: I would like to ask only one thing, is `service {{svc}} restart` portable between all unix OSes?
fedora, debian, redhat etc all support it?

dennybaa:
Awesome!

contrib/packs/actions/restart_component.yaml
dennybaa: @DoriftoShoes , @armab, @lakshmi-kannan  guys!?

What about this 


@armab check this also please https://github.com/StackStorm/st2/blame/master/st2common/bin/st2ctl#L58-L68

DoriftoShoes: The correct answer is to move that task in the workflow to using this action (linux.service):
https://github.com/StackStorm/st2/blob/master/contrib/linux/actions/service.py

It is currently broken but I will fix it today.

dennybaa: great, thank you for clarification.

armab: I was sure that `service` command is still used in CentOS7 for backward compatibility with `centos6`, it just redirects `service` -> `systemctl`.

For example my Vagrant CentOS7 has `service` command preinstalled: 
```sh
cat /etc/centos-release
CentOS Linux release 7.2.1511 (Core)

[root@centos7 vagrant]# service
Usage: service < option > | --status-all | [ service_name [ command | --full-restart ] ]
```

armab: Fix for new `st2ctl`: https://github.com/StackStorm/st2/pull/2385 to use `systemctl` when possible.

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1454451208004429

DoriftoShoes:
Closing since this was resolved by #2385

PR #2369, opened 2016-01-09
Fix runaway action triggers caused by workflow state miscalculation

m4dcoder:
…l workflow

### DIFF
--- a/st2actions/st2actions/query/mistral/v2.py
+++ b/st2actions/st2actions/query/mistral/v2.py
@@ -131,13 +131,12 @@ class MistralResultsQuerier(Querier):
         # Identify the list of tasks that are not in completed states.
         active_tasks = [t for t in tasks if t['state'] not in DONE_STATES]
 
-        # On cancellation, mistral workflow executions are paused so that tasks can
-        # gracefully reach completion. If any task is not completed, do not mark st2
-        # action execution for the workflow complete. By marking the st2 action execution
-        # as running, this will keep the query for this mistral workflow execution active.
-        if wf_state not in DONE_STATES and not active_tasks and is_action_canceled:
+        # On cancellation, mistral workflow executions are paused so that tasks
+        # can gracefully reach completion. This is only temporary until a canceled
+        # status is added to mistral.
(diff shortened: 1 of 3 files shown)


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452373309022252

st2actions/st2actions/query/mistral/v2.py
```diff
@@ -131,13 +131,12 @@ def _determine_execution_status(self, execution_id, wf_state, tasks):
...
 
-        # On cancellation, mistral workflow executions are paused so that tasks can
-        # gracefully reach completion. If any task is not completed, do not mark st2
-        # action execution for the workflow complete. By marking the st2 action execution
-        # as running, this will keep the query for this mistral workflow execution active.
-        if wf_state not in DONE_STATES and not active_tasks and is_action_canceled:
+        # On cancellation, mistral workflow executions are paused so that tasks
+        # can gracefully reach completion. This is only temporary until a canceled
+        # status is added to mistral.
+        if (wf_state in DONE_STATES or wf_state == 'PAUSED') and is_action_canceled:
```
Kami: Hm, I'm a bit confused about this condition (maybe I'm just not reading / understanding it correctly).

What happens if workflow is in a running state and user cancels an execution? It looks like that right now we won't change the status to `canceled` in that scenario? Is this correct / desired? Or should this be `wf_state not in DONE_STATES`?

m4dcoder: @Kami: that depends. In most cases if users cancels the WF from st2, it'll be canceled by st2. If a user cancels a WF from st2, the WF in mistral shouldn't be in a RUNNING state.  But this can happen, although not recommended, if the user bypasses st2 and uses the mistral CLI directly to manually change the state of the WF. Also, mistral has a debatable behavior on failure, it'll fail the WF fast even though task(s) is still running.  In these cases, we want to maintain the st2 execution status as running so that the querier will continue to query for and update tasks information in the st2 execution result.

Kami: Alright - thanks for the clarification.

PR #2372, opened 2016-01-12
RFR: Add --always-copy option to create_virtualenv method for pack virtual…

lakshmi-kannan:
…envs

See https://github.com/StackStorm/st2-packages/issues/53 for details. 

I anyways added config options for virtualenv_binary and virtualenv_opts in case we want to play around with these in the future so it's easier. The only legit change that fixed anything was `--always-copy`. 

This works on my dev env as well. Meanwhile, @Kami, @DoriftoShoes, @dennybaa and @armab please take a look at my comment on https://github.com/StackStorm/st2-packages/issues/53 and this change. 

TODO

* - [ ] Docs changes --> Not sure where to document this!
* - [X] Changelog

### DIFF
--- a/contrib/packs/actions/pack_mgmt/setup_virtualenv.py
+++ b/contrib/packs/actions/pack_mgmt/setup_virtualenv.py
@@ -125,6 +125,8 @@ class SetupVirtualEnvironmentAction(Action):
 
     def _create_virtualenv(self, virtualenv_path):
         python_binary = cfg.CONF.actionrunner.python_binary
+        virtualenv_binary = cfg.CONF.actionrunner.virtualenv_binary
+        virtualenv_opts = cfg.CONF.actionrunner.virtualenv_opts
 
         if not os.path.isfile(python_binary):
             raise Exception('Python binary "%s" doesn\'t exist' % (python_binary))
@@ -132,8 +134,15 @@ class SetupVirtualEnvironmentAction(Action):
         self.logger.debug('Creating virtualenv in "%s" using Python binary "%s"' %
                           (virtualenv_path, python_binary))
 
-        cmd = ['virtualenv', '-p', python_binary, '--system-site-packages', virtualenv_path]
-        exit_code, _, stderr = run_command(cmd=cmd)
+        cmd = [virtualenv_binary, '-p', python_binary]
+        cmd.extend(virtualenv_opts)
+        cmd.extend([virtualenv_path])
+        self.logger.debug('Running command "%s" to create virtualenv.', ' '.join(cmd))
+
+        try:
+            exit_code, _, stderr = run_command(cmd=cmd)
+        except OSError:
+            raise Exception('Virtualenv binary "%s" doesn\'t exist.' % virtualenv_binary)
 
         if exit_code != 0:
             raise Exception('Failed to create virtualenv in "%s": %s' %
--- a/contrib/packs/actions/pack_mgmt/setup_virtualenv.py
+++ b/contrib/packs/actions/pack_mgmt/setup_virtualenv.py
@@ -131,6 +131,9 @@ class SetupVirtualEnvironmentAction(Action):
         if not os.path.isfile(python_binary):
             raise Exception('Python binary "%s" doesn\'t exist' % (python_binary))
 
+        if not os.path.isfile(virtualenv_binary):
+            raise Exception('Virtualenv binary "%s" doesn\'t exist.' % (virtualenv_binary))
+
         self.logger.debug('Creating virtualenv in "%s" using Python binary "%s"' %
                           (virtualenv_path, python_binary))
 
@@ -141,8 +144,9 @@ class SetupVirtualEnvironmentAction(Action):
 
         try:
             exit_code, _, stderr = run_command(cmd=cmd)
-        except OSError:
-            raise Exception('Virtualenv binary "%s" doesn\'t exist.' % virtualenv_binary)
+        except OSError as e:
+            raise Exception('Error executing command %s. %s.' % (' '.join(cmd),
+                                                                 e.message))
 
         if exit_code != 0:
             raise Exception('Failed to create virtualenv in "%s": %s' %
--- a/st2actions/st2actions/config.py
+++ b/st2actions/st2actions/config.py
@@ -17,6 +17,7 @@
 Configuration options registration and useful routines.
 """
 
+import os
 import sys
 
 from oslo_config import cfg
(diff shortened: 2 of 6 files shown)


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452564146022529

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1452574436022548

contrib/packs/actions/pack_mgmt/setup_virtualenv.py
```diff
@@ -132,8 +134,15 @@ def _create_virtualenv(self, virtualenv_path):
...
-        exit_code, _, stderr = run_command(cmd=cmd)
+        cmd = [virtualenv_binary, '-p', python_binary]
+        cmd.extend(virtualenv_opts)
+        cmd.extend([virtualenv_path])
+        self.logger.debug('Running command "%s" to create virtualenv.', ' '.join(cmd))
+
+        try:
+            exit_code, _, stderr = run_command(cmd=cmd)
+        except OSError:
+            raise Exception('Virtualenv binary "%s" doesn\'t exist.' % virtualenv_binary)
```
Kami: Ideally we would also include original error message at the end.

st2actions/st2actions/config.py
```diff
@@ -41,11 +42,19 @@ def _register_common_opts():
...
     logging_opts = [
         cfg.StrOpt('logging', default='conf/logging.conf',
                    help='location of the logging.conf file'),
-        cfg.StrOpt('python_binary', default=sys.executable,
-                   help='Python binary which will be used by Python actions.')
+        cfg.StrOpt('python_binary', default=default_python_bin_path,
+                   help='Python binary which will be used by Python actions.'),
+        cfg.StrOpt('virtualenv_binary', default=default_virtualenv_bin_path,
+                   help='Virtualenv binary which should be used to create pack virtualenvs.'),
+        cfg.ListOpt('virtualenv_opts', default=['--always-copy', '--system-site-packages'],
```
Kami: Did we verify how much of a difference in terms of virtualenv disk usage `--always-copy` makes?

Kami: I also googled around a bit and it looks like `--always-copy` didn't always work correctly in some older versions of virtualenv.

Because of that we need to make sure the latest confirmed working version of virtualenv and installed and used.

lakshmi-kannan: Please see StackStorm/st2-packages#53 with regards to disk space requirements. I already pasted the disk spaces there. 

>Because of that we need to make sure the latest confirmed working version of virtualenv and installed and used.

Yep. I saw the issue and tested with virtualenv 12.1.1 which was five versions earlier and --always-copy seems to work. dh-virtualenv uses latest virtualenv as far as I can tell so this shouldn't be a problem.

contrib/packs/actions/pack_mgmt/setup_virtualenv.py
lakshmi-kannan: The original error message was "No such file or directory" without naming the file. This was confusing. Thus this new message.

Kami: Agreed.

But `OSError` doesn't necessary only mean that the file doesn't exist, it could also represent a permissions error, etc. That's why I think it's also good to include the original message somewhere.

lakshmi-kannan: Ok, let me log the original exception and raise this new one. Does that sound good to you?

Kami: I would prefer to include it in the exception message, but whatever makes you happy :P

lakshmi-kannan:
Validated with this that sensors and actions are able to run fine with the new packages. Also, this opened up another issue https://stackstorm.atlassian.net/browse/STORM-1949 . cc: @dennybaa @Kami

manasdk:
Yup `--always-copy` worked for me as well.

st2actions/st2actions/config.py
dennybaa: Regarding `--system-site-packages`
```
If you build with virtualenv --system-site-packages ENV, your virtual environment will inherit packages from /usr/lib/python2.7/site-packages (or wherever your global site-packages directory is).
```
It should copy global packages, but since virtualenv is not really relocatable, that just works in some weird way.

**with --system-site-packages**
```
root@3cad709e2da4:/opt/stackstorm/virtualenvs/github# bin/pip list
beautifulsoup4 (4.3.2)
bzr (2.6.0.dev2)
chardet (2.0.1)
configobj (4.7.2)
mercurial (2.2.2)
pip (7.1.2)
PyGithub (1.26.0)
python-apt (0.8.8.2)
python-debian (0.1.21)
requests (2.5.3)
setuptools (18.2)
six (1.10.0)
```
**without**
```
root@3cad709e2da4:~# /opt/stackstorm/virtualenvs/github/bin/pip list
beautifulsoup4 (4.3.2)
pip (7.1.2)
PyGithub (1.26.0)
requests (2.5.3)
setuptools (18.2)
six (1.10.0)
wheel (0.24.0)
```

So what is unique only for origin venv is, and it's not required by github pack!
```
bzr (2.6.0.dev2)
chardet (2.0.1)
configobj (4.7.2)
mercurial (2.2.2)
python-apt (0.8.8.2)
python-debian (0.1.21)
```

Kami: Yeah, as noted on Slack, we added this a while back as a quick workaround to get things working.

I would much prefer to not use this option if possible.

For sensors and actions to work though, they also need access to st2common package and all the inherited st2common dependencies (mongoengine, kombu, etc.).

If we can get it to work without `--system-site-packages`, that would be great. I think it might even be possible now with new packages since st2actions and st2reactor already depend on st2common package.

dennybaa: According to explaations of @lakshmi-kannan , `PYTHONPATH` is injected with a origin virtualenv, that's why `--system-site-packages` should be removed!?

dennybaa:
+1 to merge. One thing `--system-site-packages` should be removed, since it's not required anymore due to injection to `PYTHONPATH`.

st2actions/st2actions/config.py
Kami: To clarify - the best way to test it is to remove this option, create virtualenv and make sure actions and sensors still work.

lakshmi-kannan:
@dennybaa Unfortunately, we cannot remove --system-site-packages from the code because old packages still need them. See https://stackstorm.slack.com/archives/stackstorm/p1452708198024063

So in st2-packages, can you set the virtualenv_opts = ['--always-copy'] in /etc/st2/st2.conf. This way new packages won't use `--system-site-packages`.

lakshmi-kannan:
Seems like travis is stuck. I'll merge and deal with build failures if any.

lakshmi-kannan:
See https://github.com/StackStorm/puppet-st2/pull/148/files. Basically, --always-copy fails with RHEL boxes (https://github.com/pypa/virtualenv/issues/565#issuecomment-115036122). So I am setting a config option for existing installation to just use --system-site-packages. RHEL st2bundle won't work with --always-copy but @DoriftoShoes and I decided to deal with that later.

lakshmi-kannan:
If I don't get workroom to pass with https://github.com/StackStorm/puppet-st2/pull/148, I'll revert this change from master for 1.3 release. cc: @Kami, @DoriftoShoes, @emedvedev, @dennybaa.

PR #2373, opened 2016-01-12
Enhancements : log file for pack actions

valarmathip:
Handled pack actions to log to a file in addition to the current logging to stdout. as per the story PXPSE - 422.


Kami:
Thanks for the contribution.

Can you please provide more details (feel free to update the PR description) on what this PR is supposed to do and solve?

Thanks

st2common/st2common/logging/handlers.py
```diff
@@ -37,7 +38,8 @@ def __init__(self, filename, mode='a', maxBytes=0, backupCount=0, encoding=None,
         format_values = {
             'timestamp': timestamp,
             'ts': isotime_str,
-            'pid': pid
+            'pid': pid,
+            'pack': os.path.abspath(os.path.dirname(sys.argv[2])).split('/')[-2]
```
Kami: I'm not sure I fully understand how this is supposed to work.

It seems like we parse the pack name from the command line arguments? If so, this doesn't seem robust (not to mention, that it relies on an unstable command line arguments order, etc.).

Is there a safer and more robust way to achieve this?

st2actions/st2actions/runners/pythonrunner.py
```diff
@@ -84,17 +83,10 @@ def _set_up_logger(self):
         Set up a logger which logs all the messages with level DEBUG
         and above to stderr.
         """
+        logging.config.fileConfig('/etc/st2actions/logging.conf')
```
Kami: This path can't be hard-coded and needs to be configurable.

Please see the existing code on how we handle those paths - those values are provided in the config and can be overridden by the user.

Kami: I just had a look at the code, it looks like you should be able to use the existing configuration option - `cfg.CONF.actionrunner.logging`

st2common/st2common/logging/handlers.py
azamsheriff: Yes Kami your understanding is correct. we tried to get the pack name from the command line arguments.  As per your suggestion now We are looking into alternate way to get the pack name.

andrew-regan:
@Kami 
*"Can you please provide more details (feel free to update the PR description) on what this PR is supposed to do and solve?"*

Today python action logging is collected up into a single line of output in *st2actionrunner.[PID].log* with embedded '\n' characters instead of newlines. This not very readable when trying to look through the logging of a single action execution. The intent of this PR is to break the action logging out into their own log files per pack with each log message on its own line instead of embedded '\n''s.

This was briefly discussed in stackstorm_community on Slack between ryanms-plexxi, jfryman, and doriftoshoes for reference. :-)

lakshmi-kannan:
Though I understand the high level need, I am with kami on solution being hacky and not clean. I don't have an alternate proposal but we'll discuss internally first.

andrew-regan:
Thanks! Let us know.

st2common/st2common/logging/handlers.py
Kami: I will look into it, but it will probably require some refactoring.

azamsheriff: Thanks Kami.

andrew-regan:
@lakshmi-kannan @Kami any update on an alternative for retrieving the pack name? One thought is to leverage a similar approach to the one in PR #2396. PythonActionWrapper knows the pack and could set action_instance.pack the same way it is currently setting action_instance.datastore. Thoughts?

Kami:
@andrew-regan Sorry, no update yet, been busy with 1.3 release and related stuff lately.

Kami:
@andrew-regan I had a look at the code again and yes, you could handle that in the python action wrapper since the action wrapper is already aware of the pack name (similar to the sensor wrapper).

andrew-regan:
We'll take a pass at implementing that and see if it looks any better.

andrew-regan:
@lakshmi-kannan @Kami 

Any thoughts on the approach in this commit? https://github.com/StackStorm/st2/commit/8ced5e181d7992a2707694f55dcd57b4efd8423a

I think this is better aligned with the logging in other st2 components, but doesn't loose the PID information. The PID is part of the log message instead of being in the name of the file.

Thanks!

st2actions/conf/logging.conf
```diff
@@ -21,13 +21,13 @@ args=(sys.stdout,)
 class=st2common.log.FormatNamedFileHandler
 level=INFO
 formatter=verboseConsoleFormatter
-args=('logs/st2actionrunner.{pid}.log',)
+args=('logs/st2actionrunner.{pack}.log',)
```
manasdk: ActionRunner itself logs when not in context of running an action in a pack. Per pack logging make very little sense in that case.

st2actions/conf/logging.conf
```diff
@@ -36,7 +36,7 @@ datefmt=
 
 [formatter_verboseConsoleFormatter]
 class=st2common.logging.formatters.ConsoleLogFormatter
-format=%(asctime)s %(thread)s %(levelname)s %(module)s [-] %(message)s
+format=%(asctime)s %(process)d %(thread)s %(levelname)s %(module)s [-] %(message)s
```
manasdk: I am ok with including process in the line number. If we maintain per pid logging will end up being redundant.

st2actions/st2actions/runners/pythonrunner.py
```diff
@@ -77,6 +78,21 @@ def __init__(self, config=None):
     def run(self, **kwargs):
         pass
 
+    def _set_up_logger(self):
```
manasdk: This method is not being called anywhere. Are we missing some integration code?

Also, similar impl exists in pythonactionwrapper now so not a good idea to override here anyway. See https://github.com/StackStorm/st2/blob/master/st2actions/st2actions/runners/python_action_wrapper.py#L110

manasdk:
-1

Logs should be independent of pack running the action. I honestly believe PID is the right way to deal with log. We always control the no of action runners but packs are unbounded.

Besides, in production actionrunner logs should be sent to a rsyslog or a similar util. This way all logs can end up in a single file and be easy to search etc. Also, the idea that the process can depending on context end up writing to different log files rubs me the wrong way.

I see we are solving 2 separate issues -
1. break the action logging out into their own log files per pack
2. each log message on its own line instead of embedded '\n''s

**-1 to #1.**

 **+1 to #2**

Not sure I see how #2 is solved by the PR yet - maybe some code is missing?

manasdk:
Some offline discussions led us to come to a conclusion that this approach is not going to necessarily get us to the desired solution.

@andrew-regan is investigating alternate way to pretty print logs from action output so that the NLs are properly preserved.

PR #2376, opened 2016-01-12
Throw a more friendly error message if casting a param value fails

Kami:
With this pull request we now throw a more friendly error message when casting a parameter value to the type defined in the schema fails.

Previously, if casting would fail (e.g. value doesn't contain the right type), a very cryptic error similar to the one below was returned.

```python
    return ast.literal_eval(x)
  File "/usr/lib/python2.7/ast.py", line 80, in literal_eval
    return _convert(node_or_string)
  File "/usr/lib/python2.7/ast.py", line 79, in _convert
    raise ValueError('malformed string')
ValueError: malformed string
```

Keep in mind that the actual error message could be different because it depends on the actual value (`ast.literval_eval` will throw a different error depending on the value).

Now error now includes the offending parameter name, value and type as shown below:

```python
Failed to cast value "stringnotanarray" for parameter "arrtype" of type "array"'
```

This issue was initially encountered by one of our users and reported on Slack. Because the error message was very cryptic and didn't carry enough context (parameter name, etc.) it meant we needed to waste quite a look of time to track down the actual root cause.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -44,7 +44,9 @@ in development
 * Fix action parameters validation so that only a selected set of attributes can be overriden for
   any runner parameters. (bug fix)
 * Fix type in the headers parameter for the http-request runner. (bug fix)
-* Fix runaway action triggers caused by state miscalculation for mistral workflow. (bug fix) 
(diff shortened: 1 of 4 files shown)


jfryman:
:heart:

PR #2377, opened 2016-01-13
Fix a bug with the top-level error reporting in the new SSH runner

Kami:
Before this fix, remote script runner based on the new paramiko SSH runner would throw an exception instead of correctly reporting an error if a top-level error occurred (e.g. copying artifacts to remote hosts failed or similar).

I noticed this issue on the build server yesterday and it's quite easy to replicate it.

```bash
  error: '''bool'' object has no attribute ''get'''
  traceback: "  File "/data/stanley/st2actions/st2actions/container/base.py", line 116, in _do_run
    (status, result, context) = runner.run(action_params)
  File "/data/stanley/st2actions/st2actions/runners/remote_script_runner.py", line 97, in run
    status = self._get_result_status(result, cfg.CONF.ssh_runner.allow_partial_failure)
  File "/data/stanley/st2actions/st2actions/runners/ssh/paramiko_ssh_runner.py", line 169, in _get_result_status
    r_succeess = r.get('succeeded', False) if r else False
"
```

Note: top-level error is an error which is global and doesn't belong to a host so the result dictionary doesn't contain an entry per host.

TODO:

- [x] Tests

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -47,6 +47,8 @@ in development
 * Fix runaway action triggers caused by state miscalculation for mistral workflow. (bug fix)
 * Throw a more friendly error message if casting parameter value fails because the value contains
   an invalid type or similar. (improvement)
+* Fix a bug in the remote script runner which would throw an exception if a remote script action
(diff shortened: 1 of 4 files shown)


lakshmi-kannan:
LGTM.

PR #2380, opened 2016-01-13
Implement task rerun for mistral workflow

m4dcoder:


Kami:
Sorry for the late comments.

st2actions/st2actions/container/base.py
```diff
@@ -63,6 +67,10 @@ def _get_runner(self, runnertype_db, action_db, liveaction_db):
         runner.libs_dir_path = self._get_action_libs_abs_path(action_db.pack,
                                                               action_db.entry_point)
 
+        # For re-run, get the ActionExecutionDB in which the re-run is based on.
+        rerun_ref_id = runner.context.get('re-run', {}).get('ref')
+        runner.rerun_ex_ref = ActionExecution.get(id=rerun_ref_id) if rerun_ref_id else None
```
Kami: Can we explicitly assign all the variables which get late / dynamically assigned to the runner as class level variables?

So:

```python
class RunnerContainer(object):
    rerun_ex_ref = None
    # etc.
```

Without that it's really hard to understand the code and know which variables are available when. I know it's not your fault and some of the code has been like that from the beginning, but we should try to improve it going forward when possible.

Kami: Actually, ignore that comment, sorry.

I see you do it in the constructor bellow. It's not ideal since it's still dynamic and pylint cant detect those variables so it would be better to assign it as class level variable, but I'm fine with doing this later in another PR.

st2actions/st2actions/runners/mistral/v2.py
```diff
@@ -192,6 +180,30 @@ def run(self, action_parameters):
...
+    @retrying.retry(
+        retry_on_exception=utils.retry_on_exceptions,
+        wait_exponential_multiplier=cfg.CONF.mistral.retry_exp_msec,
+        wait_exponential_max=cfg.CONF.mistral.retry_exp_max_msec,
+        stop_max_delay=cfg.CONF.mistral.retry_stop_max_msec)
+    def run(self, action_parameters):
+        resume_options = self._get_resume_options()
+        tasks = resume_options.get('tasks', [])
+        resume = self.rerun_ex_ref and tasks
+        return self.resume(self.rerun_ex_ref, tasks) if resume else self.start(action_parameters)
```
Kami: This is just a minor style thing - short hand if notation usually only tends to be used for simple variable assignments.

For cases like this it's usually more readable if you have an explicit if and else.

st2actions/st2actions/runners/mistral/v2.py
```diff
@@ -238,6 +253,51 @@ def run(self, action_parameters):
...
+    def resume(self, ex_ref, task_names):
+        mistral_ctx = ex_ref.context.get('mistral', dict())
+
+        if not mistral_ctx.get('execution_id'):
+            raise Exception('Unable to rerun because mistral execution_id is missing.')
+
+        execution = self._client.executions.get(mistral_ctx.get('execution_id'))
+
+        # pylint: disable=no-member
+        if execution.state not in ['ERROR']:
```
Kami: I was confused for a second, but not I see this refers to the Mistral execution and not StackStorm one :)

st2actions/st2actions/runners/mistral/v2.py
```diff
@@ -238,6 +253,51 @@ def run(self, action_parameters):
 
         return (status, partial_results, exec_context)
 
+    def resume(self, ex_ref, task_names):
+        mistral_ctx = ex_ref.context.get('mistral', dict())
+
+        if not mistral_ctx.get('execution_id'):
+            raise Exception('Unable to rerun because mistral execution_id is missing.')
+
+        execution = self._client.executions.get(mistral_ctx.get('execution_id'))
```
Kami: Would this throw if mistral_ctx doesn't contain execution_id?

st2actions/st2actions/runners/mistral/v2.py
Kami: Using a top level constant for mistral_re_runable_states or similar is probably a good idea.

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -267,18 +268,26 @@ class ActionExecutionReRunController(ActionExecutionsControllerMixin, ResourceCo
...
 
+            if self.tasks:
+                assert isinstance(self.tasks, list)
+
             return self
 
-    @jsexpose(body_cls=ExecutionParametersAPI, status_code=http_client.CREATED)
-    def post(self, execution_parameters, execution_id):
+    @jsexpose(body_cls=ExecutionSpecificationAPI, status_code=http_client.CREATED)
+    def post(self, spec, execution_id):
```
Kami: Minor thing - for consistency it would be good to add `_api` suffix to the argument name (yeah, I know it wasn't not there earlier, but we have a chance to improve it now).

st2actions/st2actions/container/base.py
```diff
@@ -44,6 +44,10 @@
 
 class RunnerContainer(object):
 
+    def _get_rerun_reference(self, context):
+        execution_id = context.get('re-run', {}).get('ref')
```
Kami: Might be better to be explicit and call it `execution_id` or similar?

st2client/st2client/commands/action.py
```diff
@@ -280,16 +280,16 @@ def _print_execution_details(self, execution, args, **kwargs):
...
 
-        tasks = getattr(args, 'tasks', False)
+        show_tasks = getattr(args, 'show_tasks', False)
         raw = getattr(args, 'raw', False)
         detail = getattr(args, 'detail', False)
         key = getattr(args, 'key', None)
         attr = getattr(args, 'attr', [])
 
-        if tasks and not is_workflow_action:
-            raise ValueError('--tasks option can only be used with workflow actions')
```
Kami: That's backward incompatible / breaking change, right?

Kami:
Added some comments, mostly style / readability stuff.

:+1:

PR #2382, opened 2016-01-14
Remove --always-copy from default options

lakshmi-kannan:
--always-copy breaks on RHEL boxes. And also this shouldn't be enabled by default because old packages don't require them.


lakshmi-kannan:
Travis isn't running again. I'll merge.

PR #2383, opened 2016-01-14
Fix execution cancellation for task of mistral workflow

m4dcoder:
Cancellation of task execution was not communicated to mistral workflow which leaves the workflow execution in RUNNING state.

### DIFF
--- a/st2actions/st2actions/container/base.py
+++ b/st2actions/st2actions/container/base.py
@@ -186,6 +186,10 @@ class RunnerContainer(object):
                 liveaction_db=liveaction_db)
 
             executions.update_execution(liveaction_db)
+
+            LOG.debug('Performing post_run for runner: %s', runner.runner_id)
+            runner.post_run(liveaction_db.status, {'error': 'Execution canceled by user.'})
+            runner.container_service = None
         except:
             _, ex, tb = sys.exc_info()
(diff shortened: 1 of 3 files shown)


st2actions/st2actions/container/base.py
```diff
@@ -186,6 +186,10 @@ def _do_cancel(self, runner, runnertype_db, action_db, liveaction_db):
                 liveaction_db=liveaction_db)
 
             executions.update_execution(liveaction_db)
+
+            LOG.debug('Performing post_run for runner: %s', runner.runner_id)
```
Kami: Hm, do we also want to execute this piece of code on cancel - https://github.com/StackStorm/st2/blob/2da0bfc42f5fb36453b1cd1c463ff49cd45942c7/st2actions/st2actions/container/base.py#L154?

It seems like we do (deleting the auth token)

Current approach looks fine to me for now, but it looks like some refactoring is in order.

Kami:
LGTM :+1: for now, but it seems like some refactoring is in order soon.

As mentioned in the comment above, I believe this piece of code (https://github.com/StackStorm/st2/pull/2383/files#r49835506) should also be called on cancelation.

The issue seems to be related that cancelation was added after the fact and we didn't do any refactoring when adding initial implementation for it.

PR #2384, opened 2016-01-14
Display action parameters in "st2 execution get" output for workflow actions

Kami:
With this change we now also display action parameters by default when using "st2 execution get" on workflow actions.

This will save everyone from issuing another command or using "-d" option. It's also worth mentioning that "-d" is not exactly the same since it includes a lot of other noise and doesn't really satisfy "at a glance" requirement.

Before:

selection_322

After:

selection_323

I get irritated by this every time I'm debugging st2workroom test issues. I always need to execute another command to figure out if the execution refers to Ubuntu / RHEL. With this change, it will save me (and other people) that hassle.

### DIFF
--- a/st2client/st2client/commands/action.py
+++ b/st2client/st2client/commands/action.py
@@ -314,7 +314,7 @@ class ActionRunCommandMixin(object):
         action_exec_mgr = self.app.client.managers['LiveAction']
 
         instance = execution
-        options = {'attributes': ['id', 'action.ref', 'status', 'start_timestamp',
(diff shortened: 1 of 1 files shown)


manasdk:
+1

Best thing after sliced bread!

DoriftoShoes:
Love it!

dzimine:
+100 excellent

jfryman:
It's the small touches and little things.

stay-classy

lakshmi-kannan:
nice!

PR #2385, opened 2016-01-14
Detect service manager for new st2ctl

armab:
Provides fix for https://github.com/StackStorm/st2/pull/2367/files#r49736848

when `service` command is not available under `CentOS7` (replaced by `systemctl`).

Tested/Verified.

### DIFF
--- a/st2common/bin/st2ctl
+++ b/st2common/bin/st2ctl
@@ -68,12 +68,12 @@ function st2stop() {
 }
 
 function service_manager() {
-  if command -v systemctl 2>/dev/null; then
+  if [ -d /run/systemd/system ]; then
     systemctl "${2}" "${1}"
   elif command -v service 2>/dev/null; then
     service "${1}" "${2}"
   else
-    echo -e "\e[31mError: Unknown service manager in the system! \e[0m\n"
+    echo -e "\e[31mError: Unsupported service manager in the system! \e[0m\n"
     exit 1
   fi
 }
--- a/st2common/bin/st2ctl
+++ b/st2common/bin/st2ctl
@@ -57,14 +57,25 @@ function validate_in_components() {
 
 function st2start() {
   for COM in ${COMPONENTS}; do
-    service ${COM} start
+    service_manager ${COM} start
   done
 }
 
 function st2stop() {
   for COM in ${COMPONENTS}; do
-    service ${COM} stop
+    service_manager ${COM} stop
   done
+}
+
+function service_manager() {
+  if command -v systemctl 2>/dev/null; then
+    systemctl "${2}" "${1}"
+  elif command -v service 2>/dev/null; then
+    service "${1}" "${2}"
+  else
+    echo -e "\e[31mError: Unknown service manager in the system! \e[0m\n"
+    exit 1
+  fi
 }
 
 # Next candidate for removal


st2common/bin/st2ctl
```diff
@@ -57,16 +57,27 @@ function validate_in_components() {
...
   done
 }
 
+function service_manager() {
+  if command -v systemctl 2>/dev/null; then
+    systemctl "${2}" "${1}"
+  elif command -v service 2>/dev/null; then
+    service "${1}" "${2}"
+  else
+    echo -e "\e[31mError: Unknown service manager in the system! \e[0m\n"
```
lakshmi-kannan: s/Unknown/Unsupported

lakshmi-kannan:
lgtm. nice!

st2common/bin/st2ctl
```diff
@@ -57,16 +57,27 @@ function validate_in_components() {
...
 function st2stop() {
   for COM in ${COMPONENTS}; do
-    service ${COM} stop
+    service_manager ${COM} stop
   done
 }
 
+function service_manager() {
+  if command -v systemctl 2>/dev/null; then
+    systemctl "${2}" "${1}"
```
dennybaa: This is not reliable check for systemd!
http://unix.stackexchange.com/questions/196166/how-to-find-out-if-a-system-uses-sysv-upstart-or-systemd-initsystem

> **Detecting systemd**
To check for systemd as the running system manager in the "official" manner, one checks for the existence of /run/systemd/system. This is a directory, in /run, that systemd itself creates at boot, and that other system managers are unlikely to create.

> But that's merely unlikely. This check is already broken, because uselessd creates this directory too.

> Other, unofficial, checks won't work:

> systemd publishes a whole RPC API over D-Bus, which even contains a version name and number; but:
This is not covered by the infamous "Interface Stability Guarantee" and could change tomorrow or at whim.
So too does the lookalike D-Bus server in systemd-shim.
So too does uselessd.
The existence of /run/systemd/private is similarly not guaranteed and similarly duplicated by uselessd.

> https://github.com/StackStorm/st2-packages/blob/master/packages/st2actions/bin/runners.sh#L17

armab: > Also I check the same way,
https://github.com/StackStorm/st2-packages/blob/master/packages/st2actions/bin/runners.sh#L17

And you're running `systemctl` binary: https://github.com/StackStorm/st2-packages/blob/master/packages/st2actions/bin/runners.sh#L20


I rely on existence of the toolset, rather then on init system, because our case is really simple: if `systemctl` exists - we want to use it, otherwise `service`.
I can put `service` before `systemctl`, to be sure. That'll cover us, just `service` will redirect to `systemctl` in CentOS7.

----

Also `service` already worked without issues before: https://github.com/StackStorm/st2/blob/master/tools/st2ctl#L152-L157 in AIO installer. See my case when `service` exists by default for compatibility in `CentOS7`: https://github.com/StackStorm/st2/pull/2367#discussion_r49769025

dennybaa: Wrong assumption. ex. docker container started without systemd, you check binary it exists then you invoke a call and it fails... You *must* check whether systemd is **running**.

dennybaa: Why do you think i don't use check kinda yours? because i've already faced the issue and at the begging the check was exactly the same)

armab: So that CentOS7 screenshot is from Docker?
See my comments there: https://github.com/StackStorm/st2/pull/2367/files#r49769025

Issue #2378, opened 2016-01-13
Issue with action-chain notify not deleting properly


charlottestjohn:
Issue with action-chain notify not deleting properly

Ways to reproduce: 

Create a simple action chain like this: https://gist.github.com/charlottestjohn/3d965e0b4bca4cdd1cbd (we needed a custom notify rule to work w/ hipchat)

st2ctl reload --register-all; sudo /etc/init.d/docker-hubot restart

Run the action, verify that it comes back with the alias notify response

Update the chain's meta to add a notify portion, like so: https://gist.github.com/charlottestjohn/bdf790740661be658fc3

st2ctl reload --register-all; sudo /etc/init.d/docker-hubot restart

Run the action, verify that the notify portion is the only one that comes back in chat.

Delete the notify section

st2ctl reload --register-all; sudo /etc/init.d/docker-hubot restart

Run the action, at this point, assumption would be that the notify portion of the alias would take over - howeever the chain's old meta notify still is in effect.

For further curiosity: 
Change the name of the chain, and change what the alias points to, like so: 

st2ctl reload --register-all; sudo /etc/init.d/docker-hubot restart

Changing the name appears to let it actually pick up the new changes. If you change the name back, it will go right back to the old deleted notify.

lakshmi-kannan:
@charlottestjohn Confirmed there is a bug somewhere in action not getting updated. Was able to repro in dev environment. https://gist.github.com/lakshmi-kannan/54c68d42f61b6626ac3c

Thanks for the report and repro steps. Much appreciated. Will get back to you with a fix.

emedvedev:
Fixed by #2389

PR #2389, opened 2016-01-15
RFR: When notify is not specified, mongoengine requires an empty doc

lakshmi-kannan:
Original bug: https://github.com/StackStorm/st2/issues/2378 (Thanks @ charlottestjohn).

Summary:

1. Register action with notify
2. Remove notify section from meta. Register notify again.
3. Check action. You'll still see notify section 

Repro: 
See https://gist.github.com/lakshmi-kannan/3891705755dbafa8b82c for repro. 

What happened?

We use embedded document model for notify in action model. If notify is unspecified, we set notify to None. Mongoengine interprets `None` as unmodified field therefore doesn't delete the embedded document. 

Solution: We now explicitly set the embedded document to {} instead of None. 

TODO:

- [X] Tests
- [x] Changelog

### DIFF
--- a/st2api/st2api/controllers/v1/actions.py
+++ b/st2api/st2api/controllers/v1/actions.py
@@ -155,8 +155,10 @@ class ActionsController(resource.ContentPackResourceController):
 
         try:
             action_db = ActionAPI.to_model(action)
+            LOG.debug('/actions/ PUT incoming action: %s', action_db)
             action_db.id = action_id
             action_db = Action.add_or_update(action_db)
+            LOG.debug('/actions/ PUT after add_or_update: %s', action_db)
         except (ValidationError, ValueError) as e:
             LOG.exception('Unable to update action data=%s', action)
             abort(http_client.BAD_REQUEST, str(e))
--- a/st2common/st2common/models/api/action.py
+++ b/st2common/st2common/models/api/action.py
@@ -236,6 +236,9 @@ class ActionAPI(BaseAPI, APIUIDMixin):
         if getattr(action, 'notify', None):
             notify = NotificationsHelper.to_model(action.notify)
         else:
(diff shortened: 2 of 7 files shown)


st2common/st2common/models/api/action.py
```diff
@@ -236,7 +236,7 @@ def to_model(cls, action):
         if getattr(action, 'notify', None):
             notify = NotificationsHelper.to_model(action.notify)
         else:
-            notify = None
+            notify = NotificationsHelper.to_model({})
```
manasdk: Please add a comment here why this is required since it is peculiar.

st2common/st2common/models/api/notification.py
```diff
@@ -85,6 +85,10 @@ def from_model(notify_model):
         if getattr(notify_model, 'on_failure', None):
             notify['on-failure'] = NotificationsHelper._from_model_sub_schema(
                 notify_model.on_failure)
+
+        if not notify:
+            return None
```
manasdk: Why `None` and not `notify` which would be empty dict?

st2api/st2api/controllers/v1/actions.py
```diff
@@ -155,8 +155,10 @@ def put(self, action_ref_or_id, action):
 
         try:
             action_db = ActionAPI.to_model(action)
+            LOG.debug('/actions/ PUT incoming action: %s', action_db)
```
manasdk: :)

st2common/st2common/models/api/action.py
Kami: Agreed.

Kami:
I'm worried that there are other problems similar to this one lurking around.

In any case, we need to make sure to add a test case for it :)

PR #2395, opened 2016-01-18
Runner container improvements and fixes

Kami:
I noticed some "issues" while reviewing other PRs so I decided  to address those.

1. Refactor the code a bit to make it easier to read and follow
2. Make sure temporary action-specific token is also deleted when user cancels an action (previously token was only deleted if action succeeded / failed / timed out and wasn't canceled)


Kami:
This is mostly a refactor and not functionality changes (besides fixing the token deletion) so I will go ahead and merge it since we already have good coverage for those parts.

lakshmi-kannan:
lgtm

Issue #2398, opened 2016-01-19
RulesEngine quit due to exception


alfasin:
I suspect that there's a bug in eventlet that, once hit this edge-case, case the rule-engine to go down.

```
2015-11-14 09:59:01,864 140360772253008 ERROR log [-] (PID:12405) RulesEngine quit due to exception.
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/st2reactor/cmd/rulesengine.py", line 46, in _run_worker
    return timer_thread.wait() and rules_engine_worker.wait()
  File "/usr/local/lib/python2.7/dist-packages/eventlet/greenthread.py", line 175, in wait
    return self._exit_event.wait()
  File "/usr/local/lib/python2.7/dist-packages/eventlet/event.py", line 121, in wait
    return hubs.get_hub().switch()
  File "/usr/local/lib/python2.7/dist-packages/eventlet/hubs/hub.py", line 294, in switch
    return self.greenlet.switch()
  File "/usr/local/lib/python2.7/dist-packages/eventlet/hubs/hub.py", line 346, in run
    self.wait(sleep_time)
  File "/usr/local/lib/python2.7/dist-packages/eventlet/hubs/poll.py", line 85, in wait
    presult = self.do_poll(seconds)
  File "/usr/local/lib/python2.7/dist-packages/eventlet/hubs/epolls.py", line 62, in do_poll
    return self.poll.poll(seconds)
  File "/usr/lib/python2.7/dist-packages/st2common/signal_handlers.py", line 44, in handle_sigusr1
    reopen_log_files(handlers=handlers)
  File "/usr/lib/python2.7/dist-packages/st2common/logging/misc.py", line 50, in reopen_log_files
    handler.release()
  File "/usr/lib/python2.7/logging/__init__.py", line 705, in release
    self.lock.release()
  File "/usr/lib/python2.7/threading.py", line 203, in release
    raise RuntimeError("cannot release un-acquired lock")
RuntimeError: cannot release un-acquired lock
2015-11-14 09:59:01,920 140360772253008 INFO base [-] Scheduler has been shut down
```

Potentially related issue:
http://stackoverflow.com/questions/23116383/thread-condition-variables-un-acquired-lock

Behavioral issue:
Once an exception was raise, from line 63 in our case:
https://github.com/StackStorm/st2/blob/master/st2reactor/st2reactor/cmd/rulesengine.py#L63

The default behavior is to shutdown the rule-engine:
[...]

PR #2401, opened 2016-01-19
Ignore lock release errors when re-opening log files

Kami:
Releasing a lock is not necessary fatal - it could simple indicate that the lock wasn't acquired in the first place.

Part of #2398

### DIFF
--- a/st2common/st2common/logging/misc.py
+++ b/st2common/st2common/logging/misc.py
@@ -43,12 +43,21 @@ def reopen_log_files(handlers):
 
         LOG.info('Re-opening log file "%s" with mode "%s"\n' %
                  (handler.baseFilename, handler.mode))
+
         try:
             handler.acquire()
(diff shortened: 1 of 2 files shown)


st2common/st2common/logging/misc.py
```diff
@@ -43,12 +43,21 @@ def reopen_log_files(handlers):
 
         LOG.info('Re-opening log file "%s" with mode "%s"\n' %
                  (handler.baseFilename, handler.mode))
+
         try:
             handler.acquire()
```
Kami: Another option would be to wrap acquire in `try / except`. This way we could avoid nasty if below.

manasdk: try...except...finally would make this read a bit better. Since it somewhat stylistic your call.

manasdk:
:+1:

PR #2402, opened 2016-01-19
Rework post result

manasdk:
Instead of passing in an execution object gets the execution object in the action
to drop the dependency on the st2 pack from st2contrib.


Kami:
That's what I suggested in the past when we did "pass in execution object instead of id" change so I'm +1 for this approach.

contrib/chatops/actions/format_execution_result.py
```diff
@@ -42,3 +43,12 @@ def run(self, execution):
                     template = alias.result['format']
 
         return self.jinja.from_string(template).render(context)
+
+    def _get_execution(self, execution_id):
+        if not execution_id:
+            raise ValueError('Invalid execution_id provided.')
+        execution = self.client.liveactions.get_by_id(id=execution_id)
```
Kami: So is it retrieving an execution or live action?

This stuff still is confusing - either client manager name is incorrect or the method, argument, etc. here is incorrect.

Kami: I checked the code, looks like the client name is wrong.

Gosh, this is so annoying and confusing. Maybe to make it less confusing let's create new `Execution` resource class in `st2client/models/action|execution` and let's have `LiveAction` inherit from it and leave it there for backward compatibility reasons.

manasdk: I am ont touching client code right now. Certainly not in this PR.

PR #2404, opened 2016-01-20
Get rid of mutually exclusive args in re-run

manasdk:
Get rid of mutually exclusive args since positional parameters cannot exist in mutually exclusives.

Doing so also means that a user can provide `--task` and `--parameters` together. The API guards against this condition by throwing back an error so that nothing bad actually happens.

Perhaps a better approach would be to have a standalone `st2 execution re-run-workflow --tasks <task-name>` however that is a much bigger change and could be handled later.

In case parameters and tasks are mixed -
```
$ st2 execution re-run 569f2c7932ed352e386fc284 cmd=hostname --tasks y
MESSAGE: Parameters override is not supported when re-running task(s) for a workflow. for url: http://127.0.0.1:9101/v1/executions/569f2c7932ed352e386fc284/re_run
```

### DIFF
--- a/st2client/st2client/commands/action.py
+++ b/st2client/st2client/commands/action.py
@@ -1050,14 +1050,11 @@ class ActionExecutionReRunCommand(ActionRunCommandMixin, resource.ResourceComman
         self.parser.add_argument('id', nargs='?',
                                  metavar='id',
                                  help='ID of action execution to re-run ')
-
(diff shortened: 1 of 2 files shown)


Kami:
> ERROR: 500 Server Error: Internal Server Error

I will look into this and fix it - we need to return 400 in this case since it's a validation error.

m4dcoder:
+1

PR #2411, opened 2016-01-22
Add support to handle re-run of with-items task for mistral workflow

m4dcoder:
Add option to st2 execution re-run to either re-run all or only failed iterations for with-items task for mistral workflow.

### DIFF
--- a/st2client/st2client/commands/action.py
+++ b/st2client/st2client/commands/action.py
@@ -1055,6 +1055,11 @@ class ActionExecutionReRunCommand(ActionRunCommandMixin, resource.ResourceComman
                                       'and optional args for the action.')
         self.parser.add_argument('--tasks', nargs='*',
                                  help='Name of the workflow tasks to re-run.')
+        self.parser.add_argument('--no-reset', dest='no_reset', nargs='*',
+                                 help='Name of the with-items tasks to not reset. This only '
+                                      'applies to Mistral workflows. By default, all iterations '
+                                      'for with-items tasks is rerun. If no reset, only failed '
+                                      ' iterations are rerun.')
         self.parser.add_argument('-a', '--async',
                                  action='store_true', dest='async',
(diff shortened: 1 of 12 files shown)


st2client/st2client/commands/action.py
```diff
@@ -1055,6 +1055,11 @@ def __init__(self, resource, *args, **kwargs):
                                       'and optional args for the action.')
         self.parser.add_argument('--tasks', nargs='*',
                                  help='Name of the workflow tasks to re-run.')
+        self.parser.add_argument('--no-reset', dest='no_reset', nargs='*',
```
dzimine: `--no-reset` option is hard to grok comparing to `--reset`. I understand where you're coming from with constraints of default must be "all rerun", but please think again how to phrase / expose it. May be still have it `--reset` and default is 
"all reset" ... but it's a strange default... or call it "--rerun-failed-only"... I don't have solution yet, I'll think about it and please you, too; if you can come out for something, bring it up.

dzimine:
Good enough, one comment inline but I don't have a better answer yet, if we don't come up with such - fix it in a patch when you switch default behavior.

st2client/st2client/commands/action.py
m4dcoder: I'm willing to experiment here. The current option here take a list of task names to not reset. This is to accommodate use cases where there are multiple with-items tasks and user don't want to reset a subset of them.  A global flag would limit it.

m4dcoder:
@dzimine, just so we are on the same page. v1.3 release is already started by @manasdk. I was told this PR will be included in the next patch release.

manasdk:
If this is already working feel free to merge. The release as you know in a bit of trouble :)

manasdk:
Does this require mistral fixes? I can re-run mistral release if thats the case. LMK.

PR #2417, opened 2016-01-23
Look at first 20 lines

manasdk:
* inclusion of parameters in the output shifted the status field lower


manasdk:
Not great we should have a deterministic way to figure out status of the action. This is pretty bad.

manasdk:
Improved by not saving the result around. That is anyway in the execution list.

This should be more resilient to changing parameters etc.

PR #2418, opened 2016-01-25
Open files in read-only mode in packviews

enykeev:


Kami:
That's what we originally did, but I believe @manasdk had a specific reason to also open in write mode.

Apparently if you don't open it in write mode you can't read metadata (modified time, etc.). I personally haven't verified / tested that.

enykeev:
We can revert it if that's the case, but I want us to hit and investigate the problem once again since it's important for https://github.com/StackStorm/st2-packages/issues/99. If write permissions are required, we don't really have a choice.

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1454449537004428

PR #2419, opened 2016-01-25
Replace uwsgi with gunicorn in st2auth

enykeev:


Kami:
Yay.

Does this mean you figured out proxying remote user header, etc.?

enykeev:
I figured we don't need to proxy it since October when we stopped using web server for auth and implemented auth modules as separate modules. Purely communication issue, I guess.

enykeev:
Well, I guess we're safe to merge that since we only depend on uwsgi being present as requirement for new packages.

PR #2421, opened 2016-01-25
Use virtualenv binary from "/usr/local/bin/" by default

Kami:
We now install a newer version of virtualenv and binary gets installed into `/usr/local/bin/` by default.

The problem is that we use system Python binary from `/usr/bin` which means we will also try to use virtualenv from `/usr/bin` which is an outdated version which is causing tons of issues right now.

Depends on #2422.

### DIFF
--- a/st2actions/st2actions/config.py
+++ b/st2actions/st2actions/config.py
@@ -44,7 +44,19 @@ def _register_common_opts():
 def _register_action_runner_opts():
     default_python_bin_path = sys.executable
     base_dir = os.path.dirname(os.path.realpath(default_python_bin_path))
-    default_virtualenv_bin_path = os.path.join(base_dir, 'virtualenv')
+
+    custom_virtualenv_binary = '/usr/local/bin/virtualenv'
+    standard_virtualenv_binary = os.path.join(base_dir, 'virtualenv')
+
+    # We want to use a virtualenv binary which we install - that binary gets
+    # installed into /usr/local/bin and not /usr/bin
+    # Version inside /usr/bin ships with system and this is usually a very
+    # outdated version
+    if os.path.exists(custom_virtualenv_binary):
+        default_virtualenv_bin_path = custom_virtualenv_binary
+    else:
+        default_virtualenv_bin_path = standard_virtualenv_binary
+
     logging_opts = [
         cfg.StrOpt('logging', default='conf/logging.conf',
                    help='location of the logging.conf file'),


st2stanley:
This seems non-ideal. What happens if/when someone does not have virtualenv in
/use/local/bin
Would changing PATH not work?

-James
On Mon, Jan 25, 2016 at 5:37 AM, Tomaz Muraus <notifications@github.com> wrote:
We now install a newer version of virtualenv and binary gets installed into /usr/local/bin/ by default.

The problem is that we use system Python binary from /usr/bin which means we will also try to use virtualenv from /usr/bin which is an outdated version which is causing tons of issues right now.


--------------------------------------------------------------------------------

YOU CAN VIEW, COMMENT ON, OR MERGE THIS PULL REQUEST ONLINE AT:
[https://github.com/StackStorm/st2/pull/2421] https://github.com/StackStorm/st2/pull/2421
[https://github.com/StackStorm/st2/pull/2421]

COMMIT SUMMARY
 * Use virtualenv binary from "/usr/local/bin/" by default if available.

FILE CHANGES
 * M st2actions/st2actions/config.py
   [https://github.com/StackStorm/st2/pull/2421/files#diff-0] (14)

PATCH LINKS:
 * [https://github.com/StackStorm/st2/pull/2421.patch] https://github.com/StackStorm/st2/pull/2421.patch
   [https://github.com/StackStorm/st2/pull/2421.patch]
 * [https://github.com/StackStorm/st2/pull/2421.diff] https://github.com/StackStorm/st2/pull/2421.diff
   [https://github.com/StackStorm/st2/pull/2421.diff]

Kami:
@jfryman The deeper underlying problem is that we don't use the same version of virtualenv across all the operating system.

If we get all of my proposed changes merged in, we will explicitly install a specific version of virtualenv using pip (instead of relying on messy and inconsistent state of system packages) and that version will be installed into `/usr/local/bin. That way we will also want to use the same virtualenv version everywhere, namely the one from `/usr/local/bin`.

st2stanley:
Maybe so, but I have to think there is a better way to address this.
Either way, 0.02 provides.

-James
On Mon, Jan 25, 2016 at 8:09 AM, Tomaz Muraus <notifications@github.com> wrote:
@jfryman [https://github.com/jfryman] The deeper underlying problem is that we don't use the same version of
virtualenv across all the operating system.

If we get all of my proposed changes merged in, we will explicitly install a
specific version of virtualenv using pip (instead of relying on messy and
inconsistent state of system packages) and that version will be installed into /usr/local/bin. That way we will also want to use the same virtualenv version
everywhere, namely the one from /usr/local/bin`.

—
Reply to this email directly or view it on GitHub
[https://github.com/StackStorm/st2/pull/2421#issuecomment-174518678] .[https://github.com/notifications/beacon/AHchNmg6JM_NZyQ2U7KKuku7LYfQYJPAks5pdiQdgaJpZM4HLiWq.gif]

Kami:
As far as PATH goes - in theory it might work, but the whole "might" and not being consistent enough across different platforms is one of the problems we are supposed to solve :)

If we used PATH and user had virtualenv also installed from system packages and `/usr/bin` had precedence over `/usr/local/bin` in PATH, it would end up using a wrong version - something we don't want.

st2actions/st2actions/config.py
```diff
@@ -44,7 +44,19 @@ def _register_common_opts():
 def _register_action_runner_opts():
     default_python_bin_path = sys.executable
     base_dir = os.path.dirname(os.path.realpath(default_python_bin_path))
-    default_virtualenv_bin_path = os.path.join(base_dir, 'virtualenv')
+
+    custom_virtualenv_binary = '/usr/local/bin/virtualenv'
```
Kami: Also, yes, I'm aware this won't work on Windows (well, it would actually fall back to else clause so it might work), but that's not our problem, because:

1. We currently don't support Windows as a host operating system for StackStorm and I don't think we will do it any time soon either (we already have tons and tons of issues with getting stuff to work consistently and reliably across supported Linux distros so supporting yet another totally different OS is nothing but wishful thinking)
2. Even if we ever supported Windows, there are many other things we would need to change before this would become a problem (e.g. a lot of things assume "/opt/stackstorm" path right now, etc.).

lakshmi-kannan:
Closing.Reopen if you need it.

Kami:
Yeah, that's not needed anymore with the new packages.

PR #2422, opened 2016-01-25
Bundle new version of virtualenv with st2

Kami:
This is a st2 part of those changes - https://github.com/StackStorm/puppet-st2/pull/155


Kami:
It's also worth noting that this change is needed by new packages.

It was originally submitted by @dennybaa, but I reverted it since it horribly broke st2workroom and other things. The problem is that this change by itself won't work and requires other stackstorm-st2 and st2workroom changes.

Kami:
This is not needed anymore with the new packages, closing.

PR #2423, opened 2016-01-25
Manually specify dependencies in st2client setup.py

Kami:
dist_utils relies on some functionality which is only available in newer versions of pip. Right now there is a bug / issue where old version of pip is used on packs virtual environments on Ubuntu.

This is a temporary work around until we ensure that a more recent version of virtualenv is installed everywhere where st2client is installed (system wide and virtualenv used by packs).

### DIFF
--- a/st2client/setup.py
+++ b/st2client/setup.py
@@ -18,17 +18,20 @@ import os.path
 
 from setuptools import setup, find_packages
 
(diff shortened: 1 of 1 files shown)


Kami:
Also backported into v1.3.1 branch in #2424.

PR #2435, opened 2016-01-29
Handle exception if trigger registration fails

lakshmi-kannan:
It looks like if there is a conflict in trigger type registration, we simply bubbled up the exception to the point the API process shut down. Now we catch the exception, log as a warning and move on. 

This should address intermittent build failures in circle ci such as https://circleci.com/gh/StackStorm/st2-packages/310

### DIFF
--- a/st2common/st2common/triggers.py
+++ b/st2common/st2common/triggers.py
@@ -71,8 +71,13 @@ def register_internal_trigger_types():
             is_action_trigger = trigger_definition['name'] == ACTION_SENSOR_TRIGGER['name']
             if is_action_trigger and not action_sensor_enabled:
                 continue
-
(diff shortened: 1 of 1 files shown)


manasdk:
+1

Kami:
It seems like internal trigger registration failure should indeed be a fatal error and should bubble up.

Only time it shouldn't be a fatal error is if we get duplicate error or similar due to a race - we should explicitly catch this exception thought and bubble up others.

If internal triggers are not registered, this could cause all kinds of issues.

lakshmi-kannan:
>Only time it shouldn't be a fatal error is if we get duplicate error or similar due to a race - we should explicitly catch this exception thought and bubble up others.

You are right. We shouldn't ignore all exceptions. We were qucikly testing this out to see if it stops API process from restarting under gunicorn.

PR #2436, opened 2016-01-29
Allow user to specify a boolean for a "cacert" argument value

Kami:
Right now we only allow user to pass a string (path to the CA bundle) as a value for the `cacert` argument.

`requests` also allows user to pass in a boolean, but we don't support this use case right now. If boolean `Rrue` is passed for the `verify` argument requests will use bundled CA bundle or CA bundle from the system store (http://docs.python-requests.org/en/latest/user/advanced/#ca-certificates).

I know that the argument is named `cacert` so also allowing user to pass in boolean values makes for a bit of a confusing API, but I don't have many other options right now which don't require bigger refactoring.

Related to https://github.com/StackStorm/st2contrib/pull/376

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,10 @@ Changelog
 in development
 --------------
 
+* Allow user to pass a boolean value for the ``cacert`` st2client constructor argument. This way
+  it now mimics the behavior of the ``verify`` argument of the ``requests.request`` method.
+  (improvement)
+
 1.3.0 - January 22, 2016
 ------------------------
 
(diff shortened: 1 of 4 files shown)


Kami:
Another use case we also don't support right now which `requests` supports is passing in a list of strings - list of paths to CA bundles.

If other people think we should also support that, I can also make that change (it's simple enough). If we also support list / iterable values `cacert` argument will basically be equivalent to `requests` `verify` argument.

manasdk:
WFM. We can add support for list of paths to CA bundles as required. This should do for now.

Kami:
I'll go ahead and merge this.

I will only merge the st2contrib PR when a new version of st2client is released and published to PyPi.

PR #2437, opened 2016-01-30
RFR: Gunicorn config fixes

lakshmi-kannan:
## What got fixed? 

* The way gunicorn_config.py was written for API and AUTH, both master and workers would connect to db, register trigger types, ... etc. This results in the duplication of trigger registration which originally threw an exception and got fixed in https://github.com/StackStorm/st2/pull/2435. So we now know the root cause why there was duplication. With this PR, gunicorn_config.py doesn't call common_setup. This means master process won't set up db connections, connect to rabbit etc. We also found out that gunicorn_pecan calls into module.app.setup_app for each worker. So we moved common_setup code to there. In case of standalone http, common_setup is not run more than once. 

* With more than 1 worker for API, we were seeing queue lock issues like so:
```
ResourceLocked: Queue.declare: (405) RESOURCE_LOCKED - cannot obtain exclusive access to locked queue 'st2.trigger.watch.WebhooksController' in vhost ‘/‘ in the second process as soon as i start up
```
Manas already knew about this and he proposed using a different queue_suffix for each trigger watcher. exclusive=True should not be changed. We did that and I was able to spin more than 1 worker process just fine. 

I still don't recommend running more than 1 worker with eventlet. If need be, we should run more than one gunicorn_pecan process and take care of port issues. 

* Also with this PR, I set the default mode for API and AUTH to use gunicorn in dev env to catch issues. 
* Circular dependency between logging setup before config parsing depending on cfg.CONF variables

## What remains a problem? 

* Though gunicorn allows us to specify service hooks to catch signals, when I kill the gunicorn_process, eventlet pukes. I couldn't figure out why. So if some day we want some cleanup code to run for each worker, we got to fix this. 

## TODO
* [X] - API tests ** FIXED NOW.
* [x] - Changelog
* [x] - Limit queue name length
* [ ] - Integration tests for trigger watcher queue consumption **NOT DONE [See https://github.com/StackStorm/st2/pull/2437#issuecomment-178300317]

Luckily our rspec tests in new packages are catching this issue. Maybe we can add a test case for 2 API workers there? That would be our integration tests.  

Thanks to @manasdk for all the help with pecan and gunicorn debugging. We did a lot of group therapy aka pair programming today.


st2api/st2api/controllers/v1/webhooks.py
```diff
@@ -51,11 +51,12 @@ def __init__(self, *args, **kwargs):
         self._trigger_types = WEBHOOK_TRIGGER_TYPES.keys()
 
         self._trigger_dispatcher = TriggerDispatcher(LOG)
+        queue_suffix = self.__class__.__name__ + ':' + str(uuid.uuid4())
         self._trigger_watcher = TriggerWatcher(create_handler=self._handle_create_trigger,
                                                update_handler=self._handle_update_trigger,
                                                delete_handler=self._handle_delete_trigger,
                                                trigger_types=self._trigger_types,
-                                               queue_suffix=self.__class__.__name__,
```
Kami: Good catch :+1: 

We still use a topic exchange here so the messages should be delivered in round-robin fashion so that's fine, right?

st2api/st2api/controllers/v1/webhooks.py
```diff
@@ -51,11 +51,12 @@ def __init__(self, *args, **kwargs):
         self._trigger_types = WEBHOOK_TRIGGER_TYPES.keys()
 
         self._trigger_dispatcher = TriggerDispatcher(LOG)
+        queue_suffix = self.__class__.__name__ + ':' + str(uuid.uuid4())
```
Kami: The name might be too long, in get_queue function we actually truncate the uuid. Probably a good idea to do that here as well.

```python
    def _get_queue(queue_suffix, exclusive):
        if not queue_suffix:
            # pick last 10 digits of uuid. Arbitrary but unique enough for the TriggerWatcher.
            u_hex = uuid.uuid4().hex
            queue_suffix = uuid.uuid4().hex[len(u_hex) - 10:]
        queue_name = 'st2.trigger.watch.%s' % queue_suffix
```

Kami: Another minor refactoring thing - we could potentially refactor that function and add another argument `append_random_id` or similar and leverage that.

st2api/st2api/controllers/v1/webhooks.py
Kami: On a related note - we need some integration tests for that.

I've added some in the past (st2api/tests/integration), but they don't run on Travis since I could never get it to work there :/

manasdk: > We still use a topic exchange here so the messages should be delivered in round-robin fashion so that's fine, right?

From docs -
"The logic behind the topic exchange is similar to a direct one - a message sent with a particular routing key will be delivered to all the queues that are bound with a matching binding key."

Therefore, there will be no round-robin. This is the right behavior since we want each WebHookController to recognize any webhook.

st2api/st2api/controllers/v1/webhooks.py
manasdk: I have not been able to find a restriction of queue name length. Likely that anything more than 256 would be rejected by rabbitmq server.

On the other hand and Q name more than 50 chars makes it harder to grok. I am +1 to some truncation :)

st2api/st2api/controllers/v1/webhooks.py
Kami: Yeah, good point (that's the case if queue names are different).

If we run multiple API servers we indeed want all of them to process that messages so user can hit any API server.

st2api/st2api/controllers/v1/webhooks.py
lakshmi-kannan: Fixed in 6a08c3a04c42286a690dfc0ece5a6fa822658764

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1454373548004018

lakshmi-kannan:
@Kami I tried writing an integration test for spinning multiple trigger watchers but unfortunately the test is highly unreliable. It might have something to do with eventlets and our queue consumers. I reverted that commit. I'll look into this later and I don't want to hold up this PR because this blocks packaging. cc: @manasdk

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1454377944004024

tools/launchdev.sh
```diff
@@ -16,7 +16,7 @@ while getopts ":r:gxc" o; do
             runner_count=${OPTARG}
             ;;
         g)
-            use_gunicorn=true
+            use_gunicorn=false
```
manasdk: lol. Fine :P

manasdk:
+1

Validated locally.

st2common/st2common/log.py
```diff
@@ -118,18 +117,18 @@ def _audit(logger, msg, *args, **kwargs):
...
 
 def _redirect_stderr():
     # It is ok to redirect stderr as none of the st2 handlers write to stderr.
-    if cfg.CONF.log.redirect_stderr:
-        sys.stderr = LoggingStream('STDERR')
+    sys.stderr = LoggingStream('STDERR')
 
 
-def setup(config_file, disable_existing_loggers=False):
+def setup(config_file, redirect_stderr=True, excludes=None, disable_existing_loggers=False):
```
Kami: Hm, I don't see `excludes` argument being used. Or am I missing something?

st2common/st2common/log.py
```diff
@@ -139,7 +138,8 @@ def setup(config_file, disable_existing_loggers=False):
                                   disable_existing_loggers=disable_existing_loggers)
         handlers = logging.getLoggerClass().manager.root.handlers
         _add_exclusion_filters(handlers)
```
Kami: Looks like `excludes` should be passed to this function.

Using keyword arguments would also be a good idea :P

st2common/st2common/log.py
Kami: Docstring which documents redirect_stderr, excludes, etc. would be nice.

st2api/st2api/controllers/v1/webhooks.py
```diff
@@ -51,11 +51,12 @@ def __init__(self, *args, **kwargs):
         self._trigger_types = WEBHOOK_TRIGGER_TYPES.keys()
 
         self._trigger_dispatcher = TriggerDispatcher(LOG)
+        queue_suffix = self.__class__.__name__
```
Kami: Hm, a bit confused. Did you change back to the old behavior, or am I missing something?

Kami: Ah, I see, you now always add random suffix in sensor_watcher.py.

st2common/st2common/services/sensor_watcher.py
```diff
@@ -103,9 +103,8 @@ def stop(self):
 
     @staticmethod
     def _get_queue(queue_suffix):
-        if not queue_suffix:
-            # pick last 10 digits of uuid. Arbitrary but unique enough for the TriggerWatcher.
-            u_hex = uuid.uuid4().hex
-            queue_suffix = uuid.uuid4().hex[len(u_hex) - 10:]
-        queue_name = 'st2.sensor.watch.%s' % queue_suffix
+        queue_name = queue_utils.get_queue_name(queue_name_base='st2.sensor.watch',
```
Kami: Probably a good idea to add a docstring to this function which clarifies that we always add a random suffix to the queue name and why.

st2common/st2common/triggers.py
```diff
@@ -47,9 +47,8 @@ def _register_internal_trigger_type(trigger_definition):
                           trigger_type_db.get_reference().ref)
             raise
         except StackStormDBObjectConflictError:
-            LOG.exception('Shadow trigger creation of "%s" failed with uniqueness conflict.',
-                          trigger_type_db.get_reference().ref)
-            raise
+            LOG.debug('Shadow trigger "%s" already exists. Ignoring.',
```
Kami: :+1:

PR #2444, opened 2016-02-02
Register oslo opts in uwsgi config for auth

lakshmi-kannan:
By trying to fix a nasty import level registration in https://github.com/StackStorm/st2/pull/2437, I accidentally broke wsgi method for spinning auth. This PR fixes that. 

I also make changes to launchdev so uwsgi auth can be spun up for quick testing. With auth enabled, I am able to get token just fine.

### DIFF
--- a/st2auth/st2auth/wsgi.py
+++ b/st2auth/st2auth/wsgi.py
@@ -17,6 +17,7 @@ from pecan import load_app
 from oslo_config import cfg
 
 from st2auth import config  # noqa
+config.register_opts()
 from st2common import log as logging
 from st2common.persistence import db_init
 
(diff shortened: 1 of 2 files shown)


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1454391332004067

lakshmi-kannan:
Also, tested the fix on a workroom box and verified auth process starts up.

Kami:
Yeah, wish fixing all those import level side affects was so easy :/

If we wanted to fix / change all of those it would require use to refactor quite a bit of code and have single entry point where stuff is initialized and parsed. And we would of course need to ensure this initialization code is always called on start up.

PR #2445, opened 2016-02-02
1.3.1 fix: Run some st2ctl commands without sudo

dzimine:
Answering https://github.com/StackStorm/st2-packages/issues/134 - that's what I mean.

### DIFF
--- a/st2common/bin/st2ctl
+++ b/st2common/bin/st2ctl
@@ -1,9 +1,5 @@
 #!/bin/bash
 
-if [ $(id -u) -ne 0 ]; then
-    echo "Please run with root privileges"
-    exit 1
-fi
 
 LOGFILE="/dev/null"
 COMPONENTS="st2actionrunner st2api st2auth st2garbagecollector st2notifier st2resultstracker st2rulesengine st2sensorcontainer st2web mistral"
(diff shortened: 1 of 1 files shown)


Kami:
Which "tasks" don't require sudo / root, just `getpids` or something else as well?

IIRC, it's usually a good idea to use sudo for process listing as well - it used to be a good "hardening" practice to disallows users see processes of other users (I always enabled this on FreeBSD, but IIRC, Linux also supports that).

armab:
Part of https://github.com/StackStorm/st2-packages/issues/134

manasdk:
:+1:

manasdk:
Is it really needed? Its “only” a better error message.

I am all for adding to master btw for next major release but unsure on getting this into v1.3 where we should have changes that are truly necessary for patch.

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1454447308004378

DoriftoShoes:
I am +1 to explicitly calling out mistral-api and mistral in st2ctl status.

dzimine:
Ok, I'll keep `mistal-api` from what I see it's no harm to start/stop it explicitly, and benefit of seeing if api is up or down.

PR #2451, opened 2016-02-03
To avoid 1.3.2 package creation remove circle.yaml

manasdk:


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1454461004004504

enykeev:
Instead of removing the whole file, can we just modify `deployment.branch` to exclude patch version branches?

manasdk:
> Instead of removing the whole file, can we just modify deployment.branch to exclude patch version branches?

Would fixing the regex on 
https://github.com/StackStorm/st2/pull/2451/files#diff-29944324a3cbf9f4bd0162dfe3975d88L76
work?

Seems odd to have deplotment.branch in every circle.yml. Is the master one the only one that matters?

enykeev:
I assume it would.

And if I understood the second part correctly, you have to have a circle.yml file per repo because circleci expects it to be there and you have to have deployment section inside it because you want changes in other repos to also produce new packages. Like, changes in st2-packages may alter the behavior of the package and you want a new version of it without creating a meaningless PR in st2.

manasdk:
Per this https://circleci.com/docs/configuration#deployment section seems like it kicks in post build.

armab:
Better remove if you don't need builds for this version at all and `1.3.2` packages to be created.
However I don't really understand why you don't need `1.3.2` packages for 1.3.2 version.

Anyways, :+1: for delete, do it.

enykeev:
Ignore it I'm stating the obvious, but we need to make sure this file will stay in place when/if we decide to merge this branch somewhere. If we accidentally remove the file in the branch that should go through CI, there is no easy way to spot it. Saying that, do what you think is right.

armab:
You're right, after discussing/understanding how patched branches work: 
https://stackstorm.slack.com/archives/packages/p1454517758002810

These PR should solve issue with creating packages from "patch branch":
https://github.com/StackStorm/st2/pull/2453 (for `master` for future)
https://github.com/StackStorm/st2/pull/2455 (for `v1.3.2` as hotfix)

manasdk:
> If we accidentally remove the file in the branch that should go through CI, there is no easy way to spot it.

This is absolutely a concern. Seems like we may need a control point outside somewhere.

DoriftoShoes:
> This is absolutely a concern. Seems like we may need a control point outside somewhere.
Seems like a good use case for monitoring.  A script on the sensu box could easily do this.  Just need a list of all the branches we care about.

PR #2458, opened 2016-02-03
Call db_setup after config is parsed

manasdk:
* calling db_setup prior to parsing config means that default values of
  connection settings are used. These default value are usually correct
  in dev environments making it hard to catch the issue locally.
  Luckily we had a quirky enough CI which broke the assumption and picked
  up the failure.

### DIFF
--- a/st2actions/st2actions/runners/python_action_wrapper.py
+++ b/st2actions/st2actions/runners/python_action_wrapper.py
@@ -49,7 +49,6 @@ class PythonActionWrapper(object):
         :param parent_args: Command line arguments passed to the parent process.
         :type parse_args: ``list``
         """
-        db_setup()
 
         self._pack = pack
         self._file_path = file_path
@@ -62,6 +61,8 @@ class PythonActionWrapper(object):
             config.parse_args(args=self._parent_args)
         except Exception:
             pass
+        else:
+            db_setup()
 
     def run(self):
         action = self._get_action_instance()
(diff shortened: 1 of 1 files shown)


Kami:
Good catch :+1:

st2actions/st2actions/runners/python_action_wrapper.py
```diff
@@ -62,6 +61,8 @@ def __init__(self, pack, file_path, parameters=None, parent_args=None):
             config.parse_args(args=self._parent_args)
         except Exception:
             pass
+        else:
+            db_setup()
```
Kami: It's slightly weird that this is in the else block.

We probably also want to establish a connection if `config_parse_args()` fails since we ignore the exceptions right now and proceed normally even if it fails. Otherwise some weird issues will probably pop up due to connection not being established.

manasdk: @lakshmi-kannan and I discussed this offline as well. 

I did not much like this approach either and would prefer to better handle the config. I have left a task for myself to write tests for this class where I can take a closer look and fix this logic as well. I feel failures to parse config should now be fail-fast.

Kami: The original reason, I believe, why I wrapped it in try / except is that in some cases, config is already parses at that point (don't remember where and why) and if you try to parse it again it will throw.

Obviously this try / except is not the best (ideally we would have a consistent place where a config is parsed and we would never try to do it twice), but our config parsing and bootstrap code is less than ideal so that's not always that easy to do.

PR #2461, opened 2016-02-04
Safe kill

manasdk:
* There is some timing related to exit of gunicorn processes. Seems
  like killing the screen session will kill the gunicorn processes.
  Persumably the osx impl does not kill these chil processes.
  The guard makes it so that if killing the screen process does infact
  cleanup gunicorn then we do not necessarily race.


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1454549423004989

m4dcoder:
+1

manasdk:
And this is how code bloat happens :P

PR #2463, opened 2016-02-04
pin python-mistralclient version

manasdk:


manasdk:
This is actually automated but we have some challenges with the automation. Hopefully manual the last time :)

PR #2465, opened 2016-02-04
Changes to ST2 Packs Pack to remove requirement of Git URLs in aliases and also support CD

jjm:
Improvements to add ChatOPS aliases and supporting actions & workflow to allow installation ST2 packs without supplying Git URLs in chat (or allowing them to be over ridden).

It's also possible to use the workflow to trigger an automation installation of the pack for a set branch. This requires an additional webhook rule or an sensor.


Kami:
Great, thanks for the contribution!

We will look into it shortly :)

contrib/packs/actions/check_pack_auto_deploy.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def run(self, branch, repository):
+        exit_code = 1
+
+        # Set up the results
+        results = {}
+        results['notify_channel'] = None
+        results['deployment_branch'] = None
+
+        if self.config.has_key("repositories"):
```
Kami: Minor formatting thing - to avoid that many levels of nesting and make the code a bit easier to follow, you could do something like that:

```python
repositories = self.config.get('repositories', {})
repository = repositories.get(repository, {})
auto_deploy = repository.get('auto_deployment, False)

if auto_deploy:
....
```

contrib/packs/actions/check_pack_auto_deploy.py
```diff
@@ -0,0 +1,90 @@
...
+                    self.logger.error("Branch %s for %s does not have auto_dep_branch set" % (branch,
+                                                                                              repository) )
+        else:
+            self.logger.error("Unknown repository: %s", repository)
+            exit_code = 1
+        
+        if exit_code == 0:
+            return results
+        else:
+            print json.dumps(results, sort_keys=True)
```
Kami: Just curious - is this result used somewhere and is there a specific reason why you use `print` and not `return`?

In theory you should be able to return a result from the function using a simple return - you can either return JSON or native Python type as long as it only contains simple types (no class instances, only simple types such as dicts, lists, strings, etc.).

contrib/packs/actions/check_pack_auto_deploy.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+        sys.exit(exit_code)
+        
+if __name__ == '__main__':
+
+    parser = argparse.ArgumentParser()
```
Kami: I assume this is done to make testing the script easier? :)

Kami: If you have some time you can also look into our pack testing story - https://docs.stackstorm.com/development/pack_testing.html

It's meant more for automated testing, but nevertheless.

contrib/packs/actions/check_pack_auto_deploy.py
```diff
@@ -0,0 +1,90 @@
...
+import json
+import getopt
+import argparse
+import os
+import yaml
+
+from getpass import getpass
+from st2actions.runners.pythonrunner import Action
+
+class Packs_Check_Pack_Auto_Deploy(Action):
```
Kami: This action made me think a bit - in theory you could also achieve this via rule (no action needed), but then you couldn't utilize the config and you would need to "hard code" the values in the rule.

contrib/packs/actions/check_pack_auto_deploy.py
Kami: Ah, I see, you use `sys.exit` below.

If you explicitly use exit, `return` won't work. If you want an action to fail, you can raise an exception - runner will detect that and automatically mark action as failed.

This way action will also be more standalone and decoupled from StackStorm which means it's easier to test and maintain it.

contrib/packs/actions/pack_to_git_url.py
```diff
@@ -0,0 +1,126 @@
...
+                return None
+
+    def run(self, pack_to_expand):
+        exit_code = 0
+
+        # Set up the results object
+        results = {}
+        results['pack_to_expand'] = pack_to_expand
+
+        if pack_to_expand.count('/') == 0:
```
Kami: I believe we already have a similar functionality in one of the existing packs actions, wonder if anything can be re-used here.

contrib/packs/config.yaml
```diff
@@ -1,2 +1,30 @@
 ---
 pack_group: st2packs
+
+repositories:
+  default:
+    protocol: "ssh"
```
Kami: Would it be too much work to support remote URLs directly? It would be a bit easier for the user.

Kami:
I've added some comments.

Nice work overall :+1: 

It would also be great if you could add some docs to http://github.com/stackstorm/st2docs and explain how to configure and utilize this aliases.

contrib/packs/actions/check_pack_auto_deploy.py
jjm: Sure,

I'll clean that up and change to raise an exception so it cleanly fails.

contrib/packs/actions/check_pack_auto_deploy.py
jjm: That is indeed what I was trying to do, I'll make move this over to a try block and raise an exception.

contrib/packs/actions/check_pack_auto_deploy.py
jjm: Yes, making it work on the command line is normally the first thing I do. Making tests that can be run would be a good improvement and I'll look into doing that.

contrib/packs/actions/check_pack_auto_deploy.py
jjm: Yes, if the config could be used in the action, the workflow or a rule then it would. It would not be needed at all.

contrib/packs/actions/pack_to_git_url.py
jjm: Thanks, I'll have a look around and see what I can see.

contrib/packs/config.yaml
jjm: Good suggestion, it would make things clearer for sure. There was a reason it ended up being like this. However I can't quite remember why... 

I'll look at refactoring this too.

jjm:
Thanks, I'll look at getting the changes done and also do a pull request for st2docs on how to configure.

contrib/packs/config.yaml
jjm: I'm Going this this format, which is much easier to configure:

`st2incubator:
    repo: "https://github.com/StackStorm/st2contrib.git"
    subtree: true

  st2contrib:
    repo: "https://github.com/StackStorm/st2contrib.git"
    subtree: true
`

contrib/packs/actions/pack_to_git_url.py
jon-middleton: Thanks @Kami,

I'm assuming you mean `download.py`? There's is a large warning at the top saying the following:

```Else, be warned you might loose a part of you soul or sanity```

So I followed the advice from the community and did the Git to URL expansion in an action. Also I guess this PR would be the start of removing some of the assumptions in this script.

**Comment by jjm** [comment]
Afternoon, 

That should be everything addressed expect for the testing part. Which I do plan to have a read on however I'm out of time today....

**Comment by jjm** [comment]
The following is an example rule for triggering an auto deployment from an BitBucket server

```---
name: post_receive
pack: bitbucket
description: "Trigger ST2 packs deploy for commit on Git branch"
enabled: true

trigger:
  type: "core.st2.webhook"
  parameters:
    url: "bitbucket_post_receive"

criteria:
  trigger.body:
    pattern: "changesets"
    type: "exists"

action:
  ref: "packs.deploy"
  parameters:
    auto_deploy: true
    repo_name:  "{{trigger.body.repository.name}}"
    branch:     "{{trigger.body.refChanges[0].refId}}"
    packs:      [ "{{trigger.body.repository.name}}" ]
    message:    "{{trigger.body.changesets.get('values')[0].toCommit.message}}"
    author:     "{{trigger.body.changesets.get('values')[0].toCommit.author.name}}"
```
Which I'm planning to submit to the st2contrib repo when we have successfully merged this in set of changes into master.

jjm:
On adding to st2docs, yes I do plan to do this when we have the contents of this PR finalised, also happy to do a write up for the blog too.

Looking over the CI fail not sure if it's related to my changes, as most of the others passed.

Kami:
Just had a look at the tests (https://travis-ci.org/StackStorm/st2/jobs/107455051#L949) - it looks like it's an intermediate issue not related to your changes.

I saw this one happen a couple of times in the past, but I never managed to track down the root cause.

In any case, I will restart the tests and they should pass :)

contrib/packs/actions/auto_deploy_repo.py
```diff
@@ -0,0 +1,68 @@
...
+        super(Auto_Deploy_Repo, self).__init__(config=config)
+
+    def run(self, branch, repo_name):
+        results = {}
+
+        try:
+            results['deployment_branch'] = self.config["repositories"][repo_name]["auto_deployment"]["branch"]
+            results['notify_channel'] = self.config["repositories"][repo_name]["auto_deployment"]["notify_channel"]
+        except KeyError:
+            raise ValueError("No repositories or auto_deployment config for '%s'" % repo_name)
```
Kami: Yeah, that also works :)

Kami:
IIRC, this is also the first core action which uses Mistral and not Action Chain workflow.

I'm personally fine with that since Mistral is integral part of StackStorm and more complex workflows are a better fit for Mistral anyway, but I just want to make sure no one else has any objections.

/cc @manasdk @lakshmi-kannan @m4dcoder

Kami:
@jjm Thanks for the updates.

The changes look good to me (I still need to test them though) and if no one else has any objections, I'll go ahead and merge them next week.

As far as documentation and blog post goes - yes, blog post would be great. You can start a draft in Google Docs (or similar) and send it to us when you are ready and we will review it and publish it :)

Kami:
I also pushed a change (158ff97bf31a2e0c03eee060e61856def801a5ff) which will hopefully help with this occasional test failures.

contrib/packs/actions/expand_repo_name.py
```diff
@@ -0,0 +1,70 @@
...
+        # Set up the results object
+        results = {}
+
+        try:
+            results['repo_url'] = self.config["repositories"][repo_name]["repo"]
+            results['subtree'] = self.config["repositories"][repo_name]["subtree"]
+        except KeyError:
+            raise ValueError("Missing repositories config for '%s'" % repo_name)
+        else:
+            return results
```
jjm: Is going 'return results' the best thing to do here?

This ends up with you needing to use '...results.results.repo_url' when you access the results in mistral or the ChatOPS alias. It would be cleaner if it was just '....results.repo'. Am I missing a trick here? I'm not sure where I got that example from, but the duplication of *results* is really jarring.

jjm:
I seem to have missed out docstings, I'll sort on Tuesday as part of to PR or if you have merged it in I'll raise a new one.

contrib/packs/actions/expand_repo_name.py
emedvedev: Afair there's no easy way to avoid that. It's just a naming issue on our part, because you have `results.stdout`, `results.stderr` and, well, `results.results`. Changing it now will break backward compatibility in a lot of places, user-generated actions/aliases included :)

jjm: Thanks for clearing that up, I was thinking it was my doing with the name of what I was returning. However thinking about it further that would not be the case.

contrib/packs/actions/expand_repo_name.py
```diff
@@ -0,0 +1,70 @@
...
+import json
+import getopt
+import argparse
+import os
+import yaml
+
+from getpass import getpass
+from st2actions.runners.pythonrunner import Action
+
+class expand_repo_name(Action):
```
lakshmi-kannan: Hmm. The class names are not consistent in terms of casing. I personally do not prefer `_` in class names. So `MyAwesomeClass` is my preferred way.

lakshmi-kannan:
Sorry, I am late to the game. I am not 100% sure that this workflow belongs in core st2. This could have been a pack itself `packs_deploy`. We are actually debating how we are going to ship core packs and I personally believe that core packs need to be minimal and include testing. I am just thinking out loud here. It doesn't look like there is much needed to move this to a pack in st2contrib. One potential problem I see if is this is the pack that gets updated and we set auto-deploy to true, we have an interesting circular dependency. 

I do agree this is neat work. Now we just have to figure out where to put it :).

jjm:
Yes, the effects of updating a pack via an action contained in the pack is something I thought may go badly so I've not attempted it so far. 

With this in the Packs pack in contrib means it's not deployable via the workflow (as sub tree needs the pack directory). If it needs tests then I'm happy to supply as testing of packs is on my stack. 

The location is an interesting question, really depends how extra packs will be installed. If this makes it easier to install packs and its in st2cibtrib How will the user overcome the chicken and egg situation. Of getting it installed.

contrib/packs/actions/auto_deploy_repo.py
```diff
@@ -0,0 +1,68 @@
...
+import json
+import getopt
+import argparse
+import os
+import yaml
+
+from getpass import getpass
+from st2actions.runners.pythonrunner import Action
+
+class Auto_Deploy_Repo(Action):
```
manasdk: We try to stick to pep8 as much as possible (https://www.python.org/dev/peps/pep-0008/#class-names)

Auto_Deploy_Repo -> AutoDeployRepo

contrib/packs/actions/auto_deploy_repo.py
```diff
@@ -0,0 +1,68 @@
...
+            results['notify_channel'] = self.config["repositories"][repo_name]["auto_deployment"]["notify_channel"]
+        except KeyError:
+            raise ValueError("No repositories or auto_deployment config for '%s'" % repo_name)
+        else:
+            if branch == "refs/heads/%s" % results['deployment_branch']:
+                return results
+            else:
+                raise ValueError("Branch %s for %s should not be auto deployed" % (branch, repo_name))
+        
+if __name__ == '__main__':
```
manasdk: Is this for testing purposes?

contrib/packs/actions/auto_deploy_repo.yaml
```diff
@@ -0,0 +1,18 @@
+---
+  description: "Check if a given branch in a ST2 Pack's Git repository should be auto deployed"
+  enabled: true
+  entry_point: 'auto_deploy_repo.py'
+  name: "auto_deploy_repo"
```
manasdk: to correctly capture the nature of this action perhaps `check_auto_deploy_repo` would be a better name.

contrib/packs/actions/expand_repo_name.py
manasdk: Suggest  `ExpandRepoName`

contrib/packs/config.yaml
```diff
@@ -1,2 +1,32 @@
...
+  #If you have your own ST2 Pack
+  #my-st2:
+  #  repo: "https://github.com/<my-GitHub-user>/my-st2.git"
+  #  subtree: false
+  #  auto_deployment:
+  #    branch: "master"
+  #    notify_channel: "my-chatops-channel"
+
+  st2incubator:
+    repo: "https://github.com/StackStorm/st2contrib.git"
```
manasdk: incorrect url should be `st2incubator`

manasdk:
This is a great change!

My initial reaction to this change was that it should like be a pack of it own in st2contrib. (Pretty much similar to @lakshmi-kannan). However, as I thought further I feel that this is the right place for the auto_deploy actions. At some point content management will be a first class story with APIs etc in StackStorm and I can see that auto_deploy could be a feature we would implement natively in content-management at which point it would become a core StackStorm feature. Seeing from that lens I feel it does fit with in `packs` pack. The downside right now is that config.yaml for a core pack needs to mucked around with by a user. This is non-ideal but as I see it necessary for the time being.

This definitely needs some docs to help explain how to use the feature etc.

contrib/packs/actions/auto_deploy_repo.py
jjm: Yes, it was fit testing.  Would you like it removed? 

I'm planning to look at the testing frame work very soon. For use in our own packs, so I could migrate to that.

jon-middleton:
That should be all the comments addressed in the latest check in, also docstrings have been added and testing moved out action scripts into the pack tests as suggested by @Kami.

contrib/packs/config.yaml
```diff
@@ -1,2 +1,32 @@
...
+
+  #If you have your own ST2 Pack
+  #my-st2:
+  #  repo: "https://github.com/<my-GitHub-user>/my-st2.git"
+  #  subtree: false
+  #  auto_deployment:
+  #    branch: "master"
+  #    notify_channel: "my-chatops-channel"
+
+  st2incubator:
```
Kami: Do other people have any objections to auto deployment being enabled by default for those repos?

I'm personally fine with that since it's not really automatic and still requires user interaction on chat.

/cc @manasdk @lakshmi-kannan

Kami:
Thanks for the updates.

The changes look good to me so I will go ahead and merge them. If we decide we don't want "auto" deployment enabled by default, we can disable it before we roll out the next minor release (1.4.0).

contrib/packs/config.yaml
jjm: I've just removed them as I think they will confuse people and I only used them for testing before I had added unit tests.

The comments on using your own pack include the format and they will be covered in the docs today.

Kami:
Merged, thanks :)

Please follow up with documentation changes at your convenience.

jjm:
Your welcome, I'm working on the docs just now :-D.

jfryman:
This is wonderful! I love to see the small iteration to cut out friction. ❤️

Something to ponder about as you use this daily. Do you need auto-deployment all the time, or just when you're testing or rolling out something new?

The deployment system Heaven we used at GitHub did very much the same thing as what I propose above. Anyone could push any branch to any application. When I wanted to branch deploy to test to a percentage of servers, I would do a deploy...

`!pack deploy super-cool-pack branch=new-feature`

and from *that* point on, until one of two things happened:

* A user manually deployed back to the `master` branch, or
* The user *merged* the branch upstream, and an auto-deployment would happen back to `master`.

There were a few tricks employed as well, such as having a lock so that only the initial user testing an application (in this case, pack) would be able to do additional deploys. Essentially creating a block such that *when* (operative word) pack activity gets heavy you don't end up stepping on each other. Secondly, there was a queuing system, such that if I *was* blocked by the previous step, I'd get queued up for usage. I could check my spot in that line with a ChatOps command, and when any of the above conditions were met with regards to getting back to the `master` branch, the lock would release, and the next user in line would be notified that it's "their turn".

@jjm I didn't want to derail the scope of this PR in any way, so I'm just now dropping these notes. Really interested in your UX with this as you and your teams interact with each other as-is, and what solutions you end up employing with your bottlenecks.

jjm:
Currently its just myself doing Pack deployments, but I'm working towards being able to support all the team being able to work without stepping on each others toes.

The process I'm currently following:

1. Work on a feature branch and deploy when I need to test with:
 `! pack deploy our-awesome-pack our-awesome-pack Feature/latest-addition`
1. Following completion of the work, merge the PR to into dev (which should in time go though CD testing).
1. Manually merge into master which is then auto deployed via a post commit webhook,
1. Announcement posted to the channel which is set the config.

Which is not very far off what you were mentioning (I have a feeling I was using one of your talks as inspiration). I like the sound of locking and overriding, to prevent an roll back of next feature. 

In this initial PR my main focus was to merge in my internal deployment pack and bring it into line with the features of `pack deploy` without the need to supply Git URLs in chat. Which removes the risk of some ChatOPS user installing a malicious pack from a ransom Repo on the internet.

I'll add deployment locking to our backlog as an improvement.

jjm:
The PR to add information to st2docs can be found [here](https://github.com/StackStorm/st2docs/pull/59).

PR #2466, opened 2016-02-04
Disable auth by default in packages

lakshmi-kannan:
** I'll merge the changes to 1.3 branch as well. **

Disable auth by default.
See https://github.com/StackStorm/st2docs/pull/36#discussion_r51930668

PR #2467, opened 2016-02-04
Disable auth by default in packages

lakshmi-kannan:
Cherry picked from https://github.com/StackStorm/st2/pull/2466


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1454628436005120

manasdk:
So no auth when packages are installed? So definitely coming back full circle to st2_deploy?

+1

lakshmi-kannan:
Yep. No auth by default.

PR #2468, opened 2016-02-05
Ngnix rule for mistral_api redirect

lakshmi-kannan:
I was thinking if this is the right place for mistral ngnix configuration. Best would be to have such a thing in mistral fork and ship multiple configurations. I am not a nginx expert. It also looks like st2web configuration is here. Oh well! 

## Test
```
vagrant@pkgtestingbox:~$ http --verify=no https://192.168.1.17/mistral_api/v2/executions
HTTP/1.1 200 OK
Connection: keep-alive
Content-Length: 850
Content-Type: application/json; charset=UTF-8
Date: Fri, 05 Feb 2016 00:39:33 GMT
Front-End-Https: on
Server: nginx/1.4.6 (Ubuntu)
X-Content-Type-Options: nosniff

{
    "executions": [
        {
            "created_at": "2016-02-04 23:19:34.523955",
            "description": "",
            "id": "1c254559-a9ae-4e7e-a98a-e3b20076b495",
            "input": "{\"cmd\": \"date\"}",
            "output": "{\"stdout\": \"Thu Feb  4 15:19:35 PST 2016\"}",
            "params": "{\"env\": {\"st2_liveaction_id\": \"56b3dc855a705f396a644e64\", \"st2_execution_id\": \"56b3dc855a705f396a644e65\", \"__actions\": {\"st2.action\": {\"st2_context\": {\"auth_token\": \"98a0b5cde9ee495eb5bdd314d785c3c4\", \"endpoint\": \"http://0.0.0.0:9101/v1/actionexecutions\", \"notify\": {}, \"parent\": {\"user\": \"stanley\", \"execution_id\": \"56b3dc855a705f396a644e65\"}, \"skip_notify_tasks\": []}}}, \"st2_action_api_url\": \"http://0.0.0.0:9101/v1\"}}",
            "state": "SUCCESS",
            "state_info": null,
            "task_execution_id": null,
            "updated_at": "2016-02-04 23:19:37.085879",
            "workflow_name": "examples.mistral-basic"
        }
    ]
}

vagrant@pkgtestingbox:~$
```

PR #2469, opened 2016-02-05
Ngnix rule for mistral_api redirect

lakshmi-kannan:
Cherry pick https://github.com/StackStorm/st2/pull/2468


lakshmi-kannan:
Apparently we didn't want mistral_api nginx rule intentionally because mistral authorization story isn't figured out yet.

lakshmi-kannan:
Reverting.

PR #2473, opened 2016-02-07
Update webhooks.py to enable form POSTs

drpitman:
replaced lines 84-90 with 84-93 - this enables external webhooks to POST the request body as either application/json or regular web-form-style querystring encoded fields. I'm using this on a number of client servers for the last month and it works fine.

### DIFF
--- a/st2api/st2api/controllers/v1/webhooks.py
+++ b/st2api/st2api/controllers/v1/webhooks.py
@@ -81,13 +81,16 @@ class WebhooksController(RestController):
     def post(self, *args, **kwargs):
         hook = '/'.join(args)  # TODO: There must be a better way to do this.
         body = pecan.request.body
-        try:
+        try: # ADDITION: lines 83-91 to enable form POSTs
             body = json.loads(body)
         except ValueError:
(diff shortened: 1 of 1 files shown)


Kami:
First thanks for the contribution!

I believe I suggested something similar in the past - supporting multiple body formats for the webhook body (there might even be a ticket for it :)).

As far as implementation goes - I would prefer the code to be more strict and explicit and select the corresponding body parser based on the request's `Content-Type` header value (`application/json`, `application/x-www-form-urlencoded` /  `multipart/form-data`). This way we could also support XML and similar if needed.

drpitman:
OK, cool. This suited my immediate purposes, but I'll look at a better implementation along your thoughts there in the near future.

manasdk:
Like to PR. Agree with Kami on using headers to explicitly check for supported content-type and unmarshall accordingly.

Kami:
@drpitman Alright, thanks.

I will go ahead and make the content type changes and merge the PR into master :)

Kami:
@drpitman Alright, I made the changes I have mentioned above in #2477.

Thanks again for your PR.

PR #2474, opened 2016-02-08
Use 127.0.0.1 instead of localhost for consistency [st2]

armab:
> Adresses https://github.com/StackStorm/st2-packages/issues/157

Mixing `localhost` vs `127.0.0.1` in configs can lead to bugs/connection problems with some specific configurations in `/etc/hosts`.

Use `127.0.0.1` where possible and follow that for consistency.

### DIFF
--- a/conf/st2.conf.sample
+++ b/conf/st2.conf.sample
@@ -13,7 +13,7 @@ logging = conf/logging.conf
 
 [api]
 # List of origins allowed
-allow_origin = http://localhost:3000  # comma separated list allowed here.
+allow_origin = http://127.0.0.1:3000  # comma separated list allowed here.
 # location of the logging.conf file
 logging = conf/logging.conf
 # True to mask secrets in API responses
--- a/conf/st2.dev.conf
+++ b/conf/st2.dev.conf
@@ -46,7 +46,7 @@ action_executions_ttl = 30
 trigger_instances_ttl = 30
 
 [syslog]
-host = localhost
+host = 127.0.0.1
 port = 514
 facility = local7
 protocol = udp
@@ -61,7 +61,7 @@ user = stanley
 ssh_key_file = /home/vagrant/.ssh/stanley_rsa
 
 [messaging]
-url = amqp://guest:guest@localhost:5672/
+url = amqp://guest:guest@127.0.0.1:5672/
 
 [ssh_runner]
 remote_dir = /tmp
--- a/conf/st2.package.conf
+++ b/conf/st2.package.conf
@@ -52,7 +52,7 @@ api_url =
 base_path = /opt/stackstorm
 
 [syslog]
-host = localhost
+host = 127.0.0.1
 port = 514
 facility = local7
 protocol = udp
@@ -67,7 +67,7 @@ user = stanley
 ssh_key_file = /home/stanley/.ssh/stanley_rsa
 
 [messaging]
-url = amqp://guest:guest@localhost:5672/
+url = amqp://guest:guest@127.0.0.1:5672/
 
 [ssh_runner]
 remote_dir = /tmp
--- a/st2api/tests/unit/controllers/v1/test_base.py
+++ b/st2api/tests/unit/controllers/v1/test_base.py
@@ -22,7 +22,7 @@ class TestBase(FunctionalTest):
         response = self.app.get('/')
         self.assertEqual(response.status_int, 200)
         self.assertEqual(response.headers['Access-Control-Allow-Origin'],
-                         'http://localhost')
+                         'http://127.0.0.1:3000')
         self.assertEqual(response.headers['Access-Control-Allow-Methods'],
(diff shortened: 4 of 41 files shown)


armab:
@manasdk I changed to `127.0.0.1` it where it looks like required change, for example `conf/st2.package.conf`.

Still I didn't touch `localhost` in all test files. Do you think it's OK to replace it in tests as well?
See: https://github.com/StackStorm/st2/search?utf8=%E2%9C%93&q=localhost

manasdk:
Did you run `st2` locally in dev environment? I can test the branch and validate if it works.

manasdk:
> Still I didn't touch localhost in all test files. Do you think it's OK to replace it in tests as well?

In theory replacing in tests should be ok although t may lead to test failures which we would need to fix. I would try it out and see if travis allows the PR through.

armab:
Let me try to change everything and run tests.
I'll give you it for review then to double verify.

armab:
@manasdk please take a look.
Looks green.

st2api/tests/unit/controllers/v1/test_base.py
```diff
@@ -22,7 +22,7 @@ def test_defaults(self):
         response = self.app.get('/')
         self.assertEqual(response.status_int, 200)
         self.assertEqual(response.headers['Access-Control-Allow-Origin'],
-                         'http://localhost')
+                         'http://127.0.0.1:3000')
```
armab: This is single place where I changed the logic to make tests happy.
Verify if that's still correct.

manasdk: makes sense

manasdk:
> Left untouched st2_deploy.sh and some cases like: https://github.com/StackStorm/st2/blob/467df5c41cc1e50112052a9a80c0a1e5fef12261/st2client/tests/unit/test_formatters.py#L134

This is ok.

conf/st2.conf.sample
```diff
@@ -13,7 +13,7 @@ logging = conf/logging.conf
 
 [api]
 # List of origins allowed
-allow_origin = http://localhost:3000  # comma separated list allowed here.
+allow_origin = http://127.0.0.1:3000  # comma separated list allowed here.
```
manasdk: @enykeev Do you think this will an issue for st2web?

manasdk:
@armab Some more tests we need to perform prior to merge.

1. Use tools/launchdev.sh to validate that services start
2. Validate with st2web

If time permits I will get to this myself.

manasdk:
>1. Use tools/launchdev.sh to validate that services start

Tested. Works.

>2. Validate with st2web

Unable to get st2web to run. This is likely a local issue and nothing to do with this change.

conf/st2.dev.conf
```diff
@@ -46,7 +46,7 @@ action_executions_ttl = 30
 trigger_instances_ttl = 30
 
 [syslog]
-host = localhost
+host = 127.0.0.1
```
dennybaa: We can not be sure that other services will be listening on `127.0.0.1`, same as ours which we explicitly bind to this addres.

conf/st2.dev.conf
```diff
@@ -61,7 +61,7 @@ user = stanley
 ssh_key_file = /home/vagrant/.ssh/stanley_rsa
 
 [messaging]
-url = amqp://guest:guest@localhost:5672/
+url = amqp://guest:guest@127.0.0.1:5672/
```
dennybaa: same ^^

conf/st2.package.conf
```diff
@@ -52,7 +52,7 @@ api_url =
 base_path = /opt/stackstorm
 
 [syslog]
-host = localhost
+host = 127.0.0.1
```
dennybaa: We can not be sure about other services. Ours we explicitly bind to this address.

Kami:
I'm +1 on the "127.0.0.1" change overall since I know using localhost could cause all kinds of issue because of IPv6 and similar.

conf/st2.package.conf
dennybaa: rabbitmq I cheked by default is also there. So it makes sence to check also: mongo, postgres and rsyslog. Then it will be clear. Overall +1 for these changes.

manasdk: These are all configurable values. If something does not bind to 127.0.0.1 then change config. Since this is install specific not much we can do here - this is being done by AIO.

armab:
> Tested. Works.

manasdk:
Good enough I think we should merge. If there is a fallout we can fix/cleanup.

armab:
fallout_4 1

conf/st2.conf.sample
enykeev: Yes, it turned out to be an annoyance and took me about half an hour to figure out what happened.

PR #2477, opened 2016-02-09
Allow webhooks endpoint request body to also be url encoded form data

Kami:
This pull request works on top of #2473 and adds support for determining body type based on the `Content-Type` header value.

Now that we have this code in place we could also easily add support for other Content-Types in the future if needed. (e.g. YAML or XML).

### DIFF
--- a/st2api/st2api/controllers/v1/webhooks.py
+++ b/st2api/st2api/controllers/v1/webhooks.py
@@ -120,7 +120,7 @@ class WebhooksController(RestController):
             self._log_request('Parsing request body as JSON', request=pecan.request)
             body = json.loads(body)
         elif content_type in ['application/x-www-form-urlencoded', 'multipart/form-data']:
-            self._log_request('Parsing reqiest body as form encoded data', request=pecan.request)
+            self._log_request('Parsing request body as form encoded data', request=pecan.request)
             body = urlparse.parse_qs(body)
         else:
             # For backward compatibility reasons, try to parse any other content
--- a/st2api/st2api/controllers/v1/webhooks.py
+++ b/st2api/st2api/controllers/v1/webhooks.py
@@ -88,9 +88,9 @@ class WebhooksController(RestController):
 
         try:
             body = self._parse_request_body(content_type=content_type, body=body)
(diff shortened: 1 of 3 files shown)


st2api/st2api/controllers/v1/webhooks.py
```diff
@@ -110,6 +115,21 @@ def post(self, *args, **kwargs):
 
         return body
 
+    def _parse_request_body(self, content_type, body):
+        if content_type == 'application/json':
+            self._log_request('Parsing body as JSON', request=pecan.request)
+            body = json.loads(body)
+        elif content_type in ['application/x-www-form-urlencoded', 'multipart/form-data']:
+            self._log_request('Parsing body as form encoded data', request=pecan.request)
+            body = urlparse.parse_qs(body)
```
lakshmi-kannan: Interestingly an ill formed form payload doesn't throw. ``` In [3]: urlparse.parse_qs('i am a dumb payload')
Out[3]: {}
```

Comment (Kami) [inline st2api/st2api/controllers/v1/webhooks.py]: Heh, I was actually just thinking about the same thing so I made the change above to capture all the exception types.

File: st2api/st2api/controllers/v1/webhooks.py [hunk 2]
```diff
@@ -110,6 +115,21 @@ def post(self, *args, **kwargs):
...
+        else:
+            # For backward compatibility reasons, try to parse any other content
```
Kami: Actually, now that I think of it - most clients and services should already specify ``application/json`` Content-Type for JSON request body so maybe we can just throw if no or unsupported content type is provided.

lakshmi-kannan: Call me maybe.

Kami: I went ahead and made this change.

If it turns out that for some reasons many of our users indeed do not specify a Content-Type header and this breaks it for them, we can consider reverting this change.

Kami:
Corresponding documentation change - https://github.com/StackStorm/st2docs/pull/57

Issue #2456, opened 2016-02-03
Problem with chatops response formatting


jmound:
There is a problem with the chatops pack's formatting where some newlines are being converted to `@` characters.

For this action:
```
root@server:~# st2 run mypack.run_task  playbook=test.yml
..
id: 56b25bab8259a46e31f981a5
status: succeeded
parameters: 
  playbook: test.yml
result: 
  failed: false
  return_code: 0
  stderr: ''
  stdout: "
PLAY [test] ********************************************************************

TASK [command] *****************************************************************
changed: [10.10.5.71]

PLAY RECAP *********************************************************************
10.10.5.71                 : ok=1    changed=1    unreachable=0    failed=0   
"
  succeeded: true
root@server:~# 
```
`runner.result` is:
```
  "result": {
    "failed": false,
    "stderr": "",
    "return_code": 0,
    "succeeded": true,
    "stdout": "\nPLAY [test] ********************************************************************\n\nTASK [command] *****************************************************************\nchanged: [10.10.5.71]\n\nPLAY RECAP *********************************************************************\n10.10.5.71                 : ok=1    changed=1    unreachable=0    failed=0   \n"
  },
```
However, the chatops response is:
```
Action mypack.run_task completed.
status : succeeded
execution: 56b25b368259a46e31f98199

result :
--------
stdout :       
 PLAY [test] ********************************************************************@@TASK [command] *****************************************************************@changed: [10.10.5.71]@@PLAY RECAP *********************************************************************@10.10.5.71                 : ok=1    changed=1    unreachable=0    failed=0
```

It appears to be caused by this jinja2 macro: [chatops/actions/templates/default.j2#L6](https://github.com/StackStorm/st2/blob/master/contrib/chatops/actions/templates/default.j2#L6). Specifically the `replace('\n', '@')`

Kami:
That's for the report.

I believe we use some special markers / characters for advanced formatting functionality of Slack, but @emedvedev probably has more clue on what is going in and if this behavior is desired or not :)

emedvedev:
No idea, actually. That's @enykeev's template, and I'm pretty sure there was a reason for this but I'm not seeing it. Let's wait for him before removing something that could be important. :)

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1455024630005908

emedvedev:
Merged https://github.com/StackStorm/st2/issues/2479, closing.

PR #2479, opened 2016-02-09
ChatOps pack: don't replace /n with @

emedvedev:
As reported in https://github.com/StackStorm/st2/issues/2456. I have no idea why that replace is even there, so I’ll test for a while and merge if nothing’s broken. If anyone knows anything about this mysterious replace, speak up, but even @enykeev does not. :P


jmound:
+1 As a data point, I've been using a test instance for a week without the replace, with no problems.

emedvedev:
Well, looks fine. Merging for now, and if there's some edge case that really requires replacing newlines, then it's probably wiser to fix it in the alias anyway, not here.

PR #2482, opened 2016-02-10
Include number of elapsed seconds for completed executions in "st2 execution get" CLI command output

Kami:
The title says it all.

This should make it easier to see, at a glance, how long it took for the execution to complete and potentially adjust the action timeouts accordingly.

screenshot from 2016-02-10 18 11 51

### DIFF
--- a/st2client/st2client/commands/action.py
+++ b/st2client/st2client/commands/action.py
@@ -150,21 +150,17 @@ def format_execution_status(instance):
     executions which are in running state and execution total run time for all the executions
     which have finished.
     """
-    if instance.status == LIVEACTION_STATUS_RUNNING and instance.start_timestamp:
+    start_timestamp = getattr(instance, 'start_timestamp', None)
+    end_timestamp = getattr(instance, 'end_timestamp', None)
+
+    if instance.status == LIVEACTION_STATUS_RUNNING and start_timestamp:
         start_timestamp = instance.start_timestamp
         start_timestamp = parse_isotime(start_timestamp)
         start_timestamp = calendar.timegm(start_timestamp.timetuple())
         now = int(time.time())
         elapsed_seconds = (now - start_timestamp)
         instance.status = '%s (%ss elapsed)' % (instance.status, elapsed_seconds)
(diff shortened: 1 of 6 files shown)


st2client/st2client/commands/action.py
```diff
@@ -41,10 +41,24 @@
...
 LIVEACTION_STATUS_RUNNING = 'running'
+LIVEACTION_STATUS_SUCCEEDED = 'succeeded'
+LIVEACTION_STATUS_FAILED = 'failed'
+LIVEACTION_STATUS_TIMED_OUT = 'timeout'
+LIVEACTION_STATUS_ABANDONED = 'abandoned'
 LIVEACTION_STATUS_CANCELING = 'canceling'
 LIVEACTION_STATUS_CANCELED = 'canceled'
 
+
+LIVEACTION_COMPLETED_STATES = [
```
Kami: Needed to copy this over since we still can't have st2client to depend on st2common.

Packages will make that easier though, we just need to make sure we also publish st2common to PyPi.

Kami:
While working on this I believe a noticed a bug with live instances - we don't always set `end_timestamp` for LiveActionDB objects which represent workflow tasks.

I will look into it.

Kami:
Ignore my last comment.

It turns out it was a CLI bug which I have just fixed.

Kami:
I pushed another change so we now also display elapsed time for tasks inside the workflow actions when using `st2 execution get` CLI command.

screenshot from 2016-02-10 19 37 20

manasdk:
:+1:

lakshmi-kannan:
It's funny how the last screenshot shows 13s in one place and 14s in another place :P.

manasdk:
> It's funny how the last screenshot shows 13s in one place and 14s in another place

Isn't that expected? 14s for the parent execution and 13s for the child. I assume rounding errors kicked in since nothing takes 0s to complete.

lakshmi-kannan:
>Unsure if you point that out as a bug of just quirk of rounding errors?

Yeah, I was referring to rounding errors.

Kami:
Yep, it's total execution vs child execution run time + rounding :P

PR #2487, opened 2016-02-10
Sensor Logging Fixes

andrew-regan:
__Updates__
* Make sure the `getLogger()` call does not append more then one `st2.` prefix to the logger name to preserve the logging name hierarchy. This code from `st2reactor/st2reactor/container/sensor_wrapper.py` was creating loggers prefixed with `st2.st2.` because `self._sensor_wrapper._logger.name` is already prefixed with `st2.` and the call to `getLogger` would prefix an additional `st2.`:
```
    def get_logger(self, name):
        """
        Retrieve an instance of a logger to be used by the sensor class.
        """
        logger_name = '%s.%s' % (self._sensor_wrapper._logger.name, name)
        logger = logging.getLogger(logger_name)
        logger.propagate = True

        return logger
```
* Include the pack name in the logger named used by sensors since `self._class_name` does not have to be unique across packs

The current default logging setup shares the `st2sensorcontainer.log` across multiple processes which isn't really suppored. Using these changes its possible to break out individual log files per sensor via the `logging.sensorcontainer.conf`. Example changes:

```
[loggers]
keys=root,vSphereEventSensor

...

[logger_vSphereEventSensor]
level=DEBUG
handlers=consoleHandler, vSphereEventSensorFileHandler, vSphereEventSensorAuditHandler
propagate=0
qualname=st2.SensorWrapper.vsphere.EventSensor

...

[handler_vSphereEventSensorFileHandler]
class=handlers.RotatingFileHandler
level=DEBUG
formatter=verboseConsoleFormatter
args=("logs/vsphere_event_sensor.log",)
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,6 +19,7 @@ in development
   ``Content-Type`` header.
   Note: For backward compatibility reasons we default to JSON if ``Content-Type`` header is
   not provided. #2473 [David Pitman]
+* Bug fixes to allow Sensors to have their own log files. #2487 [Andrew Regan]
 
 1.3.0 - January 22, 2016
 ------------------------
--- a/st2common/st2common/log.py
+++ b/st2common/st2common/log.py
@@ -94,8 +94,14 @@ def decorate_logger_methods(logger):
 
 
 def getLogger(name):
-    logger_name = 'st2.{}'.format(name)
-    logger = logging.getLogger(logger_name)
+    # make sure that prefix isn't appended multiple times to preserve logging name hierarchy
+    prefix = 'st2.'
+    if name.startswith(prefix):
+        logger = logging.getLogger(name)
+    else:
+        logger_name = '{}{}'.format(prefix, name)
+        logger = logging.getLogger(logger_name)
+
     logger = decorate_logger_methods(logger=logger)
     return logger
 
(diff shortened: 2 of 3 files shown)


manasdk:
:+1:

lakshmi-kannan:
I'd appreciate if you could check in a sample logging conf in the repo under conf/ directory. This way people have a working example. We also need some basic documentation in https://github.com/StackStorm/st2docs/blob/master/docs/source/sensors.rst. Also a CHANGELOG entry in https://github.com/StackStorm/st2/blob/master/CHANGELOG.rst.

andrew-regan:
@lakshmi-kannan 

1. @manasdk delivered https://github.com/StackStorm/st2docs/issues/60 earlier which documents this info in Configuration --> Configure Logging. Is this sufficient? Or do you want something in the sensor docs that points back to this?
2. Are the docs good enough to cover the sample? If not do you want the sample in `conf/` or `st2reactor/conf/`? What pack/sensor should the sample refer to?

Thanks,
Andrew

lakshmi-kannan:
@andrew-regan: I am blind. I missed the docs changes. Looks good to me. I was wondering about the sample sensor in conf myself. Maybe when we ship something in core packs, I'll make this change.

estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1455146832006656

andrew-regan:
Sounds good. CHANGELOG.rst updated.

andrew-regan:
Don't believe anything else is needed for this at the moment. Can't access the Slack link above from @estee-tew

Kami:
Thanks! :+1:

PR #2489, opened 2016-02-12
ChatOps: Fixing null optional arguments error

### LINKED ISSUE #2295: action alias regex fails to evaluate null optional arguments when type is not string
For example, something like:  `update {{hostname}}( {{count}} times)?` 
Works for `!update myhost 2 times`
Fails for `!update myhost`
Error: `(invalid literal for int() with base 10: '')`
So the workaround is to change the type of `count` from integer to string.
Or @emedvedev provides another workaround:
```
- update {{ hostname }} {{ count }} times
- update {{ hostname }}
```

Start from the most explicit.

emedvedev:
Fixes #2295

@pixelrebel reported a use case where an argument that’s technically not optional (i.e. doesn’t have a default value) is placed inside an optional regex group:

```
update {{hostname}}( {{count}} times)? 
```

With a string like `update google.com` our parser extracts `{{count}}` from the alias and sets the value to “” because it can’t be matched to anything. While a string like this is something you shouldn’t really do in the first place, it’s pretty easy to introduce support for it, and the use case is legit, so I suggest we go for it.

This is a small fix that doesn’t affect parser logic, so it should be fine as soon as Travis is finished, but I’ll wait for someone else to LGTM this first.

### DIFF
--- a/st2common/st2common/models/utils/action_alias_utils.py
+++ b/st2common/st2common/models/utils/action_alias_utils.py
@@ -81,7 +81,7 @@ class ActionAliasFormatParser(object):
         if matched_stream:
             values = matched_stream.groupdict()
         for param in params:
-            matched_value = values[param[0]] if matched_stream else param[1] or None
+            matched_value = values[param[0]] if matched_stream else param[1]
             if matched_value:
                 result[param[0]] = matched_value
         if extra:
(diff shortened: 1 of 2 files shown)


emedvedev:
Adjusted the spacing unit test to account for the new behavior, too.

manasdk:
+1

Kami:
I'm fine with that, just wondering how ties tie into the whole "no parameter value provided vs None provided for the value" story.

emedvedev:
This case is "no value provided", because the parameter isn't even in the string. If there's a default value, it'll be used instead, as usual.

PR #2503, opened 2016-02-16
Use lint configs from the shared configs repo

Kami:
We've talked about this in the past, now we have all those files copied and duplicated all over the repos.

With this approach, all the lint configs are stored in a single place so it's easier to maintain them.

I have decided to go with subtree instead of submodules since submodules can be a lot of pain. They are not transparent (they require interaction from the user - git submodule init, etc.) and I had many issues with them in the past.

Subtree doesn't require any interaction or extra work from the user (only if you want to update the subtree revision). This does mean files from the subtree repo are duplicated in this one, but this is not an issue in this case.

Repos to update:

- [ ] https://github.com/StackStorm/st2contrib
- [ ] https://github.com/StackStorm/st2docs
- [ ] https://github.com/StackStorm/st2installer
- [ ] https://github.com/StackStorm/st2-auth-backend-pam
- [ ] https://github.com/StackStorm/st2-auth-backend-flat-file
- [ ] https://github.com/StackStorm/st2-auth-backend-keystone
- [ ] https://github.com/StackStorm/st2-auth-backend-mongodb
- [ ] https://github.com/StackStorm/st2-auth-backend-ldap
- [ ] https://github.com/StackStorm/st2-enterprise-auth-backend-ldap/
- [ ] https://github.com/StackStorm/st2sdk


tox.ini
```diff
@@ -33,18 +33,18 @@ commands = {posargs}
 setenv = VIRTUALENV_DIR = {envdir}
```
Kami: This file is not actually used anywhere anymore (afaik).

Any objections to removing it?

Kami:
After I get +1 on this change, I will also go ahead and update other repos.

manasdk:
Any reason you decided to place all lint-configs in `st2` repo and not in a separate `st2-lint-config` repo. I guess YAR can be a pain itself but at least then there would be consistency with our model of moving cross-cutting concerns out to separate repos.

How does this work with release branches etc? Ideally new lint config should not affect old code.

In general I am +1 to the idea of a shared lint configurations.

Kami:
@manasdk They are in a separate repo - https://github.com/StackStorm/lint-config (I just need to fix the typo in the repo name config -> configs).

They are in-line because that's how git subtrees work - http://blogs.atlassian.com/2013/05/alternatives-to-git-submodule-git-subtree/

Configs bundled in the repos don't get updated until `git subtree pull` is ran.

manasdk:
I was confused by seeing this PR in st2 and also the lint-configs folder. Makes sense now

Kami:
Yeah, subtree might be a bit more confusing at first - it's not immediately obvious that the folder is using a git subtree.

With submodules this is immediately obvious since only SHA is stored in the repo, but submodules can be a big pain.

tox.ini
manasdk: Thought this was added to have a way to check for py34 compat.

Kami: Ah, I can leave it here then, we just need to hook it up somewhere (Travis?), otherwise it will get even more out of date than it currently is.

lint-configs/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Usage
+
+To use those configs, add this repository as a git subtree to the repository
+where you want to utilize those configs. After that is done, update make
+targets (or similar) to correctly pass path to the configs to the tools
+in question.
+
+```bash
+git subtree add --prefix lint-configs https://github.com/StackStorm/lint-configs.git master --squash
```
manasdk: Once you merge this change all of us will have to do this right?

manasdk:
:+1:

lint-configs/README.md
Kami: Nope - only when you want to update lint-configs subtree you use pull.

There is no additional interaction required from the end user once the subtree is checked in (unlike with submodule).

PR #2505, opened 2016-02-17
Add etc folder to pack python path for testing handlers

lakshmi-kannan:
Needed for https://github.com/StackStorm/st2contrib/pull/396


Kami:
Sure, why not :P

PR #2506, opened 2016-02-17
Fix "Content-Type" parsing bug in the webhooks API endpoint

### LINKED ISSUE #2504: API returns 400 for some requests due to "bad" Content-type header
It seems that *some* API endpoints will return a HTTP 400 when presented with a header such as:

```
Content-type: application/json; charset=utf-8
```

but works fine with:

```
Content-type: application/json
```

I ran into this with a POST to the URL `/webhooks/st2`. I'm currently running version `1.4dev-52`

Kami:
This pull request fixes a bug reported in #2504 and makes sure we also correctly take into account any optional option such as charset which can also be present in the `Content-Type` header value.

Resolves #2504.

### DIFF
--- a/st2api/st2api/controllers/v1/webhooks.py
+++ b/st2api/st2api/controllers/v1/webhooks.py
@@ -21,6 +21,7 @@ except ImportError:
 import six
 import pecan
 import uuid
+import werkzeug
 from pecan import abort
 from pecan.rest import RestController
 from six.moves.urllib import parse as urlparse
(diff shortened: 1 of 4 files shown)


manasdk:
Any chance we can honor the charset if provided? Not a deal breaker or anything.

Kami:
@manasdk I believe `pecan` should already take care of that when parsing the body (we don't parse the body ourselves, just the headers).

PR #2508, opened 2016-02-17
Various improvements in the linux.wait_for_ssh action

### LINKED ISSUE #2507: linux.wait_for_ssh action improvements
Currently `linux.wait_for_ssh` action (https://github.com/StackStorm/st2/blob/master/contrib/linux/actions/wait_for_ssh.yaml) only supports authenticating with RSA SSH key.

There are multiple ways we can improve this action and make it more generic.

- Support for other key types
- Support for password based authentication
- Refactor the action to utilize the new Paramiko SSH client
- Default to using system user (stanley) username and key file. Ideally key file location should be read from the config

Kami:
@jon-middleton originally reported on the Slack community channel that this action uses an incorrect path to the default SSH key file used to authenticate.

The problem is that path is hard-coded in the action parameter default value while it should really be read from the config file since the path is user-configurable and installation specific.

In addition to that, I noticed this action needs some love so I also made some other improvements which make this action more generic and generally usable.

1. Support for password based authentication (previously only key based authentication was supported)
2. Support for non-default SSH ports (previously only 22 was supported)
3. Support for non RSA key types (previously only RSA keys were supported)
4. If neither `password` nor `keyfile` argument is explicitly provided, action will try to use path to the key file from the system user. That's the default behavior to keep the action backward compatible.
5. Refactor the code to utilize the new `ParamikoSSHClient` class which is also used by the remote runner.
6. Add new `sleep_delay` parameter which represents number of seconds to sleep / wait before each failed connection timeout and actually utilize existing `ssh_timeout` argument for the SSH connection timeout.

Resolves #2507.

### DIFF
--- a/contrib/linux/actions/wait_for_ssh.py
+++ b/contrib/linux/actions/wait_for_ssh.py
@@ -9,12 +9,14 @@ from st2actions.runners.ssh.paramiko_ssh import ParamikoSSHClient
 
 
 class BaseAction(Action):
-    def run(self, hostname, port, username, password=None, keyfile=None, ssh_timeout=30,
-            retries=10):
-        # Note: If neither password nor key file is provided, we try to use
-        # system user key file
+    def run(self, hostname, port, username, password=None, keyfile=None, ssh_timeout=5,
+            sleep_delay=20, retries=10):
+        # Note: If neither password nor key file is provided, we try to use system user 
+        # key file
         if not password and not keyfile:
             keyfile = cfg.CONF.system_user.ssh_key_file
+            self.logger.info('Neither "password" nor "keyfile" parameter provided, '
+                             'defaulting to using "%s" key file' % (keyfile))
 
         client = ParamikoSSHClient(hostname=hostname, port=port, username=username,
                                    password=password, key_files=keyfile,
@@ -28,7 +30,8 @@ class BaseAction(Action):
                 client.connect()
                 return True
             except Exception as e:
-                self.logger.info('Attempt %s failed (%s), sleeping...' % (attempt, str(e)))
-                time.sleep(ssh_timeout)
+                self.logger.info('Attempt %s failed (%s), sleeping for %s seconds...' %
+                                  (attempt, str(e), sleep_delay))
+                time.sleep(sleep_delay)
 
         raise Exception('Exceeded max retries (%s)' % (retries))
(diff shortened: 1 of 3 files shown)


contrib/linux/actions/wait_for_ssh.yaml
```diff
@@ -1,23 +1,34 @@
...
-      required: true
-    ssh_timeout: 
+    password:
+      description: "Password used to authenticate."
+      required: false
+    keyfile:
+      description: "SSH key file used to authenticate."
+      required: false
+    # TODO: We should really rename this to sleep delay or similar
+    ssh_timeout:
```
Kami: I really want to rename this argument since current name is deceiving and incorrect.

Right now it represents a sleep delay between each connection attempt and not a SSH connection timeout.

I was thinking about actually utilizing this parameter for SSH connection timeout and introducing a new `sleep_delay` (or similar) parameter which represents a sleep delay between each SSH connection attempt.

What do others think?

Kami: There we go - 7ba9b524c55e278209ac90e7e4fe342f9171e525

contrib/linux/actions/wait_for_ssh.py
```diff
@@ -2,26 +2,36 @@
...
 class BaseAction(Action):
-    def run(self, keyfile, username, hostname, ssh_timeout, retries):
-        key = paramiko.RSAKey.from_private_key_file(keyfile)
-        client = paramiko.SSHClient()
-        client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+    def run(self, hostname, port, username, password=None, keyfile=None, ssh_timeout=5,
+            sleep_delay=20, retries=10):
+        # Note: If neither password nor key file is provided, we try to use system user
+        # key file
+        if not password and not keyfile:
```
manasdk: empty password not ok? I thought `sshd` can support no passwords. perhaps put the check on username.

manasdk:
+1

contrib/linux/actions/wait_for_ssh.py
Kami: I believe it does, but not sure if we should support that (it's a bad idea to use an empty password :)).

manasdk: hmm, not really our concern but I doubt we will be dinged for being opinionated.

PR #2511, opened 2016-02-18
Python runner actions datastore access changes, action unit testing fixes and lint cleanup

Kami:
This pull request updates base `Action` class to utilize `action_service` argument and expose datastore management methods via this `action_service` instance variable.

Using `action_service` makes it consistent with sensors and also allows us to expand and add additional functionality to action service in the future.

In addition to that, it includes some other changes:

* Make sure that we pass mock action service and datastore to the Action instance which is available to action unit tests.
* Add some more test for BaseActionTestCase
* Random lint fixes and cleanup

## TODO

- [x] Update affected tests in st2contrib/ - https://github.com/StackStorm/st2contrib/pull/402
- [x] Documentation on get_action_instance - https://github.com/StackStorm/st2docs/pull/67
- [x] Update affected action docs (docs/source/actions.rst) - https://github.com/StackStorm/st2docs/pull/67
- [x] Notify plexxi about changes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,8 @@ pylint: requirements .pylint
 		. $(VIRTUALENV_DIR)/bin/activate; pylint -E --rcfile=./lint-configs/python/.pylintrc --load-plugins=pylint_plugins.api_models $$component/$$component || exit 1; \
 	done
 	# Lint Python pack management actions
-	. $(VIRTUALENV_DIR)/bin/activate; pylint -E --rcfile=./lint-configs/python/.pylintrc --load-plugins=pylint_plugins.api_models contrib/packs/actions/pack_mgmt/ || exit 1;
+	. $(VIRTUALENV_DIR)/bin/activate; pylint -E --rcfile=./lint-configs/python/.pylintrc --load-plugins=pylint_plugins.api_models contrib/packs/actions/ || exit 1;
+
 	# Lint other packs
 	. $(VIRTUALENV_DIR)/bin/activate; pylint -E --rcfile=./lint-configs/python/.pylintrc --load-plugins=pylint_plugins.api_models contrib/linux || exit 1;
 	# Lint Python scripts
@@ -84,7 +85,7 @@ flake8: requirements .flake8
 	@echo "==================== flake ===================="
 	@echo
 	. $(VIRTUALENV_DIR)/bin/activate; flake8 --config ./lint-configs/python/.flake8 $(COMPONENTS)
-	. $(VIRTUALENV_DIR)/bin/activate; flake8 --config ./lint-configs/python/.flake8 contrib/packs/actions/pack_mgmt/
+	. $(VIRTUALENV_DIR)/bin/activate; flake8 --config ./lint-configs/python/.flake8 contrib/packs/actions/
 	. $(VIRTUALENV_DIR)/bin/activate; flake8 --config ./lint-configs/python/.flake8 contrib/linux
 	. $(VIRTUALENV_DIR)/bin/activate; flake8 --config ./lint-configs/python/.flake8 scripts/
 	. $(VIRTUALENV_DIR)/bin/activate; flake8 --config ./lint-configs/python/.flake8 tools/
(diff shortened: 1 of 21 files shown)


Makefile
```diff
@@ -68,7 +68,8 @@ pylint: requirements .pylint
 		. $(VIRTUALENV_DIR)/bin/activate; pylint -E --rcfile=./lint-configs/python/.pylintrc --load-plugins=pylint_plugins.api_models $$component/$$component || exit 1; \
 	done
 	# Lint Python pack management actions
-	. $(VIRTUALENV_DIR)/bin/activate; pylint -E --rcfile=./lint-configs/python/.pylintrc --load-plugins=pylint_plugins.api_models contrib/packs/actions/pack_mgmt/ || exit 1;
+	. $(VIRTUALENV_DIR)/bin/activate; pylint -E --rcfile=./lint-configs/python/.pylintrc --load-plugins=pylint_plugins.api_models contrib/packs/actions/ || exit 1;
```
Kami: I noticed we added some new Python files there but we didn't lint files there before so some lint violations managed to slip in.

st2actions/st2actions/runners/python_action_wrapper.py
```diff
@@ -34,6 +35,38 @@
...
+        self._datastore_service = DatastoreService(logger=logger,
+                                                   pack_name=self._action_wrapper._pack,
+                                                   class_name=self._action_wrapper._class_name,
+                                                   api_username='action_service')
+
+    ##################################
+    # Methods for datastore management
+    ##################################
+
+    def list_values(self, local=True, prefix=None):
```
manasdk: Why not just let datastore_service be a pass through?

Kami: This way it's consistent with existing sensor service interface.

manasdk: Hmm, not a fan of the pattern but consistency is better in this regard.

st2actions/st2actions/runners/pythonrunner.py
```diff
@@ -69,13 +70,25 @@ def __init__(self, config=None):
         :type config: ``dict``
         """
         self.config = config or {}
-        # logger and datastore are assigned in PythonActionWrapper._get_action_instance
-        self.logger = None
+        self.logger = get_logger_for_action(action_name=self.__class__.__name__)
+
+        # action_service is late assigned post class instatiation in "setup()"
         self.datastore = None
```
manasdk: self.action_service

st2actions/st2actions/runners/pythonrunner.py
```diff
@@ -69,13 +70,25 @@ def __init__(self, config=None):
...
+        """
+        Method which performs additional class initialization.
+        """
+        self.action_service = action_service
+        self._initialized = True
+
     @abc.abstractmethod
     def run(self, **kwargs):
-        pass
+        if not self._initialized:
```
manasdk: Don't think this will be called ever if overridden. I don't think calling run of the base method is typical.

Kami: Yeah, mentioned this on Slack - it's mostly pointless since most actions don't call parent run so I can remove it.

Kami:
Also, just a heads up.

There is some similar / duplicated code in the mocks, etc. but I plan to add tests, etc. first and address / refactor that later.

manasdk:
@vCabbage @andrew-regan 

Likely some actions you have in place will be affected. Since this is not released yet we should be ok changing the interface.

Kami:
@manasdk Yeah, thanks for the heads up - I planned to ping @vCabbage once the PR is fully ready (including doc changes, etc.).

 This will most likely happen tomorrow and I will provide more details to them then :)

andrew-regan:
Thanks for the heads up!

andrew-regan:
:+1:

Kami:
@manasdk and I discussed the offline the other day and decided to go with another approach which is consistent with sensors.

We will first try to pass `action_service` argument to the `Action` class constructor and if that doesn't work (e.g. old action which hasn't been updated yet), we will simply fall back to the late assignment on the action instance post instantiation.

This is not ideal, but good thing is that a lot of Action classes doesn't override the constructor and we control `st2contrib` so it's easy to update most of the actions.

Eventually, (e.g. after next .minor) release we should also consider dropping this late assignment fall back since by then most actions should be updated.

Kami:
I will go ahead and merge those + documentation changes into master.

I updated affected actions in core packs which are bundled in this repo, but I will wait with actions in st2contrib until more users upgrade to the latest version and / or pack versioning is in place.

I could update it so it also works with older versions of StackStorm, but it would require more boiler plate code...

PR #2513, opened 2016-02-19
Normalize url in trigger parameters

manasdk:
* Leading or trailing slashes in the url coming from the rule
  would cause the API to reject webhook requests.
* Normailzing the url to not include trailing or leading / means
  that the match is / agnostic and StackStorm does not end tripping
  over some silly /.
* URL normalization basically mean multiple version of the URL map
  to the same URL.
* Each unique URL gets it own Trigger. Even though the controller
  treats `sample`, `/sample` and `sample/` different it still
  represents 3 different triggers. Therefore it is necessary to
  match the normalized version to all 3 triggers. This is done by
  holding a match between normalized to a list of triggers.

### DIFF
--- a/st2api/st2api/controllers/v1/webhooks.py
+++ b/st2api/st2api/controllers/v1/webhooks.py
@@ -44,10 +44,50 @@ LOG = logging.getLogger(__name__)
 TRACE_TAG_HEADER = 'St2-Trace-Tag'
 
 
+class HooksHolder(object):
+    """
+    Maintains a hook to Trigger mapping.
+    """
+    def __init__(self):
+        self._triggers_by_hook = {}
+
+    def __contains__(self, key):
+        return key in self._triggers_by_hook
+
+    def add_hook(self, hook, trigger):
+        if hook not in self._triggers_by_hook:
+            self._triggers_by_hook[hook] = []
+        self._triggers_by_hook[hook].append(trigger)
+
+    def remove_hook(self, hook, trigger):
+        if hook not in self._triggers_by_hook:
+            return False
+        remove_index = -1
+        for idx, item in enumerate(self._triggers_by_hook[hook]):
+            if item['id'] == trigger['id']:
+                remove_index = idx
+                break
+        if remove_index < 0:
+            return False
+        self._triggers_by_hook[hook].pop(remove_index)
+        if not self._triggers_by_hook[hook]:
+            del self._triggers_by_hook[hook]
+        return True
+
+    def get_triggers_for_hook(self, hook):
+        return self._triggers_by_hook.get(hook, [])
+
+    def get_all(self):
+        triggers = []
+        for values in six.itervalues(self._triggers_by_hook):
+            triggers.extend(triggers)
+        return triggers
+
+
 class WebhooksController(RestController):
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


st2api/tests/unit/controllers/v1/test_webhooks.py
```diff
@@ -164,6 +165,42 @@ def test_unsupported_content_type(self):
...
+            get_webhook_trigger('with_trailing_slash', '/with_trailing_slash/'),
+            get_webhook_trigger('with_leading_trailing_slash', '/with_leading_trailing_slash/'),
+            get_webhook_trigger('with_mixed_slash', '/with/mixed/slash/')
+        ]
+
+        controller = WebhooksController()
+        for trigger in test_triggers:
+            controller.add_trigger(trigger)
+
+        get_webhook_trigger('with_mixed_slash', '/with/mixed/slash/')
```
lakshmi-kannan: This is weird.

lakshmi-kannan:
lgtm. lint's failing.

Kami:
Probably still a good idea to document this behavior in the docs.

st2api/st2api/controllers/v1/webhooks.py
```diff
@@ -169,12 +169,19 @@ def update_trigger(self, trigger):
...
         # creation
-        url = trigger['parameters']['url']
+        url = self._get_normalized_url(trigger)
 
         if url in self._hooks:
             LOG.info('Stop listening to endpoint: %s', urljoin(self._base_url, url))
             del self._hooks[url]
 
+    def _get_normalized_url(self, trigger):
+        '''
```
Kami: Double quotes? :P

Kami:
Actually, now that I think of it some more - docs are definitely a must since otherwise it could be confusion.

Output of `st2 webhook list` will differ from what's defined in the rule so I could see that being slightly confusing.

Besides that, LGTM.

manasdk:
I just realized that this is not really going to work properly unless I normalize higher up when the rule is created. This is very messy unfortunately; hopefully I can actually fix this one.

Anyway docs are here https://github.com/StackStorm/st2docs/pull/65

manasdk:
@kami With new changes here is what output will look like 

```
$ st2 webhook list
+------------------+------+--------------------------------------+-------------+---------------------------------+
| type             | pack | name                                 | description | parameters                      |
+------------------+------+--------------------------------------+-------------+---------------------------------+
| core.st2.webhook | core | 2acbe30e-2e22-4798-8577-af717ef6d2ec | None        | {u'url': u'sample'}             |
| core.st2.webhook | core | 46a53a73-d67d-4fdd-aafa-f8a128382988 | None        | {u'url': u'/sample'}            |
| core.st2.webhook | core | efe4e176-f6a4-44a2-bf36-19bec6d8f040 | None        | {u'url': u'/sample_with_slash'} |
+------------------+------+--------------------------------------+-------------+---------------------------------+
```

Kami:
Cool.

Did we confirm end to end (dunno if we have any such tests right now) that the whole flow works:

1. Creating a rule with webhook
2. Sending data to that webhooks
3. Removing a rule with webhook (webhook is removed).

manasdk:
Yes works. We shockingly do have exactly that e2e

PR #2517, opened 2016-02-20
mistral-workbook-complex.yam: task create_vm never call

gaelL:
In the example, the `create_vm` task seems to be never called.
If I understand well the example, we would like to have :

  * register_dns
  * create_vm
  * configure_vm
  * ...

But currently we have `register_dns` that call `configure_vm` where as the vm is not created.

  * `register_dns` -> `configure_vm`


m4dcoder:
create_vm and register_dns in this example are executed concurrently.  configure_vm has a join statement that waits for both to complete.

gaelL:
I didn't saw it that way. Thanks

Issue #2485, opened 2016-02-10
Unclear decorate_log_method() implementaton in st2common/st2common/log.py


lakshmi-kannan:
https://stackstorm-community.slack.com/archives/community/p1455127437005176

Reported by: @andrew-regan on slack.

andrew-regan:
The `decorate_log_method()` functionality causes all logging of filename/funcName to point at `log.py:func_wrapper:68` instead of the real source where the log message originated. Here's an example using the  `decorate_log_method()` functionality:
```
2016-02-10 13:18:01,668 [st2.SensorWrapper.vsphere.EventSensor.EventSensor|68850|4483470512] | log.py:func_wrapper:68 (INFO) - Starting poll...
2016-02-10 13:18:01,687 [st2.SensorWrapper.vsphere.EventSensor.EventSensor|68850|4483470512] | log.py:func_wrapper:68 (DEBUG) - Retrieving vSphere events from 2016-02-10 18:17:56.643586+00:00 to 2016-02-10 18:18:01.737971+00:00
2016-02-10 13:18:01,726 [st2.SensorWrapper.vsphere.EventSensor|68850|4483470512] | log.py:func_wrapper:68 (AUDIT) - Setting value in the datastore (name=vsphere.EventSensor:last_vsphere_event::7e2250f2-19e3-4335-a840-e01c55d7dd56)
```
And with the call to `decorate_log_method()` removed in `getLogger(name)`:
```
2016-02-10 13:17:29,762 [st2.SensorWrapper.vsphere.EventSensor.EventSensor|68475|4393854128] | vsphere_event_sensor.py:_poll:107 (INFO) - Starting poll...
2016-02-10 13:17:29,780 [st2.SensorWrapper.vsphere.EventSensor.EventSensor|68475|4393854128] | vsphere_event_connector.py:get_events:79 (DEBUG) - Retrieving vSphere events from 2016-02-10 18:17:24.752739+00:00 to 2016-02-10 18:17:29.830475+00:00
2016-02-10 13:17:29,818 [st2.SensorWrapper.vsphere.EventSensor|68475|4393854128] | datastore.py:set_value:117 (AUDIT) - Setting value in the datastore (name=vsphere.EventSensor:last_vsphere_event::7e2250f2-19e3-4335-a840-e01c55d7dd56)
```

andrew-regan:
[...]

PR #2518, opened 2016-02-22
Fix "decorate_log_method", make sure LogRecord attributes reflect the correct caller

Kami:
This pull request fixes issue reported in #2485.

It makes sure `filename`, `module`, `funcName` and `lineno` `LogRecord` attributes correctly point to the function / method which called the underlying logger method.

Previously they incorrectly pointed to the `log.py` file and `decorate_log_method` function.

I'm not too big of a fan of the implementation, but after doing some more research online it doesn't seem like there are any better and nicer alternatives possible :/

If it ever turns out that this adds too big of a performance overhead (unlikely since the same functionality exists and it's turned on by default in Python stdlib), we can revert this change or look into alternatives (e.g. only add caller info if the format string uses any of those log record attributes).

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,6 +27,10 @@ in development
   * Support for using default system user (stanley) ssh key if neither ``password`` nor
     ``keyfile`` parameter is provided.
 * Support for leading and trailing slashes in the webhook urls. (improvement)
+* Make sure that the ``filename``, ``module``, ``funcName`` and ``lineno`` attributes which are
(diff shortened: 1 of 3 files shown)


andrew-regan:
Tested the changes using the sensor I've been working on and the logging is now working as expected. Seeing the correct filename/funcName values. :+1:

Kami:
@andrew-regan Great.

Thanks for testing it out and noticing and reporting this change :)

PR #2519, opened 2016-02-22
Introduce new "matchwildcard" criteria operator, update "matchregex" so dot (.) also matches new lines

Kami:
This pull request includes the following changes:

1. It introduces new ``matchglob`` criteria operator which provides support for Unix shell-style wildcards (* and ?). This operator should be preferred over ``matchregex`` for simple string matches since it's a lot more straight forward.

2. Update ``matchregex`` operator so `.` character also matches new lines. In most cases where you work with multi line strings you want ``.`` character to also match new lines so that's a more sensible default behavior (people got burnt by this in the past).

This change is backward incompatible (it will make existing regexes a bit less greedy), but I will document this with examples in the upgrade notes and it's unlikely that it's going to badly affect existing regexes (in most cases it will allow users to simply existing regex patterns).

## TODO

- [x] Documentation changes
- [x] Upgrade notes changes

### DIFF
--- a/st2common/st2common/operators.py
+++ b/st2common/st2common/operators.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import re
+import fnmatch
 
 from st2common.util import date as date_utils
 
@@ -113,6 +114,13 @@ def greater_than(value, criteria_pattern):
     return value > criteria_pattern
 
 
+def match_glob(value, criteria_pattern):
+    if criteria_pattern is None:
+        return False
+
+    return fnmatch.fnmatch(value, criteria_pattern)
+
+
 def match_regex(value, criteria_pattern):
     if criteria_pattern is None:
         return False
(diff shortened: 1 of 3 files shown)


st2common/st2common/operators.py
```diff
@@ -161,6 +169,7 @@ def nexists(value, criteria_pattern):
     return value is None
 
 # operator match strings
+MATCH_GLOB = 'matchglob'
```
lakshmi-kannan: I donno why someone would use match_glob in rules context unless they want glob matching on filenames which I think can be done with match_regex.

Kami: As the description says - it's simpler. For simple glob matching, regexes are can be an overkill.

lakshmi-kannan: Ok, you have my skeptic +1 on that change. The other changes lgtm.

manasdk:
+1.

Meh about the glob one but whatever makes a user happy in this case.

Kami:
I decided to go ahead and rename it from ``matchglob`` to ``matchwildcard`` since globs usually refer to wildcards in file names.

PR #2521, opened 2016-02-23
Allow user to turn off verbose flag in the "linux.rm" action

Kami:
This change makes "-v" (verbose) flag for `linux.rm` action configurable. For backward compatibility reasons it defaults to true.

I noticed `st2cd.st2workroom_st2enterprise_test` tests recently started failing with the following error - https://gist.github.com/Kami/101e707a66f54930c6cd. I believe this is related to the error @lakshmi-kannan found and worked-around in the past (we didn't fully fix it though since it would require writing our own streaming based parser which is an over kill right now).

I believe the issue pops ups because some of the file recently added to the workroom repo contains a non-ascii character in the name so a simple work-around would be to pass `verbose: false` parameter to this action in our workflows. This way what was being done (which includes file names) won't be printed to stdout.

### DIFF
--- a/st2tests/st2tests/base.py
+++ b/st2tests/st2tests/base.py
@@ -488,7 +488,7 @@ class BaseSensorTestCase(TestCase):
 
 class BaseActionTestCase(TestCase):
     """
-    Base class for action tests.
+    Base class for Python runner action tests.
     """
 
     action_cls = None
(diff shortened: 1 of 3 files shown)


Kami:
I also wanted to add some tests, but that's a bit hard right now and we need #2522 first.

st2tests/st2tests/base.py
```diff
@@ -488,7 +488,7 @@ def assertTriggerDispatched(self, trigger, payload=None, trace_context=None):
 
 class BaseActionTestCase(TestCase):
     """
-    Base class for action tests.
+    Base class for Python runner action tests.
```
Kami: I will eventually need to rename this to `BasePythonRunnerActionTestCase`, but I will do this in a different PR when adding support for tests for other runners.

I will also leave old name in place for backward compatibility reasons.

PR #2523, opened 2016-02-23
Remove index.php from default nginx for st2web

armab:
`st2web` has `index.html` as index entrypoint.
Removing unneeded pieces, probably oversight with copy-paste.


Kami:
LGTM, +1

PR #2526, opened 2016-02-24
Use query instead of objects mongoengine filter

Kami:
This operation results in the same underlying query, but it allows us to use profiling stuff I added a while back (profiling stuff doesn't work when using `.objects()` document method to retrieve stuff from the database).


manasdk:
+1

PR #2528, opened 2016-02-24
Make stream a standalone service and WSGI app, gunicorn config changes

Kami:
This pull request aims to fix the API performance issues we have identified (API throughput when using eventlet async gunicorn worker is very low). It seems like that eventlet asnyc gunicorn worker doesn't play along well with any other application code which relies on eventlet monkeypatching.

This pull request aims to fix that by doing the following:

* Update API and auth gunicorn config to use sync worker type with 1 (or more processes) and 1 thread
* Make stream a standalone service and WSGI app. We will use async worker with 1 or more worker processes.

This change is required because stream and other API endpoints have vastly different and incompatible models. API serves short-lived requests while stream serves long lived requests. This way we can tune gunicorn accordingly (sync worker for API and async one for stream) and receive normal and expected performance.

Doing those changes also allows us to remove "WSGI server shutdown" and some other work-around changes we needed because stream was running in the same process / WSGI app as the API.

We decided to only include this change in the new packages so we need to decide what to do with v1.4. If we plan to release v1.4 before new packages are out this means we either need to skip this change or make corresponding workroom changes (which is something we want to avoid).

## TODO

- [x] Add new `st2stream` changes
- [x] Update affected code
  - [x] tests
  - [x] new listen port
  - [x]  update launchdev
  - [x] logrotate config
  - [x] sample config
- [x] Related changes for packages (add new service for stream)
- [x] Gunicorn changes for packages (st2api, st2auth need to use sync worker with 1 process and 1 thread, st2stream needs to use eventlet async worker with 1 or more processes)
- [x] Pacakges nginx changes - for backward compatibility reasons we need to proxy /api/v1/stream to the stream st2gunicorn app (and not st2api)
- [x] Decide what to do with v1.4 release (wait for new packages?)


Kami:
Also just a head up - once this is merged you will need to nuke `virtualenv` dir locally in a dev VM and re-create it using `make virtualenv`.

One faster option also is to edit `virtualenv/bin/activate` script and add new `st2stream` package to the path there. This way you don't need to wait for all the Python packages to be installed again.

tools/st2ctl
```diff
@@ -2,7 +2,7 @@
 
 LOGFILE="/dev/null"
 WEBUI_LOG_FILE="/var/log/st2/st2web.log"
-COMPONENTS="st2actionrunner st2api st2auth st2sensorcontainer st2rulesengine mistral st2resultstracker st2notifier st2garbagecollector"
+COMPONENTS="st2actionrunner st2api st2stream st2auth st2sensorcontainer st2rulesengine mistral st2resultstracker st2notifier st2garbagecollector"
```
armab: There is also `st2ctl` in `st2common/bin`, which is used by the new packages.

Kami: Ah, good catch, thanks.

 Will also update that.

PR #2529, opened 2016-02-24
Add "--output" and "--existing-file" options to st2-submit-debug-info

vCabbage:
Primary purpose of this PR is to add the following options:
* `--output` - Specifies the output location of the bundle when used with `--review`.
* `--existing-file` - Allows uploading of an existing bundle created with `--review`.

I took the liberty of refactoring the code to facilitate adding this functionality. Majority of the changes center around moving the separate functions into a class in order to avoid passing large number of arguments around and consolidating duplicate logic. Some logic was also simplified. Behavior was not changed other than as noted below.

* Moved the date formatting string used in the output filename to `DATE_FORMAT` global variable.
* Removed colons from the date formatting string used in the output filename.
* Added the collection of stderr when collecting command output.
* Changed the yaml config file format to be more consistent with the script's variable names/types (ie, uses lists instead of dicts, where approriate).

Updated config.yaml format:
``` yaml
log_file_paths: 
    - /var/log/st2/*.log
    - /var/log/mistral*.log
st2_config_file_path: /etc/st2/st2.conf
mistral_config_file_path: /etc/mistral/mistral.conf
s3_bucket_url: https://st2debuginfo.s3.amazonaws.com/
gpg_key_fingerprint: BDE989A1F308B18D29789C717064B11C82F62D6F
gpg_key : |
        -----BEGIN PGP PUBLIC KEY BLOCK-----
        Version: GnuPG v1

        mQINBFTaXHIBEAC+IId30KtMKgKzaT+2Hc/svFkM46ZzG0+EF+0se5yBlOMiTJxl
        Obfuj2CLAg1QnusfefOrSG3l6MwByaQvzHwUPWx7S0Fa0N2TSVFedb9bSYByUtd0
        [truncated]
        -----END PGP PUBLIC KEY BLOCK-----
shell_commands:
    - rpm -qa
company_name: StackStorm
```

PR to update st2docs: https://github.com/StackStorm/st2docs/pull/73


Kami:
Great, thanks for the contribution.

I did a quick glance and overall I'm fine with the idea of using a class. It does indeed make code a bit nicer and simpler now that there are so many arguments.

I will do a more in depth review as soon as possible.

Kami:
I just tested the changes locally and they appear to work fine :+1: 

There is still some additional error handling we could do (e.g. if config doesn't contain valid YAML / dictionary it will throw `AttributeError: 'str' object has no attribute 'get'`), but we can do this in another PR.

I will review the changes now and if everything looks OK, go ahead and merge them :)

st2debug/tests/integration/test_submit_debug_info.py
```diff
@@ -204,22 +197,18 @@ def test_encrypt_archive(self):
                           extract_path='/tmp')
 
     def test_encrypt_archive_with_custom_gpg_key(self):
```
Kami: In a future PR it would also be great if you can add another test case which tests more things when providing a custom config - verify command name is script, verify command stuff is included in the output archive, etc.

Kami:
I've added some comments and merged PR into master since it looks good. Thanks again :+1: 

It would be great though if you can address those minor comments (notably more tests for custom config option) in a future PR when you have some time.

Overall the change are good and thanks for doing the refactoring - the class approach looks a lot better!

vCabbage:
@Kami Thanks!

I'm working on a follow up PR now.

In regards to the additional tests, can you elaborate on "verify command name is script"? I'm not clear on what you're asking for.

PR #2530, opened 2016-02-25
Make sure "--use-debugger" flag is taken into account in the sensor process

Kami:
Previously, `--use-debugger` flag was not correctly propagated and checked in the child sensor processes which means `thread` module was still patched and this broke remote debugging support (some remote debugger like the pydev and pycharm one rely on thread module not being patched).

Some references:

* https://ask.openstack.org/en/question/815/how-do-i-debug-nova-service-with-eclipse-and-pydev/
* http://debugopenstack.blogspot.si/

I didn't test this end to end since I don't have PyDev installed locally, but I tested it using netcat and it seems to be working. In any case, I will ask Ethan or someone with PyDev to test those changes.

This was reported by Ethan on the IRC.

I also asked Ethan to write a guest blog post on remote debugging sensors with PyDev once this issue has been resolved since having such post / article would be great.

### DIFF
--- a/st2reactor/st2reactor/container/sensor_wrapper.py
+++ b/st2reactor/st2reactor/container/sensor_wrapper.py
@@ -33,18 +33,14 @@ from st2common.services.triggerwatcher import TriggerWatcher
 from st2reactor.sensor.base import Sensor, PollingSensor
 from st2reactor.sensor import config
 from st2common.services.datastore import DatastoreService
+from st2common.util.monkey_patch import monkey_patch
 
 __all__ = [
     'SensorWrapper',
     'SensorService'
 ]
 
-eventlet.monkey_patch(
-    os=True,
-    select=True,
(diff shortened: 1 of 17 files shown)


st2reactor/st2reactor/container/sensor_wrapper.py
```diff
@@ -33,18 +33,14 @@
...
     'SensorService'
 ]
 
-eventlet.monkey_patch(
-    os=True,
-    select=True,
-    socket=True,
-    thread=False if '--use-debugger' in sys.argv else True,
-    time=True)
+monkey_patch()
```
Kami: Any objections to using this wrapper method everywhere else in the code where we call `eventlet.monkey_patch` right now?

manasdk: +1

PR #2537, opened 2016-02-27
Support for user-defined display timezone for timestamps in the CLI

Kami:
This pull request introduces initial support for user-defined display timezone in the CLI. User can configure display timezone in the per-user CLI config (`~/.st2/config`) and all the timestamps will be converted to that timezones when being displayed in the CLI.

@enykeev argued that we could add some kind of metadata field to the `User` object and leverage that in the CLI and Web UI. I'm personally not convinced about that approach (yet), since I believe that is a user and medium specific setting.

I do see some advantages of that approach do - all user settings are stored in a single place and we don't need per-user config in the CLI and Web UI config.

No timezone specified (defaults to UTC, same behavior as before):

selection_349

Europe/Ljubljana timezone specified in the CLI config:

selection_348

Similar as other formatting we do, this formatting is only applied when using human-friendly output format such as table (we don't mutate or format data when using JSON output mode).


ckamps:
> @enykeev argued that we could add some kind of metadata field to the User object and leverage that in the CLI and Web UI. I'm personally not convinced about that approach (yet), since I believe that is a user and medium specific setting.

> I do see some advantages of that approach do - all user settings are stored in a single place and we don't need per-user config in the CLI and Web UI config.

Since our organization has a common enterprise-wide timezone setting for all IT operations, I am very much in favor of enabling central configuration of a default display timezone so that individual users don't need to bother with such settings when they want the default enterprise reference out of the box.  
Should I file another issue or leverage the discussion associated with the web UI issue?  https://github.com/StackStorm/st2web/issues/292

PR #2546, opened 2016-03-01
Add old packaging metadata for stream service

Kami:
I would prefer to avoid all this mess and duplication, but if I don't add this, update relevant flows, etc. our CI and everything will fail.

I really hope we can eventually get rid of that and just use new packages.


Kami:
Corresponding st2cd change -  https://github.com/StackStorm/st2cd/pull/131

PR #2547, opened 2016-03-01
Ignore conflict / "object already exist" errors during internal trigger types registration

Kami:
Those operations are idempotent so we can simply ignore those errors (e.g. this could occur when using gunicorn with more than 1 worker).


st2api/st2api/cmd/api.py
```diff
@@ -25,6 +25,7 @@
 from st2common.service_setup import teardown as common_teardown
 from st2common.util.monkey_patch import monkey_patch
 from st2api import config
+config.register_opts()
```
manasdk: This is awful!

Please add a comment on why its needed.

Kami: It was already there before, I just accidentally removed it during stream changes.

I do agree that it's awful though.

Issue #2522, opened 2016-02-23
Code infrastructure and support for local and remote command runner action unit tests


Kami:
We now have an infrastructure and code in place for writing Python runner action unit tests, but we don't have such code infrastructure in place for other runners such as local and remote ones.

We should add some code infrastructure in-place which will make writing such tests easier. We can start simple by adding some code infrastructure for writing remote and local command runner action tests. In this case we can simply intercept a final shell command and return it to the test case which can assert that the command is correct. 

This is still valuable since a lot of actions (e.g. many of the actions in linux pack) consists of metadata which dynamically constructs a command based on the action parameters.

PR #2548, opened 2016-03-01
Add support for file attachments in core.sendmail

codyaray:
For my use case, I'm writing a workflow that'll

1. query an API and write the results to a file; and
2. email the file as an attachment.

This patch enables `core.sendmail` to send file attachments. There are two things of note here:

1. The `attachments` parameter is a comma-separated `string`instead of an `array`. This can be changed when we have better support for array serialization for shell scripts. (see [discussion](https://stackstorm-community.slack.com/archives/community/p1456864311005045))
2. Only files found locally on the StackStorm host may be sent as attachments currently. We could change `core.sendmail` to a `remote-shell-script` instead of `local-shell-script` with a default host of localhost to make this more flexible. Thoughts?

```
# st2 run core.sendmail from=test to=cray@peak6.com subject="Just Testing" body="Gotta get 'em good" attachments="/tmp/image1.png,/tmp/image2.png"
..
id: 56d60990fb3b871d916f8043
status: succeeded
parameters:
  attachments: /tmp/image1.png,/tmp/image2.png
  body: Gotta get 'em good
  from: test
  subject: Just Testing
  to: cray@peak6.com
result:
  localhost:
    failed: false
    return_code: 0
    stderr: ''
    stdout: Successfully sent mail to cray@peak6.com
    succeeded: true
```

The email was received with both attachments (named image1.png and image2.png in the email).

The action still works as expected when the `attachments` parameter is omitted.

```
 st2 run core.sendmail from=test to=cray@peak6.com subject="Just Testing" body="Gotta get 'em for real"
.
id: 56d60e1dfb3b871d916f804f
status: succeeded
parameters:
  body: Gotta get 'em for real
  from: test
  subject: Just Testing
  to: cray@peak6.com
result:
  failed: false
  return_code: 0
  stderr: ''
  stdout: Successfully sent mail to cray@peak6.com
  succeeded: true
```

I followed [this guide](http://backreference.org/2013/05/22/send-email-with-attachments-from-script-or-command-line/) for attachments with sendmail.

/cc @Kami @m4dcoder

### DIFF
--- a/contrib/core/actions/send_mail/send_mail
+++ b/contrib/core/actions/send_mail/send_mail
@@ -25,7 +25,13 @@ SEND_EMPTY_BODY=$1
 shift
 CONTENT_TYPE=$1
 shift
-BODY="$@"
+BODY="$1"
+shift
+IFS=',' read -ra ATTACHMENTS <<< "$@"
+
+if [[ $ATTACHMENTS == "None" ]]; then
+  ATTACHMENTS=""
+fi
 
 if [[ "${CONTENT_TYPE}" = "text/html" ]]; then
   LINE_BREAK="<br><br>"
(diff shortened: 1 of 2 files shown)


contrib/core/actions/send_mail/send_mail
```diff
@@ -25,7 +25,13 @@ SEND_EMPTY_BODY=$1
 shift
 CONTENT_TYPE=$1
 shift
-BODY="$@"
+BODY="$1"
+shift
+IFS=',' read -ra ATTACHMENTS <<< "$@"
+
+if [[ $ATTACHMENTS == "None" ]]; then
```
codyaray: Without this guard, it tries to find an attachment named "None". This can be seen on the `stderr` output.
(Should "None" be passed to a `required: false` positional parameter? What is normally passed?)

```
    stderr: 'Warning: attachment None not found, skipping'
```

Kami:
Thanks for the contribution!

We will have a look shortly.

contrib/core/actions/send_mail/send_mail
Kami: Yeah, if no value is provided for a particular parameter when using shell script runner + positional arguments we pass in `None` to the script for the parameter value.

The reason for that is, that in theory, your script could take 4 positional argument and let's say argument number 3 is optional. In this case we still need to pass in something (`None` in our case) if you still want to provide argument number 4 (e.g. `foo bar None arg4`).

I know that's not the best script design (you should put optional arguments at the end), but that's not always possible (e.g. multiple optional arguments).

contrib/core/actions/sendmail.yaml
```diff
@@ -41,5 +41,10 @@ parameters:
     type: string
   sudo:
     immutable: true
+  attachments:
+    description: Array of attachment file paths, comma-delimited.
+    position: 6
+    required: false
+    type: "string"
```
Kami: As mentioned on Slack, I would prefer to make this an array and serialize to comma delimited string in the script runner.

In any case, I can take care of that later.

Kami:
After quickly looking at it, the changes looks good to me, but I still need to test it.

It would also be good to add some tests for it, but this would be quite hard right now. Hopefully, https://github.com/StackStorm/st2/issues/2522 will make it easier in the future.

contrib/core/actions/sendmail.yaml
codyaray: Yeah, I was hoping you could break the serializing to comma delimited string into a separate story. :)

Kami: Yeah, I will do that separate + update the docs :)

codyaray:
Yeah, I was wondering about how tests may be written. Didn't see them for any other shell scripts, so I passed on it. For what its worth, before I moved on to testing with sendmail proper, I changed MAIL from sendmail to tee so that I could just see the email as text. Something like that may be doable and useful for a test like this. Much like writing "real code", though, it requires a bit of design to make sure your code is testable.

Kami:
I checked it locally and seems to be working fine so I will go ahead and merge it.

I will also try to work on the local / remoter runner testing code infrastructure and framework in the near future so we can finally add some tests for those actions.

Kami:
Merged, thanks.

PR #2550, opened 2016-03-02
Add chatops to st2ctl

dzimine:
Problem: avoid managing st2chatops service as an exception. 

I bounced at this when testing chatops docs, and was about to enable the service at boot, but realized we don't want an exception here and rather manage it just like all other Stackstorm components.

Tested on ubuntu and RHEL, unfortunately it complains on "unknown service" when chatops is not installed, but I think consistency is better. Input welcome.

### DIFF
--- a/st2common/bin/st2ctl
+++ b/st2common/bin/st2ctl
@@ -2,7 +2,7 @@
 
 
 LOGFILE="/dev/null"
-COMPONENTS="st2actionrunner st2api st2stream st2auth st2garbagecollector st2notifier st2resultstracker st2rulesengine st2sensorcontainer mistral"
+COMPONENTS="st2actionrunner st2api st2stream st2auth st2garbagecollector st2notifier st2resultstracker st2rulesengine st2sensorcontainer st2chatops mistral"
 STANCONF="/etc/st2/st2.conf"
 
 # Ensure global environment is sourced if exists
@@ -141,6 +141,7 @@ function clean_logs() {
 function getpids() {
   echo "##### st2 components status #####"
   COMPONENTS=${COMPONENTS/mistral/mistral-server mistral-api}
+  COMPONENTS=${COMPONENTS/st2chatops/hubot}
 
   for COM in ${COMPONENTS}; do
     PID=`ps ax | grep -v grep | grep -v st2ctl | grep "${COM}" | awk '{print $1}'`


emedvedev:
> unfortunately it complains on "unknown service" when chatops is not installed
I don't think that's good. Same reason I keep objecting to `st2web` still being listed with AIO: it gives you an error and you assume something's wrong, when it's actually `st2ctl` that's wrong. If a user made a choice not to install the service, they shouldn't get info about it being missing.

dzimine:
way to handle optional services in st2ctl? If so let's do it but not hold the release with it.

On a side note, when do we ditch st2/tools/st2ctl ? having two is REALLY confusing...

lakshmi-kannan:
>way to handle optional services in st2ctl? If so let's do it but not hold the release with it.

Do not hold the release. This can be fixed later. +1 to merge.

st2common/bin/st2ctl
```diff
@@ -141,6 +141,7 @@ function clean_logs() {
 function getpids() {
   echo "##### st2 components status #####"
   COMPONENTS=${COMPONENTS/mistral/mistral-server mistral-api}
+  COMPONENTS=${COMPONENTS/st2chatops/hubot}
```
dzimine: when it exists, it picks up two pids: one from the parent shell wrapper, another from actual node. Not sure if we need two pids, it's a bit confusing.
```
ps ax | grep hubot
31774 ?        S      0:00 bash -c ulimit -S -c 0 >/dev/null 2>&1 ; cd /opt/stackstorm/chatops && nohup </dev/null >/dev/null 2>&1 /opt/stackstorm/chatops/bin/hubot &
31776 ?        Sl     0:01 node /opt/stackstorm/chatops/node_modules/hubot/node_modules/.bin/coffee /opt/stackstorm/chatops/node_modules/.bin/hubot @
```

Kami:
What's our plan for this?

st2common/bin/st2ctl
DoriftoShoes: This is no longer an issue with the new st2chatops package since it is not in a container.

DoriftoShoes:
I say we merge this.

emedvedev:
Closing in favor of https://github.com/StackStorm/st2/pull/2625

PR #2551, opened 2016-03-02
fix st2ctl to handle upstart

dennybaa:


Kami:
LGTM, but I haven't tested it yet.

We also need some tests for that, I believe this would tie into tests @lakshmi-kannan is working on.

dennybaa:
@Kami do we?
https://github.com/StackStorm/st2-packages/blob/master/packages/st2/bin/runners.sh#L17-L22
it's been tested many times when we use packages. and every time we invoke start wrapper actionrunner service.

Kami:
Yeah, i do think tests would be good - and they need to cover all the service managers we support.

In any case, that's not a blocker for now, but when @lakshmi-kannan comes up with a framework for package tests, we should add those test cases there.

dennybaa:
here https://github.com/StackStorm/st2-packages/blob/master/packages/st2/debian/st2.st2actionrunner.init#L42-L45

dennybaa:
yup not now/

dennybaa:
@Kami there's actually nothing really to come up with package tests should be serverspec/rspec. It's the most easiest and appropriate framework which is available (on the it market as i'm aware) for this need.
However they should not be as bad as the current tests in st2-packages. I still see no time and no necessity to refactor them from zero.
Also it's desirable to remove all rake build code from st2-packages, it's complexity and the whole existence were defined by wrong task input --- *we want to build packages per component*.

PR #2554, opened 2016-03-03
Setup insecure and cacert properties for mistral

manasdk:
Follow up to https://review.openstack.org/#/c/286942/3

This will allow StackStorm to use mistral-api with a self signed SSL cert


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1457117315006163

m4dcoder:
Looks good! +1

Issue #2557, opened 2016-03-04
Rule criteria needs more regex operators


pixelrebel:
Need support for a case-insensitive `imatchregex` operator.

Additionally, the `matchregex` operator searches from the start of a string.  So support for `re.search` is also necessary.  I propose adding `searchregex` and `isearchregex` operators too.

pixelrebel:
Here's the PR:  https://github.com/StackStorm/st2/pull/2558

pixelrebel:
Merged, Thanks!

Kami:
Thank you :)

PR #2558, opened 2016-03-04
Add imatchregex, searchregex, and isearchregex operators.

pixelrebel:
Addresses issue https://github.com/StackStorm/st2/issues/2557

### DIFF
--- a/st2common/st2common/operators.py
+++ b/st2common/st2common/operators.py
@@ -122,30 +122,26 @@ def match_wildcard(value, criteria_pattern):
 
 
 def match_regex(value, criteria_pattern):
+    # match_regex is deprecated, please use 'regex' and 'iregex'
     if criteria_pattern is None:
         return False
     regex = re.compile(criteria_pattern, re.DOTALL)
     # check for a match and not for details of the match.
     return regex.match(value) is not None
 
-def imatch_regex(value, criteria_pattern):
-    if criteria_pattern is None:
-        return False
-    regex = re.compile(criteria_pattern, re.DOTALL | re.IGNORECASE)
-    # check for a match and not for details of the match.
-    return regex.match(value) is not None
 
-def search_regex(value, criteria_pattern):
+def regex(value, criteria_pattern):
     if criteria_pattern is None:
         return False
-    regex = re.compile(criteria_pattern, re.DOTALL)
+    regex = re.compile(criteria_pattern)
     # check for a match and not for details of the match.
     return regex.search(value) is not None
 
-def isearch_regex(value, criteria_pattern):
+
+def iregex(value, criteria_pattern):
     if criteria_pattern is None:
         return False
-    regex = re.compile(criteria_pattern, re.DOTALL | re.IGNORECASE)
+    regex = re.compile(criteria_pattern, re.IGNORECASE)
     # check for a match and not for details of the match.
     return regex.search(value) is not None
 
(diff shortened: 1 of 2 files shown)


st2common/st2common/operators.py
```diff
@@ -114,19 +113,33 @@ def greater_than(value, criteria_pattern):
...
+        return False
+    regex = re.compile(criteria_pattern)
+    # check for a match and not for details of the match.
+    return regex.search(value) is not None
 
-def match_regex(value, criteria_pattern):
+def isearch_regex(value, criteria_pattern):
     if criteria_pattern is None:
         return False
-    regex = re.compile(criteria_pattern, re.DOTALL)
```
manasdk: Why is DOTALL removed? We can look back atthe recent PR where this was added to understand why it is needed.

st2common/st2common/operators.py
```diff
@@ -169,8 +182,10 @@ def nexists(value, criteria_pattern):
     return value is None
 
 # operator match strings
-MATCH_WILDCARD = 'matchwildcard'
```
manasdk: generally not a good idea to remove operators. This will cause someones rule etc to break.

st2common/st2common/operators.py
lakshmi-kannan: I think DOTALL was added to allow regex to work with \n in the string IIRC. Please don't remove DOTALL.

pixelrebel:
Hmm, that's weird, I updated my branch to master before I started, but perhaps it was behind.  Let me check...

Kami:
Thanks for the contribution.

In general the changes look good to me, but I wonder if we really need both (match and search). I think it might be a good idea to only have one which behaves like `re.search` (that's usually what users expects anyway).

If user wants match-like behavior when using search, they can still do that by further limiting the regular expression using `^` and `$`.

Kami:
In any case, with whatever approach we end up going, we need corresponding documentation updates and make it clear what are the differences between all those operators (additional section in the docs on differences between those operators is fine).

pixelrebel:
@kami:  I agree that one implementation is probably enough.  Personally, I almost always use `re.search` and limit accordingly, like you said.  But I'm sure there are others who use `re.match` and use `.*` to open it up.  I'll leave it up to you to decide.  Changing the current `matchregex` operator to use `re.search` could break users code.  So that's why I created a new operator for myself. :P

pixelrebel:
Perhaps we create simply `regex` and `iregex`.  And deprecate `matchregex`?

Kami:
@pixelrebel I like that idea.

You can go ahead and make that change if you are up for it. This also means we need corresponding documentation changes (update operators table, update upgrade notes, document `matchregex` has been deprecated, but we will still leave the code for now to avoid breaking changes).

pixelrebel:
Sure, glad to.   Can you help me figure out what I need to do to pass travis?

Kami:
Will have a look.

st2common/tests/unit/test_operators.py
```diff
@@ -45,6 +45,51 @@ def test_matchregex(self):
...
+        
+        string = 'fooponiesbarfooooo'
+        self.assertTrue(op(string, 'ponies'), 'Failed searchregex.')
+
+    def test_searchregex_fail(self):
+        op = operators.get_operator('searchregex')
+        self.assertFalse(op('v1_foo', 'v1$'), 'Passed searchregex.')
+
+        string = 'fooPONIESbarfooooo'
+        self.assertTrue(op(string, 'ponies'), 'Passed searchregex.')
```
Kami: This test is failing. It looks like this should be `assertFalse` since it's not a case insensitive search?

Or am I missing something?

Kami: Other failures seem to be mostly minor style things - https://travis-ci.org/StackStorm/st2/jobs/113746955#L350

Kami: Let me know when those issues have been addressed and I will go ahead and merge it.

pixelrebel:
@kami, I don't think we want `re.DOTALL` for the new `regex`/`iregex` operators.  Since `re.search` scopes the whole string, I think you would want the option to explicitly match new lines.  What do you think?

Kami:
@pixelrebel I believe you are right. Just make sure we document that :)

Kami:
Thanks!

I went ahead and merged PR into master.

PR #2559, opened 2016-03-04
st2-submit-debug-info #2529 Follow Up

vCabbage:
Follow up for PR #2529, as requested by @Kami.

* Following errors are handled when specify a custom config:
    * Config file does not exist
    * Config file is not valid YAML
    * YAML is not a dict
* Changed `yaml.load` to `yaml.safe_load` to avoid potential for privilege escalation when a non-privileged user is allowed to execute the script as root.
* Added missing docstring for `add_user_info`.
* Added tests to verify that shell command files are created and that the output is correctly written to the files.
* Added tests to verify that custom config does override default options.

Let me know if there are any other test cases you'd like added.

### DIFF
--- a/st2debug/st2debug/cmd/submit_debug_info.py
+++ b/st2debug/st2debug/cmd/submit_debug_info.py
@@ -596,6 +596,8 @@ def main():
                         help='Specify an existing file to operate on')
     args = parser.parse_args()
 
+    setup_logging()
+
     # Ensure that not all options have been excluded
     abort = True
     for arg_name in ARG_NAMES:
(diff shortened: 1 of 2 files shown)


Kami:
@vCabbage great, thanks a lot for working on this!

I will have a look shortly.

st2debug/st2debug/cmd/submit_debug_info.py
```diff
@@ -607,8 +615,16 @@ def main():
 
     # Get setting overrides from yaml file if specified
     if args.config:
-        with open(args.config, 'r') as yaml_file:
-            config_file = yaml.load(yaml_file)
+        try:
+            with open(args.config, 'r') as yaml_file:
+                config_file = yaml.safe_load(yaml_file)
```
Kami: Good call, totally missed that.

st2debug/tests/integration/test_submit_debug_info.py
```diff
@@ -120,22 +120,55 @@ def test_create_archive_include_all(self):
 
     def _get_yaml_config(self):
         return {
-            'log_file_paths': [
+            'log_files_paths': [
```
Kami: Can we please change this back to `log_file_path` so it's consistent with existing `config_file_paths` (or vice versa, make sure it's `files` in both places :))

Kami: Besides that, LGTM

vCabbage: Will do, st2docs says `log_file_paths`, so I'll stick with that.

Kami:
@vCabbage Thanks for pushing that change.

I will go ahead and merge it into master once CI finishes.

Kami:
Merged, thanks again!

PR #2560, opened 2016-03-07
Fix cryptic error messages in nginx config

enykeev:
I've also made some changes to Stream API endpoint to fix backward compatibility and make it more compact.

### DIFF
--- a/conf/nginx/st2.conf
+++ b/conf/nginx/st2.conf
@@ -1,5 +1,5 @@
 #
-# nginx configuration to expose st2 webui, redirect HTTP->HTTPS, 
+# nginx configuration to expose st2 webui, redirect HTTP->HTTPS,
 # provide SSL termination, and reverse-proxy st2api and st2auth API endpoint.
 # To enable:
 #    cp ${LOCATION}/st2.conf /etc/nginx/sites-available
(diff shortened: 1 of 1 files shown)


conf/nginx/st2.conf
```diff
@@ -43,7 +43,14 @@ server {
   add_header              Front-End-Https on;
   add_header              X-Content-Type-Options nosniff;
 
+  location @apiError {
+    add_header Content-Type application/json;
+    return 200 '{ "faultstring": "Nginx is unable to reach st2api. Make sure service is running." }';
```
enykeev: 200 is of course an awful return code here, but nginx doesn't allow to change headers on 4xx and 5xx responses up until 1.7.5 version and ubuntu's current one is 1.4.6. We'll fix that in the next commit.

conf/nginx/st2.conf
```diff
@@ -62,57 +69,20 @@ server {
     proxy_set_header Host $host;
   }
 
-  location /stream/ {
```
Kami: Yeah, I just did what was the fastest so it included some code duplication.

Kami:
If you tested this config end to end, I'm LGTM, +1.

enykeev:
The latest commit require nginx 1.7.5+

enykeev:
screen shot 2016-03-08 at 3 24 48 pm

No more cryptic HTML messages.

Kami:
> The latest commit require nginx 1.7.5+

Is this a problem or not?

Kami:
I checked the docs and we don't really mention minimum version of software (MongoDB, RabbitMQ, nginx) we require in the docs.

We should fix that and make it explicit. That's something I already complained about in the past (we should ideally control the versions ourselves with docker or similar since relying on upstream packages is usually a bad idea, but anyway).

enykeev:
> Is this a problem or not?

It was a problem before https://github.com/StackStorm/st2-packages/pull/218 since ubuntu trusty ships 1.4.6 in their official repos.

enykeev:
I'm not going to backport it to 1.3 since it's mostly cosmetic. It also includes changes for stream which is not present in 1.3 either.

PR #2569, opened 2016-03-09
Add support for st2-packages branching (master)

armab:
> Part of https://github.com/StackStorm/st2-packages/pull/227

#### What this change brings
* pushing to `master` of `st2` repo will generate packages based on `master` of `st2-packages`
* pushing to `v1.3` of `st2` repo will generate packages based on `v1.3` of `st2-packages`
  * If there is no `v1.3` branch in `st2-packages` repo, then `master` will be used as a fallback
  * ^^ This race case would happen after automating branch creation in `st2-packages` repo (`st2` branch was created, but `st2-packages` branch wasn't created yet)

----
From this point `st2` and `st2-packages` repos should have the same branches `vX.Y`.

cc @manasdk to include automated branch creation in `st2-packages` repo as part of release machinery (same as you do for `st2web`).

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -21,7 +21,6 @@ machine:
     DISTROS: "wheezy jessie trusty el6 el7"
     NOTESTS: "el7"
     ST2_PACKAGES_REPO: https://github.com/StackStorm/st2-packages
-    ST2_PACKAGES_BRANCH: master
     ST2_DOCKERFILES_REPO: https://github.com/StackStorm/st2-dockerfiles
     BUILD_DOCKER: 0
     DEPLOY_DOCKER: 0
@@ -40,7 +39,10 @@ machine:
 
 checkout:
   post:
-    - git clone --depth 1 -b ${ST2_PACKAGES_BRANCH} ${ST2_PACKAGES_REPO} /home/ubuntu/st2/st2-packages
+    - |
+      git clone --depth 1 ${ST2_PACKAGES_REPO} /home/ubuntu/st2/st2-packages
+      cd /home/ubuntu/st2/st2-packages
+      git checkout ${CIRCLE_BRANCH} || true
     - .circle/buildenv.sh
 
 dependencies:


circle.yml
```diff
@@ -40,7 +39,10 @@ machine:
 
 checkout:
   post:
-    - git clone --depth 1 -b ${ST2_PACKAGES_BRANCH} ${ST2_PACKAGES_REPO} /home/ubuntu/st2/st2-packages
+    - |
+      git clone --depth 1 ${ST2_PACKAGES_REPO} /home/ubuntu/st2/st2-packages
+      cd /home/ubuntu/st2/st2-packages
+      git checkout ${CIRCLE_BRANCH} || true
```
armab: Here we suppress `git checkout` error code to avoid race, when remote `st2-packages` branch wasn't created yet.
So it works as a fall-back to default `master` branch if there is no `vX.Y` branch in packages.

enykeev: I thought of that, but what bothers me is that we're implicitly treating master as default. There's a whole set of situations when mistake may cost us putting broken package to the repo. As I see it, if we don't have a sister branch, we fail, loud.

armab: Yeah, but it's circular dependency: 

* You can't build packages in `st2` repo before you create branch in `st2-packages` repo
* You can't build packages in `st2-packages` repo before you create branch in `st2` repo

So `master` is a fallback here.


As I know current release machinery works this way: 
- Create `vX.Y` branch in `st2` repo
- Create `vX.Y` branch in dependent repos like `st2web` (`st2-packages` to add)


^^ @manasdk is this order correct or no? (to make sure and if needed - adjust the logic here).

enykeev: Well, yes, it is a circular dependency and yes, there's no way to resolve it without failing or falling back to master. My point is that it's better to fail because failing has no side effects.

enykeev: I'm still hoping someday we would just merfe st2 and st2-packages and forget about this problem once and for all.

armab: ^^ 
Yeah, initially, we couldn't think seriously about such option at all, because we couldn't touch entire existing `st2` codebase and release machinery with our `st2-packages` changes.
So we kept that package logic isolated in new repo to avoid overflowing other team members (especially considering we had a lot of experiments and pretty big code flow).

When `st2-packages` matures and stabilizes, it's good to consider pros/cons again for keeping it in one or separated repos :+1:

PR #2571, opened 2016-03-09
Add a custom gunicorn sync worker implementation which works around some issues

Kami:
This includes a custom sync worker implementation which includes a work-around for shutdown related issues when using sync worker in combination with eventlet monkey patching,.

### DIFF
--- a/st2common/st2common/util/gunicorn_workers.py
+++ b/st2common/st2common/util/gunicorn_workers.py
@@ -1,0 +1,43 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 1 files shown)


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1457555334007138

manasdk:
@kami - where did you find out that this was the workaround. Did you read gunicorn code to decide which method to overwrite?

Kami:
@manasdk Yep, I checked the sync and base worker implementation.

Without that it fails with the error below inside `handle_quit` (https://github.com/benoitc/gunicorn/blob/master/gunicorn/workers/base.py#L170).

```python
    assert hub.greenlet is not current, 'do not call blocking functions from the mainloop'
AssertionError: do not call blocking functions from the mainloop
```

I'm not too big of a fan of this work-around, but dunno. There are also other issues with async (eventlet) based worker.

manasdk:
@kami If you want to still merge this change please include reference to code from gunicorn in the new file so it is easy for someone to track this down in the future.

Kami:
@manasdk Does af54ffb8571592fec12d4da29db626f45d8a2121 make it any more clear?

PR #2572, opened 2016-03-10
[RFR] Add the `extra` field to action aliases

emedvedev:
Different chat adapters have special capabilities which are sometimes very handy and nice. Slack has attachment colors and parsing/unfurling options for better formatting; HipChat has Connect features; and so on.

It’s hard for users to add support for those right now, because you have to modify the schema and the DB connector in st2, and only then submit a PR to `hubot-stackstorm` utilizing the new parameter. More importantly, if we had a lot of requests for this, our schema would be overblown with rarely used parameters.

A solution for this,—which has been discussed some weeks ago, but I only got around to it now,—is an `extras` object inside the alias with a “free” schema. It will allow people to easily add support for extra features with relative ease and without modifying `st2` (which would imply only getting a feature out in the next `st2` release and complicating the schema). While this is not ideal architecturally (a reason the schema exists in the first place is to have order, and this breaks it a bit), I feel like pros outweigh the cons, and it’s worth to introduce this. We still maintain strict schema outside `extras`, after all.

Another neat feature this PR includes is rendering extras with Jinja, sharing the context with `ack` or `result`, depending on the placement. E.g. different colors depending on the execution status.

This can also be useful for possible usage of action aliases outside chat (e-mails, sms, tweets, and whatnot).

An example:
```
name: "tesla.honk_horn"
action_ref: "tesla.honk_horn"
description: "Honk your horn"
formats:
  - “honk-honk”
result:
  format: "**Honk Honk**"
  extras:
    color: “<% if result.succeeded %> green <% else %> red <% endif %>”
```

What do you think?

### DIFF
--- a/contrib/chatops/actions/format_execution_result.py
+++ b/contrib/chatops/actions/format_execution_result.py
@@ -42,8 +42,8 @@ class FormatResultAction(Action):
                     raise Exception("Output of this template is disabled.")
                 if 'format' in alias.result:
                     template = alias.result['format']
-                if 'extras' in alias.result:
-                    result['extras'] = jinja_utils.render_values(alias.result['extras'], context)
+                if 'extra' in alias.result:
+                    result['extra'] = jinja_utils.render_values(alias.result['extra'], context)
 
         result['message'] = self.jinja.from_string(template).render(context),
 
--- a/contrib/chatops/actions/format_execution_result.py
+++ b/contrib/chatops/actions/format_execution_result.py
@@ -26,6 +26,7 @@ class FormatResultAction(Action):
             'execution': execution
         }
         template = self.default_template
+        result = {}
 
         alias_id = execution['context'].get('action_alias_ref', {}).get('id', None)
         if alias_id:
@@ -35,14 +36,18 @@ class FormatResultAction(Action):
                 'alias': alias
             })
 
-            result = getattr(alias, 'result', None)
-            if result:
-                if not result.get('enabled', True):
+            result_params = getattr(alias, 'result', None)
+            if result_params:
+                if not result_params.get('enabled', True):
                     raise Exception("Output of this template is disabled.")
                 if 'format' in alias.result:
                     template = alias.result['format']
+                if 'extras' in alias.result:
+                    result['extras'] = jinja_utils.render_values(alias.result['extras'], context)
 
-        return self.jinja.from_string(template).render(context)
+        result['message'] = self.jinja.from_string(template).render(context),
+
+        return result
 
     def _get_execution(self, execution_id):
         if not execution_id:
--- a/st2common/st2common/models/db/actionalias.py
+++ b/st2common/st2common/models/db/actionalias.py
@@ -65,8 +65,8 @@ class ActionAliasDB(stormbase.StormBaseDB, stormbase.ContentPackResourceMixin,
     result = me.DictField(
         help_text='Parameters pertaining to the execution result message.'
     )
-    extras = me.DictField(
-        help_text='Additional parameters not covered in the schema.'
+    extra = me.DictField(
+        help_text='Additional parameters (usually adapter-specific) not covered in the schema.'
     )
 
     meta = {
--- a/st2common/st2common/models/db/actionalias.py
+++ b/st2common/st2common/models/db/actionalias.py
@@ -65,6 +65,9 @@ class ActionAliasDB(stormbase.StormBaseDB, stormbase.ContentPackResourceMixin,
     result = me.DictField(
         help_text='Parameters pertaining to the execution result message.'
     )
+    extras = me.DictField(
+        help_text='Additional parameters not covered in the schema.'
+    )
 
     meta = {
         'indexes': ['name']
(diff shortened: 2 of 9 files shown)


estee-tew:
https://stackstorm.slack.com/archives/stackstorm/p1457592739007163

Kami:
Yeah, I was in favor of `extra` approach in the past so I'm +1 on the overall idea.

contrib/chatops/actions/format_execution_result.py
```diff
@@ -35,14 +36,18 @@ def run(self, execution_id):
...
-            if result:
-                if not result.get('enabled', True):
+            result_params = getattr(alias, 'result', None)
+            if result_params:
+                if not result_params.get('enabled', True):
                     raise Exception("Output of this template is disabled.")
                 if 'format' in alias.result:
                     template = alias.result['format']
+                if 'extras' in alias.result:
+                    result['extras'] = jinja_utils.render_values(alias.result['extras'], context)
```
Kami: Minor style / consistency thing - `extras` -> `extra`

st2common/st2common/models/db/actionalias.py
```diff
@@ -65,6 +65,9 @@ class ActionAliasDB(stormbase.StormBaseDB, stormbase.ContentPackResourceMixin,
     result = me.DictField(
         help_text='Parameters pertaining to the execution result message.'
     )
+    extras = me.DictField(
+        help_text='Additional parameters not covered in the schema.'
```
Kami: Maybe clarify that those can be adapter specific attributes and things like that.

Kami:
Docs and some tests for this change would be nice, besides that, LGTM :+1:

emedvedev:
Updated the test fixtures to account for the new schema, and docs are coming along as a part of a more general chatops PR that DZ requested. I want to do an example in `hubot-stackstorm`, too, to show how to use extras with some actual code.

Changed to `extra`.

manasdk:
I see the need for `extra` and am +1 to the idea and PR.

Free schemas generally pose are good and bad for exactly the same reason i.e. you can put anything. In this case it is not really a free schema but an adapter specific schema so depending on the eventual chat service/tool the supported properties are fixed. The corollary is that it is likely that an alias will have adapter specific properties which as I understand this impl may or may not apply.  

**Thoughts on mitigation**
1. As long as alias formatting degrades gracefully it should be acceptable. Can this be guaranteed?
2. Can we document the fixed set of properties for `slack` or `hipchat`? Even if we start with slack by giving a few examples it should be acceptable.
3. `extras` is generic as it does not suggest where the property will be interpreted - can we make this specific like `extra_adapter_info` - yeah this is gross and I do not like the name but suggesting the direction.

manasdk:
> Another neat feature this PR includes is rendering extras with Jinja, sharing the context with ack or result, depending on the placement. E.g. different colors depending on the execution status.

This is really nice :)

ajstuart:
+1

pixelrebel:
I'm excited to see this, nice work Ed!

emedvedev:
> 1. As long as alias formatting degrades gracefully it should be acceptable i.e. if there is a slack specific property then will hipchat ignore or barf. Can this be guaranteed? If not can we doc this information?

It will ignore. If you want to support a certain `extra`, you extend `hubot-stackstorm` and check for the extra in a formatter for the adapter you want. Other formatters will not even know it's there.

> 2. Can we document the fixed set of properties for slack or hipchat? Even if we start with slack by giving a few examples it should be acceptable.

Yeah, I'm working at the example for attachment colors, and then we can doc it. I'll also get a post with guidelines out, and maybe we can squeeze the guidelines into the docs, too (we really should).

> 3. extras is generic as it does not suggest where the property will be interpreted - can we make this specific like extra_adapter_info - yeah this is gross and I do not like the name but suggesting the direction.

I actually did plan for it to be generic. This part is debatable, but we can add adapter-agnostic params here, too: I was considering playing with confirmations, for example. We can even add client-specific params here—like a subject line if someone wants to use chatops with e-mail or something (an unlikely case, just the first I could come up with). Hence the very generic name.

emedvedev:
Also see https://github.com/StackStorm/hubot-stackstorm/pull/101

Kami:
Feel free to merge this once you add some tests and changelog has been updated :+1:

emedvedev:
@Kami: nice! The test fixtures are already updated (and there's nothing else to test because the field doesn't really do anything), I'll update the changelog tomorrow and merge.

Kami:
@emedvedev Cool :)

I thought it might also be a good idea to just add a test case which makes sure `extra` parameter is actually stored / written in the DB, but that's not a big deal (as long as we have some higher level tests which exercise this parameter).

manasdk:
@emedvedev 

> I actually did plan for it to be generic. This part is debatable, but we can add adapter-agnostic params here, too: I was considering playing with confirmations, for example. We can even add client-specific params here—like a subject line if someone wants to use chatops with e-mail or something (an unlikely case, just the first I could come up with). Hence the very generic name.

I would recommend keeping a very limited scope of properties and say extra is only for properties we transparently pass down to an adapter. Enabling arbitrary features without taking care to define the properties leads to extremely brittle APIs. A loose definition of generic i.e. we have no clue what the schema can be since its adapter specific to magic new features are available via `extra` can get very sketch.

PR #2576, opened 2016-03-11
Add "--register--setup-virtualenvs" flag to the register-content script

Kami:
This flag is to be used in a distributed StackStorm setup where action runner services run on multiple hosts.

In the distributed deployment context, (right now) it's up to operator to ensure register-content script with this flag is ran on every host where action runner service is running. This can be tied into the whole "content distribution" story - user can hook it up as a last step of their "deploy content to the server" story.

TODO:

- [ ] Integration tests for virtualenv creation (we can do that now that code is split out)

### DIFF
--- a/st2common/st2common/content/bootstrap.py
+++ b/st2common/st2common/content/bootstrap.py
@@ -263,8 +263,8 @@ def register_content():
 
 def setup(argv):
     if '--register-setup-virtualenvs' in argv:
-        # If setup-virtualenvs option is used we also want to setup config for action runner since it
-        # contains some setup virtualenv related options.
+        # If setup-virtualenvs option is used we also want to setup config for action runner since
+        # it contains some setup virtualenv related options.
         # Keep in mind that using "setup-virtualenv" only makes sense on nodes which run action
         # runners.
         # Note: This is a hack, but we can't check if that flag is provided using oslo config since
@@ -274,7 +274,7 @@ def setup(argv):
     else:
         config = common_config
 
-    common_setup(config=common_config, setup_db=True, register_mq_exchanges=True)
+    common_setup(config=config, setup_db=True, register_mq_exchanges=True)
 
 
 def teardown():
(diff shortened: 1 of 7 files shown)


st2common/st2common/content/bootstrap.py
```diff
@@ -219,8 +257,23 @@ def register_content():
     if cfg.CONF.register.policies:
         register_policies()
 
+    if cfg.CONF.register.setup_virtualenvs:
+        setup_virtualenvs()
+
 
 def setup(argv):
+    if '--register-setup-virtualenvs' in argv:
+        # If setup-virtualenvs option is used we also want to setup config for action runner since
```
Kami: This is a hack, but it's sadly not possible to do it differently because args need to be parsed before we can check `cfg.CONF...`.

And we can't just always use action runner config since `st2actions` doesn't necessary be installed on every host (well, that will always be true with the new packages, but if someone does something funky this would not always be true).

st2common/st2common/content/bootstrap.py
```diff
@@ -219,8 +257,23 @@ def register_content():
...
+        # it contains some setup virtualenv related options.
+        # Keep in mind that using "setup-virtualenv" only makes sense on nodes which run action
+        # runners.
+        # Note: This is a hack, but we can't check if that flag is provided using oslo config since
+        # we need to parse config first.
+        from st2actions import config as action_runner_config
```
manasdk: Can you put this in a try ... except to print a useful exception message? It would be nice to know that module import failed because st2actions is not present.

Another alternative is to move the action_runner_config to st2common. We have already done that in a few cases. Again not ideal but possible.

Kami: Yeah, we could do that.

As you have said, it's not ideal but we already do it for a lot of things in st2api, etc.

I can take that approach, since I personally think it's the least hacky.

manasdk:
+1

Kami:
I'll merge this and add tests in a separate PR.

Kami:
While discussing this on IRC, I noticed I forgot to add a flag to `st2ctl`. Will do that tomorrow.

PR #2579, opened 2016-03-15
Support crontab like notation for CronTimer trigger

Kami:
This pull request updates our JSON schema validation code so that the CronTimer now also accepts other cron-like notation (`a-b`, `*/a`, etc.) which is supported by `CronTrigger` in apscheduler (http://apscheduler.readthedocs.org/en/3.0/modules/triggers/cron.html#api).

In addition to that, it also introduces additional validation for CronTimer parameters - this is required since we now allow arbitrary strings for parameter values, but not any arbitrary string is valid `CronTimer` argument value.

## TODO

- [x] Code refactor
- [x] Tests
- [x] Documentation changes

### DIFF
--- a/st2common/st2common/constants/triggers.py
+++ b/st2common/st2common/constants/triggers.py
@@ -339,7 +339,8 @@ TIMER_PAYLOAD_SCHEMA = {
     }
 }
 
-INTERVAL_TIMER_TRIGGER_REF = ResourceReference.to_string_reference(SYSTEM_PACK_NAME, 'st2.IntervalTimer')
+INTERVAL_TIMER_TRIGGER_REF = ResourceReference.to_string_reference(SYSTEM_PACK_NAME,
+                                                                   'st2.IntervalTimer')
 DATE_TIMER_TRIGGER_REF = ResourceReference.to_string_reference(SYSTEM_PACK_NAME, 'st2.DateTimer')
 CRON_TIMER_TRIGGER_REF = ResourceReference.to_string_reference(SYSTEM_PACK_NAME, 'st2.CronTimer')
 
--- a/st2common/st2common/constants/triggers.py
+++ b/st2common/st2common/constants/triggers.py
@@ -252,40 +252,64 @@ CRON_PARAMETERS_SCHEMA = {
             "type": "string"
         },
         "year": {
-            "type": "integer"
+            "anyOf": [
+                {"type": "string"},
+                {"type": "integer"}
+            ],
         },
         "month": {
-            "type": "integer",
+            "anyOf": [
+                {"type": "string"},
+                {"type": "integer"}
+            ],
             "minimum": 1,
             "maximum": 12
         },
         "day": {
-            "type": "integer",
+            "anyOf": [
+                {"type": "string"},
+                {"type": "integer"}
+            ],
             "minimum": 1,
             "maximum": 31
         },
         "week": {
-            "type": "integer",
+            "anyOf": [
+                {"type": "string"},
+                {"type": "integer"}
+            ],
             "minimum": 1,
             "maximum": 53
         },
         "day_of_week": {
-            "type": "integer",
+            "anyOf": [
+                {"type": "string"},
+                {"type": "integer"}
+            ],
             "minimum": 0,
             "maximum": 6
         },
         "hour": {
-            "type": "integer",
+            "anyOf": [
+                {"type": "string"},
+                {"type": "integer"}
+            ],
             "minimum": 0,
             "maximum": 23
         },
         "minute": {
-            "type": "integer",
+            "anyOf": [
+                {"type": "string"},
+                {"type": "integer"}
+            ],
             "minimum": 0,
             "maximum": 59
         },
         "second": {
-            "type": "integer",
+            "anyOf": [
+                {"type": "string"},
+                {"type": "integer"}
+            ],
             "minimum": 0,
             "maximum": 59
         }
--- a/st2common/st2common/constants/triggers.py
+++ b/st2common/st2common/constants/triggers.py
@@ -32,7 +32,11 @@ __all__ = [
     'WEBHOOK_TRIGGER_TYPES',
     'WEBHOOK_TRIGGER_TYPE',
     'INTERNAL_TRIGGER_TYPES',
-    'SYSTEM_TRIGGER_TYPES'
+    'SYSTEM_TRIGGER_TYPES',
+
+    'INTERVAL_TIMER_TRIGGER_REF',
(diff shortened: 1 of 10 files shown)


st2common/st2common/validators/api/reactor.py
```diff
@@ -70,4 +74,12 @@ def validate_trigger_parameters(trigger_type_ref, parameters):
     cleaned = util_schema.validate(instance=parameters, schema=parameters_schema,
                                    cls=util_schema.CustomValidator, use_default=True,
                                    allow_default_none=True)
+
+    # Additional validation fron CronTimer
+    if trigger_type_ref == ResourceReference.to_string_reference(SYSTEM_PACK_NAME, 'st2.CronTimer'):
```
Kami: Yes, I will refactor this in a constant. I just quickly wanted to test it if it works.

st2common/st2common/constants/triggers.py
```diff
@@ -252,40 +252,64 @@
...
+                {"type": "integer"}
+            ],
         },
         "month": {
-            "type": "integer",
+            "anyOf": [
+                {"type": "string"},
+                {"type": "integer"}
+            ],
             "minimum": 1,
```
manasdk: how does minimum and maximum work if these can also be strings?

st2common/st2common/validators/api/reactor.py
manasdk: Do you really have to validate in common code and look for a special trigger_type_ref?

st2common/st2common/constants/triggers.py
Kami: Not sure, I would need to test it.

In any case, it doesn't really matter since we now perform stricter validation inside the trigger validation code path which is executed on rule creation.

st2common/st2common/validators/api/reactor.py
```diff
@@ -15,7 +15,11 @@
 
 import six
 
+from apscheduler.triggers.cron import CronTrigger
```
manasdk: this is pretty bad I am not sure it is even work breaking the abstraction to provide the validation.

st2common/st2common/constants/triggers.py
manasdk: You may have guessed but I am pretty much `-1` to doing so :)

Kami:
```MESSAGE: 1000 is greater than the maximum of 6

Failed validating 'maximum' in schema['properties']['day_of_week']:
    {'anyOf': [{'type': 'string'}, {'type': 'integer'}, {'type': 'null'}],
     'maximum': 6,
     'minimum': 0}
```

Looks like it works as expected - minmax validation is only performed if parameter type is `integer`.

So this works as expected :)

st2common/st2common/validators/api/reactor.py
Kami: I mean our implementation is already quite coupled to apscheduler so it's not breaking any abstraction (imo). Or you mean something else?

And the same validation is already performed in rules engine where apscheduler jobs are created. This change just moves it up the stack so the validation is also performed earlier when rule is created.

This is something which is almost always desirable since it allows us to fail early and provide a better user-experience.

st2common/st2common/validators/api/reactor.py
Kami: I'm not actually sure what you mean.

Right now this function is the place where additional, trigger-specific parameter validation is performed (hence it's called `validate_trigger_parameters`).

st2common/st2common/validators/api/reactor.py
manasdk: Our timer impl is coupled to apscheduler.

rulesengine which implements the timer is the only place that even knows about specific of the trigger.

Validate as much as you like in RulesEngine but please don't bring it upstream. Just because `timer` have some special meaning since `st2` implements these does not mean they should get special preference. As much as possible only https://github.com/StackStorm/st2/blob/master/st2reactor/st2reactor/timer/base.py module should know about implementation details of timers. Anything else that is done in a non-generic way is a hack.

st2common/st2common/validators/api/reactor.py
manasdk: The fact that this code looks for a specific `trigger_type_ref` and does special handling in code that should have no clue about the specific nature of trigger types is unacceptable.

I guess it is the same as my point in a different comment. This code knows nothing about the nature of this specific trigger_type i.e. `core.st2.CronTimer` - that StackStorm even has such an internal trigger is an internal detail. I do not think we should be validating that here - given the decoupled nature of the system it should only be handled in RulesEngine. Yes this means rule will be create and lead to a 200 even though trigger parameters might be bad.

st2common/st2common/validators/api/reactor.py
Kami: I personally don't see this as an issue at all.

It might break some logical / code organization abstraction which doesn't provide much value here anyway.

Adding additional validation here allows us to provide a better user experience (we can throw early, it would get really messy and unfriendly if the exception is thrown inside rules engine when creating a timer) so I literally see no issues with that.

On top of that, we already perform validation here so I don't see how performing some additional validation breaks the abstraction. For a more specific validation (on top of what JSON schema provides) you need trigger-specific information and I see no issues with putting this here.

If we could do the same thing for other system triggers (perform additional validation which would allows us to catch issues early and provide a better user experience) I would be all for that as well.

manasdk: > On top of that, we already perform validation here so I don't see how performing some additional validation breaks the abstraction. For a more specific validation (on top of what JSON schema provides) you need trigger-specific information and I see no issues with putting this here.

The validation being performed here is generic based on the schema information provided. This  code nothing about the specifics of any trigger

There is a very good reason that we have implemented St2Timer as an encapsulated class and placed it in a module. The explicit reason is that only that code knows anything about the specifics of a Timer. We have gone out of our way to build out the abstraction of a trigger-type, trigger and post that onto a message bus which any interested party can consume. The reason we have built out these abstractions is so that there is a decoupled system and by adding this special validation here we are completely laying to waste the basis for these abstractions. I still see tremendous value in these abstractions. How can someone logically explain that this code is aware of `core.st2.CronTimer`? Why would a piece of code that knows nothing about the specific implementation details of any trigger suddenly start effectively "down-casting" to a specific trigger and perform special checks.

I personally see an issue with this and am not agreeable to this validation being performed in this code path.

manasdk:
To be amply clear my objection is only to where the validation is being performed as opposed to the feature being enabled. 

I am `+1` to the feature being enabled i.e. cron like syntax.

PR #2580, opened 2016-03-16
RFR: Call buildenv_st2.sh in circle

lakshmi-kannan:


lakshmi-kannan:
Merge this after https://github.com/StackStorm/st2-packages/pull/245 is merged.

PR #2581, opened 2016-03-17
Clean up the ChatOps aliases pack deployment

jjm:
A clean up to the ChatOps aliases for pack deployment in the packs pack.

The current way turned out to be a little hard of the fingers when deploying a pack which had no subtree. As you need to type `my-pack` twice. This removes this requirement.

### DIFF
--- a/contrib/packs/actions/deploy.yaml
+++ b/contrib/packs/actions/deploy.yaml
@@ -19,11 +19,11 @@
     branch:
       type: "string"
       description: "The Branch to deploy."
-      required: true
+      default: "master"
     packs:
       type: "array"
       description: "The Pack(s) to deploy."
-      required: true
+      default: []
     auto_deploy:
       type: "boolean"
       description: "Should this pack and branch be evulated for auto deployment."
--- a/contrib/packs/actions/workflows/deploy.yaml
+++ b/contrib/packs/actions/workflows/deploy.yaml
@@ -26,7 +26,23 @@ workflows:
 
         on-complete:
           - do_auto_install: <% $.auto_deploy = true %>
-          - do_manual_install: <% $.auto_deploy = false %>
+          - check_for_no_subtree: <% $.auto_deploy = false %>
+
+      check_for_no_subtree:
+        action: core.noop
+
+        on-complete:
+          - do_manual_install: <% len($.packs) != 0 %>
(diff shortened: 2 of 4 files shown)


contrib/packs/actions/deploy.yaml
```diff
@@ -19,11 +19,11 @@
     branch:
       type: "string"
       description: "The Branch to deploy."
-      required: true
+      default: "master"
     packs:
       type: "array"
       description: "The Pack(s) to deploy."
-      required: true
+      default: []
```
emedvedev: Are you sure `packs` shouldn't be required?

contrib/packs/actions/workflows/deploy.yaml
```diff
@@ -26,7 +26,23 @@ workflows:
...
+
+        on-complete:
+          - do_manual_install: <% len($.packs) != 0 %>
+          - publish_packs: <% len($.packs) = 0 %>
+
+      publish_packs:
+        action: core.noop
+
+        publish:
+          packs: [ <% $.repo_name %> ]
```
Kami: I'm slightly confused about this variable name here - why do we call it `packs` and not `repo_name`?

jjm: It's because if where is no sub tree in the repo the repo_name and pack are mostly likely to have the same name. With this check and publish in place when calling the install from ChatOps you do not need to supply the same text as repo_name and packs.

So before the ChatOps alias was:

`Pack deploy my-pack packs my-pack branch dev`

Is now now instead:

`Pack deploy my-pack branch dev`

In the case of st2contrib packs will always have at least a single pack to install, so publish_packs will never be reached. As this only happens if packs has a 0 length.

contrib/packs/actions/deploy.yaml
jjm: We talked about his on Slack, this is so that it's possible to install a pack which has no sub tree with out supplying the same string for repo_name and packs.

contrib/packs/actions/workflows/deploy.yaml
Kami: Ah, OK, that makes sense. Thanks for the clarification.

It would also be good to add a comment to the workflow which clarifies that :)

PR #2585, opened 2016-03-18
Trying to fix circleci notify

DoriftoShoes:


manasdk:
:+1:

circle.yml
```diff
@@ -15,9 +15,10 @@ general:
   build_dir: st2-packages
   artifacts:
     - ~/packages
-  notify:
-    webhooks:
-      - url: https://webhooks.stackstorm.net:8531/webhooks/build/events
+
+notify:
+  webhooks:
+    - url: https://webhooks.stackstorm.net:8531/webhooks/build/events
```
armab:
```yaml
- url: ${ST2_HOOK_URL}
```
and then you can set `ST2_HOOK_URL` value in Circle UI (Environment variables section of Settings) to avoid disclosing that URL in public.

DoriftoShoes: Ah good to know.  We were debating how to do this.  Thanks!

armab: Another option - we can hook `st2` here: https://github.com/StackStorm/st2/blob/c3fc470b75ddd2d1ac0e770417402b3d2ad74d00/circle.yml#L93
via pure Python.

I did that before in this PR: https://github.com/StackStorm/st2-packages/pull/42

See actual code: https://github.com/StackStorm/st2-packages/pull/42/commits/bf6964b721b84a24a49c870aa6af851fad311e45 which hooks `st2` box with all needed data.


But as talked with @lakshmi-kannan we changed approach from hooking `st2` box to saving payload as Circle artifact: https://circle-artifacts.com/gh/StackStorm/st2/909/artifacts/0/home/ubuntu/packages/payload.json

PR #2586, opened 2016-03-18
Add comments to explain packs being updated to repo_name in deploy wo…

jjm:
…rkflow.

As discussed with @kami and @emedvedev on PR #2581, this PR covers
when a pack has no subtree and the repo_name and packs should be the
same. However this can't be done in the alais so we handle it in the
workflow.


Kami:
Merged, thanks.

PR #2589, opened 2016-03-21
Revert "Use Environment variable for webhook URL"

armab:
Reverts StackStorm/st2#2587

env variable didn't worked in that section, see : 
https://circleci.com/gh/StackStorm/st2/911

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -18,7 +18,7 @@ general:
 
 notify:
   webhooks:
-    - url: ${ST2_HOOK_URL}
+    - url: https://webhooks.stackstorm.net:8531/webhooks/build/events
 
 machine:
   environment:


armab:
Another way, as described https://github.com/StackStorm/st2/pull/2585#discussion_r56725177

> Another option - we can "manually" hook `st2` here: https://github.com/StackStorm/st2/blob/c3fc470b75ddd2d1ac0e770417402b3d2ad74d00/circle.yml#L93
via pure Python.

> I did that before in this PR: https://github.com/StackStorm/st2-packages/pull/42

> See actual code: https://github.com/StackStorm/st2-packages/pull/42/commits/bf6964b721b84a24a49c870aa6af851fad311e45 which hooks `st2` box with all needed data.

> But as talked with @lakshmi-kannan we changed approach from hooking `st2` box to saving payload as Circle artifact: https://circle-artifacts.com/gh/StackStorm/st2/909/artifacts/0/home/ubuntu/packages/payload.json


@DoriftoShoes ^^ If that looks better for you, I can work on it: revive that old approach from the history, polish a bit and we'd use it as web hook.

But it really depends on your requirements, eg. what payload you need and what you want to achieve with that web hook.

PR #2590, opened 2016-03-21
Fix excludes handling.

andrew-regan:
Need to pass `excludes` to `_add_exclusion_filters()` 

Confirmed w/ @lakshmi-kannan that this was introduced when fixing a issue with logging and config loading here: 51204cb81bf42bd4e977711f0b8444eef4704fee


manasdk:
+1

Thanks for the change!

Kami:
Good catch, indeed. Thanks :)

In the future it would also be a good idea to use keyword args when calling a function.

PR #2594, opened 2016-03-22
Fix nginx config error issue #2593

### LINKED ISSUE #2593: Nginx config error
The latest Nginx config doesn't work on older nginx versions. 

```
2016/03/22 15:15:57 [emerg] 23052#0: invalid number of arguments in "add_header" directive in /etc/nginx/sites-enabled/st2.conf:47
```

I'm using Ubuntu 14.04 (LTS) which [packages nginx 1.4.6](http://packages.ubuntu.com/trusty/nginx) but the `always` keyword was added to the nginx `add_header` directive in [1.7.5](http://nginx.org/en/docs/http/ngx_http_headers_module.html).

This bug was introduced as part of a [recent change](https://github.com/StackStorm/st2/commit/13b3c74e5fb38fd38348333df040006fd9f80235).

I know, I know. I should just open a pull request. :)

/cc @enykeev

codyaray:
Fixes #2593

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -63,6 +63,9 @@ in development
   ``x,y,z``, etc.). (improvement)
 * Add new ``regex`` and ``iregex`` rule criteria operator and deprecate ``matchregex`` in favor of
   those two new operators. (new-feature) [Jamie Evans]
+* Record failures to enforce rules due to missing actions or parameter validation errors. A
+  RuleEnforcement object will be created for failed enforcements that do not lead to an
+  ActionExecution creation. (improvement)
 
 1.3.2 - February 12, 2016
 -------------------------


enykeev:
Yeah, either nginx need to be upgraded or return code needs to be changed to 2xx or 3xx. In the current form, the custom message won't be returned at all, I believe. And without proper return code, chatops would stop reconnecting after receiving default 502 once (as per EventStream spec):

> Any other HTTP response code not listed here, and any network error that prevents the HTTP connection from being established in the first place (e.g. DNS errors), must cause the user agent to fail the connection.

(https://www.w3.org/TR/2009/WD-eventsource-20091029/#processing-model)

codyaray:
Isn't this only going to be returned on 502 errors?

```
error_page 502 = @apiError;
<snip>
error_page 502 = @authError;
```

Will chatops not be broken anyway if one of these is down? So it seems like failing the chatops connection would be fine in this case. Otherwise if a different error code works, that seems fine too.

Otherwise, yeah, if ya'll want to make a non-packaged nginx version a hard requirement, then just need to add a note in the docs that its now required to run nginx 1.7.5+ and possibly include

```
sudo add-apt-repository ppa:nginx/stable
sudo apt-get update
sudo apt-get install nginx
```

PR #2596, opened 2016-03-23
Get Paramiko runner to support passphrase

emedvedev:
This enables a `passphrase` parameter for remote actions.


Kami:
LGTM :+1: 

(thinking out loud) As far as parameter name goes, I'm fine with `passphrase` (something like `key_file_password` might be a bit too verbose, but I could be convinced otherwise).

Kami:
It would also be good to add a unit / integration test which utilizes a dummy fixture key which is password protected.

Kami:
I also need to dig into the code a bit because most of the Paramiko stuff was taken from Libcloud and IIRC, Libcloud already supported password protected key files at some point - if a private key file based auth is used, `password` parameter refers to the key password and not password used to authenticate.

It's possible though that the code has diverged and / or my memory doesn't serves me well :)

emedvedev:
> if a private key file based auth is used, password parameter refers to the key password and not password used to authenticate

See https://github.com/StackStorm/st2/pull/2596/files#diff-482702b071548ded4f874e5f662ae19cL534: if a private key is used, no password is passed when trying to get `pkey`. Actually, using password parameter is a good approach which we could adopt, but having a separate `passphrase` parameter is a bit more intuitive to my taste.

Kami:
Yeah, I think having a separate parameter is actually better since it's more explicit and we can (and should) throw if `passphrase` is provided, but `private_key` parameter is not.

emedvedev:
@Kami: added a passphrase-protected dummy key and a test. Looks good?

st2actions/tests/unit/test_paramiko_ssh.py
```diff
@@ -122,6 +122,31 @@ def test_key_material_argument_invalid_key(self):
                                 expected_msg, mock.connect)
 
     @patch('paramiko.SSHClient', Mock)
+    def test_key_with_passphrase(self):
+        path = os.path.join(get_resources_base_path(),
+                            'ssh', 'dummy_rsa_passphrase')
```
Kami: Let's also add a negative test case when invalid passphrase is provided - I assume error should be propagated all the way up, right?

st2actions/st2actions/runners/ssh/paramiko_ssh.py
```diff
@@ -113,6 +113,7 @@ def __init__(self, hostname, port=22, username=None, password=None, bastion_host
         self.bastion_host = bastion_host
         self.bastion_client = None
         self.bastion_socket = None
+        self.passphrase = passphrase
```
Kami: I think it's a good idea to throw here if `passphrase` is provided, but private key is not.

st2actions/st2actions/runners/ssh/paramiko_ssh.py
```diff
@@ -478,14 +479,14 @@ def _get_decoded_data(self, data):
...
+    def _get_pkey_object(self, key_material, passphrase):
         """
         Try to detect private key type and return paramiko.PKey object.
         """
 
         for cls in [paramiko.RSAKey, paramiko.DSSKey, paramiko.ECDSAKey]:
             try:
-                key = cls.from_private_key(StringIO(key_material))
+                key = cls.from_private_key(StringIO(key_material), password=passphrase)
             except paramiko.ssh_exception.SSHException:
```
Kami: I wonder what kind of exception is thrown on invalid passphrase... Is it the same one with different message or?

emedvedev: It's nearly the same one, with an addition about BER error on RSA keys; it looks like it wouldn't be trivial to distinguish between an invalid key and an invalid passphrase, and Paramiko doesn't provide a separate exception.

st2actions/st2actions/runners/ssh/paramiko_ssh.py
emedvedev: Done

st2actions/tests/unit/test_paramiko_ssh.py
emedvedev: Done

st2actions/st2actions/runners/ssh/paramiko_ssh.py
Kami: That's unfortunate :/

Kami: One thing we can and should do is at least update the error message we thrown down below.

Something along the long lines of "Invalid or unsupported key type or invalid key password).

emedvedev: That's a good idea! Done.

Kami:
Let's also update changelog :)

Besides that, LGTM, +1

emedvedev:
Done, merging :)

manasdk:
:+1:

Kami:
I worked on a related change today and noticed `passphrase` argument wasn't correctly propagated to the ParallelSSHClient constructor so this functionality never really worked - a5dd9093c48a3a6ec85e01416f7edc36934900a1.

In addition to that, passphrase was only supported when providing key material directly and not when providing path to the key file - 9be9fd74f70d6b1b8ee92ba3fa0fa60d880c65d4.

I fixed both of those things and I will also add tests.

PR #2597, opened 2016-03-23
Improve output for core.sendmail when nothing sent

codyaray:
Want to be able to easily tell if an email was sent. The current message "Successfully sent mail" is misleading when we actually skip sending mail since the body is empty and we've selected not to send an empty body.

Of course, I'm open to changing the wording.


Kami:
LGTM.

Kami:
Merged, thanks.

PR #2601, opened 2016-03-25
Fix branch checkout logic

manasdk:
`--depth 1` means no branch info therefore the checkout will fail.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -39,11 +39,8 @@ machine:
 
 checkout:
   post:
-    - |
-      git clone --depth 1 ${ST2_PACKAGES_REPO} /home/ubuntu/st2/st2-packages
-      cd /home/ubuntu/st2/st2-packages
-      git checkout ${CIRCLE_BRANCH} || true
-    - .circle/buildenv.sh
+    - git clone -b ${CIRCLE_BRANCH} --single-branch --depth 1 ${ST2_PACKAGES_REPO} /home/ubuntu/st2/st2-packages
+    - /home/ubuntu/st2/st2-packages/.circle/buildenv.sh
 
 dependencies:
   cache_directories:
--- a/circle.yml
+++ b/circle.yml
@@ -40,7 +40,7 @@ machine:
 checkout:
   post:
     - git clone -b ${CIRCLE_BRANCH} --single-branch --depth 1 ${ST2_PACKAGES_REPO} /home/ubuntu/st2/st2-packages
-    - /home/ubuntu/st2/st2-packages/.circle/buildenv.sh
+    - .circle/buildenv.sh
 
 dependencies:
   cache_directories:


Kami:
Yeah, if we want to go to a different branch in a shallow clone (--depth 1) we need to use the same "hack" as we use in st2contrib.

circle.yml
```diff
@@ -39,11 +39,8 @@ machine:
 
 checkout:
   post:
-    - |
-      git clone --depth 1 ${ST2_PACKAGES_REPO} /home/ubuntu/st2/st2-packages
-      cd /home/ubuntu/st2/st2-packages
-      git checkout ${CIRCLE_BRANCH} || true
-    - .circle/buildenv.sh
+    - git clone -b ${CIRCLE_BRANCH} --single-branch --depth 1 ${ST2_PACKAGES_REPO} /home/ubuntu/st2/st2-packages
+    - /home/ubuntu/st2/st2-packages/.circle/buildenv.sh
```
armab:
```
- .circle/buildenv.sh
```
for consistency (see commands after this line)

circle.yml
```diff
@@ -39,11 +39,8 @@ machine:
 
 checkout:
   post:
-    - |
-      git clone --depth 1 ${ST2_PACKAGES_REPO} /home/ubuntu/st2/st2-packages
-      cd /home/ubuntu/st2/st2-packages
-      git checkout ${CIRCLE_BRANCH} || true
```
armab: As noted before, see the discussion why this order `|| true` hack was needed:
https://github.com/StackStorm/st2/pull/2569

circle.yml
manasdk: Are you sure? Without the `cd` the current working dir is not changed. 

The entire CWD is non-obvious in cicle ci.

circle.yml
manasdk: Actually not having the remote branch is a failure. It should fail loud and clear rather than falling back to master.

So yeah some build might fail due to a race but its easy to get out of the race.

armab: For example when branch `v1.4` in `st2` is created and changes are pushed,
CI here will try to fetch `v1.4` branch of `st2-packages` which is not yet created.

It means that first build for new major release will be always `RED`.


Also interesting coincidence that you was mentioned there: https://github.com/StackStorm/st2/pull/2569#discussion_r55529678 to clarify existing release machinery :)

manasdk: `RED` is better than building against the wrong branch. I can live with that one.

circle.yml
armab: Yes, sure.

The build dir is defined here: 
```yaml
...
general:
  build_dir: st2-packages
```

See: https://circleci.com/docs/configuration#build-dir

manasdk: Thanks!

armab: Also FYI every command in Circle is run in separate subshell, so any vars or `cd` are not preserved across different command, see: https://circleci.com/docs/configuration#phases

manasdk: That explains a bunch of behavior I am seeing.

armab: > The entire CWD is non-obvious in cicle ci.

That's only start :laughing: 
As said you before, it's not like Travis.

Issue #2600, opened 2016-03-24
Rule criteria: comparison with zero is broken


emptywee:
Hi, I was working on the Icinga2 integration pack, and ran into the following issue:

```
2016-03-24 15:33:32,423 INFO [-] Validating rule icinga2.relay_to_chat for event.state_change.
2016-03-24 15:33:32,543 INFO [-] Validation for rule icinga2.relay_to_chat failed on criteria -
  key: trigger.state_type
  pattern: 0.0
  type: gt
  payload: 1.0
2016-03-24 15:33:32,544 INFO [-] 0 rule(s) found to enforce for event.state_change.
```

So when the Icinga2 API returns check results it uses float for numbers. For some reason rule engine doesn't match it correctly. I can work around it by stripping the fraction part in the sensor, but I think the problem is generally with the type conversions made in the stackstorm internally. 

Here's the payload:

```
INFO log [-] Validating rule icinga2.relay_to_chat for event.state_change. (trigger_instance={'occurrence_time': '2016-03-24 15:30:42.995794+00:00', 'trigger': u'icinga2.event.state_change', 'id': '56f408
238e326f157bf447c2', 'payload': {'check_result': {u'check_source': u'dc1udtlhtst02.stack.qadev.corp', u'exit_status': 1.0, u'vars_before': {u'state_type': 0.0, u'attempt': 2.0, u'state': 1.0, u'reachable': True}, u'schedule_start': 14588
33443.956044, u'vars_after': {u'state_type': 1.0, u'attempt': 1.0, u'state': 1.0, u'reachable': True}, u'execution_start': 1458833443.953009, u'schedule_end': 1458833443.956044, u'performance_data': [u'/=2547MB;4735;5327;0;5919', u'/boot
=105MB;154;173;0;193', u'/home=3MB;1560;1755;0;1951', u'/opt=111MB;15268;17177;0;19086', u'/tmp=1565MB;1560;1755;0;1951', u'/var=407MB;3224;3627;0;4031', u'/var/log=1684MB;3122;3512;0;3903'], u'state': 1.0, u'command': [u'/usr/lib64/nagi
[...]

PR #2602, opened 2016-03-25
Fix rule criteria comparison for falsy criteria pattern values

Kami:
This pull request fixes a bug with criteria comparison of falsy pattern values.

Previously, if a falsy value was provided for a pattern (e.g. number 0), comparison would fail since we incorrectly assumed criteria pattern is not provided and cast it to `None`.

This fixes bug described in #2600.


manasdk:
And this is why kids checking for `falsy` values can be a trap :P

manasdk:
+1

PR #2603, opened 2016-03-28
Fix sending announcements through WebUI

emedvedev:
This PR fixes running announcements through WebUI with empty `trace-tag`. WebUI passes an empty dict into `trace_context` instead of `None`, and it can’t match the assertion in code.


Kami:
I'm fine with that, although ideally we would be consistent with None / empty dict everywhere (if possible) :)

PR #2604, opened 2016-03-30
Remove formating for old packs.pack_deploy alias as it was non-functional

jjm:
During continued use of the new deploy pack I noticed that the old formatting was not functional and all the parameters were ending up as a space separated in `{{branch}}`.

So a PR to remove the old formating and remove the unneeded `{{branch}}` from `pack deploy {{repo_name}} {{branch=master}}`.

If there is another way to achieve this, happy to update the PR....

### DIFF
--- a/contrib/packs/aliases/pack_deploy.yaml
+++ b/contrib/packs/aliases/pack_deploy.yaml
@@ -11,8 +11,7 @@ formats:
   - display: "pack deploy <repo_name> [branch <branch=master>]"
     representation:
       - "pack deploy {{repo_name}} branch {{branch=master}}"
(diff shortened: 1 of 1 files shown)


Kami:
LGTM, thanks.

Kami:
LGTM, thanks.

Kami:
On a related note - I will try to spin up some simple testing utils and conventions next week which will make it possibly to easily test aliases.

Kami:
Here is a tracking issue for action alias testing code infrastructure - #2605

It should be simple enough so I should get an initial version out next week and we can start working on some tests then :)

PR #2610, opened 2016-04-07
Initial code infrastructure for action alias unit tests

### LINKED ISSUE #2605: Code infrastructure and conventions for testing action aliases
We should make it easy for pack authors to be able to write tests and test the action aliases.

Basic functionality should be allowing the pack author to select / provide an alias format string and a list of strings with parsed values which should match that format string.

Kami:
This pull request introduces some conventions and structure for writing unit tests for action aliases. To be consistent with the existing action and sensor test approach I decided to go the unit test route so the code doesn't need access to database and other services.

In the future we might also want to introduce some framework for writing pack integration or end to end tests, but that's out of scope here.

You can find an example of an action alias test here - https://github.com/StackStorm/st2/blob/bd78a9aa8e6e4b7b7b934d3e95274bd10cd746fb/contrib/packs/tests/test_action_aliases.py. Feedback on the actual "API" for action alias tests is welcome.

## Regression with format string matching

While working on this I noticed we introduced a regression in the alias execution API with regards to the format string matching when we moved to the regular expressions approach.

Previously we first compared provided user string (command) and made sure it matches any of the format strings. If it didn't we threw an exception. Second pass was actual extraction of parameters from the matched format string.

Right now we directly try to extract parameters from the user input string and we ignore actual format string matching. The problem with that approach is that right now we don't distinguish between "command didn't match any of the format strings" and "command matched the format string, but there are no parameters".

That's problematic and potentially dangerous, especially in scenarios where either an alias or action doesn't take any parameters. In addition to that, a lot of common errors (missing parameters, etc.) will only be caught during run-time as opposed to while trying to match against the format string.

Let's say user specifies `reboot` format string (it takes no parameters) which calls an action which doesn't take any parameters. Since we don't do perform matching against actual format string, any string will match it (`extract_parameters` function will simply return empty dictionary in both scenarios).

I've briefly discussed this with @emedvedev on Slack and I plan to sync up with him again once this PR is merged so we can plan how to fix that.

## TODO

- [x] Finish it up
- [x] Refactor code (move stuff outside of tests.base module into separate modules, etc.).
- [x] Documentation update - https://github.com/StackStorm/st2docs/pull/113
- [x] Alias execution format matching related fixes

## Feature Improvements

- [ ] Support for testing rendered format string (allow user to provide dummy data)

Resolves #2605


contrib/packs/tests/test_action_aliases.py
```diff
@@ -0,0 +1,70 @@
+from st2tests.base import BaseActionAliasTestCase
+
+
+class DeployActionAliasTestCase(BaseActionAliasTestCase):
+    alias_name = 'deploy_pack'
+
+    def test_deploy_alias(self):
+        # Includes packs
+        format_string = self.action_alias_db.formats[0]['representation'][0]
```
Kami: I'm not too big of a fan of way format strings are referenced, but I couldn't come up with a better approach.

An alternative would be specifying format string in the tests directly (instead of referencing it), but that's kinda defeats the purpose since string would need to be duplicated in two places and things would get out of sync...

jjm:
Being able to test aliases is a great improvement for packs and being able to apply TDD methology!

It would be great if we had 2 additional functions:
- **AssertOnlySingleMatch** (?): Which takes _command_ and array of _format_string_'s and failure happens if more than matches.
- **AssertExpectedFormatMatch** (?): Which takes the same as the as the above, but also asserts if the expected entry in the format_string_ does not match.

An alais format being to greedy or an accediental substring of another representation is something I'm finding is hard to catch. Also testing each representation standalone is great to catch parameters are extracted correctly but will not catch an overlap or where an incorrect representation is selected.

Kami:
@jjm Thanks for the feedback.

I agree with your feedback and plan to add some more of that functionality once I address the format string match issue mentioned in the PR description. I planned to do the second thing you proposed at first, but then I noticed there is this bug (we don't throw if the format doesn't match) so I need to fix this first :)

Kami:
@jjm and others - I added two more assertion methods in 65804951f641f512f2d8571af2112f860907f0d3 (commit also includes some usage examples):

* `assertCommandMatchesFormatString`
* `assertCommandMatchesSingleFormatString`

Let me know what you think - if the API doesn't look or feel right we can always change it :)

In addition to that, I also included all the alias execution related fixes in this pull request (I tried to logically group them in separate commits) since doing it in a separate branch would be quite painful since changes are interdependent.

contrib/packs/tests/test_action_aliases.py
```diff
@@ -0,0 +1,106 @@
...
+        # Command with branch
+        command = 'pack deploy StackStorm/st2contrib packs libcloud,aws branch master'
+        expected_parameters = {
+            'repo_name': 'StackStorm/st2contrib',
+            'packs': 'libcloud,aws',
+            'branch': 'master'
+        }
+        self.assertExtractedParametersMatch(format_string=format_string,
+                                            command=command,
+                                            values=expected_parameters)
```
lakshmi-kannan: Instead of values=, would it be better to call it args=?

Kami: Yeah, I'm not too happy with the parameter name and I'm open to suggestions.

I was also thinking about maybe calling it `parameters`, `expected_parameters` or similar.

lakshmi-kannan: Yeah, I also thought of parameters. I can live with that. @manasdk ?

st2tests/st2tests/action_aliases.py
```diff
@@ -0,0 +1,125 @@
...
+        Assert that the provided command matches the provided format string.
+        """
+        try:
+            extract_parameters_for_action_alias_db(action_alias_db=self.action_alias_db,
+                                                   format_str=format_string,
+                                                   param_stream=command)
+        except ParseException as e:
+            raise AssertionError(str(e))
+
+    def assertCommandMatchesSingleFormatString(self, format_strings, command):
```
lakshmi-kannan: Sorry, this test https://github.com/StackStorm/st2/pull/2610/files#diff-b2513a4fce3afc879336c085ac31062fR4 just shows that pack authors are mostly expected to write assertExtractedParametersMatch tests. So why have the other ones?

Kami: I added those tests before adding those new assertion methods (I need to fix some alias execution related stuff before I could add those assertion methods) and I haven't updated those tests yet.

Having said that, yes, there is some overlap between the methods (`assertCommandMatchesFormatString` is a subset of `assertExtractedParametersMatch`). I would personally be fine just with `assertExtractedParametersMatch`, but if others believe (cc @jjm) having those two additional methods provides value we can leave them.

jjm: Thanks @Kami, 

I think I could live just the addition of `assertCommandMatchesSingleFormatString` if it asserted if there was no matches. Which is really what `assertCommandMatchesFormatString` is attempting to cover.

This is just to allow pack authors to test a set of aliases as whole instead of individually, thus ensuring we don't have an overlap (or incorrect ordering) of different representations.

Kami: Yeah, I can make that change (to also throw assertion error if there were no matches).

So maybe a better name would also be `assertCommandMatchesExactlyOneFormatString` or similar (I'm open to suggestions).

jjm: Going with `assertCommandMatchesExactlyOneFormatString` sounds nice and descriptive to me.

Kami: I've made those changes in 88d5d0ca780618d9ddede27086716579a6adfc6e.

contrib/packs/tests/test_action_aliases.py
Kami: Renamed from `values` to `parameters` in f0fb10168b4b4e7b54946fb3804d9cef5e18141b.

tonybaloney:
:+1:

PR #2612, opened 2016-04-08
Tools that publishes messages directly to a Q

manasdk:
* Exisiting tool queue_producer published to an exchange. This is
  for direct publishing to a queue.
* Written on top of pika instead of kombu as this was simpler to write
  given code samples on rabbitmq tutorial site

### DIFF
--- a/tools/direct_queue_publisher.py
+++ b/tools/direct_queue_publisher.py
@@ -1,0 +1,44 @@
+#!/usr/bin/env python
+
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


tools/direct_queue_publisher.py
```diff
@@ -0,0 +1,44 @@
...
+
+import pika
+
+def main(queue, payload):
+	connection = pika.BlockingConnection(pika.ConnectionParameters(
+		host='localhost',
+		credentials=pika.credentials.PlainCredentials(username='guest', password='guest')))
+	channel = connection.channel()
+
+	channel.queue_declare(queue=queue, durable=True)
```
Kami: Maybe better to declare a non-durable queue, otherwise it will stick around if we don't explicitly delete it when the tool exits, right?

manasdk: umm, yeah its a bit of a mess because I used this to work with rabbitmqqueuesensor which for some reason defines it durable. 

I will make durable an input argument.

tools/direct_queue_publisher.py
```diff
@@ -0,0 +1,44 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+
+import pika
```
lakshmi-kannan: ummm, what?

lakshmi-kannan: Saw description. Panic averted.

PR #2620, opened 2016-04-14
Include stderr in the error message in virtualenvs.py when the pip command fails.

jsjeannotte:
Include stderr in the error message in virtualenvs.py when the pip command fails.


Kami:
Thanks!

The change looks good to me and the tests pass so I will go ahead and merge it into master.

PR #2622, opened 2016-04-16
SSL support for mongodb connections

manasdk:
* Expose all parameters for SSL/TLS support with mongodb

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,8 @@ Changelog
 in development
 --------------
 
+* SSL support for mongodb connections. (improvement)
+
 
 1.4.0 - April 18, 2016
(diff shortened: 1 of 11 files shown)


manasdk:
To test this PR I am having to build mongod. This might be one of the worst rabbit holes I have ventured into!

manasdk:
Validated with mongodb that had ssl enabled. Tested only the ssl=True path for now.

manasdk:
Docs - https://github.com/StackStorm/st2docs/pull/122

Testing - Ideally we want to do some sort of infra testing however that would mean maintaining our own mongodb distribution with SSL support. I am going to punt on that until we decide that SSL/TLS si default for all connections.

PR #2625, opened 2016-04-18
Make st2ctl ignore services that aren't installed

emedvedev:
This is my proposed fix to the issue outlined in #2550: st2ctl will give an “unknown service” error if one of the packages (chatops in that case) wasn’t installed.

The script makes a `status` call before start/stop, and if the exit code is 1, the service is not shown in the list. I checked on RHEL6 and Ubuntu that `1` is only given when the service doesn’t exist, otherwise it gives other codes.

Can be tested by removing the st2chatops package or stopping the service and then trying to stop it again.

Includes st2chatops, so trumps #2550.

### DIFF
--- a/st2common/bin/st2ctl
+++ b/st2common/bin/st2ctl
@@ -71,10 +71,10 @@ function st2start() {
 
 function st2stop() {
   for COM in ${COMPONENTS}; do
-      service_manager $COM status >/dev/null 2>&1;
-      if [[ "1" == $? ]]; then
-        continue
-      fi
+    service_manager $COM status >/dev/null 2>&1;
+    if [[ "1" == $? ]]; then
+      continue
+    fi
     service_manager ${COM} stop
   done
 }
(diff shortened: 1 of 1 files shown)


st2common/bin/st2ctl
```diff
@@ -61,12 +61,20 @@ function validate_in_components() {
 
 function st2start() {
   for COM in ${COMPONENTS}; do
+    service_manager $COM status >/dev/null 2>&1;
+    if [[ "1" == $? ]]; then
```
lakshmi-kannan: not equal to zero would be better.

emedvedev: It's not exactly the same. `status` can give exit codes that are not 0 or 1; i.e. upstart gives some other code when you're trying to stop a service that's already stopped (or start a service that's already running). We probably don't want to hide those, so I only made it skip on 1 which is given when service is not found.

DoriftoShoes: incorrect service will always throw 1.

DoriftoShoes:
LGTM

manasdk:
Well this didnt make it to v1.4.

emedvedev:
Yeah, we probably should.

PR #2631, opened 2016-04-19
Secrets in kv store phase #0

lakshmi-kannan:
## What? 

This PR allows encrypted items to be stored in key value store. The encryption is symmetric AES256 encryption and we use keyczar to do this. The encryption key is global for all keys. So StackStorm admin can decrypt all the keys. However, with RBAC, we'll make sure only intended user (and admin) can decrypt the user's keys. 

## TODO
* [x] Encrypted values are non UTF-8. Mongoengine complains about that. 
* [x] CLI changes
  * [X] st2 keyvalue get changes 
  * [x] st2 keyvalue list changes
* [x] Tests
* [X] Tool for generating AES key in tools/ 249340e
* [X] Docs https://github.com/StackStorm/st2docs/pull/125

### PUT
```
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (secrets_kv_store●●)$ st2 key set foo bar                                                                                                                                                       [ruby-1.9.3p484]
+------------------+-------+
| Property         | Value |
+------------------+-------+
| name             | foo   |
| value            | bar   |
| expire_timestamp |       |
+------------------+-------+
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (secrets_kv_store●●)$ st2 key set secret_foo secret_bar --encrypt                                                                                                                               [ruby-1.9.3p484]
+------------------+--------------------------------------------------------------+
| Property         | Value                                                        |
+------------------+--------------------------------------------------------------+
| name             | secret_foo                                                   |
| value            | 0083E1033A7C286818517F1DDD2190042C6F6161A4A6318CA7A7767B97E8 |
|                  | DCD2984F36B070A60BA3BFE0ACF0C583F6DE8233B7C842E2EAD2D6       |
| expire_timestamp |                                                              |
+------------------+--------------------------------------------------------------+
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (secrets_kv_store●●)$
```

### GET
```
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (secrets_kv_store●●)$ st2 key get secret_foo --decrypt                                                                                                                                          [ruby-1.9.3p484]
+------------------+------------+
| Property         | Value      |
+------------------+------------+
| name             | secret_foo |
| value            | secret_bar |
| secret           | True       |
| encrypted        | False      |
| expire_timestamp |            |
+------------------+------------+
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (secrets_kv_store●●)$ st2 key get secret_foo                                                                                                                                                    [ruby-1.9.3p484]
+------------------+--------------------------------------------------------------+
| Property         | Value                                                        |
+------------------+--------------------------------------------------------------+
| name             | secret_foo                                                   |
| value            | 0083E1033A90EB483B7368A414E9456928FDAB3808830B553CA63691E045 |
|                  | 3990D8B4BEB0263B3F9363F1D19CF06BF70EC9B2D81F7F7F0F6199       |
| secret           | True                                                         |
| encrypted        | True                                                         |
| expire_timestamp |                                                              |
+------------------+--------------------------------------------------------------+
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (secrets_kv_store●●)$ st2 key get foo                                                                                                                                                           [ruby-1.9.3p484]
+------------------+-------+
| Property         | Value |
+------------------+-------+
| name             | foo   |
| value            | bar   |
| secret           | False |
| encrypted        | False |
| expire_timestamp |       |
+------------------+-------+
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (secrets_kv_store●●)$
```

### List
```
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (secrets_kv_store●●)$ st2 key list --decrypt                                                                                                                                                   [ruby-1.9.3p484]

+------------+------------+--------+-----------+------------------+
| name       | value      | secret | encrypted | expire_timestamp |
+------------+------------+--------+-----------+------------------+
| secret_foo | secret_bar | True   | False     |                  |
| foo        | bar        | False  | False     |                  |
+------------+------------+--------+-----------+------------------+
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (secrets_kv_store●●)$                                                                                                                                                                          [ruby-1.9.3p484]
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (secrets_kv_store●●)$ st2 key list                                                                                                                                                             [ruby-1.9.3p484]
+------------+----------------------------------------------------+--------+-----------+------------------+
| name       | value                                              | secret | encrypted | expire_timestamp |
+------------+----------------------------------------------------+--------+-----------+------------------+
| secret_foo | 0083E1033A040EA17E5D5A61E1087675EEFED8C7A1BFF0927F | True   | True      |                  |
|            | EDEF0BE9C948044A87BE35DE9238FC67E707D377999C3FEA43 |        |           |                  |
|            | 17B4EF32371572                                     |        |           |                  |
| foo        | bar                                                | False  | False     |                  |
+------------+----------------------------------------------------+--------+-----------+------------------+
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (secrets_kv_store●●)$
```
### No secret case

```
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (secrets_kv_store●●)$ http PUT http://127.0.0.1:9101/v1/keys/foo < kv_nosecret.json                                                                                                             [ruby-1.9.3p484]
HTTP/1.1 200 OK
Access-Control-Allow-Origin: http://172.168.50.50:3000
Connection: keep-alive
Content-Length: 97
Date: Thu, 21 Apr 2016 20:12:54 GMT
Server: gunicorn/19.4.5
X-Request-ID: 6244e7c9-bdd3-4040-9da5-2941e84391eb

{
    "encrypted": false,
    "name": "foo",
    "secret": false,
    "uid": "key_value_pair:foo",
    "value": "bar"
}

(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (secrets_kv_store●●)$
```

### Secret case. No encryption key setup by admin

XXX: Note I return 400 BAD REQUEST. Opinions welcome. 

```
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (secrets_kv_store●)$ http PUT http://127.0.0.1:9101/v1/keys/secret_foo < kv.json                                                                                                               [ruby-1.9.3p484]
HTTP/1.1 400 Bad Request
Access-Control-Allow-Origin: http://172.168.50.50:3000
Connection: keep-alive
Content-Length: 131
Content-Type: application/json
Date: Tue, 19 Apr 2016 23:43:26 GMT
Server: gunicorn/19.4.5
X-Request-ID: 14ae1cf2-f6be-472b-b1e4-8fa46b4d6148

{
    "faultstring": "Crypto key not found in conf/st2_kvstore_demo.crypto.key.json. Unable to encrypt value for key secret_foo."
}

(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (secrets_kv_store●)$
```

### Secret case. Encryption key setup. 

```
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (secrets_kv_store●●)$ http PUT http://127.0.0.1:9101/v1/keys/secret_foo < kv.json                                                                                                               [ruby-1.9.3p484]
HTTP/1.1 200 OK
Access-Control-Allow-Origin: http://172.168.50.50:3000
Connection: keep-alive
Content-Length: 220
Date: Thu, 21 Apr 2016 20:13:26 GMT
Server: gunicorn/19.4.5
X-Request-ID: 41cc580c-7557-4e1b-9e32-df4672da7a63

{
    "encrypted": true,
    "name": "secret_foo",
    "secret": true,
    "uid": "key_value_pair:secret_foo",
    "value": "0083E1033A66FF9293F6715BCF5B3918E3A18E2281B6A45B08603032571E3812C6D9CEE25BD1E060DADD8E951C85E4479C1C1BE91835BACEF1"
}

(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (secrets_kv_store●●)$

(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (secrets_kv_store●●)$ http GET "http://127.0.0.1:9101/v1/keys/secret_foo?decrypt=False"                                                                                                         [ruby-1.9.3p484]
HTTP/1.1 200 OK
Access-Control-Allow-Origin: http://172.168.50.50:3000
Connection: keep-alive
Content-Length: 220
Date: Thu, 21 Apr 2016 20:13:58 GMT
Server: gunicorn/19.4.5
X-Request-ID: 8ee5d905-5878-47ca-98d1-118ccddc5a3d

{
    "encrypted": true,
    "name": "secret_foo",
    "secret": true,
    "uid": "key_value_pair:secret_foo",
    "value": "0083E1033A66FF9293F6715BCF5B3918E3A18E2281B6A45B08603032571E3812C6D9CEE25BD1E060DADD8E951C85E4479C1C1BE91835BACEF1"
}

(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (secrets_kv_store●●)$ http GET "http://127.0.0.1:9101/v1/keys/secret_foo?decrypt=True"                                                                                                          [ruby-1.9.3p484]
HTTP/1.1 200 OK
Access-Control-Allow-Origin: http://172.168.50.50:3000
Connection: keep-alive
Content-Length: 117
Date: Thu, 21 Apr 2016 20:14:03 GMT
Server: gunicorn/19.4.5
X-Request-ID: 3a5ad5a8-bb68-472b-854f-3d3764b22b28

{
    "encrypted": false,
    "name": "secret_foo",
    "secret": true,
    "uid": "key_value_pair:secret_foo",
    "value": "secret_bar"
}

(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (secrets_kv_store●●)$
```

```
> db.key_value_pair_d_b.find({'name': 'secret_foo'})
{ "_id" : ObjectId("5717e445d9d7ed0ea5b07ec7"), "uid" : "key_value_pair:secret_foo", "name" : "secret_foo", "value" : "0083E1033AE5F3D1B045398FC51F1046927971A46502A5A5BEB72ED37FB32073709569E37D081D9D38BF3E96F8A9F4D1EA5FCE349D6D18B3EB", "encrypted" : true }
>
```

### DIFF
--- a/conf/st2_kvstore_demo.crypto.key.json
+++ b/conf/st2_kvstore_demo.crypto.key.json
@@ -1,0 +1,1 @@
+{"hmacKey": {"hmacKeyString": "685EWJ8U2U_c3Ulv9ibB6aa0Y9aYm-0bCwe2mqjLJuw", "size": 256}, "aesKeyString": "eKGFrfNEUhAgX0uofaU1wQ", "mode": "CBC", "size": 128}
--- a/st2api/st2api/controllers/v1/keyvalue.py
+++ b/st2api/st2api/controllers/v1/keyvalue.py
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from oslo_config import cfg
 from pecan import abort
 from pecan.rest import RestController
 import six
@@ -47,19 +46,20 @@ class KeyValuePairController(RestController):
         self.get_one_db_method = self.__get_by_name
 
     @jsexpose(arg_types=[str, str])
-    def get_one(self, name, decrypt):
+    def get_one(self, name, decrypt='false'):
         """
             List key by name.
 
             Handle:
                 GET /keys/key1
         """
-        kvp_db = self.__get_by_name(name=name)
 
         if not decrypt:
             decrypt = False
         else:
             decrypt = (decrypt == 'true' or decrypt == 'True' or decrypt == '1')
+
+        kvp_db = self.__get_by_name(name=name)
 
         if not kvp_db:
             LOG.exception('Database lookup for name="%s" resulted in exception.', name)
--- a/st2api/st2api/controllers/v1/keyvalue.py
+++ b/st2api/st2api/controllers/v1/keyvalue.py
@@ -85,12 +85,21 @@ class KeyValuePairController(RestController):
         # Prefix filtering
         prefix_filter = kw.get('prefix', None)
 
+        decrypt = kw.get('decrypt', None)
+        if not decrypt:
+            decrypt = False
+        else:
+            decrypt = (decrypt == 'true' or decrypt == 'True' or decrypt == '1')
+            del kw['decrypt']
+
         if prefix_filter:
             kw['name__startswith'] = prefix_filter
             del kw['prefix']
 
         kvp_dbs = KeyValuePair.get_all(**kw)
-        kvps = [KeyValuePairAPI.from_model(kvp_db) for kvp_db in kvp_dbs]
+        kvps = [KeyValuePairAPI.from_model(
+            kvp_db, mask_secrets=(not decrypt)) for kvp_db in kvp_dbs
+        ]
 
         return kvps
 
--- a/st2api/st2api/controllers/v1/keyvalue.py
+++ b/st2api/st2api/controllers/v1/keyvalue.py
@@ -13,12 +13,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from oslo_config import cfg
 from pecan import abort
 from pecan.rest import RestController
 import six
 from mongoengine import ValidationError
 
 from st2common import log as logging
+from st2common.exceptions.keyvalue import CryptoKeyNotSetupException
 from st2common.models.api.keyvalue import KeyValuePairAPI
 from st2common.models.api.base import jsexpose
 from st2common.persistence.keyvalue import KeyValuePair
@@ -113,7 +115,10 @@ class KeyValuePairController(RestController):
                 LOG.exception('Validation failed for key value data=%s', kvp)
                 abort(http_client.BAD_REQUEST, str(e))
                 return
-
+            except CryptoKeyNotSetupException as e:
+                LOG.exception(str(e))
+                abort(http_client.BAD_REQUEST, str(e))
+                return
         extra = {'kvp_db': kvp_db}
         LOG.audit('KeyValuePair updated. KeyValuePair.id=%s' % (kvp_db.id), extra=extra)
 
--- a/st2api/st2api/controllers/v1/keyvalue.py
+++ b/st2api/st2api/controllers/v1/keyvalue.py
@@ -46,8 +46,8 @@ class KeyValuePairController(RestController):
         self._coordinator = coordination.get_coordinator()
         self.get_one_db_method = self.__get_by_name
 
-    @jsexpose(arg_types=[str])
-    def get_one(self, name):
+    @jsexpose(arg_types=[str, str])
+    def get_one(self, name, decrypt):
         """
             List key by name.
 
@@ -56,13 +56,18 @@ class KeyValuePairController(RestController):
         """
         kvp_db = self.__get_by_name(name=name)
 
+        if not decrypt:
+            decrypt = False
+        else:
+            decrypt = (decrypt == 'true' or decrypt == 'True' or decrypt == '1')
+
         if not kvp_db:
             LOG.exception('Database lookup for name="%s" resulted in exception.', name)
             abort(http_client.NOT_FOUND)
             return
 
         try:
-            kvp_api = KeyValuePairAPI.from_model(kvp_db)
+            kvp_api = KeyValuePairAPI.from_model(kvp_db, mask_secrets=(not decrypt))
         except (ValidationError, ValueError) as e:
             abort(http_client.INTERNAL_SERVER_ERROR, str(e))
             return
--- a/st2client/st2client/models/core.py
+++ b/st2client/st2client/models/core.py
@@ -169,7 +169,7 @@ class ResourceManager(object):
         prefix = kwargs.pop('prefix', None)
         user = kwargs.pop('user', None)
 
-        params = {}
+        params = kwargs.pop('params', {})
         if limit and limit <= 0:
             limit = None
         if limit:
--- a/st2client/st2client/models/core.py
+++ b/st2client/st2client/models/core.py
@@ -240,7 +240,8 @@ class ResourceManager(object):
         if 'limit' in kwargs and kwargs.get('limit') <= 0:
             kwargs.pop('limit')
         token = kwargs.get('token', None)
-        params = {}
+        params = kwargs.get('params', {})
+        print(params)
         for k, v in six.iteritems(kwargs):
             if k != 'token':
                 params[k] = v
--- a/st2client/tests/unit/test_commands.py
+++ b/st2client/tests/unit/test_commands.py
@@ -73,22 +73,8 @@ class TestResourceCommand(unittest2.TestCase):
         self.assertEqual(actual, expected)
 
     @mock.patch.object(
-        models.ResourceManager, 'get_by_name',
-        mock.MagicMock(return_value=base.FakeResource(**base.RESOURCES[0])))
-    @mock.patch.object(
-        models.ResourceManager, 'get_by_id',
-        mock.MagicMock(return_value=None))
-    def test_command_get_by_name(self):
-        args = self.parser.parse_args(['fakeresource', 'get', 'abc'])
-        self.assertEqual(args.func, self.branch.commands['get'].run_and_print)
-        instance = self.branch.commands['get'].run(args)
-        actual = instance.serialize()
-        expected = json.loads(json.dumps(base.RESOURCES[0]))
-        self.assertEqual(actual, expected)
-
-    @mock.patch.object(
         httpclient.HTTPClient, 'get',
-        mock.MagicMock(return_value=base.FakeResponse(json.dumps([base.RESOURCES[0]]), 200, 'OK')))
+        mock.MagicMock(return_value=base.FakeResponse(json.dumps(base.RESOURCES[0]), 200, 'OK')))
     def test_command_get(self):
         args = self.parser.parse_args(['fakeresource', 'get', 'abc'])
         self.assertEqual(args.func, self.branch.commands['get'].run_and_print)
--- a/st2common/st2common/config.py
+++ b/st2common/st2common/config.py
@@ -139,6 +139,17 @@ def register_opts(ignore_errors=False):
     ]
     do_register_opts(api_opts, 'api', ignore_errors)
 
+    # Key Value store options
+    keyvalue_opts = [
+        cfg.BoolOpt('enable_encryption', default=True,
+                    help='Allow encryption of values in key value stored qualified as "secret".'),
+        cfg.StrOpt('encryption_key_path', default='conf/st2_kvstore_demo.crypto.key.json',
+                   help='Location of the symmetric encryption key for encrypting values in ' +
+                        'kvstore. This key should be in JSON and should\'ve been ' +
+                        'generated using keyczar.')
+    ]
+    do_register_opts(keyvalue_opts, group='keyvalue')
+
     # Common auth options
     auth_opts = [
         cfg.StrOpt('api_url', default=None,
--- a/st2common/st2common/config.py
+++ b/st2common/st2common/config.py
@@ -143,7 +143,7 @@ def register_opts(ignore_errors=False):
     keyvalue_opts = [
         cfg.BoolOpt('enable_encryption', default=True,
                     help='Allow encryption of values in key value stored qualified as "secret".'),
-        cfg.StrOpt('encryption_key_path', default='conf/st2_kvstore_demo.crypto.key.json',
+        cfg.StrOpt('encryption_key_path', default='',
                    help='Location of the symmetric encryption key for encrypting values in ' +
                         'kvstore. This key should be in JSON and should\'ve been ' +
                         'generated using keyczar.')
--- a/st2common/st2common/models/api/keyvalue.py
+++ b/st2common/st2common/models/api/keyvalue.py
@@ -100,7 +100,7 @@ class KeyValuePairAPI(BaseAPI):
             return key
 
     @classmethod
-    def from_model(cls, model, mask_secrets=False):
+    def from_model(cls, model, mask_secrets=True):
         if not KeyValuePairAPI.crypto_setup:
             KeyValuePairAPI._setup_crypto()
 
--- a/st2common/st2common/models/api/keyvalue.py
+++ b/st2common/st2common/models/api/keyvalue.py
@@ -110,7 +110,7 @@ class KeyValuePairAPI(BaseAPI):
             del doc['id']
 
         if not mask_secrets and model.encrypted:
-            doc[value] = symmetric_decrypt(KeyValuePairAPI.crypto_key, value)
+            doc['value'] = symmetric_decrypt(KeyValuePairAPI.crypto_key, model.value)
 
         if model.expire_timestamp:
             doc['expire_timestamp'] = isotime.format(model.expire_timestamp, offset=False)
--- a/st2common/st2common/models/api/keyvalue.py
+++ b/st2common/st2common/models/api/keyvalue.py
@@ -14,7 +14,6 @@
 # limitations under the License.
 
 import datetime
-import json
 import os
 
 from keyczar.keys import AesKey
@@ -30,6 +29,7 @@ from st2common.models.api.base import BaseAPI
 from st2common.models.db.keyvalue import KeyValuePairDB
 
 LOG = logging.getLogger(__name__)
+
 
 class KeyValuePairAPI(BaseAPI):
     crypto_setup = False
@@ -58,6 +58,11 @@ class KeyValuePairAPI(BaseAPI):
                 'required': False,
                 'default': False
             },
+            'encrypted': {
+                'type': 'boolean',
+                'required': False,
+                'default': False
+            },
             'expire_timestamp': {
                 'type': 'string',
                 'pattern': isotime.ISO8601_UTC_REGEX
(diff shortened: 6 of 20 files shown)


manasdk:
> Secret case. No encryption key setup by admin

I am ok with 400. I guess we are saying for the current configuration state this request is bad.

st2common/st2common/models/api/keyvalue.py
```diff
@@ -57,8 +76,39 @@ class KeyValuePairAPI(BaseAPI):
         'additionalProperties': False
     }
 
+    @staticmethod
+    def _setup_crypto():
```
manasdk: I would imagine encryption and decryption for kvp being handled in the persistence or the db modules. This way its not an API only feature and it would be possible to get decrypted keys even if not using the API i.e. from services do not use the API layer.

lakshmi-kannan: Yeah, I can move the code around when we address that.

st2common/st2common/models/api/keyvalue.py
```diff
@@ -44,6 +53,16 @@ class KeyValuePairAPI(BaseAPI):
                 'type': 'string',
                 'required': True
             },
+            'secret': {
```
manasdk: You have 2 properties secret and encrypted so I have a few questions -

1. If a key is a secret it is always encrypted?
2. If a key is encrypted does it have to be a secret?
3. Do we attribute any special semantic meaning to secret? i.e. because a key is a secret it will be handled differently?

manasdk:
`st2 key list --decrypt`

Would it be safer to not support a list level decrypt? I wonder if there is value in providing this in the first place.

manasdk:
Overall +1 to design and impl.

Once you add some tests and docs around this good to go :)

st2common/st2common/models/api/keyvalue.py
lakshmi-kannan: 1. Yes
2. Yes for now. I am using encrypted as a display flag to intimate CLI or UI that the value sent out is encrypted. This is more for cosmetic purposes at this point. If I don't do that, st2 keyvalue get secret_key will display a value that is not immediately obvious that is encrypted. So `encrypted` field is more for user UX. This is the thing I wanted feedback on. 
3. Yes, that flag is how we know a key user wants to store needs to be encrypted. If you look at PUT APIs for secret vs non-secret, it might make sense.

lakshmi-kannan:
>Would it be safer to not support a list level decrypt?

Well, with RBAC this might get even harder but it is something we need to solve I think. I was debating it myself too. Let's wait for another input to break the tie.

st2common/st2common/util/crypto.py
```diff
@@ -17,8 +17,38 @@
...
+    Encrypt the given message using the encrypt_key. Returns a UTF-8 str
+    ready to be stored in database. Note that we convert the hex notation
+    to a ASCII notation to produce a UTF-8 friendly string.
+
+    Also, this method will not return the same output on multiple invocations
+    of same method. The reason is that the Encrypt method uses a different
+    'Initialization Vector' per run and the IV is part of the output.
+
+    :param encrypt_key: Symmetric AES key to use for encryption.
+    :type encrypt_key: :class:`keyczar.keys.AesKey`
```
manasdk: for this method does it matter if its an AesKey? Is there is higher interface/superclass?

st2common/st2common/util/crypto.py
```diff
@@ -17,8 +17,38 @@
...
+
+    :param message: Message to be encrypted.
+    :type message: ``str``
+
+    :rtype: ``str``
+    """
     return binascii.hexlify(encrypt_key.Encrypt(message)).upper()
```
manasdk: curious upper is for consistency sake of there is a specific reason?

st2common/st2common/util/crypto.py
lakshmi-kannan: Yep. Unfortunately it does. The Encrypt method is on the key object which is keyczar and AES specific implementation.

st2common/st2common/util/crypto.py
lakshmi-kannan: It actually doesn't make a difference when unhexlify. I am just following the guide on hexlify.

Kami:
> Tool for generating AES key in tools/

Keyczar already provides a tool for managing and rotating keys. Do you plan to build a wrapper around that or something else?

conf/st2_kvstore_demo.crypto.key.json
```diff
@@ -0,0 +1 @@
+{"hmacKey": {"hmacKeyString": "685EWJ8U2U_c3Ulv9ibB6aa0Y9aYm-0bCwe2mqjLJuw", "size": 256}, "aesKeyString": "eKGFrfNEUhAgX0uofaU1wQ", "mode": "CBC", "size": 128}
```
Kami: For fixtures that's a fine path, but for actual key on production, we should advise users to put in a path with limited access (e.g. /etc/private or similar).

st2api/st2api/controllers/v1/keyvalue.py
```diff
@@ -44,14 +45,20 @@ def __init__(self):
         self._coordinator = coordination.get_coordinator()
         self.get_one_db_method = self.__get_by_name
 
-    @jsexpose(arg_types=[str])
-    def get_one(self, name):
+    @jsexpose(arg_types=[str, str])
+    def get_one(self, name, decrypt='false'):
```
Kami: Can we hook this up and make bool work so we don't need to abuse string type?

I thought we already do this somewhere, but maybe we don't...

st2common/st2common/config.py
```diff
@@ -139,6 +139,17 @@ def register_opts(ignore_errors=False):
     ]
     do_register_opts(api_opts, 'api', ignore_errors)
 
+    # Key Value store options
+    keyvalue_opts = [
+        cfg.BoolOpt('enable_encryption', default=True,
+                    help='Allow encryption of values in key value stored qualified as "secret".'),
+        cfg.StrOpt('encryption_key_path', default='conf/st2_kvstore_demo.crypto.key.json',
```
Kami: Default path should probably point to some safer location and we should override this setting in the tests.

st2common/st2common/models/api/keyvalue.py
```diff
@@ -57,8 +76,39 @@ class KeyValuePairAPI(BaseAPI):
...
+                     KeyValuePairAPI.crypto_key_path)
+            if not os.path.exists(KeyValuePairAPI.crypto_key_path):
+                msg = ('Encryption key file does not exist in path %s.' %
+                       KeyValuePairAPI.crypto_key_path)
+                LOG.exception(msg)
+                LOG.info('All API requests will now send out BAD_REQUEST ' +
+                         'if you ask to store secrets in key value store.')
+                KeyValuePairAPI.crypto_key = None
+            else:
+                KeyValuePairAPI.crypto_key = KeyValuePairAPI._read_crypto_key(
```
Kami: Unless I'm missing something it seems like this is called only once.

If that's the case, since the whole API has access to this part I think it would be safer to only load the key when needed. I know it will be slower than caching the key, but it means the key won't stuck in the memory when we don't need it (well, that's not 100% true since we don't do manual garbage collection, etc., but still).

Kami:
I'm not totally against this approach, but as mentioned on Slack, I would prefer to use public key cryptography.

This will make it easier for us to eventually move to a safer model, where we have a single service (or similar) which is responsible for decrypting data (this would be the only service which has access to the private key, reducing the surface area), but other services (e.g. API) will still be able to encrypt data using the public key...

conf/st2_kvstore_demo.crypto.key.json
lakshmi-kannan: Yep, will say that in docs.

st2api/st2api/controllers/v1/keyvalue.py
lakshmi-kannan: >I thought we already do this somewhere, but maybe we don't...

I checked we don't. 

>I don't think it should be bad, just a couple of lines of code in jsexpose..

That was my initial approach. I couldn't get it to work. I was complaining about it on slack. I can try again once PR is merged.

st2common/st2common/config.py
lakshmi-kannan: Good point. Let me think about this.

st2common/st2common/models/api/keyvalue.py
lakshmi-kannan: Yeah, I went both ways. Initially, I was loading it on every call to encrypt/decrypt but that definitely seems excessive. I do agree with your assessment of not holding onto the key in memory but then if we load it every time, people could swap out the key and not realize they changed it. If that's beyond us, I can change this to load on every encrypt/decrypt call.

lakshmi-kannan:
>I'm not totally against this approach, but as mentioned on Slack, I would prefer to use public key cryptography.

We can swap it out if needed. I don't see this as a big problem now that we have API & CLI changes.

st2client/tests/unit/test_commands.py
```diff
@@ -73,22 +73,8 @@ def test_command_get_by_id(self):
         self.assertEqual(actual, expected)
 
     @mock.patch.object(
-        models.ResourceManager, 'get_by_name',
-        mock.MagicMock(return_value=base.FakeResource(**base.RESOURCES[0])))
-    @mock.patch.object(
-        models.ResourceManager, 'get_by_id',
-        mock.MagicMock(return_value=None))
-    def test_command_get_by_name(self):
```
lakshmi-kannan: I removed this bogus test. The underlying code was anyways calling get_by_id and we are already testing that in previous test case.

lakshmi-kannan:
> Keyczar already provides a tool for managing and rotating keys. Do you plan to build a wrapper around that or something else? 

Correct - a wrapper. Some argparse + keyczar + fwrite()

st2common/st2common/config.py
manasdk: Why even have a default value? If not specified assume it does not exist and ignore entirely.

st2common/st2common/models/api/keyvalue.py
manasdk: Does `encrypted` only mean that whether the current display value or value returned from the API is encrypted? It as such provides no guarantee that it is stored in an encrypted manner in the DB since that is a function of `secret`.

dzimine:
👍  for design, thanks! didn't look into impl. yet.
> 
`st2 key list --decrypt`
>
Would it be safer to not support a list level decrypt? I wonder if there is value in providing this in the first place.
>

The use case is, for instance, migration by admin to another node. As admin, I dump all KV with `st2 key list --decrypt -j` and recreate it on the other one. Here having "encrypted" value helps too, to hint to decrypt it when json is loaded.

manasdk:
>  As admin, I dump all KV with `st2 key list --decrypt -j` and recreate it on the other one.

Would migrating the encrypted data as well as the key work just as well?

Kami:
Re saving and serializing encrypted data - I can't recall if Keyczar already does that, but if doesn't we should update serialized data so it includes header with the encryption information (which algorithm was used, etc.).

This will makes this such as migrating to a different algorithm or similar easier.

st2client/st2client/models/core.py
```diff
@@ -240,7 +240,8 @@ def query(self, **kwargs):
         if 'limit' in kwargs and kwargs.get('limit') <= 0:
             kwargs.pop('limit')
         token = kwargs.get('token', None)
-        params = {}
+        params = kwargs.get('params', {})
+        print(params)
```
Kami: Probably need to remove print here?

st2client/st2client/models/core.py
```diff
@@ -240,7 +240,8 @@ def query(self, **kwargs):
         if 'limit' in kwargs and kwargs.get('limit') <= 0:
             kwargs.pop('limit')
         token = kwargs.get('token', None)
-        params = {}
+        params = kwargs.get('params', {})
```
Kami: Sigh, yeah, wish we could just port everything to explicitly use `params` argument instead of abusing kwargs...

st2common/st2common/config.py
```diff
@@ -139,6 +139,17 @@ def register_opts(ignore_errors=False):
...
+    # Key Value store options
+    keyvalue_opts = [
+        cfg.BoolOpt('enable_encryption', default=True,
+                    help='Allow encryption of values in key value stored qualified as "secret".'),
+        cfg.StrOpt('encryption_key_path', default='',
+                   help='Location of the symmetric encryption key for encrypting values in ' +
+                        'kvstore. This key should be in JSON and should\'ve been ' +
+                        'generated using keyczar.')
+    ]
+    do_register_opts(keyvalue_opts, group='keyvalue')
```
Kami: I would prefer to call it `datastore` and not `keyvalue`, but yeah I know we already call it keyvalue in some places and datastore in others...

tools/st2-generate-symmetric-crypto-key.py
```diff
@@ -0,0 +1,45 @@
...
+
+        if not force:
+            print('Not generating a new key. Either delete the file or re-run with --force.')
+            sys.exit(2)
+
+        print('WARNING: Rewriting existing key with new key!')
+
+    with open(key_path, 'w') as key_file:
+        k = AesKey.Generate()
+        key_file.write(str(k))
```
Kami: Does this return keyczar formated JSON or just the raw key?

If it returns just the raw key, I would prefer to remove this wrapper and just use keyczar directory + file format and keyczar tool for generating / managing keys.

The reason for that is that keyczar tools allows user to handle multiple keys and makes it easy to rotate them. Having out own wrapper provides little value and breaks that functionality so I'm not a fan of it.

st2common/st2common/config.py
lakshmi-kannan: Now you are preferring the opposite of what you preferred months ago.

tools/st2-generate-symmetric-crypto-key.py
lakshmi-kannan: It returns the JSON. I don't see why this wrapper script is a problem. It hides the details of whetehr user wants to generate AES256 or something else. We know the details and letting user use the keyczar utility would mean we have to tell them what kind of key to generate etc. This will be a support problem for sure.

manasdk: If someone prefers keyczar they are free use keyczar, we have a wrapper that makes things a little bit easier and is our opinion. If we need to add more features to this script making it as complex as keyczar we can abandon this approach. At this point it certainly adds value by making things simpler for a user.

lakshmi-kannan:
>Re saving and serializing encrypted data - I can't recall if Keyczar already does that, but if doesn't we should update serialized data so it includes header with the encryption information (which algorithm was used, etc.).

It does not. It assumes symmetric crypto key to be always AES. The library doesn't support other symmetric crypto algos. https://github.com/google/keyczar/blob/master/python/src/keyczar/keys.py#L425

st2common/st2common/config.py
Kami: Unlikely, I'm 98% I was always for `datastore`.

Anyways :P

st2client/st2client/models/core.py
Kami: Removed it in 406a701e563034a053f4bd290865bb97543d936a.

PR #2634, opened 2016-04-20
Fix for publishing `extra` in post_message

emedvedev:
cc @Kami


Kami:
Thanks.

Let's please also add a test case for it because without it it's just another regression waiting to happen.

emedvedev:
The chatops pack needs a full test suite, not just for this. I'll see what we can do with it.

PR #2635, opened 2016-04-20
Fix a bug with gunicorn master process not shutting down

Kami:
The problem was that we performed eventlet monkey patching inside master instead
of only inside workers which broke signal handling (master never received
signals so it never shut down).

### DIFF
--- a/st2api/st2api/app.py
+++ b/st2api/st2api/app.py
@@ -14,7 +14,9 @@
 # limitations under the License.
 
 import os
+import sys
 
+import eventlet
 import pecan
 from oslo_config import cfg
 from pecan.middleware.static import StaticFileMiddleware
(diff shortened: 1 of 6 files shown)


st2api/st2api/app.py
```diff
@@ -52,6 +54,16 @@ def setup_app(config=None):
 
     is_gunicorn = getattr(config, 'is_gunicorn', False)
     if is_gunicorn:
+        # Note: We need to perform monkey patching in the worker. If we do it in
+        # the master process (gunicorn_config.py), it breaks tons of things
+        # including shutdown
+        eventlet.monkey_patch(
```
Kami: Yes, we need to refactor this copy pasta.

Kami: To clarify - monkey patching is really only needed if `sync` gunicorn worker is used. If async one is used, the worker class itself already performs monkey patching.

In any case, I think it's probably still a good idea to leave it here so we can easily switch between sync / async workers without making any code changes (double monkey patch is a no-op).

PR #2638, opened 2016-04-20
Register ANNOUNCEMENT_XCHG in bootstrap_utils.py

cheremushki:
This exchange was not being registered with rabbitmq, causing "no exchange s2.announcement in vhost" errors.


manasdk:
LGTM. Will merge in a bit.

Thanks for the change!

PR #2642, opened 2016-04-21
Formatting fix for the chatops message

emedvedev:
There’s a bug where Jinja renderer returns a list instead of a string, resulting in a formatting glitch. 

Reported by Tracy Teague on the community channel.

### DIFF
--- a/contrib/chatops/actions/workflows/post_result.yaml
+++ b/contrib/chatops/actions/workflows/post_result.yaml
@@ -7,7 +7,7 @@ chain:
       execution_id: "{{execution_id}}"
     publish:
       message: "{{format_execution_result.result.message}}"
-      extra: "{{format_execution_result.result.extra}}"
+      extra: "{% if 'extra' in format_execution_result.result %}{{format_execution_result.result.extra}}{% else %}{}{% endif %}"
     on-success: "post_message"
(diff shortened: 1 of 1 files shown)


Kami:
Probably a good idea to update the changelog :)

Do we plan to release v1.4.1 soon (or what's our plan for backporting those changes)?

Although, as @manasdk has said, I would feel a lot more comfortable when we have some integration tests for ChatOps otherwise this is just another regression or bug waiting to happen.

emedvedev:
It's a fix for the feature that's in the changelog already, I'm not sure if it deserves an update :) I think @DoriftoShoes said we'll just rebuild the packages and release a new revision of 1.4, so you can still use the package manager update to get the fix. There's basically one changed line of yaml, so it's the best course of action.

Tests are a major pain point. I'm dedicating time to work on them this week.

Kami:
Ah, so it has already been merged into v1.4.

I protested against this approach in the past (every commit and change should be trackable back to a specific git revision and correspond to a StackStorm version (e.g. 1.4.1)), but I'll live with it for now :)

And yay for tests.

emedvedev:
I know, not very happy with it either. But we'll make releases great again!

PR #2643, opened 2016-04-22
Acknowledge MQ message only after a TriggerInstance is stored in DB

manasdk:
With this PR we make sure that only after a TriggerInstance is saved in the DB will the message be ack-ed to RMQ. This way if the rulesengine process goes down while the event is being processed at least the event will not be lost.

There are various stages in the handling of a TriggerInstance from the perspective of the RulesEngine
1. `pending` -. RulesEngine picks up the TI from the MQ and records in the DB. At this point it is safe to ACK back to RMQ.
2. `processing` - TI is picked up for processing and figures out the applicable rules etc.
3. `processed` - RulesEngine has determined the executions to create and those are now scheduled.
4. `processing_failed` - TI is picked up but for one or another reason there is a failure to process. There should be more information about this in a corresponding `RuleEnforcement` or in the worst case logs(ugh!)

If RulesEngine crashes after firing off executions but not before moving a TriggerInstance from `processing` -> `processed` then there will be a mis-calculated step. Still there will be no TI loss.

A future improvement could be for `st2` to track such processing failures and go automatically re-process the executions.

```
$ st2 trigger-instance list --status processing
+----+---------+-----------------+--------+
| id | trigger | occurrence_time | status |
+----+---------+-----------------+--------+
+----+---------+-----------------+--------+
```

### Todo

- [x] Decouple Trace creation from message acknowledgement.
- [x] Some doc around StagedMessageHandler


Kami:
Some background / context on this change would be good.

I assume we don't want a message to be ACKed if the TriggerInstance (or underlying Trace) creation fails, right?

Also, afaik, Trace creation should be / is best-effort anyway and we proceed even if it fails, right?

manasdk:
> Some background / context on this change would be good.

Yes, will get more info out.

manasdk:
> Also, afaik, Trace creation should be / is best-effort anyway and we proceed even if it fails, right?

Yes. There are some quirks in how I handled async creation etc which makes is a bit tricky to not depend on trace create but I will fix that part. It is not something I am focusing on at this point.

st2common/st2common/transport/consumers.py
```diff
@@ -86,3 +102,17 @@ def shutdown(self):
     @abc.abstractmethod
     def process(self, message):
         pass
+
+    def _get_queue_consumer(self, connection, queues):
+        return QueueConsumer(connection, queues, self)
+
+
+@six.add_metaclass(abc.ABCMeta)
+class StagedMessageHandler(MessageHandler):
```
lakshmi-kannan: This ideally should be in its own file but whatevs.

st2common/st2common/transport/consumers.py
```diff
@@ -47,26 +47,42 @@ def get_consumers(self, Consumer, channel):
...
             if not isinstance(body, self._handler.message_type):
                 raise TypeError('Received an unexpected type "%s" for payload.' % type(body))
-
-            self._handler.process(body)
+            processed_body = self._handler.pre_ack_process(body)
+            self._dispatcher.dispatch(self._process_message, processed_body)
         except:
             LOG.exception('%s failed to process message: %s', self.__class__.__name__, body)
+        # At this point we will always ack a specific message.
+        message.ack()
```
lakshmi-kannan: I really think you should use finally clause. Otherwise someone could accidentally add a return in except clause causing all sorts of damage. Same for above class.

st2common/st2common/constants/triggers.py
```diff
@@ -36,7 +36,13 @@
 
     'INTERVAL_TIMER_TRIGGER_REF',
     'DATE_TIMER_TRIGGER_REF',
-    'CRON_TIMER_TRIGGER_REF'
+    'CRON_TIMER_TRIGGER_REF',
+
+    'TRIGGER_INSTANCE_STATUSES',
+    'TRIGGER_INSTANCE_STATUS_PENDING',
```
lakshmi-kannan: The word STATUS seems unnecessary really.

lakshmi-kannan:
Looking forward to CLI and API changes.

st2common/st2common/transport/consumers.py
manasdk: E_TOO_MUCH_EFFORT

st2common/st2common/transport/consumers.py
manasdk: done

st2common/st2common/constants/triggers.py
manasdk: Done

st2common/st2common/transport/consumers.py
```diff
@@ -86,3 +107,33 @@ def shutdown(self):
...
+    def _get_queue_consumer(self, connection, queues):
+        return QueueConsumer(connection, queues, self)
+
+
+@six.add_metaclass(abc.ABCMeta)
+class StagedMessageHandler(MessageHandler):
+    """
+    MessageHandler to deal with messages in 2 steps.
+        1. pre_ack_process : This is called on the handler before ack-ing the message.
+        2. process: Called after ack-in the messages
```
Kami: Honestly, for consistency and to make it more obvious, I would prefer `process` to be called `post_ack_process` (or similar), but I imagine you just kept it as `process`, because it was called that before.

PR #2644, opened 2016-04-22
st2 Robot Framework Demo

humblearner:
Changes:

+ Add new directory robotfm
+ Add st2_demo_rfm.rst and variables/st2_rfm_demo.yaml in robotfm/
+ Updated test-requirements.txt with "robot framework" and "docutils"
+ Updated .gitignore for RobotFM related files


lakshmi-kannan:
Nice. This should be helpful in writing e2e tests for CLI. Merge baby merge.

dzimine:
promising, looking forward to see what's coming out of this!

manasdk:
Awesome?

Would this be more of a unit test or integration/system test? e.g. does it need a running StackStorm instance to work against. It might help define where we put these tests i.e. here or in the st2tests.

humblearner:
@manasdk : RF can be used for integration or functional/system test. Yes, it will need a running StackStorm instance to work against. Also, following log reports are auto-generated every time the test cases are executed:

[Logs.zip](https://github.com/StackStorm/st2/files/232313/Logs.zip)

These can be integrated with the CI setup for reporting purpose.

manasdk:
Ah so this fits better with e2e in that case? Likely https://github.com/StackStorm/st2tests is the right home to bootstrap this effort?

I belive you and @lakshmi-kannan spoke about this so not sure how you guys see this integrating.

lakshmi-kannan:
@manasdk I was trying to figure out how this fits into our testing story. So a sample and a suite of tests will help us identify what the tool can and cannot do. I suspect that most of this would end up in st2tests but we can make that call later? I think we should merge and write more tests ourselves to find out. Wdyt?

manasdk:
> I suspect that most of this would end up in st2tests but we can make that call later?

Right.

> I think we should merge and write more tests ourselves to find out. Wdyt?

Per https://github.com/StackStorm/st2/pull/2644#issuecomment-213525328 would need a running st2 instance. This unit-tests and integration-tests we have in this repo do no start st2. Without packages built I am unclear how we can actually run any test from this repo. You see my confusion?

manasdk:
+1

PR #2645, opened 2016-04-22
[WIP] Configuring packs via datastore

Kami:
This is a first iteration of the larger story which will allows users  to configure packs via the datastore. It introduces some basic concepts such as config schema (right now it's inferred from the existing `config.yaml` so it has limitations, especially when multiple levels of object nesting are used inside the config, but that's an anti-pattern anyway) and pack configuration via the datastore.

The main goal of allowing users to configure packs via the datastore is improving our infrastructure as code story around the packs. With this approach, users won't need to edit any files directly inside the pack (eventually we will deprecate `config.yaml`). This will make it easier for the packs to be version controlled, updated, etc.

## TODO

- [x] Integration test for running a Python runner action which utilizes some config values from the datastore
- [ ] Documentation
- [ ] API for managing configuration (get, set, delete)
- [ ] CLI commands for managing config values in the datastore (those command will know how to encode the key name, serialize the value, etc.). Keep in mind that this tool will only be provide for the user convenience (it's kinda like `packs.install` - meant for quick testing, but not a good fit for production deployments). For real life production infrastructure as code deployments, users will be encouraged to manipulate values locally in a JSON file, version control this file and then load this whole JSON file on the servers on deploy. This is the similar to what others (including me) did in the past with the (encrypted) Chef data bags. Chef provides CLI command for locally manipulating databag values. Users use this command to manipulate the version controlled JSON file locally and then deploy it to the servers as part of the deployments process (as opposed to directly manipulating datastore on the servers - this is a bad practice from the infrastructure as code perspective - it's not version controlled, no audit trail, not treated as code, etc.)

## Version 2

Second iteration will build on those two existing primitives and introduce two new concepts described below.

### Config schema (`config.schema.yaml` or similar)

User will be able to specify a schema for the config. This schema will be used (and validated against) when populating / loading config values in the datastore.

You can find an example of the proposed schema here - https://gist.github.com/Kami/72e67c6462fb8d2144c0ac0ad7dc9d8b. As you can see, it can get complex quite fast so we should encourage pack authors to use flat config schemas and layouts (e.g. `user_api_key` vs `user.api_key`).

### Config template (name TBD, suggestions welcome)

Config templates will allow users to specify Jinja expressions for the config key values. Those expressions will be evaluated and used for the key name when retrieving config value from the datastore.

This feature will add more flexibility and allow users to do things such as use different config value depending on the context (e.g. different config based on the user who requested the action execution etc.).

Templates will probably look something like this:

```yaml
api_key: "system.configs.{{pack}}.api_key.{{api_user}}"
region: "system.pack_configs.{{pack}}.region"
```

The goal is for those files to be version controlled and stored somewhere outside the pack directory. This will probably be `/opt/stackstorm/configs/` (e.g. `/opt/stackstorm/configs/<pack name>.yaml`), but subject to change and suggestions :)


st2common/st2common/services/config.py
```diff
@@ -0,0 +1,148 @@
...
+
+LOG = logging.getLogger(__name__)
+
+DATASTORE_KEY_SEPARATOR = '.'
+
+# Prefix for datastore items which store config values
+# Full keys follow this format: system.pack_config.<pack name>.<config key name>
+# For example:
+# pack_config.aws.setup.region
+DATASTORE_CONFIG_KEY_PREFIX = ['system', 'pack_config']
```
lakshmi-kannan: https://github.com/StackStorm/st2/blob/master/st2common/st2common/constants/system.py#L32 Please reuse that.

lakshmi-kannan: Not a fan of pack_config. I actually want to use config.pack_name.foo instead of system.pack_config. This important UX decision needs the eyes of Sir @dzimine, Sir @manasdk.

DoriftoShoes: I agree.  I would rather use `config.pack_name.asdf`

Kami: Yeah, I'm fine with that change.

Also, to clarify, this is really just an implementation detail (key name) and in real-life user should use higher level CLI command which will look something like `st2 config set <pack name> <config key name>` and the API which will handle this command will be responsible for handling implementation details aka serializing the key name.

Having said that, I do agree that consistency and good key names are still important so I will look into it.

Kami:
Closing in favor of #2708.

PR #2658, opened 2016-04-30
Fix for data is dropped if message is not present in notification

manasdk:


lakshmi-kannan:
As kami would say, no tests!!!

manasdk:
> As kami would say, no tests!!!

Fixed b3965bc5996505481e5c1545e4e5055b5e6d3223

st2common/st2common/models/api/notification.py
```diff
@@ -94,7 +94,7 @@ def _from_model_sub_schema(notify_sub_schema_model):
 
         if getattr(notify_sub_schema_model, 'message', None):
             notify_sub_schema['message'] = notify_sub_schema_model.message
-        if getattr(notify_sub_schema_model, 'message', None):
+        if getattr(notify_sub_schema_model, 'data', None):
```
Kami: Good catch.

Kami:
Good catch, LGTM, +1.

(I believe a user on Slack had a similar issue a couple of days ago)

PR #2659, opened 2016-05-01
Update jsexpose so it knows how to handle "bool" type, various improvements and fixes

Kami:
Looks like @lakshmi-kannan missed or ignored my comments in the secrets PR so I decided to make some improvements to `jsexpose` myself :P

1. Update jsexpose so it knows how to handle `bool` type. Perhaps better and more explicit name for this argument would be `cast_funcs`, but that's something we can rename in the future aka separate PR.
2. Update jsexpose so it always consumes keyword arguments first. This way it works correctly if not all of the arguments are provided and we don't need to abuse "**kwargs" in the method signature.
3. Update jsexpose to handle more edge cases and invalid HTTP requests. The code which performs those checks is hacky, but it's better than not having anything and silently swallowing and ignoring invalid requests. As mentioned below, proper solution is moving away from jsexpose to something better and more sane.

Note: As I mentioned many times before already, I hate `jsexpose`. It's a huge hack and a mess, but I'm trying to improve it one step at a time. Ideally though, we could totally get rid of it and replace it with something more strict, correct and less fragile in the future.

## TODO

- [x] Test cases for all the jsexponse (edge) cases with arg_types
- [x] Code refactor (refactor some of the code from jsexpose into standalone utility functions)

### DIFF
--- a/st2common/st2common/models/api/base.py
+++ b/st2common/st2common/models/api/base.py
@@ -146,7 +146,7 @@ class APIUIDMixin(object):
 
 def jsexpose(arg_types=None, body_cls=None, status_code=None, content_type='application/json'):
     """
-    :param arg_types: A list of types for the function arguments.
+    :param arg_types: A list of types for the function arguments (e.g. [str, str, int, bool]).
     :type arg_types: ``list``
 
     :param body_cls: Request body class. If provided, this class will be used to create an instance
--- a/st2common/st2common/models/api/base.py
+++ b/st2common/st2common/models/api/base.py
@@ -185,21 +185,29 @@ def jsexpose(arg_types=None, body_cls=None, status_code=None, content_type='appl
                 argspec = inspect.getargspec(f)
                 names = argspec.args[1:]
 
-                for name in names:
-                    try:
-                        value = args.pop(0)
-                        value = cast_value(value_type=types.pop(0), value=value)
-                        more.append(value)
-                    except IndexError:
+                for index, name in enumerate(names):
+                    # 1. Try kwargs first
+                    if name in kwargs:
                         try:
                             value = kwargs[name]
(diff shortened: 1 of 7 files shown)


st2common/st2common/models/api/base.py
```diff
@@ -171,22 +171,43 @@ def callfunction(*args, **kwargs):
...
+                    cast_func = value_type
+
+                result = cast_func(value)
+                return result
+
             if types:
                 argspec = inspect.getargspec(f)
                 names = argspec.args[1:]
 
-                for name in names:
```
Kami: This approach with trying to consume non-keyword arguments first was very fragile.

It only really worked correctly if method didn't take more than a single parameter. If method took more than one parameter (e.g. `get_all` in this PR), we needed to abuse `kwargs` to make it work.

The problem with abusing kwargs is that it breaks whole "arg_types" functionality / abstraction and user can also pass arbitrary unsupported parameters to the API and we will just simply swallow them and ignore it (or in some cases where we pass kwargs directly to the underlying methods, things will explode in a weird manner).

Kami:
I also pushed a change which updates KeyValuePair API controller to utilize base `ResourceController` class (e42934186657a56a0ead6b1f1382cd3c751cbf44).

There was a TODO note for that and this should have been done a long time ago. Not utilizing this base class means that some things like limit, sort, etc. didn't work correctly and / or in the same manner as those things work with other API endpoints.

lakshmi-kannan:
>Looks like @lakshmi-kannan missed or ignored my comments in the secrets PR so I decided to make some improvements to jsexpose myself

No, I responded saying I tried but the changes were not simple and should be in another PR.

Kami:
@lakshmi-kannan OK, sorry, looks like I missed your comment then.

PR #2661, opened 2016-05-02
Use new Mistral task function to publish results.

jjm:
The promised PR to fix the issue with out dated Mistral publishing in `packs.deploy` workflow.

As outlined in StackStorm/st2/issues/2660.

### DIFF
--- a/contrib/packs/actions/workflows/deploy.yaml
+++ b/contrib/packs/actions/workflows/deploy.yaml
@@ -58,8 +58,8 @@ workflows:
           packs: <% $.packs %>
 
         publish:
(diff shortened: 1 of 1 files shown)


jjm:
Looks like I've stopped and error, just doing another couple of tests.

jjm:
It's good now @Kami, tested with a couple of auto deploys and manual deploy via chatOps

Kami:
Thanks.

The change looks good to me.

This also brings up a good point we have already debated in the past - we really need to come up with some framework which will making testing workflows end-to-end easier.

PR #2666, opened 2016-05-04
Remove code for deprecated fabric based remote runner

Kami:
Fabric runner has been deprecated almost half a year ago now so it's only time to get rid of the unused code.

Totals:

* - 1 direct dependency
* -728 lines of code

Resolves #2160.

Issue #2662, opened 2016-05-02
Feature request: SSH keys in rules = path or output of command.


murgurgula:
It would be nice to be able to write the path or be a result of a command like $(cat /path/to/file) in the StackStorm Rules instead of writing the actual key 
Reason: 
- Make the config more readable.
It would be easier to see if the proper key is used.

- Reason for "result of a command"
Plug stackstorm to Vault or another system destinated to manage credentials.

Kami:
Thanks for the feedback.

The reason why we have decided to go with the actual key content and not path is because action runner which executes actions can be deployed on many different servers. If we went with the path approach, you would need to ensure the key file is located in the same location on every server where action runner component is running.

One option for you would be to save the raw key content in the datastore and reference this datastore value in the rule. Right now we are also working on improved support for secrets in the datastore - we will allow you to encrypt secret values in the datastore and this will make datastore more suitable for storing secrets like that.

As far as result of the command goes - you can achieve that today by using a workflow - one action talks to vault or similar and then you pass this action's result to another action (in fact, that's what some of our users already do today).

Allowing user to execute arbitrary code or similar in rules would change the whole (security) model and of rules and that's something you can already achieve today using workflows and actions which were designed specially for that. You can see more background on why I don't think this is a good fit for rules in my comment on this issue - https://github.com/StackStorm/st2/issues/2654.

murgurgula:
I understand, but having the private key in a rule is a security issue.
Rules, custom packs, etc most likely are stored in git. In a company, where infra is split (you do windows, I do linux, bobby takes care of the mail servers, Teddy is the network guy, etc), git may be in the control of someone else and having the keys in there may not be that suitable.

If it was a path the sysadmin would have to worry only about securing the actual key file, however, when the content of the actual key is spread across numerous files, it will become hard to track.

Is it hard to do both options?

Kami:
@murgurgula Yes I agree that this approach is not ideal.

We could support both (also support path to the key file and document limitations of that approach - key needs to be deployed on all the servers, etc.), but I'm not too sure this will still be needed once support for encrypted datastore values is in place.

murgurgula:
Not for someone who wants security, I would prefer encrypted datastore.

Kami:
This has been resolved by #2668 and it will be included in the next stable release. Closing.

PR #2668, opened 2016-05-04
Support for path to the private key file for the "private_key" argument

Kami:
Note: This PR is based on #2666.

This pull request updates remote runner so user can now also specify a path to the private key file for the `private_key` remote runner argument.

This resolves security concerns from #2662.

We are currently working on secrets in datastore story which goal is to solve some of the issues like that, but I still see value in allowing users to specify a path to the key file. A lot of organizations already have a code / configuration management in place or managing secret files on disk so it makes sense to allow users to re-use that functionality instead of forcing them to use datastore secrets which is StackStorm specific functionality.

In addition to that, I also fixed support for password protected keys (see https://github.com/StackStorm/st2/pull/2596#issuecomment-216768910 for details) and removed support for the deprecated `key` argument since the whole logic for that was kinda convoluted and made it hard to follow / understand.

## TODO:

- [x] Docs
- [x] Tests

### DIFF
--- a/st2common/st2common/bootstrap/runnersregistrar.py
+++ b/st2common/st2common/bootstrap/runnersregistrar.py
@@ -134,8 +134,8 @@ RUNNER_TYPES = [
                 'secret': True
             },
             'private_key': {
-                'description': ('Private key material to log in. Note: This needs to be actual '
-                                'private key data and NOT path.'),
+                'description': ('Private key material or path to the private key file on disk '
+                                'used to log in.'),
                 'type': 'string',
                 'required': False,
                 'secret': True
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,7 @@ in development
 * Fix for `data` is dropped if `message` is not present in notification. (bug-fix)
 * Remove now deprecated Fabric based remote runner and corresponding
   ``ssh_runner.use_paramiko_ssh_runner`` config option. (cleanup)
+* Fix support for password protected private key files in the remote runner. (bug-fix)
 
 1.4.0 - April 18, 2016
 ----------------------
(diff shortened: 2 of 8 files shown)


Kami:
I've added some tests for `ParallelSSHClient` instantiation inside the runner `pre_run` method - a9df32e701b7e9518055141cc5886f50bbe8e072

Previously we had no tests for this code path so we didn't tests the runner argument passing end to end (we only tested ParallelSSHClient and ParamikoSSHClient independently) which means runner could be broken end to end and our tests wouldn't catch that.

That's also the reason why we didn't catch that key passpharse support never worked in the first place - we didn't have any tests for that code path.

st2common/st2common/bootstrap/runnersregistrar.py
```diff
@@ -134,8 +134,8 @@
                 'secret': True
             },
             'private_key': {
-                'description': ('Private key material to log in. Note: This needs to be actual '
-                                'private key data and NOT path.'),
+                'description': ('Private key material or path to the private key file on disk '
```
lakshmi-kannan: Depending on what user the action runner runs as, this is a huge problem without documentation. If the action runner process runs as ``st2``, st2 needs to be able to read this key. There should be docs around the risk of making the key readable by st2. Without a corrresponding change to deprecate core.local, this IMO is a huge issue. I am -1 to supporting path to key file.

Kami: I agree about the risk and the need for the documentation on it.

Kami: In any case, I will add some docs on the limitations and security implications of this approach.

Then we can decide if we are fine with it or not. If we are not, I will open a new PR and at least cherry-pick some fixes, improvements and tests from this PR.

Kami: Added docs which clarify the requirements and potential security risks when using this feature - https://github.com/StackStorm/st2docs/pull/147.

Kami:
I added the docs in https://github.com/StackStorm/st2docs/pull/147 so I will go ahead and merge this.

PR #2669, opened 2016-05-04
Introduce 'scope' field in key value model

lakshmi-kannan:
## TODO:

* [X] GET on user scoped variables  ** Hits V1 APIs with ?scope= query param. 
  * [x] Validate allowed scopes 8de9c81
* [X] LIST on scoped variables ** Hits V1 APIs with ?scope= query param.
  * [x] Validate allowed scopes 8de9c81
* [X] V2 APIs for scoped key value access 

* [x] Backward compatibility tests
  * [x] Migration script if needed
  * [ ] st2-packages changes to include migration script. DECIDED TO WRITE DOCS FOR MIGRATION
* [X] KeyValueLookup should work with user scopes
  * [X] User scoped variables need to be written to user name as prefix.
  * [ ] User context should be passed to all places where  jinja template {{user.foo}} is supported See https://github.com/StackStorm/st2/issues/2678
  * [ ] Decide how to handle {{user.foo}} and implement it See https://github.com/StackStorm/st2/issues/2678
* [x] Tests 
* [ ] Docs -->  Decided to write this after we support resolving jinja resolution.
* [x] CLI shows 400: Bad request but doesn't tell what is bad about the request
* [x] Changelog

Prefix based RBAC will be in another PR. Trying to scope things down. 

### GET

```
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●●)$ st2 key set company brocade     [ruby-1.9.3p484]
+------------------+---------+
| Property         | Value   |
+------------------+---------+
| name             | company |
| value            | brocade |
| expire_timestamp |         |
+------------------+---------+
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●●)$ st2 key get company             [ruby-1.9.3p484]
+------------------+---------+
| Property         | Value   |
+------------------+---------+
| name             | company |
| value            | brocade |
| secret           | False   |
| encrypted        | False   |
| scope            | system  |
| expire_timestamp |         |
+------------------+---------+
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●●)$ st2 key set boss lakstorm --scope=user
+------------------+----------+
| Property         | Value    |
+------------------+----------+
| name             | boss     |
| value            | lakstorm |
| expire_timestamp |          |
+------------------+----------+

## XXX: Get needs to be fixed so asking for something without scope "user" doesn't return it. 

(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●●)$ st2 key get boss                [ruby-1.9.3p484]
+------------------+----------+
| Property         | Value    |
+------------------+----------+
| name             | boss     |
| value            | lakstorm |
| secret           | False    |
| encrypted        | False    |
| scope            | user     |
| expire_timestamp |          |
+------------------+----------+
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●●)$ st2 key set manas grump --scope=nsfw
MESSAGE: Invalid scope "nsfw"! Allowed scopes are ['system', 'user']. for url: http://127.0.0.1:9101/v1/keys/manas

(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●●)$
```

### LIST 

```
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●●)$ st2 key list --scope=user       [ruby-1.9.3p484]
+------+-------+--------+-----------+------------------+
| name | value | secret | encrypted | expire_timestamp |
+------+-------+--------+-----------+------------------+
+------+-------+--------+-----------+------------------+
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●●)$ st2 key list --scope=system     [ruby-1.9.3p484]
+------+-------+--------+-----------+------------------+
| name | value | secret | encrypted | expire_timestamp |
+------+-------+--------+-----------+------------------+
| foo  | bar   | False  | False     |                  |
+------+-------+--------+-----------+------------------+
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●●)$ st2 key set boss lakstorm --scope=user
+------------------+----------+
| Property         | Value    |
+------------------+----------+
| name             | boss     |
| value            | lakstorm |
| expire_timestamp |          |
+------------------+----------+
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●●)$ st2 key list --scope=user       [ruby-1.9.3p484]
+------+----------+--------+-----------+------------------+
| name | value    | secret | encrypted | expire_timestamp |
+------+----------+--------+-----------+------------------+
| boss | lakstorm | False  | False     |                  |
+------+----------+--------+-----------+------------------+
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●●)$
```

### PUT
```
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●)$ http PUT "http://localhost:9101/v2/keys/user/company" < kv.json
HTTP/1.1 200 OK
Access-Control-Allow-Origin: http://172.168.50.50:3000
Connection: keep-alive
Content-Length: 195
Date: Sat, 07 May 2016 01:07:40 GMT
X-Request-Id: 24183b1b-30f1-45e3-a4ab-4460a7fa620d

{
    "encrypted": false,
    "id": "572d3fdcd9d7ed14d88a3e23",
    "name": "company",
    "scope": "user",
    "secret": false,
    "uid": "key_value_pair:user:company",
    "value": "BROcade"
}

(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●)$
```

### DELETE

```
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●)$ http DELETE "http://localhost:9101/v2/keys/user/company"
HTTP/1.1 204 No Content
Access-Control-Allow-Origin: http://172.168.50.50:3000
Connection: keep-alive
Content-Length: 4
Date: Sat, 07 May 2016 01:08:02 GMT
X-Request-Id: 673981e0-c1f0-4117-80cf-7d539567e64d



(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●)$
```

### GET

```
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●)$ http "http://localhost:9101/v2/keys/user/foo"
HTTP/1.1 200 OK
Access-Control-Allow-Origin: http://172.168.50.50:3000
Connection: keep-alive


  1 {
Content-Length: 187
Date: Sat, 07 May 2016 01:06:02 GMT
X-Request-Id: 1a1bfc27-c54f-4ac6-8ab9-5107954c2dd2

{
    "encrypted": false,
    "id": "572d3dd7d9d7ed0c17a9857c",
    "name": "foo",
    "scope": "user",
    "secret": false,
    "uid": "key_value_pair:user:foo",
    "value": "userbar"
}

(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●)$ http "http://localhost:9101/v2/keys/system/foo"
HTTP/1.1 200 OK
Access-Control-Allow-Origin: http://172.168.50.50:3000
Connection: keep-alive
Content-Length: 187
Date: Sat, 07 May 2016 01:06:07 GMT
X-Request-Id: 493f459f-b9df-4fc6-a200-d122298e61d6

{
    "encrypted": false,
    "id": "572d3dbed9d7ed0c17a9857b",
    "name": "foo",
    "scope": "system",
    "secret": false,
    "uid": "key_value_pair:system:foo",
    "value": "bar"
}

(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●)$
```

### v2 APIs with secrets

```
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●)$ http PUT "http://localhost:9101/v2/keys/user/company" < kv_secret.json
HTTP/1.1 200 OK
Access-Control-Allow-Origin: http://172.168.50.50:3000
Connection: keep-alive
Content-Length: 300
Date: Sat, 07 May 2016 01:10:30 GMT
X-Request-Id: 33cc6994-f140-42a9-a3b4-31211d79220c

{
    "encrypted": true,
    "id": "572d4086d9d7ed14d88a3e24",
    "name": "company",
    "scope": "user",
    "secret": true,
    "uid": "key_value_pair:user:company",
    "value": "0083E1033A54F9FAB3336FDA3F6FE0A658A02A21DC6BA530A5FB062B13999A85418906DEE8D62CFD13BDF64863013DBEDAB998C742D4767A10"
}

(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●)$ http "http://localhost:9101/v2/keys/user/company"
HTTP/1.1 200 OK
Access-Control-Allow-Origin: http://172.168.50.50:3000
Connection: keep-alive
Content-Length: 300
Date: Sat, 07 May 2016 01:10:45 GMT
X-Request-Id: 45f12a60-8593-4ae1-800d-befa8590f8c0

{
    "encrypted": true,
    "id": "572d4086d9d7ed14d88a3e24",
    "name": "company",
    "scope": "user",
    "secret": true,
    "uid": "key_value_pair:user:company",
    "value": "0083E1033A54F9FAB3336FDA3F6FE0A658A02A21DC6BA530A5FB062B13999A85418906DEE8D62CFD13BDF64863013DBEDAB998C742D4767A10"
}

(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●)$ http "http://localhost:9101/v2/keys/user/company?decrypt=true"
HTTP/1.1 200 OK
Access-Control-Allow-Origin: http://172.168.50.50:3000
Connection: keep-alive
Content-Length: 194
Date: Sat, 07 May 2016 01:10:53 GMT
X-Request-Id: 623d82d8-bc63-45e1-a4a7-4c0743e93c72

{
    "encrypted": false,
    "id": "572d4086d9d7ed14d88a3e24",
    "name": "company",
    "scope": "user",
    "secret": true,
    "uid": "key_value_pair:user:company",
    "value": "BROcade"
}

(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●)$
```


st2common/st2common/models/api/keyvalue.py
```diff
@@ -114,17 +120,22 @@ def from_model(cls, model, mask_secrets=True):
...
+        secret = getattr(model, 'secret', False)
+        if secret:
             encrypted = True
 
-        if not mask_secrets and model.secret:
+        if not mask_secrets and secret:
             doc['value'] = symmetric_decrypt(KeyValuePairAPI.crypto_key, model.value)
             encrypted = False
 
+        scope = getattr(model, 'scope', SYSTEM_KV_PREFIX)
```
Kami: I personally view scope and key prefix as two different things and they don't need to be the same (scope is a user visible feature and key prefix for a particular scope is an implementation detail).

In any case, I would have those scopes:

* `user` - User scopes, keys are prefixed with some common prefix for this scope (`user`.<user pk aka username>.foo` or `_user.<user pk aka username>.foo`)
* `global` - Anything which doesn't have a prefix (so any other existing key). We could explicitly prefix keys with a global scope, but the question is how to handle the keys without the prefix then.

Or do you want to introduce also a concept of no scope?

st2common/st2common/constants/system.py
```diff
@@ -30,3 +30,11 @@
 
 # Used to prefix all system variables stored in the key-value store.
 SYSTEM_KV_PREFIX = 'system'
+
+# Used to prefix all user scoped variables in key-value store.
+USER_KV_PREFIX = 'user'
+
+ALLOWED_KV_PREFIXES = [
+    SYSTEM_KV_PREFIX,
```
Kami: I think we should think about it some more on how to handle system / global scope.

Right now this is using `SYSTEM_KV_PREFIX` prefix which is totally unrelated to scopes and used for a different purpose. We only use this feature so we can identify, that inside a Jinja variable expression, user wants to reference value from a datastore.

So we should think if it makes sense to conflate / combine those things together or it will just make it more confusing.

st2common/st2common/models/api/keyvalue.py
lakshmi-kannan: I like "user" and "global" scopes as well. I even anticipate more scopes like "environment" etc. Our current story for key value store is everything is global and you can "refer" to a variable via  {{system.foo}} even though "foo" is the actual key name. So for backward compatibility, I am setting the scope to "system" and don't want to introduce another name "global". If you guys think this is a bad idea, we should talk. /cc: @manasdk

st2common/st2common/constants/system.py
lakshmi-kannan: See my response to prev comment.

st2api/st2api/controllers/v1/keyvalue.py
```diff
@@ -42,25 +43,36 @@ class KeyValuePairController(ResourceController):
...
         from_model_kwargs = {'mask_secrets': not decrypt}
-        kvp_db = super(KeyValuePairController, self)._get_one_by_name_or_id(name_or_id=name,
-            from_model_kwargs=from_model_kwargs)
+        if scope:
+            kwargs = {'name': name, 'scope': scope}
+            kvp_db = super(KeyValuePairController, self)._get_all(
+                from_model_kwargs=from_model_kwargs,
+                **kwargs
+            )
+            kvp_db = kvp_db[0] if kvp_db else None
```
manasdk: why/how would get_one by name+scope return multiple values?

I feel this code needs to be in the service layer where there is a method `get_one(scope, name)`

st2api/st2api/controllers/v1/keyvalue.py
```diff
@@ -42,25 +43,36 @@ class KeyValuePairController(ResourceController):
...
 
     def __init__(self):
         super(KeyValuePairController, self).__init__()
         self._coordinator = coordination.get_coordinator()
         self.get_one_db_method = self._get_by_name
 
-    @jsexpose(arg_types=[str, bool])
-    def get_one(self, name, decrypt=False):
+    @jsexpose(arg_types=[str, str, bool])
+    def get_one(self, name, scope=SYSTEM_KV_PREFIX, decrypt=False):
```
manasdk: It is weird to see scope after name but backwards compatbility.

st2common/st2common/models/api/keyvalue.py
manasdk: If we did design this from scratch we would have the `scope` property and assign it to `system`/`global` by default. The other choice would be to have `{{system.foo}}` and `{{system.user.foo}}` where `system` is still a prefix and what follows is possibly a scope i.e. `user` or something else in the future if we allow for user configurable scopes. I actually do prefer `system` transitioning to a scope although it is really from the scope perspective better named global. I wish I had been prescient and made space for this when I added ability to look up kv-store values in rules, actions etc.

 Taking this forward it would mean that when `{{system.foo}}` and `{{user.foo}}` we now grab 2 top level words i.e. `system` and `user` for the context addressable space. Not ideal but not really much worse that what we have already. I feel we should only grab `{{$user.foo}}` or some such obscure token where 2 purposes are served -
1. `$` suggest it will be replaced
2. `$user` is less-likely to appear in an addressable context

st2common/st2common/models/db/keyvalue.py
```diff
@@ -32,12 +33,13 @@ class KeyValuePairDB(stormbase.StormBaseDB, stormbase.UIDFieldMixin):
...
 
     RESOURCE_TYPE = ResourceType.KEY_VALUE_PAIR
-    UID_FIELDS = ['name']
+    UID_FIELDS = ['scope', 'name']
 
     name = me.StringField(required=True, unique=True)
     value = me.StringField()
     secret = me.BooleanField(default=False)
     expire_timestamp = me.DateTimeField()
+    scope = me.StringField(default=SYSTEM_KV_PREFIX)
```
manasdk: What happens to old data in DB with new code? Still works?

lakshmi-kannan: I captured it in a TODO item. Backward compatibility. Yet to be done.

manasdk: This is what happens when I read code and dont read the description :P

st2api/st2api/controllers/v1/keyvalue.py
lakshmi-kannan: >why/how would get_one by name+scope return multiple values?

It is not returning more than one value. It simply returns an array with one item which is both counter intuitive and blows up CLI.

manasdk:
@lakshmi-kannan What does `st2 key list` show? Would there be a scope column?

lakshmi-kannan:
@manasdk Not yet

```
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●●)$ st2 key list                    [ruby-1.9.3p484]
+------+-------+--------+-----------+------------------+
| name | value | secret | encrypted | expire_timestamp |
+------+-------+--------+-----------+------------------+
| foo  | bar   | False  | False     |                  |
+------+-------+--------+-----------+------------------+
(virtualenv)vagrant@st2dev /mnt/src/storm/st2 (issue_167/user_scoped_vars●)$
```

st2api/st2api/controllers/resource.py
```diff
@@ -242,6 +242,25 @@ def _get_from_model_kwargs_for_request(self, request):
         """
         return self.from_model_kwargs
 
+    def _get_one_by_scope_and_name(self, scope, name, from_model_kwargs=None):
```
lakshmi-kannan: @manasdk FYI

st2api/st2api/controllers/resource.py
```diff
@@ -246,6 +246,25 @@ def _get_from_model_kwargs_for_request(self, request):
...
+
+        :param scope: Scope the key belongs to.
+        :type scope: ``str``
+
+        :param name: Name of the key.
+        :type name: ``str``
+        """
+        instance = self.access.get_by_scope_and_name(scope=scope, name=name)
+        if not instance:
+            return instance
```
Kami: `_get_one_*` methods usually return 404 (`pecan.abort`) if a object is not found (and `_get_by_*` throw `StackStormDBObjectNotFoundError` if an object is not found, and `_get_one` methods usually call `_get_by` methods).

We should do the same here for consistency and also add corresponding test case.

Kami: Also seems like this is kinda key value pair controller specific so maybe it should live in base key value pair controller or similar?

st2client/st2client/commands/keyvalue.py
```diff
@@ -70,13 +70,17 @@ def __init__(self, *args, **kwargs):
...
+        self.parser.add_argument('-s', '--scope', default='system', dest='scope',
+                                 help='Scope variable is under. Example: "user".')
 
     def run_and_print(self, args, **kwargs):
         if args.prefix:
             kwargs['prefix'] = args.prefix
 
         decrypt = getattr(args, 'decrypt', False)
         kwargs['params'] = {'decrypt': str(decrypt).lower()}
+        scope = getattr(args, 'scope', 'system')
```
Kami: Maybe want to use constant here?

(I know we can't re-use st2common stuff here...)

st2client/st2client/commands/resource.py
```diff
@@ -154,6 +155,7 @@ def get_resource_by_pk(self, pk, **kwargs):
         try:
             instance = self.manager.get_by_id(pk, **kwargs)
         except Exception as e:
+            traceback.print_exc()
```
Kami: Is this intentional or you just added it for debugging only purposes?

If it's intentional, we should probably be consistent with the rest of the code and only print it if `--debug` flag is passed to the CLI.

st2common/st2common/models/api/keyvalue.py
```diff
@@ -111,20 +118,27 @@ def from_model(cls, model, mask_secrets=True):
...
-        if not mask_secrets and model.secret:
+        if not mask_secrets and secret:
             doc['value'] = symmetric_decrypt(KeyValuePairAPI.crypto_key, model.value)
             encrypted = False
 
+        scope = getattr(model, 'scope', SYSTEM_SCOPE)
+        if scope:
+            doc['scope'] = scope
+
+        key = doc.get('name', None)
```
Kami: Hm, how would this work if `name` is not present?

st2common/st2common/models/api/keyvalue.py
```diff
@@ -145,16 +160,24 @@ def to_model(cls, kvp):
...
                        (KeyValuePairAPI.crypto_key_path, name))
                 raise CryptoKeyNotSetupException(msg)
             value = symmetric_encrypt(KeyValuePairAPI.crypto_key, value)
-            secret = True
 
-        model = cls.model(name=name, description=description, value=value,
-                          secret=secret,
+        scope = getattr(kvp, 'scope', SYSTEM_SCOPE)
+
+        if scope not in ALLOWED_SCOPES:
```
Kami: Is this really the right place for that?

Imo, it kinda belongs in the API controller and we should return 400 if user provides an invalid scope...

st2common/st2common/services/keyvalues.py
```diff
@@ -42,15 +46,72 @@ def _get(self, name):
...
+
+    def __getitem__(self, key):
+        return self._get(key)
+
+    def __getattr__(self, name):
+        return self._get(name)
+
+    def _get(self, name):
+        # get the value for this key and save in value_cache
+        if not self._key_prefix:
```
Kami: Don't we want to re-use `UserKeyReference` class here?

st2common/st2common/services/keyvalues.py
```diff
@@ -42,15 +46,72 @@ def _get(self, name):
...
-            # ValueErrors are expected in case of partial lookups
-            pass
-        # A good default value for un-matched value is empty string since that will be used
-        # for rendering templates.
+        scope = self._scope
+        kvp = KeyValuePair.get_by_scope_and_name(scope=scope, name=key)
         return kvp.value if kvp else ''
+
+
+class UserKeyValueLookup(object):
```
Kami: I would personally move those look-up classes somewhere outside of `services` module since it's more like Jinja specific functionality...

st2common/st2common/util/templating.py
```diff
@@ -56,7 +56,7 @@ def render_template_with_system_context(value):
     :type context: ``dict``
     """
     context = {
-        SYSTEM_KV_PREFIX: KeyValueLookup(),
+        SYSTEM_SCOPE: KeyValueLookup(scope=SYSTEM_SCOPE)
```
Kami: Yeah, I like this a lot more (renaming prefix with scope).

Kami:
I've added some in-line comments.

Overall, I'm fine with other changes, but as mentioned on Slack, I would remove v2 stuff from this PR.

I think we all need to get on the same page and have a consistent story for the v2 API so we don't need to deprecate v2 yet again in the near future when we plan to work on an improved API.

Also, what's our plan for RBAC here? Will this be done separately, or?

manasdk:
>  I would remove v2 stuff from this PR.

What is wrong with it? is the v2 API bad already?

Kami:
@manasdk The problem is that it's a one off-thing.

We need a consistent story and plan for it (that means how paths are going to look like, responses, etc.). I'm fine with adding it as soon as we have a consistent story and plan for it, because then we can make sure this new v2 API follows all of those conventions.

Tomorrow I plan to start a discussion thread on v2 API and we can continue discussion about that there.

manasdk:
> The problem is that it's a one off-thing.

Is it ad-hoc? Isn't this exactly the reason i.e. we are changing a model therefore need to have a new endpoint. I am fine with the discussion driving the impl and then coming around to picking up changes here but why would you suggest reverting the changes?

Kami:
@manasdk Yes, and the same end result can be achieved using v1 API so I see no point in rushing v2 API until we have a clear picture on how we want it to look.

st2api/st2api/controllers/resource.py
lakshmi-kannan: You made that change already. I'll wait for you to merge that and make the corresponding change here.

st2client/st2client/commands/resource.py
lakshmi-kannan: It's intentional. Dude, we don't print the exception and bubble up a weird exception. This is not just a --debug case. This is useful.

st2common/st2common/services/keyvalues.py
lakshmi-kannan: I am not a fan either. Moving this around is a pain now to fix imports. Let me give it a shot.

st2common/st2common/models/api/keyvalue.py
lakshmi-kannan: 5c64e1090bea1f2e62d0f86ff729aff7a5371a62

st2common/st2common/services/keyvalues.py
lakshmi-kannan: 81b76ef43eedeff33c81aea1566b63c40e8525e7

st2client/st2client/commands/keyvalue.py
lakshmi-kannan: e113f08fffdd46ecc9be8eb2c841b3dec3d118d5

st2common/st2common/models/api/keyvalue.py
lakshmi-kannan: I just did that for safety. If name isn't present, we won't be able to even write to db so it's fine.

st2api/st2api/controllers/resource.py
lakshmi-kannan: >Also seems like this is kinda key value pair controller specific so maybe it should live in base key value pair controller or similar?

I had it in KeyValuePairController and manas said move it to some base layer. Geez. Ignoring.

lakshmi-kannan: Addressed StackStormDBObjectNotFoundError - feef877eee0c91fb8906c32ec0d5a0caa92c3d60

lakshmi-kannan:
@manasdk @Kami : Removed v2 APIs here 67d7ed9

lakshmi-kannan:
@Kami: > Also, what's our plan for RBAC here? Will this be done separately, or?

You should read the description sometimes before complaining everywhere.

st2api/st2api/controllers/resource.py
Kami: Lesson -> don't listen to manas :P

j/k

st2api/st2api/controllers/root.py
```diff
@@ -33,9 +34,12 @@ class RootController(BaseRootController):
 
     def __init__(self):
         v1_controller = v1_root.RootController()
+        v2_controller = v2_root.RootController()
```
Kami: Probably forgot to remove this?

PR #2671, opened 2016-05-05
Remove sudo support from the local runner, allow administrator to disable the local runner

Kami:
This pull request removes `sudo` attribute from the local runner and allows administrator to completely disable the local runner.

We have discussed this in the past since allowing sudo access for local runner means any user which has access to `core.local_sudo` action can run arbitrary command as root. In many cases that's undesirable and can present a security risk.

We have also discussed potentially fully removing local runner in the future, but that's something which needs more work since we will need to find good alternatives / replacements for some common things where local runner comes handy and it's used a lot today (sleep, echo, exit, noop, etc.).

## TODO

- [x] Documentation (notably, upgrade notes) - https://github.com/StackStorm/st2docs/pull/146
- [x] Tests
- [x] Update affected packs in st2contrib (remove `sudo: immutable: true` parameter) - https://github.com/StackStorm/st2contrib/pull/489
- [ ] Utilize `RunnerTypeDB.enabled` attribute for disabling a runner instead of the config option
  - [ ] Make sure enabled attribute doesn't get overriten on runner registration
  - [ ] Add API endpoint for disabling / enabling a runner
  - [ ] Add related CLI commands
- [ ] Check and update affected st2tests tests, if needed
- [ ] Update installer and docs (remove paswordless sudo requirement for stanley)


Kami:
Looks like there is one blocker for this change right now - ansible pack (https://github.com/StackStorm/st2contrib/pull/489#issuecomment-217102811).

I'm not sure how to proceed, any suggestions?

manasdk:
LGTM

CHANGELOG.rst
```diff
@@ -15,6 +15,11 @@ in development
 * Fix for `data` is dropped if `message` is not present in notification. (bug-fix)
 * Remove now deprecated Fabric based remote runner and corresponding
   ``ssh_runner.use_paramiko_ssh_runner`` config option. (cleanup)
+* Remove ``sudo`` support for local runner actions. This also means removing ``sudo`` parameter
+  from ``core.local`` action and removing ``core.local_sudo`` action completely.
+* Allow operator to disable local runner actions by setting ``local_runner.enable`` config option
```
manasdk: I thought we could enable/disable runners

Kami: Yeah, that's a good point - we do actually have `enabled` attribute on the `RunnerTypeDB` model, but it gets overwritten when running register content when runners are registered.

I will look into it and see if I can make this existing attribute work without overwriting it on runner registration. I would definitely prefer to get this to work over re-inventing a new, not so consistent approach for disabling a runner.

conf/st2.conf.sample
```diff
@@ -83,14 +83,26 @@ lock_timeout = 60
 username = None
 # Connection retry total time (minutes).
 connection_retry_max_delay_m = 3
+# ca_certs file contains a set of concatenated CA certificates, which are                          used to validate certificates passed from MongoDB.
+ssl_ca_certs = None
+# Certificate file used to identify the localconnection
+ssl_certfile = None
 # Connection retry backoff max (seconds).
 connection_retry_backoff_max_s = 10
+# Specifies whether a certificate is required from the other side of the                          connection, and whether it will be validated if provided
```
lakshmi-kannan: what's that weird space?

CHANGELOG.rst
lakshmi-kannan: Meh. I am fine with the config option but whatever you guys decide is fine by me.

conf/st2.conf.sample
manasdk: Multiline comment being rendered poorly but auto generator I guess

Kami: What he said :P

Kami:
Closing this since it's not relevant anymore.

Runners can now be disabled using an API and removing sudo support is something we need to revisit in the future.

PR #2673, opened 2016-05-06
[WIP] Allow administrators to disable a runner

Kami:
This pull request introduces a new API and corresponding CLI commands which allows administrator to disable a particular runner.

In addition to that, it includes other related changes and improvements:

1. It ports RunnerType API endpoint to utilize `ResourceController` class. This way this API endpoint behaves the same as other API endpoints (limit works, filtering works, etc.) and means we can get rid of the duplicated code.
2. Introduces RBAC support for `RunnerType` resource. This is need because only admins (or people with an explicit permission grant) are allowed to disable / enable a runner (previously we could get away with not having RunnerType API endpoint behind RBAC wall, not we can't anymore).
3. Fixes runner registration - if user has explicitly disabled a runner through the API, runner won't get re-enabled when running register content and registering runners (aka the expected behavior).

## Background

I initially went with the config approach, but in https://github.com/StackStorm/st2/pull/2671#discussion_r62194745, @manasdk pointed out that we already have a concept for disabling a particular runner (aka `enabled` attribute on the `RunnerTypeDB` model). This attribute was there, but it wasn't hooked up and working so I hooked it up.

The whole change did require more work compared to the config approach (rbac, etc.), but imo it's worth it (RBAC is something we would eventually need to get done anyway) and this approach is generic so it applies to every runner, not just local one.

TODO:

- [ ] Update BaseRunner `pre_run` method to throw and abort action execution if a runner is disabled. Also update all the affected runner classes (aka all classes which inherit from the base class).
- [ ] Tests for RBAC permissions resolver
- [ ] Only allow "enabled" attribute of the runner to be changed through the modify / PUT API endpoint
- [ ] Tests for a disabled runner

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,9 @@ in development
 * Fix for `data` is dropped if `message` is not present in notification. (bug-fix)
 * Remove now deprecated Fabric based remote runner and corresponding
   ``ssh_runner.use_paramiko_ssh_runner`` config option. (cleanup)
+* Fix support for password protected private key files in the remote runner. (bug-fix)
+* Allow user to provide a path to the private SSH key file for the remote runner ``private_key``
+  parameter. Previously only raw key material was supported. (improvement)
 
 1.4.0 - April 18, 2016
 ----------------------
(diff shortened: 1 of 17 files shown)


Kami:
Closing this PR. 

I will open a new one, since I missed this one up (I forgot to push some local changes to master so this diff is now accurate).

Issue #2675, opened 2016-05-06
Alias names should be scoped to pack


codyaray:
I've written some aliases for `github` (still testing, that's why I haven't pushed back yet). Since I'm hitting some issues, I decided to try with the JIRA pack to see if its pack-specific or maybe my setup. So I tried writing a `jira.get_issue` however the name `get_issue` apparently conflicts with the one from `github` that already exists. I would expect the alias name to be scoped to the pack.

Here's the aliases before I try to sync the jira rule:

```
$ st2 action-alias list
+----------------------+--------+---------------------------+---------+
| ref                  | pack   | description               | enabled |
+----------------------+--------+---------------------------+---------+
| github.add_comment   | github | Add a comment to the      | True    |
|                      |        | provided issue / pull     |         |
|                      |        | request.                  |         |
| github.add_status    | github | Add a commit status for a | True    |
|                      |        | provided ref.             |         |
| github.get_issue     | github | Retrieve information      | True    |
|                      |        | about a particular Github |         |
|                      |        | issue.                    |         |
| github.get_user      | github | Get a user from the       | True    |
|                      |        | Github user database      |         |
| github.list_issues   | github | Retrieve a list of issues | True    |
|                      |        | (including pull requests) |         |
|                      |        | for a particular          |         |
|                      |        | repository.               |         |
| github.list_teams    | github | List teams in             | True    |
|                      |        | organization              |         |
| packs.deploy_pack    | packs  | Download StackStorm packs | True    |
|                      |        | via ChatOps               |         |
| packs.pack_deploy    | packs  | Download StackStorm packs | True    |
|                      |        | via ChatOps               |         |
| packs.pack_info      | packs  | Get StackStorm pack       | True    |
|                      |        | information via ChatOps   |         |
| packs.show_git_clone | packs  | Show the Git clone        | True    |
|                      |        | command for a given pack  |         |
+----------------------+--------+---------------------------+---------+
```

I'm trying to add this alias for `jira.get_issue`:

```
---
  name: "get_issue2"
  pack: "jira"
  action_ref: "jira.get_issue"
  description: "Retrieve information about a particular JIRA issue."
  formats:
    - "jira issue get {{issue_key}}"
```

But when I try creating the alias, I get this error:

```
$ st2 action-alias create /opt/stackstorm/packs/jira/aliases/get_issue.yaml
ERROR: 409 Client Error: Conflict
[...]

PR #2677, opened 2016-05-11
[RFR] actionalias.pack unique with name

manasdk:
Should fix https://github.com/StackStorm/st2/issues/2675

### todo
- [x] remove old index on name
- [x] add unit test
- [x] Make Kami happy


manasdk:
@lakshmi-kannan @Kami Thoughts? Specifically on the migration script.

Kami:
One alternative would be to remove old / incorrect indexes in the `ensure_db_indexes` method where we ensure all the necessary indexes exist.

In fact, that was the alternative I proposed in the past when we had a similar situation (I believe there should still be a PR with prototype code somewhere), but we decided not to merge it since it was kinda complicated.

It might be worth re-considering that approach again, but that approach also has some disadvantages (in case user has created a custom index, it will be dropped in they don't white-list it).

In any case, I'm not really happy with so many hacky one-time scripts we have in `tools/` so it's probably also a good time to consider investing some time in a proper data / schema migration framework.

manasdk:
> One alternative would be to remove old / incorrect indexes in the ensure_db_indexes method where we ensure all the necessary indexes exist.

I am glad you brought that up as it was an option that I dismissed very quickly assuming I would get heat from you guys :)

The downside to doing so in `ensure_db_indexes` is that this is a migration i.e. needs to be applied once while ensure_db_indexes runs everytime a service starts etc. Actually, I am not very clear any longer where all the times this method runs. Also, often I've found it better to keep version specific migration out of the main code-base and have a capability of running it independently. 

Having said that we don't have any rational thought behind migrations of schema (schema-less n all) as well as indexes so perhaps it is worth exploring that at some point rather than making ad-hoc changes.

Kami:
Yes, it runs on every service startup, but that addition shouldn't add a lot of overhead (in case there are no indexes to remove, the part which rectifies the indexes and deletes the removed ones simply wouldn't do anything).

The problem I'm having with ad-hoc stuff is that, because we have no other mechanism right now we need to (probably?) run this inside the package post-inst or let user know to manually running it. In addition to that, we don't have schema and migration versioning which is also not ideal.

manasdk:
Ok. As a middle ground I am willing to move this to ensure_db_indexes with the caveat that we really need to figure this stuff out fast before it gets entirely out of hand. Agreeable?

Kami:
@manasdk I'm fine with that as long as we document that behavior and we add a whitelist option to the config (this way, in case user creates custom indexes, they can whitelist them so they don't get removed on rectification).

Also, IIRC, the code for that was a bit convoluted since I needed to use pymongo directly to retrieve existing indexes, etc. and IIRC, there were also some other caveats.

manasdk:
> as long as we document that behavior and we add a whitelist option to the config

Isn't this rather arbitrary?

> Also, IIRC, the code for that was a bit convoluted since I needed to use pymongo directly to retrieve existing indexes, etc. and IIRC, there were also some other caveats.

Yeah I got that part figured out already. Using pymongo.

Kami:
@manasdk I think it's a good idea to add that and document it. Otherwise, if user creates a custom index (for whatever reason they might have), it will be removed on service startup and that's undesired from the user perspective so we need a way to allow user to whitelist those custom indexes.

manasdk:
@Kami Can't help but think that the migration script which is controllable would be a better alternative. Anyway, adding this code will push the work out to early next week since I am off to dealing with some other tasks. I will do this but only because you think it is a good idea :P

manasdk:
@kami I have decided to not care about custom indexes. It should be possible to fix that part later if the need arises.

st2common/st2common/models/db/__init__.py
```diff
@@ -97,9 +99,28 @@ def db_ensure_indexes():
...
+    """
+    extra_indexes = model_class.compare_indexes().get('extra', None)
+    if not extra_indexes:
+        return
+    # mongoengine does not have the necessary method so we need to drop to
+    # pymongo interfaces via some private methods.
+    c = model_class._get_collection()
+    for extra_index in extra_indexes:
+        try:
+            c.drop_index(extra_index)
```
Kami: Probably a good idea to log (under debug) which indexes we remove...

st2common/st2common/models/db/__init__.py
```diff
@@ -97,9 +99,28 @@ def db_ensure_indexes():
...
+        cleanup_extra_indexes(model_class=model_class)
+        LOG.debug('Ensuring indexes for model "%s"...' % (model_class.__name__))
+        model_class.ensure_indexes()
+
+
+def cleanup_extra_indexes(model_class):
+    """
+    Finds any extra indexes and removes those from mongodb.
+    """
+    extra_indexes = model_class.compare_indexes().get('extra', None)
```
Kami: Nice - when I was working on that I didn't know this `extra` attribute is available (my approach was a lot more complicated).

Is this new mongoengine functionality or just something I have missed?

Kami:
Nice :+1:

st2common/st2common/models/db/__init__.py
manasdk: Will fix in a separate PR.

st2common/st2common/models/db/__init__.py
manasdk: Not sure. I was surprised that this exact method even existed but am glad it does.

PR #2679, opened 2016-05-12
Add "get_fixture_content" method to the base pack resource test classes

Kami:
This PR adds "get_fixture_content" method to the base pack resource test classes. This method enforces fixture file location conventions and allows user to load raw fixture content from a fixture file on disk.

We discussed this on a st2contrib PR recently - the functionality is very simple, but we still think there is value in it since it enforces fixture file location convention (`<pack path>/tests/fixtures/`).

### DIFF
--- a/st2tests/st2tests/pack_resource.py
+++ b/st2tests/st2tests/pack_resource.py
@@ -1,0 +1,52 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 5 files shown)


jjm:
This will be really useful, I'm finishing up a PR for st2contrib which I'm happy to up date to use these helper functions prior to it being merged. 

Thanks @Kami!

lakshmi-kannan:
👍 , I like the convention as well, personally.

st2tests/st2tests/pack_resource.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class BasePackResourceTestCase(TestCase):
+    """
+    Base test class for all the pack resource test classes.
+
+    Contains some utility methods for loading fixtures from disk, etc.
+    """
+
+    def get_fixture_content(self, fixture_path):
```
lakshmi-kannan: I don't think we should allow nesting. Better to support fixture_file_name. Wdyt?

Kami: I think it's fine if user wants to store fixture in a sub-directory (e.g. `tests/fixtures/api_responses/my_fixture.json` or similar)

PR #2683, opened 2016-05-13
Pack config schema support

Kami:
This is a first iteration of the improvement pack configuration story.

It includes support for optional pack config schema. Pack configuration which will be supplied by the user (that's the second iteration of the improved pack configuration story and is TBD) will be validated against the schema provided by the pack author when loading pack configuration.

Pack schema follows action parameters format and the schema file looks like that:

```yaml
---
  attributes:
    api_key:
      type: "string"
     secret: true
     required: true
    api_secret:
      type: "string"
     secret: true
     required: true
    region:
      type: "string"
      required: false
```

(I'm open to feedback, we could rename `attributes` to something else, we could also get rid of it and have config items as top-level object, but I'm not too big of a fan of that approach.)

Schema is registered (and corresponding `ConfigSchemaDB` object is created in the database) when registering packs when running st2-register-content script.

This pull request also includes API endpoint for retrieving config schema (read-only).

## Open Questions / To Decide

- [x] Config schema file name - right now it's `config.schema.yaml`, but we could also call it `config.meta.yaml` or similar. I'm open to feedback.
- [x] API endpoint path - `/v1/packs/config_schema/<pack ref>` vs `/v1/packs/views/config_schema/<pack ref>` (ideally, it would be `/packs/<pack ref>/config_schema`, but because of our pecan abuse that's not possible right now and something we should do once we move to v2 and clean up our API).

## TODO

- [ ] Documentation
  - [ ] Basic docs on config schema format
  - [ ] Config best practices (prefer flat, avoid deeply nested objects, etc.)
- [x] Update affected `packs.uninstall` command
- [x] Tests with config schema examples fixtures
  - [x] Basic tests (registration, API retrieval)
  - [x] RBAC tests

### DIFF
--- a/st2api/st2api/controllers/resource.py
+++ b/st2api/st2api/controllers/resource.py
@@ -204,6 +204,22 @@ class ResourceController(rest.RestController):
 
         return result
 
+    def _get_one_by_pack_ref(self, pack_ref, exclude_fields=None, from_model_kwargs=None):
+        LOG.info('GET %s with pack_ref=%s', pecan.request.path, pack_ref)
+
+        instance = self._get_by_pack_ref(pack_ref=pack_ref, exclude_fields=exclude_fields)
+
+        if not instance:
+            msg = 'Unable to identify resource with pack_ref "%s".' % (pack_ref)
+            pecan.abort(http_client.NOT_FOUND, msg)
+
+        from_model_kwargs = from_model_kwargs or {}
+        from_model_kwargs.update(self._get_from_model_kwargs_for_request(request=pecan.request))
+        result = self.model.from_model(instance, **from_model_kwargs)
+        LOG.debug('GET %s with pack_ref=%s, client_result=%s', pecan.request.path, id, result)
+
+        return result
+
     def _get_by_id(self, resource_id, exclude_fields=None):
         try:
             resource_db = self.access.get(id=resource_id, exclude_fields=exclude_fields)
@@ -216,6 +232,15 @@ class ResourceController(rest.RestController):
         try:
             resource_db = self.access.get(name=resource_name, exclude_fields=exclude_fields)
         except Exception:
+            resource_db = None
+
+        return resource_db
+
+    def _get_by_pack_ref(self, pack_ref, exclude_fields=None):
+        try:
+            resource_db = self.access.get(pack=pack_ref, exclude_fields=exclude_fields)
+        except Exception as e:
+            print e
             resource_db = None
 
         return resource_db
--- a/st2api/st2api/controllers/resource.py
+++ b/st2api/st2api/controllers/resource.py
@@ -240,8 +240,11 @@ class ResourceController(rest.RestController):
         try:
             resource_db = self.access.get(pack=pack_ref, exclude_fields=exclude_fields)
         except Exception as e:
-            print e
             resource_db = None
+
+        if not resource_db:
+            msg = 'Resource with a pack_ref "%s" not found' % (pack_ref)
+            raise StackStormDBObjectNotFoundError(msg)
 
         return resource_db
 
--- a/st2api/st2api/controllers/resource.py
+++ b/st2api/st2api/controllers/resource.py
@@ -59,6 +59,8 @@ class ResourceController(rest.RestController):
     filter_transform_functions = {}
 
     # Method responsible for retrieving an instance of the corresponding model DB object
+    # Note: This method should throw StackStormDBObjectNotFoundError if the corresponding DB
+    # object doesn't exist
     get_one_db_method = None
 
     def __init__(self):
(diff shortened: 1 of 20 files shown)


lakshmi-kannan:
>Config schema file name - right now it's config.schema.yaml, but we could also call it config.meta.yaml or similar. I'm open to feedback.

I prefer config.schema.yaml

> /v1/packs/config_schema/<pack ref> vs /v1/packs/views/config_schema/<pack ref> (ideally, it would be /packs/<pack ref>/config_schema

I ran into similar problems with pecan when I tried to do this scoped key value controller. If 
/packs/<pack ref>/config_schema is not possible, I propose moving config_schema to own URL base. So /v1/config_schema/<pack_ref>. That is better than  /v1/packs/config_schema/<pack ref> but I am not sure if it's better than  /v1/packs/views/config_schema/<pack ref>. Views controller is there to provide alternate views into the same data. So putting schema there doesn't seem right.

lakshmi-kannan:
BTW I already introduced a v2 API for key value. https://github.com/StackStorm/st2/pull/2669/files#diff-8beb8eec6c835f872e21cecf1a18e620R1
 It's a giant hack because of pecan limitations. FYI.

Kami:
I can do that (separate top-level API endpoint), but I'm not too big of a fan of it.

And yeah, @manasdk and I discussed this earlier on Slack. I encountered this issue many times before and sadly because of the way we (ab)use pecan right now, we can't achieve using that URL path structure without adding a lot more hacks on top of the existing hacks...

Kami:
> BTW I already introduced a v2 API for key value. https://github.com/StackStorm/st2/pull/2669/files#diff-8beb8eec6c835f872e21cecf1a18e620R1
It's a giant hack because of pecan limitations. FYI.

Can't we put this under /v1 or something? I'd really like us to invest some time in a proper RESTful API and ideally that would be API v2 (I don't want us to deprecate 3 different API versions if that's not needed).

lakshmi-kannan:
/v1/ already supports scope via query parameter and @manasdk insisted on a v2 API to do /keys/${scope}/key. We discussed this on slack and I spent hours on adding v2 api and tests. I can remove it but it seems like we have to get on the same page.

Kami:
I've moved API endpoint to be top-level (`/v1/config_schema/<pack ref>`).

I'm really not fan of it and I hate our API paths, but we plan to have a discussion about API v2 today where we plan to improve the API paths, etc.

Kami:
I decided I will work on the documentation in another PR where I will introduce support for dynamic configs.

Schema is tightly related to that feature so docs will make more sense where both features are available.

st2api/st2api/controllers/resource.py
```diff
@@ -220,6 +238,18 @@ def _get_by_name(self, resource_name, exclude_fields=None):
 
         return resource_db
 
+    def _get_by_pack_ref(self, pack_ref, exclude_fields=None):
+        try:
+            resource_db = self.access.get(pack=pack_ref, exclude_fields=exclude_fields)
+        except Exception:
+            resource_db = None
+
+        if not resource_db:
```
lakshmi-kannan: It's a little silly to not do this in the except block and using a variable to capture and then raise. Any particular reasons?

st2api/st2api/controllers/resource.py
```diff
@@ -204,6 +206,22 @@ def _get_one_by_name_or_id(self, name_or_id, exclude_fields=None, from_model_kwa
 
         return result
 
+    def _get_one_by_pack_ref(self, pack_ref, exclude_fields=None, from_model_kwargs=None):
+        LOG.info('GET %s with pack_ref=%s', pecan.request.path, pack_ref)
+
+        instance = self._get_by_pack_ref(pack_ref=pack_ref, exclude_fields=exclude_fields)
```
lakshmi-kannan: This can throw StackStormDBObjectNotFoundError. I am confused by the `if not instance` and abort. Actually I am confused more by why there is both _get_one_by_pack_ref and _get_by_pack_ref. Seems like there is a good scope for abuse.

st2api/tests/unit/controllers/v1/test_pack_config_schema.py
```diff
@@ -0,0 +1,35 @@
...
+
+class PackConfigSchemaControllerTestCase(FunctionalTest):
+    register_packs = True
+
+    def test_get_all(self):
+        resp = self.app.get('/v1/config_schema')
+        self.assertEqual(resp.status_int, 200)
+        self.assertEqual(len(resp.json), 1, '/v1/config_schema did not return all schemas.')
+
+    def test_get_one(self):
```
lakshmi-kannan: Another test for 404 :P

st2api/st2api/controllers/resource.py
Kami: It's just an existing code pattern I followed (yeah, I could just move throw inside the except block).

st2api/st2api/controllers/resource.py
Kami: I will change / fix this code, convention we tend to follow is that `_get_by_` returns None and doesn't throw.

Kami: There we go - 6adffc20c1c2b586f2f44f8ffcd52f69147ae9a8

PR #2684, opened 2016-05-17
Update the base model code to throw StackStormDBObjectNotFoundError exception if an object is not found, update affected code

Kami:
I noticed this while working on some related changes.

Right now we throw `ValueError` instead of `StackStormDBObjectNotFoundError` inside base model `.get` method when object is not found. This means that in many places we need to capture this exception and then re-throw `StackStormDBObjectNotFoundError`.

This can be solved by throwing `StackStormDBObjectNotFoundError` directly. In comparison to `ValueError`, this exception is also more specific.

If the `ValueError` was selected intentionally (e.g. some other code path also throws ValueError and we want to capture all those exceptions), please let me know.


lakshmi-kannan:
I am ok with this change. LGTM.

Kami:
@manasdk any objections? do you happen to recall if we had some specific reason to utilize `ValueError` in the past?

Kami:
I contained everything in a single commit so it's easy to revert it if needed. I will go ahead and merge it.

PR #2685, opened 2016-05-17
Fix delay caused by loading a huge set of actions

enykeev:
Timing graph for GET https://192.168.10.16/api/v1/actions API request that returns around 1600 actions looks like that:

Applying throttling to it (Good 3G preset, 40ms latency, 1.5Mbps in, 750Kbps out) to get the picture better representing real world use:

In general, TTFB can be round down to the time server takes to prepare the answer (contact db, query data, apply transforms) and `content download` is obviously the time it takes for browser to pull this data from server. Browsers can't parse JSON from a stream so the time to first pixel rendered is always bigger than that.

By excluding fields like `parameters` and `notify`, you can reduce it to 

Personally, I'd rather prefer to use `.only()` mongo filter instead of `.exclude()` since it would help us maintain certain performance level in the long run without constantly catching up on the client side, but I guess we can solve it in a separate PR. 

Further optimization would require us to use some form of caching or pagination.

Caching is relatively easy and will only require us to track the last time we've changed an action (in simplest case, registered them) and then return proper http code. Downside is that it will still require user to download the list at least once.

Pagination on the other hand is a somewhat permanent and consistent solution, but it implies quite a lot of work both client and server side (besides the obvious "pagination" stuff, the filtering would have to be re-implemented server side).

### DIFF
--- a/st2api/st2api/controllers/v1/actions.py
+++ b/st2api/st2api/controllers/v1/actions.py
@@ -78,8 +78,13 @@ class ActionsController(resource.ContentPackResourceController):
 
     @request_user_has_permission(permission_type=PermissionType.ACTION_LIST)
     @jsexpose()
-    def get_all(self, **kwargs):
-        return super(ActionsController, self)._get_all(**kwargs)
+    def get_all(self, exclude_attributes=None, **kwargs):
+        if exclude_attributes:
+            exclude_fields = exclude_attributes.split(',')
+        else:
+            exclude_fields = None
+
+        return super(ActionsController, self)._get_all(exclude_fields=exclude_fields, **kwargs)
 
     @request_user_has_resource_db_permission(permission_type=PermissionType.ACTION_VIEW)
     @jsexpose(arg_types=[str])


Kami:
I'm fine with that change, but I do agree that some kind of pagination would be better.

st2api/st2api/controllers/v1/actions.py
```diff
@@ -78,8 +78,13 @@ def __init__(self, *args, **kwargs):
 
     @request_user_has_permission(permission_type=PermissionType.ACTION_LIST)
     @jsexpose()
-    def get_all(self, **kwargs):
-        return super(ActionsController, self)._get_all(**kwargs)
+    def get_all(self, exclude_attributes=None, **kwargs):
+        if exclude_attributes:
```
Kami: In action executions controller we validate that a user provided a valid exclude attribute (`_validate_exclude_fields`) so probably not a bad idea to refactor some of that functionality on a base controller and also do that here (and introduce a whitelist of attributes which can be excluded).

enykeev: I'd rather prefer us to choose the list of fields we need instead of the one we exclude. Reason being, client developer knows for sure what kind of data he wants to show. Any field we add to the model on top of that will make response bigger and will require client developer to add it to exclusion list. By using `.only()` instead of `.exclude()` we can ensure performance won't degrade over time. Saying that, yeah, the refactoring is needed.

Kami: Yeah, I agree, but doing that would require bigger code changes.

dzimine:
with product owner hat on: this solves the immediate problem: with your latency assumptions, under 3 sec is acceptable. Let's do it.
off this hat: `only()` looks more convenient for a client, indeed.

st2api/st2api/controllers/v1/actions.py
lakshmi-kannan: I like .only() over .exclude() but when we expose this to an user not from stackstorm team, neither approach would make sense without knowing the schema for the object itself which needs to be solved in first place.

Kami:
I'll go ahead and merge this since it's a good change and add validation in a separate PR.

I think validation is good idea, since AFAIK, if we don't do any validation and user passes in unsupported attribute it will result in internal server error which is not ideal.

Kami:
There we go - 55a9c49aea05bbfe23153325d7762a98753f45db, db4603b3add3d9f6acf18a37259f205bc2005418.

PR #2687, opened 2016-05-17
Register all RMQ exchanges

manasdk:
This used to work because whenever a Q is created on an exchange it ends up being registered if the exchange is not already present. However, always better to explicitly register these exchanges up-front.


m4dcoder:
+1

Kami:
Good catch :+1: 

A couple of days ago I fixed a similar issue for models (we didn't update a variable which holds a list of all the model modules when adding a new model a while back).

Wish there was a less error prone way of doing that.

PR #2688, opened 2016-05-18
Revert #2625

emedvedev:
This PR reverts #2625, since RHEL6 isn’t compatible. We still need a solution of the original problem (services that aren’t installed showing up as errors on st2ctl commands), but currently I don’t have a better solution than checking exit codes, and this is a blocker, so we should revert now and worry about error messages later.

I’ll wait for @DoriftoShoes before merging, hopefully he’ll have some ideas. :)

### DIFF
--- a/st2common/bin/st2ctl
+++ b/st2common/bin/st2ctl
@@ -32,7 +32,7 @@ function print_usage() {
     echo "  --register-aliases            Register all aliases."
     echo "  --register-policies           Register all policies."
     echo "  --register-setup-virtualenvs  Create Python virtual environments for all the registered packs."
-    echo "  --verbose                     Output additional debug and informational messages."
+    echo "  --verbose           Output additional debug and informational messages."
     echo ""
     echo "Most commands require elevated privileges."
 }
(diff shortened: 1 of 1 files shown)


DoriftoShoes:
What is the failure on EL6?  It seems like `service` should work fine.  Either way, I agree that we should merge this PR then look for a better solution to the earlier problem.

manasdk:
@DoriftoShoes A combination of https://github.com/StackStorm/st2-packages/blob/master/packages/st2/bin/runners.sh#L48 and https://github.com/StackStorm/st2/blob/master/st2common/bin/st2ctl#L66
mean that on RHEL6/centos66 exit code is 1 and actionrunners are not started.

Anyway the assumption that exit_code will be 1 when service is not present isn't inaccurate since the exit code is as per the init script.

manasdk:
+1

emedvedev:
@manasdk: should we also merge this into 1.4? What would be best?

manasdk:
@emedvedev v1.4 looks fine https://github.com/StackStorm/st2/blob/v1.4/st2common/bin/st2ctl#L63

Issue #2689, opened 2016-05-18
Missing dependency pytz in requirements.txt


ericcitaire:
Here's the trace after a fresh install of `st2client` using `pip` :

```python
Traceback (most recent call last):
  File "/usr/local/bin/st2", line 9, in <module>
    load_entry_point('st2client==1.4.0.8', 'console_scripts', 'st2')()
  File "/usr/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 542, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2569, in load_entry_point
    return ep.load()
  File "/usr/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2229, in load
    return self.resolve()
  File "/usr/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2235, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/local/lib/python2.7/site-packages/st2client/shell.py", line 39, in <module>
    from st2client.commands import action
  File "/usr/local/lib/python2.7/site-packages/st2client/commands/action.py", line 36, in <module>
    from st2client.utils.date import format_isodate_for_user_timezone
  File "/usr/local/lib/python2.7/site-packages/st2client/utils/date.py", line 18, in <module>
    from pytz import timezone as TZ
ImportError: No module named pytz
```

`pytz` should be added to `dependencies.txt`.

Kami:
Thanks for the report.

I believe one of our dependencies is depending on `pytz` so we in-directly depended on `pytz`. It's possible though that this dependency is used by some other component of st2 and not st2client. In this case it would break if you manually install `st2client` from PyPi or similar.

I will dig in and update `st2client` requirements.txt if needed.

Kami:
Fix in #2690 - good catch & thanks!

Kami:
@ericcitaire I've merged PR into master.

The issue will be resolved once we publish new client to PyPi (this will happen when we release StackStorm v1.4.1 or v1.5.0).

ericcitaire:
Thanks guys! That was fast!! :)

manasdk:
Fixed.

PR #2690, opened 2016-05-18
Add missing pytz dependency to st2client

Kami:
Originally reported in #2689 (thanks!).

We never encountered this issue (and tests didn't catch it) since when using package installation method, `st2client` uses the same virtual environment as other components and `appscheduler` (used by `st2reactor`) depends on `pytz` so the dependency was also satisfied for `st2client`.

The issue would only manifest itself if user manually installed `st2client` outside st2 virtual environment.


manasdk:
Good catch!

Also, the same value needs to be in 3 places - something is wrong here isn't it?

Kami:
It's a "feature" (and one of them is auto-generated).

Top-level requirements.txt is there for backward compatibility reasons, ideally it would only require one manual change in `st2client/in-requirements.txt` and other requirements.txt files would be auto-generated (that's true for top-level one, but not requirements.txt in st2client directory).

PR #2691, opened 2016-05-18
Drop support for JSON format for resource metadata files

Kami:
We have introduced support for YAML in StackStorm v0.6 and about the same time we have also deprecated JSON support.

It's about time to drop support for JSON.

I will also open corresponding st2docs PR (upgrade notes) and double check st2contrib and this repo that we don't have any JSON files anymore.


manasdk:
> and about the same time we have also deprecated JSON support.

I thought we decided to support both. I am not sure why this even matters?

Btw, as of now since there is no way to backup rules.

`st2 rule get rule-ref -j > rule.json` and `st2 rule update rule-ref rule.json`

manasdk:
I am -1

Kami:
@manasdk I should have made it more clear (I thought there will be some confusion about that).

CLI still supports JSON, but for resource metadata files contained in the pack we now only support YAML.

Also as far as backing up goes, that's unrelated to that change, but for true infrastructure as code approach, people should backup their content directory and not what is returned by the API.

manasdk:
> people should backup their content directory and not what is returned by the API.

since we support rule creation from the API often the only way to back up content is by using the CLI to send to a file and then save that file in a content folder which could then be version controller.

lakshmi-kannan:
Summarizing @DoriftoShoes and my opinions here:

st2 foo get --yaml or st2 foo get -y support is critical to deprecate JSON support. Otherwise, we won't have a path to dump a rule in YAML after creating it from UI. This is already a point of friction. So let's wait until we do that to merge.

Kami:
I've added support for formatting CLI output as YAML in 164c049f38f6343944115245d0b742d3c9c9fd30.

Issue #2694, opened 2016-05-18
Inconsistency in self._sensor_service and self.action_service


lakshmi-kannan:

Kami:
Yep, there is a same thing with config (`self.config` and `self._config` - it's inconsistency between actions and sensors).

Since they are "public attributes" we should get rid of `_` for consistency (well, we should leave it there as a @property for backward  compatibility, but that's it).

What do others think?

Issue #2693, opened 2016-05-18
ST2_AUTH_TOKEN env variable set in sensor container but not in action leading to issues


lakshmi-kannan:
User was following this doc to access key value items instead of datastore services but it looks like we don't set the ST2_AUTH_TOKEN env variable in action runners. This means the same code works in sensor but fails in action complaining about auth. 

```
lkannan@HQ2-LC02QV2NDFVH8 /Volumes/workspace/src/storm/st2 (master●)$ ag "AUTH_TOKEN_ENV_VARIABLE_NAME"       [ruby-2.0.0p648]
st2common/st2common/constants/system.py
22:    'AUTH_TOKEN_ENV_VARIABLE_NAME',
29:AUTH_TOKEN_ENV_VARIABLE_NAME = 'ST2_AUTH_TOKEN'

st2reactor/st2reactor/container/process_container.py
30:from st2common.constants.system import AUTH_TOKEN_ENV_VARIABLE_NAME
297:        env[AUTH_TOKEN_ENV_VARIABLE_NAME] = temporary_token.token
lkannan@HQ2-LC02QV2NDFVH8 /Volumes/workspace/src/storm/st2 (master●)$
```
Also related: https://github.com/StackStorm/st2docs/issues/156

Kami:
We do set it for actions, it's called `ST2_ACTION_AUTH_TOKEN`. All the public variables StackStorm sets and which are available to actions are prefixed with `ST2_ACTION_` (that's documented in actions section of the documentation).

The `ST2_AUTH_TOKEN` variable you are talking about is actually a "private" variable set for the st2client so SensorWrapper can access the datastore via API.

(And yes, I'm not a fan of environment variable approach either, but the reason we do that is because st2client API is less than ideal - client constructor doesn't accept the token).

Kami:
Actually, please ignore my comment above (I need to stop responding to comments before I have my coffee :P)

Looks like the issue you are talking about is accessing the datastore via the new `action_service` and not directly. In that case, we need to set that environment variable, yeah (or ideally, make client work with a token which is provided via constructor, but this would require bigger client refactorings).

On a related note - we also need some integration tests for that.

PR #2695, opened 2016-05-19
Set same env variables as sensor process and s/_sensor_service/sensor_service

lakshmi-kannan:
Fixes https://github.com/StackStorm/st2/issues/2694 and https://github.com/StackStorm/st2/issues/2693

Need feedback on https://github.com/StackStorm/st2docs/issues/156

### DIFF
--- a/st2actions/st2actions/runners/__init__.py
+++ b/st2actions/st2actions/runners/__init__.py
@@ -21,6 +21,8 @@ import six
 from st2actions import handlers
 from st2common import log as logging
 from st2common.constants.pack import DEFAULT_PACK_NAME
+from st2common.constants.system import API_URL_ENV_VARIABLE_NAME
+from st2common.constants.system import AUTH_TOKEN_ENV_VARIABLE_NAME
 from st2common.exceptions.actionrunner import ActionRunnerCreateError
 from st2common.util import action_db as action_utils
 from st2common.util.api import get_full_public_api_url
@@ -136,9 +138,13 @@ class ActionRunner(object):
         result['ST2_ACTION_PACK_NAME'] = self.get_pack_name()
         result['ST2_ACTION_EXECUTION_ID'] = str(self.execution_id)
         result['ST2_ACTION_API_URL'] = get_full_public_api_url()
+        # Compatibility with sensors
+        result[API_URL_ENV_VARIABLE_NAME] = get_full_public_api_url()
 
         if self.auth_token:
             result['ST2_ACTION_AUTH_TOKEN'] = self.auth_token.token
+            # Compatibility with sensors
+            result[AUTH_TOKEN_ENV_VARIABLE_NAME] = self.auth_token.token
 
         return result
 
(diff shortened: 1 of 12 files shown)


lakshmi-kannan:
Docs: https://github.com/StackStorm/st2docs/pull/157

st2actions/st2actions/runners/__init__.py
```diff
@@ -136,9 +138,13 @@ def _get_common_action_env_variables(self):
         result['ST2_ACTION_PACK_NAME'] = self.get_pack_name()
```
Kami: Let's please move this to a different place.

As I noted here https://github.com/StackStorm/st2/issues/2693#issuecomment-220262497, there are two different things:

1. Environment variables which are set for st2client to work (as mentioned, that's a hack, I don't like it but that's limitation of st2client right now)
2. Environment variables which are set as a convenience and available to actions.

In this case we are fixing 1), but you are doing in in code for 2).

st2reactor/st2reactor/sensor/base.py
```diff
@@ -23,8 +23,10 @@ def __init__(self, sensor_service, config=None):
         :keyword config: Sensor config.
         :type config: ``dict`` or None
         """
-        self._sensor_service = sensor_service
-        self._config = config or {}
+        self._sensor_service = sensor_service  # Deprecate in the future
+        self.sensor_service = sensor_service
+        self._config = config or {}  # Deprecate in the future
+        self.config = self._config
```
Kami: I originally missed this PR (I blame thunderbird sync!) and that's how I've done it utilizing @property - https://github.com/StackStorm/st2/pull/2698/files#diff-f120a0125bcd3ec86f1d85020ed530ffR26

Kami:
Let's please get this addressed - https://github.com/StackStorm/st2/pull/2695#discussion_r63854788

Besides that, LGTM, :+1:

Kami:
As mentioned in #2699 (yes, that PR I opened because I missed this one), it would also be good to add some (integration / end to end) tests which test datastore operations to st2tests or similar.

The unit tests which we have right now only test implementation details - they test correct environment variables are set. What we really are interested in and should be testing is "do datastore related methods" work. For that, we need end to end tests.

Kami:
Minor comment wrt to PR title and description - imo, it's better if the PR title includes a more high-level description of what is being fixed / changed. This makes it easier for anyone checking the PR to understand what is going on and what is being changed.

Right now it includes implementation detail ("setting same environment variables"and  by itself this doesn't tell much). Imo, a better title would be something along the lines of "Fix datastore access in the Python runner actions" or something along those lines since that what we are doing - we are fixing datastore access for Python runner actions (and "setting same environment variables" is just an implementation detail of the fix which can be included in the description).

lakshmi-kannan:
As per slack discussion, here is an integration test for datastore access in action - https://github.com/StackStorm/st2tests/pull/47

Kami:
Thanks for moving the code and adding a test :+1:

PR #2697, opened 2016-05-19
Add basic autocomplete support to st2 CLI client

enykeev:
To activate autocomplete, perform `eval "$(register-python-argcomplete st2client/st2client/shell.py)"` or `eval "$(register-python-argcomplete st2)"` in your shell (depending on how you're planning to call it).

It will become a part packaging process when we figure out we're happy with autocomplete in general.

Please note that I've added +x and hashbang to shell.py. You may have your own opinion on the matter.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 # Don't edit this file. It's generated automatically!
 apscheduler<3.1,>=3.0.5
+argcomplete
 bcrypt
 bencode<1.1,>=1.0
 eventlet<0.19,>=0.18.4
--- a/st2client/in-requirements.txt
+++ b/st2client/in-requirements.txt
@@ -1,4 +1,5 @@
 # Remeber to list implicit packages here, otherwise version won't be fixated!
+argcomplete
 prettytable
 pytz
 python-dateutil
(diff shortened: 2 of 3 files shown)


Kami:
Sweet, that's something I wanted for a while now.

And yeah, ideally we would also ship a bash script which activates the auto-completion. I think this script should be part of this repository (st2client) and as part of st2client and package installation we copy it to the correct place.

manasdk:
Awesome! This will be life changing for me :)

lakshmi-kannan:
I think bash script is the way to go when we ship with packages. I am not sure if this eval thingy would work?

lakshmi-kannan:
I mean, I will take this instead of nothing :). So +1 to merge from me.

Kami:
It's slightly slow, but I've confirmed it's working fine :+1: 

I'm fine with merging this now and submitting another PR which includes activate script + documentation in stdocs.

Kami:
Also, it looks like it's quite easy to register custom complete functions so it would be awesome if we could eventually also introduce support for auto-completing action names for `st2 run` and `st2 action execute`.

It shouldn't be too hard to do, we would just need to cache action list otherwise it would probably be too slow.

manasdk:
Slow but works!

+1

manasdk:
Next step to get to tab complete on content :D

PR #2699, opened 2016-05-19
Fix action service datastore access

Kami:
This fixes issue described in #2693.

I've added a simple unit tests, but ideally we would also have some integration tests (a Python runner action which uses operations which manipulate a datastore, probably a good candidate for st2tests).


Kami:
Looks like there already is #2695, closing.

PR #2700, opened 2016-05-19
Adding linux.sleep_until action

codyaray:
## Linux Pack

Added a new action that just sleeps until a specified time using the Linux `sleep` and `date` commands.

### Status 

- existing pack, new action

### Checklist (tick everything that applies)

- [X] Metadata: pack.yaml, icon, structure (required for new packs)
- [ ] Version bump (required for changed packs)
- [ ] Code linting (required, can be done after the PR checks)
- [ ] [Tests](https://docs.stackstorm.com/development/pack_testing.html) (not required but really recommended)

### DIFF
--- a/contrib/packs/linux/actions/sleep_until.yaml
+++ b/contrib/packs/linux/actions/sleep_until.yaml
@@ -10,7 +10,7 @@ entry_point: ""
 parameters:
   time:
     type: string
-    description: 'the time to wake up; must be parsable by "date -d"'
+    description: 'the time to wake up; must be smaller than the action timeout; must be parsable by "date -d" like "5pm today" or "Thu May 19 18:04:39 UTC 2016"'
     required: true
   cmd:
     default: 'sleep $(expr `date -d "{{ time }}" +%s` - `date -d "now" +%s`)'
--- a/contrib/packs/linux/actions/sleep_until.yaml
+++ b/contrib/packs/linux/actions/sleep_until.yaml
@@ -1,0 +1,17 @@
+---
+name: "sleep_until"
+pack: "linux"
+description: "Sleeps until a specified time"
+enabled: true
+
+runner_type: "local-shell-cmd"
+entry_point: ""
+
+parameters:
+  time:
+    type: string
+    description: 'the time to wake up; must be parsable by "date -d"'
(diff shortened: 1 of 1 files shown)


contrib/packs/linux/actions/sleep_until.yaml
```diff
@@ -0,0 +1,17 @@
...
+description: "Sleeps until a specified time"
+enabled: true
+
+runner_type: "local-shell-cmd"
+entry_point: ""
+
+parameters:
+  time:
+    type: string
+    description: 'the time to wake up; must be parsable by "date -d"'
```
manasdk: Can you provide a sample? Like "Thu May 19 18:04:39 UTC 2016"

Kami:
Thanks!

Probably also a good idea to add a note that sleep must be smaller than the action timeout, otherwise the action will time out (or potentially increasing default timeout or similar).

lakshmi-kannan:
At some point, I'll wish this were a python action.

Kami:
I honestly don't think it would make a difference if it's a Python runner or a local runner action (well, at least Python runner one would be easier to test). Only thing which would make a difference would be if this was a platform feature - which we could argue, we already have - timer triggers.

Also, as discussed with @codyaray on Slack, long sleep intervals have some caveats and depending on the use-case / workflow, event-driven approach might be better (in this case event-driven meaning emitting a trigger on a particular time, instead of having a long sleep inside a workflow).

codyaray:
Hm. If I were thinking about this as a platform feature, I actually think it would look somewhat different.

What I _really_ want would be to have a single logical workflow that can have triggers to "resume" in the middle of a workflow. So I can still see the big picture with all the moving pieces (relative to breaking it up into multiple sub-workflows), but I don't have to worry about a long sleep which can be potentially disrupted (relative to having one master workflow with sleeps).

lakshmi-kannan:
>I honestly don't think it would make a difference if it's a Python runner or a local runner action (well, at least Python runner one would be easier to test).

I was looking for something more testable indeed.

codyaray:
I haven't forgotten about this PR, guys. Just haven't gotten back to it yet. :)

Issue #2682, opened 2016-05-12
Process the triggers folder like actions and sensors


codyaray:
Follow-up issue to StackStorm/st2contrib#475.

We should support registering triggers which don't belong to the any sensor. Creating webhooks means creating standalone triggers. The most obvious convention would be to process YAML files in each pack's `<pack name>/triggers/` directory like we do with sensors, actions, rules, etc.

Acceptance Criteria:
* `st2ctl reload --register-triggers` new command exists 
* `st2ctl reload --register all` command also registers triggers

There's probably more. I'm not sure what is run for testing packs, but I'd also include whatever that mechanism is (to close StackStorm/st2contrib#475) in the acceptance criteria. :)

Kami:
Thanks for opening this ticket.

We've briefly talked about this in the past and I'm +1 on that change (this way we can stop "abusing" the API in some scenarios like sensu pack, etc.).

Kami:
This has been addressed by #2752.

Closing & thanks!

PR #2701, opened 2016-05-19
Add "--register-triggers" option to register content script and st2ctl

codyaray:
Follow-up PR for StackStorm/st2#2682.

This is entirely untested... but it sure looks thorough!

I don't have a StackStorm dev environment setup yet. 😢  What's the latest suggestion with [st2workroom/st2dev](https://docs.stackstorm.com/development/index.html#setting-up-a-development-environment) in a [native packaged](https://docs.stackstorm.com/install/deb.html) world? What's the team using nowadays? Or is this still valid?

### DIFF
--- a/st2common/st2common/bootstrap/triggersregistrar.py
+++ b/st2common/st2common/bootstrap/triggersregistrar.py
@@ -1,0 +1,182 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+import six
+
+from st2common import log as logging
+from st2common.constants.meta import ALLOWED_EXTS
+from st2common.bootstrap.base import ResourceRegistrar
+import st2common.content.utils as content_utils
+from st2common.models.api.trigger import TriggerTypeAPI
+from st2common.persistence.trigger import TriggerType
+
+__all__ = [
+    'TriggersRegistrar',
+    'register_triggers'
+]
+
+LOG = logging.getLogger(__name__)
+
+
+class TriggersRegistrar(ResourceRegistrar):
+    ALLOWED_EXTENSIONS = ALLOWED_EXTS
+
+    def register_triggers_from_packs(self, base_dirs):
+        """
+        Discover all the packs in the provided directory and register triggers from all of the
+        discovered packs.
+
+        :return: Number of triggers registered.
+        :rtype: ``int``
+        """
+        # Register packs first
+        self.register_packs(base_dirs=base_dirs)
+
+        registered_count = 0
+        content = self._pack_loader.get_content(base_dirs=base_dirs,
+                                                content_type='triggers')
+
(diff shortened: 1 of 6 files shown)


codyaray:
@Kami @lakshmi-kannan I somehow fell into this rabbit hole and this is what came out.

🐇

Kami:
Awesome, thanks!

Will have a look tomorrow.

Kami:
As far as dev environment goes - I personally still use devenv (st2dev repo), but in short, you need a Ubuntu VM with dependencies (MongoDB, Python, etc.) installed and st2 repo cloned.

You could even start on a fresh Ubuntu VM (install dependencies yourself) or use script for package based installation and clone the repo yourself.

st2common/st2common/bootstrap/triggersregistrar.py
```diff
@@ -0,0 +1,182 @@
...
+                            (pack, pack_field))
+
+        entry_point = content.get('entry_point', None)
+        if not entry_point:
+            raise ValueError('Trigger definition missing entry_point')
+
+        triggers_dir = os.path.dirname(trigger_metadata_file_path)
+        trigger_file_path = os.path.join(triggers_dir, entry_point)
+        artifact_uri = 'file://%s' % (trigger_file_path)
+        content['artifact_uri'] = artifact_uri
```
Kami: TriggerType DB model doesn't take those two attributes so you should remove them.

tools/st2ctl
```diff
@@ -51,6 +51,7 @@ function print_usage() {
     echo "usage: st2ctl {reload, clean}"
     echo "optional arguments:"
     echo "  --register-all                Register all."
+    echo "  --register-triggers           Register all triggers."
```
Kami: Also need to update `st2common/bin/st2ctl`, I believe this file is unused now.

@manasdk @lakshmi-kannan can some verify that this file is indeed unused and we can get rid of it to avoid duplication and confusion?

st2common/st2common/bootstrap/triggersregistrar.py
```diff
@@ -0,0 +1,182 @@
...
+
+__all__ = [
+    'TriggersRegistrar',
+    'register_triggers'
+]
+
+LOG = logging.getLogger(__name__)
+
+
+class TriggersRegistrar(ResourceRegistrar):
```
Kami: Base resource registrar class definitely needs some more love in the future :)

It's better than it was originally, but there is still way too much duplicated code.

Anyway, nothing for you to do right now, more of a note for us to fix / improve it.

tools/st2ctl
codyaray: Already updated `st2common/bin/st2ctl`. Its in this commit. Wasn't sure which was used, so I just updated both. (And I saw that your recent changes for --register-setup-virtualenvs were in both)

But yeah, seems like one should go bye bye. 👍

st2common/st2common/bootstrap/triggersregistrar.py
codyaray: Agreed, there was a lot of copy and paste the find and replace in my PR. :)

tools/st2ctl
Kami: Looks like I missed it, sorry.

Kami:
Can you please look into the Travis CI failure and also add some tests? :)

After that, I will manually test it and if everything looks OK, I will go ahead and merge it.

Kami:
@codyaray How is this PR looking?

We would love to include in v1.5 (should be out in week or so), so it would be great if you could get if finished. If you can't or don't have time, please let us know and I will take over.

codyaray:
Go ahead and take it. This should either give you a good head start or cause you a lot of problems. ;)

Yesterday I started setting up st2devenv but was reading through to figure out how to setup the code dir to install my local st2 download for editing, etc. Didn't make it through.

Anyway, hopefully this helps. If not, I'm still glad to see it in good hands! :)

tools/st2ctl
armab: +1 to remove this file. It's not used in new packages (only old deprecated system).

Apart of that, we can remove directories like https://github.com/StackStorm/st2/tree/master/st2auth/packaging, since it's part of old deprecated process as well.

PR #2703, opened 2016-05-19
JUST_FOR_FUN: Add colors and timing info to nosetests

lakshmi-kannan:
By default, I am turning on colors so I can spot an error easily. 

Optionally, by specifying NOSE_TIME=1 you get detailed drill down of how much time each test took!
Yellow indicates a bad test and red indicates an abomination. 

https://www.evernote.com/l/ANGmYg7IwnJGaYrloBvo_ItyD3GmrbzJ5j4

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,13 @@ PYTHON_TARGET := 2.7
 REQUIREMENTS := test-requirements.txt requirements.txt
 PIP_OPTIONS := $(ST2_PIP_OPTIONS)
 
+NOSE_OPTS := --rednose --immediate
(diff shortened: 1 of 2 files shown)


manasdk:
This look great. 

+1

Kami:
Yay, colors!

Nice, :+1:

Kami:
Wonder if we should also create an alias for nosetests command or similar.

A lot of times when I want to run a particular tests I invoke `notests` manually. I'm kinda too lazy to always pass in those flags. In any case, it's not a big deal.

PR #2708, opened 2016-05-23
Pack Configuration v2

Kami:
Note: This pull request is still WIP so please don't comment on the diff until the WIP tag has been removed.

This pull request introduces support for pack configs located in `/opt/stackstorm/configs/<pack name>.yaml` files. Those config values can contain static or dynamic values. Dynamic values are the ones which key references a value in a datastore (this value could either be scoped to the user or be in a global scope).

Here is an example config which utilizes all the available approaches:

```yaml
---
  api_key: "some_api_key" # static value
  api_secret: "{{user.api_secret}}"  # user scoped datastore value - for user joe this would resolve to "pack_configs:dummy_pack:joe:api_secret"
  regions:
    - "us-west-1"
  private_key_path: "{{system.private_key_path}}"  # global pack datastore value - this would resolve to "pack_configs:dummy_pack:private_key_path"
```

Keep in mind that all the pack config related values in the datastore contain a special prefix (right now that is `pack_config:<pack name>:<config key name>`) so we can differentiate between pack config and other datastore values.

## Backward Compatibility

If a pack local config is found (`<pack path>/config.yaml`), it's loaded first. If no `<pack name>.yaml` config is found, the behavior is exactly the same as before.

If a `<pack name>.yaml`  is found, it's loaded and values from this config have a precedence over the values from the pack local config (they are merged from left to right and they override pack local config items).

## Other Notes

To make getting started for a common scenario easier, we will begin with config schema being optional. If config schema is not provided, we won't know about the value type and we will simply assume that a value type is string.

This will work for simple case when datastore values are strings, but if user wants to use different types, they will need to define the schema.

Down the road (once all the st2contrib packs are sanitized and have a corresponding schema) we should also make schema mandatory if user wants to utilize values from a datastore.

### 1. Should configs be registered by default when running register-content script?

Right now sensors, actions and rules (and packs and config schemas implicitly) are registered by default. The question is if configs should also be registered by default when register-content script runs?

If not, user will need to manually pass `--register-configs` flag to the `st2-register-content` script.

### 2. Should datastore items which correspond to pack config values be "pack local"?

Right now all the dynamic config items values are local to a pack. This means that when user references a datastore variable (e.g. ``{{user.api_api}}`` or ``{{system.some_value}}`), a datastore value which is local to this pack is returned. This was done intentionally to avoid conflicts and collisions (e.g. user wants to use a different value for a config item which has the same name across different packs, etc.).

Internally, that's implemented by prefixing config related datastore values with a special prefix (`pack_configs:<pack name>:<username>:<item name>`).

For example, for pack "libcloud" and a config item "api_secret" the key internal name would be `pack_configs:libcloud:api_secret` (or `pack_configs:libcloud:<user>:api_secret` for user-scoped value).

The question is if we want this behavior or not (meaning config related datastore values are not prefixed). Both approaches have advantages and disadvantages and might be desired or not desired in some scenarios.

Even if values were not prefixed / namespaced, user would still need to use a special CLI command for setting those datastore values because we support arbitrary types for config items and the API which is responsible for storing this values knows how to handle serialization.

One of the problems if we make those values global also is that they aren't that much use outside of the pack config context - as mentioned above, we support arbitrary types and right now only config related code knows how to handle those types. If we want to make those values useful outside of the config context, this means we would also need to make other code paths where those values can be used aware of this serialization format so the valued can be de-serialized correctly.

One approach where we could avoid that would be to only support string types for dynamic template values and secrets. This means all non-string values would need to be explicitly declared in the config file and couldn't reference a value from a datastore. This has limitations, but it makes it a bit simpler and more straight forward.

## TODO

- [x] Add support for secret values
- [x] Simply config schema file format - make it flat and get rid of one level of nesting
- [x] API endpoints for setting and viewing config values in the datastore
- [x] Utilize config schema when configuring dynamic pack values using the API
- [x] Allow administrator to configure user-scoped dynamic configuration value for any system user
- [ ] New RBAC PermissionType which allows configuration of dynamic pack configuration values
- [x] Allow administrator to configure user-scoped variable for arbitrary user
- [ ] CLI commands for setting and viewing config values in the datastore
- [ ] Hook up actions and sensors to utilize ConfigLoader
- [x] Update register-content scripts, add --register-configs flag
- [ ] Deprecate pack local configs (`<pack path>/config.yaml file`)
- [ ] Code refactor and improvements (refactor prefix and key_prefix in lookup classes, etc.)
- [ ] Documentation - https://github.com/StackStorm/st2docs/pull/163


manasdk:
> Down the road (once all the st2contrib packs are sanitized and have a corresponding schema) we should also make schema mandatory if user wants to utilize values from a datastore.

Do watch out for making choices like this that can make it a little bit harder for both pack author and pack consumer.

manasdk:
> If a <pack name>.yaml is found, it's loaded and values from this config have a precedence over the values from the pack local config (they are merged from left to right and they override pack local config items).

+1.

st2actions/st2actions/runners/actionchainrunner.py
```diff
@@ -180,11 +180,11 @@ def _is_valid_node_name(self, all_node_names, node_name):
             # This task name needs to be resolved during run time so we cant validate the name now
             return True
 
-        for jinja_start_marker in JINJA_START_MARKERS:
-            if jinja_start_marker in node_name:
-                # This task name needs to be resolved during run time so we cant validate the name
-                # now
-                return True
+        is_jinja_expression = jinja_utils.is_jinja_expression(value=node_name)
```
lakshmi-kannan: return jinja_utils.is_jinja_expression(value=node_name) should be sufficient, no?

st2api/st2api/controllers/v1/pack_config_schemas.py
```diff
@@ -24,17 +24,17 @@
 
 
 __all__ = [
-    'PackConfigSchemaController'
+    'PackConfigSchemasController'
 ]
 
 
-class PackConfigSchemaController(ResourceController):
+class PackConfigSchemasController(ResourceController):
```
lakshmi-kannan: I like the singular version better

st2api/st2api/controllers/v1/pack_configs.py
```diff
@@ -0,0 +1,106 @@
...
+from st2common.rbac.types import PermissionType
+from st2common.rbac.decorators import request_user_has_permission
+from st2common.rbac.decorators import request_user_has_resource_db_permission
+
+__all__ = [
+    'PackConfigsController'
+]
+
+
+class PackConfigsController(ResourceController):
```
lakshmi-kannan: There is only one config per pack. Singular version is better.

st2api/st2api/controllers/v1/pack_configs.py
```diff
@@ -0,0 +1,106 @@
...
+    def __init__(self):
+        super(PackConfigsController, self).__init__()
+
+        # Note: This method is used to retrieve object for RBAC purposes and in
+        # this case, RBAC is checked on the parent PackDB object
+        self.get_one_db_method = packs_service.get_pack_by_ref
+
+    @request_user_has_permission(permission_type=PermissionType.PACK_LIST)
+    @jsexpose()
+    def get_all(self, **kwargs):
```
lakshmi-kannan: I am not sure why this should be exposed now. Is there even a use case to get configs for all the packs?

st2common/st2common/content/bootstrap.py
```diff
@@ -248,6 +252,28 @@ def register_policies():
...
+        registered_count = configs_registrar.register_configs(pack_dir=pack_dir,
+                                                              fail_on_failure=fail_on_failure)
+    except Exception as e:
+        exc_info = not fail_on_failure
+        LOG.warning('Failed to register configs: %s', e, exc_info=exc_info)
+
+        if fail_on_failure:
+            raise e
+
+    LOG.info('Registered %s configs.' % (registered_count))
```
lakshmi-kannan: %d

manasdk: Always prefer the lazy construction of the string.

LOG.info('Registered %d configs.', registered_count)

st2common/st2common/models/api/pack.py
```diff
@@ -120,5 +125,74 @@ def to_model(cls, config_schema):
...
+        "properties": {
+            "id": {
+                "description": "The unique identifier for the config.",
+                "type": "string"
+            },
+            "pack": {
+                "description": "The content pack this config belongs to.",
+                "type": "string"
+            },
+            "values": {
```
lakshmi-kannan: I prefer to call them configs. cc: manas?

st2api/st2api/controllers/v1/pack_configs.py
```diff
@@ -0,0 +1,106 @@
...
+        Retrieve config for a particular pack.
+
+        Handles requests:
+            GET /configs/<pack_ref>
+        """
+        # TODO: Make sure secret values are masked
+        return self._get_one_by_pack_ref(pack_ref=pack_ref)
+
+    @jsexpose(arg_types=[str], body_cls=ConfigItemSetAPI)
+    def put(self, pack_ref, config_item_api):
```
lakshmi-kannan: I am not sure if this is a good idea to do this without locks. In HA, this is obviously a problem. Just noting it so you capture in your TODO. It's probably easier to fix it now.

st2common/st2common/services/config.py
```diff
@@ -0,0 +1,200 @@
...
+DATASTORE_CONFIG_KEY_PREFIX = 'pack_config'
+
+
+def get_datastore_key_prefix_for_pack(pack_name):
+    prefix = [DATASTORE_CONFIG_KEY_PREFIX, pack_name]
+    prefix = DATASTORE_KEY_SEPARATOR.join(prefix)
+    return prefix
+
+
+def get_datastore_key_name(pack_name, key_name, user=None):
```
lakshmi-kannan: I am really confused why you are bringing in the notion of pack scoped config variables again. We explicitly decided not to support it now. 

People can specify only 3 things in the config.yaml file in /opt/stackstorm/configs/${pack}/config.yaml:

k1: v1
k2: {{system.foo}}
k3: {{user.bar}}

With this in mind, we just have the problem of config resolution (lookup). We just need to save the flat list of config item as an object in db after resolution. That can just be a DB model.

st2common/st2common/models/api/pack.py
Kami: Imo, `configs` would be a weird name. That's jut the attribute which holds the actual config values (since we obviously can't store those values as a top-level attributes on the model).

st2common/st2common/content/bootstrap.py
Kami: Sigh, always getting distracted by minor details (which were actually copied, for better or worse) from other registrar classes :P

st2actions/st2actions/runners/actionchainrunner.py
Kami: No (check a couple of lines bellow).

st2api/st2api/controllers/v1/pack_config_schemas.py
Kami: I've added "s" suffix to make it consistent with other top-level controllers. First it had no suffix since it lived as a nested controller under packs, but when moving it to a top-level I forgot to add "s" suffix.

All other top-level controllers use plural form, so we should also use it here for consistency. But yeah, I'm all for getting rid of it in API v2 where the path will probably look something like `/v2/packs/<pack ref>/config_schema`.

st2api/st2api/controllers/v1/pack_configs.py
Kami: Yes, but since this is a top-level controller (I wish it wasn't, but we are all aware of our pecan limitations right now...), same reason as here applies - https://github.com/StackStorm/st2/pull/2708#discussion_r64530912

st2api/st2api/controllers/v1/pack_configs.py
Kami: It's a top-level controller so if nothing else, it makes sense to have it for consistency reasons.

Kami:
> Do watch out for making choices like this that can make it a little bit harder for both pack author and pack consumer.

@manasdk Yeah, I followed progressive  enhancement / graceful degradation approach. Config schema is only required if user wants to utilize dynamic config values.

If no schema is present and config only contains static values, it still works.

In addition to that, users can simply move their existing configs outside of pack directory (e.g. `mv my_pack/config.yaml -> /opt/stackstorm/configs/my_pack.yaml`) to configs directory and everything will work without any additional input or changes (beyond running register-configs that is).

Should probably also document this somewhat prominently in the docs.

st2common/st2common/services/config.py
Kami: Let's sync up about the prefix and values being scoped to the pack later (we should also get feedback from other people on that).

In any case, that's a minor change and easy to do.

Kami: I've added "Open Questions" section to this PR. 2) refers to this.

As mentioned on Slack, let's make sure everyone understands the question / problem and understands the use-cases behind it and then we can decide which approach to take.

manasdk:
> 2. Should datastore items which correspond to pack config values be "pack local"?

I guess the system admin could setup these values as `{{user.libcloud_api_key}}` or `{{system.libcloud.api_key}}` i.e. making an explicit choice on the scoping and uniqueness. In the user mental model do we expect a user to naturally expect a property to be namespaced by a pack? If so then we must make the value "pack local" as you suggest. If not then we would have surprising behavior. I am leaning towards a user not having such an expectation plus "pack local" is not a pervasive concept anyway i.e. is it possible to have pack local variable anywhere else?

> One of the problems if we make those values global also is that they aren't that much use outside of the pack config context

This is a valid point and perhaps drives the temptation to provide the explicit scope. This thought is still conjecture and we cannot absolutely claim this to be true since we cannot always tell the nature of all config properties and in what cases a value can be shared between 2 config properties while maintaining the semantic.

dzimine:
> The question is if we want this behavior or not (meaning config related datastore values are not prefixed). 

DZ: No, we don't. A: user may have reasons to use the same piece of data in config and outside of config (e.g. in rule, in workflow?) and B: more importantly - it makes system less predictable thus confusion:  `{user.foo`} and `{user.bar}` look same but behave differently the moment I use `{user.bar}` in config. 

Users will do the prefixes themselves as a matter of convention, and watch for collisions themselves. E.g. I use `{user.aws_api_key}` and `{user.openstack_api_key}`, and know what I am doing, if I used `{user.api_key}` I wouldn't expect it to work everywhere :)  The mental model is env variables, it's familiar,  and folks know how to deal with this. 


> Even if values were not prefixed / namespaced, user would still need to use a special CLI command for setting those datastore values because we support arbitrary types for config items and the API which is responsible for storing this values knows how to handle serialization.

Two different CLI for same functionality doesn't seem right. What's wrong here though is that we don't support different types in KV store to begin with. Can we uniform it and support diff types everywhere (global, user, config)? if not, why, if yes, how much would it cost.

> One approach where we could avoid that would be to only support string types for dynamic template values and secrets. This means all non-string values would need to be explicitly declared in the config file and couldn't reference a value from a datastore. This has limitations, but it makes it a bit simpler and more straight forward.

This seems like a reasonable if I understand it correctly. If we support strings (and numbers) only, can config file still use, and config template break it down to supply hardcoded values, and templated strings? In the example of config (below, modified) it doesn't look like any limitation. This part I may be missing.

```yaml
---
  api_key: "some_api_key" # static value
  api_secret: "{{user.my_pack.api_secret}}"  
 # DZ here I assume user scoped datastore value don't auto-prefixed with pack 
 # and no magic resolution happened, it would be just "users:joe:my_pack.api_secret"
  regions:
    - "us-west-1"
  private_key_path: "{{system.private_key_path}}"  
  # global pack datastore value - this would resolve (NO SURPRISE!) to "system:private_key_path"
  # as if user wants to use private key for all over. 
  # And if not, user will name the value "my_pack.private_key_path" 
```

DoriftoShoes:
> 1. Should configs be registered by default when running register-content script?

I think so, yes.  Pack config is a crucial piece to making the pack work correctly for the user.

> 2. Should datastore items which correspond to pack config values be "pack local"?

This is tough.  I think not though.  Similar to how I have common.yaml set up in hiera, I assume users will want access to some of the same data across packs (the monitoring server address for example).  I think if someone needs a pack specific setting, they will use a special key (system.puppet_puppetmaster for the puppet pack).

Kami:
Thanks to everyone for their feedback

@dzimine Your example is correct. And in this case since variables only refer to string types, it would work just fine.

And yes, users can still supply "static" non-strings values in the same manner as they do today in `config.yaml` file - besides the support for dynamic values, new config format is fully backward compatible and users can easily migrate to it by moving existing config outside of the pack to `/opt/stackstorm/configs/<pack name>.yaml`.

st2api/st2api/controllers/v1/pack_configs.py
lakshmi-kannan: I see, got it.

st2common/st2common/models/api/pack.py
lakshmi-kannan: I don't think it's weird. cc: @manasdk

lakshmi-kannan:
I am with DZ on not support pack scope configs now and I already registered my comment. And regarding types, I do want to support all simple types for values in configs but I think this is addressed.

Kami:
Alright, so we decided to not support pack scope and only support string values in the datastore.

This reduces the scope of this change quite a bit. It also means we don't need a separate API endpoint for setting those values, etc.

After I'm done with those changes I will start working on RBAC and related changes for "set keyvalue" API endpoint. This a change I actually already started to work on in this PR (inside the API endpoint handler for setting a config value, but now it's not needed here anymore.

jjm:
This sounds great, when your ready for schema's for packs to hit st2contrib I'm happy update any of ours to use this method as it'll make things so much easier. The same applies for packs.deploy config too.

Kami:
I made the necessary changes and for all purposes this PR is now complete. I will go ahead and merge it.

As far as the whole feature goes, it's not yet complete and still requires some work (I'm on it):

1. RBAC and other datastore API related changes - this was previously done in this PR, but since we decided to not support pack local config / datastore values and only support string values in a datastore, some of those changes now need to be done inside the datastore API controller (that's not necessary bad, in fact it's good since it's generic and same functionality can be re-used for pack config and other datastore values)
2. Documentation changes

Issue #2692, opened 2016-05-18
Request: Ability to migrate api keys to new installs


sixth:
Hi there, 

As stackstorm is going through updating its installation and upgrade processes, I've been having to create fresh new st2 instances as a fix for different issues i've encountered. 

My production stackstorm instance is mostly used via its webhook and accessed through an api key. Since at most times the api key is hardcoded to my developer's code, It would be a nice feature to have a way to migrate the api keys between different st2 instances.

Keep up the great work!
-E

manasdk:
### Extract API keys
`st2 apikey list -dj/-dy`

### Upload API keys
`st2 apikey create -f apikeys.json/apikeys.yaml`

lakshmi-kannan:
@manasdk Can you document this in st2docs?

manasdk:
> Can you document this in st2docs?

Sure, once I implement it :)

manasdk:
Done.

PR #2710, opened 2016-05-24
[RFR] Ability to load apikeys from a file.

manasdk:
Implements feature request https://github.com/StackStorm/st2/issues/2692

### Extract API keys
`st2 apikey list -dy > apikeys.yaml`

### Load API keys
`st2 apikey load apikeys.yaml`

Note : Going with `load` instead of re-using `create` is to avoid any confusion. Creation will always create a new ApiKey while load will assume the key_hash was properly generated by a StackStorm instance and that the user already has a matching ApiKey. To preserve the distinction between a new key is desired and that an existing key is being loaded the 2 separate verbs are preferred.

Appropriate fixes made in the API layer.


manasdk:
Docs PR https://github.com/StackStorm/st2docs/pull/162

PR #2712, opened 2016-05-25
Add API key support to st2client

enykeev:
As requested in https://github.com/StackStorm/st2-packages/issues/283, the PR adds support for API key to st2client and allows user to store said key in `~/.st2/config`.

### DIFF
--- a/st2client/st2client/client.py
+++ b/st2client/st2client/client.py
@@ -37,7 +37,7 @@ DEFAULT_API_VERSION = 'v1'
 
 class Client(object):
     def __init__(self, base_url=None, auth_url=None, api_url=None, api_version=None, cacert=None,
-                 debug=False, token=None):
+                 debug=False, token=None, api_key=None):
         # Get CLI options. If not given, then try to get it from the environment.
         self.endpoints = dict()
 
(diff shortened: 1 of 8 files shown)


st2client/st2client/client.py
```diff
@@ -37,7 +37,7 @@
 
 class Client(object):
     def __init__(self, base_url=None, auth_url=None, api_url=None, api_version=None, cacert=None,
-                 debug=False, token=None):
+                 debug=False, token=None, apikey=None):
```
Kami: For consistency, I think it would be better to call it `api_key`.

st2client/st2client/config_parser.py
```diff
@@ -79,6 +79,10 @@
         'password': {
             'type': 'string',
             'default': None
+        },
+        'apikey': {
```
Kami: Do we even want to support API key in the config?

Config is mostly meant to be used by users and API keys are mostly meant to be used for integrating with 3rd party services so I would also be OK with not supporting API key in the CLI config.

enykeev: The whole story started with https://github.com/StackStorm/st2-packages/issues/283 which implies there is a valid usecase to have api key in config.

st2client/st2client/client.py
enykeev: I was under the impression it would be inconsistent with variable name in some other place, but I went and checked and could not find one. So here you go.

enykeev:
@Kami so do I have a green light on that one?

Kami:
LGTM, +1

Please also open a corresponding st2docs PR (I believe there should already be a chapter / section on CLI config somewhere).

PR #2714, opened 2016-05-26
Remove temp files created by submit_debug_info

vcabbage:
Currently `submit_debug_info.py` doesn't clean up the the temp files it creates. This PR adds a finally statement to remove the temp files created by the `create_archive` function.

Added a remove_dir function to `st2debug/utils/fs.py` for consistency with the `remove_file` function.

### DIFF
--- a/st2debug/st2debug/cmd/submit_debug_info.py
+++ b/st2debug/st2debug/cmd/submit_debug_info.py
@@ -241,12 +241,12 @@ class DebugInfoCollector(object):
         try:
             # 1. Create temporary directory with the final directory structure where we will move
             # files which will be processed and included in the tarball
-            temp_dir_path = self.create_temp_directories()
+            self._temp_dir_path = self.create_temp_directories()
 
             # Prepend temp_dir_path to OUTPUT_PATHS
             output_paths = {}
             for key, path in OUTPUT_PATHS.iteritems():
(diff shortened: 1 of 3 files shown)


Kami:
Great, thanks.

We also need to make sure directory and file is not removed when using `--review` option, right?

vcabbage:
According to the description, `--review` generates the tarball. This only removes the temp dir, the tarball will still be output.

``` python
    parser.add_argument('--review', action='store_true', default=False,
                        help='Generate the tarball, but don\'t encrypt and upload it')
```

Kami:
@vcabbage Alright, great - then we should be fine.

On a related note - it would be great if you can add a test case for this change.

vcabbage:
Difficult to test as written. I could check the number of files int /tmp and verify there are no additional files after `create_archive` is run, though that sounds a bit flaky. Or I could save the temp dir path to an attribute and check for it's existence directly.

Do you have a preference (or better approach)?

Kami:
Would need to dig in, but in short, we would probably need to modify some methods to store a list of directories and files which are created...

vcabbage:
Since it deletes the entire directory I added a test to check if the directory exists after running. Let me know if that works.

Kami:
Yep, that works - thanks!

Kami:
Merged, thanks.

Issue #2706, opened 2016-05-20
Add another "timestamp" attribute to the execution object when represents a time execution has been started


Kami:
Right now we have `start_timestamp` which is set when an execution is scheduled (aka requested by the API or other service) and `stop_timestamp` which is set by the runner container when execution finishes.

There could be a delay (for whatever reason) between when execution is scheduled (`start_timestamp`) and when it actually starts being executed by the runner. We should add another timestamp attribute to the `ActionExecutionDB` model (dunno, `execution_start_timestamp` or similar, but ideally we would probably rename existing `start_timestamp` to `scheduled_timestamp`, but this is a bit of a painful change) which indicates a timestamp when action execution has been started by the runner.

Originally reported by Sayli K. on Slack.

manasdk:
I cannot help but agree with @enykeev. We need some sort of audit-trail of activities on a single execution.

Kami:
I agree, ideally we would log all the state transitions and events related to executions.

We already kinda do that in the logs, but it's not really in a consistent and structured format and that makes it hard to use and query.

In any case, I think this additional timestamp would still be beneficial, even if we had this even stream, or whatever we want to call it.

enykeev:
Note that we change status to `running` when we dispatch it to MQ [[1]](https://github.com/StackStorm/st2/blob/b74067d4afd9f871f42876e93c633883fa38af68/st2actions/st2actions/worker.py#L107-L107) and there is really no limit on how long it might be there before it gets picked up by the worker. Also note that the way st2 is designed, worker has a pretty decent chance to be running on the different machine with local time being skewed one way or the other.

So where exactly do we want to sample this execution_timestamp for it to be useful for debugging purposes? On dispatch, on worker run or on both? Do we want to start solving the problem if we're pretty much agree we can't solve it good enough without per-execution event log?

lakshmi-kannan:
>Do we want to start solving the problem if we're pretty much agree we can't solve it good enough without per-execution event log?

We can punt it but I am pretty sure we won't work on event log in next few months. I'd rather have a timestamp now and build event log later. As far as where you should start ticking the time for running, it looks like we actually have a bug that we set status to ``running`` even before dispatching. So I think we need a new state ``dispatched``. This is anyways needed and the right thing to do. We should mark the action as running when the worker picked it up. I know this is more work but I don't think it's a week of work. Thoughts everyone?

Kami:
This has been resolved by #2718, closing.

PR #2718, opened 2016-05-27
Execution log

enykeev:
While discussing #2706 we figured there is a few new timestamps we would like to add to execution record to improve user's understanding of the flow and make it easier to debug indeterminate set of problems (primarily performance related I would guess).

Problem is, there is quite a few execution states already and there's going to be even more if we are to focus on the problem seriously. Storing timestamps related to this state changes as a properties of the execution itself will quickly bloat the execution object beyond reason.

An obvious solution to the problem is to create another data structure that would hold this expanding set of data in a linear fashion.

The idea is, for every execution state mutation, we creating a log entry with the state execution transitions to and the timestamp when it happened. For some time, this this structure may exist in parallel to our current `state`, `start_timestamp` and `end_timestamp` to ensure backward compatibility and while providing additional context for new executions.

For now, user would have to access api directly or use st2client debug mode to get to this data, but as soon as we happy with it, we can easily add a command for it to cli and a view in web.

### DIFF
--- a/st2common/st2common/models/api/execution.py
+++ b/st2common/st2common/models/api/execution.py
@@ -137,7 +137,7 @@ class ActionExecutionAPI(BaseAPI):
             end_timestamp = isotime.format(end_timestamp, offset=False)
             doc['end_timestamp'] = end_timestamp
 
-        for entry in doc['log']:
+        for entry in doc.get('log', []):
             entry['timestamp'] = isotime.format(entry['timestamp'], offset=False)
 
         attrs = {attr: value for attr, value in six.iteritems(doc) if value}
--- a/st2common/st2common/models/api/execution.py
+++ b/st2common/st2common/models/api/execution.py
@@ -107,6 +107,7 @@ class ActionExecutionAPI(BaseAPI):
                 "uniqueItems": True
             },
             "log": {
+                "description": "Contains information about execution state transitions.",
                 "type": "array",
                 "items": {
                     "type": "object",
--- a/st2common/st2common/models/api/execution.py
+++ b/st2common/st2common/models/api/execution.py
@@ -105,6 +105,22 @@ class ActionExecutionAPI(BaseAPI):
                 "type": "array",
                 "items": {"type": "string"},
                 "uniqueItems": True
+            },
+            "log": {
+                "type": "array",
+                "items": {
+                    "type": "object",
+                    "properties": {
+                        "timestamp": {
+                            "type": "string",
+                            "pattern": isotime.ISO8601_UTC_REGEX
+                        },
+                        "status": {
+                            "type": "string",
+                            "enum": LIVEACTION_STATUSES
+                        }
+                    }
+                }
             }
         },
         "additionalProperties": False
@@ -120,6 +136,9 @@ class ActionExecutionAPI(BaseAPI):
         if end_timestamp:
             end_timestamp = isotime.format(end_timestamp, offset=False)
             doc['end_timestamp'] = end_timestamp
+
+        for entry in doc['log']:
+            entry['timestamp'] = isotime.format(entry['timestamp'], offset=False)
 
         attrs = {attr: value for attr, value in six.iteritems(doc) if value}
         return cls(**attrs)
--- a/st2common/st2common/services/executions.py
+++ b/st2common/st2common/services/executions.py
@@ -148,10 +148,14 @@ def _get_parent_execution(child_liveaction_db):
 def update_execution(liveaction_db, publish=True):
     execution = ActionExecution.get(liveaction__id=str(liveaction_db.id))
     decomposed = _decompose_liveaction(liveaction_db)
+
     kw = {}
     for k, v in six.iteritems(decomposed):
         kw['set__' + k] = v
+
     if liveaction_db.status != execution.status:
+        # Note: If the status changes we store this transition in the "log" attribute of action
+        # execution
         kw['push__log'] = _create_execution_log_entry(liveaction_db.status)
     execution = ActionExecution.update(execution, publish=publish, **kw)
     return execution
--- a/st2common/st2common/services/executions.py
+++ b/st2common/st2common/services/executions.py
@@ -56,7 +56,10 @@ __all__ = [
 
 LOG = logging.getLogger(__name__)
 
-SKIPPED = ['id', 'callback', 'action', 'runner_info', 'parameters', 'notify']
+# Attributes which are stored in the "liveaction" dictionary when composing LiveActionDB object
+# into a ActionExecution compatible dictionary.
+# Those attributes are LiveAction specified and are therefore stored in a "liveaction" key
+LIVEACTION_ATTRIBUTES = ['id', 'callback', 'action', 'runner_info', 'parameters', 'notify']
 
 
 def _decompose_liveaction(liveaction_db):
@@ -66,7 +69,7 @@ def _decompose_liveaction(liveaction_db):
     decomposed = {'liveaction': {}}
     liveaction_api = vars(LiveActionAPI.from_model(liveaction_db))
     for k in liveaction_api.keys():
-        if k in SKIPPED:
+        if k in LIVEACTION_ATTRIBUTES:
             decomposed['liveaction'][k] = liveaction_api[k]
         else:
             decomposed[k] = getattr(liveaction_db, k)
@@ -74,6 +77,9 @@ def _decompose_liveaction(liveaction_db):
 
 
 def _create_execution_log_entry(status):
+    """
+    Create execution log entry object for the provided execution status.
+    """
     return {
         'timestamp': date_utils.get_datetime_utc_now(),
         'status': status
--- a/st2common/st2common/services/executions.py
+++ b/st2common/st2common/services/executions.py
@@ -73,7 +73,7 @@ def _decompose_liveaction(liveaction_db):
     return decomposed
 
 
-def _create_timestamp(status):
+def _create_execution_log_entry(status):
     return {
         'timestamp': date_utils.get_datetime_utc_now(),
         'status': status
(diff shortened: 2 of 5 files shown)


enykeev:
Lets say I just want to get a discussion going and not really expect we merge it soon. Still, to me it seems like a better way to solve #2706 than adding another property.

Question is, do you see any particular problems with that approach... well, besides the fact we haven't really discussed and designed it as a team before implementing. Tests and docs are coming as soon as I hear anything on the substance of the change.

st2common/st2common/models/db/liveaction.py
```diff
@@ -75,6 +75,7 @@ class LiveActionDB(stormbase.StormFoundationDB):
         default={},
         help_text='Information about the runner which executed this live action (hostname, pid).')
     notify = me.EmbeddedDocumentField(NotificationSchema)
+    log = me.ListField(field=me.DictField())
```
enykeev: I'm not a fan of that part, but the way we implemented execution update [[1]](https://github.com/StackStorm/st2/blob/b3b394fd45db38c1aa4cd02e833498c15e4752db/st2common/st2common/services/executions.py#L132-L138) we would have to do some guessing to figure out whether update should or should not add a log entry.

Remind me once again, why do we need liveactions? =)

lakshmi-kannan: We used to have three models a) LiveAction b) Execution c) History. We realized we need only two things: 1. A lightweight model as a control message to track lifetime of action until it begins execution and 2) An actual execution model that contains the data and the context for the execution. Now we have come close to deprecating LiveAction multiple times but have realized cases where we need two. I cannot recollect a specific example now. The rule of thumb we now have is LiveAction is a system internal that should never be exposed to users and execution is a user consumable model.

st2common/st2common/services/action.py
```diff
@@ -98,6 +98,11 @@ def create_request(liveaction):
     liveaction.status = action_constants.LIVEACTION_STATUS_REQUESTED
     liveaction.start_timestamp = date_utils.get_datetime_utc_now()
 
+    liveaction.log = [{
```
lakshmi-kannan: I'd prefer a util method that either adds to array or appends it. The current way is prone to errors (programmer).

st2common/st2common/models/db/liveaction.py
lakshmi-kannan: All that said, just add log to execution model. You don't need this in liveaction.

manasdk:
I don't see the value of having this property on liveaction. Having on ActionExecution is fine.

st2common/st2common/util/action_db.py
```diff
@@ -195,6 +196,11 @@ def update_liveaction_status(status=None, result=None, context=None, end_timesta
     old_status = liveaction_db.status
     liveaction_db.status = status
 
+    liveaction_db.log.append({
```
Kami: To avoid race issues, we should use MongoDB method for automatically publishing an item to an array.

manasdk: Checkout https://github.com/StackStorm/st2/blob/1df8f4cdd4fc4e3bd1eac79732d1b3aac1be4745/st2common/st2common/persistence/trace.py#L28

Idea is to use the push operator. Likely applies here as well.

enykeev: How it relates to our plans to ditch MongoDB? Are we cool about using mongo-specific features?

enykeev:
I've reworked the PR by removing log from LiveAction and instead trying to figure out if status have changed during execution update. Doing so, I've substantially shortened window of opportunity for race condition. Using atomic updates would require me to use mongo specific api and then manually publish updates and dispatch triggers. Doesn't seem necessary to me, though, I might be missing something. How big of a change we get race condition between [this lines](https://github.com/StackStorm/st2/blob/21fb998af876b801598e741aed7073dcef75121b/st2common/st2common/services/executions.py#L143-L149) and how different the situation of `log` getting a race versus any other field?

st2common/st2common/util/action_db.py
lakshmi-kannan: We already have mongo specific code. I am divided honestly. I don't see the need here to use mongo specific atomic updates but I might be missing something. cc: @manasdk, @Kami

Kami: Yes, I agree.

Ideally we would have a better abstraction (at least a service function or similar), but I don't think DB specific code is too big of a deal right now.

A (DB neutral) alternative is to use locking, but this would require more code.

st2common/st2common/models/api/execution.py
```diff
@@ -105,6 +105,22 @@ class ActionExecutionAPI(BaseAPI):
                 "type": "array",
                 "items": {"type": "string"},
                 "uniqueItems": True
+            },
+            "log": {
```
Kami: If we only store status information in this attribute, I would prefer to have  some more self-explanatory name (status_transition_log or similar).

If we plan to expand it later, I'm fine with this name though.

Kami:
It's also worth nothing that re-using existing model could mean potential negative performance overhead, but that really depends on the way MongoDB does locking and handles atomic operations.

Some of the locking granularity has been improved in MongoDB 3.x (which we don't require yet, but we should) so we should research how it handles it in such scenarios. This attribute doesn't change a lot, but it would still be terrible and have bad impact if MongoDB acquired a collection-wide lock when modifying a single document.

manasdk:
> I've reworked the PR by removing log from LiveAction and instead trying to figure out if status have changed during execution update. Doing so, I've substantially shortened window of opportunity for race condition.

Whenever we identify that a race could it is best to eliminate the race by proper code design.

> Using atomic updates would require me to use mongo specific api and then manually publish updates and dispatch triggers. 

Yes and we hide those behind service classes as the DB specific abstraction. Using mongo specific features is ok if we can somehow localize the code.

st2common/st2common/models/api/execution.py
enykeev: At that point, I don't see what else could we add here. At the same time, I see no problem with expanding the log objects providing we keep in mind performance implications of transferring the execution down the wire.

lakshmi-kannan:
So @enykeev, we have two people insisting that the atomicity issue needs to be addressed. Can you add a commit and then we can merge this feature?

st2common/st2common/models/api/execution.py
```diff
@@ -121,6 +137,9 @@ def from_model(cls, model, mask_secrets=False):
             end_timestamp = isotime.format(end_timestamp, offset=False)
             doc['end_timestamp'] = end_timestamp
 
+        for entry in doc['log']:
```
Kami: Will this work since we don't default it to an empty array? Maybe safer to do `entry.get('log', []) or []` or something like that.

st2common/st2common/services/executions.py
```diff
@@ -72,6 +73,13 @@ def _decompose_liveaction(liveaction_db):
     return decomposed
 
 
+def _create_timestamp(status):
```
Kami: I would prefer a more explenatory name - `create_execution_log_entry` or similar :P

st2common/st2common/services/executions.py
```diff
@@ -132,9 +142,12 @@ def _get_parent_execution(child_liveaction_db):
 def update_execution(liveaction_db, publish=True):
     execution = ActionExecution.get(liveaction__id=str(liveaction_db.id))
     decomposed = _decompose_liveaction(liveaction_db)
+    kw = {}
     for k, v in six.iteritems(decomposed):
-        setattr(execution, k, v)
-    execution = ActionExecution.add_or_update(execution, publish=publish)
+        kw['set__' + k] = v
```
Kami: Is this required or you just did it to match rest of the code?

In any case, I think it shouldn't affect the end result so I'm fine with it.

Kami:
Besides a couple of comments, LGTM, :+1:

Kami:
Please also update changelog and corresponding documentation.

st2common/st2common/models/api/execution.py
enykeev: It seems like mongoengine does it for us since we've defined the field to be `me.ListField(field=me.DictField())`. I'm going to play safe and update it per your request anyway.

st2common/st2common/services/executions.py
enykeev: Sure you don't want us to switch to objc? =)

st2common/st2common/services/executions.py
enykeev: Surprisingly, `update` routine varies from `add_or_update` in most unexpected ways. It didn't work for me when I tried it initially and I've decided not to waste my time figuring out why.

manasdk: why are mongo specific literals in this module? They should be in persistence which is where we should try to hide the mongo specific stuff.

enykeev: Because it wasn't particularly clear it is the goal. `liveaction__id` few lines above is also mongo-specific and if you do `ag "\w__\w"`, you soon realize we don't really follow this convention outside of `trace` and `trigger` both written by you. Saying that, I'm ok with that convention and will try to figure out how to rewrite this function to match it, but we need everyone else to understand and comply with it too for it to have any effect.

enykeev:
> Please also update changelog and corresponding documentation.

As far as I can tell we don't really have any documentation on debugging state changes (mostly a performance topic I guess). We can add an article to Troubleshooting section, although I'd prefer to dark ship this function until we create proper way to access that information through cli or ui.

st2common/st2common/services/executions.py
manasdk: `persistence` is for all DB specific abstractions. Limits itself to a single model type like `ActionDB` etc. In theory replacing MongoDB with say Postgres should have required the bulk of the changes to be under the `persistence` layer only however due to our failure to maintain abstractions faithfully that will not be true. 

`service` - provides a higher level feature which can use persistence to retrieve data from the DB. Works simultaneously with multiple models and generally capture the meat of the what has been traditionally thought of as "business logic". https://github.com/StackStorm/st2/blob/master/st2common/st2common/services/action.py#L47 is a pretty decent example of what comprises service and why it is not necessary part of persistence.

Often abstraction do not make sense in all contexts however the lack of consistent abstractions make the code base challenging to reason about which is why I personally prefer to maintain them even if they might seem superfluous.

manasdk:
@enykeev Don't worry about my comments or don't let them block your current work. If you feel confident in what you have do merge the change. I hope I have at least conveyed why the distinction exists.

st2common/st2common/services/executions.py
Kami: :D

PR #2719, opened 2016-05-27
Pin kombu and amqp dependency

Kami:
New version of amqp was released (https://pypi.python.org/pypi/amqp/2.0.0) and it appears to be breaking some end to end tests.


Kami:
I also need to / fix change how `requirements.txt` generation works.

Right now we only explicitly pin and include direct dependencies in `requirements.txt`. `amqp` is a transitive dependency of `kombu` so we don't include it in generated `requirements.txt`.

lakshmi-kannan:
The pinning lgtm.

PR #2720, opened 2016-05-27
String casting improvements

Kami:
This pull request introduces some improvements in the string parameter casting:

* Treat `None` as-is and return early
* On non-string and None type, throw a more user-friendly exception (previously, `to_unicode` would throw a very cryptic exception)
* Include additional info in the existing log message

Reported by @jjm. Looking at the code and history, it looks like this bug was there for a long time, but it only manifested under specific conditions so we never noticed it before.

@jjm was also "lucky" to encounter this issue, since it would only manifest itself under a specific scenario - the way we do ChatOps string matching right now is a bit nasty (we also match a whitespace and then strip additional whitespace so for a matched parameter value of `" "` actual value of `""` would be passed in as an action parameter).

### DIFF
--- a/st2common/st2common/util/casts.py
+++ b/st2common/st2common/util/casts.py
@@ -62,6 +62,12 @@ def _cast_number(x):
 
 
 def _cast_string(x):
+    if x is None:
+        # Preserve None as-is
+        return x
+
+    # TODO: Throw on non None and string value? (otherwise to_unicode will
+    # throw)
     x = to_unicode(x)
     x = _cast_none(x)
     return x
--- a/st2common/st2common/util/casts.py
+++ b/st2common/st2common/util/casts.py
@@ -66,8 +66,11 @@ def _cast_string(x):
         # Preserve None as-is
         return x
 
-    # TODO: Throw on non None and string value? (otherwise to_unicode will
-    # throw)
+    if not isinstance(x, six.string_types):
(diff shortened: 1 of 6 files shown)


Kami:
Should we also back-port this into v1.4 branch?

If yes, we should roll out v1.4.1 release (I'm not really a fan of abusing package build numbers which is what we have done a couple of times now already... Every release should correspond to a specific StackStorm version).

st2common/st2common/util/casts.py
```diff
@@ -62,6 +62,15 @@ def _cast_number(x):
 
 
 def _cast_string(x):
+    if x is None:
+        # Preserve None as-is
+        return x
```
Kami: This is the actual fix for None type handling and issue @jjm encountered.

Rest of the changes are related user-friendliness improvements.

jjm:
I'll like to see this back ported :-D Otherwise I'll need to patch our 2 systems when the next package update happens,

Kami:
I will merge this into master and for now, cherry-pick those changes directly into v1.4 branch.

Kami:
@jjm I've merged changes directly into v1.4 branch so you should be able to upgrade to latest v1.4 packages once they are built (I will let you know when that happens).

jjm:
Thanks @Kami.

PR #2723, opened 2016-05-30
Access control and other new features and improvements for the datastore API endpoints

Kami:
This pull request includes access control and other new features, related improvements and fixes for  the datastore API endpoints.

1. Allow administrators to list all the datastore items (including user scoped ones) by specifying `all` for the `scope` query parameter.
2. When using decrypt option in the list API endpoint, require the user to be an administrator or the user is listing user-scoped items for themselves.
3. When using decrypt option in the get one API endpoint, require the user to be an administrator or the user is retrieving a user-scoped item for themselves.
4. Allow administrator to set a user-scoped item for an arbitrary user.
5. Allow administrator to delete a user-scoped item for an arbitrary user.
6. Some code cleanup (we don't need to catch `StackStormDBObjectNotFoundError` exceptions, jsexpose hooks does that and maps those errors to 404).
7. Fix get all items API endpoint - when retrieving user-scoped items, non-admin user should only be able to view / retrieve items scoped to themselves.
7. Some CLI improvements and fixes (add missing scope argument to `st2 key delete` command, etc.)
9. Various other improvements and fixes

## TODO

- [x] More tests
- [x] Some code refactor

### DIFF
--- a/st2api/st2api/controllers/v1/keyvalue.py
+++ b/st2api/st2api/controllers/v1/keyvalue.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import pecan
 from pecan import abort
 import six
 from mongoengine import ValidationError
@@ -28,6 +29,8 @@ from st2common.persistence.keyvalue import KeyValuePair
 from st2common.services import coordination
 from st2common.services.keyvalues import get_key_reference
 from st2common.util.api import get_requester
+from st2common.exceptions.rbac import AccessDeniedError
+from st2common.rbac.utils import request_user_is_admin
 
 http_client = six.moves.http_client
 
@@ -63,6 +66,7 @@ class KeyValuePairController(ResourceController):
             Handle:
                 GET /keys/key1
         """
+        # either admin or user requesting it for self
         self._validate_scope(scope=scope)
         key_ref = get_key_reference(scope=scope, name=name, user=get_requester())
         from_model_kwargs = {'mask_secrets': not decrypt}
@@ -86,6 +90,15 @@ class KeyValuePairController(ResourceController):
             Handles requests:
                 GET /keys/
         """
+        requester_user = get_requester()
+
+        # Note: Decrypt option requires admin access or listing datstore values
+        # for current user( (scope == user and user == authenticated user)
+        is_admin = request_user_is_admin(request=pecan.request)
+        if decrypt and (scope != USER_SCOPE and not is_admin):
+            msg = 'decrypt option requires administrator access'
+            raise AccessDeniedError(message=msg, user_db=requester_user)
+
         from_model_kwargs = {'mask_secrets': not decrypt}
         kwargs['prefix'] = prefix
         if scope:
@@ -94,7 +107,7 @@ class KeyValuePairController(ResourceController):
 
         if scope == USER_SCOPE and kwargs['prefix']:
             kwargs['prefix'] = get_key_reference(name=kwargs['prefix'], scope=scope,
-                                                 user=get_requester())
+                                                 user=requester_user)
 
         kvp_apis = super(KeyValuePairController, self)._get_all(from_model_kwargs=from_model_kwargs,
                                                                 **kwargs)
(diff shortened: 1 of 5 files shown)


st2api/st2api/controllers/v1/keyvalue.py
```diff
@@ -55,26 +59,37 @@ def __init__(self):
...
+
+        requester_user = get_requester()
+        user = user or requester_user
+        is_admin = request_user_is_admin(request=pecan.request)
+
+        # User needs to be either admin or requesting item for itself
+        self._validate_decrypt_query_parameter(decrypt=decrypt, scope=scope, is_admin=is_admin)
+
+        # Validate that the authenticated user is admin if user query param is provided
+        self._validate_user_query_parameter(user=user, is_admin=is_admin)
```
lakshmi-kannan: Shouldn't this validation happen before 81?

st2api/st2api/controllers/v1/keyvalue.py
```diff
@@ -55,26 +59,37 @@ def __init__(self):
...
-        try:
-            kvp_api = self._get_one_by_scope_and_name(
-                name=key_ref,
-                scope=scope,
-                from_model_kwargs=from_model_kwargs
-            )
-        except StackStormDBObjectNotFoundError as e:
-            abort(http_client.NOT_FOUND, e.message)
-            return
+        kvp_api = self._get_one_by_scope_and_name(
```
lakshmi-kannan: I don't know why the except clause was removed without a corresponding change in _get_one_by_scope_and_name? Am I missing something?

st2client/st2client/commands/keyvalue.py
```diff
@@ -135,6 +136,8 @@ def __init__(self, resource, *args, **kwargs):
         self.parser.add_argument('-s', '--scope', dest='scope', default=DEFAULT_SCOPE,
                                  help='Specify the scope under which you want ' +
                                       'to place the variable.')
+        self.parser.add_argument('-u', '--user', dest='user', default=None,
+                                 help='User for user scoped variables ')
```
lakshmi-kannan: Period at end. Also, note in help that it is admin access only.

st2api/st2api/controllers/v1/keyvalue.py
lakshmi-kannan: NVM. I see the description.

lakshmi-kannan:
Overall looks good. 

1. When I was about to work on this, I was wondering how the RBAC PermissionsResolver class would look like? It looks like you don't use and instead use custom methods in the controller. I am calling it out so everyone knows about that. 

2. I am not sure I got why you wanted KeyValuePairSetAPI as a separate model. I can only guess. 

I think generally the code & the feature is progressing nicely. So 👍

st2api/st2api/controllers/v1/keyvalue.py
lakshmi-kannan: After reading the methods, I think it's safer to do this before 81.

Kami:
@lakshmi-kannan 

1. Yeah, decorator and resolver doesn't always work that nicely. It mostly works fine when the resolver code only deals with the input (request) body, but in this case it mostly deals with request query parameters and we don't have a pattern to handle this yet with the resolver.

In any case, I would still prefer to move this into resolver classes at some point though (it's better to have all access control code in one place). For one, I could simply move "_validate" methods on a new resolver class and call that (decorator won't always work and it's always a good fit anyways) inside the API controller methods.

2. Should have done this in the past already, but at that time we haven't established that pattern yet (different API model for different operations).

The problem is that "set" endpoint takes some additional parameters and abusing `KeyValuePairAPI` model which is only meant for output is not good. Anyways, this deals with the whole "input is not always the same as output"  problem we have identified (and need a better approach for) in the past.

st2api/st2api/controllers/v1/keyvalue.py
Kami: You mean before `self._validate_decrypt_query_parameter` call?

If so, I don't think the order really matters that much.

Kami:
Re RBAC resolver refactor - we have extensive test cases for API endpoints now for (hopefully I didn't miss something) all the edge cases, so I will probably go ahead, merge this into master and do a refactor in a separate PR.

Having the tests means the refactor shouldn't be too painful.

PR #2724, opened 2016-05-31
Add new actions for IP address valuation to Linux Pack.

jjm:
Complete!

- Add ipaddress module to requirements.txt in linux pack.
- Add new actions is_valid_ip, is_valid_ipv4 and is_valid_ipv6
  to linux pack.
- Add tests for is_valid_ip.py action script.
- Add entry to for new actions to the CHANGELOG.rst.


contrib/linux/requirements.txt
```diff
@@ -1,3 +1,4 @@
 # used by file watcher sensor
 pyinotify>=0.9.5,<=0.10
 -e git+https://github.com/Kami/logshipper.git@stackstorm_patched#egg=logshipper
+ipaddress>=1.0.16
```
Kami: IIRC (but I need to double check how new packages handle that) we need to also put this in global requirements.txt file.

Kami: (Just thinking out loud) - Module license is Python Software Foundation License v2 (https://github.com/phihag/ipaddress/blob/master/LICENSE) so it's OK for us to bundle this dependency.

contrib/linux/actions/is_valid_ip.py
```diff
@@ -0,0 +1,62 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import ipaddress
+
+from st2actions.runners.pythonrunner import Action
+
+
+class IsValidIp(Action):
```
Kami: Minor style thing - we usually suffix action classes with `Action`.

Kami:
Thanks.

The changes look good to me :+1:

Kami:
Please also sync this branch with latest master since there are some conflicts it looks like.

jjm:
Sorry, I seem to have sync'ed our master branch and then branched from the wrong place :-( I'll fix.

Kami:
Since this package is a port of Python 3 functionality, I believe it should also work on Windows.

In that case, `core` pack is maybe a better place for this action. What do others think (in any case, it's not such a big deal right now since we don't really officially support running StackStorm on Windows anyway)?

/cc @lakshmi-kannan @manasdk

jjm:
I've corrected the Action suffix and brought my branch up to date, so should be fine to merge.

Kami:
Thanks.

I'll wait for the feedback on the pack and merge it after that :)

contrib/linux/requirements.txt
lakshmi-kannan: >IIRC (but I need to double check how new packages handle that) we need to also put this in global requirements.txt file.

Yep. That or move linux pack back to st2contrib.

lakshmi-kannan:
I like this in linux pack just fine but I have problems with linux pack being part of st2 code. I lost that argument before but I am not giving up.

lakshmi-kannan:
It appears no one has a problem moving linux pack back to st2contrib. Once this PR gets merged, I'll own moving it to st2contrib.

manasdk:
I am not very clear what these actions have to do anything with linux or with core StackStorm functionality. Would it make more sense to place them in st2contrib in a standalone pack called `ip_utils`?

jjm:
They are general utility / helper functions for use in other workflows,so following a quick discussion on Slack it was thought the linux pack in ST2 would be best location.

I think the real question is _Where do you put general propose helper actions?_ They are of use only in other workflows, if they are not the main st2 repo will folks be allowed to use them in other st2contrib workflows? As the normal review comment is often _don't depend on other packs_ if you use action from another pack.

My suggestion is move them to a new *utils* pack in st2, which purpose is useful helper actions for workflows. We can then add actions like the following:

- is_valid_hostname
- is_valid_dns_entry
- is_in_dns

Which would all be every useful for adding checks within workflows, I guess the other option wold be aiming to push them into Mistral? Which would reduce the load on the action runners..

These came from my need to add a check for a valid IP in one of our troubleshooting flows and I thought it made sense to share with the wider community as it's action I see others will need to do, so why not same them the effort.

Kami:
I'm personally fine with either moving it to a separate pack (e.g. `networking` / `ip_utils`) or leaving it in a `linux` pack.

We already have some "similar" functionality in Linux pack (e.g. wait_for_ssh) so I don't see it as a big problem.

Kami:
> I like this in linux pack just fine but I have problems with linux pack being part of st2 code. I lost that argument before but I am not giving up.

This also means updating installer script and other stuff to install this pack by default? Or?

manasdk:
> We already have some "similar" functionality in Linux pack (e.g. wait_for_ssh) so I don't see it as a big problem.

I guess we all draw our personal lines very differently. What is the similarity between `wait_for_ssh` and `is_valid_ip ` etc? The fact that it has something to do with ip addresses or is related to networking?

Kami:
> I guess we all draw our personal lines very differently. What is the similarity between wait_for_ssh and is_valid_ip etc? The fact that it has something to do with ip addresses or is related to networking?

Both.

Anyway, to unblock this PR, is everyone OK with moving this to new "networking" or similar pack in st2contrib?

jjm:
If you let me know where it would go, I'll open a new PR.

jjm:
How about *networking_utils* in st2contrib?

manasdk:
@jjim sounds good! New networking_utils it is

jjm:
I've moved this to StackStorm/st2contrib/#498.

jjm:
I've moved these actions to the networking-utils pack in StackStorm/st2contrib#498.

PR #2725, opened 2016-06-01
Move file to st2common/bin to work with st2-packages

manasdk:
* All tools we would like to ship actually live in st2common/bin
  and we don't care much about the tools except for internal dev

packaging : https://github.com/StackStorm/st2-packages/pull/307
docs : https://github.com/StackStorm/st2docs/pull/170


lakshmi-kannan:
lgtm

Kami:
:+1:

Issue #2686, opened 2016-05-17
[RFC] [PROPOSAL] API version 2


Kami:
## Background

From the end-user perspective, API v1 is not too bad (there are definitely worse RESTful APIs out there), but from developer and maintainability perspective, API v1 is quite horrible (it has been causing us a lot of pain).

A lot of that is related to our (ab)use of Pecan. We use Pecan in ways it was never really meant to be used. This results in a lot of hacks (jsexpose is a terrible monstrosity) and hard-to-maintain controller code. In addition to that, using Pecan ties us to a particular not-too-restful URL scheme (in theory, we could actually make it work with any kind of URL scheme we want, but this would require horrible hacks to the code and code would be even harder to follow and maintain).
## Goals / requirements

New API should address the following things:
- Nicer and more "RESTful" paths (e.g. `/v1/packs/views/files/<pack name>` vs `/v2/packs/<pack name>/files`, `/v1/config_schema/<pack name>` vs `/v2/packs/<pack name>/config_schema`, etc.)
- Controller related code should be easy understand and to maintain
- It should be possible and easy to auto-generate API documentation from the code and specification
- It should be possible and easy to auto-generate low-level API clients for multiple programming languages. We might still want to write more idiomatic and higher-level client libraries ourselves, but auto-generated ones should be a good start.
## Open questions and Implementation details

There are some open questions and implementation details we need to decide about:
1. Should we switch away from Pecan (we also need to check the latest version since some things have been improved)? What are the good alternatives for light-weight API services?

I've use Django, Flask and Tornado for such purposes in the past. Django is obviously not a good fit and over-kill and Tornado requires Tornado-specific async code which is no go for us since we are already bought into the eventlet ecosystem.

Flask might be an OK middle-ground, but I still think it's a slight over-kill and too bloated for simple API services.
1. Should we use some kind of framework (e.g. Swagger or grpc) which allows us to describe API schema and then auto generates client libraries and docs?

I do think we should consider using something like Swagger or grpc. Not matter what kind of approach we go with, having some kind of description / "schema" for API is a good thing. Manually writing low-level API clients and API docs is error prone and a waste of time (it's too easy for things to get out of sync, hard to maintain, etc.).

Using something like Swagger / Thrift / grpc would potentially also allow us to version models and API responses, but I think that's out of scope of this change (doing all of that mentioned above is already quite a lot of work and we can't do and fix everything as part of this change).

[...]

PR #2727, opened 2016-06-02
REST Services v2

enykeev:
As I said in #2686, refactoring st2api and starting to work on API v2 is essentially two different tasks that might be done at the same time, but doesn't have to. This PR shows an example of how we can make our web services easier to understand and to work with.

As we continue to add more different endpoints to the API, we figured our current approach of using `@jsexpose` decorator is "too magical" and requires us to dive deep into pecan code to figure out what's going on.

Web frameworks (pecan, flask, any other really) are full of features to appease a wide variety of users with even wider variety of needs (templating, out the box caching solutions, static serving, logging support, you name it). We've already figured out what exactly do we need and mostly reimplemented it on top of pecan anyway.

At the same time, we've reached the point where we want to stabilize our API and make it easier for both us and our users to work with it. We want to have a single place we can define our API and then generate routing table, docs, clients and tests from it.

Instead of monolithic framework, I propose to use a number of smaller single purpose libraries and build our router and controller around them. In this case, I've replaced pecan with three libraries:
- **WebOb** provides a number of user-friendly utility functions to work with HTTP requests and responses. For example, it allows you to access json-encoded request data using `req.json()` instead of parsing it manually from WSGI `environ` object. It also allows you to construct the response and only call `start_response` at the last possible moment.
- **Routes** handles URL matching and templating allowing us to comply with OpenAPI spec regarding path templating without diving deep into regex replacements.
- **swagger-spec-validator** handles spec validation and helps us spot an error in our spec during initialization

Some features that are missing in this implementation:
- Versioning. As I see it, we should have a separate `spec.yml` for every version with [basePath](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields) field defining a proper prefix. Default (non-prefixed) version should be defined by router argument.
- Request parameter validation and type casting. Parameter spec is JSONSchema-complaint so it's pretty straightforward. We also have a [code](https://github.com/StackStorm/st2/blob/master/st2common/st2common/models/api/base.py#L224-L232) for type casting already, but I have to admit, it was one of this confusing parts of `jsexpose` so maybe we should just do it inside controller instead.
- Response validation. OpenAPI has a section defining possible responses for the endpoint. The question is how should we react when this validation had failed. One hand, we're breaking our own convention and returning an answer user don't expect to get. On the other, all the work that should have been done during the request is already done and returning an error to the user only hides the results of this work.
- Using spec definitions as API models. They are serving the same purpose, really.
- Unit and integration tests based on spec. We have schemes for both request and response parameters. Generating request from the first and matching the response with the second makes a very cheap test which can later be extended with more complex logic when it makes sense.
- Automatic reloading for spec and\or controller code. Since the service is much more decoupled now, we can reload controller modules and specs without reloading the whole service. Productivity impact will depend on personal preferences, though.
- Expose spec files on a separate endpoint. st2web and st2client may became much more flexible if they will be able to determine API models and endpoints on the fly instead of relying on hardcoded values. If we put an effort to it, we can really reduce the number of places you need to make a change to implement particular feature.


enykeev:
As you see, I've tried `flask` and `connexion` first but figured you can't redefine error reporting in this case.

With this approach, we not only having smaller codebase while keeping full control, we can also transition our services to the new spec\router one controller at a time. And if we are willing to play safe, it won't be too hard to ensure responses for existing API won't change at all.

manasdk:
I like the overall approach. It is very clear and does not involve a great deal of magic :)

It is a shame that we would need to write `st2common/st2common/router.py` but perhaps that is the cost of picking specific libraries.

lakshmi-kannan:
TBH, I am unable to make up my mind. The custom 3 piece thing seems to work but I am unsure what issues we'll hit when we migrate all the endpoints. The problem is we won't know unless we do it. One thing though is libraries like routes aren't super active in terms of contributions as opposed to things like flask. So even if flask seems like excessive or is missing out on some features, it seems more battle tested.

enykeev:
As for being battle tested, both WebOb and routes are used by reddit and are also a parts of other frameworks (pecan uses webob and routes is a part of pylons) so they are getting plenty. 

> The problem is we won't know unless we do it.

Yep, we don't. But we can mitigate the risk by transitioning one controller at a time, first checking the request against controllers that have already been rewritten and then falling back to pecan if it could not be resolved there. There is nothing in Pecan that we can't do ourselves inside the controller or as a middleware on top of it, but even if we find something complex enough, it won't invalidate all the effort we already put into the transition.

Simme212:
[5ad542588c3c0117273e8f8975eb9027-4fcc6f65b928f293c80c60dc1ac0c3a791502784 (1).zip](https://github.com/StackStorm/st2/files/417111/5ad542588c3c0117273e8f8975eb9027-4fcc6f65b928f293c80c60dc1ac0c3a791502784.1.zip)

enykeev:
I've force-pushed changes from `feature/openapi` here to avoid creating duplicate PR. This is how test coverage looks at the moment:

```
(virtualenv)vagrant@arkham:/dock/st2$ for file in st2api/tests/unit/controllers/v1/test_*.py; do nosetests $file > /dev/null 2>&1 && echo $file ok || echo $file fail; done
st2api/tests/unit/controllers/v1/test_action_alias.py fail
st2api/tests/unit/controllers/v1/test_actions.py ok
st2api/tests/unit/controllers/v1/test_actions_rbac.py fail
st2api/tests/unit/controllers/v1/test_action_views.py ok
st2api/tests/unit/controllers/v1/test_alias_execution.py fail
st2api/tests/unit/controllers/v1/test_alias_execution_rbac.py fail
st2api/tests/unit/controllers/v1/test_auth_api_keys.py ok
st2api/tests/unit/controllers/v1/test_auth.py fail
st2api/tests/unit/controllers/v1/test_base.py fail
st2api/tests/unit/controllers/v1/test_executions_filters.py ok
st2api/tests/unit/controllers/v1/test_executions_simple.py ok
st2api/tests/unit/controllers/v1/test_jsexpose_decorator.py fail
st2api/tests/unit/controllers/v1/test_kvps.py fail
st2api/tests/unit/controllers/v1/test_kvps_rbac.py fail
st2api/tests/unit/controllers/v1/test_pack_config_schema.py fail
st2api/tests/unit/controllers/v1/test_pack_configs.py fail
st2api/tests/unit/controllers/v1/test_packs.py fail
st2api/tests/unit/controllers/v1/test_packs_views.py fail
st2api/tests/unit/controllers/v1/test_policies.py fail
st2api/tests/unit/controllers/v1/test_rbac_for_supported_endpoints.py fail
st2api/tests/unit/controllers/v1/test_rule_enforcements.py fail
st2api/tests/unit/controllers/v1/test_rules.py ok
st2api/tests/unit/controllers/v1/test_rules_rbac.py fail
st2api/tests/unit/controllers/v1/test_rule_views.py ok
st2api/tests/unit/controllers/v1/test_runnertypes.py ok
st2api/tests/unit/controllers/v1/test_runnertypes_rbac.py fail
st2api/tests/unit/controllers/v1/test_sensortypes.py ok
st2api/tests/unit/controllers/v1/test_timers.py fail
st2api/tests/unit/controllers/v1/test_traces.py fail
st2api/tests/unit/controllers/v1/test_triggerinstances.py fail
st2api/tests/unit/controllers/v1/test_triggers.py fail
st2api/tests/unit/controllers/v1/test_triggertypes.py fail
st2api/tests/unit/controllers/v1/test_webhooks.py fail
```

Besides converting all the endpoints to the OpenAPI spec, there's two major points under consideration:
 - current implementation of RBAC doesn't work for OpenAPI as it heavily relies on positional arguments and method names neither of which has any value in context of OpenAPI spec. We have other reasons to want to review our approach to RBAC so it's hard for me to justify spending time on making current implementation work.
 - OpenAPI provides means for validation not only input data, but also endpoint's response. Problem is, by the time the response is ready for validation, all the work related to that response was already completed and in most cases can not be easily undone. Raising an error in such circumstances doesn't look right, but ignoring the fact response doesn't match the scheme isn't ok either as API consumers would rely on the spec. The only option I see is to set a header that would warn consumer of response being invalid. We would then check for this header in our tests to make sure we don't get a regression. I'm open for other suggestions, though.

enykeev:
Notable changes to controllers:
- Pecan expects controller to be a class of certain methods (get_one, post, delete). OpenAPI deals with functions. Instead of instantiating controller classes implicitly during initialization as Pecan does it, OpenAPI router expects `operationId` to point to a static method of a class or a method of a class instance. See [actions.py#L316](https://github.com/StackStorm/st2/blob/1791d8b56542cbbede0d0c67c78259e2d198e7f8/st2api/st2api/controllers/v1/actions.py#L316) and [openapi.yaml#L12](https://github.com/StackStorm/st2/blob/1791d8b56542cbbede0d0c67c78259e2d198e7f8/st2api/st2api/controllers/openapi.yaml#L12) for example.
- `jsexpose` decorator is out. We have no use for it anymore. See [actions.py#L103](https://github.com/StackStorm/st2/blob/1791d8b56542cbbede0d0c67c78259e2d198e7f8/st2api/st2api/controllers/v1/actions.py#L103)
- Controller functions should return an instance webob.Response (see [actions.py#L146-L149](https://github.com/StackStorm/st2/blob/1791d8b56542cbbede0d0c67c78259e2d198e7f8/st2api/st2api/controllers/v1/actions.py#L146-L149)). While they may still return arbitrary python object as a response (which will be implicitly converted to webob.Response by router [router.py#L264-L265](https://github.com/StackStorm/st2/blob/1791d8b56542cbbede0d0c67c78259e2d198e7f8/st2common/st2common/router.py#L264-L265)), there is no way to set status code or headers on the response.
- If controller response need to be modified by some kind of middleware before it will be returned to the user, said middleware would have to repackage the response object. See [resource.py#L414-L419](https://github.com/StackStorm/st2/blob/1791d8b56542cbbede0d0c67c78259e2d198e7f8/st2api/st2api/controllers/resource.py#L414-L419) for example.
- RBAC decorators should be commented out until we figure out what to do with them ([actions.py#L87](https://github.com/StackStorm/st2/blob/1791d8b56542cbbede0d0c67c78259e2d198e7f8/st2api/st2api/controllers/v1/actions.py#L87)). There's relatively good chance that we would need them to be exposed in OpenAPI spec so we should probably start adding them there now ([openapi.yaml#L13](https://github.com/StackStorm/st2/blob/1791d8b56542cbbede0d0c67c78259e2d198e7f8/st2api/st2api/controllers/openapi.yaml#L13)).
- All mentions of pecan should be removed from controllers and replaced with OpenAPI analogs. `pecan.response` should be integrated to webob.Response (replacing [this](https://github.com/StackStorm/st2/pull/2727/files#diff-0d66808a8d3531eaee33359ca1826173L188) with [that](https://github.com/StackStorm/st2/pull/2727/files#diff-0d66808a8d3531eaee33359ca1826173R216)). `pecan.request` should be exposed through OpenAPI parameters or x-parameters.

Later, I want us to think about getting rid of controller classes completely as it might provide better readability. Doing it now as a part of this PR will just complicate the diff.

enykeev:
Note: despite the fact that tests are passing on webhook endpoint, it requires additional attention due to lacking coverage and complex design. Avoid merging the PR until the problems are addressed.

st2api/st2api/controllers/v1/keyvalue.py
```diff
@@ -116,21 +117,22 @@ def get_all(self, prefix=None, scope=FULL_SYSTEM_SCOPE, user=None, decrypt=False
             scope = FULL_USER_SCOPE
 
         scope = get_datastore_full_scope(scope)
-        requester_user = get_requester()
+        requester_user = cfg.CONF.system_user.user
```
enykeev: I have a feeling you've simplified this part too much. In this controller, we're dealing with two kinds of user: one requests the key and another one is on behalf of whom the key is requested. You provide username for the second one through `user` query parameter. First one, according to `get_requester()` is either user from auth context (one you've authenticated with) or system_user (stanley) for the `auth=false` case.

To fix that, you would need to get second user from WSGI context like https://github.com/StackStorm/st2/blob/feature/rest_services_v2/st2api/st2api/controllers/openapi.yaml#L408-L411. Problem is, now they have conflicting variable names.

enykeev:
```
(virtualenv)vagrant@arkham:/dock/st2$ for file in st2api/tests/unit/controllers/v1/test_*.py; do nosetests $file > /dev/null 2>&1 && echo $file ok || echo $file fail; done
st2api/tests/unit/controllers/v1/test_action_alias.py ok
st2api/tests/unit/controllers/v1/test_actions.py ok
st2api/tests/unit/controllers/v1/test_actions_rbac.py ok
st2api/tests/unit/controllers/v1/test_action_views.py ok
st2api/tests/unit/controllers/v1/test_alias_execution.py ok
st2api/tests/unit/controllers/v1/test_alias_execution_rbac.py ok
st2api/tests/unit/controllers/v1/test_auth_api_keys.py ok
st2api/tests/unit/controllers/v1/test_auth.py ok
st2api/tests/unit/controllers/v1/test_base.py fail
st2api/tests/unit/controllers/v1/test_executions_filters.py ok
st2api/tests/unit/controllers/v1/test_executions_simple.py ok
st2api/tests/unit/controllers/v1/test_kvps.py ok
st2api/tests/unit/controllers/v1/test_kvps_rbac.py ok
st2api/tests/unit/controllers/v1/test_pack_config_schema.py ok
st2api/tests/unit/controllers/v1/test_pack_configs.py ok
st2api/tests/unit/controllers/v1/test_packs.py ok
st2api/tests/unit/controllers/v1/test_packs_views.py ok
st2api/tests/unit/controllers/v1/test_policies.py ok
st2api/tests/unit/controllers/v1/test_rbac_for_supported_endpoints.py ok
st2api/tests/unit/controllers/v1/test_rule_enforcements.py ok
st2api/tests/unit/controllers/v1/test_rules.py ok
st2api/tests/unit/controllers/v1/test_rules_rbac.py ok
st2api/tests/unit/controllers/v1/test_rule_views.py ok
st2api/tests/unit/controllers/v1/test_runnertypes.py ok
st2api/tests/unit/controllers/v1/test_runnertypes_rbac.py ok
st2api/tests/unit/controllers/v1/test_sensortypes.py ok
st2api/tests/unit/controllers/v1/test_timers.py ok
st2api/tests/unit/controllers/v1/test_traces.py ok
st2api/tests/unit/controllers/v1/test_triggerinstances.py ok
st2api/tests/unit/controllers/v1/test_triggers.py ok
st2api/tests/unit/controllers/v1/test_triggertypes.py ok
st2api/tests/unit/controllers/v1/test_webhooks.py ok
```

It seems like there's two more endpoints that are not covered by tests (rbac.py and ruletypes.py) and at least one more that's not covered by rbac tests (auth.py).

st2api/st2api/app.py
```diff
@@ -67,34 +49,22 @@ def setup_app(config=None):
...
+    }
+    router.add_spec(spec, transforms=transforms)
 
-    # Static middleware which servers common static assets such as logos
-    static_root = os.path.join(BASE_DIR, 'public')
-    app = StaticFileMiddleware(app=app, directory=static_root)
+    app = router.as_wsgi
 
-    LOG.info('%s app created.' % __name__)
+    app = CorsMiddleware(app)
```
Kami: I like this better.

I assume order matters, same as before?

st2api/st2api/config.py
```diff
@@ -64,6 +64,7 @@ def _register_app_opts():
     CONF.register_opts(pecan_opts, group='api_pecan')
 
     logging_opts = [
+        cfg.BoolOpt('debug', default=False),
```
Kami: There should already be existing system wide debug option (IIRC, `system.debug` which also gets set when `--debug` CLI flag is used).

st2api/st2api/controllers/base.py
```diff
@@ -71,24 +66,18 @@ def _get_query_param_value(self, request, param_name, param_type, default_value=
 
         return value
 
-    def _get_mask_secrets(self, request):
+    def _get_mask_secrets(self, requester_user, show_secrets=None):
```
Kami: Do we need to distinguish between None and False for `show_secrets`?

st2api/st2api/controllers/exp/validation.py
```diff
@@ -13,28 +13,28 @@
...
 
-    @pecan.expose(content_type='application/json')
-    def post(self):
-        def_yaml = pecan.request.text
+    def post(self, def_yaml):
         result = self.validator.validate(def_yaml)
-        return jsonify.json_encode(result)
+
+        for error in result:
+            if not error.get('path', None):
```
Kami: Just curious why we need this piece of code - it looks like previously it wasn't needed? Or that happened somewhere else?

st2api/st2api/controllers/resource.py
```diff
@@ -171,92 +187,87 @@ def _get_all(self, exclude_fields=None, sort=None, offset=0, limit=None, query_o
...
-        return result
+        resp = Response(json=result)
+        resp.headers['X-Total-Count'] = str(instances.count())
+        if limit:
+            resp.headers['X-Limit'] = str(limit)
+
+        return resp
 
-    def _get_one(self, id, exclude_fields=None):
+    def _get_one(self, id, exclude_fields=None, **kwargs):
```
Kami: Oh no, not kwargs again :P

Looking at the code below, I would prefer to call it `from_model_kwargs` or something more explicit and making it a dict.

st2api/st2api/controllers/resource.py
```diff
@@ -171,92 +187,87 @@ def _get_all(self, exclude_fields=None, sort=None, offset=0, limit=None, query_o
...
         # Note: This is here for backward compatibility reasons
-        return self._get_one_by_id(id=id, exclude_fields=exclude_fields)
+        return self._get_one_by_id(id=id, exclude_fields=exclude_fields, **kwargs)
 
-    def _get_one_by_id(self, id, exclude_fields=None, from_model_kwargs=None):
+    def _get_one_by_id(self, id, exclude_fields=None, from_model_kwargs=None,
+                       requester_user=None, permission_type=None, **kwargs):
```
Kami: If `None` valid value for `permission_type` aka this argument is always required?

If this argument is mandatory we should throw or similar when it's not provided to catch programmer errors as early as possible.

st2api/st2api/controllers/resource.py
```diff
@@ -171,92 +187,87 @@ def _get_all(self, exclude_fields=None, sort=None, offset=0, limit=None, query_o
...
         from_model_kwargs = from_model_kwargs or {}
-        from_model_kwargs.update(self._get_from_model_kwargs_for_request(request=pecan.request))
+        from_model_kwargs.update(self._get_from_model_kwargs_for_request(**kwargs))
         result = self.model.from_model(instance, **from_model_kwargs)
-        LOG.debug('GET %s with id=%s, client_result=%s', pecan.request.path, id, result)
 
         return result
 
-    def _get_one_by_name_or_id(self, name_or_id, exclude_fields=None, from_model_kwargs=None):
+    def _get_one_by_name_or_id(self, name_or_id, exclude_fields=None, from_model_kwargs=None,
```
Kami: Same here wrt to `kwargs` and `permission_type`.

st2api/st2api/controllers/resource.py
```diff
@@ -171,92 +187,87 @@ def _get_all(self, exclude_fields=None, sort=None, offset=0, limit=None, query_o
...
         if not instance:
             msg = 'Unable to identify resource with name_or_id "%s".' % (name_or_id)
-            pecan.abort(http_client.NOT_FOUND, msg)
+            abort(http_client.NOT_FOUND, msg)
 
         from_model_kwargs = from_model_kwargs or {}
-        from_model_kwargs.update(self._get_from_model_kwargs_for_request(request=pecan.request))
+        from_model_kwargs.update(self._get_from_model_kwargs_for_request(**kwargs))
         result = self.model.from_model(instance, **from_model_kwargs)
-        LOG.debug('GET %s with name_or_id=%s, client_result=%s', pecan.request.path, id, result)
```
Kami: Does this logging happen on the app / router level now or why did we remove it?

(I do prefer to have it centralized, but just want to make sure it's still logged somewhere)

st2api/st2api/controllers/resource.py
Kami: Looking at the code below I now see `from_model_kwargs` and `kwargs` are two different, albeit related thing.

Having said that, I would still prefer to have some more explicit name, perhaps `request_kwargs` or similar to make it obvious it's referring to additional arguments which can be passed in via request.

st2api/st2api/controllers/v1/actionalias.py
```diff
@@ -67,18 +63,15 @@ def _match_tuple_to_dict(self, match):
...
-    @request_user_has_permission(permission_type=PermissionType.ACTION_ALIAS_LIST)
-    @jsexpose()
     def get_all(self, **kwargs):
         return super(ActionAliasController, self)._get_all(**kwargs)
 
-    @request_user_has_resource_db_permission(permission_type=PermissionType.ACTION_ALIAS_VIEW)
-    @jsexpose(arg_types=[str])
-    def get_one(self, ref_or_id):
-        return super(ActionAliasController, self)._get_one(ref_or_id)
+    def get_one(self, ref_or_id, requester_user):
```
Kami: I personally liked the decorator approach more for checking permissions - it was more explicit and obvious.

I'm also fine with current approach as long as we have some "easy" way to (automatically) validate / ling that all the methods apply some permission check (previously that was easy to verify by checking that decorator was in place).

st2api/st2api/controllers/v1/actionalias.py
```diff
@@ -104,80 +97,85 @@ def match(self, action_alias_match_api, **kwargs):
...
-    @jsexpose(arg_types=[str, str, int], body_cls=ActionAliasHelpAPI,
-    status_code=http_client.ACCEPTED)
-    def help(self, action_alias_help_api, **kwargs):
+    def help(self, filter, pack, limit, offset, **kwargs):
         """
             Get available help strings for action aliases.
 
             Handles requests:
-                POST /actionalias/help
+                GET /actionalias/help
```
Kami: :+1:

st2api/st2api/controllers/v1/actionalias.py
```diff
@@ -104,80 +97,85 @@ def match(self, action_alias_match_api, **kwargs):
...
             LOG.exception('Command "%s" matched (%s) patterns.', e.command, len(e.matches))
-            pecan.abort(http_client.BAD_REQUEST, str(e))
-            return [self._match_tuple_to_dict(match) for match in e.matches]
+            return abort(http_client.BAD_REQUEST, str(e))
 
-    @request_user_has_permission(permission_type=PermissionType.ACTION_ALIAS_HELP)
-    @jsexpose(arg_types=[str, str, int], body_cls=ActionAliasHelpAPI,
-    status_code=http_client.ACCEPTED)
-    def help(self, action_alias_help_api, **kwargs):
+    def help(self, filter, pack, limit, offset, **kwargs):
```
Kami: Yeah, I like those arguments being explicitly declared more more than previous approach.

Having said that, those `limit` and `offset` arguments / query params should really be global since they apply to every "list" / get_all controller, right?

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -86,32 +85,35 @@ class ActionExecutionsControllerMixin(BaseRestControllerMixin):
...
+        return {'mask_secrets': not kwargs.get('show_secrets', False)}
 
-    def _handle_schedule_execution(self, liveaction_api):
+    def _handle_schedule_execution(self, liveaction_api, requester_user=None, **kwargs):
         """
         :param liveaction: LiveActionAPI object.
         :type liveaction: :class:`LiveActionAPI`
         """
 
+        if not requester_user:
```
Kami: Can we use some utility method for that (IIRC we already use it in some other place).

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -126,7 +128,7 @@ def _handle_schedule_execution(self, liveaction_api):
             LOG.exception('Unable to execute action. Unexpected error encountered.')
             abort(http_client.INTERNAL_SERVER_ERROR, str(e))
 
-    def _schedule_execution(self, liveaction, user=None):
+    def _schedule_execution(self, liveaction, user=None, **kwargs):
```
Kami: Not a fan of this kwargs approach since it's not explicit which arguments can be present here.

I know this is not your fault because previously code did a very similar thing, it just didn't use `kwargs`, but it would still be better if we can come up with some more explicit and obvious approach.

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -277,17 +294,35 @@ def validate(self):
...
+    def post(self, spec_api, id, requester_user=None, no_merge=False, **kwargs):
         """
         Re-run the provided action execution optionally specifying override parameters.
 
         Handles requests:
 
             POST /executions/<id>/re_run
         """
+
+        if (spec_api.tasks or spec_api.reset) and spec_api.parameters:
```
Kami: It seems like you moved this outside API model `.validate()` method to the controller method?

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -401,26 +433,30 @@ def get_one(self, id, exclude_attributes=None, **kwargs):
...
+    def delete(self, id, requester_user, **kwargs):
         """
         Stops a single execution.
 
         Handles requests:
             DELETE /executions/<id>
 
         """
-        execution_api = self._get_one(id=exec_id)
+        if not requester_user:
```
Kami: Yeah, should probably use a utility function for that since we do a similar thing in multiple places.

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -445,19 +481,15 @@ def delete(self, exec_id):
 
         try:
             (liveaction_db, execution_db) = action_service.request_cancellation(
-                liveaction_db, get_requester())
+                liveaction_db, requester_user.name or cfg.CONF.system_user.user)
```
Kami: Same here - I assume you had some specific reason to replace utility function with this line?

st2api/st2api/controllers/v1/actions.py
```diff
@@ -141,14 +137,17 @@ def post(self, action):
         LOG.audit('Action created. Action.id=%s' % (action_db.id), extra=extra)
         action_api = ActionAPI.from_model(action_db)
 
-        return action_api
+        return Response(json=action_api, status=http_client.CREATED)
```
Kami: I like this, much more explicit what is going on.

st2api/st2api/controllers/v1/actionviews.py
```diff
@@ -98,32 +95,36 @@ class OverviewController(resource.ContentPackResourceController):
...
+        action_apis = resp.json
+        result = map(self._transform_action_api, action_apis)
+        resp.json = result
+        return resp
 
     @staticmethod
     def _transform_action_api(action_api):
-        action_id = action_api.id
-        action_api.parameters = ParametersViewController._get_one(action_id).get('parameters')
+        action_id = action_api['id']
```
Kami: I noticed this in other places as well - does this need to be a dict or it can be an object (aka specific class instance)?

I would much more prefer it to be an object - this way it's much easier to do static analysis and similar things (IDE auto-completion, etc.) and see what attributes are available.

I'm really not a fan of dicts, especially when dealing with objects like our API model which have a fixed limited set of possible attributes.

"Magical" dicts make programmer's life much harder and also make it easier to introduce bugs which would otherwise be caught if we used an object or similar.

I didn't dig in too much yet, but I assume that's swaggers thing (please correct me if I'm wrong) and I don't like it. I like what protobufs, thrift and other similar frameworks do much more - "definitions" are compiled and models are automatically generated from the definitions. This way it's same as any other static class definition in the code. It involves no magic, it makes static analysis and other things possible.

st2api/st2api/controllers/v1/packviews.py
```diff
@@ -174,8 +172,7 @@ class FileController(BaseFileController):
     Controller which allows user to retrieve content of a specific file in a pack.
     """
 
-    @expose()
-    def get_one(self, ref_or_id, *file_path_components):
+    def get_one(self, ref_or_id, file_path, requester_user, **kwargs):
```
Kami: Great, can find get rid of this nasty pecan file_path workaround.

st2api/st2api/controllers/v1/triggers.py
```diff
@@ -395,7 +388,13 @@ def get_all(self, **kw):
         return trigger_instances
 
     def _get_trigger_instances(self, **kw):
-        kw['limit'] = int(kw.get('limit', self.default_limit))
+        kw['limit'] = int(kw.get('limit') or self.default_limit)
```
Kami: Ah, so it looks like now it contains `None` if value is not supplied by the user, or?

st2api/st2api/controllers/v1/webhooks.py
```diff
@@ -119,70 +106,41 @@ def get_one(self, name):
...
-
-        try:
-            body = self._parse_request_body(content_type=content_type, body=body)
-        except Exception as e:
-            self._log_request('Failed to parse request body: %s.' % (str(e)), pecan.request)
-            msg = 'Failed to parse request body "%s": %s' % (body, str(e))
-            return pecan.abort(http_client.BAD_REQUEST, msg)
-
-        headers = self._get_headers_as_dict(pecan.request.headers)
+    # @request_user_has_webhook_permission(permission_type=PermissionType.WEBHOOK_SEND)
```
Kami: It looks like we are not performing RBAC check here anymore? Or am I missing something?

st2api/st2api/controllers/v1/actionviews.py
enykeev: Not that I had a moment to check the code you're commenting, this particular place have an asterisk to it, really. We accessing it as a dict and not an object because at that point we're already working not with user input, but with server response. Method `resp = super(OverviewController, self)._get_all(**kwargs)` already created a response for us and now we're trying to modify it.

Now, if you want to work with the model here, `._get_all` method should also return model, which is possible, yet would require us to make sure `.get_all` of every other controller in the whole st2api will convert it to Response at the end.

Although we still support controllers returning models, I find this practice highly detrimental to us better understanding how http stack works. Which was part of the reason I'm doing it to begin with. Pecan was complex and involved too many features we don't need and don't care. Router is short enough anyone on the team can read through it and understand what's going on in between of user making a request and receiving the answer to it. I consider understanding of the Router.__call__ method an absolute requirement for anyone touching an endpoint on this project.

st2api/st2api/controllers/v1/webhooks.py
enykeev: Yep, looks like I've missed rbac for webhooks

st2api/st2api/controllers/v1/actionexecutions.py
enykeev: this utility function expected to magically get `pecan.request` global ten levels into the call stack. This way you can clearly see where this requester comes from.

st2api/st2api/controllers/v1/actionexecutions.py
enykeev: personally, I don't think we do. Couple extra lines would not break the bank, but this way is much more expressive.

st2api/st2api/controllers/v1/actionexecutions.py
enykeev: At that point I wasn't sure I'll be using models anymore. I'm still not sure about that, honestly, I don't like us doing stuff like https://github.com/StackStorm/st2/blob/master/st2common/st2common/models/api/action.py#L109 and I see an explicit controller with no hidden logic as the way to achieve that. I do not plan to fix every problem in this PR, though, it's huge already.

st2api/st2api/controllers/v1/actionexecutions.py
enykeev: I'm using kwargs here as a replacement for pecan.request to use in methods like `_get_from_model_kwargs_for_request`. Fixing problems like this requires complete refactoring of st2api. It's not what I'm doing here.

st2api/st2api/controllers/v1/actionexecutions.py
enykeev: As I said, I see it more explicit this way. Having 2-line utility method instead of a simple if makes reading code harder and requires you fo constantly jump between the files.

st2api/st2api/controllers/v1/actionalias.py
enykeev: Nope, they absolutely should never be global. Pecan tried to play with global variables and we end up with things like get_requester() when you absolutely can not change the logic of the method without passing variable 10 levels down the stack.

Global variables are awful way of writing code and should be avoided unless absolutely required (read singleton).

st2api/st2api/controllers/v1/actionalias.py
enykeev: Well, that's simple. Every controller should have rbac definition of some kind. Let's make it a part of TestApp base class.

st2api/st2api/controllers/resource.py
enykeev: The reason I removed this particular line is that it uses pecan.request.path which can't be exposed through the means of openapi. Saying that, yes, we do have logging middleware https://github.com/StackStorm/st2/pull/2727/files/d802d6afbe760a4e05944214389d16616d963331#diff-f745fc7cee78f050f04933589c1924e9

st2api/st2api/controllers/resource.py
enykeev: `**kwargs` is a direct equivalent for `...rest` in es6. It means "everything that was defined in spec other than that I've described previously". The benefit of openapi is that it strictly defines the list of arguments your controller would get and there will be no surprises, no additional stuff router puts there on his own initative.

With this in mind, if you don't intend to use some of the arguments inside this method and instead are planning to pass it deeper into the stack, there's really no reason for you to be explicit there and write down every single parameter just so you can write them again in a next line and then change them on every spec change. You just pass `**kwarg` and let underlying function to pick what it needs out of the bucket.

st2api/st2api/controllers/exp/validation.py
enykeev: Scheme we validating against doesn't like path parameter to be None. Previously we implicitly defaulted them to ''.

st2api/st2api/controllers/base.py
enykeev: Why we may need to do that?

st2api/st2api/app.py
enykeev: Sure it does. If you look at how middleware work, they just call each other in sequence.

enykeev:
Just to point out, with the changes made, there's now a need for refactoring some of the controller methods and utility functions used. I tried to avoid getting into that out of fear of further inflate this PR. If you see some code inefficiencies and you know you can fix that with refactoring, don't comment on it, instead let us merge that think, ensure that everything that supposed to be working is indeed working, then we'll be fixing stuff that just plain ugly.

st2api/st2api/controllers/resource.py
enykeev: If I had to guess, we did it this way to make test pass. If `cfg.CONF.rbac.enable`, the assert would never raise so tests could get away with not providing permission type. Once again, some refactoring is required.

st2api/st2api/controllers/v1/actionexecutions.py
Kami: The problem with that is that API is not the only way for us to get data into the system.

Some resources are also registered using `st2-register-content` script, etc. and having some centralized place (in this case a model) where a common validation happens, etc. is better - this way we can be sure it's consistently applied everywhere where it needs to be.

Having said that, yes, this is not really an issue for `re_run` itself so I'm fine with it being the way it is, but it is when "creating" other resources.

st2api/st2api/controllers/v1/actionalias.py
Kami: With global I really meant something which is applied to every "get_all" and other list like methods.

The way I mentioned it is because I didn't see those arguments declared in some "get_all" methods - it's possible that I missed it or that it's hidden under "**kwargs".

st2api/st2api/controllers/resource.py
Kami: Yeah, but I would argue this makes it much harder for developer and people reading code - you don't really know which arguments can and will be available and you constantly need to refer back to the spec.

st2api/tests/unit/controllers/v1/test_timers.py
```diff
@@ -91,13 +97,13 @@ def setUpClass(cls):
...
 
-    def test_get_one_with_id(self):
+    def test_timerscontroller_get_one_with_id(self):
         model = TestTimersController.MODELS['interval1.yaml']
         get_resp = self._do_get_one(model.id)
         self.assertEqual(get_resp.status_int, 200)
         self.assertEqual(get_resp.json['parameters'], model['parameters'])
 
-    def test_get_one_with_ref(self):
+    def test_timerscontroller_get_one_with_ref(self):
```
Kami: Yeah, I like this naming scheme more - easier and fast to see in Circle CI output which test failing with a quick glace.

st2auth/st2auth/controllers/v1/auth.py
```diff
@@ -53,32 +49,41 @@ def post(self, request, **kwargs):
...
-                                         remote_addr=pecan.request.remote_addr,
-                                         remote_user=pecan.request.remote_user,
-                                         authorization=pecan.request.authorization,
+        headers = {}
+        if 'x-forwarded-for' in kwargs:
+            headers['x-forwarded-for'] = kwargs.pop('x-forwarded-for')
+
+        authorization = kwargs.pop('authorization', None)
+        if authorization:
+            authorization = tuple(authorization.split(' '))
```
Kami: I don't think base64 encoded string can contain a whitespace, but maybe still safer to split on the first whitespace aka `.split(' ', 1)`.

st2common/st2common/hooks.py
```diff
@@ -1,376 +0,0 @@
...
-        if isinstance(e, auth_exceptions.TokenNotProvidedError):
-            LOG.exception('Token is not provided.')
-            return self._abort_unauthorized(str(e))
-        if isinstance(e, auth_exceptions.TokenNotFoundError):
-            LOG.exception('Token is not found.')
-            return self._abort_unauthorized(str(e))
-        if isinstance(e, auth_exceptions.TokenExpiredError):
-            LOG.exception('Token has expired.')
-            return self._abort_unauthorized(str(e))
-        if isinstance(e, auth_exceptions.ApiKeyNotProvidedError):
```
Kami: What happened to this code, where do we handle this right now?

I've seen some code in `abort_request`, but I guess I'm still missing where exception to friendly user message translation happens.

st2common/st2common/models/db/__init__.py
```diff
@@ -114,8 +114,6 @@ def db_ensure_indexes():
         class_name = model_class.__name__
 
         # Note: We need to ensure / create new indexes before removing extra ones
-        LOG.debug('Ensuring indexes for model "%s"...' % (model_class.__name__))
```
Kami: I assumed you removed it since it's already logged in `ensure_indexes` method?

st2common/st2common/router.py
```diff
@@ -0,0 +1,386 @@
...
+
+
+def abort_unauthorized(msg=None):
+    raise exc.HTTPUnauthorized('Unauthorized - %s' % msg if msg else 'Unauthorized')
+
+
+def extend_with_default(validator_class):
+    validate_properties = validator_class.VALIDATORS["properties"]
+
+    def set_defaults(validator, properties, instance, schema):
```
Kami: I know we already do similar thing for action parameters, etc. but the actual "extend with default" snippet posted on jsonschema webpage doesn't work so we need to implement our own thing to correctly handle default values everywhere.

It looks like that is not needed here and the "official" version is sufficient?

st2common/st2common/router.py
```diff
@@ -0,0 +1,386 @@
...
+class Response(webob.Response):
+    def __init__(self, body=None, status=None, headerlist=None, app_iter=None, content_type=None,
+                 *args, **kwargs):
+        # Do some sanity checking, and turn json_body into an actual body
+        if app_iter is None and body is None and ('json_body' in kwargs or 'json' in kwargs):
+            if 'json_body' in kwargs:
+                json_body = kwargs.pop('json_body')
+            else:
+                json_body = kwargs.pop('json')
+            body = json_encode(json_body).encode('UTF-8')
```
Kami: Do we still correctly handle "user-friendly" response JSON encoding?

IIRC, in the past, by default we encoded it friendly for computer consumption by default so it was more compact, but if ?debug flag was used, we returned it user-friendly formatted.

st2common/st2common/router.py
```diff
@@ -0,0 +1,386 @@
...
+
+                    m = self.routes.submapper(_api_path=path, _api_method=method,
+                                              conditions=conditions)
+                    for transform in transforms[filter]:
+                        m.connect(None, re.sub(filter, transform, path), **connect_kw)
+
+        for route in sorted(self.routes.matchlist, key=lambda r: r.routepath):
+            LOG.debug('Route registered: %+6s %s', route.conditions['method'][0], route.routepath)
+
+    def match(self, req):
```
Kami: Is this method called on every request to figure out to which controller to dispatch it (it looks like it)?

If yes, just curious how performant it is...

st2common/st2common/router.py
```diff
@@ -0,0 +1,386 @@
...
+
+        if len(path) > 1 and path.endswith('/'):
+            path = path[:-1]
+
+        match = self.routes.match(path, req.environ)
+
+        if match is None:
+            raise NotFoundException('No route matches "%s" path' % req.path)
+
+        # To account for situation when match may return multiple values
```
Kami: Hm wouldn't multiple values indicate a programmer error - aka the route is not set up correctly since there should only be one controller match for each path?

st2common/st2common/router.py
```diff
@@ -0,0 +1,386 @@
...
+                                    'Only one of Token or API key expected.')
+
+                            auth_func = op_resolver(definition['x-operationId'])
+                            auth_resp = auth_func(token)
+
+                            context['user'] = User.get_by_name(auth_resp.user)
+
+                if 'user' not in context:
+                    raise auth_exc.NoAuthSourceProvidedError('One of Token or API key required.')
+            except (auth_exc.NoAuthSourceProvidedError,
```
Kami: Ignore my comment above.

I see exception translation for auth related errors happens here now. Perhaps now a bad idea to refactor this in a special error translation / handling middleware or similar.

st2common/st2common/router.py
```diff
@@ -0,0 +1,386 @@
...
+                LOG.exception('API key is not found.')
+                return abort_unauthorized(str(e))
+            except auth_exc.ApiKeyDisabledError as e:
+                LOG.exception('API key is disabled.')
+                return abort_unauthorized(str(e))
+
+        if cfg.CONF.rbac.enable:
+            user_db = context['user']
+
+            permission_type = endpoint.get('x-permissions', None)
```
Kami: Ahh, I see.

So the permission checking which was previously performed on a decorator level for "global" controller related permissions now happens here?

st2common/st2common/router.py
```diff
@@ -0,0 +1,386 @@
...
+                resolver = resolvers.get_resolver_for_permission_type(permission_type)
+                has_permission = resolver.user_has_permission(user_db, permission_type)
+
+                if not has_permission:
+                    raise rbac_exc.ResourceTypeAccessDeniedError(user_db,
+                                                                 permission_type)
+
+        # Collect parameters
+        kw = {}
+        for param in endpoint.get('parameters', []) + endpoint.get('x-parameters', []):
```
Kami: Quite some "magic" (aka new `jsexpose` in some ways) :P

It's definitively an improvement over `jsexpose` because now at least it uses global definitions, but I it's still a lot going on here so I would also prefer some standalone explicit tests for this functionality if we don't have some already.

st2common/st2common/openapi.yaml
```diff
@@ -0,0 +1,3781 @@
...
+      operationId: st2api.controllers.root:root_controller.index
+      description: General API info.
+      responses:
+        '200':
+          description: General API info.
+          schema:
+            type: object
+            properties:
+              version:
+                type: string
```
Kami: Perhaps also a good idea to make it a bit more friendly and also add description string to those two response attributes.

st2common/st2common/openapi.yaml
```diff
@@ -0,0 +1,3781 @@
...
+          description: Action pack name filter
+          type: string
+      responses:
+        '200':
+          description: List of actions
+          schema:
+            type: array
+            items:
+              $ref: '#/definitions/Action'
+          examples:
```
Kami: I like this :+1:

st2common/st2common/openapi.yaml
```diff
@@ -0,0 +1,3781 @@
...
+            application/json:
+              ref: 'core.local'
+              # and stuff
```
Kami: Ideally we could re-use existing test fixture or similar though.

If we already do and I misunderstood how it works, please ignore it.

st2common/st2common/openapi.yaml
```diff
@@ -0,0 +1,3781 @@
...
+          description: Action content
+          schema:
+            $ref: '#/definitions/ActionUpdateRequest'
+      x-parameters:
+        - name: user
+          in: context
+          x-as: requester_user
+          description: User running the action
+      responses:
+        '200':
```
Kami: Just trying to understand how this works for non default, non-200 responses (e.g. 400 for conflict on save, etc.). I only see definition for unexpected error.

Do we need to define such things here or not?

st2common/st2common/openapi.yaml
```diff
@@ -0,0 +1,3781 @@
...
+              ref: 'core.local'
+              # and stuff
+        default:
+          description: Unexpected error
+          schema:
+            $ref: '#/definitions/Error'
+  /api/v1/actions:
+    get:
+      operationId: st2api.controllers.v1.actions:actions_controller.get_all
+      x-permissions: {{ PERMISSION_TYPE.ACTION_LIST }}
```
Kami: Bit confused, I see `x-permissions` here but not for some other controllers (e.g. create action, etc.).

Is this intentional for some reason or it still needs to be finished?

st2api/st2api/controllers/resource.py
lakshmi-kannan: I agree with kami's assessment. We have tried kwargs before in CLI and it was a big PITA. It led to programmer errors and ended up creating regressions.

st2api/st2api/controllers/v1/actionviews.py
lakshmi-kannan: kami After reading enykeev's explanation and looking at the code, I think that _get_all() should really return classes and not JSON. I think I prefer to work with classes than dicts as kami outlined and I am also of the opinion that class.to_json() should be implemented. The serialization and deserialization from/to JSON should happen only at the HTTP interface/handler.

>Although we still support controllers returning models, I find this practice highly detrimental to us better understanding how http stack works. Which was part of the reason I'm doing it to begin with. 

Is the problem because of our convoluted code or a design issue? In previous jobs etc, what typically happened was the API payload was validated against and a schema. You would have another library that converted these objects to classes and you'd have a from_json and to_json method defined on the classes. Any code that worked with these objects dealt only with the models. The JSON serialization/deserialization is restricted to HTTP handler. We never had any confusions when we did that. I've seen this pattern at multiple places in different languages.

st2auth/st2auth/handlers.py
```diff
@@ -23,14 +23,18 @@
 from st2common.exceptions.auth import NoNicknameOriginProvidedError, AmbiguousUserError
```
lakshmi-kannan: See https://github.com/StackStorm/st2/pull/3262/files

st2auth/st2auth/controllers/v1/auth.py
lakshmi-kannan: Yep, see https://github.com/StackStorm/st2/pull/3262/files for something similar.

st2api/st2api/controllers/resource.py
enykeev: While I get the sentiment, I can argue that the damage done to productivity by requiring developer to constantly keep argument list in sync with the spec is much greater. Pretty sure that statical analysis allows you to get the list of keys you're trying to get from the particular kwarg and then compare the list with endpoint parameters.

st2api/st2api/controllers/v1/actionalias.py
enykeev: Well, you could use `allOf` construct for parameters too, I believe, so you can create a list of default common params and then extend it with additional ones on a case by case basis. Problem is, at that point we don't really have any indication on the list of filters the particular get_all controller supports so I had to rely on existing tests to infer it. If our coverage is lacking, I might have missed a few. Potentially, quite a few.

st2api/st2api/controllers/v1/actionexecutions.py
enykeev: I believe we've already discussed and even worked on making `st2-register-content` work through the API rather than accessing DB directly. Yep, we're not there yet and we would need to figure out how to minimize the risk in the mean time. But in the long run, it seems to me a right way to go.

st2api/st2api/controllers/v1/actionviews.py
enykeev: ```class.to_json()``` is already implemented by means of `__json__()` method. Almost the entire purpose of `json_encode()` utility is to properly unwrap it.

As far as what controller should return, well, yes, you're right and we should probably keep using models till the very end. I might be just trying to express my frustration at the fact that we as a team sometimes forget what exactly our API is returning to the user and that it is our job to design api model properly so it could be serialized to json without additional fiddle on Router side.

st2auth/st2auth/controllers/v1/auth.py
enykeev: https://github.com/StackStorm/st2/pull/2727/files#diff-b94a6aca675fab2ff48c6ef80601b8b5R141

At the point you commented on, `authorization` is not just base64 string yet, it also contains auth type like ```basic c29tZSBzdHVmZg==``` or ```bearer c29tZSBzdHVmZg==```

st2common/st2common/hooks.py
enykeev: Authorization routine is a part of Router now, the rest of the hooks were converted to wsgi middleware.

st2common/st2common/models/db/__init__.py
enykeev: Yep, I removed it because it makes no sense to me to announce some action if we're intending to respond with the result of it the very next line.

st2common/st2common/router.py
enykeev: I have no idea why official snippet doesn't work for action parameters, never had any problems with it.

st2common/st2common/router.py
enykeev: Nope, currently we always return it formatted. Any numbers that make us think that saving on spaces is a good idea?

st2common/st2common/router.py
enykeev: Well, reddit uses it so I guess we're fine. They have benchmarks in the repo, but I'm not sure with what we should compare it, really.

st2common/st2common/router.py
enykeev: Potentially. The url `/some/views/` may match both `/some/{ref_or_id}` and `/some/views/{empty_string}`. I'm hesitant of making it a problem for our users, though. We need to catch this things during development time and try to gracefully recover in runtime.

st2common/st2common/router.py
enykeev: I though about using middleware for that but decided against it at the end. I see authentication an inalienable part of the application and if we decide to extend it, we have other means for that.

st2common/st2common/router.py
enykeev: Yes. It only works for most simple RBAC case though. We would have to figure complex cases during RBAC 2.0, but I want us to move in this general direction so we could display rbac requirements in API docs.

st2common/st2common/router.py
enykeev: This is where OpenAPI really comes into play, the rest was just foreplay.

And sadly we can't avoid this complexity if we want controllers to get the list of arguments instead of dry (request, response, next) notation of proper http routers. The worst part is surely the 'body' type, but I still can sign under every single line in it, we **need** them all.

st2common/st2common/openapi.yaml
enykeev: We need A LOT more descriptions and examples in this file. It's going to be a long journey.

st2common/st2common/openapi.yaml
enykeev: We *could* try and reuse them. I'm hesitant because essentially, examples and fixtures serves different purposes and it would be a constant struggle to compose them in a way that would fit both.

st2common/st2common/openapi.yaml
enykeev: Once again, we *could* define every single possible error for the particular endpoint or we could use an umbrella `default` scheme to signify any possible error since we stick to the same response scheme for them anyway (an object with "faultstring" property). It's up to us. I personally don't think it's worth the hassle to do it now, maybe in the release or two when we populate the rest of the file with proper descriptions or get this requested by users.

st2common/st2common/openapi.yaml
enykeev: `x-permissions` only check that user has particular permission, it doesn't factor the data. Methods like POST, PUT or DELETE requires user to have particular permission to a particular entity and are not covered by our OpenAPI spec yet. Fixing it will be one of the problems we would have to solve during RBAC 2.0 work.

st2auth/st2auth/handlers.py
enykeev: now we'll see which one lands first =)

st2common/st2common/router.py
Kami: Then we are fine :+1:

st2common/st2common/router.py
Kami: IIRC, only of the problems with the default snippet is that it only handles validation for default values, but it doesn't actually manipulate the instance passed to the validator aka also assign the default values.

But yeah, if assigning is not needed, then it should work fine.

st2api/st2api/controllers/v1/actionviews.py
enykeev: Actually, I remembered why we returning Response from `._get_all()` and `._get_one()`. Rather fancy problem really.

We want to be able to set headers on the Response (in particular, 'X-Total-Count' and 'X-Limit'). And since we don't have `pecan.response` global for that anymore and we can't attach headers to model, we have to have a Request already present at that point.

st2api/st2api/controllers/resource.py
lakshmi-kannan: I still side with kami. I’ll make a tradeoff on dev productivity there. It won’t be worse than waiting for builds to pass. Seriously, we had numerous bugs in CLI code because of kwargs abuse.

PR #2728, opened 2016-06-02
Allow admin to specify an user who's context will be used when running / re-running an action

Kami:
This ties to the pack configuration v2 story.

This pull request allows administrator to specify a user who's context will be used when running and re-running an action.

Right now inheriting a user context basically only means using datastore values for that user when using a dynamic config value, but now we have a framework in place so we can expand it in the future (e.g. it could also means inheriting that user's permissions, etc.).

I also refactored some of the RBAC code since the same logic is now used in two places so it makes sense to refactor it in some common place.

In addition to that, I also cleaned / fixed some code (One of the methods expected an API object, but in one place we passed in a DB object, it mostly worked out of pure luck because objects look mostly the same. That's just one of the left-overs from the execution -> live action change...)

### DIFF
--- a/st2api/st2api/controllers/v1/actionexecutions.py
+++ b/st2api/st2api/controllers/v1/actionexecutions.py
@@ -39,7 +39,6 @@ from st2common.models.api.action import LiveActionAPI
 from st2common.models.api.action import LiveActionCreateAPI
 from st2common.models.api.base import jsexpose
 from st2common.models.api.execution import ActionExecutionAPI
-from st2common.models.db.liveaction import LiveActionDB
 from st2common.persistence.liveaction import LiveAction
 from st2common.persistence.execution import ActionExecution
 from st2common.services import action as action_service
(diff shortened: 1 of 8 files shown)


st2common/st2common/models/api/action.py
```diff
@@ -410,6 +411,18 @@ def to_model(cls, live_action):
         return model
 
 
+class LiveActionCreateAPI(LiveActionAPI):
```
manasdk: Why a new model and not add a property to an existing model?

manasdk:
> This pull request allows administrator to specify a user who's context will be used when running and re-running an action.

Why is this a good idea? What usecase does it solve?

st2common/st2common/models/api/action.py
Kami: See my comment here - https://github.com/StackStorm/st2/pull/2723#issuecomment-222937619

In short, that's the model for input, otherwise is for output.

Kami:
> Why is this a good idea? What usecase does it solve?

Impersonation is one of the things which we discussed in the past. I believe collab people and some other users asked about it.

Now that we have improved configuration in place we can start working towards that, but obviously we are still far away from full end to end impersonation.

In addition to that, it allows makes testing and replicating some issues easier (now admin can run an action as arbitrary user).

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -54,6 +54,7 @@
 from st2common.rbac.decorators import request_user_has_permission
 from st2common.rbac.decorators import request_user_has_resource_db_permission
 from st2common.rbac.utils import assert_request_user_has_resource_db_permission
+from st2common.rbac.utils import assert_request_user_is_admin_if_user_query_param_is_provider
```
lakshmi-kannan: lmao. it's def funny.

Kami: longer is better :P

Open to suggestions though.

lakshmi-kannan:
I am definitely with manas on the impersonation part. It's a scope creep and isn't something that we planned as part of this deliverable. What is done is done but we need to be prudent about making such changes because impersonation as a whole is a critical piece and needs collaborative design process.

lakshmi-kannan:
lgtm

Kami:
I agree that the impersonation is a bigger story and needs more thought.

I think this feature is important because it's a part of debugging / testing story - it allows admin to run action as any user which makes it easier to replicate user-related problems.

In addition to that, I also think that we need a command which allows admin to view fully resolved config for any user, but that's out of the scope of this change.

PR #2729, opened 2016-06-03
Add script to initialize autocomplete for new shell

enykeev:


Kami:
I would put it in scripts or a similar directory instead since it's not really a config.

Anyway, LGTM, :+1:

enykeev:
> I would put it in scripts or a similar directory instead since it's not really a config.

It would end up in `/etc/bash_complete.d/` after installation so it's not that crystal clear.

Kami:
Did we also decide how (if anything) we will handle this when installing st2client from PyPi?

Simplest thing to being with would be to make sure this file is included in the final tarball (so potentially needs updates to MANIFEST.in and / or setup.py)

PR #2731, opened 2016-06-04
include encrypt/decrypt and scope in dataservice for actions

manasdk:


Kami:
Good catch and thanks for doing that :+1:

Kami:
Would also be good to add some tests for that to the existing datastore service tests.

And we also need to have a look if corresponding documentation needs to be updated (don't remember if @lakshmi-kannan initially added any documentation about secrets in the datastore).

st2common/st2common/services/datastore.py
```diff
@@ -79,7 +85,8 @@ def get_value(self, name, local=True):
         self._logger.audit('Retrieving value from the datastore (name=%s)', name)
 
         try:
-            kvp = client.keys.get_by_id(id=name)
+            params = {'decrypt': str(decrypt).lower(), 'scope': scope}
+            kvp = client.keys.get_by_id(id=name, params=params)
```
lakshmi-kannan: I wonder how this would work with user scoped variables? The context of user is not defined here, right?

manasdk: Nope. I do not know the answer :)

lakshmi-kannan: Then for now, can you throw an exception if someone passes scope=USER_SCOPE to the API call?

lakshmi-kannan:
BTW e2e tests are here - https://github.com/StackStorm/st2tests/pull/47/files

st2common/st2common/services/datastore.py
manasdk: k.

manasdk: Fixed dc35f901049178fd250809f75cf44f6cdd78236b

Issue #2730, opened 2016-06-03
Creating a rule with month: "*" fails


codyaray:
Specifying the month in a CronTimer rule results in an Internal Server Error. The below rule code works fine without the `month: "*"` line.

```
---
name: "remote_test"
pack: "default"
description: ""
enabled: true

trigger:
    type: "core.st2.CronTimer"
    parameters:
        minute: "3"
        hour: "15"
        day: "*"
        month: "*"
        day_of_week: "*"

action:
    ref: "core.remote_sudo"
    parameters:
        cmd: "whoami"
        hosts: "myhost"
```

Try to load it

```
$ st2 rule create remote_test.yaml
ERROR: 500 Server Error: Internal Server Error
MESSAGE: Internal Server Error for url: http://127.0.0.1:9101/v1/rules
```

We can see what happens in the logs here:

```
==> /plogs/st2/st2api.log <==
2016-06-03 15:26:55,881 140553046116784 ERROR hooks [-] API call failed: 'NoneType' object has no attribute 'update'
Traceback (most recent call last):
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/pecan/core.py", line 631, in __call__
    self.invoke_controller(controller, args, kwargs, state)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/pecan/core.py", line 531, in invoke_controller
    result = controller(*args, **kwargs)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/models/api/base.py", line 230, in callfunction
    result = f(*args, **kwargs)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/rbac/decorators.py", line 85, in func_wrapper
    return func(*args, **kwargs)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2api/controllers/v1/rules.py", line 96, in post
    increment_trigger_ref_count(rule_api=rule)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/services/triggers.py", line 202, in increment_trigger_ref_count
    Trigger.update(trigger_db, inc__ref_count=1)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/persistence/base.py", line 200, in update
    cls._get_impl().update(model_object, **kwargs)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/models/db/__init__.py", line 178, in update
[...]

PR #2732, opened 2016-06-06
Implement a work-around for CronTimer rule creation

Kami:
This is a terrible work-around for bug reported in #2730.

As described in that bug report, sometimes rule creation fails because finding a corresponding TriggerDB object by trigger parameters fails. I managed to track it down to an issue when unicode-literals are used for Python values (it's easy to reproduce it, check the bug report and insert debugger when trigger look up fails - if you try to look it up when parameters contain unicode literals it fails, but if you encode it to string literals it passes).

The most bizarre thing is that this doesn't happen always - sometimes dictionary can contain all unicod e literals and it works just fine - I have no idea why, but it looks like a weird bug with unicode literals and internal query cache handling in pymongo / mongoengine.

Yes, I know the work-around is terrible and I hate it, but I couldn't come up with a nicer solution. At least I tried to scope it down so we only re-run the query if trigger is not found and we are trying to find a cron trigger with parameters.

### DIFF
--- a/st2common/st2common/services/triggers.py
+++ b/st2common/st2common/services/triggers.py
@@ -13,7 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import six
+
 from st2common import log as logging
+from st2common.constants.triggers import CRON_TIMER_TRIGGER_REF
 from st2common.exceptions.sensors import TriggerTypeRegistrationException
 from st2common.exceptions.triggers import TriggerDoesNotExistException
 from st2common.exceptions.db import StackStormDBObjectNotFoundError
@@ -45,6 +48,27 @@ def get_trigger_db_given_type_and_params(type=None, parameters=None):
                                     parameters=parameters)
 
         trigger_db = trigger_dbs[0] if len(trigger_dbs) > 0 else None
+
+        # Work around for cron-timer when in some scenarios finding an object fails when Python
+        # value types are unicode :/
+        if not trigger_db and six.PY2 and type == CRON_TIMER_TRIGGER_REF and parameters:
+            non_unicode_literal_parameters = {}
+            for key, value in six.iteritems(parameters):
+                key = key.encode('utf-8')
+                value = value.encode('utf-8')
+                non_unicode_literal_parameters[key] = value
+            parameters = non_unicode_literal_parameters
+
+            trigger_dbs = Trigger.query(type=type,
+                                        parameters=non_unicode_literal_parameters).no_cache()
+
+            # Note: We need to directly access the object, using len or accessing the query set
+            # twice won't work - there seems to bug a bug with cursor where accessing it twice
+            # will throw an exception
+            try:
+                trigger_db = trigger_dbs[0]
+            except IndexError:
+                trigger_db = None
 
         if not parameters and not trigger_db:
             # We need to do double query because some TriggeDB objects without
--- a/st2common/st2common/services/triggers.py
+++ b/st2common/st2common/services/triggers.py
@@ -49,13 +49,22 @@ def get_trigger_db_given_type_and_params(type=None, parameters=None):
 
         trigger_db = trigger_dbs[0] if len(trigger_dbs) > 0 else None
 
+        # NOTE: This is a work-around which we might be able to remove once we upgrade 
+        # pymongo and mongoengine
         # Work around for cron-timer when in some scenarios finding an object fails when Python
         # value types are unicode :/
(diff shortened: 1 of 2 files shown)


st2common/st2common/services/triggers.py
```diff
@@ -46,6 +49,27 @@ def get_trigger_db_given_type_and_params(type=None, parameters=None):
...
+        # value types are unicode :/
+        if not trigger_db and six.PY2 and type == CRON_TIMER_TRIGGER_REF and parameters:
+            non_unicode_literal_parameters = {}
+            for key, value in six.iteritems(parameters):
+                key = key.encode('utf-8')
+                value = value.encode('utf-8')
+                non_unicode_literal_parameters[key] = value
+            parameters = non_unicode_literal_parameters
+
+            trigger_dbs = Trigger.query(type=type,
```
Kami: If i don't disable query set cache it fails (it will just try to retrieve an item which was previously not found from an internal cache). I basically need to force a cache and only way to do that is by disabling the internal queryset cache.

st2common/st2common/services/triggers.py
```diff
@@ -46,6 +49,27 @@ def get_trigger_db_given_type_and_params(type=None, parameters=None):
...
+                                        parameters=non_unicode_literal_parameters).no_cache()
+
+            # Note: We need to directly access the object, using len or accessing the query set
+            # twice won't work - there seems to bug a bug with cursor where accessing it twice
+            # will throw an exception
+            try:
+                trigger_db = trigger_dbs[0]
```
Kami: This is also terrible, but if I try to access the QuerySet twice (perform the same check as above or similar), it throws:

```python
  File "/data/stanley/virtualenv/local/lib/python2.7/site-packages/pymongo/cursor.py", line 660, in __getitem__
    raise IndexError("no such item for Cursor instance")
```

Basically, it looks like that accessing query set in any kind of way (even just by checking the length) causes cursor to advance to the very end. This is another, what looks like a bizarre bug with cursor handling in pymongo / mongoengine.

Kami:
I also checked and it appears we are running the latest version if pymongo 2.9.x release branch.

Maybe this is fixed in newer versions of pymongo / mongoengine, but upgrade is not that straight forward so I wasn't able to confirm that.

st2common/st2common/services/triggers.py
```diff
@@ -46,6 +49,27 @@ def get_trigger_db_given_type_and_params(type=None, parameters=None):
 
         trigger_db = trigger_dbs[0] if len(trigger_dbs) > 0 else None
 
+        # Work around for cron-timer when in some scenarios finding an object fails when Python
+        # value types are unicode :/
+        if not trigger_db and six.PY2 and type == CRON_TIMER_TRIGGER_REF and parameters:
+            non_unicode_literal_parameters = {}
```
manasdk: these are unicode since you encode them so why this name?

manasdk:
You are right this does look pretty horrible.

st2common/st2common/services/triggers.py
manasdk: Ignore me, i don't understand how encode(...) works.

Kami: `.encode()` returns str in Python 2.x and bytes in 3.x. `.decode()` on string returns unicode in 3.x (in short, it's vice-versa as commonly though).

manasdk:
Maybe there are better hacks in our codebase that could be done but given that the real fix is in pymongo/mongoengine there is very little value in spending more time apart from this change.

+1

st2common/st2common/services/triggers.py
```diff
@@ -46,6 +49,27 @@ def get_trigger_db_given_type_and_params(type=None, parameters=None):
 
         trigger_db = trigger_dbs[0] if len(trigger_dbs) > 0 else None
 
+        # Work around for cron-timer when in some scenarios finding an object fails when Python
+        # value types are unicode :/
+        if not trigger_db and six.PY2 and type == CRON_TIMER_TRIGGER_REF and parameters:
```
lakshmi-kannan: Maybe it's better to stub this out into its own method?

lakshmi-kannan: While with it, since this access is always guaranteed to work, can you add tests? :P

Kami:
Looks like test failure is related to mistral / Travis CI timeout and not to my changes.

Kami:
It's also worth noting that even though issue, or at least part of the issue seems to be related to QuerySet caching, adding `no_cache()` to the initial look-up made no difference - object was not found so two work-arounds are required (using no_cache and casting unicode type to str).

PR #2733, opened 2016-06-06
Bump timeout to 120 for reload in packs.install

DoriftoShoes:
The VDX pack takes roughly 80 seconds to load with 4000 actions.  Bumping this up to accommodate larger packs.

### DIFF
--- a/contrib/packs/actions/workflows/install.yaml
+++ b/contrib/packs/actions/workflows/install.yaml
@@ -27,6 +27,7 @@
       ref: "packs.load"
       parameters:
         register: "{{register}}"
+        timeout: 120
       on-success: "restart-sensor-container"
     -
       name: "restart-sensor-container"


lakshmi-kannan:
sketchy but lgtm

DoriftoShoes:
I agree that the ACTUAL fix is to make the registration go faster but this unblocks large packs in the mean time.

manasdk:
👍 

Thoughts on exposing this all the way to the user so they can optionally bump up the values?

DoriftoShoes:
I like that better.  I'll expose it.

Kami:
LGTM.

But yeah, we should measure it (I assume it spends most time talking to the DB, but who knows) and see how we can speed it up - for one, we might be able to perform some kind of multi insert instead of inserting object one by one...

PR #2735, opened 2016-06-06
Fix mocks

manasdk:


lakshmi-kannan:
lgtm

Kami:
Did tests catch that?

If not, we should maybe consider making a base impl class with abstract methods from which both classes (real and mock one) inherit. This way this should be auto-detected by lint step.

manasdk:
This was caught by st2tests not by the unit and integration tests in the st2 repo.

PR #2738, opened 2016-06-09
Support --register-fail-on-failure in st2ctl

codyaray:
The `--register-fail-on-failure` option that I [learned about yesterday|https://stackstorm-community.slack.com/archives/community/p1465382685000395] is pretty useful. We should expose this option through `st2ctl reload`.

Our content management story currently deploys packs with ansible and triggers `st2ctl reload --register-all` (and in some playbooks `--register-setup-virtualenvs` as well). So right now it fails silently if the content can't be loaded. Would be nice if we could just pass this flag through directly so it returns the error to ansible to display. 

Or you can say, no, just use the `st2-register-content` script directly for more flags. That's fine too. :)

### DIFF
--- a/st2common/bin/st2ctl
+++ b/st2common/bin/st2ctl
@@ -33,6 +33,7 @@ function print_usage() {
     echo "  --register-policies           Register all policies."
     echo "  --register-configs            Register all configuration files."
     echo "  --register-setup-virtualenvs  Create Python virtual environments for all the registered packs."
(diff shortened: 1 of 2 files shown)


Kami:
Thanks - I agree that this functionality is useful.

Please make sure to also update st2ctl in st2common/bin/ directory.

lakshmi-kannan:
lgtm.

PR #2739, opened 2016-06-13
Update remote (command) runner to only establish SFTP connection if one is required

Kami:
This pull request updates ParamikoSSHRunner to lazily establish SFTP connection. This means that connection will only be established if SFTP related functionality (file upload, file download, etc.) is required.

Common case for remote runner is that SFTP connection is not required (it's only required for script, but not for command runner). In addition to that, remote command runner should now also work under Cygwin on Windows.

Cygwin Windows issue was originally reported by @codyaray (https://github.com/StackStorm/st2contrib/issues/505.).

This resolved https://github.com/StackStorm/st2contrib/issues/505.


Kami:
@codyaray Can you please test this change on your Windows box and confirm that it works?

Thanks! (and thanks for reporting the issue :))

lakshmi-kannan:
What a weird issue? 👍 . @codyaray: Good catch.

Kami:
I've added some tests in 89d727abd7e1bb13b52cce7c03cde6af0ffec2ce.

I will go ahead and merge it once the build passes.

Kami:
@lakshmi-kannan Yeah, it's probably related to some permissions thing (you probably need some additional permissions for SFTP access, or maybe cygwin OpenSSH doesn't even support that).

In any case, it's also not such a bad idea to only establish a connection if it's needed (although, yeah, if this issue wouldn't arise I would probably argue it's a premature optimization since there are many other more important places to optimize / improve :)).

Kami:
@codyaray I've merged this change into master. This means new packages should be built shortly and you should be able to use them to test this change.

PR #2744, opened 2016-06-14
Add token validation endpoint to st2auth

m4dcoder:


st2auth/st2auth/controllers/v1/auth.py
```diff
@@ -47,6 +49,25 @@ def post(self, request, **kwargs):
         elif cfg.CONF.auth.mode == 'standalone':
             return self._handle_standalone_auth(request=request, **kwargs)
 
+    @jsexpose(arg_types=[str])
+    def get_one(self, token):
+        try:
+            token_db = get_token(token)
+
+            # Check if token has expired.
+            if token_db.expiry < date_utils.get_datetime_utc_now():
```
Kami: IIRC, there is already `validate_auth_token` method in `st2common/st2common/util/auth.py` so we should ideally re-use that if possible (and if necessary, make changes so we can re-use it).

I think that's especially important since it's security sensitive code.

st2common/st2common/hooks.py
```diff
@@ -114,6 +115,12 @@ def before(self, state):
         if state.request.method == 'OPTIONS':
             return
 
+        # Token request is authenticated separately.
+        if (state.request.method == 'POST' and (
+                state.request.path == '/tokens' or
+                re.search('/v\d+/tokens$', state.request.path))):
```
Kami: Yeah, that's kinda nasty, but I don't think we can do much else at this point. 

For one, we could maybe make path regex a constant or something, but that's about it.

st2auth/st2auth/controllers/v1/auth.py
```diff
@@ -47,6 +49,25 @@ def post(self, request, **kwargs):
         elif cfg.CONF.auth.mode == 'standalone':
             return self._handle_standalone_auth(request=request, **kwargs)
 
+    @jsexpose(arg_types=[str])
+    def get_one(self, token):
```
Kami: Since this method validates the token, I would prefer to call it and such and if possible (if that doesn't make integration too harder) also utilize POST method.

So something like: `POST /v1/tokens/validate`.

st2common/st2common/hooks.py
m4dcoder: Ok I will make the regex pattern as a constant.

st2auth/st2auth/controllers/v1/auth.py
m4dcoder: Sure.  What's the concern though?  Security related?

st2auth/st2auth/controllers/v1/auth.py
m4dcoder: Ok. I'll take a look at the auth utility.

st2auth/st2auth/controllers/v1/auth.py
Kami: Nope, more from an REST API perspective.

lakshmi-kannan: +1 on validate endpoint. Good call @Kami.

st2common/st2common/util/auth.py
```diff
@@ -26,14 +26,38 @@
...
+    :rtype: :class:`.TokenDB`
+    """
+    token = Token.get(token_string)
+
+    if token.expiry <= date_utils.get_datetime_utc_now():
+        # TODO: purge expired tokens
+        LOG.audit('Token with id "%s" has expired.' % (token.id))
+        raise exceptions.TokenExpiredError('Token has expired.')
+
+    LOG.audit('Token with id "%s" is validated.' % (token.id))
```
lakshmi-kannan: Let's not log the token anywhere.

lakshmi-kannan:
nice. i don't know if we want to advertise this via CHANGELOG. Probably not?

enykeev:
Mind providing some context for the change? Why do we need this endpoint? _Not against the change, just pure curiosity._ =)

m4dcoder:
@enykeev: We need to add authentication to the BWC Inventory Service. In BWC related workflows, we will pass a st2 token to the inventory action which calls this inventory service.  The inventory service will then validate the token before authorizing the request.

Issue #2543, opened 2016-03-01
Paramiko SSH runner does not use bastion host correctly.


lakshmi-kannan:
Reported by @emptywee on slack.

PR #2545, opened 2016-03-01
WIP: Actually pass bastion_host param to ParamikoRunner

lakshmi-kannan:
## TODO

* [ ] Write unit tests atleast cc: @Kami :P
* [ ] Changelog 
* [ ] Docs

cc: @lattwood 

The feature doesn't really work.

PR #2745, opened 2016-06-14
fix bastion ssh support

acm1:
Bastion support is currently broken because the bastion_host parameter isn't being used when creating a ParamikoSSHClient.

Fixes https://github.com/StackStorm/st2/issues/2543.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -64,6 +64,8 @@ in development
   functionality (file upload, directory upload, etc.) is not used. (improvement)
   Reported by  Cody A. Ray
 * API and CLI allow rules to be filtered by their enable state. (improvement)
+* Fix SSH bastion host support by ensuring the bastion parameter is passed to the paramiko ssh
+  client. (bug-fix) #2543 [Adam Mielke]
 
 1.4.0 - April 18, 2016
 ----------------------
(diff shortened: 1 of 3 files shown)


manasdk:
👍 

/cc @lakshmi-kannan Can you take a look as well.

lakshmi-kannan:
I started working on it https://github.com/StackStorm/st2/pull/2545/files. The problem is without unit tests or real infra on which the change is tested, it's impossible to know if this fixed the issue. @acm1 Did this fix actually work on your infra? Were you able to SSH via bastion?

acm1:
Yes, I can confirm this does work in my environment.

acm1:
I haven't looked closely at the unit tests in this repo yet -- I'll see if I can take a crack at adding a test for this.

lakshmi-kannan:
@acm1 Would you be willing to write unit tests to make sure the param was passed correctly? And thanks for confirming that the fix actually worked on your infra.

acm1:
I've added a unit test. Let me know what you think.

lakshmi-kannan:
@acm1 LGTM. Would you mind adding an entry to CHANGELOG at the bottom of the list with your name? https://github.com/StackStorm/st2/blob/master/CHANGELOG.rst#in-development.

lakshmi-kannan:
Thanks for changelog entry and the contribution! You are awesome! Merging.

Issue #2746, opened 2016-06-14
Fabric SSH runner hangs when using `use_ssh_config = True` and ProxyCommand


acm1:
If you are using the Fabric SSH runner with your user ssh config, and if your ssh config file uses `ProxyCommand` lines to connect through bastions, st2actionrunner will hang indefinitely, eventually using 100% CPU. This looks to be caused by calling `eventlet.monkey_patch()` too late. See https://github.com/paramiko/paramiko/issues/389 for discussion of a similar problem. Monkey patching earlier fixes the issue.

PR #2747, opened 2016-06-14
monkey patch actionrunner earlier to fix ssh proxycommand support

acm1:
Fixes https://github.com/StackStorm/st2/issues/2746.

### DIFF
--- a/st2actions/st2actions/cmd/actionrunner.py
+++ b/st2actions/st2actions/cmd/actionrunner.py
@@ -1,3 +1,8 @@
+# Monkey patching should be done as early as possible.
+# See http://eventlet.net/doc/patching.html#monkeypatching-the-standard-library
+from st2common.util.monkey_patch import monkey_patch
+monkey_patch()
+
 import os
 import signal
 import sys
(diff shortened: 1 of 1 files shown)


acm1:
I don't know how to write a test for this, though.

lakshmi-kannan:
@acm1 Fabric runner is deprecated in master. So this patch is probably irrelevant? FWIW, it was deprecated fairly recently. 

Master: https://github.com/StackStorm/st2/tree/master/st2actions/st2actions/runners/ssh
v1.4: https://github.com/StackStorm/st2/tree/v1.4/st2actions/st2actions/runners/ssh

If you agree, I can close the PR.

acm1:
Ahh, I didn't realize Fabric was deprecated in master. This patch is also needed if you use the Paramiko ssh feature in https://github.com/StackStorm/st2/pull/2748 and use ProxyCommand. I can make this part of that PR, though, if you'd like to close this.

lakshmi-kannan:
@acm1 Nah, we can merge this but a comment in the code on why you moved it to top will help us in the future. Since there are no tests someone might move it back and we'll hit the same bug :). Sorry for my silly comments. I hope you don't mind this kind of feedback.

acm1:
Not at all -- feedback is good :) I'll add the comment above.

lakshmi-kannan:
@acm1 We are planning a release of 1.5 this week and it would be good to ship this patch with that. Could you make the quick fix?

acm1:
Will do.

Issue #2544, opened 2016-03-01
Paramiko should respect SSH configuration from ~$SYSTEM_USER/.ssh/config 


lakshmi-kannan:
Feature requested by: @emptywee on slack. 

Fabric supports it right now.

manasdk:
For some weird reason I thought this was already done. I guess I remembered support from fabric base remote runner.

PR #2748, opened 2016-06-14
add ssh config file support to paramiko ssh runner

acm1:
This is based on https://gist.github.com/acdha/6064215 and potentially fixes https://github.com/StackStorm/st2/issues/2544.

This uses the ssh config file from the user that the actionrunner runs as, which on my test system is root. That's also how the fabric ssh runner works.

### DIFF
--- a/st2actions/st2actions/runners/ssh/paramiko_ssh.py
+++ b/st2actions/st2actions/runners/ssh/paramiko_ssh.py
@@ -560,6 +560,24 @@ class ParamikoSSHClient(object):
         if socket:
             conninfo['sock'] = socket
 
+        if cfg.CONF.ssh_runner.use_ssh_config:
+            ssh_config = paramiko.SSHConfig()
+            user_config_file = os.path.expanduser("~/.ssh/config")
+            if os.path.exists(user_config_file):
+                with open(user_config_file) as f:
+                    ssh_config.parse(f)
+
+            user_config = ssh_config.lookup(conninfo['hostname'])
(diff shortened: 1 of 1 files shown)


st2actions/st2actions/runners/ssh/paramiko_ssh.py
```diff
@@ -560,6 +560,24 @@ def _connect(self, host, socket=None):
         if socket:
             conninfo['sock'] = socket
 
+        if cfg.CONF.ssh_runner.use_ssh_config:
+            ssh_config = paramiko.SSHConfig()
+            user_config_file = os.path.expanduser("~/.ssh/config")
```
lakshmi-kannan: Just noting that the actionrunner would run as root if people use PAM authentication as the auth backend. In this case, expanduser for ~ will be /root. Usually, actionrunner process runs as ``st2`` in which case ~ would be /home/st2/. I honestly prefer to pass the full path to ssh config file via config. Also, log an exception with appropriate error message on lack of permissions to read would be critical. 

I also think in terms of code, you could stub it out into a method so it's more unit testable. So lines 564 - 579 could be a method _get_conninfo_from_ssh_config() and this method could return a conninfo dict. 

Thoughts? I am on slack if you want to talk this out. stackstorm.com/community-signup

lakshmi-kannan:
Completed in https://github.com/StackStorm/st2/pull/3058. Closing this PR. cc: @acm1

PR #2754, opened 2016-06-16
Use mask_secrets properly

manasdk:
* Use the mask_secrets property from API config
* Provide API support for admin to override masking

Note : `--show-secrets` is not supported for `st2 apikey get`. This feature is meant to enable token export.

### CLI

With and without `--show-secrets`

```
(virtualenv)vagrant@st2dev:~/st2$ st2 apikey list -dy
-   created_at: '2016-06-17T00:08:05.555610Z'
    enabled: true
    id: 57633f65d9d7ed1fc39c019e
    key_hash: '********'
    metadata: {}
    uid: '********'
    user: stanley

(virtualenv)vagrant@st2dev:~/st2$ st2 apikey list -dy --show-secrets
-   created_at: '2016-06-17T00:08:05.555610Z'
    enabled: true
    id: 57633f65d9d7ed1fc39c019e
    key_hash: f157ebc6ae10f87af9205b31e9492d156fbd036b12686d97c69536b04d8c05327d80fc467ffc7de6f0f518b86e860a8d06aff289ed81a66bdcb160b8f23a1add
    metadata: {}
    uid: api_key:f157ebc6ae10f87af9205b31e9492d156fbd036b12686d97c69536b04d8c05327d80fc467ffc7de6f0f518b86e860a8d06aff289ed81a66bdcb160b8f23a1add
    user: stanley
```

### Todo
- [x] cli updates
- [x] tests
- [x] docs (https://github.com/StackStorm/st2docs/pull/186)

### DIFF
--- a/st2api/st2api/controllers/base.py
+++ b/st2api/st2api/controllers/base.py
@@ -22,7 +22,8 @@ from st2api.controllers.controller_transforms import transform_to_bool
 from st2common.rbac.utils import request_user_is_admin
 
 __all__ = [
-    'BaseRestControllerMixin'
+    'BaseRestControllerMixin',
+    'SHOW_SECRETS_QUERY_PARAM'
 ]
 
 
--- a/st2api/st2api/controllers/base.py
+++ b/st2api/st2api/controllers/base.py
@@ -24,6 +24,9 @@ __all__ = [
 ]
 
 
+SHOW_SECRETS_QUERY_PARAM = 'show_secrets'
+
+
 class BaseRestControllerMixin(RestController):
     """
     Base REST controller class which contains various utility functions.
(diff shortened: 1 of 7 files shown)


Kami:
I'm personally not too opinionated about this - I think displaying a hash itself is fine since it's not a secret.

st2api/st2api/controllers/base.py
```diff
@@ -64,3 +70,25 @@ def _get_query_param_value(self, request, param_name, param_type, default_value=
             value = transform_to_bool(value)
 
         return value
+
+    def _get_mask_secrets(self, request):
```
Kami: :+1:

manasdk:
> I think displaying a hash itself is fine since it's not a secret.

I was starting to question that assumption as well. Anyway, this is more or less a continuation of a previous decision so keeping with it until we conclude that `key_hash` is not a secret anyway.

Kami:
Yeah, hash is by definition one-way so there should be no reason to hide it or treat is as a secret.

Having said that, I'm also fine with doing it, but I guess it could confuse some users and make them thing like we use symmetric encryption or something instead of hash if we mask it...

manasdk:
> Having said that, I'm also fine with doing it

Keeping it as is for now.

PR #2757, opened 2016-06-20
Remove old packaging

armab:
> Part of @Kami's cleanup [Remove obsolete tools/st2ctl file](https://github.com/StackStorm/st2/commit/90839f68c98766434fdc4b4cb747b36ebf36331a)

After moving packaging to [`st2-packages`](https://github.com/stackstorm/st2-packages), looks this is not used anymore and could be deleted to avoid confusion.

----
@DoriftoShoes's :church: blessing is needed here.


Kami:
Thanks.

If everything pass we should go and merge this and then see what happens to the build system. We might need to remove some old rules otherwise stuff will fail.

And yeah, @DoriftoShoes probably has a better idea about that.

lakshmi-kannan:
I <3 this. I am glad you took the initiative @armab.

manasdk:
:+1:

DoriftoShoes:
LGTM

armab:
Let's see what happens

PR #2764, opened 2016-06-22
Fix: Array With Variable

bigmstone:
When attempting to use an array that contains variables a `maximum recursion depth exceeded` error would be raised from Jinja. This commit fixes that bug and extends the unit test to cover lists with variables.


Kami:
Awesome, nice catch and fix!

(I had no idea this didn't work :))

manasdk:
Don't forget to CP the change to v1.5

lakshmi-kannan:
+1. Like manas said, please add a CHANGELOG entry!

PR #2767, opened 2016-06-22
Add migration script

lakshmi-kannan:


st2common/setup.py
```diff
@@ -50,5 +50,6 @@
         'bin/st2-run-pack-tests',
         'bin/st2ctl',
         'bin/st2-generate-symmetric-crypto-key',
+        'bin/migrations/',
```
manasdk: might need some sort of glob pattern? see https://docs.python.org/2/distutils/sourcedist.html#manifest

lakshmi-kannan:
I am gonna merge and test to see how the packages look.

lakshmi-kannan:
It ended up in /opt/stackstorm/st2/bin but the relative folder structure wasn't maintained.

vagrant@st2test:~$ /opt/stackstorm/st2/bin/st2-migrate-datastore-to-include-scope-secret.py
vagrant@st2test:~$

PR #2763, opened 2016-06-21
RFR: Render user params (jinja) in API

lakshmi-kannan:
For background, see https://stackstorm.atlassian.net/browse/STORM-2195

## Before
```
root@st2test:/home/vagrant# st2 run core.local cmd="user.date_cmd"
ERROR: 400 Client Error: Bad Request
MESSAGE: Dependecy unsatisfied in user for url: http://127.0.0.1:9101/v1/executions
root@st2test:/home/vagrant# s
```

## After

```
(virtualenv)vagrant@st2dev /m/s/s/st2 ❯❯❯ st2 run core.local cmd="{{user.date_cmd}}"                              master ✭ ✱ ◼
.
id: 5769c625d9d7ed15e61f17eb
status: succeeded
parameters:
  cmd: '{{user.date_cmd}}'
result:
  failed: false
  return_code: 0
  stderr: ''
  stdout: Tue Jun 21 22:56:37 UTC 2016
  succeeded: true
(virtualenv)vagrant@st2dev /m/s/s/st2 ❯❯❯
```

## TODO

* [x] tests
* [x] Changelog
* [X] Docs https://github.com/StackStorm/st2docs/pull/189

PR #2772, opened 2016-06-23
Using 'system' or 'user' reserved names as param names breaks execution

lakshmi-kannan:
It looks like even in 1.4, if you used a param name "system", it will blow up execution API because of our jinja resolution code. 

Now that we allow jinja expression like {{user.}} too with https://github.com/StackStorm/st2/pull/2763, 'user' becomes a reserved param name. 

Our chatops CI exposed this issue after we merged https://github.com/StackStorm/st2/pull/2763 (thanks @humblearner and @manasdk). We are using param 'user' in chatops.post_message https://github.com/StackStorm/st2/blob/master/contrib/chatops/actions/post_message.yaml#L13.

I amended the unit tests fixtures to include params 'system' and 'user' so they fail. See travis failure. 

## Solutions

1. We have to live with the fact that we want UX in terms of  {{system.}} and {{user.}} and disallow param names 'system' and 'user'. In this case, we would fix https://github.com/StackStorm/st2/blob/master/contrib/chatops/actions/post_message.yaml#L13.

2. If we want to allow those param names, we have to introduce alternate jinja expressions like {{system_kv.}} and {{user_kv.}}. This only reduces the chance of collision.

### DIFF
--- a/st2tests/st2tests/fixtures/generic/actions/action_system_default.yaml
+++ b/st2tests/st2tests/fixtures/generic/actions/action_system_default.yaml
@@ -11,4 +11,7 @@ parameters:
   actionstr:
     default: '{{system.actionstr}}'
     type: string
+  system:
+    default: 'this will break things'
+    type: string
 runner_type: test-runner-1
--- a/st2tests/st2tests/fixtures/generic/actions/action_user_default.yaml
+++ b/st2tests/st2tests/fixtures/generic/actions/action_user_default.yaml
@@ -8,4 +8,7 @@ parameters:
   actionstr:
     default: '{{user.foo}}'
     type: string
+  user:
+    default: 'unholy mess'
+    type: string
 runner_type: test-runner-1


dzimine:
I'm ok with 2 (just name `st2system` and `st2user`) - but this comes with deprecation of `system` and migration to anyone who is already using system and user.

Kami:
First question - can we delay #2763 for post v1.5 (if that's indeed the only change affecting that)?

This way we don't need to rush and have a bit more time to decide how to proceed.

I'm personally OK with 2) - changing the prefix used for accessing the datastore. In addition to that, we should also make it more user friendly and throw when registering actions in case a reserved parameter name is used. I think second part is also important, otherwise users will only get notified during run-time and that's bad UX...

emedvedev:
I'm ok with 2, better to use a prefix or even a param (`kv.user` and `kv.system`). We can still allow using `system` for compatibility, but deprecate it eventually.

`user.` and `system.` aren't really descriptive, too. I wouldn't tell where the data from `user.` comes from. A clear prefix would solve this problem.

lakshmi-kannan:
We decided that we'll revert #2763 until after v1.5 which gives us extra time. We are going to leave {{user.}} as such for now. We'll migrate {{system.}} and {{user.}} in next release.

PR #2777, opened 2016-06-23
Self check v1.5

DoriftoShoes:


Kami:
We should probably also update package definitions so self  check script is also installed into /usr/local/bin. 

I will look into it.

Issue #2779, opened 2016-06-28
`st2ctl status` reports st2chatops is not running


LindsayHill:
Using v1.5 packages on Ubuntu 14.04:
```
root@chatops:/var/run# dpkg -l|grep st2
ii  st2                                  1.5.0-6                          amd64        StackStorm Event-driven automation
ii  st2chatops                           1.5.0-2                          amd64        St2Chatops - StackStorm Chatops
ii  st2mistral                           1.5.0-7                          amd64        St2 Mistral workflow service
ii  st2web                               1.5.0-1                          amd64        St2Web - StackStorm Web UI
```

`st2ctl status` reports that st2chatops is not running:
```
root@chatops:/var/run# st2ctl status
##### st2 components status #####
st2actionrunner PID: 22197
st2api PID: 22205
st2api PID: 22424
st2stream PID: 22212
st2stream PID: 22428
st2auth PID: 22219
st2auth PID: 22429
st2garbagecollector PID: 22226
st2notifier PID: 22233
st2resultstracker PID: 22240
st2rulesengine PID: 22247
st2sensorcontainer PID: 22254
st2chatops is not running.
mistral-server PID: 22285
mistral-api PID: 22277
mistral-api PID: 22415
mistral-api PID: 22416
root@chatops:/var/run#
```

But it is running:
```
root@chatops:/var/run# service st2chatops status
st2chatops start/running, process 22261
root@chatops:/var/run# ps 22261
  PID TTY      STAT   TIME COMMAND
22261 ?        Ssl    0:02 node /opt/stackstorm/chatops/node_modules/.bin/coffee /opt/stackstorm/chatops/node_modules/.bin/hubot @
root@chatops:/var/run#
```

Looks like this is the problem lines are this: https://github.com/StackStorm/st2/blob/master/st2common/bin/st2ctl#L5 in conjunction with https://github.com/StackStorm/st2/blob/master/st2common/bin/st2ctl#L162

It's searching the process list for `st2chatops` but it might be better to search for `hubot`

PR #2780, opened 2016-06-28
Fix st2ctl status for st2chatops

armab:
> Addresses #2779

As @LindsayHill noted, one of the fast fixes is to use `hubot` instead of `st2chatops` for pid discovery.

```
$ st2ctl status
##### st2 components status #####
st2actionrunner PID: 929
st2api PID: 843
st2api PID: 1192
st2stream PID: 846
st2stream PID: 1193
st2auth PID: 832
st2auth PID: 1191
st2garbagecollector PID: 830
st2notifier PID: 835
st2resultstracker PID: 833
st2rulesengine PID: 838
st2sensorcontainer PID: 824
hubot PID: 9907
mistral-server is not running.
mistral-api is not running.
```

### DIFF
--- a/st2common/bin/st2ctl
+++ b/st2common/bin/st2ctl
@@ -158,6 +158,7 @@ function clean_logs() {
 function getpids() {
   echo "##### st2 components status #####"
   COMPONENTS=${COMPONENTS/mistral/mistral-server mistral-api}
+  COMPONENTS=${COMPONENTS/st2chatops/hubot}
 
   for COM in ${COMPONENTS}; do
     PID=`ps ax | grep -v grep | grep -v st2ctl | grep "${COM}" | awk '{print $1}'`


Kami:
Good catch @LindsayHill and thanks @armab!

Kami:
On a related note - we also need to add some tests for this to st2tests and or robotfm tests suite.

/cc @emedvedev

Kami:
Btw, maybe this was done (semi) intentionally? Do we support installations without ChatOps / hubot?

If we do, we also need to handle that case as well - make sure st2ctl doesn't fail if hubot is intentionally not installed. IIRC, that was the scenario we indeed supported at some point, but not 100% sure right now.

/cc @lakshmi-kannan @DoriftoShoes @emedvedev

emedvedev:
> make sure st2ctl doesn't fail if hubot is intentionally not installed

It doesn't fail, but it does give an error that can be confusing. We had this covered, but apparently that code wasn't working for RHEL6 (IIRC), so it would make sense to rewrite it and include the fix into `st2ctl` again.


I'm closing this in favor of https://github.com/StackStorm/st2chatops/pull/45 which is somewhat hackier but better from the UX perspective.

PR #2786, opened 2016-06-30
Correctly resolve parameter's required property

enykeev:
Back in https://github.com/StackStorm/st2/pull/690/files#diff-950f03378fda3eb80bbc01142a430fd7R31 @m4dcoder presumably decided that if certain parameter is set as required on a runner level, it should end up being required even after we add action parameters to the mix. In reality, user can set default value for this parameter on an action level and thus lift the requirement.

Strictly speaking, parameter should never have truthy values of `require` and `default` simultaneously, but that change is more on par with linting actions during registration and therefore outside the scope of this PR.

Resolves https://github.com/StackStorm/st2web/issues/307

STORM-2201

### DIFF
--- a/st2common/st2common/models/utils/action_param_utils.py
+++ b/st2common/st2common/models/utils/action_param_utils.py
@@ -40,7 +40,7 @@ def _merge_param_meta_values(action_meta=None, runner_meta=None):
         elif key in runner_meta_keys and key not in action_meta_keys:
             merged_meta[key] = runner_meta[key]
         else:
-            if key in ['immutable', 'required']:
+            if key in ['immutable']:
                 merged_meta[key] = runner_meta.get(key, False) or action_meta.get(key, False)
             else:
                 merged_meta[key] = action_meta.get(key)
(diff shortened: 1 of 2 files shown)


Kami:
Good catch - thanks!

As far as required and default goes - I've updated the JSON schema validate function (`util.schema.validate`) in the past so if a parameter was required, but had a default value set it wouldn't throw an exception. In short, you didn't need to set `required` to `False` if default value was provided.

From this PR it looks like this functionality got changed at some point though :/

Kami:
Actually, looking at #307 some more, this really only seems to be "display" issue. It sounds like backend functionality still works fine.

To keep it consistent with the backend, would it be possible to update st2web code to not display asterik next to the parameter name if required is true, but default value is also provided (this way it will be consistent with the backend behavior)?

lakshmi-kannan:
>Strictly speaking, parameter should never have truthy values of require and default simultaneously,

I don't think that is a valid assumption although I can see it's a less common use case. Manas and I have argued about it and figured it's a less common but valid use case.

PR #2787, opened 2016-07-01
Exclude parameters during rerun to fallback to default value

enykeev:
Currently, if the initial action was executed with particular parameter, there is no way to remove this parameter during rerun. Best you can do is to replace it with default value manually.

Since it seems like in practice we don't really have another use for parameter's `null` value, I'm proposing to use it as an indicator to fallback to default value during rerun. This way, execution with parameters

```yaml
a: 'b'
c: 'd'
```

being rerun with parameters

```yaml
c: null
```

will result in a new execution with parameters

```yaml
a: 'b'
```

which would be later resolved as if it was an initial execution run (with default value of `c` potentially being resolved). In practice, this is how it works today. The change only ensures resulting execution having proper set of parameters instead of

```yaml
a: 'b'
c: null
```

UI changes are pending.

### DIFF
--- a/st2api/st2api/controllers/v1/actionexecutions.py
+++ b/st2api/st2api/controllers/v1/actionexecutions.py
@@ -22,6 +22,7 @@ import traceback
 import jsonschema
 import pecan
 from pecan import abort
+import six
 from six.moves import http_client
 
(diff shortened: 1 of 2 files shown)


enykeev:
Closed in favor of #2790

PR #2790, opened 2016-07-05
Add flag to prevent parameter merging during rerun

enykeev:
That's another way of solving https://github.com/StackStorm/st2/pull/2787 that plays much better with specifics of rerun implementation in st2web.

By avoiding the merge on an api side, user is getting full control over execution parameters without introduction of any additional special types.

### DIFF
--- a/st2api/st2api/controllers/v1/actionexecutions.py
+++ b/st2api/st2api/controllers/v1/actionexecutions.py
@@ -276,7 +276,7 @@ class ActionExecutionReRunController(ActionExecutionsControllerMixin, ResourceCo
             return self
 
     @jsexpose(body_cls=ExecutionSpecificationAPI, status_code=http_client.CREATED)
-    def post(self, spec, execution_id):
+    def post(self, spec, execution_id, no_merge=None):
         """
         Re-run the provided action execution optionally specifying override parameters.
 
@@ -290,7 +290,9 @@ class ActionExecutionReRunController(ActionExecutionsControllerMixin, ResourceCo
             raise ValueError('Task option is only supported for Mistral workflows.')
 
         # Merge in any parameters provided by the user
-        new_parameters = copy.deepcopy(getattr(existing_execution, 'parameters', {}))
+        new_parameters = {}
+        if not no_merge:
+            new_parameters.update(getattr(existing_execution, 'parameters', {}))
         new_parameters.update(spec.parameters)
 
         # Create object for the new execution


st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -276,7 +276,7 @@ def validate(self):
             return self
 
     @jsexpose(body_cls=ExecutionSpecificationAPI, status_code=http_client.CREATED)
-    def post(self, spec, execution_id):
+    def post(self, spec, execution_id, nomerge=None):
```
Kami: Let's please call it `no_merge` or similar for consistency.

Kami:
I guess I'm still missing some context on why that is needed / desired (the idea with the re-run is that you want to re-run execution as-is + potentially override some parameters), but I won't block it and I'm fine with this query parameter approach.

enykeev:
Kami, simple example:
- you have an action with parameter that has default value

```
a:
    default: 'b'
```
- you are initially running the action with `a='c'`
- you want to rerun the execution with default `a`

Before this PR or #2787, you can't, the only option for you is to manually copy value from action spec into rerun dialog and that's obviously a bad UX.

Now the difference between PR's is that in #2787 we're introducing new parameter value with implicit meaning of "discard parent execution value and fallback to default" while in this PR we're making it a default behavior for all parameters.

The reason I think this PR better solves the problem is the fact that when you try to rerun an execution in Web UI, fields of the form are already come pre-filled with values from parent execution. We did it to make it easier for user to make a small change in parameters (like fix a single character typo).

screen shot 2016-07-06 at 11 06 19 am

If we would decide to implement #2787 support in Web UI, we would then need to track the state of every field explicitly to figure out whether user wants to keep a previous value or fallback to default one. While implementing this PR only requires to change initial value for the payload object. Doing it this way is simply cheaper, both computationally and cognitively.

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -276,7 +276,7 @@ def validate(self):
             return self
 
     @jsexpose(body_cls=ExecutionSpecificationAPI, status_code=http_client.CREATED)
-    def post(self, spec, execution_id):
+    def post(self, spec, execution_id, no_merge=None):
```
Kami: Minor thing - you can also default this argument to `False` and then specify `arg_types`  argument in `jsexpose` and it will correctly handle and cast the query param.

Kami:
I'm going to add a changelog entry and cherry-pick this into v1.5.1 branch for v1.5.1 release.

PR #2800, opened 2016-07-11
Upgrade to the latest version of pymongo and mongoengine

Kami:
This pull request upgrades to latest version of `pymongo` and `mongoengine`.

We wanted to do this many times before, but there are some backward incompatible changes in the newer versions and we never spent enough time digging in and addressing those issues so it never happened.

With the latest version of those libraries, StackStorm now also supports MongoDB 3.x. I think this is very important since MongoDB 3.x addresses and improves a lot of things (especially with a new storage engine). In the near future we should recommend MongoDB 3.x for new users, install MongoDB 3.x with the installer and come up with an upgrade scenario and document for an existing users which want to upgrade from 2.x to 3.x (there are some backward incompatible changes which require some work on the user side).

## Mongoengine Strict Mode

New version of mongoengine now ships with a "strict" mode which is enabled by default. When this mode is enabled, document constructor method will throw an exception if an argument / field which is not declared on the model definition is passed in.

For example:
    
```python
class MyModel(me.Document):
    field1 = me.StringField()

instance = MyModel(field1='foo', field2='bar')
```

This snippet would throw `FieldDoesNotExist` exception because `MyModel` doesn't declare that field.

I of course left this thing enabled because it's a good thing - it will allow us to more easily catch programmer errors and typos. If you check the diff you can see we had some issues like that in our code, but previous versions of mongoengine didn't throw an exception so those issues were silently ignored.

# Roll out scenario

All the tests pass locally, but I would still prefer to roll this out gradually and I want us to first test / run it on our build servers for a while.

As far as StackStorm version goes, I propose to ship v1.5.1 with the current bug-fixes from master and then ship and include this change in v1.6.0.

## TODO

- [ ] Update installer documentation and scripts to install MongoDB 3.2 for new installations (not part of this PR)

### DIFF
--- a/st2common/st2common/models/db/stormbase.py
+++ b/st2common/st2common/models/db/stormbase.py
@@ -55,10 +55,9 @@ class StormFoundationDB(me.Document, DictSerializableClassMixin):
     # We explicitly assign the manager so pylint know what type objects is
     objects = me.queryset.QuerySetManager()
 
-    # ObjectIdField should be not have any constraints like required,
-    # unique etc for it to be auto-generated.
-    # TODO: Work out how we can mark this as a unique primary key.
-    id = me.ObjectIdField()
+    # Note: In mongoengine >= 0.10 "id" field is automatically declared on all
+    # the documents and declaring it ourselves causes a lot of issues so we
+    # don't do that
 
     # see http://docs.mongoengine.org/guide/defining-documents.html#abstract-classes
     meta = {
--- a/st2common/st2common/models/db/stormbase.py
+++ b/st2common/st2common/models/db/stormbase.py
@@ -65,6 +65,13 @@ class StormFoundationDB(me.Document, DictSerializableClassMixin):
         'abstract': True
     }
 
+    def __init__(self, *args, **kwargs):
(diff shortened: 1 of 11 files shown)


st2common/st2common/models/db/stormbase.py
```diff
@@ -55,10 +55,9 @@ class StormFoundationDB(me.Document, DictSerializableClassMixin):
     # We explicitly assign the manager so pylint know what type objects is
     objects = me.queryset.QuerySetManager()
 
-    # ObjectIdField should be not have any constraints like required,
-    # unique etc for it to be auto-generated.
-    # TODO: Work out how we can mark this as a unique primary key.
-    id = me.ObjectIdField()
+    # Note: In mongoengine >= 0.10 "id" field is automatically declared on all
```
Kami: I tried to get it work by having `id` explicitly declared (that's my preferred approach) and without needing to have pylint plugin (explicit is better than implicit), but I couldn't get it to work without patching mongoengine - it breaks tons of things and part of it seems to be related to https://github.com/MongoEngine/mongoengine/pull/961.

PR #2801, opened 2016-07-11
Various fixes

Kami:
- Correctly cast `no_merge` query param to boolean
- Update regular expression so we correctly handle some more pecan edge cases and return 404 instead of 500 on invalid paths

### DIFF
--- a/st2api/tests/unit/controllers/v1/test_base.py
+++ b/st2api/tests/unit/controllers/v1/test_base.py
@@ -65,3 +65,12 @@ class TestBase(FunctionalTest):
         self.assertEqual(response.status_int, 200)
         self.assertEqual(response.headers['Access-Control-Allow-Origin'],
                          '*')
+
+    def test_valid_status_code_is_returned_on_invalid_path(self):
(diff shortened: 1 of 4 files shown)


st2api/tests/unit/controllers/v1/test_base.py
```diff
@@ -65,3 +65,13 @@ def test_wildcard_origin(self):
...
         self.assertEqual(response.headers['Access-Control-Allow-Origin'],
                          '*')
+
+    def test_valid_status_code_is_returned_on_invalid_path(self):
+        # TypeError: get_all() takes exactly 1 argument (2 given)
+        resp = self.app.get('/v1/executions/577f775b0640fd1451f2030b/re_run', expect_errors=True)
+        self.assertEqual(resp.status_int, 404)
+
+        # get_one() takes exactly 2 arguments (4 given)
+        resp = self.app.get('/v1/executions/577f775b0640fd1451f2030b/re_run/a/b',
```
lakshmi-kannan: Man, this should be pecan's job. Oh well.

PR #2808, opened 2016-07-12
Model migrations -- bash script

lakshmi-kannan:
## The plan

First, I'll be writing a migration wrapper script that calls out specific migration scripts for each version. Each version specific migration script will call the python migration scripts. 

This script will be used only for upgrade_testing workflows. 

## Assumptions

1. Patch versions won't have data model migrations. Only major and minor versions would.

2. We won't support downgrade migrations i.e. if you are going from 1.5 to 1.4, you can deal with shit yourself.

## Random musings

~~I don't yet know how to address upgrading to skipped versions. Like say 1.4 to 1.6. This would mean that we should run migration scripts for 1.4 -> 1.5 and 1.5 -> 1.6. Maybe we can skip this for initial pass and abort the script with a positive exit code.~~

^ I solved this by some bash magic. Now I call migrate_to_${major}_${minor} in a loop starting from 
"from version" to "to version". I assume (1) is true. 

## How do I include a version specific migration? 

1. You first write a specific python script that does some data model migration and place it in appropriate version folder. For example, see https://github.com/StackStorm/st2/tree/master/st2common/bin/migrations/v1.5

2. You'd then edit this bash script and include a migrate_to_major.minor method and call out the name of the scripts to run as a bash array def. For example:

```
migrate_to_1.5() {
  local scripts=("st2-migrate-datastore-to-include-scope-secret.py")
  run_migration_scripts $scripts
}
```

### DIFF
--- a/st2common/bin/st2-perform-model-migrations
+++ b/st2common/bin/st2-perform-model-migrations
@@ -4,8 +4,49 @@ set -eu
 
 FROM_VERSION=''
 TO_VERSION=''
+MIGRATION_MIN_VERSION_SUPPORTED='1.5'
+MIGRATION_FUNC_BASE_NAME="migrate_to_"
 
-MIGRATION_MIN_VERSION_SUPPORTED="1.5.0"
+# http://stackoverflow.com/questions/59895/can-a-bash-script-tell-what-directory-its-stored-in
+MIGRATION_SCRIPT_BASE_PATH="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# Version check util methods
+
+verlte() {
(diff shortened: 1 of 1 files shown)


DoriftoShoes:
Me likey.

Kami:
My feedback on the big picture - I would really prefer this to be a Python script / framework.

That way it would be easier to write it, test it, etc. And we can also call and re-use any of the existing StackStorm Python code.

I would also prefer to store metadata about migration stuff in the database (current schema version, current data version, etc.).

This would of course require more work, but imo we should do it right so we should as much time on it as needed. In addition to that, there are already a bunch of existing Python libraries for migrations so maybe we can re-use one of them or at least use it as a starting point.

PR #2810, opened 2016-07-13
Mongoengine related fixes

Kami:
This pull request fixes some issued introduced by #2800.

Sadly those issues were not caught by unit and integration tests, they were only caught by end to end tests which don't run for every PR.

The root cause of some of the issues is that the code relied on the internal mongoengine functionality which has changed.

### DIFF
--- a/st2common/st2common/models/api/trigger.py
+++ b/st2common/st2common/models/api/trigger.py
@@ -152,6 +152,10 @@ class TriggerAPI(BaseAPI):
                           parameters=parameters)
         return model
 
+    def to_dict(self):
+        # Return dictionary version of the trigger
+        return vars(self)
+
 
 class TriggerInstanceAPI(BaseAPI):
     model = TriggerInstanceDB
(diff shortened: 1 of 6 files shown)


st2common/st2common/models/api/trigger.py
```diff
@@ -152,6 +152,10 @@ def to_model(cls, trigger):
                           parameters=parameters)
         return model
 
+    def to_dict(self):
+        # Return dictionary version of the trigger
+        return vars(self)
```
lakshmi-kannan: I am concerned this might return internal fields but other than explicit construction of the object, I don't see a way out. I think we need this to_dict() for all models too. So a base class abstract method might be better.

Kami: Yeah, that's actually an API model and not a DB one so it's less of an issue.

It's also worth nothing that that's pretty much how our API response serialization works right now - calling `json.dumps` on API model works almost in the same manner as that code here :)

Issue #2797, opened 2016-07-07
Delayed actions processed in wrong order


bri365:
When actions are delayed for policy concurrency (and concurrency.attr), they appear to be retried in newest to oldest order, rather than oldest to newest.  The file attached is a trimmed down actionrunner log showing delayed actions and their subsequent dispatch.  When multiple actions are delayed, they appear to be dispatched in the opposite order intended.

[2600.txt](https://github.com/StackStorm/st2/files/352933/2600.txt)

st2actions/policy/concurrency.py:88 and concurrency_by_attr.py:115 both include order_by['start_timestamp'], but does not indicate ascending or descending.

Kami:
Thanks for the report.

We don't officially say that we process delayed executions in the FIFO order, but I agree with you that LIFO behavior is confusing and making it FIFO would indeed be better and make more sense.

As far as the actual code / implementation goes - I believe the default mongoengine order (unless explicitly specified) is to sort in ascending order so there must be something else going on and I will dig into the code.

Kami:
I quickly looked at the code and the concurrency stuff looks correct - we don't explicitly specify ASC order, but that's the default behavior (so doing `+start_timestamp` is optional).

There might be an issue with the way requested and scheduled are processed, but looking at the code nothing stands out. By default RabbitMQ messages are processes in FIFO order, but there are many edge-cases (in a distributed system) where this doesn't always hold true - if there are multiple consumers (we run multiple action runners by default), if for some reason message is re-queued (timeout, ...), etc.

Do you happen to have a simple workflow / action example which I can use to try to reproduce this issue?

/cc @m4dcoder

bri365:
Thanks, Kami.  I'll try to see if we can reproduce this issue.  Also, our system is defaulting to a single action runner (I think four cores or less yields a single one) - any known concerns?  Would we be better/safer with two?

Kami:
No, it should be fine with a single action runner - if anything it should be better with regards to the race conditions (but there shouldn't be any, more runners simply means more actions will be executed in parallel).

And yeah, if you can come up with a simple workflow with reproduces a problem, please gist it here and I will have a look.

m4dcoder:
@Kami, nope, will have to define one to test this.

bri365:
[...]

PR #2811, opened 2016-07-13
Make sure that we don't apply policies which are disabled

Kami:
This was originally reported at https://github.com/StackStorm/st2/issues/2797#issuecomment-232146181

# TODO:

- [x] Tests
- [x] Changelog entry


Kami:
I just added some tests (it wasn't that easy) and also did some refactoring for better readability and consistency.

PR #2817, opened 2016-07-16
Allow user to specify which action to perform on threshold reached when using concurrency policies

### LINKED ISSUE #2794: Feature Request: action.concurrency.attr that will cancel instead of delay
There are situations when I want to cancel a duplicate action instead of just delaying it.

This could be in the form of an optional parameter for the policy that would let me pick delay or cancel as the action to perform.

A workaround is to have an initial task that will see if the workflow is needed before running.

Examples:

Two users requests the same thing, but the workflow is already running because of the first request.

If an problem alert is triggering a workflow and it fires once per minute but the workflow takes a few minutes to fix the problem.

Kami:
This change allows user to specify which action to perform (delay, cancel) when a concurrency policy is used and a defined threshold has been reached.

Previously, those executions were delayed by default, but now user can also select to cancel those actions. The change is fully backward compatible and `action` defaults to `delay`.

While working on this change I also noticed some of the code is not so bullet proof and fail safe so I also did some refactoring and improvements (more to come).

The actual dynamic class loading is quite unsafe right now and it could break if a third party module which is imported by the concurrency module changes (adds a class which inherits from ResourcePolicyApplicator).

In addition to that, some of the fixtures are bad - we should avoid copying base policies to fixtures directory. We should simply load default policy types in the database using resource registrar instead of re-defined and copy and pasting base policies over to fixtures. This duplication means those definitions are hard to maintain and easily get out of sync - the source of truth for the actual functionality should always be code and not fixtures!

Besides those changes, I also made various other improvements (getting rid of some *args and **kwargs abuse which could cause us to miss real programmer errors, etc) and added additional tests.

Resolves #2794

### Delay execution on threshold reached (default and existing / old behavior)

```yaml
---
name: action-2.concurrency.
pack: wolfpack
description: Limits the concurrent executions for the fake action.
enabled: true
resource_ref: wolfpack.action-2
policy_type: action.concurrency
parameters:
    threshold: 3
```

```yaml
---
name: action-2.concurrency.
pack: wolfpack
description: Limits the concurrent executions for the fake action.
enabled: true
resource_ref: wolfpack.action-2
policy_type: action.concurrency
parameters:
    action: delay
    threshold: 3
```

This one is the same as above, but `action: delay` is explicitly provided.

### Cancel execution on threshold reached (new feature)

```yaml
---
name: action-2.concurrency.
pack: wolfpack
description: Limits the concurrent executions for the fake action.
enabled: true
resource_ref: wolfpack.action-2
policy_type: action.concurrency
parameters:
    action: cancel
    threshold: 3
```

## TODO

- [x] Tests
- [x] Documentation updates
- [x] More refactoring and robustness improvements


Kami:
Here is the code where I removed the fixtures which shouldn't be fixtures and instead registered policies from st2common (which are the ones which are used otherwise) - 35c574cd78a1e709234a7af617b53271497065ae

Kami:
And here is the change which makes dynamic loading more robust / less error prone - 2a56d4c5bd338f1fceaa14aac0d9b521465e36b4

st2common/st2common/policies/meta/concurrency.yaml
```diff
@@ -6,5 +6,13 @@ resource_type: action
 module: st2actions.policies.concurrency
 parameters:
     threshold:
+        description: Concurrency threshold.
         type: integer
         required: true
+    action:
```
lakshmi-kannan: This actually might end up confusing people because ``action`` is a content type in st2 world. I am trying hard to come up with an alternate but I can't.

lakshmi-kannan:
I really like the change and the related refactoring and fixtures cleanup. Let's settle on the ``action`` name and then we can merge this.

st2common/st2common/policies/meta/concurrency.yaml
Kami: Yeah I had / have the same doubts / feelings.

I first called it `operation`, but I liked this even less than action so I changed it to action...

Kami: /cc @m4dcoder @enykeev what others think about the attribute name?

I can't come up with anything better than `action`.

Kami:
I will go ahead and merge this into trunk.

If we decide for a better / different key / attribute name, I will go ahead and change it in a new PR (it's a simple change).

PR #2820, opened 2016-07-20
Execution model to contain web_url field for chatops sake

lakshmi-kannan:
## What? 

We want to be able to display the execution history web url in chatops messages on ACK as well as after final results. 
https://github.com/StackStorm/hubot-stackstorm/issues/120

## Context 

1. When an action is kicked off we do give out the execution id link. This is because https://github.com/StackStorm/hubot-stackstorm/blob/master/scripts/stackstorm.js#L245 but it unfortunately depends on an env variable https://github.com/StackStorm/st2chatops/blob/master/st2chatops.env#L36. So mostly we get “localhost” based URL. See  https://stackstorm.slack.com/archives/chatopsci/p1468963585000061 (private link). 

2. When an action completes, we don’t share the execution URL. This is obv looking at this https://github.com/StackStorm/st2/blob/master/contrib/chatops/actions/templates/default.j2#L15. 

So it looks like we need two changes:

1. A response for alias execution POST should include the history URL 

2. An execution model which is used in post_message should carry the URL with it for final result to show the URL. 

For both, we need the web URL generated in backend. I have no good way to get the external IP of the box from within st2. So I rely on a config parameter which defaults to https://localhost but users can change it to point to Web UI host. The execution model is now changed to contain ``web_url`` field which will then be used in both hubot-stackstorm (See https://github.com/StackStorm/hubot-stackstorm/pull/122) and post_message. 

## TODO

* [x] Tests
* [x] Docs changes https://github.com/StackStorm/st2docs/pull/215

### DIFF
--- a/st2common/st2common/config.py
+++ b/st2common/st2common/config.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import os
+import socket
 import sys
 
 from oslo_config import cfg
@@ -83,7 +84,7 @@ def register_opts(ignore_errors=False):
     do_register_opts(content_opts, 'content', ignore_errors)
 
     webui_opts = [
-        cfg.StrOpt('webui_base_url', default='https://localhost',
+        cfg.StrOpt('webui_base_url', default='https://%s' % socket.gethostname(),
                    help='Base https URL to access st2 Web UI. This is used to construct' +
                         'history URLs that are sent out when chatops is used to kick off ' +
                         'executions.')
@@ -109,12 +110,12 @@ def register_opts(ignore_errors=False):
         cfg.StrOpt('ssl_certfile', help='Certificate file used to identify the localconnection',
                    default=None),
         cfg.StrOpt('ssl_cert_reqs', choices='none, optional, required',
-                   help='Specifies whether a certificate is required from the other side of the \
-                         connection, and whether it will be validated if provided',
+                   help='Specifies whether a certificate is required from the other side of the ' +
+                        'connection, and whether it will be validated if provided',
                    default=None),
         cfg.StrOpt('ssl_ca_certs',
-                   help='ca_certs file contains a set of concatenated CA certificates, which are \
-                         used to validate certificates passed from MongoDB.',
+                   help='ca_certs file contains a set of concatenated CA certificates, which are' +
+                        ' used to validate certificates passed from MongoDB.',
                    default=None),
         cfg.BoolOpt('ssl_match_hostname',
                     help='If True and `ssl_cert_reqs` is not None, enables hostname verification',
--- a/st2common/st2common/config.py
+++ b/st2common/st2common/config.py
@@ -82,6 +82,14 @@ def register_opts(ignore_errors=False):
     ]
     do_register_opts(content_opts, 'content', ignore_errors)
 
+    webui_opts = [
+        cfg.StrOpt('webui_base_url', default='https://localhost',
+                   help='Base https URL to access st2 Web UI. This is used to construct' +
+                        'history URLs that are sent out when chatops is used to kick off ' +
+                        'executions.')
+    ]
+    do_register_opts(webui_opts, 'webui', ignore_errors)
+
     db_opts = [
         cfg.StrOpt('host', default='0.0.0.0', help='host of db server'),
         cfg.IntOpt('port', default=27017, help='port of db server'),
--- a/st2common/st2common/config.py
+++ b/st2common/st2common/config.py
@@ -84,7 +84,7 @@ def register_opts(ignore_errors=False):
     do_register_opts(content_opts, 'content', ignore_errors)
 
     webui_opts = [
-        cfg.StrOpt('webui_base_url', default='https://%s' % socket.gethostname(),
+        cfg.StrOpt('webui_base_url', default='https://%s' % socket.getfqdn(),
                    help='Base https URL to access st2 Web UI. This is used to construct' +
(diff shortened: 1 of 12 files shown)


st2common/st2common/config.py
```diff
@@ -82,6 +82,14 @@ def register_opts(ignore_errors=False):
     ]
     do_register_opts(content_opts, 'content', ignore_errors)
 
+    webui_opts = [
+        cfg.StrOpt('webui_base_url', default='https://localhost',
```
emedvedev: `localhost` is never a good option: nobody ever installs St2 on a local machine, and even if they would, it would only be valid for one user in a channel.

What about using the output of `/bin/hostname` as a default? Yes, it will still be wrong in a lot of cases, but still better than `localhost` and seems to be the least bad option overall.

Kami: I'm personally not a big fan of that - it's yet another setting which needs to be maintain in sync and it doesn't seem to belong here.

Can we move this logic to st2chatops package / bot since it's ChatOps specific (user already needs to configure some environment variables to get bot to work)?

armab:
:-1: about including this as config parameter in `st2`

> So mostly we get “localhost” based URL.

If `ST2_WEBUI_URL` defaults to `localhost` in `st2chatops.env`, doesn't mean that you get always `localhost`. This setting is for users to configure.

For example I always change `ST2_WEBUI_URL` in my `st2chatops.env`

------
My point as a user that `st2chatops.env` should be used as Single Point of ChatOps Configuration.
So rely on `ST2_WEBUI_URL` from https://github.com/StackStorm/st2chatops/blob/master/st2chatops.env#L36 without any additional `st2` config parameters.

st2common/st2common/config.py
lakshmi-kannan: >Can we move this logic to st2chatops package / bot since it's ChatOps specific (user already needs to configure some environment variables to get bot to work)?

It was already there and there are two use cases one of them which it wasn't solving. I called it out in the description.

lakshmi-kannan:
@armab You probably overlooked the use cases and the implementation I mentioned in description. The use case (2) where a jinja template is rendered by the *backend* when there are final results (See https://github.com/StackStorm/st2/blob/master/contrib/chatops/actions/templates/default.j2#L15) cannot use an env variable set for chatops. Chatops package could be installed on an entirely different host. So there is a config parameter needed in backend. If you are telling me setting a config value in st2.conf is harder than setting an env variable, I won't buy that argument. Anyways, @emedvedev and I settled on this approach after discussion.

armab:
@lakshmi-kannan As said, you can keep & pass that URL when `hubot-stackstorm` initially sends trigger to run an action on st2 side. So the same env URL used in `ack` will be re-used in `result`.

I'm saying that for ChatOps the ideal is to set everything in one single place, rather then get into several configs: 
* Before: edit `st2chatops.env`
* After: edit both `st2chatops.env` and `st2.conf`

what's easier?

lakshmi-kannan:
>After: edit both st2chatops.env and st2.conf 

Here was my gist from prev comment:

```Once this is merged and tested, we'll eventually deprecate the env variable in st2chatops.env because it won't be needed.```

>As said, you can keep & pass that URL when hubot-stackstorm initially sends trigger to run an action on st2 side

Yes, this was considered. What we can pass is the base_url because the entir url needs the execution id which is generated by backend. So if you are going to pass just the base url on every http request, it seems rather wasteful when you can get it via just configuration.

armab:
>> After: edit both st2chatops.env and st2.conf 

> Here was my gist from prev comment:
> `Once this is merged and tested, we'll eventually deprecate the env variable in st2chatops.env because it won't be needed.`

This is not answering question about the usability.
Even if that setting will be moved from `st2chatops.env` to `st2.conf`, it won't be easier, since other env variables in `st2chatops.env` still exist and needs touching.

armab:
Apart of that, what if there are 2 different `st2chatops` hubot instances on different hosts and configured to work *in different network environments* talking to 1 central `st2` installation?
Previously they could pass 2 different envs: 
`ST2_WEBUI_URL=https://chatops.internal`
and
`ST2_WEBUI_URL=https://chatops.stackstorm.com`
for each `st2chatops` installation and benefit from it.

lakshmi-kannan:
>Apart of that, what if there are 2 different st2chatops hubot instances on different hosts and configured to work in different network environments talking to 1 central st2 installation?
Previously they could pass 2 different envs: 
ST2_WEBUI_URL=https://chatops.internal
and
ST2_WEBUI_URL=https://chatops.stackstorm.com
for each st2chatops installation and benefit from it.

ST2_WEBUI_URL should point to the web ui host and not chatops host. So in this case, both should point to https://webui.stackstorm.com.

>since other env variables in st2chatops.env still exist and needs touching.

Yes, those are adapter specific settings. Having the WEBUI_URL in chatops was already a confusing thing. Why should chatops even know about where webui is hosted? By having the backend store the info about history URL, we can also make things like CLI display the URL on st2 run. I see a lot of good things by having the URL in backend. Anyways, I am going to proceed my route and if people are still unhappy, they can fix it.

lakshmi-kannan:
>If Mongo doesn't validate a URL where the hostname is not an FQDN, this method still can (and will) crash it.

Yep, I just wanted to know if FQDN was the issue and it looks like it. I now changed URLField to StringField so things don't crash when FQDN is not available. It's not ideal but better than failing to write execution model to disk.

emedvedev:
LGTM

PR #2822, opened 2016-07-21
Oh my space

lakshmi-kannan:
Our Slack CI tests caught it but really we should have unit tests to validate it. Let me unbreak master and then here is a JIRA  https://stackstorm.atlassian.net/browse/STORM-2236 (sorry, internal link) to add a test for this jinja that I'll pick next.


Kami:
Yeah, agreed unit tests for this would be nice :)

PR #2823, opened 2016-07-21
[WIP] Use two BufferedDispatcher pools inside the action runner

### LINKED ISSUE #2814: Workflows with action policies can deadlock
Given the attached pack (useful for meditation), the system deadlocks when the action _dispatcher_pool (50 entries) gets filled with workflows, and all workflow actions are subsequently held in the _work_buffer queue indefinitely.

One possible solution would be to limit the number of workflows in the _dispatcher_pool to one less than half the total entries, in order to leave room for action execution.

[deadlock.tar.gz](https://github.com/StackStorm/st2/files/361933/deadlock.tar.gz)

# Deadlock pack

This pack contains actions for testing policy deadlock and other behaviors

## Actions

* ``delay2`` - a two second delay
* ``delay20`` - a twenty second delay
* ``delay2test`` - a workflow with delay2 followed by delay20

## Policies

* ``delay20`` - limit concurrency to 1
* ``delay2test`` - limit concurrency to 10

## Testing

Launch multiple instances of the workflow:

* ``for i in `seq 1 100`; do st2 action execute deadlock.delay2test; done``

Without the ``delay2test`` policy, the actions deadlock

Kami:
This pull request updates action runner message handler class to utilize two `BufferedDispatcher` pools - one of regular (non-workflow) actions and one for workflow actions. Size of both of those pools is also configurable by the operator.

This will increase the safety margins and make hitting deadlocks and issues like the one described in #2814 less likely.

It's also worth pointing out again that the actual pool sizes are very much user-specific (hardware on which the action runner process runs, number of action runners processes and workload aka type of actions).

Making those pool sizes configurable will also allow users to more efficiently utilize action runners and resources - e.g. if user has many concurrent lightweight actions and server CPU, etc. is far from the upper limit, the user can now bump the pool size and receive a better throughput without needing to spin up a new action runner process (of course that also depends on the actual scenario and workload, sometimes it's better and more effective to spin up a new action runner, etc.).

In addition to that I also updated `BufferedDispatcher` class to log an INFO message if a potential deadlock has been detected (if a pool is busy with no free works for more than some pre-defined period of seconds).

Resolves #2814.

## TODO

- [ ] Documentation

### DIFF
--- a/st2common/st2common/models/db/liveaction.py
+++ b/st2common/st2common/models/db/liveaction.py
@@ -59,6 +59,9 @@ class LiveActionDB(stormbase.StormFoundationDB):
     action = me.StringField(
         required=True,
         help_text='Reference to the action that has to be executed.')
+    is_action_workflow = me.BooleanField(
+        default=False,
+        help_text='A flag indicating whether the references action is a workflow.')
     parameters = stormbase.EscapedDynamicField(
         default={},
         help_text='The key-value pairs passed as to the action runner & execution.')
--- a/st2common/st2common/services/action.py
+++ b/st2common/st2common/services/action.py
@@ -98,6 +98,9 @@ def create_request(liveaction):
     liveaction.status = action_constants.LIVEACTION_STATUS_REQUESTED
     liveaction.start_timestamp = date_utils.get_datetime_utc_now()
 
+    # Set the "is_action_workflow" attribute
+    liveaction.is_action_workflow = action_db.is_workflow()
+
     # Publish creation after both liveaction and actionexecution are created.
     liveaction = LiveAction.add_or_update(liveaction, publish=False)
 
--- a/st2common/st2common/util/greenpooldispatch.py
+++ b/st2common/st2common/util/greenpooldispatch.py
@@ -16,16 +16,22 @@
 import eventlet
 import Queue
 
+__all__ = [
+    'BufferedDispatcher'
+]
+
 
 class BufferedDispatcher(object):
 
     def __init__(self, dispatch_pool_size=50, monitor_thread_empty_q_sleep_time=5,
-                 monitor_thread_no_workers_sleep_time=1):
+                 monitor_thread_no_workers_sleep_time=1, name=None):
         self._pool_limit = dispatch_pool_size
         self._dispatcher_pool = eventlet.GreenPool(dispatch_pool_size)
         self._dispatch_monitor_thread = eventlet.greenthread.spawn(self._flush)
         self._monitor_thread_empty_q_sleep_time = monitor_thread_empty_q_sleep_time
         self._monitor_thread_no_workers_sleep_time = monitor_thread_no_workers_sleep_time
+        self._name = name
+
         self._work_buffer = Queue.Queue()
 
     def dispatch(self, handler, *args):
@@ -49,3 +55,11 @@ class BufferedDispatcher(object):
         while not self._work_buffer.empty() and self._dispatcher_pool.free() > 0:
             (handler, args) = self._work_buffer.get_nowait()
             self._dispatcher_pool.spawn(handler, *args)
+
+    def __repr__(self):
+        name = self._name or id(self)
+        values = (name, self._pool_limit, self._monitor_thread_empty_q_sleep_time,
+                  self._monitor_thread_no_workers_sleep_time)
+        return ('<BufferedDispatcher name=%s,dispatch_pool_size=%s,'
+                'monitor_thread_empty_q_sleep_time=%s,monitor_thread_no_workers_sleep_time=%s>' %
+                values)
--- a/st2common/st2common/util/greenpooldispatch.py
+++ b/st2common/st2common/util/greenpooldispatch.py
@@ -13,12 +13,27 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import time
+
 import eventlet
 import Queue
+
+from st2common import log as logging
 
 __all__ = [
     'BufferedDispatcher'
 ]
+
+# If the thread pool has been occupied with no empty threads for more than this number of seconds
+# a message will be logged
+POOL_BUSY_THRESHOLD_SECONDS = 60
+
+POOL_BUSY_LOG_MESSAGE = """
+BufferedDispatcher pool "%s" has been busy with no free threads for more than %s seconds. If there \
+are server resources available, consider increasing the dispatcher pool size in the config.
+""".strip()
+
+LOG = logging.getLogger(__name__)
 
 
 class BufferedDispatcher(object):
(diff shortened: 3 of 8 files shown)


Kami:
On thing we could eventually also do in such scenario is return message back to the queue.

If user runs multiple action runner processes this should cause RabbitMQ to deliver this message to another runner...

Kami:
Docs now at https://github.com/StackStorm/st2docs/pull/217.

CHANGELOG.rst
```diff
@@ -16,6 +16,12 @@ In development
   is now exposed to let st2 know about the location of web ui. This is used to construct the
   web url. Chatops messages can now display the history url from the execution model they
   get as response to API calls.
+* Update action runner to use two internal green thread pools - one for regular (non-workflow) and
```
lakshmi-kannan: s/non-workflow/non-workflows :)

CHANGELOG.rst
```diff
@@ -16,6 +16,12 @@ In development
...
+  one for workflow actions. Sizes of both of those pools are also user configurable inside the
```
lakshmi-kannan: s/workflow actions/workflows :)

st2common/st2common/models/db/liveaction.py
```diff
@@ -59,6 +59,9 @@ class LiveActionDB(stormbase.StormFoundationDB):
     action = me.StringField(
         required=True,
         help_text='Reference to the action that has to be executed.')
+    is_action_workflow = me.BooleanField(
```
lakshmi-kannan: Dude, is_workflow is better. is_action_workflow is confusing. Sorry for the nit :).

st2common/st2common/models/db/liveaction.py
```diff
@@ -59,6 +59,9 @@ class LiveActionDB(stormbase.StormFoundationDB):
...
+        default=False,
+        help_text='A flag indicating whether the references action is a workflow.')
```
lakshmi-kannan: s/references/referenced

st2common/st2common/services/action.py
```diff
@@ -98,6 +98,9 @@ def create_request(liveaction):
     liveaction.status = action_constants.LIVEACTION_STATUS_REQUESTED
     liveaction.start_timestamp = date_utils.get_datetime_utc_now()
 
+    # Set the "is_action_workflow" attribute
+    liveaction.is_action_workflow = action_db.is_workflow()
```
lakshmi-kannan: See, you used is_workflow 🎯

st2common/st2common/util/greenpooldispatch.py
```diff
@@ -45,7 +73,23 @@ def _flush(self):
 
     def _flush_now(self):
         if self._dispatcher_pool.free() <= 0:
+            now = time.time()
+
+            if (now - self._pool_last_free_ts) >= POOL_BUSY_THRESHOLD_SECONDS:
+                LOG.info(POOL_BUSY_LOG_MESSAGE % (self.name, POOL_BUSY_THRESHOLD_SECONDS))
```
lakshmi-kannan: Should this be warning?

lakshmi-kannan:
>On thing we could eventually also do in such scenario is return message back to the queue.

That's an excellent point.

lakshmi-kannan:
It's funny there are no end-to-end tests to cover the case where we can show deadlock won't occur. It seems doable :).

conf/st2.conf.sample
```diff
@@ -6,17 +6,21 @@
 enable = True
```
lakshmi-kannan: Can you add a commented config section in other conf files as well? We've been repeatedly been asked about it in support.

lakshmi-kannan:
Aside from the nits and one comment about end-to-end tests, change is awesome! 👍

CHANGELOG.rst
LindsayHill: How about we re-write these sentences like this:
* Both pool sizes are user-configurable. This should help increase the throughput of a single action runner when the system is not over-utilized. It can also help prevent deadlocks which may occur when using delay policies with action-chain workflows. (improvement)

conf/st2.conf.sample
Kami: That would be a bigger change and not part of this PR since it needs to be automated like this config.

Kami: Well, first I also need to fully automate this one.

This means adding a step to Makefile so we don't forget to update it like we did many times now :P

CHANGELOG.rst
Kami: Changed, thanks.

st2common/st2common/models/db/liveaction.py
Kami: Yeah, I was thinking about it, but the attribute refers to the related action and not LiveAction. Another option also is `action_is_workflow` :P

WDYT?

We could also do something like we do with `runner_info` attribute (e.g. `action_info`), but I'm not too big of a fan of free for all dictionaries since it's hard to see what are possible attributes, etc.

st2common/st2common/models/db/liveaction.py
Kami: Fixed, thanks.

st2common/st2common/util/greenpooldispatch.py
Kami: I was thinking about it, but I decided to use INFO.

The thing is that this could represent a problem (warning), but this message being logged doesn't always represent an actual problem so I decided to use INFO.

st2common/st2common/models/db/liveaction.py
Kami: Now that I think about it again, I do like `action_is_workflow` more.

conf/st2.conf.sample
Kami: Here we go - 40018836044f1942208c7bff23678aa170358cd0, ccbfb0e566272c26e069ee584125fab064831725

Kami:
Re end to end tests - I was thinking about it, but it's a bit hard, even just the setup.

To be able to reproduce it in a reasonable time frame, etc. we would need to spin up only one action runner with a small max pool size and then asserting that it dead locked is also a bit hard...

Makefile
```diff
@@ -62,6 +62,17 @@ checklogs:
 .PHONY: pylint
 pylint: requirements .pylint
 
+.PHONY: configgen
+configgen:
```
manasdk: finally!

Kami: It's alive!

Lol, yeah, I was actually hoping someone else will do it before I need to do it :P

manasdk:
Sorry for chiming in super late to this party!

I suspect the real issue is with ActionChainRunner impl(who implemented that crap) which needs to really stop spinning on action updates. Specifically https://github.com/StackStorm/st2/blob/master/st2actions/st2actions/runners/actionchainrunner.py#L540  is bad. Might we worth looking at fixing ActionChain to suspend/resume a WF following the completion of an Action.

A crazy option would also be to continue supporting the ActionChain syntax but use mistral to run by translating AC to mistral WF.

> This will increase the safety margins and make hitting deadlocks and issues like the one described in #2814 less likely.

This almost always means its a problem for future selves 👯

PR #2824, opened 2016-07-22
Pluggable runners

bigmstone:
This will add the ability to treat runners as plugins which should simplify the development process for creating new runners where needed.

- [x] Move runners to and load runners from contrib
- [x] Move existing dictionary to individual manifest files describing each runner
- [x] Modify action execution to call runners from new location
- [x] Finish Mistral Query
- [x] Acceptance testing
- [x] Add testing where needed
- [x] Get tests passing
- [x] Clean up code/add comments where needed/prep for master


contrib/runners/actionchainrunner/runner.yaml
```diff
@@ -0,0 +1,14 @@
+- aliases: []
+  description: A runner for launching linear action chains.
+  enabled: true
+  name: action-chain
+  runner_module: st2actions.runners.actionchainrunner
+  runner_parameters:
```
lakshmi-kannan: We should call it parameters and in the code we can call it runner_parameters. UX feedback.

contrib/runners/actionchainrunner/runner.yaml
```diff
@@ -0,0 +1,14 @@
+- aliases: []
+  description: A runner for launching linear action chains.
+  enabled: true
+  name: action-chain
+  runner_module: st2actions.runners.actionchainrunner
```
lakshmi-kannan: We call it entry_point for other content. But if you are going to use stevedore approach, some of this would change.

contrib/runners/cloudslang/runner.yaml
```diff
@@ -0,0 +1,16 @@
+- aliases: []
+  description: A runner to execute cloudslang flows.
+  enabled: true
+  name: cloudslang
```
m4dcoder: Can we drop cloud slang and move it to st2contrib?

m4dcoder:
We should also abstract callback handlers for the runners (i.e. mistral callback handler).

st2common/st2common/content/loader.py
```diff
@@ -204,6 +240,36 @@ def _get_folder(self, pack_dir, content_type):
             return None
         return path
 
+    def _get_runners_from_dir(self, base_dir):
+        result = {}
+        for runner_name in os.listdir(base_dir):
```
Kami: Perhaps we should sort the `os.listdir` result to always register runners in the same order? Or maybe it's not a big deal.

st2common/st2common/content/loader.py
```diff
@@ -204,6 +240,36 @@ def _get_folder(self, pack_dir, content_type):
...
+            runner_dir = os.path.join(base_dir, runner_name)
+            runner_manifest_file = os.path.join(runner_dir, RUNNER_MANIFEST_FILE_NAME)
+
+            if os.path.isdir(runner_dir) and os.path.isfile(runner_manifest_file):
```
Kami: Probably a good idea to add a DEBUG log statement when we found a valid directory and also when we don't (e.g. missing manifest file).

st2common/st2common/content/loader.py
Kami: Do we need to do some sanitation with the directory name (we do it for packs)?

Kami: On a related note, we have a similar "issue" with packs - we actually use pack directory name as pack name ID, even though there is a "name" attribute in the metadata file (the latter is only used for display in API).

It's slightly confusing, but yeah for consistency we should probably do the same for runners, but the problem is runners names right now contain dashes, but for consistency with packs, etc. we should only allow underscores in the directory name.

Anyway, we can probably come up with some compromise, similar as we do with pack name alias right now for backward compatibility.

contrib/runners/remote_command_runner/runner.yaml
```diff
@@ -0,0 +1,80 @@
+- aliases:
+  - run-remote
```
Kami: While we are at it, we should also get rid of the old aliases which have been deprecated for a while now.

Let's just make sure upgrade notes is updated and all the code in st2contrib as well (if there is any action which still uses old and deprecated runner name).

contrib/runners/mistral/runner.yaml
```diff
@@ -2,8 +2,8 @@
   description: A runner for executing mistral v2 workflow.
   enabled: true
   name: mistral-v2
-  query_module: st2actions.query.mistral.v2
-  runner_module: st2actions.runners.mistral.v2
+  query_module: mistralv2
+  runner_module: mistralv2
```
Kami: Does the new infrastructure support sub-directories?

E.g. mistral.v2 would map to `<runners path>/mistral/v2.py` or similar.

If not, it would be nice if we can also make support that, but in the worst case we can call it mistral_v2 for now :)

contrib/runners/pythonrunner/pythonrunner.py
```diff
@@ -0,0 +1,247 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
```
Kami: Another thing - let's make sure that those files are still linted (might need to update lint targets in the makefile, but we already lint some other Python files in that directory, might just need to add contrib/runners/*/.py to it or similar).

bigmstone:
After fighting with my env most of today to get tests running faster I am stuck at how to best move the tests around for runners.

One problem is the runners are in `contrib` now so they must be imported by explicit path. I could use the loader to do this, but the loader by default points to the base_dir known by st2_config. So I'd have to have a loader for tests that uses relative paths. Not the end of the world, but not ideal.

Second problem is using mock's patch decorator. It takes module import path as a string. Don't have a solution for this one yet, so open to suggestions.

I also lean towards having runner test in the directory for the runner. But, assuming that's not part of the solution to the above, doesn't have to be done in this PR.

bigmstone:
I've worked through most of the testing issues. Still have a few hanging around. Any insights welcome.

bigmstone:
The tests are expecting the action to be in a different state from `succeeded` but they're all succeeding (Instead of `failed`, `delayed`, etc.). Only tests that are failing have the `mock.patch.object` of the `LocalShellRunner` object in common. I've looked over all the changes I've made now 3 times and can't find anything that would screw up concurrency, but that's where my gut is taking me right now. Will continue looking, but would love an extra set of eyes.

m4dcoder:
@bigmstone here's a pointer to the problem you're seeing.  the mock @ https://github.com/StackStorm/st2/blob/master/st2actions/tests/unit/policies/test_concurrency.py#L59 is suppose to keep the action in a running state.  this makes sure we can test the threshold and concurrency.  in this branch, the action executions are moving to succeeded state.   so the changes has affected how that mock is suppose to work.  i'll let you pick this up from here.

bigmstone:
After much digging it would appear I'm no longer patching the same instance of `LocalSSHRunner` which is a side effect of loading the runners via `imp`. So I need a way to patch the instance of the runner being imported by the runner loader.

bigmstone:
Tests finally nailed down. (Halalurrya)

Still TODO:
* Finish Mistral Query
* Acceptance testing
* Clean up/Add tests

Maybe, needs discussion on if we do this in the current PR:
* Migrate away from `register_runner` and use existing loader to return class.

st2actions/st2actions/runners/pythonrunner.py
```diff
@@ -1,272 +0,0 @@
-# Licensed to the StackStorm, Inc ('StackStorm') under one or more
```
Kami: Yeah we need to verify the symlink works as expected end to end (it might cause issues with virtual environments...).

One option would also be to have a Python module here which basically just imports everything from new location and re-exports it, so:

```python
__all__ = [
    'Action'
]


from st2actions.runners.base_action import Action
```

This is probably also better from cross-OS compatibility perspective :)

bigmstone: Yes, I like this. I also think I will do this for the test runners. This is much cleaner than throwing symlinks around everywhere.

st2common/st2common/content/loader.py
```diff
@@ -15,22 +15,59 @@
...
 __all__ = [
+    'ContentRunnerLoader',
     'ContentPackLoader',
     'MetaLoader'
 ]
 
 LOG = logging.getLogger(__name__)
 
 
+class ContentRunnerLoader(object):
```
bigmstone: As discussed stevedore is -1 and at the very least not happening in this PR. (I realize this is already decided, but I'm leaving comments on all existing comments so we can get this ready to close out.)

contrib/runners/cloudslang/runner.yaml
bigmstone: IMO this PR is big enough. I agree this should/could be dropped, but should be done in a different PR.

st2common/st2common/content/loader.py
bigmstone: I can't conceive of a time when this would be a big deal, but would be trivial to implement.

st2actions/st2actions/runners/pythonrunner.py
bigmstone: Done, and unfortunately can't do it for test fixtures because of manifest file.

st2common/st2common/content/loader.py
bigmstone: Done

contrib/runners/remote_command_runner/runner.yaml
bigmstone: Done

bigmstone: I removed this, but it had a cascading effect. Would result in a few more files being changed. I changed it back because I think this PR is big enough as it is. Can make this change in another PR.

contrib/runners/mistral/runner.yaml
bigmstone: It doesn't currently support sub-directories because the modules are loaded by `<base_path>/runners/<runner_module>/<runner_module>.py`. We could change it to use `<base_path>/runners/<runner_name>/<runner_module>.py`

What is the goal? Is the goal to load various modules from the same folder or to be able to have sub-directories inside of runners? For instance `<base_path>/runners/<runner_name>/<some>/<new>/<path>/<runner_module>.py`?

st2common/st2common/content/loader.py
bigmstone: We have runner alias as well. So it would be trivial to name, for instance, `mistral-v2` to `mistral_v2` and add an alias for `mistral-v2`. This also ties into [your other comment](https://github.com/StackStorm/st2/pull/2824#pullrequestreview-967994).

What type sanitation of the directory name are you thinking? Enforcing not using dash, for instance? I didn't see anything obvious in `loader.py` that adds any additional sanitation for pack loading.

st2common/st2common/content/loader.py
Kami: Maybe just RunnersLoader / RunnerMetadataLoader or similar is a better name? "ContentRunner" makes it sound a bit confusing, at least to me.

With prefix pack loader with content because we used to call them "content packs" (these days days we mostly just call them packs or integration packs).

st2common/st2common/constants/runners.py
```diff
@@ -44,7 +49,7 @@
 
 REMOTE_RUNNER_PRIVATE_KEY_HEADER = 'PRIVATE KEY-----'.lower()
 
-# Python runner
+# Python runner'
```
Kami: How did this ' get here :P

st2common/st2common/content/loader.py
Kami: Yep, not allowing dash at least.

Here is a regex we use for pack names - `PACK_NAME_WHITELIST = r'^[A-Za-z0-9_-]+'`.

st2common/st2common/config.py
```diff
@@ -75,11 +75,16 @@ def register_opts(ignore_errors=False):
...
+    system_runners_base_path = os.path.join(cfg.CONF.system.base_path, 'runners')
     content_opts = [
         cfg.StrOpt('system_packs_base_path', default=system_packs_base_path,
                    help='Path to the directory which contains system packs.'),
+        cfg.StrOpt('system_runners_base_path', default=system_runners_base_path,
+                   help='Path to the directory which contains system runners.'),
         cfg.StrOpt('packs_base_paths', default=None,
-                   help='Paths which will be searched for integration packs.')
+                   help='Paths which will be searched for integration packs.'),
+        cfg.StrOpt('runners_base_paths', default=None,
```
Kami: I'm fine with two settings for consistency with packs, but I'm also OK with just having a single setting for simplicity (user runners can just live in the same directory as system ones).

Anyways, it's fine as it is now.

st2common/st2common/content/bootstrap.py
```diff
@@ -57,6 +57,7 @@ def register_opts():
         cfg.BoolOpt('configs', default=False, help='Register and load pack configs.'),
 
         cfg.StrOpt('pack', default=None, help='Directory to the pack to register content from.'),
+        cfg.StrOpt('runner', default=None, help='Directory to load runners from.'),
```
Kami: Hm, should this be called `runners-path` or similar instead?

st2common/st2common/runners/__init__.py
```diff
@@ -0,0 +1,207 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
```
Kami: Minor thing, I would prefer to put this into `base.py` file and leave __init__.py empty.

bigmstone: Would you like this changed in this PR or a later one? I was just leaving it as it was for now, but happy to change it.

st2common/st2common/content/bootstrap.py
bigmstone: That makes more sense to me, I was just following the contention set by `pack`.

st2common/st2common/content/loader.py
bigmstone: Done. Will push in next commit. (Putting this here now to keep track)

st2common/st2common/content/loader.py
bigmstone: Got it. Didn't break open the `setup_pack_virtualenv` function which would appear where we do this (In addition to when you setup venv with pack mgmt).

Kami:
Thanks.

It looks good to me right now, but before merging I would prefer to spend some time on it and test it locally. In addition to that, I think we should merge it once at least one of us has a day to spend on the things which might break after the merge (aka end to end stuff).

bigmstone:
@kami agree. I won't get a chance to work on it this weekend but I will plan to start this on Monday. I'm going to begin work on the new runner next week so I can have both going at the same time. If you or someone else can do the same I would appreciate it. If we find anything I can spend time fixing it and will plan to be ready to focus exclusively on it if needed when we merge to master.

contrib/runners/remote_command_runner/runner.yaml
```diff
@@ -15,11 +15,11 @@
       description: Arbitrary Linux command to be executed on the remote host(s).
       type: string
     cwd:
-      default: "~/"
+      default: "/tmp"
```
Kami: How does this change affect things?

I would imagine it's backward incompatible in some scenarios - /tmp is not necessarily always accessible to all the users. In any case, I believe it should be at least fine for our bootstrap installations.

st2common/st2common/content/bootstrap.py
```diff
@@ -172,19 +194,10 @@ def register_actions():
...
         LOG.info('############## Registering actions ######################')
         LOG.info('=========================================================')
-        runners_registrar.register_runner_types(experimental=cfg.CONF.experimental)
-    except Exception as e:
-        LOG.warning('Failed to register runner types: %s', e, exc_info=True)
-        LOG.warning('Not registering stock runners .')
-        return
-
-    # 2. Register actions
-    try:
```
bigmstone: https://github.com/StackStorm/st2/commit/0c5f5aab8fc8ab819889a65d9dab6bca815c45d7#diff-499d9f5cfe40e44d821453ced745e3b1R352

PR #2826, opened 2016-07-24
Add test for ActionsQueueConsumer

Kami:
Simple test for #2823 which verifies that the right dispatcher class is used.


lakshmi-kannan:
sweet! lgtm.

Issue #2806, opened 2016-07-12
Inconsistent behavior from st2 execution list with -n flag


johandahlberg:
Setting `-n 0` does not list all executions. See discussion in: https://github.com/StackStorm/st2/pull/2803

```
usage: st2 execution list [-h] [-t TOKEN] [-j] [-n LAST] [--action ACTION]
                          [--status STATUS]
                          [--trigger_instance TRIGGER_INSTANCE]
                          [-tg TIMESTAMP_GT] [-tl TIMESTAMP_LT] [-l]
                          [-a ATTR [ATTR ...]] [-w WIDTH [WIDTH ...]]

Get the list of the 50 most recent action executions.

optional arguments:
[...]
  -n LAST, --last LAST  List N most recent action executions; list all if 0.
[...]
```

Kami:
To get back to the comment in https://github.com/StackStorm/st2/pull/2803#issuecomment-231974048.

I just had a look and if you don't provide a limit or provide a value which is > 100, you will get back a maximum of 100 items since this is the value for the max limit.

Same goes if you specify 0 for limit - you get back last `100` executions

As discussed in the other ticket, I will make this value user-configurable in st2.conf, but default it to 100 for safety and backward compatibility reasons.

Kami:
@johandahlberg #2832 should work for your use case, right?

And of course updating the argument comment (removing "list all if 0").

johandahlberg:
@Kami It looks like it should work for us. Thanks!

Kami:
You are welcome and glad to hear it.

(#2832 has been merged into master)

PR #2803, opened 2016-07-11
Pass limit 0 on in parameters to client

johandahlberg:
I noticed in the CLI that if I ran: `st2 --debug execution list -n 0`, no limit was set in the called url, and only the last 100 executions were returned. According to the documentation if the `-n/--last` parameter is set to 0 all executions should be returned. I believe that this PR should fix this. 

I was also wondering if perhaps settling the limit to less than zero should raise an exception, but now I left the behavior as was.

Have a look and let me know what you think.

PR #2832, opened 2016-07-26
Allow max_page size (value of ?limit param) to be configured by the administrator

Kami:
As discussed in #2803 and #2806, this PR makes maximum page size aka maximum value of `?limit` query parameter configurable by the StackStorm administrator.

For backward compatibility and safety reasons (to prevent DDoS, etc.) we still default this value to `100`, but if user as a need to use a lower / higher value, they can now tune this setting.

### DIFF
--- a/st2api/st2api/controllers/resource.py
+++ b/st2api/st2api/controllers/resource.py
@@ -18,6 +18,7 @@
 import abc
 import copy
 
+from oslo_config import cfg
 from mongoengine import ValidationError
 import pecan
 from pecan import rest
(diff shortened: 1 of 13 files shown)


st2api/st2api/controllers/resource.py
```diff
@@ -119,6 +123,9 @@ def _get_all(self, exclude_fields=None, sort=None, offset=0, limit=None, query_o
         offset = int(offset)
 
         if limit and int(limit) > self.max_limit:
+            # TODO: We should throw here, I don't like this.
```
Kami: As mentioned in the comment, I'm not fan of this "silently ignore invalid value" approach.

Imo, we should throw `ValueError` with appropriate message which results in 400 status code.

What do others think?

Kami: Here we go - 3d26b4826d9d023bc1694b7e68568b3b4aff94df.

lakshmi-kannan:
LGTM. Nice!

Issue #2742, opened 2016-06-13
python action should be allowed to return status


manasdk:
Reporting back an error status along with the some data is useful for a python action. 

```
class MyAction(Action):
    ...
    def run(self, param):
        ...
        return False, result
```

This would allow extra data to be returned even if there were a failure in the action.

PR #2835, opened 2016-07-27
[RFR]: Enhancement: python action should be allowed to return status #2742

humblearner:
**Changed files:**
- st2actions/st2actions/runners/python_action_wrapper.py
- st2actions/st2actions/runners/pythonrunner.py
- st2tests/st2tests/resources/packs/pythonactions/actions/pascal_row.py
- st2actions/tests/unit/test_pythonrunner.py
 

**Test Cases:**
_Modified_:
- test_simple_action_without_status
- test_stdout_interception_and_parsing

_New_:
- test_simple_action_with_status_succeeded
- test_simple_action_with_status_failed
- test_simple_action_with_status_failed_result_none

**Caveats:**
~~- Assumption: User must provide status as "succeeded" or "failed" (case insensitive). It may fail otherwise. No check for this.~~

### DIFF
--- a/st2actions/st2actions/runners/python_action_wrapper.py
+++ b/st2actions/st2actions/runners/python_action_wrapper.py
@@ -114,14 +114,27 @@ class PythonActionWrapper(object):
     def run(self):
         action = self._get_action_instance()
         output = action.run(**self._parameters)
+        action_status = None
+
+        if type(output) is tuple and len(output) is 2:
+            action_status = output[0]
+            action_result = output[1]
+        else:
+            action_result = output
+
+        action_output = {"result": action_result}
 
         # Print output to stdout so the parent can capture it
         sys.stdout.write(ACTION_OUTPUT_RESULT_DELIMITER)
         print_output = None
         try:
-            print_output = json.dumps(output)
+            if action_status:
+                action_output['status'] = action_status
+                print_output = json.dumps(action_output)
+            else:
+                print_output = json.dumps(action_output)
         except:
-            print_output = str(output)
+            print_output = str(action_output)
         sys.stdout.write(print_output + '\n')
         sys.stdout.write(ACTION_OUTPUT_RESULT_DELIMITER)
 
@@ -169,7 +182,6 @@ if __name__ == '__main__':
     parent_args = json.loads(args.parent_args) if args.parent_args else []
 
     assert isinstance(parent_args, list)
-
     obj = PythonActionWrapper(pack=args.pack,
                               file_path=args.file_path,
                               parameters=parameters,
--- a/st2actions/st2actions/runners/python_action_wrapper.py
+++ b/st2actions/st2actions/runners/python_action_wrapper.py
@@ -16,7 +16,7 @@
 import sys
 import json
 import argparse
-
+import traceback
 from oslo_config import cfg
 
 from st2common import log as logging
@@ -133,8 +133,12 @@ class PythonActionWrapper(object):
             action_output['status'] = action_status
             print_output = json.dumps(action_output)
         else:
-            raise InvalidStatusException('Status returned from the action must'
-                                         ' either be True or False.')
+            try:
+                raise InvalidStatusException('Status returned from the action'
+                                             ' must either be True or False.')
+            except:
+                traceback.print_exc()
+                sys.exit(220)
 
         sys.stdout.write(print_output + '\n')
         sys.stdout.write(ACTION_OUTPUT_RESULT_DELIMITER)
--- a/st2actions/st2actions/runners/python_action_wrapper.py
+++ b/st2actions/st2actions/runners/python_action_wrapper.py
@@ -66,7 +66,7 @@ class ActionService(object):
 
     def set_value(self, name, value, ttl=None, local=True, scope=SYSTEM_SCOPE, encrypt=False):
         return self._datastore_service.set_value(name, value, ttl, local, scope=scope,
-            encrypt=encrypt)
+                                                 encrypt=encrypt)
 
     def delete_value(self, name, local=True, scope=SYSTEM_SCOPE):
         return self._datastore_service.delete_value(name, local)
@@ -115,8 +115,7 @@ class PythonActionWrapper(object):
         action = self._get_action_instance()
         output = action.run(**self._parameters)
         action_status = None
-
-        if type(output) is tuple and len(output) is 2:
+        if type(output) is tuple and len(output) == 2:
             action_status = output[0]
             action_result = output[1]
         else:
@@ -127,14 +126,14 @@ class PythonActionWrapper(object):
         # Print output to stdout so the parent can capture it
         sys.stdout.write(ACTION_OUTPUT_RESULT_DELIMITER)
         print_output = None
-        try:
-            if action_status:
-                action_output['status'] = action_status
-                print_output = json.dumps(action_output)
-            else:
-                print_output = json.dumps(action_output)
-        except:
-            print_output = str(action_output)
+        if action_status is None:
+            print_output = json.dumps(action_output)
+        elif type(action_status) is bool:
+            action_output['status'] = action_status
+            print_output = json.dumps(action_output)
+        else:
+            raise Exception('Status should either be True or False.')
+
         sys.stdout.write(print_output + '\n')
         sys.stdout.write(ACTION_OUTPUT_RESULT_DELIMITER)
 
--- a/st2actions/st2actions/runners/python_action_wrapper.py
+++ b/st2actions/st2actions/runners/python_action_wrapper.py
@@ -30,6 +30,7 @@ from st2common.constants.action import ACTION_OUTPUT_RESULT_DELIMITER
 from st2common.constants.keyvalue import SYSTEM_SCOPE
 from st2common.service_setup import db_setup
 from st2common.services.datastore import DatastoreService
+from st2common.exceptions.invalidstatus import InvalidStatusException
 
 __all__ = [
     'PythonActionWrapper',
@@ -132,7 +133,8 @@ class PythonActionWrapper(object):
             action_output['status'] = action_status
             print_output = json.dumps(action_output)
         else:
-            raise Exception('Status should either be True or False.')
+            raise InvalidStatusException('Status returned from the action must'
+                                         ' either be True or False.')
 
         sys.stdout.write(print_output + '\n')
         sys.stdout.write(ACTION_OUTPUT_RESULT_DELIMITER)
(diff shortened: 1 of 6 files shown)


Kami:
Awesome, thanks, will look into it shortly.

Kami:
As far as returning the status goes - I thought we only want to allow user to return a flag which indicates a success or a failure.

So (succeeded, result), e.g. `(True, result)`. Allowing user to return other statuses beyond succeeded and failed could cause issues and so using a string instead of a boolean flag is probably not needed.

st2actions/st2actions/runners/pythonrunner.py
```diff
@@ -158,32 +157,50 @@ def run(self, action_parameters):
...
         else:
-            result = None
-
+            action_result = None
         try:
-            result = json.loads(result)
+            action_result = json.loads(action_result)
         except:
             pass
+        action_status = None
```
Kami: This could perhaps be refactored in a separate utility method to make it a bit easier to read and test.

st2actions/st2actions/runners/pythonrunner.py
```diff
@@ -158,32 +157,50 @@ def run(self, action_parameters):
...
-
         if error:
             output['error'] = error
-
-        if exit_code == 0:
-            status = LIVEACTION_STATUS_SUCCEEDED
-        elif timed_out:
-            status = LIVEACTION_STATUS_TIMED_OUT
+        status = None
+        if action_status:
```
Kami: Same here.

I think it could also be simplified a bit (shouldn't need that many if / else branches).

Kami:
I've added some comments.

Overall it looks good. :+1: 

It just needs some small polish and we need to decide how to handle the status (do we want to use a string or should we just use a boolean flag).

After that's done, we also need to update the documentation.

st2actions/st2actions/runners/python_action_wrapper.py
```diff
@@ -114,14 +114,27 @@ def __init__(self, pack, file_path, parameters=None, user=None, parent_args=None
     def run(self):
         action = self._get_action_instance()
         output = action.run(**self._parameters)
+        action_status = None
+
+        if type(output) is tuple and len(output) is 2:
```
Kami: Safer to use `==` when comparing strings and numbers (`is` should only be used for built in types such as True, False, None).

Well in theory `is` should also work fine for small integers since they tend to always contain the same reference, but it's still a bit dangerous and implementation specific.

st2actions/st2actions/runners/python_action_wrapper.py
```diff
@@ -114,14 +114,27 @@ def __init__(self, pack, file_path, parameters=None, user=None, parent_args=None
...
+
+        action_output = {"result": action_result}
 
         # Print output to stdout so the parent can capture it
         sys.stdout.write(ACTION_OUTPUT_RESULT_DELIMITER)
         print_output = None
         try:
-            print_output = json.dumps(output)
+            if action_status:
+                action_output['status'] = action_status
```
Kami: This could perhaps also be simplified a bit by putting `if action_status: ...` clause above, then we don't need the else here and can always just call `json.dumps` on action_output.

st2actions/st2actions/runners/python_action_wrapper.py
```diff
@@ -114,14 +114,27 @@ def __init__(self, pack, file_path, parameters=None, user=None, parent_args=None
...
+                print_output = json.dumps(action_output)
+            else:
+                print_output = json.dumps(action_output)
         except:
-            print_output = str(output)
+            print_output = str(action_output)
```
Kami: Hm, do we indeed want to use `action_output` here or would the right thing be to just use `action_result`?

Kami:
And let's also not forget to update changelog once everything else is addressed :)

Even though this change is not backward-incompatible, it might still be a good idea to update "Upgrade Notes" section in the documentation as well (encourage users to utilize new "notation"...).

humblearner:
Thank you for review @Kami. I will make the changes and get back to you.

humblearner:
### 	Updates as per the comments:

- https://github.com/StackStorm/st2/pull/2835#issuecomment-235508641: Changed the status to flag instead of string.
- Refactored as per: https://github.com/StackStorm/st2/pull/2835#discussion_r72392855, https://github.com/StackStorm/st2/pull/2835#discussion_r72392855, https://github.com/StackStorm/st2/pull/2835#discussion_r72392855

**Test cases:**
Modified test cases accordingly and added two new test cases: 
- test_simple_action_with_result_as_None_no_status
- test_simple_action_timeout

**TODO**:
- Update Documentation and Upgrade Notes

st2actions/st2actions/runners/python_action_wrapper.py
```diff
@@ -114,14 +114,26 @@ def __init__(self, pack, file_path, parameters=None, user=None, parent_args=None
...
-            print_output = json.dumps(output)
-        except:
-            print_output = str(output)
+        if action_status is None:
+            print_output = json.dumps(action_output)
+        elif type(action_status) is bool:
+            action_output['status'] = action_status
+            print_output = json.dumps(action_output)
+        else:
+            raise Exception('Status should either be True or False.')
```
Kami: Can you please make exception message a bit more explicit and user-friendly (e.g. "Status returned from the action must either be True or False").

In addition to that, let's please also add a test case for  this scenario (invalid status being returned from the Python action).

st2actions/st2actions/runners/pythonrunner.py
```diff
@@ -158,34 +168,61 @@ def run(self, action_parameters):
...
+            action_result = None
 
         try:
-            result = json.loads(result)
+            action_result = json.loads(action_result)
         except:
             pass
+        action_status = None
+        if action_result and action_result != "None":
+            result = action_result['result']
```
Kami: This could also throw KeyError in case `result` key is not present in the dictionary, right?

st2actions/st2actions/runners/pythonrunner.py
```diff
@@ -158,34 +168,61 @@ def run(self, action_parameters):
...
+            try:
+                action_status = action_result['status']
+            except:
```
Kami: Probably a good idea to explicitly catch `KeyError` here.

st2actions/st2actions/runners/pythonrunner.py
humblearner: `result` key should always be there in dictionary : https://github.com/StackStorm/st2/pull/2835/files#diff-f1f34011cf21592061e7b15f398f4e3fR124

st2actions/st2actions/runners/python_action_wrapper.py
humblearner: I was about to add test case for that scenario. But the execution fails at : https://github.com/StackStorm/st2/pull/2835/files#diff-2d7a5f0f5688752b443886b6f4c3c78cL160. Let me try this again today

st2actions/st2actions/runners/python_action_wrapper.py
```diff
@@ -132,7 +133,8 @@ def run(self):
             action_output['status'] = action_status
             print_output = json.dumps(action_output)
         else:
-            raise Exception('Status should either be True or False.')
+            raise InvalidStatusException('Status returned from the action must'
```
Kami: Could also just use `ValueError` or similar, but that's OK.

st2actions/st2actions/runners/pythonrunner.py
```diff
@@ -165,6 +166,9 @@ def _get_output_values(self, exit_code, stdout, stderr, timed_out):
         else:
             error = None
 
+        if "InvalidStatusException" in stderr:
```
Kami: That's not 100% fail-safe and robust so we need to be a bit careful (regular stdout could also contain this string).

Kami: One option would be to serialize it as a special result (like we do with regular result - we wrap it in a special start and end markers which are very unlikely to occur in regular output).

Kami: Anyway, let's get merged what we have right now, but keep an eye on this going forward (e.g. if a user reports an issue with it or similar).

humblearner: Hmm, I thought only time we will get `InvalidStatusException` in stderr is : https://github.com/StackStorm/st2/pull/2835/files#diff-f1f34011cf21592061e7b15f398f4e3fR135. i.e when action_status in python_action_wrapper.py is anything but bool or None. That is why i went with this comparison and have user defined exception. 

I will definitely keep an eye on the issues related to this change. Thank you for reviewing this and being patient!! :)

Kami: No problem and nice work on the PR overall :+1:

st2actions/st2actions/runners/pythonrunner.py
lakshmi-kannan: .get('result', None) is always the preferred approach.

st2actions/st2actions/runners/python_action_wrapper.py
```diff
@@ -133,8 +133,12 @@ def run(self):
             action_output['status'] = action_status
             print_output = json.dumps(action_output)
         else:
-            raise InvalidStatusException('Status returned from the action must'
-                                         ' either be True or False.')
+            try:
```
Kami: try / except is a bit redundant, you could just do `sys.exit` :)

It would also be a good idea to define exit code in some constants module. E.g. `PYTHON_ACTION_INVALID_STATUS_EXIT_CODE = 220` so it can be re-used across the code base.

Besides that, LGTM :+1:

humblearner:
Tried without try/except it doesn't print the exception message:
May be i am missing something :)

```
In [5]: try:
   ...:     raise Exception("error message")
   ...: except:
   ...:     traceback.print_exc()
   ...:     sys.exit(220)
   ...:
   ...:
  File "<ipython-input-5-beb3060981eb>", line 2, in <module>
    raise Exception("error message")
Exception: error message
SystemExit: 220
```
```
In [6]: raise Exception("error message",traceback.print_exc(),sys.exit(220))
None
SystemExit: 220
```

Kami:
Ah, right you also want to include message in stderr. That's fine but it could also just be done in the parent process.

In any case, LGTM :)

Kami:
We've missed a scenario / edge case where this change is not a backward compatible - if an existing Python runner action returns a tuple with two items.

I documented this scenario in upgrade notes with information on how to resolve it - https://github.com/StackStorm/st2docs/pull/221.

I also updated code to throw a more user-friendly exception in case such scenario is encountered - ce1b72f10813f2e53e2076dd07aef9842a64ad3e

```python
status: failed
error: Status returned from the action run() method must either be True or False, got: chef
If this is an existing action which returns a tuple with two items, it needs to be updated to 
either:

1. Return a list instead of a tuple
2. Return a tuple where a first items is a status flag - (True, ('item1', 'item2'))
```

PR #2836, opened 2016-07-27
[WIP] [NSFW] StackStorm Exchange

emedvedev:
Before I start describing what this PR is for, let me draw attention to this wonderful comment, a model example of thrilling writing worthy of being inscribed in front of a dragon cave if only said caves existed in real life:

```
#####
# !!!!!!!!!!!!!!
# !!! README !!!
# !!!!!!!!!!!!!!
#
# This NEEDS a rewrite. Too many features and far too many assumption
# to keep this impl straight any longer. If you only want to read code do
# so at your own peril.
#
# If you are here to fix a bug or add a feature answer these questions -
# 1. Am I fixing a broken feature?
# 2. Is this the only module in which to fix the bug?
# 3. Am I sure this is a bug fix and not a feature?
#
# Only if you can emphatically answer 'YES' to allow about questions you should
# touch this file. Else, be warned you might loose a part of you soul or sanity.
#####
```

This PR is partially dedicated to rewriting and refactoring this file, and while I might need to get my soul checked afterwards, it still needs to be done.

Let's get to the point: since we're moving to the "one pack = one repo" model and StackStorm Exchange instead of st2contrib, I'm rewriting `packs.install` in a way that would install packs either from a default org (specified in the config), or from a git repo, no longer assuming that a repo might have multiple packs.

The only two parameters we need for `packs.install` in this PR are `name_or_url` and `ref`. A temporary regression is that only one pack at a time can be installed right now, but for multiple packs I'll probably need to convert the chain into a `with-items` workflow and then @enykeev will punch me in the face. Ouch.

Combined with @enykeev's new API endpoints, we can now set actions for creating and installing packs from the Exchange through CLI and—some time later—Web UI. There are still things to be worked out, but it works for the most part. It's just a start though.

Exchange is at https://github.com/StackStorm-Exchange right now. I'll add the team to owners once I'm done with testing, and at some point we'll have to start transferring packs.


emedvedev:
Reload is also something that has to be reworked in both `packs.install` and `packs.uninstall` as Kirill pointed out, but the scope of this particular PR is mostly confined to `packs.download` changes.

contrib/packs/actions/download.yaml
```diff
@@ -5,23 +5,15 @@
...
-    packs:
-      type: "array"
-      items:
-        type: "string"
-    repo_url:
+    name_or_url:
       type: "string"
-      default: "StackStorm/st2contrib"
-    branch:
+    ref:
```
enykeev: Can we call it `git_ref` or `gitref`, please? For me, it has a strong mental connection to `ref_or_id` from API layer and causes lots of confusion even after I figured what you meant by that.

contrib/packs/actions/download.yaml
```diff
@@ -5,23 +5,15 @@
   enabled: true
   entry_point: "pack_mgmt/download.py"
   parameters:
-    packs:
-      type: "array"
-      items:
-        type: "string"
-    repo_url:
+    name_or_url:
```
enykeev: `name_or_url` at the same time looks awfully verbose to me. How about `pack`?

contrib/packs/actions/pack_mgmt/download.py
```diff
@@ -66,50 +40,28 @@ def __init__(self, config=None, action_service=None):
         self._subtree = None
         self._repo_url = None
 
-    def run(self, packs, repo_url, abs_repo_base, verifyssl=True, branch='master', subtree=False):
+    def run(self, name_or_url, ref, abs_repo_base, verifyssl=True, branch='master'):
```
enykeev: `branch` is still in signature

contrib/packs/actions/pack_mgmt/download.py
```diff
@@ -66,50 +40,28 @@ def __init__(self, config=None, action_service=None):
         self._subtree = None
         self._repo_url = None
```
enykeev: both are not needed anymore

contrib/packs/actions/download.yaml
emedvedev: Yeah, `ref` is pretty much a working title. I like `version` the most, honestly, but to use this word we need to make people follow at least some basic convention with branching/tagging. :) Think `version` would be ok? I'm fine with `gitref`, too, but it's slightly more confusing for new users.

enykeev: `rev` as in revision - that's the word we're looking for. It can be particular version, hash, branch or tag.

contrib/packs/actions/pack_mgmt/download.py
```diff
@@ -120,47 +70,43 @@ def _clone_repo(repo_url, verifyssl=True, branch='master'):
         Repo.clone_from(repo_url, abs_local_path, branch=branch, depth=1)
         return abs_local_path
 
-    def _move_packs(self, abs_repo_base, packs, abs_local_path, subtree):
+    def _move_pack(self, abs_repo_base, pack_name, abs_local_path):
```
dzimine: The logic remains "clone to temp, move to `opt/stackstorm/packs`? If so, why? why not clone to the location? Again, the requirement is I go to `/opt/stackstorm/packs/mypack` and do `git status`there, and all the other git operations. It may still  work as we move `.git`, just want to better understand what is a good reason of intermediate copy?

emedvedev: We move `.git`, so you'll be able to—I double-checked it. :)

One simple reason for cloning to temp first — if the pack is already there, we don't want to mess up the config (see L86 and below).

contrib/packs/actions/download.yaml
```diff
@@ -5,23 +5,16 @@
   enabled: true
   entry_point: "pack_mgmt/download.py"
   parameters:
-    packs:
-      type: "array"
-      items:
-        type: "string"
-    repo_url:
+    pack:
```
dzimine: Do we assume StackStorm exchange as the only source of packs? How do run pack install from my local git? Getting it from some random git server is a major use case; and unlike github, we don't even have a concept of org in this case. User will install packs from a mixture of Github/st2exchange, Github/foo, Github/bar, and private git server.

emedvedev: https://github.com/StackStorm/st2/blob/8e32faf5cfbb7ea747f088ef2d82c980e7d978ec/contrib/packs/actions/install.meta.yaml#L13

>       description: "Name of the pack in Exchange or a git repo URL"

dzimine: You absolutely need a description to `pack`. I looked down to code and looks like it may be quite a few things :)

emedvedev: And this: https://github.com/StackStorm/st2/blob/8e32faf5cfbb7ea747f088ef2d82c980e7d978ec/contrib/packs/actions/pack_mgmt/download.py#L170

So we don't, of course. If it's just a name of the pack, like `sensu`, it'll be downloaded from exchange. If it's a git url, and not a name, it'll be downloaded from git.

emedvedev: We have a description in `install.meta.yaml`, you're not really supposed to launch `download` on its own. :) Still, fair point.

contrib/packs/actions/workflows/install.yaml
```diff
@@ -4,10 +4,8 @@
       name: "download"
       ref: "packs.download"
       parameters:
-        repo_url: "{{repo_url}}"
-        branch: "{{branch}}"
-        packs: "{{packs}}"
-        subtree: "{{subtree}}"
+        pack: "{{pack}}"
```
dzimine: Install will need more modifications.

dzimine:
I only looked briefly so if I missed a point or two, apologies. 

In addition to inline notes, more on the spec. As I mentioned in https://github.com/StackStorm/discussions/issues/188
>
Most packs require configuration, how to balance the "ease" of pack installation with the need to configure? One idea: 
		* `st2 pack get` - git clone pack, setup virtual env
		* `st2 pack configure` - CLI offers to input config params interactively and writes to stackstorm/config/<pack>.yaml, and also tells user "you can as well do it manually, link to docs".
		* `st2 pack install` is "`get`, `configure` (here it asks user for parameters), `load`" 
>

Just think through getting a pack that requires installation (you may write docs first :) )

dzimine:
`StackStorm-Exchange` - do we really need a `-`? Opinions?

emedvedev:
> StackStorm-Exchange - do we really need a -? Opinions?

https://github.com/sensu-plugins
https://github.com/hubot-scripts

Seems to be the norm :) I was considering having no dash, and also something like St2-Exchange, but didn't like either.

emedvedev:
> * st2 pack configure - CLI offers to input config params interactively and writes to stackstorm/config/.yaml, and also tells user "you can as well do it manually, link to docs".

This is a great idea, especially with the new config schema changes. I think we shouldn't really have a separate command for `get` though: if the user wants to go the manual way, they're on their own. A flag for non-interactive installation would be better (i.e. leave the config empty and—optionally—not register a pack if something's required but not filled in).

To sum it up, definitely something we need to consider, but I'm not sure it's a priority for step 1. If we start nitpicking at this point, the scope of the first release will become monstrous and get out of control. I'd rather keep design discussions in https://github.com/StackStorm/discussions/issues/188 unless it's something that's already in the PR.

enykeev:
> Seems to be the norm :) I was considering having no dash, and also something like St2-Exchange, but didn't like either.

`st2exchange`. Why complicate things?

contrib/packs/actions/pack_mgmt/download.py
enykeev: > The logic remains "clone to temp, move to opt/stackstorm/packs? If so, why? why not clone to the location?

From what I can tell, we were cloning to temp location to be able to work with subtree. Since we don't have a subtree option now, we should indeed clobne to packs folder directly, but this would mean you can't install the same pack twice (think of git error 'directory is not empty' on clone). Which may not actually be a bad thing considering there is no obvious way to go about it.

emedvedev:
> st2exchange. Why complicate things?

Because our main org (not repo) is `StackStorm`, not `st2` :)

contrib/packs/actions/pack_mgmt/download.py
emedvedev: We can either check for the dir and abort the install if it's not empty, as @enykeev says (a sensible approach), or update the pack and keep the config (that's what we do now, and that's why we have temp clone). I'd go for the latter, at least for now—makes things easier for user, and even a bit easier for us.

enykeev:
Closing in favor of https://github.com/StackStorm/st2/pull/2842

Issue #2837, opened 2016-07-27
Request for sort ordering execution list results


bri365:
Allow users to specify sort order when requesting a list of executions.

When troubleshooting slow/overloaded systems, getting the oldest action/workflow in a given state is very helpful, especially as the API will only return a limited number of executions, so finding the oldest running action can required going directly to mongo.

Perhaps '-r' or '--reverse' for oldest to newest?

Kami:
I believe API already supports that. In case it doesn't, it should be easy to add support for it (just need to update a class variable on the API controller class).

Same goes for CLI - should be quite easy to add support for a new flag. Do you want to take a stab? :)

In any case, I'm fine with "-r" flag (if it's not already used for something else). Otherwise perhaps `--sort desc/asc`.

bri365:
I'll drop it in my queue :-)

PR #2839, opened 2016-07-27
Enhancement: add option to sort in descending order - ascending is default

bri365:
Add execution list sort ordering as requested in [issue 2837](https://github.com/StackStorm/st2/issues/2837) 

Added command line argument -s | --sort asc|desc (ascending is the current default sort order)

Added check for 'desc' in api (to mesh with overrides for -tg or -tl)

### DIFF
--- a/st2api/st2api/controllers/v1/actionexecutions.py
+++ b/st2api/st2api/controllers/v1/actionexecutions.py
@@ -376,7 +376,7 @@ class ActionExecutionsController(ActionExecutionsControllerMixin, ResourceContro
         if 'timestamp_lt' in kw:
             query_options = {'sort': ['-start_timestamp', 'action.ref']}
             kw['query_options'] = query_options
-        elif 'timestamp_gt' in kw:
+        elif 'timestamp_gt' in kw or 'sort_desc' in kw:
             query_options = {'sort': ['+start_timestamp', 'action.ref']}
             kw['query_options'] = query_options
 
--- a/st2api/st2api/controllers/v1/actionexecutions.py
+++ b/st2api/st2api/controllers/v1/actionexecutions.py
@@ -373,10 +373,10 @@ class ActionExecutionsController(ActionExecutionsControllerMixin, ResourceContro
 
         # Use a custom sort order when filtering on a timestamp so we return a correct result as
         # expected by the user
-        if 'timestamp_lt' in kw:
+        if 'timestamp_lt' in kw or 'sort_desc' in kw:
             query_options = {'sort': ['-start_timestamp', 'action.ref']}
             kw['query_options'] = query_options
-        elif 'timestamp_gt' in kw or 'sort_desc' in kw:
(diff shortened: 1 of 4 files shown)


bri365:
@Kami I'm a little confused by test_default_sort in st2api/tests/unit/controllers/v1/test_executions_filters.py:
it appears to be verifying the last json item has the lowest timestamp, which is not the observed behavior from st2 commands (with -j to see raw json)

Kami:
Great, thanks for this PR.

As far as tests go and CLI ordering goes - I believe that's a CLI only ordering. If you hit API directly (`curl http://127.0.0.1:9101/v1/executions`) you should get back the response in the default sort order.

Kami:
So far the PR looks good, it just needs some tests :)

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -376,7 +376,7 @@ def get_all(self, exclude_attributes=None, **kw):
         if 'timestamp_lt' in kw:
```
Kami: For explicitness sake, I would also add `or 'sort_desc' in kw:` here so it's consistent with the elif clause below.

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -376,7 +376,7 @@ def get_all(self, exclude_attributes=None, **kw):
...
             query_options = {'sort': ['-start_timestamp', 'action.ref']}
             kw['query_options'] = query_options
-        elif 'timestamp_gt' in kw:
+        elif 'timestamp_gt' in kw or 'sort_desc' in kw:
```
Kami: Actually, shouldn't this be `sort_asc`, since `+start_timestamp` implies sorting in ascending order (smallest aka oldest executions first)?

I know it's probably a bit confusing right now since CLI options and filters don't always map directly to the API ones and we do some additional formatting and sorting in the CLI...

bri365: yes, I think I can wrap my head around that.  I like explicitly including them as well.

st2client/st2client/commands/action.py
```diff
@@ -975,6 +980,11 @@ def run(self, args, **kwargs):
             kwargs['timestamp_gt'] = args.timestamp_gt
         if args.timestamp_lt:
             kwargs['timestamp_lt'] = args.timestamp_lt
+        if args.sort_order:
+            if args.sort_order in ['asc', 'ascending']:
```
lakshmi-kannan: I have a feeling this asc and desc might get confusing for user. So maybe add help strings - "earliest first" and "latest first" will make it less so.

Kami: What @lakshmi-kannan said (CLI only).

Besides that, LGTM :+1:

bri365: Great point @lakshmi-kannan - I'll get it updated today or tomorrow

Kami:
Merged, thanks!

PR #2840, opened 2016-07-28
Validate config values against the config schema

Kami:
This pull request adds `--register-validate-configs` flag to the `st2-register-content` script.

If this flag is available, new-style config values are validated against the config schema if config schema is available. As mentioned in the past, this works along the whole gradual / optional typing story so the validation is not enabled by default right now, but a user can opt-in using a flag mentioned above.

It's also to important to note that only "static" config values are validated during the registration phase. Dynamic ones are resolved during run-time so those can only be validated during run-time (when a sensor is instantiated for sensors and when action is ran for actions). That's not something I will do in this PR, but I will introduce support for that at some point in the future (baby steps).

Usage:

```bash
st2-register-content --config-file conf/st2.dev.conf --register-configs --register-validate-configs
```

Example output on failure:

```bash
2016-07-29 14:20:50,296 INFO [-] Connecting to database "st2" @ "0.0.0.0:27017" as user "None".
2016-07-29 14:20:50,410 INFO [-] =========================================================
2016-07-29 14:20:50,411 INFO [-] ############## Registering configs ######################
2016-07-29 14:20:50,411 INFO [-] =========================================================
2016-07-29 14:20:50,975 WARNING [-] Failed to register configs: Failed validating config for pack "aws": 1 is not of type u'string'

Failed validating u'type' in schema['properties'][u'region']:
    {u'default': u'us-east-1',
     u'description': u'AWS region to use.',
     u'required': True,
     u'type': u'string'}

On instance[u'region']:
    1
  File "./st2common/bin/st2-register-content", line 22, in <module>
    sys.exit(content_loader.main(sys.argv[1:]))
  File "/data/stanley/st2common/st2common/content/bootstrap.py", line 359, in main
    register_content()
  File "/data/stanley/st2common/st2common/content/bootstrap.py", line 343, in register_content
    register_configs()
  File "/data/stanley/st2common/st2common/content/bootstrap.py", line 307, in register_configs
    raise e
jsonschema.exceptions.ValidationError: Failed validating config for pack "aws": 1 is not of type u'string'

Failed validating u'type' in schema['properties'][u'region']:
    {u'default': u'us-east-1',
     u'description': u'AWS region to use.',
     u'required': True,
     u'type': u'string'}

On instance[u'region']:
    1
```

## TODO

- [x] Tests
- [x] Documentation updates - https://github.com/StackStorm/st2/pull/2840

### DIFF
--- a/st2common/st2common/content/bootstrap.py
+++ b/st2common/st2common/content/bootstrap.py
@@ -56,10 +56,6 @@ def register_opts():
         cfg.BoolOpt('policies', default=False, help='Register policies.'),
         cfg.BoolOpt('configs', default=False, help='Register and load pack configs.'),
 
-        # Config related options
-        cfg.BoolOpt('validate-configs', default=False,
-                    help=('Validate config values against config schema if schema is available.')),
-
         cfg.StrOpt('pack', default=None, help='Directory to the pack to register content from.'),
         cfg.BoolOpt('setup-virtualenvs', default=False, help=('Setup Python virtual environments '
                                                               'all the Python runner actions.')),
@@ -288,7 +284,6 @@ def register_policies():
 def register_configs():
     pack_dir = cfg.CONF.register.pack
     fail_on_failure = not cfg.CONF.register.no_fail_on_failure
-    validate_configs = cfg.CONF.register.validate_configs
 
     registered_count = 0
 
@@ -298,7 +293,7 @@ def register_configs():
         LOG.info('=========================================================')
         registered_count = configs_registrar.register_configs(pack_dir=pack_dir,
                                                               fail_on_failure=fail_on_failure,
-                                                              validate_configs=validate_configs)
+                                                              validate_configs=True)
     except Exception as e:
         exc_info = not fail_on_failure
         LOG.warning('Failed to register configs: %s', e, exc_info=exc_info)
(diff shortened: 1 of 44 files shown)


st2common/st2common/content/bootstrap.py
```diff
@@ -56,10 +56,15 @@ def register_opts():
         cfg.BoolOpt('policies', default=False, help='Register policies.'),
         cfg.BoolOpt('configs', default=False, help='Register and load pack configs.'),
 
+        # Config related options
+        cfg.BoolOpt('validate-configs', default=False,
```
Kami: I'm also open to a better flag name - suggestions welcome.

Kami:
Docs at https://github.com/StackStorm/st2/pull/2840

st2common/st2common/content/bootstrap.py
lakshmi-kannan: I've no idea why we want a flag for this? If there is a schema, we should by default validate and throw an error. We now have flags for skip failure or abort (default). I think those should be enough.

lakshmi-kannan:
I really like the validation change but i think the flag to turn on the validation is excessive IMO. We should try and reduce the number of flags and not add more :P.

Kami:
I'm fine with making validation mandatory if config schema exists.

Kami:
I will make the changes tomorrow - turn it on by default and remove the flag.

PR #2841, opened 2016-07-29
Making jsexpose more flexible

enykeev:
Allows creating controllers with optional body and reuse the same controller for url's of any depth.

Controller with this signature
```
@jsexpose(body_cls=..., arg_types=[str, int, str])
def post(self, body, type, id, property):
```
will successfully handle all this rest calls
```
POST /document/112/header/ -> post(body=({...} or None), type='document', id=112, property='header')
POST /document/112/ -> post(body=({...} or None), type='document', id=112, property=None)
POST /document/ -> post(body=({...} or None), type='document', id=None, property=None)
```

### DIFF
--- a/st2api/st2api/controllers/v1/actions.py
+++ b/st2api/st2api/controllers/v1/actions.py
@@ -144,7 +144,7 @@ class ActionsController(resource.ContentPackResourceController):
 
     @request_user_has_resource_db_permission(permission_type=PermissionType.ACTION_MODIFY)
     @jsexpose(arg_types=[str], body_cls=ActionCreateAPI)
-    def put(self, action_ref_or_id, action):
+    def put(self, action, action_ref_or_id):
         action_db = self._get_by_ref_or_id(ref_or_id=action_ref_or_id)
 
         # Assert permissions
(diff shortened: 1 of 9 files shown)


enykeev:
see https://github.com/StackStorm/st2/commit/704f40bed690b404b9991143b994c05ec1f414c7 for more discussion on the topic

Kami:
LGTM.

Can we also add some specific cases for the scenario mentioned in the description?

Kami:
We should also document this (second paragraph in your comment) -https://github.com/StackStorm/st2/commit/704f40bed690b404b9991143b994c05ec1f414c7#commitcomment-18444450 somewhere in the developer section of our documentation.

enykeev:
> Can we also add some specific cases for the scenario mentioned in the description?

We will, with pack API.

Kami:
Looks like we missed a spot - https://github.com/StackStorm/st2/issues/2878

In any case, I'm on it. Will also add a test case to make sure it doesn't happen again in the future (although, I was like 90% we already have a test case for that scenario :/).

PR #2842, opened 2016-08-01
Pack management

enykeev:
ezgif com-resize


contrib/packs/actions/pack_mgmt/download.py
```diff
@@ -26,141 +25,94 @@
...
-# If you are here to fix a bug or add a feature answer these questions -
-# 1. Am I fixing a broken feature?
-# 2. Is this the only module in which to fix the bug?
-# 3. Am I sure this is a bug fix and not a feature?
-#
-# Only if you can emphatically answer 'YES' to allow about questions you should
-# touch this file. Else, be warned you might loose a part of you soul or sanity.
-#####
-
+PACK_VERSION_SEPARATOR = '#'
```
emedvedev: Any reason not to go with `=`? :) An existing convention, personal preference, etc?

enykeev: No one use `=` for this purpose as it seems. NPM uses `#` and PIP sticks to `@`. I like `#` more as it correlates with anchor part of the URL and also doesn't repeat throughout the rest of the url so you can just split by character.

emedvedev: Alright, makes sense. :)

lakshmi-kannan: Yeah, `#` makes sense to me.

contrib/packs/actions/pack_mgmt/download.py
```diff
@@ -16,7 +16,6 @@
 import os
```
lakshmi-kannan: Please make sure there is good test coverage for all the changes. I don't see much tests for new code added but I haven't looked at it in detail. This is a dangerous file to make changes because of absence of tests. I'd say this is a PR blocker.

conf/st2.conf.sample
```diff
@@ -73,8 +73,12 @@ backend = flat_file
 home_dir = /opt/cslang
 
 [content]
+# A URL pointing to the pack index. StackStorm Exchange is used by default. Use a comma-separated list for multiple indexes if you want to get other packs discovered with "st2 pack search".
+index_url = https://index.stackstorm.org/v1/index.json # comma separated list allowed here.
```
Kami: How does the ordering work if the same pack is available on multiple indexes? Does it work from left to right aka first one has priority?

It would be good to document this behavior somewhere in the docs.

emedvedev: See https://github.com/StackStorm/st2/pull/2842/files#diff-3a3bf900cae2541d75a791bc40814158R66

It works from right to left (the last one has priority), but could be changed to left-to-right. Pack names are assumed to be unique, and if you have overlapping pack names in different indexes, they'll be overwritten. I.e. if I maintain my own fork of the `sensu` pack and keep the name, my fork will be pulled when I do `st2 install sensu` if my index comes last in the list.

We should definitely document this, because it's also useful when maintaining local mirrors, etc.

contrib/packs/actions/pack_mgmt/download.py
enykeev: Done https://github.com/StackStorm/st2/pull/2842/commits/ee861a2fdd9bdfead3e76d362c3586b302b1915e

st2common/st2common/models/api/base.py
```diff
@@ -245,16 +243,6 @@ def cast_func(value):
...
-
-                    # Set default pack if one is not provided for resource create
-                    if function_name == 'post' and not hasattr(obj, 'pack'):
-                        extra = {
-                            'resource_api': obj,
-                            'default_pack_name': DEFAULT_PACK_NAME
-                        }
-                        LOG.debug('Pack not provided in the body, setting a default pack name',
-                                  extra=extra)
-                        setattr(obj, 'pack', DEFAULT_PACK_NAME)
```
enykeev: I moved this logic to https://github.com/StackStorm/st2/pull/2842/files#diff-530450a16546013ae50b71cd07dcb2b4R111 to avoid it leaking into every post request. I'm not entirely certain we don't rely on it anywhere else. I'm not sure we should have ever done anything like that in the first place. I'm absolutely positive that this is one of the worst places it could have been done.

@kami thoughts? https://github.com/StackStorm/st2/commit/39e5ac529b10cbc25bd1c4ddf58fac312bf33f20

st2api/st2api/controllers/v1/packs.py
```diff
@@ -32,6 +55,128 @@
...
+            return packs_service.search_pack_index(pack_search_request.query)
+        else:
+            return packs_service.get_pack_from_index(pack_search_request.pack)
+
+
+class IndexHealthController(RestController):
+
+    @jsexpose()
+    def get(self):
+        return packs_service.check_index_health()
```
enykeev: @emedvedev my personal take on that one is that if you absolutely positive you would need this functions to be services so you can reuse them somewhere else, you need to put an additional effort into separating parts that worth reusing from parts that are only make sense in API controller. For example, `search_pack_index` and `get_pack_from_index` probably should not return PackAPI instance, but should be converted to one in the API controller.

For one, I just feel uneasy with controllers as one liners. I have to go to packs_service to figure out what I'm going to get as a response. In an st2api v2 we're going to document response for every endpoint so it would be better for anyone if you would have an IndexHealthAPI model with the scheme describing the response from day one. Saying that, I too have this problem in PackRegisterController and should fix it before we merge.

enykeev:
I'd say that we need to create a parity between st2contrib and stackstorm-exchange before we merge the PR (so it would be easier for us to test it). We may have to recreate stackstorm-exchange another time for the official release to get all the latest changes.

conf/st2.conf.sample
Kami: Yep. And I think left-to-right is probably better for consistency since we already do this in other places - where you specify multiple MongoDB and RabbitMQ URLs, etc.

contrib/packs/actions/pack_mgmt/download.py
```diff
@@ -16,152 +16,102 @@
...
 GITINFO_FILE = '.gitinfo'
 PACK_RESERVE_CHARACTER = '.'
-
-STACKSTORM_CONTRIB_REPOS = [
-    'st2contrib',
-    'st2incubator'
-]
-
-#####
-# !!!!!!!!!!!!!!
```
Kami: Yay, go away and don't come back.

contrib/packs/actions/pack_mgmt/register.py
```diff
@@ -0,0 +1,127 @@
...
+
+        client_kwargs = {}
+        if cacert:
+            client_kwargs['cacert'] = cacert
+
+        return self._client(base_url=base_url, api_url=api_url,
+                            auth_url=auth_url, token=token,
+                            **client_kwargs)
+
+    def _get_st2_urls(self):
```
Kami: This looks familiar, probably from st2 pack? :P

contrib/packs/actions/render_template.py
```diff
@@ -0,0 +1,37 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+from st2actions.runners.pythonrunner import Action
+from st2common.util import jinja as jinja_utils
+
+
+class RenderTemplateAction(Action):
```
Kami: Now that we support pack unit tests, unit test for this would be nice.

contrib/packs/actions/templates/pack.yaml.j2
```diff
@@ -0,0 +1,13 @@
+---
+name: {{ name }}
+description: {{ description }}
+{% if keywords %}
+keywords:
+{% for keyword in keywords %}
```
Kami: Do we need to use `-` at the end (-%}) to make sure there is no extra blank line where the control statement is?

st2api/st2api/controllers/v1/packs.py
```diff
@@ -32,6 +55,128 @@
...
+
+class PackInitController(ActionExecutionsControllerMixin, RestController):
+
+    @jsexpose(body_cls=PackInitRequestAPI, status_code=http_client.ACCEPTED)
+    def post(self, args):
+        parameters = vars(args)
+
+        new_liveaction_api = LiveActionCreateAPI(action='packs.create',
+                                                 parameters=parameters,
+                                                 user=None)
```
Kami: Can we please correctly carry the user context over if possible?

This way RBAC, {{user.}}, will work as expected (I know it's not used right now in that action, but still).

st2api/st2api/controllers/v1/packs.py
```diff
@@ -32,6 +55,128 @@
...
+                'packs': pack_uninstall_request.packs
+            }
+
+        new_liveaction_api = LiveActionCreateAPI(action='packs.uninstall',
+                                                 parameters=parameters,
+                                                 user=None)
+
+        execution = self._handle_schedule_execution(liveaction_api=new_liveaction_api)
+
+        return PackAsyncAPI(execution_id=execution.id)
```
Kami: As mentioned in the other PR, to be able to keep the same good user-experience we will either need to make API sync or use event-stream / poll on the client side...

st2api/st2api/controllers/v1/packs.py
```diff
@@ -32,6 +55,128 @@
...
+
+
+class PackRegisterController(RestController):
+
+    @jsexpose(body_cls=PackRegisterRequestAPI)
+    def post(self, pack_register_request):
+        if pack_register_request and hasattr(pack_register_request, 'types'):
+            types = pack_register_request.types
+        else:
+            types = ['runner', 'action', 'trigger', 'sensor', 'rule', 'rule_type', 'alias',
```
Kami: Should we use some constant for that or similar?

st2client/st2client/commands/pack.py
```diff
@@ -0,0 +1,247 @@
...
+        if not instance:
+            raise Exception('Server did not create instance.')
+
+        parent_id = instance.execution_id
+
+        stream_mgr = self.app.client.managers['Stream']
+
+        execution = None
+
+        with term.TaskIndicator() as indicator:
```
Kami: Ah, cool I see how you solved the async API in the client.

This works for me :+1:

st2client/st2client/models/core.py
```diff
@@ -385,3 +386,95 @@ def re_emit(self, trigger_instance_id, **kwargs):
...
+
+
+class StreamManager(object):
+    def __init__(self, endpoint, cacert, debug):
+        self._url = httpclient.get_url_without_trailing_slash(endpoint) + '/stream'
+        self.debug = debug
+        self.cacert = cacert
+
+    @add_auth_token_to_kwargs_from_env
+    def listen(self, events, **kwargs):
```
Kami: Can we explicitly declare arguments here? :P

It seems like we should only add token, api_key and then we can get rid of kwargs...

st2common/st2common/models/api/base.py
Kami: Yeah, I never liked it, it was hacky as hell.

I think we rely on it for actions and rules.

st2client/st2client/models/core.py
enykeev: We probably can, but the next time we decide to add some other kind of token, I'd rather be on the safe side. We're passing kwargs throughout the whole st2client. Why this place should be different?

st2api/st2api/controllers/v1/packs.py
enykeev: As I said previously, I'd rather we have all our content types in form of a tree with explicit dependencies (like runner before rule and rule_type before rule).

st2api/st2api/controllers/v1/packs.py
enykeev: Here he realizes https://github.com/StackStorm/st2/pull/2842#discussion_r84045133 :P

st2client/st2client/models/core.py
enykeev: I disagree. Bad consistency seem better to me than no consistency.

st2api/st2api/controllers/v1/packs.py
enykeev: The idea for this health check is to allow st2 to seamlessly switch to hot backup if something would happen with github or the link to it. It surely seems like an overkill, but since we can get it cheap enough, I don't see a reason not to.

enykeev: have I? Well, that's unfortunate :P

emedvedev: Anyway, I'll fix.

emedvedev:
> I'd say that we need to create a parity between st2contrib and stackstorm-exchange before we merge the PR (so it would be easier for us to test it). We may have to recreate stackstorm-exchange another time for the official release to get all the latest changes.

@enykeev you mean recreate all packs? I don't have tagging implemented yet, but once I get it done, I'll run the script.

Kami:
For a more consistent and explicit API, I propose changing:

/v1/packs/search -> /v1/packs/index/search
/v1/packs/health -> /v1/packs/index/health

What do others think?

enykeev:
Personally, I don't see the need in it, but when I try to find a problem with it I fell like I'm starting nitpicking. If you sure you feel strongly about it, go ahead, but I surely hoping you have a plan in mind 😛

Kami:
I think it's a good idea because the API operation searches the registry (it doesn't search locally installed pack, we have list for that) and it queries health of the indexes / registries specified in the config.

Not having health somewhere in the URL makes it confusing, imo (it makes it seem like operation is performed on the packs, but as mentioned above, it's not).

Kami:
Another small naming thing - do we all like the idea of calling it "index" everywhere"? Or do we think something else would be better (e.g. registry)?

emedvedev:
@Kami index is fine, I think. Registry might have some weird connotations, since we also have pack registration, --register-all, etc. It's not a bad choice, just might get confusing in the context. Index is clear and has no other meaning

```
/v1/packs/search -> /v1/packs/index/search
/v1/packs/health -> /v1/packs/index/health
```
Gfi.

enykeev:
ok, seems green enough to me 😛 Merging!

PR #2847, opened 2016-08-03
Add mistral integration tests for various st2kv scenarios

m4dcoder:


contrib/examples/actions/workflows/mistral-yaql-st2kv-user-scope.yaml
```diff
@@ -0,0 +1,12 @@
+version: '2.0'
+
+examples.mistral-yaql-st2kv-user-scope:
+    vars:
+        polo: unspecified
+    tasks:
+        task1:
+            action: std.noop
+            publish:
+                polo: <% st2kv('marco') %>
```
lakshmi-kannan: This is slightly confusing for me. I think we should explicitly use st2kv.get(). There can be cases when we want to publish a variable to kv store. @m4dcoder tells me that YAQL really doesn't like `.` because it has a special meaning in YAQL world. I don't quite get that. I think st2kv_get is another option or if we can do st2kv['marco'], that is another option.

lakshmi-kannan: @dzimine @Kami: Your thoughts?

m4dcoder:
YAQL evaluate any dot as a dict.  So `st2kv.get()` won't work.  st2kv[key] will assume st2kv is a list because of the bracket.  Otherwise we will have to hack YAQL to achieve this result.  Also, YAQL is read/eval only.  I don't recommend doing set in YAQL even if it's possible.  That will be very inconsistent as anyplace where YAQL is evaluated in mistral will now be able to write out to st2 datastore.  I prefer not to support this. User should explicitly have an action to set the key.  Given that YAQL is read/eval only, it's not necessary to explicitly say st2kv_get or st2kv[key].  I think st2kv() as a function is fine the way it is.

contrib/examples/actions/workflows/mistral-yaql-st2kv-user-scope.yaml
m4dcoder: YAQL evaluate any dot as a dict. So st2kv.get() won't work. st2kv[key] will assume st2kv is a list because of the bracket. Otherwise we will have to hack YAQL to achieve this result. Also, YAQL is read/eval only. I don't recommend doing set in YAQL even if it's possible. That will be very inconsistent as anyplace where YAQL is evaluated in mistral will now be able to write out to st2 datastore. I prefer not to support this. User should explicitly have an action to set the key. Given that YAQL is read/eval only, it's not necessary to explicitly say st2kv_get or st2kv[key]. I think st2kv() as a function is fine the way it is.

m4dcoder: So, I tested this st2kv.get quickly.  To be exact, because of the `()`. YAQL evaluated st2kv as an object and thinks there's a get method for st2kv object. The exact error is "No method get for receiver st2kv matches supplied arguments". Since only pluggable YAQL function is supported, I don't see an option where I can register st2kv as a module in YAQL.

PR #2853, opened 2016-08-06
Use and run unit and integration tests with MongoDB 2.4, 2.6 and 3.2

Kami:
This pull request updates Travis file so we now run unit and integration tests with MongoDB 2.4, 2.6 and 3.2.

In the future we should drop official support for versions < 3.2 since they are quite old now.

In addition to the MongoDB change, I also updated the script to create a ramdisk and use this for MongoDB data dir. In theory, this should speed things up a bit, but probably not by much since we are running on a shared environment.

Note: If it turns out that the whole build now takes too long because we run 4 additional jobs, we can update the build configuration and only run tests with MongoDB 3.2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ before_script:
   - wget http://fastdl.mongodb.org/linux/mongodb-linux-x86_64-${MONGODB}.tgz -O /tmp/mongodb.tgz
   - tar -xvf /tmp/mongodb.tgz
   - mkdir /tmp/data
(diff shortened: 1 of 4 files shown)


Kami:
So yeah, sadly running takes many jobs does slow down the whole build quite a bit so for now I will only make sure we run tests with MongoDB 3.2.

The code itself is ready so it's easy to get it up and running with multiple versions in case we upgrade to get more parallel builds or similar.

PR #2854, opened 2016-08-08
Make fixate-requirements.py PIP 8.1.2 Compatible

andrew-regan:
Using PIP 8.1.2 with _fixate-requirements.py_ currently results in the following due to internal changes in PIP:

```
  File "scripts/fixate-requirements.py", line 174, in <module>
    skip=skip)
  File "scripts/fixate-requirements.py", line 117, in write_requirements
    requirements = merge_source_requirements(sources)
  File "scripts/fixate-requirements.py", line 96, in merge_source_requirements
    if req.req.project_name in projects:
AttributeError: 'Requirement' object has no attribute 'project_name'
```

See: https://github.com/pypa/pip/issues/3687

This change should be backward compatible with earlier versions of pip. Tested with 7.1.2 and 8.1.2.

### DIFF
--- a/scripts/fixate-requirements.py
+++ b/scripts/fixate-requirements.py
@@ -93,9 +93,9 @@ def merge_source_requirements(sources):
             # Requirements starting with project name "project ..."
             if req.req:
                 # Skip already added project name
-                if req.req.project_name in projects:
+                if req.name in projects:
                     continue
-                projects.add(req.req.project_name)
+                projects.add(req.name)
                 merged_requirements.append(req)
 
             # Requirements lines like "vcs+proto://url"
@@ -120,7 +120,7 @@ def write_requirements(sources=None, fixed_requirements=None, output_file=None,
     # Make sure there are no duplicate / conflicting definitions
     fixedreq_hash = {}
     for req in fixed:
-        project_name = req.req.project_name
+        project_name = req.name
 
         if not req.req:
             continue
@@ -133,7 +133,7 @@ def write_requirements(sources=None, fixed_requirements=None, output_file=None,
     lines_to_write = []
     links = set()
     for req in requirements:
-        if req.req.project_name in skip:
+        if req.name in skip:
             continue
 
         # we don't have any idea how to process links, so just add them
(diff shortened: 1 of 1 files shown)


lakshmi-kannan:
https://github.com/StackStorm/st2/blob/master/Makefile#L204
https://github.com/StackStorm/st2/blob/master/fixed-requirements.txt#L30
makes me think twice. We should test with latest versions of pip for sure. You have my skeptic +1. Please wait for @Kami to confirm.

andrew-regan:
Yeah....we hit this explicitly trying to pull in 8.1.2 w/o using that Makefile.

andrew-regan:
Are there other known issues when using pip>8.0.0 w/ StackStorm?

Kami:
Thanks for the PR.

I'm fine with this change, but can we please make the code more fail-safe so it also works with older versions?

Besides that, LGTM.

Kami:
@andrew-regan 

> Are there other known issues when using pip>8.0.0 w/ StackStorm?

Not that I'm aware off and fixate-requirements is a build-time only dependency anyway :)

andrew-regan:
@Kami How far back do you want to go? I tested with 7.1.2 and 8.1.2 which looks to be as far back as StackStorm currently supports according to the Makefile?

`$(VIRTUALENV_DIR)/bin/pip install --upgrade "pip>=7.1.2,<8.0.0"`

andrew-regan:
`InstallRequirement.name` looks to exist in PIP 6.0 as well.

Kami:
@andrew-regan pip>=7.1.2 is fine.

So value of req.req.project_name and req.name is always the same?

Kami:
I just checked it locally and it looks like the values are indeed the same.

I will go ahead and merge it. Thanks!

andrew-regan:
If 7.1.2 is used it will behave exactly the same because `name` returns `req.project_name`:
```
    @property
    def name(self):
        if self.req is None:
            return None
        return native_str(self.req.project_name)
```

8.1.2 returns the exact same information, but it is stored internally in a different location:
```
    @property
    def name(self):
        if self.req is None:
            return None
        return native_str(pkg_resources.safe_name(self.req.name))
```

PR #2856, opened 2016-08-08
Custom jinja functions

lakshmi-kannan:
## What? 

Include a ``to_human_time`` jinja function that can be used to fuzzy render time value in seconds. 
For example {{ elapsed_seconds || to_human_time }} will convert an integer (or long) 181 to string '3m1s'.

## Why?
See https://github.com/StackStorm/hubot-stackstorm/issues/123

## TODO

- [X] Refactor existing jinja filters to own files
- [X] Add a to_human_time method (** I think I'll need to_human_time_from_secs to be explicit)
- [X] Fix a bug in jinja rendering where jinja filter expressions weren't detected to be jinja renderable (See https://gist.github.com/lakshmi-kannan/0af3004652c12eaa9d5002cb0903a058, eaa7420, refactored 8c95a6d)
- [x] Unit tests
- [x] Documentation for jinja filters (https://github.com/StackStorm/st2docs/pull/228)
- [x] Changelog

### DIFF
--- a/contrib/chatops/actions/format_execution_result.py
+++ b/contrib/chatops/actions/format_execution_result.py
@@ -12,7 +12,7 @@ class FormatResultAction(Action):
         api_url = os.environ.get('ST2_ACTION_API_URL', None)
         token = os.environ.get('ST2_ACTION_AUTH_TOKEN', None)
         self.client = Client(api_url=api_url, token=token)
-        self.jinja = jinja_utils.get_jinja_environment()
+        self.jinja = jinja_utils.get_jinja_environment(allow_undefined=True)
         self.jinja.tests['in'] = lambda item, list: item in list
 
         path = os.path.dirname(os.path.realpath(__file__))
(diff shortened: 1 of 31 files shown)


contrib/chatops/actions/format_execution_result.py
```diff
@@ -12,7 +12,7 @@ def __init__(self, config=None, action_service=None):
         api_url = os.environ.get('ST2_ACTION_API_URL', None)
         token = os.environ.get('ST2_ACTION_AUTH_TOKEN', None)
         self.client = Client(api_url=api_url, token=token)
-        self.jinja = jinja_utils.get_jinja_environment()
+        self.jinja = jinja_utils.get_jinja_environment(allow_undefined=True)
```
emedvedev: As mentioned in Slack earlier, I have some reservations about this, but it should be safe and might even lead to better UX. Shouldn't be a problem.

PR #2860, opened 2016-08-10
Allow user to cancel multiple executions using st2 execution cancel command

Kami:
This change allows user to cancel multiple executions using a single invocation of ``st2 execution cancel`` command.

It uses the same notation as many other well known CLI commands (e.g. `git cherry-pick sha1 sha2 sha2`, etc.).

For example:

```bash
st2 execution cancel 57a9707b0640fd0a5f8b15c4 57a96e070640fd0a5f8b158e 57a96e270640fd0a5f8b1594
```

This would save me some typing yesterday when dealing with the workflows. The change itself is fully backward compatible and works the same as before when only a single execution id is provided.


lakshmi-kannan:
👍

Issue #2861, opened 2016-08-10
Rule - `st2ctl reload` Reloads all except rules


vaishalipavashe:
- Intervaltimer rule was created and tested. Tried disabling it by updating rule yaml file and reloaded st2ctl as `st2ctl reload` but still rule is enabled:

![image](https://cloud.githubusercontent.com/assets/10110869/17563463/7e2da484-5efd-11e6-9e4e-d3015a31c859.png)

- From `st2ctl reload` output found that it reloads all except rules. Performed `st2ctl reload --register-rules` to reload rules and now it shows disabled:

![image](https://cloud.githubusercontent.com/assets/10110869/17563523/d66f35b8-5efd-11e6-9625-91cbcb3a4e70.png)

armab:
Yeah, you have to `reload` with `--register-rules` manually.

This is something that Manas described previously in https://github.com/StackStorm/st2/issues/1619#issuecomment-111215415

The reason is that different `packs` can contain default rules and it's safer for administrator not to load them.
If all rules are registered - then you have many triggers enabled, running different actions (which most of us probably don't want).

@kami will correct me if I'm wrong.

armab:
I think there is definitely a user & usability problem here, since it's something that raised many times.

The good approach for me feels like: 
- pack install should load all content within a pack (and not load everything else, out of current pack scope)
- there should be ability to (re)load files for specific pack (isolated)
- make the interaction interface easy for users/predictable/obvious

samirsss:
Isn't it the users responsibility to make sure that the rules they want to not trigger be saved in disabled state? 

I think this ask is a fair thing. I'm getting used to the system and I didn't realize that the 1 rule I had written up didn't get reloaded when I did st2ctl reload. (it wasn't an important rule) but still the expectation that I had was rules would get reloaded.

@lakshmi-kannan @Kami

Kami:
Yeah the idea behind rules being disabled by default was that unlike actions, rules are usually not totally generic and they need to be modified / customized by the user before they can be used.

That's why we decided not to register rules by default - if user didn't modify the rule or disable it, they would get surprised.

samirsss:
Then should we close this PR? I dont have the history on this and hence just thinking from users perspective.

Kami:
We can all think about it some more about re-evaluate our decision aka if we still think that's the best default behavior.

/cc @lakshmi-kannan

lakshmi-kannan:
[...]

PR #2864, opened 2016-08-11
st2ctl reload to register all enabled rules as well

samirsss:
This is for the issue: https://github.com/StackStorm/st2/issues/2861


Kami:
:+1:

PR #2868, opened 2016-08-13
Fix trigger-instance re-emit

dzimine:
one-line fix for STORM-2263: re-emit the trigger fails with 401 because the auth token is not passed.

PS. I don't know enough of CLI unit tests to see how to get it covered, so skipping this part, sorry.


samirsss:
+1

Kami:
Oh wow, haven't seen a commit here from Dmitri for a while now.

LGTM, but where are tests, etc. ? :P

j/k :+1:

PR #2872, opened 2016-08-17
Fix validation of "type" attribute for action parameters

Kami:
This pull request fixes validation for `type` attribute of action parameters.

Existing JSON schema was incorrect so we allowed any string value and not just valid types. Since this wasn't caught before it looks like we didn't have any tests for this edge case.

This was originally discovered and reported by @LindsayHill in https://github.com/StackStorm/st2contrib/pull/532.

## TODO

- [x] Changelog entry
- [x] Tests


LindsayHill:
👍 Thanks for addressing this

Issue #2869, opened 2016-08-16
Removing a trigger can break st2 rule list


DoriftoShoes:
ERROR: 400 Client Error: Bad Request
MESSAGE: Missing TriggerDB object for rule 57ae57614f83fa2f415b0556 for url: http://127.0.0.1:9101/v1/rules/?limit=50

The above is an example message.  If a user removes a pack with packs.uninstall and that removes a trigger, any existing rules that reference that trigger will cause st2 rule list as well as the rules pane of the UI to break.

Kami:
Thanks for the report.

It looks like this is a related issue - https://gist.github.com/Kami/a7279dc3f0b3acb07436b161f309da78.

I will look into it next week - API needs to be made a but more robust so it doesn't break if some data is missing or similar.

Kami:
This has been fixed and fix included in v2.0.0.

PR #2875, opened 2016-08-22
Rules API robustness improvements

Kami:
This pull request addresses #2869 and allows users to still list and retrieve rules even if a rule in a database references a trigger which has been deleted because a pack which provides that trigger has been removed.

This is a pretty common use case which we don't handle right now - we only clean up triggers for rules which belong to the pack which is uninstalled. If there is a pack A and pack B with a rule which references a trigger inside a pack A and user uninstalls pack A, rule inside a Pack B won't be deleted (that's of course when user is working via the API and / or using `packs.uninstall` command).

One option would be to also clean up all the rules which reference a trigger from the pack which is being uninstalled (even the rules from other packs). I personally think that's a reasonable behavior for API (but not for on disk content, of course).

What do others think?

### DIFF
--- a/st2api/st2api/controllers/v1/rules.py
+++ b/st2api/st2api/controllers/v1/rules.py
@@ -76,7 +76,8 @@ class RuleController(resource.ContentPackResourceController):
     @request_user_has_resource_db_permission(permission_type=PermissionType.RULE_VIEW)
     @jsexpose(arg_types=[str])
     def get_one(self, ref_or_id):
-        return super(RuleController, self)._get_one(ref_or_id)
+        from_model_kwargs = {'ignore_missing_trigger': True}
+        return super(RuleController, self)._get_one(ref_or_id, from_model_kwargs=from_model_kwargs)
 
     @jsexpose(body_cls=RuleAPI, status_code=http_client.CREATED)
     @request_user_has_resource_api_permission(permission_type=PermissionType.RULE_CREATE)
--- a/st2api/st2api/controllers/v1/rules.py
+++ b/st2api/st2api/controllers/v1/rules.py
@@ -70,7 +70,8 @@ class RuleController(resource.ContentPackResourceController):
     @request_user_has_permission(permission_type=PermissionType.RULE_LIST)
     @jsexpose()
     def get_all(self, **kwargs):
-        return super(RuleController, self)._get_all(**kwargs)
+        from_model_kwargs = {'ignore_missing_trigger': True}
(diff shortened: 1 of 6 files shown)


Kami:
Actually, now that I think about it some more - I don't think we should do the cascading delete since that's unexpected behavior.

If you delete (rm -rf) pack from disk, you also don't expect rules from other packs which reference triggers from deleted pack to be deleted.

st2api/st2api/controllers/v1/rules.py
```diff
@@ -70,12 +70,14 @@ class RuleController(resource.ContentPackResourceController):
...
     def get_all(self, **kwargs):
-        return super(RuleController, self)._get_all(**kwargs)
+        from_model_kwargs = {'ignore_missing_trigger': True}
+        return super(RuleController, self)._get_all(from_model_kwargs=from_model_kwargs, **kwargs)
 
     @request_user_has_resource_db_permission(permission_type=PermissionType.RULE_VIEW)
     @jsexpose(arg_types=[str])
     def get_one(self, ref_or_id):
-        return super(RuleController, self)._get_one(ref_or_id)
+        from_model_kwargs = {'ignore_missing_trigger': True}
```
lakshmi-kannan: Instead of ignore_missing_trigger, use `list_rules_with_missing_triggers` to be more explicit.

Kami: It's the "generic" argument to `from_model` which also applies to get one, not just list...

lakshmi-kannan:
>I don't think we should do the cascading delete since that's unexpected behavior.

I agree. We shouldn't do this. We should however warn user though that the trigger is used in pack B. We'll get that as part of pack management.

st2api/st2api/controllers/v1/rules.py
lakshmi-kannan: By "list", I didn't mean to imply REST. I can see the confusion :). Then `include_rules_with_missing_triggers` may be better. I'll let you decide.

Kami:
Yeah, for one we could just add a simple logging statement to `packs.uninstall` for now :)

Kami:
Here we go - d7cdaeb.

>
  stderr: 'st2.actions.python.UnregisterPackAction: DEBUG    Removing pack aws.

>    st2.actions.python.UnregisterPackAction: WARNING  Rule "sample_rule_sqs" references deleted trigger "aws.sqs_new_message"

>    st2.actions.python.UnregisterPackAction: INFO     Removed pack aws.

PR #2882, opened 2016-08-24
CLI common functionality refactor

Kami:
This pull request refactors some common functionality so it can be re-used inside other StackStorm CLI based apps (e.g. BWC CLI) without code duplication and copy and paste.


Kami:
Added some more tests, etc. Considering this as done.

PR #2887, opened 2016-09-06
Fix handling of result and trigger_instance attribute in st2 execution get command

### LINKED ISSUE #2858: CLI: `--attr` couldn't retrieve result attributes
Reference: https://docs.stackstorm.com/cli.html
Steps with Screenshots:
- Execute local action
image

- Use `-k` to retrieve result attribute, which works without any issues:
image

- As per documentation, we can use `-attr` to retrieve execution attributes:

image

- In my case it prints whole result set with `--attr` but fails when I try to retrieve result attributes individually. It prints `None`:

image

Kami:
We didn't correctly handle `exclude_attributes` API query parameter if user selected to display a child attribute of either `result` or `trigger_instance` object when using `st2 execution get` command.

For example, `--attr result` worked, but `--attr result.stdout` didn't because we incorrectly excluded `result` attribute from the API response in the second scenario.

This PR fixes that.

selection_098

Resolves #2858.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,10 @@ Changelog
 In development
 --------------
 
+* Fix ``st2 execution get`` command so now ``--attr`` argument correctly works with child
+  properties of the ``result`` and ``trigger_instance`` dictionary (e.g. ``--attr
+  result.stdout result.stderr``). (bug fix)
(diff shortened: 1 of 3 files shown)


emedvedev:
LGTM

PR #2895, opened 2016-09-12
Refactor chatops to process input alias matching server side

tonybaloney:
This PR is proposing to allow clients to submit text captured via another method (e.g. Python bot or direct input) into the StackStorm chatops system. To cater for this, StackStorm needs to be changed to:
* Move the business logic of flattening action alias format strings into regex patterns into an API controller
* Move the matching logic from hubot-stackstorm into a server-side controller

Strategy:

- [x] Create a utility submodule in `st2common.util` for the alias matching/regex logic
- [x] Create a series of test scenarios for checking regex logic in the `re` stdlib compared with the existing hubot code.
- [x] Create an API action for matching a text input from a chatops window to an alias/action reference
- [x] Create a helper function on the `st2client` project, e.g. **`st2 alias match "honk-honk"`**

Functionality in future PRs:
- [ ] Create an API view for formatted alias patterns
- [ ] Create an API view for showing help strings of aliases
- [ ] Modify the `st2stream` API to support server side filtering (not required but preferred).

This is a collaboration with @nzlosh


estee-tew:
https://stackstorm.slack.com/archives/community/p1473660729000786

nzlosh:
@estee-tew I don't have access to the url you posted.  Would you be able to share the comments here?

emedvedev:
This is amazing and long due; one of our chatops roadmap items, actually. 

The regex is already built here on the back-end: https://github.com/StackStorm/st2/blob/c013fc72111ae9141b5c474f32ec85dae3660498/st2common/st2common/models/utils/action_alias_utils.py
You should be able to make it give you a full list of compiled regexes mapped to aliases without much effort. Overall, the plan looks solid. :heart:

enykeev:
@nzlosh seems like a bug in @estee-tew. The correct URL is https://stackstorm-community.slack.com/archives/community/p1473660729000786.

st2api/st2api/controllers/v1/chatops.py
```diff
@@ -0,0 +1,95 @@
...
+from st2common.util.alias_matching import (list_format_strings_from_aliases,
+                                           match_command_to_alias)
+
+
+http_client = six.moves.http_client
+
+LOG = logging.getLogger(__name__)
+
+
+class ChatopsController(resource.ContentPackResourceController):
```
tonybaloney: opinions here?

Kami:
Nice, indeed!

That's been our goal for a while now, but we never got to it. Moving all the parsing and matching to the server will allow user to make clients a lot thinner / dumb which should make it easier to develop clients for other protocols and also reduce the client maintenance burden.

Well, at least the basic implementation. For performance reasons, in some scenarios (a lot of messages per second and a lot of aliases) it might make sense to move some pre-parsing / matching to the client.

st2api/st2api/controllers/v1/chatops.py
Kami: Is it possible to keep those methods / operations on the alias controller?

The reason is that right now StackStorm concept is called action alias and the code doesn't use "ChatOps" term anywhere else (afaik) so it might be a bit confusing to call it ChatOps here...

estee-tew:
https://stackstorm.slack.com/archives/community/p1473748033001093

tonybaloney:
this is essentially done. I'm going to keep this PR simple and just add the API method to complete matching. 

Ready for review.

estee-tew:
https://stackstorm.slack.com/archives/community/p1473815171001196

tonybaloney:
ok pending tests this is done. The `st2 action-alias execute` command is a bit wonky. The pattern expects the model to be the same as the command, but because its action-alias returning an execution object it looks a little strange (but similar to the action execute command).

st2api/st2api/controllers/v1/actionalias.py
```diff
@@ -58,6 +73,31 @@ def get_all(self, **kwargs):
     def get_one(self, ref_or_id):
         return super(ActionAliasController, self)._get_one(ref_or_id)
 
+    @jsexpose(arg_types=[str], status_code=http_client.ACCEPTED)
+    def match(self, command, **kwargs):
+        """
+            Run a chatops command
+
+            Handles requests:
+                POST /chatops/match
```
Kami: This should say `POST /actionalias/match`, right?

st2api/st2api/controllers/v1/actionalias.py
```diff
@@ -58,6 +73,31 @@ def get_all(self, **kwargs):
...
+
+                command=hello%20world
+        """
+        try:
+            # 1. Get aliases
+            aliases = super(ActionAliasController, self)._get_all(**kwargs)
+            # 2. Match alias(es) to command
+            matches = match_command_to_alias(command, aliases)
+            if len(matches) > 1:
+                raise ActionAliasAmbiguityException("Command matched more than 1 pattern",
```
Kami: Nice :+1: 

Probably also a good idea to include the actual command on the exception instance and inside the exception error message.

st2api/st2api/controllers/v1/actionalias.py
```diff
@@ -58,6 +73,31 @@ def get_all(self, **kwargs):
...
+                                                    matches=matches)
+            return [self._match_tuple_to_dict(match) for match in matches]
+        except (ActionAliasAmbiguityException) as e:
+            # TODO : error on unmatched alias
```
Kami: We should also throw the same exception here as above, right?

st2client/st2client/commands/action.py
```diff
@@ -246,22 +246,6 @@ class ActionRunCommandMixin(object):
     def get_resource(self, ref_or_id, **kwargs):
         return self.get_resource_by_ref_or_id(ref_or_id=ref_or_id, **kwargs)
 
-    @add_auth_token_to_kwargs_from_cli
```
Kami: Hm, any particular reason why you re-ordered those methods? It's slightly confusing, or did you also make any other changes?

Kami: Also, minor style / consistency thing - we tend to put "public" methods at the top and "private" at the bottom / after "public" ones.

st2client/st2client/commands/action.py
```diff
@@ -781,8 +742,48 @@ def _get_inherited_env_vars(self):
...
+            sys.stdout.write('\n')
+
+            if execution.status == LIVEACTION_STATUS_CANCELED:
+                return execution
+
+        return execution
+
 
-class ActionRunCommand(ActionRunCommandMixin, resource.ResourceCommand):
+class ActionRunCommand(ActionRunCommandMixin,
```
Kami: We use 100 characters long lines so those breaks are not needed.

st2client/st2client/commands/action_alias.py
```diff
@@ -40,3 +54,106 @@ class ActionAliasGetCommand(resource.ContentPackResourceGetCommand):
...
+    def run_and_print(self, args, **kwargs):
+        instances = self.run(args, **kwargs)
+        self.print_output(instances, table.MultiColumnTable,
+                          attributes=args.attr, widths=args.width,
+                          json=args.json, yaml=args.yaml)
+
+
+class ActionAliasExecuteCommand(ActionRunCommandMixin,
+                                resource.ResourceCommand):
+    def __init__(self, resource, *args, **kwargs):
```
Kami: Nice, I like the idea of having a CLI command for those two operations - should make testing some scenarios easier :+1:

st2client/st2client/commands/action_alias.py
```diff
@@ -40,3 +54,106 @@ class ActionAliasGetCommand(resource.ContentPackResourceGetCommand):
...
+        if len(matches) > 1:
+            raise ActionAliasAmbiguityException("Too many matches for provided command",
+                                                matches=matches)
+        elif len(matches) == 0:
+            raise ActionAliasAmbiguityException("Could not locate an ActionAlias with a "
+                                                "matching format command", matches=matches)
+        match = matches[0]
+        action_alias_db = match[0]
+
+        if not action_alias_db.enabled:
```
Kami: Hm, looks like we are encoding some server side logic inside the CLI command...

Can't we just call the API operation (alias execute) here so we don't need to duplicate code in multiple places?

st2common/st2common/exceptions/actionalias.py
```diff
@@ -0,0 +1,26 @@
...
+# limitations under the License.
+
+from st2common.exceptions import StackStormBaseException
+
+__all__ = [
+    'ActionAliasAmbiguityException'
+]
+
+
+class ActionAliasAmbiguityException(StackStormBaseException):
```
Kami: Should probably inherit from `ValueError`, this way the status code returned by the API on this exception will be correct (400 - bad request).

st2common/st2common/models/api/base.py
```diff
@@ -283,6 +282,7 @@ def cast_func(value):
             try:
                 result = f(*args, **kwargs)
             except TypeError as e:
+                LOG.debug(traceback.format_exc())
```
Kami: Is this needed here or you just add it to debug things locally?

Just wondering since this exception should get logger elsewhere - in one layer above where this function is used. If it's not, then I'm fine with this change.

st2common/st2common/util/actionalias_matching.py
```diff
@@ -0,0 +1,90 @@
...
+
+    :param aliases: The list of aliases
+    :type  aliases: ``list`` of :class:`st2common.models.api.action.ActionAliasAPI`
+
+    :return: A description of potential execution patterns in a list of aliases.
+    :rtype: ``list`` of ``list``
+    '''
+    patterns = []
+    for alias in aliases:
+        for _format in alias.formats:
```
Kami: `format_` (yeah, blame pep8 - https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments :D)

st2common/tests/unit/test_util_actionalias_matching.py
```diff
@@ -0,0 +1,129 @@
...
+
+import unittest2
+import mongoengine as me
+
+from st2common.constants.types import ResourceType
+from st2common.models.db import stormbase
+import st2common.util.actionalias_matching as matching
+
+
+class MemoryActionAliasDB(stormbase.StormFoundationDB, stormbase.ContentPackResourceMixin,
```
Kami: Just curious - why do we need this class here?

Can't we re-use one from the existing model? If we have another model here, it means another place to maintain in sync and things will get out of date...

Kami:
Added some comments in-line.

Some minor things need to be updated / changed, but overall great work :+1:

st2client/st2client/commands/action.py
tonybaloney: reverted changes. I did split up the mixin, but later decided against it.

st2client/st2client/commands/action_alias.py
tonybaloney: the tracing logic is different, I looked at what was done in the action.py execute command module and refactored for this scenario. Agreed it's not ideal, the alias execute API doesn't wait for the execution to be completed from what I saw.

st2common/st2common/models/api/base.py
tonybaloney: the stack trace gets swallowed here, i uncovered a bug in the decorator jsexpose function, but nothing critical. I'll remove.

st2common/tests/unit/test_util_actionalias_matching.py
tonybaloney: because the `__init__` for `ActionAliasDB` does a foreign key lookup. I'll see if I can monkey patch the lookup out instead.

tonybaloney:
done.

estee-tew:
https://stackstorm.slack.com/archives/community/p1473913804001384

st2api/st2api/controllers/v1/actionalias.py
```diff
@@ -58,6 +73,37 @@ def get_all(self, **kwargs):
     def get_one(self, ref_or_id):
         return super(ActionAliasController, self)._get_one(ref_or_id)
 
+    @jsexpose(arg_types=[str], status_code=http_client.ACCEPTED)
+    def match(self, command, **kwargs):
```
Kami: Can user also pass other arguments to this function? If not, we should probably get rid of `**kwargs`.

st2client/st2client/commands/action_alias.py
```diff
@@ -40,3 +49,98 @@ class ActionAliasGetCommand(resource.ContentPackResourceGetCommand):
...
+        execution.format = representation
+        execution.command = args.command_text
+        execution.source_channel = 'cli'  # ?
+        execution.notification_channel = None
+        execution.notification_route = None
+        execution.user = args.user
+
+        action_exec_mgr = self.app.client.managers['ActionAliasExecution']
+
+        execution = action_exec_mgr.create(execution, **kwargs)
```
Kami: Nice, much better :+1:

st2client/st2client/models/core.py
```diff
@@ -337,6 +339,20 @@ def __init__(self, resource, endpoint, cacert=None, debug=False):
         self.debug = debug
         self.client = httpclient.HTTPClient(root=endpoint, cacert=cacert, debug=debug)
 
+    @add_auth_token_to_kwargs_from_env
+    def match(self, command, **kwargs):
+        url = '/%s/match' % self.resource.get_url_path_name()
+        query_str = urlencode({'command': command})
+        response = self.client.post_raw(url, query_str, **kwargs)
```
Kami: Hm, this is sent as query string, right?

If so, requests library should encode it correctly as long as the requests.post method is called with the params=params argument.

Kami:
Nice, thanks, LGTM!

I will check it out and play a bit with it locally and if everything looks good merge it into master.

st2api/st2api/controllers/v1/actionalias.py
tonybaloney: they could for the purposes of reducing the criteria of the underlying `get_all` call, which is where they are passed. So filtering by pack for example.

st2client/st2client/models/core.py
tonybaloney: It wasn't 100% clear on the docs (I can't stand that documentation!). I know this works now :-)

st2api/st2api/controllers/v1/actionalias.py
```diff
@@ -48,6 +52,17 @@ class ActionAliasController(resource.ContentPackResourceController):
         'sort': ['pack', 'name']
     }
 
+    _custom_actions = {
```
Kami: Hm, it appears `_custom_actions` is only supported in newer versions of pecan because I get 404 in execute :/

We still use 0.7.0 because of backward incompatible changes in newer versions which make it hassle to upgrade.

Kami: Actually, it looks like old versions support it as well. Must be something else - looking into it.

Kami: Actually, it's a CLI issue - CLI doesn't correctly send query param to the method...


```python
match() takes exactly 2 arguments (1 given)
```

I will make `urlencode` and requests changes mentioned below, I believe this should fix it.

Kami: Actually, since this is a post we should probably send data as body and not via query string.

This way it will be consistent with other models. Yes, it's a bit "heavy" and we need a wrapper API model, but we already do that in other places.

I will make that change, hope you don't mind :)

Kami: There we go - 4afaa54f7e980a2fd5d66c9f82dda127b4387366

It works now, woo!

```bash
(virtualenv)vagrant@vagrant-ubuntu-trusty-64:/data/stanley$ python ./st2client/st2client/shell.py action-alias match "run uptime on localhost"
2016-09-15 12:37:32,670  WARNING - Auth API server is not available, skipping authentication.
+------------------+---------------------------------------------+
| name             | description                                 |
+------------------+---------------------------------------------+
| remote_shell_cmd | Execute a command on a remote host via SSH. |
+------------------+---------------------------------------------+
(virtualenv)vagrant@vagrant-ubuntu-trusty-64:/data/stanley$ python ./st2client/st2client/shell.py action-alias execute "run uptime on localhost"
2016-09-15 12:37:34,473  WARNING - Auth API server is not available, skipping authentication.
Started execution, id '57da960e0640fd107ab26409'
Hold tight while I run command: *uptime* on hosts *localhost*
```

Minor thing - what do other things, I think `execute` could be blocking by default (similar as `st2 run` command) and optionally user can specify `--async` flag to make it non-blocking. WDYT?

st2client/st2client/commands/action_alias.py
```diff
@@ -40,3 +49,98 @@ class ActionAliasGetCommand(resource.ContentPackResourceGetCommand):
...
+                                 help=('List of attributes to include in the '
+                                       'output. "all" will return all '
+                                       'attributes.'))
+        self.parser.add_argument('-w', '--width', nargs='+', type=int,
+                                 default=None,
+                                 help=('Set the width of columns in output.'))
+
+    @resource.add_auth_token_to_kwargs_from_cli
+    def run(self, args, **kwargs):
+        action_alias, representation = self.manager.match(args.command_text, **kwargs)
```
Kami: I'm fine with having this command for now, but for true server side matching and execution which would allow us to make clients event dumber I think we should also move this to the server side (e.g. "match and execute" endpoint or similar).

This is not a blocker right now though. We can leave it here and introduce this new API endpoint in the future. The CLI command is forward compatible anyway (once the feature is available on the server, we just need to change the code here).

This would also save client / user one API call which is actually good in this case, especially if you send every single line posted to chat to the API.

estee-tew:
https://stackstorm.slack.com/archives/community/p1473943558001392

Kami:
Merged into master via #2904.

Thanks again for great work and a PR! :+1: 

And in the future if you find yourself with nothing to do (yeah, I know, that never happens :D) it would be great if you could also look into this https://github.com/StackStorm/st2/pull/2895#r78954366.

st2client/st2client/commands/action_alias.py
emedvedev: You should only send the lines that are addressed to Hubot, so that's not a huge concern, but match-and-execute would be pretty good regardless.

emedvedev:
:heart: :heart: :heart: 
Awesome job!

st2client/st2client/commands/action_alias.py
tonybaloney: the patterns got a bit wonky crossing resource types between controllers. I'll have a think about it.

PR #2897, opened 2016-09-13
IPv6 binding support for the dev script

tonybaloney:


estee-tew:
https://stackstorm.slack.com/archives/community/p1473727725001089

bigmstone:
I wonder. Should we listen on v6 and v4 by default? Having it be a binary option v4 or v6 feels weird. They're not usually mutually exclusive (in production).

tonybaloney:
I tried that. Gunicorn broke. 

Get Outlook for iOS




On Tue, Sep 13, 2016 at 11:02 AM +1000, "Matthew Stone" <notifications@github.com> wrote:












I wonder. Should we listen on v6 and v4 by default? Having it be a binary option v4 or v6 feels weird. They're not usually mutually exclusive (in production).

bigmstone:
Did you try this?
http://docs.gunicorn.org/en/stable/settings.html#server-socket
Specifically: `gunicorn -b 127.0.0.1:8000 -b [::1]:8000 test:app`
Also, I  do realize this is for a dev script and thus not all that important but would be good to know it works or breaks in dev.

tonybaloney:
In production, all of the service bindings get proxies through NGINX anyway, so having local IPv6 bindings wouldn't make much difference (unless I guess for those people addressing the API locally). I figured I was in the minority wanting proper IPv6 bindings on any host address (not localhost) so that's why I didn't make it default. 
There was a thread on gunicorn about multiple bindings, some systems consider [::] to mean any host, 4 or 6, some don't. 

Get Outlook for iOS




On Tue, Sep 13, 2016 at 11:11 AM +1000, "Matthew Stone" <notifications@github.com> wrote:












Did you try this?

http://docs.gunicorn.org/en/stable/settings.html#server-socket

Specifically: gunicorn -b 127.0.0.1:8000 -b [::1]:8000 test:app

Also, I  do realize this is for a dev script and thus not all that important but would be good to know it works or breaks in dev.

tonybaloney:
this is the article, its system dependent, so yuck. https://github.com/benoitc/gunicorn/issues/1138

bigmstone:
Yeah that's nasty. Also looks like the v6 addr will make it listen on both addresses on the platforms we support.

This seems to be the best way to approach it for now.

bigmstone:
I'm +1 on the change. Will let someone else put eyes on it and merge.

emedvedev:
+1. Shouldn't be a problem. I'll merge and then check what our CI says as well.

PR #2898, opened 2016-09-13
Fix logrotate bug

Kami:
This fixes the "mysterious" st2actionrunner service manager file is deleted issue.

It was originally caught by vp on Slack (thanks!).

I will also cherry-pick this into v2.0.0 and publish new packages once they are built.

### DIFF
--- a/conf/logrotate.conf
+++ b/conf/logrotate.conf
@@ -105,7 +105,7 @@ notifempty
   postrotate
     # Delete all files that haven't been modified in over 30 days
     # Clean up stale PID logs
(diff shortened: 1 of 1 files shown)


Kami:
One option is to cherry-pick this into v2.0.0 and build and publish new v2.0.0, other is to create v2.1.0 patch release.

@lakshmi-kannan @m4dcoder WDYT?

estee-tew:
https://stackstorm.slack.com/archives/community/p1473760228001097

PR #2904, opened 2016-09-15
Refactor chatops to process input alias matching server side

Kami:
Some fixes and changes on top of #2895.

See https://github.com/StackStorm/st2/pull/2895#r78949824 for details.

Closes #2904.


st2api/st2api/controllers/v1/actionalias.py
```diff
@@ -58,6 +74,39 @@ def get_all(self, **kwargs):
...
+            elif len(matches) == 0:
+                raise ActionAliasAmbiguityException("Command '%s' matched no patterns" %
+                                                    command,
+                                                    matches=[],
+                                                    command=command)
+            return [self._match_tuple_to_dict(match) for match in matches]
+        except (ActionAliasAmbiguityException) as e:
+            LOG.exception('Command "%s" matched (%s) patterns.', e.command, len(e.matches))
+            pecan.abort(http_client.BAD_REQUEST, str(e))
+            return [self._match_tuple_to_dict(match) for match in e.matches]
```
lakshmi-kannan: I don't think this line will ever be hit because you have pecan.abort()

st2api/st2api/controllers/v1/actionalias.py
```diff
@@ -58,6 +74,39 @@ def get_all(self, **kwargs):
     def get_one(self, ref_or_id):
         return super(ActionAliasController, self)._get_one(ref_or_id)
 
+    @jsexpose(arg_types=[str], body_cls=ActionAliasMatchAPI, status_code=http_client.ACCEPTED)
+    def match(self, action_alias_match_api, **kwargs):
```
lakshmi-kannan: I am not sure I like match. I'd prefer `validate` or `test`.

lakshmi-kannan:
Nice!

st2api/st2api/controllers/v1/actionalias.py
emedvedev: It's neither validation nor testing though, because you would validate or test against something. "Match" is probably both the closest and the most clear word for what's happening. :)

PR #2906, opened 2016-09-15
Add support for Jinja template notation for default parameter values for object parameters

Kami:
This pull request adds support for using Jinja notation for action default parameter values for parameters of type object.

This issue (lack of feature / support for objects) was caught and reported by @code-ape on Slack.

He had an action similar to the one below:

```yaml
---
name: example
pack: examples
description: Example
runner_type: http-request
enabled: true
parameters:
    driveraddress:
        type: string
        required: true
        default: "10.255.4.15:3000" 
    host:
        type: string
        required: true
    port:
        type: integer
        required: true
    params:
        type: object
        immutable: true
        default:
            host: "{{host}}"
            port: "{{port}}"
    url:
        type: string
        immutable: true
        default: "http://{{driveraddress}}/api/gen-payload/ve"
```

This didn't work since default values for `params` parameter was not rendered since existing code didn't support objects.

This pull request fixes that and adds support for rendering default parameter values for objects with a single level of nesting (like the one in the example above).

We don't support arbitrary level of nesting since this would increase computation time and potentially lead to resource exhaustion, infinite recursion or similar. It's the same story with lists which we already supported - we only render first level items in the list and don't support nested lists.

Not supporting objects basically means that in a scenario like the one above, user won't be able to leverage the HTTP runner, but they will need to write code for Python runner action or similar.

## TODO

- [x] Tests
- [x] Changelog

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -22,6 +22,8 @@ In development
   ``st2 alias-execution execute``. (new feature) #2895
 
   Contributed by Anthony Shaw
+* Fix a bug with action default parameter values not supporting Jinja template
+  notation for parameters of type ``object``. (bug fix, improvement)
 
 2.0.0 - August 31, 2016
 -----------------------
--- a/st2common/st2common/util/param.py
+++ b/st2common/st2common/util/param.py
@@ -135,7 +135,7 @@ def _validate(G):
     '''
     for name in G.nodes():
         if 'value' not in G.node[name] and 'template' not in G.node[name]:
-            msg = 'Dependecy unsatisfied for parameter "%s"' % name
+            msg = 'Dependecy unsatisfied in %s' % name
             raise ParamException(msg)
 
     if not nx.is_directed_acyclic_graph(G):
--- a/st2common/st2common/util/param.py
+++ b/st2common/st2common/util/param.py
@@ -162,19 +162,33 @@ def _resolve_dependencies(G):
     for name in nx.topological_sort(G):
         node = G.node[name]
         try:
-            if 'template' in node and isinstance(node.get('template', None), list):
-                rendered_list = list()
-                for template in G.node[name]['template']:
-                    rendered_list.append(
-                        _render(dict(template=template), context)
-                    )
-                context[name] = rendered_list
+            template = node.get('template', None)
+
+            # Special case for non simple types which contains Jinja notation (lists, dicts)
+            if 'template' in node and isinstance(template, (list, dict)):
+                if isinstance(template, list):
+                    rendered_list = list()
+
+                    for template in G.node[name]['template']:
+                        rendered_list.append(
+                            _render(dict(template=template), context)
+                        )
+                    context[name] = rendered_list
+                elif isinstance(template, dict):
+                    rendered_dict = dict()
+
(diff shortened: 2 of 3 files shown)


Kami:
If I get this finished by tomorrow (tests) I propose to also include it in v2.0.1. What do others think?

code-ape:
Attempted to test this change with use-case shown. This will now be accepted when running `st2 run packs.load` however attempting to run the action inside stackstorm results in the following:

```json
{
  ...,
  "parameters": {
    "host": "10.255.4.15",
    "driver-address": "localhost:3000",
    "port": 22
  },
  "runner": {
    "runner_module": "st2actions.runners.httprunner",
    "uid": "runner_type:http-request",
    "enabled": true,
    "name": "http-request",
    "runner_parameters": {
      ...
    },
    "id": "57d95dc00640fd1e8d6dd431",
    "description": "A HTTP client for running HTTP actions."
  },
  "web_url": "https://vagrant-ubuntu-trusty-64/#/history/57dae8ad0640fd283a76d4c9/general",
  "result": {
    "traceback": "  File \"/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2api/controllers/v1/actionexecutions.py\", line 154, in _schedule_execution\n    liveaction_db.context)\n  File \"/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/util/param.py\", line 218, in render_live_params\n    _validate(G)\n  File \"/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/util/param.py\", line 139, in _validate\n    raise ParamException(msg)\n",
    "error": "Dependecy unsatisfied for parameter \"driver\""
  },
  "context": {
    "trace_context": {},
    "user": "st2admin"
  },
  "action": {
    "runner_type": "http-request",
    "name": "edit_ve_gen_payload",
    "parameters": {
      "url": {
        "default": "http://{{ driver-address }}/api/gen-payload/ve",
        "type": "string",
        "immutable": true
      },
      "host": {
        "required": true,
        "type": "string"
      },
      "driver-address": {
        "required": true,
        "type": "string"
      },
      "params": {
        "default": {
          "host": "10.255.4.15",
          "port": 22
        },
        "type": "object",
        "immutable": true
      },
      "port": {
        "required": true,
        "type": "integer"
      }
    },
 ...
}
```

Kami:
@code-ape Yeah, that's actually another issue - replace dash (`-`) with underscore (`_`) and it should work.

To fix the actual object issue, I just removed dash all together in the example I used to test it. Tomorrow I will also look into dash issue, might just be a YAML think and you need to quote it, but need to dig in.

Kami:
@code-ape Yeah, I confirmed that's a Jinja limitation:

> Variable names should never include dashes, they will be interpreted as a subtraction. This is pretty much true for any language, so stick with underscores.

In short, replace dash with underscore to solve the second problem.

Kami:
Merged into master, cherry picked into v2.0.1 release branch.

Kami:
@code-ape and here is "fix" for dash in the parameter name - #2908

We shouldn't have allowed dashes in the parameters in the first place, but there was a bug in the code so every character was allowed.

PR #2908, opened 2016-09-16
Only allow word characters and don't allow slashes in action parameter names

### LINKED ISSUE #2907: Drop support for dashes in action parameter names
Dashes are not supported everywhere in the pipeline (for example, Jinja variable template notation doesn't support it so users can't use Jinja notation to reference parameters with dashes in the name) so we should remove support for it to reduce confusion and issues.

We already discussed this a year or so ago since it was already causing confusion and issues back then.

I propose to update packs in st2contrib which utilize dashes (if any, I think there was an issue related to ansible or salt pack because it used dashes) and remove support for it in the parameter names in v2.1.

Kami:
While looking into #2907 I noticed our code is broken / incorrect everywhere where we use `patternProperties` functionally of json schema.

We limit valid names for property to any valid word character (`\w` - `a-Z0-0_`), but we don't specify `'additionalProperties': False`. Not specifying that means that any other name which doesn't fall into that (e.g. parameter name which contains a dash or any other character which falls outside of `\w`) will be considered as additional property and no validation will be performed for that property/

For one, this means we allow dashes and other special characters which causes issue because for example, Jinja doesn't support dashes in the variable names, etc. In addition to that, it's dangerous (and not user-friendly) since we don't perform any validation for those attributes.

Since this change is potentially backward incompatible, I propose to include it in v2.1. We should also go over st2contrib and make sure we don't have any parameters with invalid names.

Resolves #2907.

## TODO

- [x] Tests
- [x] Changelog entry
- [ ] Upgrade notes entry

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -24,6 +24,12 @@ In development
   Contributed by Anthony Shaw
 * Fix a bug with action default parameter values not supporting Jinja template
   notation for parameters of type ``object``. (bug fix, improvement)
+* Only allow valid word characters (``a-z``, ``0-9`` and ``_``) to be used for action parameter
+  names. Previously, due to bug in the code, any character was allowed.
+
+  If you have an existing action definition which uses parameter with an invalid character, the
+  parameter name needs to be adjusted otherwise action registration will fail. (bug fix,
+  improvement)
 
 2.0.0 - August 31, 2016
 -----------------------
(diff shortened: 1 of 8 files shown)


Kami:
I'll merge this into master and look for failures (if any) in Travis CI for st2contrib repo.

Kami:
It looks like there was only one affected pack in st2contrib which I already fixed - 014ffa1ed2320031736968598901eeeeba88513b.

PR #2913, opened 2016-09-19
Throw more user-friendly exception on invalid action parameter name

Kami:
This pull request builds on top of #2908 and update the exception to be more user-friendly.

Original json schema exception is very unfriendly and it's not immediately obvious what is going on (invalid parameter name).

### DIFF
--- a/st2common/st2common/bootstrap/actionsregistrar.py
+++ b/st2common/st2common/bootstrap/actionsregistrar.py
@@ -14,8 +14,10 @@
 # limitations under the License.
 
 import os
+import re
 
 import six
+import jsonschema
 
 from st2common import log as logging
 from st2common.constants.meta import ALLOWED_EXTS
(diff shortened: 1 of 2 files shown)


st2common/st2common/bootstrap/actionsregistrar.py
```diff
@@ -124,7 +126,24 @@ def _register_action(self, pack, action):
...
+            action_api.validate()
+        except jsonschema.ValidationError as e:
+            # We throw a more user-friendly exception on invalid parameter name
+            msg = str(e)
+
+            is_invalid_parameter_name = 'Additional properties are not allowed' in msg
+            is_invalid_parameter_name &= 'in schema[\'properties\'][\'parameters\']' in msg
+
+            if is_invalid_parameter_name:
+                parameter_name = re.search('\'(.+?)\' was unexpected', msg).groups()[0]
```
Kami: That's nasty, I don't like it, but sadly the exception class instance doesn't include variable with offending parameter :/

lakshmi-kannan:
lgtm

st2common/st2common/bootstrap/actionsregistrar.py
Kami: At some point I will also look into upstream repo, probably shouldn't be too hard (famous last words) to make a change to include offending attribute on the exception class instance.

PR #2916, opened 2016-09-19
Refactor mistral to support new CANCELLED status

m4dcoder:
Use the newly introduced CANCELLED state in mistral for workflow cancellation. Currently, st2 put the workflow in a PAUSED state in mistral.

### DIFF
--- a/contrib/runners/mistral_v2/query/mistral_v2.py
+++ b/contrib/runners/mistral_v2/query/mistral_v2.py
@@ -17,7 +17,8 @@ LOG = logging.getLogger(__name__)
 
 DONE_STATES = {
     'ERROR': action_constants.LIVEACTION_STATUS_FAILED,
-    'SUCCESS': action_constants.LIVEACTION_STATUS_SUCCEEDED
+    'SUCCESS': action_constants.LIVEACTION_STATUS_SUCCEEDED,
+    'CANCELLED': action_constants.LIVEACTION_STATUS_CANCELED
 }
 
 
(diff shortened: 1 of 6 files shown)


Kami:
Nice, LGTM :+1:

Kami:
@m4dcoder what's happening with this PR - do we still need / want it?

andrew-regan:
Is this related to PR #3125 at all?

m4dcoder:
None what so ever

Kami:
@m4dcoder Is this PR still active (aka needed / wanted)?

m4dcoder:
This PR is still active. Once our mistral fork catches up w/ upstream mistral, I will refactor this branch.

contrib/runners/mistral_v2/query/mistral_v2.py
```diff
@@ -134,15 +135,15 @@ def _determine_execution_status(self, execution_id, wf_state, tasks):
...
-            status = action_constants.LIVEACTION_STATUS_RUNNING
+        # Keep the execution in running state if there are active tasks.
+        # In certain use cases, Mistral sets the workflow state to
+        # completion prior to task completion.
+        if is_action_canceled and active_tasks:
+            status = action_constants.LIVEACTION_STATUS_CANCELING
         elif wf_state not in DONE_STATES:
             status = action_constants.LIVEACTION_STATUS_RUNNING
+        elif wf_state in DONE_STATES and active_tasks:
+            status = action_constants.LIVEACTION_STATUS_RUNNING
```
Kami: Is it possible that we would some how end here and st2 execution would get stuck in "running" state?

I guess this could happen if Mistral task never finishes, which means we would need to manually cancel Mistral task, right?

If that's the case, if needed, we should perhaps consider adding `st2 execution cancel --force` or similar which would propagate this force cancel to Mistral so user doesn't need to interact directly with Mistral...

I'm just "thinking out loud" about this scenario, because we had this issue quite many times in the past (executions getting "stuck" in running state).

m4dcoder: Having a cancel with force options sounds good. I'll add this feature in a separate PR.

PR #2917, opened 2016-09-20
s/{{system.}}/{{st2kv.system.}} and s/{{user.}}/{{st2kv.user.}}

lakshmi-kannan:
## What?

PR tries to move away from {{system.}} and {{user.}} jinja template to access datastore items to {{st2system.}} and {{st2user.}}

## Why?

See conversation here https://github.com/StackStorm/st2/pull/2772

## TODO:

* [x] Backward compatibility tests
* [x] Migration tool for datastore
* [ ]  ~~st2contrib changes~~
* [X] Docs changes https://github.com/StackStorm/st2docs/pull/286
  - [X] Migration docs
  - [X] Jinja change in docs
* [x] Changelog


Kami:
Nice :+1: 

Looks like your TODO already covers all the stuff I wanted to comment on so nothing from me for now :)

emedvedev:
I still want `st2kv.system` and `st2kv.user`. :cry: I remember the discussion about the extra dot, but I think this syntax is the clearest semantically. It also corresponds to `st2kv()`.

Kami:
> I still want st2kv.system and st2kv.user. :cry: I remember the discussion about the extra dot, but I think this syntax is the clearest semantically. It also corresponds to st2kv().

I do agree that this would be ideal and more explicit... :)

lakshmi-kannan:
@emedvedev @Kami I now have the changes to get {{st2kv.system.}} and {{st2kv.user.}} to work. It was a pain slightly but it's done now. Have a look and if you guys are cool, I'll work on the docs changes.

emedvedev:
Looks good. You make me happy :heart:

st2api/st2api/controllers/v1/keyvalue.py
```diff
@@ -61,19 +63,22 @@ def __init__(self):
...
+    def get_one(self, name, scope=None, user=None, decrypt=False):
         """
             List key by name.
 
             Handle:
                 GET /keys/key1
         """
-        self._validate_scope(scope=scope)
+        if not scope:
+            scope = SYSTEM_SCOPE
```
bigmstone: Why's this better?

bigmstone: Assumption: If `NoneType` is assigned to method call it still gets replaced with `SYSTEM_SCOPE`?

st2common/tests/unit/test_config_loader.py
```diff
@@ -11,7 +11,7 @@
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
-# limitations under the License.
+# limitations under the License.s
```
bigmstone: Whoops.

st2actions/st2actions/runners/actionchainrunner.py
```diff
@@ -464,6 +470,11 @@ def _render_publish_vars(action_node, action_parameters, execution_result,
         context.update(chain_vars)
         context.update({RESULTS_KEY: previous_execution_results})
         context.update({SYSTEM_SCOPE: KeyValueLookup(scope=SYSTEM_SCOPE)})
+        context.update({
+            DATASTORE_PARENT_SCOPE: {
```
Kami: Looks like this piece is repeated / duplicated multiple time across the code-base.

Perhaps "get_base_datastore_jinja_context" or similar utility function is in order? :P

st2api/st2api/controllers/v1/keyvalue.py
Kami: Yeah, I would prefer to do both if we have to - default in the method signature and also leave this part here.

And in short I don't think it's better, we should always assign default values to the arguments in the function signature (that's how it was designed and supposed to be used), but I assume it has something to do with the way pecan handles argument so the actual default argument value is overridden with `None` when calling the method.

In any case, as mentioned above, I would still prefer to do both in such scenario - specify default argument value in the method signature and do additional default assignment in the method body, if needed.

st2api/st2api/controllers/v1/keyvalue.py
```diff
@@ -96,20 +101,24 @@ def get_one(self, name, scope=SYSTEM_SCOPE, user=None, decrypt=False):
...
+
         if user:
             # Providing a user implies a user scope
             scope = USER_SCOPE
 
+        scope = get_datastore_full_scope(scope)
         requester_user = get_requester()
         user = user or requester_user
-        is_all_scope = (scope == 'all')
+        is_all_scope = (scope == ALL_SCOPE)
```
Kami: :+1:

st2api/st2api/controllers/v1/keyvalue.py
```diff
@@ -144,15 +153,18 @@ def get_all(self, prefix=None, scope=SYSTEM_SCOPE, user=None, decrypt=False, **k
         return kvp_apis
 
     @jsexpose(arg_types=[str, str, str], body_cls=KeyValuePairSetAPI)
-    def put(self, kvp, name, scope=SYSTEM_SCOPE):
+    def put(self, kvp, name, scope=None):
```
Kami: Same here, let's please also default the argument value here. Or do you need to differentiate between system scope being a default value and a system scope requested by the user?

st2api/st2api/controllers/v1/keyvalue.py
```diff
@@ -203,13 +215,17 @@ def put(self, kvp, name, scope=SYSTEM_SCOPE):
         return kvp_api
 
     @jsexpose(arg_types=[str, str, str], status_code=http_client.NO_CONTENT)
-    def delete(self, name, scope=SYSTEM_SCOPE, user=None):
+    def delete(self, name, scope=None, user=None):
```
Kami: Same here.

st2api/tests/unit/controllers/v1/test_kvps.py
```diff
@@ -107,65 +113,86 @@ def test_put(self):
     def test_put_with_scope(self):
         self.app.put_json('/v1/keys/%s' % 'keystone_endpoint', KVP,
                           expect_errors=False)
-        self.app.put_json('/v1/keys/%s?scope=system' % 'keystone_version', KVP_2,
+        self.app.put_json('/v1/keys/%s?scope=st2kv.system' % 'keystone_version', KVP_2,
```
Kami: Ah, so the query param value changed as well.

Do we still handle "system" and "user" for backward compatibility reasons? If we do, we should add some explicit backward compatibility tests if we don't have any already...

st2common/bin/migrations/v1.5/st2-migrate-datastore-to-include-scope-secret.py
```diff
@@ -36,11 +36,10 @@ def migrate_datastore():
         for kvp in key_value_items:
             kvp_id = getattr(kvp, 'id', None)
             secret = getattr(kvp, 'secret', False)
-            encrypted = getattr(kvp, 'encrypted', False)
             scope = getattr(kvp, 'scope', SYSTEM_SCOPE)
             new_kvp_db = KeyValuePairDB(id=kvp_id, name=kvp.name,
                                         expire_timestamp=kvp.expire_timestamp,
-                                        value=kvp.value, secret=secret, encrypted=encrypted,
```
Kami: Hm, why was this removed?

Kami:
Added some comments, mostly small stuff.

Overall nice work! :+1:

st2api/st2api/controllers/v1/keyvalue.py
lakshmi-kannan: > but I assume it has something to do with the way pecan handles argument so the actual default argument value is overridden with None when calling the method.

Yes, this is what I noticed. The behavior is inconsistent. Even if supply a value for the API call using a query param, pecan defaulted to SYSTEM_SCOPE. I didn't chase this but I decided to be on safe side.

st2actions/st2actions/runners/actionchainrunner.py
lakshmi-kannan: Yeah, I thought about this but there are cases where we want just SYSTEM_SCOPE inside DATASTORE_PARENT_SCOPE and in some cases we also want USER_SCOPE (config loader case). So then we need three get_system_datastore_jinja_context, get_user_datastore_jinja_context and get_datastore_jinja_context. It seemed more indirection to me.

st2api/st2api/controllers/v1/keyvalue.py
lakshmi-kannan: Fixed in 77bb7f8

st2common/tests/unit/test_config_loader.py
lakshmi-kannan: Fixed in 77bb7f8

st2api/st2api/controllers/v1/keyvalue.py
lakshmi-kannan: 77bb7f8

st2api/st2api/controllers/v1/keyvalue.py
lakshmi-kannan: 77bb7f8

st2api/tests/unit/controllers/v1/test_kvps.py
lakshmi-kannan: Already have them
https://github.com/StackStorm/st2/pull/2917/files#diff-8191f7b7ffefbccfae995461e8925299R169
https://github.com/StackStorm/st2/pull/2917/files#diff-8191f7b7ffefbccfae995461e8925299R177

st2common/bin/migrations/v1.5/st2-migrate-datastore-to-include-scope-secret.py
lakshmi-kannan: It appears I changed the data model but the migration script was never changed. This is an issue now because of "additionalProperties: False" change.

Kami:
LGTM :+1: 

Let's not forget to update upgrade notes if it's not already updated. And let's keep the upgrade notes entry verbose and user-friendly with some examples of the old and new notation...

PR #2927, opened 2016-09-27
Adding connection retry for RabbitMQ

bigmstone:
This PR adds connection retry to RabbitMQ transport. This will recover from issues where RabbitMQ is unavailable immediately on startup.

- [x] Add config options for parameters.

### DIFF
--- a/st2common/st2common/transport/bootstrap_utils.py
+++ b/st2common/st2common/transport/bootstrap_utils.py
@@ -14,6 +14,8 @@
 # limitations under the License.
 
 import retrying
+import socket
+from oslo_config import cfg
 from kombu import Connection
 from st2common import log as logging
 from st2common.transport import utils as transport_utils
@@ -57,10 +59,7 @@ def _do_register_exchange(exchange, connection, channel, retry_wrapper):
         LOG.exception('Failed to register exchange : %s.', exchange.name)
 
 
-@retrying.retry(
-    wait_fixed=10000,
-    stop_max_attempt_number=10)
-def register_exchanges():
+def _register_exchanges():
     LOG.debug('Registering exchanges...')
     connection_urls = transport_utils.get_messaging_urls()
     with Connection(connection_urls) as conn:
@@ -73,3 +72,12 @@ def register_exchanges():
                                       retry_wrapper=retry_wrapper)
 
         retry_wrapper.run(connection=conn, wrapped_callback=wrapped_register_exchanges)
+
+
+def register_exchanges():
+    retrying_obj = retrying.Retrying(
+        retry_on_exception=socket.error,
+        wait_fixed=cfg.CONF.messaging.connection_retry_wait,
+        stop_max_attempt_number=cfg.CONF.messaging.connection_retries
+    )
+    return retrying_obj.call(_register_exchanges)
--- a/st2common/st2common/transport/bootstrap_utils.py
+++ b/st2common/st2common/transport/bootstrap_utils.py
@@ -59,7 +59,7 @@ def _do_register_exchange(exchange, connection, channel, retry_wrapper):
         LOG.exception('Failed to register exchange : %s.', exchange.name)
 
 
-def _register_exchanges():
+def register_exchanges():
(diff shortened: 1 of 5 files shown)


bigmstone:
I've left the parameters hard-coded for now. Want to get a :+1: on the method and then I'll add config params for this.

st2common/st2common/transport/bootstrap_utils.py
```diff
@@ -56,6 +57,9 @@ def _do_register_exchange(exchange, connection, channel, retry_wrapper):
         LOG.exception('Failed to register exchange : %s.', exchange.name)
 
 
+@retrying.retry(
```
Kami: Is it possible to only catch more specific exceptions aka connection errors?

Otherwise it could potentially mask some other / real issues...

bigmstone: Yes, good catch. I can use `retry_on_exception=exception`.

Kami: Yeah, please do :)

st2common/st2common/transport/bootstrap_utils.py
```diff
@@ -69,3 +72,12 @@ def wrapped_register_exchanges(connection, channel):
                                       retry_wrapper=retry_wrapper)
 
         retry_wrapper.run(connection=conn, wrapped_callback=wrapped_register_exchanges)
+
+
+def register_exchanges():
+    retrying_obj = retrying.Retrying(
```
Kami: For consistency with `db_setup` can you please add `register_exchanges_with_retry` method which calls this one with `retrying` stuff?

Thanks!

bigmstone: Per in-person discussion this change request would lose consistency with existing functionality with `db_setup` so we'll keep it the same.

Kami:
LGTM :+1:

armab:
Awesome!
The right thing :+1: 


Just to know, will it try to reconnect for example on RabbitMQ service restart or during temporary networking issue (when MQ is external service) or self-healing retry in this PR works only during "bootstrap"?

bigmstone:
@armab It works during both. If it fails for any `socket.error` reason. There's also restart mechanisms for each channel which can handle other types of failures, but that's not new functionality. The number of times it will retry, and the time between each try is configurable in `st2actions/config.py`

st2common/st2common/transport/bootstrap_utils.py
```diff
@@ -69,3 +72,12 @@ def wrapped_register_exchanges(connection, channel):
                                       retry_wrapper=retry_wrapper)
 
         retry_wrapper.run(connection=conn, wrapped_callback=wrapped_register_exchanges)
+
+
+def register_exchanges_with_retry():
+    retrying_obj = retrying.Retrying(
+        retry_on_exception=socket.error,
+        wait_fixed=cfg.CONF.messaging.connection_retry_wait,
+        stop_max_attempt_number=cfg.CONF.messaging.connection_retries
```
manasdk: If this property is only used for retries in register_exchanges why is it called connection_retries? The issue with a broadly named property like this is that it is bound to confuse someone as to why this is getting st2 to retry the specified no of times.

Perhaps there is value in wiring these properties up to https://github.com/StackStorm/st2/blob/master/st2common/st2common/transport/connection_retry_wrapper.py#L30. The semantics there are slightly different but worth looking into to bring together to mean the same thing.

bigmstone: The main issue boils down to [this line](https://github.com/StackStorm/st2/blob/6a530bc5a6053d8cfd36b61903af3f2845c7bdbd/st2common/st2common/transport/bootstrap_utils.py#L65). The retry logic is backed in below this line; however, the connection is made on line 65. If rabbitmq server is down when that line is executed it will fail without retry.

As for the config options I tend to agree that those options should be brought together. I was unaware that they exist. I would be more in favor of moving `connection_retry_wrapper`'s properties to the config file since those properties seem like things that should be encouraged to modify.

PR #2931, opened 2016-10-03
WIP: Render user params (jinja) in API

lakshmi-kannan:
See https://github.com/StackStorm/st2/pull/2772#issuecomment-228106632
https://github.com/StackStorm/st2/pull/2770

Tests will still fail because we still support {{user.}} and {{system.}} notations for backward compatibility. We should deprecate the support soon. I just opened the PR to capture the commits so I can reuse this when we deprecate {{user.}} and {{system.}}


lakshmi-kannan:
Closing. Will open when actively working on this.

PR #2932, opened 2016-10-04
Pack intialization

enykeev:
```
(virtualenv)vagrant@arkham:/dock/st2$ st2 pack create thing --author="enykeev" --version="0.0.1" --description="some description"

	[ succeeded ] create directory
	[ succeeded ] change permissions
	[ succeeded ] change ownership
	[ succeeded ] render pack.yml
	[ succeeded ] save pack.yml
	[ succeeded ] create subdirectories

(virtualenv)vagrant@arkham:/dock/st2$ cat /opt/stackstorm/packs/thing/pack.yml
---
name: thing
description: some description
version: 0.0.1
author: enykeev

(virtualenv)vagrant@arkham:/dock/st2$ ls -lah /opt/stackstorm/packs/thing
total 24K
drwxrwxr-x 5 vagrant stanley 4.0K Oct  4 10:54 .
drwxr-xr-x 7 vagrant vagrant 4.0K Oct  4 10:54 ..
drwxrwxr-x 2 vagrant stanley 4.0K Oct  4 10:54 actions
-rwxrwxr-x 1 vagrant stanley   78 Oct  4 11:22 pack.yml
drwxrwxr-x 2 vagrant stanley 4.0K Oct  4 10:54 rules
drwxrwxr-x 2 vagrant stanley 4.0K Oct  4 10:54 tests
```

### DIFF
--- a/contrib/packs/actions/create.meta.yaml
+++ b/contrib/packs/actions/create.meta.yaml
@@ -1,0 +1,41 @@
+---
+  name: "create"
+  runner_type: "action-chain"
+  description: "Creates a new pack"
+  enabled: true
+  entry_point: "workflows/create.yaml"
+  parameters:
+    name:
+      type: "string"
+      required: true
+      description: "Name of the pack to create"
+    description:
+      type: "string"
+      description: "Pack description"
+    keywords:
+      type: "array"
+      items:
+        type: "string"
+      description: "Search keywords"
+    version:
+      type: "string"
+      description: "Initial version of the pack"
+      required: true
+    author:
+      type: "string"
+      description: "Pack's author"
+      required: true
+    email:
+      type: "string"
+      description: "Pack's author email"
+    directories:
+      type: "array"
+      default:
+        - actions
+        - rules
+        - tests
+      description: "List of internal directories to create within the pack"
+    env:
+      type: "object"
+      description: "Optional environment variables"
+      required: false
--- a/contrib/packs/actions/create.meta.yaml
+++ b/contrib/packs/actions/create.meta.yaml
@@ -11,6 +11,7 @@
       description: "Name of the pack to create"
     description:
       type: "string"
+      required: true
       description: "Pack description"
     keywords:
       type: "array"
@@ -20,7 +21,7 @@
     version:
       type: "string"
       description: "Initial version of the pack"
-      required: true
+      default: "0.0.1"
     author:
       type: "string"
       description: "Pack's author"
--- a/contrib/packs/actions/workflows/create.yaml
+++ b/contrib/packs/actions/workflows/create.yaml
@@ -1,0 +1,51 @@
+---
+vars:
+  pack_path: '/opt/stackstorm/packs/{{ name }}'
+  pack_group: stanley
+  pack_permissions: '775'
+
+chain:
+  -
+    name: "create directory"
+    ref: "core.local_sudo"
+    parameters:
+      cmd: "mkdir -p {{ pack_path }}"
+    on-success: "change permissions"
+  -
+    name: "change permissions"
+    ref: "core.local_sudo"
+    parameters:
+      cmd: "chmod -R {{ pack_permissions }} {{ pack_path }}"
+    on-success: "change ownership"
+  -
+    name: "change ownership"
+    ref: "core.local_sudo"
+    parameters:
+      cmd: "chgrp -R {{ pack_group }} {{ pack_path }}"
+    on-success: "render pack.yml"
+  -
+    name: "render pack.yml"
+    ref: "packs.render_template"
+    parameters:
+      template_path: "templates/pack.yml.j2"
+      context:
+        name: "{{ name }}"
+        description: "{{ description or '' }}"
+        keywords: "{{ keywords or '' }}"
+        version: "{{ version }}"
+        author: "{{ author }}"
+        email: "{{ email or '' }}"
+    on-success: "save pack.yml"
+  -
+    name: "save pack.yml"
+    ref: "core.local"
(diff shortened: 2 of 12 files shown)


contrib/packs/actions/create.meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+      description: "Initial version of the pack"
+      required: true
+    author:
+      type: "string"
+      description: "Pack's author"
+      required: true
+    email:
+      type: "string"
+      description: "Pack's author email"
+    directories:
```
emedvedev: I wonder if we should have an archived boilerplate instead, and then just unpack it and change `pack.yaml`. It would be really nice to place READMEs inside the dirs, like we used to do for old st2contrib packs, a default icon, etc. Does it seem reasonable?

contrib/packs/actions/workflows/create.yaml
```diff
@@ -0,0 +1,51 @@
...
+      cmd: "chmod -R {{ pack_permissions }} {{ pack_path }}"
+    on-success: "change ownership"
+  -
+    name: "change ownership"
+    ref: "core.local_sudo"
+    parameters:
+      cmd: "chgrp -R {{ pack_group }} {{ pack_path }}"
+    on-success: "render pack.yml"
+  -
+    name: "render pack.yml"
```
emedvedev: We should probably make it `.yaml` and not `.yml`, for consistency's sake. All the packs we have use `pack.yaml`, and even if we rename meta in every pack to `.yml` (for whatever reason), everything else in StackStorm has the `.yaml` extension, too.

So if there's no strong point for it, I'd just change the default file to `.yaml`. `yml` is nice, but more trouble than it's worth.

st2client/st2client/commands/pack.py
```diff
@@ -150,10 +150,28 @@ def __init__(self, resource, *args, **kwargs):
...
+                                 help='Author of the %s.' %
+                                 resource.get_display_name().lower())
+        self.parser.add_argument('--email',
+                                 help='%s author\'s email.' %
+                                 resource.get_display_name())
 
     @resource.add_auth_token_to_kwargs_from_cli
     def run(self, args, **kwargs):
-        return self.manager.create(args.name, **kwargs)
+        return self.manager.create(args, **kwargs)
```
emedvedev: How hard would it be to make an interactive command here, like `adduser`? I.e. `st2 pack create` or `st2 pack create blabla`, then the user just answers questions instead of writing a huge one-liner? I'm fine with not doing it, but it's a nice feature.

contrib/packs/actions/create.meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+  runner_type: "action-chain"
+  description: "Creates a new pack"
+  enabled: true
+  entry_point: "workflows/create.yaml"
+  parameters:
+    name:
+      type: "string"
+      required: true
+      description: "Name of the pack to create"
+    description:
```
emedvedev: I feel like the description should be required at this point: we'll have the index, and the front-end, and the description is prominent in both, so the packs without description will suck. Wdyt?

contrib/packs/actions/create.meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+      type: "string"
+      description: "Pack description"
+    keywords:
+      type: "array"
+      items:
+        type: "string"
+      description: "Search keywords"
+    version:
```
emedvedev: Shouldn't it just default to `0.1` or `0.0.1`?

contrib/packs/actions/workflows/create.yaml
```diff
@@ -0,0 +1,51 @@
...
+  pack_permissions: '775'
+
+chain:
+  -
+    name: "create directory"
+    ref: "core.local_sudo"
+    parameters:
+      cmd: "mkdir -p {{ pack_path }}"
+    on-success: "change permissions"
+  -
```
emedvedev: I think we should init an empty git repo during creation (DZ proposed this as something we should have for v1.0 scope AFAIR), and also have git remote as an optional parameter. We're assuming git all the way, right? And it won't break anything for people who aren't gonna use it, too.

contrib/packs/actions/create.meta.yaml
enykeev: Probably. This spec is a conversation piece more than proposal, really, I don't have strong feelings on pack.yml content.

contrib/packs/actions/create.meta.yaml
enykeev: sure

contrib/packs/actions/workflows/create.yaml
enykeev: Yeah, knock yourself out. Personally, I'm using them pretty inconsistently throughout my code so I'd rather we support both... not sure it worth the effort though. In this particular case, I'm ok with `.yaml`.

contrib/packs/actions/workflows/create.yaml
enykeev: Yeah, we should.

contrib/packs/actions/create.meta.yaml
emedvedev: Yeah, I mean the rest: directories, READMEs, sample files, etc. A sample config schema would be useful, too. In the future, even more stuff. Pack.yaml can (and even should) be generated from a template. :)

emedvedev:
Awesome! Most of the comments are addressed, and the interactive mode can be implemented later. The only other thing I'd add in this PR is a pack template for better structure, instead of dir list + pack.yaml.

PR #2933, opened 2016-10-04
Addition of pause core action

paul-mulvihill:
### Status 
- Expansion of Core Actions

### Description
Addition of pause action allowing for a random pause between steps of a workflow.
action can take fixed duration and wait that long or using boolean random field will pause for up to that time in seconds

### Checklist (tick everything that applies)

- [x] Metadata: pack.yaml, icon, structure (required for new packs)
- [x] Version bump (required for changed packs)
- [x] Code linting (required, can be done after the PR checks)
- [ ] [Tests](https://docs.stackstorm.com/development/pack_testing.html) (not required but really recommended)

### DIFF
--- a/contrib/core/actions/pause.py
+++ b/contrib/core/actions/pause.py
@@ -1,0 +1,30 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import time
+from random import randint
+from st2actions.runners.pythonrunner import Action
+
+
+class pause(Action):
+    def run(self, max_pause, random):
+        if random:
+            delay = randint(1, int(max_pause))
+        else:
+            delay = max_pause
+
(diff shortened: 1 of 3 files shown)


contrib/core/actions/pause.yaml
```diff
@@ -0,0 +1,18 @@
+---
+  name: "pause"
+  runner_type: "run-python"
+  description: "Action to Pause workflow"
```
dzimine: well it will pause "everything", not just a workflow :) Pls craft a description that also hints an intended use.

contrib/core/actions/pause.py
```diff
@@ -0,0 +1,30 @@
...
+class pause(Action):
+    def run(self, max_pause, random):
+        if random:
+            delay = randint(1, int(max_pause))
+        else:
+            delay = max_pause
+
+        time.sleep(delay)
+
+        return
```
dzimine: May be return something? Like actual delay? Again you know better given you use this for real.

dzimine:
👍

contrib/core/actions/pause.yaml
paul-mulvihill: Unless i'm mistaken it just pauses the current thread of the workflow/subworkflow. 
I've had it running 10 concurrently and it doesn't do them one at a time. but i'll update the description.

contrib/core/actions/pause.py
paul-mulvihill: Add it to return the delay value.. makes no difference to me. but may be useful for some.

contrib/core/actions/pause.yaml
Kami: Yeah, it will pause "everything" - well just the actual action execution will behave as sleep, but it has little use outside of workflows... :)

Kami:
As discussed on Slack, I'm fine with it being Python runner action, but this does involve some overhead (spawning of a new process, even though it will just idle there since stuff in the subprocess is evenlet monkey patch so time.sleep shouldn't consume any resources), but  eventually I would prefer to add it as a functionality to the core or similar.

Kami:
Merged, thanks!

I will update the changelog and related stuff.

PR #2938, opened 2016-10-05
Pack search and working with pack indexes

emedvedev:
This PR adds:

1. A config entry that allows you to specify a pack index or a list of indexes. It defaults to our main index in the Exchange, but if there's some security concerns in a large enterprise, they can maintain their own index, or even a local one on the same host, or use both at the same time and have both their own packs and Exchange packs available for discovery. (Oh boy, @armab will geek out about the multiple indexes part.)

2. Helper functions in `st2common` for (1) fetching and combining the indexes, and (2) the index search.

3. The `pack/search` API endpoint that supports getting a single pack object by its name, or listing all packs matching a query.

4. The corresponding CLI commands (`st2 pack search <query>` and `st2 pack show <pack>`).

5. Changes to the `packs.download` action, which is the cool part. Now, if you want to download a pack by its short name, the action won't just assume the pack is hosted on the Exchange github org: it will query the index instead. This has the added benefit of using local indexes, or enterprise-only indexes, or listing external packs in our own index in the future, and the end user won't need the full URL to install any of them via CLI or `packs.install`. Of course, the full URLs are still supported, since maintaining an index is unnecessary in most cases.

### DIFF
--- a/st2common/st2common/services/packs.py
+++ b/st2common/st2common/services/packs.py
@@ -39,9 +39,9 @@ def fetch_pack_index(index_url=None):
     """
     if not index_url:
         index_urls = cfg.CONF.content.index_url
-    else if isinstance(index_url, str):
+    elif isinstance(index_url, str):
         index_urls = [index_url]
-    else if hasattr(index_url, '__iter__'):
+    elif hasattr(index_url, '__iter__'):
         index_urls = index_url
     else:
         raise TypeError('"index_url" should either be a string or an iterable object.')
@@ -72,5 +72,5 @@ def search_pack_index(query=None, pack=None):
             if query in value:
                 matches.append(pack)
                 break
-                
+
     return matches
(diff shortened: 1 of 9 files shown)


emedvedev:
I'll wait for @enykeev's review and feedback, and then it can be merged into the main pack management PR (#2842).

st2common/st2common/services/packs.py
```diff
@@ -26,3 +31,48 @@ def get_pack_by_ref(pack_ref):
...
+    return result
+
+
+def search_pack_index(query=None, pack=None):
+    """
+    Search the pack index either by pack name or by query.
+    Returns a pack object if the pack name is specified, otherwise returns
+    a list of matches for a query.
+    """
+    if (not query and not pack) or (query and pack):
```
enykeev: So xor? :P

emedvedev: Yeah, but there's no xor operator in python (except for `^`), and this is the most explicit way that doesn't involve imports or weird casting :)

PR #2939, opened 2016-10-05
Add ability to decrypt kv item in Jinja via filters

lakshmi-kannan:
## What? 

We want the ability to use decrypted items like password etc in rules, actions etc. 

Desired syntax: {{st2kv.system.password | decrypt_kv}}

## What actually got fixed?

* dectrypt_kv filter is now available for use
* Action chain init vars could actually throw an exception especially when using jinja templates. The exception raised was unclear and did not say anything about Jinja. I noticed this when testing and fixed it. 
* I introduced a bug in scope migration PR https://github.com/StackStorm/st2/pull/2917. Basically, the default scope in KeyValuePairDB still was "system" as opposed to "st2kv.system". This and other bugs in lookup code resulted in "st2kv.system" not resolving correctly. The tests were writing to DB with scope=system and the read code path was also doing the same even when I wanted the scope to be"st2kv.system". I now fixed the DB model and the lookup code. 

## Sample run of action chain example

```
(virtualenv)vagrant@st2dev /m/s/s/st2 ❯❯❯ st2 key list --scope=system                                   STORM-2398/decrypt_datastore_item ✭
+--------+----------------------+--------+-----------+--------------+------+------------------+
| name   | value                | secret | encrypted | scope        | user | expire_timestamp |
+--------+----------------------+--------+-----------+--------------+------+------------------+
| my_cmd | 0083E1033A1216710B74 | True   | True      | st2kv.system |      |                  |
|        | 3F77A94AD70E1D8EB9BA |        |           |              |      |                  |
|        | FD55DDDCD710650AE5DE |        |           |              |      |                  |
|        | 2779365D256875DAB23B |        |           |              |      |                  |
|        | B70F6D21EF495817C7C9 |        |           |              |      |                  |
|        | B7342C911B39A6       |        |           |              |      |                  |
+--------+----------------------+--------+-----------+--------------+------+------------------+
(virtualenv)vagrant@st2dev /m/s/s/st2 ❯❯❯ st2 run examples.                                             STORM-2398/decrypt_datastore_item ✭
(virtualenv)vagrant@st2dev /m/s/s/st2 ❯❯❯ st2 run examples.decrypt_kv_jinja_filter                    ⏎ STORM-2398/decrypt_datastore_item ✭
..
id: 5807a770d9d7ed087e225752
action.ref: examples.decrypt_kv_jinja_filter
parameters: None
status: succeeded
start_timestamp: 2016-10-19T17:03:44.637437Z
end_timestamp: 2016-10-19T17:03:46.902712Z
+--------------------------+------------------------+------+------------+----------------------+
| id                       | status                 | task | action     | start_timestamp      |
+--------------------------+------------------------+------+------------+----------------------+
| 5807a771d9d7ed0838c04985 | succeeded (1s elapsed) | c1   | core.local | Wed, 19 Oct 2016     |
|                          |                        |      |            | 17:03:45 UTC         |
+--------------------------+------------------------+------+------------+----------------------+
(virtualenv)vagrant@st2dev /m/s/s/st2 ❯❯❯ st2 execution get 5807a771d9d7ed0838c04985                    STORM-2398/decrypt_datastore_item ✭
id: 5807a771d9d7ed0838c04985
parameters:
  cmd: echo foo
result:
  failed: false
  return_code: 0
  stderr: ''
  stdout: foo
  succeeded: true
```

## Docs

https://github.com/StackStorm/st2docs/pull/299


Kami:
I'm personally fine with this change, but we need to think (and document) how it affects RBAC.

From what it looks like right now, every person who has either create / edit & execute access to actions / rules can potentially read a decrypted value from a datastore. That's probably fine, but we just need to think about it some more and document the behavior to avoid surprises.

PR #2941, opened 2016-10-07
paramiko-ssh-config-support

eedgar:


eedgar:
I don't expect this to be merged yet.  I was going to work with lakstorm on the unit tests.

dzimine:
Thank you for your PR. 
+1, let's have @lakshmi-kannan take another look and we'll get it in.

dzimine:
Thank you for your PR. 
+1, let's have @lakshmi-kannan take another look and we'll get it in.

st2actions/st2actions/runners/ssh/paramiko_ssh.py
```diff
@@ -579,13 +579,39 @@ def _connect(self, host, socket=None):
...
 
+        conninfo = self._get_conninfo_from_ssh_config(conninfo)
         client = paramiko.SSHClient()
         client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
         client.connect(**conninfo)
 
         return client
 
     @staticmethod
+    def _get_conninfo_from_ssh_config(conninfo):
```
lakshmi-kannan: A minor thing that's concerning is that you are passing an object and mutating it inside a static method. Plus the method is called _get_conninfo... 

Should we refactor this to something like:

```
if cfg.CONF.ssh_runner.use_ssh_config:
  conninfo_ssh_config = self._get_conninfo_from_ssh_config_for_host(host)
  conninfo.update(conninfo_ssh_config)
```

lakshmi-kannan:
You should also edit CHANGELOG.rst and call it as a new feature and put your name next to it :)
https://github.com/StackStorm/st2/blob/master/CHANGELOG.rst#in-development

lakshmi-kannan:
Besides those comments, LGTM. Once you have them addressed, I'll merge.

st2actions/st2actions/runners/ssh/paramiko_ssh.py
Kami: Yeah, what lakshmi has said - In fact, I would prefer this method to just retrieve info from the SSH config and return a dictionary and the merging to happen somewhere else.

Kami:
Nice work, thanks.

It would also be great if you can add corresponding documentation to https://github.com/stackstorm/st2docs

st2actions/st2actions/config.py
```diff
@@ -73,7 +73,10 @@ def _register_action_runner_opts():
                         'Works only with Paramiko SSH runner.'),
         cfg.BoolOpt('use_ssh_config',
                     default=False,
-                    help='Use the .ssh/config file. Useful to override ports etc.')
+                    help='Use the .ssh/config file. Useful to override ports etc.'),
+        cfg.StrOpt('ssh_config_path',
+                   default='.ssh/config',
```
Kami: Shouldn't this default to `~/.ssh/config` if we want it to use config from the user under which action runner process is running (stanley)?

Otherwise it seems it will default to a relative path based on the cwd where action runner is started.

humblearner:
Fixed here: https://github.com/StackStorm/st2/pull/3032

Kami:
Thanks to both of you - let's continue in #3032 then.

PR #2947, opened 2016-10-13
st2-check-license added

punalpatel:
st2-check-license will verify whether license is valid or not


st2common/bin/st2-check-license
```diff
@@ -0,0 +1,17 @@
+#! /bin/bash
+if [[ $# -eq 0 ]] ; then
+echo 'No License Key Entered'
```
lakshmi-kannan: Indent these two lines

st2common/bin/st2-check-license
```diff
@@ -0,0 +1,17 @@
...
+echo 'Enter License Key'
+exit 0
+fi
+LICENSE=$@
+BWC_LICENSE_CHECK_URL="https://${LICENSE}:@packagecloud.io/install/repositories/StackStorm/enterprise/script.deb.sh"
+echo $LICENSE
```
lakshmi-kannan: Remove these two echo lines

st2common/bin/st2-check-license
```diff
@@ -0,0 +1,17 @@
...
+echo $BWC_LICENSE_CHECK_URL
+LICENSE_CURL_TEST=`curl --output /dev/null --silent --fail ${BWC_LICENSE_CHECK_URL}`
+if [ $? -ne 0 ]; then
```
lakshmi-kannan: Instead of $?, shouldn't you be checking the value of variable LICENSE_CURL_TEST?

Kami:
I agree with lakshmi - good overall idea.

Another thing which needs to be done when this is finished is packaging changes, but I can take over that if necessary.

Kami:
Another thing - Python interpreter is required for StackStorm installation so if you feel more comfortable with Python I'm also fine with using Python instead of bash.

In any case, it's your call, I'm fine with both :)

st2common/bin/st2-check-license
```diff
@@ -0,0 +1,15 @@
+#! /bin/bash
+if [[ $# -eq 0 ]] ; then
+   echo 'No License Key Entered'
+   echo 'Enter License Key'
```
Kami: Maybe better to say something like:

```bash
echo "No license key entered"
echo "Usage: $0 <license key to check>"
```

lakshmi-kannan:
>Another thing which needs to be done when this is finished is packaging changes, but I can take over that if necessary.

@Kami This is the reason I asked him to put the file in st2common/bin/. Anything there is already packaged and put in /usr/bin/ I believe.

Kami:
@lakshmi-kannan Yeah, you are correct - it's put in the `/opt/stackstorm`, but not symlinked into a PATH directory (e.g. /usr/local/bin) or similar.

I do think it would be a good idea to symlink it, but yeah, that's a simple st2-packages change I / we can do once this is merged.

Kami:
Another place which needs to be updates is `st2common/setup.py`

Kami:
I went ahead and merge changes into master - thanks!

I will make the setup.py and other change locally. Thanks again!

PR #2958, opened 2016-10-17
API exception handling fixes, allow user to filter get all responses on multiple resources id

Kami:
This pull request addresses two issues:

1. It fixes handling of `ValidationError` exception in the. Previously if exception was thrown due to an invalid ID being provided in the query string internal error was returned, not bad request error is returned.

2. It allows users to filter "get all" API responses on multiple resource ids by providing a comma delimited string value for the query parameter (e.g. `?id=1,2,3,4`).

This allows users to write more efficient clients and also reduces some overhead on the API. Previously user would either need to request all the objects and do late filtering locally or by performing multiple get one requests (one for each resource / object).

I'm personally not a fan id comma delimited string approach (HTTP already solves that by allowing client to provide the same key multiple time - `?id=1&id=2&id=3`) but sadly we already use comma delimited approach in some other places and it looks like pecan doesn't directly support multiple keys notation so I went with this approach for now.

### DIFF
--- a/st2common/st2common/util/api.py
+++ b/st2common/st2common/util/api.py
@@ -16,6 +16,7 @@
 import re
 
 import pecan
+import mongoengine
 from oslo_config import cfg
 from webob import exc as webob_exc
 
@@ -29,8 +30,8 @@ __all__ = [
     'get_mistral_api_url',
 
     'get_requester',
-    'get_exception_for_type_error'
-
+    'get_exception_for_type_error',
+    'get_exception_for_uncaught_api_error'
 ]
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


st2common/st2common/util/api.py
```diff
@@ -140,3 +141,17 @@ def get_exception_for_type_error(func, exc):
...
+def get_exception_for_uncaught_api_error(func, exc):
+    """
+    Utility function which tries to map an uncaught exception throwed inside an
+    API to a more user-friendly exception which is returned instead of returning
+    internal server error.
+    """
+
+    if isinstance(exc, mongoengine.ValidationError):
+        result = webob_exc.HTTPBadRequest(detail=exc.message)
+        return result
```
enykeev: Should have been a part of https://github.com/StackStorm/st2/blob/master/st2common/st2common/hooks.py#L245. Took me quite a while to figure out the hell is going on.

PR #2959, opened 2016-10-17
Add migration script to register runners

lakshmi-kannan:
Docs: https://github.com/StackStorm/st2docs/pull/296


st2common/bin/migrations/v2.1/st2-migrate-runners.py
```diff
@@ -0,0 +1,49 @@
...
+from st2common.util.green.shell import run_command
+
+
+def main():
+    timeout = 180
+
+    args = [
+        'st2ctl',
+        'reload',
+        '--register-all',
```
lakshmi-kannan: Yeah, I thought about it and I could switch back to registering runners alone.

Kami: This would probably be better and fast, since if user has a lot of resources it can take a long time.

PR #2960, opened 2016-10-18
Funnel pack registration through the API

enykeev:
Still somewhat WIP and there is a few questions unanswered, but I was hoping to get your eyes on that one as soon as possible.

With this PR we're steering towards registering all the content through API calls rather than doing it locally.

Known problems:
- registration of even a default set of content takes more than 30 seconds on devbox with nfs share. There might be a problem if user decides to run the command remotely.
- current content bootstrapping includes setting up venvs for python actions. We haven't moved this part yet and it is not clear to me how often the feature is used in the wild and whether it would be ok to keep it separate of the registration process.

By the time we're done with this PR, we're likely to merge `feature/packs`, so PR would have to be recreated against master anyway.

### DIFF
--- a/st2common/st2common/content/bootstrap.py
+++ b/st2common/st2common/content/bootstrap.py
@@ -19,21 +19,12 @@ import logging
 
 from oslo_config import cfg
 
-import st2common
 from st2common import config
 from st2common.script_setup import setup as common_setup
 from st2common.script_setup import teardown as common_teardown
 from st2common.bootstrap.base import ResourceRegistrar
-import st2common.bootstrap.triggersregistrar as triggers_registrar
-import st2common.bootstrap.sensorsregistrar as sensors_registrar
-import st2common.bootstrap.actionsregistrar as actions_registrar
-import st2common.bootstrap.aliasesregistrar as aliases_registrar
-import st2common.bootstrap.policiesregistrar as policies_registrar
-import st2common.bootstrap.runnersregistrar as runners_registrar
-import st2common.bootstrap.rulesregistrar as rules_registrar
-import st2common.bootstrap.ruletypesregistrar as rule_types_registrar
-import st2common.bootstrap.configsregistrar as configs_registrar
 import st2common.content.utils as content_utils
+from st2client.client import Client
 from st2common.util.virtualenvs import setup_pack_virtualenv
 
 __all__ = [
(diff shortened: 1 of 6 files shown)


enykeev:
Setting up venvs takes an awful lot of time and together with the first point, it would no doubt require us to make /packs/register api async too. We need to decide whether we deem this to be a problem for this release.

Kami:
Nice work overall.

Here are some of my high level comments:

1. I think running register-content should still be transparent to the end user and work the same as in the past.

This means we need to generate a token for the admin (root) user and some how pass it to the register content script. One option also is to add support for `~/.st2/config` to the register-content script. This token should also be scoped to a user which only has pack_manage or similar role assigned via RBAC.

2. RBAC

I know this really related to #/2842, but we need for RBAC support for the new pack management API.

Overall, as far as code goes, I don't have too many comments (yet), but as mentioned above, my main "worry" right now is end-to-end experience and deployment (making sure that for the common use-case experience is seamless).

st2common/st2common/content/bootstrap.py
```diff
@@ -19,21 +19,12 @@
 
 from oslo_config import cfg
 
-import st2common
 from st2common import config
 from st2common.script_setup import setup as common_setup
 from st2common.script_setup import teardown as common_teardown
 from st2common.bootstrap.base import ResourceRegistrar
-import st2common.bootstrap.triggersregistrar as triggers_registrar
```
Kami: Can we now also nuke the code for all those registrars or not (yet)?

Kami:
> Setting up venvs takes an awful lot of time and together with the first point, it would no doubt require us to make /packs/register api async too. We need to decide whether we deem this to be a problem for this release.

I think this API being async is fine for now. If needed, we can add (internal) `?sync=1` flag or similar.

Kami:
Actually, the stuff discussed on Slack brings up a good point - for a better user experience (not swallowing errors, etc.) and for backward compatible experience we do need a sync API - we wan't to return an error back if something fails (validation, etc.).

st2common/st2common/content/bootstrap.py
enykeev: We're using the same registrars in #2842, but instead of calling them from bootstrap script, we're importing them in api. At some point I would like us to rewrite the whole thing with networkx to properly assign the dependencies between different types of content.

enykeev:
Choice between sync or async API is not one we can do purely based on what we want. If request takes longer than 30 seconds, it can't be sync (or at least until someone calls me on this bs and points me to why it's not the case anymore). We can synchronize this call on a client level by listening to stream and blocking until we get a message that the content was registered or failed to register and is now in a broken state, but a) we need to start publish this events before we can consume them and b) we still don't have a way to retrieve events happened during reconnect so it might block forever. It's not a huge problem for client, but for scripts that supposed to run automatically, it might eventually became one.

Kami:
Yeah, I'm personally fine with either approach as long as the end user experience is the same.

And yes, even if we do blocking / sync implementation, for reliability we need to implement retry / reconnect to the client.

enykeev:
The spec for SSE already has a way of requesting the missing messages (I believe it is called `event_id`). To support it, we need to keep a list of recent messages on the st2stream (probably just store some configurable amount, say 100, in memory). It is still nowhere near fault-tolerant, but hey, baby steps.

PR #2962, opened 2016-10-18
Remove enum to allow for extra content-type arguments, like charset

pixelrebel:
Having `enum` in the `content_type` schema, restricts users from adding extras like `text/plain; charset="utf-8"`  

If we want to keep the restriction, I suppose we could add a `charset` argument to the send_mail script?

### DIFF
--- a/contrib/core/actions/sendmail.yaml
+++ b/contrib/core/actions/sendmail.yaml
@@ -30,9 +30,6 @@ parameters:
   content_type:
     type: string
     description: Content type of message to be sent
-    enum:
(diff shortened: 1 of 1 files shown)


Kami:
I'm fine with removing a restriction - I believe we added it there so it shows as a drop-down inside WebUI, but as you said, this restricts users from adding a custom content type.

emedvedev:
@pixelrebel thanks!
I'm fine with it, too, especially since there's a default value in place so the expected format is clear. Merging.

PR #2964, opened 2016-10-19
Interactive pack configuration

enykeev:
Due to a multitude of requests both from inside and outside the team, I humbly present you interactive pack configuration.

[asciicast](https://asciinema.org/a/8dnn9oenfiokqwsojybbnxva7)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -26,6 +26,7 @@ prettytable
 pyinotify<=0.10,>=0.9.5
 pymongo==3.2.2
 python-dateutil
+python-editor
 python-gnupg<0.4,>=0.3.7
 python-json-logger
 python-keyczar==0.715
--- a/requirements.txt
+++ b/requirements.txt
@@ -26,7 +26,7 @@ prettytable
 pyinotify<=0.10,>=0.9.5
 pymongo==3.2.2
 python-dateutil
-python-editor
+python-editor==1.0.1
 python-gnupg<0.4,>=0.3.7
 python-json-logger
 python-keyczar==0.715
--- a/st2api/st2api/controllers/v1/pack_configs.py
+++ b/st2api/st2api/controllers/v1/pack_configs.py
@@ -13,18 +13,32 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
+import six
+
+import jsonschema
+from oslo_config import cfg
+import yaml
+
+from st2common import log as logging
 from st2common.models.api.base import jsexpose
 from st2api.controllers.resource import ResourceController
+from st2common.exceptions.apivalidation import ValueValidationException
 from st2common.services import packs as packs_service
 from st2common.models.api.pack import ConfigAPI
+from st2common.models.api.pack import ConfigUpdateRequestAPI
 from st2common.persistence.pack import Config
 from st2common.rbac.types import PermissionType
 from st2common.rbac.decorators import request_user_has_permission
 from st2common.rbac.decorators import request_user_has_resource_db_permission
 
+http_client = six.moves.http_client
+
 __all__ = [
     'PackConfigsController'
 ]
+
+LOG = logging.getLogger(__name__)
 
 
 class PackConfigsController(ResourceController):
@@ -63,3 +77,28 @@ class PackConfigsController(ResourceController):
         """
         # TODO: Make sure secret values are masked
         return self._get_one_by_pack_ref(pack_ref=pack_ref)
+
+    @request_user_has_permission(permission_type=PermissionType.PACK_CREATE)
+    @jsexpose(body_cls=ConfigUpdateRequestAPI, arg_types=[str])
+    def put(self, pack_uninstall_request, pack_ref):
+        """
+            Create a new config for the action.
+
+            Handles requests:
+                POST /configs/<pack_ref>
+        """
+
+        try:
+            config_api = ConfigAPI(pack=pack_ref, values=vars(pack_uninstall_request))
+            config_api.validate(validate_against_schema=True)
+        except jsonschema.ValidationError as e:
+            raise ValueValidationException(str(e))
+
+        config_content = yaml.safe_dump(config_api.values, default_flow_style=False)
+
+        configs_path = os.path.join(cfg.CONF.system.base_path, 'configs/')
+        config_path = os.path.join(configs_path, '%s.yaml' % config_api.pack)
+        with open(config_path, 'w') as f:
+            f.write(config_content)
+
+        return config_api
--- a/st2client/requirements.txt
+++ b/st2client/requirements.txt
@@ -1,11 +1,11 @@
 # Don't edit this file. It's generated automatically!
+argcomplete
 jsonpath-rw>=1.3.0
 prettytable
 python-dateutil
+python-editor==1.0.1
 pytz
 pyyaml<4.0,>=3.11
-requests<3.0,>=2.7.0
+requests[security]<2.12,>=2.11.1
 six==1.9.0
-argcomplete==1.4.1
-prompt_toolkit==1.0.7
-jsonschema==>=2.5.0,<2.6
+sseclient==0.0.12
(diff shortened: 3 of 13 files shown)


enykeev:
The thing still lacks tests and doesn't support some types (both primitives as `boolean` and complex ones as `array of objects`, for example), but still probably enough to have a glimpse of how it would look and feel when it's done.

Kami:
Awesome, nice work!

st2api/st2api/controllers/v1/pack_configs.py
```diff
@@ -63,3 +77,28 @@ def get_one(self, pack_ref):
...
+            config_api = ConfigAPI(pack=pack_ref, values=vars(pack_uninstall_request))
+            config_api.validate(validate_against_schema=True)
+        except jsonschema.ValidationError as e:
+            raise ValueValidationException(str(e))
+
+        config_content = yaml.safe_dump(config_api.values, default_flow_style=False)
+
+        configs_path = os.path.join(cfg.CONF.system.base_path, 'configs/')
+        config_path = os.path.join(configs_path, '%s.yaml' % config_api.pack)
+        with open(config_path, 'w') as f:
```
Kami: How will this affect file permissions and cross-server deployments?

st2api/st2api/controllers/v1/pack_configs.py
```diff
@@ -63,3 +77,28 @@ def get_one(self, pack_ref):
...
+
+        try:
+            config_api = ConfigAPI(pack=pack_ref, values=vars(pack_uninstall_request))
+            config_api.validate(validate_against_schema=True)
+        except jsonschema.ValidationError as e:
+            raise ValueValidationException(str(e))
+
+        config_content = yaml.safe_dump(config_api.values, default_flow_style=False)
+
+        configs_path = os.path.join(cfg.CONF.system.base_path, 'configs/')
```
Kami: Minor thing - I believe we do this in multiple places so we should probably and and use a utility function to retrieve a path. Similar as we do for other paths.

st2client/st2client/commands/pack.py
```diff
@@ -243,3 +249,57 @@ def __init__(self, resource, *args, **kwargs):
...
+        self.parser.add_argument('name',
+                                 help='Name of the %s(s) to configure.' %
+                                      resource.get_display_name().lower())
+
+    @resource.add_auth_token_to_kwargs_from_cli
+    def run(self, args, **kwargs):
+        schema = self.app.client.managers['ConfigSchema'].get_by_ref_or_id(args.name, **kwargs)
+
+        if not schema:
+            raise resource.ResourceNotFoundError("%s doesn't have config schema defined" %
```
Kami: "Pack doesn't have a config schema defined - consider writing and contributing one" :P

Kami:
Support for boolean should be simple right? Or am I missing something?

I'm also fine with not supporting complex types since those will always be hard to enter during interactive mode anyway...

st2api/st2api/controllers/v1/pack_configs.py
Kami: Also - I assume it's still the same as issue we have in other places where we write content to the disk from inside the API.

So probably we don't have a good solution for it (yet), but adding more things which write to disk directly just increases the scope of the problem (which I guess is fine for now as long as we tell users how to sync the changes and work on multi node systems).

Kami:
I assume API controller tests are still WIP? :)

enykeev:
> I assume API controller tests are still WIP? :)

You trying to offend me with this kind of questions aren't you? =)

> Support for boolean should be simple right?

It is, I just didn't have time to implement it.

> I'm also fine with not supporting complex types since those will always be hard to enter during interactive mode anyway...

I feel like we need to try and implement at least some of most common ones and try to discourage writing anything more challenging. Array of objects is right between the two.

In general, I'm planning to focus on getting #2842 out the door before I get back to this PR and polish it.

st2client/st2client/commands/pack.py
enykeev: I like the idea, but I don't think it suits us to whine about it in error messages.

st2api/st2api/controllers/v1/pack_configs.py
enykeev: Well, we would need to figure out permissions, but as far as HA setups go, it doesn't affect it in any way. Until we figure out the problem in general (as a part of StackStorm/discussions#204), it's up to a user to distribute configs and packs between the nodes.

st2api/st2api/controllers/v1/pack_configs.py
enykeev: See StackStorm/discussions#204 for a possible solution to multi-node.

dzimine:
This is awesome! 
and because it is, I'll be critical :) 

This needs: 
1) defaults
2) way to jump to the file-based from the middle.

I really liked the "in the file" at the end of it as the most convinient user interaction. 

3) When user "cancels" (e.g., exists without saving) - need to print a clear instructions on where to go to complete the configuration. 

4) (FOR FUTURE) ideally we need to test that configuration works (at minimal, make a connection), and have a way to mark "misconfigured" packs. By mark I mean, a field in pack model, available for API, CLI, UI.

Kami:
> 4) (FOR FUTURE) ideally we need to test that configuration works (at minimal, make a connection), and have a way to mark "misconfigured" packs. By mark I mean, a field in pack model, available for API, CLI, UI.

Looking at the code, the API does validate the syntax against the schema on save, but that's it.

Doing actual "it works" test (e.g. establish a connection or similar) would be much more complex and since it's integration specific we would need to introduce new kind of API on actions or similar for it (e.g. test_connection / test_config).

Kami:
On a related note - did we investigate how long would it take for us to write a custom wrapper around YAML parser which preserves ordering?

enykeev:
The only way you could try to solve this problem is by intercepting yaml code before it goes to parsing, substituting every object property
```
# type: object
some:
    type: 'thing'
```
with array of an object of single property
```
# type: array
# items:
#     type: object
- some:
    type: 'thing'
```
then parse the resulting code and convert this arrays back to OrderedDict's. Then hoping they would not get messed up by json serializer or a client-side parser.

It's probably possible, but damn if I ever want to see it implemented. You can try and fork pyyaml and then make it use OrderedDict there, and it would be a slightly cleaner solution...

But at the end of the day, IF someone would ask me, I'd would tell them that you should never rely on dict-like structures to have an order. Never. If we want parameters to be ordered, they should be in form of a list.

enykeev:
> This needs: 
> 1) defaults

Defaults are already there in square brackets after parameter name. The library we're using to draw prompts allow us to define initial value for the prompt, but it would then require user to manually remove it before typing out own value. Pretty poor UX if you ask me. The way we implemented it, empty (as in `''`) value gets replaced with default.

> way to jump to the file-based from the middle

From the top of my head, I can't remember any CLI tool that allows you to do something like that. We could probably use single `^C` to interrupt interactive mode and another `^C` to break out of the configuration completely, but that certainly require some stretch of imagination on a user side. Is it worth it to you?

> When user "cancels" (e.g., exists without saving) - need to print a clear instructions on where to go to complete the configuration.

For now, we assume that if you quit, you quit and you need to start over. That's a CLI tool, it is supposed to be punishing =)

requirements.txt
```diff
@@ -26,6 +26,7 @@ prettytable
 pyinotify<=0.10,>=0.9.5
 pymongo==3.2.2
 python-dateutil
+python-editor
```
Kami: This file should get auto-generated during some make step and the version should be pinned as long as we put a pinned version into fixed-requierments.txt

Kami:
Besides the missing fixed-config.txt entry, LGTM, +1

Nice work.

st2client/requirements.txt
```diff
@@ -1,11 +1,11 @@
 # Don't edit this file. It's generated automatically!
+argcomplete
 jsonpath-rw>=1.3.0
 prettytable
 python-dateutil
+python-editor==1.0.1
 pytz
 pyyaml<4.0,>=3.11
-requests<3.0,>=2.7.0
```
enykeev: That stuff is autogenerated. I'm not editing this file manually.

enykeev: And the change was done well in advance, people just forgot to push it there.

PR #2971, opened 2016-10-25
Fix pack registration

enykeev:


st2api/st2api/controllers/v1/packs.py
```diff
@@ -131,7 +131,7 @@ def post(self, pack_register_request):
         if 'policy_type' in types or 'policy' in types:
             result['policy_types'] = policies_registrar.register_policy_types(st2common)
 
-        use_pack_cache = True
+        use_pack_cache = False
```
Kami: I'm fine with this for now.

PR #2972, opened 2016-10-25
Add support for "ref" attribute to the pack metadata file (pack.yaml)

Kami:
This attribute will replace directory name which is now gone in the single repo per pack model.

Previously, we used pack directory name as a unique identifier (ref) for packs, but not that this is gone, we are going to utilize `ref` field from `pack.yaml` file for it. "name" field is human readable version of the pack name and doesn't satisfy the unique id criteria pack directory name satisfied (only a-z0-9_- characters) so we need to go with this approach.

On a related note - we need to get rid of https://github.com/StackStorm-Exchange/ci/blob/7776bd2dc07965df7e0676460b0fbc3f81f292ab/.circle/pack.yaml.schema this file and replace it with a call to register-content --register-pack script, otherwise we have same information duplication in multiple places and things will get out of sync.

## TODO

- [x] Update docs (notably upgrade notes - validation changes, new ref attribute, etc.) - https://github.com/StackStorm/st2docs/pull/304
- [x] Add ref to all the pack yaml files in the new stackstorm-exchange repos
- [x] Update affected and existing packs to pass all the validation rules
- [ ] Update CI exchange repo to utilize register-content script to validate pack.yaml instead of duplicating the logic across multiple repos


Kami:
To make the migration easier, as discussed with @emedvedev we now also use `name` attribute from pak metadata file if it doesn't contain `ref` attribute and if pack is not in a sub-directory (that's true for packs in the new pack per repo model).

st2common/st2common/bootstrap/base.py
```diff
@@ -149,13 +152,38 @@ def _register_pack_db(self, pack_name, pack_dir):
...
+        # 2hich are in sub-directories)
+        # 2. If attribute is not available, but pack name is and pack name meets the valid name
+        # criteria, we use that
+        if content.get('ref', None):
+            content['ref'] = content['ref']
+        elif re.match(PACK_NAME_WHITELIST_REGEX, pack_name):
+            content['ref'] = pack_name
+        else:
+            if re.match(PACK_NAME_WHITELIST_REGEX, content['name']):
+                content['ref'] = content['name']
```
emedvedev: Should also do `.lower()` in this path, probably, since ref would ideally be lowercase.

Kami: I would prefer to avoid additional magic though - one option would be to update the `ref` regex and only allow lowercase letters.

And this way if pack name is not lowercase already it's up to the user to add an explicit ref :)

PR #2975, opened 2016-10-26
Replace bencode parameters serialization with simple json.dumps call

Kami:
When retrieving Trigger UID we need to ensure that UID for all the triggers with the same parameters is the same.

Internally Python dictionaries aren't ordered so when calculating a UID and hash for parameters we serialized them using `bencode` since bencode ensures parameters are ordered before serialization - this ensures that UID for triggers with the same parameters is always the same.

We only support simple types for trigger parameters so we can achieve the same thing by using `json.dumps` in combination with `sort_keys=True` parameter. This way we can get rid of external dependency.

## Backward compatibility

The change is backward compatible because the `uid` attribute is assigned inside the model constructor where it's also dynamically calculated based on the model field values.

This could only cause issues if user would run different versions of StackStorm on different nodes, but doing that in long term (longer than short upgrade phase) would probably cause issues with other things as well.


Kami:
Re backward compatibility - to be on the safe side I would still prefer to test this by upgrading an existing installation and making sure things don't break in a weird way.

Kami:
I checked the code and verified that this shouldn't affect thing, but we should still keep an eye for potential issues (e.g. failing to find trigger) when a new version is out.

PR #2981, opened 2016-10-28
Pack management: branch checkouts

emedvedev:
Two changes here:
1. Versions can also point to branches now (i.e. `st2 pack install sensu=development`).
2. Checking out commits / tags that aren't tips of the branches will not detach HEAD. Instead it will put you on the right branch and point to N commits back. 

(2) is important to figure out where in the history you actually are. If I checked out a specific version, `Your branch is behind 'origin/master' by 3 commits` is a pretty useful message. `HEAD detached at c46a25c` is not a useful message.

It also helps you work with your pack more effectively, since it's pretty easy to update your pack with fast-forwarding or even start a new branch from that particular commit if you want to. Detached HEAD would require you to figure out which branch it belonged to, do a merge or a checkout to master, or make commits that are gonna be garbage-collected if you forget to create a new branch.

In other words, working with packs as repos (navigating, editing, etc.) is much easier now.

### DIFF
--- a/contrib/packs/actions/pack_mgmt/download.py
+++ b/contrib/packs/actions/pack_mgmt/download.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 # Licensed to the StackStorm, Inc ('StackStorm') under one or more
 # contributor license agreements.  See the NOTICE file distributed with
 # this work for additional information regarding copyright ownership.
@@ -22,7 +23,7 @@ import re
 import six
 import yaml
 from git.repo import Repo
-from gitdb.exc import BadName
+from gitdb.exc import BadName, BadObject
 from lockfile import LockFile
 
 from st2common.runners.base_action import Action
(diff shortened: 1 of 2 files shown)


contrib/packs/actions/pack_mgmt/download.py
```diff
@@ -51,46 +52,63 @@ def run(self, packs, abs_repo_base, verifyssl=True):
...
+                "\"%s\" is not a valid version, hash, tag, or branch in %s." % (ref, repo_url)
+            )
+
+        # We're trying to figure out which branch the ref is actually on,
+        # since there's no direct way to check for this in git-python.
+        branches = repo.git.branch('--color=never', '--all', '--contains', gitref.hexsha)
+        branches = branches.replace('*', '').split()
+        if 'master' not in branches:
+            branch = branches[0]
+            repo.git.checkout('--track', branches[0])
```
Kami: I hope git library properly escapes argument underneath :D

emedvedev: `gitpython` is quite horrible (also I don't like the absence of a plumbing command for `git branch --contains`, but yeah, this works. :)

Kami: There are also native bindings which don't shell out to git (https://github.com/libgit2/pygit2), but probably not worth swapping it out right now...

emedvedev: Yeah, there's also dulwich and gittle, definitely worth a try when we get to the next stage, but not now, you're right. :)

PR #2982, opened 2016-10-31
Rework the packs pack (more packs-related ChatOps!)

emedvedev:
1. Aliases for the packs pack:
— `pack install` for installing a pack or multiple packs (just a minor rework here);
— `pack search` to query the index;
— `pack show` to display information about an installed pack, including HEAD location in the git tree (N commits ahead/behind);
— `pack show available` to fetch information about a pack from the index.

2. Most of `packs.deploy` is now obsolete, so I'm removing it and the helper actions for now, but I would appreciate @jjm finding some time to talk it over with me. A thing or two that was possible with `packs.deploy` might not be covered by our new pack management, and if that's the case, I'm fine with keeping the deploy workflow or breaking it down into multiple actions.

3. There was a `linux` pack directory and an action. Inside the `packs` pack. Apparently, for quite some time, too (I remember @enykeev noticing it). I removed it. Sorry, not sorry.

4. Some minor cleanups (this PR started as a minor cleanup, actually, but then things just got more and more interesting).


emedvedev:
WIP! Like, really WIP. Don't test it yet, there's still some work to be done.

emedvedev:
Btw, `pack show available <pack>` shows content count if it's present in the index, but `pack show <pack>` doesn't, because we don't have helper functions for content counting yet. @Kami promised to get it done :)

jjm:
@emedvedev Yes, I think `packs.deploy` and it's associated actions can be removed. In fact GitHub pack [deployment_event](https://github.com/StackStorm/st2contrib/blob/master/packs/github/actions/workflows/deployment_event.yaml) workflow that I added earlier in October does not call `packs.deploy` and calls `packs.install` directly.

I think with the release of _Stackstorm-exchange_ the _bitbucket_ rule [post_receive_webhook.yaml](https://github.com/StackStorm/st2contrib/blob/master/packs/bitbucket/rules/post_receive_webhook.yaml)

emedvedev:
> In fact GitHub pack deployment_event workflow that I added earlier in October does not call packs.deploy and calls packs.install directly.

Once we make the move, that rule and workflow will become a bit simpler, too! 

```

      input:
        packs: [  <% $.ssh_url %>=<% $.deploy_ref %> ]
```

I'll skim through all the packs to make sure all the `packs.install` and `packs.download` references are updated when the packs are moved.

contrib/packs/actions/pack_mgmt/lookup_remote.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class LookupRemote(Action):
+    """Get detailed information about an available pack from the pack index"""
+    def run(self, pack):
+        pack_meta = get_pack_from_index(pack)
+        pack_formatted = None
+        # Try to get the original yaml file because we know
+        # how it's stored in the Exchange, otherwise fall back
+        # to transforming json.
```
armab: @emedvedev 
As I see, we already getting pack meta from index via `pack_meta = get_pack_from_index(pack)` from endpoint: https://index.stackstorm.org/v1/index.json
And then we're downloading https://index.stackstorm.org/v1/packs/github.yaml again with the same meta.

What was initial idea here? Because it seems a bit dirty with all the hardcoding, additional request, regex.

I'm thinking about removing this second `pack.yaml` part and rely only on `get_pack_from_index(pack)` meta.

emedvedev: > I'm thinking about removing this second pack.yaml part and rely only on get_pack_from_index(pack) meta.

:+1:

Problem with json->yaml is that it's not ordered, but that doesn't really matter much.

armab:
So it's fixed & Polished & Acceptance tested in Slack.
The new list of ChatOps Action Aliases is adjusted to conform with `CLI` commands for consistency.

* `!pack install` ([screenshot](https://i.imgur.com/gWBpRfW.png)) - Install pack(s)
* `!pack get` ([screenshot](https://i.imgur.com/qjBfUKq.png)) - Get info about locally installed pack
* `!pack search` ([screenshot](https://i.imgur.com/jdoVfsy.png)) - Search for pack(s) in StackStorm Exchange
* `!pack show` ([screenshot](https://i.imgur.com/NCeqHDj.png)) - Show pack info from StackStorm Exchange

The [Aliases automated tests](https://docs.stackstorm.com/development/pack_testing.html) as well as Integration Tests (e2e, robo) are in my TODO list and will be added, but It's good to merge now to see things in pre-production (packages).
> Additionally, I'll need staging-unstable packages to make sure that it works in `EL6` (outdated `git`).

emedvedev:
:heart: :heart: :heart: :heart: 

LGTM.

enykeev:
```
enykeev	[10:30] 
!pack show st2

stuffBOT	[10:30]  
@enykeev: Getting back to you about the `st2` pack:
Here's the full index entry at StackStorm Exchange:
{author: st2-dev, description: StackStorm pack management, email: info@stackstorm.com,
    name: st2, repo_url: 'https://github.com/StackStorm-Exchange/stackstorm-st2',
    version: 0.1.2}

To install the pack: `pack install st2`
If the pack is already installed, I will upgrade it to the latest version, keeping the config.
```

I'm kinda missing hubot alias before the command, though I'm not sure we can get it now in any sane way.

Also, https://github.com/StackStorm/st2/pull/3051 to prevent different kinds of output formatting for different packs.

Other than that, looks good to me.

contrib/packs/actions/install.meta.yaml
```diff
@@ -17,6 +18,10 @@
       type: "string"
       default: "actions,aliases,sensors,triggers"
       description: "Possible options are all, triggers, sensors, actions, rules, aliases."
+    upgrade:
+      type: "boolean"
+      default: false
+      description: "Upgrade the pack if already installed."
```
enykeev: Wait a second. I don't see us using it in any way.

emedvedev: I'm not sure what that flag is doing here 😛 Should just remove.

armab: Hm, but this is interesting.

In future it would be good to have pack install really idempotent.
So if pack is already installed - it won't be reinstalled again, unless some special flag like that specified. Re-downloading/re-installing by default takes time and it's in fact "upgrade" or "reinstall".

This makes sense especially for configuration management tools, when you apply playbook/cookbook, which apart of deploying `st2` installs a bunch of st2 packs on the same machine. And it's normal to run same playbook on the same machine several times.

emedvedev: This is a bigger discussion that applies to package management in general. It will be idempotent if you specify a precise version, and we already support that, but a repeated `install` is normally regarded as upgrade, we're certainly not alone here. :)

armab: Installing + Downloading over and over again (even if it reaches the same state) is not considered really "Idempotent" in configuration management.
If resources is in desired state - we shouldn't repeat the same operation again.

emedvedev: Fair. They do stay in that state though 😛 

Let's make sure we get to that issue after the release, it's pretty important.

armab:
> I'm kinda missing hubot alias before the command, though I'm not sure we can get it now in any sane way.

Yeah, @enykeev I just verified, and there is no hubot alias variable in context. I bet none could think that it might be useful to have in context.  Might be something good to adjust in future.

Kami:
Nice work everyone!

PR #2984, opened 2016-10-31
Add proper callout box for Python dependencies section

Mierdin:
Minor fix to CONTRIBUTING.rst so that a proper callout box is shown. The sphinx-style callout doesn't show up well when viewed in Github (which presumably is the primary use case for this).

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -11,10 +11,7 @@ http://docs.stackstorm.com/development/index.html
 Managing Python dependencies
 ----------------------------
 
-.. note::
-
-    ``requirements.txt`` files are generated automatically using
-    ``scripts/fixate-requirements.py`` script and should't be edited manually.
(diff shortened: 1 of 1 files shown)


lakshmi-kannan:
lgtm.

CONTRIBUTING.rst
```diff
@@ -11,10 +11,7 @@ http://docs.stackstorm.com/development/index.html
 Managing Python dependencies
 ----------------------------
 
-.. note::
-
-    ``requirements.txt`` files are generated automatically using
-    ``scripts/fixate-requirements.py`` script and should't be edited manually.
+    ``requirements.txt`` files are generated automatically using ``scripts/fixate-requirements.py`` script and should't be edited manually.
```
LindsayHill: you gonna fix the other typo in there while you're at it? /s/should't/shouldn't/.

Mierdin: Nice catch, fixed in https://github.com/StackStorm/st2/pull/2986

Issue #2923, opened 2016-09-23
execution canceled by policy does not set end_timestamp which causes notifier failure


JohnWelborn:
When a policy cancels a request due to concurrency, it leaves end_timestamp set to None which the notifier expects to be a date.

```
2016-09-23 17:44:56,767 140603018531216 ERROR notifier [-] An exception occurred while applying policy "REDACT.concurrency".
Traceback (most recent call last):
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2actions/notifier/notifier.py", line 254, in _apply_post_run_policies
    liveaction_db = driver.apply_after(liveaction_db)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2actions/policies/concurrency_by_attr.py", line 136, in apply_after
    with self.coordinator.get_lock(lock_uid):
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/tooz/locking.py", line 35, in __enter__
    self.acquire()
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/tooz/drivers/zookeeper.py", line 56, in acquire
    return _lock()
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/retrying.py", line 49, in wrapped_f
    return Retrying(*dargs, **dkw).call(f, *args, **kw)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/retrying.py", line 206, in call
    return attempt.get(self._wrap_exception)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/retrying.py", line 247, in get
    six.reraise(self.value[0], self.value[1], self.value[2])
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/retrying.py", line 200, in call
    attempt = Attempt(fn(*args, **kwargs), attempt_number, False)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/tooz/drivers/zookeeper.py", line 49, in _lock
    timeout=0):
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/kazoo/recipe/lock.py", line 170, in acquire
    six.reraise(exc_info[0], exc_info[1], exc_info[2])
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/kazoo/recipe/lock.py", line 160, in acquire
    blocking=blocking, timeout=timeout)
[...]

PR #2985, opened 2016-10-31
Add protection for a null timestamp in notifier

Mierdin:
When a policy cancels a request due to concurrency, it leaves
end_timestamp set to None which the notifier expects to be a date.

This patch adds exception handling to address this properly

Fixes: https://github.com/StackStorm/st2/issues/2923

Signed-off-by: Matt Oswalt <oswaldm@brocade.com>


Mierdin:
Please hold; working on unit test for this patch

Mierdin:
Okay, with that, this should be ready to review.

st2actions/st2actions/notifier/notifier.py
```diff
@@ -167,7 +167,15 @@ def _post_notify_subsection_triggers(self, liveaction=None, execution=None,
             payload['execution_id'] = execution_id
             payload['status'] = liveaction.status
             payload['start_timestamp'] = isotime.format(liveaction.start_timestamp)
-            payload['end_timestamp'] = isotime.format(liveaction.end_timestamp)
+
+            try:
+                payload['end_timestamp'] = isotime.format(liveaction.end_timestamp)
+            except AttributeError:
+                # This can be raised if liveaction.end_timestamp is None, which is caused
```
emedvedev: Shouldn't we just check whether `liveaction.end_timestamp` is set instead of throwing? EAFP is very Pythonic, I know, but no `end_timestamp` is expected and predictable behavior. Anyway, I'm fine with both, mostly just wondering :)

lakshmi-kannan: I am ok with catching AttributeError.

st2actions/st2actions/notifier/notifier.py
```diff
@@ -167,7 +167,15 @@ def _post_notify_subsection_triggers(self, liveaction=None, execution=None,
...
+                # when policy cancels a request due to concurrency
+                # In this case, use the start_timestamp instead
+                payload['end_timestamp'] = isotime.format(liveaction.start_timestamp)
```
lakshmi-kannan: I am not so sure this is a good default. I'd rather not include ``end_timestamp`` in the payload. I'll let @Kami also weigh in.

lakshmi-kannan:
This is a simple fix so a CHANGELOG entry is not really required but preferred. https://github.com/StackStorm/st2/blob/master/CHANGELOG.rst.

st2actions/st2actions/notifier/notifier.py
Kami: I would personally just set it to the time when the execution was canceled / this method is called (aka now). I know this is not the most accurate, but yeah, other alternative is leaving it empty which is probably the most correct since execution never started.

Mierdin: Good idea - updated in https://github.com/StackStorm/st2/pull/2985/commits/5f58b0414c0906ffb102350b09a115098d5abcd4

Mierdin:
@lakshmi-kannan No problem, updated in https://github.com/StackStorm/st2/pull/2985/commits/afc1bb64e3bfdb7e5557d898a50fea1963ea765b

lakshmi-kannan:
LGTM.

st2actions/st2actions/notifier/notifier.py
```diff
@@ -167,7 +168,15 @@ def _post_notify_subsection_triggers(self, liveaction=None, execution=None,
...
             payload['start_timestamp'] = isotime.format(liveaction.start_timestamp)
-            payload['end_timestamp'] = isotime.format(liveaction.end_timestamp)
+
+            try:
+                payload['end_timestamp'] = isotime.format(liveaction.end_timestamp)
+            except AttributeError:
+                # This can be raised if liveaction.end_timestamp is None, which is caused
+                # when policy cancels a request due to concurrency
+                # In this case, use datetime.now() instead
+                payload['end_timestamp'] = isotime.format(datetime.now())
```
lakshmi-kannan: I think this should be utcnow(). Is there a utcnow()?

PR #2987, opened 2016-11-01
Add `pack create` generators

enykeev:
[asciicast](https://asciinema.org/a/47pv75tjdoa132oi5ew2oc9yo)

### DIFF
--- a/st2client/in-requirements.txt
+++ b/st2client/in-requirements.txt
@@ -11,4 +11,3 @@ sseclient
 python-editor
 prompt-toolkit
 gitpython
-cookiecutter
--- a/st2client/in-requirements.txt
+++ b/st2client/in-requirements.txt
@@ -10,3 +10,5 @@ six
 sseclient
 python-editor
 prompt-toolkit
+gitpython
+cookiecutter
--- a/st2client/requirements.txt
+++ b/st2client/requirements.txt
@@ -1,6 +1,5 @@
 # Don't edit this file. It's generated automatically!
 argcomplete
-cookiecutter
 gitpython==0.3.2.1
 jsonpath-rw>=1.3.0
 prettytable
--- a/st2client/requirements.txt
+++ b/st2client/requirements.txt
@@ -1,5 +1,7 @@
 # Don't edit this file. It's generated automatically!
 argcomplete
+cookiecutter
+gitpython==0.3.2.1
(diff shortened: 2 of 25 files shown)


Kami:
Just a heads up - some of that functionality used to be part of st2sdk - https://github.com/stackstorm/st2sdk

We need to decide on how to proceed with st2sdk.

enykeev:
@Kami we should probably just move it there completely. I somehow entirely missed the fact that we even have st2sdk.

enykeev:
Oh, ok, so apparently st2sdk implementation is quite minimalistic so the question what to do is still open.

As for this PR, there's quite a few things I personally don't like in it and I'm hesitant to even try to fit it in this release.

For one, although cookiecutter takes some pain away, it's extremely limited and we're already using it pretty much to its limits.

I also don't like that you need to have a bunch of separate repos for templates. It's hard for me to estimate how often we would like to change this templates and that's an important part of consideration how we want to ship them.

Moreover, I personally never use content generators as they, in my experience, do more harm than good.

For one proposing the PR, I surely do have a lot of negative about it.

Moving it to a separate repo\tool might be actually a good idea. It would allow us to release it outside of st2 schedule, the moment we would actually think it's ready. We may also investigate possibility of connecting it to `st2client` in a more dynamic matter which would allow us to iterate on templates easily and more frequently.

emedvedev:
> Moving it to a separate repo\tool might be actually a good idea

I think that would be the most sensible approach. I like the pack template generator, and I would probably use it (like, say, I'd use a boilerplate for a small React page rather than set everything up myself), but on my local machine. Creating a pack template on a StackStorm server only makes sense when there's a good way to sync, edit and share that pack (in our case, that would be a git repo).

enykeev:
> I like the pack template generator, and I would probably use it, but on my local machine.

The idea was that you'll start using st2client on your machine too. You surely would not want to have them both required to write a pack, would you?

emedvedev:
> The idea was that you'll start using st2client on your machine too. You surely would not want to have them both required to write a pack, would you?

Yeah, but if there's no easy way to push this pack to a server and synchronize, then cookiecutter is still nice, but kinda goes into the "bells and whistles" category.

nmaludy:
+1

lakshmi-kannan:
Closing. If you are working on this actively, please reopen @enykeev

PR #2988, opened 2016-11-01
Change the pack schema

emedvedev:
As discussed with @Kami, we need to do something about packs meant for newer StackStorm versions than whatever the user might have. Ideally, StackStorm should be able to figure it out and ask the user to update if they want to install a newer pack: 

```
$ st2 pack install sensu

ERROR: You are running StackStorm 2.1dev, but this pack requires StackStorm 3.14. 

Consider updating or install an older version of the pack: sensu=940faea3 is 
the latest version that would work for you (requires 1.6).
```

This PR does not implement this feature, but it lays the groundwork by preventing StackStorm from throwing on unknown `pack.yaml` fields. We also add properties that are very likely to appear in the next version: `dependencies` and `system`. Even if there's no implementation for these fields right now, they are pretty self-explanatory (and there's a description in the schema, too) and helpful to the user, so it's better if 2.1 still allows to install a pack meant for whenever those features are implemented. Then users can handle versions and deps themselves or consider updating.

`dependencies` will be implemented either in the content management release (see https://github.com/stackstorm/discussions/issues/204), or earlier, but in a basic form: for instance, conflicting dependencies won't be resolved.

`system` is a way to specify system requirements. We could use it to require a minimal StackStorm version, but also for packages. Even if StackStorm can't check and install system requirements right now, at least with this field users would know where to look for them. I'll try to take a stab at it it as soon as I can, but only after everything critical is done for the current release.


emedvedev:
Discussed this schema change with @bigmstone, I think we came up with a decent structure for storing system requirements:
```
name: mypack
description: A pack that does things
[...]
system:
  stackstorm: 2.1
  packages:
    - libxml
    - pkg2
    - pkg3
  deb-packages:
    - libpcap-dev
    - happy-package
  rpm-packages:
    - libpcap-devel
    - happy_package
```

StackStorm can act upon the params inside `system` that it knows about (i.e. if `stackstorm` is specified, it would check if the version matches), and even if handling of some parameters isn't implemented yet because of an older version, it's still pretty self-explanatory for a user.

st2tests/st2tests/fixtures/packs/dummy_pack_9/pack.yaml
```diff
@@ -0,0 +1,14 @@
+---
+name : dummy_pack_9_deps
+description : dummy pack with pack and engine dependencies
+version : 0.1.0
+author : st2-dev
+email : info@stackstorm.com
+dependencies :
+  - core=0.2.0
+system :
+  stackstorm : 1.6dev
```
lakshmi-kannan: Looks like we’ll need a range and not a simple number for this. So when we actually start validating these fields, we should make sure that we don't lock ourselves on specifying an exact version. 

I also think that st2_version should be a top level field and system should just be used for OS level package deps. @Kami what are your thoughts?

emedvedev: > Looks like we’ll need a range and not a simple number for this. So when we actually start validating these fields, we should make sure that we don't lock ourselves on specifying an exact version.

The number was supposed to be a minimum version, but you're right, we could find ourselves deprecating or removing stuff, so I'm adding full range support.

emedvedev:
> I also think that st2_version should be a top level field and system should just be used for OS level package deps. @Kami what are your thoughts?

Yeah, could use @Kami's opinion on this. My thinking is `system` would collect not just packages, but various other system-level dependencies: additional packages, StackStorm version, required action runners if we decide to pack them separately, Ruby version if the pack implements a Ruby runner, etc. I can get behind making StackStorm version special, so I'm not exactly opposed, but I'd still rather put all external requirements in one place to have consistency. Otherwise we'll just find ourselves wondering if we should promote every new kind of requirement. Is the ruby version special for a ruby runner pack? Is it "top level" special or just sorta special? Etc.

So I'm fine with either, but I think `system` for requirements and `system.stackstorm` for st2 version is better because there's consistency and even some degree of tidiness, so to speak. There's not enough consistency and tidiness in the package management world as it is, let's introduce some 😛

st2common/st2common/bootstrap/base.py
```diff
@@ -177,11 +177,6 @@ def _register_pack_db(self, pack_name, pack_dir):
         pack_file_list = get_file_list(directory=pack_dir, exclude_patterns=EXCLUDE_FILE_PATTERNS)
         content['files'] = pack_file_list
 
-        # Note: If some version values are not explicitly surrounded by quotes they are recognized
-        # as numbers so we cast them to string
-        if 'version' in content:
```
Kami: Any particular reason why this was removed?

Comment mentioned why it was needed :D Or it's not needed anymore now that we use semver everywhere?

st2common/st2common/models/api/pack.py
```diff
@@ -48,67 +49,95 @@ class PackAPI(BaseAPI):
...
+            },
+            'dependencies': {
+                'type': 'array',
+                'description': 'A list of other StackStorm packs this pack depends upon. '
+                               'The same format as in "st2 pack install" is used: '
+                               '"<name or full URL>[=<version or git ref>]".',
                 'items': {'type': 'string'},
                 'default': []
+            },
+            'system': {
```
Kami: Yeah - I think this name is much better than engines. We could perhaps also call it `system_version` or `st2_version` to be more explicit and obvious what is going on.

Kami: I would also simplify it and make it a string, I think there is little need to have an object since we have only one platform - StackStorm.

In Node.JS land it's a bit different since there are multiple executions environments using different engines / JITs. Having said that, I also very rarely seen people specify anything else than nodejs there :)

st2common/st2common/bootstrap/base.py
emedvedev: Shouldn't be needed anymore. Can get it back just in case :)

st2common/st2common/models/api/pack.py
```diff
@@ -48,67 +49,95 @@ class PackAPI(BaseAPI):
...
-        ref = pack.ref
-        keywords = getattr(pack, 'keywords', [])
-        version = str(pack.version)
-        author = pack.author
-        email = pack.email
-        files = getattr(pack, 'files', [])
-
-        model = cls.model(name=name, description=description, ref=ref, keywords=keywords,
-                          version=version, author=author, email=email, files=files)
+        parameters = pack.__dict__
```
Kami: I'm personally not a fan of this __dict__ approach, too much magic and could cause us to store weird stuff.

I would prefer to explicitly assign it as we did before. As far as forward compatibility goes - I thought we will just accept them without throwing, but not actually store them in the database (which I think it's the right thing to do).

st2common/st2common/models/api/pack.py
emedvedev: > I would also simplify it and make it a string, I think there is little need to have an object since we have only one platform - StackStorm.

Yeah, if we stuck to the initial "engines" purpose, that would be "st2_version", but later (https://github.com/StackStorm/st2/pull/2988#issuecomment-257704846 and https://github.com/StackStorm/st2/pull/2988#issuecomment-257731149) there was an idea to utilize it for storing system-level requirements like packages and versions of runners, etc.

st2common/tests/integration/test_register_content_script.py
```diff
@@ -144,5 +144,5 @@ def test_register_setup_virtualenvs(self):
         # All packs
         cmd = BASE_CMD_ARGS + ['--register-setup-virtualenvs', '--register-no-fail-on-failure']
         exit_code, stdout, stderr = run_command(cmd=cmd)
-        self.assertTrue('Setup virtualenv for 8 pack(s)' in stderr)
+        self.assertTrue('Setup virtualenv for 10 pack(s)' in stderr)
```
Kami: Yeah, this is a bit error prone, for got to update this in my PR so we don't need to update it every time new pack fixtures are added (I did do this in one place, but forgot some others :)).

Will push a commit to this branch which changes that.

emedvedev: Haha. I thought it's written like this on purpose, to manually check the pack number or something. 😛

Kami:
> I also think that st2_version should be a top level field and system should just be used for OS level package deps. @Kami what are your thoughts?

Yeah, I agree :)

As mentioned in my comment above, I would also prefer `st2_version` to be a simple string since there is not much need for anything else.

Kami:
I added some comments - some things which still need to be decided and changes, but overall nice work :+1: 

I also just checked out this branch and will make the version changes  we discussed yesterday directly in this branch so there won't be any merge conflicts :)

st2common/st2common/bootstrap/base.py
Kami: Yeah, you can actually just leave it as is - I will make this change in the versioning change and move it to a better place anyway (API model) :)

armab:
```
  deb-packages:
    - libpcap-dev
    - happy-package
  rpm-packages:
    - libpcap-devel
    - happy_package
```

Even if that part is not implemented, just a thought: I think it will be a bit dangerous to specify/install OS-level dependencies, because it could be harder than we initially think.

Different packages can provide same libraries(package1|package2), different flavors of distros can have different package naming, dependency could be installed/configured from sources, add here OS package versioning requirements and so on underwater rocks.

`pip` doesn't install OS-level dependencies, why should we? Maybe it could be a bit of a nightmare, I don't know. But it's good to think about the change like that carefully.

Kami:
@emedvedev I merged my PR into this one so only thing we need to decide is version normalization (cc @m4dcoder, @lakshmi-kannan).

Besides that, we should be good to go.

Oh, and of course we also need some docs + upgrade notes on it (I can tackle upgrade notes once it's merged).

Kami:
@armab 

> Even if that part is not implemented, just a thought: I think it will be a bit dangerous to specify/install OS-level dependencies, because it could be much harder than we initially think.

I definitely think StackStorm should *not* be be touching or handling OS-level dependencies. This info is more just like a metadata for packaging (and perhaps package maintainers can use it to auto-generate files based on this metadata).

emedvedev:
Yeah, what that guy said. :)

armab:
@emedvedev @Kami ^^ Huh, I thought you smoked something, but looks like I've smoked something wrong :smile: 

Just to make sure the naming is not confusing, so users won't think that os-level packages will be really installed. Something along _recommends_ keyword.

emedvedev:
@Kami I've removed the .0 stuff as I think it should be a separate discussion in another PR, and the version validation should probably be done during the installation for a few reasons, so also removed the check (but not utils). So normalization can be returned later, I just didn't want to hold off this PR until everyone has time for discussions.
Merging.

PR #2989, opened 2016-11-01
Fix runner loader

m4dcoder:
The frequent reloading of runner modules at runtime leads to random errors complaining the self instance of super(RunnerType, self).pre_run(...) is not instance type of the runner type. This PR address this issue by loading the runner modules only once and storing a copy of the module for reuse.

### DIFF
--- a/st2common/st2common/util/loader.py
+++ b/st2common/st2common/util/loader.py
@@ -176,10 +176,11 @@ def register_runner(module_name):
         "%s/runners/%s/%s.py" % (base_path, module_name, module_name)
     )
 
-    LOG.debug('Loading runner from: %s', module_path)
-
     if module_name not in RUNNER_MODULES:
+        LOG.info('Loading runner module from "%s".', module_path)
         RUNNER_MODULES[module_name] = imp.load_source(module_name, module_path)
+    else:
+        LOG.info('Reusing runner module "%s" from cache.', module_path)
 
     return RUNNER_MODULES[module_name]
 
@@ -190,10 +191,11 @@ def register_query_module(module_name):
         "%s/runners/%s/query/%s.py" % (base_path, module_name, module_name)
     )
 
-    LOG.debug('Loading query module from: %s', module_path)
-
     if module_name not in QUERIER_MODULES:
+        LOG.info('Loading query module from "%s".', module_path)
         QUERIER_MODULES[module_name] = imp.load_source(module_name, module_path)
+    else:
+        LOG.info('Reusing query module "%s" from cache.', module_path)
 
     return QUERIER_MODULES[module_name]
 
--- a/st2common/st2common/util/loader.py
+++ b/st2common/st2common/util/loader.py
@@ -32,6 +32,8 @@ __all__ = [
 
 LOG = logging.getLogger(__name__)
 PYTHON_EXTENSIONS = ('.py')
+RUNNER_MODULES = {}
+QUERIER_MODULES = {}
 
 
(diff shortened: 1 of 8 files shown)


.gitignore
```diff
@@ -49,3 +49,4 @@ nosetests.xml
 
 # st2 specific
 contrib/runners/*/dist_utils.py
+st2common/tests/fixturesc
```
lakshmi-kannan: Typo? Should be /fixtures?

m4dcoder: It is not typo.  This is actually generated for whatever reason...

Kami:
That's really weird - I could never reproduce it, but it could be due to local Python module cache or similar.

One thing which I just reminded we could and probably should do is call `.reload()` on the module after we load it.

/cc @bigmstone

st2common/st2common/util/loader.py
```diff
@@ -176,9 +178,10 @@ def register_runner(module_name):
 
     LOG.debug('Loading runner from: %s', module_path)
 
-    module = imp.load_source(module_name, module_path)
+    if module_name not in RUNNER_MODULES:
```
Kami: Yeah, I'm personally fine with caching an existing module. It's also worth noting that this requires process restart (action runner) in case module changes and user wants those changes to take an effect, but imo that's fine and just something we need to document.

One thing we could probably to avoid the need of restart (well in most cases) or avoid the need to cache is to use `.reload()` (or delete module from internal module cache), but that's not Erlang and it's not really that robust and good idea to use it in most cases so process restart is more robust.

Anyway, just something to document in pluggable runner section of documentation :P

/cc @bigmstone

st2common/st2common/util/loader.py
```diff
@@ -189,9 +192,10 @@ def register_query_module(module_name):
 
     LOG.debug('Loading query module from: %s', module_path)
 
-    module = imp.load_source(module_name, module_path)
+    if module_name not in QUERIER_MODULES:
+        QUERIER_MODULES[module_name] = imp.load_source(module_name, module_path)
```
Kami: Let's please add an `else` and LOG.info / debug statement which says that module x for runner y is already cached - this way user won't be that surprised if they made a change without restarting action runner and wondering why the change is not taking effect.

Kami:
I added some comments - nice catch and good work overall :+1: 

There are just some things we need to document and all be aware of since it does change the behavior a bit (restart vs no restart, etc.).

jbrownsc:
I'm seeing an issue that seems very similar to this one. This stackstorm env hasn't had any executions in a while.  The first run of of a core.local action raised an exception around LocalShellRunner, and the first python action (st2.kv.set) raised an exception around PythonRunner.    I suspect something is getting garbage collected or some lower level construct that I don't understand is cleaning something up.  If I re-run the execution, it works just fine.

```super(type, obj): obj must be an instance or subtype of type
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2actions/container/base.py", line 110, in _do_run
    runner.pre_run()
  File "/opt/stackstorm/runners/local_runner/local_runner.py", line 84, in pre_run
    super(LocalShellRunner, self).pre_run()

super(type, obj): obj must be an instance or subtype of type
File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2actions/container/base.py", line 110, in _do_run
runner.pre_run()
File "/opt/stackstorm/runners/python_runner/python_runner.py", line 86, in pre_run
super(PythonRunner, self).pre_run()```

PR #2990, opened 2016-11-02
Various pack schema changes

Kami:
This pull request works on top of https://github.com/StackStorm/st2/pull/2988 and makes some changes @emedvedev and I discussed.

- [x] Support for old style version specifiers which don't match semver format (e.g. 1.0). We did update all the affects packs in st2contrib, but this was done to make migration of other non-st2contrib packs easier. If such version is encountered, we simply append ".0" to the end to make it a valid semver specifier (e.g. 1.0 -> 1.0.0, 0.2 -> 0.2.0, etc.)
- [x] Throw more user-friendly exception if pack version defined in schema doesn't contain valid semver specifier
- [x] Make `stackstorm_version` attribute in pack.yaml top-level
- [x] Throw when registering a pack if the pack `stackstorm_version` property is not satisfied by currently running StackStorm version
- [x] Update changelog

### DIFF
--- a/st2common/st2common/models/api/pack.py
+++ b/st2common/st2common/models/api/pack.py
@@ -152,6 +152,23 @@ class PackAPI(BaseAPI):
 
         super(PackAPI, self).__init__(**values)
 
+    def validate(self):
+        # We wrap default validate() implementation and throw a more user-friendly exception in
+        # case pack version doesn't follow a valid semver format
+        try:
+            super(PackAPI, self).validate()
+        except jsonschema.ValidationError as e:
+            msg = str(e)
+
+            if "Failed validating 'pattern' in schema['properties']['version']" in msg:
+                new_msg = ('Pack version "%s" doesn\'t follow a valid semver format. Valid '
+                           'versions and formats include: 0.1.0, 0.2.1, 1.1.0, etc.' %
+                           (self.version))
+                new_msg += '\n' + msg
+                raise jsonschema.ValidationError(new_msg)
+
+            raise e
+
     @classmethod
     def to_model(cls, pack):
         ref = pack.ref
--- a/st2common/st2common/models/api/pack.py
+++ b/st2common/st2common/models/api/pack.py
@@ -164,7 +164,7 @@ class PackAPI(BaseAPI):
                 new_msg = ('Pack version "%s" doesn\'t follow a valid semver format. Valid '
                            'versions and formats include: 0.1.0, 0.2.1, 1.1.0, etc.' %
                            (self.version))
-                new_msg += '\n' + msg
+                new_msg += '\n\n' + msg
                 raise jsonschema.ValidationError(new_msg)
 
             raise e
--- a/st2common/st2common/models/api/pack.py
+++ b/st2common/st2common/models/api/pack.py
@@ -92,6 +92,12 @@ class PackAPI(BaseAPI):
                 'pattern': PACK_VERSION_REGEX,
                 'required': True
             },
+            'stackstorm_version': {
+                'type': 'string',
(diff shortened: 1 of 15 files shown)


st2common/st2common/models/api/pack.py
```diff
@@ -117,27 +126,46 @@ class PackAPI(BaseAPI):
...
+        version = str(pack.version)
+
+        # Note: If some version values are not explicitly surrounded by quotes they are recognized
+        # as numbers so we cast them to string
+        if getattr(pack, 'version', None):
+            version = str(pack.version)
+
+        # Special case for old version which didn't follow semver format (e.g. 0.1, 1.0, etc.)
+        # In case the version doesn't match that format, we simply append ".0" to the end (e.g.
+        # 0.1 -> 0.1.0, 1.0, -> 1.0.0, etc.)
```
emedvedev: I'm still against this change. There should be validation in place, and incorrect version should be an error. Otherwise nobody will start using correct versioning, because "I can't be bothered with this" or "I'm too busy" or "it's working fine now" or whatever else people come up with, and whenever we deprecate this behavior, people will complain. Since we're making a few more breaking changes in 2.1, the best time to introduce string versioning is now: as a user you'll probably have to restructure your packs anyway, might as well spend 15 more seconds to append ".0" to your pack version.

emedvedev:
Oh no, @Kami reduced test coverage by 0.26%!

st2common/st2common/models/api/pack.py
Kami: Re validation - we still perform validation after the version is normalized.

As far as the actual normalization goes - I'm usually for being explicit and require user to specify a correct version, but it sounded that @lakshmi-kannan and @m4dcoder were against it. I do see some value in this approach since it makes migration a bit easier.

emedvedev: > I do see some value in this approach since it makes migration a bit easier.
My point is, there will be no "migration," people will just stick to the old versioning for as long as we allow it.

Kami: @emedvedev Well, that's debatable of course :D

I personally agree with that, but let's wait on @lakshmi-kannan and @m4dcoder before we decide how to proceed.

Kami:
Here is an example of a new more user-friendly error message (open to feedback / improvements of course):

```bash
Pack version "0.1.2.3.4" doesn't follow a valid semver format. Valid versions and formats include: 0.1.0, 0.2.1, 1.1.0, etc.
'0.1.2.3.4' does not match '^(?:0|[1-9]\\d*)\\.(?:0|[1-9]\\d*)\\.(?:0|[1-9]\\d*)(?:-[\\da-z\\-]+(?:\\.[\\da-z\\-]+)*)?(?:\\+[\\da-z\\-]+(?:\\.[\\da-z\\-]+)*)?$'

Failed validating 'pattern' in schema['properties']['version']:
    {'description': 'Pack version. Must follow the semver format (for instance, "0.1.0").',
     'pattern': '^(?:0|[1-9]\\d*)\\.(?:0|[1-9]\\d*)\\.(?:0|[1-9]\\d*)(?:-[\\da-z\\-]+(?:\\.[\\da-z\\-]+)*)?(?:\\+[\\da-z\\-]+(?:\\.[\\da-z\\-]+)*)?$',
     'required': True,
     'type': 'string'}

On instance['version']:
    '0.1.2.3.4'"
```

Kami:
> Oh no, @Kami reduced test coverage by 0.26%!

All lies! :P

Kami:
Finished the PR.

I will merge it into other PR to move forward and prevent conflicts and then we can decide how to proceed with the normalized version stuff there.

PR #2991, opened 2016-11-02
Improve descriptions in pack CLI

emedvedev:
```
usage: st2 pack [-h] {list,get,show,search,install,remove,register,config} ...

A group of related integration resources: actions, rules, and sensors.

positional arguments:
  {list,get,show,search,install,remove,register,config}
                        List of commands for managing packs.
    list                Get the list of packs.
    get                 Get information about an installed pack.
    show                Get information about an available pack from the
                        index.
    search              Search the index for a pack with any attribute
                        matching the query.
    install             Install new packs.
    remove              Remove packs.
    register            Register a pack: sync all file changes with DB.
    config              Configure a pack based on config schema.

optional arguments:
```

The changes we've been discussing with @humblearner, @Mierdin and @lakshmi-kannan: added `st2 pack get` and updated descriptions.

@Mierdin If you want to make the descriptions better or change something else in the packs command list, let's discuss the changes here, or you can just commit into this branch. 

I'll let you run point on this: I won't merge anything before you give an explicit approval, but once you do, we're assuming `st2 pack` command names and descriptions are final.

### DIFF
--- a/st2client/st2client/commands/pack.py
+++ b/st2client/st2client/commands/pack.py
@@ -135,8 +135,8 @@ class PackListCommand(resource.ResourceListCommand):
 
 class PackGetCommand(resource.ResourceGetCommand):
     pk_argument_name = 'ref'
-    display_attributes = ['all']
-    # attribute_display_order = [] ResourceCommand
+    display_attributes = ['name', 'version', 'author', 'email', 'keywords', 'description']
+    attribute_display_order = ['name', 'version', 'author', 'email', 'keywords', 'description']
     help_string = 'Get information about an installed pack.'
 
 
--- a/st2client/st2client/commands/pack.py
+++ b/st2client/st2client/commands/pack.py
@@ -23,7 +23,6 @@ from st2client.models import Pack
 from st2client.models import LiveAction
 from st2client.commands import resource
 from st2client.commands.action import ActionRunCommandMixin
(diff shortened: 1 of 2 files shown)


lakshmi-kannan:
@emedvedev Can I see some sample executions of st2 pack get calls in the description, please?

Kami:
LGTM :+1: 

Maybe also not such a bad idea to add some st2client CLI tests with fixtures (this will also satisfy @lakshmi-kannan sample output question :P)l

armab:
`st2 pack get` looks kind of excessive.
Also `get` is pretty vague keyword which can bring confusion (eg. download).

Probably `st2 pack show` is enough to show info about both local and remote packs. Maybe it would be nice to show additional flag that pack is already installed on fs or not.

emedvedev:
> Probably st2 pack show is enough to show info about both local and remote packs. Maybe it would be just nice to show additional flag for st2 pack show that pack is already installed on FS or not.

Problem is, if the pack is both installed locally and available in the index, it's going to be a lot of unnecessary information, since the user probably was looking for one or the other, not both at the same time. `get` does bring confusion, so does `show`, but at least `show` is an apt convention. If you have better suggestions instead of `get`, that'd be great.

lakshmi-kannan:
Hmm. I thought st2 pack get is already merged. Can you close this PR if it's irrelevant?

PR #2993, opened 2016-11-03
Simplify command override decisions

Mierdin:
Just cutting down on some redundant code with a simpler structure for overriding commands


emedvedev:
LGTM.

Kami:
Nice :+1: 

There are definitely quite many other places in st2client which need similar (refactoring) love :)

PR #2994, opened 2016-11-03
Adding timeout Datastore Client object based on Token TTL

### LINKED ISSUE #2983: Sensor fails to retrieve kv object after some time.
st2 2.0.1
mongo 3.2.10

I have a polling sensor that queries an ES database and when a datapoint crosses a threshold, it sends an alert.  Once alerted, it stores a key with a 24-hour ttl so that the sensor knows to skip alerts for the rest of the day.  I also have a mechanism for silencing alerts with the same approach.  This used to work well before I upgraded st2, but now after a few days of running st2 the sensor keeps alerting even though there’s an `alerted` key in the keystore.  When I restart st2 everything goes back to normal for a few days then the problem comes back.   

I can probably get permission to publish my entire pack, but here's the gist of the dispatch_trigger logic in my sensor:

https://gist.github.com/pixelrebel/240ca95c480d1377175b3667de0411cb

And here's the gist of my action chain that sets the `alerted` key.

https://gist.github.com/pixelrebel/6def658dfd2ee43fe71ce7735db76b29

bigmstone:
Fixes #2983

- [x] Test

### DIFF
--- a/st2common/st2common/services/datastore.py
+++ b/st2common/st2common/services/datastore.py
@@ -94,7 +94,11 @@ class DatastoreService(object):
             params = {'decrypt': str(decrypt).lower(), 'scope': scope}
             kvp = client.keys.get_by_id(id=name, params=params)
         except Exception as e:
-            self._logger.exception('Exception retrieving value from datastore (name=%s): %s', name, e)
+            self._logger.exception(
+                'Exception retrieving value from datastore (name=%s): %s',
+                name,
+                e
+            )
             return None
 
         if kvp:
@@ -190,7 +194,11 @@ class DatastoreService(object):
             params = {'scope': scope}
             client.keys.delete(instance=instance, params=params)
         except Exception as e:
-            self._logger.exception('Exception deleting value from datastore (name=%s): %s', name, e)
+            self._logger.exception(
+                'Exception deleting value from datastore (name=%s): %s',
+                name,
+                e
+            )
             return False
 
         return True
--- a/st2common/st2common/services/datastore.py
+++ b/st2common/st2common/services/datastore.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from datetime import datetime, timedelta
 from st2client.client import Client
 from st2client.models import KeyValuePair
 from st2common.services.access import create_token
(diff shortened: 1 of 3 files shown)


Kami:
Nice - good catch!

And yeah, tests for this would be great :)

st2common/st2common/services/datastore.py
```diff
@@ -194,8 +197,12 @@ def _get_api_client(self):
         """
         Retrieve API client instance.
         """
-        if not self._client:
+        token_expire = self._token_expire <= datetime.now()
+
+        if not self._client or token_expire:
```
lakshmi-kannan: Nice catch!

bigmstone:
@Kami added test when you get a sec.

st2common/st2common/services/datastore.py
```diff
@@ -90,7 +92,8 @@ def get_value(self, name, local=True, scope=SYSTEM_SCOPE, decrypt=False):
         try:
             params = {'decrypt': str(decrypt).lower(), 'scope': scope}
             kvp = client.keys.get_by_id(id=name, params=params)
-        except Exception:
+        except Exception, e:
```
Kami: You can use Python >= 2.7 notation aka `except Exception as e` since we use it everywhere else.

st2common/st2common/services/datastore.py
```diff
@@ -185,7 +188,8 @@ def delete_value(self, name, local=True, scope=SYSTEM_SCOPE):
         try:
             params = {'scope': scope}
             client.keys.delete(instance=instance, params=params)
-        except Exception:
+        except Exception, e:
+            self._logger.warning('Exception deleting value from datastore (name=%s): %s', name, e)
```
Kami: Perhaps a good idea to use log.exception so the actual exception / stacktrace is also logged.

Kami:
Added some comments - besides that LGTM :+1: 

Please also update changelog before merging :)

pixelrebel:
Thanks @bigmstone and the rest for moving so quickly on this!  I still can't believe you found the bug so fast.  I tell all my colleagues that st2 is the best OSS project I've worked with.  This dev team is the best!  Cheers!

bigmstone:
@pixelrebel Thanks! And thanks for reporting the issue, too. Your use case for it I would assume to be fairly common. So good to have it fixed.

PR #2995, opened 2016-11-04
Updated single quotes to escaped doubled quotes for loadavg workflow

vishnu81:
Updated single quotes to escaped doubled quotes to prevent bash syntax error for diag_loadavg workflow.


Kami:
Good catch - thanks :+1:
 
For consistency with the second part of the command we could also just use single quotes, but I'm also fine with escaping it.

PR #3003, opened 2016-11-08
Change timeout to sane default for long-running action

bigmstone:


lakshmi-kannan:
OMG. Ship this already.

Kami:
:+1:

Kami:
I'm still thinking how we could improve that - we have the same constant repeated on two places now which means it's just a matter of time when it will get out of sync again :/

PR #3006, opened 2016-11-08
Fix old git syntax compatibility

armab:
> Fixes bug described in StackStorm/st2-packages#363

`--all` flag is not available for older `git` versions shipped with `EL6`, so use `-a` instead.

* git version 1.7.1 (`CentOS6`):
```sh
    -a                    list both remote-tracking and local branches
```

* version 2.7.4:
```sh
      -a, --all
           List both remote-tracking branches and local branches.
```

### DIFF
--- a/contrib/packs/actions/pack_mgmt/download.py
+++ b/contrib/packs/actions/pack_mgmt/download.py
@@ -116,7 +116,7 @@ class DownloadGitRepoAction(Action):
 
         # We're trying to figure out which branch the ref is actually on,
         # since there's no direct way to check for this in git-python.
-        branches = repo.git.branch('--color=never', '--all', '--contains', gitref.hexsha)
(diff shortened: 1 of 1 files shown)


Kami:
LGTM - good catch!

On a related note - it would be great if we could add a check for the minimum required version to those actions and throw a more user-friendly error message if one is not available.

In addition to that, we should also update installer if possible and make sure we install the latest version.

armab:
For `EL6` it's less likely asking users to add additional repository just to install latest `git` version.
Ideally if we make it working with default software that comes with distro.

If that's not possible - then yes, `st2-packages` deb/rpm requirements, like `git >= 2.0.0` + [Readme changes](https://docs.stackstorm.com/install/rhel6.html).


Let's merge and see if it will error in any other place.

Kami:
@armab I agree.

Nevertheless, I think it would still be a good idea to add a check for a minimum working version in case user has older version running on system for some reason.

PR #3007, opened 2016-11-08
Use utility function for retrieving pack metadata

Kami:
Just a small thing so we don't need to duplicate the same logic in multiple places.


emedvedev:
Ah, nice. +1.

PR #3009, opened 2016-11-09
Various pack related changes

Kami:
1. Add new `contributors` field to the pack model and metadata file. This field follows the format of the same named field in the Node.js / npm package.json file. The value is optional and can contain a list of people who have contributed to the pack.

Before that, we only had an author field and a lot of pack these days have more than one contributor, but we had no good way to indicate that and give credit where credit is due.

2. Update `force` parameter in the `packs.download` and `packs.install` action to also remove the lock file if one exists. This allows user to force install a pack and clean-up a stuck lock without needing to SSH into the server in case for some reason lock wasn't correctly cleaned up / released.
3. Allow user to pass `--force` flag to `st2 pack install` command.

### DIFF
--- a/contrib/packs/actions/pack_mgmt/download.py
+++ b/contrib/packs/actions/pack_mgmt/download.py
@@ -55,8 +55,19 @@ class DownloadGitRepoAction(Action):
             pack_url, pack_version = self._get_repo_url(pack)
 
             temp_dir = hashlib.md5(pack_url).hexdigest()
+            lock_file = LockFile(temp_dir)
+            lock_file_path = lock_file.lock_file
 
-            with LockFile('/tmp/%s' % (temp_dir)):
+            if force:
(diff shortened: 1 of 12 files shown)


contrib/packs/actions/pack_mgmt/download.py
```diff
@@ -55,8 +55,19 @@ def run(self, packs, abs_repo_base, verifyssl=True, force=False):
             pack_url, pack_version = self._get_repo_url(pack)
 
             temp_dir = hashlib.md5(pack_url).hexdigest()
+            lock_file = LockFile(temp_dir)
```
armab: I think It will try to create `/12312.lock` in system root dir,
while before it was `/tmp/12323.lock`

Kami: @armab Yeah, good catch - I agree. I was actually playing with it when testing it and forget to change it back.

PR #3008, opened 2016-11-08
Support ancient git versions in packs.download

emedvedev:
Expanded all the ugly git shortcuts, this should fix RHEL6.

PR #3010, opened 2016-11-09
Fix ancient git EL6 bug with newer gitpython

armab:
After @emedvedev fixing https://github.com/StackStorm/st2/pull/3008,
there was another one with `gitpython` library infinite loop (starting from [this line](https://github.com/StackStorm/st2/blob/ca19e2045f941196e3817eb19f5ddf15c4f6098f/contrib/packs/actions/pack_mgmt/download.py#L100)) which blocked the builds for `EL6` with timeout during the pack download.

Luckily, spending entire day with @enykeev debugging it, simple `gitpython` library upgrade worked as a last minute fix.

Let's see If this finishes [EL6 epopeya](https://github.com/StackStorm/st2-packages/issues/363).

### DIFF
--- a/fixed-requirements.txt
+++ b/fixed-requirements.txt
@@ -12,7 +12,7 @@ six==1.9.0
 pyyaml>=3.11,<4.0
 requests[security]>=2.11.1,<2.12
 apscheduler>=3.0.5,<3.1
-gitpython>=0.3.7,<1.0
+gitpython==1.0.2
 jsonschema>=2.5.0,<2.6
 mongoengine==0.10.6
 pymongo==3.2.2
--- a/fixed-requirements.txt
+++ b/fixed-requirements.txt
@@ -12,7 +12,7 @@ six==1.9.0
 pyyaml>=3.11,<4.0
 requests[security]>=2.11.1,<2.12
 apscheduler>=3.0.5,<3.1
-gitpython==0.3.2.1
+gitpython>=0.3.7,<1.0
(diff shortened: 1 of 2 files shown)


fixed-requirements.txt
```diff
@@ -12,7 +12,7 @@ six==1.9.0
 pyyaml>=3.11,<4.0
 requests[security]>=2.11.1,<2.12
 apscheduler>=3.0.5,<3.1
-gitpython==0.3.2.1
+gitpython>=0.3.7,<1.0
```
Kami: Can we please pin it to a more specific version?

Kami:
Besides the comment thanks and a good catch!

fixed-requirements.txt
armab: I just saw we follow that thingy everywhere.

Will `gitpython>=1.0.0,<1.1` work better? It's a bit more strict. Something in between the breaking change and bug-fix patch.

Kami:
:+1:

fixed-requirements.txt
```diff
@@ -12,7 +12,7 @@ six==1.9.0
 pyyaml>=3.11,<4.0
 requests[security]>=2.11.1,<2.12
 apscheduler>=3.0.5,<3.1
-gitpython==0.3.2.1
+gitpython==1.0.2
```
lakshmi-kannan: Woo!

PR #3012, opened 2016-11-10
Support for basic and digest auth in core.http action

warrenvw:
TODO:
  - Update documentation
  - Add tests


contrib/runners/http_runner/http_runner.py
```diff
@@ -191,6 +200,12 @@ def run(self):
             else:
                 data = self.body
 
+            # TODO: What should be done if username is specified
```
lakshmi-kannan: I don't see how this would happen reading the code.

contrib/core/actions/http.yaml
```diff
@@ -38,8 +38,14 @@ parameters:
...
     description: Query params to be used with the HTTP request.
     type: object
+  password:
+    description: Password for username.
+    type: string
   timeout:
     default: 60
     description: Timeout for the HTTP request.
     type: number
+  username:
```
lakshmi-kannan: You don't need to repeat these parameters here. Actions inherit runner's parameters.

warrenvw:
This PR is associated with https://github.com/StackStorm/st2docs/pull/331

Issue #2946, opened 2016-10-13
DELETE /v1/executions/57fee1b6c506f479b9c58e97 returns 200 even if already canceled


dbavery:
I found that issuing a cancel execution rest call to an execution which is already canceled returns a 200 response code ("OK") even though it also sends back a faultstring. 

In this case the fault string make sense - you attempted to do something but we are not going to do it since we are already canceled. 

However the return code of 200 implies that we attempted to issue the cancel and that it worked. 

I think a response code of 409 would be more appropriate if we determine that we can even attempt what is being requested.

```
409 Conflict
Indicates that the request could not be processed because of conflict in the request, such as an edit conflict between multiple simultaneous updates.
```

The 409 in this case would imply that we took no action because we were not in a valid state for the given request.

If folks feel that it should really stay in the 2xx, then 202 might be more appropriate to imply that we accept the request and it may or may not be appropriate. Even then, if you know enough to send back a failure case as part of the request, then I would expect a "failure" return code.

Here is the curl output illustrating.

```
~>curl -X DELETE -H  'Connection: keep-alive' -H  'Accept-Encoding: gzip, deflate' -H  'Accept: */*' -H  'User-Agent: python-requests/2.11.1' -H  'X-Auth-Token: c64c7c82be38408f9dc406ceb2640d77' -H  'Content-Length: 0' http://127.0.0.1:9101/v1/executions/57fee1b6c506f479b9c58e97
{
    "faultstring": "Action is already in \"canceled\" state."
}~>
~>
~>curl -I -X DELETE -H  'Connection: keep-alive' -H  'Accept-Encoding: gzip, deflate' -H  'Accept: */*' -H  'User-Agent: python-requests/2.11.1' -H  'X-Auth-Token: c64c7c82be38408f9dc406ceb2640d77' -H  'Content-Length: 0' http://127.0.0.1:9101/v1/executions/57fee1b6c506f479b9c58e97
HTTP/1.1 200 OK
Server: gunicorn/19.6.0
Date: Thu, 13 Oct 2016 01:27:58 GMT
Connection: keep-alive
Content-Type: application/json
Content-Length: 65
Access-Control-Allow-Origin: http://127.0.0.1:3000
Access-Control-Allow-Methods: GET,POST,PUT,DELETE,OPTIONS
Access-Control-Allow-Headers: Content-Type,Authorization,X-Auth-Token,St2-Api-Key,X-Request-ID
[...]

PR #3017, opened 2016-11-10
Return entire execution object if execution is already canceled

Mierdin:
This commit makes changes to the st2api so that the entire execution
object is returned to API calls that attempt to cancel execution
objects that are already cancelled, instead of returning a simple
error message.

Previously, an API call to this API would result in the following error message:

```
~>curl -X DELETE -H  'Connection: keep-alive' -H  'Accept-Encoding: gzip, deflate' -H  'Accept: */*' -H  'User-Agent: python-requests/2.11.1' -H  'X-Auth-Token: c64c7c82be38408f9dc406ceb2640d77' -H  'Content-Length: 0' http://127.0.0.1:9101/v1/executions/57fee1b6c506f479b9c58e97
{
    "faultstring": "Action is already in \"canceled\" state."
}
```

Now, the API will return a full execution object, as shown [here](https://gist.github.com/Mierdin/cf97f23ddd9ada1d4965b089b6561d24).

This fixes https://github.com/StackStorm/st2/issues/2946

Signed-off-by: Matt Oswalt <oswaldm@brocade.com>

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -60,6 +60,8 @@ In development
   install workflow. This step hasn't been needed for a while now because sensor container
   dynamically reads a list of available sensors from the database and starts the sub processes.
   (improvement)
+* Change st2api so that a full execution object is returned instead of an error message, when an
+  API client requests cancellation of an execution that is already canceled
 
 2.0.1 - September 30, 2016
 --------------------------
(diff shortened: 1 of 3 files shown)


Mierdin:
Please do not merge yet, currently looking at unit testing for this change.

Kami:
LGTM

Mierdin:
@Kami Okay this should be ready for a final check. I expected to also have to make a change to st2client to reflect this change, but my testing seems to work just fine right now

lakshmi-kannan:
LGTM

PR #3018, opened 2016-11-11
Adding complex type Jinja rending

bigmstone:
When you have a complex type (list, object) that includes a jinja template it will cause a non-terminating loop and fail after too many iterations. This PR fixes this issue and allows complex types to be properly rendered.

- [x] Tests
- [x] Changelog


Kami:
So far LGTM, but that's one of the PR's which would be easier to parse if there were some tests up front :P

st2common/tests/unit/test_param_utils.py
```diff
@@ -362,15 +362,17 @@ def test_get_finalized_params_with_list(self):
             'a2': 'Test',
             'a3': 'Test2',
             'a4': '{{a1}}',
-            'a5': ['{{a2}}', '{{a3}}']
+            'a5': ['{{a2}}', '{{a3}}'],
+            'a6': [['{{r1}}', '{{a2}}'], ['{{a3}}', '{{a1}}']]
```
lakshmi-kannan: I honestly think there is no way to sanely solve this in Jinja for non-homogenous types of items in the array. https://spacetelescope.github.io/understanding-json-schema/reference/array.html#list-validation
Though JSON schema allows heterogenous types inside the array, Jinja is the main problem here. The way we do Jinja rendering is we convert types to string and we convert strings back to appropriate types back based on user letting us know what type the variable is. This information is lost for list of heterogenous types where user doesn't specify what each item is like. We simply cannot solve this. @Kami Wdyt?

bigmstone: It might be possible to capture the type of the object being rendered in to the template then cast to that, but that would only be relevant for single variable templates. We obviously can't cast `My variable is: 1` as an int. But if the template was `{{my_var}}` we can. I can determine this from a simple regex but we need to decide if it's worth it. I lean towards yes but I'm willing to get vetoed.

bigmstone: I would also note here that my reason for fixing this (expect runner) just needs this to work for strings. I'd just like this to be as clean as possible for future uses.

bigmstone:
We're not going to support this. The available solutions are sub-optimal.

bigmstone:
@m4dcoder @lakshmi-kannan @Kami Please make sure I've not gone Christmas insane on this, but basically I assume anything that's exactly matched in `"{{[A-z0-9_-]}}"` should attempt to be rendered with whatever JSON type it has. So that's precisely what I attempt to do w/ the `complex_type` jinja filter. It seems to work in the tests. If the item is surrounded by additional elements ie: "My fancy sentence has a {{variable}}" it will render it as a string (probably need to add another test to cover this. Will do that.). This covers all the corner cases I could think of though there's a possibility there's another one out there so please view this with a critical eye.

st2common/st2common/jinja/filters/complex_type.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def to_complex(value):
+    result = json.dumps(value)
+
+    if isinstance(result, unicode):
+        try:
+            result = str(result)
+        except UnicodeEncodeError:
+            pass
```
bigmstone: Should I raise here w/ a user friendly error message or allow this to be unicode?

Kami:
From high-level the change looks good to me, but I would feel more comfortable with a lot more tests for different types and more edge cases.

st2common/st2common/jinja/filters/complex_type.py
Kami: IIRC, we allow unicode in other places (I did a bug fix for allowing unicode quite a long time ago) so we should probably also allow unicode here.

st2common/tests/unit/test_param_utils.py
```diff
@@ -343,34 +343,90 @@ def test_get_finalized_params_with_dict(self):
...
             'r1': '{{r2}}',
             'r2': ['1', '2'],
             'a1': True,
             'a2': 'Test',
             'a3': 'Test2',
             'a4': '{{a1}}',
-            'a5': ['{{a2}}', '{{a3}}']
+            'a5': ['{{a2}}', '{{a3}}'],
+            'a6': [
+                ['{{r2}}', '{{a2}}'],
```
Kami: Just curious - how many levels deep do we render stuff?

Or is it recursive aka we support arbitrary levels of nesting?

st2common/st2common/jinja/filters/complex_type.py
Kami: Is this unicode stuff still relevant?

st2common/tests/unit/test_param_utils.py
bigmstone: Supports arbitrary levels of nesting since it just dumps to JSON then looks for the specific regex.

st2common/st2common/util/param.py
```diff
@@ -150,8 +152,32 @@ def _render(node, render_context):
     Render the node depending on its type
     '''
     if 'template' in node:
+        complex_type = False
+
+        LOG.debug("Rendering param of type: %s", type(node['template']))
+
+        if isinstance(node['template'], list) or isinstance(node['template'], dict):
+            node['template'] = json.dumps(node['template'])
```
Kami: Just thinking out loud.

I assume this will always work since `node['template']` can only contain simple types which are JSON serializable, right?

Because if not, this could cause issues with types which are not JSON serializable (datetime, class instances, etc).

st2common/st2common/jinja/filters/complex_type.py
bigmstone: Probably not. I will probably just remove it. Let me do that and see what break. :trollface:

st2common/st2common/util/param.py
```diff
@@ -150,8 +152,32 @@ def _render(node, render_context):
     Render the node depending on its type
     '''
     if 'template' in node:
+        complex_type = False
+
+        LOG.debug("Rendering param of type: %s", type(node['template']))
```
Kami: Do we have access to the param name or similar here? If not the log message seems bit confusing / not so useful since it only contains a type and not any other additional context which could help track it back or similar...

st2common/st2common/util/param.py
bigmstone: Yeah this is getting fed by our supported types (Unless I'm mistaken here) which are all supported in JSON.

st2common/st2common/util/param.py
Kami: Never mind, I see you do log the whole template couple if lines below.

I'm fine with that, although it could potentially cause problems if template contains sensitive data (although this shouldn't be an issue since it's a DEBUG log level).

bigmstone: Meant to take this out. I added it in when I was debugging why context was getting rendered incorrectly

Kami:
Added some comments / questions.

Besides that + changelog entry, LGTM.

st2common/st2common/util/param.py
Kami: Sounds good :+1:

st2common/st2common/util/param.py
```diff
@@ -150,8 +152,28 @@ def _render(node, render_context):
     Render the node depending on its type
     '''
     if 'template' in node:
+        complex_type = False
+
+        if isinstance(node['template'], list) or isinstance(node['template'], dict):
+            node['template'] = json.dumps(node['template'])
+            node['template'] = re.sub(
```
lakshmi-kannan: I think a comment explaining this magic would be helpful with an example.

lakshmi-kannan:
I still can't believe you solved it :). Looking at the tests, it looks like you've. So from my side, we are good 👍. Awesome work, @bigmstone!

PR #3019, opened 2016-11-11
Remove packs.info action

Kami:
With new pack management code, `.gitinfo` file has been deprecated (it's not used anymore) and the actual pack directories are checked out git repository so this file and action is not needed anymore.

If user wants to see which version is installed or similar, they can use existing `pack get` CLI command. In the future, if need arises, we can also add more information to the `pack get` command output (e.g. currently checked out git revision, etc.).


emedvedev:
I'll change it in https://github.com/StackStorm/st2/pull/2982.

Kami:
@emedvedev Works for me.

On a related note - we could also leave it there for backward compatibility reasons for a while, but we probably want users to migrate to the new system so getting rid of it right now is not a bad idea.

What's our plan for that PR - do we plan to merge it soon and include it in v2.1?

emedvedev:
Yeah, I'm planning to get to it soon and have it ready for 2.1, and then I'll comb through the st2contrib packs using the old `packs` actions (like github/bitbucket rules), and change or disable those rules in their Exchange versions.

emedvedev:
So the backward compatibility here is achieved by the fact that old st2, having the old `packs` pack, will get packs from st2contrib (and packs there will reference the old syntax), and new st2 will use newer versions from Exchange, that will reference new pack commands.

lakshmi-kannan:
Nice. We should capture this deprecation in the blog for new pack management

Issue #2260, opened 2015-12-02
Clean-up old rules on update of a pack


manasdk:
Rules that are removed from a pack are not removed from StackStorm when the pack is updated. Similar behavior for other content.

Current work around is to delete old content from StackStorm manually.

PR #3021, opened 2016-11-14
Fix/pack.unload

userlocalhost2000:
Hi, 

In the current processing of unload action, the only rules which are belonged to the pack are removed.
And the rules that have association with the packs in trigger or action are remained.

Both rules should be removed as mentioned in #2260.

This patch additionally remove rules that are set the pack in trigger or action.

Thank you.

### DIFF
--- a/contrib/packs/actions/pack_mgmt/unload.py
+++ b/contrib/packs/actions/pack_mgmt/unload.py
@@ -94,7 +94,7 @@ class UnregisterPackAction(BaseAction):
 
         # delete rules that has an association with the pack in trigger or action.
         def _match_trigger(r):
-            return (r.trigger.split('.')[0] == pack)
+            return r.trigger.split('.')[0] == pack
 
         def _match_action(r):
             return r.action.ref.split('.')[0] == pack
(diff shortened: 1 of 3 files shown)


Kami:
Thanks for the pull request.

This is an intentional behavior we have decided on a while back - when user uninstalls a pack we only remove content which belongs to that particular pack. This is to avoid surprises - if user wants to remove rules from other packs which reference trigger from a pack which has been deleted, the rule needs to be removed manually.

Another reason for that is that even if we removed rules from other packs, we would only remove rules from the database (we wouldn't touch content from other packs on disk because this violates infrastructure as code principles) and the rule would be registered again when user ran st2ctl reload or similar.

I do agree though that we should document this behavior better and perhaps also log rules from other packs which are left and let user know they should be cleaned up.

/cc @lakshmi-kannan

manasdk:
I agree with Kami. This will lead to some really unexpected behavior.

Kami:
Thanks for this PR but as described above this is the desired behavior right now so I will go ahead and close it.

userlocalhost2000:
Thank you for your reply.
I don't mind. I also hope that this is to be better.

Thank you.

PR #3031, opened 2016-11-17
Make Runner Loader Have Consistent Output

bigmstone:
Change `info` level logging of runner register to `debug` so that output is consistent with other loaders.


Kami:
printf debugging ftw! :P

PR #3033, opened 2016-11-17
Add new CLI tool for validating config against a specific schema

Kami:
This pull request adds a script which allows user to validate config file against a specific config schema file.

By default, the same thing is performed when user runs ``st2-register-content --register-configs`` script, but that validation is only performed on the actual pack config file in ``/opt/stackstorm/config/<pack name>.yaml`` (if one exists).

In a lot of cases you want the flexibility and perform validation on a file outside the configs directory. For example, you want to validate that the config file is correct (passes validation) before you deploy your configs and move them to the final location in ``/opt/stackstorm/configs/`` and similar.

On a similar note - we have a bunch of sample configs inside pack directories and right now we don't perform any validation on those example files which means they could be invalid or out of date.

In the future, I also plan to add "context" or similar option which will allow user to validate config with dynamic configuration values - user will be able to provide mock context for the dynamic config values which are read from the datastore.

Example usage:

```bash
(virtualenv)vagrant@vagrant-ubuntu-trusty-64:/data/stanley$ ./st2common/bin/st2-validate-config --config-file conf/st2.dev.conf --schema-path /data/st2contrib/packs/cubesensors/config.schema.yaml  --config-path valid.yaml
Validating config "valid.yaml" against schema in "/data/st2contrib/packs/cubesensors/config.schema.yaml"
Config "valid.yaml" successfuly validated against schema in /data/st2contrib/packs/cubesensors/config.schema.yaml.
```

```bash
(virtualenv)vagrant@vagrant-ubuntu-trusty-64:/data/stanley$ ./st2common/bin/st2-validate-config --config-file conf/st2.dev.conf --schema-path /data/st2contrib/packs/cubesensors/config.schema.yaml  --config-path invalid.yaml 
Validating config "invalid.yaml" against schema in "/data/st2contrib/packs/cubesensors/config.schema.yaml"
Failed to validate pack config: Failed validating attribute "sensor" in config for pack "unknown" (/data/stanley/invalid.yaml): Additional properties are not allowed ('boo' was unexpected)

Failed validating 'additionalProperties' in schema['properties']['sensor']:
     'description': 'Sensor specific settings.',
     'properties': {'device_uids': {'description': 'A list of device UIDs to poll metrics for.',
                                    'items': {'type': 'string'},
                                    'required': False,
                                    'type': ['array', 'null']}},
     'required': False,
     'type': ['object', 'null']}

On instance['sensor']:
    {'boo': 'a', 'device_uids': ['a', 'b']}

```

## TODO

- [x] Add tests
- [x] Update changelog
- [x] Update packages to copy the new tool over - https://github.com/StackStorm/st2-packages/pull/368
- [x] Update st2contrib to perform example config validation - https://github.com/StackStorm/st2contrib/pull/617
- [ ] Port st2contrib changes into new StackStorm-Exchange


dzimine:
Cool. Will you add the docs on how to use it?

Kami:
Yeah, I will look into it at some point, but it's not a priority for me right now - I just wanted to get it working on CI first so we actually test / validate the examples configs we ship...

PR #3034, opened 2016-11-17
Reduce Python action runner script run-time

Kami:
Victory!

I finally tracked down the issue which causes Python action runner wrapper process take 2 seconds or so. As suspected before, the root cause are (slow) imports.

Notably the root cause is `jsonschema` import. jsonschema dynamically reads, loads and parses schema definition from JSON file on-disk (https://github.com/Julian/jsonschema/blob/master/jsonschema/validators.py#L162). This adds around a second of so to the run-time to any module or package which directly or in-directly imports that module.

I accidentaly encoutered this again today when I added a simple new CLI tool / script which took almost a second to run even though it's very simple.

It turns out that that this script and python action process wrapper module both indirectly import jsonschema (it's like 10 level deep in-direct import so it wasn't that easy to track it). Sadly there is nothing I can do about the new tool because it relies on jsonschema, but python action wrapper doesn't need it so I can reduce the run-time by a good second, by re-organizing imports.

Before

```bash
(virtualenv)vagrant@vagrant-ubuntu-trusty-64:/data/stanley$ time python st2common/st2common/runners/base_action.py

real	0m1.500s
user	0m0.512s
sys	0m0.753s
```

```bash
(virtualenv)vagrant@vagrant-ubuntu-trusty-64:/data/stanley$ time python st2common/st2common/runners/base_action.py

real	0m0.103s
user	0m0.019s
sys	0m0.065s
```

As you can see in 521d8020f7cb55225b76cb9c8dbcaedf94bd9ecb, the problem lies inside runners.init module which in-directly imports jsonschema which is then imported by base_action.py which is used inside the action wrapper.

That simple one-liner here adds more than a second to the import time - https://github.com/StackStorm/st2/blob/master/st2common/st2common/runners/base_action.py#L20 (you can easily replicable it by commenting it out and timing it).

The "funny" thing is that this line doesn't even directly import `__init__.py`, but because how Python module loading system works, `__init__.py` is loaded in-directly (__init__.py makes directory a package). 

This is exactly the same reason why I warned everyone many times in the past to put code in `base.py` or similar module instead. Putting constants in __init__.py is fine, but putting anything else can make it a ticking time bomb like in this case - everyone please keep this in mind going forward.

It's also worth noting that this is not an issue with services and long runing processes, but it's an issue with any short-lived script and Python action wrapper where 1 second of import time is a lot of overhead.

## TODO

- [x] Finish it, update existing code to import from base, not init
- [x] Tests for python runner wrapper script run time to prevent accidental regressions in the future


Kami:
Edit: It's not that easy - it turns out there are more in-direct import which are causing issues, I'm working on it... 

The solution won't be nice, but that's the most we can do right now. Anything else would require massive well planned and though out refactor.

Kami:
It's not pretty, but it works - c1f6833a7bcdcac9d5d36cbf1069846d9a4c079f

As mentioned above, other solution would involve a big well thought out and planned refactor which is really out of scope now and for the short and long term future.

Kami:
I've added some more work-arounds and optimizations, but that's about a low as I can get right now - it's still a pretty substantial improvement over the previous status quote - 50% - 70% for short lived actions.

Currently the part which takes the most is "db_setup" - establishing database connection inside the subprocess and retrieving config schema + config items.

This could be improved in the future by moving config parsing and loading to action container which would then parse already loaded values to the subprocess. This would allow us to reduce the time by another 200-300ms because action runner is a long running process and already has a connection to the database established (unlike the process wrapper which is a short lived one where establishing a connection for short lived actions represents a big overhead).

armab:
> This is exactly the same reason why I warned everyone many times in the past to put code in base.py or similar module instead. Putting constants in init.py is fine, but putting anything else can make it a ticking time bomb like in this case - everyone please keep this in mind going forward.

ACK :+1:

Kami:
Just for the completeness, here is an example of how much overhead pkg_resources adds and how slow it is https://gist.github.com/Kami/bfc893b52f38b50a709c0b4caa6a3351.

lakshmi-kannan:
This is awesome! Thanks for the PR @Kami!

PR #3035, opened 2016-11-17
Cache filters API response for web UI

### LINKED ISSUE #2876: Cache filters API response for web UI
As the number of actions, rules, etc has grown in our StackStorm cluster, the UI is becoming slower. In particular, the filters view (which we rely on heavily with this many executions) is very slow to become visible and usable. It might help to cache this `/api/v1/executions/views/filters` API response and asynchronously reload this value outside of the request thread, since this is slowly-changing data.

Our typical workflow is to
1. load the main execution page
2. wait for the filters to appear
3. apply some set of filters
4. wait for the next page to load

Here's a waterfall showing a 20s load time for the filters response. This is pretty common for us now.
screen shot 2016-08-22 at 12 20 14 pm

For reference, we have 572 rules, 1200 actions, 143 triggers, 19 trigger types, and 600k+ executions.

codyaray:
Use a TTL-based cache since we don't have a better cache invalidation mechanism here.

Hardcoded parameters and other yuck. But thought I'd get the ball rolling. It works for me. :)

Closes #2876

### DIFF
--- a/st2api/in-requirements.txt
+++ b/st2api/in-requirements.txt
@@ -11,3 +11,4 @@ six
 git+https://github.com/StackStorm/pecan.git@st2-patched#egg=pecan
 git+https://github.com/StackStorm/python-mistralclient#egg=python-mistralclient
 gunicorn
+cachetools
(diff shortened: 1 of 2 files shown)


enykeev:
Closing in favor of #3038.

PR #3037, opened 2016-11-19
Update various dependencies to latest versions

Kami:
Just some house keeping.

I tested the changes locally and went over the changelogs - new versions include various fixes, etc.

Most upgrades should be harmless, the biggest one is paramiko to 2.0 (http://bitprophet.org/blog/2016/04/23/paramiko-2.0-is-coming/). In version 2.0 PyCrypto dependency which requires C extensions to build is replaced with cryptography.

I haven't encountered any issues yet locally, but in case it cause issues we can revert that upgrade.


Kami:
I also tried to upgrade tooz, but any version after 1.20.0 causes our tests to hang.

I believe I narrowed down the issue to the code change inside tooz, but I will leave it as is for now.

Kami:
OK, all the tests passed locally and on the CI and there appears to be no issues with paramiko upgrade (I also did manual testing locally).

In case you do encounter issues or weirdness with remote SSH runner, let me / us know.

PR #3040, opened 2016-11-21
Uncomment [webui] section in st2.conf

armab:
Originally reported in #community.

User uncommented only `webui_base_url = https://mywebhost.domain` without uncommenting `[webui]` section, ex:
```ini
# [webui]
webui_base_url = https://mywebhost.domain
```
See: https://github.com/StackStorm/hubot-stackstorm/issues/129#issuecomment-262007128

With this small change it looks safer, more obvious and less error-prone default.

### DIFF
--- a/conf/st2.dev.conf
+++ b/conf/st2.dev.conf
@@ -68,7 +68,7 @@ port = 514
 facility = local7
 protocol = udp
 
-# [webui]
+[webui]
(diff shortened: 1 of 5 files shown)


Kami:
Thanks.

It doesn't seem to break tests so I will go ahead and merge it. Will also verify locally it doesn't cause issues with dev environment (it shouldn't, but you never know).

PR #3044, opened 2016-11-23
[WIP] Moving Core Runner Tests

bigmstone:
Moving core runner tests to `st2common` to prevent these tests from being copied with packages.


armab:
Just files don't take too much space. Why it's important not to copy tests with packages?
Required by some compliance stuff?

Kami:
Yeah, I personally don't see a problem with including tests in the final distribution (packages).

Space is cheap in fact it's usually is a good idea to include tests with the final distribution - if someone unpacks the distribution or similar, the tests are there for a specific version and should pass.

We did this in Libcloud in the past, but then people complained they want us to include tests to confirm that the actual release artifact is working correctly (albeit, it's a slightly different scenario since for Libcloud we distribute source code and Python packages, but still).

bigmstone:
Okay. There was a jira ticket out. I'll close this PR and mark that as "won't-fix YOLO DGAF GTFO". Which I wish was an actual label now.

PR #3048, opened 2016-11-23
Fix Core Runner Tests

bigmstone:
- [ ] Mistral Tests


bigmstone:
@m4dcoder Might need your help with some of the mistral tests. Will pick back up on it after Thanksgiving.

PR #3050, opened 2016-11-24
Python runner env description

enykeev:
The description is confusing to some users since it only reflects CLI specifics.

Fixes https://github.com/StackStorm/st2web/issues/280


Kami:
Agreed.

LGTM, +1

PR #3054, opened 2016-11-28
Various pack API changes

Kami:
* Improved error message
* Pack search is now case insensitive by default. This makes it more useful and easier to find things.


emedvedev:
Good catch about the case sensitiveness. :heart:

PR #3062, opened 2016-11-29
Fix pack uninstall action to remove pack db object based on the pack ref

Kami:
This pull request fixes "pack uninstall" action to remove pack based on the pack ref and not the name.

If we would use the name, it wouldn't work correctly with new pack management for packs where name != ref (e.g. travis_ci, etc.).

# TODO

- [x] Integration tests for unload / uninstall


contrib/packs/actions/pack_mgmt/unload.py
```diff
@@ -109,9 +109,22 @@ def _unregister_pack(self, pack):
...
+
+        # 1. Try by ref
         try:
-            pack_db = Pack.get_by_name(value=pack)
+            pack_db = Pack.get_by_ref(value=pack)
         except StackStormDBObjectNotFoundError:
+            pack_db = None
+
+        # 2. Try by name (here for backward compatibility)
+        if not pack_db:
```
Kami: For just in case aka backward compatibility I left this code in place.

Plan to remove it in 2.1 or similar.

PR #3066, opened 2016-11-30
Fix description for 'pack install' Alias

armab:
`=` is not acceptable in Action Alias as separator between `pack=version` because of https://docs.stackstorm.com/chatops/aliases.html#key-value-parameters

So adjusting description accordingly.

### DIFF
--- a/contrib/packs/aliases/pack_install.yaml
+++ b/contrib/packs/aliases/pack_install.yaml
@@ -2,10 +2,10 @@
 name: "pack_install"
 action_ref: "packs.install"
 pack: "packs"
-description: "Install/upgrade StackStorm packs (pack = name or git URL, gitref = tag/branch/commit)."
+description: "Install/upgrade StackStorm packs."
 formats:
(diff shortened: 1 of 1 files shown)


contrib/packs/aliases/pack_install.yaml
```diff
@@ -2,10 +2,10 @@
 name: "pack_install"
 action_ref: "packs.install"
 pack: "packs"
-description: "Install/upgrade StackStorm packs (pack = name or git URL, gitref = tag/branch/commit)."
+description: "Install/upgrade StackStorm packs."
 formats:
   - display: "pack install <pack>[,<pack>]"
-  - display: "pack install <pack>=<gitref>[,<pack>=<gitref>]"
+  - display: "pack install <gitref>[,<gitref>]"
```
emedvedev: `<git URL>` then? `ref` is usually a commit/branch/tag reference.

armab: Yeah +1

PR #3068, opened 2016-12-01
Remove config on pack unload

enykeev:
Remove the config alongside the pack. Requested by @lakshmi-kannan.


Kami:
I'll merge this into #3068 since I plan to add tests there (the other then when I noticed another unintall bug I noticed we have no direct tests for it).

Kami:
And a good catch and thanks btw!

Kami:
Continuing in #3062 (with tests).

PR #3070, opened 2016-12-01
WIP: Rename runners/base to runners/base_runner

lakshmi-kannan:


lakshmi-kannan:
@Kami and I have concluded the problem is adding st2common/runers to sys.path for python actions is causing the issue. https://gist.github.com/lakshmi-kannan/e7aa1cddd0e9e3ae73a748892b433503. There is no need for st2common/runners to be added to sys.path. This was added when we move action_wrapper from st2actions to st2common.

lakshmi-kannan:
So the fix is to pop st2common/runners from sys.path. @Kami will get this in another PR. I am closing this PR as it is not needed.

PR #3075, opened 2016-12-01
Workaround for Python action runner subprocess sys.path pollution

Kami:
TBW.

## TODO

- [x] Tests
- [x] Changelog


Kami:
I added a test case, etc.

Will go ahead and merge it into master and v2.1 once the build passes.

@lakshmi-kannan After it has been merged we still need to merge st2tests branch and also test that it works with packages (paths might be a bit different there and I also just want to verify it works correctly there - st2tests should catch that).

PR #3080, opened 2016-12-02
Pack register API should fail on pack register failure

Kami:
This PR updates pack register API so it fails on failure in case a pack registration fails (error is returned to the client).

Before that, it didn't fail so it was hard to debug pack installation issue (`st2 pack install` would succeed even though if pack registration failed). This way it's also consistent with `st2ctl reload` / register content which fails on failure by default.

If for some reason user wants not to fail on failure, they can pass "fail_on_failure" option in the request body set to False.

While working on that I noticed there is another issue - we didn't actually fail on failure before if registration of pack itself failed, we only failed if pack resource registration failed.

## TODO

- [x] Tests (pack registration fail, pack registration fail)
- [x] Make sure all the "failed to register" exception messages include name of the resource and / or pack which caused the failure


Kami:
Note: We decided to include this in the next patch release (v2.0.1).

enykeev:
I'd say that fail_on_failure should be a client side feature. The way it is implemented in this PR gives user zero insight into what kind of content had been registered before the failure occured and essentially leaves st2 in a broken state. It was working for st2ctl because we had an intermediate results flushed to stdout. With st2api, we don't have this luxury. 

Instead, API endpoint should attempt to register everything this instance has, gather all the errors happened and return them for client to decide on what to do with it.

The only use for server side `fail_on_failure` is for the case when it takes too long to register all the content every time, but I'd say we need to treat it as a dangerous feature one should be fully aware of consequences of using it. Also, the output format should be shared with the one proposed earlier (result + list of errors rather than a single faultstring).

Also, let me remind you that we still have fully functionsl implementation of fail_on_failure in st2ctl and that we intend on stop exposing registration process to the user during stage two and we'd better start thinking from the perspective "how it should work in the future" rather than "how it worked previously".

Kami:
@enykeev Yes, it leaves stuff in a "semi consistent state", but this will always be the case no matter what approach we go with unless we implement some kind of transactions for registration (register all or nothing if a failure occurs).

This would be ideal, but imo that's overkill and users should be encouraged to run those checks themselves on their CI / CD servers to catch such issues before pushing things to production (that's what we do on Travis with our and community contributed packs).

Having said that, I think that failing early is good enough. I do agree though that there is a small difference between st2ctl and register API - st2ctl prints some logs so use has more context (if required), but this only happens when running it with `-v` flag which is not a default behavior.

To make it easier to troubleshoot when registering multiple packs using the API we could add `debug=True` or similar parameter which would cause it to collect and returns the logs and similar.

And it's also worth nothing that all of that only applies when registering multiple packs at once. On `packs.install`, we only register a single pack so this is not an issue. As @dzimine said, I would personally also be fine with this API just supporting registration of a single pack at once then this problem goes away completely.

> Also, let me remind you that we still have fully functional implementation of fail_on_failure in st2ctl and that we intend on stop exposing registration process to the user during stage two and we'd better start thinking from the perspective "how it should work in the future" rather than "how it worked previously".

One of the problems with that is that pack install command already uses register API which currently doesn't correctly report errors back to the users which makes a bad UX and also makes it hard to troubleshoot what is going on.

enykeev:
> but this only happens when running it with -v flag which is not a default behavior.

I'm talking about the stage registration fails on. With the approach you proposed I can't tell what types of content has been registered and where in the process the fail had happened (eg have all the actions been registered). And after registration is over, there is no way to figure what was registered and what wasn't either...

> To make it easier to troubleshoot when registering multiple packs using the API we could add debug=True or similar parameter which would cause it to collect and returns the logs and similar.

No point of hiding it behind the flag, you want to get this info 100% of times. It's never ok if your registration failed. You want to get it fixed asap, not running it again with some magic flag.

We're going to spend more time arguing than actually making the change. Again. And you know it's better UX to give user all the information upfront rather than making him rerun registration until he fixes all the errors one by one.

Kami:
Will also cherry-pick into v2.1.1 branch.

PR #3081, opened 2016-12-02
submit-debug-info: pass the `output_path` as the shell command working directory

tim-ireland:
In addition to log files and stdout/stderr, I had the need to run a shell script as a `shell_command` when running submit_debug_info.py to generate a debug bundle.  The use case for the shell script is to create and then collect various additional files outside of the purview of st2 and place them in the tmp****** directory to be included in the archive. 

In order to do so effectively, I needed the shell script to reliably know the path to the temporary directory.  This change sets the `output_path` as the working directory of the subprocess which executes the shell command.  Ex: `/tmp/tmpp0qiO5/commands`

### DIFF
--- a/st2debug/st2debug/cmd/submit_debug_info.py
+++ b/st2debug/st2debug/cmd/submit_debug_info.py
@@ -424,7 +424,7 @@ class DebugInfoCollector(object):
         LOG.debug('Including the required shell commands output files')
         for cmd in self.shell_commands:
             output_file = os.path.join(output_path, '%s.txt' % self.format_output_filename(cmd))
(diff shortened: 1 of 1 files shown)


lakshmi-kannan:
@tim-ireland Thanks for the PR. From the surface, it looks like a fine change but we aren't picking any more external PRs until we do a release (should happen early next week). Once release is behind, we'll get this merged. Tests would've made this decision simpler but we are where we are :)

Kami:
Sorry for the delay.

The change look good to me so I will go ahead and merge it into master - thanks again!

Kami:
Just a heads up - added some tests in https://github.com/StackStorm/st2/pull/3114.

PR #3084, opened 2016-12-04
Add step to delete st2tests dir. in st2-self-check script

humblearner:
Git clone fails on second attempt as the s2tests folder is not deleted: `fatal: destination path 'st2tests' already exists and is not an empty directory.`

### DIFF
--- a/st2common/bin/st2-self-check
+++ b/st2common/bin/st2-self-check
@@ -59,25 +59,22 @@ elif [ -e "${ST2_CLI_CONFIG_FILE}" ]; then
   fi
 fi
 
-echo -n "Cloning st2tests in /tmp.."
-echo ""
-cd /tmp
+pushd ~
+echo "Cloning st2tests in `pwd`."
 git clone --depth 1 https://github.com/StackStorm/st2tests.git
 
-echo -n "Copying asserts, fixtures, tests and examples packs.."
-echo ""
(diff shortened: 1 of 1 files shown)


st2common/bin/st2-self-check
```diff
@@ -59,17 +59,26 @@ elif [ -e "${ST2_CLI_CONFIG_FILE}" ]; then
   fi
 fi
 
+echo -n "Cloning st2tests in /tmp.."
+echo ""
+cd /tmp
```
lakshmi-kannan: You shouldn't clone to /tmp. Clone to ~

st2common/bin/st2-self-check
```diff
@@ -59,17 +59,23 @@ elif [ -e "${ST2_CLI_CONFIG_FILE}" ]; then
   fi
 fi
 
+pushd ~
```
Kami: Probably safer to clone stuff to `/tmp/`, just in case.

humblearner: @kami: https://github.com/StackStorm/st2/pull/3084#discussion_r90782303 and https://github.com/StackStorm/st2/pull/3084/commits/b6a6de7ef7e05c0a56fa204596b938d1da6fd111#diff-94c8a3b53a504995343d0f973102c9f1R64 😆

PR #3085, opened 2016-12-04
Add pack back pack version normalization

Kami:
This adds back pack version normalization (+ extra tests, etc.) from https://github.com/StackStorm/st2/pull/2990.

This way the upgrade to v2.1.0 will be more smooth and break less backward compatibility. We will still advise users to update their pack metadata when they get a chance since non-valid semver version strings are deprecated and we will fully drop support for it in one of the next major releases.

Will cherry pick into v2.1.

### DIFF
--- a/st2common/st2common/constants/pack.py
+++ b/st2common/st2common/constants/pack.py
@@ -19,6 +19,7 @@ __all__ = [
     'PACK_RESERVED_CHARACTERS',
     'PACK_VERSION_SEPARATOR',
     'PACK_VERSION_REGEX',
+    'NORMALIZE_PACK_VERSION',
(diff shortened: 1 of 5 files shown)


Kami:
Here is the deprecation message please see on invalid semver version:

```bash
2016-12-04 12:07:21,569 INFO [-] =========================================================
2016-12-04 12:07:21,570 INFO [-] ############## Registering actions ######################
2016-12-04 12:07:21,570 INFO [-] =========================================================
2016-12-04 12:07:22,521 INFO [-] Pack "Travis CI" contains invalid semver version specifer, casting it to a full semver version specifier (0.2 -> 0.2.0)
```

PR #3088, opened 2016-12-04
Workaround for Python action runner subprocess sys.path pollution (take 2)

Kami:
This pull request works on top of https://github.com/StackStorm/st2/pull/3075.

The work-around in that PR worked fine locally, but not with actual packages inside production. The problem was that I used a very safe, but also a very restrictive path check which works locally, but not with packages which use a virtualenv  for st2 so paths are different.

Here is an example of actual BASE_DIR and SCRIPT_PATH with packages:

* /opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/runners
* /opt/stackstorm/st2/lib/python2.7/site-packages/st2common/runners

Luckily we now also have en end to end test in addition to unit test which caught this.

### DIFF
--- a/st2common/st2common/runners/python_action_wrapper.py
+++ b/st2common/st2common/runners/python_action_wrapper.py
@@ -13,17 +13,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import os
 import sys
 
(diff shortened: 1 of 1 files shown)


Kami:
Will cherry pick this into v2.1.

Issue #3077, opened 2016-12-02
[st2reactor] Issue about no validation checking before dispatching trigger


userlocalhost:
The `dispatch` method on the `SensorService` class doens't check validation of `payload` argument which is passed to TriggerDispatcher with `payload_schema` of Trigger.

So a sensor can send a payload which is in violation of the `payload_schema` to a Trigger without warning or error, like following.

```yaml
---
class_name: 'TestSensor'
entry_point: 'test.py'
description: 'A test sensor'
poll_interval: 5
```

```python
### excerpt of test.py
class TestSensor(PollingSensor):
    def poll(self):
        payload = {
            'hoge': [1,2,3],
        }
        self._sensor_service.dispatch(trigger="linux.file_watch.line", payload=payload)
...
```

This is `payload_scheme` of `linux.file_watch.line` which is a pre-installed Trigger in StackStorm.
```
$ st2 trigger get linux.file_watch.line
...
| payload_schema    | {                                                    |
|                   |     "type": "object",                                |
|                   |     "properties": {                                  |
|                   |         "file_name": {                               |
|                   |             "type": "string"                         |
|                   |         },                                           |
|                   |         "line": {                                    |
|                   |             "type": "string"                         |
|                   |         },                                           |
|                   |         "file_path": {                               |
|                   |             "type": "string"                         |
|                   |         }                                            |
|                   |     }                                                |
|                   | }                                                    |
...
```

Then we could conforme that this invalid payload could be get from the properties of `linux.file_watch.line` as below.

```yaml
---
name: 'test-rule'
description: 'A test rule'
enabled: True
trigger:
  type: linux.file_watch.line

  action:
    ref: 'core.local'
    parameters:
        cmd: 'echo {{trigger}} >> /tmp/test'
```

And the output is here.

```
{hoge: [1, 2, 3]}
{hoge: [1, 2, 3]}
...
```

  I think the validation processing is needed to guarante that trigger output properties are based on `payload_schema` before dispatching Trigger.

Thank you.

Kami:
#3142 has been merged so I will close this.

If you have any objections or additional comments, let us know (and thanks again for your initial PR!).

PR #3094, opened 2016-12-05
added the function to verify that the payload format is complied with the schema of Trigger (#3077)

userlocalhost:
This adds the function to verify that the payload format is complied with the schema of specified Trigger.
In the past, the validation checking for payload format has never been, as mentioned in #3077.

The main changes of this patch are
- adding the function to verify the payload which will be the trigger output is complied with the `payload_schema` of the corresponding Trigger.
- adding the configuration parameter of `permit_invalid_payload` to permit dispatching trigger forcibly even if the payload doesn't match the Trigger's schema.

The former is the feature for the future implementation. And the later is the one for the backward compatibility.
Even if there is a sensor that dispatch a payload which violate TriggerType, the SensorService only output warning message.

Sometime in near future, I hope that all trigger output is guaranteed that the format of payload is same with the `payload_schema` by this.

Thank you.

### DIFF
--- a/st2common/st2common/validators/api/reactor.py
+++ b/st2common/st2common/validators/api/reactor.py
@@ -22,6 +22,7 @@ from st2common.constants.triggers import SYSTEM_TRIGGER_TYPES
 from st2common.constants.triggers import CRON_TIMER_TRIGGER_REF
 from st2common.util import schema as util_schema
 import st2common.operators as criteria_operators
+from st2common.services import triggers
 
 __all__ = [
     'validate_criteria',
@@ -65,11 +66,14 @@ def validate_trigger_parameters(trigger_type_ref, parameters):
     if not trigger_type_ref:
         return None
 
-    if trigger_type_ref not in SYSTEM_TRIGGER_TYPES:
-        # Not a system trigger, skip validation for now
+    trigger_type = triggers.get_trigger_type_db(trigger_type_ref)
+    if trigger_type_ref in SYSTEM_TRIGGER_TYPES:
+        parameters_schema = SYSTEM_TRIGGER_TYPES[trigger_type_ref]['parameters_schema']
+    elif trigger_type and trigger_type.payload_schema:
+        parameters_schema = trigger_type.payload_schema
+    else:
         return None
 
-    parameters_schema = SYSTEM_TRIGGER_TYPES[trigger_type_ref]['parameters_schema']
     cleaned = util_schema.validate(instance=parameters, schema=parameters_schema,
                                    cls=util_schema.CustomValidator, use_default=True,
                                    allow_default_none=True)
--- a/st2reactor/st2reactor/container/sensor_wrapper.py
+++ b/st2reactor/st2reactor/container/sensor_wrapper.py
@@ -33,6 +33,7 @@ from st2reactor.sensor.base import Sensor, PollingSensor
 from st2reactor.sensor import config
 from st2common.services.datastore import DatastoreService
 from st2common.util.monkey_patch import monkey_patch
+from st2common.services import triggers
 
 __all__ = [
     'SensorWrapper',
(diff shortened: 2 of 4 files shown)


Kami:
Thanks for this PR.

We are a bit busy with v2.1.0 release at the moment, but just in short, nice work - this should be a great PR.

In fact that's something we already planned in the past. We already validate parameters for system trigger types when sending payloads via webhook (timer, etc.), but we don't do that yet for user-defined triggers types. The idea behind it was to make the whole validation gradual so it won't break too much stuff - in first release introduce validation and make it optional (opt-in) and in a future release make it enabled by default, but allow user to disable it (opt-out).

I quickly glanced over the code - it looks like you have done it for the sensor container, but it would also be great to do it for webhooks. Look for `validate_trigger_parameters` function in the code - you should be able to extend it with validation for user-defined trigger types and then also re-use it inside the sensor container.

userlocalhost:
Thank you for your reply and telling about `validate_trigger_parameters`.

I learned about the past background and existing verification checker.
I'll correct this patch to use it.

Thank you.

userlocalhost:
@Kami 
I corrected the previous patch to use `validate_trigger_parameters` instead of reinventing validation checker.
Could you please review it?

Thank you.

Kami:
@userlocalhost Thanks and sorry for the delay - will look into it today.

st2common/st2common/validators/api/reactor.py
```diff
@@ -65,11 +66,14 @@ def validate_trigger_parameters(trigger_type_ref, parameters):
     if not trigger_type_ref:
         return None
 
-    if trigger_type_ref not in SYSTEM_TRIGGER_TYPES:
-        # Not a system trigger, skip validation for now
+    trigger_type = triggers.get_trigger_type_db(trigger_type_ref)
+    if trigger_type_ref in SYSTEM_TRIGGER_TYPES:
+        parameters_schema = SYSTEM_TRIGGER_TYPES[trigger_type_ref]['parameters_schema']
```
Kami: Thanks.

This change looks good to me now, but it's worth noting that this method is also to validate parameters when creating a rule so we also need some flag to disable validation during rule creation as well (similar to what you did with `permit_invalid_payload`).

In addition to that, we also need some tests for rule creation with invalid parameters when validation is turned on.

Kami: After some more thought, I think we should have a single config option.

Perhaps, something along the lines of:

```ini
[system]
validate_trigger_parameters = False
```

The description should also be explicit and say something along those lines of "True to enable trigger parameter validation during rule creation and when dispatching triggers inside a sensor".

For backward compatibility, we should default it to false like you did with the config option you added.

We also need to document this new flag and behavior in the upgrade notes section of document (clarify how it works and affects the system).

Kami:
@userlocalhost Added some comments.

Some small things which need to be addressed, besides that, LGTM :+1:

st2common/st2common/validators/api/reactor.py
userlocalhost: > In addition to that, we also need some tests for rule creation with invalid parameters

I had never realized that case, thank you. I'll cover it in the test.

> I think we should have a single config option.

Sure, I'll correct the configuration parameter and make it translate to the `system` section because this flag will affect system wide.

> We also need to document this new flag and behavior in the upgrade notes section of document (clarify how it works and affects the system).

I got it. I'll make another PR to [Upgrade Notes](https://github.com/StackStorm/st2docs/blob/598d7f82e9fba0d8b199a4a5bb15509e1ed051b7/docs/source/upgrade_notes.rst) in st2docs.

userlocalhost:
Thank you for your reviewing and giving me many advices.

> sorry for the delay

Please don't be concerned it. I think this patch requires much careful consideration because this makes much impact.
(If I were a maintainer, I might put it off)

BTW, I'll address the points that you mentioned.

Thank you.

Kami:
@userlocalhost Btw, if you are busy in the near future, let me know.

It would be great to get those changes merged as soon as possible - I'm also happy to do those small changes mentioned above myself and take it across the finish line :)

userlocalhost:
@Kami 
That would be great! Thank you for taking care of it.

Kami:
Just a heads up - I'm working on this.

I also want to let you know that I originally missed that and I was wrong. In fact, we do need two functions and config options - one for validating trigger parameters and one for trigger payload.

In any case, thanks again and I'm taking care of those changes - will open a PR and let you know once it's ready.

Kami:
Closing this in favor of https://github.com/StackStorm/st2/pull/3142.

Thanks again.

userlocalhost:
@Kami 
Thank you for finishing it off (and sorry for my late response).

I'll catch up with your commits for future reference.

Thank you.

Issue #2320, opened 2015-12-22
User-friendly YAML Validation


armab:
> Report from the #community

It's obvious that single-line Jinja variables like `{{ playbook }}` should be enclosed with `"{{ playbook }}"` or they treated as dict.
Ex:
```yaml
params:
    playbook: {{playbook}}
```

`st2` shows this error: 
```py
st2 run st2-chatops-aliases.deploy_and_run_task playbook=test-slack.yml 
..
id: 56787954089f67469a3d7627
action.ref: st2-chatops-aliases.deploy_and_run_task
status: failed
error: Failed to run task "run_task". Parameter rendering failed: Failed rendering value for action parameter "playbook" in task "run_task" (template string={{[playbook}}): unexpected '}', expected ']'
traceback: Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/st2actions/runners/actionchainrunner.py", line 307, in run
    action_params=action_parameters, context_result=context_result)
  File "/usr/lib/python2.7/dist-packages/st2actions/runners/actionchainrunner.py", line 504, in _get_next_action
    chain_context={'parent': parent_context})
  File "/usr/lib/python2.7/dist-packages/st2actions/runners/actionchainrunner.py", line 486, in _resolve_params
    raise ParameterRenderingFailedException(msg)
ParameterRenderingFailedException: Failed rendering value for action parameter "playbook" in task "run_task" (template string={{[playbook}}): unexpected '}', expected ']'
```

Here is how Ansible handles it:
![stackstorm_ansible-st2_ 24_-_circleci_-_2015-12-22_04 07 23](https://cloud.githubusercontent.com/assets/1533818/11946322/0fcee356-a862-11e5-88aa-8f78ef61c9e9.png)
I was literally shocked and how many possible corner cases they handle, catching even upstream lib errors and how easy/useful were such error messages. Love to user is part of success :heart:

----

Just believe it's good example where to aim and what approach to follow someday with more user-readable errors, better overall error reporting (and less raw Python tracebacks). Honestly, not too much cases when full Python traceback in `st2` was useful for me personally (as user).

Leaving it here as example and food for thinking.

Kami:
I do agree that we should always throw the most user friendly exception as possible.

As far as this specific case goes - is it possible to always throw an exception in such scenario? In some cases, a dictionary can be a valid value (when parameter type is `object`).

[...]

PR #3101, opened 2016-12-07
Throw more user-friendly exception on invalid pack name

Kami:
Default jsonschema exceptions are not that friendly so this pull request updates the code to throw a more user-friendly error in case pack ref / name contains invalid characters.

It's also worth noting that this pattern is becoming more common  in our code base (wrap original jsonschema exception and throw something more user-friendly) so we should probably come up with some kind of consistent "approach" / "conventions" for handling it consistently through the code base.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,8 @@ Changelog
 in development
 --------------
 
+* Throw a more user-friendly exception when registering packs (``st2ctl reload``) if pack ref /
+  name is invalid. (improvement)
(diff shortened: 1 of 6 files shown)


armab:
> It's also worth noting that this pattern is becoming more common in our code base (wrap original jsonschema exception and throw something more user-friendly)

:+1: for this good practice to catch all strange upstream errors and throw own [more user-friendly](https://github.com/StackStorm/st2/issues/2320).

PR #3106, opened 2016-12-12
Update changelog for core.http action change

Kami:
Changelog entry for change in https://github.com/StackStorm/st2/pull/3012.


armab:
> also update changelog

I just noticed we don't list that in [CONTRIBUTING](https://github.com/StackStorm/st2/blob/master/CONTRIBUTING.rst). I think we should.

@warrenvw Motivated to make that CONTRIBUTING.rst more contributor-friendly with your fresh vision?

PR #3108, opened 2016-12-13
Use the same registration order inside the packs register API as we do in st2-register-content

Kami:
This pull request fixes pack register API endpoint to register different resource types in the exact same order as we use in `st2-register-content`.

Not registering it in the same order can cause multiple issues:

1. Output which it not consistent with `st2-register-content`. This could be confusing because both scripts fail on failure by default, but they could fail with a different error in case a different resource register order is used (e.g. one fails with an error inside action metadata first and other fails with an error inside rule).
2. Rules could get registered before sensors / triggers which would cause an error (trigger doesn't exist).

This addresses an issue indirectly reported in https://gist.github.com/lakshmi-kannan/6493c0d303df0622145730cf6b8e2b22. @lakshmi-kannan reported an issue, but it looks like a problem lied elsewhere (it was an "old" bug unrelated to my changes) which is nevertheless still a good catch :+1: 

## TODO

- [x] Tests
- [x] Changelog entry

### DIFF
--- a/st2api/st2api/controllers/v1/packs.py
+++ b/st2api/st2api/controllers/v1/packs.py
@@ -16,6 +16,7 @@
 import re
 
 from collections import defaultdict
+from collections import OrderedDict
 
 import pecan
 from pecan.rest import RestController
@@ -58,15 +59,18 @@ __all__ = [
 
 LOG = logging.getLogger(__name__)
 
-ENTITIES = {
-    'action': (ActionsRegistrar, 'actions'),
-    'trigger': (TriggersRegistrar, 'triggers'),
-    'sensor': (SensorsRegistrar, 'sensors'),
-    'rule': (RulesRegistrar, 'rules'),
-    'alias': (AliasesRegistrar, 'aliases'),
-    'policy': (PolicyRegistrar, 'policies'),
-    'config': (ConfigsRegistrar, 'configs')
-}
+# Note: The order those are defined it's important so they are registered in the same order as
+# they are in st2-register-content.
+# We also need to use list of tuples to preserve the order.
+ENTITIES = OrderedDict([
+    ('trigger', (TriggersRegistrar, 'triggers')),
+    ('sensor', (SensorsRegistrar, 'sensors')),
+    ('action', (ActionsRegistrar, 'actions')),
+    ('rule', (RulesRegistrar, 'rules')),
+    ('alias', (AliasesRegistrar, 'aliases')),
+    ('policy', (PolicyRegistrar, 'policies')),
+    ('config', (ConfigsRegistrar, 'configs'))
+])
 
 
 class PackInstallController(ActionExecutionsControllerMixin, RestController):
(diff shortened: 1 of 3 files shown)


Kami:
While working on this I encountered another "issue" - well, it's not an issue because the same behavior exists in register content and it has been then since the begining, but it's something we might need to document or similar.

Right now, we only register "dependent" resources when registering actions and rules - we also register runners when user requests to register actions and we register rule types when user requests to register rules.

Keep in mind that this only applies when user requests to register a single resource type.

Another thing we could do (but it's debatable and maybe current behavior is fine since it has been like that from the beginning) - when user requests to register just rules we could potentially also register other dependent types - notably sensors and triggers.

Imo though, this is not the best behavior, especially because sensors also contain actual sensors in additional to triggers.

st2api/tests/unit/controllers/v1/test_packs.py
```diff
@@ -169,6 +170,22 @@ def test_show(self):
         self.assertEqual(resp.status_int, 200)
         self.assertEqual(resp.json, PACK_INDEX['test2'])
 
+    def test_packs_register_endpoint_resource_register_order(self):
```
Kami: This is not ideal, but I can't come up with a better test case which doesn't require bigger changes.

Kami:
For other supposed bug (not a bug, same behavior existed since the beginning), please refer to the comment here - https://gist.github.com/lakshmi-kannan/6493c0d303df0622145730cf6b8e2b22

> The reason why pack.install doesn't fail, but pack register and st2ctl reload --register-all do is because packs.register action only registers actions, aliases and sensors by default (see packs.load action metadata). It has been like that since the beginning and nothing has changed and exact same behavior existed before the packs API change.

> So until we decide how to proceed there is nothing for me to do with regard to that - one option would be to update packs.load to register all resource types, but again, this is a "breaking" change since it changes an old default behavior so probably not a good idea to do it in v2.1.1.

> Note: This only applies if pack github is not already installed, if it's installed this error won't appear of course.

st2api/st2api/controllers/v1/packs.py
```diff
@@ -53,20 +54,24 @@
...
+# We also need to use list of tuples to preserve the order.
+ENTITIES = OrderedDict([
+    ('trigger', (TriggersRegistrar, 'triggers')),
+    ('sensor', (SensorsRegistrar, 'sensors')),
+    ('action', (ActionsRegistrar, 'actions')),
+    ('rule', (RulesRegistrar, 'rules')),
+    ('alias', (AliasesRegistrar, 'aliases')),
+    ('policy', (PolicyRegistrar, 'policies')),
+    ('config', (ConfigsRegistrar, 'configs'))
+])
```
enykeev: For the record, I'm unhappy with this approach and I want us instead to use networkx to define all kinds of dependencies between system entities and the order in which they should be registered. I'm ok with this as a temporary solution though.

Kami: I agree - ideally it would also be some system defined constant which we can also re-use in some way inside st2ctl and other places (right now tht's not the case so it relies on good test coverage to catch such issues).

PR #3109, opened 2016-12-13
Fix "get one" packs API endpoint

Kami:
This pull request fixes "get one" packs API endpoint - it would return no items if the pack name didn't match a ref.

The method itself was fine, but the actual problem was RBAC - rbac defaulted to using "get_by_name_or_id" instead of "get_by_ref_or_id".

## TODO

- [x] Tests

### DIFF
--- a/st2api/st2api/controllers/v1/packs.py
+++ b/st2api/st2api/controllers/v1/packs.py
@@ -43,6 +43,7 @@ from st2common.models.api.pack import PackInstallRequestAPI
 from st2common.models.api.pack import PackRegisterRequestAPI
 from st2common.models.api.pack import PackSearchRequestAPI
 from st2common.models.api.pack import PackAsyncAPI
+from st2common.exceptions.db import StackStormDBObjectNotFoundError
 from st2common.persistence.pack import Pack
 from st2common.rbac.types import PermissionType
 from st2common.rbac.decorators import request_user_has_permission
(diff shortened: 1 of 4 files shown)


st2api/st2api/controllers/v1/packs.py
```diff
@@ -284,6 +289,10 @@ class PacksController(BasePacksController):
     views = PackViewsController()
     index = PacksIndexController()
 
+    def __init__(self):
+        super(PacksController, self).__init__()
+        self.get_one_db_method = self._get_by_ref_or_id
```
enykeev: Shouldn't we fix RBAC instead of hacking around the particular endpoint?

Kami: That's how RBAC decorator for "get one" works - it uses method defined on the constructor / class.

It's also defined in other places, but it defaults to "get by name or id" which is not what we want here.

enykeev: How about we use another decorator? My point, if that's the problem with RBAC, we need to fix it RBAC side. If we have enough RBAC problems, we should rewrite it. We would need to rewrite it anyway for st2api v2 so we better start gathering what's wrong with it.

lakshmi-kannan: +1 to enykeev's sentiment. It seems like a hack now.

lakshmi-kannan:
+1 to changes.

st2api/st2api/controllers/v1/packs.py
Kami: Yes, one option would be to pass the method name / callable to the decorator or similar, but I decided to go with current approach because it's more generic.

No matter what approach we go with, we will need to specify this information (how to retrieve the item by PK) in some form, and the current approach is that it's specified on the class as a class / instance variable.

PR #3113, opened 2016-12-14
Run circle for every PR

enykeev:
This change is needed to start experimenting with moving away from Travis. Circle would refuse to run PRs with updated circle.yml unless we remove `branches.only` from the master version.


Kami:
LGTM

enykeev:
@Kami please repeat that in commanding voice

Issue #3112, opened 2016-12-13
pack install does not register rules in 2.1


emptywee:
@bigmstone asked me to file an issue about it.
`st2 pack install <git repo url>` does not register rules. 

Either documentation should mention it, or user should be allowed to add some argument to the install command (like we had register=all option before), or explicitly tell users to use the `st2 pack register <packname>` command if you want anything other than actions to get [re]registered.

Kami:
@bigmstone This is the same "issue" as the one I mentioned on Slack and here - https://github.com/StackStorm/st2/pull/3108.

As far as I know, it's not related to v2.1.0, but the same behavior existed before. By default, `packs.load` command only loads actions aliases and sensors - https://github.com/StackStorm/st2/blob/master/contrib/packs/actions/load.yaml#L10.

We intentionally didn't load rules by default in the past because rules usually need to be configured and enabled by the users before they make sense (they are more installation specific then other resources).

If we all agree that for some reason we should need load rules by default (at the very least we should fix it to register triggers by default - I got this), I'm fine with making the change, but it should probably be part of v2.2 or similar.

armab:
Since with a new pack management we register content per-pack now (which is a big improvement), maybe it's a good time to register all content coming with a pack?

This always was confusing for users and they kept asking over and over and created duplicated Issues.

armab:
If registering rules by default, we maybe should double-check that all packs in `Exchange` have `disabled` flag for rules. So the transition will be safer.

At the same time, with change like that, users may have better infra as a code and better flexibility.
They can place "enabled" rules in custom packs and assume that everything is registered, eg. no futher steps needed like `register`.
Better and more predictable deployments, especially when using configuration management tools.

Kami:
@armab We have many discussion in the past but perhaps it's not bad to re-evaluate the decision.

Most rule actually do ship disabled by default this days so it might be OK to just register them on install. It would only be problematic with rules which refer to triggers from other packs, but I don't think we actually have any such rules in Exchange (we do have them in st2cd and st2ci though).

/cc @lakshmi-kannan @dzimine

LindsayHill:
Not sure exactly which change it came in, but it appears that rules are now registered by default.
e.g.
```bash
vagrant@bwcvagrant:/opt/stackstorm/packs$ st2 --version
st2 2.4.1
vagrant@bwcvagrant:/opt/stackstorm/packs$ st2 pack install twitter

For "twitter" pack following content will be registered:

[...]

PR #3115, opened 2016-12-14
Update packs.load to register triggers by default

Kami:
Part of https://github.com/StackStorm/st2/issues/3112.

Not loading rules by default was intentional, but not loading triggers was an oversight - we missed it when we added support for new triggers resource type.

## TODO

- [x] Tests

### DIFF
--- a/contrib/packs/actions/load.yaml
+++ b/contrib/packs/actions/load.yaml
@@ -8,7 +8,7 @@
     register:
       type: "string"
       default: "actions,aliases,sensors,triggers"
-      description: "Possible options are all, sensors, actions, rules, aliases."
+      description: "Possible options are all, sensors, actions, rules, aliases, triggers, configs."
     packs:
       type: "array"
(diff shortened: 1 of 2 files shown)


contrib/packs/actions/load.yaml
```diff
@@ -7,8 +7,8 @@
   parameters:
     register:
       type: "string"
-      default: "actions,aliases,sensors"
-      description: "Possible options are all, sensors, actions, rules, aliases."
+      default: "actions,aliases,sensors,triggers"
```
armab: BTW, `triggers` are already in place here: 
https://github.com/StackStorm/st2/blob/8df8c5adaf8ec843702411fa8a00b029d898345c/contrib/packs/actions/install.meta.yaml#L19

so seems like it was unsynced someday.

Kami: @armab True, good catch - thanks!

PR #3117, opened 2016-12-14
Replace Travis with Circle as PR CI

enykeev:


enykeev:
I've moved tests but will need @humblearner help on getting coverage reports moved too.

We would also need to experiment moving around jobs between parallel containers to ensure utilization\acceptable timing.

.circle/install-mongo.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -e
+
+sudo apt-get purge mongodb-org*
+sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 7F0CEB10
+echo "deb http://repo.mongodb.org/apt/ubuntu trusty/mongodb-org/3.2 multiverse" | sudo tee /etc/apt/sources.list.d/mongodb-org-3.0.list
```
lakshmi-kannan: Should be mongodb-org-3.2.list :D. I didn't fix the typo. Me being lazy and all.

enykeev:
As far as I can tell, we're at parity between circle and travis. @humblearner please 👀 to make sure we reporting coverage properly. @lakshmi-kannan you've seem to volunteer for review 😛

.circle/install-mongo.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -e
+
+sudo apt-get purge mongodb-org*
```
Kami: Can't we use the similar ramdisk approach as we used on Travis?

It speeds things up and the script also allows us to install and run multiple and specific versions of MongoDB.

> -  - if [ ${TASK} = 'unit' ] || [ ${TASK} = 'integration' ] || [ ${TASK} = 'packs-tests' ]; then sudo -E ./scripts/travis/create-and-mount-ramdisk.sh; fi
> -  - if [ ${TASK} = 'unit' ] || [ ${TASK} = 'integration' ] || [ ${TASK} = 'packs-tests' ]; then ./scripts/travis/install-and-run-mongodb.sh; fi

circle.yml
```diff
@@ -22,24 +21,24 @@ machine:
...
     - sudo .circle/configure-services.sh
     - sudo .circle/add-itest-user.sh
     - sudo apt-get -y install parallel jq
-  post:
-    - cd ${ST2_REPO_ROOT_DIR} && hostname && make compile requirements
+  # Let's install Mongo 3 manually because circle doesn't support mongo3 yet
+    - sudo .circle/install-mongo.sh
+    - sudo pip install codecov
+  override:
+    - make compile requirements
```
Kami: Edit: Never mind, compile doesn't actually depend on requirements being present.

.circle/install-mongo.sh
Kami: And yes, to clarify - ideally we would always be able to use native service functionality without needing to hack stuff on top ourselves, but sadly this is usually only possible for very simple and basic use cases.

If you want better performance and / or control version, sadly this won't work anymore.

circle.yml
enykeev: That's how it was done previously and I don't really understand it deep enough to mess with it. Can you elaborate?

Kami:
Re better utilization and faster tests: Yeah, we should try to run those tasks in parallel and see how it goes:

* lint
* unit tests
* integration tests
* pack tests

That's basically what we did on Travis. In theory, as long as dependencies are cached, running those tasks in parallel should be quite faster, but yeah still good to try it out and actually measure it.

enykeev:
We already running them in parallel. Kami, please read the code before review 😛

enykeev:
Just want to point out that the purpose of this PR is not to make CI faster or achieve great utilization (although both are desirable), but to switch from Travis CI to Circle CI for PR checks, keep feature parity between the two and, hopefully, make it so it would take the same amount of time.

The last travis test took us 18 minutes of actual time and about 32 minutes of total time including waiting for containers to free. The test for this PR took 8 minutes total. Clearly we're not running travis anymore. So the only thing I need your help in determining is whether we achieved feature parity?

We can work on further speeding up build process and getting better utilization in the other PR.

.circle/install-mongo.sh
enykeev: Also, keep in mind that the way circle works, containers would block until all the parallel steps are finished so there is no gain in only installing mongo on some containers but not the other.

As far as ramdisk goes, I would need your explanation on why this change is needed and how it should affect the build. I'd rather we do it in a separate PR.

armab:
I don't think there is a significant change in speed between Travis and Circle for things like Unit/Integration tests.

`18 mins` in last Travis check is an exception from the rule.
Circle can be slow too from time to time too: ex `21 mins` here https://circleci.com/gh/StackStorm/st2/2609#build-timing

And "Total time" in Travis is how they summ up time from all parallel executions, makes almost no sense.

Here is the usual `10.5 mins` Travis build: https://travis-ci.org/StackStorm/st2/builds/183518380


So the diff in Circle vs Travis here I believe is mostly about usability. Example like that: 

> We already running them in parallel. Kami, please read the code before review

is a sign that it's hard to understand how stuff works in parallel in Circle (it's known issue).

enykeev:
> 18 mins in last Travis check is an exception from the rule.

Units, being the longest ones, take about 10+ minutes to execute. The result still qualifies as "not slower than Travis".

armab:
BTW, just remembered. There are CircleCI metrics with all bells & whistles.

Yeah, timing stability for `st2` looks not that bad for simple unit & integration tests. Only `4` regressions for last `2 months`: 
https://circleci.com/build-insights/gh/StackStorm/st2/master




Much worse for `st2-packages`, due to increased complexity and more upstream interactions:
https://circleci.com/build-insights/gh/StackStorm/st2-packages/master

circle.yml
```diff
@@ -22,24 +21,24 @@ machine:
...
-    - echo "deb http://repo.mongodb.org/apt/ubuntu trusty/mongodb-org/3.2 multiverse" | sudo tee /etc/apt/sources.list.d/mongodb-org-3.0.list
-    - sudo apt-get update
-    - sudo apt-get install -y mongodb-org
-    - sudo service mongod restart
   cache_directories:
     - ~/.cache/pip
   pre:
     - sudo .circle/configure-services.sh
     - sudo .circle/add-itest-user.sh
     - sudo apt-get -y install parallel jq
```
armab: Both packages seems to be unused. Not needed

circle.yml
```diff
@@ -13,7 +13,6 @@ notify:
 
 machine:
   environment:
-    ST2_REPO_ROOT_DIR: ${HOME}/${CIRCLE_PROJECT_REPONAME}
     BUILD_DOCKER: 0
   services:
     - docker
```
armab: Docker is not used

circle.yml
```diff
@@ -13,7 +13,6 @@ notify:
 
 machine:
   environment:
-    ST2_REPO_ROOT_DIR: ${HOME}/${CIRCLE_PROJECT_REPONAME}
     BUILD_DOCKER: 0
```
armab: Can be removed as well

armab:
Apart of that, remove first comment lines of `circle.yml` to avoid confusion:
```yml
# Setup in CircleCI account the following ENV variables:
# IS_PRODUCTION (default: 0)
# IS_ENTERPRISE (default: 0)
# PACKAGECLOUD_ORGANIZATION (default: stackstorm)
# PACKAGECLOUD_TOKEN
# DOCKER_USER
# DOCKER_EMAIL
# DOCKER_PASSWORD
```
everything that is not needed.

armab:
@enykeev Can you also verify if https://github.com/StackStorm/st2/blob/c3655c5d82210d6a184a70eb2bc4c44e4ad10e93/circle.yml#L10-L12 notify webhook to st2 is actually used and remove it if there is no need?

I guess it's from some old plans, but not sure.

.circle/install-mongo.sh
lakshmi-kannan: +1. Seems like an optimization. Can come in another PR.

lakshmi-kannan:
Awesome, besides the minor comments from others, I think we should forward here. You have my +1.

enykeev:
@armab even if we don't use this hook now (and I'm reasonably sure we do), we should start using it in a nearest future. So it's in. Minor things are fixed.

armab:
I'm OK with that, since everything was cleaned :+1:

PR #3123, opened 2016-12-16
Adding support for `st2 login` and `st2 whoami` commands

### LINKED ISSUE #3110: Update CLI authentication to store login
We use LDAP for authentication. I've been sharing this one-liner with all of our internal users:

```
export ST2_AUTH_TOKEN=$(st2 auth `whoami` -t)
```

As a new user this is definitely non-intuitive. It'd be really nice to support a mechanism like

```
st2 login
```

While you could require your username, you could also just default to the current user name since thats how many (most?) corporate environments would be setup with ldap anyway.

Or you can prompt with default like:

```
$ st2 login
Username [default <user>]: <prompt for input>
Password: <prompt for input>
```

Or skip straight to username and optional password

```
$ st2 login `whoami` -p $PASSWORD
```

Lots of ways to skin the cat. But the business objective here is to make it easier for new/occasional end-users to authenticate.

Mierdin:
# Overview

This PR introduces a new `st2 login` command to the StackStorm client. This command is similar to `st2 auth`, as it does generate a user token, but it also modifies `~/.st2/config` and adds a token file to the `~/.st2/config` directory. Further `st2` commands will use the new username/token combination that was written to disk.

This is in response to https://github.com/StackStorm/st2/issues/3110, where a user asked if there was a more friendly way of specifying the current user (i.e. without explicitly modifying configuration files or setting environment variables). This command is simply an alternative to those options.

Initially, I considered adding a flag to `st2 auth` that caches the token in `~/.st2/token-<username>` but realized that `~/st2/config` would also need to be modified to specify the "current" user. This seemed like a bit too much functionality to have in a single "flag" off of `st2 auth`, so I decided a separate command was ideal, and more self-explanatory.

By default, this command will write only the username to the configuration, for security purposes (not needed if a token file is present). The user can use the `-w` option to override this default and write the password into the config file as well.

> Note, an st2docs PR will follow this, to document the new command. I just wanted to make sure the approach was valid before I started updating docs.

# Example

Here's the new command in action. Note that the configuration file is updated and only the relevant section ("credentials") is written to - the rest is preserved.

```
(virtualenv)vagrant@st2dev:~/st2$ ll ~/.st2
total 12
drwxrwxr-x 2 vagrant vagrant 4096 Dec 16 17:42 ./
drwxr-xr-x 9 vagrant vagrant 4096 Dec 16 17:42 ../
-rw-rw-r-- 1 vagrant vagrant   40 Dec 16 17:42 config
(virtualenv)vagrant@st2dev:~/st2$ cat ~/.st2/config
[extrasection]
foo1 = bar1
foo2 = bar2

(virtualenv)vagrant@st2dev:~/st2$ st2 login st2admin --password Password1!
Logged in as st2admin
(virtualenv)vagrant@st2dev:~/st2$ cat ~/.st2/config
[extrasection]
foo1 = bar1
foo2 = bar2

[credentials]
username = st2admin

(virtualenv)vagrant@st2dev:~/st2$ cat ~/.st2/token-st2admin
{"expire_timestamp": 1481996584, "token": "a89b4a5fcc0542bcac48d3d9ef67c8c9"}(virtualenv)vagrant@st2dev:~/st2$
```

On the other hand, a failed login attempt will do nothing:

```
(virtualenv)vagrant@st2dev:~/st2$ ll ~/.st2/
total 12
drwxrwxr-x 2 vagrant vagrant 4096 Dec 16 17:45 ./
drwxr-xr-x 9 vagrant vagrant 4096 Dec 16 17:45 ../
-rw-rw-r-- 1 vagrant vagrant   40 Dec 16 17:45 config
(virtualenv)vagrant@st2dev:~/st2$ cat ~/.st2/config
[extrasection]
foo1 = bar1
foo2 = bar2

(virtualenv)vagrant@st2dev:~/st2$ st2 login st2admin --password wrongpassword
Failed to log in as st2admin
(virtualenv)vagrant@st2dev:~/st2$ ll ~/.st2/
total 12
drwxrwxr-x 2 vagrant vagrant 4096 Dec 16 17:45 ./
drwxr-xr-x 9 vagrant vagrant 4096 Dec 16 17:45 ../
-rw-rw-r-- 1 vagrant vagrant   40 Dec 16 17:45 config
(virtualenv)vagrant@st2dev:~/st2$ cat ~/.st2/config
[extrasection]
foo1 = bar1
foo2 = bar2
```

StackStorm documentation has been updated in https://github.com/StackStorm/st2docs/pull/380

**UPDATE**: Based on some of the review comments, I've added a few tasks for myself:

- [x] Add `st2 whoami` command ([thanks for idea](https://github.com/StackStorm/st2/pull/3123#issuecomment-268281752) @Kami)
- [x] Address logging configuration brought up in https://github.com/StackStorm/st2/pull/3123#discussion_r93160449
- [x] Update changelog
- [x] It's been a while since I looked at this PR - after all the above has been done, need to go through a full end-to-end test and ensure that everything shown in the examples above still works, and also verify that it still handles minor issues like `~/.st2/` not existing, etc.
- [x] Documentation tasks. Make sure the descriptions between `login` and `auth` commands are sufficiently distinct. Also at least open a st2docs PR and link here

Closes #3110


st2client/st2client/commands/auth.py
```diff
@@ -70,6 +73,65 @@ def run_and_print(self, args, **kwargs):
...
+        cli = BaseCLIApp()
+        logging.basicConfig(
+            level=logging.CRITICAL,
+            handlers=[logging.NullHandler]
+        )
+        cli.LOG = logging.getLogger("st2client.base")
+        cli._cache_auth_token(token_obj=manager)
+
+        # Update existing configuration with new credentials
+        config_file = "%s/.st2/config" % expanduser("~")
```
bigmstone: Should use `os.path.join` or `expanduser('~.st2/config')` here.

st2client/st2client/commands/auth.py
```diff
@@ -70,6 +73,65 @@ def run_and_print(self, args, **kwargs):
...
+        config = ConfigParser()
+        config.read(config_file)
```
bigmstone: What happens if this failes?

st2client/st2client/commands/auth.py
```diff
@@ -70,6 +73,65 @@ def run_and_print(self, args, **kwargs):
...
+        config['credentials'] = {
+            "username": args.username,
+            "password": args.password
+        }
+        with open(config_file, "w") as cfg_file_out:
```
bigmstone: Should be `w+` to create file if it doesn't exist.

st2client/st2client/commands/auth.py
Mierdin: No exception is raised, just an empty list is returned.

```
(Pdb) config.read("/tmp/foodoesntexist.txt")
[]
```

This just means there wasn't a file there to populate the config object.

If I run this with the real code, and the config file already exists, you see this:

```
(Pdb) config.read(config_file)
['/home/vagrant/.st2/config']
```

Either way, we're always overwriting the two parameters in the "credentials" section, and writing it to disk.

st2client/st2client/commands/auth.py
Mierdin: My understanding is that `w` will still create the file; `w+` is used if you want to also read from the file, which in this case I don't care about.

st2client/st2client/commands/auth.py
Mierdin: Done (03fb3f6).

Mierdin:
@bigmstone:

> Also what do you do in the event that the ~/.st2 folder doesn't exist? I'm of the opinion this tool should create it.

The `_cache_auth_token` method already has lots of checks like this, and I am doing that first, so if that has an error, the rest of the code won't run; if it doesn't, everything should exist.

You think this still warrants more checks within this function, or is that sufficient?

st2client/st2client/commands/auth.py
bigmstone: Yep, my mistake.

bigmstone:
@Mierdin Looked over `_cache_auth_token`. :+1:

Mierdin:
@bigmstone wanna take another quick look? Lots of changes in https://github.com/StackStorm/st2/pull/3123/commits/eb3aedcd99a6a8593851e3adce5179af51020e47 so maybe you might want to take a peek again?

st2client/st2client/commands/auth.py
```diff
@@ -70,6 +73,76 @@ def run_and_print(self, args, **kwargs):
...
+
+        logging.basicConfig(
+            level=logging.CRITICAL,
+            handlers=[logging.NullHandler]
+        )
+        cli.LOG = logging.getLogger("st2client.base")
+        cli._cache_auth_token(token_obj=manager)
+
+        # Update existing configuration with new credentials
+        config_file = expanduser('~/.st2/config')
```
armab: There is probably constant somewhere to store the `~/.st2/config` default value.

Additionally, here might be edge cases: 
what if env `ST2_CONFIG_FILE` or `--config-file=` flag is provided? See the docs: https://docs.stackstorm.com/reference/cli.html?highlight=cli%20authentication#configuration-file

I think we should support that for consistency.

st2client/st2client/commands/auth.py
```diff
@@ -70,6 +73,76 @@ def run_and_print(self, args, **kwargs):
...
+        config = ConfigParser()
+        config.read(config_file)
+
+        # Other st2 commands error out if the "password" field is missing from the "credentials"
+        # section. So, here we will write it to "notarealpassword", unless the args.write_password
+        # option is provided, in which case we'll write the real password.
+        config['credentials'] = {
+            "username": args.username,
+            "password": args.password if args.write_password else "notarealpassword"
```
armab: Writing `notarealpassword` to conf file looks really really strange

st2client/st2client/commands/auth.py
```diff
@@ -70,6 +73,76 @@ def run_and_print(self, args, **kwargs):
...
+        }
+
+        with open(config_file, "w") as cfg_file_out:
+            config.write(cfg_file_out)
```
armab: What if correct credentials already exist in the file. Do we overwrite it with "fake password"?

armab:
If `~/.st2` directory doesn't exist, will it be created by `st2 login`?

We create that dir via `curl | bash` script, but it's not single point of truth.
People can install StackStorm with configuration management or manually via `apt-get/yum install`.

For me, `st2 login -w` could be useful as a way to run it once after clean st2 installation to store the password. See example: https://github.com/StackStorm/chef-stackstorm/issues/41

Additionally, we could replace this bashery: https://github.com/StackStorm/st2-packages/blob/master/scripts/st2bootstrap-deb.sh#L219-L259 with new functionality provided in this PR.

st2client/st2client/commands/auth.py
Mierdin: What else should I write? I originally tried omitting the field altogether, but it seems that other st2 commands actually require this field, yet don't use it. Writing an obviously fake password to this field seemed like the simplest approach.

st2client/st2client/commands/auth.py
Mierdin: I don't think so, the whole point of  `st2 login` is to change this configuration, so the user should expect it to get overwritten with whatever they put into the args or getpass

Mierdin:
@armab See https://github.com/StackStorm/st2/pull/3123#issuecomment-267532280 - the method that I'm calling to generate a new token does all kinds of useful checks to make sure the directory is created if it doesn't exist. I am making sure that call is done first, so I know what files/directories will be on the system at that point.

m4dcoder:
I'm going to need more clarification from the team here. I looked at the original request. Wouldn't writing/caching the token be sufficient to meet the requirement? How come we are also rewriting the client config and saving the client credential in the config file. That seems insecure and not obvious to the user that their credential will be written to disk.

Mierdin:
@m4dcoder the reason I'm modifying the config is all about instructing other st2 commands which token file to use, via the username. Tokens are cached by username (i.e. `~/.st2/token-st2admin`), and my understanding is that the client uses the username in the configuration to determine which token file to reference (helpful when there are multiple).

So, `st2 login` modifies the configuration by writing the username specified via args. By doing this, other st2 commands know which token file to read to get the token.

Note also **by default** it writes a dummy/fake password to the config file. Other st2 commands apparently require that this field is present in the config (I tested without), but the field is not used if the token file is found, so the password doesn't need to be correct.

The user request was all about being able to quick switch between users, and this accomplishes this in the most secure way I know how, and doesn't rely on being able to modify environment variables in the calling shell (which you can't do in Python anyways). The config file in this case just acts as a pointer to which token file to use at a given time.

Hope that helps clear things up

m4dcoder:
Can you figure out why the password is needed (even a fake password) in the config file? I prefer we fix this and don't write a fake password. It's just confusing code wise and users who is looking at the config will question that anyway.

st2client/st2client/commands/auth.py
Mierdin: This is a good idea, and I used the function `_get_config_file_path` in BaseCLI to ensure this consistency. Implemented in 683cb4d

This is how it works now (note if the `--config` option is omitted, that still works and writes to the default location:

```
(virtualenv)vagrant@st2dev:~/st2/st2client$ st2 --config /tmp/st2config login st2admin --password Password1!
Logged in as st2admin
(virtualenv)vagrant@st2dev:~/st2/st2client$ cat /tmp/st2config
[credentials]
username = st2admin
password = notarealpassword
```

Mierdin:
I agree - will hunt that down next.

Mierdin:
@m4dcoder It seems I was mistaken about the password field. I must have lost track of my steps during initial testing; I can perform queries without the password field in the config:

```
(virtualenv)vagrant@st2dev:~/st2/st2client$ cat ~/.st2/config
[extrasection]
foo1 = bar1
foo2 = bar2

[credentials]
username = st2admin

(virtualenv)vagrant@st2dev:~/st2/st2client$ cat ~/.st2/token-st2admin
{"expire_timestamp": 1482014621, "token": "ee46b13293b345c4a9f7bcd955854b3e"}(virtualenv)vagrant@st2dev:~/st2/st2client$ st2 execution list
+----+------------+--------------+--------+-----------------+---------------+
| id | action.ref | context.user | status | start_timestamp | end_timestamp |
+----+------------+--------------+--------+-----------------+---------------+
+----+------------+--------------+--------+-----------------+---------------+
```

In the next few minutes I'll push another commit that removes the password functionality from this patch.

Mierdin:
@armab @m4dcoder I have made the requested changes, can I get another review at your convenience? Thanks and have a great weekend!

st2client/st2client/commands/auth.py
```diff
@@ -70,6 +73,84 @@ def run_and_print(self, args, **kwargs):
...
+        instance = self.resource(ttl=args.ttl) if args.ttl else self.resource()
+
+        cli = BaseCLIApp()
+
+        # Determine path to config file
+        try:
+            config_file = cli._get_config_file_path(args)
+        except ValueError:
+            # config file not found in args or in env, defaulting
+            config_file = expanduser('~/.st2/config')
```
m4dcoder: Can you use os.path.expanduser instead and import os above?  Also, can you use the constants at https://github.com/StackStorm/st2/blob/master/st2client/st2client/config_parser.py#L36 instead of duplicating and hard coding the path value here?

st2client/tests/unit/test_auth.py
```diff
@@ -39,6 +41,108 @@
...
+            },
+            {
+                "args": ['login', 'st2admin', '--password', 'Password1!', '-w'],
+                "expected_username": 'st2admin'
+            },
+            {
+                "args": ['--config', '/tmp/st2config', 'login', 'st2admin', '-w'],
+                "expected_username": 'st2admin'
+            }
+        ]
```
m4dcoder: Can you separate into different test methods especially that you've additional things to test for each test case below?

st2client/st2client/commands/auth.py
```diff
@@ -70,6 +73,84 @@ def run_and_print(self, args, **kwargs):
...
+        cli._cache_auth_token(token_obj=manager)
+
+        # Update existing configuration with new credentials
+        config = ConfigParser()
+        config.read(config_file)
+
+        # Modify config (and optionally populate with password)
+        if 'credentials' not in config:
+            config.add_section('credentials')
+            config['credentials'] == {}
```
m4dcoder: Why double equal sign here?  typo?

st2client/st2client/commands/auth.py
```diff
@@ -70,6 +73,84 @@ def run_and_print(self, args, **kwargs):
...
+            config_file = cli._get_config_file_path(args)
+        except ValueError:
+            # config file not found in args or in env, defaulting
+            config_file = expanduser('~/.st2/config')
+
+        logging.basicConfig(
+            level=logging.CRITICAL,
+            handlers=[logging.NullHandler]
+        )
+        cli.LOG = logging.getLogger("st2client.base")
```
m4dcoder: How come we are configuring LOG like this?  Why set the path to st2client.base?  We can't set up log like other commands and than pass the logger here?  https://github.com/StackStorm/st2/blob/master/st2client/st2client/commands/action.py#L40

st2client/st2client/commands/auth.py
Mierdin: Definitely typo. Will fix and will also make sure proper assertions are added to unit test to actually check for this value in `config`

st2client/tests/unit/test_auth.py
Mierdin: I can split it up; the reason I did it this way was to save a little bit of line count considering that the tests cases have a lot in common. Would result in a lot of redundant test code if I split it up

st2client/st2client/commands/auth.py
Mierdin: Can do

st2client/st2client/commands/auth.py
m4dcoder: Please make sure there's a unit test for this if statement since you missed this.

st2client/tests/unit/test_auth.py
m4dcoder: Yeah, i rather you duplicate here to keep the tests clear.  And you can always refactor into private methods for obvious reduction.

Kami:
On a related note (something I wanted to do myself already), It would also be nice to have `st2 whoami` or a similar command which prints the username of the user which is currently being used for API operations :)

st2client/tests/unit/test_auth.py
```diff
@@ -39,6 +41,108 @@
...
+        super(TestLogin, self).tearDown()
+
+    @mock.patch.object(
+        requests, 'post',
+        mock.MagicMock(return_value=base.FakeResponse(json.dumps({}), 200, 'OK')))
+    @mock.patch("st2client.commands.auth.ConfigParser")
+    @mock.patch("st2client.commands.auth.open")
+    @mock.patch("st2client.commands.auth.BaseCLIApp")
+    @mock.patch("st2client.commands.auth.getpass")
+    def test_login(self, mock_gp, mock_cli, mock_open, mock_cfg):
```
Kami: Nice work on the tests :+1:

st2client/tests/unit/test_auth.py
```diff
@@ -39,6 +41,108 @@
...
+class TestLogin(base.BaseCLITestCase):
+
+    def __init__(self, *args, **kwargs):
+        super(TestLogin, self).__init__(*args, **kwargs)
+        self.parser = argparse.ArgumentParser()
+        self.parser.add_argument('-t', '--token', dest='token')
+        self.parser.add_argument('--api-key', dest='api_key')
+        self.shell = shell.Shell()
+
+    def setUp(self):
```
Kami: setUp and tearDown aren't needed here since we are not overriding a default behavior.

st2client/st2client/commands/auth.py
```diff
@@ -70,6 +73,84 @@ def run_and_print(self, args, **kwargs):
...
+            config.write(cfg_file_out)
+
+        return manager
+
+    def run_and_print(self, args, **kwargs):
+        try:
+            self.run(args, **kwargs)
+            print("Logged in as %s" % args.username)
+        except Exception:
+            print("Failed to log in as %s" % args.username)
```
Kami: Should still print the error / exception, even if `self.app.client.debug:` is False otherwise it will be hard to debug.

st2client/st2client/commands/auth.py
```diff
@@ -70,6 +73,84 @@ def run_and_print(self, args, **kwargs):
...
+
+        with open(config_file, "w") as cfg_file_out:
+            config.write(cfg_file_out)
+
+        return manager
+
+    def run_and_print(self, args, **kwargs):
+        try:
+            self.run(args, **kwargs)
+            print("Logged in as %s" % args.username)
```
Kami: Minor style thing - it's usually a good idea to always wrap format string arguments in parenthesis - `print("Logged in as %s" % (args.username))`.

This way it's impossible for thing to break in case someone adds another format value parameter, but forgets to add parenthesis. We have lint, etc. which should, in theory, catch this, but I've still seen it happen before.

st2client/st2client/commands/auth.py
Kami: Yeah, I need to check what is going on - in theory this logging stuff here shouldn't be needed.

st2client/tests/unit/test_auth.py
Mierdin: You're right; total habit, since I historically have done some stuff here, but as-is, they're pretty pointless

st2client/tests/unit/test_auth.py
Mierdin: Thanks!

st2client/st2client/commands/auth.py
Mierdin: FWIW, here's the error I get when I don't set up logging. It seemed strange that the log was being defined on the instance and not with the typical `LOG = logging.getLogger(__name__)` but figured it was for some reason.

```
(virtualenv)vagrant@st2dev:~/st2/st2client$ python setup.py install && st2 --debug login st2admin -p Password1!
# -------- begin 140500329551632 request ----------
curl -X POST -H  'Connection: keep-alive' -H  'Accept-Encoding: gzip, deflate' -H  'Accept: */*' -H  'User-Agent: python-requests/2.11.1' -H  'content-type: application/json' -H  'Content-Length: 2' -H  'Authorization: Basic c3QyYWRtaW46UGFzc3dvcmQxIQ==' --data-binary '{}' http://127.0.0.1:9100/tokens
# -------- begin 140500329551632 response ----------
{
    "user": "st2admin",
    "token": "11d044175f514b7dbd5b3b15557f09fe",
    "expiry": "2016-12-21T18:37:53.506004Z",
    "id": "58597a8132ed3534e5d0a586",
    "metadata": {}
}
# -------- end 140500329551632 response ------------

Failed to log in as st2admin
ERROR: 'NoneType' object has no attribute 'debug'

CLI settings:
----------------
Config file path: /home/vagrant/.st2/config
Client settings:
----------------
ST2_BASE_URL: http://127.0.0.1
ST2_AUTH_URL: http://127.0.0.1:9100
ST2_API_URL: http://127.0.0.1:9101/v1
ST2_AUTH_TOKEN: None

Proxy settings:
---------------
HTTP_PROXY:
HTTPS_PROXY:

  File "/home/vagrant/st2/virtualenv/lib/python2.7/site-packages/st2client-2.2.dev0-py2.7.egg/st2client/shell.py", line 282, in run
    args.func(args)
  File "/home/vagrant/st2/virtualenv/lib/python2.7/site-packages/st2client-2.2.dev0-py2.7.egg/st2client/commands/auth.py", line 146, in run_and_print
    self.run(args, **kwargs)
  File "/home/vagrant/st2/virtualenv/lib/python2.7/site-packages/st2client-2.2.dev0-py2.7.egg/st2client/commands/auth.py", line 125, in run
    cli._cache_auth_token(token_obj=manager)
  File "/home/vagrant/st2/virtualenv/lib/python2.7/site-packages/st2client-2.2.dev0-py2.7.egg/st2client/base.py", line 321, in _cache_auth_token
    self.LOG.debug('Token has been cached in "%s"' % (cached_token_path))
AttributeError: 'NoneType' object has no attribute 'debug'
```

st2client/st2client/commands/auth.py
Mierdin: Fixed in dafb4e1 - the problem wasn't that the conditional wasn't being tested, but that I wasn't checking for the presence of a dict as the value for `config['credentials']`. I added an assertion to check for this

st2client/st2client/commands/auth.py
Mierdin: Good idea - fixed in 2a4d38d

st2client/st2client/commands/auth.py
Mierdin: Added in 9ca6a40

st2client/tests/unit/test_auth.py
Mierdin: Okay I split up these tests and added a function `_login_common_assertions` to save a little bit of linecount on common things between the unit tests. Committed in d45b9af

st2client/st2client/commands/auth.py
```diff
@@ -70,6 +73,84 @@ def run_and_print(self, args, **kwargs):
...
+        config = ConfigParser()
+        config.read(config_file)
+
+        # Modify config (and optionally populate with password)
+        if 'credentials' not in config:
+            config.add_section('credentials')
+            config['credentials'] = {}
+        config['credentials']['username'] = args.username
+        if args.write_password:
+            config['credentials']['password'] = args.password
```
armab: :+1:

st2client/st2client/commands/auth.py
```diff
@@ -70,6 +73,84 @@ def run_and_print(self, args, **kwargs):
...
+
+        logging.basicConfig(
+            level=logging.CRITICAL,
+            handlers=[logging.NullHandler]
+        )
+        cli.LOG = logging.getLogger("st2client.base")
+
+        # Retrieve token
+        manager = self.manager.create(instance, auth=(args.username, args.password), **kwargs)
+        cli._cache_auth_token(token_obj=manager)
```
armab: :+1: 

I see there is a logic in [`st2client/base.py/_cache_auth_token()`](https://github.com/StackStorm/st2/blob/ed6fdb4043647b766551c25f365782a58abce019/st2client/st2client/base.py#L280-L281
) to create the config dir if it doesn't exist.

@Mierdin But just to be sure can you double check and confirm that config dir is created with `st2 login` when installing pure packages without `curl | bash` script?

st2client/st2client/commands/auth.py
```diff
@@ -70,6 +73,84 @@ def run_and_print(self, args, **kwargs):
...
+        self.parser.add_argument('username',
+                                 help='Name of the user to authenticate.')
+
+        self.parser.add_argument('-p', '--password', dest='password',
+                                 help='Password for the user. If password is not provided, '
+                                      'it will be prompted.')
+        self.parser.add_argument('-l', '--ttl', type=int, dest='ttl', default=None,
+                                 help='The life span of the token in seconds. '
+                                      'Max TTL configured by the admin supersedes this.')
+        self.parser.add_argument('-w', '--write-real-password', action='store_true', default=False,
```
armab: Additionally, don't forget to update Pull Request description (first message) with new syntax/logic, since we don't save fake password: `notarealpassword` anymore

st2client/st2client/commands/auth.py
Kami: I just had a look at the code and I believe simply adding new `LOG` attribute to the class as class level-variable should work.

Basically what happens on the `st2client.base.Shell` class.

One option, which might be better would also be to refactor those utility methods for retrieving config, etc. in a separate module so you don't need to instantiate `BaseCLIApp` class here, but this might require too much refactoring so it's not necessary.

Mierdin: @Kami Right - and changing all references of `self.LOG` to `LOG` as well, right?

Kami: Yeah, this should work - just need to verify --debug logging, etc still works :)

st2client/st2client/commands/auth.py
Mierdin: Updated the description, and changed the argument in https://github.com/StackStorm/st2/pull/3123/commits/7a2ce3952dd39f3d00475ca338ccb4de3213e983

st2client/st2client/commands/auth.py
Mierdin: So after a few months I wanted to make sure I understood what we discussed. I think I misread your comment slightly in December, and assumed you had recommended I place `LOG` as a module variable, instead of a class variable (thus my comment about removing `self.LOG`).

Anyways, I ended up fixing this in 079c89b26 by putting the `LOG` definition within the `BaseCLIApp` class definition, and left all of the `self.LOG` definitions as they were, so they could continue to work. Seemed like the best approach based on this thread, and the [debug output](https://gist.github.com/Mierdin/3d6431a2f999a8281682ce2a7e4be0c5) seems to work just fine with this approach. Let me know if anything should change

Mierdin:
@Kami Okay - introduced the `st2 whoami` command in 5f8771963, would love to hear your thoughts on the implementation

st2client/st2client/commands/auth.py
```diff
@@ -70,6 +73,121 @@ def run_and_print(self, args, **kwargs):
...
+        # Update existing configuration with new credentials
+        config = ConfigParser()
+        config.read(config_file)
+
+        # Modify config (and optionally populate with password)
+        if 'credentials' not in config:
+            config.add_section('credentials')
+            config['credentials'] = {}
+        config['credentials']['username'] = args.username
+        if args.write_password:
```
dzimine: scenario: configure .st2/config with `user1` with password, rewriting it with `st2 login user2`, user get rewriten, password remains from user1, token expires, user1 no longer works.

Mierdin: Good point - so perhaps I can default to deleting the "password" field (or if `--write-password` is provided, overwrite)?

Mierdin: @dzimine I took a stab at this in da93e1b07 - let me know what you think. [Seems to work well now](https://gist.github.com/Mierdin/2ca2ee3a58239a5b25bac62ccc79535e) for this case

CHANGELOG.rst
```diff
@@ -49,6 +49,10 @@ in development
   put the workflow in a PAUSED state in mistral. (improvement)
 * Add support for evaluating jinja expressions in mistral workflow definition where yaql
   expressions are typically accepted. (improvement)
+# Added support for `st2 login` and `st2 whoami` commands. These add some additional functionality
```
Kami: Minor thing - s/#/*/

st2client/st2client/commands/auth.py
```diff
@@ -70,6 +73,124 @@ def run_and_print(self, args, **kwargs):
...
+        try:
+            self.run(args, **kwargs)
+            print("Logged in as %s" % (args.username))
+        except Exception as e:
+            print("Failed to log in as %s: %s" % (args.username, str(e)))
+            if self.app.client.debug:
+                raise
+
+
+class WhoamiCommand(resource.ResourceCommand):
```
Kami: Just to confirm - does this follow the same logic as other commands - aka it takes `ST2_CONFIG_PATH` environment variable, `--st2-config (or whatever the argument name is), --user, etc. and other things into account when determining the username?

If not, it would be a good idea to do that otherwise it might be confusing to the user.

Kami:
Overall, nice work and LGTM :+1: 

Only "blocker" for me right now which needs some clarification is https://github.com/StackStorm/st2/pull/3123/files#r99538549

Besides that, LGTM.

st2client/st2client/commands/auth.py
Kami: To clarify, that are the options I had in mind - https://docs.stackstorm.com/reference/cli.html#configuration-file.

Mierdin: @Kami Okay, I tested both the ST2_CONFIG_PATH` and `--config-file` options, and other than running into [this bug](
https://github.com/StackStorm/st2/issues/3186) (unrelated to this patch) everything seems to work using a totally different config path.

It should be noted that this tool doesn't actually create the configuration file - as with other commands, if the file doesn't exist at the referenced path, it will raise an exception. I did this to be as predictable compared to other st2 commands as possible.

CHANGELOG.rst
Mierdin: Doh! Fixed in https://github.com/StackStorm/st2/pull/3123/commits/3da2b42f05de1fc1c922a5a5c01c8bbb3088feff

PR #3124, opened 2016-12-17
Feature Request - Actionalias help API

nzlosh:
With the action-alias processing on the server side since 2.1, it is possible to send strings to the API and have it return success or failure when matching against an action alias.  It is still a client side task to display action alias help, which requires fetching the list of action-aliases' and processing them by applying the same logic as the server side to action alias `format` field.

To fully capitalise on server side action alias processing, it would be useful to have an API that would return available help to the client side without the need for client side processing.  I submit this PR as a candidate for the implementation and for your comment/feedback on the idea.

The API supports

- filtering by help string (regular expression)
- selecting a pack (string comparison with available packs)
- limit the size of the resulting help strings
- offset in the list or returned help strings (to allow pagination via the API)

Here is an example API response
`curl -H 'Content-Type: application/json' localhost:9101/v1/actionalias/help -d '{"filtering": "^[a-z]", "pack": "packs", "limit":2, "offset": 4 }'`
```
{
    "packs": [
        {
            "display": "pack show <pack>",
            "description": "Show information about the pack from StackStorm Exchange."
        }
    ]
}
```
and with no limit/offset
`curl -H 'Content-Type: application/json' -XPOST localhost:9101/v1/actionalias/help -d '{"filtering": "^[a-z]", "pack": "packs"}'`
```
{
    "packs": [
        {
            "display": "pack get <pack>",
            "description": "Get information about installed StackStorm pack."
        },
        {
            "display": "pack install <pack>[,<pack>]",
            "description": "Install/upgrade StackStorm packs."
        },
        {
            "display": "pack install <gitUrl>[,<gitUrl>]",
            "description": "Install/upgrade StackStorm packs."
        },
        {
            "display": "pack search <query>",
            "description": "Search for packs in StackStorm Exchange and other directories."
        },
        {
            "display": "pack show <pack>",
            "description": "Show information about the pack from StackStorm Exchange."
        }
    ]
}
```


st2common/st2common/models/api/action.py
```diff
@@ -682,3 +683,51 @@ def to_model(cls, aliasexecution):
...
+    API model used to display action-alias help API endpoint.
+    """
+    model = None
+
+    schema = {
+        "title": "ActionAliasHelpAPI",
+        "description": "ActionAliasHelpAPI.",
+        "type": "object",
+        "properties": {
+            "filtering": {
```
lakshmi-kannan: Please use a noun form - ``filter``.

st2common/st2common/util/actionalias_helpstring.py
```diff
@@ -0,0 +1,73 @@
...
+        if not alias.enabled:
+            continue
+        # Skip packs which don't explicitly match the requested pack.
+        if pack != alias.pack and pack != "":
+            continue
+        for format_ in alias.formats:
+            display, _ = normalise_alias_format_string(format_)
+            if display:
+                # Skip over help strings until the requested offset is reached.
+                if cnt < offset:
```
lakshmi-kannan: alias.formats[offset:] in line 53 would've worked, right? Of course you'll need try catch for index out of bounds.

st2common/st2common/models/api/action.py
```diff
@@ -682,3 +683,51 @@ def to_model(cls, aliasexecution):
...
+                "type": "string",
+                "description": "Find help strings containing keywords.",
```
lakshmi-kannan: Is your intentional to support only one string? Why is this not a list?

st2common/st2common/util/actionalias_helpstring.py
```diff
@@ -0,0 +1,73 @@
...
+                    matches[alias.pack] = []
+                matches[alias.pack].append({
+                    "display": display,
+                    "description": alias.description
+                })
+                cnt += 1
+            # Return the list of help strings when we reach the requested limit.
+            if (cnt >= offset + limit) and limit > 0:
+                return matches
+    return matches
```
lakshmi-kannan: There needs to be an indicator to the user about the pagination. So if limit=10 and total items=50, we have to tell the user that there are 50 items. So this function should not only return matches but also total count IMO.

lakshmi-kannan:
I think my code comments are minor. I am sure those can be addressed. I think the PR in general is in the right direction and I am super happy with this entire PR. Enough context in the description describing the change, good tests and pretty much no hand holding from us :). You rock!

st2common/st2common/models/api/action.py
nzlosh: The intention was to have the string represent a regular expression pattern which provides very powerful search features.

You can get a similar effect as a list by using regex pipes. e.g. `item1|item2|item3`.  There is a limitation however, the API doesn't allow for passing regex flags with the request, so using case insensitive search isn't currently possible.

That said, I'll use whatever search mechanism is recommended by the st2 devs.

st2common/st2common/util/actionalias_helpstring.py
nzlosh: Sorry, I didn't make the intention of offset very clear in my initial post.  I'll try to explain it better here.  This feature was added after discussing the idea of a help api with @emedvedev.

The intention of offset was to allow paging through a large list of help strings.  If an st2 installation had a large number of packs and a client wanted to "scroll" through the list of helpstrings,  a client could make consecutive calls using the offset/limit pair.  E.g. if there were 122 help strings and the display interface was large enough for 20 items, the first call would contain `offset:0, limit: 20`, the second `offset:20, limit:20` etc.

Unfortunately, applying offset at line 53 wouldn't produce the result mentioned above because at that point in the code, `alias.formats[offset:]` is being applied to a single action alias and not the entire list of help strings.

st2common/st2common/util/actionalias_helpstring.py
nzlosh: OK, that sounds reasonable, I'll include the total count in the return too.

nzlosh:
Glad to hear it.  I'll add more unit tests in next PR to cover the last ~28%.  Thanks for the code review and the feedback.

emedvedev:
We should just set the default flags to make the search case-insensitive then, imo

emedvedev:
Carlos rocks!

nzlosh:
Hello,
I've pushed the changes as per the discussions in this issue.  I don't understand why CircleCI is failing, is it something I've done?

enykeev:
Seems like you need to rebase your PR to include https://github.com/StackStorm/st2/pull/3155

nzlosh:
Thanks for the hint @enykeev.  I've rebased.

nzlosh:
Things have gone quiet, is the PR still of interest?

emedvedev:
Definitely! We just need a bit more time to get back to it :) Sorry for the delay!

lakshmi-kannan:
LGTM. I am divided on a changelog entry - reason is I still consider this beta and not ready for public consumption. In any case, we can always add a changelog entry later if needed. So good to merge from my side.

st2api/st2api/controllers/v1/actionalias.py
```diff
@@ -106,6 +109,28 @@ def match(self, action_alias_match_api, **kwargs):
             pecan.abort(http_client.BAD_REQUEST, str(e))
             return [self._match_tuple_to_dict(match) for match in e.matches]
 
+    @request_user_has_permission(permission_type=PermissionType.ACTION_ALIAS_HELP)
```
Kami: A bit late to the game, but perhaps we could / should just re-use existing list or get permission here.

In any case, not a big deal :)

st2common/st2common/util/actionalias_helpstring.py
```diff
@@ -0,0 +1,77 @@
...
+    :return: A list of aliases help strings.
+    :rtype: ``list`` of ``list``
+    """
+    matches = {}
+    count = 0
+    if not (isinstance(limit, int) and isinstance(offset, int)):
+        raise TypeError
+    for alias in aliases:
+        # Skip disable aliases.
+        if not alias.enabled:
```
Kami: It seems like we could off-load some of that filtering and perform it on the database level (at the very least the enabled and pack name).

Not a blocker for now, but it would be great to do that in the future.

st2common/st2common/util/actionalias_helpstring.py
```diff
@@ -0,0 +1,77 @@
...
+    :param offset: The offset in the list to start returning help strings.
+    :type  limit: ``integer``
+
+    :return: A list of aliases help strings.
+    :rtype: ``list`` of ``list``
+    """
+    matches = {}
+    count = 0
+    if not (isinstance(limit, int) and isinstance(offset, int)):
+        raise TypeError
```
Kami: Please throw a more specific error message - e.g. `raise TypeError('limit or offset argument is not an integer')`.

Kami:
Sorry for the delay (I know the PR has already been merged) - from a high level, change looks good to me :+1: 

I added some comments for some small things which can be improved in the future, but nothing major and blocking.

st2api/st2api/controllers/v1/actionalias.py
```diff
@@ -106,6 +109,28 @@ def match(self, action_alias_match_api, **kwargs):
...
+        """
+            Get available help strings for action aliases.
+
+            Handles requests:
+                POST /actionalias/help
+        """
+        filter_ = action_alias_help_api.filter
+        pack = action_alias_help_api.pack
+        limit = action_alias_help_api.limit
+        offset = action_alias_help_api.offset
```
enykeev: The rest of the API endpoints expect `pack`, `limit` and `offset` to be sent as query parameters. We need to follow the same logic here too. Frankly, `filter` also looks like query parameter, but we would need to ensure we can pass regex string in query or came up with a way to escape it.

st2api/st2api/controllers/v1/actionalias.py
```diff
@@ -106,6 +109,28 @@ def match(self, action_alias_match_api, **kwargs):
             pecan.abort(http_client.BAD_REQUEST, str(e))
             return [self._match_tuple_to_dict(match) for match in e.matches]
 
+    @request_user_has_permission(permission_type=PermissionType.ACTION_ALIAS_HELP)
+    @jsexpose(arg_types=[str, str, int], body_cls=ActionAliasHelpAPI,
+    status_code=http_client.ACCEPTED)
```
enykeev:
```
10.2.3 202 Accepted

The request has been accepted for processing, but the processing has not been completed. The request might or might not eventually be acted upon, as it might be disallowed when processing actually takes place. There is no facility for re-sending a status code from an asynchronous operation such as this.

The 202 response is intentionally non-committal. Its purpose is to allow a server to accept a request for some other process (perhaps a batch-oriented process that is only run once per day) without requiring that the user agent's connection to the server persist until the process is completed. The entity returned with this response SHOULD include an indication of the request's current status and either a pointer to a status monitor or some estimate of when the user can expect the request to be fulfilled.
```

Status code 202 should only be used in cases when server can not answer within 30 second response window. In this case,  `200 OK` should be used instead.

enykeev: Also, `arg_types` in @jsexpose defines types for query parameters for the controller. It should not be used without query parameters defined in arguments list and I'm quite surprised the endpoint works at all.

enykeev:
Updated version of the endpoint is available at https://github.com/StackStorm/st2/pull/2727/commits/520b9fef57123fedc275c6c261711e189baca99e. We would still need to see what can be done to leverage database-level querying.

Also, endpoint output is still up for discussion. In other endpoints, we don't return the number of available documents in the response but rather use headers for that purpose (see https://github.com/StackStorm/st2/blob/01e7da684d26e6c4995c2ad96421d60d69b828b3/st2api/st2api/controllers/resource.py#L186-L188).

PR #3125, opened 2016-12-17
Fix cancel policy

m4dcoder:
Cancels mistral workflow when a subtask has been cancelled due to concurrency policy. Previously, the mistral workflow is not cancelled and orphaned as running in mistral.

### DIFF
--- a/st2actions/st2actions/policies/concurrency.py
+++ b/st2actions/st2actions/policies/concurrency.py
@@ -61,7 +61,8 @@ class ConcurrencyApplicator(BaseConcurrencyApplicator):
                       count, target.action, self._policy_ref, action)
             status = self._get_status_for_policy_action(action=self.policy_action)
 
-        # Update the status in the database but do not publish.
+        # Update the status in the database and publish status so the appropriate runner can
+        # cancel the execution appropriately.
         target = action_service.update_status(target, status, publish=True)
 
         return target
--- a/st2actions/st2actions/policies/concurrency.py
+++ b/st2actions/st2actions/policies/concurrency.py
@@ -62,7 +62,7 @@ class ConcurrencyApplicator(BaseConcurrencyApplicator):
             status = self._get_status_for_policy_action(action=self.policy_action)
 
         # Update the status in the database but do not publish.
-        target = action_service.update_status(target, status, publish=False)
+        target = action_service.update_status(target, status, publish=True)
 
         return target
 
--- a/st2actions/tests/unit/policies/test_concurrency_by_attr.py
+++ b/st2actions/tests/unit/policies/test_concurrency_by_attr.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 
 import mock
+from mock import call
 
 import st2common
 from st2common.bootstrap.policiesregistrar import register_policy_types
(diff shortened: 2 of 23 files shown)


st2actions/st2actions/policies/concurrency.py
```diff
@@ -62,7 +62,7 @@ def _apply_before(self, target):
             status = self._get_status_for_policy_action(action=self.policy_action)
 
         # Update the status in the database but do not publish.
```
andrew-regan: `# Update the status in the database and publish.`

Applies to next file as well.

Kami:
Good catch and thanks - will have a look shortly.

In the future it would make it a bit easier to review if cc4285381a291bb375d30d7b6b47fa73888a05fd was part of a separate PR / branch (now two changes are intertwined which makes it a bit hard to review).

st2actions/st2actions/policies/concurrency.py
Kami: Yeah, looks like we either need to update the comment or the code below :)

st2actions/tests/unit/policies/test_concurrency_by_attr.py
```diff
@@ -136,8 +143,14 @@ def test_over_threshold_cancel_executions(self):
         # Execution is expected to be delayed since concurrency threshold is reached.
         liveaction = LiveActionDB(action='wolfpack.action-2', parameters={'actionstr': 'fu'})
         liveaction, _ = action_service.request(liveaction)
-        delayed = LiveAction.get_by_id(str(liveaction.id))
-        self.assertEqual(delayed.status, action_constants.LIVEACTION_STATUS_CANCELED)
+
+        # Assert the canceling state is being published.
+        calls = [call(liveaction, action_constants.LIVEACTION_STATUS_CANCELING)]
+        LiveActionPublisher.publish_state.assert_has_calls(calls)
```
Kami: Thanks for adding this additional asserts.

st2common/st2common/policies/concurrency.py
```diff
@@ -35,6 +35,6 @@ def _get_status_for_policy_action(self, action):
         if action == 'delay':
             status = action_constants.LIVEACTION_STATUS_DELAYED
         elif action == 'cancel':
-            status = action_constants.LIVEACTION_STATUS_CANCELED
+            status = action_constants.LIVEACTION_STATUS_CANCELING
```
Kami: Just trying to understand this and thinking out loud - how will this work and affect non-mistral tasks?

For mistral actions this will eventually be changed to canceled once Mistral updates internal task status and once we query it, right? How about non-mistral actions?

Kami:
Added some comments / questions.

Let's please also update changelog.

contrib/runners/mistral_v2/tests/unit/test_mistral_v2_auth.py
```diff
@@ -30,7 +30,7 @@
 
 from mistral_v2 import MistralRunner
 from st2common.bootstrap import actionsregistrar
-import st2common.bootstrap.runnersregistrar as runners_registrar
+from st2common.bootstrap import runnersregistrar
```
Kami: Just could leave the alias "as runners_registrar" there :trollface:

st2common/st2common/policies/concurrency.py
m4dcoder: The st2action worker https://github.com/StackStorm/st2/blob/master/st2actions/st2actions/worker.py will process the execution and let the appropriate runner do the cancel. For most non-mistral actions, there's no specific cancel procedures and it just inherits and runs from the base runner cancel method which is empty. https://github.com/StackStorm/st2/blob/master/st2common/st2common/runners/base.py#L106 After the cancel is run, the status of execution will be changed to CANCELED. https://github.com/StackStorm/st2/blob/master/st2actions/st2actions/container/base.py#L154

m4dcoder:
@Kami concerns and input addressed

st2common/st2common/policies/concurrency.py
Kami: Thanks for clarifying - looks good to me then.

andrew-regan:
Picked up this fix locally and verified. Thanks!

Issue #3128, opened 2016-12-19
Pack icon.png does not display


skjbulcher:
I'm having trouble getting my pack's [icon.png](https://drive.google.com/file/d/0B3QFylZ1CuvsYTVJR1dzODR5YjA/view?usp=sharing) to display.

Running st2 version 2.1.1

**permissions** `/opt/stackstorm/packs`
```
root@stackstorm:/opt/stackstorm/packs# ls -rlth
total 24K
drwxrwxr-x 5 root st2packs 4.0K Dec 16 21:36 packs
drwxrwxr-x 4 root st2packs 4.0K Dec 16 21:36 linux
drwxrwxr-x 5 root st2packs 4.0K Dec 16 21:36 default
drwxrwxr-x 5 root st2packs 4.0K Dec 16 21:36 chatops
drwxrwxr-x 4 root st2packs 4.0K Dec 19 19:09 core
drwxrwxr-x 4 root st2packs 4.0K Dec 19 19:29 grade_us
```

**permissions** `/opt/stackstorm/packs/grade_us`
```
root@stackstorm:/opt/stackstorm/packs/grade_us# ls -rlth
total 28K
-rwxrwxr-x 1 root st2packs   22 Dec 19 19:29 requirements.txt
-rwxrwxr-x 1 root st2packs  206 Dec 19 19:29 pack.yaml
-rwxrwxr-x 1 root st2packs 3.8K Dec 19 19:29 icon.png
-rwxrwxr-x 1 root st2packs   52 Dec 19 19:29 grade_us.yaml.example
-rwxrwxr-x 1 root st2packs   94 Dec 19 19:29 config.schema.yaml
drwxrwxr-x 3 root st2packs 4.0K Dec 19 19:29 actions
```

**output** `st2api.log` after loading URL https://<mystackstorm>.com/api/v1/packs/views/file/grade_us/icon.png?x-auth-token=TOKEN
```
root@stackstorm:/var/log/st2# tail -f -n 0 st2api.log
2016-12-19 19:45:06,990 140160140703856 INFO hooks [-] 1b150dd3-fbaf-485c-9e06-afae8e3e5433 -  GET /v1/packs/views/file/grade_us/icon.png with filters={} (remote_addr='127.0.0.1',method='GET',filters={},request_id='1b150dd3-fbaf-485c-9e06-afae8e3e5433',path='/v1/packs/views/file/grade_us/icon.png')
[...]

PR #3131, opened 2016-12-20
Fix pack files and file controller when pack name != pack ref and similar + other various pack management related bug fixes

Kami:
This pull request fixes issue with pack files and file controller returning 404 (pack file not found) when pack name didn't match pack ref.

Before this change, the controller incorrectly used pack.name attribute when determining the pack on-disk directory name. This wouldn't work when pack ref != pack name or when pack name contains special characters.

This pull request is not related just to the ref change, but to the whole "getting rid of pack directory name" concept which I was very skeptical about (at least doing it in a rushed manner we did - even if we didn't make the ref change issues like that would still appear because originally pack.download action "magically" inferred directory name from pack name).

I personally never liked using directory name for UID, but we didn't have anything better and I was worried exactly about issues like this happening - sadly a lot more code relies on this old behavior of pack directory name having special meaning and acting as pack UID.

Fixes issue reported in https://github.com/StackStorm/st2/issues/3128.

## TODO

- [x] Refactor code (use pack_ref instead of pack_name to be explicit and obvious)
- [x] Tests
- [x] Changelog entry


Kami:
While working on this I found some more issues related to the new pack management stuff :/

Another issue is that `pack_db.files` field now also contains all the git metadata files (files in `.git` directory) which breaks pack files controller.

I'm also working on a fix for that.

Kami:
While working and testing those changes I found some other bugs related to pack management which I also fixed in this PR:

1. I updated a list of exclude patterns for files which are not included in `pack_db.files` attribute. Not pack directories are git repositories which contain .git/ directory. Before this change, this directory was added to `pack_db.files` field which was causing some pack view controllers to fail.

2. Improved binary file detection so binary files are not included in "pack files" API controller response which caused response serialization to fail and API to include error 500.

Kami:
Another "limitation" exposed by this changes is that those controllers will only work if pack reference is the same as directory name on disk

This is the default behavior for packs installed using `st2 pack install` and it only affects packs which are manually placed into `/opt/stackstorm/packs/`.

In fact, this limitation is not really new, it was the same before this new pack management changes, but just something we need to be more explicit about and point out somewhere in the docs.

PR #3132, opened 2016-12-20
Include original traceback when re-throwing an exception inside the sensor wrapper

Kami:
When we re-throw a more user-friendly exception we didn't include the original stack trace which would make some issues harder to debug (it didn't include additional stack trace which would point user to the offending line in the sensor file and similar).

Caught by @enykeev.

### DIFF
--- a/st2reactor/st2reactor/container/sensor_wrapper.py
+++ b/st2reactor/st2reactor/container/sensor_wrapper.py
@@ -269,7 +269,8 @@ class SensorWrapper(object):
             msg = ('Failed to load sensor class from file "%s" (sensor file most likely doesn\'t '
                    'exist or contains invalid syntax): %s' % (self._file_path, str(e)))
             msg += '\n\n' + tb_msg
-            raise ValueError(msg)
+            exc_cls = type(e)
+            raise exc_cls(msg)
 
         if not sensor_class:
             raise ValueError('Sensor module is missing a class with name "%s"' %
--- a/st2reactor/st2reactor/container/sensor_wrapper.py
+++ b/st2reactor/st2reactor/container/sensor_wrapper.py
@@ -17,6 +17,7 @@ import os
 import json
 import atexit
 import argparse
+import traceback
 
 from oslo_config import cfg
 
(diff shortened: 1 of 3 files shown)


lakshmi-kannan:
lgtm.

st2reactor/st2reactor/container/sensor_wrapper.py
```diff
@@ -264,8 +265,10 @@ def _get_sensor_instance(self):
                                                         file_path=self._file_path,
                                                         class_name=self._class_name)
         except Exception as e:
-            msg = ('Failed to load sensor class from file "%s"'
-                   ' (sensor file most likely doesn\'t exist): %s' % (self._file_path, str(e)))
+            tb_msg = traceback.format_exc()
+            msg = ('Failed to load sensor class from file "%s" (sensor file most likely doesn\'t '
+                   'exist or contains invalid syntax): %s' % (self._file_path, str(e)))
+            msg += '\n\n' + tb_msg
             raise ValueError(msg)
```
lakshmi-kannan: While we are on this, do you want to raise a specific exception than ValueError?

Kami: We could just re-throw original exception type or simply Exception.

Kami: Also added tests - 486c4af185c052481e617d522dcff702e0d191f8, d463290b48aded5f1c973db1ce325a9d334de393.

st2reactor/tests/unit/test_sensor_wrapper.py
```diff
@@ -107,9 +107,31 @@ def test_sensor_init_fails_file_doesnt_exist(self):
...
+        self.assertRaisesRegexp(NameError, expected_msg, SensorWrapper,
+                                pack='core', file_path=file_path,
+                                class_name='TestSensor',
+                                trigger_types=trigger_types,
+                                parent_args=parent_args)
+
+        # Verify error message also contains traceback
+        try:
+            SensorWrapper(pack='core', file_path=file_path, class_name='TestSensor',
+                          trigger_types=trigger_types, parent_args=parent_args)
```
lakshmi-kannan: You want a self.fail('Should have thrown an exception'). Otherwise, tests might still pass on an edge case :P

Kami: Yep, forgot to add else :)

Kami: Actually, there is else 3 lines below :P

lakshmi-kannan: Missed it baby one more time :D

PR #3140, opened 2016-12-22
Fix retry policy

m4dcoder:
Retry policy is not supported for action executions under workflows. Let specific workflow engine handles retries where applicable. Currently, action execution will retry on failure and timeout but workflow fails immediately.

### DIFF
--- a/st2actions/st2actions/policies/retry.py
+++ b/st2actions/st2actions/policies/retry.py
@@ -63,13 +63,8 @@ class ExecutionRetryPolicyApplicator(ResourcePolicyApplicator):
         self.max_retry_count = max_retry_count
         self.delay = delay or 0
 
-    def apply_before(self, target):
-        # Nothing to do here
-        target = super(ExecutionRetryPolicyApplicator, self).apply_before(target=target)
-        return target
-
     def apply_after(self, target):
-        target = super(ExecutionRetryPolicyApplicator, self).apply_before(target=target)
+        target = super(ExecutionRetryPolicyApplicator, self).apply_after(target=target)
 
         live_action_db = target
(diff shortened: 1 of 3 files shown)


Kami:
Thanks - LGTM!

Let's please also make sure to document this in the policies section of documentation.

st2actions/st2actions/policies/retry.py
```diff
@@ -63,15 +63,20 @@ def __init__(self, policy_ref, policy_type, retry_on, max_retry_count=2, delay=0
...
         self.delay = delay or 0
 
-    def apply_before(self, target):
-        # Nothing to do here
-        target = super(ExecutionRetryPolicyApplicator, self).apply_before(target=target)
-        return target
-
     def apply_after(self, target):
-        target = super(ExecutionRetryPolicyApplicator, self).apply_before(target=target)
+        target = super(ExecutionRetryPolicyApplicator, self).apply_after(target=target)
```
Kami: Good catch - I guess none of the existing tests caught it because it uses a default parent implementation which doesn't do anything right now (so mostly just out of luck and it would break if we changed the parent behavior).

st2actions/st2actions/policies/retry.py
```diff
@@ -117,6 +122,16 @@ def apply_after(self, target):
 
         return target
 
+    def _get_live_action_parent(self, live_action_db):
```
Kami: Perhaps `_is_live_action_part_of_workflow_action` might be a bit more clear and obvious (it would make it more generic and hide the abstraction that child workflow actions have `parent` attribute in context), but it's not a big deal.

Kami:
LGTM :+1:

PR #3142, opened 2016-12-27
Introduce validation for trigger parameters and payload for non-system (user-defined) triggers

Kami:
This pull request builds on top of a great foundation started at https://github.com/StackStorm/st2/pull/3094 and introduces two new things.

## 1. Trigger parameter validation for non-system (user-defined) triggers

This validation is performed when creating a rule. Previously we only performed this validation for system triggers (webhook, timer, etc.).

For backward compatibility reasons, this validation is only performed for user-defined triggers if ``system.validate_trigger_parameters`` config option is enabled (it's disabled by default) and if a trigger type object in a database contains ``parameters_schema`` field.

## 2. Trigger payload validation for system and user-defined triggers.

This validation is performed when dispatching a trigger inside a sensor and when dispatching a trigger using a webhook.

For backward compatibility reasons, this validation is only performed if ``system.validate_trigger_payload`` config option is enabled (it's disabled by default) and if a trigger type object in a database contains ``payload_schema`` field.

Before this change we didn't perform any validation for triggers dispatched into the system. This is not just bad from "UX" perspective (not failing fast, hard to debug what is going on, etc.), but also because of potential security and DDoS issues (we didn't validate the payload so user could potentially dispatch really large payloads or similar).

Note: Since most of the system-triggers are only dispatched directly by StackStorm right now, issue with the validation would indicate a programmer error (trigger payload schema is out of sync or trigger is including attributes which shouldn't be included in the payload) and something which needs to be fixed on our side.

## Background and reasoning

The aim of those two changes is to improve the user experience and follow the "fail fast" and "optional typing" mantra - they allow users and developers to identify and fix issues earlier and thus save them time and frustration. "Optional typing" approach is followed because ``parameters_schema`` and ``payload_schema`` fields are optional and validation is only performed if those fields are available.

They also follow the standard approach of shipping a new backward incompatible feature which is making sure that the feature is opt-in by default (no breaking changes in non-major releases and giving the user time to adapt their code).

In a future major release (after we announce this change of course, etc.) we should set both of those options to enabled by default, but still allow users to opt-out. In fact, the idea at the beginning was to perform validation for all the parameters and payloads, but we simply didn't have the time to implement it.

**Note: I propose including this change in v2.3.0, but not in v2.2.2 to better follow semver and be on the safe side.**

## TODO

- [x] Tests for validation when paramters schema and payload schema fields are not available, but validation is enabled
- [x] Tests for creating a rule for user-defined trigger
- [x] Upgrade notes entry


Kami:
@userlocalhost This PR is now finished and ready for review from your and st2 side :)

lakshmi-kannan:
I think validating payload is indeed the right behavior and should be turned on by default for user triggers as well. One question I have is whether we should stop a sensor on first instance of invalid trigger payload. Letting the sensor run and log warn a bunch of times is kinda bad when people aren't expecting this - They would just see that the process is running and won't bother to even look at logs. Besides this, the change LGTM. Also, in next major release or so, we should plan for actively deprecating old behaviors. The code gets really confusing with those if blocks pretty soon.

Kami:
> Also, in next major release or so, we should plan for actively deprecating old behaviors. The code gets really confusing with those if blocks pretty soon.

Agreed. We have a lot of notes in the code of stuff which can get removed.

We just need to make sure we document it in upgrade notes in case someone still relies on some behavior which has already been deprecated a long time ago.

Kami:
> I feel that validate_trigger_parameters and validate_trigger_payload could be refactored to use a common function since they look similar. But that's minor.

Yeah, I agree - I was looking into it, but refactoring it didn't seem to buy much, it just made it slightly confusing.

Having said that, I can still refactor it.

PR #3144, opened 2017-01-03
Make st2-run-pack-tests install dependencies only on virtualenv creation

dzimine:
A pack developer needs a faster way to run tests.

When running without `-x` (the first time), setup everything. 
On subsequent runs, provide `-x`, activate the virtualenv, skip dependencies, and just run tests.

Sample output:
```
$ /opt/stackstorm/st2/bin/st2-run-pack-tests  -p /opt/stackstorm/packs/csv/
Running tests for pack: csv
Creating virtualenv in /tmp/st2-pack-tests-virtualenvs/csv...
New python executable in /tmp/st2-pack-tests-virtualenvs/csv/bin/python
Installing setuptools, pip, wheel...done.
Requirement already up-to-date: pip in ./st2-pack-tests-virtualenvs/csv/lib/python2.7/site-packages
Installing dependencies from st2 repository...
Installing global pack test dependencies...
Running tests...
test_run_comma_delimiter (test_action_parse.ParseCSVActionTestCase) ... ok
test_run_pipe_delimiter (test_action_parse.ParseCSVActionTestCase) ... ok

Ran 2 tests in 0.352s

OK

$ /opt/stackstorm/st2/bin/st2-run-pack-tests  -x -p /opt/stackstorm/packs/csv/
Running tests for pack: csv
Activating virtualenv in /tmp/st2-pack-tests-virtualenvs/csv...
Running tests...
test_run_comma_delimiter (test_action_parse.ParseCSVActionTestCase) ... ok
test_run_pipe_delimiter (test_action_parse.ParseCSVActionTestCase) ... ok

Ran 2 tests in 0.323s

OK

### CODE REVIEW

**Comment by Kami** [comment]
To give some background - we are intentionally installing dependency each time because they can change and we always want to use the latest and correct version (that's also important for CI of course).

The problem of "installing dependencies is slow" is "solved" on the CI by caching the dependencies on each build. Same thing should also happen by default locally - it will still take a while because cache needs to be read, etc. but it will be much faster than without using a cache.

I'm fine with this change as long as we add a flag or similar which causes all the dependencies to be (re)installed and update affected docs and CI.

**Comment by Kami** [comment]
Also, congrats on your first PR in 2017 :champagne:

**Comment by dzimine** [comment]
Before I redo it with an extra flag:

is there a better direction for a pack developer writing unit tests? 
The current one (or at least what I figured, as we don't have it documented yet) will require a user to 1) `git clone http://.../st2.git /tmp/st2` 2) `export ST2_REPO_PATH=/tmp/st2`
If not too hard, I'd rather spare them from this setup step, assuming they do the tests on machine with st2 already installing. Give me a hint I'll explore.

**Comment by Kami** [comment]
If st2 is already installed, they just need to ensure st2 components are in PYTHONPATH.

This basically means they can just use `/opt/stackstorm/bin/st2-run-pack-tests` binary and it should already work (it's using binary from stackstorm virtualenv so all the components should already be in PYTHONPATH).

**Comment by Kami** [comment]
And if the written above is not totally clear from the documentation we should improve it and add it there :)

**Comment by dzimine** [comment]
> This basically means they can just use /opt/stackstorm/bin/st2-run-pack-tests binary and it should already work 

Either I miss something, or the script is not built this way: 
1) it relies on `ST2_REPO_PATH` to install st2 requirements into the virtualenv; 
2) `$PYTHONPATH` is not set on my system after installing st2 and I afraid it's not supposed to, so [this assumption](https://github.com/StackStorm/st2/blob/master/st2common/bin/st2-run-pack-tests#L148-L149) may only hold true for st2 dev env, not for pack dev env (think st2vagrant). 

> And if the written above is not totally clear from the documentation we should improve it and add it there :)

Bummer: This page is not visible in TOC https://docs.stackstorm.com/development/pack_testing.html and [I can't figure why](https://raw.githubusercontent.com/StackStorm/st2docs/master/docs/source/development/index.rst). I was not able to find the page until today. 

I suggest we take this as an opportunity and drive to fully documented pack unit-testing. CC @Mierdin

**Comment by Kami** [comment]
Yeah, sorry, I need to clarify it better. I was talking about `PYTHONPATH` used by st2 components (`st2-run-pack-tests`) and the one used by the pack.

So you have multiple options:

1.Let the script create virtualenv for you - the best approach for dev environment and testing, imo, since it doesn't mess up with StackStorm pack virtualenv (you don't really want to install test dependencies into StackStorm pack virtualenv). That's also the default option used by the `st2-run-pack-tests` script.
2. Create virtualenv manually yourself - this way you can control caching, etc. That's what we do on CI to speed things up.
3. Re-use StackStorm pack virtualenv (`/opt/stackstorm/virtualenvs/<pack name>`). As mentioned above, I'm personally not a fan of this approach since it's messing with pack's virtualenv and also installing test dependencies there.

Re TOC - yeah I noticed recently, it probably got away while we were re-organizing it for BWC...

**Comment by m4dcoder** [comment]
Just jumping in to provide some friendly feedback and not trying to be nosy. I think the test script should create a separate venv for testing per pack. If test venv doesn't exists, just create and install deps. If test venv exists, use the existing venv but update the deps.  Always let user know where the test venv is located and how they can recreate it manually in case it gets messed up.  As for deps to st2common and st2tests, add it to the PYTHONPATH somehow (disclaimer, haven't tested this last part so just feedback based on code).

**Comment by Kami** [comment]
> I think the test script should create a separate venv for testing per pack. If test venv doesn't exists, just create and install deps. If test venv exists, use the existing venv but update the deps. Always let user know where the test venv is located and how they can recreate it manually in case it gets messed up. 

That's exactly the default behavior right now :)

> Also, may be helpful to add a CLI args to let user disable updating venv if exists to run tests faster if they know req hasn't changed.

That's basically what @dzimine wanted to achieve with his PR, but I described my reasoning above why this might be problematic, at least for the default behavior.

>As for deps to st2common and st2tests, add it to the PYTHONPATH somehow (disclaimer, haven't tested this last part so just feedback based on code).

Yes, as described above, either use st2 components which are installed using packages (use that PYTHONPATH) or clone git repo and add components from there to PYTHONPATH - that's what we do in CI right now to make sure we always use latest version of StackStorm.

**Comment by Mierdin** [comment]
Here are my thoughts after spending .5 day on this or so.

It seems there's not a consistent developer experience between pack development and st2 development in general. St2 devs are encouraged to use `make requirements && launchdev.sh`, but for pack development, it seems that the st2 libraries must be installed on the system itself. Following [the instructions for testing a pack](https://docs.stackstorm.com/development/pack_testing.html did not result in success for me. There are a few things missing there:

- I think it should be stated clearly what needs to be installed, and where. Even after running `make requirements`, and then following the documentation, it was still asking for stackstorm libs (i.e. `st2tests`, `st2common`, etc). If the "canonical" approach is to test within st2vagrant (and as a result, stackstorm should **actually be installed at a system level**, we should at least state it (though I still disagree with this constraint because again it seems to be contrary to the mobility offered via the launchdev.sh approach)
- There's no mention of ST2_REPO_PATH anywhere in that doc - @dzimine had to tell me about it

There are a number of virtualenvs in use here, and I would think that it would be easier to simply ensure that all of the stackstorm libs (`st2common`, etc) are installed there alongside all of the other pack dependencies when `/tmp/st2-pack-tests-virtualenvs/<pack>` is set up.

**Comment by dzimine** [comment]
We miss `st2tests` module on the "normal user's" machine, along with it's dependencies.

Everything else can be made working by a script. 

What are our options here? 
1) getting st2tests from github by a script, seting PYTHONPATH, and installing test-requirements? Dirty hack IMO. On the other hand, users will have to do exactly this as of today, anyways.
2) installing st2tests with st2 and adding what's missed (like psutils) to `PACK_TEST_PYTHON_DEPENDENCIES`? 
3) moving all pack dev dependencies to st2sdk as @m4dcoder brought up earlier - what ARE "all" dependencies? But what ARE they exactly? And again, installing st2sdk with st2 along with test-requirements?
4) What else? What do you guys think?

This needs a fix: unittesting flow for st2 user (not st2 developer) is not working ATM, and to get unit tests we shall have no friction.

**Comment by lakshmi-kannan** [comment]
I think we are trying to fix a lot of things in one go. The only we should have right now is to enable easy pack testing in st2vagrant environment (as @dzimine has pointed out). st2vagrant comes with st2 installed but it doesn't contain st2tests from st2 repo as it's not part of packaging. I think we have two options 1. Push a st2sdk with artifacts from st2tests and st2common to pypi (Likely 2 weeks of work easily looking at st2common. We've always wanted to do this for various reasons.) 2. Just git clone st2repo and setup virtualenv and PYTHONPATH. I lean towards (2). This is probably 2 days of work and will not change much. There are several ways to optimize this path like @Kami has pointed out. I'd go with two virtualenvs created by the script A. st2 parent virtualenv where st2 artifacts are available and B. Pack's test virtualenv (pack dependencies + pack test dependencies). PYTHONPATH now has to include both (A) and (B). git clone could be git pull if we already know there is a cloned dir. Those are optimizations. Let's revisit to see if the slowness is indeed a real issue.

**Comment by Kami** [comment]
> We miss st2tests module on the "normal user's" machine, along with it's dependencies.

Now that StackStorm components are installed inside a virtual environment, I see no more issues with including `st2tests` package as part of the official StackStorm packages / distribution.

This could only become problematic in case user, for some reason, wants to use latest version from git (we would just need to document how to handle that scenario by putting package from git as a first entry of the PYTHONPATH).

**Comment by dzimine** [comment]
Ok file updated now, bringing it back to life. 
I am so irritated by waiting for python dependencies to update.
With a new parameter the existing CI won't be affected. 

This will help folks (and keep me sane) right now.
```
Usage: ./st2-run-pack-tests [-x -s] -p <path to pack>
  -x : Do not create virtualenv, e.g. when running in existing one.
  -j : Just run tests (use previously installed dependencies
       and virtualenv if any. For subsequent test runs.
```

**Comment by Kami** [comment]
@dzimine On a related note - when I just want to run a single test or tests from a single file and pack dependencies are already installed in the currently used virtualenv, I just use this command:

```bash
PYTHONPATH=packs/time/actions:$PYTHONPATH nosetests -v --nocapture packs/time/tests/test_action_get_week_boundaries.py 
```

In short, as mentioned above, you just need to add pack actions directory to PYTHONPATH (or sensors for sensors).

Kami:
But as far as `-j` flag goes and your change - I'm fine with that change and will have a look at the code change later.

Kami:
The tests are failing, it looks like you need to sync your branch with the latest master to pick up the Mistral changes.

Kami:
Made a minor change so we only activate virtualenv if it doesn't exist - dd577c13dd60ae0511a01e1f7c1216ee18829e12.

Before your change, specifying `-x` flag would mean it doesn't create and not activate the virtualenv, but it would just try to install dependencies inside the current PYTHONPATH (or virtual environment if one is activated).

With this change, we also support running pack tests outside a pack-specific environment. I do this often - I tend to have one virtual environment which contains dependencies for all st2 components + various packs and I just want to re-use that (aka don't want to activate and update pack specific virtual environment).

I think this should be sufficient for v2.2.0 and we can make additional improvements and fixes in v2.2.1 if needed / necessary.

Will also open separate PR with docs, but we can also do this post v2.2.0 release since merging docs in v2.2 branch will cause build to happen and to docs to be automatically updated so it's not a blocker for v2.2.0.

PR #3145, opened 2017-01-04
Fix mistral API call to get task result

m4dcoder:
Mistral list function for task execution no longer returns result. This patch iterates through the list and make additional calls to get task result. Merge this patch just before st2's mistral fork syncs with upstream.

### DIFF
--- a/contrib/runners/mistral_v2/query/mistral_v2.py
+++ b/contrib/runners/mistral_v2/query/mistral_v2.py
@@ -102,7 +102,10 @@ class MistralResultsQuerier(Querier):
         :type exec_id: ``str``
         :rtype: ``list``
         """
-        wf_tasks = tasks.TaskManager(self._client).list(workflow_execution_id=exec_id)
+        wf_tasks = [
+            tasks.TaskManager(self._client).get(task.id)
+            for task in tasks.TaskManager(self._client).list(workflow_execution_id=exec_id)
+        ]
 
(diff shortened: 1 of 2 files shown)


Kami:
LGTM.

So from the description it sounds like this only affects newer version of Mistral.

contrib/runners/mistral_v2/query/mistral_v2.py
```diff
@@ -102,7 +102,10 @@ def _get_workflow_tasks(self, exec_id):
         :type exec_id: ``str``
         :rtype: ``list``
         """
-        wf_tasks = tasks.TaskManager(self._client).list(workflow_execution_id=exec_id)
+        wf_tasks = [
+            tasks.TaskManager(self._client).get(task.id)
```
m4dcoder: This returns exactly the same dict as individual item of the old list function.  The new list function no longer returns the result attribute.

m4dcoder:
@Kami yes, this only affects newer version of mistral

Kami:
@m4dcoder Thanks for clarifying.

"Batch" mode as before would be nice to have to save on HTTP requests, but I assume Mistral people had good reasons to do what they did so I won't poke into it any further :)

m4dcoder:
Yeah batch mode would have been nice. But the list function was returning way too much data if it includes the result. That's why the mistral team made this change to the list function.

m4dcoder:
@Kami btw, you should poke when it doesn't make sense. Your insight/input/feedback is always appreciated. And mistral peps or any developers are not always correct.

PR #3151, opened 2017-01-06
Fix duplication action execution in concurrency policies

m4dcoder:
While evaluating the concurrency policies, the live action was published again and led to duplicate executions. This patch fixes the concurrency policies to only publish if the live action state is changed to cancelling.

### DIFF
--- a/st2actions/st2actions/policies/concurrency.py
+++ b/st2actions/st2actions/policies/concurrency.py
@@ -61,9 +61,10 @@ class ConcurrencyApplicator(BaseConcurrencyApplicator):
                       count, target.action, self._policy_ref, action)
             status = self._get_status_for_policy_action(action=self.policy_action)
 
-        # Update the status in the database and publish status so the appropriate runner can
-        # cancel the execution appropriately.
-        target = action_service.update_status(target, status, publish=True)
+        # Update the status in the database. Publish status for cancellation so the
+        # appropriate runner can cancel the execution.
+        publish = (status == action_constants.LIVEACTION_STATUS_CANCELING)
+        target = action_service.update_status(target, status, publish=publish)
 
         return target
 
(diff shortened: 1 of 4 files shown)


st2actions/st2actions/policies/concurrency.py
```diff
@@ -61,9 +61,10 @@ def _apply_before(self, target):
                       count, target.action, self._policy_ref, action)
             status = self._get_status_for_policy_action(action=self.policy_action)
 
-        # Update the status in the database and publish status so the appropriate runner can
-        # cancel the execution appropriately.
-        target = action_service.update_status(target, status, publish=True)
+        # Update the status in the database. Publish status for cancellation so the
+        # appropriate runner can cancel the execution.
+        publish = (status == action_constants.LIVEACTION_STATUS_CANCELING)
```
Kami: Heh, so `publish` was intentionally set to `False` before previous change. That makes sense now :)

st2actions/tests/unit/policies/test_concurrency_by_attr.py
```diff
@@ -99,25 +99,24 @@ def test_over_threshold_delay_executions(self):
...
         scheduled = [item for item in LiveAction.get_all() if item.status in SCHEDULED_STATES]
         self.assertEqual(len(scheduled), policy_db.parameters['threshold'])
+        self.assertEqual(len(scheduled), runner.MockActionRunner.run.call_count)
 
         # Execution is expected to be delayed since concurrency threshold is reached.
         liveaction = LiveActionDB(action='wolfpack.action-1', parameters={'actionstr': 'fu'})
         liveaction, _ = action_service.request(liveaction)
 
-        # Assert the delayed state is being published.
-        calls = [call(liveaction, action_constants.LIVEACTION_STATUS_DELAYED)]
```
Kami: Perhaps some more "explicit" tests (explicitly asserting that call count is zero for non-cancel publish method call) would make it more obvious and harder in the future to make the same mistake.

A comment in the tests would also be nice.

Kami:
Good catch - LGTM.

But as mentioned in the comment, it would be good to add some more explicit tests for where status should not be publish to avoid same mistake accidentally happening again in the future.

m4dcoder:
@Kami addressed your comments. I added more checks in the tests and more comments.

Kami:
Thanks for adding those additional assertions :+1:

PR #3154, opened 2017-01-10
Fix mistral upstream integration

m4dcoder:
Mistral added jinja evaluation and breaks st2kv rendering via jinja. This patch wraps any jinja as raw if there is any st2kv reference.

### DIFF
--- a/st2common/st2common/util/workflow/mistral.py
+++ b/st2common/st2common/util/workflow/mistral.py
@@ -61,6 +61,11 @@ SPEC_TYPES = {
     }
 }
 
+JINJA_WITH_ST2KV = '{{st2kv\..*?|.*?\sst2kv\..*?}}'
+JINJA_WITH_ST2KV_PTRN = re.compile(JINJA_WITH_ST2KV)
+JINJA_WITH_LOCAL_CTX = '{{.*?_\..*?}}'
+JINJA_WITH_LOCAL_CTX_PTRN = re.compile(JINJA_WITH_LOCAL_CTX)
+
 
 def _parse_cmd_and_input(cmd_str):
     cmd_matcher = CMD_PTRN.search(cmd_str)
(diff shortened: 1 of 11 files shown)


st2common/st2common/util/workflow/mistral.py
```diff
@@ -61,6 +61,11 @@
     }
 }
 
+JINJA_WITH_ST2KV = '{{.*?st2kv\..*?}}'
```
Kami: Only whitespace can be located before `st2kv\.` string, right?

If so, we should probably make it more restrictive using `\s*` or similar, or am I missing something?

st2tests/st2tests/fixtures/generic/workflows/wf_has_jinja_st2kv_post_xform.yaml
```diff
@@ -0,0 +1,41 @@
...
+            action: st2.action
+            input:
+                ref: wolfpack.a1
+        task_with_inputs:
+            action: st2.action
+            input:
+                ref: wolfpack.a2
+                parameters:
+                    strtype: '{% raw %}{{ st2kv.system.var1 }}{% endraw %}'
+                    inttype: '{% raw %}{{ st2kv.user.var2 }}{% endraw %}'
```
Kami: Just curious / confirming that user context (the user which executed the action) is indeed correctly passed down to Mistral.

Kami:
Added a couple of comments, besides that, LGTM.

st2common/st2common/util/workflow/mistral.py
m4dcoder: Sure. I'll make this more restrictive.

st2tests/st2tests/fixtures/generic/workflows/wf_has_jinja_st2kv_post_xform.yaml
m4dcoder: User scoped variables are not supported. Waiting for another PR that @lakshmi-kannan is working on. I'll remove from test here to avoid confusion.

m4dcoder:
@Kami your new comments are addressed

dzimine:
+1. Please do a doc PR to explain the rules on when Jinga is resolved and when not.

PR #3155, opened 2017-01-11
Fix usage of mistral client in mistral runner

m4dcoder:
Fix how mistral client and resource managers are being used in the mistral runner. Authentication has changed in the mistral client. Fix unit test accordingly.


Kami:
LGTM.

e2e tests are failing - I assume that's because of this change and they will start passing after it's merged? Or is it something else?

PR #3158, opened 2017-01-12
Clean up pyc files

m4dcoder:


Kami:
LGTM, good catch :+1:

Kami:
Looks like tests are failing, need to re-run them and make sure they pass (this way we can be sure none of the tests actually utilized one of those pyc files).

m4dcoder:
I didn't rebase and so the commit was based off an older commit with mistral related problems that had been fixed.

Issue #3157, opened 2017-01-12
Getting an error while passing a single rbridge_id value as input


poorani612:
Getting an error while passing a single rbridge_id value as input. rbridge_id type is array and legnth of the array is 1.  I am not getting any error while running the same action with single element in Web UI.

```
st2 run network_essentials.modify_arp_nd_aging_ve mgmt_ip=10.20.34.46 arp_aging_type=arp_aging vlan_id=10 rbridge_id=52
ERROR: 400 Client Error: Bad Request
MESSAGE: 52 is not of type 'array', 'null' for url: http://127.0.0.1:9101/v1/executions
```

Mierdin:
I am noting that @poorani612 also mentioned that this behavior is new to 2.1 - this worked in 2.0.1. I'm looking over the `st2client` code today to see if I can identify what changed

Mierdin:
@poorani612 The parsing of these parameters occurs in https://github.com/StackStorm/st2/commits/master/st2client/st2client/commands/action.py and this file hasn't been changed since before the 2.0.1 release

Have you been able to confirm that this worked in a previous release?

poorani612:
Everyone upgraded to 2.1 and not able to verify in 2.0 :-(

Mierdin:
@poorani612 no problem - based on my findings I'm guessing this has always existed. Regardless, thanks for the find, and watch #3159 for the fix

PR #3159, opened 2017-01-13
Add handler for single-element (int) arrays without brackets on CLI

Mierdin:
> Due to some CI issues, I am closing this PR and re-opening as a PR from the main repo, instead of my fork: https://github.com/StackStorm/st2/pull/3161

If a user wants to use an action with an array parameter, and passes a single int to this parameter thusly:

	st2 run foopack.fooaction arrayparam=51

the API throws an error because the client renders this as an int, and not an array of length 1.

This was because the transform_array function in actions.py first tries to
load the given value as JSON. In the example above, 51 is a valid int in JSON,
so this results in an error because of a mismatch between what the API expected
(array) and what was actually sent (int).

Note that this isn't a problem if the single element in this array is a string; in that case, the exception causes the function to split on commas, and since there aren't any, it returns a list containing a single string member. However, if it's an int, for some reason, json.loads works fine and this bug appears.

To get around this, the user **could** run something like:

	st2 run foopack.fooaction arrayparam=[51]

However, this is a bit unintuitive. Because of the action metadata, we already know
that this **should** be a list, so we can force it to be. So, this commit adds a
step to check if the value is an int, and if so, force it to be a Python
list. This allows the user to also use the more intutive former example.

Closes https://github.com/StackStorm/st2/issues/3157


Kami:
LGTM :+1:

st2client/st2client/commands/action.py
```diff
@@ -503,6 +503,26 @@ def transform_object(value):
...
+            #
+            # Normally, json.loads can handle this, but not if the single element is
+            # an int.
+            #
+            # We want to try to handle this first, so it doesn't get accidentally
+            # sent to the API as an int, instead of an array of single-element int.
+            try:
+                # Force this to be a list containing the single int, then
+                # cast the whole thing to string so json.loads can handle it
+                value = str([int(value)])
```
Kami: Another way we could improve handling of the array parameters in the future -

Users can also specify type for each array parameter inside an action metadata (`items: type: int` for example) and this information is available here so we could also leverage that to cast each array item to the correct type.

If nothing else we should probably just open a feature request issue for that change :)

Mierdin: That's a good idea! I opened https://github.com/StackStorm/st2/issues/3160 to track this.

Mierdin:
Due to some CI issues, I am closing this PR and re-opening as a PR from the main repo, instead of my fork: https://github.com/StackStorm/st2/pull/3161

PR #3161, opened 2017-01-13
Add handler for single-element (int) arrays without brackets on CLI 

Mierdin:
> Continuation of closed PR https://github.com/StackStorm/st2/pull/3159

If a user wants to use an action with an array parameter, and passes a single int to this parameter thusly:

	st2 run foopack.fooaction arrayparam=51

the API throws an error because the client renders this as an int, and not an array of length 1.

This was because the transform_array function in actions.py first tries to
load the given value as JSON. In the example above, 51 is a valid int in JSON,
so this results in an error because of a mismatch between what the API expected
(array) and what was actually sent (int).

Note that this isn't a problem if the single element in this array is a string; in that case, the exception causes the function to split on commas, and since there aren't any, it returns a list containing a single string member. However, if it's an int, for some reason, json.loads works fine and this bug appears.

To get around this, the user **could** run something like:

	st2 run foopack.fooaction arrayparam=[51]

However, this is a bit unintuitive. Because of the action metadata, we already know
that this **should** be a list, so we can force it to be. So, this commit adds a
step to check if the value is an int, and if so, force it to be a Python
list. This allows the user to also use the more intutive former example.

Closes https://github.com/StackStorm/st2/issues/3157


CHANGELOG.rst
```diff
@@ -3,7 +3,8 @@ Changelog
 
 in development
 --------------
-
+* Fixed issue where passing a single integer member for an array parameter for an action would
```
Kami: Minor thing - we tend to follow chronological order and put latest changes at the end :)

Mierdin:
@Kami Whoops! Sorry about that, thought it was the other way around. Fixed

PR #3163, opened 2017-01-20
Add option to authenticate with mistral using st2 auth token

m4dcoder:
Mistral can now authenticate using st2 auth token. This patch adds an option to configure st2 to pass the auth token when invoking the mistral client.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -49,6 +49,8 @@ in development
   (new feature)
 * Linux file watch sensor is now disabled by default. To enable it, set ``enabled: true`` in
   ``/opt/stackstorm/packs/linux/sensors/file_watch_sensor.yaml``
+* Add option to use st2 auth token to authenticate with mistral. Previously, OpenStack Keystone
+  is recommended for securing mistral. (improvement)
 
 2.2.0 - February 27, 2017
 -------------------------
--- a/contrib/runners/mistral_v2/mistral_v2.py
+++ b/contrib/runners/mistral_v2/mistral_v2.py
@@ -19,17 +19,15 @@ import uuid
 import retrying
 import six
 import yaml
-from mistralclient.api import client as mistral
 from oslo_config import cfg
 
 from st2common.runners.base import AsyncActionRunner
 from st2common.constants.action import LIVEACTION_STATUS_RUNNING
 from st2common import log as logging
 from st2common.models.api.notification import NotificationsHelper
+from st2common.util import api as api_utils
+from st2common.util import url as url_utils
 from st2common.util.workflow import mistral as utils
-from st2common.util.url import get_url_without_trailing_slash
-from st2common.util.api import get_full_public_api_url
-from st2common.util.api import get_mistral_api_url
 
 
 LOG = logging.getLogger(__name__)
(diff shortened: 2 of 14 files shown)


Kami:
High level question / feedback (please correct me if I'm wrong / missing something):

Auth tokens are short lived, what happens when they expire? Wouldn't it make more sense to support API keys which are not short lived and designed for "long lived" integrations?

m4dcoder:
It's a good question. If the token expired, then the execution fails because mistral can't validate the token and auth fails. I use token currently because 1) we already generate a temporary token on start of workflow execution so mistral can auth w/ st2 to invoke task execution. 2) When using mistral client to troubleshoot, users can acquire an auth token like they do with the st2 client, export it to ST2_AUTH_TOKEN, and mistral client will use that token. I'm not oppose to using an API key but we also haven't been receiving complains of token being too short lived.  And security wise, I rather we use a short live token for each transaction than a long live API key.

lakshmi-kannan:
@m4dcoder A workflow might run for a duration > expiry time for a token (which again is configurable). I think the st2 <-> mistral integration should just use API keys or have logic to refresh tokens in the context of a workflow. I am personally leaning towards API key myself. It makes the surface area for bugs lesser at some increased risk in terms of security.

m4dcoder:
If we are using st2 API key to authenticate at mistral, this will require the st2auth controller to open an endpoint to validate API key.  Also, I would have to keep both st2 token and API key validation in the st2 auth plugin at mistral because the mistral client should be able to authenticate w/ ST2_AUTH_TOKEN.

CHANGELOG.rst
```diff
@@ -49,6 +49,8 @@ in development
   (new feature)
 * Linux file watch sensor is now disabled by default. To enable it, set ``enabled: true`` in
   ``/opt/stackstorm/packs/linux/sensors/file_watch_sensor.yaml``
+* Add option to use st2 auth token to authenticate with mistral. Previously, OpenStack Keystone
+  is recommended for securing mistral. (improvement)
```
lakshmi-kannan: * was the recommended option to secure mistral.

contrib/runners/mistral_v2/mistral_v2.py
```diff
@@ -69,6 +59,15 @@ def pre_run(self):
...
         self._skip_notify_tasks = self.runner_parameters.get('skip_notify', [])
 
+    def _get_client(self, ctx=None):
+        auth_token = (
+            ctx['auth_token']
+            if ctx and 'auth_token' in ctx and ctx.get('auth_token')
+            else getattr(self.auth_token, 'token', None)
+        )
+
+        return utils.get_client(self._base_url, auth_token=auth_token)
```
lakshmi-kannan: So if an auth token is in the context, this will use auth_token over API key? When is a legit case when we would use token vs a key? Is it when API key is not supplied in config?

contrib/runners/mistral_v2/mistral_v2.py
```diff
@@ -400,4 +407,6 @@ def _build_mistral_context(parent, current):
             else:
                 context['mistral'] = current
 
+        context['mistral']['auth_token'] = getattr(self.auth_token, 'token', None)
```
lakshmi-kannan: Can you just create this member only if token exists?

contrib/runners/mistral_v2/mistral_v2.py
m4dcoder: In the get_client method, if cfg.CONF.mistral.st2_api_key is defined, it'll use the API key over the auth_token supplied here.  If user wants to use token vs. key, the st2_api_key option needs to be remove from the config file.

contrib/runners/mistral_v2/mistral_v2.py
m4dcoder: Yeah I guess. May need to check if it breaks anything first...

m4dcoder:
@lakshmi-kannan @Kami this is the corresponding change to st2mistral. There's client and server side plugin for Mistral to use StackStorm for auth provider. https://github.com/StackStorm/st2mistral/pull/27

lakshmi-kannan:
Closing. Please reopen if you are actively working on this @m4dcoder.

PR #3164, opened 2017-01-20
Deprecate {{system.}} and {{user.}} notations

lakshmi-kannan:
With this PR, we are deprecating {{user.}} and {{system.}} notations for accessing items from datastore. We only support {{st2kv.user.}} and {{st2kv.system.}} notations. 

## TODO

* [x] Deprecate {{user.}}
* [X] Fix docs (https://github.com/StackStorm/st2docs/pull/367)
* [X] Deprecation notice in Upgrade Notes (https://github.com/StackStorm/st2docs/pull/367)
* [x] Changelog

### DIFF
--- a/st2common/st2common/models/api/keyvalue.py
+++ b/st2common/st2common/models/api/keyvalue.py
@@ -134,7 +134,7 @@ class KeyValuePairAPI(BaseAPI):
             doc['value'] = symmetric_decrypt(KeyValuePairAPI.crypto_key, model.value)
             encrypted = False
 
-        scope = getattr(model, 'scope', SYSTEM_SCOPE)
+        scope = getattr(model, 'scope', FULL_SYSTEM_SCOPE)
         if scope:
             doc['scope'] = scope
 
--- a/st2common/st2common/models/api/keyvalue.py
+++ b/st2common/st2common/models/api/keyvalue.py
@@ -21,7 +21,6 @@ from oslo_config import cfg
 import six
 
 from st2common.constants.keyvalue import FULL_SYSTEM_SCOPE, FULL_USER_SCOPE, ALLOWED_SCOPES
-from st2common.constants.keyvalue import USER_SCOPE
 from st2common.exceptions.keyvalue import CryptoKeyNotSetupException, InvalidScopeException
 from st2common.log import logging
 from st2common.util import isotime
@@ -139,7 +138,7 @@ class KeyValuePairAPI(BaseAPI):
             doc['scope'] = scope
 
         key = doc.get('name', None)
-        if (scope == USER_SCOPE or scope == FULL_USER_SCOPE) and key:
+        if (scope == FULL_USER_SCOPE) and key:
             doc['user'] = UserKeyReference.get_user(key)
             doc['name'] = UserKeyReference.get_name(key)
 
--- a/st2common/st2common/models/api/keyvalue.py
+++ b/st2common/st2common/models/api/keyvalue.py
@@ -21,7 +21,7 @@ from oslo_config import cfg
 import six
 
 from st2common.constants.keyvalue import FULL_SYSTEM_SCOPE, FULL_USER_SCOPE, ALLOWED_SCOPES
-from st2common.constants.keyvalue import SYSTEM_SCOPE, USER_SCOPE
+from st2common.constants.keyvalue import USER_SCOPE
 from st2common.exceptions.keyvalue import CryptoKeyNotSetupException, InvalidScopeException
 from st2common.log import logging
 from st2common.util import isotime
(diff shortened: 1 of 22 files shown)


Kami:
LGTM :+1: 

If not done already, it's probably also a good idea to grep all our repos for "{{user.}}" and "{{system}}" strings, it's probably still used in some examples.

st2common/st2common/models/api/keyvalue.py
```diff
@@ -134,12 +133,12 @@ def from_model(cls, model, mask_secrets=True):
             doc['value'] = symmetric_decrypt(KeyValuePairAPI.crypto_key, model.value)
             encrypted = False
 
-        scope = getattr(model, 'scope', SYSTEM_SCOPE)
```
Kami: Wait, this should still be `user` and `system`, we only use ``st2kv`` notation in templates, right?

Or am I missing something?

st2common/st2common/services/datastore.py
```diff
@@ -61,7 +61,7 @@ def list_values(self, local=True, prefix=None):
         kvps = client.keys.get_all(prefix=key_prefix)
         return kvps
 
-    def get_value(self, name, local=True, scope=SYSTEM_SCOPE, decrypt=False):
+    def get_value(self, name, local=True, scope=FULL_SYSTEM_SCOPE, decrypt=False):
```
Kami: Same here and in many other modified files - valid values for "scope" option should still be "user" and "system".

I assume you just used sed to replace those occurrences in all the files or similar, but I don't think we should replace the actual valid values for "scope" option.

One thing is the template notation (e.g. "{{st2kv.user.foo}}") and the other thing is valid valued for "scope" (e.g. "user") option.

Kami:
Added some comments - looks like you also (I assume unintentionally) replaced valid values for "scope" argument / option which I don't think should be changed.

As mentioned above, one thing is template notation and the other thing is a datastore item option. Those values don't need to be (and I don't think they should be) the same since they don't represent the same thing.

st2common/st2common/util/param.py
```diff
@@ -76,8 +76,7 @@ def _create_graph(action_context):
     Creates a generic directed graph for depencency tree and fills it with basic context variables
     '''
     G = nx.DiGraph()
-    G.add_node(SYSTEM_SCOPE, value=KeyValueLookup(scope=SYSTEM_SCOPE))
-    system_keyvalue_context = {SYSTEM_SCOPE: KeyValueLookup(scope=SYSTEM_SCOPE)}
+    system_keyvalue_context = {SYSTEM_SCOPE: KeyValueLookup(scope=FULL_SYSTEM_SCOPE)}
     G.add_node(DATASTORE_PARENT_SCOPE, value=system_keyvalue_context)
```
m4dcoder: How come we are not supporting FULL_USER_SCOPE here?

lakshmi-kannan:
<internal> https://stackstorm.slack.com/archives/stackstorm/p1485440942023760 </internal>. Question is do we want API for datastore to be ?scope=system or ?scope=st2kv.system since jinja template is {{st2kv.system.foo}}. I wrote DB migration models in v2.1 so scope in DB will be st2kv.system. I personally perfer to consistently use st2kv.system but kami says we should use st2kv.system for jinja and everything else should be just ``system``. I think this will be confusing when people dump the DB contents. So please vote so I can make progress on this PR. cc: @mstone, @m4dcoder, @dzimine

st2common/st2common/util/param.py
lakshmi-kannan: I assume you mean why we don't do FULL_SYSTEM_SCOPE: KeyValueLookup(scope=FULL_SYSTEM_SCOPE)? 

This has to do with Jinja and this graph rendering code. If you see line 80, DATASTORE_PARENT_SCOPE is `st2kv`. So the key for the dictionary inside should just be `system` and not `st2kv.system`. In object notations, what we want is

```
st2kv: { 
   system: KeyValueLookup(scope=st2kv.system)
}
```
When we want to support user scoped variables resolution (which will be done after this PR gets merged), we'd have

```
st2kv: { 
   system: KeyValueLookup(scope=st2kv.system)
   user: KeyValueLookup(scope=st2kv.user)
}
```

m4dcoder:
I think I prefer scope=system.  The user already knows they're accessing the st2 datastore API, there's no need to specify it's st2kv again.  If we want consistency, can we support both scope=system and scope=st2kv.system?

Kami:
What's our plan for this?

I would like us to finish it and include it in v2.2.0.

As mentioned above - I'm quite opinionated about it and I think the API parameter values and DB model values should still be user / system (without st2kv prefix).

Because it seems like I missed the DB model change in the past, we might need to support both values for backward compatibility reasons.

lakshmi-kannan:
> Because it seems like I missed the DB model change in the past, we might need to support both values for backward compatibility reasons.

We cannot go back now and we've to support both.

Kami:
@lakshmi-kannan Alright.

Are you going to finish this PR (make scope=user/system change for API and model and also support st2kv.system/user on the model for backward compatibility)?

Kami:
From Slack:

> I will close the PR and redo the changes from scractch to just deprecate {{system}} and {{user}} jinja.

Yeah, sounds good to me.

PR #3177, opened 2017-02-01
Add integration test for jinja evaluation in mistral

m4dcoder:
Add an integration test to test jinja expressions in mistral workflows.


m4dcoder:
Yes, it passes. Thanks.

lakshmi-kannan:
Thanks for these examples. Super useful. 👍

PR #3180, opened 2017-02-03
[NO MERGE] RBAC refactoring

enykeev:
Tried to clean it up a little bit and bring `pecan.request` closer to the action.

### DIFF
--- a/st2common/st2common/rbac/utils.py
+++ b/st2common/st2common/rbac/utils.py
@@ -19,6 +19,7 @@ RBAC related utility functions.
 
 import six
 
+import pecan
 from oslo_config import cfg
 
 from st2common.exceptions.rbac import AccessDeniedError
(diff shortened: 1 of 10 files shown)


Kami:
Overall, I'm fine with this RBAC API change.

I assume you made this change so the utility functions are less coupled to the request / pecan API?

st2common/st2common/rbac/utils.py
```diff
@@ -63,158 +56,82 @@
 ]
 
 
-def request_user_is_admin(request):
```
Kami: Actually, now that I have checked more code - those functions were intended and there as convenience wrappers, but If removing them and directly retrieving user inside API controller some how makes it easier for you to make your other changes I'm fine with it (but would still like to hear reasoning :)).

PR #3182, opened 2017-02-03
Let querier plugin decide to delete state object on error

m4dcoder:
Add option in querier plugin to specify whether state object should be deleted if an error occur during query. This allows querier to recover in case of unexpected environment related exceptions. Mistral querier is refactored to delete state object on workflow completion or when the workflow or task references no longer exists.

### DIFF
--- a/contrib/runners/mistral_v2/query/mistral_v2.py
+++ b/contrib/runners/mistral_v2/query/mistral_v2.py
@@ -1,11 +1,13 @@
 import uuid
 
+from mistralclient.api import base as mistralclient_base
 from mistralclient.api import client as mistral
 from oslo_config import cfg
 import retrying
 
 from st2common.query.base import Querier
 from st2common.constants import action as action_constants
+from st2common.exceptions import resultstracker as exceptions
 from st2common import log as logging
 from st2common.services import action as action_service
 from st2common.util import jsonify
@@ -31,6 +33,8 @@ def get_instance():
 
 
 class MistralResultsQuerier(Querier):
+    delete_state_object_on_error = False
+
     def __init__(self, id, *args, **kwargs):
         super(MistralResultsQuerier, self).__init__(*args, **kwargs)
         self._base_url = get_url_without_trailing_slash(cfg.CONF.mistral.v2_base_url)
@@ -66,6 +70,9 @@ class MistralResultsQuerier(Querier):
         try:
             result = self._get_workflow_result(mistral_exec_id)
             result['tasks'] = self._get_workflow_tasks(mistral_exec_id)
+        except exceptions.ReferenceNotFoundError as exc:
+            LOG.exception('[%s] Unable to find reference. %s', execution_id, exc.message)
+            return (action_constants.LIVEACTION_STATUS_FAILED, exc.message)
         except Exception:
             LOG.exception('[%s] Unable to fetch mistral workflow result and tasks. %s',
                           execution_id, query_context)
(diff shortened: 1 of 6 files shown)


Kami:
From high level - LGTM.

Looking at the code, it doesn't seem to be a way for user / operator to enable this option for Mistral since it's disabled by default. Or am I missing something?

contrib/runners/mistral_v2/query/mistral_v2.py
```diff
@@ -87,7 +94,14 @@ def _get_workflow_result(self, exec_id):
         :type exec_id: ``str``
         :rtype: (``str``, ``dict``)
         """
-        execution = self._client.executions.get(exec_id)
+        try:
+            execution = self._client.executions.get(exec_id)
+        except mistralclient_base.APIException as mistral_exc:
+            if 'not found' in mistral_exc.message:
+                raise exceptions.ReferenceNotFoundError(mistral_exc.message)
```
Kami: :+1:

contrib/runners/mistral_v2/query/mistral_v2.py
```diff
@@ -105,10 +119,17 @@ def _get_workflow_tasks(self, exec_id):
         :type exec_id: ``str``
         :rtype: ``list``
         """
-        wf_tasks = [
-            self._client.tasks.get(task.id)
-            for task in self._client.tasks.list(workflow_execution_id=exec_id)
-        ]
+        try:
+            wf_tasks = [
```
Kami: I would personally prefer to make this a regular for loop, because list comprehensions are usually only intended to be used for simple transformations and in cases like that when you have exception handling, etc. regular for loop is more clear and obvious.

contrib/runners/mistral_v2/query/mistral_v2.py
```diff
@@ -87,7 +94,14 @@ def _get_workflow_result(self, exec_id):
...
         :rtype: (``str``, ``dict``)
         """
-        execution = self._client.executions.get(exec_id)
+        try:
+            execution = self._client.executions.get(exec_id)
+        except mistralclient_base.APIException as mistral_exc:
+            if 'not found' in mistral_exc.message:
+                raise exceptions.ReferenceNotFoundError(mistral_exc.message)
+            raise mistral_exc
+        except Exception:
```
Kami: This is redundant since we are not doing any handling of the exception here so it can be removed.

contrib/runners/mistral_v2/query/mistral_v2.py
```diff
@@ -105,10 +119,17 @@ def _get_workflow_tasks(self, exec_id):
...
+        try:
+            wf_tasks = [
+                self._client.tasks.get(task.id)
+                for task in self._client.tasks.list(workflow_execution_id=exec_id)
+            ]
+        except mistralclient_base.APIException as mistral_exc:
+            if 'not found' in mistral_exc.message:
+                raise exceptions.ReferenceNotFoundError(mistral_exc.message)
+            raise mistral_exc
+        except Exception:
```
Kami: Same here.

st2actions/tests/integration/test_resultstracker.py
```diff
@@ -58,16 +110,168 @@ def test_querier_started(self):
...
+
+        exec_id = str(ResultsTrackerTests.states['state2.yaml'].execution_id)
+        exec_db = LiveAction.get_by_id(exec_id)
+        self.assertTrue(exec_db.result['called_with'][exec_id] is not None,
+                        exec_db.result)
+
+        tracker.shutdown()
+
+        # Ensure state objects are deleted
+        self.assertRaises(
```
Kami: Should we also verify somewhere at the top that this object actually exists in the database in the first place?

m4dcoder:
It's not an option for the user/operator.  It's a querier plugin level option.

contrib/runners/mistral_v2/query/mistral_v2.py
m4dcoder: Will it still raise any general exception if I remove it?

contrib/runners/mistral_v2/query/mistral_v2.py
m4dcoder: Sure can change to a for loop

m4dcoder:
@Kami concerns/comments addressed.

PR #3183, opened 2017-02-03
Config Render Support in Action

bigmstone:
This adds support for rendering configuration elements inside action definitions for standalone actions and action-chains.

- [x] Tests
- [x] St2 Docs Update
- [x] Changelog Update
- [x] Explore Mistral Support


bigmstone:
Added StackStorm/st2docs PR StackStorm/st2docs#395

st2common/st2common/constants/pack.py
```diff
@@ -32,6 +32,9 @@
     'CONFIG_SCHEMA_FILE_NAME'
 ]
 
+# Prefix for render context w/ config
+PACK_CONTEXT_KV_PREFIX = 'pack_context'
```
bigmstone: Thinking this might should be called `config_context` now. Thoughts?

armab:
:+1:
For history, can you show a piece of `yaml` with complete action example?

bigmstone:
@armab added full example to docs https://github.com/StackStorm/st2docs/pull/395/files#diff-157d9a0f26f16d59cb8468550a11f495R224

contrib/runners/action_chain_runner/action_chain_runner.py
```diff
@@ -506,6 +508,20 @@ def _render_publish_vars(action_node, action_parameters, execution_result,
     @staticmethod
     def _resolve_params(action_node, original_parameters, results, chain_vars, chain_context):
         # setup context with original parameters and the intermediate results.
+        chain_parent = chain_context.get('parent', {})
+        pack = chain_parent.get('pack')
+        user = chain_parent.get('user')
+
+        if pack and user:
+            config_loader = ContentPackConfigLoader(
```
Kami: Maybe a good idea to add some log statements here so it's easier to know that config is used (and also the values and what config is being used).

contrib/runners/mistral_v2/query/mistral_v2.py
```diff
@@ -101,11 +101,13 @@ def _get_workflow_result(self, exec_id):
                 raise exceptions.ReferenceNotFoundError(mistral_exc.message)
             raise mistral_exc
 
-        result = jsonify.try_loads(execution.output) if execution.state in DONE_STATES else {}
+        result = jsonify.try_loads(
```
Kami: Minor style thing - since it splits over multiple lines I would prefer to just use a normal and not short hand if.

In any case, not a blocker.

Kami: Now back to the disable no-member pragma - just curious why this is needed now, but it wasn't needed in the past.

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -126,12 +128,13 @@ def _handle_schedule_execution(self, liveaction_api):
             LOG.exception('Unable to execute action. Unexpected error encountered.')
             abort(http_client.INTERNAL_SERVER_ERROR, str(e))
 
-    def _schedule_execution(self, liveaction, user=None):
+    def _schedule_execution(self, liveaction, user=None, pack=None):
```
Kami: Perhaps a good idea to add a docstring here which documents the parameters.

st2api/tests/unit/controllers/v1/test_executions_simple.py
```diff
@@ -385,27 +385,28 @@ def test_post_parameter_validation_explicit_none(self):
     def test_post_with_st2_context_in_headers(self):
         resp = self._do_post(copy.deepcopy(LIVE_ACTION_1))
         self.assertEqual(resp.status_int, 201)
-        parent_user = resp.json['context']['user']
-        parent_exec_id = str(resp.json['id'])
+        parent_user = unicode(resp.json['context']['user'])
```
Kami: Would need to research but perhaps safer and more cross-compatible to use `st2common.util.compat.to_unicode` (just using `unicode` won't work with all Python 2 and 3 versions.

contrib/runners/action_chain_runner/action_chain_runner.py
Kami: Actually, it looks like you probably also wanted to use `_get_config` utility function here?

Kami:
Overall, nice work :+1: 

I added some small comments, but no major blockers from my side.

As far as tests go - there is a good amount of unit tests, but it would also be great to add, if possible, "end to end" integration test (perhaps in st2tests repo).

contrib/runners/action_chain_runner/action_chain_runner.py
```diff
@@ -321,7 +323,7 @@ def run(self, action_parameters):
             }
 
         parent_context = {
-            'execution_id': self.execution_id
+            'execution_id': self.execution_id,
```
m4dcoder: What's with the comma here?

estee-tew:
https://stackstorm.slack.com/archives/community/p1492201362523180

contrib/runners/action_chain_runner/action_chain_runner.py
Mierdin: Wonder if this is a habit from writing Go. :) The compiler makes you do this for literals

estee-tew:
https://stackstorm.slack.com/archives/community/p1494869108843734

ytjohn:
lakstorm pointed me here.

For me, my use case is that I'd like to configure some defaults, specifically a default_url and allow the end user or workflow to override it. I can do this in python code (`if url is None: self.url = self.config['default_url']`), but it would be nicer to have stackstorm handle this when I define the parameter.

My example would look something like this:

```
parameters:
    url:
      type: string
      default: "{{ config.default_url }}"
      description: Service URL
```

ytjohn:
Today, I wanted to add an IntervalTimer trigger, and I wanted to have the delta set in configuration.

I think it would be advantageous to allow end-users of my pack to set the interval delta (and unit) in configuration instead of having to edit the pack's yaml file.

```
trigger:
  type: "core.st2.IntervalTimer"
  parameters:
    unit: "hours"
    delta: 20
```

contrib/runners/action_chain_runner/action_chain_runner.py
Kami: ^Bump

Besides that, LGTM.

Kami:
Let's include this in v2.4.0 :+1:

Kami:
Let's please also add a changelog entry and clarify what support for config values for actions means aka where those values are supported (default values for parameters inside action metadata, parameter values which are rendered during action execution time, action chain workflows context).

Issue #3170, opened 2017-01-27
action-alias update : 500 Internal Server Error


humblearner:
`$ st2 --debug action-alias update examples.remote_shell_cmd remote_shell_cmd.yaml`
```
# -------- begin 140245180802000 request ----------
curl -X GET -H  'Connection: keep-alive' -H  'Accept-Encoding: gzip, deflate' -H  'Accept: */*' -H  'User-Agent: python-requests/2.11.1' -H  'X-Auth-Token: 45baab7d9d4b4124aed3fb6781c9e9dd' 'http://127.0.0.1:9101/v1/actionalias/?name=examples.remote_shell_cmd'
# -------- begin 140245180802000 response ----------
[]
# -------- end 140245180802000 response ------------

# -------- begin 140245180802000 request ----------
curl -X GET -H  'Connection: keep-alive' -H  'Accept-Encoding: gzip, deflate' -H  'Accept: */*' -H  'User-Agent: python-requests/2.11.1' -H  'X-Auth-Token: 45baab7d9d4b4124aed3fb6781c9e9dd' http://127.0.0.1:9101/v1/actionalias/examples.remote_shell_cmd
# -------- begin 140245180802000 response ----------
{"description": "Execute a command on a remote host via SSH.", "extra": {}, "ack": {"format": "Hold tight while I run command: *{{execution.parameters.cmd}}* on hosts *{{execution.parameters.hosts}}*"}, "enabled": true, "name": "remote_shell_cmd", "result": {"format": "Ran command *{{execution.parameters.cmd}}* on *{{ execution.result | length }}* hosts.\n\nDetails are as follows:\n{% for host in execution.result -%}\n    Host: *{{host}}*\n    ---> stdout: {{execution.result[host].stdout}}\n    ---> stderr: {{execution.result[host].stderr}}\n{%+ endfor %}\n"}, "formats": ["run {{cmd}} on {{hosts}}"], "action_ref": "core.remote", "pack": "examples", "ref": "examples.remote_shell_cmd", "id": "588a97e255fc8c34bd710e02", "uid": "action:examples:remote_shell_cmd"}
# -------- end 140245180802000 response ------------

# -------- begin 140245180802000 request ----------
[...]

PR #3184, opened 2017-02-06
Fix action alias update API endpoint

Kami:
This fixes bug reported in https://github.com/StackStorm/st2/issues/3170.

It looks like this regression was introduced as part of https://github.com/StackStorm/st2/pull/2841, but we didn't have any tests for this API endpoint so the regression wasn't caught earlier.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -49,6 +49,7 @@ in development
   put the workflow in a PAUSED state in mistral. (improvement)
 * Add support for evaluating jinja expressions in mistral workflow definition where yaql
   expressions are typically accepted. (improvement)
+* Fix action alias update API endpoint. (bug fox)
 
 2.1.1 - December 16, 2016
 -------------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -49,7 +49,7 @@ in development
   put the workflow in a PAUSED state in mistral. (improvement)
 * Add support for evaluating jinja expressions in mistral workflow definition where yaql
   expressions are typically accepted. (improvement)
-* Fix action alias update API endpoint. (bug fox)
+* Fix action alias update API endpoint. (bug fix)
 
 2.1.1 - December 16, 2016
 -------------------------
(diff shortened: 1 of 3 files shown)


m4dcoder:
It's not obvious from the original problem posted at https://github.com/StackStorm/st2/issues/3170 to the fix here.  Maybe more explanation in the summary will help.  Otherwise the code LGTM.

CHANGELOG.rst
```diff
@@ -49,6 +49,7 @@ in development
   put the workflow in a PAUSED state in mistral. (improvement)
 * Add support for evaluating jinja expressions in mistral workflow definition where yaql
   expressions are typically accepted. (improvement)
+* Fix action alias update API endpoint. (bug fox)
```
m4dcoder: Do you mean bug fix?

humblearner: minor: s/fox/fix

st2api/st2api/controllers/v1/actionalias.py
```diff
@@ -156,13 +156,16 @@ def post(self, action_alias):
 
         return action_alias_api
 
-    @request_user_has_resource_db_permission(permission_type=PermissionType.ACTION_MODIFY)
+    @request_user_has_resource_db_permission(permission_type=PermissionType.ACTION_ALIAS_MODIFY)
     @jsexpose(arg_types=[str], body_cls=ActionAliasAPI)
-    def put(self, action_alias_ref_or_id, action_alias):
+    def put(self, action_alias, action_alias_ref_or_id):
```
lakshmi-kannan: Interesting. I didn't realize body is first arg and ref is second.

CHANGELOG.rst
Kami: lol, yeah, good catch :D

Kami:
> It's not obvious from the original problem posted at #3170 to the fix here. Maybe more explanation in the summary will help. Otherwise the code LGTM.

@m4dcoder Yeah, regression / bug was introduced as part of - https://github.com/StackStorm/st2/pull/2841

That PR changed the arguments order in the method signature and the fix is changing the arguments order in the method signature (previously first argument was id, now it's the actual request payload).

st2api/st2api/controllers/v1/actionalias.py
Kami: Yeah, I was personally not a fan of that change and I prefer id to be first (this way it's consistent with jsexpose and everything and less magic, etc.), but @enykeev some how managed to continence me in https://github.com/StackStorm/st2/pull/2841 that this is good idea :P

enykeev: In st2api v2, position of the argument would not matter at all, all of them will be passed as kwargs.

PR #3185, opened 2017-02-06
Add support for MongoDB 3.4

### LINKED ISSUE #3100: Support MongoDB 3.4
MongoDB 3.4 is the current latest stable release and we should eventually add support for it.

Sadly it won't be totally straight- forward, because that version includes some backward incompatible changes and we make sure it doesn't break stuff for existing users (for new installations it will probably be easier although there are still some pymongo changes we need to evaluate).

Kami:
This pull request updates pymongo and mongoengine dependencies to latest versions which also support MongoDB 3.4.

In addition to that it also makes various other changes to the code to make it work with newest versions of those libraries (some backward incompatible changes were introduced).

Closes #3100.

## TODO

- [x] Automated tests for 3.2 on Circle CI
- [x] Automated tests for 3.4 on Circle CI
- [x] Docs which clarify that 3.2 and 3.4 are supported versions of MongoDB - https://github.com/StackStorm/st2docs/pull/384


Kami:
@enykeev I will probably need to update Circle CI scripts to they support installing and testing with multiple MongoDB versions.

I spent time on that in the past for Travis CI and it worked (we had tests running with MongoDB 2.6 and 3.2), but we introduced a "regression" when moving to Circle CI with regards to that aka we now only test with one version :/

Kami:
I got the builds with two different MongoDB versions working on Circle CI (see diff for corresponding circle.yml and script changes).

Sadly Circle CI user interface is very user-unfriendly and terrible when running multiple things in parallel (it doesn't support build Matrix like Travis CI does which makes it hard to see what is going on).

I will see what I can do to to make it more user-friendly and easier to see which MongoDB version is being used, etc.

In addition to that, it looks like there are still some MongoDB 3.4 related code issues I need to fix (3.4 perform more script index validation, etc. - https://docs.mongodb.com/manual/release-notes/3.4-compatibility/#stricter-validation-of-index-specifications)

Kami:
All the unit and integration tests are passing with MongoDB 3.2 and 3.4 on CircleCI now.

I will do some more polish and work on a corresponding documentation PR then I will consider this done and ready for the review.

I plan to update the installer scripts to install 3.4 by default in the near future when we have more testing with 3.4 under the belt, but we should still update the blog and release announcement and let people know that since StackStorm v2.2.0 we now also support MongoDB 3.3.

.circle/install-and-run-mongodb.sh
```diff
@@ -0,0 +1,33 @@
...
+# Script which installs versions of MongoDB specified using an environment variable
+
+if [ ! "${MONGODB_VERSION}" ]; then
+    echo "MONGODB_VERSION environment variable not set"
+    exit 2
+fi
+
+# Note: MongoDB 2.4 and 2.6 don't work with ramdisk since they don't work with
+# small files and require at least 3 GB of space
+# TODO: Use ramdisk
```
Mierdin: Is this something you wanted to do in this PR or is this for future?

.circle/install-and-run-mongodb.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+# Script which installs versions of MongoDB specified using an environment variable
+
+if [ ! "${MONGODB_VERSION}" ]; then
+    echo "MONGODB_VERSION environment variable not set"
```
m4dcoder: It would be nice to provide a list of supported versions in this error message.

.circle/install-and-run-mongodb.sh
Kami: Yeah in the future - but I will actually remove this note all together since on Circle CI ramdisk doesn't seem to make much difference (probably because of virtualzed memory)

.circle/install-and-run-mongodb.sh
Kami: Agreed - will update it.

Kami: Actually, that's a bit hard since it's not a static list - it will work with any version which is available on MongoDB mirrors...

Issue #3186, opened 2017-02-06
"--config-file" option doesn't behave as expected when passing in home dir shortcut


Mierdin:
The [CLI documentation](https://docs.stackstorm.com/reference/cli.html#configuration-file) gives an example of being able to pass a different configuration file into the `--config-file` argument, but this appears to just append this string to the current directory (implying a relative path is required here), rather than evaluate the provided argument as an absolute path:

```
(virtualenv)vagrant@st2dev:~/st2/st2client$ st2 --config-file=~/.st2/prod-config action list
Traceback (most recent call last):
  File "/home/vagrant/st2/virtualenv/bin/st2", line 11, in <module>
    load_entry_point('st2client==2.2.dev0', 'console_scripts', 'st2')()
  File "/home/vagrant/st2/st2client/st2client/shell.py", line 331, in main
    return Shell().run(argv)
  File "/home/vagrant/st2/st2client/st2client/shell.py", line 272, in run
    config = self._parse_config_file(args=args)
  File "/home/vagrant/st2/st2client/st2client/base.py", line 160, in _parse_config_file
    config_file_path = self._get_config_file_path(args=args)
  File "/home/vagrant/st2/st2client/st2client/base.py", line 179, in _get_config_file_path
    raise ValueError('Config "%s" not found' % (path))
ValueError: Config "/home/vagrant/st2/st2client/~/.st2/prod-config" not found
```

Mierdin:
Indeed, this seems to work:

```
st2 --config-file=../../.st2/prod-config action list
```

I would prefer that this uses the absolute path, or if this was done by design, then the help output and st2 docs should be updated to be clearer here.

Kami:
@Mierdin Yeah, I'm fine with calling `os.path.expanduser` on the path provided by the user via command line argument.

I believe doing that should make the example you pasted work (we might also need to call abspath on it, but not 100% sure).

Kami:
Also, good catch - it seems like `os.path.abspath` call was added at some point later when docs were added so it broke.

Will make sure we have tests for it going forward :)

lakshmi-kannan:
This seems fixed now. Closing. cc: @Kami

Kami:
Yep, fixed in #3192. Thanks for closing.

PR #3192, opened 2017-02-07
Make config file paths which are relative to the user's home directory work inside st2 CLI

Kami:
This should fix the issue reported in https://github.com/StackStorm/st2/issues/3186.

## TODO

- [x] Tests


Mierdin:
Nice, thanks! I'll 👍  once there's tests - just hit me up on slack

Kami:
Tests added in c9fbc152932281e7f867c7b5ca2282d8f8b3b308.

PR #3195, opened 2017-02-07
Add more Jinja examples for Mistral

m4dcoder:
Add examples on how to use Jinja expresssions to branch in workflow, iterate thru with-items, and access the variables returned from the env() filter.


bigmstone:
Thanks for this. :100:

PR #3199, opened 2017-02-09
Deprecate {{system}} and {{user}} jinja notations

lakshmi-kannan:
Redo https://github.com/StackStorm/st2/pull/3164 after we decided to only deprecate {{system}} and {{user}} jinja notations. 

Deprecation notice in Upgrade Notes - https://github.com/StackStorm/st2docs/pull/367

### DIFF
--- a/contrib/runners/action_chain_runner/action_chain_runner.py
+++ b/contrib/runners/action_chain_runner/action_chain_runner.py
@@ -193,7 +193,6 @@ class ChainHolder(object):
         if not vars:
             return {}
         context = {}
-        context.update({SYSTEM_SCOPE: KeyValueLookup(scope=SYSTEM_SCOPE)})
         context.update({
             DATASTORE_PARENT_SCOPE: {
                 SYSTEM_SCOPE: KeyValueLookup(scope=FULL_SYSTEM_SCOPE)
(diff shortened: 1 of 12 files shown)


contrib/runners/action_chain_runner/action_chain_runner.py
```diff
@@ -193,7 +193,6 @@ def _get_rendered_vars(vars, action_parameters):
         if not vars:
             return {}
         context = {}
-        context.update({SYSTEM_SCOPE: KeyValueLookup(scope=SYSTEM_SCOPE)})
```
Kami: I assume this change was intentional aka old notation is not supported anymore here (and not just deprecated)?

Kami:
LGTM although I assumed we will still support old notation for a while long - only deprecate it first and fully remove support for it in a future release.

If we already intend to fully remove it in this release (and not just deprecate it) I would prefer to make wording in changelog and upgrade nodes more explicit - e.g. "This notation has been removed in StackStorm v2.2, you need to update your code for it to work, etc."

contrib/runners/action_chain_runner/action_chain_runner.py
lakshmi-kannan: Correct. This is the exact change in https://github.com/StackStorm/st2/pull/3164.

Issue #3200, opened 2017-02-09
Packs install (dowhload) is broken when repo's default branch is not master


Kami:
If user specifies a "branch" parameter when installing a pack it doesn't work.

From Slack:

```bash
[14:19] root@ip-10-1-0-25:~/aws/roles/stackstorm/tasks# st2 execution get 589c7a79ecf50606a1314870
id: 589c7a79ecf50606a1314870
status: failed (2s elapsed)
parameters: 
  force: false
  packs:
  - git@github.com:pearsontechnology/stackstorm-bitesize.git
result: 
  exit_code: 1
  result: None
  stderr: "Traceback (most recent call last):
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/runners/python_action_wrapper.py", line 238, in <module>
    obj.run()
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/runners/python_action_wrapper.py", line 152, in run
    output = action.run(**self._parameters)
  File "/opt/stackstorm/packs/packs/actions/pack_mgmt/download.py", line 75, in run
    verifyssl=verifyssl, ref=pack_version)
  File "/opt/stackstorm/packs/packs/actions/pack_mgmt/download.py", line 139, in _clone_repo
    repo.git.checkout('-b', short_branch, branch)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/git/cmd.py", line 423, in <lambda>
    return lambda *args, **kwargs: self._call_process(name, *args, **kwargs)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/git/cmd.py", line 868, in _call_process
    return self.execute(call, **_kwargs)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/git/cmd.py", line 686, in execute
    raise GitCommandError(command, status, stderr_value, stdout_value)
git.exc.GitCommandError: Cmd('git') failed due to: exit code(128)
  cmdline: git checkout -b dev dev
  stderr: 'fatal: A branch named 'dev' already exists.'
"
  stdout: ''
```

We need some unit and integration tests for this as well.

Kami:
Edit: It looks like it might only be related to an edge case when using a non-master default branch - confirming...

Kami:
Confirmed - it's related to the default branch not being master - will work on a fix.

Kami:
Fixed by #3201.

PR #3201, opened 2017-02-09
Fix packs.download action so it works if default repo branch is not master

Kami:
packs.download and as such `pack install` command would not work if git repository uses a default branch which is not master.

This pull request fixes that.

The whole gitref and checkout code is quite convoluted so I won't accept this as done until we have good coverage for all the ref edge cases - it's possible that there are more bugs lurking in that code.

I understand the reason why the code is doing what is doing (we try to determine if ref / commit is part of a branch, but ref could also be a branch), but I will also think about it some more and see if I can simply some of that logic. Simpler logic means less chance for bugs.

This fixes a bug reported by a user in the community - https://github.com/StackStorm/st2/issues/3200.

## TODO

- [x] Tests for all the ref related edge cases (we are lacking that but we need it)
- [x] Changelog

### DIFF
--- a/contrib/packs/actions/pack_mgmt/download.py
+++ b/contrib/packs/actions/pack_mgmt/download.py
@@ -110,7 +110,7 @@ class DownloadGitRepoAction(Action):
             gitref = repo.active_branch.object
         else:
             # Try to match the reference to a branch name (i.e. "master")
-            gitref = DownloadGitRepoAction._get_gitref(repo, "origin/%s" % ref)
+            gitref = DownloadGitRepoAction._get_gitref(repo, 'origin/%s' % ref)
             if gitref:
                 use_branch = True
 
@@ -120,7 +120,7 @@ class DownloadGitRepoAction(Action):
 
         # Try to match the reference to a "vX.Y.Z" tag
         if not gitref and re.match(PACK_VERSION_REGEX, ref):
-            gitref = DownloadGitRepoAction._get_gitref(repo, "v%s" % ref)
+            gitref = DownloadGitRepoAction._get_gitref(repo, 'v%s' % ref)
 
         # Giving up ¯\_(ツ)_/¯
         if not gitref:
@@ -142,7 +142,7 @@ class DownloadGitRepoAction(Action):
         branches = branches.replace('*', '').split()
 
         if active_branch.name not in branches or use_branch:
-            branch = "origin/%s" % ref if use_branch else branches[0]
+            branch = 'origin/%s' % ref if use_branch else branches[0]
             short_branch = ref if use_branch else branches[0].split('/')[-1]
             repo.git.checkout('-b', short_branch, branch)
             branch = repo.head.reference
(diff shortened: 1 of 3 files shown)


Kami:
As mentioned on Slack, I tested all the "edge cases" locally, which include installing pack from a repo where:

* default branch is master, ref is pack version
* default branch is master, ref is branch name
* default branch is master, ref is default branch name
* default branch is not master, ref is pack version
* default branch is not master, ref is branch name
* default branch is not master, ref is default branch name

This all appears to be working and I plan to work on the tests tomorrow, but I would still prefer if we can maybe re-think the whole gitref stuff and simply it / get rid of it if possible.

contrib/packs/tests/test_action_download.py
```diff
@@ -289,3 +289,58 @@ def test_resolve_urls(self):
 
         url = DownloadGitRepoAction._eval_repo_url("file:///home/vagrant/stackstorm-test")
         self.assertEqual(url, "file:///home/vagrant/stackstorm-test")
+
+    def test_run_pack_download_edge_cases(self):
```
Kami: Thanks for working on the tests.

And yeah, I knew that sadly testing will be quite a hassle because it's hard to replicate a default branch problem without mocking things out...

I will also look into if it's possible to write some more tests which require less mocking so we can exercise more code paths...

Kami:
I added another test case for scenario where `ref` is None - this is a very common scenario during regular code flow because if user doesn't specify a pack version, method which parses version of from the provided string returns `None` for `pack_version` and as such, `ref` is None - a2e3e9057522c13788aedbb84b3f31322e801a7b

Sadly it turns out it's quite hard to test things automatically without mocking stuff and mocking is not ideal because we mock some things we would actually want to test...

Kami:
In any case, I think that's about as much as we can do here.

We should consider adding more of the tests to st2tests though - https://github.com/StackStorm/st2tests/blob/master/packs/tests/actions/chains/test_packs_pack.yaml

contrib/packs/actions/pack_mgmt/download.py
```diff
@@ -100,20 +100,27 @@ def _clone_repo(temp_dir, repo_url, verifyssl=True, ref='master'):
         # because we want the user to work with the repo in the
         # future.
         repo = Repo.clone_from(repo_url, temp_dir)
+        active_branch = repo.active_branch
```
enykeev: Looks fine on the glance, but I'm slightly worried there might be cases when `active_branch` may not point to remote's default branch, but rather just be local checkout or something else entirely.

Kami:
I again manually confirmed the fix and changed by running pack install for various conditions listed below and then also verifying that the repo is checked out at the correct revision by doing running `cd /opt/stackstorm/packs/test ; git status ; git branch ; git log`.

contrib/packs/actions/pack_mgmt/download.py
Kami: We always clone the repo so this shouldn't matter, right?

And that's really what we are interested in - repo's active branch (which happens to be repo's default remote branch as well).

Kami:
Also added "end to end" test case here - https://github.com/StackStorm/st2tests/pull/93

emedvedev:
LGTM. 

As discussed yesterday, this whole approach should eventually be simplified along with switching to a another git library, but right now there's nothing we can do without sacrificing the UX (even if in small edge cases) or breaking RHEL6, where the default git version doesn't support some of the newer flags.

For now, this PR should be enough to fix the behavior :)

PR #3202, opened 2017-02-09
Ignore "index with different options already exists" mongoengine errors

Kami:
This pull request updates the code to ignore the errors about which we can't do anything right now.

Correct solution we should eventually do is to re-creating offending indexes. The problem with re-creation is that the operation is blocking and depending on the dataset size, it could take a very long time.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -22,7 +22,7 @@ dependencies:
         background: true
     # We sleep a bit to wait for the background process to start and script to
     # finish
-    - case $CIRCLE_NODE_INDEX in [1,2]) sleep 10 ;; esac:
+    - case $CIRCLE_NODE_INDEX in [1,2]) sleep 10 ;; esac
     # Tail the logs so it's easier to see what is going on. Sadly when using "background: true"
     # whole output is ignored.
     - case $CIRCLE_NODE_INDEX in [1,2]) tail -50 /tmp/mongodb-install.log ;; [3,4]) tail -50 /tmp/mongodb-install.log ;; esac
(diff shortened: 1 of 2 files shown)


Kami:
To give some context - MongoDB 3.4 is much stricter with index validation and uid field index creation would fail on bootstrap when running on existing installation because index with the same name, but different fields already exist (`types` index option was removed from MongoDB 3.4).

This would cause a failure, but the error is not actually fatal - old / existing index still works just fine.

As mentioned above, real long-term solution is to re-work the ensure_indexes code to re-create offending indexes, but we will still need to be careful we don't mess with indexes in such manner unless is really needed because it re-creation would be very slow and expensive.

circle.yml
```diff
@@ -20,6 +20,9 @@ dependencies:
     # We install and test with all the supported MongoDB versions
     - case $CIRCLE_NODE_INDEX in [1,2]) sudo service mongodb stop ; MONGODB_VERSION=3.2.12 sudo -E .circle/install-and-run-mongodb.sh > /tmp/mongodb-install.log 2>&1 ;; [3,4]) sudo service mongodb stop ; MONGODB_VERSION=3.4.2 sudo -E .circle/install-and-run-mongodb.sh > /tmp/mongodb-install.log 2>&1 ;; esac:
         background: true
+    # We sleep a bit to wait for the background process to start and script to
+    # finish
+    - case $CIRCLE_NODE_INDEX in [1,2]) sleep 10 ;; esac
```
enykeev: Can't we just make it a part of the previous line?

Kami: Sadly no because it runs in a background (background: true) and that's why we need it to be a separate step.

enykeev: What about the next one? =)

Kami: It can but I prefer separate lines to make it more readable (flow it's more sequential).

I already hate all the case lines which make it hard to read and follow so I plan to look into circleci-matrix or similar solution in the future so we can at the very least keep circle.yml more sane.

Kami:
It looks like there is sadly no way around it but to drop and re-create the offending index - ebcf8cb20401ec6235143057e02530e4e9dc012e.

Luckily this only affects the "uid" index.

See https://docs.mongoengine.org/upgrade.html#inheritance for details on the types stuff.

Kami:
It looks like the e2e issue was the same issue as one here - https://github.com/StackStorm/st2/pull/3211#issuecomment-279321093

Syncing with latest master should solve it.

PR #3208, opened 2017-02-10
Remove obsolete coverall badge from readme

Kami:
We use codecov now, not coveralls anymore.


Kami:
It looks like the e2e issue was the same issue as one here - https://github.com/StackStorm/st2/pull/3211#issuecomment-279321093

Syncing with latest master solved it.

PR #3211, opened 2017-02-10
Fix st2 login and st2 whoami commands to work

Kami:
Randomly noticed it's not working while testing it - https://gist.github.com/Kami/f03360208e2e8599f42d3152df054842

It looks like the problem is that code tries to work with the `ConfigParser` object as a dictionary, but `ConfigParser` object doesn't behave and support dictionary-like access notation.

Edit: Looked at the tests and it looks like they didn't catch this issue because the hide it because they mock `ConfigParser` instance. I know that's not always possible, but we should try to avoid mocking as much as possible because many times we end up mocking too much and we actually end up testing mocks and not the actual code...

## TODO

- [x] Tests (and figure out why existing tests passed, maybe they mock too much. Imo, better and safer approach would be to use actual config file fixtures from disk - this way we don't need to mock stuff.)

### DIFF
--- a/st2client/tests/unit/test_auth.py
+++ b/st2client/tests/unit/test_auth.py
@@ -60,19 +60,21 @@ class CaptureStdout(list):
 
 
 class TestLoginBase(base.BaseCLITestCase):
-    '''A base class for testing related to 'st2 login' commands
+    """
+    A base class for testing related to 'st2 login' commands
 
     This exists primarily to ensure that each specific test case is kept atomic,
     since the tests create actual files on the filesystem - as well as to cut down
     on duplicate code in each test class
-    '''
+    """
 
     DOTST2_PATH = os.path.expanduser('~/.st2/')
 
-    def __init__(self, *args, **kwargs):
-        self.config_file = kwargs.pop('config_file', '~/.st2/config')
-        self.config_contents = kwargs.pop('config_contents', None)
-        super(TestLoginBase, self).__init__(*args, **kwargs)
+    def __init__(self, config_file='~/.st2/config', config_contents=None):
+        super(TestLoginBase, self).__init__()
+
+        self.config_file = config_file
+        self.config_contents = config_contents
 
         self.parser = argparse.ArgumentParser()
         self.parser.add_argument('-t', '--token', dest='token')
@@ -87,7 +89,6 @@ class TestLoginBase(base.BaseCLITestCase):
             os.remove(self.config_file)
 
         with open(self.config_file, 'w') as cfg:
-
             # If a test passes in it's own config, we write that instead
             if self.config_contents:
                 for line in self.config_contents.split('\n'):
@@ -125,8 +126,7 @@ class TestLoginPasswordAndConfig(TestLoginBase):
 
     def __init__(self, *args, **kwargs):
         super(TestLoginPasswordAndConfig, self).__init__(
-            config_file=self.CONFIG_FILE, *args, **kwargs
-        )
+            config_file=self.CONFIG_FILE)
 
     @mock.patch.object(
         requests, 'post',
@@ -171,8 +171,7 @@ class TestLoginIntPwdAndConfig(TestLoginBase):
 
     def __init__(self, *args, **kwargs):
         super(TestLoginIntPwdAndConfig, self).__init__(
-            config_file=self.CONFIG_FILE, *args, **kwargs
-        )
+            config_file=self.CONFIG_FILE)
 
     @mock.patch.object(
         requests, 'post',
@@ -219,8 +218,7 @@ class TestLoginWritePwdOkay(TestLoginBase):
 
     def __init__(self, *args, **kwargs):
         super(TestLoginWritePwdOkay, self).__init__(
-            config_file=self.CONFIG_FILE, *args, **kwargs
-        )
+            config_file=self.CONFIG_FILE)
 
     @mock.patch.object(
         requests, 'post',
@@ -265,8 +263,7 @@ class TestLoginUncaughtException(TestLoginBase):
 
     def __init__(self, *args, **kwargs):
         super(TestLoginUncaughtException, self).__init__(
-            config_file=self.CONFIG_FILE, *args, **kwargs
-        )
+            config_file=self.CONFIG_FILE)
 
     @mock.patch.object(
         requests, 'post',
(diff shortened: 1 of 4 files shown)


Kami:
@Mierdin I'm a bit confused about this - it seems like it should have never worked, or am I missing something?

Kami:
Edit: So it seems like the problem is that only ConfigParser instance in Python 3 works like a dict (https://docs.python.org/3/library/configparser.html#module-configparser), but not one available in Python 2 (https://docs.python.org/2/library/configparser.html).

Did you happen to test it with Python 3 locally (I used Python 2.7 which is the version we officially support)?

We don't fully support Python 3 yet and if anything, we should use the approach which works on both versions where possible :)

Mierdin:
Hrm, this is all very strange. FWIW I was not using Python 3 - my dev box runs 2.7.6.

I'll dig into this today and see if I can't get it straightened out. Apologies if I'm over-relying on `mock` - I was always told to use `mock` to avoid testing things you weren't trying to test. In this case, I don't want to test ConfigParser, I want to test the code that's using it. However, I can see your perspective and will try to find some middle ground.

Kami:
> Apologies if I'm over-relying on mock - I was always told to use mock to avoid testing things I wasn't trying to test. In this case, I don't want to test ConfigParser, I want to test the code that's using it. However, I can see your perspective and will try to find some middle ground.

No problem, sometimes it's hard to find the balance what to mock and what to not in the unit tests.

> Hrm, this is all very strange. FWIW I was not using Python 3 - my dev box runs 2.7.6.

It looks like this could also happen if you have `configparser` package installed locally - https://pypi.python.org/pypi/configparser

Afaik, none of our packages and components rely and use that package, but having this package installed locally would cause this behavior.

Mierdin:
@Kami figured it out. This is the culprit - https://github.com/StackStorm/st2/pull/3194

I was 99% sure I sufficiently tested everything manually after that change, but the only possible explanation is, no I did not. :-/

Anyways - this one's on me - I'll fix it.

Mierdin:
So....after looking into this a bit more, here's my working theory. It's true that the built-in `ConfigParser` doesn't allow dict-like references in until Python 3. However, it appears that you can install [this package](https://pypi.python.org/pypi/configparser) and get this functionality when running Python 2.7.

Looks like this does get installed in the virtualenv when running `make requirements`, as it is a dependency of `flake8`:

```
~$ pipdeptree
....
flake8==3.0.4
  - configparser [required: Any, installed: 3.5.0]
  - enum34 [required: Any, installed: 1.1.6]
  - mccabe [required: >=0.5.0,<0.6.0, installed: 0.5.3]
  - pycodestyle [required: >=2.0.0,<2.1.0, installed: 2.0.0]
  - pyflakes [required: <1.3.0,!=1.2.1,!=1.2.2,>=0.8.1,!=1.2.0, installed: 1.2.3]
....
```

So, this explains why it was working before the move to using the `ConfigParser` in `six`. Before that, the commands worked just fine, because they were using this third-party library that acted like the built-in Python3 `ConfigParser`.

Now, regarding the errors that came up once #3194 was merged, I'm guessing this is because `six`'s `ConfigParser` is designed to act like the built-in version for Python 2, which doesn't have __getitem__ access either. 

All that said, the changes you made to auth.py are definitely better, and my mocks were clearly wrong as they were built to mimic the wrong `ConfigParser` version. I re-did all of the unit tests for this in https://github.com/StackStorm/st2/pull/3211/commits/ea4c1212d64beadb720a5339c59db84c9f922e7c; hopefully they'll help protect from this a bit better.

Kami:
Yes, I also agree with your assessment :)

And yeah, it's a bit nasty that dev dependencies can interfere with things in suck a sneaky manner...

st2client/tests/unit/test_auth.py
```diff
@@ -41,277 +41,393 @@
...
+                    self.assertEquals(line.split(' ')[2][:-1], expected_username)
 
+            # validate token was created
+            self.assertTrue(os.path.isfile('%stoken-%s' % (self.DOTST2_PATH, expected_username)))
 
-class TestLogin(base.BaseCLITestCase):
+
+class TestLoginUncaughtException(TestLoginBase):
+
+    CONFIG_FILE = '/tmp/logintest.cfg'
```
Kami: Might be a bit safer to use a random temporary path (e.g. mkstemp), besides that, LGTM.

Kami:
It appears e2e tests failed because this branch doesn't contain packs.download fix / changes (the tests are passing on master).

I've merged latest master into this branch, this should fix it.

st2client/tests/unit/test_auth.py
```diff
@@ -22,9 +21,10 @@
 import requests
 import argparse
 import logging
+from cStringIO import StringIO
```
Kami: Minor style / import ordering thing - we tend to put stdlib imports at the top then followed by 3rd party package imports then followed by local components imports.

Kami: For cross version compatibility it would be better to use `six.StringIO` (yeah, this one shouldn't cause issues :D)

st2client/tests/unit/test_auth.py
```diff
@@ -41,277 +41,393 @@
...
+    '''A base class for testing related to 'st2 login' commands
+
+    This exists primarily to ensure that each specific test case is kept atomic,
+    since the tests create actual files on the filesystem - as well as to cut down
+    on duplicate code in each test class
+    '''
+
+    DOTST2_PATH = os.path.expanduser('~/.st2/')
 
     def __init__(self, *args, **kwargs):
```
Kami: Instead of using kwargs, it would be a bit more readable to just directly declare arguments in the constructor signature.

Kami:
I pushed some changes (41090ab3719beb8f564729980dd5586323e88eec, c1923f64dbae13f318a6aab5aa9dc1eac71ba754, ba85739bbb964ceb01653670f0cd7319ee9b5660, 339af8950e799ae97c946675126c84391c1a628b, 2602577b2935a587e79744db01eec0f347365e0b).

I think we might also be able to re-use stdout capturing from base class (no need to create and use CaptureStdout), but I would need to dig in to confirm and something we can simplify in the future (if possible).

Kami:
I pushed some changes (41090ab3719beb8f564729980dd5586323e88eec, c1923f64dbae13f318a6aab5aa9dc1eac71ba754, ba85739bbb964ceb01653670f0cd7319ee9b5660, 339af8950e799ae97c946675126c84391c1a628b, 2602577b2935a587e79744db01eec0f347365e0b).

I think we might also be able to re-use stdout capturing from base class (no need to create and use CaptureStdout), but I would need to dig in to confirm and something we can simplify in the future (if possible).

Kami:
@Mierdin when you get a chance can you please have a look and approve it (I forgot I opened it and the PR hasn't been approved yet so I can't merge it :))?

Kami:
It looks like I was also able to replace `CaptureStdout` with existing stdout capture mechanism in the base test class - 072bd3ddcaa4e4e4368346902b7d74b75542a940. It just required a small fix to the base test class otherwise it wouldn't work when multiple classes inherited from the same bass class - 380b958dd52bb10934e8cc43e9fb19d585262b34 (internal state was not reset for each test run).

Kami:
Some of the e2e tests failure appear to be related to intermediate upstream / networking issues.

st2client/tests/unit/test_auth.py
Mierdin: I follow the same standard, and I know better. Need to figure out why my editor didn't catch this

Mierdin:
@kami I re-ran the e2e test and it passed. Not sure why Codecov is throwing a fit though.

PR #3189, opened 2017-02-07
Upgrade various dependencies

Kami:
Just some house keeping.

It updates the following dependencies to the latest stable version.

## General dependencies

* paramiko
* appscheduler
* pyyaml
* semver
* retrying
* kombu
* amqp
* networkx
* jsonschema
* sseclient
* jsonpath-rw

## Dev / test dependencies

* pylint, astroid (done in a separate PR which fixes some lint issues)
* flake8 (done in a separate PR which fixes some lint issues)
* mock (done in a separate PR which fixes some lint issues)

I verified the checked and all the changes are backward compatible, but I still think it's better to merge this post v2.2.0.

PR #3215, opened 2017-02-12
Fix pylint on runner Python modules

Kami:
I noticed it was broken when I was upgrading pylint as part of #3189.

We passed invalid (in-existent) path to pylint but older versions simply silently ignored that.

For example, it got passed in `contrib/runners/action_chain_runner/contrib/runners/action_chain_runner/` instead of `contrib/runners/action_chain_runner/` so files in those directories were not actually pylinted.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -116,7 +116,7 @@ flake8: requirements .flake8
 	@echo "==================== flake ===================="
 	@echo
 	. $(VIRTUALENV_DIR)/bin/activate; flake8 --config ./lint-configs/python/.flake8 $(COMPONENTS)
-	. $(VIRTUALENV_DIR)/bin/activate; flake8 --config ./lint-configs/python/.flake8 $(COMPONENTS_RUNNER)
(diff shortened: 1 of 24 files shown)


Kami:
Discovered another issue - f1cf556d6aea42a6a24ff123afd307022603410e.

Kami:
Re line break change - it was needed because of new version of flake8 which improves detection of "2 line breaks after a class and function definition" rule.

This rule was there from the beginning, but it didn't correctly detect all the places and new version of flake8 fixes that so we need to update our code.

Kami:
Since we will be merging https://github.com/StackStorm/st2/pull/3219 into master, I will go ahead and merge this one as well.

This one doesn't affect st2 code, just the build and improves lint detection.

Kami:
e2e test failures are related to external service being offline. 

https://github.com/StackStorm/st2cd/pull/243 should hopefully improve the situation.

PR #3219, opened 2017-02-15
Make 'st2 login' actually work

Mierdin:
In `st2client/base.py`, there's a conditional that detects if the username and password
is provided, and if so, try to retrieve a token and update the token attribute of the
client before returning. However, in a traditional `st2 login` scenario, only the
`username` field is populated, so this code will never run.

Previous tests were done on systems that effectively had auth disabled (doh!), so while
the config file and token files were being changed/generated correctly, API requests
did not carry the `X-Auth-Token` header. Since auth was disabled, this was no problem.
However, on a real installation, this feature just plainly did not work.

In short, this allows the 'st2 login' feature to work, and it also provides unit tests
to help prevent this kind of thing

- [x] Add another unit test to check that other commands also send auth header after being configured with `st2 login`


Kami:
Good catch - thanks.

Will test and confirm locally it works (with auth enabled :)).

st2client/st2client/shell.py
```diff
@@ -65,8 +64,7 @@ class Shell(BaseCLIApp):
     LOG = LOGGER
 
     SKIP_AUTH_CLASSES = [
-        TokenCreateCommand.__name__,
-        LoginCommand.__name__
```
Kami: LoginCommand needs to be here otherwise it won't work - I confirmed it locally.

Kami:
I confirmed it's working locally with this change - 4b8a1460ec7190f494fd4ccb2224dc3485fc8e62.

st2client/st2client/shell.py
Kami: On a related note - we need a test case to catch this.

If `LoginCommand` class is not here, st2 login won't work so we should have a test which catches this :)

Kami:
I pushed another change to skip auth api call when running `st2 whoami` command - 7d09f74002032ac083d49db41004f71b648c9c47

Imo (please correct me if I'm wrong), there is no need to try to authenticate and hit the auth API when running `st2 whoami` command - we just want to read the username from config file / environment variable.

st2client/tests/unit/test_auth.py
```diff
@@ -154,6 +157,14 @@ def runTest(self, mock_gp):
 
         self.shell.run(args)
 
+        # TODO(mierdin): This tests that this particular command sends X-Auth-Token but you should
+        # also test other commands after this token has been installed
+        kwargs = {
+            'headers': {'X-Auth-Token': self.TOKEN['token'], 'content-type': 'application/json'},
```
Kami: This test was correct since this request is hitting auth api (and not regular API) to authenticate to retrieve the token and as such, only username and password should be sent (and token returned in the response and used on subsequent requests to the st2 api).

Kami: Alright, I belive this should do it - cd89193fd69bda32a1f9591faa1e9ea6f8e953a0, 545c7baef5225e9e9c0a10d97c25dec3b822c87c.

Kami:
I believe this should do it as far as current state of st2 tests goes - https://github.com/StackStorm/st2/pull/3219/files#r101242702

Some additional st2tests (aka integration tests) would also be nice at some point.

Kami:
Same story for e2e tests here - https://github.com/StackStorm/st2/pull/3215#issuecomment-280003567

st2client/st2client/base.py
```diff
@@ -124,7 +124,7 @@ def get_client(self, args, debug=False):
         password = credentials.get('password', None)
         cache_token = rc_config.get('cli', {}).get('cache_token', False)
 
-        if username and password:
+        if credentials:
```
lakshmi-kannan: One thing I realized is if password is not written to config file and token expires, the code doesn't handle the case correctly. So if password is not present, we have to throw an exception that user needs to re-login with credentials or write password to config.

Kami: As far as "not handling that case correctly" - that's an expect behavior, but yes, as discussed on Slack, we should improve on it and print a message if user doesn't use `-w` option that token will expire in X hours and user will need to re-login.

And yeah, we can perhaps do the same on "token expired error" (advise to re-login), but we don't really know if the exception is related to expired token or it's simply an invalid token so we need to keep that in mind so the message needs to be more advisory.

https://github.com/StackStorm/st2docs/pull/380#discussion_r99888376

lakshmi-kannan: > it's simply an invalid token so we need to keep that in mind so the message needs to be more advisory

Yes, this is what I want. Sorry to confuse with the word "exception". Right now the message we get when a token expires with this change is not ideal. See https://gist.github.com/lakshmi-kannan/31c4f5cbc10f81f382ae76380ae1994c#file-gistfile1-txt-L9

Kami: Updated the code to print a message on successful login in 3418ffea3f2e0edce4d82ef0b8fe67c282ce9330.

```bash
st2 login testu -p testp
Logged in as testu

Note: You didn't use --write-password option so the password hasn't been stored in the client config and you will need to login again in 24 hours when the auth token expires.
As an alternative you can run st2 login command with the "--write-password" flag, but keep it mind this will cause it to store the password in plain-text in the client config file.
```

Mierdin:
@kami @lakshmi-kannan was CI finished when this was merged?

armab:
I think we should enable in GitHub setting which will block merging if some tests are still red/unfinished.
Hopefully this will push us more to improve the build machinery.

Bug introduced after merging this PR was actually caught in `st2-packages` build: https://circleci.com/gh/StackStorm/st2-packages/1886

```sh
/bin/sh -c st2\ run\ core.local\ --\ hostname

Failed to authenticate with credentials provided in the config.
ERROR: 401 Client Error: Unauthorized
MESSAGE: Invalid or missing credentials for url: http://127.0.0.1:9100/tokens
```

when `auth` is disabled in `st2.conf`.

Mierdin:
@armab just for my own sanity - after circle runs basic testing like unit and integration, things eventually go back to circle for actual package build right? Just curious we have checks in place for this or if this would have only shown up post-merge.

armab:
@Mierdin You're right.
Actually in your PR https://github.com/StackStorm/st2/pull/3226 `8 successful checks` is a good demonstration how it should work.

PR #3223, opened 2017-02-15
Adjust Codecov behavior

armab:
Avoid Codecov spammy behavior with inaccurate coverage% comments and wrong red GitHub PR statuses.

See https://docs.codecov.io/docs/commit-status


Kami:
Thanks.

I do like the comments, but yeah, they do use no good until the numbers are more accurate.

armab:
I'd like https://github.com/codecov/support/wiki/Pull-Request-Comments too.

Maybe it's a signal of worse problems somewhere inside st2 or in a way how we're using testing framework?
Because it always shows 2 different coverage numbers when re-running the tests.

bigmstone:
@armab created #3227 since it will go unnoticed if it's not captured anywhere.

PR #3226, opened 2017-02-16
Change conditional to point to username field in client config

Mierdin:
https://github.com/StackStorm/st2/pull/3219 changed the logic of when a token is retrieved from the API by looking for the 'credentials' field of the configuration.

However, when this section is missing (or when the config is omitted entirely, like it is in CI), this field is set to an empty dictionary, and then because both `username` and `password` default to None and then are added to this dict, the statement `if credentials:` will never evaluate to False, because this dict will never be empty.

In CI, this causes problems We want to skip this step when the config is missing, because in CI this is expected and other API methods will work just fine because auth is disabled. This change proposes looking at just specifically the `username` field, because even though `st2 login` omits the `password` field by default, it still writes the `username` field so can be relied upon.

### DIFF
--- a/st2client/st2client/base.py
+++ b/st2client/st2client/base.py
@@ -124,7 +124,7 @@ class BaseCLIApp(object):
         password = credentials.get('password', None)
         cache_token = rc_config.get('cli', {}).get('cache_token', False)
 
-        if credentials:
+        if username:
             # Credentials are provided, try to authenticate agaist the API
             try:
                 token = self._get_auth_token(client=client, username=username, password=password,


armab:
Considering all efforts and [discussions](https://stackstorm.slack.com/archives/stackstorm/p1487201208639889) around this bug, do you think it's possible to include a test case for this particular scenario?

Kami:
Thanks, good catch.

And yes, I agree, we should add a test case for that (so it doesn't try to authenticate when: 1. no credentials are set, 2. only password is set but not the username).

st2client/st2client/base.py
```diff
@@ -124,7 +124,7 @@ def get_client(self, args, debug=False):
         password = credentials.get('password', None)
         cache_token = rc_config.get('cli', {}).get('cache_token', False)
 
-        if credentials:
+        if username:
```
Kami: Now that I think more about it - this will probably work for v2.2.0, but we need to add some more robustness and logic to handle more scenarios.

Good example of scenario which is not handled correctly right now is a case when only a `username` is provided in the config and there is no cached token. In such scenario we don't want to perform an authentication with the auth API since password is mandatory (but that's what happens right now).

Kami:
In short, if we are "in a rush" with v2.2.0, I would be fine with merging this and then work on additional robustness + test cases mentioned in https://github.com/StackStorm/st2/pull/3226/files#r101514299 in the very near future (and we can release that as v2.2.1).

(I looked into more "quick fixes" + tests myself, but it requires more time to do it correctly and not something which should be rushed so we either ignore that for now or make it a blocker for v2.2.0 and delay the release)

st2client/st2client/base.py
lakshmi-kannan: This entire piece of code is a mess. We've added several layers of "patch" and we should refactor this post release.

Kami: Agreed - and we should start with tests and making sure we handle all the edge cases.

Mierdin:
Thanks all for the discussion. In the interest of moving 2.2 forward, I'm going to merge this, but I created https://github.com/StackStorm/st2/issues/3232 to ensure we address additional testing and corner cases in 2.2.1

PR #3204, opened 2017-02-10
Initial draft of the st2 sdk module [WIP][DO NOT MERGE]

m4dcoder:
Introduce a new st2forge library that will be uploaded to pypi. The st2forge library contains a base python Action class that the python runner will recognize. The st2common.runners.base_action.Action class will be deprecated in future releases. Sample pack using st2forge and tox to run pep and unit tests: https://github.com/m4dcoder/stackstorm-influxdb

PR #3238, opened 2017-02-21
Fix st2-run-pack-tests script

Kami:
This pull request fixes two issued I discovered while testing the packages:

1. Make sure we activate virtualenv before we try to upgrade pip (previously it would try to install it into global namespace).
2. Make sure we add /opt/stackstorm/st2/bin to $PATH so it works on package installations (previously this wasn't added to $PATH so the script wouldn't work out of the box on package based installations)/

### DIFF
--- a/st2common/bin/st2-run-pack-tests
+++ b/st2common/bin/st2-run-pack-tests
@@ -55,10 +55,11 @@ STACKSTORM_VIRTUALENV_BIN="/opt/stackstorm/st2/bin"
 ####################
 
 function usage() {
-    echo "Usage: $0 [-x] [-j] -p <path to pack>" >&2
+    echo "Usage: $0 [-x] [-j] [-v] -p <path to pack>" >&2
     echo "  -x : Do not create virtualenv, e.g. when running in existing one." >&2
     echo "  -j : Just run tests (use previously installed dependencies" >&2
     echo "       and virtualenv if any. For subsequent test runs." >&2
+    echo "  -v : Verbose mode (log debug messages to stdout)." >&2
 }
 
 while getopts ":p:xjv" o; do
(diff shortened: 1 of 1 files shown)


Kami:
On a related note - I also need to make a change to install `st2tests` with packages.

This way running `st2-run-pack-tests` will work out of the box on package based installations which means we don't need to rush #3204 too much (imo).

Kami:
^ I will work on thing described above for v2.2.1 (it should be quite a straight forward change).

st2common/bin/st2-run-pack-tests
```diff
@@ -146,19 +162,24 @@ else
...
-    # Activate created virtualenv
     if [ ! -d "${VIRTUALENV_DIR}" ]; then
         echo "Virtual environment ${VIRTUALENV_DIR} doesn't exist"
         echo "Run this script without the -j flag to create it or run it with -x flag to not use" \
              " a pack specific virtual environment."
         exit 2
     fi
 
+    # Activate created virtualenv
     source ${VIRTUALENV_DIR}/bin/activate
```
bigmstone: I believe this source will always run in addition to the one above. Can we check if activated already and skip this if it's already active?

Kami: Yes, in some situations it's already ran in the if above, but the operation of activating a virtualenv is idempotent so it doesn't hurt if it's activated twice (and it's a bit simpler than using additional if check, but I'm fine either way) :)

Kami: Made a small change to use a function to activate it - 457d718bcedd646e91030bbb5932e9cef0ddbef0

As mentioned above, this might actually be and overkill and harder to read than the previous version (activate operating is idempotent so multiple activate calls don't break / hurt anything).

PR #3242, opened 2017-02-21
Fix st2ctl reload so it preserves exit code from "st2-register-content" script and fails on failure by default

Kami:
There was an inconsistency between default `st2-register-content` script and `st2ctl reload` command behavior.

The former would correctly exit on failure by default, but `st2ctl reload` wouldn't because it didn't correctly preserve and propagate the exit code from the `st2-register-content`.

Issue originally reported by tanya on Slack (good catch!)

## TODO

- [ ] Add st2tests tests - not a blocker for v2.2.0 (we already have tests for st2-register-content, but none for this specific scenario for st2ctl reload)

### DIFF
--- a/st2common/st2common/content/bootstrap.py
+++ b/st2common/st2common/content/bootstrap.py
@@ -63,6 +63,8 @@ def register_opts():
                                                               'all the Python runner actions.')),
 
         # General options
+        # Note: This value should default to False since we want fail on failure behavior by
+        # default.
         cfg.BoolOpt('no-fail-on-failure', default=False,
                     help=('Don\'t exit with non-zero if some resource registration fails.')),
         # Note: Fail on failure is now a default behavior. This flag is only left here for backward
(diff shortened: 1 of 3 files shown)


st2common/st2common/content/bootstrap.py
```diff
@@ -63,11 +63,13 @@ def register_opts():
...
 
         # General options
+        # Note: This value should default to False since we want fail on failure behavior by
+        # default.
         cfg.BoolOpt('no-fail-on-failure', default=False,
                     help=('Don\'t exit with non-zero if some resource registration fails.')),
         # Note: Fail on failure is now a default behavior. This flag is only left here for backward
         # compatibility reasons, but it's not actually used.
-        cfg.BoolOpt('fail-on-failure', default=False,
+        cfg.BoolOpt('fail-on-failure', default=True,
```
lakshmi-kannan: Damn. How did we miss this?

Kami: This is not actually a real issue - real issue was with `st2ctl reload` not propagating this exit code.

This flag is not used anymore at all, but it was left there for backward compatibility so it didn't break thing. I only changed it to `True` and added more explicit comment to make the whole thing less confusing to people reading the code.

Kami:
@Mierdin OK to merge this into master and include in v2.2.0?

Or I need to wait for v2.1.1?

Kami:
Also confirmed this change works with latest packages.

Kami:
Merging this now that v2.2.0 has been released (after syncing changelog).

Issue #3243, opened 2017-02-21
Remove medium-strength ciphers from default nginx configuration


LindsayHill:
We should update the ssl_ciphers line in our nginx configuration: 
https://github.com/StackStorm/st2/blob/master/conf/nginx/st2.conf#L35

We should remove EDH-RSA-DES-CBC3-SHA, DES-CBC3-SHA and ECDHE-RSA-DES-CBC3-SHA 

This currently triggers medium-severity Nessus alerts.

LindsayHill:
Addressed in #3244

PR #3244, opened 2017-02-21
remove medium-strength ciphers

LindsayHill:
Changes our default NGINX configuration to not accept medium-strength ciphers

Addresses #3243


Kami:
LGTM.

How does this affect browser compatibility?

LindsayHill:
As best as I can tell, it would only be IE on XP that would be affected (https://github.com/client9/sslassert/wiki/IE-Supported-Cipher-Suites). I think we can safely let those clients go? 

Those people who really need to support older ciphers can still go and re-enable them, but I think it's important for us to have strong defaults.

PR #3246, opened 2017-02-22
Disable CORS in development config

enykeev:
This madness needs to end.


Kami:
O, hai, CORS is back.

enykeev:
It never went away. I've been stumbling upon this config option for years now. Every time I need to consume an API from fresh devbox in browser, I have to change this properly locally and then restart the server. Drives me insane every single time.

Kami:
LGTM

Issue #3250, opened 2017-02-25
StackStorm-Exchange hosted packs using BaseActionAliasTestCase fail with an exception


jjm:
When running alias tests for a pack on StackStorm Exchange they fail with the following error:

```
======================================================================
ERROR: test_geoip (test_action_aliases.GeoIP)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/st2/st2tests/st2tests/action_aliases.py", line 44, in setUp
    self.action_alias_db = self._get_action_alias_db_by_name(name=self.action_alias_name)
  File "/tmp/st2/st2tests/st2tests/action_aliases.py", line 108, in _get_action_alias_db_by_name
    action_alias=alias_path)
  File "/tmp/st2/st2common/st2common/bootstrap/aliasesregistrar.py", line 118, in _get_action_alias_db
    (pack, pack_field))
Exception: Model is in pack "stackstorm-networking_utils" but field "pack" is different: networking_utils
```

I've encountered this on a PR StackStorm-Exchange/stackstorm-networking_utils#1.

Kami:
Thanks again for reporting this.

As commented on the other issue, I found the root cause and I will work on a fix on Monday :)

PR #3256, opened 2017-02-27
Fix base action alias test case so it works if local pack directory name doesn't match the pack name

Kami:
This pull request fixes a bug which would cause action alias tests to fail when local pack directory name doesn't match `pack` attribute in different resource metadata files.

This can be the case with the new pack management when you are locally developing a pack and local git repository directory name is not the same as `pack` / `ref` attribute (e.g. `stackstorm-network_utils` vs `network_utils`).

Again, this only affects local development and not actual StackStorm installations when packs are installed into `/opt/stackstorm/packs`. When pack is downloaded (git clone), correct directory name is used (one which matches `pack` / `ref` attribute) - if you run tests on `/opt/stackstorm/packs/<pack>` directory, they will work just fine.

Resolves https://github.com/StackStorm/st2/issues/3250 (https://github.com/StackStorm-Exchange/stackstorm-networking_utils/pull/1#issuecomment-282509188).

Thanks to @jjm for catching and reporting this issue.

## TODO

- [x] Tests

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,9 @@ Changelog
 in development
 --------------
 
+* Fix base action alias test class (``BaseActionAliasTestCase``) so it also works if the local pack
+  directory name doesn't match the pack name (this might be the case with new pack management
+  during development where local git repository directory name doesn't match pack name) (bug fix)
 
 2.2.0 - February 22, 2017
 -------------------------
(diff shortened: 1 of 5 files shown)


armab:
> everything is green


https://images-na.ssl-images-amazon.com/images/I/613teLHSSCL._UX385_.jpg
:v:

Kami:
@armab 

I'm +1 for that once we fix the last two or so major intermediate failures which sadly still happen quite often on e2e tests (mistral cancel stuff and network downloading package from pkgcloud repos).

PR #3259, opened 2017-02-28
Fix workflow status when task is canceled

m4dcoder:
Fix Mistral workflow status when task is canceled. Currently, when a task is canceled, the workflow status is set to error. The workflow status should be set to canceled. Also, when a canceled action execution completes, the action execution will be updated from canceled to its new status. This should not be the case because the action execution has already been canceled. Before merging this patch, please ensure mistral fork is caught up with mistral upstream master. This requires another fix from mistral upstream.

### DIFF
--- a/st2common/st2common/util/action_db.py
+++ b/st2common/st2common/util/action_db.py
@@ -20,7 +20,7 @@ from mongoengine import ValidationError
 import six
 
 from st2common import log as logging
-from st2common.constants.action import LIVEACTION_STATUSES
+from st2common.constants.action import LIVEACTION_STATUSES, LIVEACTION_STATUS_CANCELED
 from st2common.exceptions.db import StackStormDBObjectNotFoundError
 from st2common.persistence.action import Action
 from st2common.persistence.liveaction import LiveAction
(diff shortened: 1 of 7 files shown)


Kami:
Good catch - the st2 code change looks good to me :+1:

st2common/st2common/util/action_db.py
```diff
@@ -192,6 +192,12 @@ def update_liveaction_status(status=None, result=None, context=None, end_timesta
     LOG.debug('Updating ActionExection: "%s" with status="%s"', liveaction_db.id, status,
               extra=extra)
 
+    # If liveaction is already canceled, then do not allow status to be updated.
+    if liveaction_db.status == LIVEACTION_STATUS_CANCELED and status != LIVEACTION_STATUS_CANCELED:
```
Kami: I guess this would probably indicate a programmer / user error or similar so throwing might be more correct (and propagate this error to the API / user if user tried to cancel it via API), but then again I could see the way everything it's currently implemented just making it a no-op if execution is already canceled is simpler and better.

Kami:
Just to confirm - this should fix intermediate `examples.mistral-test-cancel` Robot test failures we are seeing?

I assume it should, but I still see those tests intermediately failing on build server...

PR #3262, opened 2017-03-01
st2auth handler - Constrain split on : to first instance

nzlosh:
When a user password contains `:` the authorisation fails because the split breaks the string into list with more than 2 elements.  Adding 1 to the split constrains it to the first occurrence of the `:`.  Caveat, this will still break if the users name contains a `:`.

Thanks to @ftoussenel for finding this bug.


Kami:
Thanks that's a good catch - can you please also add a test case for it?

nzlosh:
No problem, I'll add a test case as soon as I can.

PR #3263, opened 2017-03-02
Make st2tests directory a proper Python package

Kami:
This is a first step needed which will allow us to create Python package from `st2tests` and include it in the rpm / deb distributions.

Including it in those distributions, will among other things, allow users to run `st2-run-pack-tests` directly on StackStorm package based installations.


enykeev:
What I don't like about that all is that we have too much st2tests that has little to do with each other. I wonder if we should merge it somehow with e2e tests.

PR #3269, opened 2017-03-04
Fix concurrency unit tests to allow tooz upgrade

m4dcoder:
Due to unit tests running logic on the same process, locking via older version of tooz result in unit tests being blocked and unable to complete.


Kami:
Great work on the fix :)

I spent a lot of time tracking it down two times overall. Last time, I tracked it down to the MockLiveActionPublisher running in the same process which causes a dead-lock because of double locking, but gave up after that because it was too time consuming and too big of a rabbit hole (thought it was a small thing I could fix quickly on the side :D)

st2tests/st2tests/mocks/liveaction.py
```diff
@@ -57,3 +58,36 @@ def publish_update(cls, payload):
...
             print(payload)
+
+
+class MockLiveActionPublisherNonBlocking(object):
+
+    @classmethod
+    def publish_create(cls, payload):
+        try:
+            if isinstance(payload, LiveActionDB):
+                eventlet.spawn(scheduler.get_scheduler().process, payload)
```
Kami: :+1:

Kami:
I pushed a small change to make the Circle CI tests pass - 98d148691d5604667848a5021314efb764db7843

With new versions of `tooz`, tooz doesn't depend on 3rd party libraries for different backends anymore and as such, they must be explicitly installed.

Kami:
On a related note - now that we support new version of tooz we should also document how users can configure it to use other new supported backends such as consul or etcd.

Basically what user has to do is:

1. Install library for the backend on which the backend depends on into st2 virtual environment - e.gt.`/opt/stackstorm/st2/bin/pip install consul`
2. Update `coordination.url` in st2.conf
3. Profit!

Edit: I just had a look and we already document that some backends require 3rd party libraries to be installed manually (e.g. redis). So only thing we should probably do is link to tooz docs and / or list supported backends and which corresponding library needs to be installed for each backend in question.

PR #3273, opened 2017-03-06
Adding trigger to render context for rules

### LINKED ISSUE #3272: Failed rendering payload value into pattern
Hi,

Following is a rule's example. In this snippet, I would like to compare two elements from payload. In this specific case, **host** and **master**. Log shows `UndefinedError: 'trigger' is undefined`. It looks like trigger is not being rendered.

```
trigger:
  type: "core.st2.webhook"
  parameters:
    url: "serf"
criteria:
  trigger.body.event:
    type: "equals"
    pattern: "member-failed"
  trigger.body.host:
    type: "equals"
    pattern: "{{ trigger.body.master }}"
```

Thanks!

bigmstone:
Fixes #3272


Kami:
Let's please add some test cases for that.

Besides that, + changelog entry I see no other issues :+1:

bigmstone:
@kami working on them. Types aren't handled correctly (which my tests caught) so working through that. Everything is rendered as a string. Probably going to have to use the same voodoo I used for complex type rendering in params unless you have a better idea.

Kami:
Thanks for working on that and sadly I don't have a better idea :+1:

st2reactor/st2reactor/rules/filter.py
```diff
@@ -149,7 +154,36 @@ def _render_criteria_pattern(self, criteria_pattern):
...
+            criteria_pattern,
+            criteria_context
+        )
+
+        to_complex = False
+
+        if len(re.findall(r'{{[A-z0-9_.-]+}}', criteria_pattern)) > 0:
+            LOG.debug("Rendering Complex")
+            criteria_pattern = re.sub(
+                r'{{([A-z0-9_.-]+)}}', r'{{\1 | to_complex}}',
```
lakshmi-kannan: 🤕

lakshmi-kannan: As per offline conversation, to_complex transformation is needed to preserve the data type of the pattern so the operator works the right data types.

st2reactor/st2reactor/rules/filter.py
```diff
@@ -149,7 +154,36 @@ def _render_criteria_pattern(self, criteria_pattern):
...
 
-        criteria_pattern = render_template_with_system_context(value=criteria_pattern)
+        LOG.debug(
+            'Rendering criteria pattern (%s) with context: %s',
+            criteria_pattern,
+            criteria_context
+        )
+
+        to_complex = False
+
```
m4dcoder: Can you provide description here why we are looking for jinja templates here and what is the expected outcome for the substitution? It's not straightforward to understand.

st2reactor/st2reactor/rules/filter.py
```diff
@@ -149,7 +154,36 @@ def _render_criteria_pattern(self, criteria_pattern):
...
+        if len(re.findall(r'{{[A-z0-9_.-]+}}', criteria_pattern)) > 0:
```
m4dcoder: Also, what if there are spaces in between the curly braces (i.e. `{{ foobar }}`)?

st2reactor/st2reactor/rules/filter.py
```diff
@@ -149,7 +157,38 @@ def _render_criteria_pattern(self, criteria_pattern):
...
+        )
+
+        to_complex = False
+
+        # Check if jinja variable is in criteria_pattern and if so lets ensure
+        # the proper type is applied to it using to_complex jinja filter
+        if len(re.findall(r'^{{\s*[A-z0-9_.-]+\s*}}$', criteria_pattern)) > 0:
+            LOG.debug("Rendering Complex")
+            criteria_pattern = re.sub(
+                r'^{{\s*([A-z0-9_.-]+)\s*}}$', r'{{\1 | to_complex}}',
```
m4dcoder: Can you give me an example of this substitution?  Sorry for being a pain in the behind. This does not immediately click so need to see an example. Thanks.

st2reactor/st2reactor/rules/filter.py
bigmstone: Good catch.

st2reactor/st2reactor/rules/filter.py
bigmstone: Done

st2reactor/st2reactor/rules/filter.py
bigmstone: It will match `{{ test.var }}` and `{{ one_var.two_var }}`. It will not match `{{ variable | my_filter }}` or anything more complex. It also won't match `pre{{ variable }}post`

bigmstone: Also, anything not matched is still rendered, just rendered as a string.

CHANGELOG.rst
```diff
@@ -23,6 +23,9 @@ in development
   tooz for coordination. (improvement)
 * Allow user to specify which branch of ``st2tests`` repository to use by passing ``-b`` option to
   ``st2-self-check`` script. (improvement)
+* Allow use of webhook payload in value context.
```
Kami: Please provide more context (now it's not really clear to what this applies to).

E.g. "Add webhook payload to the context which is available to Jinja variable inside rule criteria section" or similar.

st2reactor/st2reactor/rules/filter.py
```diff
@@ -149,7 +157,38 @@ def _render_criteria_pattern(self, criteria_pattern):
...
+            'Rendering criteria pattern (%s) with context: %s',
+            criteria_pattern,
+            criteria_context
+        )
+
+        to_complex = False
+
+        # Check if jinja variable is in criteria_pattern and if so lets ensure
+        # the proper type is applied to it using to_complex jinja filter
+        if len(re.findall(r'^{{\s*[A-z0-9_.-]+\s*}}$', criteria_pattern)) > 0:
```
Kami: I believe I've seen similar piece of code and regex in your other PR - is it possible to refactor some of that in common function / similar or was it a similar but different thing?

Kami: It also seems like we use the same regular expression multiple times so it would make sense to compile it once and re-use compiled version (top-level on module load) to speed things up a bit.

On a related note - any ideas if this adds any significant negative performance overhead (I believe I asked a similar thing in other PR, but don't remember the answer)?

st2reactor/st2reactor/rules/filter.py
```diff
@@ -149,7 +157,38 @@ def _render_criteria_pattern(self, criteria_pattern):
...
+            context=criteria_context
+        )
+
+        LOG.debug(
+            'Rendered criteria pattern: %s',
+            criteria_pattern
+        )
+
+        if to_complex:
+            return json.loads(criteria_pattern)
```
m4dcoder: So are we sure that criteria_pattern here is always a json string?

st2reactor/st2reactor/rules/filter.py
m4dcoder: So i have an issue here, maybe just need more explanation.  So are we sure that the result of the render for the variable substitution will be a json object that can then be json.dumps in to_complex?

m4dcoder:
So from what i can tell, this patch only support simple variable substitution in the jinja rendering in the criteria pattern. Again, once user starts using complex jinja expressions, this falls apart pretty quickly. If we are ok w/ this, please make sure user is aware of this limitation in the documentation.

bigmstone:
Latest change supports complex jinja. So long as the resulting complex jinja is valid JSON it will preserve types and return a value of the correct type. Including dicts, lists, etc. If it not valid JSON if falls back to just treating it as a string.

m4dcoder:
LGTM. To be honest, it's hard to follow up on since it's been awhile since this was last reviewed. I see that you added some tests to support complex jinja. Good. Please make sure docs is updated to explain this feature clearly. Thanks.

estee-tew:
https://stackstorm.slack.com/archives/community/p1492789516864680

PR #3275, opened 2017-03-07
Fix a bug with the remote command and script runner not closing SSH connections after execution has finished

### LINKED ISSUE #3274: remote-shell-cmd does not close ssh connections
Hi,

If I run a `core.remote` action an ssh connection opens and it does not close after the action finishes successfully. Running the same action again and again opens new connections until the remote server refuses to accept more.

A simple bash one-liner to reproduce this:

```
for i in {1..1000}; do st2 run core.remote hosts=127.0.0.1 whoami; done
```

And the result:

```
$ netstat -a | grep "tcp \+0 \+0 \+localhost:ssh" | wc -l
1000
```

I tried it on a fresh Ubuntu 16.04 machine with a fresh install of StackStorm 2.2.0.

BR,
k.

Kami:
This pull request fixes a bug with remote command and script runner not closing SSH connections after the action execution has finished.

The issue was easy to reproduce, but I imagine the reason why we didn't see more reports about this is that default SSH max connection limit is quite high and using a default config, SSH connections are automatically killed if they are idle for more than some default amount of time.

 /cc @lakshmi-kannan Just wanted to confirm we are on the same page since you primarily worked on this change.

Right now we instantiate new `ParallelSSHClient` instance for each action invocation which means it makes no sense to keep the connections open at the end because we instantiate new instance for each action run. There is still some benefit with re-using connections inside the same action execution because single remote script action run consists of running multiple commands over SSH and not opening a new connection for each command run during the same action execution is beneficial and faster.

One option would also be to have a global singleton instance of `ParallelSSHClient` with some maximum number of hosts and this way we could potentially cache and keep connections open for longer time across different action executions. This would cause some benefits in case user runs multiple remote command / script actions in a short-time frame, but it has many edge cases (making sure this singleton has some limit of connections so it doesn't cause memory leaks, making sure we handle timeouts and reconnect correctly, etc.). so I would like to avoid making those changes now if it's not necessary.

Resolves #3274.


lakshmi-kannan:
What a bug?! Kinda shameful. Thanks for the fix. I agree that caching connections might be beneficial but it introduces more code logic. We can do that optimization if needed in the future.

Kami:
@m4dcoder Yeah I was thinking about tests which should ensure `post_run` is always called but this is really runner container behavior and something which should be tested there (and not part of the specific implementation test which relies on that contract / functionality).

Having said that, I will have a look and if we dont have such tests for runner container yet, I will add some.

Kami:
@m4dcoder I checked that code again - all the stuff which is called inside finally either does:

1. Re-throw the exception - indicates a bigger issue so we intentionally don't call `post_run`.
2. `self._clean_up_auth_token` already ensures is never throws an exception because it catches all the exceptions internally.

One thing I will do is add a comment to clarify that behavior. I could also add another try / except around `self._clean_up_auth_token` but this would be weird and confusing because `self._clean_up_auth_token` should never throw.

m4dcoder:
+1 for now. This patches for the most part fixes what it's intended. But there's exception here which will still lead to ssh session not closed. Please track this somewhere so we can come back to it.

PR #3277, opened 2017-03-07
Refactor the action execution asynchronous callback

m4dcoder:
Refactor the action execution asynchronous callback functionality into the runner plugin architecture. Currently the mistral callback handler is under the st2actions module whereas the rest of the mistral runner is separated in a plugin directory. The asynchronous callback is runner specific and hence the logic should be located with the runner.


m4dcoder:
This PR is dependent on https://github.com/StackStorm/st2mistral/pull/26.

st2common/st2common/util/loader.py
```diff
@@ -202,6 +203,21 @@ def register_query_module(module_name):
     return QUERIER_MODULES_CACHE[module_name]
 
 
+def register_callback_module(module_name):
+    base_path = cfg.CONF.system.base_path
+    module_path = os.path.join(
+        "%s/runners/%s/callback/%s.py" % (base_path, module_name, module_name)
```
Kami: This will actually combine everything as a string and not use `os.path.join` since you are passing a single argument (string) to `os.path.join`.

Should be:

```python
os.path.join(base_path, "runners", module_name, "callbacks", module_name + '.py')
```

Kami: Just noticed we do the same thing in over place so it was probably copied over and needs to be fixed.

/cc @bigmstone

PR #3279, opened 2017-03-08
[RFR] New style changelog

LindsayHill:
I'd like to add a little bit of structure to our changelogs. Not a huge difference, just a simple structural change, using the Ansible Changelog as a model (https://github.com/ansible/ansible/blob/devel/CHANGELOG.md)

For each version section, we should have up to 3 sub-sections:
- "Major Changes" - anything that we want to highlight to users. They don't have to necessarily be huge code changes, but they could be anything of interest to users.
- "Minor Changes" - smaller bug fixes, many of them only relevant to a small group, or maybe not even noticed by users.
- "Deprecations" - anything where we are removing support for something. This section will include both giving notice of upcoming changes, and notifying when we do actually remove support for something.

Note: If we don't have any relevant entries for that section, we can just skip it - e.g. if there are no deprecations, then leave the section out altogether.

We could also have "API Changes" as a section, or just put it in "Major Changes" section.

Each entry in the changelog should be similar to what we do now - i.e. a short description, along with "bug fix" / "improvement", and an issue or PR #.

I've updated the in development and 2.2 sections as an example. I'll probably retrospectively change v2.1 and v2.0 once we agree on a format.


CHANGELOG.rst
```diff
@@ -3,46 +3,58 @@ Changelog
 
 in development
 --------------
+
+Major Changes:
+~~~~~~~~~~~~~~
+
+* Fix ``st2ctl reload`` command so it preserves exit code from `st2-register-content` script and
```
armab: As you said: 
> "Major Changes" - anything that we want to highlight to users. They don't have to necessarily be huge code changes, but they could be anything of interest to users.

It's a bit confusing and counter-intuitive for me. Eg. these changes are Minor and we call them "Major".
Maybe any other ideas how to name this section?

LindsayHill: "Important" changes? Dunno

armab: +1 "Important" sounds better

lakshmi-kannan:
I am personally meh on this change. This might make things easier for you when you want to write a blog but as an engineer I find it hard to classify a change as a "major" vs "minor" when we iterate slowly for a "major" change.

LindsayHill:
@lakshmi-kannan Our current style is to have a wall of text for all changes, listed in reverse chronological order. That is fine for developers adding comments as they go, but it's not just about me writing a release blog - this is about StackStorm users trying to figure out what changes they really need to pay attention to, and which ones are less important. Right now that's really hard for people to figure out - from a quick look at docs.stackstorm.com/changelog.html, can you easily identify the changes that you need to pay attention to?

bigmstone:
I think it's grey on how to define major/minor changes. Amount of code changed? Impact to user? I can make a one-line change that greatly impacts user. Or rewrite python wrapper to make it 10x faster w/ no impact to user. I'm fine with this change but we need clear criteria on what is major and what's not if we want to have success with it.

LindsayHill:
To me, major is more about impact rather than lines of code. Decision criteria might be "Is this something that's important to draw the user's attention to?" Could be security fixes, could be notable bug fixes - e.g. something that affected quite a few ppl, would definitely be things like APIv2. Would also be new features that many users could benefit from - e.g. the recent Jinja/Mistral changes. 

I'll take a look at some other projects to see if I can find a sensible set of guidelines to follow.

bigmstone:
That'd be great. I also like the question "is this something the user should be aware of." but we have to be careful because we don't want every little feature in there.

LindsayHill:
Here's an alternative approach. Note that it does not separate minor/major, but breaks out by Added/Changed/Deprecated/etc. 

From http://keepachangelog.com/en/0.3.0/ 

> Each version should:
> - List its release date in the above format.
> - Group changes to describe their impact on the project, as follows:
> - `Added` for new features.
> - `Changed` for changes in existing functionality.
> - `Deprecated` for once-stable features removed in upcoming releases.
> - `Removed` for deprecated features removed in this release.
> - `Fixed` for any bug fixes.
> - `Security` to invite users to upgrade in case of vulnerabilities.

bigmstone:
I like that better. It's more "things" but it's not a mystery where things fit.

LindsayHill:
I'll push another version in the next day or so that matches that style for the last couple of releases, so we can see how it looks.

If that style works for us, it will be handy to be able to point to that site as our reference style. Can put it in our contribution guidelines

armab:
> Here's an alternative approach.
> ...

Nice!
We follow similar approach in [`ansible-st2` CHANGELOG](https://github.com/StackStorm/ansible-st2/releases/tag/v0.6.0) where list is sorted by: 

- Changes
- New
- Enhancements
- Fixes

Where `Changes` are in the top and they really mean "Breaking Changes" which can affect users.
Then it goes list of `New` features.

LindsayHill:
I've re-worked all versions from v2.0 onwards to match the "keepachangelog.com" style, with Added/Changed/Fixed/Removed/Deprecated/Security as required.

How does this work for people? Think we can manage to do that level of categorization?

bigmstone:
@LindsayHill I think this is a good change. You'll get my +1 when conflicts are resolved w/ `master`.

bigmstone:
That also means @lakshmi-kannan needs to give his -1 before me if he doesn't want it in. :trollface:

LindsayHill:
Thanks @bigmstone. I'll do a rebase against latest changelog once we get one more +1

LindsayHill:
Once we've released v2.2.1, I'll update this to resolve conflicts, and to reflect updated ordering if required (#3288)

bigmstone:
@LindsayHill Do we have consensus? Reason I ask is this file is ever flowing so as soon as you edit it the next time we should merge asap. Otherwise this will be a never ending cycle. Just want us all to be :+1: to the change before you put in the effort again.

LindsayHill:
@bigmstone I think we have sufficient +1s, and lack of -1s, so I think we can do it. My plan is to update this once we release v2.2.1, and then merge. You're quite right about not getting stuck in endless cycle. Had hoped we'd have 2.2.1 out the door last week, but it's pretty close now.

bigmstone:
Good with me. Just wanted to prevent doubling up work for you. :taco:

LindsayHill:
OK, this has now been brought up to date, **and** I've gone right back to the start, and changed the format for the entire changelog. 

Cleaned up a handful of other duplicate, typo and formatting issues along the way.

If there are no complaints, I will merge this by Wednesday at the latest.

LindsayHill:
Sigh. Looks like the RST checker doesn't like my re-use of headings. I'll have to do some investigation into how to modify the RST checker options

Kami:
Looks like that's just an info message so perhaps just changing rstcheck target to run with `--report warning`` so only issues treated as warning and above are treated as fatal errors should do it.

CHANGELOG.rst
```diff
@@ -4,6 +4,9 @@ Changelog
 in development
 --------------
 
+Changed:
```
Kami: Small more of a personal preference thing - maybe get rid of semi-colon?

Besides that, LGTM

Kami: Also, since changelog entries tend to contain present form and not past, maybe we should use same format for those section names as well - New features / Improvements / Additions / Fixes

LindsayHill: Yeah, actually I might get rid of semicolon. Will push a new commit for that

For the headings (Added/Changed/Fixed/etc), I'm using the format from here: http://keepachangelog.com/en/0.3.0/ - it's not perfect, but at least it's a referenceable style guide

LindsayHill:
@kami agreed with changing rst-checker to fail on warning & above. Should do the trick

CHANGELOG.rst
Kami: If we are following some kind of standard them I'm kinda OK with it even though "Changes" is a bit weird once since it's more like a parent category since every other thing (fix, security) is also a change that's why I would prefer "Improvements" and "New features" and a categories instead.

PR #3280, opened 2017-03-08
Introduce "capabilities" and expand functionality provided by the auth backends

Kami:
This pull request expands functionality offered by auth backends and introduces a concept of "capabilities".

Right now every available auth backend provides user authentication information and has a single capability (can authenticate user).

With this "capability" functionality now in place, auth backends can advertise and provide additional functionality such as "provide additional information about the user" (`get_user(username)` method) and "provide information about the user group membership" (`get_groups(self, username)`) method.

The latter is needed for the LDAP Group to RBAC role synchronization which will happen automatically on user authentication.

The idea behind capabilities it to make auth backends more generic and generally applicable and not couple it to the LDAP backend implementation.

In the future, users could extend other backends to provide user group membership information (e.g. flat file backend could read it from flat file, MongoDB one could read it from MongoDB and some other backend could read user group membership information from some third party service or product).

As mentioned above, right now each existing backend provides "authenticate user" functionality, but in the future it would also be possible to have a backend which, for example, just provides user group membership information.


st2auth/st2auth/backends/base.py
```diff
@@ -28,10 +30,31 @@ class BaseAuthenticationBackend(object):
     Base authentication class.
     """
 
+    # Capabilities offered by the auth backend
+    CAPABILITIES = (
+        AuthBackendCapability.CAN_AUTHENTICATE_USER
```
Kami: This is done for backward compatibility reasons (so we don't need to update each existing backend).

Having said that, I also plan to update each existing backend in the near future to explicitly spell out capabilities offered by it.

lakshmi-kannan:
I get where you are going with this. Your APIs get_user() and get_groups() kinda assume unix style permissions model for auth backends. This may be true for the likes of LDAP and PAM but I am not sure if these are generally applicable. Did you research on some other backends to get a sense? 

I think having capabilities exposed and a base class is fairly straight forward way to implement this. I am trying to make sure the base class has the right methods. Some research before doing the initial commit would be nice.

Kami:
@lakshmi-kannan Yeah, `get_user` was already in place previously, but it wasn't really implemented yet (it's implemented in enterprise LDAP backend now, pr to follow soon).

I took some inspiration from Django and other frameworks which expose similar methods for auth backends.

As far as other implementations goes - yes, same approach could work with other backends as well (e.g. with file backend you read groups from some file and there are many SaaS authentication and user management services out there which also fit this model well and also the PAM model you mentioned).

Kami:
@m4dcoder it looks like an intermediate error / failure.

I'm personally getting sick of re-running stuff manually when it fails. It's a stupid task, and also an immensely stupid waste of everyone's time.

We already spent tons of time fixing intermediate test failures and fixed most of it, but it looks like some times it still happens. 

For the tests which still occasionally fail - we should prioritize tracking down the root cause and fix it (or if we can't, disable / remove those tests)..

For other intermediate test failures - I added policy in place in the past which seemed to have alleviated the problem, but you removed it so we are back to step one.

Kami:
I made a small change to the API - renamed `get_groups()` to `get_user_groups()` to make it immediately clear from the method name that this method retrieves groups for a particular user (and not, for example, retrieve all the available groups or similar).

dzimine:
Read, understood, and OK with the approach: 

while I don't think there is more in auth backends than users and groups I am most likely wrong :) and +1 on looking for inspirations with common frameworks as this problem has been long solved. 

Also, consider glancing at SaltStack and how they solve the same.

Kami:
And tests passed now after some new changes and re-runs :)

/cc @m4dcoder

PR #3281, opened 2017-03-09
Update st2-run-pack-tests to only install global test dependencies when ST2_REPO_PATH variable is set

Kami:
This pull request updates `st2-run-pack-tests` script to only install global pack test dependencies when `ST2_REPO_PATH` environment variable is set.

When this variable is set, it means the script is likely running standalone outside of StackStorm package installation environment.

With recent changes and improvements to this script, the script now runs out of the box on new StackStorm package based installations where those global pack test dependencies are already available (they are installed and provided by `st2tests` package).

This change speeds things up when running multiple pack tests on a box where global pack test dependencies are available (e.g. `make .pack-tests` is now 10 seconds or so faster).

### DIFF
--- a/st2common/bin/st2-run-pack-tests
+++ b/st2common/bin/st2-run-pack-tests
@@ -30,6 +30,10 @@ function join { local IFS="$1"; shift; echo "$*"; }
 ###########
 
 # A list of dependencies which are installed and available to every pack tests
+# Note: Those only need to be installed when running this script standalone
+# outside of StackStorm package installation.
+# When using StackStorm package installation, those dependencies are already
+# available (they are installed by st2tests package).
 PACK_TEST_PYTHON_DEPENDENCIES=(
     'mock>=1.3.0,<2.0'
     'unittest2>=1.1.0,<2.0'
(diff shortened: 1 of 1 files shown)


st2common/bin/st2-run-pack-tests
```diff
@@ -205,9 +214,14 @@ if [ "${JUST_TESTS}" = false ]; then
     fi
 
     # Install global test dependencies
+    INSTALLED_PIP_PACKAGES=$(pip list)
```
Kami: Just a heads up - invoking pip is slow so we should try to avoid invoking it multiple times and try to do everything in one go where it makes sense and it doesn't sacrifice readability (similar to `apt-get` command runs, it involves reading a local database, etc. which is slow).

st2common/bin/st2-run-pack-tests
```diff
@@ -205,9 +214,14 @@ if [ "${JUST_TESTS}" = false ]; then
...
+
     echo "Installing global pack test dependencies..."
-    for dependency in ${PACK_TEST_PYTHON_DEPENDENCIES[@]}; do
-        pip install --cache-dir ${HOME}/.pip-cache ${ST2_PIP_OPTIONS} "${dependency}"
+    for((i = 0; i < ${#PACK_TEST_PYTHON_DEPENDENCIES_NAMES[@]}; i++)); do
+        DEPENDENCY_NAME="${PACK_TEST_PYTHON_DEPENDENCIES_NAMES[$i]}"
+        DEPENDENCY_VERSION="${PACK_TEST_PYTHON_DEPENDENCIES_VERSIONS[$i]}"
+
+        echo ${INSTALLED_PIP_PACKAGES} | grep "${DEPENDENCY_NAME}"  > /dev/null || pip install --cache-dir ${HOME}/.pip-cache ${ST2_PIP_OPTIONS} "${DEPENDENCY_NAME}${DEPENDENCY_VERSION}"
```
bigmstone: :heart:

Kami: Yeah, not ideal, `pip install` is already idempotent if package is installed, but it still needs to read the local database, cache, etc. so actual pip invocation is slow as hell (https://github.com/StackStorm/st2/pull/3281#discussion_r105163117).

PR #3286, opened 2017-03-10
Make mistral querier into a polling sensor

m4dcoder:
Prototype to move the very sluggish Mistral querier into a smarter polling sensor.

### DIFF
--- a/contrib/mistral/actions/update_workflow_status.py
+++ b/contrib/mistral/actions/update_workflow_status.py
@@ -1,0 +1,85 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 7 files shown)


bigmstone:
This is kinda awesome. Good work.

Kami:
I personally like the idea of "dog fooding" and re-using existing primitives where appropriate instead of (re)inventing some new approach :+1: 

If we dog food and we find issues (e.g. rules engine can't handle that many rules per second or similar) we will make it better for everything and everyone (not just that particular use case) when we fix the issue / deficiency / whatever we have found.

On the contrary - introducing other approaches aka one-offs means more code to write and maintain which definitely means less robust code and more bugs. Yes, I know it sometimes makes sense (mostly for performance and related reasons) to go with a different "new" approach, but I still prefer to re-use existing strong and stable primitives where possible (think about Redis - it provides very few commands and basic primitives but they are very robust and stable and you can build almost anything on top).

That's why 2 years or so ago when we worked on initial ChatOps implementation I was against the current approach and for using a sensor + rules but in the end my approach didn't win the argument :)

PR #3270, opened 2017-03-05
Actionable readme

dzimine:

PR #3266, opened 2017-03-02
Changelog update for medium-strength cipher removal

LindsayHill:

PR #3265, opened 2017-03-02
Make st2-run-pack-tests command work out of the box on servers with StackStorm installed

Kami:
More fixes needed to get it to work:

1. Make sure we run `nose` with `--exe` binary so it also runs tests in test files which are executable (packs.download automatically makes all files executable)
2.  Add missing `psutil` dependency
3. Various improvements to `st2-run-pack-tests` script (make sure StackStorm virtualenv is in PYTHONPATH for test run, etc.)

## TODO

- [ ] Add st2tests for st2-run-pack-tests script

PR #3287, opened 2017-03-12
Branch for v2.2.1 release

Kami:
Pull request with changes for v2.2.1 release (bug fixes and improvements).

It includes the following changes:

* Fix st2ctl reload so it preserves exit code from "st2-register-content" script and fails on failure by default - #3242
* Remove medium-strength ciphers - #3244, #3266
* Fix base action alias test case so it works if local pack directory name doesn't match the pack name - #3256
* Fix a bug with default values from config schema not being used if pack doesn't have a config - #3260
* #3263, #3264
* Make st2-run-pack-tests command work out of the box on servers with StackStorm installed  - #3265
* Make st2tests directory a proper Python package - #3268
* Fix concurrency unit tests and upgrade tooz -  #3269
* Fix a bug with authentication not working if password contained a colon (:) - #3271
* Fix a bug with the remote command and script runner not closing SSH connections after execution has finished - #3275
* Update st2-run-pack-tests to only install global test dependencies when ST2_REPO_PATH variable is set - #3281
* Fix "PUT /v1/configs" API endpoint not working when RBAC is enabled - #3285
* Actionable readme - #3270


Kami:
@hagay3 Yes, v2.2.1 release should be out in a week or two.

PR #3288, opened 2017-03-12
Fix changelog ordering

Kami:
Noticed some things are not ordered correctly while working on #3287 (probably due to bad merges).


armab:
BTW take a look at https://github.com/StackStorm/st2/pull/3279

PR #3294, opened 2017-03-16
Support for automatic RBAC role assignments based on the remote auth backend user group membership information

### LINKED ISSUE #2509: StackStorm should support LDAP groups for RBAC
StackStorm RBAC roles should mirror LDAP roles when using ldap. At least it should be possible to do so with some configuration and mapping.

In general too it would be nice if StackStorm could "slurp" up roles from an external provider for RBAC purposes.

Kami:
This pull request introduces support for automatically assigning RBAC roles to users based on the remote auth backend groups user group membership information (#2509) - for example, based on which LDAP groups a particular user is a member of.

To get all of that to work, a couple of new concepts have been introduced.

## Auth Backend Capabilities and User Group Membership Information

Each auth backend can now advertise which capabilities it offers. In addition to that, new "has groups" capability has been introduced. This capability allows auth backend to return which remote groups a particular user is a member of.

This has been implemented as part of #3280 and this functionality is also already implemented for enterprise LDAP auth backend (the only one which is a priority for us right now).

## Remote Auth Backend Group to StackStorm RBAC Role Mappings

Auth backend now provides us with the user remote group membership information, but to be able to utilize this in combination with RBAC roles a new concept has been introduced.

This concept allows operators to specify how those remote groups map to StackStorm roles.

This mapping is, like other RBAC definitions, contained in a YAML file inside ``/opt/stackstorm/rbac/maps`` directory (actual directory name, etc. is TBD).

The definition currently looks like this:

`/opt/stacsktorm/rbac/maps/stormers.yaml`

```yaml
---
  group: "CN=stormers,OU=groups,DC=stackstorm,DC=net"
  description: "Automatically grant st2_admin and role_one to all stormers group members."
  roles:
    - "role_one"
    - "st2_admin"
```

In the example above, each StackStorm user which is part of the `CN=stormers,OU=groups,DC=stackstorm,DC=net` LDAP group will automatically be granted two StackStorm roles (`role_one`, `st2_admin`) when they successfully authenticate.

Those role assignments are special in a sense that they are remote - unlike local role assignments which are read from files on-disk and are synced when user runs ``st2-apply-rbac-definitions`` tool, those are synced when user successfully authenticates.

We could potentially also sync those when user runs that tool, but syncing it when user authenticates makes more sense and re-uses the existing access token system (the tool itself really shouldn't be doing many network calls to LDAP and other external systems) and that's how most other similar systems works as well (sync and operations like that happen on authentication).

Those remote role assignments are valid for as long as the user access token is valid. When access token expires, user needs to re-authenticate to retrieve a new one and then those remote roles assignments are synchronized again.

If for some reason, someone needs this to happen more often, they can decrease the default access token TTL. This is also no different if a user is removed from an external auth system (e.g. user leaves the company and is removed from LDAP). That user will still be able to access StackStorm for as long as access token is still valid. If that's a problem, operator can always purge all the currently valid access token, or just an access token for user which has left.

## Operator / admin workflow

The general workflow for the operator to get this functionality working and configured would look something like this:

1. Operator writes role definitions or simply re-uses existing system roles if those are sufficient. Nothing new, same as before.

2. Operator enables remote group to RBAC role sync in ``st2.conf`` (``rbac.sync_remote_group`` needs to be set to ``True``).

3. Operator writes remote group to local StackStorm role mapping files.

4. Operator runs ``st2-apply-rbac-definitions`` tool so the all mapping file definitions are loaded in the database. This tool needs to run each time a mapping is updated (created / updated / removes).

5. Profit! When user authenticates, they will automatically get StackStorm RBAC roles assignments based on the mappings specified in those files and based on the remote auth backend group membership.

Note: This whole workflow can, like most everything else, of course also be automated (e.g. ise some config management tool to write those files on disk and run ``st2-apply-rbac-definitions`` when operator pushes those to some git repo or similar).

## Next steps - Let's agree on the high level concepts and naming

Before I proceed with finishing and polishing the code, I would ask you all to please look at the high level concepts and terms, think about it and then come back and leave your feedback about the actual high level concepts and terms.

The code itself is just an early prototype and not finished and polished yet. Please **DO NOT** comment on the actual code diff and code related stuff yet and focus on high level concepts and terminology / naming (commenting on it now would be wasting everyone time because the concepts are not fully finalized yet).

I opened this PR early on so we have a basic prototype people can play with to get an idea of how it works and so we can all think and agree about the new terminology and naming before I can actually work on finishing this prototype and making it production ready.

Once that is done and I will finalize the code I will let you all know and then you can start commenting on the actual code stuff and go crazy about it :P

The prototype still needs some changes so it will work end to end. It should work end to end by Monday EOD, but I would still prefer to agree on the concepts, naming, etc. before we share actual working prototype (with full test coverage, documentation, etc.) with our users and customers.

## TODO

- [x] Decide on the concepts, naming, paths, etc. (making sure naming, etc. is consistent with existing primitives, etc.)
- [x] Update all the new code so it uses consistent naming for variables, method, models, concepts, etc.
- [x] Polish and refactor the code
- [x] Finish the code
- [x] Tests (yes, this needs a lot of tests)
- [x] Documentation
  - [x] Document new concepts, terms, etc.
  - [x] Document operator flow
  - [x] Document how to enable and configure this functionality

Resolves #2509.


dzimine:
OK by me, but one thing not clear: 
> Operator enables remote group to RBAC role sync in st2.conf (rbac.sync_remote_group needs to be set to True).

1) Looking at the code sketch, the idea seem to be to get the groups if the flag is on, and skip it all together if not. But is there a case for not needing to get the groups?  Can we rely on backend capabilities instead, and fetch the groups if backend is capable? 

2) the word remote" groups smells: if there are local groups do we work with them differently?

lakshmi-kannan:
I like how this is shaping up. But I have some minor concerns: 

1. I don't think we should create another folder for maps. It can go in /opt/stackstorm/rbac/assigments/. But converting the file to appropriate model would require us to check if ``user`` is specified or ``group`` is specified. I don't think this should be a problem. This will simplify UX a little because I already have two folders to deal with. 

2. The word "sync" is overused so things get slightly confusing.There are multiple syncs happening - 1. Sync between DB and disk. (st2-apply-rbac-definitions) 2. Sync between auth backend and auth system to get groups info for user. (LDAP query once until auth token is valid) 3. Sync auth groups to DB so we can use this to validate RBAC permissions (LDAP query once until auth token is valid). Did I get (3) right? This piece of code is TODO right now so I wanna make sure I got this part right. I think what you did with tying the lifetime of user group to TTL of auth token is pretty smart.

3. I do not think this config ``rbac.sync_remote_group`` should even exist. I don't see why anyone would run it as disabled. So in short, I agree with dzimine's (1).

Kami:
@dzimine 

> Looking at the code sketch, the idea seem to be to get the groups if the flag is on, and skip it all together if not. But is there a case for not needing to get the groups? Can we rely on backend capabilities instead, and fetch the groups if backend is capable?

We do actually rely on both (capabilities and additional flag to enable the actual group sync). The idea behind also having a flag is that it's possible that backend can provide user group membership information (e.g. LDAP), but the operator doesn't want to use the actual group to role synchronization functionality (in addition to that it also acts as a feature flag).

In theory, we could just rely on the capabilities and still retrieve user groups, but because there are no mappings defined no role assignments would be performed. The ideal behind having a flag is totally skip the user group retrieval in such situation (it speeds things up and avoids unnecessary network call).

> the word remote" groups smells: if there are local groups do we work with them differently?

Yeah, I'm also not a big fan of the term, but I couldn't come up with anything better.

It basically needs to be a term which makes it clear those are not local StackStorm roles, but remote ones provided by auth backend. Ideas for better name are welcome. Perhaps `is_auth_backend` group or similar, dunno.

Kami:
@lakshmi-kannan 

> I don't think we should create another folder for maps. It can go in /opt/stackstorm/rbac/assigments/. But converting the file to appropriate model would require us to check if user is specified or group is specified. I don't think this should be a problem. This will simplify UX a little because I already have two folders to deal with.

Yes I was thinking about re-using the existing directory, but this would mean we would need to support two different file formats in this directory which might cause even more confusion.

I'm still open to this idea though so if you have some suggestions on how this definition for "group to role" mappings file in the `/opt/stackstorm/rbac/assigments/` should look like, let me know.

> The word "sync" is overused so things get slightly confusing.There are multiple syncs happening - 1. Sync between DB and disk. (st2-apply-rbac-definitions) 2. Sync between auth backend and auth system to get groups info for user. (LDAP query once until auth token is valid) 3. Sync auth groups to DB so we can use this to validate RBAC permissions (LDAP query once until auth token is valid). Did I get (3) right? This piece of code is TODO right now so I wanna make sure I got this part right. I think what you did with tying the lifetime of user group to TTL of auth token is pretty smart.

Correct.

All that `st2-apply-rbac-definitions` does is synchronizes database state so it matches the definitions specific on disk. So this tool only performs local sync.

But yes, now that we introduced those group to role mappings we need to perform another sync which happens on authentication - we need to read the mappings and then write additional use role assignments to the database based on those mappings.

As mentioned above, those assignments are special in a sense that they are "remote" (ideas for better term are welcome).

So in short, `st2-apply-rbac-definitions` behavior doesn't change so we can keep the sync term and as fa as documenting how the whole flow works, yeah, we should maybe think about it and come up with some other term or similar so it will be very clear how it works and what is going on.

> I do not think this config rbac.sync_remote_group should even exist. I don't see why anyone would run it as disabled. So in short, I agree with dzimine's (1).

Describes my reasoning in reply to dzmine.

In short - sometimes you want to disable it and also acts as a feature flag.

Kami:
The code is getting close to the stage when it can be reviewed (it has a good amount of tests so I'm feeling enough confident to make any refactoring if needed)  and later on, shared with our users and customers who wish to be early testers for this feature.

I plan to work on the documentation once the PR is reviewed and any feedback / comments are addressed.

st2auth/st2auth/handlers.py
```diff
@@ -161,17 +163,53 @@ def handle_auth(self, request, headers=None, remote_addr=None, remote_user=None,
...
+                except Exception as e:
+                    # Note: Failed sync is not fatal
+                    LOG.exception('Failed to synchronize remote groups for user "%s"' % (username),
+                                  extra=extra)
+                else:
+                    LOG.debug('Successfuly synchronized groups for user "%s"' % (username),
+                              extra=extra)
+
+                return token
+            return token
```
lakshmi-kannan: Can you return early for the case when sync_groups is not enabled? I feel like there is a scope for future programmer error here.

st2auth/st2auth/validation.py
```diff
@@ -0,0 +1,44 @@
...
+
+    # 2. Verify that auth backend used by the user exposes group information
+    if cfg.CONF.rbac.enable and cfg.CONF.rbac.sync_remote_groups:
+        auth_backend = get_auth_backend_instance(name=cfg.CONF.auth.backend)
+        capabilies = getattr(auth_backend, 'CAPABILITIES', ())
+        if AuthBackendCapability.HAS_GROUP_INFORMATION not in capabilies:
+            msg = ('Configured auth backend doesn\'t expose user group information. Disable '
+                   'remote group synchronization or use a different backend which exposes '
+                   'user group membership information.')
+            raise ValueError(msg)
```
lakshmi-kannan: Is there a reason you didn't define custom exceptions and relied on ValueError?

st2auth/st2auth/validation.py
```diff
@@ -0,0 +1,44 @@
...
+    # 1. Verify correct mode is specified
+    if cfg.CONF.auth.mode not in VALID_MODES:
+        msg = ('Invalid auth mode "%s" specified in the config. Valid modes are: %s' %
+               (cfg.CONF.auth.mode, ', '.join(VALID_MODES)))
+        raise ValueError(msg)
+
+    # 2. Verify that auth backend used by the user exposes group information
+    if cfg.CONF.rbac.enable and cfg.CONF.rbac.sync_remote_groups:
+        auth_backend = get_auth_backend_instance(name=cfg.CONF.auth.backend)
+        capabilies = getattr(auth_backend, 'CAPABILITIES', ())
```
lakshmi-kannan: Why is this an attr and not get_capabilities()? Did you not @property it?

st2common/st2common/models/db/rbac.py
```diff
@@ -76,9 +82,26 @@ class PermissionGrantDB(stormbase.StormFoundationDB):
...
+    Attribute:
+        group: Name of the remote auth backend group.
+        roles: A reference to the local RBAC role names.
+        description: Optional description for this mapping.
+    """
+    group = me.StringField(required=True, unique=True)
+    roles = me.ListField(field=me.StringField())
+    description = me.StringField()
+    enabled = me.BooleanField(required=True, default=True,
+                             help_text='A flag indicating whether the mapping is enabled.')
```
lakshmi-kannan: Surprise linter ain't puke at you.

st2common/st2common/rbac/loader.py
```diff
@@ -114,6 +117,26 @@ def load_user_role_assignments(self):
...
+
+        :rtype: ``dict``
+        """
+        LOG.info('Loading group to role map definitions from "%s"' % (self._role_maps_path))
+        file_paths = self._get_group_to_role_maps_file_paths()
+
+        result = {}
+        for file_path in file_paths:
+            LOG.debug('Loading group to role mapping from: %s' % (file_path))
+            group_to_role_map_api = self.load_group_to_role_map_assignment_from_file(
```
lakshmi-kannan: Dude, this is disaster waiting to happen without explicit try catch.

st2common/st2common/rbac/syncer.py
```diff
@@ -206,6 +211,28 @@ def sync_users_role_assignments(self, role_assignment_apis):
         LOG.info('User role assignments synchronized')
         return results
 
+    def sync_group_to_role_maps(self, group_to_role_map_apis):
+        LOG.info('Synchronizing group to role maps...')
+
+        # Retrieve all the mappings currently in the db
+        group_to_role_map_dbs = rbac_services.get_all_group_to_role_maps()
```
lakshmi-kannan: This is not a paginated call. It might be fine for now but I want to call out there is a potential problem here.

st2common/st2common/rbac/loader.py
Kami: Wdym? It follows exactly the same pattern as other loaders.

And failing to load a mapping is fatal anyway so the exception should be propagated - aka fail loudly and not silently ignore real issues.

st2common/st2common/rbac/syncer.py
Kami: It's intentional - this call should not be paginated since it's used internally by the syncer and not exposed externally via the API.

st2auth/st2auth/validation.py
Kami: It's really a fatal run-time error aka something which is fatal and not intended to be caught so it doesn't make much sense to introduce a new exception class for it.

So if anything, because it's a runtime error I could simple use `RuntimeError` instead of `ValueError`.

st2auth/st2auth/handlers.py
Kami: I actually personally usually prefer to return early, but I just followed the existing style so I went with this approach.

I'm fine either way aka fine with changing it.

st2auth/st2auth/validation.py
Kami: @lakshmi-kannan It's for backward compatibility reasons - e.g. you use old auth backend class version which doesn't declare `CAPABILITIES` class variable.

And because of historical cross-module related dependency issues, auth backend classes don't inherit from any base class so that's about only (fully backward compatible) option we have at the moment.

st2common/st2common/models/db/rbac.py
Kami: Yeah, should have, will fix the indentation.

st2auth/st2auth/validation.py
lakshmi-kannan: Interesting point and a legit one. Thanks for explaining.

st2common/st2common/rbac/loader.py
lakshmi-kannan: Don't we usually have logic around fail on failure and no fail on failure?

lakshmi-kannan: Anyways, I think I see your point and agree it should fail loudly.

st2common/st2common/models/db/rbac.py
```diff
@@ -76,9 +82,26 @@ class PermissionGrantDB(stormbase.StormFoundationDB):
     permission_types = me.ListField(field=me.StringField())
 
 
+class GroupToRoleMappingDB(stormbase.StormFoundationDB):
```
lakshmi-kannan: I didn't see a stand alone unit test for this in st2common/unit/test_db_group_role_mapping.py. Did I miss it? I thin k it's important to add one.

st2common/st2common/rbac/syncer.py
```diff
@@ -267,3 +295,105 @@ def _sync_user_role_assignments(self, user_db, role_assignment_dbs, role_assignm
...
+        role_assignment_dbs_to_delete = [role_assignment_db for role_assignment_db
+                                         in remote_assignment_dbs
+                                         if role_assignment_db.role in role_names_to_delete]
+
+        UserRoleAssignment.query(user=user_db.name, role__in=role_names_to_delete,
+                                 is_remote=True).delete()
+
+        # 3. Create role assignments for all the current groups
+        created_assignments_dbs = []
+        for mapping_db in enabled_mapping_dbs:
```
m4dcoder: Why not update role for all_mapping_dbs? If the mapping is enabled, the information is already there and as recent as possible.

m4dcoder:
Conceptually, this looks ok. So if I understand this correctly, on auth request, there will be a remote group lookup and the user's role assignment will be updated per the group to role mapping. For LDAP specially, does this support group nesting (i.e. user is member of groupB, groupB is member of groupA, and groupA is mapped to a st2 role)?

st2common/st2common/rbac/syncer.py
Kami: I follow the approach of the existing syncer, but I plan to refactor it the near future - https://github.com/StackStorm/st2/issues/3310

Keep in mind that the end result is the same, the current approach just creates a bit more object churn but will refactor that in the near future.

Kami:
@m4dcoder 

> Conceptually, this looks ok. So if I understand this correctly, on auth request, there will be a remote group lookup and the user's role assignment will be updated per the group to role mapping.

That's correct.

> For LDAP specially, does this support group nesting (i.e. user is member of groupB, groupB is member of groupA, and groupA is mapped to a st2 role)?

Nested groups are currently not supported, but if needed, implementing it would mean changing the implementation of `get_user_groups` method on the LDAP auth backend to also return all the nested groups.

st2common/st2common/models/db/rbac.py
Kami: I can add it if needed, but those DB models itself don't implement any custom functionality or methods so I think such tests are very low value.

We already have tests for other important stuff and with testing model create / save / update we are basically just testing mongoengine functionality and we already test that in other places.

Kami:
@lakshmi-kannan To make you happy I added model db tests in - a58b3b3cdecfffd7aa8c7fbaa38007fcbea3061a.

Again, as mentioned before, I believe those tests are basically useless and high cost to maintain. They basically test mongoengine functionality. We should either have single test if we don't trust mongonegine API contract or similar or we should have higher level tests which actually test things and provide value (well, that's already the case as mentioned before - I already added high value tests).

In any case, I couldn't stand this awful copy and paste for DB model tests so I refactored it in a base class - 791e04cff6f34cf9b827d87bd0ca7aa1456b572a.

I also plan to update other DB model tests to use that class instead of copy and pasting things over.

estee-tew:
https://stackstorm.slack.com/archives/community/p1491236836148106

PR #3300, opened 2017-03-23
Refactor endpoints to get rid of kwargs

enykeev:
As @Kami pointed in https://github.com/StackStorm/st2/pull/2727#discussion_r103643612, using `**kwargs` in controllers require user to constantly keep in mind the list of parameters available for the particular controller or get back and forth between code and spec.

Luckily, as it turns out, we no longer need to use `**kwargs` in most cases and it's just a matter of refactoring.

The point of this PR is to replace the use of kwargs in every controller it is possible to do and justify it's use in the rest of them.

### DIFF
--- a/st2api/st2api/controllers/resource.py
+++ b/st2api/st2api/controllers/resource.py
@@ -118,19 +118,22 @@ class ResourceController(object):
 
         self.get_one_db_method = self._get_by_name_or_id
 
-    def get_all(self, **kwargs):
-        return self._get_all(**kwargs)
+    def get_all(self, sort=None, offset=0, limit=None, **raw_filters):
+        return self._get_all(sort=sort,
+                             offset=offset,
+                             limit=limit,
+                             raw_filters=raw_filters)
 
     def get_one(self, id):
         return self._get_one_by_id(id=id)
 
     def _get_all(self, exclude_fields=None, sort=None, offset=0, limit=None, query_options=None,
-                 from_model_kwargs=None, **raw_filters):
+                 from_model_kwargs=None, raw_filters=None):
         """
         :param exclude_fields: A list of object fields to exclude.
         :type exclude_fields: ``list``
         """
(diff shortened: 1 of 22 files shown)


enykeev:
We can not completely get rid of `**raw_filters` in `.get_all` due to how filters are currently implemented in v1 API. We can however be mindful of it and save this issue for v2.

Kami:
Nice, better :+1:

st2api/st2api/controllers/resource.py
```diff
@@ -312,16 +311,6 @@ def _get_by_name_or_id(self, name_or_id, exclude_fields=None):
 
         return resource_db
 
-    def _get_from_model_kwargs_for_request(self, **kwargs):
```
Kami: IIRC, there was a reason I added this method in the past (something with the messy cross-test state being preserved because it first was a class variable and not instance variable), but if tests and everything pass I'm fine with removing it.

Kami:
LGTM :+1:

st2api/st2api/controllers/resource.py
```diff
@@ -118,19 +118,22 @@ def __init__(self):
...
 
     def _get_all(self, exclude_fields=None, sort=None, offset=0, limit=None, query_options=None,
-                 from_model_kwargs=None, **raw_filters):
+                 from_model_kwargs=None, raw_filters=None):
         """
         :param exclude_fields: A list of object fields to exclude.
         :type exclude_fields: ``list``
         """
-        raw_filters = copy.deepcopy(raw_filters)
+        raw_filters = copy.deepcopy(raw_filters) or {}
```
Kami: I guess `copy.deepcopy` also works on `None` type so it's probably fine and we don't need additional if.

And I would imagine it also doesn't cause any negative performance overhead since `None` is a built-in type and `copy.deepcopy` probably has some optimization built-in for such types.

PR #3303, opened 2017-03-24
Set loglevel for nosetests to INFO

enykeev:


enykeev:
I don't like the fact that we have to add `setup.cfg` to the repo root either, but that's how `nosetests` work, it only looks for `./setup.cfg`, `~/.noserc` or `~/nose.conf` by default.

Kami:
Will check it out locally.

Kami:
Yeah, much better, thanks.

Can you please also update developer documentation (testing section) and clarify how to set log level to debug if needed (e.g. `nosetests --logging-level DEBUG` ...)?

Kami:
> I don't like the fact that we have to add setup.cfg to the repo root either, but that's how nosetests work, it only looks for ./setup.cfg, ~/.noserc or ~/nose.conf by default.

Can't we use ``--logging-level`` CLI argument?

enykeev:
We can, but you'll also have to do that for every manual call of nosetests which may prove annoying. Better to have sane default and go from there.

PR #3304, opened 2017-03-24
Fix traceback duplication in ErrorHandlingMiddleware

enykeev:
`LOG.exception` already includes a traceback. No need to log it a second time.


Kami:
LGTM

PR #3308, opened 2017-03-25
Preinitialize modules during route registration

enykeev:
That was the initial intention, but it somehow slipped through the cracks.

If you don't import controller modules during server initialization, it would happen on the first request to the controller within the module. It would a) create a delay on this first request and b) cause a race on endpoints like `/webhooks` or `/stream` which require more time to properly initialize.

### DIFF
--- a/st2common/st2common/router.py
+++ b/st2common/st2common/router.py
@@ -178,6 +178,9 @@ class Router(object):
                     for transform in transforms[filter]:
                         m.connect(None, re.sub(filter, transform, path), **connect_kw)
 
+                    module_name = endpoint['operationId'].split(':', 1)[0]
+                    __import__(module_name)
+
         for route in sorted(self.routes.matchlist, key=lambda r: r.routepath):
             LOG.debug('Route registered: %+6s %s', route.conditions['method'][0], route.routepath)
 


armab:
:+1:
But IMO for such perf improvements, a small 1-line code comment with reasoning would be nice.

st2common/st2common/router.py
```diff
@@ -178,6 +178,9 @@ def add_spec(self, spec, transforms):
                     for transform in transforms[filter]:
                         m.connect(None, re.sub(filter, transform, path), **connect_kw)
 
+                    module_name = endpoint['operationId'].split(':', 1)[0]
```
Kami: Let's please also add a comment here and clarify why this is needed (basically what you wrote in the PR description).

Kami: Just curious how much "overhead" this adds to st2api and st2auth startup time.

enykeev: You talking about overhead comparing to state before or after 2727? We've imported them during init before the spec thing to we should be back to the same level we were before the PR.

enykeev: As for comment, I'll added with some other change if I don't forget. But hey, there's always `git blame`, use it.

PR #3309, opened 2017-03-27
Fix router to only check rbac on endpoints with security

enykeev:
Auth service were raising 500 on endpoint with no security definition (`/auth/v1/tokens`) when RBAC is enabled.

Since RBAC always require auth to be enabled, no reason to check RBAC and authentication flags independently.

Also, all RBAC related tests should now flip auth flag too.

### DIFF
--- a/st2api/tests/base.py
+++ b/st2api/tests/base.py
@@ -63,6 +63,8 @@ class APIControllerWithRBACTestCase(FunctionalTest, CleanDbTestCase):
     Base test case class for testing API controllers with RBAC enabled.
     """
 
+    enable_auth = True
+
     @classmethod
     def setUpClass(cls):
         super(APIControllerWithRBACTestCase, cls).setUpClass()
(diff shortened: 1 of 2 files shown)


st2api/tests/base.py
```diff
@@ -63,6 +63,8 @@ class APIControllerWithRBACTestCase(FunctionalTest, CleanDbTestCase):
     Base test case class for testing API controllers with RBAC enabled.
     """
 
+    enable_auth = True
```
Kami: A bit confused, I don't see this variable being used anywhere in this code.

Kami:
Is it also possible to add some kind of test case for it?

st2api/tests/base.py
enykeev: The variable is used in FunctionalTest._do_setUpClass few lines highier. It mocks auth config option.

enykeev:
You want me to do a test case for a test case? With the changes I made in `st2common/st2common/router.py ` the test now fails and this is the reason I've added the property - to fix broken test.

Issue #3307, opened 2017-03-24
Erroneously trying to parse blank pack configs


Mierdin:
When a pack contains a configuration schema, but that file is blank (has no contents), st2 will still try to parse it, which causes a traceback for the pack.

The workaround right now is simply to have the user delete the schema file - but we should consider implementing some exception handling here.

PR #3311, opened 2017-03-27
Fix a bug with empty config schemas

Kami:
This pull request fixes an issue reported by @bigmstone (good catch!)

Config registration would fail if a config schema for a particular pack was empty or invalid.

The actual root cause was that we didn't call `.validate()` on config schema API object on registration and as such we didn't validate config schemas so this way it was possible for bad or empty config schemas to get into the system.

## TODO

- [x] Tests

Closes https://github.com/StackStorm/st2/issues/3307

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -49,6 +49,8 @@ in development
   (new feature)
 * Linux file watch sensor is now disabled by default. To enable it, set ``enabled: true`` in
   ``/opt/stackstorm/packs/linux/sensors/file_watch_sensor.yaml``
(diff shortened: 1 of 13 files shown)


Kami:
The fix already caught an issue with a typo in config schema for `excel` pack I happen to have locally:

```ValueError: Failed to register pack "excel": Additional properties are not allowed ('requried' was unexpected)
```

Should be `required`, not `requried`.

Kami:
Will merge and also cherry-pick into v2.2 branch.

Thanks.

PR #3318, opened 2017-03-29
Fix dependency version specified in st2-run-pack-tests-tool

Kami:
Noticed a small issue while testing latest v2.2.1 packages.

The issue would only affect people who directly use this tool from git master, but not people who run it on a system where StackStorm was installed using packages.

In any case, I would still propose to include this in v2.2.1.


warrenvw:
@Kami Thanks for comment on how to test! Very helpful.

PR #3328, opened 2017-04-03
Create forloop chain in the example pack

Carles-Figuerola:
This is an example chain for a job that needs to be done in several steps because the bulk of each of the passes is bottlenecked (in the case I found, arguments between python calls cannot exceed 131072 characters). The "get_data" and the "push_data" actions can be from different packs if need be if they have support for "paging" on their requests.

If this feels valuable I could write something up more detailed as a blogpost to show how each of the steps work and post an example run and why anyone would want to use it. In this case, this is how it looks for now:

forloop_chain

### DIFF
--- a/contrib/examples/actions/chains/forloop_chain.yaml
+++ b/contrib/examples/actions/chains/forloop_chain.yaml
@@ -1,0 +1,43 @@
+---
+  chain:
+    -
+      name: "get_github_page"
+      ref: "examples.forloop_get_github_page"
+      parameters:
+        url: "{{github_organization_url}}"
+        page: "{% if increase_index_and_check_condition is defined %}{{increase_index_and_check_condition.result}}{% endif %}"
+      on-success: "parse_github_repos"
+      on-failure: "failed"
+    -
+      name: "parse_github_repos"
+      ref: "examples.forloop_parse_github_repos"
+      parameters:
+        content: "{{get_github_page.result}}"
+      on-success: "push_github_repos"
+      on-failure: "failed"
+    -
+      name: "push_github_repos"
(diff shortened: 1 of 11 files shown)


estee-tew:
https://stackstorm.slack.com/archives/community/p1491250378854634

Kami:
Thanks for the contribution - will review it shortly.

Kami:
From a high level the change looks good to me, but since we also use code from `examples` pack in various tests it would be better, if possible, that actions in those pack wouldn't perform HTTP requests and only depend on core.local actions.

This way if we ever add tests for those new actions they will be more reliable and won't depend on 3rd party service which could be offline or unavailable because of various issues.

Kami:
> If this feels valuable I could write something up more detailed as a blogpost to show how each of the steps work and post an example run and why anyone would want to use it. 

That would be great. We would be happy to feature a guest post from you about this particular looping use case on our blog.

Feel free to start a draft in google docs / github gist / similar and share it with us once you make some progress and we will work from there :)

Kami:
@lakshmi-kannan I'm personally fine with this change, but as mentioned above, I have some concerns with including it in the default `examples` pack - it depends on external service and parsing HTML which is not so robust (it would break if github changes HTML structure and potentially also fails if github is unavailable / there is a networking issue / similar).

What do others think?

Kami:
I will go ahead and merge this for now, but eventually it would be better if we moved to an example which is less error prone and doesn't depend on external service.

Thanks again.

Kami:
Pushes a small style clean-up - https://github.com/StackStorm/st2/pull/3428

In the future, please use 4 spaces for a tab (some how our automatic lint check didn't catch this).

PR #3333, opened 2017-04-05
Improve spec validation

enykeev:


st2common/st2common/openapi.yaml
```diff
@@ -1247,8 +1245,13 @@ paths:
         '200':
           description: Pack search results.
           schema:
-            items:
-              - $ref: '#/definitions/PacksList'
+            type:
+              - array
+              - object
+            # Ideally, we should return array every time and when we request a pack, it should be an
```
Kami: I agree with your "ideal" scenario :)

Kami:
LGTM.

Is it possible to add some kind of test for it (spec failing validation)?

enykeev:
We need more tests for router in general. Problem is, I really have my hands full at the moment so I don't know if I'll be able to do them before the release. And really, I'm not sure, maybe we should postpone the release due to missing tests.

PR #3334, opened 2017-04-05
Filter out null from /executions/views/filters results (with tests)

Kami:
This pull request works on top of #3193 and adds tests so we can merge that change.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -26,8 +26,8 @@ in development
   tokens and services use special service access tokens with no max TTL limit. (bug fix)
 
   Reported by Jiang Wei. #3314 #3315
-* Fix ``/executions/views/filters`` API endpoint to exclude empty (None / null) filter values.
-  (bug fix)
+* Update ``/executions/views/filters`` API endpoint so it excludes null / None from filter values
+  for fields where ``null`` is not a valid field value. (improvement)
 
   Contributed by Cody A. Ray. #3193
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -26,6 +26,10 @@ in development
   tokens and services use special service access tokens with no max TTL limit. (bug fix)
 
   Reported by Jiang Wei. #3314 #3315
+* Fix ``/executions/views/filters`` API endpoint to exclude empty (None / null) filter values.
+  (bug fix)
+
+  Contributed by Cody A. Ray. #3193
 
 2.2.1 - April 3, 2017
 ---------------------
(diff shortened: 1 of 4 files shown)


codyaray:
Hurray! I've had a ticket floating in my queue to get back to this and write some tests ever since I submitted the PR. I'm glad you beat me to it. Thanks!

Kami:
Just discussion this with @enykeev, we believe that in some cases `null` is indeed a valid value (e.g. `parent`) so we should also modify the code to handle that scenario.

Kami:
@codyaray See the comment above - I need to make some modifications to the code to correctly handle scenarios where null is indeed a valid value:

* parent
* rule
* user

`null` is a valid value for those fields and it's indeed one of the filters on which we rely quite often (e.g. `parent: null` for non-workflow executions, etc).

So basically, we only want to skip `null` values for fields where that is not a valid value (e.g. action).

Kami:
I believe this should do it - 65412b7ae0be19b0d1081add25e51e6c712f2998.

/cc @codyaray

codyaray:
I'm not sure that `null` is valid for Rule. That's where I see the exception in the javascript console most of the time, IIRC.

Kami:
@codyaray I believe rule is not for action executions which are triggered directly by the API.

enykeev:
I can confirm that `rule.name = null` is a valid value due to executions being run manually (with `st2 run` or st2web action tab) will have their `rule` equal to an empty object.

As for the issue (I'm assuming it's https://github.com/StackStorm/st2web/issues/343), I'll need to see where this trace points to in code. Screenshot of pointers to minified code isn't particularly useful in this case. And yeah, I'm very sorry I've dropped the ball on st2web issues. Next time please just ping me on community channel if you think there's something wrong with st2web.

_I really need to fix this github notifications!_

Kami:
@enykeev Thanks for looking / confirming this.

Will add some additional tests to verify that `null` is correctly returned for entities where null is a valid filter value.

Kami:
@codyaray so I was digging into code and tests again and now I'm like 80% convinced it's actually just an UI issue (st2web) so not totally sure we need or want this PR anymore.

enykeev:
Since we already approaching the release, I'd say we should postpone this one at least until the next patch. I was planning to focus on st2web in the next release anyway so you can just assign this one to me and I'll deal with it as soon as I have a minute.

Kami:
@codyaray and others - I was able to "replicate" this issue today.

I'm working on a computer which still has old MongoDB installed (v2.4) and filters tests fail there without this change. I believe it's related to a change in 3.2 because it works just fine there (it seems like MongoDB 3.2 doesn't return None anymore, it could be a bug in v2.4 or similar).

@codyaray which MongoDB version do you use?

Kami:
As noted above, it seems like this null issue only appears with MongoDB 2.4.x, but I will still go ahead and merge this PR.

Thanks again and we will also look into st2web since it seems like another bug lies there.

PR #3335, opened 2017-04-05
Fix / update linux.scp action #1236

### LINKED ISSUE #1236: Fix / update linux.scp action
Currently, `linux.scp` action is mostly useless if you want to download files from a remote server to a local one since it expects a `dest_server` argument.

We have a couple of ways to solve this:
1. Make `scp` action "desination agnostic" - instead of taking `dest_server` parameter, make it just take `source` and `destination` arguments. This way it works both ways (upload, download) since source and destination can either be a remote host with a path or just a path.
2. Rename existing `scp` action to `scp_upload` and add a new inverse `scp_download` action.

nmaludy:
PR to fix #1236

I liked Kami's approach 1). In my mind this is more flexible because the author using the ```linux.scp``` action may not know if this will be an "upload" or a "download" or potentially a "transfer" (```scp user@host1:a.txt user@host2:```).

By using ``` source``` and ```destination```  this accounts of all scenarios and allows for each to be accomplished using the same action. Also, this mimics the way that a user would interact with ```scp``` on the command line.

Now, i know this breaks backwards compatibility. Is this OK, or should i mark ```dest_server``` as  ```required: false``` for the time being and some Jinja logic to insert it in the command if it's specified?

Thoughts?

### DIFF
--- a/contrib/linux/actions/scp.yaml
+++ b/contrib/linux/actions/scp.yaml
@@ -13,10 +13,6 @@
             type: 'string'
             description: 'Destination of files/directories'
             required: true
-        dest_server:
-            type: 'string'
-            description: 'Destination of files/directories'
-            required: true
         recursive:
             type: 'boolean'
             description: 'Boolean flag for recursive copy'
@@ -35,7 +31,7 @@
             default: '/home/stanley/.ssh/stanley_rsa'
         cmd:
             immutable: true
-            default: 'scp {{args}} -i {{keyfile}} {{source}} {{dest_server}}:{{destination}}'
+            default: 'scp {{args}} -i {{keyfile}} {{source}} {{destination}}'
         args:
             description: 'Command line arguments passed to cp'
             default: '-o stricthostkeychecking=no -v{% if recursive == true %} -r{% endif %}{% if force == true %} -f{% endif %}'
--- a/contrib/linux/actions/scp.yaml
+++ b/contrib/linux/actions/scp.yaml
@@ -7,11 +7,11 @@
     parameters:
         source:
             type: 'string'
-            description: 'List of files/directories to to be copied'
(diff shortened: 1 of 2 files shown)


estee-tew:
https://stackstorm.slack.com/archives/community/p1491431996752869

Kami:
Thanks for this change.

I'm fine with it as long as we document this "breaking" change in CHANGELOG.rst and in upgrade notes in st2docs repo (https://github.com/StackStorm/st2docs/blob/master/docs/source/upgrade_notes.rst).

contrib/linux/actions/scp.yaml
```diff
@@ -13,10 +13,6 @@
             type: 'string'
             description: 'Destination of files/directories'
```
Kami: Let's please add an example to the description to make it clear destination can be relative, absolute, or full server:path notation.

Kami: @nmaludy Can you please look into this change and the one mentioned in changelog so we can get it merged once that is resolved?

Thanks :)

nmaludy:
Thanks for the comments. Should be updated now based on your requests.

Kami:
Related st2docs PR - https://github.com/StackStorm/st2docs/pull/500

We also need to update our build servers, etc. before we can merge this because it's a backward incompatible change.

Kami:
As mentioned in https://github.com/StackStorm/st2docs/pull/500#issuecomment-307728838, I decided to make this change backward compatible (so it will work using either approach for a next release or two) because otherwise merging it would be quite messy and we would need to update a lot of our CI / CD related code.

Kami:
Let's continue in https://github.com/StackStorm/st2/pull/3463.

Thanks again.

PR #3341, opened 2017-04-07
Fix secrets

enykeev:


Kami:
Good catch, thanks.

If you need help or want me to work on additional tests for this functionality, let me know and I will do it.

PR #3344, opened 2017-04-10
Allow user to specify a blacklist of attributes which are masked in the log messages

Kami:
By default we use a static blacklist of parameter names (password, token, etc.) which are masked in the log message `extra` if `log.mask_secrets` config option is set to `True`.

This pull request allows administrator to define an additional custom list of attributes which are masked in the log messages.

Related PR https://github.com/StackStorm/st2docs/pull/459.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -26,6 +26,8 @@ in development
   tokens and services use special service access tokens with no max TTL limit. (bug fix)
 
   Reported by Jiang Wei. #3314 #3315
+* Allow user to specify a custom list of attribute names which are masked in the log messages by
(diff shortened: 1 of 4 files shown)


lakshmi-kannan:
We might need this at a per action level. Base runner should expose this attribute. Can you file an issue for this so we can add the feature later?

Kami:
@lakshmi-kannan Yes, we use secret parameters for actions and API level secrets masking, but that's slightly different than log message secrets masking.

It would simply be too expensive if we needed to do a DB query or similar for each log message call.

Also, see related explanation in https://github.com/StackStorm/st2docs/pull/459.

PR #3347, opened 2017-04-11
Fix cors middleware conflicting with logging and error handling

enykeev:
The PR solves a number of problems introduced by #2727.

First, logging middleware fails during OPTIONS preflight request due to us not defining OPTIONS requests in spec. As far as logging concerned, every request to non-existing endpoint still needs to be logged.

Second, cors and error_handling middleware both depends on each other. CORS intercepts and reraises the NotFoundException, but don't provide its headers for endpoints that don't exist (preventing browser from properly handling the error) and error_handling middleware intercepts all the errors making it impossible for cors middleware to properly handle OPTIONS requests.

### DIFF
--- a/st2auth/st2auth/app.py
+++ b/st2auth/st2auth/app.py
@@ -65,9 +65,9 @@ def setup_app(config={}):
 
     app = router.as_wsgi
 
+    app = ErrorHandlingMiddleware(app)
     app = CorsMiddleware(app)
     app = LoggingMiddleware(app, router)
-    app = ErrorHandlingMiddleware(app)
     app = RequestIDMiddleware(app)
 
     return app
--- a/st2auth/st2auth/app.py
+++ b/st2auth/st2auth/app.py
@@ -65,6 +65,7 @@ def setup_app(config={}):
 
     app = router.as_wsgi
 
+    # Order is important. Check middleware for detailed explanation.
     app = ErrorHandlingMiddleware(app)
     app = CorsMiddleware(app)
     app = LoggingMiddleware(app, router)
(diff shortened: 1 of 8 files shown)


st2auth/st2auth/app.py
```diff
@@ -65,9 +65,9 @@ def setup_app(config={}):
 
     app = router.as_wsgi
 
+    app = ErrorHandlingMiddleware(app)
```
Kami: If ordering is important, let's please add a comment which clarifies that.

Kami:
LGTM, but some tests would also be good.

st2auth/st2auth/app.py
enykeev: Do you want me to add `# Ordering is important` or do you want me to explain why exactly one middleware should go before the other? Don't you think that adding such a detailed description in three different places would require us to constantly keep them up to date and will introduce even more confusion than simply reading the code?

Kami: Yeah, adding what and why would be great.

And yes, having it in 3 places in slightly redundant, but this could also mean code could be refactored (if possible).

In any case, I still think it would be good to add it because without a comment (and tests) it's just a question of time until someone starts moving things around when refactoring it or similar...

st2tests/st2tests/api.py
```diff
@@ -41,4 +41,12 @@ def do_request(self, *args, **kwargs):
                 raise ResponseLeakError('Endpoint response contains secret parameter. '
                                         'Find the leak.')
 
+        if 'Access-Control-Allow-Origin' not in res.headers:
```
Kami: :+1:

Issue #3348, opened 2017-04-12
With RBAC enabled, trigger get gives 500:Internal server error 


humblearner:
```
$ st2 trigger get core.st2.webhook
Traceback (most recent call last):
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2client/commands/resource.py", line 162, in get_resource_by_pk
    instance = self.manager.get_by_id(pk, **kwargs)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2client/models/core.py", line 39, in decorate
    return func(*args, **kwargs)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2client/models/core.py", line 204, in get_by_id
    self.handle_error(response)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2client/models/core.py", line 165, in handle_error
    response.raise_for_status()
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/requests/models.py", line 862, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
HTTPError: 500 Server Error: Internal Server Error
MESSAGE: Internal Server Error for url: http://127.0.0.1:9101/v1/triggertypes/core.st2.webhook
Trigger "core.st2.webhook" is not found.
```

st2api.log:
--------------

```
2017-04-12 00:33:57,775 140260897577712 INFO logging [-] 36d75dfb-c2ee-46d0-a375-50f2eba40702 - GET /v1/triggertypes/core.st2.webhook with query={} (remote_addr='127.0.0.1',method='GET',request_id='36d75dfb-c2ee-46d0-a375-50f2eba40702',query={},path='/v1/triggertypes/core.st2.webhook')
2017-04-12 00:33:57,777 140260897577712 AUDIT auth [-] Token with id "58ed5ff055fc8c1ec02d788d" is validated.
2017-04-12 00:33:57,780 140260897577712 ERROR error_handling [-] API call failed: 'NoneType' object has no attribute 'split'
Traceback (most recent call last):
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/middleware/error_handling.py", line 39, in __call__
    return self.app(environ, start_response)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/middleware/logging.py", line 69, in __call__
    retval = self.app(environ, custom_start_response)
[...]

PR #3349, opened 2017-04-12
(RBAC enable) Fix when permission_type is not provided with API call

humblearner:
Fix for #3348

### DIFF
--- a/st2api/st2api/controllers/resource.py
+++ b/st2api/st2api/controllers/resource.py
@@ -222,9 +222,10 @@ class ResourceController(object):
 
         instance = self._get_by_id(resource_id=id, exclude_fields=exclude_fields)
 
-        rbac_utils.assert_user_has_resource_db_permission(user_db=requester_user,
-                                                          resource_db=instance,
-                                                          permission_type=permission_type)
+        if permission_type:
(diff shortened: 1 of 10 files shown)


enykeev:
It seems to me that instead of fixing the problem, it masks it. The problem is that we forget to provide a permission_type for some endpoints. Essentially, `permission_type` method argument should be mandatory. Removing RBAC check from code path creates a security regression.

humblearner:
Couldn't find permission_type for Triggers: https://github.com/StackStorm/st2/blob/master/st2common/st2common/rbac/types.py

Are we using some other type for triggers ?

enykeev:
Yep, I've just stumble on that myself. We indeed don't have RBAC definitions for triggers, rbac roles and pack config schemas - the only three endpoints that use that method. I still think that we should  make `permission_type` argument mandatory and manually set it to None in this three cases to ensure we don't accidentally leak something we do have rbac definition for. I'll add it to this PR.

Kami:
Yeah, I agree with last comment - I think explicitly passing in `None` for endpoints where we don't have RBAC (yet) is better.

We also need tests for this - I can work on it later today.

PR #3351, opened 2017-04-12
Add RBAC implementation for API endpoints which were missing it

Kami:
This pull request adds RBAC implementation for API endpoints which were missing it (endpoints which were added recently where we got sloppy and didn't implement when we were implementing that feature - we should be careful this doesn't repeat again in the future):

- [x] Traces (get all, get one, add RBAC tests for API endpoints)
- [ ] Webhooks (get one, get all, add RBAC tests for API endpoints)
- [x] Timers (get all, get one, add RBAC tests for API endpoints)
- [x] API keys (get one, create, add RBAC tests for API endpoints)

## TODO

- [ ] Docs

### DIFF
--- a/st2api/tests/unit/controllers/v1/test_rbac_for_supported_endpoints.py
+++ b/st2api/tests/unit/controllers/v1/test_rbac_for_supported_endpoints.py
@@ -33,7 +33,7 @@ TEST_FIXTURES = OrderedDict([
     ('sensors', ['sensor1.yaml']),
     ('actions', ['action1.yaml', 'local.yaml']),
     ('aliases', ['alias1.yaml']),
-    ('triggers', ['trigger1.yaml']),
+    ('triggers', ['trigger1.yaml', 'cron1.yaml']),
     ('rules', ['rule1.yaml']),
     ('triggertypes', ['triggertype1.yaml']),
     ('executions', ['execution1.yaml']),
@@ -121,6 +121,7 @@ class APIControllersRBACTestCase(APIControllerWithRBACTestCase):
         enforcement_model = self.models['enforcements']['enforcement1.yaml']
         execution_model = self.models['executions']['execution1.yaml']
         trace_model = self.models['traces']['trace_for_test_enforce.yaml']
+        timer_model = self.models['triggers']['cron1.yaml']
 
         supported_endpoints = [
             # Runners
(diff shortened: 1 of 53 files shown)


st2common/st2common/openapi.yaml
```diff
@@ -3348,7 +3360,13 @@ paths:
   /api/v1/webhooks:
     get:
       operationId: st2api.controllers.v1.webhooks:webhooks_controller.get_all
+      x-permissions: {{ PERMISSION_TYPE.WEBHOOK_LIST }}
       description: Returns a list of all webhooks.
+      x-parameters:
+        - name: user
+          in: context
+          x-as: requester_user
+          description: User performing the operation
```
enykeev: I don't see why you need both. `x-permissions` should be more than enough here.

st2common/st2common/rbac/resolvers.py
```diff
@@ -769,58 +771,136 @@ def user_has_permission(self, user_db, permission_type):
 
     def user_has_resource_api_permission(self, user_db, resource_api, permission_type):
         assert permission_type in [PermissionType.API_KEY_CREATE]
-        # TODO:
-        return True
+        return self._user_has_global_permission(user_db=user_db, permission_type=permission_type)
```
enykeev: Can't find where we using this one.

Kami: It's not being used by WebhooksController.get_all method - https://github.com/StackStorm/st2/pull/3351/commits/aef02a481fb29cfbc45e216eeac3a77782147454#diff-91ccd907e72f23cb1463520e892c92e6R98.

st2common/st2common/openapi.yaml
Kami: Will try it out and verify it works without this parameter.

st2common/st2common/rbac/resolvers.py
enykeev: you've answered on the wrong one

enykeev: I also found what's going on in this one. You can ignore the thread.

st2common/st2common/openapi.yaml
Kami: Good catch, it's indeed not needed - 07e06d4.

st2api/tests/unit/controllers/v1/test_rbac_for_supported_endpoints.py
```diff
@@ -13,13 +13,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from collections import OrderedDict
 import httplib
+from collections import OrderedDict
```
enykeev: I've never been able to understand the logic behind ordering the imports

Issue #3053, opened 2016-11-28
Process config attributes / options in the order defined in the config schema when using "pack config" CLI command


Kami:
Right now attributes are not processed / displayed in the order defined in the config schema when using `st2 pack config` command. This could be a bit confusing to the user.

This ties into the age old "preserve ordering when reading and writing out YAML files" issue.

This is not a big blocker for now, but something we should fix in the future to provide a better user experience.

userlocalhost:
I'm annoyed by this problem. And I want to solve it.
Would you please assign this issue to me?

emedvedev:
@userlocalhost you are more than welcome to take a stab at this issue! Thank you for volunteering :)

userlocalhost:
Thank you for your prompt reply.
I got it. I'll make a PR after finding the way to solve it.

userlocalhost:
If the config data that `st2api` received can store to the datastore, and load it in order. This problem can be resolved.

There is a [PR](https://github.com/MongoEngine/mongoengine/pull/1497) to let `mongoengine` to store object data (e.g. the fields which are defined in 'st2common/model/db/*') to the datastore in order.
If this feature is merged, this problem could be solved with small changes of st2.

So please wait until that PR is settled.

PR #3353, opened 2017-04-14
This PR enables to configure pack based on config schema in the defined order at the config.schema.yaml.

userlocalhost:
This is an implementation for #3053.

Previously, the prompt order which is displayed by the `st2 pack config ~` didn't same with the defined order in the config.schema.yaml.

To guarantee the displayed order will be the same with the defined order, we have to maintain the order of data during processing
- from loading data from `config.schema.yaml` to storing it to the Datastore. (af74d45)
- from loading data from Datastore to sending it as a response of st2api. (ac643fe)
- at loading response data from st2api. (d0bc450)

By this PR, st2client displays prompt for pack configuration in the defined order at the schema file.

Thank you.


requirements.txt
```diff
@@ -17,7 +17,7 @@ jsonschema<2.6,>=2.5.0
 kombu==3.0.37
 lockfile<0.11,>=0.10.2
 mock
-mongoengine==0.11.0
+mongoengine==0.12.0
```
Kami: Is this intentional?

Would need to check and verify that there are no backward incompatible changes, etc.

Kami:
Thanks a lot for this PR.

Ordering is one of the problems we were aware of for a while now. YAML and (default) CPython dictionaries are not ordered so to be able to preserve order we would need to write a custom YAML parser / loader.

That's also the reason why we haven't done it yet - writing a custom parser is not totally trivial and we would need to make quite some code changes across the board.

enykeev:
We've internally discussed it on multiple occasions. My take on the problem is that expecting fundamentally unordered type (dict) to retain a particular order after being processed by multiple levels of dependencies is going to be a constant exercise in frustration.

If we want parameters inside config schema to be ordered, we should avoid using dictionaries and switch to collections (list of dicts). YAML notation adds very little overhead to the schema itself and this approach would ensure that every tool written in every language will keep the initial order without any additional effort from the tool author.

armab:
I like the change at a high level, - it's very helpful and liquidates some users pain points, also +1 for OrderedDict.

BTW, "dicts" become somehow ordered in `Python 3.6`: https://mail.python.org/pipermail/python-dev/2016-September/146327.html.

userlocalhost:
Thank you for a lot of comments and explaining in detail.

I could found out that the StackStorm team considers to solve this problem with multiple approaches.

And I could understand that this PR is inadequate because of the lack of verification for the mongoengine v0.12.0.
That comment made me notice a very important thing to maintain a software safely (Thank you @kami).

I hope that this problem will be gone in some way.
Please let me know if I can do anything to help.

Thank you.

PR #3357, opened 2017-04-17
Fix RBAC syncer so it correctly removes assignments for users with no assignment file on disk

Kami:
This pull request fixes a bug in `st2-apply-rbac-definitions` tool.

Previously, if an existing assignment file for a user in `/opt/stackstorm/rbac/assignments/` directory was deleted and `st2-apply-rbac-definitions` tool was ran afterwards, this wasn't correctly reflected in the database - previous / old assignments would still stay in the database instead of being deleted.

This pull request fixes that.

Found this bug while working on #3351.

### DIFF
--- a/st2common/st2common/rbac/syncer.py
+++ b/st2common/st2common/rbac/syncer.py
@@ -188,9 +188,7 @@ class RBACDefinitionsDBSyncer(object):
 
         username_to_user_db_map = dict([(user_db.name, user_db) for user_db in user_dbs])
         username_to_role_assignment_api_map = dict([(role_assignment_api.username,
-                                                     role_assignment_api) for
-                                                     role_assignment_api in
-                                                     role_assignment_apis])
+            role_assignment_api) for role_assignment_api in role_assignment_apis])
 
         # Note: We process assignments for all the users (ones specified in the assignment files
         # and ones which are in the databse)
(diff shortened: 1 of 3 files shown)


st2common/st2common/rbac/syncer.py
```diff
@@ -185,11 +185,22 @@ def sync_users_role_assignments(self, role_assignment_apis):
...
+        username_to_role_assignment_api_map = dict([(role_assignment_api.username,
+            role_assignment_api) for role_assignment_api in role_assignment_apis])
+
+        # Note: We process assignments for all the users (ones specified in the assignment files
+        # and ones which are in the databse). We want to make sure assignments are correctly
+        # deleted from the databse for users which existing in the databse, but have no assignment
+        # file on disk.
+        all_usernames = (username_to_user_db_map.keys() +
+                         username_to_role_assignment_api_map.keys())
+        all_usernames = list(set(all_usernames))
```
lakshmi-kannan: Why do you need to convert them to a list?

m4dcoder: Could do the set in the operation above or in the for statement below? It's minor though.

Kami: It's not necessary to convert it to a list - just a personal preference.

And yes, I could expand list comprehension into a set and then use set.add...

PR #3358, opened 2017-04-17
Webhook URL Update for new build node

humblearner:


LindsayHill:
Looks like we should remove the old webhook URL now? It's returning 503 errors now.

PR #3360, opened 2017-04-17
Fix liveaction rerun with no parameters

m4dcoder:
If no parameters is provided, NoneType is passed to the server side resulting in error.


enykeev:
If it only started to fail after 2727, it's probably a thing that should be fixed on st2api end.

PR #1291, opened 2015-03-23
[DONT MERGE] Update FileWatchSensor to utilize trigger parameters

Kami:
Note: This is only here to help people understand the problem better.

PR #3361, opened 2017-04-18
File watch sensor to utilize trigger with parameters

lakshmi-kannan:
Cherry-picked https://github.com/StackStorm/st2/pull/1291

Then, I made the following changes:
* I enabled the file_watch_sensor by default. 
* I edited the sensor YAML file to reflect pack name explicitly
* I added a sample rule which would register a trigger with parameter (file_path). Sensor automatically picks up the file name from rule and starts to listen to file and emit lines from file.

To test, do the following:

* Spin st2 up and place the linux pack from this branch to /opt/stackstorm/packs
* Register contents of linux pack - st2ctl reload --register-all
* Restart st2 so file_watch_sensor is run - st2ctl restart
* Verify file_watch_sensor is running and enabled
```
(virtualenv)vagrant@st2dev /m/s/s/st2 ❯❯❯ ps auxww | grep file_watch | grep -v grep                                      tt_with_params ✭ ◼
vagrant  15142  0.1  2.0 111104 42520 pts/5    S+   07:19   0:04 /mnt/src/storm/st2/virtualenv/bin/python /mnt/src/storm/st2/st2reactor/st2reactor/container/sensor_wrapper.py --pack=linux --file-path=/opt/stackstorm/packs/linux/sensors/file_watch_sensor.py --class-name=FileWatchSensor --trigger-type-refs=linux.file_watch.file_path,linux.file_watch.line --parent-args=["--config-file", "/mnt/src/storm/st2/tools/../conf/st2.dev.conf"]
```

```
(virtualenv)vagrant@st2dev /m/s/s/st2 ❯❯❯ st2 sensor list --pack=linux                                                   tt_with_params ✭ ◼
+-----------------------+-------+-------------------------------------+---------+
| ref                   | pack  | description                         | enabled |
+-----------------------+-------+-------------------------------------+---------+
| linux.FileWatchSensor | linux | Sensor which monitors files for new | True    |
|                       |       | lines                               |         |
+-----------------------+-------+-------------------------------------+---------+
```
* Now copy sample rule that will register the rule 
```
(virtualenv)vagrant@st2dev /m/s/s/st2 ❯❯❯ st2 rule create /opt/stackstorm/packs/examples/rules/sample_rule_file_watch.yaml
+-------------+-----------------------------------------------------------+
| Property    | Value                                                     |
+-------------+-----------------------------------------------------------+
| id          | 58f620c6d9d7ed3af4554f4d                                  |
| name        | sample_rule_file_watch                                    |
| pack        | examples                                                  |
| description | Sample rule customer trigger type.                        |
| action      | {                                                         |
|             |     "ref": "core.local",                                  |
|             |     "parameters": {                                       |
|             |         "cmd": "echo "{{trigger}}""                       |
|             |     }                                                     |
|             | }                                                         |
| criteria    |                                                           |
| enabled     | False                                                     |
| ref         | examples.sample_rule_file_watch                           |
| tags        |                                                           |
| trigger     | {                                                         |
|             |     "type": "linux.file_watch.file_path",                 |
|             |     "ref": "linux.1c65e441-fd18-4217-bc53-8ac616a31074",  |
|             |     "parameters": {                                       |
|             |         "file_path": "/var/log/dmesg"                     |
|             |     }                                                     |
|             | }                                                         |
| type        | {                                                         |
|             |     "ref": "standard",                                    |
|             |     "parameters": {}                                      |
|             | }                                                         |
| uid         | rule:examples:sample_rule_file_watch                      |
+-------------+-----------------------------------------------------------+
```

### DIFF
--- a/contrib/examples/rules/sample_rule_file_watch.yaml
+++ b/contrib/examples/rules/sample_rule_file_watch.yaml
@@ -1,0 +1,17 @@
+---
+name: sample_rule_file_watch
+pack: "examples"
+description: Sample rule custom trigger type - add a file to be watched by file_watch_sensor in linux pack.
+enabled: false
+
+trigger:
+  parameters:
+    file_path: /var/log/dmesg
+  type: linux.file_watch.file_path
+
+criteria: {}
+
+action:
+  parameters:
+    cmd: echo "{{trigger}}"
+  ref: core.local
--- a/contrib/linux/sensors/file_watch_sensor.py
+++ b/contrib/linux/sensors/file_watch_sensor.py
@@ -9,17 +9,14 @@ class FileWatchSensor(Sensor):
     def __init__(self, sensor_service, config=None):
         super(FileWatchSensor, self).__init__(sensor_service=sensor_service,
                                               config=config)
-        self._config = self._config['file_watch_sensor']
-
-        self._file_paths = self._config.get('file_paths', [])
         self._trigger_ref = 'linux.file_watch.line'
+        self._logger = self._sensor_service.get_logger(__name__)
+
+        self._file_paths = []  # stores a list of file paths we are monitoring
         self._tail = None
 
     def setup(self):
-        if not self._file_paths:
-            raise ValueError('No file_paths configured to monitor')
-
-        self._tail = Tail(filenames=self._file_paths)
+        self._tail = Tail(filenames=[])
         self._tail.handler = self._handle_line
         self._tail.should_run = True
 
@@ -36,13 +33,25 @@ class FileWatchSensor(Sensor):
                 pass
 
     def add_trigger(self, trigger):
-        pass
+        if trigger['type'] not in ['linux.file_watch.file_path']:
+            return
+
+        file_path = trigger['parameters']['file_path']
+        self._tail.add_file(filename=file_path)
+
+        self._logger.info('Added file "%s"' % (file_path))
 
     def update_trigger(self, trigger):
         pass
 
     def remove_trigger(self, trigger):
-        pass
+        if trigger['type'] not in ['linux.file_watch.file_path']:
+            return
+
+        file_path = trigger['parameters']['file_path']
+        self._tail.remove_file(filename=file_path)
+
+        self._logger.info('Removed file "%s"' % (file_path))
 
     def _handle_line(self, file_path, line):
         trigger = self._trigger_ref
--- a/contrib/linux/sensors/file_watch_sensor.py
+++ b/contrib/linux/sensors/file_watch_sensor.py
@@ -47,7 +47,10 @@ class FileWatchSensor(Sensor):
         pass
 
     def remove_trigger(self, trigger):
-        if trigger['type'] not in ['linux.file_watch.file_path']:
+        file_path = trigger['parameters'].get('file_path', None)
+
+        if not file_path:
+            self._logger.error('Received trigger type without "file_path" field.')
             return
 
         file_path = trigger['parameters']['file_path']
--- a/contrib/linux/sensors/file_watch_sensor.py
+++ b/contrib/linux/sensors/file_watch_sensor.py
@@ -33,10 +33,12 @@ class FileWatchSensor(Sensor):
                 pass
 
     def add_trigger(self, trigger):
-        if trigger['type'] not in ['linux.file_watch.file_path']:
+        file_path = trigger['parameters'].get('file_path', None)
+
+        if not file_path:
+            self._logger.error('Received trigger type without "file_path" field.')
             return
 
-        file_path = trigger['parameters']['file_path']
         self._tail.add_file(filename=file_path)
 
         self._logger.info('Added file "%s"' % (file_path))
--- a/contrib/linux/sensors/file_watch_sensor.py
+++ b/contrib/linux/sensors/file_watch_sensor.py
@@ -53,7 +53,6 @@ class FileWatchSensor(Sensor):
             self._logger.error('Received trigger type without "file_path" field.')
             return
 
-        file_path = trigger['parameters']['file_path']
         self._tail.remove_file(filename=file_path)
 
         self._logger.info('Removed file "%s"' % (file_path))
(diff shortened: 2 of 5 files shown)


contrib/linux/sensors/file_watch_sensor.py
```diff
@@ -36,13 +33,25 @@ def cleanup(self):
                 pass
 
     def add_trigger(self, trigger):
-        pass
+        if trigger['type'] not in ['linux.file_watch.file_path']:
```
Kami: As discussed on Slack - I prototyped it this way for easiness.

If we decide to go with this approach I would propose to move this filtering to the base sensor class so those trigger handler methods only get called with triggers which belong to that sensor (so no additional filtering needed in sensor class trigger handler methods).

Another, even better future optimization (better than this suggested later filtering in the base class) would be to use a more advanced message routing so each sensor class would only receive triggers it should receive. Problem with that approach is that it's not fully backward compatible and would need to research how expensive it is to use queue per sensor (probably not too expensive).

dzimine:
This looks right. 
I don't see any draw-backs with this approach, if there are, pls list. Regardless, let's move forward.

To complete this PR, let's modify the line-watching sensor to take path as a trigger parameter, not a global parameter, too.

> If we decide to go with this approach I would propose to move this filtering to the base sensor class so those trigger handler methods only get called with triggers which belong to that sensor (so no additional filtering needed in sensor class trigger handler methods).
👍

lakshmi-kannan:
>To complete this PR, let's modify the line-watching sensor to take path as a trigger parameter, not a global parameter, too.

Dude, that's what this PR does. I am confused. 

>I don't see any draw-backs with this approach, if there are, pls list

I do see a drawback which is the sensor code is now responsible for some kind of scheduling handling the trigger types. In this particular case, this logic is hidden luckily https://github.com/StackStorm/st2/pull/3361/files#diff-414b51e67b8a124ee29083fc2bbce3a8R19. But it also is a poor example for people. We might want to add an example where we loop over the trigger types (may come from different rules) and do something.

contrib/linux/sensors/file_watch_sensor.py
lakshmi-kannan: @Kami I think the trigger check is not strictly required. We already filter in TriggerWatcher. 

```
(virtualenv)vagrant@st2dev /m/s/s/st2 ❯❯❯ ps auxww | grep file_watch | grep -v grep                                      tt_with_params ✭ ◼
vagrant  18758  1.1  2.1 115520 45008 pts/4    S+   14:25   0:02 /mnt/src/storm/st2/virtualenv/bin/python /mnt/src/storm/st2/st2reactor/st2reactor/container/sensor_wrapper.py --pack=linux --file-path=/opt/stackstorm/packs/linux/sensors/file_watch_sensor.py --class-name=FileWatchSensor --trigger-type-refs=linux.file_watch.file_path,linux.file_watch.line --parent-args=["--config-file", "/mnt/src/storm/st2/tools/../conf/st2.dev.conf"]
```

Note we pass --trigger-type-refs to the sensor wrapper. This is then passed to TriggerWatcher
https://github.com/StackStorm/st2/blob/master/st2reactor/st2reactor/container/sensor_wrapper.py#L354

https://github.com/StackStorm/st2/blob/master/st2reactor/st2reactor/container/sensor_wrapper.py#L192

https://github.com/StackStorm/st2/blob/master/st2common/st2common/services/triggerwatcher.py#L93

So I thought you originally did it for safety measure.

Kami: Good call - this was probably added after my original trigger with parameters change was implemented so it looks like my comment is obsolete :)

In any case, would be good to double check and verify that is indeed the case.

lakshmi-kannan: I verified and we are filtering right in TriggerWatcher and sensors don't see triggers that they don't register.

dzimine:
> Dude, that's what this PR does. I am confused.

I think you're adding file_watcher which takes parameters, but the line_watcher still relies on the global, no?

Kami:
Let's please also update the changelog.

Kami:
@dzimine 

> I think you're adding file_watcher which takes parameters, but the line_watcher still relies on the global, no?

Not exactly sure what you mean with line watcher? There is no such thing as line watcher (or am I missing something) - file watcher emits a trigger each time a change (new line) is detected on monitored file.

contrib/linux/sensors/file_watch_sensor.py
Kami: Cool, in this case we should get rid of checks in the sensor code :)

contrib/linux/sensors/file_watch_sensor.py
```diff
@@ -36,13 +33,29 @@ def cleanup(self):
...
+        if not file_path:
+            self._logger.error('Received trigger type without "file_path" field.')
+            return
+
+        self._tail.remove_file(filename=file_path)
+
+        self._logger.info('Removed file "%s"' % (file_path))
 
     def _handle_line(self, file_path, line):
         trigger = self._trigger_ref
```
dzimine: Will it be always emitting `'linux.file_watch.line`, see line 12?

contrib/examples/rules/sample_rule_file_watch.yaml
```diff
@@ -0,0 +1,17 @@
+---
+name: sample_rule_file_watch
+pack: "examples"
+description: Sample rule custom trigger type - add a file to be watched by file_watch_sensor in linux pack.
+enabled: false
+
+trigger:
+  parameters:
+    file_path: /var/log/dmesg
+  type: linux.file_watch.file_path
```
dzimine: The emitted trigger is `linux.file_watch.line` (I checked the logs, too) and here it is `linux.file_watch.file_path`. Funniest is that it works.

contrib/linux/sensors/file_watch_sensor.yaml
```diff
@@ -5,7 +5,20 @@
...
+          file_path:  # User ``st2`` should have permissions to be able to read this file.
+            description: "Path to the file to monitor"
+            type: "string"
+            required: true
+        additionalProperties: false
+    -
       name: "file_watch.line"
+      pack: "linux"
       description: "Trigger which indicates a new line has been detected"
       payload_schema:
```
dzimine: need to add `parameters_schema` here, too

PR #3364, opened 2017-04-19
RBAC for rest of the API endpoints

Kami:
This pull request implements RBAC for rest of the API endpoints which were missing it:

- [x] Policy type (get all, get one)
- [x] Policy (get all, get one, create, update, delete)
- [x] Action entry_point view
- [x] Action parameters view
- [x] Rule views
- [x] Introduce new ephemeral "Timer" model which is only used for permission grants (same as we do for webhooks)

### DIFF
--- a/st2common/st2common/rbac/types.py
+++ b/st2common/st2common/rbac/types.py
@@ -118,6 +118,10 @@ class PermissionType(Enum):
     WEBHOOK_DELETE = 'webhook_delete'
     WEBHOOK_ALL = 'webhook_all'
 
+    TIMER_LIST = 'timer_list'
+    TIMER_VIEW = 'timer_view'
+    TIMER_ALL = 'timer_all'
+
     API_KEY_LIST = 'api_key_list'
     API_KEY_VIEW = 'api_key_view'
     API_KEY_CREATE = 'api_key_create'
@@ -242,6 +246,7 @@ class ResourceType(Enum):
     EXECUTION = SystemResourceType.EXECUTION
     KEY_VALUE_PAIR = SystemResourceType.KEY_VALUE_PAIR
     WEBHOOK = SystemResourceType.WEBHOOK
+    TIMER = SystemResourceType.TIMER
     API_KEY = SystemResourceType.API_KEY
     TRACE = SystemResourceType.TRACE
     TRIGGER = SystemResourceType.TRIGGER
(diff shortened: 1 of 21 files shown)


Kami:
I introduced a new "TimerDB" meta model and resource type which is used to grant permissions on timers.

This way it's consistent with how we handle webhooks. Before that change, user needed to grant permissions on Triggers which is confusing and not really obvious.

Another small future improvement would be to allow users to specify optional "name" attribute (unique) for each timer rule trigger. This would make it a bit more user-friendly when working with timers and granting permissions on them. Right now user needs to use a system generation name which is not that user-friendly.

st2common/st2common/rbac/types.py
```diff
@@ -91,6 +94,7 @@ class PermissionType(Enum):
     EXECUTION_RE_RUN = 'execution_rerun'
     EXECUTION_STOP = 'execution_stop'
     EXECUTION_ALL = 'execution_all'
+    EXECUTION_VIEWS_FILTERS_LIST = 'execution_view_filters_list'
```
enykeev: `EXECUTION_FILTERS_LIST` seems sufficiently unique

Kami: Yeah, was thinking about it quite a bit, but I prefer to be explicit :P

"pack health" vs "pack index health" - first one is a bit confusing and not really clear, imo.

enykeev: is it a problem with the naming or the concept of index health by itself?

Kami: Naming - pack health would imply something about pack health while pack index health implies something about pack index health (which is more correct).

enykeev: I'm saying that `VIEW` part is not needed, not that we should remove `INDEX` part. The whole idea of views was added out of sheer necessity due to limitations of pecan we no longer have. That's minor enough, yet kinda annoying.

st2common/st2common/rbac/types.py
```diff
@@ -438,6 +491,8 @@ class SystemRole(Enum):
     PermissionType.EXECUTION_STOP: 'Ability to stop (cancel) a running execution.',
     PermissionType.EXECUTION_ALL: ('Ability to perform all the supported operations on a '
                                    'particular execution.'),
+    PermissionType.EXECUTION_VIEWS_FILTERS_LIST: ('Ability vew all the distinct execution '
```
enykeev: typo

st2common/st2common/rbac/types.py
```diff
@@ -468,6 +523,10 @@ class SystemRole(Enum):
     PermissionType.WEBHOOK_ALL: ('Ability to perform all the supported operations on a particular '
                                  'webhook.'),
 
+    PermissionType.TIMER_LIST: 'Ability to list (view all) timers.',
```
enykeev: inconsistency

st2common/st2common/rbac/types.py
Kami: Ah, I see.

Will think about it a bit, from a quick glance I agree with you, but I originally went with "view" for explicitness sake.

st2common/st2common/rbac/types.py
Kami: Good catch, will fix.

PR #3375, opened 2017-04-24
Cancel subworkflow actions

m4dcoder:
When a Mistral workflow is cancelled, any actions that is a Mistral workflow is also cancelled.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -41,6 +41,7 @@ in development
   ``TRIGGER_*`` RBAC permission constants (improvement)
 * Implement RBAC for webhooks get all and get one API endpoint. (improvement)
 * Add webhook payload to the Jinja render context when rendering Jinja variable inside rule criteria section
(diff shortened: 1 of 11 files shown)


lakshmi-kannan:
@m4dcoder I still see this not being fixed. https://github.com/StackStorm/st2/blob/master/contrib/runners/mistral_v2/query/mistral_v2.py#L162. The case when WF state is "CANCELLED" and there are no active_tasks, we should return status CANCELED as opposed to CANCELING. Same for the case when the WF state is "CANCELLED" but there are active tasks which are also "CANCELLED". The current status quo would mean any workflow that is canceled, the querier will never delete the state object.

m4dcoder:
> The case when WF state is "CANCELLED" and there are no active_tasks, we should return status CANCELED as opposed to CANCELING

This is set to canceled. Where are you seeing that it is not?

> Same for the case when the WF state is "CANCELLED" but there are active tasks which are also "CANCELLED". 

Why? There are active tasks... Technically, it's not cancelled. So the question is why are the tasks still active?

PR #3381, opened 2017-04-29
Fix get_roles_for_user and get_role_assignments_for_user

Kami:
Recently a branch introduced a new `is_remote` field on the `UserRoleAssignmentDB` model.

To accommodate for this field and change, existing service functions were updated to handle local (non-remote) and remote role assignments.

This new field has a default value of false, but mongoengine doesn't lazily populate this field for existing documents which don't contain this field with a default value. This means that the queryset filter and service function didn't work correctly - it didn't correctly return old / existing documents without this field.

This pull request fixes that.

This would only affect users upgrading from older release to v2.3.0 (not released yet).

On a related note - I'm sure we made similar changes in the past and we were probably just lucky we haven't encountered similar issue / bug yet.

Note: This issue was discovered by one of our users.

### DIFF
--- a/st2common/st2common/models/db/auth.py
+++ b/st2common/st2common/models/db/auth.py
@@ -45,13 +45,16 @@ class UserDB(stormbase.StormFoundationDB):
     nicknames = me.DictField(required=False,
                              help_text='"Nickname + origin" pairs for ChatOps auth')
 
-    def get_roles(self):
+    def get_roles(self, include_remote=True):
         """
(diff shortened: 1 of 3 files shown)


lakshmi-kannan:
>On a related note - I'm sure we made similar changes in the past and we were probably just lucky we haven't encountered similar issue / bug yet.

We usually write migration scripts to handle such a case. There is a possibility we did miss some of these.

PR #3382, opened 2017-05-01
WIP: Add a 'enabled' field for triggers

lakshmi-kannan:


st2common/st2common/models/api/rule.py
```diff
@@ -239,7 +239,8 @@ def to_model(cls, rule):
 
         # Create a trigger for the provided rule
         trigger_db = TriggerService.create_trigger_db_from_rule(rule)
-        kwargs['trigger'] = reference.get_str_resource_ref_from_model(trigger_db)
+        if trigger_db:
```
Kami: Just curious why this change is needed.

st2common/st2common/models/db/trigger.py
```diff
@@ -77,6 +77,8 @@ class TriggerDB(stormbase.StormBaseDB, stormbase.ContentPackResourceMixin,
 
     name = me.StringField(required=True)
     pack = me.StringField(required=True, unique_with='name')
+    enabled = me.BooleanField(required=True, default=True,
```
Kami: I know it's WIP, but before we merge it we should also document what "disabled trigger" implies (what kind of behavior, etc.).

Kami:
^ Bump let's get that finished.

lakshmi-kannan:
This PR is not needed anymore.

Issue #3385, opened 2017-05-02
Config option system.validate_trigger_payload not being honored


Mierdin:
https://github.com/StackStorm/st2/pull/3142 introduced a new option to disable validation for trigger payloads, among other things. The advertised default behavior is that validation is disabled by default, but can be explicitly enabled with [this option](https://github.com/StackStorm/st2/blob/master/conf/st2.conf.sample#L265).

However, several users have reported that [validation is still being done](https://github.com/StackStorm-Exchange/stackstorm-kubernetes/issues/11) regardless of the presence or even the value of this configuration option.

My theory is that, because [the configuration definition for these options](https://github.com/StackStorm/st2/pull/3142/files#diff-f7f00f2d9fa85d6b39dc2def0f872561R77) set up these two options as `StrOpt`, [this conditional](https://github.com/StackStorm/st2/pull/3142/files#diff-fcc1652906182b402530d76b68e1733aR91) always evaluates to False, since `cfg.CONF.system.validate_trigger_parameters` is always populated with a value, and will always evaluate to True in this expression. Therefore, the following validation logic isn't skipped and always takes place.

So, a few things need done:

- [x] confirm my theory, and that this is indeed a bug
- [x] fix le bug
- [x] find out why this wasn't caught in testing and amend things as needed

Kami:
Good catch - this should totally be `BoolOpt` :)

And yeah, would need to dig in why the existing tests didn't catch this (we probably have no tests which directly use default value from the config).

PR #3390, opened 2017-05-03
Fix trigger parameter and payload validation bug

Mierdin:
This PR fixes the bug found in https://github.com/StackStorm/st2/issues/3385 by converting the trigger payload and parameter validation configuration parameters from `StrOpt` to `BoolOpt`, as well as adding unit tests for both to ensure the default configuration allows validation for both to be bypassed (which is the behavior we currently advertise).

## TODOs

- [x] Fix bug
- [x] Add test for payload validation with default config
- [x] Add test for parameter validation with default config

Closes https://github.com/StackStorm/st2/issues/3385

### DIFF
--- a/st2reactor/tests/unit/test_sensor_service.py
+++ b/st2reactor/tests/unit/test_sensor_service.py
@@ -36,8 +36,15 @@ class SensorServiceTestCase(unittest2.TestCase):
         self.sensor_service._dispatcher.dispatch = mock.MagicMock(side_effect=side_effect)
         self._dispatched_count = 0
 
-        # reset default configuration value
-        cfg.CONF.system.validate_trigger_payload = False
+        # Previously, cfg.CONF.system.validate_trigger_payload was set to False explicitly
+        # here. Instead, we store original value so that the default is used, and if unit
+        # test modifies this, we can set it to what it was (preserve test atomicity)
+        self.validate_trigger_payload = cfg.CONF.system.validate_trigger_payload
+
+    def tearDown(self):
+
+        # Replace original configured value for payload validation
+        cfg.CONF.system.validate_trigger_payload = self.validate_trigger_payload
 
     @mock.patch('st2common.services.triggers.get_trigger_type_db',
                 mock.MagicMock(return_value=TriggerTypeMock(TEST_SCHEMA)))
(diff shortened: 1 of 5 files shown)


Kami:
Great, thanks - LGTM :)

And yeah, should figure out how to add unit test for default values which would have caught this.

Mierdin:
@Kami Found some time to add tests. I'll pay attention to CircleCI but if you don't hear from me, assume this is RFR

st2reactor/tests/unit/test_sensor_service.py
```diff
@@ -36,8 +36,14 @@ def side_effect(trigger, payload, trace_context):
         self.sensor_service._dispatcher.dispatch = mock.MagicMock(side_effect=side_effect)
         self._dispatched_count = 0
 
-        # reset default configuration value
-        cfg.CONF.system.validate_trigger_payload = False
+        # Previously, cfg.CONF.system.validate_trigger_payload was set to False explicitly
+        # here. Instead, we store original value so that the default is used, and if unit
+        # test modifies this, we can set it to what it was (preserve test atomicity)
+        self.validate_trigger_payload = cfg.CONF.system.validate_trigger_payload
```
Kami: This should work fine as long as each test method correctly sets the value to the one which is being tested, right?

Kami:
Hm, looks like tests are failing now - any clue *didn't really dig in, but the trigger param validation test was failing)?

st2reactor/tests/unit/test_sensor_service.py
Mierdin: If they want, yeah. The idea here is to preserve the original/default value for `cfg.CONF.system.validate_trigger_payload` so that each unit test can do whatever they want with it. So, a unit test can change the value of `cfg.CONF.system.validate_trigger_payload` before testing (as many of them already do) or they can not, and use the default value (as my added test does).

Mierdin:
Yeah, I finally had a chance to look into it, I was re-using a rule from a similar unit test but that was throwing a duplicate key error in mongo driver. Makes me 🤔  a bit since it means the tests aren't atomic, but I just added a new test rule anyways.

Mierdin:
Everything is passing except for Xenial, and it looks like it's exactly like this reported issue: https://github.com/StackStorm/st2-packages/issues/445

Kami:
Yeah, some "old" test cases use `FunctionalTest` and `DbTestCase` base class.

With those base classes, DB is only destroyed once per whole test class run. This is obviously not ideal from "each test method should start in a clean state" perspective.

Quite a long time ago, I added `CleanDbTestCase` class where DB is destroyed before and after each test method run - this way tests can't accidentally and intentionally rely on some shared state across test method runs. Most new test cases (at least the ones I work on) use this base class :D

Kami:
:+1:

estee-tew:
https://stackstorm.slack.com/archives/community/p1494453032636641

estee-tew:
https://stackstorm.slack.com/archives/community/p1495558209860466

PR #3394, opened 2017-05-05
Add sys.exit() to rulesengine

andrew-regan:
The entry points for all of the st2 processes `sys.exit()` except for `st2rulesengine`. If `st2rulesengine` currently fails and exits it will exit successfully and won't be restarted by systemd. This change makes  `st2rulesengine` consistent with the entry points for all of the other st2 processes.


Kami:
Good catch - thanks!

Can you please also update changelog? And test would also be great (but yeah, not totally straight forward since it's an integration one)

Kami:
Merged, thanks!

PR #3372, opened 2017-04-21
Upgrade dependencies

Kami:
Various dependencies upgrades again to get the latest changes and bug fixes:

* pyyaml,
* requests
* appscheduler
* gitpython
* paramiko
* mongoengine
* tooz

PR #3400, opened 2017-05-10
Pin requests to the latest version (v2.14.1)

Kami:
As per https://github.com/StackStorm/st2/pull/3372#discussion_r115702199 (thanks)

Should fix the https://github.com/shazow/urllib3/issues/1104 and broken Githubwebhooks on `st2cicd`.

### DIFF
--- a/fixed-requirements.txt
+++ b/fixed-requirements.txt
@@ -10,7 +10,7 @@ oslo.config>=1.12.1,<1.13
 oslo.utils<3.1.0
 six==1.10.0
 pyyaml>=3.12,<4.0
-requests[security]>=2.13.0,<2.14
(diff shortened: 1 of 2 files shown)


armab:
Additionally, I suspect that previous PR with updated dependencies https://github.com/StackStorm/st2/pull/3372 brings some more regressions related to timeouts.

armab:
I just rebuilt the cicd server with recent packages, new one is `st2cicd003` and the fix worked. Githubwebhook `/webhooks/github/events` is back alive (`POST` URL works).

Thanks @Kami cc @LindsayHill

Kami:
Sweet, thanks for confirming :+1:

PR #3402, opened 2017-05-10
Revert mongoengine dependency upgrade

armab:
Reverts StackStorm/st2#3372

For now just a PR to experiment if `pip` packages might be related to re-appeared Ubuntu16 regression (even with timeouts) https://github.com/StackStorm/st2-packages/issues/445.

### DIFF
--- a/st2actions/tests/unit/policies/test_concurrency.py
+++ b/st2actions/tests/unit/policies/test_concurrency.py
@@ -248,7 +248,7 @@ class ConcurrencyPolicyTest(EventletTestCase, DbTestCase):
         # Since states are being processed asynchronously, wait for the
         # liveactions to go into scheduled states.
         for i in range(0, 100):
-            eventlet.sleep(2)
+            eventlet.sleep(1)
             scheduled = [item for item in LiveAction.get_all() if item.status in SCHEDULED_STATES]
             if len(scheduled) == policy_db.parameters['threshold']:
                 break
@@ -272,7 +272,7 @@ class ConcurrencyPolicyTest(EventletTestCase, DbTestCase):
         # Since states are being processed asynchronously, wait for the
         # liveaction to go into delayed state.
         for i in range(0, 100):
-            eventlet.sleep(2)
+            eventlet.sleep(1)
             liveaction = LiveAction.get_by_id(str(liveaction.id))
             if liveaction.status == action_constants.LIVEACTION_STATUS_DELAYED:
                 break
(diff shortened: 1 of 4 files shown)


Mierdin:
Good idea. Will keep an eye on this. Curious how you knew to look for this - how long has this been an issue? I was going to start looking through history last night but that was at about 1AM so I bailed

Mierdin:
Okay, u16 was passing with your reverts, so I set mongoengine back to 0.13 in https://github.com/StackStorm/st2/pull/3402/commits/31c60658bfc96653e9b74d6fb1cea95718ee35e9

Mierdin:
Just changing mongoengine back to 0.13 re-broke things. So it seems pretty clear that this upgrade is the culprit

Mierdin:
Okay I reverted the reverts except for the one for mongoengine. Will watch this make sure this passes, then we should consider merging this for now so that tests for other PRs can go through.

Someone knowledgeable about the database layer (not me) should review http://docs.mongoengine.org/changelog.html#changes-in-0-12-0 and see if anything impacts us; it seems obvious that it's something, but I am not sure.

Kami:
:+1: 

I did check the changelog before bumping, but nothing stood out and everything also passed locally and on Circle CI.

Having said that, it does look like some change is negatively impacting us so I will need to dig deeper in the mongoengine code.

armab:
As discussed with @Mierdin we :fire_engine: the :fire: first and merge the fix to unblock the CI. According to the commit log, it's clearly `mongoengine` guilty.

And @Kami feel free to dig in deeper into the `mongoengine` changelog or ideally if you can, - please help investigate the https://github.com/StackStorm/st2-packages/issues/445 which still remains the bug if we remove `sleep 5` from the existing code (see https://github.com/StackStorm/st2-packages/blob/ffe4f7bb9addd777f0a5dd33602ff4d9e759378c/scripts/st2bootstrap-deb.sh#L331-L336).
You're more knowledgeable about st2 core mechanics & mongo.

st2actions/tests/unit/policies/test_concurrency.py
```diff
@@ -248,7 +248,7 @@ def test_on_cancellation(self):
         # Since states are being processed asynchronously, wait for the
         # liveactions to go into scheduled states.
         for i in range(0, 100):
-            eventlet.sleep(2)
```
Kami: This was actually one because concurrency tests were failing, but if they are not failing anymore, it's fine.

That failure was probably related to Circle CI since it was failing on master as well, before any other stuff was merged.

PR #3404, opened 2017-05-10
make sphinx vers match st2docs

LindsayHill:
Updated the sphinx and sphinx-autobuild versions in `test-requirements.txt` to match what we currently use in st2docs. 

We don't need to track our own version of sphinx-autobuild anymore, and we should pin the sphinx version.


Kami:
:+1:

PR #3406, opened 2017-05-11
Fix ?pack query param filter in /v1/triggertypes API endpoint

### LINKED ISSUE #3405: st2 trigger list doesn’t respect —pack
It says it all.

`st2 trigger list --pack=mypack` doesn't filter packs. 

Version 2.3dev

Kami:
It looks like a small regression introduced during move to OpenAPI and we didn't have any tests for this functionality.

Resolves #3405


Kami:
@enykeev We should probably throw / return 400 status code in case user passes in unsupported query param filter, wdyt?

enykeev:
> We should probably throw / return 400 status code in case user passes in unsupported query param filter, wdyt?

I don't think we should. The request itself (list all triggers that belongs to certain pack) has finished successfully (there's exactly zero of this triggers currently present). If we are to return 404 on zero result response, we then need to do the same when we don't have any entities of a particular type registered at all (no executions or no actions). And that's clearly not what the status code is for.

PR #3408, opened 2017-05-12
Properly cast numbers in keyvalues

### LINKED ISSUE #3407: st2kv does not resolve in Jinja cast expression
In the rule: 
` {{ "2" | int }} - 2` Dah
`{{ trigger.count | int }} -> 2` OK
`{{ st2kv.system.count | int }} -> 0` Wrong.

enykeev:
Fixes #3407


Kami:
Good catch - can you please also update the changelog.

PR #3410, opened 2017-05-15
Include RBAC related info in action context when running action through the API

Kami:
This pull request updates action executions handler to include user RBAC related information (username and roles) when an execution is scheduled through the API.

It was already previously possible to retrieve user roles inside the action / workflow, but it was more convoluted and involved more steps since user needed to write a custom action which hits StackStorm API to retrieve current user roles.

This also meant one additional task / action inside the workflow chain now it's possible to reference it using ``action_context`` context (e.g. ``action_context.rbac.roles``) which is available to actions and action parameters.

Example action execution context when RBAC is enabled:

```json
{
	"user": "multiple_roles",
	"rbac": {
		"user": "multiple_roles",
		"roles": ["admin", "role_1", "role_2", "role_3"]
	}
}
```

Example of using it inside action metadata definition using the Jinja notation:

```yaml
...
parameters:
...
  action_user_rbac_roles:
    default: "{{action_context.rbac.rolesd}}"
    type: array
    immutable: true
...

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -62,6 +62,8 @@ in development
 * Update ``/v1/rbac/roles`` API endpoint so it includes corresponding permission grant objects.
   Previously it only included permission grant ids. (improvement)
 * Fix a bug where keyvalue objects weren't properly cast to numeric types. (bug fix)
+* When RBAC is enabled and action is scheduled (ran) through the API, include ``rbac`` dictionary
+  with ``user`` and ``roles`` ``action_context`` attribute. (improvement)
 
 2.2.1 - April 3, 2017
 ---------------------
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: st2api/tests/unit/controllers/v1/test_executions_simple.py [hunk 1]
```diff
@@ -202,7 +235,7 @@ def tearDownClass(cls):
         cls.app.delete('/v1/actions/%s' % cls.action2['id'])
```
Kami: Note: Those existing tests are old and messy - they don't use fixtures loader so we eventually need to port them.

CHANGELOG.rst
```diff
@@ -62,6 +62,8 @@ in development
 * Update ``/v1/rbac/roles`` API endpoint so it includes corresponding permission grant objects.
   Previously it only included permission grant ids. (improvement)
 * Fix a bug where keyvalue objects weren't properly cast to numeric types. (bug fix)
+* When RBAC is enabled and action is scheduled (ran) through the API, include ``rbac`` dictionary
```
lakshmi-kannan: Newline goober

PR #3411, opened 2017-05-16
Update st2-apply-rbac-definitions to remove assignments from database for users which don't exist in the database

Kami:
This pull request updates `st2-apply-rbac-definitions` tools so it also removes role assignments from the database for users which don't exist in the database.

Before that change we already supported pre-creation of role assignments, meaning user could write role assignments for users which don't yet exist in the database and those assignments would get stored in the database (StackStorm user accounts are lazy created on first successful authentication with st2auth).

We didn't do that for removal (we only removed assignments for users which exist in the database) so the behavior was inconsistent (there are no negative security implications because assignments existing in the database for non-existent user is basically the same as if assignments wouldn't exist).


Kami:
Still need a review.

/cc @lakshmi-kannan

PR #3412, opened 2017-05-16
Fix mistral abandoned tasks

m4dcoder:
When action runner is shutdown, any liveactions that are still running will be abandoned. On abandon, the runner's post run method is not called so operations such as callback is not performed. In the case of Mistral, since callback is not performed, Mistral does not know that the task has completed and so the workflow is stuck in running state.

### DIFF
--- a/st2common/st2common/runners/utils.py
+++ b/st2common/st2common/runners/utils.py
@@ -15,7 +15,11 @@
 
 import logging as stdlib_logging
 
+from st2actions.container.service import RunnerContainerService
 from st2common import log as logging
+from st2common.runners import base as runners
+from st2common.util import action_db as action_db_utils
+
 
 __all__ = [
     'get_logger_for_python_runner_action',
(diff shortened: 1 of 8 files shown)


Kami:
I have the context from Slack, but in the future please try to add some description, context and background so other people without the context about the change also know what is going on :)

Kami:
LGTM, good catch :+1:

st2common/st2common/runners/utils.py
```diff
@@ -78,3 +82,38 @@ def get_action_class_instance(action_cls, config=None, action_service=None):
         action_instance.action_service = action_service
 
     return action_instance
+
+
+def invoke_post_run(liveaction_db, action_db=None):
```
Kami: Let's please add some standalone "unit" tests for this functions for all the scenarios and edge cases:

* action_db is None and action exists in db
* action_db is None and action doesn't exist in db

m4dcoder: These tests are added.

Kami:
:+1:

PR #3413, opened 2017-05-16
Some changes to Mistral querier to get better CPU performance

lakshmi-kannan:
## Issue:

This issue came up in st2cicd003 box where a bunch of workflows are stuck in RUNNING state (mistral <-> st2 integration bug, suspected to be addressed by https://github.com/StackStorm/st2/pull/3412). The results tracker keep querying the status of these workflows forever and doesn't bail. The CPU is constantly at 100% on the box. 

## Details

We seem to be querying mistral state over and over again for 9 workflows stuck in RUNNING. On an average, it takes some queries 3s to complete. We make a query every 1s i.e. if I query and I realized that the query didn’t clear the state, it goes back to the queue and we check if current_time - last_query_time > 1 (query_interval). If yes, we re-query. So we were constantly chugging CPU. Unfortunately, the query_interval is hard coded to 1. When I set that to something like 20s, I see the right pattern. When there are threads querying, we hit 100% CPU and then the CPU goes back to normal.

## What's the change? 

As you can see, I cleaned up the while loop. Previously, we were adding back a query to be rescheduled back in the same queue as the job queue. This meant that in this particular case, the queue was never empty and therefore the CPU kept spinning without a cooperative yield. This is a nasty oversight from my original implementation which took 3 years for us to hit it :D. I now cleaned up the code to capture the queries to be rescheduled in a different list and add it back to job queue and also cooperatively yield CPU. 

Along with this change, I also bumped the query_interval to 20s as default. This means we are introducing a delay in detecting end of workflows but with light to reasonable loads, we shouldn't see the CPU issue. This number is entirely load dependent which is why I made the interval and thread pool size configurable so people who know what they are doing can tweak it. 

The CPU profile is much better now. I'd love to share graphs but it's a WIP. 

Please register your thoughts and this IMO should be merged while we think about the callback based redesign which is WIP. /cc @dzimine @Kami @m4dcoder @bigmstone @enykeev

P.S.: The reason for the mistral queries to take 3s is because we not only query execution result but also query individual tasks for a workflow in a serial manner. Here's timing info https://stackstorm.slack.com/files/lakstorm/F5EE3HD61/timing.txt and here's the link to code where we do this serially - https://github.com/StackStorm/st2/blob/master/contrib/runners/mistral_v2/query/mistral_v2.py#L124.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -77,6 +77,13 @@ in development
   support non-ascii (unicode) characters. (bug fix)
 * When RBAC is enabled and action is scheduled (ran) through the API, include ``rbac`` dictionary
   with ``user`` and ``roles`` ``action_context`` attribute. (improvement)
+* Fix a bug in query base module when outstanding queries to mistral or other workflow engines 
+  could cause a tight loop without cooperative yield leading to 100% CPU usage by st2resultstracker
+  process. (bug-fix)
+* Make the query interval to third party workflow systems (including mistral) a configurable
+  value. You can now set ``query_interval`` in ``[results_tracker]`` section in ``/etc/st2/st2.conf``.
+  With this, the default query interval is set to 20s as opposed to 0.1s which was rather aggressive
+  and could cause CPU churn when there is a large number of outstanding workflows. (improvement)
 
 2.2.1 - April 3, 2017
 ---------------------
--- a/conf/st2.dev.conf
+++ b/conf/st2.dev.conf
@@ -100,3 +100,9 @@ logging = st2exporter/conf/logging.exporter.conf
 
 [content]
 pack_group = stanley
+
+[results_tracker]
+query_interval = 0.1
+
+[mistral]
+jitter_interval = 0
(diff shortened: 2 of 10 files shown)


dzimine:
ack

Kami:
Thanks and good catch both (@enykeev and @lakshmi-kannan) :+1: 

I'm personally with @enykeev and I think it would be good to randomly spread out requests a bit (to avoid thundering herd affect and overloading Mistral API):

> @lakstorm can we actually spread the requests around this 20s window instead of bursting them all at once?

m4dcoder:
LGTM I assume mistral-itests passed

Kami:
Please update changelog.

Besides that, LGTM :+1:

armab:
Because it's a forked PR, some integration checks like st2-packages + e2e were not triggered :/

Just a reminder to use a branch instead when possible.

lakshmi-kannan:
Updated CHANGELOG. Also, docs are here https://github.com/StackStorm/st2docs/pull/503

PR #3414, opened 2017-05-17
Add new core.uuid action which generates a random uuid4

lampwins:
Based on recommendation from @lakshmi-kannan  on slack, here is a new core action for generating new UUIDs.


Kami:
Thanks - looks good to me :)

Small improvements and  things to do:

* Bump version in contrib/core/pack.yaml
* Add a small unit test for action in in contrib/tests/

armab:
The action is good :+1: 

But IMO this would be much better to implement via custom Jinja 2 filter:

```yaml
{{ something | to_uuid }}
```

See: https://docs.stackstorm.com/reference/jinja.html#custom-jinja-filters and https://github.com/StackStorm/st2/tree/master/st2common/st2common/jinja/filters

Kami:
Filter is usually for transforming an existing value and not for generating a new one.

We could implement a custom Jinja function in addition to the action so you can use either, depending on the context aka which one is easier to use / fits better.

lampwins:
This action was to cover the case when a new random UUID is needed. Not one created from user data. I can see the value in a jinja filter but that is separate from what this action is intended for.

@Kami I will create the unit test(s) but in relation to the version bump, I did bump from 0.2.1 to 0.3.1, was this not what you had intended?

Kami:
@lampwins I agree, filter can be a separate PR.

0.3.1 is fine

contrib/core/actions/uuid.yaml
```diff
@@ -0,0 +1,6 @@
+---
+name: uuid
```
Kami: Just thinking out loud - would we ever want to add support for other types of UUID? If so it, might make sense to call this action `uuid4`, but in any case, not a big deal - we can always change it in the future in a backward compatible manner if necessary.

emedvedev: FWIW, I'm also in favor of `uuid4`

lakshmi-kannan:
>This action was to cover the case when a new random UUID is needed. Not one created from user data. I can see the value in a jinja filter but that is separate from what this action is intended for. 

+1

Also, +1 to calling it uuid4 if it's not a hassle. Thanks for the PR! As you can see, we have a critical mass of reviewers when diff count is enticing ;).

contrib/core/actions/uuid.yaml
lampwins: So I had this same conversation in my head at the time. My descision was to leave it uuid to allow for a a `type` parameter to be added later on. This would default to uuid4 but would allow the user choose, say uuid1 all in the same action. Thoughts on that? Also ping @lakshmi-kannan

lakshmi-kannan: Yeah, "type" may not be the most ideal name for that param. It will for sure clash with the "type" field that st2 exposes to declare "data type" for parameters. I'd call it uuid_type and it's a good idea. +1 to that.

lampwins: Ah that's a good point, `uuid_type` it is then. I will put this to bed by implementing this with support for uuid1 and uuid4. `uuid_type` will be optional so it will default to uuid4. I will try to get this done today with associated test(s).

lampwins:
Not really sure what I broke. The error in CI is rather vague... `make: *** [.packs-tests] Error 123`

contrib/core/actions/generate_uuid.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class GenerateUUID(Action):
+    def run(self, uuid_type):
+        if uuid_type == 'uuid1':
+            return str(uuid.uuid1())
+        elif uuid_type == 'uuid4':
+            return str(uuid.uuid4())
+        else:
+            return False, "Unknown uuid_type. Only uuid1 and uuid4 are supported"
```
Kami: Would be better and more Pythonic to throw an exception here - `raise ValueError('Unknown uuid_type. Only uuid1 and uuid4 are supported')`

Returning a tuple with a status code is mostly meant to be used when you want action to fail, but also want to return some structured data (e.g. dict, array, etc.).

If you just want to return error message, exceptions are usually a better fit.

contrib/core/actions/uuid.yaml
```diff
@@ -0,0 +1,11 @@
...
+runner_type: "python-script"
+description: Generate a new UUID (default uuid4)
+enabled: true
+entry_point: generate_uuid.py
+parameters:
+  uuid_type:
+    type: string
+    default: uuid4
```
Kami: Small thing, should also add `enum` property:

```yaml

parameters:
  uuid_type:
    type: string
    default: uuid4
    enum:
      - uuid1
      - uuid4
```

This way other, non-valid values won't be accepted and a nice drop-down with possible values will be shown in the UI.

contrib/core/tests/test_action_uuid.py
```diff
@@ -0,0 +1,39 @@
...
+        result = action.run(uuid_type='uuid1')
+        self.assertTrue(result)
+
+        # accepts uuid4 as a type
+        result = action.run(uuid_type='uuid4')
+        self.assertTrue(result)
+
+        # fails on incorrect type
+        result = action.run(uuid_type='foobar')
+        self.assertFalse(result)
```
Kami: Looks like this test is failing.

Can you please update it to thrown an exception and then update this test case?

Once that is done, I will go ahead and merge it into master :)

Kami: And while you are at it, also feel free to update `CHANGELOG.rst` file.

Kami:
Merged, thanks!

PR #3416, opened 2017-05-17
Prevent UnicodeDecodeErrors on stdout/stderr when packs.setup_virtualenv fails

Kami:
This pull request works on top of #3338 and adds / changes:

* Moves utility function into `st2common.util.compat` module
* Adds test case for it
* Adds changelog entry


Kami:
Still need a review.

/cc @lakshmi-kannan

PR #3420, opened 2017-05-18
Optimize workflow task queries in st2resultstracker

m4dcoder:


Kami:
https://github.com/StackStorm/st2/pull/3412#issuecomment-302027574 - same here, please :)

contrib/runners/mistral_v2/query/mistral_v2.py
```diff
@@ -115,19 +143,28 @@ def _get_workflow_result(self, exec_id):
...
-        wf_tasks = []
+        result = []
 
         try:
-            for task in self._client.tasks.list(workflow_execution_id=exec_id):
-                task_result = self._client.tasks.get(task.id)
-                wf_tasks.append(task_result)
+            query_filters = {}
+
+            if last_query_time:
```
Kami: Cool, I assume this should be a lot more efficient :)

contrib/runners/mistral_v2/query/mistral_v2.py
```diff
@@ -55,24 +61,43 @@ def __init__(self, id, *args, **kwargs):
...
         :param query_context: context for the query to be made to mistral. This contains mistral
                               execution id.
-        :type query_context: ``objext``
+        :type query_context: ``object``
+        :param last_query_time: Timestamp of last query.
+        :type last_query_time: ``float``
         :rtype: (``str``, ``object``)
         """
+        dt_last_query_time = (
+            time.strftime('%Y-%m-%d %H:%M:%S', time.gmtime(last_query_time))
```
Kami: It looks like actual date and time string and not timestamp is used, let's hope it correctly handles DST :)

Kami: Actually, does Mistral API accept timezone at the end of the date and time string? Aka can we just pass UTC to avoid DST and other timezone related issues?

lakshmi-kannan: +1 to using UTC if supported. This is a bug waiting to happen

lakshmi-kannan:
Except the UTC comment, this PR looks good to go with a CHANGELOG update.

contrib/runners/mistral_v2/query/mistral_v2.py
m4dcoder: Mistral saves timestamps as UTC. time.gmtime returns a UTC timestamp. DST flag is not set.  https://docs.python.org/2/library/time.html#time.gmtime AFAIK, the two system are using the same time format. Where's the issue?

m4dcoder: https://docs.python.org/2/library/time.html#time.gmtime says time.gmtime returns UTC and daylight savings time not represented. i assume it takes the epoch seconds and just return the UTC, doesn't care about DST. epoch seconds is already in UTC per https://en.wikipedia.org/wiki/Unix_time#Encoding_time_as_a_number. and since mistral already saves timestamp as UTC. so afaik, DST doesn't apply here...

Kami: It sounds like we should be fine then so thanks for clarifying and LGTM.

PR #3451, opened 2017-06-07
Fix duplicate remove from running liveactions set

m4dcoder:
On action runner shutdown, item in the _running_liveactions set is removed twice, once during the abandon process and a second time on completion of _run_action.

### DIFF
--- a/st2actions/tests/unit/test_worker.py
+++ b/st2actions/tests/unit/test_worker.py
@@ -98,7 +98,7 @@ class WorkerTestCase(DbTestCase):
         # over a separate thread, run the shutdown sequence on the main thread, and then let
         # the local runner to exit gracefully and allow _run_action to finish execution.
         with tempfile.NamedTemporaryFile() as fp:
-            params = {'cmd': 'while [ -e \'%s\' ]; do sleep 1; done' % fp.name}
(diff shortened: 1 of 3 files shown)


st2actions/tests/unit/test_worker.py
```diff
@@ -71,19 +84,41 @@ def test_non_utf8_action_result_string(self):
...
+        with tempfile.NamedTemporaryFile() as fp:
+            params = {'cmd': 'while [ -e \'%s\' ]; do sleep 1; done' % fp.name}
+            liveaction_db = self._get_liveaction_model(WorkerTestCase.local_action_db, params)
+            liveaction_db = LiveAction.add_or_update(liveaction_db)
+            executions.create_execution_object(liveaction_db)
+            runner_thread = eventlet.spawn(action_worker._run_action, liveaction_db)
+
+            # Wait for the worker to add the liveaction to _running_liveactions.
+            while len(action_worker._running_liveactions) <= 0:
+                eventlet.sleep(1)
```
lakshmi-kannan: Do something like 0.1. Otherwise, this test might take foreva. Unfortunately, I don't see a way to test this without sleep.

m4dcoder: Ok, I'll change this. Yeah, this is hard to test...

lakshmi-kannan:
LGTM, add a changelog and address one minor comment.

Issue #3445, opened 2017-06-01
Inconsistency in Pack register content


armab:
### 1) st2 pack install
When installing a pack `st2 pack install` the following content is registered by default: [`actions,aliases,sensors,triggers`](https://github.com/StackStorm/st2/blob/master/contrib/packs/actions/install.meta.yaml#L19) for a specific pack.

### 2) st2 pack register
When registering a pack with `st2 pack register` we pick up everything that's possible in this pack by default:

```
$ st2 pack register github
+--------------+-------+
| Property     | Value |
+--------------+-------+
| actions      | 20    |
| aliases      | 7     |
| configs      | 0     |
| policies     | 0     |
| policy_types | 3     |
| rule_types   | 2     |
| rules        | 1     |
| runners      | 13    |
| sensors      | 1     |
| triggers     | 0     |
+--------------+-------+
```

### 3) st2ctl reload
One more example of inconsistency is when we do `st2ctl reload` which registers by default another set of components for packs: `actions, aliases, sensors, triggers, configs, rules` https://github.com/StackStorm/st2/blob/b30008cdc777d46d40e13609b874210d3a9d0b36/st2common/bin/st2ctl#L133

---

So we have 3 situations where we register different content by default for a pack. This leads to a different confusing error-prone scenarios when you got different functionality working for a pack, depending on situation.

IMO at minimum we should pick for `st2 pack install` at least the same amount of content as we do for `st2ctl reload`, at maximum: register all content, same as for st2 pack register. Still `st2ctl reload` defaults probably needs a revision, WDYT?
cc @Kami @emedvedev @enykeev please advice, since you're more aware about the corner cases/context.

enykeev:
Frankly, I don't remember why we leave it that way during Pack API PR. Most likely we just deemed the change to be outside of PR scope. We also had plans to make `st2 pack register` the only way of registering new content.

From the top of my head, the only reason I see not to register everything on `packs.install` is that it could take too much time and potentially exceed default action timeout.

PR #3452, opened 2017-06-07
Consistent register content during 'pack install'

armab:
Fixes https://github.com/StackStorm/st2/issues/3445

The problem with the current approach is that `st2ctl reload` & `st2 pack install` & `st2 pack register` register different type of content by default, leading to confusion and error-prone scenarios when you got different functionality working for a pack, depending on situation.

---

With this PR we register `all` available content in pack by default during the `st2 pack install`.
This is already default behavior for `st2 pack register`, - follow that.


Kami:
Thanks.

I'm fine with that change - I originally left is so it only registered this content type because that's what happened in the past (pre-content management).

Let's please also add a changelog entry.

Issue #3377, opened 2017-04-26
st2 pack show in master is broken


lakshmi-kannan:
```
(virtualenv)vagrant@st2dev /m/s/s/st2 ❯❯❯ st2 pack show core                                                                     master ✭ ◼
ERROR: Expecting value: line 1 column 1 (char 0)
```

with --debug

```
(virtualenv)vagrant@st2dev /m/s/s/st2 ❯❯❯ st2 --debug pack show core                                                           ⏎ master ✭ ◼
# -------- begin 139705409745296 request ----------
curl -X POST -H  'Connection: keep-alive' -H  'Accept-Encoding: gzip, deflate' -H  'Accept: */*' -H  'User-Agent: python-requests/2.13.0' -H  'content-type: application/json' -H  'Content-Length: 16' --data-binary '{"pack": "core"}' http://127.0.0.1:9101/v1/packs/index/search
# -------- begin 139705409745296 response ----------

# -------- end 139705409745296 response ------------

ERROR: Expecting value: line 1 column 1 (char 0)
CLI settings:
----------------
Config file path: /home/vagrant/.st2/config
Client settings:
----------------
ST2_BASE_URL: http://127.0.0.1
ST2_AUTH_URL: http://127.0.0.1:9100
ST2_API_URL: http://127.0.0.1:9101/v1
ST2_AUTH_TOKEN: None

Proxy settings:
---------------
HTTP_PROXY:
HTTPS_PROXY:

Traceback (most recent call last):
  File "/mnt/src/storm/st2/st2client/st2client/shell.py", line 289, in run
    args.func(args)
  File "/mnt/src/storm/st2/st2client/st2client/commands/pack.py", line 85, in run_and_print
    raise OperationFailureException(message)
OperationFailureException: Expecting value: line 1 column 1 (char 0)
(virtualenv)vagrant@st2dev /m/s/s/st2 ❯❯❯
```

Mierdin:
Seems like this is because `core` isn't in the index (naturally, it's installed with stackstorm) and this functionality works by searching the index only. Wondering if right approach is to maintain a list of preinstalled packs in the API controller and if requests come in for one of them, don't look at the index, rather pull metadata from filesystem. Thoughts?

bigmstone:
@Mierdin I believe the bug is the traceback/unfriendly error handing and not the need to add pre-installed packs to the index (which I'm -1 on, if that's what we're thinking).

Mierdin:
[...]

PR #3458, opened 2017-06-08
Return empty list when pack not found in index

Mierdin:
Fixes https://github.com/StackStorm/st2/issues/3377


Kami:
LGTM - the API result is indeed more consistent this way.

Please just document it in changelog and if you feel the need also upgrade notes entry since it's in theory a breaking change, yet a very small and non-significant one.

armab:
@Mierdin Also please re-run the failed test to make sure we're all green.

Mierdin:
@armab Re-ran the u16 execution and it succeeded - not sure why Github isn't showing the full 8 checks though.

armab:
@Mierdin Thanks! Seems it's green now :+1:

PR #3461, opened 2017-06-09
Add PUT method to action executions API

m4dcoder:
Allow action executions to be updated over API. The API will only allow status and result to be updated for incomplete action executions. This patch is required for 1) mistral->st2 callback on workflow completion and 2) st2 response to update st2.ask action execution.

### DIFF
--- a/st2api/st2api/controllers/v1/actionexecutions.py
+++ b/st2api/st2api/controllers/v1/actionexecutions.py
@@ -29,7 +29,7 @@ from st2api.controllers.v1.executionviews import ExecutionViewsController
 from st2api.controllers.v1.executionviews import SUPPORTED_FILTERS
 from st2common import log as logging
 from st2common.constants.action import LIVEACTION_STATUS_CANCELED, LIVEACTION_STATUS_FAILED
-from st2common.constants.action import LIVEACTION_CANCELABLE_STATES
+from st2common.constants.action import LIVEACTION_CANCELABLE_STATES, LIVEACTION_COMPLETED_STATES
 from st2common.exceptions.param import ParamException
 from st2common.exceptions.apivalidation import ValueValidationException
 from st2common.exceptions.trace import TraceNotFoundException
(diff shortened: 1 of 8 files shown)


st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -469,6 +469,58 @@ def post(self, liveaction_api, requester_user, context_string=None, show_secrets
...
+
+        liveaction_id = execution_api.liveaction['id']
+        if not liveaction_id:
+            abort(http_client.INTERNAL_SERVER_ERROR,
+                  'Execution object missing link to liveaction %s.' % liveaction_id)
+
+        try:
+            liveaction_db = LiveAction.get_by_id(liveaction_id)
+        except:
+            abort(http_client.INTERNAL_SERVER_ERROR,
```
Kami: Perhaps 404 would be better there?

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -469,6 +469,58 @@ def post(self, liveaction_api, requester_user, context_string=None, show_secrets
...
+                  'Execution object missing link to liveaction %s.' % liveaction_id)
+
+        if liveaction_db.status in LIVEACTION_COMPLETED_STATES:
+            abort(http_client.BAD_REQUEST, 'Execution is already in completed state.')
+
+        try:
+            liveaction_db = action_service.update_status(
+                liveaction_db,
+                liveaction_api.status,
```
Kami: Wonder if we should explicitly throw or similar if user provides LiveActionAPI which has any other attribute then `status` updates.

I know we will document that only status can be changed, but I'm sure someone will still try to change some other attribute and be confused why it's not working. So in short, I would prefer to throw 400 if user tries to update other attribute then `status`.

Kami: Ignore this comment above.

I finally got to the end of diff and saw you limited possible attributes via openapi schema - https://github.com/StackStorm/st2/pull/3461/files#diff-c6bc5bca320e32b130cd88ba7a961ad9R3956.

Kami:
LGTM.

st2api/st2api/controllers/v1/actionexecutions.py
m4dcoder: Yep, I limited it via the schema and there are unit tests for these scenarios.

st2api/st2api/controllers/v1/actionexecutions.py
m4dcoder: The delete method is already using 500 for similar cases at https://github.com/StackStorm/st2/blob/master/st2api/st2api/controllers/v1/actionexecutions.py#L501. Also, it's technically not 404 because the execution which the user is updating/deleting is found but the reference to liveaction in the execution object is missing. That's an internal system error for whatever reason.

PR #3466, opened 2017-06-12
Fix while loop in worker shutdown unit test

m4dcoder:
Change the while loop into a for loop to ensure the test exit after set time.


Kami:
LGTM

PR #3472, opened 2017-06-13
Fix syntax in changelog.rst, validate changelog rst file syntax on CI

Kami:
I also added rstcheck Make target which runs on Circle CI on every build.

This will prevent such issues from happening in the future - https://github.com/StackStorm/st2docs/pull/521#issuecomment-308170615.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -912,21 +912,6 @@ in development
   easier since user simply needs to start sensor container with ``--debug`` flag and all the sensor
   logs with level debug or higher will be routed to the container log. (improvement)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -912,21 +912,6 @@ in development
   easier since user simply needs to start sensor container with ``--debug`` flag and all the sensor
   logs with level debug or higher will be routed to the container log. (improvement)
 
-0.13.2 - September 09, 2015
```
Kami: It was a duplicate (detected by the new lint target).

Probably from  a bad merge or similar.

PR #3473, opened 2017-06-13
Remove packs config

bigmstone:
Now that deprecation notice is place we need to remove empty `config.yaml` so warning isn't thrown by default.


CHANGELOG.rst
```diff
@@ -117,6 +117,8 @@ in development
   (``required: false``). (improvement)
 
   Reported by nmlaudy. #3468 #3469
+* Removing empty pack config.yaml so warning isn't thrown by default now that deprecation warning
+  is in place. (fix)
```
Kami: fix -> improvement? :trollface:

Kami: would also be good to clarify from which pack

armab: It's a fix :trollface:
Files missing from the previous PR which introduced that deprecation notice :)

Issue #3477, opened 2017-06-14
pack install fails with git > 1.9.0 


lakshmi-kannan:
```
(virtualenv)vagrant@st2dev /o/s/p/vsphere ❯❯❯ st2 pack install vyatta                                                              master ✱

        [  failed   ] download pack

id: 59417552d9d7ed69b9b2d9c3
action.ref: packs.install
parameters:
  packs:
  - vyatta
status: failed
error: Traceback (most recent call last):
  File "/mnt/src/storm/st2/st2common/st2common/runners/python_action_wrapper.py", line 259, in <module>
    obj.run()
  File "/mnt/src/storm/st2/st2common/st2common/runners/python_action_wrapper.py", line 155, in run
    output = action.run(**self._parameters)
  File "/opt/stackstorm/packs/packs/actions/pack_mgmt/download.py", line 75, in run
    verifyssl=verifyssl, ref=pack_version)
  File "/opt/stackstorm/packs/packs/actions/pack_mgmt/download.py", line 150, in _clone_repo
    repo.git.checkout('-b', short_branch, branch)
  File "/mnt/src/storm/st2/virtualenv/lib/python2.7/site-packages/git/cmd.py", line 425, in <lambda>
    return lambda *args, **kwargs: self._call_process(name, *args, **kwargs)
  File "/mnt/src/storm/st2/virtualenv/lib/python2.7/site-packages/git/cmd.py", line 877, in _call_process
    return self.execute(call, **exec_kwargs)
  File "/mnt/src/storm/st2/virtualenv/lib/python2.7/site-packages/git/cmd.py", line 688, in execute
    raise GitCommandError(command, status, stderr_value, stdout_value)
git.exc.GitCommandError: Cmd('git') failed due to: exit code(128)
  cmdline: git checkout -b master master
  stderr: 'fatal: Cannot update paths and switch to branch '?[7;33mmaster?[m' at the same time.
Did you intend to checkout '?[7;33mmaster?[m' which can not be resolved as commit?'

traceback: None
failed_on: download pack
start_timestamp: 2017-06-14T17:41:38.485470Z
end_timestamp: 2017-06-14T17:41:43.411575Z
+--------------------------+---------------------+---------------+----------------+----------------------+
| id                       | status              | task          | action         | start_timestamp      |
+--------------------------+---------------------+---------------+----------------+----------------------+
[...]

PR #3478, opened 2017-06-14
Only checkout if short_branch != branch

lakshmi-kannan:
Fixes: https://github.com/StackStorm/st2/issues/3477


Kami:
Did you figure this one out?

Was it related to a non-bash shell being used?

lakshmi-kannan:
>Was it related to a non-bash shell being used?

Yeah, it's my zshell. Nevertheless, this fix is a reasonable one IMO.

Kami:
I'm fine with it as long as we test it well.

We had a lot of bugs and edge cases related to this code.

If you verified all the various scenarios (checking out non-default branch, default branch, non-master branch is a default one, etc). and we add a test case for it I'm fine with merging it.

Kami:
Is this still relevant?

lakshmi-kannan:
I'll close this. Will reopen when I have time to actually test this for all cases.

Issue #3343, opened 2017-04-08
New regex filter for Jinja2 expressions


Mierdin:
In https://github.com/StackStorm/st2/blob/master/st2common/st2common/jinja/filters/regex.py, there are 3 possible filters that can be used in a Jinja2 expression in a rule. Unfortunately, the only function that returns a string is `regex_replace`. A user in the slack community wanted to be able to filter some text from a trigger through a filter, and use the resulting value - sort of like `regex_search` but returning the matched string, instead of a boolean.

So basically we need to add a filter that does this. The functionality of `regex_search` but returning the matched string instead of a boolean - and it would be nice if it would honor capture groups as well.

armab:
:+1: for more useful filters. That's one of the advantages that makes Ansible so useful, [for example](http://docs.ansible.com/ansible/playbooks_filters.html).

> Also not similar, but related: https://github.com/StackStorm/st2/issues/2956 will be to allow users defining their cusom filters "on fly".

PR #3482, opened 2017-06-15
Implement new regex_substring jinja filter

Mierdin:
Closes https://github.com/StackStorm/st2/issues/3343

### DIFF
--- a/contrib/examples/actions/chains/regex_jinja_filter.yaml
+++ b/contrib/examples/actions/chains/regex_jinja_filter.yaml
@@ -18,7 +18,7 @@
 
         # Searches for pattern (first param) and replaces all matches with the provided
         # string (second param) and returns the final result
-        cmd: 'echo {{sample_str | regex_match("([0-9]{3} \w+ (?:Ave|St|Dr))", "<redacted>")}}'
+        cmd: 'echo {{sample_str | regex_replace("([0-9]{3} \w+ (?:Ave|St|Dr))", "<redacted>")}}'
       on-success: "search_for_str"
     -
       name: "search_for_str"
--- a/contrib/examples/actions/chains/regex_jinja_filter.yaml
+++ b/contrib/examples/actions/chains/regex_jinja_filter.yaml
@@ -1,0 +1,47 @@
+---
+  vars:
+    sample_str: 'Your address is 567 Elsewhere Dr. My address is 123 Somewhere Ave.'
+  chain:
+    -
+      name: "is_match"
+      ref: "core.local"
+      parameters:
+
+        # Just returns True if match is found at beginning, False if not (uses re.match)
+        # Since the address is not at the beginning of the string, this will return False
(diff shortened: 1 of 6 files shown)


st2common/tests/unit/test_jinja_render_regex_filters.py
```diff
@@ -68,3 +68,24 @@ def test_filters_regex_search(self):
         actual = env.from_string(template).render({'k1': 'v0.10.1'})
         expected = 'True'
         self.assertEqual(actual, expected)
+
+    def test_filters_regex_substring(self):
+        env = jinja_utils.get_jinja_environment()
+
+        # Normal (match)
+        template = '{{k1 | regex_substring("(127.0.0.1)")}}'
```
Kami: I assume user can also user other regex modifiers such as `\w+`, `.*`, etc. in the substring, right?

Otherwise functionality makes little sense since it just returns an original static sub-string as passed in.

If that's the case, it would also be good to add a test case with a more "realistic" scenario which uses `.*` or similar

Kami: Edit: I also see user can pass in `group_index` - would also be good to add a test case when an on-zero group index is used and when there is a match and when there is none.

Kami:
Added a couple of comments, would also be good to open a corresponding PR with docs in st2docs.

Mierdin:
@Kami yes I have "opening a PR for st2docs" in my TODO list, will do that as soon as my two filters are merged.

st2common/tests/unit/test_jinja_render_regex_filters.py
Mierdin: Good ideas - did this in https://github.com/StackStorm/st2/pull/3482/commits/223b0a33f910feeaaaf6bb3bb732a2b478848e98 (as well as refactored to use `re.findall` instead

Mierdin:
Okay I took the liberty of adding all regex examples in contrib, and I should have already addressed your other comments.

contrib/examples/actions/chains/regex_jinja_filter.yaml
```diff
@@ -0,0 +1,47 @@
...
+        cmd: 'echo {{sample_str | regex_match("([0-9]{3} \w+ (?:Ave|St|Dr))")}}'
+      on-success: "replace_string"
+    -
+      name: "replace_string"
+      ref: "core.local"
+      parameters:
+
+        # Searches for pattern (first param) and replaces all matches with the provided
+        # string (second param) and returns the final result
+        cmd: 'echo {{sample_str | regex_match("([0-9]{3} \w+ (?:Ave|St|Dr))", "<redacted>")}}'
```
lakshmi-kannan: This should probably be regex_replace. match takes only one argument.

Mierdin: Oops, yeah that's a typo. Good catch

Issue #3369, opened 2017-04-19
/api/v1/apikeys list does not have total in result


m4dcoder:
Can /api/v1/apikeys paginate?

humblearner:
@m4dcoder: Are you looking for something similar to: https://github.com/StackStorm/st2/blob/master/st2common/st2common/openapi.yaml#L234, adding limit to number of keys that can be viewed?

PR #3486, opened 2017-06-16
Add pagination support for /api/v1/apikeys endpoints

humblearner:
Addresses: https://github.com/StackStorm/st2/issues/3369
```
$  curl -v -X GET -H  'Connection: keep-alive' -H  'Accept-Encoding: gzip, deflate' -H  'Accept: */*' -H  'User-Agent: python-requests/2.14.2' -H  'X-Auth-Token: <token>' 'http://127.0.0.1:9101/v1/apikeys/?limit=1'

*   Trying 127.0.0.1...
* Connected to 127.0.0.1 (127.0.0.1) port 9101 (#0)
> GET /v1/apikeys/?limit=1 HTTP/1.1
> Host: 127.0.0.1:9101
> Connection: keep-alive
> Accept-Encoding: gzip, deflate
> Accept: */*
> User-Agent: python-requests/2.14.2
> X-Auth-Token: <token>
>
< HTTP/1.1 200 OK
< Server: gunicorn/19.6.0
< Date: Fri, 16 Jun 2017 16:15:07 GMT
< Connection: keep-alive
< Content-Type: application/json
< Content-Length: 245
< X-Total-Count: 2
< X-Limit: 1
< Access-Control-Allow-Origin: http://127.0.0.1:3000
< Access-Control-Allow-Methods: GET,POST,PUT,DELETE,OPTIONS
< Access-Control-Allow-Headers: Content-Type,Authorization,X-Auth-Token,St2-Api-Key,X-Request-ID
< Access-Control-Expose-Headers: Content-Type,X-Limit,X-Total-Count,X-Request-ID
< X-Request-ID: 658ab5e1-b874-4acb-b0af-d47ece4ff7a1
<
[
    {
        "uid": "********",
        "key_hash": "********",
        "created_at": "2017-06-14T19:44:47.496745Z",
        "enabled": true,
        "user": "st2admin",
        "id": "5941922fde59a4212d147155",
        "metadata": {}
    }
* Connection #0 to host 127.0.0.1 left intact
```
To do:
- [x]  Fix corner cases of invalid limit and offset count
- [x]   Add Unit tests

### DIFF
--- a/st2api/st2api/controllers/v1/auth.py
+++ b/st2api/st2api/controllers/v1/auth.py
@@ -105,7 +105,7 @@ class ApiKeyController(BaseRestControllerMixin):
         resp.headers['X-Total-Count'] = str(api_key_dbs.count())
 
         if limit:
-             resp.headers['X-Limit'] = str(limit)
+            resp.headers['X-Limit'] = str(limit)
 
         return resp
 
--- a/st2api/st2api/controllers/v1/auth.py
+++ b/st2api/st2api/controllers/v1/auth.py
@@ -88,7 +88,7 @@ class ApiKeyController(BaseRestControllerMixin):
             LOG.exception('Failed to serialize API key.')
             abort(http_client.INTERNAL_SERVER_ERROR, str(e))
 
-    def get_all(self, requester_user, show_secrets=None):
+    def get_all(self, requester_user, show_secrets=None, limit=None, offset=0):
         """
             List all keys.
 
@@ -97,11 +97,17 @@ class ApiKeyController(BaseRestControllerMixin):
         """
         mask_secrets = self._get_mask_secrets(show_secrets=show_secrets,
                                               requester_user=requester_user)
-        api_key_dbs = ApiKey.get_all()
+        api_key_dbs = ApiKey.get_all(limit=limit, offset=offset)
         api_keys = [ApiKeyAPI.from_model(api_key_db, mask_secrets=mask_secrets)
                     for api_key_db in api_key_dbs]
 
-        return api_keys
+        resp = Response(json=api_keys)
+        resp.headers['X-Total-Count'] = str(api_key_dbs.count())
+
+        if limit:
+             resp.headers['X-Limit'] = str(limit)
+
+        return resp
 
     def post(self, api_key_api, requester_user):
         """
--- a/st2api/st2api/controllers/v1/auth.py
+++ b/st2api/st2api/controllers/v1/auth.py
@@ -112,7 +112,7 @@ class ApiKeyController(BaseRestControllerMixin):
             api_keys = [ApiKeyAPI.from_model(api_key_db, mask_secrets=mask_secrets)
                         for api_key_db in api_key_dbs]
         except OverflowError:
-            msg = 'Offset "%s" specified is more than 8-byte ints' % (offset)
+            msg = 'Offset "%s" specified is more than 32 bit int' % (offset)
             raise ValueError(msg)
 
         resp = Response(json=api_keys)
--- a/st2api/st2api/controllers/v1/auth.py
+++ b/st2api/st2api/controllers/v1/auth.py
@@ -88,6 +88,10 @@ class ApiKeyController(BaseRestControllerMixin):
             LOG.exception('Failed to serialize API key.')
             abort(http_client.INTERNAL_SERVER_ERROR, str(e))
 
+    @property
+    def max_limit(self):
+        return cfg.CONF.api.max_page_size
+
     def get_all(self, requester_user, show_secrets=None, limit=None, offset=0):
         """
             List all keys.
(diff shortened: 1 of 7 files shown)


st2api/st2api/controllers/v1/auth.py
```diff
@@ -97,11 +101,27 @@ def get_all(self, requester_user, show_secrets=None):
         """
         mask_secrets = self._get_mask_secrets(show_secrets=show_secrets,
                                               requester_user=requester_user)
-        api_key_dbs = ApiKey.get_all()
-        api_keys = [ApiKeyAPI.from_model(api_key_db, mask_secrets=mask_secrets)
-                    for api_key_db in api_key_dbs]
 
-        return api_keys
+        if limit and int(limit) > self.max_limit:
```
Kami: Hm, just curious, why does this need to be duplicated here (we already do it in some base controller)?

Or is it because that base controller only applies and is used with pack resources?

st2api/st2api/controllers/v1/auth.py
```diff
@@ -97,11 +101,27 @@ def get_all(self, requester_user, show_secrets=None):
...
+            msg = 'Limit "%s" specified, maximum value is "%s"' % (limit, self.max_limit)
+            raise ValueError(msg)
+
+        api_key_dbs = ApiKey.get_all(limit=limit, offset=offset)
+
+        try:
+            api_keys = [ApiKeyAPI.from_model(api_key_db, mask_secrets=mask_secrets)
+                        for api_key_db in api_key_dbs]
+        except OverflowError:
```
Kami: Hm, why do we need to do that here? IIRC, we don't do something like that in the base controller where we handle pagination for most other resources.

Kami:
Thanks.

Added a couple of comments. Mostly looks good to me, just need to clarify some things in regards to code duplication.

st2api/st2api/controllers/v1/auth.py
Kami: And if that is required, we should also add it to the base controller where we handle pagination for other stuff.

Kami: On a related note - is overflow error also thrown if user provides offset which is bigger than number of items? If it does, then we need do modify the exception message.

st2api/st2api/controllers/v1/auth.py
humblearner: Yes, we are using this here in the resource controller: https://github.com/StackStorm/st2/blob/1043ed8b2659625bb02b58e52639d7832a7a8974/st2api/st2api/controllers/resource.py#L160, but are not using it as a base class in this case.

st2api/st2api/controllers/v1/auth.py
humblearner: This is because of mongo limitation: ```2017-06-19 03:49:58,210 139885476219952 ERROR router [-] Failed to call controller function "get_all" for operation "st2api.co
ntrollers.v1.auth:api_key_controller.get_all": Python int too large to convert to C long
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/router.py", line 414, in __call__
    resp = func(**kw)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2api/controllers/v1/auth.py", line 114, in get_all
    for api_key_db in api_key_dbs:
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/mongoengine/queryset/queryset.py", line 97, in _iter_results
    self._populate_cache()
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/mongoengine/queryset/queryset.py", line 116, in _populate_cache
    self._result_cache.append(self.next())
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/mongoengine/queryset/base.py", line 1407, in next
    raw_doc = self._cursor.next()
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/pymongo/cursor.py", line 1114, in next
    if len(self.__data) or self._refresh():
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/pymongo/cursor.py", line 1036, in _refresh
    self.__collation))
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/pymongo/cursor.py", line 873, in __send_message
    **kwargs)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/pymongo/mongo_client.py", line 905, in _send_message_with_respon
se
    exhaust)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/pymongo/mongo_client.py", line 916, in _reset_on_error
    return func(*args, **kwargs)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/pymongo/server.py", line 124, in send_message_with_response
    set_slave_okay, sock_info.is_mongos, use_find_cmd)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/pymongo/message.py", line 291, in get_message
    spec, self.fields, self.codec_options)
OverflowError: Python int too large to convert to C long```. 
For these endpoints we don't have pagination handling in the base controller. No, overflow error is only thrown when the offset > 8 byte int

st2api/tests/unit/controllers/v1/test_auth_api_keys.py
```diff
@@ -72,12 +72,47 @@ def test_get_all(self):
...
         retrieved_ids = [apikey['id'] for apikey in resp.json]
-
         self.assertEqual(retrieved_ids,
                          [str(self.apikey1.id), str(self.apikey2.id), str(self.apikey3.id),
                           str(self.apikey4.id), str(self.apikey5.id)],
                          'Incorrect api keys retrieved.')
 
+    def test_get_all_with_pagnination_with_offset_and_limit(self):
+        resp = self.app.get('/v1/apikeys?offset=2&limit=1')
+        self.assertEqual(resp.status_int, 200)
```
Kami: Please also add additional asserts to every test case for values of `X-Total-Count` and `X-Limit` header.

Besides that, LGTM :+1:

st2api/st2api/controllers/v1/auth.py
Kami: Thanks for clarification.

That works for not, but if we need to do that in more places, refactor is in order :)

humblearner: @Kami: Good point. Apikeys endpoint use https://github.com/StackStorm/st2/blob/master/st2common/st2common/persistence/auth.py#L84. Not sure if we can refactor here. However, all other endpoints were not even addressing OverflowError, fixed here: https://github.com/StackStorm/st2/pull/3486/commits/489fa2d357bbd339f887f69da2d7e4df798214dc. Also, some endpoints on openapi.yaml didn't have offset as a parameter.

PR #3489, opened 2017-06-19
Python runner should throw / log an error when failing to de-serialize action result

Kami:
@tonybaloney reported this issue (user unfriendly behavior) on Slack.

Right now, if Python runner actions returns a result which can't be serialized (e.g. non complex type such as class instance, datetime object or similar), Python runner will "silently" swallow this issue and simply return `None` as the result.

This is obviously very unfriendly and confusing.

The code change I included just logs a message and still returns `None`, but I think a better and more user-friendly approach is to simply throw a very explicit and user-friendly error message - something along the lines of "Failed to de-serialize action result, it most likely contains a non-simple type or similar, see docs at YYY on info on Python runner action results".

The thing is if we can't serialize and de-serialize the result user can't do anything with it anyway so throw makes sense because user / developer interaction is required (updating the action code so it only returns simple types).

@tonybaloney asked if we recently made a change which affects that, but AFAIK this behavior has been there since the beginning - I checked the code and we did the same thing in v1.6 as well (https://github.com/StackStorm/st2/blob/v1.6/st2actions/st2actions/runners/pythonrunner.py#L190).

Example action which reproduces this issue - https://gist.github.com/Kami/7e654186f95eda654589a171c9f90d12

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,13 +6,13 @@ in development
 
 * Update ``st2 run`` / ``st2 execution run`` command to display result of workflow actions when
   they finish. In the workflow case, result of the last task (action) of the workflow is used.
-  (improvement)
+  (improvement) #3481
 * Update Python runner so it mimics behavior from StackStorm pre 1.6 and returns action result as
   is (serialized as string) in case we are unable to serialize action result because it contains
   non-simple types (e.g. class instances) which can't be serialized.
 
   In v1.6 we introduced a change when in such instances, we simply returned ``None`` as result
-  and didn't log anything which was confusing. (improvement)
+  and didn't log anything which was confusing. (improvement) #3489
 
   Reported by Anthony Shaw.
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,14 @@ Changelog
 in development
 --------------
 
+* Update Python runner so it mimics behavior from StackStorm pre 1.6 and returns action result as
+  is (serialized as string) in case we are unable to serialize action result because it contains
+  non-simple types (e.g. class instances) which can't be serialized.
+
+  In v1.6 we introduced a change when in such instances, we simply returned ``None`` as result
+  and didn't log anything which was confusing. (improvement)
+
+  Reported by Anthony Shaw.
 
 2.3.0 - June 19, 2017
 ---------------------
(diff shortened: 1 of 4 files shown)


Kami:
Managed to track down old behavior @tonybaloney was described to v1.5 - https://github.com/StackStorm/st2/blob/v1.5/st2actions/st2actions/runners/pythonrunner.py#L158.

The different was that at that point in time, we only tried to de-serialize it if user returned JSON from action, otherwise we always simply returned a string (which wasn't all that useful).

I will look if we can replicate pre v1.6 behavior, but I still think correct long term approach would be throwing.

Kami:
I pushed a change to utilize pre-v1.6 behavior in -

Now if we fail to de-serialize the result because it contains a non-simple type we simply return result as is aka serialized as a string.

contrib/runners/python_runner/python_runner.py
```diff
@@ -178,14 +179,22 @@ def _get_output_values(self, exit_code, stdout, stderr, timed_out):
...
+            LOG.warning('Failed to de-serialize result "%s": %s' % (str(action_result), str(e)))
             pass
 
         if action_result and isinstance(action_result, dict):
             result = action_result.get('result', None)
             status = action_result.get('status', None)
         else:
-            result = 'None'
+            # Failed to de-serialize return result as is aka as a string
+            match = re.search("'result': (.*?)$", action_result).groups()
```
Kami: Yes, this is horrible but we have a safe-guard below - if there are no results we simply return result as is.

And as mentioned in other comment, I believe that correct long term approach would be to throw.

Kami:
P.S. Thanks again for @tonybaloney for catching and reporting this.

tonybaloney:
:+1:

CHANGELOG.rst
```diff
@@ -7,6 +7,14 @@ in development
 * Update ``st2 run`` / ``st2 execution run`` command to display result of workflow actions when
   they finish. In the workflow case, result of the last task (action) of the workflow is used.
   (improvement)
+* Update Python runner so it mimics behavior from StackStorm pre 1.6 and returns action result as
+  is (serialized as string) in case we are unable to serialize action result because it contains
+  non-simple types (e.g. class instances) which can't be serialized.
+
+  In v1.6 we introduced a change when in such instances, we simply returned ``None`` as result
+  and didn't log anything which was confusing. (improvement)
```
armab: let's please include the PR# in the changelog.

Could be a good tradition to follow, since it's easier to find the change later.

Kami: Will do - I agree it's a good idea and we should do it for all the changes :)

Kami: Done - 21b8fa8.

@lakshmi-kannan @m4dcoder @bigmstone let's all try to go doing going forward :)

Issue #3487, opened 2017-06-16
Jinja templating broken in action notify section


cwkatzmann:
When I run an action and try to reference any of its output values in jinja template format, the literal string of the template itself is used. For example, the message output is `"@channel: Action run by {{action_context.user}} succeeded. Cmd run was {{cmd}}."`

The documentation demonstrates this feature here: https://docs.stackstorm.com/chatops/notifications.html#jinja-templating-in-message-and-data

lakshmi-kannan:
@cwkatzmann When referring to the output and parameters, are you using {{ action_parameters.cmd }} and {{ action_parameters.output }}. The context "action_parameters" is required.

lakshmi-kannan:
Related: https://github.com/StackStorm/st2/issues/3403

nmaludy:
@lakshmi-kannan i've been doing some digging on this and noticed the same thing.

@cwkatzmann original problem came from the following action yaml:

```
---
name: "guest_wifi"
runner_type: "local-shell-cmd"
parameters:
  cmd:
    type: "string"
    default: "cat /home/stanley/guest_wifi.txt"
notify:
  on-complete:
    routes:
    - chatops_override
    message: "{{action_results.stdout}}"
```


I did some digging and printed out the Jinja context before rendering the message and noticed that `action_results` is empty:
```
st2notifier.log:2017-06-16 16:00:24,003 44250928 INFO notifier [-] nmaludy: context = {'action_context': {u'user': u'st2admin'}, 'action_results': {}, 'action_parameters': {u'cmd': u'cat /etc/hosts'}, 'st2kv': {'system': <st2common.services.keyvalues.KeyValueLookup object at 0x2b0e250>}}
```

nmaludy:
https://github.com/StackStorm/st2/blob/3a615564218323be52d6a541a4eb249b7a0590cd/st2actions/st2actions/notifier/notifier.py#L210

It looks like execution.result is being used to populate the `action_result` context and, in my case, that result is empty in the `execution` object on the line of code i linked to above.

```
[...]

PR #3496, opened 2017-06-20
Notifier to scan executions queue and not liveaction queue

lakshmi-kannan:
Fixes: https://github.com/StackStorm/st2/issues/3487

### DIFF
--- a/st2actions/st2actions/notifier/notifier.py
+++ b/st2actions/st2actions/notifier/notifier.py
@@ -33,7 +33,7 @@ from st2common import policies
 from st2common.models.system.common import ResourceReference
 from st2common.persistence.execution import ActionExecution
 from st2common.services import trace as trace_service
-from st2common.transport import consumers, execution, liveaction, publishers
+from st2common.transport import consumers, execution, publishers
 from st2common.transport import utils as transport_utils
 from st2common.transport.reactor import TriggerDispatcher
 from st2common.util import isotime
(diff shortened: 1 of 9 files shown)


Kami:
So far LGTM, but needs some good test cases :)

Kami:
For other's people reference, more context in https://github.com/StackStorm/st2/issues/3487#issuecomment-309509757.

st2actions/st2actions/notifier/notifier.py
```diff
@@ -74,29 +75,24 @@ def __init__(self, connection, queues, trigger_dispatcher=None):
             pack=ACTION_TRIGGER_TYPE['pack'],
             name=ACTION_TRIGGER_TYPE['name'])
 
-    def process(self, liveaction):
-        live_action_id = str(liveaction.id)
-        extra = {'live_action_db': liveaction}
-        LOG.debug('Processing liveaction %s', live_action_id, extra=extra)
+    def process(self, execution):
```
Kami: While making changes, we could be more explicit and consistent and call argument `execution_db`.

st2actions/st2actions/notifier/notifier.py
```diff
@@ -50,8 +51,8 @@
 
 LOG = logging.getLogger(__name__)
 
-ACTIONUPDATE_WORK_Q = liveaction.get_queue('st2.notifiers.work',
-                                           routing_key=publishers.UPDATE_RK)
+ACTIONUPDATE_WORK_Q = execution.get_queue('st2.notifiers.work',
```
Kami: Hm, I think this should be backward compatible (no messages will be lost and there should be no duplication) because of the way how live actions and executions are updated, right?

lakshmi-kannan: I'll have to test but it's probably safe to call this new something else. In a HA setup, people might do rolling restarts and this could be an issue.

Kami:
LGTM :+1:

PR #3504, opened 2017-06-22
Fix boolean config value race condition

geriBatai:
This PR fixes race condition for boolean config values. When pack schema had

```
type: boolean
default: true
```
Overriding it in pack's config with `value: false`, would still give `value: true`.

I apologize for not including tests in this PR, but I could not quickly find a way how I could pass boolean value to `set_datastore_value_for_config_key` in `st2common/tests/unit/test_config_loader.py` . If someone could point me the right direction, I would be glad to add unit test for that.

### DIFF
--- a/st2common/st2common/util/config_loader.py
+++ b/st2common/st2common/util/config_loader.py
@@ -156,8 +156,9 @@ class ContentPackConfigLoader(object):
             default_value = schema_item.get('default', None)
             is_object = schema_item.get('type', None) == 'object'
             has_properties = schema_item.get('properties', None)
+            config_value = config.get(schema_item_key, None)
 
-            if default_value and not config.get(schema_item_key, None):
+            if default_value and config_value is None:
                 config[schema_item_key] = default_value
 
             # Inspect nested object properties
--- a/st2common/st2common/util/config_loader.py
+++ b/st2common/st2common/util/config_loader.py
@@ -156,15 +156,12 @@ class ContentPackConfigLoader(object):
             default_value = schema_item.get('default', None)
             is_object = schema_item.get('type', None) == 'object'
             has_properties = schema_item.get('properties', None)
-            config_value = config.get(schema_item_key, None)
 
-            if default_value and config_value is None:
-                config[schema_item_key] = default_value
+            config.setdefault(schema_item_key, default_value)
 
             # Inspect nested object properties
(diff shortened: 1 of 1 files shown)


st2common/st2common/util/config_loader.py
```diff
@@ -156,8 +156,9 @@ def _assign_default_values(self, schema, config):
             default_value = schema_item.get('default', None)
             is_object = schema_item.get('type', None) == 'object'
             has_properties = schema_item.get('properties', None)
+            config_value = config.get(schema_item_key, None)
```
lakshmi-kannan: config_value = config.get(schema_item_key, default_value) should do it, no?

geriBatai: Ah! TBH, I wasn't really looking into assign_default_values logic and didn't realize that config is a simple dictionary, I was just fixing distinction between None and False retvals in conditional. PR updated.

st2common/st2common/util/config_loader.py
```diff
@@ -157,13 +157,11 @@ def _assign_default_values(self, schema, config):
             is_object = schema_item.get('type', None) == 'object'
             has_properties = schema_item.get('properties', None)
 
-            if default_value and not config.get(schema_item_key, None):
-                config[schema_item_key] = default_value
+            config.setdefault(schema_item_key, default_value)
```
bigmstone: 👍

Kami:
Thanks, good catch, but can you please add test cases for it?

Kami:
And please also update changelog :)

Kami:
@geriBatai  How is this looking?

Would be great if we can get it finished soon so we can get it included in v2.3.1.

If you don't have time in the near time, let me know and I will pick it over and finish it.

Kami:
Also, looking at the failing tests a bit, looks like this change will be a bit more involved and we need to be careful to not break something.

I think the best way to start with this change would be to add failing test cases (scenarios which are not correctly handled right now) and work from there to make all the tests pass.

Kami:
Here we go - #3531.

Kami:
I will go ahead and merge #3531 because we are in a bit of a rush to make sure that change makes it into the next release (v2.3.1).

Thanks again for catching, reporting and working on this :+1:

Issue #3500, opened 2017-06-21
Config now has resultstracker and results_tracker groups. 


lakshmi-kannan:
https://github.com/StackStorm/st2/blob/master/st2actions/st2actions/resultstracker/config.py#L47

and 

https://github.com/StackStorm/st2/blob/master/st2common/st2common/config.py#L279

PR #3505, opened 2017-06-22
s/results_tracker/resultstracker in config

lakshmi-kannan:
Fixes: https://github.com/StackStorm/st2/issues/3500

Context: 

There was already a ``resultstracker`` config group but I accidentally shipped the new configs in ``results_tracker`` config section. With this PR, I am adding a deprecation warning for ``results_tracker`` section but we will still parse it. Going forward, 

The config will stay 

```
[resultstracker]
query_interval = 0.1
```

Since we introduced ``results_tracker`` as a config group, we'll support it for this version and then deprecate the config. So we'll pick up 

```
[results_tracker]
query_interval = 0.1
```

This will however spit out a warning to remind people.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -44,6 +44,9 @@ Fixed
   resolved. (bug-fix) #3487 #3496
 
   Reported by Chris Katzmann, Nick Maludy.
+* Deprecate ``results_tracker`` config group and move configuration variables to ``resultstracker``
+  group instead. If you have ``results_tracker`` config group in the config, it is recommended
+  to switch to ``resultstracker`` instead. (bug-fix) #3500
 
 
 2.3.0 - June 19, 2017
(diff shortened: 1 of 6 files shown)


lakshmi-kannan:
Will update CHANGELOG before merge to master.

m4dcoder:
Btw, does it make sense to move the configs at https://github.com/StackStorm/st2/blob/v2.2/st2actions/st2actions/resultstracker/config.py to the same location as the configs here?

lakshmi-kannan:
>Btw, does it make sense to move the configs at https://github.com/StackStorm/st2/blob/v2.2/st2actions/st2actions/resultstracker/config.py to the same location as the configs here?

I thought about it but the jitter_interval etc are specific to mistral runner and they are in st2common. So from a programmer perspective, it made sense to lump them all in st2common.

Kami:
We still have some inconsistencies like `system_user`, etc.

Down the road, I would prefer to make it consistent and just use `_` everywhere.

PR #3507, opened 2017-06-23
Implement pause and resume for Mistral runner

m4dcoder:
Mistral already supports pause and resume of workflow execution. This patch exposes the pause and resume functionality in the Mistral runner.


contrib/runners/mistral_v2/mistral_v2.py
```diff
@@ -363,6 +363,74 @@ def resume(self, ex_ref, task_specs):
...
+
+        # If workflow is executed under another parent workflow, pause the corresponding
+        # action execution for the task in the parent workflow.
+        if 'parent' in getattr(self, 'context', {}):
+            mistral_action_ex_id = mistral_ctx.get('action_execution_id')
+            self._client.action_executions.update(mistral_action_ex_id, 'PAUSED')
+
+        # Identify the list of action executions that are workflows and cascade pause.
+        for child_exec_id in self.execution.children:
+            child_exec = ActionExecution.get(id=child_exec_id)
```
Kami: I assume this could throw up in some edge case where execution doesn't exist?

contrib/runners/mistral_v2/mistral_v2.py
```diff
@@ -363,6 +363,74 @@ def resume(self, ex_ref, task_specs):
...
+            if (child_exec.runner['name'] in action_constants.WORKFLOW_RUNNER_TYPES and
```
Kami: So it looks like this will also be run for action-chain workflows?

st2actions/st2actions/container/base.py
```diff
@@ -60,12 +60,21 @@ def dispatch(self, liveaction_db):
...
+        funcs = {
+            action_constants.LIVEACTION_STATUS_REQUESTED: self._do_run,
+            action_constants.LIVEACTION_STATUS_SCHEDULED: self._do_run,
+            action_constants.LIVEACTION_STATUS_RUNNING: self._do_run,
+            action_constants.LIVEACTION_STATUS_CANCELING: self._do_cancel,
+            action_constants.LIVEACTION_STATUS_PAUSING: self._do_pause,
+            action_constants.LIVEACTION_STATUS_RESUMING: self._do_resume
+        }
+
+        liveaction_db = funcs[liveaction_db.status](
```
Kami: Maybe try / catch which throws a more friendly exception in case of invalid status?

I know this shouldn't really happen since it would indicate a programmer error, but still.

st2actions/st2actions/container/base.py
```diff
@@ -184,6 +193,93 @@ def _do_cancel(self, runner, runnertype_db, action_db, liveaction_db):
...
+            LOG.debug('Performing resume for runner: %s', (runner.runner_id), extra=extra)
+
+            (status, result, context) = runner.resume()
+
+            try:
+                result = json.loads(result)
+            except:
+                pass
+
+            action_completed = status in action_constants.LIVEACTION_COMPLETED_STATES
```
Kami: It looks like a lot of code is copied here from "run" method.

Since it looks like the same logic which happens at the end of the run when execution finished should also happen here (https://github.com/StackStorm/st2/blob/0536b072dd3a45c1cd55e827e3b8a93b23e59c85/st2actions/st2actions/container/base.py#L116) we should refactor that in some common method and call it in "run" and this method.

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -502,22 +502,57 @@ def put(self, id, liveaction_api, requester_user, show_secrets=False):
...
-        if liveaction_db.status in LIVEACTION_COMPLETED_STATES:
+        if liveaction_db.status in action_constants.LIVEACTION_COMPLETED_STATES:
             abort(http_client.BAD_REQUEST, 'Execution is already in completed state.')
 
+        if (getattr(liveaction_api, 'result', None) is not None and
+                liveaction_db.status in [
+                    action_constants.LIVEACTION_STATUS_PAUSING,
+                    action_constants.LIVEACTION_STATUS_PAUSED,
+                    action_constants.LIVEACTION_STATUS_RESUMING]):
+            LOG.warning(
```
Kami: Shouldn't we return 400 bad request here?

Or you have a specific reason to silently ignore it?

st2actions/st2actions/container/base.py
Kami: I mean something along the lines of: `raise KeyError('Unsupported status %s. Valid supported statuses are: "%s')` instead of a generic non-friendly KeyError.

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -502,22 +502,57 @@ def put(self, id, liveaction_api, requester_user, show_secrets=False):
...
+            )
+
         try:
-            liveaction_db = action_service.update_status(
-                liveaction_db,
-                liveaction_api.status,
-                result=getattr(liveaction_api, 'result', None)
+            if (liveaction_api.status == action_constants.LIVEACTION_STATUS_PAUSING or
+                    liveaction_api.status == action_constants.LIVEACTION_STATUS_PAUSED):
+                liveaction_db, actionexecution_db = action_service.request_pause(
```
Kami: Hm, so if execution is already in paused state we will still call request_pause?

I assume we don't do anything if it's already in pause state inside a container, but maybe we could also already bail out already and not call `request_pause` if it's already in paused state.

Or am I missing something?

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -502,22 +502,57 @@ def put(self, id, liveaction_api, requester_user, show_secrets=False):
...
+                    liveaction_api.status,
+                    result=getattr(liveaction_api, 'result', None)
+                )
+
+                actionexecution_db = ActionExecution.get(liveaction__id=str(liveaction_db.id))
+        except runner_exc.InvalidActionRunnerOperationError as e:
+            LOG.exception('Failed updating liveaction %s. %s', liveaction_db.id, str(e))
+            abort(
+                http_client.BAD_REQUEST,
+                'Failed updating execution because the specific type of request is not supported.'
```
Kami: Could we return a more specific and friendly error message here?

E.g. Something along the lines of `"PAUSE" operation is not supported for actions with runner type "foo"`.

I know this would probably require changing a messaging of the underlying exception thrown by the container / runner.

st2common/st2common/runners/base.py
```diff
@@ -102,6 +102,12 @@ def pre_run(self):
     def run(self, action_parameters):
         raise NotImplementedError()
 
+    def pause(self):
+        raise NotImplementedError()
```
Kami: Maybe we could update all those places where we throw `NotImplementedError` to include a message which is more user-friendly.

E.g.

```python
runner_name = getattr(self.runner_type_db, 'name', 'unknown')
raise NotImplementedError('"pause" operation is not supported for runner %s' % (runner_name)
```

Then we could also leverage this exception message higher up in the stack.

st2api/st2api/controllers/v1/actionexecutions.py
Kami: It looks like exception thrown by the underlying method is already quite user-friendly so we could perhaps utilize that and append it to this generic message.

https://github.com/StackStorm/st2/pull/3507/files#diff-21bb35b73032d803366de9b6ff330a9fR239

st2common/st2common/query/base.py
```diff
@@ -126,7 +126,9 @@ def _query_and_save_results(self, query_context, last_query_time=None):
                 LOG.debug('Removed state object %s.', query_context)
             return
 
-        if status in action_constants.LIVEACTION_COMPLETED_STATES:
+        if (status in action_constants.LIVEACTION_COMPLETED_STATES or
+                status == action_constants.LIVEACTION_STATUS_PAUSED):
```
Kami: Hm so it looks like we will also invoke post_run when execution is paused - https://github.com/StackStorm/st2/blob/0536b072dd3a45c1cd55e827e3b8a93b23e59c85/st2common/st2common/query/base.py#L133

Is this desired?

contrib/runners/mistral_v2/mistral_v2.py
m4dcoder: Action chain needs to support pause/resume first. I likely will implement that in a separate PR. Going to be tough making AC support pause/resume though. And the reason we should support this in AC is to be consistent so when st2.ask action is run, the workflow is put into paused regardless of Mistral or AC and don't keep action runner occupied.

contrib/runners/mistral_v2/mistral_v2.py
m4dcoder: An edge case = bug. So if that happens, this will error out and let user know.

st2actions/st2actions/container/base.py
m4dcoder: Sure.

st2actions/st2actions/container/base.py
m4dcoder: Sure.

st2api/st2api/controllers/v1/actionexecutions.py
m4dcoder: I guess I can throw an error.  I was just going to ignore the result if it's not applicable.

st2api/st2api/controllers/v1/actionexecutions.py
m4dcoder: You are missing something. The liveaction_api is what the user wants to change status to.

st2api/st2api/controllers/v1/actionexecutions.py
m4dcoder: Ok...

st2common/st2common/runners/base.py
m4dcoder: Sure.

st2common/st2common/query/base.py
m4dcoder: I have to double check. Right now post_run is only supported by Mistral runner and it communicates final status (even paused) back to Mistral.

m4dcoder: The invoke post run is needed here. In short, this will tell a super workflow that this task that is a st2 action of another mistral workflow has paused.

contrib/runners/mistral_v2/mistral_v2.py
```diff
@@ -363,6 +363,80 @@ def resume(self, ex_ref, task_specs):
...
+        # If workflow is executed under another parent workflow, pause the corresponding
+        # action execution for the task in the parent workflow.
+        if 'parent' in getattr(self, 'context', {}):
+            mistral_action_ex_id = mistral_ctx.get('action_execution_id')
+            self._client.action_executions.update(mistral_action_ex_id, 'PAUSED')
+
+        # Identify the list of action executions that are workflows and cascade pause.
+        for child_exec_id in self.execution.children:
+            child_exec = ActionExecution.get(id=child_exec_id, raise_exception=True)
+            if (child_exec.runner['name'] in action_constants.WORKFLOW_RUNNER_TYPES and
```
Kami: Shouldn't we check here for mistral-v2 runner type since action-chain is also in `WORKFLOW_RUNNER_TYPES`.

I know it should also work this way because we already filter on execution id, but still.

contrib/runners/mistral_v2/mistral_v2.py
```diff
@@ -363,6 +363,80 @@ def resume(self, ex_ref, task_specs):
...
+            self._client.action_executions.update(mistral_action_ex_id, 'RUNNING')
+
+        # Pause the main workflow execution. Any non-workflow tasks that are still
+        # running will be allowed to complete gracefully.
+        self._client.executions.update(mistral_ctx.get('execution_id'), 'RUNNING')
+
+        # Identify the list of action executions that are workflows and cascade resume.
+        for child_exec_id in self.execution.children:
+            child_exec = ActionExecution.get(id=child_exec_id, raise_exception=True)
+            if (child_exec.runner['name'] in action_constants.WORKFLOW_RUNNER_TYPES and
```
Kami: Same here.

contrib/runners/mistral_v2/query/mistral_v2.py
```diff
@@ -155,12 +171,14 @@ def _get_workflow_tasks(self, exec_id, last_query_time=None):
         result = []
 
         try:
-            query_filters = {}
+            wf_tasks = self._client.tasks.list(workflow_execution_id=exec_id)
```
Kami: How is expensive is this now that we don't so filtering server-side (aka on Mistral API) anymore?

It there any pagination or something else we need to be careful of? Or the client library takes care of that?

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -515,22 +515,48 @@ def put(self, id, liveaction_api, requester_user, show_secrets=False):
...
+
         try:
-            liveaction_db = action_service.update_status(
-                liveaction_db,
-                liveaction_api.status,
-                result=getattr(liveaction_api, 'result', None)
-            )
+            if (liveaction_api.status == action_constants.LIVEACTION_STATUS_PAUSING or
+                    liveaction_api.status == action_constants.LIVEACTION_STATUS_PAUSED):
+                liveaction_db, actionexecution_db = action_service.request_pause(
```
Kami: So I assume we intentionally allow user to request pause multiple times, even if the execution is in one of those states?

Aka because operation is idempotent or to allow user to "rectify" the situation in case something gets stuck / fails half way through?

st2common/st2common/services/action.py
```diff
@@ -216,6 +220,85 @@ def request_cancellation(liveaction, requester):
...
+    # Validate that the runner type of the action supports pause.
+    action_db = action_utils.get_action_by_ref(liveaction.action)
+
+    if not action_db:
+        raise ValueError(
+            'Unable to pause liveaction "%s" because the action "%s" '
+            'is not found.' % (liveaction.id, liveaction.action)
+        )
+
+    if action_db.runner_type['name'] not in action_constants.WORKFLOW_RUNNER_TYPES:
```
Kami: Same here - should probably only check for Mistral until we also implement it for action-chain workflows.

Kami:
Some comments and questions, but overall LGTM :+1:

contrib/runners/mistral_v2/mistral_v2.py
m4dcoder: I am using WORKFLOW_RUNNER_TYPES because we're going to support pause/resume for Action Chain as well.

contrib/runners/mistral_v2/mistral_v2.py
m4dcoder: I am using WORKFLOW_RUNNER_TYPES because we're going to support pause/resume for Action Chain as well.

st2common/st2common/services/action.py
m4dcoder: I am using WORKFLOW_RUNNER_TYPES because we're going to support pause/resume for Action Chain as well.

contrib/runners/mistral_v2/query/mistral_v2.py
m4dcoder: Not very expensive. It's still filtering by the execution ID which limits the result set. I had to remove the filter because there's an edge case where the query filter doesn't work.

st2api/st2api/controllers/v1/actionexecutions.py
m4dcoder: Not sure what lead you to this idea. But in the code here, liveaction_api contains what the user wants to change. So in this case, if the status is PAUSING or PAUSED, then we request_pause. Regarding your question, we can make it idempotent. I have to check and add some tests.

m4dcoder: So I rechecked the code. Pause and resume is not idempotent. Right now, I coded such that a workflow has to be in running state to be paused and vice versa to be resumed.  If someone tries to pause the workflow again, it'll just throw an exception in the API and thus in the CLI that the workflow is not running.

Kami:
Small thing - it looks like you forgot to add and commit automatically generated openapi.yaml file - https://github.com/StackStorm/st2/pull/3676/commits/a70d4d5cb3e413dfe6197b8d02b165ed9132b400#diff-c6bc5bca320e32b130cd88ba7a961ad9L3939

Kami:
@StackStorm/team another heads up from Slack:

> because of the change above^ (3d7aa9491631a69cf2ad56498e9b6fc252ac523d), you need to run `ST2_PIP_OPTIONS=-U make requirements` to update to latest python-mistralclient otherwise some mistral tests will fail locally. this is required because we install python-mistralclient from git and we don't pin it to a specific revision / commit so it won't get updated if pip -U flag is not used.

/cc @armab

Issue #3488, opened 2017-06-17
'st2 rule list' is limited to 50 results


armab:
By default `st2` cli command like `st2 rule list` displays only the first `50` records.
This is so error-prone when you can't find desired content in output and there is no mention that command result is limited.

Ideally is to list all results in CLI (if possible). If not, display in the table output and `--help` a note that only first 50 records are showed up.

For example `st2 action list` outputs all possible records, so at least some consistency is expected.
TODO: Check other CLI commands.

humblearner:
rule-enforcemnent has default set to 50 as well: https://github.com/StackStorm/st2/blob/master/st2common/st2common/openapi.yaml#L2859 and the CLI has `Get the list of the 50 most recent rule enforcements.` in the `--help` too. It definitely makes sense to have default limit of 50 in rule-enforcement.

armab:
I think it's not enough for a good UX. What I mean to show a hint in CLI.

### Less than 50 results: 
```
$ st2 rule list
+---------------------------------------+----------+---------------------------------------+---------+
| ref                                   | pack     | description                           | enabled |
+---------------------------------------+----------+---------------------------------------+---------+
| chatops.notify                        | chatops  | Notification rule to send results of  | True    |
|                                       |          | action executions to stream for       |         |
|                                       |          | chatops                               |         |
| examples.notify_slack                 | examples | Sample rule firing on action          | True    |
|                                       |          | completion.                           |         |
| examples.sample_rule_file_watch       | examples | Sample rule custom trigger type - add | True    |
|                                       |          | a file to be watched by               |         |
|                                       |          | file_watch_sensor in linux pack.      |         |
| examples.sample_rule_with_actiontrigg | examples | Sample rule firing on action          | True    |
| er                                    |          | completion.                           |         |
| examples.sample_rule_with_timer       | examples | Sample rule using an Interval Timer.  | False   |
| fixtures.tests.interval_timer         | fixtures | Test rule using an Interval Timer.    | False   |
+---------------------------------------+----------+---------------------------------------+---------+
```

### If there are more than 50 results:
```
$ st2 rule list
+---------------------------------------+----------+---------------------------------------+---------+
| ref                                   | pack     | description                           | enabled |
+---------------------------------------+----------+---------------------------------------+---------+
 [...]
| examples.sample_rule_with_timer       | examples | Sample rule using an Interval Timer.  | False   |
[...]

PR #3514, opened 2017-06-23
Add "Note" in the CLI if API has default query "limit=50" and update rule endpoint help

humblearner:
Changes:

```
$ st2 rule list
+---------------------------------+--------+---------------------------------+---------+
| ref                             | pack   | description                     | enabled |
+---------------------------------+--------+---------------------------------+---------+
| github.deploy_pack_on_deploymen | github | Trigger an ST2 pack.install     | False   |
| t_event                         |        | when receiving a deployment     |         |
                ~~~~~~~~~~~~~~~~~~~~~~~
                ~~~~~~~~~~~~~~~~~~~~~~~
| st2cd.st2_pkg_test_stable_u16_e | st2cd  | Test stable packages            | True    |
| nterprise                       |        |                                 |         |
+---------------------------------+--------+---------------------------------+---------+
+-----------------------------------------------------------------------------------------+
| Note: Only first 50 results are displayed. Use -n/--last flag for more results, if any. |
+-----------------------------------------------------------------------------------------+
```

```
$ st2 rule list -h
usage: st2 rule list [-h] [-t TOKEN] [--api-key API_KEY] [-j] [-y]
                     [-a ATTR [ATTR ...]] [-w WIDTH [WIDTH ...]] [-n LAST]
                     [--iftt] [-p PACK] [-c ACTION] [-g TRIGGER]
                     [--enabled | --disabled]

Get the list of the 50 most recent rules.

optional arguments:
                ~~~~~~~~~~~~~~~~~~~~~~~
                ~~~~~~~~~~~~~~~~~~~~~~~
  -n LAST, --last LAST  List N most recent rules. Default is 50.
```

CLI Updated:
- [x] rule
- [x] execution
- [x] triggerinstance
- [x] trace
- [x] rule_enforcement

Functionality:
- Dynamically display note (in stderr) if  # of entries are greater  than limit (default or set by -n flag).
- Only in (default) table format, not in json or yaml.

Addresses: https://github.com/StackStorm/st2/issues/3488

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,11 @@ Added
   for password protected SSH key files. (improvement)
 
   Reported by Sibiraja L, Nick Maludy.
+* Append note to the output, if there are more entries for the following ``list`` CLIs: ``rule``,
+  ``execution``, ``rule-enforcment``, ``trace`` and ``trigger-instance`` as they are not displayed.
+  Default limit is 50. (improvement)
+
+  Reported by Eugen C. #3488
 
 Changed
 ~~~~~~~
@@ -176,7 +181,7 @@ Fixed
   on the action executions to ensure operations such as callback is performed. (bug fix)
 * Fix action chain runner workflows so variables (vars) and parameter values
   support non-ascii (unicode) characters. (bug fix)
-* Fix a bug in query base module when outstanding queries to mistral or other workflow engines 
+* Fix a bug in query base module when outstanding queries to mistral or other workflow engines
   could cause a tight loop without cooperative yield leading to 100% CPU usage by st2resultstracker
   process. (bug-fix)
 * Ignore unicode related encoding errors which could occur in some circumstances when
@@ -435,7 +440,7 @@ Changed
 ~~~~~~~
 
 * Improved pack validation - now when the packs are registered we check that:
-  
+
   - ``version`` attribute in the pack metadata file matches valid semver format (e.g
     ``0.1.0``, ``2.0.0``, etc.)
   - ``email`` attribute (if specified) contains a valid email address. (improvement)
@@ -500,7 +505,7 @@ Added
   of an intermediate network error or similar. (improvements)
 * Allow jinja expressions ``{{st2kv.system.foo}}`` and ``{{st2kv.user.foo}}`` to access
   datastore items from workflows, actions and rules. This is in addition to supporting
-  expressions ``{{system.foo}}`` and ``{{user.foo}}``. 
+  expressions ``{{system.foo}}`` and ``{{user.foo}}``.
 
 Changed
 ~~~~~~~
@@ -688,7 +693,7 @@ Added
 * Ability to migrate api keys to new installs. (new feature)
 * Introduce a new concept of pack config schemas. Each pack can now contain a
   ``config.schema.yaml`` file. This file can contain an optional schema for the pack config.
-  Site-specific pack configuration is then stored outside the pack directory, in 
+  Site-specific pack configuration is then stored outside the pack directory, in
   ``/opt/stackstorm/configs/<pack name>.yaml``. Those files are similar to the existing pack
   configs, but in addition to the static values they can also contain dynamic values. Dynamic value
   is a value which contains a Jinja expression which is resolved to a datastore item during
@@ -850,7 +855,7 @@ Deprecated
 * Mistral has deprecated the use of task name (i.e. ``$.task1``) to reference task result. It is
   replaced with a ``task`` function that returns attributes of the task such as id, state, result,
   and additional information (i.e. ``task(task1).result``).
- 
+
 Fixed
 ~~~~~
 
@@ -1349,7 +1354,7 @@ Changed
   date formats - previously it only worked with ISO8601 date strings. (improvement)
 * API server now gracefully shuts down on SIGINT (CTRL-C). (improvement)
 * Single sensor mode of Sensor Container uses ``--sensor-ref`` instead of ``--sensor-name``.
-* Move ``/exp/actionalias/`` and ``/exp/aliasexecution`` to ``/v1/actionalias/`` and 
+* Move ``/exp/actionalias/`` and ``/exp/aliasexecution`` to ``/v1/actionalias/`` and
   ``/v1/aliasexecution/`` respectively. (upgrade)
 * Display friendly message for error in parameters validation on action execution. (improvement)
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,8 +11,9 @@ Added
   for password protected SSH key files. (improvement)
 
   Reported by Sibiraja L, Nick Maludy.
-* Append note to the output, if there are more entries for the following ``list`` CLIs: ``rule``,
-  ``execution``, ``rule-enforcment``, ``trace`` and ``trigger-instance`` as they are not displayed.
+* Print a note to stderr if there are more entries / results on the server side which are displayed
+  to the user for the following ``list`` CLI commands: ``rule``, ``execution``,
+  ``rule-enforcment``, ``trace`` and ``trigger-instance``.
   Default limit is 50. (improvement)
 
   Reported by Eugen C. #3488
--- a/st2client/st2client/commands/action.py
+++ b/st2client/st2client/commands/action.py
@@ -1085,9 +1085,9 @@ class ActionExecutionListCommand(ActionExecutionReadCommand):
                               attributes=args.attr, widths=args.width,
                               attribute_transform_functions=self.attribute_transform_functions)
 
-            if args.last >= self.default_limit and int(count) > args.last:
-                table.SingleRowTable.note_box("Note: Only first %s results are displayed. Use -n/"
-                                              "--last flag for more results." % args.last)
+            if args.last >= self.default_limit and count and int(count) > args.last:
+                table.SingleRowTable.note_box("Note: Only first %s results are displayed. "
+                                              "Use -n/--last flag for more results." % args.last)
 
 
 class ActionExecutionGetCommand(ActionRunCommandMixin, ActionExecutionReadCommand):
--- a/st2client/st2client/commands/action.py
+++ b/st2client/st2client/commands/action.py
@@ -991,9 +991,10 @@ class ActionExecutionListCommand(ActionExecutionReadCommand):
             resource.get_plural_display_name().lower(),
             *args, **kwargs)
 
+        self.default_limit = 50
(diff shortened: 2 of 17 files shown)


Kami:
LGTM, but we should also do the change @armab suggested aka printing out "there are more results" when using CLI in user-friendly (table) mode (we don't want to print that when using --json or --yaml flag).

st2client/st2client/commands/rule.py
```diff
@@ -34,20 +34,26 @@ def __init__(self, description, app, subparsers, parent_parser=None):
...
 
         self.group = self.parser.add_argument_group()
         self.parser.add_argument('-n', '--last', type=int, dest='last',
                                  default=50,
-                                 help=('List N most recent %s.' %
+                                 help=('List N most recent %s. Default is 50.' %
                                        resource.get_plural_display_name().lower()))
         self.parser.add_argument('--iftt', action='store_true',
                                  help='Show trigger and action in display list.')
+        self.parser.add_argument('-p', '--pack', type=str,
```
Kami: Perhaps we should add changelog entry for this change.

st2client/st2client/commands/rule.py
```diff
@@ -34,20 +34,26 @@ def __init__(self, description, app, subparsers, parent_parser=None):
         self.commands['disable'] = RuleDisableCommand(self.resource, self.app, self.subparsers)
 
 
-class RuleListCommand(resource.ContentPackResourceListCommand):
+class RuleListCommand(resource.ResourceTableCommand):
```
Kami: Just curious why the inherit class change was needed.

armab:
True, without explicit note that only first 50 results are showed up, - it's not user friendly and doesn't solve the confusion.
As said in https://github.com/StackStorm/st2/issues/3488 to be consistent we need to check if all CLI commands have this feature as well.

humblearner:
Besides rules, the default limit=50 is set for: `rule_enforcement`, `execution`, `triggerinstance` and  `trace`. If above `Note` change is ok, then I'll add it to the above CLI as well.

st2client/st2client/commands/rule.py
humblearner: Yes, once all the changes are ok, I'll update the changelog.

Kami:
Yeah, that note works for me, but maybe we could also make it more dynamic and only display the note if there are in fact more results.

We could do that by utilizing X- headers returned by the API, bit it could also make it a bit more complex and require more refactoring, because headers and raw response objects are not available in those methods right now.

st2client/st2client/commands/trace.py
```diff
@@ -163,22 +164,31 @@ def run(self, args, **kwargs):
...
             # user has overriden the attributes to display
             if args.attr == self.display_attributes:
                 args.attr = ['all']
             self.print_trace_details(trace=instances[0], args=args)
         else:
-            self.print_output(reversed(instances), table.MultiColumnTable,
-                              attributes=args.attr, widths=args.width,
-                              json=args.json, yaml=args.yaml,
-                              attribute_transform_functions=self.attribute_transform_functions)
+            if args.json or args.yaml:
```
Kami: It looks like this is repeated in quite some places so eventually it will need some refactor, but it should be fine for now.

st2client/st2client/models/core.py
```diff
@@ -274,11 +274,14 @@ def query(self, **kwargs):
             self.handle_error(response)
         items = response.json()
         instances = [self.resource.deserialize(item) for item in items]
-        return instances
+        if 'X-Total-Count' in response.headers:
```
Kami: It looks like query is used in quite some places, were all the affected places updated?

st2client/st2client/commands/action.py
```diff
@@ -1062,20 +1063,31 @@ def run(self, args, **kwargs):
...
+            self.print_output(reversed(instances), table.MultiColumnTable,
+                              attributes=args.attr, widths=args.width,
+                              attribute_transform_functions=self.attribute_transform_functions)
 
-        self.print_output(reversed(instances), table.MultiColumnTable,
-                          attributes=args.attr, widths=args.width,
-                          json=args.json,
-                          yaml=args.yaml,
-                          attribute_transform_functions=self.attribute_transform_functions)
+            if args.last >= self.default_limit and int(count) > args.last:
```
Kami: It looks like `query` method returns None` if response doesn't contain X-Total header so this `int(count)` would fail.

So probably need to do something along the lines of `and count is not None and int(count)`...

st2client/st2client/models/core.py
humblearner: Yes, I have updated all the places, where query was called. However, unit tests are not happy, as we are mocking the response w/o headers.

st2client/st2client/formatters/table.py
```diff
@@ -248,3 +248,13 @@ def _get_attribute_value(subject, attribute):
...
+
+
+class SingleRowTable(object):
+
+    @staticmethod
+    def note_box(message):
+        note = PrettyTable([""])
+        note.header = False
+        note.add_row([message])
+        print note
```
Kami: Minor API consistency thing - all other formatter class methods return a value instead of printing it so it would be better for this one to do that as well.

Kami: And perhaps, as @armab said earlier, maybe even print this note to stderr instead of stdout.

st2client/st2client/commands/action.py
humblearner: Agree, however, is it safe to assume if `args.last` is present `X-Total-Count` will be present in the header?

Kami: I believe so, but would need to confirm.

And perhaps better to be on the safe side :)

st2client/st2client/formatters/table.py
```diff
@@ -248,3 +249,13 @@ def _get_attribute_value(subject, attribute):
...
+
+
+class SingleRowTable(object):
+
+    @staticmethod
+    def note_box(message):
+        note = PrettyTable([""])
+        note.header = False
+        note.add_row([message])
+        return sys.stderr.write(str(note))
```
Kami: Missing line break at the end.

It would also be good if you could align note box table with actual content table (so it's the same number of columns / width).

screenshot from 2017-06-29 11-43-54

st2client/st2client/commands/triggerinstance.py
```diff
@@ -117,14 +118,24 @@ def run(self, args, **kwargs):
...
+        if args.json or args.yaml:
+            self.print_output(reversed(instances), table.MultiColumnTable,
+                              attributes=args.attr, widths=args.width,
+                              json=args.json, yaml=args.yaml,
+                              attribute_transform_functions=self.attribute_transform_functions)
+        else:
+            self.print_output(reversed(instances), table.MultiColumnTable,
+                              attributes=args.attr, widths=args.width,
+                              attribute_transform_functions=self.attribute_transform_functions)
+            if args.last >= self.default_limit and count and int(count) > args.last:
```
Kami: It looks like it doesn't correctly handle a scenario where you pass in `-n 1` for example but there are more results.

So this logic needs to be changed to handle that scenario as well.

humblearner: Hmm, it is working for me. If n > default limit then show the note, otherwise, just show the result.
```
$ st2 rule-enforcement list -n 1
+--------------------------+----------------------+----------------------+----------------------+----------------------+
| id                       | rule.ref             | trigger_instance_id  | execution_id         | enforced_at          |
+--------------------------+----------------------+----------------------+----------------------+----------------------+
| 5952ab31909a5027b65c037b | st2ci.st2_pkg_test_a | 5952ab30909a5027b65c | 5952ab30909a5027b65c | 2017-06-27T19:00:00. |
|                          | nd_promote_unstable_ | 036e                 | 037a                 | 858196Z              |
|                          | u16                  |                      |                      |                      |
+--------------------------+----------------------+----------------------+----------------------+----------------------+

$ st2 execution list -n 1
+--------------------------+------------+--------------+------------------------+------------------+---------------+
| id                       | action.ref | context.user | status                 | start_timestamp  | end_timestamp |
+--------------------------+------------+--------------+------------------------+------------------+---------------+
| 595467b3909a5015dab1c8b9 | core.local | stanley      | succeeded (1s elapsed) | Thu, 29 Jun 2017 | Thu, 29 Jun   |
|                          |            |              |                        | 02:36:35 UTC     | 2017 02:36:36 |
|                          |            |              |                        |                  | UTC           |
+--------------------------+------------+--------------+------------------------+------------------+---------------+

$ st2 trigger-instance list -n 1
+--------------------------+---------------------------+---------------------------+-----------+
| id                       | trigger                   | occurrence_time           | status    |
+--------------------------+---------------------------+---------------------------+-----------+
| 59546ab5909a5045fc0d0882 | core.st2.key_value_pair.c | Thu, 29 Jun 2017 02:49:25 | processed |
|                          | reate                     | UTC                       |           |
+--------------------------+---------------------------+---------------------------+-----------+
```

CHANGELOG.rst
```diff
@@ -11,6 +11,11 @@ Added
   for password protected SSH key files. (improvement)
 
   Reported by Sibiraja L, Nick Maludy.
+* Append note to the output, if there are more entries for the following ``list`` CLIs: ``rule``,
+  ``execution``, ``rule-enforcment``, ``trace`` and ``trigger-instance`` as they are not displayed.
+  Default limit is 50. (improvement)
+
+  Reported by Eugen C. #3488
```
armab: Just `#3514 #3488` is enough there.

I think we include "reported by" if there was an external contribution.

st2client/st2client/formatters/table.py
```diff
@@ -143,11 +146,16 @@ def format(cls, entries, *args, **kwargs):
...
                     transform_function = attribute_transform_functions.get(field_name,
-                                                                          lambda value: value)
+                                                                           lambda value: value)
                     value = transform_function(value=value)
                     value = strutil.unescape(value)
                     values.append(value)
             table.add_row(values)
+
+        # width for the note
+        global table_width
```
Kami: I will test it and go ahead and merge it if everything look OK so it will end up in v2.3.1, but we really need to refactor and fix this global variable thing since it's not robust and could cause issues.

Kami:
I'll go ahead and merge this so we can start testing changes for v2.3.1 release, but please look into getting rid of the global variable as soon as you get a chance.

Kami:
@humblearner It would be great if you can also work on automated robot tests for this CLI functionality - so testing all the common scenarios + edge cases.

CLI is one of the places where we lack tests and regressions are easy to introduce so we should invest some time in more automated tests.

PR #3524, opened 2017-06-26
Datastore Service - Get Token TTL from configured max

### LINKED ISSUE #3523: Issue with TTL
Our StackStorm config had `token_ttl = 14400`, but we got this error below:
```    
return self.datastore_service.get_value(name, local, scope=scope, decrypt=decrypt)
  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/services/datastore.py\", line 90, in get_value
    client = self._get_api_client()
  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/services/datastore.py\", line 216, in _get_api_client
    temporary_token = create_token(username=self._api_username, ttl=ttl)
  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/services/access.py\", line 58, in create_token
    raise TTLTooLargeException(msg)
st2common.exceptions.auth.TTLTooLargeException: TTL specified 86400 is greater than max allowed 14400.
```
It looks like the TTL is hardcoded in https://github.com/StackStorm/st2/blob/master/st2common/st2common/services/datastore.py#L214

bigmstone:
Fixes #3523 

Datastore should get its TTL from the configured max instead of hard-coded value. Without this change if `token_ttl` is set to a value `<` the default any python action that access the data service will throw an error.

### DIFF
--- a/st2common/st2common/config.py
+++ b/st2common/st2common/config.py
@@ -210,8 +210,10 @@ def register_opts(ignore_errors=False):
         cfg.StrOpt('api_url', default=None,
                    help='Base URL to the API endpoint excluding the version'),
         cfg.BoolOpt('enable', default=True, help='Enable authentication middleware.'),
-        cfg.IntOpt('token_ttl', default=86400, help='Access token ttl in seconds.'),
-        cfg.IntOpt('service_token_ttl', default=86400, help='Service token ttl in seconds.')
+        cfg.IntOpt('token_ttl', default=(24 * 60 * 60), help='Access token ttl in seconds.'),
+        # This TTL is used for tokens which belong to StackStorm services
+        cfg.IntOpt('service_token_ttl', default=(24 * 60 * 60),
+                   help='Service token ttl in seconds.')
     ]
     do_register_opts(auth_opts, 'auth', ignore_errors)
 
--- a/st2common/st2common/config.py
+++ b/st2common/st2common/config.py
@@ -210,7 +210,8 @@ def register_opts(ignore_errors=False):
         cfg.StrOpt('api_url', default=None,
                    help='Base URL to the API endpoint excluding the version'),
         cfg.BoolOpt('enable', default=True, help='Enable authentication middleware.'),
-        cfg.IntOpt('token_ttl', default=86400, help='Access token ttl in seconds.')
+        cfg.IntOpt('token_ttl', default=86400, help='Access token ttl in seconds.'),
+        cfg.IntOpt('service_token_ttl', default=86400, help='Service token ttl in seconds.')
     ]
     do_register_opts(auth_opts, 'auth', ignore_errors)
 
--- a/st2common/st2common/services/datastore.py
+++ b/st2common/st2common/services/datastore.py
@@ -214,9 +214,9 @@ class DatastoreService(object):
 
         if not self._client or token_expire:
             self._logger.audit('Creating new Client object.')
-            ttl = cfg.CONF.auth.token_ttl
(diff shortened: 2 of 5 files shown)


bigmstone:
I don't have an immediate solution as to how the auth service would validate if the request is coming from a "service" or not - and therefore know to use a different MAX TTL. It also seems to break what a MAX TTL setting would be.

Perhaps if we used the concept of a `client` similar to oAuth this could be achieved, but the added complexity isn't required for what our auth service does and this use-case doesn't warrant prioritizing that feature.

I think setting the service TTL to the max available TTL possible is a reasonable solution. Otherwise we'll have to touch the auth service and come up with some scheme for denoting when to use the `service_token_ttl`. As max ttl is usually configured for security reasons we'd need to have a reasonably valid way of trusting a service is a service (`complexity: medium`).

Thoughts?

Kami:
With service I really just meant every token we generate outside st2auth API endpoint for service purpose (e.g. the one in the datastore used by actions and sensors, temporary token we generate for actions, etc.).

st2common/st2common/services/datastore.py
```diff
@@ -211,7 +214,7 @@ def _get_api_client(self):
 
         if not self._client or token_expire:
             self._logger.audit('Creating new Client object.')
-            ttl = (24 * 60 * 60)
+            ttl = cfg.CONF.auth.token_ttl
             self._token_expire = get_datetime_utc_now() + timedelta(seconds=ttl)
             temporary_token = create_token(username=self._api_username, ttl=ttl)
```
Kami: Just adding `service=True` argument to `create_token` function call should do the trick. Then we can revert the TTL change.

Recently I made a change so we support arbitrary TTLs for service tokens - https://github.com/StackStorm/st2/pull/3315

st2common/st2common/services/datastore.py
```diff
@@ -211,9 +214,9 @@ def _get_api_client(self):
 
         if not self._client or token_expire:
             self._logger.audit('Creating new Client object.')
-            ttl = (24 * 60 * 60)
+            ttl = cfg.CONF.auth.service_token_ttl
             self._token_expire = get_datetime_utc_now() + timedelta(seconds=ttl)
-            temporary_token = create_token(username=self._api_username, ttl=ttl)
+            temporary_token = create_token(username=self._api_username, ttl=ttl, service=True)
```
Kami: Passing additional metadata aka the name of the service (action, sensor) where this token is used would be great, but this would require small refactoring - adding additional argument to `DatastoreService` constructor and updating places where it's used (should only be two places - sensor wrapper and python action wrapper).

st2common/st2common/config.py
```diff
@@ -210,7 +210,8 @@ def register_opts(ignore_errors=False):
         cfg.StrOpt('api_url', default=None,
                    help='Base URL to the API endpoint excluding the version'),
         cfg.BoolOpt('enable', default=True, help='Enable authentication middleware.'),
-        cfg.IntOpt('token_ttl', default=86400, help='Access token ttl in seconds.')
+        cfg.IntOpt('token_ttl', default=86400, help='Access token ttl in seconds.'),
+        cfg.IntOpt('service_token_ttl', default=86400, help='Service token ttl in seconds.')
```
Kami: Would be good to add a comment which clarifies a bit what service token is and where it's used for (basically what we debated in this PR :D)

Kami:
I finished this one so it can be included in v2.3.1.

Will go ahead and merge it once checks pass.

PR #3525, opened 2017-06-28
Add missing passpharse argument to the remote script runner

Kami:
It looks like remote shell script runner was missing `passphrase` parameter and as such, password / passphrase supported private SSH key files were not supported (remote command runner supports it just fine).


Kami:
@armab Yeah, that file is auto-generated so can only open st2docs once this PR is merged :)

PR #3531, opened 2017-06-29
Update config loader to correctly handle default values which are falsey (None, False, {}, etc.)

Kami:
This pull request works on top of #3504.

I decided to start with the test approach (failing test) and work from there to be sure we don't introduce any regressions.

Instead of using `setdefault` which caused a bunch of other issues with tests I decided to go with a "magic" value which is very unlikely to appear in the config as a default value or otherwise.

### DIFF
--- a/st2common/st2common/util/config_loader.py
+++ b/st2common/st2common/util/config_loader.py
@@ -35,8 +35,6 @@ __all__ = [
 
 LOG = logging.getLogger(__name__)
 
-DEFAULT_VALUE_MAGIC_MARKER = '~~***---%%--%%--%--%%%***~~~+++'
-
 
 class ContentPackConfigLoader(object):
     """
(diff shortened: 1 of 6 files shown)


st2common/st2common/util/config_loader.py
```diff
@@ -35,6 +35,8 @@
 
 LOG = logging.getLogger(__name__)
 
+DEFAULT_VALUE_MAGIC_MARKER = '~~***---%%--%%--%--%%%***~~~+++'
```
armab: Whoa, what's this magic?

armab: I'm starting getting it :spades: 
Also a small `#` code comment would be useful to clarify.

Kami: I was actually just working to use a more robust approach - ef8b87a5666edca38d928e242cd20f959338f21d.

Will push more changes soon.

PR #3537, opened 2017-06-30
"Inquirer" Runner

Mierdin:
# TODO

- [ ] Changelog
- [ ] Required fields for trigger payload schema?
- [ ] update the Mistral querier and Action Chain runner to   process `pending` state correctly. Right now, they don't understand what `pending` means. Also make `pending` an inactive state - right now the parent workflow is set to "pausing" because `pending` is an active state, so the workflow thinks its still doing work.
- [ ] [update the way parent execution is determined (get true parent)](https://github.com/StackStorm/st2/pull/3537#issuecomment-322332248)

This PR creates the `inquirer` runner, which forms the basis for "asking a question" in the middle of a Workflow.

Based on previous and current (see below) design discussions, this runner is fairly simple:
- Ensure this liveaction is running within a workflow (has a parent)
- Dispatch trigger indicating a new inquiry
- Request that the parent execution (workflow) is paused
- Return a `pending` status

The entire act of handling a response (including validation) and resuming the workflow is handled by the API, and is outside the scope of this runner. This runner's sole purpose is to pause the workflow and provide enough context for a 3rd party to make a proper response.

# Usage

> The runner/action will return `pending` but until https://github.com/StackStorm/st2/pull/3507 and upstream Mistral changes are merged, the workflow will continue to run

```
(virtualenv) vagrant@st2dev:~/st2$ st2 run examples.mistral-ask-basic
......^C
(virtualenv) vagrant@st2dev:~/st2$ sel
+----------------------------+----------------------------+--------------+--------------------------+-------------------------------+---------------+
| id                         | action.ref                 | context.user | status                   | start_timestamp               | end_timestamp |
+----------------------------+----------------------------+--------------+--------------------------+-------------------------------+---------------+
| + 5983d94332ed351244873fac | examples.mistral-ask-basic | stanley      | running (75613s elapsed) | Fri, 04 Aug 2017 02:17:39 UTC |               |
| + 5985001e32ed3541f991fc67 | examples.mistral-ask-basic | stanley      | running (130s elapsed)   | Fri, 04 Aug 2017 23:15:42 UTC |               |
| + 5985008732ed3541f991fc6c | examples.mistral-ask-basic | stanley      | running (25s elapsed)    | Fri, 04 Aug 2017 23:17:27 UTC |               |
+----------------------------+----------------------------+--------------+--------------------------+-------------------------------+---------------+
(virtualenv) vagrant@st2dev:~/st2$ seg 5985008732ed3541f991fc6c
id: 5985008732ed3541f991fc6c
action.ref: examples.mistral-ask-basic
parameters: None
status: running (30s elapsed)
start_timestamp: 2017-08-04T23:17:27.017198Z
end_timestamp: None
+--------------------------+---------+-------+----------+-------------------------------+
| id                       | status  | task  | action   | start_timestamp               |
+--------------------------+---------+-------+----------+-------------------------------+
| 5985008732ed3541f991fc6f | pending | task1 | core.ask | Fri, 04 Aug 2017 23:17:27 UTC |
+--------------------------+---------+-------+----------+-------------------------------+
```


contrib/runners/ask_runner/ask_runner.py
```diff
@@ -0,0 +1,71 @@
...
+
+from st2common import log as logging
+from st2common.runners.base import ActionRunner
+from st2common.runners import python_action_wrapper
+
+LOG = logging.getLogger(__name__)
+
+__all__ = [
+    'get_runner',
+    'AskRunner',
```
m4dcoder: I prefer to name this something like Inquirer.

contrib/runners/ask_runner/ask_runner.py
```diff
@@ -0,0 +1,71 @@
...
+    def pre_run(self):
+        super(AskRunner, self).pre_run()
+
+        # TODO :This is awful, but the way "runner_parameters" and other variables get
+        # assigned on the runner instance is even worse. Those arguments should
+        # be passed to the constructor.
+        self._schema = self.runner_parameters.get(RUNNER_SCHEMA, self._schema)
+
+    def run(self, action_parameters):
+
```
m4dcoder: The sequence below isn't complete. Please review the design discussion. Let's start from when the st2.ask action is initially run and sequence from there.  Also we already discussed that there's no need to use the execution context to store the response.

Mierdin:
Previously I had a note in here:

> Should this runner also be written to handle an "st2.respond" action in addition to "st2.ask"?

I don't think this is a good idea - and tbh I can't remember what made me have this thought originally. This can/should be done by the API, which will take care of providing response data and re-running the execution (this runner)

Mierdin:
Okay, I **believe** this is a more complete example. A few notes:

- There were parts of this I just found easier to define in Python, so I just knocked that part out (i.e. the `has_permission` function.
- Because I started in on the implementation for a few things, I know you wanted to see a plain-english explanation of the steps, so I moved this into the docstring for `run`
- I am still fuzzy on how you're proposing we use results in lieu of context for storing response data. `self.context` is easily addressable, so I am using that for now. I am of course open to alternatives, but I think I'm missing something to fully understand what you're getting at (also @lakshmi-kannan shared some concerns in https://github.com/StackStorm/discussions/issues/236#issuecomment-303497051 on this, and I don't know enough about what you're suggesting to have an opinion either way).

m4dcoder:
- When this st2.ask action is run, how will the runner post the trigger instances for the rules?
- Also, I'm adamant about not allowing API to update execution context. Just post the answer from st2.response to the result of the action execution. The result is an attribute of the action execution, easily accessible, no different than the context attribute. Take a look at https://github.com/StackStorm/st2/blob/master/st2common/st2common/models/db/execution.py#L60 and https://github.com/StackStorm/st2/blob/master/st2common/st2common/models/db/execution.py#L63.

Mierdin:
> When this st2.ask action is run, how will the runner post the trigger instances for the rules?

My understanding is that this is outside the scope of the runner, since `core.st2.generic.actiontrigger` will produce instances whenever any action changes status. Let me know if I have something wrong there.

> The result is an attribute of the action execution, easily accessible, no different than the context attribute.

Interesting, didn't know that. I was just looking at the [runner base class](https://github.com/StackStorm/st2/blob/master/st2common/st2common/runners/base.py#L86) and only saw `context` there. I'll review the code you pointed to and see what I can come up with.

Mierdin:
Kind of a "duh" moment - if we're using result to store the information, the runner just needs to return the response data, right? Then the code in `st2actions` will persist it. Wondering if I need to do any more than that.

contrib/runners/inquirer/inquirer.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def run(self, action_parameters):
+
+        # Retrieve existing result (initialize if needed)
+        liveaction_db = action_utils.get_liveaction_by_id(self.liveaction_id)
+        response_data = liveaction_db.result.get("response_data")
+        if not response_data:
+            response_data = {"response": {}}
+
+        # Request pause of parent execution
```
m4dcoder: Requesting the parent workflow to pause should be last. We want to make sure the core logic of st2.ask (dispatch trigger, etc.) is successfully executed.

contrib/runners/inquirer/inquirer.py
```diff
@@ -0,0 +1,100 @@
...
+        parent = liveaction_db.context.get("parent")
+        if parent:
+            # TODO get current user
+            action_service.request_pause(parent, "st2admin")
+        else:
+            LOG.error("Inquiries must be run within a workflow.")
+            return (LIVEACTION_STATUS_FAILED, response_data, None)
+
+        # Validate repsonse and return
+        if action_service.validate_response(self.schema, response_data):
```
m4dcoder: I thought we are validating the response in the inquiry API endpoint?

contrib/runners/inquirer/inquirer.py
```diff
@@ -0,0 +1,100 @@
...
+            return (LIVEACTION_STATUS_SUCCEEDED, response_data, None)
+        else:
+            trigger_ref = ResourceReference.to_string_reference(
+                pack=INQUIRY_TRIGGER['pack'],
+                name=INQUIRY_TRIGGER['name']
+            )
+            trigger_payload = {
+                "response": response_data
```
m4dcoder: Why is the response data to trigger payload?

contrib/runners/inquirer/inquirer.py
Mierdin: We are doing it in both places, thus offloading to action_service (that way both can use the same code). The API will use this to get quick responses back to the user, but initial validation will not come from the API, so still needs to be done in the runner at some point.

contrib/runners/inquirer/inquirer.py
Mierdin: Ack

contrib/runners/inquirer/inquirer.py
Mierdin: If we're going to fire a trigger indicating that an inquiry has entered `pending` state, I figure it would be worth having the current response body in the payload, at the very least for debugging purposes, and potentially for any 3rd party dashboards that might consume it

m4dcoder: Is response the only data going into the payload? How will the rule or the triggered notification action know what to do with this payload?

contrib/runners/inquirer/inquirer.py
m4dcoder: I prefer to focus on the essential code path at this point. So if we think about the sequences of steps that take place when st2.ask is run, on the 1st pass, there's always no result, trigger is dispatched, and this liveaction is put into pending state. We will be using the API endpoint to post and validate response.  So, how will this liveaction be put back into running state for a 2nd pass thru this run method for this validate to be relevant?

Mierdin: The API endpoint for providing a response will be responsible for re-running this liveaction once the validation checks out. When we talked previously, we decided that shared validation code should be used so the runner could also perform this validation, since it wouldn't know if it was the first or second time it was being run.

This particular validation call actually *wouldn't* be relevant on the second pass-through, it would just be a repeat of the validation that already took place when the API called this function.

contrib/runners/inquirer/inquirer.py
Mierdin: Ah, you're talking about the other params like roles, users, etc. I will add those in, I took those for granted when we were using the other trigger type.

contrib/runners/inquirer/inquirer.py
m4dcoder: Yep, when response is posted to the API endpoint, the endpoint should validate the response first.  Since it's a specialized endpoint, and if the response passes validation, why can't we mark the action execution complete and save the result there from the endpoint? Why pass it back to the runner to be re-run?  Passing it back to the inquiry runner to re-run is fine, it'll work but it'll introduce inefficiency and a 2nd unnecessary validation.  So i'm trying to understand what's the value.

And say we do pass it back to the inquiry runner for re-run, where's the logic to resume the workflow?

I'm not concern that the validation code is shared.  We should because I think action result should be defined by schema and validated at some point and so putting the validation in some shared module is fine with me.

Mierdin: I think I see what you're getting at. Let me summarize and you tell me where/if I'm wrong:

Inquiry runner only every runs once, and therefore doesn't need to do validation, since it will always just be the first time, and we can just return pending right away. API will continue to validate response, but instead of re-running the liveaction for the inquiry runner, it will just update its status to successful, and request that the workflow continue with the next task

If that's right, my only question will be, will Mistral/AC honor the `successful` status the API puts into place and go to the *next* task, or will it try to re-run the inquiry?

contrib/runners/inquirer/inquirer.py
Mierdin: Done in a6df0de17 - let me know if you see any fields I missed, but I *think* these should provide enough for notification purposes.

contrib/runners/inquirer/inquirer.py
m4dcoder: When the inquiry POST endpoint mark the st2.ask action successful, if done right, should trigger callback to the Mistral workflow to update the result for the task.  I'll let you identify this as an exercise.  And then we'll have to explicitly resume the Mistral/AC workflow using `action_service.request_resume` in the inquiry POST endpoint.

contrib/runners/inquirer/inquirer.py
```diff
@@ -0,0 +1,104 @@
...
+                name=INQUIRY_TRIGGER['name']
+            )
+            trigger_payload = {
+                "liveaction_id": self.liveaction_id,
+                "response": response_data,
+                "schema": self.schema,
+                "roles": self.roles_param,
+                "users": self.users_param,
+                "tag": self.tag
+            }
```
m4dcoder: I don't see payload schema in the trigger ref defined.

contrib/runners/inquirer/inquirer.py
Mierdin: Sounds good, I think I didn't realize we could/should overwrite the status for a liveaction, as opposed to relying on the runner to return the status. I'll modify the runner under the assumption that it will ONLY be run once - the heavy lifting will be moved to the POST API endpoint.

m4dcoder: You have to use the action service to update/publish status and result so things get triggered properly.  But it's a specialized endpoint, I'm ok with it doing more work then a typical endpoint.  But then again, I'm also ok with it being passed back to the inquiry runner.  But in the latter, you'll have to change the status from pending to running anyway.  So what's the point?  Why not just change it to pending to success?

Mierdin: No, you're right - I was just operating under a false assumption that re-running was the only (or at least, best) way to update the status.

contrib/runners/inquirer/inquirer.py
Mierdin: Okay, re-ordered in adc00747d

contrib/runners/inquirer/inquirer.py
Mierdin: Ah, good catch. I would have caught it if I hadn't fixed that enforce-by-default bug a few releases back :)

Mierdin: Fixed in 11977ee5b, but before I did, I tried to get validation to break by adding `system.validate_trigger_parameters` to True in the config and wasn't able to. Probably beyond the scope of this PR but might be worth looking into - maybe I'm missing something.

contrib/runners/inquirer/inquirer.py
Mierdin: Okay, validation call removed in de7d303f2. The assumption now is that the runner will only run once (for a given reference, of course). The API will drive everything after that.

Mierdin:
@m4dcoder Okay, I believe I've addressed your comments thus far, and added appropriate tests. The runner is fairly simple now that a lot of the work is being moved to the API (which I'll be starting in on now).

CircleCI is having some kind of issue with RabbitMQ, and Pylint is throwing a fit now because I removed the stub `request_pause` function (and obviously the reference is still there) but I suspect we weren't planning to merge this until the pause functionality is there anyways, so not too worried about it.

I am also not 100% sure if I've added **all** appropriate statuses to constants in `st2common.constants.action.py` but I am sure that I will discover if I didn't when I make API changes. :) Of course if there's anything else I can change, let me know. In the meantime, I'll be turning my attention towards the API.

contrib/runners/inquirer/tests/unit/test_inquirer_runner.py
```diff
@@ -0,0 +1,116 @@
...
+        )
+        mock_action_service.request_pause.assert_called_once_with(
+            test_parent_id,
+            test_user
+        )
+
+    @mock.patch('inquirer.TriggerDispatcher', mock_trigger_dispatcher)
+    @mock.patch('inquirer.action_utils', mock_action_utils)
+    @mock.patch('inquirer.action_service', mock_action_service)
+    def test_inquiry_failed_no_parent(self):
```
m4dcoder: We discussed this in our F2F. For testing purposes, an st2.ask action can be run individually and so it's possible to not have a parent (under a workflow).

st2common/st2common/constants/triggers.py
```diff
@@ -100,6 +101,23 @@
...
 }
 
+INQUIRY_TRIGGER = {
+    'name': 'st2.generic.inquiry',
+    'pack': SYSTEM_PACK_NAME,
+    'description': 'Trigger indicating a new "inquiry" has entered "pending" status',
+    'payload_schema': {
+        'type': 'object',
+        'properties': {
+            'liveaction_id': {},
```
m4dcoder: Are you going to define the types for these properties?

Mierdin: Yeah, I can do that.

contrib/runners/inquirer/tests/unit/test_inquirer_runner.py
Mierdin: Hrm, correct me if I'm remembering it incorrectly, but I thought we decided that while it was **possible** to run these actions standalone, we should throw an error in that case. To facilitate testing, maybe we should have an optional parameter you can pass that causes the runner to ignore this check? WDYT?

m4dcoder: Nah, I don't think it's necessary to add an optional parameter. It's like any other action. We can execute it individually or under a workflow.

Mierdin: Okay - sorry for the misunderstanding, will remove this and will allow it to be run individually

Mierdin: Done in https://github.com/StackStorm/st2/pull/3537/commits/9dc3868c8cc306eeb100bb629c74da43775b43a3 and modified unit tests to accept a `pending` status, and that the trigger still fires, but that the request for pause still doesn't.

st2common/st2common/constants/triggers.py
Mierdin: Added more detail in c1c93da30. Also using `id` instead of `liveaction_id` to keep confusion down if we ever create a proper data model for inquiries

m4dcoder:
Request:
You'll have to make some changes to the workflow pause/resume.

Background:
If a workflow/chain is paused, it will cascade the pause both upward (to parent workflow/chain) and downward (to subworkflows/subchains). However on resume, it will only cascade downward. There's a explanation for that but requires a longer conversation which I'm happy to discuss if needed.

What you need to do:
For st2.ask, you will need to identify the root parent and then request the pause/resume on the root parent.

Mierdin:
@m4dcoder Ack that, will do. I'll modify the runner to check to see if the currently identified parent ID has its own parent, and continue until I get the root.

What should the limitation be there? I *think* I remember hearing that we have established limitations on how much nesting is permitted in Workflows (this may be different depending on WF type) but can't remember off the top of my head

Mierdin:
It's becoming a pain to manage two open PRs. The intention for opening this PR first was to 1) give folks visibility into how I was building this (since it was my first runner, really) and also 2) I had hoped to merge this before moving on with the rest of the work. However, as I discovered, it's really hard to vet the work in this PR without also having the API functionality for getting and listing outstanding Inquiries as well as responding to them.

I recently rebased the branch in https://github.com/StackStorm/st2/pull/3653, (the PR that introduces all of the aforementioned API functionality) to include the commits from this PR, for this reason - so I could work on all of this all at once and see it work end-to-end. I realize that PR is getting pretty huge as a result, but I think it's a bit easier on me and on reviewers to not have to deal with two moving targets.

So, I'm closing this PR and designating https://github.com/StackStorm/st2/pull/3653 as the sole PR for introducing Inquiry functionality, since it already includes this work and more anyways. I'll of course honor all comments and TODOs from this PR.

Mierdin:
@m4dcoder FYI I implemented a function for finding the root parent as requested, over in the other PR: https://github.com/StackStorm/st2/pull/3653/commits/c5cd023fa439f6d0ec3ac11ffe472db945e9bdcf

PR #3538, opened 2017-06-30
Upgrade various dependencies (v2.4.0)

Kami:
Just some house keeping and another attempt after https://github.com/StackStorm/st2/pull/3189 didn't make it.

It updates the following dependencies to the latest stable version.

## General dependencies

* paramiko
* appscheduler
* pyyaml
* semver
* retrying
* kombu
* amqp
* networkx
* jsonschema
* sseclient
* jsonpath-rw
* requests
* git-python
* stevedore
* passlib
* webob
* tooz
* routes
* flex
* prance
* greenlet
* ~eventlet~
* gunicorn

The plan is to include those changes in v2.4.0 after some more testing is done.

I did go over all the changelog entries and now I'm looking into unit tests to make sure that all of them pass (there are some changes needed, but mostly because some libraries doesn't expose good enough APIs so we need to write some less than ideal and robust code to make things happen - e.g. how we throw more user-friendly exceptions for some jsonschema validation failures).

## TODO

- [x] Research issue related to mongoengine upgrade - https://github.com/StackStorm/st2/pull/3538#discussion_r125010183


fixed-requirements.txt
```diff
@@ -1,40 +1,40 @@
...
 pyyaml>=3.12,<4.0
-requests[security]>=2.14.1,<2.15
+requests[security]>=2.18.1,<2.19
 apscheduler==3.3.1
-gitpython==2.1.3
-jsonschema>=2.5.0,<2.6
-mongoengine==0.11.0
+gitpython==2.1.5
+jsonschema==2.6.0
+mongoengine==0.13.0
```
armab: This is the same change which we reverted before: https://github.com/StackStorm/st2/pull/3402
It caused the regression for Xenial when 'st2ctl register' failed to register rules on a fresh DB. 

If you could investigate and fix the root cause of https://github.com/StackStorm/st2-packages/issues/445 without adding timeouts, - I'm good with that.
Otherwise let's please avoid the same  `mongoengine` update.

Kami: Yeah, plan is to merge it once more testing is done and that issue is resolved.

Kami:
Sadly we still can't upgrade to the latest version of eventlet.

v0.20.0 removes `select.poll()` which affects us. I managed to fix some of the related issue, but it still breaks `packs` pack tests which use `gitpython` so it needs more work to make it work. To clarify, it only breaks tests because we don't do monkey patch there, but the actual python runner actions work fine.

fixed-requirements.txt
Kami: I believe this work-around should do it - 60a9dde458ed472799df0205098312ba0196e948.

As mentioned in that ticket and discussed on Slack, it seems like there is some weird race and internal trigger types are not registered yet when the script runs.

Registering internal triggers is an idempotent operation we do at every service setup / init phase so doing it again here should have no negative consequences (perhaps just a small performance overhead, but should be fine).

Kami: Also, as we just discussed offline, you are correct, this script shouldn't rely on other services to start first and with this change this is now indeed the case.

In the past we didn't do that, because registering internal trigger types talks to the database and we thought register-content will also be able to run on other servers without st2 and talk to st2 over API, but that's not the case and other functionality of the script also requires to be run from the server where st2 components are running.

And if we ever move it to the API, that's fine as well because then script won't require DB access  anymore.

Kami: We believe #3542 should resolve the actual underlying race issue so we should be good, but we will still of course do more testing.

armab: 👍

nmaludy:
+1 interested in this based on a [conversation in slack](https://stackstorm-community.slack.com/archives/C066APT88/p1499871773777335) about HA and coordination backends. It appears there is a bug in the current version of tooz that prevents the postgres driver from connecting properly.

Kami:
Upgrading eventlet is quite challenging because of removal of `select.poll()` (we can't control 3rd party libraries which we use and rely on this functionality) so I will go ahead, merge this and then working on upgrading eventlet in a separate PR / branch.

Kami:
@nmaludy also updated tooz to latest version, those changes should be included in v2.4.0 (should be out some time in the next couple of months).

If it's a small bug fix, we can potentially also pick that change in v2.3.2.

armab:
Merging this update broke our builds on `EL6` & `EL7` with the following error:

```
[root@st2-ent-pkg-unstable-el7 stanley]# st2 run core.http url=https://google.com/
.
id: 5971d6acd549c538fa717614
status: failed
parameters: 
  url: https://www.google.com
result: 
  error: '("bad handshake: Error([(''SSL routines'', ''SSL3_GET_SERVER_CERTIFICATE'', ''certificate verify failed'')],)",)'
  traceback: "  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/st2actions/container/base.py\", line 105, in _do_run\n    (status, result, context) = runner.run(action_params)\n  File \"/opt/stackstorm/runners/http_runner/http_runner.py\", line 93, in run\n    result = client.run()\n  File \"/opt/stackstorm/runners/http_runner/http_runner.py\", line 217, in run\n    verify=self.verify\n  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/requests/api.py\", line 58, in request\n    return session.request(method=method, url=url, **kwargs)\n  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/requests/sessions.py\", line 502, in request\n    resp = self.send(prep, **send_kwargs)\n  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/requests/sessions.py\", line 612, in send\n    r = adapter.send(request, **kwargs)\n  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/requests/adapters.py\", line 514, in send\n    raise SSLError(e, request=request)\n"
```

After reverting back to revision before this upgrade everything is OK:
```
[root@st2-ent-pkg-unstable-el7 stanley]# st2 run core.http url=https://google.com/
.
id: 5971d94ed549c543310da786
status: failed
parameters: 
  url: https://google.com/
result: 
  body: "<HTML><HEAD><meta http-equiv=\"content-type\" content=\"text/html;charset=utf-8\">\n<TITLE>301 Moved</TITLE></HEAD><BODY>\n<H1>301 Moved</H1>\nThe document has moved\n<A HREF=\"https://www.google.com/\">here</A>.\r\n</BODY></HTML>\r\n"
  headers:
    Alt-Svc: quic=":443"; ma=2592000; v="39,38,37,36,35"
    Cache-Control: public, max-age=2592000
    Content-Length: '220'
    Content-Type: text/html; charset=UTF-8
    Date: Fri, 21 Jul 2017 10:37:02 GMT
    Expires: Sun, 20 Aug 2017 10:37:02 GMT
    Location: https://www.google.com/
    Server: gws
    X-Frame-Options: SAMEORIGIN
    X-XSS-Protection: 1; mode=block
  parsed: false
  status_code: 301
```

st2client/st2client/base.py
```diff
@@ -23,6 +23,9 @@
 
 import six
 import requests
+import urllib3
```
armab: @Kami If we revert the requests back to previous version in https://github.com/StackStorm/st2/pull/3597, this part probably should be reverted as well.

Issue #3534, opened 2017-06-29
st2 pack install fails unless git repo has .git suffix (SSH)


carbineneutral:
Our private git repository has an SSH URL that does not end in ".git"

When we attempt to install a pack it fails because the git command appends a ".git" to the URL, which is not valid for our repo.

If we artificially create a new repo with ".git" in it's URL, it works as expected.

If we manually execute the git command from the cmdline excerpt (below), but remove the ".git" suffix, the clone works as expected.

st2 2.3.0
Linux st2-local 4.4.0-81-generic #104-Ubuntu SMP Wed Jun 14 08:17:06 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

Trace included (with obfuscations).

```
ubuntu@st2-local:~$ st2 pack install ssh://<user@host>/AutomationStackStorm

        [       failed       ] download pack

id: 595525a6a5fb1d0755645a1e
action.ref: packs.install
parameters:
  packs:
  - ssh://<user@host>/AutomationStackStorm
status: failed
error: Traceback (most recent call last):
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/runners/python_action_wrapper.py", line 259, in <module>
    obj.run()
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/runners/python_action_wrapper.py", line 155, in run
    output = action.run(**self._parameters)
  File "/opt/stackstorm/packs/packs/actions/pack_mgmt/download.py", line 75, in run
    verifyssl=verifyssl, ref=pack_version)
  File "/opt/stackstorm/packs/packs/actions/pack_mgmt/download.py", line 102, in _clone_repo
    repo = Repo.clone_from(repo_url, temp_dir)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/git/repo/base.py", line 942, in clone_from
    return cls._clone(git, url, to_path, GitCmdObjectDB, progress, **kwargs)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/git/repo/base.py", line 897, in _clone
    finalize_process(proc, stderr=stderr)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/git/util.py", line 341, in finalize_process
    proc.wait(**kwargs)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/git/cmd.py", line 292, in wait
    raise GitCommandError(self.args, status, errstr)
[...]

PR #3544, opened 2017-07-03
Fix git URL mangling StackStorm/st2#3534

carbineneutral:
Fixes StackStorm/st2#3534.

Remove the automatic suffixing of ".git" to any pack download git repo URL that doesn't already have it.


Kami:
Thanks for your contribution.

Can you please update affected tests and add new ones (especially the ones which would catch this issue) here -
 https://github.com/carbineneutral/st2/blob/0e7bbf7dce9a5706b48c67ff3387794ecd5d186e/contrib/packs/tests/test_action_download.py#L269

armab:
We also need a [CHANGELOG](
https://github.com/StackStorm/st2/blob/master/CHANGELOG.rst) update

PR #3545, opened 2017-07-03
Incorporate unit + integration test experiments (timing improvement)

armab:
> Related to https://github.com/StackStorm/st2ci/issues/94

Resulting config from the https://github.com/armab/st2/pull/3 prev. week experiments to improve the unit+integration test timing.

The build timing is now `7-10mins` for unit + integration + lint checks, comparing to old `25-30mins`.

As suggested 1yr. ago https://github.com/StackStorm/discussions/issues/143#issuecomment-184856516  avoid putting all eggs in Circle which was there only for packaging because of really advanced and unique features. But we don't need all that + cryptic UI for unit + integration tests. Also we'll suffer from the queue.

---

This change is the part of the bigger game/plan, but for now it's experimental and I'm not changing any workflows yet. Just seeing how it behaves under the load. We'll see just another :white_check_mark: in GitHub status which is OK to ignore. I may experiment with more checks later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ env:
     - TASK=ci-unit NODE_INDEX=0 NODE_TOTAL=2
     - TASK=ci-unit NODE_INDEX=1 NODE_TOTAL=2
     - TASK=ci-integration
-    - TASK=ci-checks ci-packs-tests
+    - TASK="ci-checks ci-packs-tests"
 
 addons:
   apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,7 @@ env:
     - TASK=ci-unit NODE_INDEX=0 NODE_TOTAL=2
     - TASK=ci-unit NODE_INDEX=1 NODE_TOTAL=2
     - TASK=ci-integration
-    - TASK=ci-checks
-    - TASK=ci-packs-tests
+    - TASK=ci-checks ci-packs-tests
 
 addons:
   apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,8 +50,8 @@ after_success:
   - if [ ${TASK} = 'ci-unit' ] || [ ${TASK} = 'ci-integration' ]; then codecov; fi
 
 # https://docs.travis-ci.com/user/notifications/#Webhooks-Delivery-Format
-notifications:
-  webhooks:
-    #- https://ci-webhooks.stackstorm.net/webhooks/build/events
-    # See: webhook.site/#/06fde88c-1610-4c85-ba4e-d9bcacfefe4c
-    - http://webhook.site/06fde88c-1610-4c85-ba4e-d9bcacfefe4c
+#notifications:
+#  webhooks:
+#    #- https://ci-webhooks.stackstorm.net/webhooks/build/events
+#    # See: webhook.site/#/06fde88c-1610-4c85-ba4e-d9bcacfefe4c
+#    - http://webhook.site/06fde88c-1610-4c85-ba4e-d9bcacfefe4c
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ python:
 branches:
   only:
     - master
+    - /^v[0-9]+\.[0-9]+$/
 
 env:
   global:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -0,0 +1,56 @@
...
+      - mongodb-org-shell
+
+services:
+  - mongodb
+  - rabbitmq
+
+cache:
+  pip: true
+  directories:
+    - virtualenv/
```
armab: we cache entire `vietualenv` dir, saves dramatic 2 minutes.

Kami:
Nice work!

I wasn't a fan of moving unit and integration tests to Circle CI in the past either (bad UI, hard to see what is going on, not so suitable for tests, etc. but other people wanted to run everything on one platform).

This means we are kinda back where we started, but I'm +1 for moving unit and integration tests back to Travis.

.travis.yml
Kami: Yeah, we just need to be careful, we had a lot of issues with this in the past - e.g. when Travis upgraded Python version or some system library on which some Python library with C extension depends on.

So if similar issues appear again, we will need to wipe this directory and re-create it on Travis.

Kami:
We also need to add Travis badge back to the readme, etc. and update development section of st2docs (if necessary).

Makefile
```diff
@@ -30,15 +30,15 @@ PYTHON_TARGET := 2.7
...
+NOSE_OPTS := --rednose --immediate --with-parallel
 NOSE_TIME := $(NOSE_TIME)
 
 ifdef NOSE_TIME
-	NOSE_OPTS := --rednose --immediate --with-timer
+	NOSE_OPTS := --rednose --immediate --with-parallel --with-timer
 endif
 
 ifndef PIP_OPTIONS
-	PIP_OPTIONS := -U -q
```
armab: `removed --upgrade` flag which really slow & eats big chunk of timing.
This especially makes sense if we follow strict (`==`) pip fixation in `fixed-requirements`.

test-requirements.txt
```diff
@@ -15,10 +15,12 @@ sphinx-autobuild
...
 rednose
 nose-timer
+# splitting tests run on a separate CI machines
+nose-parallel
 # Required by st2client tests
-pyyaml
+pyyaml<4.0,>=3.12
 RandomWords
-gunicorn>=19.3.0,<19.4
+gunicorn==19.6.0
```
armab: this is synced with main `requirements`, so we don't do pip install -> reinstall because the version diff.

armab:
Not adding Travis badge yet, because I just enable Travis for some time to experiment with it and see if build timing is really consistent at this point.

Makefile
```diff
@@ -30,15 +30,15 @@ PYTHON_TARGET := 2.7
 REQUIREMENTS := test-requirements.txt requirements.txt
 PIP_OPTIONS := $(ST2_PIP_OPTIONS)
 
-NOSE_OPTS := --rednose --immediate
+NOSE_OPTS := --rednose --immediate --with-parallel
 NOSE_TIME := $(NOSE_TIME)
 
 ifdef NOSE_TIME
-	NOSE_OPTS := --rednose --immediate --with-timer
+	NOSE_OPTS := --rednose --immediate --with-parallel --with-timer
```
armab: Just FYI, adding `--with-parallel` is not enough, it will work only if `NODE_TOTAL` and `NODE_INDEX` env variables provided according to https://github.com/dlanger/nose-parallel

So the dev env remains the same as before.

Kami:
^cc @lakshmi-kannan @m4dcoder @dzimine

.travis.yml
```diff
@@ -0,0 +1,56 @@
...
+  only:
+    - master
+    - /^v[0-9]+\.[0-9]+$/
+
+env:
+  global:
+    - CACHE_NAME=JOB1
+  matrix:
+    - TASK=ci-unit NODE_INDEX=0 NODE_TOTAL=2
+    - TASK=ci-unit NODE_INDEX=1 NODE_TOTAL=2
```
Kami: Did you also try splitting in in more than two chunks?

But I guess right now the lower time limit is the time it takes to run integration tests which we can't split into chunks that easily.

armab: I tried splitting Unit tests into 3 chunks, - but that fails because there are some tests which rely on the state of each other.

Integration tests are good - `4-5mins`, - we don't need to split them.

PR #3552, opened 2017-07-06
Update: Add note when "-n" is less than default limit

humblearner:


Kami:
So far so good, but as mentioned the other day - let's please also add some robotfm tests in addition to unit tests.

humblearner:
Robot tests: https://github.com/StackStorm/st2tests/pull/106 and st2cd action for the tests: https://github.com/StackStorm/st2cd/pull/292

Kami:
Nice work on the tests and thanks for getting rid of the global variable. :+1: 

Well test it a bit later.

Kami:
I've tested it locally and it seems to work fine so I will go ahead and merge it :+1: 

Feel free to merge the tests once new packages are built.

PR #3566, opened 2017-07-12
Mirror user's auth token in cookies

enykeev:
During the https://github.com/StackStorm/discussions/issues/242, @armab voiced an opinion that sending auth token through query parameters constitute a security risk in a number of cases involving logging capabilities of intermediate proxies.

This PR provides a solution for the problem by implementing an additional channel for client to pass auth token through - cookies. The main benefit of using cookies instead of query params is that intermediate proxies _usually_ don't log them in any way. The main benefit of using cookies instead of auth headers is that they work universally well for any type of content you may want to access with browser (including images and eventsource to which you can not add a header).

The downside of using a cookies is that they could be used to exploit in a CSRF attacks to execute an arbitrary command impersonating user that's already logged into the system. The CORS mechanics st2 provides from day one can protect from almost every vector of exploitation I can imagine, but on practice, very few st2 users bother themselves to properly set it up.

CSRF is a blind attack and will require a sizable amount of effort to pull off, but the chance is there.

Before the 2.4 is out the door, I'll try to perform it myself and there's still additional protective measures we could add, but for now, this is what we have.


enykeev:
As it turns out we do indeed ship st2 with reasonable defaults in regards to CORS policies. However, if user decides to screw with it for some reason (happened before in cases of custom deployments paired with the lack of interest to CORS and browser security in general), it might lead to him opening the system to CSRF vulnerability.

To illustrate the attack, I've compiled a small example anyone interested could run and see for yourself.

1) Install dev build [#2578](https://circleci.com/gh/StackStorm/st2-packages/2578) as the only one currently vulnerable.

```
curl -sSL https://stackstorm.com/packages/install.sh | bash -s -- --user=st2admin --password='Ch@ngeMe' --dev=2578
```

2) Log into the WebUI

3) Open https://codepen.io/enykeev/pen/zzeKXQ and enter the IP of the service. Press the button. Web Tools Console will tell you familiar:

```
XMLHttpRequest cannot load https://192.168.10.14/api/v1/executions. Response to preflight request doesn't pass access control check: The 'Access-Control-Allow-Origin' header contains the invalid value 'null'. Origin 'https://s.codepen.io' is therefore not allowed access.
```

4) Add `allow_origin = *` to `/etc/st2/ct2.conf`. Do `sudo st2ctl restart`.

5) Repeat step 3. This time Web Tools will tell you nothing, but your execution history will feature a new `core.local` execution.

--

At that point I'm having a rather loud internal debate where:
- one side provides a decent solution for the problem by adding another cookie containing the hash of the token and then adding this hash to every request as a query parameter.
- another side argues that the attack surface is so thin, the attack itself hard to pull of because it would require not only a knowledge of the internal address of the st2 installation, but also involves luring an authenticated user to an infected site. Besides, for vulnerability to be present, user itself would have to manually disable a security feature that's created to protest from exactly that kind of attacks.

armab:
How about adding additional https://www.owasp.org/index.php/SameSite cookie? See https://security.stackexchange.com/a/154108 and https://www.netsparker.com/blog/web-security/same-site-cookie-attribute-prevent-cross-site-request-forgery/

Kami:
Do we also want this in v2.3.2?

enykeev:
@Kami I'd say the change is too disruptive for patch release. No rush either.

enykeev:
@armab given the support of SameSite cookie is still quite spotty (https://caniuse.com/#search=samesite) we can not rely on it anyway. I'll keep my eye on the ball and add it as soon as makes sense. For now, though, we'll have to continue to rely on `allow_origin` config property for custom deployments.

Issue #3575, opened 2017-07-17
Improve API validation so openapi.yaml file utilizes schemas from API model Python classes


Kami:
We need to improve / fix API validation so all the inputs are correctly validated based on the schema defined in openapi.yaml / API model.

Right now we have schemas from API models copy and pasted to openapi.yaml file and that's a maintenance nightmare (things will get out of sync, etc.).

We need to decide which place we want as an authoritative source (either API model Python classes or openapi.yaml file) and then write a generator which either writes schema to openapi.yaml file based on API model class definitions or vice-versa.

See https://github.com/StackStorm/st2/pull/3573#issuecomment-315697541 for context.

Kami:
I would personally prefer to use API model as authoritative source, because:

* We already have a process which generates final openapi.yaml file based on the Jinja template definition.
* API model schemas definitions are already used in other places - e.g. when content comes into the system outside API (via st2-register-content).

If we want with the other approach, we would then have multiple files which are auto-generated and I would prefer to limit auto-generation to openapi file.

PR #3580, opened 2017-07-18
Perform API input data validation using API model definitions

Kami:
This is a stop gap for regression described in #3575 until we implement a long term solution described in that issue (making sure each openapi entity definition contains corresponding JSON schema which is generated and copied over from API model schema attribute).

It should fix the regression (we didn't perform any API validation for endpoints which didn't have entity json schema defined in openapi.yaml).

The code basically does what we did before move to API v2, but we still need to be careful it doesn't break something.

While working on that, I thought we should actually still call `model.validate()` inside the router when performing API input data validation, even when we automatically generate and move all JSON schema definitions to openapi.yaml.

The reason for that is that `model.validate()` doesn't just validate values, but it also returns a cleaned object with default values assigned (we rely on this functionality in some places). That way, definition in `openapi.yaml` would only be used for documentation and auto-generation purposes.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -31,6 +31,8 @@ Fixed
   (bug fix)
 
   Contributed by carbineneutral. #3534 #3544
+* Fix API validation regression so all input data sent to some POST and PUT API endpoints is
+  correctly validated. (bug fix)
 
 2.3.1 - July 07, 2017
 ---------------------
--- a/st2api/tests/unit/controllers/v1/test_policies_rbac.py
+++ b/st2api/tests/unit/controllers/v1/test_policies_rbac.py
@@ -42,7 +42,8 @@ FIXTURES_PACK = 'generic'
 TEST_FIXTURES = {
     'policytypes': [
         'fake_policy_type_1.yaml',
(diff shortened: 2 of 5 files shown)


Kami:
In theory, it should be fine to include this in v2.3.2 since it's not a new feature but a bug fix for a regression, but we should still keep an eye on tests, etc.

enykeev:
> The reason for that is that model.validate() doesn't just validate values, but it also returns a cleaned object with default values assigned (we rely on this functionality in some places).

We need to identify this places and mark them for refactoring/ If we expect a default value, it should be a part of schema. I suspect there might be some nastier things, though.

Kami:
@enykeev Yes, default values are already part of schema, but the problem is how JSON schema validation works.

It doesn't assign default values (just doesn't fail if default: attribute is there and no value is provided) on validation. For default value assignment to work, we needed to implement our own code.

vincent-legoll:
Just fyi, this breaks something that was previously working fine. I gave as packs.install.packs parameter, a single string representing repo URL.

Now I get a cryptic error saying invalid character at line 1 column 1...

I managed to find what the problem was, and now I have to give ["$URL"] instead...

PR #3582, opened 2017-07-19
Add Travis build status image to the readme

Kami:
The title says it all.


armab:
I think it's undecided yet, since we're running that experiment/race between CircleCI & Travis.
Let's see in a couple of weeks running that CI under the prod load & observing the timing & queue & false negatives %.

PR #3583, opened 2017-07-19
Travis CI build speed optimizations

Kami:
Just experimenting a bit if we can also split integration Tests into multiple chunks to speed up the run time.

In addition to that, I also added back some "performance optimizations" I added to Travis long time ago but were lost once we moved to Circle CI (disable journaling for mongodb, etc.).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,10 @@ PYTHON_TARGET := 2.7
 REQUIREMENTS := test-requirements.txt requirements.txt
 PIP_OPTIONS := $(ST2_PIP_OPTIONS)
 
+NODE_INDEX ?= 0
+NODE_TOTAL ?= 1
+TEST_DB_NAME := st2-test-$(NODE_INDEX)
+
 NOSE_OPTS := --rednose --immediate --with-parallel
 NOSE_TIME := $(NOSE_TIME)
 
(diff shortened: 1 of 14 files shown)


armab:
A few points:
- Travis gives about 5 containers for OSS projects.
- Integration tests are running in `~5mins`.
- Unit tests (split into 2 chunks) are running in ~`7-8mins`.
- Entire build timing is the same as the slowest job.

Don't see any benefit of paralleling Integration tests in 2 pieces, because 1 piece of Integration runs in `5mins` which is *already* faster than slowest Unit test container. 

So considering all above, I think it only makes sense to parallel the slowest Unit tests into 3 chunks, but that's failing on some tests which rely on the state of the previous tests (we'll hit that problem in future anyways).

Makefile
```diff
@@ -30,6 +30,15 @@ PYTHON_TARGET := 2.7
 REQUIREMENTS := test-requirements.txt requirements.txt
 PIP_OPTIONS := $(ST2_PIP_OPTIONS)
 
+NODE_INDEX ?= 0
+NODE_TOTAL ?= 1
+
+ifeq ($(shell test $(NODE_TOTAL) -gt 1; echo $$?),0)
+  TEST_DB_NAME := st2-test-$(NODE_INDEX)
```
armab: What's the point of several DBs?
Containers are already isolated, eg. there is no interference between the jobs running in different containers.

Kami: Experimenting if we can perhaps run tests in parallel inside a single vm / container, but yeah, it's probably too complicated (there are other shared resources) and not worth the effort.

Kami:
Closing this for now, it was an experiment.

Kami:
test issue comment

Kami:
test issue comment

PR #3589, opened 2017-07-20
Fix post to action execution with nonexistent action

m4dcoder:
When liveaction reference an action that does not exists, the POST to action execution results in a 500 internal server error. The error should be 400 bad request. A check is added to check that the action_db is not type of NoneType before proceeding.

### DIFF
--- a/st2api/st2api/controllers/v1/actionexecutions.py
+++ b/st2api/st2api/controllers/v1/actionexecutions.py
@@ -96,16 +96,21 @@ class ActionExecutionsControllerMixin(BaseRestControllerMixin):
         if not requester_user:
             requester_user = UserDB(cfg.CONF.system_user.user)
 
-        # Assert the permissions
+        # Assert action ref is valid
         action_ref = liveaction_api.action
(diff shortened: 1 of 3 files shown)


st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -96,16 +96,21 @@ def _handle_schedule_execution(self, liveaction_api, requester_user, context_str
...
-        # Assert the permissions
+        # Assert action ref is valid
         action_ref = liveaction_api.action
         action_db = action_utils.get_action_by_ref(action_ref)
-        user = liveaction_api.user or requester_user.name
-        pack = action_db.pack
 
+        if not action_db:
+            message = 'Action "%s" cannot be found.' % action_ref
+            LOG.error(message)
```
Kami: This should be warning and not error since it's not really a fatal error.

Kami: Besides that, LGTM - good catch.

Kami:
Please also add a changelog entry.

PR #3590, opened 2017-07-20
Fixed a very trivial problem with the processing to show a message

userlocalhost:
## Abstruct
I found a very trivial problem with the processing to show a message in the st2client, and fixed it.

## Background / Problem
I got following message which is (maybe) not intended format by the configuration command in some pack (e.g. `acos`).
```
$ st2 pack config acos
('%s. Skipping...', 'Interactive mode does not support arrays of object type yet')
```
## Solution
I fixed it to substitute variable. By this change, the above output message will be as follows.
```
$ st2 pack config acos
Interactive mode does not support arrays of object type yet. Skipping...
```

### DIFF
--- a/st2client/st2client/utils/interactive.py
+++ b/st2client/st2client/utils/interactive.py
@@ -389,7 +389,7 @@ class InteractiveForm(object):
                 try:
                     result[field] = self._read_field(field)
                 except ReaderNotImplemented as e:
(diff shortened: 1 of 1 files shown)


Kami:
Good catch - thanks.

That's also the reason why I always use `%` and wrap format string arguments in parenthesis - easier to avoid such mistakes :)

PR #3591, opened 2017-07-20
Fix Error definition being too lax

enykeev:
This resulted in unit tests not catching a number of obvious validation errors.

### DIFF
--- a/st2common/st2common/openapi.yaml
+++ b/st2common/st2common/openapi.yaml
@@ -288,7 +288,7 @@ paths:
           x-as: requester_user
           description: User performing the operation.
       responses:
-        '200':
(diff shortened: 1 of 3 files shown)


armab:
Alo just FYI that OpenAPIspec check in CI always exits with success.

`st2-validate-api-spec || true`: https://github.com/StackStorm/st2/blob/e733e5e4809e5843a2b1129dbc8bc33cfbf1e736/Makefile#L138

I don't know if that's intentionally/temporary before we fix all the errors or not.

Kami:
Good catch :+1: 

Can you please also add corresponding regression unit test for every scenario you have fixed? Thanks!

Issue #3578, opened 2017-07-18
`st2 apikey create -k` fails when authentication disabled


LindsayHill:
If authentication is disabled, `st2 apikey create -k` fails with this error:

```sh
vagrant@bwcvagrant:~$ st2 apikey create -k
ERROR: 500 Server Error: Internal Server Error
MESSAGE: Internal Server Error for url: http://127.0.0.1:9101/v1/apikeys
```

st2api.log contains this:

```
2017-07-18 05:36:03,099 140399285519696 ERROR router [-] Failed to call controller function "post" for operation "st2api.controllers.v1.auth:api_key_controller.post": 'NoneType' object has no attribute 'name'
Traceback (most recent call last):
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/router.py", line 414, in __call__
    resp = func(**kw)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2api/controllers/v1/auth.py", line 140, in post
    api_key_api.user = requester_user.name or cfg.CONF.system_user.user
AttributeError: 'NoneType' object has no attribute 'name'
2017-07-18 05:36:03,103 140399285519696 ERROR error_handling [-] API call failed: 'NoneType' object has no attribute 'name'
Traceback (most recent call last):
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/middleware/error_handling.py", line 46, in __call__
    return self.app(environ, start_response)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/router.py", line 445, in as_wsgi
    resp = self(req)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/router.py", line 418, in __call__
    raise e
AttributeError: 'NoneType' object has no attribute 'name' (_exception_data={},_exception_class='AttributeError',_exception_message="'NoneType' object has no attribute 'name'")
2017-07-18 05:36:03,105 140399285519696 INFO logging [-] 07293032-81fe-41ea-8d4d-cdd4e0e222c6 - 500 46 7.573ms (content_length=46,request_id='07293032-81fe-41ea-8d4d-cdd4e0e222c6',runtime=7.573,remote_addr='127.0.0.1',status=500,method='POST',path='/v1/apikeys')
```

Calling `st2 apikey create -k -u st2admin` works.

[...]

PR #3593, opened 2017-07-20
Fix st2 apikey generation when auth is disabled

humblearner:
Addresses: #3578


enykeev:
Nice find! There's a good chance I've used this pattern in other parts of the code base, so we need to be on a lookout. And to save @Kami some breath, tests are very welcome =)

humblearner:
Yup, tests are on the way even codecov/patch is not happy. :) (_although not sure if it is ever happy_)

armab:
Would be nice to have the respective changelog entry next time ;)

humblearner:
Whoops, here you go: https://github.com/StackStorm/st2/pull/3599 . Thanks! 🙂

PR #3594, opened 2017-07-21
[st2client] Implemented ArrayObjectReader to set configuration interactively from the schema of array which has elements of object type.

userlocalhost:
## Abstract
I implemented a ArrayObjectReader for InteractiveForm in the st2client to set configuration from the schema of array which has elements of dicts interactively.

## Background / Problem
As [enykeev said](https://github.com/StackStorm/st2/pull/2964#issuecomment-254734078), the InteractiveForm of st2client has a restriction that couldn't set configuration from the `array of objects` type.
By this restriction, some pack (e.g. [git](https://github.com/StackStorm-Exchange/stackstorm-git), [napalm](https://github.com/StackStorm-Exchange/stackstorm-napalm) and [acos](https://github.com/StackStorm-Exchange/stackstorm-acos)) couldn't interactively set complete configuration by `st2 pack config` command.

## Solution
I implemented another Reader that dedicate to set configuration sets continuously from the schema of `array of objects` type.

### DIFF
--- a/st2client/st2client/utils/interactive.py
+++ b/st2client/st2client/utils/interactive.py
@@ -317,16 +317,18 @@ class ArrayObjectReader(StringReader):
 
     def read(self):
         results = []
-        prefix = u'{}.'.format(self.name)
         properties = self.spec.get('items', {}).get('properties', {})
-        message = 'Would you like to continue to set other settings of "%s"?' % self.name
+        message = '~~~ Would you like to add another item to  "%s" array / list?' % self.name
 
         is_continue = True
+        index = 0
         while is_continue:
+            prefix = u'{name}[{index}].'.format(name=self.name, index=index)
             results.append(InteractiveForm(properties,
                                            prefix=prefix,
                                            reraise=True).initiate_dialog())
 
+            index += 1
             if Question(message, {'default': 'y'}).read() != 'y':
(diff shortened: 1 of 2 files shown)


Kami:
Can you please point me to an example pack which uses such schema so I can test it out?

Kami:
I tried it out and it seems to work OK, but the UX is a bit confusing.

If you don't know it's expecting an array of dicts, you don't really know what is going on. We should some how try to modify the printed output to make this more clear.

Maybe something like that:

```bash
devices[0].credentials: aaa
devices[0].hostname: ttt
devices[0].driver: ttt
~~Would you like to continue to set other settings of "devices"? [y]: y~~
Would you like to add another item to "devices" array / list? [y]
devices[1].credentials: eee
devices[1].hostname: 2
devices[1].driver: 
Would you like to continue to set other settings of "devices"? [y]: n
```

userlocalhost:
Thank you for your review and advices. I'll consider the way to improve the UX.

userlocalhost:
I corrected the prompt messages based on your advice, @Kami.
Would you please review it again?

Kami:
@userlocalhost Thanks - looks great :)

userlocalhost:
Thank you for merging!

PR #3605, opened 2017-07-24
Create CODE_OF_CONDUCT.md

Kami:
We didn't have one yet so it's not such a bad idea to add one.


Kami:
@LindsayHill Yeah good adding on adding it to the community / contributing section.

PR #3609, opened 2017-07-25
Fix invocation of Mistral workflow from Action Chain

m4dcoder:
If the parent action chain contains jinja expression in the execution parameters, the jinja expression will be included in the context begin passed into the Mistral subworkflow. Mistral will try to evaluate the expression when it is evaulating input args and if the expression contains references to local data context, the evaulation fails because the data context is already out of scope. This patch converts the jinja expression into raw jinja blocks when passing thru the context.

### DIFF
--- a/contrib/runners/mistral_v2/tests/unit/test_mistral_v2.py
+++ b/contrib/runners/mistral_v2/tests/unit/test_mistral_v2.py
@@ -283,6 +283,81 @@ class MistralRunnerTest(DbTestCase):
     @mock.patch.object(
         executions.ExecutionManager, 'create',
         mock.MagicMock(return_value=executions.Execution(None, WF1_EXEC)))
+    def test_launch_workflow_under_parent_chain_with_jinja_params(self):
+        ac_ctx = {
+            'chain': {
+                'params': {
+                    'var1': 'foobar',
+                    'var2': '{{foobar}}',
+                    'var3': ['{{foo}}', '{{bar}}'],
+                    'var4': {
+                        'foobar': '{{foobar}}'
+                    },
+                }
+            }
+        }
+
+        liveaction = LiveActionDB(action=WF1_NAME, parameters=ACTION_PARAMS, context=ac_ctx)
+        liveaction, execution = action_service.request(liveaction)
+        liveaction = LiveAction.get_by_id(str(liveaction.id))
+        self.assertEqual(liveaction.status, action_constants.LIVEACTION_STATUS_RUNNING)
+
+        mistral_context = liveaction.context.get('mistral', None)
+        self.assertIsNotNone(mistral_context)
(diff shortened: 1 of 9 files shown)


Kami:
Good catch.

Do we also want this in v2.3.2, since we are still waiting on st2client fix from @humblearner?

st2common/st2common/util/jinja.py
```diff
@@ -180,3 +187,17 @@ def is_jinja_expression(value):
...
+
+def convert_jinja_to_raw_block(value):
+    if isinstance(value, dict):
+        return {k: convert_jinja_to_raw_block(v) for k, v in six.iteritems(value)}
+
+    if isinstance(value, list):
+        return [convert_jinja_to_raw_block(v) for v in value]
+
+    if isinstance(value, six.string_types):
+        if JINJA_REGEX_PTRN.findall(value) or JINJA_BLOCK_REGEX_PTRN.findall(value):
```
Kami: Hm, so we wrap whole thing in raw if we find Jinja start marker - is this desirable? Or do we only want to wrap a place where Jinja expression is located?

Kami: For example, if variable contains something like that: "foo bar {{ some_jinja_var }} bar foo" whole thing will be wrapped into "raw".

Just trying to understand if this is desired and if it matters if we wrap the whole string or just the part (or multiple parts) where Jinja expression is located?

contrib/runners/mistral_v2/tests/unit/test_mistral_v2.py
```diff
@@ -283,6 +283,81 @@ def test_launch_workflow(self):
     @mock.patch.object(
         executions.ExecutionManager, 'create',
         mock.MagicMock(return_value=executions.Execution(None, WF1_EXEC)))
+    def test_launch_workflow_under_parent_chain_with_jinja_params(self):
+        ac_ctx = {
+            'chain': {
+                'params': {
+                    'var1': 'foobar',
+                    'var2': '{{foobar}}',
```
Kami: And while we are at it, let's also add some test cases for scenarios for variables which are composed of more complex strings and don't just contain a single Jinja expression and nothing else, e.g. `foo bar {{ var }} bar {{ varr }}{% if ...%}`, etc.

st2common/st2common/util/jinja.py
m4dcoder: @Kami it doesn't matter if this whole value is in raw to be honest. It's not being used in Mistral. If used at all, it's only for references. I don't expect people to use the action chain parameters from the context. I don't even think they should be in the context passing into Mistral. I don't want to remove it yet as I don't know which poor soul out there is referencing them.

Kami: Thanks for clarifying / confirming.

PR #3614, opened 2017-07-26
Minor fix for format string in timer api

humblearner:


Kami:
LGTM, good catch :+1:

PR #3617, opened 2017-07-26
Check for NoneType when converting Jinja to raw blocks in Mistral runner

m4dcoder:
When converting the jinja expression in the raw blocks in Mistral runner for a parent Action Chain, check for NoneType in the context dictionary.

### DIFF
--- a/contrib/runners/mistral_v2/mistral_v2.py
+++ b/contrib/runners/mistral_v2/mistral_v2.py
@@ -170,7 +170,10 @@ class MistralRunner(AsyncActionRunner):
         # into raw block. If there is any jinja expressions, Mistral will try to evaulate
         # the expression. If there is a local context reference, the evaluation will fail
         # because the local context reference is out of scope.
-        for k, v in six.iteritems(parent_context.get('chain', {}).get('params', {})):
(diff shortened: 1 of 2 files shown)


contrib/runners/mistral_v2/mistral_v2.py
```diff
@@ -170,7 +170,10 @@ def _construct_workflow_execution_options(self):
         # into raw block. If there is any jinja expressions, Mistral will try to evaulate
         # the expression. If there is a local context reference, the evaluation will fail
         # because the local context reference is out of scope.
-        for k, v in six.iteritems(parent_context.get('chain', {}).get('params', {})):
+        chain_ctx = parent_context.get('chain') or {}
```
Kami: `parent_context.get('chain', {}) or {}` would make it even safer :)

m4dcoder: Lol. by default, dict.get will return None if the key doesn't exist.  so with `or {}`, the variable will be assigned {} anyway.  Am i missing something?

Kami: You are right - my bad, end result would be the same :)

PR #3620, opened 2017-07-27
Added [resultstracker] query_interval to st2.conf.sample

nmaludy:
The `[resultstracker] query_interval` option was added st2. It was documented in the changelog but was not added to `st2.conf.sample`. This adds it to `st2.conf.sample`


Kami:
Just a heads up - this file is auto-generated using `configgen` make target which runs as part of `make all`.

I will also make sure it runs as part of other target so we don't forget to run it - #3624.

PR #3624, opened 2017-07-28
Fix config gen script, re-generate sample config

Kami:
The title says it all.


conf/st2.conf.sample
```diff
@@ -29,7 +29,7 @@ max_page_size = 100
 # True to mask secrets in the API responses
 mask_secrets = True
 # StackStorm API server host
-host = 127.0.0.1
```
armab: Does it makes sense to change the config defaults to `127.0.0.1` so we'd generate that value, instead of 0.0.0.0?

Kami: Yeah we can do that - I don't think it should change / break anything, but will see.

Kami: Pushed some changes - 9784bde065f43c8e210d3164e1be94847aa78688, ea66b86fabcf1898f09d59ecf3d6d288d8bf8f87, 971582612a1ad234fb9e95387db26a07d599d7f9.

Will see if e2e tests pass, then we should be fine.

PR #3631, opened 2017-07-29
Spelling fix - 'perhaphs' -> 'perhaps'

mxmader:


LindsayHill:
Thanks - probably a few mistakes like that in there

mxmader:
np -- it happens -- it's the least i can do to complement the effort in the core distro... cheers

PR #3632, opened 2017-07-30
dispacthed -> dispatched

mxmader:


Kami:
Good catch - thanks.

Looks like one of us (probably me) has wrong version of "dispatcher" in spell checker and autocomplete dictionary :)

Issue #3633, opened 2017-07-31
Bug - Logrotate script generates emails


nmaludy:
### Problem
On a daily basis the default logrotate configuration of StackStorm causes mail to be generated with the following body:

```
/etc/cron.daily/logrotate:

Sending SIGUSR1 to st2auth PID: 35143
Sending SIGUSR1 to st2auth PID: 57605
##### st2 components status #####
st2actionrunner PID: 35117
st2actionrunner PID: 35119
st2actionrunner PID: 35121
st2actionrunner PID: 35123
st2actionrunner PID: 35125
st2actionrunner PID: 35127
st2actionrunner PID: 35129
st2actionrunner PID: 35131
st2actionrunner PID: 35133
st2actionrunner PID: 35135
st2api PID: 35921
st2api PID: 35938
st2stream PID: 35831
st2stream PID: 35865
st2auth PID: 35143
st2auth PID: 57605
st2garbagecollector PID: 35147
st2notifier PID: 35151
st2resultstracker PID: 35156
st2rulesengine PID: 35166
st2sensorcontainer PID: 35172
st2chatops is not running.
mistral-server PID: 37041
mistral-api PID: 37040
mistral-api PID: 37071
mistral-api PID: 37072
Sending SIGUSR1 to st2api PID: 35921
Sending SIGUSR1 to st2api PID: 35938
##### st2 components status #####
st2actionrunner PID: 35117
st2actionrunner PID: 35119
st2actionrunner PID: 35121
st2actionrunner PID: 35123
st2actionrunner PID: 35125
st2actionrunner PID: 35127
st2actionrunner PID: 35129
st2actionrunner PID: 35131
st2actionrunner PID: 35133
st2actionrunner PID: 35135
st2api PID: 35921
st2api PID: 35938
st2stream PID: 35831
st2stream PID: 35865
st2auth PID: 35143
st2auth PID: 57605
st2garbagecollector PID: 35147
st2notifier PID: 35151
st2resultstracker PID: 35156
st2rulesengine PID: 35166
st2sensorcontainer PID: 35172
st2chatops is not running.
mistral-server PID: 37041
mistral-api PID: 37040
mistral-api PID: 37071
mistral-api PID: 37072
Sending SIGUSR1 to st2stream PID: 35831
Sending SIGUSR1 to st2stream PID: 35865
##### st2 components status #####
st2actionrunner PID: 35117
st2actionrunner PID: 35119
st2actionrunner PID: 35121
[...]

PR #3634, opened 2017-07-31
Quieted logrotate script

nmaludy:
This is a fix for https://github.com/StackStorm/st2/issues/3633

The following updates have been made:

1. Redirect `stdout` to `/dev/null` for every `st2ctl reopen-log-files` command
2. Added `missingok` to `/var/log/st2/st2web.log` because this file does not seem to be created (at least on my Red Hat test box)

### DIFF
--- a/conf/logrotate.conf
+++ b/conf/logrotate.conf
@@ -4,8 +4,15 @@ copytruncate
 notifempty
 
 ## Rules engine
-/var/log/st2/st2rulesengine.log
 /var/log/st2/st2rulesengine.audit.log {
+  rotate 5
+  size 10M
+  postrotate
+    st2ctl reopen-log-files st2rulesengine > /dev/null
+  endscript
+}
+
+/var/log/st2/st2rulesengine.log {
   size 100M
   rotate 5
   postrotate
@@ -14,7 +21,14 @@ notifempty
 }
 
 ## AUTH
-/var/log/st2/st2auth.log
+/var/log/st2/st2auth.log {
+  daily
+  rotate 5
+  postrotate
+    st2ctl reopen-log-files st2auth > /dev/null
+  endscript
+}
+
 /var/log/st2/st2auth.audit.log {
   daily
   rotate 5
@@ -24,7 +38,14 @@ notifempty
 }
 
 ## API
-/var/log/st2/st2api.log
+/var/log/st2/st2api.log {
+  daily
+  rotate 5
+  postrotate
+    st2ctl reopen-log-files st2api > /dev/null
+  endscript
+}
+
 /var/log/st2/st2api.audit.log {
   daily
   rotate 5
@@ -34,7 +55,14 @@ notifempty
 }
 
 ## Stream
-/var/log/st2/st2stream.log
+/var/log/st2/st2stream.log {
+  daily
+  rotate 5
+  postrotate
+    st2ctl reopen-log-files st2stream > /dev/null
+  endscript
+}
+
 /var/log/st2/st2stream.audit.log {
   daily
   rotate 5
--- a/conf/logrotate.conf
+++ b/conf/logrotate.conf
@@ -4,15 +4,8 @@ copytruncate
 notifempty
 
 ## Rules engine
+/var/log/st2/st2rulesengine.log
 /var/log/st2/st2rulesengine.audit.log {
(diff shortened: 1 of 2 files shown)


nmaludy:
I just tried booting up an st2vagrant instance and `/var/log/st2/st2web.log` is missing there also.

Is this a legacy log file that needs to be removed from the rotate script?

Kami:
Hm, is this a desired behavior?

In case log rotation fails, I imagine service could print some useful info to stdout (in addition to stderr) which could help with troubleshooting things.

Kami:
As far as st2web goes - Yeah i believe it's an old service since now st2web files are served directly by nginx.

Need to check default nginx config to see where st2web vhost stuff is logged.

nmaludy:
Normally we direct any "standard" information to log files and then any errors get raised so an email is generated.

Working in an environment right now where we send all emails generated by `root` user on linux hosts to a sysadmin distribution list. Any cronjob that writes to `stdout` generates an email and spams the whole DL. In our environment we have about several StackStorm VMs, each sending us a nightly email with the content shown in the issue linked in the original post. This spam behavior is not desired, especially since there are no errors and this is simply rotating logs. We would want to be notified however if the rotation fails and we need to take some sort of action (example, maybe the service dies when logs are rotated).

Kami:
@nmaludy Yeah, something like that could work :+1:

conf/logrotate.conf
```diff
@@ -115,15 +115,16 @@ notifempty
...
-    st2ctl reopen-log-files st2garbagecollector
+    st2ctl reopen-log-files st2garbagecollector > /dev/null
   endscript
 }
 
 ## WebUI
 /var/log/st2/st2web.log {
   daily
   rotate 5
+  missingok
```
armab: :+1:

conf/logrotate.conf
```diff
@@ -115,15 +115,16 @@ notifempty
   daily
   rotate 5
   postrotate
-    st2ctl reopen-log-files st2garbagecollector
+    st2ctl reopen-log-files st2garbagecollector > /dev/null
   endscript
 }
 
 ## WebUI
 /var/log/st2/st2web.log {
```
armab: Right, it's outdated.
Please remove entire `st2web` block.

Current `st2web` logs are placed in `/var/log/nginx/` directory: https://github.com/StackStorm/st2/blob/master/conf/nginx/st2.conf#L21-L22 which is logrotated by `nginx`'s rules.

nmaludy:
@armab @Kami Small update based on your comments.

I tried initially to go with @Kami 's approach of redirecting both `stderr` and `stdout` to a file, but also emit `stderr` when `logrotate` was invoked so that an error email was generated. However, this turned out to be a big black hole of redirection and subprocesses invoking `tee`. There ended up being more code in the `postrotate` block to handle log redirections than to restart the service. Also, this code was quite nasty and hard to follow.

I went with @armab 's approach because it was much simpler and straight forward to implement. 

The final change i made was to consolidate the standard and audit logs for a few of the processes. This reduces duplicate code and also allows for a single restart of the service instead of one restart for the standard log rotation and another restart for the audit log rotation.

Kami:
If @armab is fine with just redirecting it to /dev/null, I'm fine with it as well :)

Kami:
As far as "consolidating it" goes - we intentionally had two sections because those are two very different log files.

For the audit ones, users might want to rotate it less often and have different log rotation rules apply to it.

Having said that, if the default values are the same, I'm fine with consolidating it as long as we add a comment and document how to "decouple it" (aka add new section and have different rules apply to audit log files).

nmaludy:
@Kami i've split the configs for the various log files out again. Let me know if you have any other thoughts.

Kami:
Thanks for working on this :+1:

Kami:
Also, to clarify:

> The final change i made was to consolidate the standard and audit logs for a few of the processes. This reduces duplicate code and also allows for a single restart of the service instead of one restart for the standard log rotation and another restart for the audit log rotation.

We don't actually restart a service - we send it `SIGUSR1` signal which causes the service just to re-open the log files without a restart.

armab:
Looks like we reached consensus here.
Thanks everyone, merging :+1:

Issue #3630, opened 2017-07-28
Retry delay is set to a maximum of 5 seconds


dead10ck:
If you try to make a retry policy with a delay of longer than 5 seconds, it will not register. With this policy:

```
---
name: test-retry
description: Retry test if it fails
enabled: true
resource_ref: examples.test
policy_type: action.retry
parameters:
  retry_on: failure
  delay: 30
```

When you try to register it, it complains:

```
2017-07-28 22:30:59,712 INFO [-] =========================================================
2017-07-28 22:30:59,712 INFO [-] ############## Registering policies #####################
2017-07-28 22:30:59,712 INFO [-] =========================================================
2017-07-28 22:30:59,720 WARNING [-] Failed to register policies: Failed to register policy "/opt/stackstorm/packs.dev/examples/policies/test-retry-policy.yaml" from pack "examples": 30 is greater than the maximum of 5

Failed validating u'maximum' in schema['properties'][u'delay']:
    {u'description': u'Number of seconds to wait before retrying the execution.',
     u'maximum': 5,
     u'minimum': 0,
     u'required': False,
     u'type': [u'number', 'null']}

On instance[u'delay']:
    30
Traceback (most recent call last):
  File "/usr/bin/st2-register-content", line 22, in <module>
    sys.exit(content_loader.main(sys.argv[1:]))
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/content/bootstrap.py", line 387, in main
    register_content()
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/content/bootstrap.py", line 341, in register_content
    register_policies()
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/content/bootstrap.py", line 303, in register_policies
    raise e
ValueError: Failed to register policy "/opt/stackstorm/packs.dev/examples/policies/test-retry-policy.yaml" from pack "examples": 30 is greater than the maximum of 5

Failed validating u'maximum' in schema['properties'][u'delay']:
    {u'description': u'Number of seconds to wait before retrying the execution.',
     u'maximum': 5,
     u'minimum': 0,
     u'required': False,
     u'type': [u'number', 'null']}

On instance[u'delay']:
    30
```

5 seconds is incredibly and arbitrarily short. Honestly, I don't think there should be a maximum at all. If I want to delay my action's retries for 4 days, I should be able to.

Kami:
[...]

PR #3637, opened 2017-08-01
Increase max retry delay for retry policy from 5 to 120 seconds

Kami:
User reported this limitation in #3630.

Because of the way retries are currently implemented, long retries are not really recommended, but @lakshmi-kannan and I agreed it still makes sense to support longer retry delays / intervals as long as we document this limitation and suggest alternative (e.g. using an interval trigger instead of retry).

Part of #3630.

## TODO

- [x] Documentation change


lakshmi-kannan:
I am +1 to merging this. @m4dcoder wdyt?

Issue #3635, opened 2017-08-01
Bug: Retrying "register exchanges" causes TypeError exception, using Redis as message queue


johnarnold:
OS: Ubuntu 16.04
st2 version:  2.3.2

Problem description:
Testing redis as alternative message queue. Registering exchanges in kombu connection throws a TypeError.  There's a code problem here that I don't think is related to transport protocol.

```
2017-08-01 02:07:26,696 139712428401040 DEBUG bootstrap_utils [-] Registering exchanges...
2017-08-01 02:07:26,705 139712428401040 ERROR st2notifier [-] (PID=17199) Results tracker quit due to exception.
Traceback (most recent call last):
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2actions/cmd/st2notifier.py", line 56, in main
    _setup()
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2actions/cmd/st2notifier.py", line 27, in _setup
    register_signal_handlers=True)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/service_setup.py", line 117, in setup
    register_exchanges_with_retry()
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/transport/bootstrap_utils.py", line 83, in register_exchanges_with_retry
    return retrying_obj.call(register_exchanges)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/retrying.py", line 205, in call
    if not self.should_reject(attempt):
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/retrying.py", line 189, in should_reject
    reject |= self._retry_on_exception(attempt.value[1])
TypeError: unsupported operand type(s) for |=: 'bool' and 'error'
```

Ref:
https://github.com/StackStorm/st2/issues/3601

lakshmi-kannan:
@johnarnold Thanks for the issue but if the service you use is not an officially supported one, we'll not prioritize it over other issues. Also, the bug you are pointing to is in a third party library. If you can help them out with a fix, that'd be great. Since this is not a direct st2 issue, I am going to go ahead and close the issue. Supporting redis for message queue is not supported and not something we'll support in the near future.

johnarnold:
@lakshmi-kannan 

I'm pointing to an issue with register_exchanges in bootstrap_utils.py returning an 'error' instead of a bool -- I don't believe that's proper behavior for the callback.

[...]

PR #3638, opened 2017-08-02
Transport util fixes

Kami:
Various transport util fixes:

1. Fix usage of `retrying` library inside `register_exchanges_with_retry` function. We incorrectly passed in an exception for `retry_on_exception` argument which actually expects a callable.

Part of #3635.


lakshmi-kannan:
👍

Issue #3290, opened 2017-03-15
[Xenial] Execution stuck in 'requested' state on a fresh RabbitMQ


armab:
When executing any st2 action for the first time on a fresh & clean `RabbitMQ` immediately after st2 startup, - it runs forewer and stuck in `requested` state: 

```sh
root@ubuntu16:~# st2 run core.local echo 123
....................................................................................................................................................
```

```sh
root@ubuntu16:~# st2 execution list
+--------------------------+------------+--------------+------------------------+------------------------+----------------------+
| id                       | action.ref | context.user | status                 | start_timestamp        | end_timestamp        |
+--------------------------+------------+--------------+------------------------+------------------------+----------------------+
| 58c96cadc8980518d7cf8ada | core.local | st2admin     | requested              | Wed, 15 Mar 2017       |                      |
|                          |            |              |                        | 16:32:45 UTC           |                      |
+--------------------------+------------+--------------+------------------------+------------------------+----------------------+


root@ubuntu16:~# st2 execution get 58c96cadc8980518d7cf8ada
id: 58c96cadc8980518d7cf8ada
status: requested
parameters: 
  cmd: echo 123
result: None

```

I can only guess that at early point st2 is busy with RabbitMQ bootstrapping and for some reason can't trigger an action (since topic/queue is not yet created/message is lost or something like that ?) when running things for the first time.

#### Requirements to reproduce:
- OS is Ubuntu Xenial
- RabbitMQ is clean
- st2 was just started
- action runs immediately after st2 start

#### Script to reproduce
I could reproduce it every time with this script: 
```sh
#!/bin/bash

# output executed commands
set -o xtrace

sudo st2ctl stop
# emulate fresh & clean RabbitMQ
rabbitmqctl stop_app
sudo rabbitmqctl reset
rabbitmqctl start_app

# isolate & make sure the problem is not with RabbitMQ startup
sleep 30
# but with StackStorm startup itself
sudo st2ctl start

# The command is stuck and runs forever
# See: st2 execution list
st2 run core.local echo 123
```
----

This is similar to https://github.com/StackStorm/st2-packages/issues/445#issuecomment-286778253
The problem is more serious than it looks like, being blocker for Automation, when deploying StackStorm in prod. Stuck execution immediately after startup is pretty much a bad thing.

I originally thought this could be solved with packaging, but after repro it sounds like more about StackStorm core.
cc @m4dcoder @Kami @lakshmi-kannan.

[...]

Issue #3622, opened 2017-07-27
st2.notifiers.work queue is growing. 


lakshmi-kannan:
Reported by: @emptywee

From slack:
```
it was st2.notifiers.work queue


[5:18] 
with over 747k messages


[5:18] 
purged it


[5:21] 
I see no consumers for that queue


[5:21] 
can anyone here point me in the right direction with this? :slightly_smiling_face:
```

lakshmi-kannan:
I think this is happening after we switched notifier to scan execution queue as opposed to liveactions queue. There is no consumer for liveactions queue.

lakshmi-kannan:
Notifiers now consume "ActionExecution" RabbitMQ exchange with queue name "st2.notifiers.execution.work". Notifier used to scan "LiveAction" exchange with
"st2.notifiers.work" queue name. When you upgrade from older versions, make sure
"st2.notifiers.work" queue size is 0. If you upgrade when it's non-empty, you might
 miss notifications. Post upgrade, please delete queue "st2.notifiers.work" queue manually
 using ``rabbitmqadmin purge queue name=st2.notifiers.work``. If not, this queue will
 grow indefinitely and rabbitmq would eat up a lot of disk space.

lakshmi-kannan:
Added info in Upgrade Notes https://github.com/StackStorm/st2/issues/3622

armab:
"As seen on all production servers."

expecting to see that issue in our fleet as well

PR #3639, opened 2017-08-02
Add ability to pre-declare all the used message bus queues on service startup (allow users to use non RabbitMQ kombu transports)

### LINKED ISSUE #3635: Bug: Retrying "register exchanges" causes TypeError exception, using Redis as message queue

Kami:
This pull request adds ability to predeclare all the used message bus queues (and as such, exchanges) on service startup and allows users to use other non-RabbitMQ kombu transports which are not officially supported.

## Background

#3635 inspired me to dig in a little to try to find out what is going on. After spending whole morning digging in and tracking what is going on, I managed to track down the root cause on why Redis kombu transport doesn't work.

#3638 fixed a small issue which was just related to retrying and error propagation, but the fix for the actual root cause of things not working is included in this PR.

When trying to use Redis kombu transport instead of RabbitMQ one, you got an error like this on each service start up:

```bash
2017-08-02 10:35:44,107 ERROR [-] Rabbitmq connection error: 
Cannot route message for exchange 'st2.liveaction': Table empty or key no longer exists.
Probably the key (u'_kombu.binding.st2.liveaction') has been removed from the Redis database.
```

If you inspect Redis keys you actually find out that none of the exchanges were pre-declared and that's why the issue arises and publishing a message to message bus exchange fails.

```bash
127.0.0.1:6379> keys *
 1) "_kombu.binding.st2.liveaction.status"
 2) "st2.trigger.watch.TimersController-5ad3e912b6"
 3) "st2.trigger.watch.WebhooksController-4e7cd714fd"
 4) "st2.trigger.watch.TimersController-efc54cba86"
 6) "st2.trigger.watch.TimersController-f76e523afb"
 7) "st2.trigger.watch.WebhooksController-d1ae5a0ee9"
 8) "st2.trigger.watch.WebhooksController-24c40d85d4"
```

It turned out that only exchanges for queues which already exist were pre-declared (aka exchanges for which we already have consumers online). That's problematic and obviously won't work in distributed HA workers model.

In this model, producer can be online before the consumer (e.g. API is online before action runner and other services which actually bind a queue to a particular exchange). Another example of late binding is the stream service - we actually only bind to the exchange once first request hits `/v1/stream` endpoint aka stream service.

In fact this happens very often in such setups and it's one of the benefits and reasons of using a worker model - you want messages to still be queued even if the consumer is not online yet and delivered / processed when consumer comes online (normal scenario for such model).

In our example this could mean action runner and other services not being online yet / being restarted / similar, but API service is online and user queues action execution. We obviously want to allow user to perform this operation and queue execution to be run which will be picked up and processed when the action runner service is online.

kombu documentation specifically clarifies this needs to be done in such scenarios (http://docs.celeryproject.org/projects/kombu/en/latest/userguide/producers.html), but because of the implementation details this is not actually required with RabbitMQ transport (nasty!) and our code works just fine with RabbitMQ transport (I guess we were just lucky).

> The declare argument lets you pass a list of entities that must be declared before sending the message. This is especially important when using the retry flag, since the broker may actually restart during a retry in which case non-durable entities are removed.

> Say you are writing a task queue, and the workers may have not started yet so the queues aren’t declared. In this case you need to define both the exchange, and the declare the queue so that the message is delivered to the queue while the workers are offline

## Proposed solution

The solution which fixes this problem is to pre-declare all the queues which causes exchanges to be pre-declared in Redis.

After this change, Redis keyspace looks like this after service set up is ran.

```bash
127.0.0.1:6379> keys *
1) "_kombu.binding.st2.liveaction"
2) "_kombu.binding.st2.liveaction.status"
3) "_kombu.binding.st2.execution"
4) "_kombu.binding.st2.actionexecutionstate"
5) "_kombu.binding.st2.trigger_instances_dispatch"
6) "_kombu.binding.st2.announcement"
7) "_kombu.binding.st2.sensor"
8) "_kombu.binding.st2.trigger"
```

After this change I tested basic functionality (run an action, etc.) and it seems to work fine, but who knows how many different edge case related issues could still be hiding underneath. 

Having said that - **RabbitMQ is still only officially supported and recommended backend by us and using any other backend is at user's own risk. You have been warned - if you use a non-officially supported backend we don't provide support for it and we can't guarantee everything will work. If you are fine with that, go ahead, but don't complain to us if things don't work.**

Even though running this pre-declare code shouldn't in any negative way affect other existing stuff, I still decided to put it behind a feature flag (`messaging.predeclare_queues` config option).

This whole pre-init / pre-declare phase is also nothing special and follows exactly the same pre-initialization approach which is required in a distributed system and we do for other things.

Resolves #3635.

### DIFF
--- a/st2common/st2common/config.py
+++ b/st2common/st2common/config.py
@@ -155,7 +155,11 @@ def register_opts(ignore_errors=False):
         cfg.IntOpt('connection_retries', default=10,
                    help='How many times should we retry connection before failing.'),
         cfg.IntOpt('connection_retry_wait', default=10000,
-                   help='How long should we wait between connection retries.')
+                   help='How long should we wait between connection retries.'),
+        cfg.BoolOpt('predeclare_queues', default=False,
+                   help=('True to pre-declare all the queues on service setup.'
+                         'This is required with some non-standard transports such as Redis one.'))
+
     ]
     do_register_opts(messaging_opts, 'messaging', ignore_errors)
 
--- a/st2common/st2common/config.py
+++ b/st2common/st2common/config.py
@@ -157,8 +157,8 @@ def register_opts(ignore_errors=False):
         cfg.IntOpt('connection_retry_wait', default=10000,
                    help='How long should we wait between connection retries.'),
         cfg.BoolOpt('predeclare_queues', default=False,
-                   help=('True to pre-declare all the queues on service setup.'
-                         'This is required with some non-standard transports such as Redis one.'))
+                   help=('True to pre-declare all the queues on service setup. This is required '
+                         'with some non-standard kombu transports such as Redis one.'))
 
     ]
     do_register_opts(messaging_opts, 'messaging', ignore_errors)
--- a/st2common/st2common/config.py
+++ b/st2common/st2common/config.py
@@ -157,9 +157,9 @@ def register_opts(ignore_errors=False):
         cfg.IntOpt('connection_retry_wait', default=10000,
                    help='How long should we wait between connection retries.'),
         cfg.BoolOpt('predeclare_queues', default=False,
-                   help=('True to pre-declare all the queues on service setup. This is required '
-                         'with some non-standard kombu transports such as Redis one.'))
-
+                   help=('Set this to True to pre-declare all the AMQP queues on service startup. '
+                         'This is required with some non-standard kombu transports such as the '
+                         'Redis one.'))
     ]
     do_register_opts(messaging_opts, 'messaging', ignore_errors)
 
--- a/st2common/st2common/transport/bootstrap_utils.py
+++ b/st2common/st2common/transport/bootstrap_utils.py
@@ -27,15 +27,37 @@ from st2common.transport.liveaction import LIVEACTION_XCHG, LIVEACTION_STATUS_MG
 from st2common.transport.reactor import SENSOR_CUD_XCHG
 from st2common.transport.reactor import TRIGGER_CUD_XCHG, TRIGGER_INSTANCE_XCHG
 
+from st2common.transport import actionexecutionstate
+from st2common.transport import announcement
+from st2common.transport import execution
+from st2common.transport import liveaction
+from st2common.transport import reactor
+
 LOG = logging.getLogger('st2common.transport.bootstrap')
 
 __all__ = [
     'register_exchanges'
 ]
 
+# List of exchanges which are pre-declared on service set up.
 EXCHANGES = [ACTIONEXECUTIONSTATE_XCHG, ANNOUNCEMENT_XCHG, EXECUTION_XCHG, LIVEACTION_XCHG,
              LIVEACTION_STATUS_MGMT_XCHG, TRIGGER_CUD_XCHG, TRIGGER_INSTANCE_XCHG,
              SENSOR_CUD_XCHG]
+
+# List of queues which are pre-declared on service set up.
+# Because of the worker model used, this is required with some non-standard transports such as
+# Redis one. Even though kombu requires queues to be pre-declared upfront in such scenarios,
+# RabbitMQ transport doesn't require that.
+QUEUES = [
+    actionexecutionstate.get_queue(name='init', routing_key='init'),
+    announcement.get_queue(name='init', routing_key='init'),
+    execution.get_queue(name='init', routing_key='init'),
+    liveaction.get_queue(name='init', routing_key='init'),
+    liveaction.get_status_management_queue(name='init', routing_key='init'),
+    reactor.get_trigger_cud_queue(name='init', routing_key='init'),
+    reactor.get_trigger_instances_queue(name='init', routing_key='init'),
+    reactor.get_sensor_cud_queue(name='init', routing_key='init'),
+]
 
 
 def _do_register_exchange(exchange, connection, channel, retry_wrapper):
@@ -59,6 +81,17 @@ def _do_register_exchange(exchange, connection, channel, retry_wrapper):
         LOG.exception('Failed to register exchange: %s.', exchange.name)
 
 
+def _do_predeclare_queue(channel, queue):
+    LOG.debug('Predeclaring queue for exchange "%s"' % (queue.exchange.name))
+
+    try:
+        bound_queue = queue(channel)
+        bound_queue.declare()
+        LOG.debug('Predeclared queue for exchange "%s"' % (queue.exchange.name))
+    except Exception:
+        LOG.exception('Failed to predeclare queue for exchange "%s"' % (queue.exchange.name))
+
+
 def register_exchanges():
     LOG.debug('Registering exchanges...')
     connection_urls = transport_utils.get_messaging_urls()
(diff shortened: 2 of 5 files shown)


st2common/st2common/transport/bootstrap_utils.py
```diff
@@ -27,16 +27,38 @@
...
+# List of exchanges which are pre-declared on service set up.
 EXCHANGES = [ACTIONEXECUTIONSTATE_XCHG, ANNOUNCEMENT_XCHG, EXECUTION_XCHG, LIVEACTION_XCHG,
              LIVEACTION_STATUS_MGMT_XCHG, TRIGGER_CUD_XCHG, TRIGGER_INSTANCE_XCHG,
              SENSOR_CUD_XCHG]
 
+# List of queues which are pre-declared on service set up.
+# Because of the worker model used, this is required with some non-standard transports such as
+# Redis one. Even though kombu requires queues to be pre-declared upfront in such scenarios,
+# RabbitMQ transport doesn't require that.
+QUEUES = [
```
armab: Interesting.
Curious if that can fix https://github.com/StackStorm/st2/issues/3290

Will test.

st2common/st2common/config.py
```diff
@@ -155,7 +155,11 @@ def register_opts(ignore_errors=False):
         cfg.IntOpt('connection_retries', default=10,
                    help='How many times should we retry connection before failing.'),
         cfg.IntOpt('connection_retry_wait', default=10000,
-                   help='How long should we wait between connection retries.')
+                   help='How long should we wait between connection retries.'),
+        cfg.BoolOpt('predeclare_queues', default=False,
```
lakshmi-kannan: What queues? :P. I'd prefer us to be explicit and say AMQP queues or similar.

st2common/st2common/config.py
```diff
@@ -155,7 +155,11 @@ def register_opts(ignore_errors=False):
...
+                   help=('True to pre-declare all the queues on service setup. This is required '
```
lakshmi-kannan: Nit: Set this to ``True`` to pre-declare all AMQP queues on service startup.

st2common/st2common/transport/bootstrap_utils.py
lakshmi-kannan: I am curious as well :). Thanks for watching out @armab

st2common/st2common/transport/bootstrap_utils.py
```diff
@@ -26,16 +28,47 @@
...
+    actionexecutionstate.get_queue(name='st2.preinit', routing_key='init'),
+    announcement.get_queue(name='st2.preinit', routing_key='init'),
+    execution.get_queue(name='st2.preinit', routing_key='init'),
+    liveaction.get_queue(name='st2.preinit', routing_key='init'),
+    liveaction.get_status_management_queue(name='st2.preinit', routing_key='init'),
+    reactor.get_trigger_cud_queue(name='st2.preinit', routing_key='init'),
+    reactor.get_trigger_instances_queue(name='st2.preinit', routing_key='init'),
+    reactor.get_sensor_cud_queue(name='st2.preinit', routing_key='init'),
```
lakshmi-kannan: Are these fake queues that will never be consumed? Because the acutal work queues are named differently. What does rabbitmqadmin tell you about the size of these queues? Do they grow? See https://github.com/StackStorm/st2/issues/3622. I am worried that the queue sizes would grow without consumers. Can you double check, please?

st2common/st2common/config.py
Kami: It's part of the `messaging` config option through so I think it's fine as it is (we also don't include more verbose info in description for other messaging group options :)

st2common/st2common/transport/bootstrap_utils.py
Kami: Yes, they will never be consumed - in fact only one queue which will never be used and grow will be created. Nothing publishes to them so they never grow.

We only do that, because that's the only way to make sure exchanges are created.

```bash
(virtualenv)vagrant@vagrant-ubuntu-trusty-64:/data/stanley$ sudo rabbitmqctl list_queues
Listing queues ...
init	0
st2.actionrunner.canel	0
st2.actionrunner.req	0
st2.actionrunner.work	0
st2.notifiers.execution.work	4092
st2.notifiers.work	3070
st2.preinit	0
....
```

st2common/st2common/config.py
Kami: Done.

st2common/st2common/transport/bootstrap_utils.py
armab: Just checked.

Sadly, https://github.com/StackStorm/st2/issues/3290 MQ-related bug is still clearly reproducible with this new `predeclare_queues=True`

st2common/st2common/transport/bootstrap_utils.py
lakshmi-kannan: At least the names are fixed. That's good. Thanks for digging in.

st2common/st2common/transport/bootstrap_utils.py
lakshmi-kannan: Le sigh.

Kami: Just for the fun of it - you could also try setting this config option value to True and using Redis transport to see if this makes a difference (if the issue is RabbitMQ related).

johnarnold:
Thank you.

Issue #3628, opened 2017-07-28
Sensor container creates way too many rabbitmq queues


lakshmi-kannan:
Reported by: @emptywee on slack

From slack:
```
I have so many queues like `st2.sensor.watch.sensor_container-0ddd32c70e`


[6:58] 
the last part is changing


[6:58] 
maybe over 600 of them


[6:58] 
is it normal?


[6:59] 
most of them are idling, but one


[6:59] 
they have no consumers as well, and they are bound to st2.sensor
```

Kami:
If the queues contain random names they should be deleted after the sensor wrapper process exists.

We do need to investigate though, maybe it makes more sense to make those queues persistent across restarts so sensor wrapper can continue where it left off on restart. But at least now it seems, with the random name that the queue should not be permanent and should only be there for the duration of the sensor wrapper process life time (aka while connection is established to RabbitMQ).

lakshmi-kannan:
Validate if there are unacknowledged messages for queues with name st2.sensor.watch.sensor_container-${UUID}. If there are unacknowledged messages, then wait until it goes down to zero. 

```
rabbitmqadmin list queues vhost name node messages messages_unacknowledged
```

Now stop st2 and delete the sensor_watch queues
 
```
st2ctl stop
```

Delete sensor_watch queues. 

```
for q in `rabbitmqadmin -f long -d 3 list queues | grep name | grep sensor | awk '{print $2}'`;do rabbitmqadmin delete queue name=$q;done
```

emptywee:
Thank you @lakshmi-kannan  :) Deleting it, right now...

PR #3640, opened 2017-08-02
Delete sensor watcher queue

lakshmi-kannan:
Closes: https://github.com/StackStorm/st2/issues/3628

TODO:

* [x] Integration tests ** I had to enable rabbitmq_management in travis because https://stackoverflow.com/questions/4287941/how-can-i-list-or-discover-queues-on-a-rabbitmq-exchange-using-python ** 
* [x] Document how to delete the queues that are left around. (https://github.com/StackStorm/st2/issues/3628#issuecomment-319801557)


CHANGELOG.rst
```diff
@@ -25,6 +25,12 @@ Fixed
 
   Reported by John Arnold.
 
+* Fix a bug where sensor watch queues were not deleted after sensor container process was shut
```
Kami: Should we add a small bash script which deletes those and tell users in upgrade notes on how to run it?

st2common/st2common/services/sensor_watcher.py
```diff
@@ -95,9 +95,18 @@ def start(self):
             self.connection.release()
 
     def stop(self):
+        LOG.debug('Shutting down sensor watcher.')
         try:
             if self._updates_thread:
                 self._updates_thread = eventlet.kill(self._updates_thread)
+
+            if self.connection:
+                channel = self.connection.channel()
```
Kami: Because it runs inside `stop` it means this code would still not run in case sensor is not shut down cleanly and this would mean we will still have dead queues left behind.

Another, probably better approach, would be to use a non-durable and auto deletable queue (http://docs.celeryproject.org/projects/kombu/en/1.0/reference/kombu.entity.html#kombu.entity.Queue.durable).

st2reactor/tests/integration/test_rules_engine.py
```diff
@@ -11,6 +11,7 @@
 # distributed under the License is distributed on an "AS IS" BASIS,
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
+# limitations under the License.
```
Kami: lol, good catch :D

CHANGELOG.rst
lakshmi-kannan: Yeah, I documented that script in the issue already. I just didn't know where to place that documentation. st2docs seemed too much. Ideas welcome.

st2common/st2common/services/sensor_watcher.py
lakshmi-kannan: The sensor watcher isn't tied to the sensors life cycle at all. See code here https://github.com/StackStorm/st2/blob/master/st2reactor/st2reactor/container/manager.py#L66. From my reading, it seems like the stop code path will be invoked when the st2sensorcontainer process is shutdown by user running sudo service/systemctl stop or killing the process with Ctrl + C. So I think it's safe to delete the queue here. I can try with durable and see if that works really. I tested the explicit delete I have in dev env and it definitely works.

Kami: Yeah, I'm all for explicitly deleting it (normal scenario), but I was also talking about the edge case scenario where we can't ensure this code will run - e.g. if process crashes or isn't shut down cleanly.

In that case this code won't run so queue will stay there. We can avoid that by using a non durable and auto deletable queue - in this case queue will be deleted when connection to the broken is lost.

lakshmi-kannan: >in this case queue will be deleted when connection to the broken is lost.

This is exactly the problem I have. They detect server restarts with connection drops which is entirely incorrect. I have seen connection drops in AWS so many times and we have code to reestablish connection. Deleting the queue could mean we lose information about some sensors. In the explicit delete case, a queue is deleted exactly when the sensor container is stopped. And on restart, we anyways load all the sensors and therefore we won't miss anything. The only downside is stray queues maybe left around if the process crashes the chance of which is fairly low.

Kami: But the queue should get re-created when connection is successfully re-established, right?

If not, we can / should utilize `declare=` functionality I mentioned in #3648

Kami:
OK, let's go ahead and merge this explicit delete for now.

We can always include durable and auto delete in a new PR if we confirm it works as it should and it's a good idea.

PR #3643, opened 2017-08-03
Various debug mode logging improvements

Kami:
This pull request improves various debug mode related logging:

1. When debug mode is enabled, also set paramiko and kombu logger level to `DEBUG`
2. When `disable_logging` function is called, actually reset logger levels back to `INFO`
3. When debug mode is enabled, don't filter out all log messages using level `DEBUG` generated by `pyamqp`, but only the ones generated by `heartbit_tick` function which is called every 2 seconds. Disabling all the DEBUG messages means we lose some valuable debugging information.

### DIFF
--- a/st2common/st2common/log.py
+++ b/st2common/st2common/log.py
@@ -25,7 +25,7 @@ from functools import wraps
 
 import six
 
-from st2common.logging.filters import ExclusionFilter
+from st2common.logging.filters import LoggerNameExclusionFilter
 
 # Those are here for backward compatibility reasons
 from st2common.logging.handlers import FormatNamedFileHandler
@@ -168,7 +168,7 @@ logging.Logger.audit = _audit
 def _add_exclusion_filters(handlers, excludes=None):
     if excludes:
         for h in handlers:
-            h.addFilter(ExclusionFilter(excludes))
+            h.addFilter(LoggerNameExclusionFilter(excludes))
 
 
 def _redirect_stderr():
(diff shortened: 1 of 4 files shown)


Kami:
With this stuff I managed to track down root cause for #3290.

The problem is that API starts listening and accepting connections before we fully establish connection to RabbitMQ.

If you check st2api.log you will see that this message is only printed later one, after executions API has already been hit which means first execution will get "stuck" in requested state since it will never actually reach the exchange.

```
2017-08-03 10:38:30,940 140523324677648 DEBUG connection [-] Start from server, version: 0.9, properties: {'information': 'Licensed under the MPL.  See http://www.rabbitmq.com/', 'product': 'RabbitMQ', 'copyright': 'Copyright (C) 2007-2015 Pivotal Software, Inc.', 'capabilities': {'exchange_exchange_bindings': True, 'connection.blocked': True, 'authentication_failure_close': True, 'basic.nack': True, 'per_consumer_qos': True, 'consumer_priorities': True, 'consumer_cancel_notify': True, 'publisher_confirms': True}, 'cluster_name': 'rabbit@ip-172-31-15-77.us-west-2.compute.internal', 'platform': 'Erlang/OTP', 'version': '3.5.7'}, mechanisms: ['AMQPLAIN', 'PLAIN'], locales: [u'en_US']
2017-08-03 10:38:30,942 140523324677648 INFO mixins [-] Connected to amqp://guest:**@127.0.0.1:5672//
```

PR #3645, opened 2017-08-03
Implement pause and resume for Action Chain

m4dcoder:
Allows Action Chain to be paused and then resumed. Pause and resume will cascade down to sub-chain or sub-workflow. Pausing a sub-chain or sub-workflow will cascade the pause up to the parent Action Chain. Resuming a sub-chain or sub-workflow on the other hand will not cascade up up to the parent Action Chain to allow troubleshooting/debugging purposes.

### DIFF
--- a/contrib/runners/action_chain_runner/action_chain_runner.py
+++ b/contrib/runners/action_chain_runner/action_chain_runner.py
@@ -388,6 +388,10 @@ class ActionChainRunner(ActionRunner):
                 LOG.exception(m)
                 top_level_error = self._format_error(e, m)
 
+            # If there are no action node to run next, then mark the chain successful.
+            if not action_node:
+                chain_status = action_constants.LIVEACTION_STATUS_SUCCEEDED
+
         # Otherwise, figure out the last task executed and
         # its state to determine where to begin executing.
         else:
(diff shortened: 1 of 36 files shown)


Kami:
Would be better if you open this pr against #3507 branch, would be easier to review it.

contrib/runners/action_chain_runner/action_chain_runner.py
```diff
@@ -278,104 +278,210 @@ def pre_run(self):
...
     def run(self, action_parameters):
-        # holds final result we store.
+        # Run the action chain.
+        return self._run_chain(action_parameters)
+
+    def pause(self):
+        # Identify the list of action executions that are workflows and cascade pause.
+        for child_exec_id in self.execution.children:
+            child_exec = ActionExecution.get(id=child_exec_id, raise_exception=True)
+            if (child_exec.runner['name'] in action_constants.WORKFLOW_RUNNER_TYPES and
```
Kami: Shouldn't we just check for action-chain workflows here?

Or do we also want to pause a child task in case it's an action which is a Mistral workflow?

contrib/runners/action_chain_runner/action_chain_runner.py
```diff
@@ -278,104 +278,210 @@ def pre_run(self):
...
+            action_constants.LIVEACTION_STATUS_RUNNING,
+            publish=False
+        )
+
+        # Run the action chain.
+        return self._run_chain(action_parameters, resuming=True)
+
+    def _run_chain(self, action_parameters, resuming=False):
+        # Set chain status initially to success until chain cancels, fails, or times out below.
+        chain_status = action_constants.LIVEACTION_STATUS_SUCCEEDED
```
Kami: I think setting it to failure by default would be safer and also act as "catch all" in case we accidentally fall through and don't explicitly set a status.

Kami:
There is a lot of going on, but from a first high-level review, it mostly looks good to me.

I still need to do another more detailed pass, especially through the tests to make sure we have a test case for every possible normal and edge case which could arise (and there are many possible scenarios which can arise).

contrib/runners/action_chain_runner/action_chain_runner.py
```diff
@@ -13,39 +13,36 @@
...
 import traceback
 import uuid
 import datetime
 
-from jsonschema import exceptions as json_schema_exceptions
+from jsonschema import exceptions as json_schema_exc
 
 from st2common.runners.base import ActionRunner
 from st2common import log as logging
-from st2common.constants.action import ACTION_CONTEXT_KV_PREFIX
```
Kami: I prefer the existing approach, easier to grep for things :P

contrib/runners/action_chain_runner/action_chain_runner.py
```diff
@@ -278,104 +278,210 @@ def pre_run(self):
...
+            if self.liveaction and hasattr(self.liveaction, 'result') and self.liveaction.result:
+                result = self.liveaction.result
+
+            # Initialize or rebuild existing context_result from liveaction
+            # which holds the result used for resolving context in Jinja template.
+            for task in result.get('tasks', []):
+                context_result[task['name']] = task['result']
+
+            # Restore or initialize the top_level_error
+            # thata stores a reference to a top level error.
```
Kami: Small typo

Kami:
Also needs a corresponding changelog and documentation entry (if we don't have one already).

contrib/runners/action_chain_runner/action_chain_runner.py
m4dcoder: We want to pause Mistral workflow as well.

contrib/runners/action_chain_runner/action_chain_runner.py
m4dcoder: Ok. I will change it.

PR #3648, opened 2017-08-04
Pre-declare (bind) all static message bus queues on service startup - fix a race condition which could cause first message(s) to get lost under some conditions

Kami:
This fixes those two issues - #3290 and #3635.

## Background

Our code didn't correctly pre-declare static queues used by the various services before they were used so there was a race condition which could, under some scenarios, cause first message(s) published on the exchange to get lost.

This would happen in a scenario where publisher published a message, but there were no consumers online yet (and as such, queue hasn't been declared / bound yet). In such scenario, there are no routes so RabbitMQ simply ignores / drops the message.

An example of that is API service starting and serving a request before action runner service is online. There was a very short time window when this could happen and that's why #3290 was very hard to track down (I spend quite a lot of time digging in and trying to find the root cause).

## Proposed Solution

In my proposed solution I pre-declare all the static queues upfront on service startup. Imo, that's the best approach and the same pattern we follow for other "init" / "ensure it exists before it's used" stuff.

Another approach would be to pass `declare=[exchange, queue_which_will_eventually_consume_this_message] argument` to `Producer.publish` method which would cause it to call `maybe_declare` to ensure those entities are declared before the message is published (see kombu docs and examples for more details).

In our case, this would need to happen here - https://github.com/StackStorm/st2/blob/master/st2common/st2common/transport/publishers.py#L44.

I avoided this approach because it's, imo, a bad one - it would break the whole abstraction and cause a cycle. Now each Publisher would need to also be aware of all the possible consumers which is obviously not ideal - that's the whole behind the Publisher / Consumer abstraction - Publisher doesn't necessary need to be aware or care who will consume the message.

As a "bonus" - #3635 was basically the same underlying issue so it also fixes that and means we can get rid of flag added in #3639 (this pull request superseeds that one).

Also, a fun fact is that this race was basically there since the beginning - we were just lucky #3290 exposed it.


Kami:
I took some liberty to rename some things to make it easier to find / grep.

I also needed to move some things around (queue declaration) so they could be referenced in multiple places without causing a cyclic import issue.

PR #3653, opened 2017-08-07
Implement runner, API, and client code for Inquiries

Mierdin:
This PR introduces a new runner, `inquirer`, as well as API endpoints and `st2client` changes necessary for working with Inquiries.

# Usage

Please review the corresponding [**st2docs PR**](https://github.com/StackStorm/st2docs/pull/595) for full usage documentation and examples.

[asciicast](https://asciinema.org/a/QUasgDVnUufT5ZltHYGSpFGSC)

# New "Inquiry" Runner

This PR creates the `inquirer` runner, which forms the basis for "asking a question" in the middle of a Workflow.

Based on previous and current (see below) design discussions, this runner is fairly simple:
- Dispatch trigger indicating a new inquiry
- Request that the parent execution (workflow) is paused (for nested workflows, this is done on the root workflow)
- Return a `pending` status

The entire act of handling a response (including validation) and resuming the workflow is handled by the API, and is outside the scope of this runner. This runner's sole purpose is to pause the workflow and provide enough context for a 3rd party to make a proper response.

The `inquirer` runner supports a number of parameters with sensible defaults. These defaults can of course be overridden by passing them into the action upon invocation. The parameters used for an Inquiry will be placed in the action `result`.

> Currently the `ttl` parameter does nothing, and will be used in a future PR where `st2garbagecollector` will use it to clean up old Inquiries

# API

For the time being, Inquiries are effectively treated as ActionExecutions, with a bit of additional logic, meaning we haven't built a full data model for Inquiries yet. This actually works pretty well, but there are a few "weird" things in this PR, especially around RBAC, because of this decision.

For the time being, the API is (hopefully) designed under the assumption that Inquiries may be their own data model in the future, and acts like it is today.

This API will provide:

- `GET /api/v1/inquiries`: Retrieve all Inquiries
- `GET /api/v1/inquiries/{id}`: Retrieve a specific inquiry by ID
- `PUT /api/v1/inquiries/{id}`: Provide response data to an Inquiry

# `st2client` Updates

There are three new commands, one for each new API function:

- `st2 inquiry list`
- `st2 inquiry get <id>`
- `st2 inquiry respond <id> <response json>`

I opted to keep command-line options to a minimum, as most of the options for the similar `st2 executions list` command are focused mainly on filtering, and Inquiries shouldn't be that long-lived. Inquiries (for now) are basically just executions with a certain status, and they shouldn't be in that status for long. So I added the `limit` option, but not much else. Let me know if you feel other filters, like datetime filters are required, but I'm thinking they shouldn't be necessary.

# Testing Instructions

> NOTE that this is not a complete picture of everything we want to do with Inquiries. This PR focused on an end-to-end implementation of the core functionality. There are a few other misc. things that need to be done to really round this feature out, like adding garbage collection and ensuring things work well with chatops. For this round of testing I want to focus on the core functionality, like the way the API and CLI work, and the invocation of the action

This is a large and significant feature so manual testing before the merge is a good idea. Here are some things to test

Please first review the corresponding [st2docs](https://github.com/StackStorm/st2docs/pull/595/files) PR containing formal descriptions of what Inquiries are and how to use them (and please leave comments on that PR as needed too)

The following commands will check this branch out, and spin up a development instance of st2, with mistral, and install the client:

```
git clone -branch api-ask-response git@github.com:StackStorm/st2.git
cd st2
make requirements
source virtualenv/bin/activate
tools/launchdev.sh stop && tools/launchdev.sh startclean -m -x && python st2client/setup.py install > /dev/null
```

From there, you can create and execute workflows and rules that test Inquiries. Note the existing [Mistral](https://github.com/StackStorm/st2/pull/3653/files#diff-e207a2840069179835bbcbc4ff503c0e) and [ActionChain](https://github.com/StackStorm/st2/pull/3653/files#diff-4ce4523b69f176080c6739fda0ed64f0) workflows in this branch that may serve as a good place to start.

There are a number of things that I feel would be useful to test, but by all means, go beyond this list:

- Basic testing. Run a simple workflow with an Inquiry (`core.ask` action), and respond to it. Confirm that the workflow initially pauses, then respond to the inquiry. Then confirm the workflow resumes. Also confirm that the response fails with invalid data (and workflow does not resume)
- Test with various parameters. Each parameter has an assumed default, override these and confirm that the behavior is as expected. For instance, override the `schema` parameter with your own, and supply data that would only validate against that
- Nested workflows. Confirm that pauses and resumes cascade up the chain to the root workflow
- Standalone Inquiries (not in a workflow). Not a common use case, but in previous discussions, we decided to support this. Confirm you can run `core.ask` on its own, and that it supports the same response behavior.
- Send slack notifications using a rule that consumes the `core.st2.generic.inquiry` trigger. Note that this is just a PoC - a future PR will introduce proper chatops integration once this core functionality is vetted
- Test all client commands and flags (`st2 inquiry get/list/respond`). Ensure the output is as you would expect
- RBAC (This is my first time doing anything with RBAC so I would appreciate some eyes on this) - Ensure you can lock down Inquiry resources the way you'd expect from other resources. I played around with this a bit myself, feel free to [use my example](https://gist.github.com/Mierdin/5cce5ec644a839d0580f4998ec59c991) as a starting point (slightly out of date - no longer necessary to use `inquiry:ask`; just use `inquiry`). This was one area where not having a dedicated model for Inquiries made things difficult (though not impossible).
- Response permissions using `users` or `roles` runner parameters. This goes beyond RBAC and actually permits/denies a per-Inquiry response

Again, please review the corresponding [st2docs PR](https://github.com/StackStorm/st2docs/pull/595) for full usage documentation and examples.

# TODOs

- [x] Test RBAC manually to ensure things still work
- [ ] (winson) Are there integration tests for the st2cd.st2_e2e_tests? Please include one for action chain and one for mistral. Also provide PR to st2cd (and st2ci?) to run this just like the mistral itests are being run
- [ ] Finish docs
- [ ] Check coverage
- [ ] Ensure example chain and mistral workflow is solid


Kami:
I also updated TODO and added RBAC there.

Every time we add a new API endpoint we also need to implement corresponding RBAC rules and checks.

Mierdin:
Good call, thank you!

st2api/st2api/controllers/v1/inquiries.py
```diff
@@ -0,0 +1,76 @@
...
+    """
+    supported_filters = {}
+    model = None
+    access = None
+
+    def get_all(self, requester_user=None):
+        return [test_inquiries]
+
+    def get_one(self, id, requester_user=None):
+        return [i for i in test_inquiries if i["id"] == id][0]
```
m4dcoder: This is fragile. This list needs to be check for empty before accessing by index `[0]`.

st2common/st2common/openapi.yaml
```diff
@@ -4002,6 +4075,16 @@ definitions:
     additionalProperties: False
   ExecutionFilters:
     type: object
+  Inquiry:
+    type: object
```
m4dcoder: Since this is a user facing API, we need to be more specific w/ the schema.

st2common/st2common/openapi.yaml.j2
```diff
@@ -3999,6 +4072,16 @@ definitions:
     additionalProperties: False
   ExecutionFilters:
     type: object
+  Inquiry:
+    type: object
```
m4dcoder: Since this is a user facing API, we need to be more specific w/ the schema.

st2api/st2api/controllers/v1/inquiries.py
Mierdin: Understood and agreed, this won't exist after I make the function actually do something - this is just there to get data into the response so I can ensure the API works.

st2common/st2common/openapi.yaml
Mierdin: Ack that, will follow your advice from https://github.com/StackStorm/st2/pull/3537 and add more detail here as well.

Mierdin:
I've decided to rebase this branch so that the work in https://github.com/StackStorm/st2/pull/3537 is available for testing. Now that I'm actually implementing a lot of the back-end logic, it's becoming difficult to not have an actual Inquiry execution to work with.

st2client/st2client/commands/inquiry.py
```diff
@@ -0,0 +1,140 @@
...
+                                 help=('List of attributes to include in the '
+                                       'output. "all" will return all '
+                                       'attributes.'))
+        self.parser.add_argument('-w', '--width', nargs='+', type=int,
+                                 default=None,
+                                 help=('Set the width of columns in output.'))
+
+    @resource.add_auth_token_to_kwargs_from_cli
+    def run(self, args, **kwargs):
+        return self.manager.query_with_count(limit=args.last, **kwargs)
```
humblearner: Can use [query()](https://github.com/StackStorm/st2/blob/master/st2client/st2client/models/core.py#L280)  here, unless we plan to use limit note, eventually.

st2client/st2client/commands/inquiry.py
```diff
@@ -0,0 +1,140 @@
...
+        'id',
+        'roles',
+        'users',
+        'tag',
+        'ttl'
+    ]
+
+    def __init__(self, resource, *args, **kwargs):
+
+        self.default_limit = 20
```
humblearner: Any reason why `default_limit=20`? Can it be 50, to be consistent with other commands?

st2common/st2common/openapi.yaml
```diff
@@ -1190,6 +1190,91 @@ paths:
...
+  /api/exp/inquiries:
+    get:
+      operationId: st2api.controllers.exp.inquiries:inquiries_controller.get_all
+      description: Returns a list of all inquiries
+      parameters:
+        - name: limit
+          in: query
+          description: Number of executions to get
+          type: integer
+          default: 20
```
humblearner: Same as https://github.com/StackStorm/st2/pull/3653/files#r137388337, however, here it should be `100`. This is till under discussion: #3708

contrib/runners/inquirer/runner.yaml
```diff
@@ -0,0 +1,37 @@
...
+  runner_parameters:
+    schema:
+      default:
+        title: response_data
+        type: object
+        properties:
+          continue:
+           type: boolean
+           description: Would you like to continue the workflow?
+        required: ["continue"]
```
Kami: We actually use a different / updated json schema where "required" attribute is part of the property (originally it's a global attribute like you have done it, here but we made changes to the schema a long time ago so we can use per attribute / property definition).

In this case this means it would be part of "continue" property:

```python
continue:
  type: boolean
   description: Would you like to continue the workflow?
   required: true
```

contrib/runners/inquirer/runner.yaml
```diff
@@ -0,0 +1,37 @@
...
+      required: true
+      description: A JSON schema that will be used to validate the response data
+      type: object
+    tag:
+      default: ""
```
Kami: Maybe we should default to None instead?

contrib/runners/inquirer/runner.yaml
```diff
@@ -0,0 +1,37 @@
...
+      type: array
+    users:
+      default: []
+      required: false
+      description: A list of usernames that are permitted to respond to the action (if nothing provided, all are permitted)
+      type: array
+    ttl:
+      default: 1440
+      required: true
+      description: Time (in minutes) that an unacknowledged Inquiry is cleaned up
```
Kami: Maybe re-word it a bit - something along the lines of "Number of minutes to wait before timing out the inquiry if no response / ack is received by then".

st2api/st2api/controllers/exp/inquiries.py
```diff
@@ -0,0 +1,295 @@
...
+    supported_filters = copy.deepcopy(SUPPORTED_FILTERS)
+
+    # No data model currently exists for Inquiries, so we're "borrowing" ActionExecutions
+    model = ActionExecutionAPI
+    access = ActionExecution
+
+    def get_all(self, requester_user=None, limit=None, **raw_filters):
+        """Retrieve multiple Inquiries
+
+           TODO(mierdin): This function doesn't currently make RBAC assertions for two reasons:
```
Kami: For "get all" and some other API endpoints, RBAC permission type is now defined inside the OpenAPI schema file (https://github.com/StackStorm/st2/blob/master/st2common/st2common/openapi.yaml.j2#L772) and the check is performed inside the router (this applied to simple checks).

I was personally never a fan of that approach - previously we used decorator for that so it was very explicit where RBAC check is performed and, imo, that's important with things like RBAC.

In any case, I lost that battle / argument with @enykeev when we moved to openapi, so we are where are are right now.

st2api/st2api/controllers/exp/inquiries.py
```diff
@@ -0,0 +1,295 @@
...
+           - I am not sure the best approach, since I would have to make an assertion
+             inside the below loop?
+
+            Handles requests:
+                GET /inquiries/
+        """
+
+        raw_inquiries = super(InquiriesController, self)._get_all(
+            limit=limit,
+            raw_filters={'status': 'pending', 'runner': 'inquirer'}
```
Kami: Should probably reference a constant for "pending" here.

st2api/st2api/controllers/exp/inquiries.py
```diff
@@ -0,0 +1,295 @@
...
+            the parent workflow.
+
+            Handles requests:
+                PUT /inquiries/<inquiry id>
+        """
+
+        LOG.debug("Inquiry %s received response payload: %s" % (inquiry_id, response_data.response))
+
+        # Retrieve details of the inquiry via ID (i.e. params like schema)
+        inquiry_execution = self._get_one_by_id(
```
Kami: How does permission check work right now? Does user need to have a permission to a specific inquiry id to respond?

Kami: Please ignore that - I see "_can_respond" method below.

st2api/st2api/controllers/exp/inquiries.py
```diff
@@ -0,0 +1,295 @@
...
+            id=inquiry_id,
+            requester_user=requester_user,
+            permission_type=PermissionType.INQUIRY_RESPOND
+        )
+
+        if inquiry_execution.runner.get('runner_module') != "inquirer":
```
Kami: Seems like those two if checks could be refactored in a separate utility method.

st2api/st2api/controllers/exp/inquiries.py
```diff
@@ -0,0 +1,295 @@
...
+            need to pick and choose fields from this to form a new franken-model for
+            Inquiries so we don't return a bunch of extra crap. The idea is to
+            provide data in response to requests that make it look like Inquiry
+            is its own data model.
+        """
+
+        # The "put" and "get" commands will have an actual ActionExecutionAPI object, so
+        # if that's what's being passed in, let's convert it to a dict first.
+        # ("get_all" provides a dict natively)
+        if isinstance(raw_inquiry, ActionExecutionAPI):
```
Kami: Hm, so we can't use API model for the response here?

Even if we can't use it directly it would, imo, make sense to utilize `InquityAPI` / `InquiryResponse` API model and perform that transformations inside the API model.

This way we "follow the abstraction" since the API models should be responsible for transforming data for either input or output.

st2api/st2api/controllers/exp/inquiries.py
```diff
@@ -0,0 +1,295 @@
...
+
+        if raw_inquiry.status != "pending":
+            abort(http_client.BAD_REQUEST, 'Inquiry %s has already been responded to' % inquiry_id)
+            return
+
+        new_inquiry = self._transform_inquiry(raw_inquiry)
+
+        return new_inquiry
+
+    def put(self, inquiry_id, response_data, requester_user):
```
Kami: Same here, we should also utilize API model for input / request data.

st2api/st2api/controllers/exp/inquiries.py
```diff
@@ -0,0 +1,295 @@
...
+        inquiry = {
+            "id": raw_inquiry["id"]
+        }
+        for field in desired_fields:
+            inquiry[field] = raw_inquiry["result"][field]
+
+        return inquiry
+
+    def _mark_inquiry_complete(self, inquiry_id, result):
+        """Mark Inquiry as completed
```
Kami: It looks like there is a small possible race in the "PUT" method (if multiple client send payload to that API endpoint in a short time frame) because we don't do any locking there, but that's probably fine, for now.

st2api/st2api/controllers/exp/inquiries.py
```diff
@@ -0,0 +1,295 @@
...
+
+        :rtype: bool - True if requester_user is able to respond. False if not.
+        """
+
+        # Deny by default
+        roles_passed = False
+        users_passed = False
+
+        # Determine role-level permissions
+        roles = inquiry.get('roles')
```
Kami: I prefer to do `roles = inquiry.get('roles', [])` and then just `for role in roles` to avoid if and additional level of nesting, but that's just a personal preference and current approach works as well :)

st2api/st2api/controllers/exp/inquiries.py
```diff
@@ -0,0 +1,295 @@
...
+        if roles:
+            for role in roles:
+                LOG.debug("Checking user %s is in role %s" % (requester_user, role))
+                LOG.debug(rbac_utils.user_has_role(requester_user, role))
+                # TODO(mierdin): Note that this will always return True if Rbac is not enabled
+                # Need to test with rbac enabled and configured
+                if rbac_utils.user_has_role(requester_user, role):
```
Kami: OK, so it looks like the behavior is "OR".

Is there a reason when we would want "AND" behavior instead?

(just thinking out loud)

st2client/st2client/commands/inquiry.py
Kami: We should use limit note here, like we should in any list api endpoint, imo :)

st2common/st2common/openapi.yaml
Kami: As mentioned above, I believe we should utilized API model instead.

st2common/tests/unit/test_rbac_resolvers_inquiry.py
```diff
@@ -0,0 +1,272 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
```
Kami: Also need to update "rbac_for_supported_api_endpoints" tests and add additional RBAC tests for inquiry API endpoints.

Kami:
I like how you implemented RBAC by utilizing action input parameters :thumbsup:

contrib/runners/inquirer/runner.yaml
lakshmi-kannan: Oh boy, the old syntax is confusing as hell.

st2api/st2api/controllers/exp/inquiries.py
lakshmi-kannan: +1 to kami's comment

m4dcoder:
I'm trying to create my own custom action.  This is what I am experiencing. 

### required value is not boolean
This is because we are using draft 3 of the JSON schema and `required` is a boolean under each properties. However, the default schema in the inquirer is also wrong.

```
---
name: inquire
pack: sandbox
description: Mock action to ask user for more information.
runner_type: inquirer
enabled: true
parameters:
    schema:
        type: object
        properties:
            region:
                type: string
                description: Select a region to deploy resources.
                enum:
                    - us-east
                    - us-west
                    - singapore
                    - shanghai
            environment:
                type: string
                description: Select an environment to deploy resources.
                enum:
                    - production
                    - development
                    - staging
        required:
            - region
            - environment
```

```
ubuntu@cadmus:~/st2$ st2 action create /opt/stackstorm/packs/sandbox/actions/inquire.yaml 
MESSAGE: [u'region', u'environment'] is not of type 'boolean' for url: http://127.0.0.1:9101/v1/actions
```

### schema property cannot be overwritten
So how can user define custom actions w/ schema already defined?
```
---
name: inquire
pack: sandbox
description: Mock action to ask user for more information.
runner_type: inquirer
enabled: true
parameters:
    schema:
        type: object
        properties:
            region:
                type: string
                description: Select a region to deploy resources.
                enum:
                    - us-east
                    - us-west
                    - singapore
                    - shanghai
                required: True
            environment:
                type: string
                description: Select an environment to deploy resources.
                enum:
                    - production
                    - development
                    - staging
                required: True
```

```
ubuntu@cadmus:~/st2$ st2 action create /opt/stackstorm/packs/sandbox/actions/inquire.yaml 
MESSAGE: The attribute "properties" for the runner parameter "schema" in action "sandbox.inquire" cannot be overridden. for url: http://127.0.0.1:9101/v1/actions
```

st2client/st2client/commands/inquiry.py
Mierdin: I'm torn on this. On one hand I agree, consistency is a good idea. However, Inquiries are meant to be short-lived, relatively speaking. If the default limit is 50, then it would almost encourage a regular practice of balancing 50 concurrent executions that are waiting for a response. Even 20 is a bit extreme.

Mierdin:
@m4dcoder Ah - since the `schema` parameter should be passed by value and not interpreted by the action validation code, putting your schema under `default` seems to work. Here's a revised copy that works:

```
---
name: inquire
pack: testpack
description: Mock action to ask user for more information.
runner_type: inquirer
enabled: true
parameters:
    schema:
      default:
        title: response_data
        type: object
        properties:
            region:
                type: string
                description: Select a region to deploy resources.
                enum:
                    - us-east
                    - us-west
                    - singapore
                    - shanghai
                required: true
            environment:
                type: string
                description: Select an environment to deploy resources.
                enum:
                    - production
                    - development
                    - staging
                required: true

```

m4dcoder:
If placed under default, does the schema get checked whether it's malformed or not?

Mierdin:
Not until runtime, when `jsonschema.validate` is called in the respond function. That's not ideal, but the response schema isn't meant to be interpreted as part of the overall action/runner parameter schema, so checking it on action creation doesn't seem right either.

Kami:
@Mierdin Also need to add `allowAdditionalProperties: false` to each property which is an object, otherwise users can specify arbitrary properties which are not part of the schema.

We should also have a look if it's possible to make this a default. By default it allows additional properties and if you want not careful when working on a schema, it's easy to forget to add that (and we had bugs related to that in the past).

Mierdin:
@Kami Which schema are you referring to?

Kami:
@Mierdin Basically any schema where we don't want to allow arbitrary properties which are not part of the schema and doesn't have `additionalProperties: false` defined.

Notably the runner one and iirc there is another one.

Mierdin:
Okay. To be clear, I'm assuming you're talking about the action/runner parameters schema. The actual `schema` parameter shouldn't have its attributes inspected by the action/runner parameter validation code. Having `additionalProperties: false` on that seems like it would defeat the purpose, as the intent is to allow the user to pass an arbitrary schema into that parameter for validating the Inquiry response data

m4dcoder:
So thanks for clarifying that the response schema needs to be placed in `default` field.

However, this highlights an issue that st2 has in general. That is the value of `default` is not validated on action registration. This is actually fixable because the JSON schema for the runner/action parameter is provided here in the action meta file and that can be used to validate against the default value if there is one. 

And this can be applied to the `schema` field of the inquiry runner.  The meta schema for schema is defined at https://github.com/StackStorm/st2/blob/master/st2common/st2common/util/schema/custom.json.  It is possible to use this to define the inquirer's runner parameter for the response schema.  And then if we fixed the issue on validating default value, then on action registration, the default response schema can be validated.

m4dcoder:
On a separate issue, I think the response schema should conform to our `custom` JSON schema. User is already used to defining our custom schema for action parameters.  Otherwise, it'll just confuses users.  Anyway, this is my opinion. But I'm open to using draft4 if the team argues for it.

m4dcoder:
Regarding https://github.com/StackStorm/st2/pull/3653#issuecomment-328651262 on validating `default` value on action registration, it's OK to fix in a separate PR after this. But it should be addressed.

st2api/st2api/controllers/exp/inquiries.py
Mierdin: Hrm, so I would have to create a `st2common.models.api.inquiry` for this - wouldn't I have to also create a DB model for Inquiries, since the API models seem to refer to the DB models?

Mierdin: And what about `st2common.persistence.inquiry`?

Mierdin: Nvm, I think I can see how I can just create a new `st2common.models.api.inquiry`, but it's already taken some tinkering and will require some more. Will take a shot at this and let you know what I come up with.

m4dcoder:
The `st2 inquiry respond` does not meet spec. I thought we decide that it'll interactively ask the questions. How come it's expecting a JSON as required parameter?

m4dcoder:
How come there's create, update, delete for `st2 inquiry`?  If I run the command to delete the inquiry, it tells me the inquiry has been deleted.  But then in actual, it's not.  These operations should not apply to `st2 inquiry`.

m4dcoder:
Setting ttl for an inquiry does not affect the state of the execution. I set ttl to 3 minutes and after that ttl has passed, the inquiry is still pending.

m4dcoder:
Other than RBAC, I played w/ this on a number of scenarios. I prefer the default schema to be validated at time of action creation. Although I can live with it being implemented after this PR is merged. The other issues I raised on `st2 inquiry respond`, unnecessary API/CLI operations on `st2 inquiry`, and the bug w/ TTL need to be fixed prior to merging this PR. I only looked at functionalities this round. I still need to review the code more closely.

Mierdin:
@m4dcoder 

> The st2 inquiry respond does not meet spec. I thought we decide that it'll interactively ask the questions. How come it's expecting a JSON as required parameter?

You're right. I was just focusing on getting things working with passing a JSON object directly in via CLI (which will make sense for those wanting to script responses). I'll change so that the JSON can be provided directly via an optional parameter and make the interactive mode the default, similar to how `st2 login` works.

> How come there's create, update, delete for st2 inquiry? If I run the command to delete the inquiry, it tells me the inquiry has been deleted. But then in actual, it's not. These operations should not apply to st2 inquiry.

Right again; this is not intentional. Will prune so that only `get`, `list`, and `respond` show.

> Setting ttl for an inquiry does not affect the state of the execution. I set ttl to 3 minutes and after that ttl has passed, the inquiry is still pending.

I wrote this in comments but could have been more explicit in the PR description. I implemented the TTL field to get completeness for the API, but it currently does nothing. My plan is to open a second PR to extend `st2garbagecollector` to clean these up after a period of time. I held off on that for now to keep the PR tidy.

st2api/st2api/controllers/exp/inquiries.py
Mierdin: @Kami @lakshmi-kannan Okay I pushed https://github.com/StackStorm/st2/pull/3653/commits/920e88f81777eff5d0e5dbf6f86fe21f1f814415 as a first attempt at this. It seems to work. It's a **bit** unorthodox since I'm only implementing the API models, but I think it's a good first step. Let me know if you see anything heinous.

st2api/st2api/controllers/exp/inquiries.py
Mierdin: Fixed in https://github.com/StackStorm/st2/pull/3653/commits/920e88f81777eff5d0e5dbf6f86fe21f1f814415

st2common/st2common/openapi.yaml
Mierdin: This has been resolved in https://github.com/StackStorm/st2/pull/3653/commits/920e88f81777eff5d0e5dbf6f86fe21f1f814415. In short, the `Inquiry` definition in the OpenAPI spec is now equivalent to the `InquiryResponseAPI` model implemented in that commit.

st2common/st2common/openapi.yaml.j2
Mierdin: This has been resolved in https://github.com/StackStorm/st2/pull/3653/commits/920e88f81777eff5d0e5dbf6f86fe21f1f814415. In short, the `Inquiry` definition in the OpenAPI spec is now equivalent to the `InquiryResponseAPI` model implemented in that commit.

st2client/st2client/commands/inquiry.py
Mierdin: Had to play surgeon on the API side a bit, but finally got the right headers in place, and implemented this table in https://github.com/StackStorm/st2/pull/3653/commits/984fe1f4430930067bdd743594036ead58d705b1

contrib/runners/inquirer/runner.yaml
Mierdin: Fixed in https://github.com/StackStorm/st2/commit/aa7b1d479a439c8502b1a8775534f157b6a35917

contrib/runners/inquirer/runner.yaml
Mierdin: Fixed in fee4118b7

st2api/st2api/controllers/exp/inquiries.py
Mierdin: Got it. Thanks for the background. I removed this TODO note and added the appropriate field to the OpenAPI def in 4135836a2. I'll make sure to test this when I finish up the RBAC unit tests for this API

st2api/st2api/controllers/exp/inquiries.py
Mierdin: Fixed in 66f3fb77a

st2api/st2api/controllers/exp/inquiries.py
Mierdin: Changed in 66f3fb77a

st2api/st2api/controllers/exp/inquiries.py
Mierdin: Ack that. Will leave it for now but will keep this in mind if I hear about any issues. I'd assume the two client payloads would have to arrive in pretty close succession for this to happen

st2api/st2api/controllers/exp/inquiries.py
Mierdin: The catch here was that breaking out of the loop meant you were permitted, so it was difficult to JUST rely on the loop for this. However, I think I carried the spirit of what you're saying into 39a17fa5d - I'm fairly confident the logic is the same but wouldn't mind a sanity check

st2api/st2api/controllers/exp/inquiries.py
Mierdin: If you're talking about the roles list, yes, in the current approach, you would only need to be a member of one of the listed roles. I don't **think** there's a solid use case for specifying that the user must be in ALL of the listed roles but am open to hearing them. For now, the approach is OR (with both users and roles)

st2common/tests/unit/test_rbac_resolvers_inquiry.py
Mierdin: I'm assuming you're talking about something like `st2api/tests/unit/controllers/exp/test_inquiries_rbac.py`, which I have had as an untracked file for about a week now. :smile: Will get on that as soon as I'm done with the existing comments

Mierdin:
@m4dcoder Okay commits have been pushed to fix the two client issues. https://github.com/StackStorm/st2/pull/3653/commits/aa9a437d247f964c924ea1f0854d628dec48a73d moves the existing full-object mode to an optional flag `-r`, and adds an interactive mode for responses, which is the default. https://github.com/StackStorm/st2/pull/3653/commits/310db0e5998038e672b7794021d0cf5b4a718415 removes those extra subcommands.

contrib/runners/inquirer/runner.yaml
Mierdin: Actually I misunderstood this and realized later that I needed to use the built in schema validation code. I did this (and corrected any straggling v4-style required fields) in https://github.com/StackStorm/st2/pull/3653/commits/e618dee8ccb7496431fabbd3b54fc9c60db02029 - hopefully that's a bit better.

contrib/runners/inquirer/runner.yaml
m4dcoder: There's no advantage and you can find other inconsistencies around. But I agree w/ @Kami.  For the most part, we default value to None if no value should be assigned. To me, an empty string is still a value.  Sometimes we default an empty string and empty dict because processing None and processing empty list/dict causes a lot of unnecessary logic. Whereas processing a None on a variable of string type is relatively straight forward.

contrib/runners/inquirer/tests/unit/test_inquirer_runner.py
```diff
@@ -0,0 +1,153 @@
...
+                'tag': "developers",
+                'schema': {},
+                'ttl': 1440
+            }
+        )
+        mock_action_service.request_pause.assert_not_called()
+        mock_trigger_dispatcher.reset_mock()
+        mock_action_utils.reset_mock()
+        mock_action_service.reset_mock()
+
```
m4dcoder: How about test inquiry w/ a grandparent?

contrib/runners/inquirer/tests/unit/test_inquirer_runner.py
```diff
@@ -0,0 +1,153 @@
...
+
+class InquiryTestCase(RunnerTestCase):
+
+    def test_runner_creation(self):
+        runner = inquirer.get_runner()
+        self.assertTrue(runner is not None, 'Creation failed. No instance.')
+        self.assertEqual(type(runner), inquirer.Inquirer, 'Creation failed. No instance.')
+
+    @mock.patch('inquirer.TriggerDispatcher', mock_trigger_dispatcher)
```
m4dcoder: If these mocks are the same for every test cases in this module, you can move the mocks to the class level.  You can then do the mock directly in the mock.patch.object method. Examples: https://github.com/StackStorm/st2/blob/master/contrib/runners/mistral_v2/tests/unit/test_mistral_v2.py#L150

st2api/st2api/controllers/exp/inquiries.py
```diff
@@ -0,0 +1,273 @@
...
+            limit=limit,
+            raw_filters={
+                'status': action_constants.LIVEACTION_STATUS_PENDING,
+                'runner': INQUIRY_RUNNER
+            }
+        )
+
+        # Transform to InquiryResponseAPI model
+        inquiries = [InquiryResponseAPI.from_inquiry_api(InquiryAPI.from_dict(raw_inquiry))
+                     for raw_inquiry in json.loads(raw_inquiries.body)]
```
m4dcoder: This is the first time I've seen that we have to handle json load for a record individually. Can you elaborate?

st2api/st2api/controllers/exp/inquiries.py
```diff
@@ -0,0 +1,273 @@
...
+            return
+
+        if not requester_user:
+            requester_user = UserDB(cfg.CONF.system_user.user)
+
+        # Determine permission of this user to respond to this Inquiry
+        if not self._can_respond(inquiry, requester_user):
+            abort(
+                http_client.FORBIDDEN,
+                'Insufficient permission to respond based on Inquiry parameters.'
```
m4dcoder: `User does not have permission to respond to the inquiry.` is more clear...

st2api/st2api/controllers/exp/inquiries.py
```diff
@@ -0,0 +1,273 @@
...
+        :rtype: bool - True if a valid Inquiry. False if not.
+        """
+
+        if inquiry_candidate.runner.get('runner_module') != "inquirer":
+            abort(http_client.BAD_REQUEST, '%s is not an Inquiry.' % inquiry_candidate.id)
+            return False
+
+        if inquiry_candidate.status != "pending":
+            abort(http_client.BAD_REQUEST,
+                  'Inquiry %s has already been responded to' % inquiry_candidate.id)
```
m4dcoder: Need to distinguish cases where it timed out due to TTL?

st2api/st2api/controllers/exp/inquiries.py
```diff
@@ -0,0 +1,273 @@
...
+        It must use the "inquirer" runner, and it must currently be in a "pending" status
+
+        :param inquiry_candidate: The inquiry to check
+
+        :rtype: bool - True if a valid Inquiry. False if not.
+        """
+
+        if inquiry_candidate.runner.get('runner_module') != "inquirer":
+            abort(http_client.BAD_REQUEST, '%s is not an Inquiry.' % inquiry_candidate.id)
+            return False
```
m4dcoder: After abort, does this method still return False?

st2api/st2api/controllers/exp/inquiries.py
```diff
@@ -0,0 +1,273 @@
...
+
+        if inquiry_candidate.status != "pending":
+            abort(http_client.BAD_REQUEST,
+                  'Inquiry %s has already been responded to' % inquiry_candidate.id)
+            return False
```
m4dcoder: After abort, does this method still return False?

st2api/st2api/controllers/resource.py
```diff
@@ -211,6 +211,15 @@ def _get_one_by_id(self, id, requester_user, permission_type, exclude_fields=Non
 
         instance = self._get_by_id(resource_id=id, exclude_fields=exclude_fields)
 
+        # A bit gross, but it seems necessary at the moment in order to
+        # make RBAC happy, since Inquiries don't currently have their own
+        # DB model
+        # TODO remove this atrocity once InquiryDB is a thing
```
m4dcoder: Please explain how this make RBAC happy.

st2api/tests/unit/controllers/exp/test_inquiries.py
```diff
@@ -0,0 +1,385 @@
...
+        old_user = cfg.CONF.system_user.user
+        cfg.CONF.system_user.user = "foo"
+        post_resp = self._do_create_inquiry(INQUIRY_2, RESULT_2)
+        inquiry_id = self._get_inquiry_id(post_resp)
+        response = {"continue": True}
+        put_resp = self._do_respond(inquiry_id, response)
+        self.assertEqual(response, put_resp.json.get("response"))
+
+        # Clean up
+        cfg.CONF.system_user.user = old_user
```
m4dcoder: test respond to expired inquiry?

st2client/st2client/commands/inquiry.py
m4dcoder: I agree w/ @humblearner here. I prefer consistency.

st2common/st2common/services/action.py
```diff
@@ -315,6 +315,28 @@ def request_resume(liveaction, requester):
     return (liveaction, execution)
 
 
+def get_root_liveaction(liveaction_db):
```
m4dcoder: Unit test for this method?

m4dcoder:
Are there integration tests for the st2cd.st2_e2e_tests?  Please include one for action chain and one for mistral.

st2api/st2api/controllers/exp/inquiries.py
Mierdin: I agree it's a bit wordy; I was trying to distinguish that this message was not coming from RBAC.

st2api/st2api/controllers/exp/inquiries.py
Mierdin: @Kami recommended that I use the timeout status in the GC so yeah I'll add this as a second case once I do that

st2api/st2api/controllers/exp/inquiries.py
Mierdin: It doesn't; I only included this to be a bit more explicit. Elsewhere in this controller and in other controllers, a return statement is included after an `abort()` call, presumably for the same reason. So I was just trying to keep it consistent.

st2api/st2api/controllers/exp/inquiries.py
Mierdin: It doesn't; I only included this to be a bit more explicit. Elsewhere in this controller and in other controllers, a return statement is included after an `abort()` call, presumably for the same reason. So I was just trying to keep it consistent.

st2api/st2api/controllers/exp/inquiries.py
Mierdin: I am trying to get the link to the actual thread that I was in with @Kami and @lakshmi-kannan further up in this PR but Github isn't letting me see back that far right now for some reason.

Basically there are two API models for Inquiries - one that actually satisfies the API definition, and another that has a few extra fields that I use to make decisions (like how [`_inquiry_sanity_check` uses `runner` and `status`](https://github.com/StackStorm/st2/pull/3653/files#diff-361872b73d29fa927dc3081ae5f6c082R190)). Since _get_all has already converted to JSON, and the `InquiriesController.model` is `InquiryAPI`, I needed to convert them all to `InquiryResponseAPI` instances. This seems the most succinct way to do that.

That's a lot of detail that the comment `Transform to InquiryResponseAPI model` doesn't quite cover, so I'll add something more there.

st2api/st2api/controllers/resource.py
Mierdin: As with most strange things I had to do in with RBAC for Inquiries, it all comes down to the fact that Inquiries don't have their own DB model. I also would like to say this is pretty much the only place where that fact made things difficult. For the most part, not having an Inquiries DB model didn't really affect my work. But in RBAC it reared its head a few times.

When I run `_get_one` from the Inquiries controller, the functions in the `Resource` base class pull a resource [directly off of the database by resource id](https://github.com/StackStorm/st2/blob/master/st2api/st2api/controllers/resource.py#L268) **and then** does its own RBAC assertions all before getting back to the Inquiries controller. So without intervention, the resource that makes it into RBAC is of the format `execution:<id>`. That's when my tests start failing:

```
FAIL: test_user_has_resource_db_permission (tests.unit.test_rbac_resolvers_inquiry.InquiryPermissionsResolverTestCase)
  File "/home/vagrant/st2/st2common/tests/unit/test_rbac_resolvers_inquiry.py", line 256, in test_user_has_resource_db_permission
    permission_type=PermissionType.INQUIRY_VIEW)
  File "/home/vagrant/st2/st2common/tests/unit/test_rbac_resolvers.py", line 105, in assertUserHasResourceDbPermission
    raise AssertionError(msg)
AssertionError: Expected permission grant "inquiry_view" for user "custom_role_inquiry_view_grant" on resource DB "execution:59caf55032ed3503f189573a", but no grant was found
```

Since all of that is done before even getting back to the Inquiries controller code, the only option I have is to intercept that resource right before the assertion and make this admittedly duct-tape-esque conditional, replacing the "get_uid" function with this one, so that RBAC calls the correct resolver.

Hopefully that makes sense, and I will certainly improve the comment with some of this detail. And if you know of a better way, do let me know. I realize this is a bit icky.

contrib/runners/inquirer/tests/unit/test_inquirer_runner.py
Mierdin: Since `action_service` is being mocked out for these tests, adding another test for testing this would seemingly prove nothing - the mock will obviously return whatever I tell it to. As you correctly point out, the `action_service.get_root_liveaction()` could use some dedicated testing, so I'll test for complex inputs like this there.

st2common/st2common/services/action.py
Mierdin: Ack that, and as mentioned in a [previous comment](https://github.com/StackStorm/st2/pull/3653#discussion_r141243809) I will test for a variety of parent depths to ensure this works as desired.

contrib/runners/inquirer/tests/unit/test_inquirer_runner.py
Mierdin: Ack, fixed in 479a997fa

st2api/st2api/controllers/exp/inquiries.py
Mierdin: I addressed this in what I think is a better way in be287e75c. In short, I got rid of the extra `from_dict` function. It was basically `from_model`, just without the call to `cls._from_model`. So I just went with a sort of hybrid function `from_model` function that allows this step to be skipped, and that way I can convert these `InquiryAPI` JSON objects directly into `InquiryResponseAPI` instances directly, simplifying this comprehension quite a bit.

st2api/st2api/controllers/exp/inquiries.py
Mierdin: Improved in ae4a81ad4

st2api/st2api/controllers/resource.py
m4dcoder: @Kami This implementation here don't look right. I need to review the RBAC code for this. Have you looked at this?

st2api/st2api/controllers/exp/inquiries.py
m4dcoder: So I would make this function either return True or False. If we're aborting on check failure, then by not erring/aborting, this mean check succeed. I know you're trying to keep it consistent w/ other functions. My guess is those functions are not implemented correctly if that's the case.

st2api/st2api/controllers/exp/inquiries.py
m4dcoder: This is a potential performance problem to have to iterate thru the record and do a json loads...

st2api/st2api/controllers/exp/inquiries.py
Mierdin: Okay. Another reason this was done was to raise different error messages depending on the different failure scenarios. I could move the `abort()` call back outside the function so the function always returns True/False, and also return the error message if any, so I can still provide feedback to the user.

st2api/st2api/controllers/resource.py
Kami: It looks like you could perhaps re-implement `_get_one_by_id` in your controller class and that way we can avoid this "hack" here?

Would that work?

st2common/st2common/rbac/resolvers.py
```diff
@@ -1047,6 +1048,76 @@ def user_has_permission(self, user_db, permission_type):
...
+            return True
+
+        resource_types = [ResourceType.INQUIRY]
+        permission_types = [
+            PermissionType.INQUIRY_VIEW,
+            PermissionType.INQUIRY_RESPOND,
+            PermissionType.INQUIRY_ALL
+        ]
+        permission_grants = get_all_permission_grants_for_user(user_db=user_db,
+                                                               resource_types=resource_types,
```
Kami: Since no resource_id is passed here it looks like those permissions should be global (don't apply to a specific resource) if that's not already the case.

Kami: Actually, I just saw the code below. I see the approach you went with.

Another approach would be to make those permission types global (don't apply to a specific resource), then you could get rid of that if len check below.

st2api/st2api/controllers/exp/inquiries.py
Mierdin: Done in 577d019f0

st2api/st2api/controllers/exp/inquiries.py
Mierdin: Fixed in 577d019f0

st2api/st2api/controllers/exp/inquiries.py
Mierdin: I added this conditional when cleaning up the sanity check function in 577d019f0. Also added a test for it

st2api/st2api/controllers/resource.py
Mierdin: That works. I changed this in 5596bfa9e, so the scope of the hack is limited. I also improved the comment so it was more obvious why I went this way

st2api/tests/unit/controllers/exp/test_inquiries.py
Mierdin: Among other things, I added a test for this in https://github.com/StackStorm/st2/commit/577d019f0ebd54af33d22a8d108674710e9e8748

st2common/st2common/services/action.py
Mierdin: Added in 5c9be2438

contrib/runners/inquirer/tests/unit/test_inquirer_runner.py
Mierdin: Added test for the underlying function in 5c9be2438

st2common/st2common/rbac/resolvers.py
Mierdin: This was mostly because Inquiries are still just executions under the hood so I wasn't sure of a better way. I'll see if I can convert the permissions to a global type and get rid of this hack.

Mierdin: I took a stab at this in 7d828a2f5. It seems to work (at least the tests are happy, will do some additional testing later) and I'm very happy to `s/inquiry:ask/inquiry` everywhere but I'm not confident that I satisfied your statement: `you could get rid of that if len check below`. For now, I'm assuming by that you meant getting rid of the list comprehension, since there are other resolvers that do a `len()` check, just in a much simpler way (just checking total len of `permission_grants`). Let me know if I missed something there.

Mierdin:
Quick update on some somewhat major changes:

- (in 545e3b66e) The trigger definition for Inquiries now **only** includes the Inquiry ID. Also fixed a bug in that commit where the ID being placed in the trigger payload was accidentally the liveaction ID. Corrected this to use the ActionExecution ID (which is equal to the Inquiry ID), and fixed test to enforce this
- (in d17ad68f5) Renamed all instances of the `tag` parameter to `route` to keep things more consistent with notifications

armab:
Congrats! :clap:

Issue #2175, opened 2015-11-05
[PROPOSAL] [RFC] Streaming action output and other intermediate results


Kami:
Multiple people have asked for this in our community so I'm bring this proposal back to life (I'm actually moving it from a private repo to a public one so other community members can participate as well).

Keep in mind that this proposal is almost 11 months old and a lot has change then. The biggest change which can affect this proposal is that now we have a /stream endpoint and mechanism for streaming data.

I will explain differences between my original proposed approach and utilizing stream for that in a new comment. In short, each approach has it's own advantages and disadvantages, but I personally think the original proposal is still better since it provides intermediate result (if we go with stream, user will only be able to receive new output / data, but if we go with my proposal, API endpoint will return everything which has been buffered so far).
## Problem

Currently if you run a long-running action or an action which consists of multiple steps (action-chain or a work-flow), you only know if an action is scheduled, running or finished, but you have no visibility into what is going on with the running action until it completes (either succeeds or fails).

Good examples of a long-running actions include building and compiling stuff,d running tests, etc.
### Use cases
1.  Long running shell or python actions: replacing Jenkins - action is a build, produces long output, a minute, user wants to see chunks of the output, rather than wait for full completion.
2. Long running workflow - ActionChain, especially Mistral. The "workflow" action execution should show workflow progress and the tasks as they change state (new tasks scheuduled, running, complete...).
3. Short tasks - majority of the cases - don't bother to do intermediate output, wait till complete. 

For this purpose, define:  short < 5 sec < Long. 
## Goals

The goal is to increase the visibility into running actions and stream output of the running actions and other intermediate results. Additionaly, we want to support better integration with Mistral.

With "output" I'm referring to the action's stdout and stderr.  With "other intermediate results" I'm referring to `ActionExecution` and potentially other intermediate objects (@enykeev is already working on the second part).

I separated those two because I think they are two different things. Output is opaque to us and we have no idea how it looks so we should treat it as such. This means streaming it as soon as it us available. For performance and conveniece reasons we should probably still buffer it. For line-delimited stuff we could simply send a line as soon as it's available.

And as far as "other intermediate results" go - here we know the structure and what is an atomic unit which means we can stream those things as a whole as new line delimited JSON (one JSON serialized object per line).
## Implementation Proposal

[...]

PR #3657, opened 2017-08-08
Real-time streaming output for Python runner actions

Kami:
This pull request finally starts implementation for #2175.

Please don't comment on the actual implementation yet because it's very much work in progress. Just comment on the higher level interface and open questions.

## Description

This pull request adds support for consuming and storing stdout and stderr for Python runner actions in real-time.

stdout and stderr for Python runner actions is consumed as it's available (aka when action produces it) and then stored inside two special MongoDB collections.

This allows users to see and consume output for Python runner actions in real-time. This is very handy for long running actions since you can immediately see what is going on compared to needed to wait (potentially for a very long time) for action to finish.

It also makes the whole common "CI" pattern much more useful since output is available immediately.

## Implementation Details

Implementation is very similar to my original proposal in #2175. stdout and stderr is consumed as available and stored in special MongoDB collections.

This output can then be consumed in real-time either by using two of the blocking / long open API endpoints or by utilizing the event stream. Event stream will probably be utilized by WebUI and CLI, but raw API endpoint access comes handy when you don't have CLI or WebUI available and you simply want to "curl" the endpoint.

Because this could potentially result in a lot of data and large collection, I will update garbage collector service which will also allow operator to set rules to periodically purge those new collections.

# Open Questions

* This feature should probably be first shipped behind a st2.conf config feature flag (aka disabled by default). Down the road it probably makes sense to enable it by default, but allow user to disable it on per action execution invocation basis (or perhaps leave it disabled by default and allow user to enable it on per invocation basis).

## TODO

- [x] API endpoints
  - [x] `/v1/executions/<execution id>/stdout`
  - [x] `/v1/executions/<execution id>/stderr`
- [x] Update `store_stdout,stderr_line` function to also publish an event on the message bus
- [x] Update /v1/stream API endpoint to handle new events
- [ ] ~Allow user to enable / disable this real time output capture via config option / on per invocation basis~ (later on once it's enabled by default)
- [x] Put it behind a feature flag (opt-in, off by default)
- [x] Tests
- [ ] Documentation
- [x] `st2 execution tail <execution id>` CLI command
- [x] Support for garbage collector to periodically empty those two new collections
- [x] RBAC
- [x] Support for local runner
- [x] Support for remote runner

## Related PRs

Those PRs started as changes in this branch, but they were split out to make it easier to review and make sure they get included in v2.4.0.

- https://github.com/StackStorm/st2/pull/3676
- https://github.com/StackStorm/st2/pull/3677


Kami:
Right now I'm adding those blocking / streaming API endpoints to `st2api` and we will lunch this feature as experimental.

If it turns out that performance is not so great, we can move it to `st2stream` and handle request path proxying on nginx level.

Kami:
## Security Implications

Almost every change / feature we add to the code has some kind of security implications and this one is no different so I want to point out those implications.

Action stdout / stderr can potentially contain sensitive data so ability to view / stream this data can expose sensitive information. This information is already available right now when the execution finishes on the execution db object and each user which has `EXECUTION_VIEW` permission can view it.

This means that this change doesn't expose any new information, it just adds additional vector to retrieve / view this information. Right now it's behind the same RBAC permission type wall (`EXECUTION_VIEW`) as "get one execution", but if it turns out that it's better to have a separate permission type for it, we can add it.

Having said that - it's nothing new (and some could also argue that it's common sense), but we of course need to document this well so there won't be any surprises.

As mentioned above, this feature will first also be opt-in and shipped behind a feature flag.

Kami:
As far as stability / production readiness goes - I don't expect it to be production ready (turned on by default) until v2.5 / v2.6.

I do think working prototype which is opt-in (off by default) will likely be ready for v2.4, but the feature is quite complex and touches a lot of code so it will need a lot of testing (I will of course spend a lot of time on automated tests, but some things are hard to automate so we should also test it on our infra and run it there for a month or so).

Kami:
Already, both API endpoints and ``st2 execution tail [--stdout] [--stderr] <id>`` CLI command has been implemented and are far enough so I can show a simple demo.

By default if ``--stdout`` and ``--stderr` flag is not provided, the CLI command will tail both - stdout and stderr.

For convenience, you can also pass ``last`` for execution id argument and the command will automatically tail last execution - this comes very handy during testing - you can just use ``last`` instead of needing to always find and copy and paste last execution id.

Here is a short recorded demo - first I show how to utilize new API endpoints and later I show how to utilize new CLI command - https://www.youtube.com/watch?v=f9RZUG6QPAM&feature=youtu.be (also available in 4K - sadly I had some issues with recorder crashing so I couldn't just record a single window, but needed to record the whole desktop).

Example action I wrote for demonstration purposes can be found here - https://github.com/StackStorm/st2/blob/python_runner_actions_real_time_output/contrib/examples/actions/print_to_stdout_and_stderr.py#L22

Kami:
While working on this I also noticed, ``/v1/stream`` API endpoint is not behind RBAC wall so I fixed that.

Now ``STREAM_VIEW`` global permission type is required to be able to view ``/v1/stream`` API endpoint. As this is a "view" permission type it means ``observer`` system role also grants this permission.

Kami:
Alright, this PR is not fully finished yet, but it's far enough to get a basic review.

Having said that, there are still some open questions:

## 1. How to handle output from the remote runner?

Remote runners are "unique snowflakes" because they allow user to run script / command on multiple hosts. Something I never liked because it breaks the runner model and rest of the runners work differently. I would much prefer to have a single action invocation on a single host, because the same end result could be achieved by simply running multiple actions (one per host) and this wouldn't break the runner and execution abstraction, but we are where we are now.

The problem is that "primary key" for new stdout and stderr objects is execution id. That is, stdout and stderr is accessed by execution id which makes total sense. The problem is how to handle that for remote runners where result could potentially contain stdout and stderr for multiple hosts:

Here are couple of options from top of my heads:

1. Utilize exactly the same approach as we do for other runners, but prefix "line" (stdout / stderr) line DB model attribute with the hostname (e.g. `[hostname] <actual line>`). This follow the abstraction, but it means that the actual "line" is not a "raw" stdout / stderr line so user would need to parse it or similar if they wanted access to raw data.

2. Make primary key a composite one - execution id and hostname. I'm personally not a fan of this approach, because it makes remote runner implementation different than other ones and breaks the whole execution and runner abstraction.

If we went with this approach, we would also need to figure out how the API would look like - e.g. `/v1/executions/<execution id>/stdout/[hostname]`. And this would only apply to remote runner based executions - different API for different runner type is something I would like to avoid, but it might not be possible.

## 2. Enabling / disabling storing and streaming of the output on per action invocation basis

In the first iteration we will only support output streaming for Python runner actions (and remote runner actions soon thereafter). For a major release or two, the feature will be behind a feature flag and disabled by default (opt-in).

The question is if we should support per-invocation enabling / disabling of storing / streaming the action output.

The idea is that storing output adds some overhead (we need to write every line to the database) and means wasting resources in case you are not actually interested in some action output in real-time.

So feedback on those open questions is welcome, and something I would like to address soon.

/cc @StackStorm/team

Kami:
@bigmstone That's one of my issues of how runner search path works right now - b741d7e2eb6772a9b80bf5f3d16f79b84beacd5d

Stuff either needs to be in the runner module (makes for big runner module files) or in st2common. I would much prefer to make runner directory a package and have it inside the runner directory. This way we don't need to keep it inside a single file and "abuse" st2common.

I believe we already discussed this in the past and one thing we would need to do is prefix runner modules with runner name so they wouldn't clash with other modules (e.g. `from utils import foo` vs `from runner_name.utils import foo`).

enykeev:
I think that if remote ssh breaks the abstraction, we should clearly say that it won’t get real time results because A and B and create another action that follows the sane logic. At the end, I still think that the way our core.local and core.remote actions work does not scale and we should not became too attached to them in a long run.

As of right now, I’ll be perfectly happy if we get real time results to work on python runner only

Kami:
@enykeev suggestion give me an idea about the approach I haven't though about yet - one approach which wouldn't break the model would be to only support streaming stdout and stderr with remote runner actions when len(hosts) == 1.

I think this would be a good compromise and wouldn't break the abstraction - user still achieve the same end result if needed (invokes action multiple times with a different host each time).

bigmstone:
My thinking lines up with @enykeev here - on both real time results as well as our attachment to them long-term.

I can appreciate how disruptive changing those runners can be, but as it sits not it's not overly explicit from an ST2 context. What I mean is the action is "throwing the execution pig over the fence" and you lose some visibility that would typically be kept in the workflow. If you had to lay out the commands that are run, or the hosts that its executed on (even in a loop) it is more explicit to the consumer of that action than how it is set up currently. We still treat the remote runners as we do any other - atomic - but by its nature it's not atomic. Which makes this hard and I'm also okay with deprecating that in the future to get back to the atomic runner model. So This is the long winded way of me saying I agree with @enykeev.

As for runners @Kami, I think turning runners into a package is a good idea, but when we added `runners` to `contrib` we began abusing that folder. Runners folder is the only folder in there that's not a pack - not to mention entire ` contrib` folder isn't a package. So it might make sense to move runners out into root and make it a package as you mention, but we should kill two birds w/ one stone in that move which is to think through how runners can/will be extended by the end user. To be clear - I don't think we need to have that solution 100% ready, but we should at least have a vision for how it will manifest itself in the future so that hopefully we can put it in a place that reduces the likelihood of it needing to move again.

Kami:
Alright, this functionality has now also been implemented for local and remote runners.

This means it's now available for:

  - local command runner
  - local script runner
  - remote command runner
  - remote script runner
  - python runner

I still need to work on more tests and docs, but if you feel adventurous you can already try it out (don't forget to set `actionrunner.stream_output` config option to True since it's disabled by default).

You can test it using arbitrary commands, scripts and actions, but there are also some actions for testing it out:

* examples.local_command_runner_print_to_stdout_and_stderr
* examples.local_script_runner_print_to_stdout_and_stderr
* examples.remote_command_runner_print_to_stdout_and_stderr
* examples.remote_script_runner_print_to_stdout_and_stderr
* examples.python_runner_print_to_stdout_and_stderr

And here is a short demo of using it with the local runner (it's the same end result when using remote runner) - https://www.youtube.com/watch?v=D4Ai4ImVsh4&feature=youtu.be

st2common/st2common/util/green/shell.py
```diff
@@ -70,54 +72,98 @@ def run_command(cmd, stdin=None, stdout=subprocess.PIPE, stderr=subprocess.PIPE,
...
+                read_stdout_thread.kill()
+                read_stderr_thread.kill()
+
+    LOG.debug('Setting up process and callback.')
     timeout_thread = eventlet.spawn(on_timeout_expired, timeout)
-    LOG.debug("Attaching to process.")
-    stdout, stderr = process.communicate()
+    LOG.debug('Attaching to process.')
+
+    if read_stderr_func and read_stderr_func:
```
m4dcoder: There's two read_stderr_func here.  Should it be `if read_stdout_func and read_stderr_func`?

st2common/st2common/util/green/shell.py
```diff
@@ -70,54 +72,98 @@ def run_command(cmd, stdin=None, stdout=subprocess.PIPE, stderr=subprocess.PIPE,
...
+        LOG.debug('Using real-time stdout and stderr read mode, calling process.wait()')
+        process.wait()
+    else:
+        LOG.debug('Using delayed stdout and stderr read mode, calling process.communicate()')
+        stdout, stderr = process.communicate()
+
     timeout_thread.cancel()
     exit_code = process.returncode
 
+    if read_stderr_func and read_stderr_func:
```
m4dcoder: There's two read_stderr_func here. Should it be if read_stdout_func and read_stderr_func?

Kami: Good catch.

To clarify - it actually doesn't change the behavior because the code inside the if does the right thing and both arguments are required (above in the function you can see we throw if only one is provided).

In any case, I will still of course fix it :)

Kami:
@m4dcoder 

> Why are we creating new DB collections for the ActionExecutionStdoutOutput and ActionExecutionStderrOutput? Why not stream the stdout and stderr into the ActionExecution.result? stdout and stderr is shell/process specific

It's mainly for performance reasons. It's much more efficient to use a special dedicated collection as a "write ahead / only log" (we only ever insert / append to that collection) compared to constantly concurrently updating an attribute or a document in an existing collection (that's very in efficient in most of the database, but especially in the MongoDB because of the way locking is performed).

> stdout and stderr is shell/process specific. For other runners like Action Chain and Mistral (if and when we want to support streaming for them) does not fit into this stdout/stderr model.

I don't think it makes much sense to support streaming for AC / Mistral workflows, because in the end, those workflows call StackStorm actions which use one of the runners which supports streaming.

And the goal for that was also "action" and not "runner" output streaming so I don't really don't know how AC / Mistral output streaming would look like. I guess this would be some debug / meta runner output, but that's already available in action runner and Mistral logs.

m4dcoder:
> I don't think it makes much sense to support streaming for AC / Mistral workflows, because in the end, those workflows call StackStorm actions which use one of the runners which supports streaming.

> And the goal for that was also "action" and not "runner" output streaming so I don't really don't know how AC / Mistral output streaming would look like. I guess this would be some debug / meta runner output, but that's already available in action runner and Mistral logs.

I disagree here. Action Chain, Mistral, and even st2.ask actions are as much an action as they are python, shell, etc.  They have output as well.  For workflows in general, the streaming output of the underlying tasks are different than the output of the workflow itself.  We should be generic in handling streaming of these output, which in these cases, the output is written into the result of the action execution.

m4dcoder:
Copied from Slack so there's information for folks when reviewing.

1. We should not discriminate runners.  If we can stream output for 1 type of runner then why not other types. It's not about what is the schema of their output and what is the content of their output. It doesn't matter if they are workflow runners (atm, these are the only outliers). What happens if we have other runners in the future that doesn't output in stdout or stderr?

2. Also the other disagreement is that the result of `st2 execution get` doesn't show the streamed result. I think it should return a snapshot of the streamed result. The argument was that user can use the UI or the new `st2 execution tail` command. But we have already trained users for many version to use `st2 execution get` and this creates the perception that the data between the two CLI commands are out of sync.

contrib/runners/local_runner/local_runner.py
```diff
@@ -31,7 +32,10 @@
 from st2common.util.misc import strip_shell_chars
 from st2common.util.green import shell
 from st2common.util.shell import kill_process
-import st2common.util.jsonify as jsonify
+from st2common.util import jsonify as jsonify
```
bigmstone: Still need as?

Kami: lol yeah, good catch - we don't :)

Kami:
We had a lot of discussion about this on Slack, so to sum it up and add my most recent thoughts here.

@m4dcoder's main issue with it is that it's not generic enough because it discriminates some runners / is biased towards some - notably that we don't support action-chain and Mistral runner.

I do understand his point of view, but I don't think that's a big issue because there is a little need for "streaming output" for those two runners. First, we don't really have a strict definition for "output" for those runners.

One definition could be meta information about the runner (what's happening inside the runner - we already have that in the action runner log file) and the other could be information about which tasks are running, etc. (that's also already available via the event stream since we publish each action execution state change there).

So from my point of view, we already have this "problem" solved better and it makes a lot more sense to utilize action executions published to the event stream for workflow runners meta info (which tasks, are running, etc.) as opposed to trying to force it on the output model (imo, that's not output, that's runner / workflow meta / log information).

Having said that, I could probably make everything more generic and come up with a solution which @m4dcoder would accept. This solution would involve only having one model (e.g. "ActionExecutionOutput") with a "chunk" field or similar. We would still need to decide what is an atomic piece of output for each runner though - line or similar.

As you can see, this approach would still expose some differences between the runners (what is a chunk, etc.) and that's really a fact we can't avoid - runners are different, especially workflow ones, so sometimes there is no one size fits all solution and no way to hide that.

My main problem with this approach would be that we would lose distinction between stdout and stderr because we would only have one stream. So personally I think that's not worth it.

This PR solves a very specific problem (one many users asked about). Instead of solving an actual problem, making it generic would make the end solution worse (no distinction between stdout and stderr streams) and basically we would make it generic just for the sake of it - no one asked for action chain and Mistral workflow out and as mentioned before, "meta" information 
@m4dcoder sees as workflow runner output is already available in the event stream. So again, I think making it more generic would be a lot more complicated for a little gain and solving a problem which doesn't exist / no one asked about.

@bigmstone also proposed some things, but from what I understand (please correct me if I'm wrong), his actual core solution and problem would be more or less the same, the only different thing would be the API - using query param for filtering instead of two separate API endpoints, potentially storing more information on DB model so we would have one instead of two models.

I would like to move on with this as soon as possible, because it has been stuck in limbo for 2+ weeks.

lakshmi-kannan:
@Kami @m4dcoder IMO, the streaming feature is best implemented for simple runners only. I am also reasonably sure that for some runners, we will never be able to implement streaming. "Streaming" requires everything to work async. Not all code is written that way. For example, HTTP runner- it's an all or nothing runner. I could probably come up with other examples.

Also, we will learn new things as we build this feature. Since this is behind a feature flag, we can decide to change behaviors (we have to be upfront in docs that this feature is still alpha/beta). I am more worried that we have fit everything into "stdout" and "stderr" streams which is rather limiting. Maybe we should just call it "output" and "error"? I don't know. This will also confuse users. I think we can go with what we have and collect some early feedback. I am going back to enykeev's original comment. We should have done this only for python runners and tested it with users. It's too late for that now.

m4dcoder:
I agree that supporting output streaming is decided case by case. I like to see a more general design that allows us to quickly support output streaming for other runner(s). We just going to end up redesigning. Again, I don't like to hardcode stdout/stderr in the code.  I don't like that we default the chuck to a line.  Maybe ActionExecutionResult with additional column to tag the streaming channel?  In this case, each entry has a tag that says it's stdout or stderr.  Instead of line, change the field to chunk.

Mierdin:
Finally caught up on the discussion here and in Slack. Two main thoughts here, most of which mirrors other already shared opinions:

To me, if we take a step back and focus on the key use case, which is to get visibility into what running actions are doing, I see no reason to even expose `stdout` or `stderr` in any way from an API perspective. Those are runner/action implementation details to me. The API should only expose `output`. `stdout` is basically the same thing anyways, and I only care about `stderr` when things have gone wrong, and in that case, the vast majority of the time, the action will be failed anyways. So my vote is that we leave `stdout` and `stderr` out of it entirely, even as a filter, and **only** after being presented with a compelling use case, add metadata and filtering capabilities (i.e. `/output?filter=stderr`). I don't see it right now.

Also, since we're past the point where this is only a beta feature being tested out on one runner type, I think there's an implication that this feature should be available everywhere. It doesn't make much sense to me that this would only be available for a subset of runners. It would be different if we explicitly said "this is only for the Python runner for now", but since we're beyond that, I would lean towards making it available to all runners unless there's a hard technical reason why not.

Kami:
@lakshmi-kannan @m4dcoder @bigmstone @Mierdin Thanks again to everyone for their feedback.

I'm happy we were able to find good middle ground / compromise everyone is (mostly) happy with :)

The changes we discussed have been made so the PR is ready for a final review and hopefully approval and merge soon.

As discussed, other things like UX, etc. can be changed and improved later on if needed and shouldn't be a blocker to get this initial version into master.

st2api/st2api/app.py
```diff
@@ -68,6 +69,8 @@ def setup_app(config={}):
     app = router.as_wsgi
 
     # Order is important. Check middleware for detailed explanation.
+    app = StreamingMiddleware(app, path_whitelist=['/v1/executions/*/output*',
+                                                  '/v1/executions/*/stderr*'])
```
m4dcoder: These endpoints correspond to changes in the the single output collection in the database?

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -286,6 +291,84 @@ def get(self, id, attribute, requester_user):
...
+            yield six.binary_type(output.encode('utf-8'))
+
+        def new_output_iter():
+            def noop_gen():
+                yield ''
+
+            # Bail out if execution has already completed
+            # TODO: Should we also bail out in PAUSE? Perhaps not a bad idea to avoid potentially
+            # very long opened connection
+            if execution_db.status in action_constants.LIVEACTION_COMPLETED_STATES:
```
m4dcoder: Yeah, definitely should bail out in PAUSE since the execution is not active.

st2client/st2client/commands/action.py
```diff
@@ -1348,5 +1350,71 @@ def run_and_print(self, args, **kwargs):
...
+        else:
+            execution = self.manager.get_by_id(execution_id, **kwargs)
+
+        if not execution:
+            raise ResourceNotFoundError('Execution  with id %s not found.' % (args.id))
+
+        # Execution has already finished
+        if execution.status in LIVEACTION_COMPLETED_STATES:
+            print('Execution %s has completed.' % (execution_id))
+            return
```
m4dcoder: So if execution is completed, this tail command will print this message and exit?  I'm kinda expecting this command to work like the linux `tail` command.  So why can't we just display the data in the ActionExecutionOutput and if it's already been garbage collected, then from the final result?

st2common/st2common/garbage_collection/executions.py
```diff
@@ -94,6 +104,23 @@ def purge_executions(logger, timestamp, action_ref=None, purge_incomplete=False)
     else:
         logger.info('Deleted %s liveaction objects' % (deleted_count))
 
+    # 3. Delete ActionExecutionOutputDB objects
+    to_delete_exection_ids = [str(execution_db.id) for execution_db in to_delete_execution_dbs]
+    output_dbs_filters = {}
+    output_dbs_filters['execution_id'] = {'$in': to_delete_exection_ids}
+
+    try:
+        deleted_count = ActionExecutionOutput.delete_by_query(**output_dbs_filters)
```
m4dcoder: Why not reuse purge_execution_output_objects from below?

m4dcoder:
How come st2common/util/green/shell.py uses eventlet to process stdout/stderr but then the paramiko ssh client is fetching them serially?  Can they share common lib/funcs to read from the stdout/stderr?  I don't know enough about paramiko so I'm asking.  But I'm seeing different implementation for them and curious why.

st2api/st2api/app.py
Kami: Good catch - still need to remove /stderr one.

st2api/st2api/controllers/v1/actionexecutions.py
Kami: k, will make that change.

st2common/st2common/garbage_collection/executions.py
Kami: I agree - will make this change.

This was written before that function, first I didn't allow output objects to be purged independently from the execution ones.

st2client/st2client/commands/action.py
Kami: It's worth noting that this tail is a bit different because there is no concept of "file has completed" (no one will ever write to it again or similar) on file system.

Having said that, I do think that makes some sense and I was thinking about it before so I will make that change.

In case execution has completed, I will hit the `/output` API endpoint to retrieve whole result (for real-time stuff, it uses event stream API).

Kami:
> How come st2common/util/green/shell.py uses eventlet to process stdout/stderr but then the paramiko ssh client is fetching them serially? Can they share common lib/funcs to read from the stdout/stderr? I don't know enough about paramiko so I'm asking. But I'm seeing different implementation for them and curious why.

Not exactly sure I understand that.

Both of the approaches are async and sadly we can't share the same functions to read the response because it's read in a different manner from a subprocess and from a paramiko channel.

The "difference" is that for subprocess we use "blocking" wait and as such we need to read the output in an eventlet, but inside the paramiko code we have a while loop where we read data when it's available).

We could also change the subprocess to the "while loop" approach, but it would mean additional code.

st2api/st2api/app.py
Kami: c18d9ee476c67d2194364ee9fea20ddeb0d767d9.

st2api/st2api/controllers/v1/actionexecutions.py
Kami: ac6c00a129e1dc29f8a84a4fab33b94ff3382cd5

st2client/st2client/commands/action.py
Kami: 7e579bb22dc0ff957b7ce24727beceac4fbb4b21

st2common/st2common/garbage_collection/executions.py
Kami: Actually, ignore that - can't make that change since the code doesn't do exactly the same thing.

The `purge_execution_output_objects` function itself purges based just on the timestamp and the `purge_executions` one purges output objects which belong to the executions which are to be purged.

In theory, the end result set should be similar, but it's safer this way to ensure consistency - so we don't have orphan output objects without executions laying around.

Kami:
@m4dcoder I believe I addressed all of your comments.

> The feature is not complete IMO.

I agree with that - rarely any feature is ever fully complete :P

Having said that, we definitely need to eventually integrate it into WebUI and of course we can adjust CLI based on the actual user feedback.

m4dcoder:
@Kami Thanks for making the changes to `st2 execution tail`.  I think it'll be easier on the users.  As for reading stdout/stderr from subprocess and paramiko, thanks for the explanation. I figure it's something different between them which leads to different implementation albeit the same result.  👍

Issue #3490, opened 2017-06-19
packs - ability to share code between actions/ and sensors/


nmaludy:
There has been some discussion on Slack about the following so i decided to create an issue to track requirements and investigations.

**Use Case:**
- There is a pack deployed that has some code in `actions/lib/common.py` that implements some common functionality
- Within this pack an action exists `actions/example_action.py` that utilizes the common code in `actions/lib/common.py`
- Within the same pack a sensor exists in `sensors/example_sensor.py` that also needs to utilize the common code in `actions/lib/common.py`

**Problem:**
- This common code is not available to both the sensor and action at the same time within the same pack

**Ideas:**

1. Place the code in some "common" area (`/opt/stackstorm/common/xxx`) and setup symlinks to the code in both the `/opt/stackstorm/packs/<pack_name>/actions/lib` and `/opt/stackstorm/packs/<pack_name>/sensors/lib` directories.
2. Place the code into a "common" python module that can be installed via `pip` using `requirements.txt` (this can support both public modules and installing from private git repos)
3. Something else (may allowing packs to access code from other packs?)

cognifloyd:
Here's what I'm doing right now:

cognifloyd:
Here's what I'm doing right now:
```
/opt/stackstorm/packs/<pack>/
  <pack>/
      __init__.py
      common_foobar.py
  actions/
    lib/ <= directory
      __init__.py <= symlink to ../../<pack>/__init__.py
      common_foobar.py <= symlink to ../../<pack>/common_foobar.py
      action.py <= not shared, not a symlink
  sensor/
    lib/ <= directory
      __init__.py <= symlink to ../../<pack>/__init__.py
      common_foobar.py <= symlink to ../../<pack>/common_foobar.py
      sensor.py <= not shared, not a symlink
```

andrew-regan:
Also see https://github.com/StackStorm/st2/issues/2220

cognifloyd:
I looked into using relative links in requirements.txt to grab some lib folder in a different directory, but pip doesn't support relative links in any sane fashion right now (see pypa/pip#328 which is fixed partially by pypa/pip#4208, but is not in any released version yet).
It looks like, if the `common_lib` dir (or whatever you choose to call it) is included in requirements.txt as `./common_lib` (not `./common_lib#egg....`) then it should work (I haven't tested yet, I'm just noting my thoughts).
[...]

Issue #2220, opened 2015-11-18
Feature Request: Standard mechanism to share python code within a pack


tim-ireland:
Often times actions and sensors might need to share logic that would ideally not be duplicated within a pack, i.e., support DRY principles.  Today, the action folder has the concept of a lib directory that is accessible to all actions to import from, however the sensors folder does not have access to this import location (in a standardized way).  

_First Example:_ When using 3rd party python bindings to establish a session to a remote api, the code for connection logic needs to be implemented in both the sensor which receives events and the actions which makes changes in the api.  Ideally they would both import from the same shared module.

_Second Example:_ Common functions and type definitions etc should be able to be shared between the different modules within a pack.

_Suggestion:_ Move the lib folder location from within actions to the top level of the pack and have that be the standard way to share code within a pack.

Kami:
Thanks for the feature request / proposal.

I do agree that something like that would indeed be useful (repeating and duplication is bad!). It's in fact something we have already talked about in the past, but we didn't have the chance to do it yet.

I personally like the "pack level" lib/ directory approach. We don't have an ETA for this feature yet, but if others agree that something like this would be useful, we should be able to do it relatively fast since it shouldn't require too many changes. There are some caveats (we also support copying lib/ directory over for remote runner actions and there are is some backward compatibility stuff we need to keep in mind), but it shouldn't be too bad.

bri365:
@Kami any idea if this proposal is moving forward?  Its beginning to account for a vast number of TODOs in our code :-)

Kami:
@bri365 Hey, sorry, we still didn't get a change to look into this yet.

I will try to look into it and come up with some easy and simple 80% solution (e.g. maybe ignore remote runner to begin with and only solve it for python runner actions) once I'm done with improved pack configuration (probably some time next week).

In the mean time - another potential solution would be to distribute common code as a Python package which gets installed into your pack virtual environment (I do know that's not always the best approach though).

bri365:
Thanks, @Kami, much appreciated!
We do package up some of our shared code (mainly code shared between packs at this point).

nmaludy:
+1

ytjohn:
For me, I do create "common" code for my stackstorm actions and sensors. This code is often highly integrated with the stackstorm environment. 

[...]

PR #3658, opened 2017-08-08
Common pack lib for sensors and actions

lakshmi-kannan:
closes: #3490, closes: #2220 

### Common lib code in /opt/stackstorm/packs/examples/lib/
```
(virtualenv)vagrant@st2dev /m/s/s/st2 ❯❯❯ cat /opt/stackstorm/packs/examples/lib/base.py                 issues_3490/pack_common_code ✭ ✱ ◼
import os

def get_environ(env_var):
    return os.environ.get(env_var, None)
```

### Using the common lib code in an action
```
(virtualenv)vagrant@st2dev /m/s/s/st2 ❯❯❯ cat /opt/stackstorm/packs/examples/actions/pythonactions/isprime.py
import math
import os

from base import get_environ


class PrimeCheckerAction(Action):
    def run(self, value=0):
        self.logger.info(get_environ('PYTHONPATH'))
        self.logger.debug('value=%s' % (value))
        if math.floor(value) != value:
            raise ValueError('%s should be an integer.' % value)
        if value < 2:
            return False
        for test in range(2, int(math.floor(math.sqrt(value)))+1):
            if value % test == 0:
                return False
        return True

if __name__ == '__main__':
    checker = PrimeCheckerAction()
    for i in range(0, 10):
        print '%s : %s' % (i, checker.run(value=1))
```

## TODO
* [X] Examples
* [X] Docs https://github.com/StackStorm/st2docs/pull/576
* [x] Unit tests No unit tests possible.  
* [X] Integration tests in st2tests https://github.com/StackStorm/st2tests/pull/111
* [x] Changelog

## Feedback needed

1) Import style 

As you can see in the action consuming the lib, the import line looks so

from base import get_environ

I want all the feature requesters like @andrew-regan @ytjohn @nmaludy to weigh in on if this works for them or do they need something like

from lib.base import get_environ

2) Do you want to allow lib code to contain dependencies from pack requirements? 

I think this should work the way we have done sandboxing but I want to know if this is something you really want it to work? 

3) Anything else?


st2common/st2common/util/pack.py
```diff
@@ -112,6 +113,31 @@ def validate_config_against_schema(config_schema, config_object, config_path,
     return cleaned
 
 
+def get_pack_common_libs_path(pack_db):
+    """
+    Return the pack's common lib path. This is the path where common code for sensors
+    and actions are placed.
+
+    For example, if the pack is at /opt/stackstorm/packs/my_pack, you can place
+    common library code for actions and sensors in /opt/stackstorm/packs/my_pack/lib/.
```
armab: :+1:

andrew-regan: 👍

nmaludy:
@lakshmi-kannan Great work!

### Import Style

I prefer `lib.base` to `base` because it seems more "explicit" to me. However, i'm not sure how this will play out for existing packs that have python files in `actions/lib/` and `sensors/lib/`?

With this import style, will python actions be able to reference this directory like so : `entry_point: ../lib/operation.py`?

### Dependencies in requirements

Yes, i plan on having common code reference dependencies defined in the pack's requirements.txt. 
My use case for this `lib/` directory are "utility" functions that aggregate several common API call chains or transform input/output datatypes in the underlying integration library.


Thank you very much for this feature!

Kami:
I would also prefer `lib.base` over `base`, but this might be harder to achieve because this means we would need to add `<pack path>/actions/` to PYTHONPATH and make sure `actions/` directory is also a package (contains `__init__.py`).

contrib/runners/python_runner/python_runner.py
```diff
@@ -128,8 +129,13 @@ def run(self, action_parameters):
...
         env['PATH'] = get_sandbox_path(virtualenv_path=virtualenv_path)
-        env['PYTHONPATH'] = get_sandbox_python_path(inherit_from_parent=True,
-                                                    inherit_parent_virtualenv=True)
+
+
+        sandbox_python_path = get_sandbox_python_path(inherit_from_parent=True,
+                                                      inherit_parent_virtualenv=True)
+        pack_common_libs_path = get_pack_common_libs_path(pack_db=pack_db)
+
+        env['PYTHONPATH'] = pack_common_libs_path + ':' + sandbox_python_path
```
Kami: Just thinking out loud - is adding it to the front of the python path the right thing to do? :)

st2common/st2common/bootstrap/base.py
```diff
@@ -174,6 +174,7 @@ def _register_pack_db(self, pack_name, pack_dir):
         # Include a list of pack files
         pack_file_list = get_file_list(directory=pack_dir, exclude_patterns=EXCLUDE_FILE_PATTERNS)
         content['files'] = pack_file_list
+        content['pack'] = pack_dir
```
Kami: Should probably be `pack` -> `path` :D

st2common/st2common/util/pack.py
```diff
@@ -112,6 +113,31 @@ def validate_config_against_schema(config_schema, config_object, config_path,
...
+    This common library code is only available for python sensors and actions. The lib
+    structure also needs to follow a python convention with a __init__.py file.
+
+    :param pack_db: Pack DB model
+    :type pack_db: :class:`PackDB`
+
+    :rtype: ``str``
+    """
+    pack_dir = getattr(pack_db, 'path', None)
```
Kami: I guess having this path here static is fine, although in other places we dynamically resolve entry point, etc.

st2common/st2common/bootstrap/base.py
lakshmi-kannan: Yeah, I've the fix locally. Haven't pushed yet.

lakshmi-kannan:
> I prefer lib.base to base because it seems more "explicit" to me. However, i'm not sure how this will play out for existing packs that have python files in actions/lib/ and sensors/lib/

@nmaludy That's why supporting it would be a pain. Plus, for remote SSH actions, we copy over actions/lib. I don't want people to get confused with that lib with this lib dir :). I'll document this but still this is going to be a source of confusions. Therefore, I am going to not support lib.base but allow modules in ${pack_dir}/lib/ to be imported by name like import foo (and not import lib.foo)

nmaludy:
@lakshmi-kannan makes total sense as to the supportability and backwards compatibility of existing code.

Although I prefer the name lib/, I like the idea of renaming it to common/ to avoid confusion.

When you say "lib cannot contain actual st2 content", does this mean it cannot contain action and sensor code? Will the code in lib/ still be able to access st2 functions like the pack's config and the key/value store?

One final question, will lib/ require an __init__.py file to work?

lakshmi-kannan:
>When you say "lib cannot contain actual st2 content", does this mean it cannot contain action and sensor code? Will the code in lib/ still be able to access st2 functions like the pack's config and the key/value store?

Code in lib/ cannot call functions in action or sensor. This would be poor code design IMO. Code in lib/ can be called by both sensor and action. That's why it's called "lib". I assume data mangling functions, common client library code not pushed to pypi etc, common utility functions will go in files in lib. We will not support calling sensor or action code in lib. We can never do that with the way we isolate the sanboxed env for actions and sensors. 

>Will the code in lib/ still be able to access st2 functions like the pack's config and the key/value store?

I can see this can be useful down the road but the initial pass will not address this. This is a complete feature by itself and I won't take it on in my first pass here. 

>One final question, will lib/ require an init.py file to work?

Yes, I answered this already somewhere but yes, lib should a python package. Otherwise, imports won't work.

armab:
@Kami @lakshmi-kannan are we going to include this change as part of `2.4` or `2.5`?

lakshmi-kannan:
@armab This needs a review from @Kami. I plan to ship this for 2.5 now.

Kami:
> I would also prefer lib.base over base, but this might be harder to achieve because this means we would need to add <pack path>/actions/ to PYTHONPATH and make sure actions/ directory is also a package (contains __init__.py).

Looks like you answered this (but you mentioned @nmaludy and not my nick so I missed it). will look at the changes again.

st2common/st2common/models/api/pack.py
```diff
@@ -137,6 +137,11 @@ class PackAPI(BaseAPI):
                 'description': 'Specification for the system components and packages '
                                'required for the pack.',
                 'default': {}
+            },
+            'path': {
```
Kami: This would cause issues if content on different servers is stored in a different location, but that's kinda bad idea anyway and probably many other things would break as well.

Kami: On a related note - I believe we an an existing function which gets pack directory path (it searches all the search paths).

Should we utilize that here as well? Or deprecate it and just always utilize this db model value?

st2common/st2common/util/deprecation.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    This is a decorator which can be used to mark functions
+    as deprecated. It will result in a warning being emitted
+    when the function is used.
+    """
+    def new_func(*args, **kwargs):
+        warnings.warn("Call to deprecated function {}.".format(func.__name__),
+                      category=DeprecationWarning)
+        return func(*args, **kwargs)
+    new_func.__name__ = func.__name__
```
Kami: There is also `functools.wraps` which does this for you, but that also works.

Kami:
I have a small question / need some clarification on the function which returns pack path, besides that, LGTM.

Kami:
Another thing - can we please also add a test case (probably should go into `st2tests`) which tries to import module from `lib/` which has the same name as a built-in module.

Just to make sure things work as expected and that the precedence and relative imports work correctly.

So basically one which does something like "from os import foo" and then one which does "from __future__ import absolute_import ; from os import foo".

The behavior really depends on where we insert lib into `PYTHONPATH` (at the end or at the beginning) and this should of course also be documented, if not already :)

st2common/st2common/models/api/pack.py
lakshmi-kannan: I think I looked into this and settled on pack path being part of model made more sense and was cleaner.

lakshmi-kannan:
@Kami I added a test case where I put a file os.py inside lib/ folder and it doesn't matter the order how I construct the PYTHONPATH (tried both lib before and after sandbox python path), things break gloriously. I think supporting a module name that has same name as std module inside lib/ is something we can't do. Maybe I can document this? What are your thoughts?

Kami:
@lakshmi-kannan I'm fine if we can document it.

And ideally, if we can also make it fail loudly in scenarios like that (but that might not be possible, at least easily - would need to dig in to say for sure though).

lakshmi-kannan:
@Kami Here is an example of how it breaks

```
result_task: test_stdout_python_action
result:
  exit_code: 1
  result: None
  stderr: "Traceback (most recent call last):\n  File \"/opt/stackstorm/virtualenvs/fixtures/lib/python2.7/site.py\", line 703, in <module>\n    main()\n  File \"/opt/stackstorm/virtualenvs/fixtures/lib/python2.7/site.py\", line 670, in main\n    virtual_install_main_packages()\n  File \"/opt/stackstorm/virtualenvs/fixtures/lib/python2.7/site.py\", line 553, in virtual_install_main_packages\n    f = open(os.path.join(os.path.dirname(__file__), 'orig-prefix.txt'))\nAttributeError: 'module' object has no attribute 'path'\n"
  stdout: ''
error: Traceback (most recent call last):
  File "/opt/stackstorm/virtualenvs/fixtures/lib/python2.7/site.py", line 703, in <module>
    main()
  File "/opt/stackstorm/virtualenvs/fixtures/lib/python2.7/site.py", line 670, in main
    virtual_install_main_packages()
  File "/opt/stackstorm/virtualenvs/fixtures/lib/python2.7/site.py", line 553, in virtual_install_main_packages
    f = open(os.path.join(os.path.dirname(__file__), 'orig-prefix.txt'))
AttributeError: 'module' object has no attribute 'path'
```

If we want to do runtime check, this is going to be hard. Maybe we can add this to the linter to stop people from doing this?

Kami:
I think having a linter or similar would be hard, because we would probably manually keep a list of the "reserved" / system module names.

So I'm not aware of any easy solution besides namespacing / prefixing user modules with lib. or similar.

The more problematic thing is that it's not just stdlib package / module names, but any package / module which is available in the virtualenv which could make it quite annoying and action developers would need to get quite creative with the names.

lakshmi-kannan:
@Kami and I discussed this in slack. In conclusion, a lot of packs already have a 'lib' folder and do imports like from lib.action import foo. By merging this PR, we will break things I think. It's better to not merge this until we can confirm we won't break existing packs that use "lib".

Kami:
Yeah, sadly this feature is a lot more complicated then we would hope for - if we continued with the current approach it has a potential to break existing code (potentially in obscure manner which would cause unnecessary confusion and bad UX).

Only good option we can think of is namespacing the imports - e.g. something like `from <pack name>.lib.<module name> import foo` and / or `from <pack name>/actions/lib/<module name> import foo`.

The problem with the name spacing is that they are two ways to achieve it:

1. Some PATH hackery so it only works inside st2 even if pack directory is not a package
2. Making sure pack root directory and `<pack root>/lib/` directories are packages (contain `__init__.py` file)

I'm strongly against the first approach. The code should also work if user runs it outside StackStorm environment (possible to re-use it, easier to develop and troubleshoot, etc.).

Second approach would mean it only works with new code and packs which are also Python packages. I'm fine with that, the problem is that now package root would also need to contain `__init__py` and I'm not a big fan of that - packs can contain actions for different runners and this would make it somewhat biased against the Python runner.

lakshmi-kannan:
@Kami I created this pack to test our theory that these changes might break if actions and sensors have "lib" folder inside them. I was thinking it shouldn't and it looks like it doesn't. Here is a sample pack https://github.com/lakshmi-kannan/lib_test with a pack level lib folder and a lib folder inside sensors/ and actions/. Both actions and sensors import from pack lib and lib folders inside actions/ and sensors/ respectively. The sensor runs fine and also the actions. I verified that there are no import issues. I'll scan exchange and see if there are any other type of imports I need to test. Also, I'll port these tests to st2tests. Summary: Looks like there is still hope!

conf/st2.dev.conf
```diff
@@ -106,3 +106,6 @@ pack_group = stanley
 [mistral]
 v2_base_url = http://127.0.0.1:8989/v2
 jitter_interval = 0
+
+[packs]
+enable_common_libs = True
```
Kami: We have `[system]` where we have other packs related options there so there are some inconsistencies.

Anyway, I'm fine with this for now.

conf/st2.conf.sample
```diff
@@ -292,3 +292,6 @@ local_timezone = America/Los_Angeles
 # Base https URL to access st2 Web UI. This is used to constructhistory URLs that are sent out when chatops is used to kick off executions.
 webui_base_url = https://localhost
 
+[packs]
+# Enable/Disable support for pack common libs
```
Kami: Needs some clarification and link to the docs.

contrib/runners/python_runner/python_runner.py
```diff
@@ -105,7 +109,8 @@ def pre_run(self):
     def run(self, action_parameters):
         LOG.debug('Running pythonrunner.')
         LOG.debug('Getting pack name.')
-        pack = self.get_pack_name()
+        pack = self.get_pack_ref()
+        pack_db = Pack.get_by_ref(pack)
```
Kami: I guess there could be some races now - e.g. when pack is being removed and the model is removed first, but then again, we can't really guarantee proper behavior during pack removal process anyway.

contrib/runners/python_runner/python_runner.py
```diff
@@ -156,8 +161,15 @@ def run(self, action_parameters):
...
         env['PATH'] = get_sandbox_path(virtualenv_path=virtualenv_path)
-        env['PYTHONPATH'] = get_sandbox_python_path(inherit_from_parent=True,
-                                                    inherit_parent_virtualenv=True)
+
+        sandbox_python_path = get_sandbox_python_path(inherit_from_parent=True,
+                                                      inherit_parent_virtualenv=True)
+        pack_common_libs_path = get_pack_common_libs_path(pack_db=pack_db)
+
+        if self._enable_common_pack_libs and pack_common_libs_path:
+            env['PYTHONPATH'] = pack_common_libs_path + ':' + sandbox_python_path
```
Kami: Can you please also add some unit tests which assert on the actual `PYTHONPATH` for various scenarios (libs path is enabled, disabled, etc.)?

If they are already there, please ignore my comment, but I didn't see any. I only see `st2tests` test, but it only seems to be testing one code path (common libs enabled). Should be straight forward to add a unit test - Python runner action can just print PYTHONPATH and we can assert on that.

lakshmi-kannan: Fixed in ce6fb89

PR #3662, opened 2017-08-09
Added a floating point timestamp to GELF formatter

nmaludy:
### Problems

1. Previously the GELF formatter was only outputting integer timestamps. This can causes issues in logging aggregation software that's trying to sort by time. Because many log messages can contain the same timestamp, things get displayed out-of-order resulting in an incorrect chain of events. Alternatively the log shipping software could generate its own timestamps, but why not use the actual timestamp if we can?

2. The timestamp being utilized by the GELF formatter was generated within the `format()` function instead of utilizing the time in the logging record in the `created` instance variable. This is not a huge problem, however if we can use a consistent time why not? This is the time field that is utilized by the default formatter when generating timestamps 

### Enhancements

1. To be consistent between releases i've kept the integer timestamp in the field `timestamp` and created a new field named `timestamp_f` that is the floating point time. I named it `timestamp_f` because `time.time()` can have various amounts of accuracy depending on the system so i didn't want to label it `timestamp_ms` or `timestamp_ns` and not be able to provide that level of accuracy.

Example: 
```
$ tail -n 1 st2api.gelf.log
==> st2api.gelf.log <==
{"version": "1.1", "level": 6, "timestamp": 1502305447, "_python": {"name": "kombu.mixins", "process": 15333, "module": "mixins", "funcName": "Consumer", "processName": "MainProcess", "lineno": 231, "filename": "mixins.py"}, "timestamp_f": 1502305447.887307, "host": "stackstorm2.xxx.xxx", "full_message": "Connected to amqp://guest:**@127.0.0.1:5672//", "short_message": "Connected to %s"}
```

2. I've converted over to utilizing the `record.created` value for the timestamp so that the time between the `st2api.log` and any outputted GELF logs are consistent. In the example below notice the same fractional component in the `timestamp_f` field in the GELF log and the fractional component in the timestamp in the `st2api.log`:

```
$  tail -n 1 st2api.gelf.log st2api.log 
==> st2api.gelf.log <==
{"version": "1.1", "level": 6, "timestamp": 1502305447, "_python": {"name": "kombu.mixins", "process": 15333, "module": "mixins", "funcName": "Consumer", "processName": "MainProcess", "lineno": 231, "filename": "mixins.py"}, "timestamp_f": 1502305447.887307, "host": "stackstorm2.xxx.xxx", "full_message": "Connected to amqp://guest:**@127.0.0.1:5672//", "short_message": "Connected to %s"}

==> st2api.log <==
2017-08-09 15:04:07,887 54550928 INFO mixins [-] Connected to amqp://guest:**@127.0.0.1:5672//
```

###  Suggestions?

I'm open to further suggestions? 

Not sure if the `timestamp` field to be updated to a floating point value or if they should remain separate for backwards-compatibility?


Kami:
Thanks, the approach and change itself looks good to me :+1:

Kami:
Merged - thanks again!

Issue #3663, opened 2017-08-10
Sensors in "Examples" pack should do something meaningful


lakshmi-kannan:
The sensors in examples pack all have pass implementations. Create an active and a passive sensor that does some real work.

lakshmi-kannan:
As part of PR https://github.com/StackStorm/st2/pull/3658, I added a sample working polling sensor https://github.com/StackStorm/st2/pull/3658/commits/7749edd8b8d419fd9f01abe22957e5ca5af623c9

lakshmi-kannan:
A passive sensor still needs to be written. Write a simple flask app and just relay the incoming payload to the trigger could be an example.

echo_flask_sensor.py

armab:
Passive sensor implemented in https://github.com/StackStorm/st2/pull/3667 by @warrenvw.

@lakshmi-kannan @warrenvw Can we close this now is resolved?

lingfish:
Hi @warrenvw and @armab - in order to use this, is the recommendation to then adjust nginx appropriately to point to it?

PR #3667, opened 2017-08-11
Add passive Flask-based sensor that echoes input

warrenvw:
See #3663 

A simple, Flask-based sensor that relays the incoming payload to the trigger.

To Do:
 - [x] Testing


contrib/examples/sensors/echo_flask_app.py
```diff
@@ -0,0 +1,49 @@
...
+        @self._app.route(self._path, methods=['POST'])
+        def echo():
+            payload = request.get_json(force=True)
+            self._sensor_service.dispatch(trigger="examples.echo_flask",
+                                          payload=payload)
+            return request.data
+
+        self._log.info('Listening for payload on http://%s:%s%s' %
+                       (self._host, self._port, self._path))
+        self._app.run(host=self._host, port=self._port)
```
armab: Use `threaded=True` for Flask app, otherwise it will block on 1+ requests,
see similar: https://github.com/StackStorm-Exchange/stackstorm-activecampaign/pull/3

LindsayHill:
@warrenvw looks like the trigger name dispatched in the .py file doesn't match the trigger definition in the YAML?

NikosVlagoidis:
So I guess Flask is already installed. How can I install my requirements for a sensor? Should I make a pack?

LindsayHill:
@NikosVlagoidis Regardless of requirements, yes, your sensor will need to be in a pack, usually your own. That pack should have a `requirements.txt` file. When you install the pack, ST2 will create a virtualenv for that pack, and install the requirements

NikosVlagoidis:
thank you, sir.

NikosVlagoidis:
Tried to use that inside a Docker container. Flask complains. Port must be an integer. Can someone confirm that?

LindsayHill:
Did you try changing it to an integer? Did that work?

NikosVlagoidis:
Yes changing to integer worked.

LindsayHill:
Good to hear. Would you mind submitting a PR with a fix?

PR #3668, opened 2017-08-11
Fix CircleCI docker volumes cleanup

armab:
Associated change in st2-packages: https://github.com/StackStorm/st2-packages/pull/481

Recently we got more failed builds. After more investigation it was "No space left on device" errors which resulted that `mongodb`, `rabbitmq`, `postgresql` failed to start.
See: https://discuss.circleci.com/t/no-space-left-on-device-while-creating-mongo/11532/13 for more context.

Fixing "No space left on device" in CircleCI infrastructure via additional Cleanup, caused by volumes persisting between the builds.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -153,7 +153,6 @@ jobs:
           name: Pre Docker cleanup
           command: |
             set -x
-            docker ps -a
             # Clean-up running containers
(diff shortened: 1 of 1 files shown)


Kami:
Good catch :+1:

Issue #3659, opened 2017-08-08
st2ctl clean fails to delete mongo db when on a separate server


skjbulcher:
Originial bug report [here ](https://github.com/StackStorm/st2-docker/issues/32)(example is taken from a docker network).

When running st2ctl clean, command fails with `mong: command not found`. After installing mongo (`apt-get install -y mongodb-clients`, the error changes to:

```
Dropping st2 Database...
MongoDB shell version: 2.4.9
connecting to: st2
Tue Aug  8 18:28:04.691 Error: couldn't connect to server 127.0.0.1:27017 at src/mongo/shell/mongo.js:145
exception: connect failed
```

`st2ctl clean` appears to require a mongo client, and fails to obtain the db connection information from st2.conf

armab:
Fixed in https://github.com/StackStorm/st2/pull/3669.
Thanks to @nmaludy `st2cl clean` now relies on python script that's using mongo creds from `st2.conf`.

PR #3669, opened 2017-08-13
Bugfix for #3659 st2ctl clean not reading the database connection

nmaludy:
Fix for #3659 

This adds functionality of reading database connection information from the config and passing it to the `mongo` command during `st2ctl clean` when dropping the database.

For simplicity i've used `crudini` to parse the config since `crudini` is installed by one-liner install method. I found several other methods to do this in pure bash but they were all very complicated. If you would like me to remove the `crudini` dependency i'd be happy to do so.

If `crudini` is not present in PATH when invoked then the following warning will be printed out:
```
WARNING: The 'crudini' command is not present in PATH, using default database connection options. To use database connection options from the config, and suppress this warning, please install 'crudini'
```
In this case we'll fall back to use the default database name of `st2` with no additional arguments. This is the exact behavior that was seen prior to this change, so this should be backwards.

 **Note** 
The only thing i wasn't sure about  was `ssl_certfile` and `ssl_keyfile`. In the `mongo` CLI there appears to only be one option to encompass both of these `--sslPEMKeyFile` . From the man page:

```
       --sslPEMKeyFile <filename>
              New in version 2.4.

              Specifies  the  .pem  file  that contains both the SSL certificate and key. Specify the file name of the .pem
              file using relative or absolute paths.

              This option is required when using the --ssl option to connect to a mongod or mongos that has CAFile  enabled
              without allowConnectionsWithoutCertificates.

              The  default  distribution  of MongoDB does not contain support for SSL.  For more information on MongoDB and
              SSL, see http://docs.mongodb.org/manual/tutorial/configure-ssl.
```

I've implemented logic to first use `ssl_certfile` if specified in the config, otherwise use `ssl_keyfile` if specified. Any guidance here is appreciated.


armab:
I was thinking about re-using existing python functionality (`st2common` and so on) which will purge the DB, based on existing `st2.conf` settings. And then just invoke that python script from `st2ctl`, like we do with `st2-register-content`. This way we get rid of mongo shell dependency and don't add another dependency like crudini.

Curious what others @StackStorm/team think about it? Is this PR good enough for the time being or we need something better?

Kami:
Yeah, another option which would allow us to re-use code from st2 would be to add this functionality to st2common and script and utilize that inside `st2ctl`.

That approach should be more robust because it would rely on the underlying mongoengine and / or pymongo functionality. In addition to that it also allows us to test that code and all other benefits which go with the "code" approach.

Also, thanks for bringing this up and working on it :+1:

nmaludy:
@Kami @armab That makes a lot of sense. I especially like it from a testing perspective. Let me re-work this PR and get back with you. FYI, going to take me a few days because i'm traveling this week.

Kami:
Great, thanks - and no rush :)

Mierdin:
I’d say `st2common/persistence/cleanup.py` probably makes the most sense

I was also going to say look at how the bootstrapping code works, since that’s kind of the same model re: st2ctl, but that code leverages `persistence` anyways, and this is closer to the DB

So yeah, a new file in persistence makes sense. It’s **kind of** icky because the majority of the files in there are pertinent to specific DB models, and this is more of a housekeeping function, but there’s already a precedent for that kind of thing (db_init) so I say GFI. And as @kami pointed out, this way the core of the functionality is in "core" st2, which means it can be properly tested, and `st2ctl` can simply refer to it.

nmaludy:
@armab and @Mierdin based on your suggestions i've implemented a new command `st2-cleanup-db` that drops the MongoDB database using the information in the `/etc/st2/st2.conf` config (or whatever config you specify on the CLI).

This is my first "real" PR to the main StackStorm repo, so feedback is appreciated.

I've also created a corresponding PR in st2-packages repo https://github.com/StackStorm/st2-packages/pull/494. Are there any other places that i need to update with the new command?

st2common/st2common/persistence/cleanup.py
```diff
@@ -0,0 +1,77 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+
+from st2common import config
+from st2common import log as logging
+from st2common.database_setup import db_config
+from st2common.models.db import db_cleanup as db_cleanup_func
```
lakshmi-kannan: This method doesn't seem to exist

 /Volumes/workspace/src/storm/st2 ⮀ master ⮀ 9h2m ⮀ ag "db_cleanup_func"    ⮂
 /Volumes/workspace/src/storm/st2 ⮀ master ⮀ 9h2m ⮀

st2common/st2common/models/db/__init__.py
```diff
@@ -87,6 +88,18 @@ def db_setup(db_name, db_host, db_port, username=None, password=None, ensure_ind
                                                 port=db_port, tz_aware=True,
                                                 username=username, password=password,
                                                 **ssl_kwargs)
+    return connection
```
lakshmi-kannan: This connection object is never used anywhere else. Shouldn't you be disconnecting somewhere?

st2common/st2common/models/db/__init__.py
```diff
@@ -199,6 +212,23 @@ def db_teardown():
...
+                             password=password, ssl=ssl, ssl_keyfile=ssl_keyfile,
+                             ssl_certfile=ssl_certfile,
+                             ssl_cert_reqs=ssl_cert_reqs, ssl_ca_certs=ssl_ca_certs,
+                             ssl_match_hostname=ssl_match_hostname)
+
+    LOG.info('Dropping database "%s" @ "%s:%s" as user "%s".',
+             db_name, db_host, db_port, str(username))
+
+    connection.drop_database(db_name)
+    return connection
```
lakshmi-kannan: I do not like that this method is returning a connection object. It's a cleanup method :). The API seems less ideal. What does connection.drop_database(db_name) return? Do you want to just do return connection.drop_database(db_name)?

st2common/st2common/models/db/__init__.py
nmaludy: I was modeling this after the already existing `db_setup()` method (just below this one).

I believe the connection is disconnected in `st2common/st2common/persistence/cleanup.py` `teardown()`

If you would like me to change this behavior i'm fine with it, please advise on what you'd like to see.

st2common/st2common/models/db/__init__.py
nmaludy: It looks like connection.drop_database() is a PyMongo call that has no return statement.

st2common/st2common/persistence/cleanup.py
nmaludy: I'm importing the `db_cleanup()` function from `st2common.models.db.__init__.py` as an alias `db_cleanup_func()`

lakshmi-kannan: Ahh. I see it now. Sorry.

st2common/st2common/models/db/__init__.py
lakshmi-kannan: Okay, I see how the original db_setup returns a connection. Ignore my comment.

st2common/st2common/models/db/__init__.py
lakshmi-kannan: You are right. I looked into this and the method does not return anything. In this case, I'll make note of this as comment and change this method to not return anything.

lakshmi-kannan:
You have my +1. Let's wait for @Kami or @m4dcoder to +1 as well.

st2common/setup.py
```diff
@@ -43,6 +43,7 @@
     packages=find_packages(exclude=['setuptools', 'tests']),
     scripts=[
         'bin/st2-bootstrap-rmq',
+        'bin/st2-cleanup-db',
```
Kami: Is it a good idea to include this as part of standard distribution (need to also check st2-packages).

This is mostly meant for dev purposes and if we ship it with the actual package distribution I'm worried someone could accidentally run it and remove all data in the database.

So in short, I think we should make it more fail-safe - perhaps the best way is move it under tools/ or similar and not include it with packages.

armab: `st2ctl clean` relies on that file, therefore it should be part of the packaging.

There is a https://github.com/StackStorm/st2-packages/pull/494/ which adds symlink `/usr/bin/st2-cleanup-db` via `st2-packages`. I don't think we should do that, but "hide it" via `/opt/stackstorm/st2/bin`

nmaludy: @armab i've closed the packaging PR StackStorm/st2-packages#494 and also updated `st2common/bin/st2ctl` to execute the command using the full path instead if relying on it being available in `$PATH`

nmaludy:
Is there something else i need to do for this? If i've missed something in the review comments, please let me know.

Thanks!

lakshmi-kannan:
@nmaludy We've exceeded the time to get this merged. Sorry about dropping the ball here. I went ahead and approved the PR. I see no strong objection to merging this from either @Kami or @m4dcoder

nmaludy:
@lakshmi-kannan @Kami Looks like i waited too long and there was a merge conflict from another PR. I've resolved the conflict and this should be good to merge

Kami:
Merged, yay! :)

Thanks again.

Issue #3646, opened 2017-08-04
[st2client/RFI] It's hard to set `array of object` in an action parameter.


userlocalhost:
## Background / Problem
We can easily set object type values in a action parameter by the past efforts of the Team, like this.
```
$ st2 run some_pack.action param_obj='foo=bar,hoge=fuga'
```

In the similar way, we can do it for array type values, as follows.
```
$ st2 run some_pack.action param_arr='foo,bar,baz'
```

But it's quite difficult to set 'array of object' value, like following.
```
$ st2 run some_pack.action param_arr_obj=['{"foo": "bar"}','{"hoge": "fuga"}']
```
(Even if only a white space after comma is not permitted)

Currently, I can seldom see the action that requires`array of object` type value in a parameter.
But I think this restriction prevents to make it.

## Suggestion
One of an idea to improve it is permitting to specify values more than once only if the array type parameter, as follows.
```
$ st2 run some_pack.action \
    param_arr_obj='foo=bar'
    param_arr_obj='hoge=fuga'
```

Internally, the `param_arr_obj` transforms as follows.
```
[{'foo': 'bar'}, {'hoge': 'fuga'}]
```

In this solution, I think there are following advantages.
- (User friendly) This easily makes `array of object` using past assets.
- (Backward compatibility) This doesn't change current UI.

Could you please let me know what you think about it?

Thank you

Mierdin:
I like the idea, though I would prefer to see different operators inside vs outside the dict definitions. I.e.:

```
$ st2 run some_pack.action \
    param_arr_obj='foo:bar'
    param_arr_obj='hoge:fuga'
```

I am imagining that in more complex scenarios, it would be difficult to keep track of the equal signs.

userlocalhost:
Thank you for your reply!
I should be able to make a PR for it based on your comment.

LindsayHill:
Addressed by #3670

PR #3670, opened 2017-08-14
[st2client] Updated ActionRunCommand to enable specifying multiple array type value in the action parameter

userlocalhost:
This is the change for #3646. The main points of this PR are

* extending `transform_array` which is the internal method to parse the argument of `array` type, to be able to parse dict data when user declared that type parameter.
* changing `_get_action_parameters_from_args` to accept multiple declaration of same argument.

Thank you.


Kami:
Thanks and nice work!

Can you please also open corresponding st2docs PR?

st2client/tests/unit/test_command_actionrun.py
```diff
@@ -55,3 +56,81 @@ def test_get_params_types(self):
...
+        mockarg = mock.Mock()
+        mockarg.inherit_env = False
+        mockarg.parameters = [
+            'param_string=hoge',
+            'param_integer=123',
+            'param_number=1.23',
+            'param_object=hoge=1,fuga=2',
+            'param_boolean=False',
+            'param_array=foo,bar,baz',
+            'param_array_of_dicts=foo:1,bar:2'
```
Kami: How does it work if you want to specify multiple objects for an array?

Would something like that work: `param_array_of_dicts=foo:1,bar:2,aram_array_of_dicts=foo:3,bar:4`.

If yes, let's please add a test case or it and also document it.

userlocalhost:
Thank you for your review and merging.
I got it. I'll make a PR which describes this feature to st2docs and do another one to add tests for array of ducts which have multiple type value.

It takes time because I've been offline until the end of this week, sorry.

PR #3674, opened 2017-08-15
Remove local pack config

m4dcoder:
The feature to use the local config.yaml in packs is removed. The config.yaml is deprecated and replaced by the config schema in earlier version.

On content loading in bootstrap, the warning message is replaced with the following error message.

2017-08-17 19:48:50,326 ERROR [-] Pack "demo" contains a deprecated config.yaml file (/opt/stackstorm/packs/demo/config.yaml). Support for "config.yaml" files has been deprecated in StackStorm v1.6.0 in favor of config.schema.yaml config schema files and config files in /opt/stackstorm/configs/ directory. Support for config.yaml files has been removed in the release (v2.4.0) so please migrate. For more information please refer to https://docs.stackstorm.com/reference/pack_configs.html

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -38,6 +38,11 @@ Changed
   official instructions at https://docs.mongodb.com/v3.4/release-notes/3.4-upgrade-standalone/.
   (improvement)
 
+Removed
+~~~~~~~
+
+* The feature to use local config.yaml in packs is removed.
+
 Fixed
 ~~~~~
 
--- a/st2common/st2common/util/config_loader.py
+++ b/st2common/st2common/util/config_loader.py
@@ -60,13 +60,6 @@ class ContentPackConfigLoader(object):
     def get_config(self):
         result = {}
 
-        # 1. Retrieve values from pack local config.yaml file
-        config = self._config_parser.get_config()
-
-        if config:
-            config = config.config or {}
-            result.update(config)
-
         # Retrieve corresponding ConfigDB and ConfigSchemaDB object
         # Note: ConfigSchemaDB is optional right now. If it doesn't exist, we assume every value
         # is of a type string
--- a/st2common/st2common/bootstrap/base.py
+++ b/st2common/st2common/bootstrap/base.py
@@ -143,15 +143,13 @@ class ResourceRegistrar(object):
         # will be fully removed in v2.4.0.
         config_path = os.path.join(pack_dir, 'config.yaml')
         if os.path.isfile(config_path):
-            LOG.warning('Pack "%s" contains a deprecated config.yaml file (%s). '
-                        'Support for "config.yaml" files has been deprecated in StackStorm v1.6.0 '
-                        'in favor of config.schema.yaml config schema files and config files in '
-                        '/opt/stackstorm/configs/ directory.'
-                        'Support for config.yaml files will be removed in next major release '
-                        ' (v2.4.0) so you are strongly encouraged to migrate. '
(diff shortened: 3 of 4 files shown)


CHANGELOG.rst
```diff
@@ -20,6 +20,11 @@ Changed
   official instructions at https://docs.mongodb.com/v3.4/release-notes/3.4-upgrade-standalone/.
   (improvement)
 
+Deprecated
+~~~~~~~~~~
```
LindsayHill: I think this section heading should be "Removed"
We Deprecated it in an earlier version, now it should be Removed.

LindsayHill: But then again, maybe it still belongs in "Deprecated" since we're making it hard to use, but not impossible, not yet anyway

st2common/st2common/util/config_loader.py
```diff
@@ -60,12 +60,16 @@ def __init__(self, pack_name, user=None):
...
-        # 1. Retrieve values from pack local config.yaml file
-        config = self._config_parser.get_config()
-
-        if config:
-            config = config.config or {}
-            result.update(config)
+        # The use of config.yaml local to the pack is deprecated. The use_local_pack_config
+        # option is default to False and will only be available for a short time in case
+        # some users are caught off guard.
+        if cfg.CONF.content.use_local_pack_config:
```
LindsayHill: How does this behave if that option is unset? The `if` statement fails? (which is good, if that's what it does)

Kami:
I'm fine with that, but as I said before, I would also be fine with totally removing support for config.yaml in v2.4 (been deprecated, documentation wise, for a very long time already).

Adding a new flag also means small, but additional thing to maintain and might just add more overhead compared to completely removing it. Now when we actually drop support, we probably need to leave this flag there to keep it backward compatible but throw or make it a no-op.

st2common/st2common/util/config_loader.py
Kami: We should also have tests to answer that question :P

m4dcoder: There is a unit test to test the `if` that is included in this PR.

m4dcoder:
When we are ready to remove it, we can just remove the flag from config as well. oslo config just ignores any undefined config option.

Kami:
Actually, looks like there is some confusion.

We already deprecated it in v2.3.0 (see changelog) so now it's time to fully remove it. This means removing all the, code, tests, docs, etc. related to it and clarifying that in the upgrade notes.

m4dcoder:
@Kami There was no confusion on our part. We know it was marked for deprecation last release. @LindsayHill suggested that we use a flag to disable this deprecated feature and make it hard to use of now but just in case users are caught off guard.

LindsayHill:
My original thinking was: Deprecate in 2.3 (which we did, with WARN logs), then Remove in 2.4 (Fail pack registration with ERROR).

We should have made that change to fail on `config.yaml` in master very soon after we released 2.3. 

But we didn't, and I'm a little nervous that there's something we've missed somewhere else, since we haven't been running this change on our internal infrastructure. Since it's now close to the planned 2.4 release time, I thought we should have the option to re-enable if we have to.

I can go either way on this one though. I'm OK with removing it completely. 

We should probably have a more consistent deprecation policy. We've been a bit inconsistent with that in the past.

LindsayHill:
LGTM

armab:
+1 and ack

Kami:
Thanks :+1:

Issue #3586, opened 2017-07-19
st2 pack install should show how many actions, rules and sensors are being installed


lakshmi-kannan:
This came from AWS pack discussion where AWS pack has 2000+ actions. st2 pack install aws is so slow and as a user I begin to think that there is something wrong. This happened to me personally multiple times. So if we just show the number of artifacts being registered, we mentally setup the user to expect a delay. 

Apparently, pack index already has this info. Let's show it in CLI then!

armab:
https://github.com/StackStorm-Exchange/index/blob/87d6ac55709695cb62886760ceeba304c15671b1/v1/index.json#L138 :smiling_imp:

warrenvw:
I guess it's just coincidence that this is issue 3586, and there are 3581 AWS actions. Just off by 5. :)

PR #3685, opened 2017-08-18
Increase 'pack register content' timeout

armab:
Our [AWS pack](https://github.com/stackstorm-exchange/stackstorm-aws) ballooned into `3581` actions which sometimes results in `st2 pack install aws` failures on slower systems (reported in #community).

For now "temporary" increasing pack register timeout.

The proper fix is to investigate the performance and research the pack content registration optimizations 
StackStorm by design should handle big packs https://github.com/StackStorm/st2/issues/3686.

Also there is a related cosmetic enhancement https://github.com/StackStorm/st2/issues/3586 which helps to additionally signalize if there is a *huge* amount of content in a pack.


LindsayHill:
There may be a difference in behavior between CentOS & Ubuntu here. Needs more testing to confirm, but initial indications are that these timeouts are more likely to occur on CentOS systems.

PR #3688, opened 2017-08-18
Fix timeout listening for events with client side async execution

m4dcoder:
On pack install, the command listens for event from the API server. If the execution takes too long, the SSEClient which listens for events on the client side returns an empty string which leads to JSON load error.

### DIFF
--- a/st2client/st2client/models/core.py
+++ b/st2client/st2client/models/core.py
@@ -524,4 +524,10 @@ class StreamManager(object):
         url = url + query_string
 
         for message in SSEClient(url):
+
+            # If the execution on the API server takes too long, the message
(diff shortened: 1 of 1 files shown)


st2client/st2client/models/core.py
```diff
@@ -524,4 +524,10 @@ def listen(self, events=None, **kwargs):
         url = url + query_string
 
         for message in SSEClient(url):
+
+            # If the execution on the API server takes too long, the message
+            # can be empty. In this case, rerun the query.
+            if not message.data:
```
Kami: Good catch :+1: 

And yeah we periodically send empty message / line break on hear beat to keep the connection open.

PR #3704, opened 2017-08-28
[st2client] Corrected the implementation of transformer of array to parse the values in object as the type which is specified in the action metadata

userlocalhost:
This is the correction based on [the suggestion from Kami](https://github.com/StackStorm/st2/pull/3670/files#r134171370).

This patch corrects the `_get_action_parameters_from_args` method, which parses `array of dicts` value from the command-line arguments, to parse each values in the object as the type which is specified in the action metadata.

### DIFF
--- a/st2client/st2client/commands/action.py
+++ b/st2client/st2client/commands/action.py
@@ -523,7 +523,7 @@ class ActionRunCommandMixin(object):
                     result[key] = value
             return result
 
-        def transform_array(value):
+        def transform_array(value, action_params = {}):
             # Sometimes an array parameter only has a single element:
             #
             #     i.e. "st2 run foopack.fooaction arrayparam=51"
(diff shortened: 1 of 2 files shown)


st2client/st2client/commands/action.py
```diff
@@ -523,7 +523,7 @@ def transform_object(value):
                     result[key] = value
             return result
 
-        def transform_array(value):
+        def transform_array(value, action_params={}):
```
Kami: Please default it to `None` in the function signature and default it to empty dict inside a function otherwise you will be in for a surprise - https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html.

st2client/st2client/commands/action.py
```diff
@@ -570,28 +575,30 @@ def transform_array(value):
             'string': str
         }
 
-        def get_param_type(key):
+        def get_param_type(key, action_params=action.parameters):
```
Kami: Same here, assign a default value inside the function.

Kami:
Thanks.

The issue with a mutable default value needs to be fixed, besides that, LGTM :+1:

userlocalhost:
Thank you for your review and giving an advice. I'll fix it soon.

st2client/st2client/commands/action.py
Kami: This one still needs to be fixed^

Besides that, LGTM.

userlocalhost: I'm so sorry for overlooking the point to fix.
I'll do it right now and recheck others.

Kami:
Thanks - I will add a changelog entry and merge it into master :)

PR #3715, opened 2017-09-05
CI check which verifies that all the automatically generated files are up to date

Kami:
This pull request adds a new CI check which verifies that all the automatically generated files are up to date (this means that if the file source file has been changed, automatically generated file has been re-generated and included in the changeset).

This way we can be sure we won't forget to run the auto generation script (they are part of all the Make targets, but it still happens people forget to include and commit it as part of their changes).

Thanks to @enykeev for the suggestion.


Kami:
It needed a bit more work (0daa95ad1539d8c6c61033f19cb5c748f82adc24, 93c02fb8d72c3fc60113ad38aaa7013cd50d9d9f), but it's done and working now so I'm going ahead and merging it.

PR #3721, opened 2017-09-07
Fix mistral callback failure when result contains unicode

m4dcoder:
When result of action execution contains unicode, Mistral callback fails and leads to orphaned workflow execution.

### DIFF
--- a/contrib/runners/mistral_v2/tests/unit/test_mistral_v2_callback.py
+++ b/contrib/runners/mistral_v2/tests/unit/test_mistral_v2_callback.py
@@ -110,6 +110,12 @@ class MistralRunnerCallbackTest(DbTestCase):
                                      action_constants.LIVEACTION_STATUS_SUCCEEDED,
                                      '<html></html>')
 
+        action_executions.ActionExecutionManager.update.assert_called_with(
+            '12345',
+            state='SUCCESS',
+            output='<html></html>'
+        )
+
     @mock.patch.object(
         action_executions.ActionExecutionManager, 'update',
         mock.MagicMock(return_value=None))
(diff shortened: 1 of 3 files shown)


contrib/runners/mistral_v2/tests/unit/test_mistral_v2_callback.py
```diff
@@ -142,6 +173,94 @@ def test_callback_handler_with_result_as_list_str(self):
...
+            '12345',
+            state='SUCCESS',
+            output='{"a": "\\\\u4ec0\\\\u9ebc"}'
+        )
+
+    @mock.patch.object(
+        action_executions.ActionExecutionManager, 'update',
+        mock.MagicMock(return_value=None))
+    def test_callback_handler_with_result_as_dict_with_unicode_type(self):
+        self.callback_class.callback('http://127.0.0.1:8989/v2/action_executions/12345', {},
```
Kami: It would be good to test unicode for all the various type (string, list, dict) and also add a test case with actual unicode string (`💩`) in addition to existing tests with the unicode escape sequences.

Kami: Besides that, LGTM :+1:

Kami:
Added v2.4.1 milestone since I assume this will / should go into v2.4.1.

armab:
👍 

There are also some failed Unit tests.

contrib/runners/mistral_v2/tests/unit/test_mistral_v2_callback.py
m4dcoder: @Kami added more unit tests as requested

PR #3726, opened 2017-09-08
Fix cancellation of delayed action execution

m4dcoder:
On cancellation request, if the liveaction has a parent, set the status to canceling to trigger post_run for the liveaction. Currently, the action execution is set to canceled and callback to Mistral is not performed, thus leading to workflows stuck in running state.

### DIFF
--- a/st2actions/tests/unit/test_execution_cancellation.py
+++ b/st2actions/tests/unit/test_execution_cancellation.py
@@ -15,6 +15,7 @@
 
 import mock
 import six
+import uuid
 
 from oslo_config import cfg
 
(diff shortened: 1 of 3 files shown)


st2actions/tests/unit/test_execution_cancellation.py
```diff
@@ -13,61 +13,93 @@
...
-import st2common.bootstrap.runnersregistrar as runners_registrar
-from st2common.runners.base import ActionRunner
-import local_runner
 from st2common.constants import action as action_constants
+from st2common.models.api.action import ActionAPI, RunnerTypeAPI
 from st2common.models.db.liveaction import LiveActionDB
-from st2common.models.api.action import ActionAPI
-from st2common.persistence.action import Action
-from st2common.persistence.liveaction import LiveAction
+from st2common.persistence.action import Action, LiveAction
```
Kami: Nooo :P

One import per line makes grepping for imports easier.

Kami:
LGTM.

Can you please also add a changelog entry.

lakshmi-kannan:
I am amazed at how many cancel bugs we've found so far! LGTM.

PR #3729, opened 2017-09-11
Action runner output streaming single collection approach

Kami:
This pull request works on top off #3657 and changes two collection approach with single "generic" collection approach - 03c37d4fb05119c7d66592d8c9c6aab5b2764699.

For some reason the actual diff looks a bit messed up, but the change is located in 03c37d4fb05119c7d66592d8c9c6aab5b2764699.

It addresses the feedback people had and makes the whole approach a bit more generic so it can work with other / future runners which produce just a single output stream.

The main difference is that this approach utilizes single "action_output_d_b" collection with the following fields:

* `execution_id`
* `action_ref`
* `runner_ref`
* `timestamp`
* `type`
* `data`

Most fields are self explanatory, but two fields of interest are "data" and "type".

"data" contains actual partial execution output. Depending on the runner, this could either be chunk, line or similar. For all the existing runners for which we implement streaming right now this is a line. I was thinking about a different name (e.g. `chunk`), but in the end I settled with `data`.

`type` contains type of the output. Right now that is either `stdout` or `stderr`, but for runners with just a single output stream, this could simply be `output` or similar. I'm open to suggestions for a better name for this field.

I was thinking about "stream" / "stream_type" / "output_type", but in the end I settled with "type". If we can't come up with anything better, I will probably change it to "output_type" so it doesn't clash with Python built-in `type` function.

As for now, I only implemented model changes so I won't cause unnecessary additional work for myself in case people don't agree with this approach.

Once we agree that's the model we all like, I will go ahead and implement rest of the changes. 

If we agree on this approach, actual API endpoint and CLI command will look like this:

* `/v1/executions/<execution id>/output[?type=stdout/stderr/etc]`
* `st2 execution tail [--type=stdout/stderr/etc]`

## TODO

- [x] Update affected API endpoints
- [x] Update affected event stream code
- [x] Update affected cli docs
- [x] Update affected tests
- [ ] Update affected docs


Kami:
@lakshmi-kannan @m4dcoder @bigmstone Let's please agree on this data model today if possible so I can finish rest of the changes and get this feature out.

st2common/st2common/models/db/execution.py
```diff
@@ -123,49 +122,39 @@ def get_masked_parameters(self):
...
-
-class ActionExecutionStderrOutputDB(stormbase.StormFoundationDB):
-    """
-    Stores stderr output of a particular action.
-
-    New document is inserted dynamically when a new line is received which means you can simulate
-    tail behavior by periodically reading from this collection.
+    Stores output of a particular execution.
+
+    New document is inserted dynamically when a new chunk / line is received which means you can
```
Mierdin: From this, I gather that this model just represents one "write"? Meaning, for each instance of data that an execution wishes to stream about itself, each would be a unique instance of this model?

st2common/st2common/models/db/execution.py
```diff
@@ -123,49 +122,39 @@ def get_masked_parameters(self):
...
+        timestamp: Timestamp when this output has been produced / received.
+        type: Type of the output (e.g. stdout, stderr, output)
+        data: Actual output data. This could either be line, chunk or similar, depending on the
+              runner.
     """
     execution_id = me.StringField(required=True)
     action_ref = me.StringField(required=True)
+    runner_ref = me.StringField(required=True)
     timestamp = me.DateTimeField(required=True, default=date_utils.get_datetime_utc_now)
+    type = me.StringField(required=True, default='output')
```
Mierdin: I am not a fan of using names of builtins as variable names

m4dcoder:
Do you need rebase this branch? It's including some unrelated commits/fixes from master (unicode, travis, etc.)

st2common/st2common/models/db/execution.py
```diff
@@ -123,49 +122,39 @@ def get_masked_parameters(self):
...
+    Stores output of a particular execution.
+
+    New document is inserted dynamically when a new chunk / line is received which means you can
+    simulate tail behavior by periodically reading from this collection.
+
+    Attribute:
+        execution_id: ID of the execution to which this output belongs.
+        action_ref: Parent action reference.
+        runner_ref: Parent action runner reference.
+        timestamp: Timestamp when this output has been produced / received.
```
m4dcoder: Does it make sense to add an auto-increment sequence field in the collection?  In case the timestamps are the same for one or more entries.

m4dcoder:
LGTM except for a few minor questions regarding unrelated commits in this branch and adding an auto increment sequence field in the ActionExecutionOutput collection.

st2common/st2common/models/db/execution.py
Kami: Correct - mentioned the reason for that in the other PR (performance - updating single document is not a good idea because of how locking works).

st2common/st2common/models/db/execution.py
Kami: I agree, will rename it to `output_type`.

Kami:
> Do you need rebase this branch? It's including some unrelated commits/fixes from master (unicode, travis, etc.)

It was already synced with master and the other branch is based on so no idea why Github is showing this diff the way it is.

st2common/st2common/models/db/execution.py
Kami: Actually, I will just change it to use a custom `ComplexDateTimeField` field we wrote which has micro-second precision and which we use for executions.

I believe microsecond precision should be fine.

Kami:
Will merge those changes into existing PR and work from there.

PR #3734, opened 2017-09-11
Add changelog entry for the fix to Mistral shutdown issue

m4dcoder:
Add a changelog entry for the upstream fix on the Mistral slow shutdown issue with systemd service.


m4dcoder:
Added

PR #3750, opened 2017-09-18
Add support for workflows to the st2 execution tail command

### LINKED ISSUE #3751: 'st2 execution tail --help' doesn't work
When running `st2 execution tail --help` it works as `st2 execution tail`, giving output for the latest execution (I guess).

It should print the help/usage instead.

Kami:
This pull request adds support for workflows to the `st2 execution tail` command.

Right now only single level deep / nested workflows are supported. This was done intentionally, because I think CLI is a bad medium to represent arbitrary nested / deep workflows. It's very hard to display arbitrary nested workflows in the CLI and still make easy to understand what is going on and to which parent workflow a particular execution belongs.

That and other more complicated "visualization" fit better in WebUI, imo.

### Example workflow developed for testing / demo purposes

[asciicast](https://asciinema.org/a/2KSLmUkszPlXlVzCCZcpJlvWV)

### Real-life workflow which is used by "st2 pack install" command

[asciicast](https://asciinema.org/a/79tjBZtshlXosdojKnAFbqn06)

## TODO

- [x] Support for Mistral workflows
- [x] Tests (will be a bit hard and require quite some mocking)
- [x] Robot tests

Also resolves #3751.


Kami:
I hate this piece of code - https://github.com/StackStorm/st2/pull/3750/files#diff-b45a67522cb13a1e385fc5d1b663eed8R1461.

I already mentioned this multiple times in the past. This "add various random attributes to the dictionary without any kind of schema" approach is very error prone and unmaintainable.

Instead of having a dictionary without a schema, we should have a strict schema for `context` attribute. This means that inside the code we should have a class (or similar) which contains all the available context attributes.

For example, something like:

```python
ActionExecutionContect(object):
  parent_execution_id = None
  execution_id = None
  task_name = None
  ...
```

We should then use this class and serialize it as a dictionary and store it on an execution db object. This way we can be sure the schema is stable and the same attributes are always present and we don't need to resort to `.get()` hackkery or similar.

In addition to that, we should store a this normalized context dictionary on the execution. If for some reason, some other runner specific attributes need to be available in the system, we should store them under a special hidden context attribute which is not visible / exposed to the end user.

This way we prevent users from relying on a internal attributes which could change at any point in time.

Code like this is also one of the reasons why dynamic languages get a bad rep.

st2client/tests/unit/test_execution_tail_command.py
```diff
@@ -0,0 +1,310 @@
...
+
+MOCK_OUTPUT_2 = {
+    'execution_id': 'idfoo3',
+    'timestamp': '1505732598',
+    'output_type': 'stderr',
+    'data': 'line 2\n'
+}
+
+
+class ActionExecutionTailCommandTestCase(BaseCLITestCase):
```
Kami: I'm not too happy with the tests because it results in a lot of mocking and that's fragile in case the the format changes.

We have similar problem in many other places so we should figure out a better way to test it. Another option would be to add some robot tests, perhaps in addition to those unit tests.

Kami:
Robot tests in https://github.com/StackStorm/st2tests/pull/120.

m4dcoder:
It's not clear from this PR how workflow executions include events of subtasks.  Can you clarify?  Otherwise, LGTM for 1st pass right now. Long term, I probably want the workflow engine (Mistral, AC, or native) to write/stream events to the ActionExecutionOutput and the tail command can just read from that. For streamed output of individual tasks, user can just tail those separately.

Kami:
> It's not clear from this PR how workflow executions include events of subtasks. Can you clarify? Otherwise, LGTM for 1st pass right now. 

It utilizes the existing event-stream functionality. We dispatch an event for each action_execution and live_action object change.

Based on those events and context / attributes available in those events we then figure out which objects are children objects / tasks of a workflow.

So in short, the event-stream provides a generic interface for that and it's up to consumer to assemble those events together / figure out what is going on (WebUI displays workflows in a similar manner).

We had this discussion in the streaming PR quite a bit already - it really seems you want to add some functionality to streaming which is already available (and was for a while now) as part of the event stream.

I personally don't think that's needed because we already have a good abstraction (event stream) for that.

Kami:
@armab 

> There is a changelog entry missing for this particular sub-feature: it includes 2 bugfixes, but not st2 execution tail in workflows changelog.

I intentionally didn't add changelog entry for this change since we only add those for new changes since the last stable release. In short we don't add changelog entry for things which change as part of the dev versions - users only care about what's new / changed since the last stable release (dev stuff changes all the time).

PR #3753, opened 2017-09-18
Fix cancellation of subworkflow and subchain

m4dcoder:
Cancel of Mistral workflow or Action Chain is cascaded down to subworkflows appropriately. Cancel from tasks in the workflow or chain is cascaded up to the parent.


Kami:
Will this go into v2.5.0 or 2.4.2?

Kami:
LGTM

Issue #3737, opened 2017-09-12
Bug - Untyped enums (core.http.method) error out on the CLI in v2.4


nmaludy:
# Problem

```
[root@stackstorm ~]# st2 --version
st2 2.4.0
[root@stackstorm ~]# st2 run core.http url="https://docs.stackstorm.com/" method="GET"
ERROR: 'type'
```

This problem only exists on v2.4

```
[root@stackstormpuppet4 actions]# st2 --version
st2 2.3.2
[root@stackstormpuppet4 actions]# st2 run core.http url="https://docs.stackstorm.com" method="GET"
.
id: 59b7d90dd6552e551d05ba9f
status: succeeded
parameters: 
  method: GET
  url: https://docs.stackstorm.com
result: 
  body: "\n\n<!DOCTYPE html>\n<!--[if IE 8]><html class=
...
```

# Workaround

By modifying the `core.http` action metadata YAML file and adding `type: string` to the `method` parameter, the CLI works as expected in v2.4.

```
---
description: Action that performs an http request.
enabled: true
entry_point: ''
name: http
parameters:
  auth:
    description: Auth string to be used. This can be like 'x-auth-token=XYZ'.
    type: string
  body:
    description: Body to send with the request
    type: string
  file_content:
    description: Magic attribute which is automatically populated when file_path is
      specified
    type: string
  file_content_type:
    description: Optional content type for the uploaded file
    type: string
  file_name:
    description: Magic attribute which is automatically populated when file_path is
      specified
    type: string
  file_path:
    description: Path to the file to upload
    type: string
  headers:
    description: 'HTTP headers to use with the request.'
    type: object
  method:
    description: HTTP method to use.
    type: string                                         # Add this line to get core.http working with a method='xxx' on the CLI
    enum:
    - HEAD
    - GET
    - POST
    - PUT
    - DELETE
  params:
    description: Query params to be used with the HTTP request.
    type: object
  timeout:
    default: 60
    description: Timeout for the HTTP request.
    type: number
runner_type: "http-request"
```

PR #3754, opened 2017-09-19
Add missing type for the parameters with enum

humblearner:
Fix for: #3737


Kami:
Good catch.

Please also add a corresponding changelog entry.

PR #3756, opened 2017-09-19
Fix delays in st2resultstracker on querying workflow status

m4dcoder:
Fix delays in st2resultstracker on querying workflow status from Mistral. Make sleep time for empty queue and no workers configurable. Reduce the default sleep times to 1 second. StackStorm instances that handle more workflows should consider increasing the query interval and jitter interval for better CPU utilization. For reference, `make mistral-itest` is reduced from 800+s to 500+s to complete.

### DIFF
--- a/conf/st2.conf.sample
+++ b/conf/st2.conf.sample
@@ -180,7 +180,7 @@ api_url = None
 # Multiplier for the exponential backoff.
 retry_exp_msec = 1000
 # Jitter interval to smooth out HTTP requests to mistral tasks and executions API.
-jitter_interval = 1
+jitter_interval = 0
 # Allow insecure communication with Mistral.
 insecure = False
 # Username for authentication.
@@ -205,10 +205,14 @@ retry_exp_max_msec = 300000
 logging = conf/logging.notifier.conf
 
 [resultstracker]
-# Time interval between subsequent queries for a context to external workflow system.
-query_interval = 20
+# Time interval between queries to external workflow system.
+query_interval = 1
+# Sleep delay in between queries when query queue is empty.
+empty_q_sleep_time = 1
 # Location of the logging configuration file.
 logging = conf/logging.resultstracker.conf
+# Sleep delay for query when there is no more worker in pool.
+no_workers_sleep_time = 1
 # Number of threads to use to query external workflow systems.
 thread_pool_size = 10
 
--- a/conf/st2.conf.sample
+++ b/conf/st2.conf.sample
@@ -180,7 +180,7 @@ api_url = None
 # Multiplier for the exponential backoff.
 retry_exp_msec = 1000
 # Jitter interval to smooth out HTTP requests to mistral tasks and executions API.
-jitter_interval = 0
+jitter_interval = 0.1
 # Allow insecure communication with Mistral.
 insecure = False
 # Username for authentication.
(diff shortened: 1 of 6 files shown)


st2common/st2common/config.py
```diff
@@ -262,26 +262,37 @@ def register_opts(ignore_errors=False):
...
         cfg.BoolOpt('insecure', default=False, help='Allow insecure communication with Mistral.'),
-        cfg.FloatOpt('jitter_interval', default=1,
-                   help='Jitter interval to smooth out HTTP requests ' +
-                        'to mistral tasks and executions API.'),
-
-        cfg.StrOpt('api_url', default=None, help=('URL Mistral uses to talk back to the API.'
-            'If not provided it defaults to public API URL. Note: This needs to be a base '
-            'URL without API version (e.g. http://127.0.0.1:9101)'))
+        cfg.FloatOpt(
+            'jitter_interval', default=0,
```
Kami: If needed, one option would also be to use float values (e.g. `0.1`) instead of totally disabling it.

st2common/st2common/config.py
```diff
@@ -262,26 +262,37 @@ def register_opts(ignore_errors=False):
...
+                 'version (e.g. http://127.0.0.1:9101)')
     ]
     do_register_opts(mistral_opts, group='mistral', ignore_errors=ignore_errors)
 
     # Results Tracker query module options
     # Note that these are currently used only by mistral query module.
     query_opts = [
-        cfg.IntOpt('thread_pool_size', default=10,
-                   help='Number of threads to use to query external workflow systems.'),
-        cfg.FloatOpt('query_interval', default=20,
```
Kami: Yeah, I could see how 20 seconds could slow things down a lot :)

Kami:
LGTM.

Perhaps makes sense to also include this in v2.4.2?

conf/st2.conf.sample
```diff
@@ -205,10 +205,14 @@ retry_exp_max_msec = 300000
 logging = conf/logging.notifier.conf
 
 [resultstracker]
-# Time interval between subsequent queries for a context to external workflow system.
-query_interval = 20
+# Time interval between queries to external workflow system.
+query_interval = 1
```
armab: 👍 20s is indeed too much

st2common/st2common/config.py
m4dcoder: Ok. I'll test 0.1 first and then default if all is well

lakshmi-kannan:
@armab @Kami @m4dcoder I shipped with a conservative 20s because this will make users notice and tweak their settings based on their CPU profile. I am fine with 1s but be prepared for support questions because people will complain. Especially, automation suites.

armab:
See https://stackstorm.slack.com/archives/C029K4CBU/p1507113470000011 concern.

Before `query_interval` was `20s`, now it's `1s`.

From the tests it looks like too big diff on CPU utilization on recommended instance https://docs.stackstorm.com/install/system_requirements.html. I think we need to find the value that'll work on that instance type more or less smoothly.

armab:
Created https://github.com/StackStorm/st2/pull/3776 to balance the settings.

`20s` is too conservative, `1s` is too aggressive. With `5s` `st2resultstracker` runs more or less smoothly with 4 workflows on `xlarge` instance we recommend in docs.

PR #3757, opened 2017-09-21
Move Mistral workflows for testing into separate directory

m4dcoder:
Clean up the examples pack and move workflows for integration tests into a separate directory.


lakshmi-kannan:
I am ok with this but what is the reason? Is it because we have too many tests in examples?

m4dcoder:
In the start, we put test workflows in examples so users have more references/examples to look at.  But the number of tests have grew.  So we have started moving the tests workflows out into a separate directory. I may consider moving them to a separate pack altogether in near future.

Issue #3755, opened 2017-09-19
LDAP Group to Role issue with user in two groups that have overlapping roles


krusolu:
Given two group mapping files and a user that is in both LDAP groups.
The user authenticates to StackStorm and only has the roles from groupB.
```
/opt/stackstorm/rbac/mappings/groupA.yaml
---
  group: CN=groupA,OU=Groups,DC=corp,DC=company,DC=test
  description: GroupA Users
  roles:
    - base_permissions
    - role1
    - role2
    - role3

/opt/stackstorm/rbac/mappings/groupB.yaml
---
  group: CN=groupB,OU=Groups,DC=corp,DC=company,DC=test
  description: GroupB Users
  roles:
    - base_permissions
    - role4
    - role5
```
This is evident by running `st2 role-assignment list` commands as well as the action_context.rbac.roles within an action:
```
$ st2 role-assignment list -t 786689b58f5747139c18780c3237d40d|grep userA
| 59c19829968d09215be615e8| base_permissions       | userA             | True      | Automatic role         |
| 59c19829968d09215be615e9| role4                  | userA             | True      | Automatic role         |
| 59c19829968d09215be615ea| role5                  | userA             | True      | Automatic role         |
```

Mongo queries:
```
> db.user_role_assignment_d_b.find({"user": "userA"})
{ "_id" : ObjectId("59c19829968d09215be615e8"), "user" : "userA", "role" : "base_permissions", "description" : "Automatic role assignment based on the remote user membership in group \"CN=groupB,OU=Groups,DC=corp,DC=company,DC=test\"", "is_remote" : true }
{ "_id" : ObjectId("59c19829968d09215be615e9"), "user" : "userA", "role" : "role4", "description" : "Automatic role assignment based on the remote user membership in group \"CN=groupB,OU=Groups,DC=corp,DC=company,DC=test\"", "is_remote" : true }
{ "_id" : ObjectId("59c19829968d09215be615ea"), "user" : "userA", "role" : "role5", "description" : "Automatic role assignment based on the remote user membership in group \"CN=groupB,OU=Groups,DC=corp,DC=company,DC=test\"", "is_remote" : true }
```

If the common `base_permissions` role is removed from one of the groups and `st2-apply-rbac-definitions`  is run again and user logs out/in again, all roles will be activated:
```
$ st2 role-assignment list -t 786689b58f5747139c18780c3237d40d|grep userA
[...]

PR #3763, opened 2017-09-25
Support overlapping RBAC remote role assignments

Kami:
Right now we don't support overlapping role assignments (each role can only be granted by one assignment / mapping file) so it's always clear where a particular role assignment comes from.

It looks like our users have a valid use case for overlapping role assignments (single role being granted by multiple assignment or mapping files - #3755) so we will add support for that.

Once we will add support for that, it will be less obvious and potentially more confusing where a particular role assignment comes from.

Because of that, I modified existing code to store and include information on where a particular assignment come. This should hopefully make it obvious and help with "I removed the role assignment file, but assignment is still there" and related issues.

Before (once we add support for overlapping assignments):

```bash
+--------------------------+-------------+-------+-----------+-------------+
| id                       | role        | user  | is_remote | description |
+--------------------------+-------------+-------+-----------+-------------+
| 59c8ce8d0640fd1d06bac3d0 | observer    | tomaz | False     |             |
| 59c8ce8d0640fd1d06bac3d1 | admin       | testu | False     |             |
| 59c8ce8d0640fd1d06bac3d1 | admin       | testu | False     |             |
| 59c8ce8d0640fd1d06bac3d2 | rule_create | testu | False     |             |
| 59c8ce8d0640fd1d06bac3d2 | rule_delete | testu | False     |             |
| 59c8ce8d0640fd1d06bac3d2 | rule_delete | testu | True      |             |
+--------------------------+-------------+-------+-----------+-------------+
```

After:

```
+--------------------------+-------------+-------+-----------+-------------+------------------------+
| id                       | role        | user  | is_remote | description | metadata.source        |
+--------------------------+-------------+-------+-----------+-------------+------------------------+
| 59c8ce8d0640fd1d06bac3d0 | observer    | tomaz | False     |             | assignments/tomaz.yaml |
| 59c8ce8d0640fd1d06bac3d1 | admin       | testu | False     |             | assignments/testu.yaml |
| 59c8ce8d0640fd1d06bac3d1 | admin       | testu | False     |             | assignments/admin.yaml |
| 59c8ce8d0640fd1d06bac3d2 | rule_create | testu | False     |             | assignments/testu.yaml |
| 59c8ce8d0640fd1d06bac3d2 | rule_delete | testu | False     |             | assignments/testu.yaml |
| 59c8ce8d0640fd1d06bac3d2 | rule_delete | testu | True      |             | mappings/mapping1.yaml |
+--------------------------+-------------+-------+-----------+-------------+------------------------+
```

Part of #3755.

### DIFF
--- a/st2common/st2common/models/db/rbac.py
+++ b/st2common/st2common/models/db/rbac.py
@@ -118,7 +118,6 @@ class GroupToRoleMappingDB(stormbase.StormFoundationDB):
         source: Source where this assignment comes from. Path to a file for local assignments
                 and "API" for API assignments.
         description: Optional description for this mapping.
-        metata: Optional metadata for this mapping.
     """
     group = me.StringField(required=True, unique=True)
     roles = me.ListField(field=me.StringField())
--- a/st2common/st2common/models/db/rbac.py
+++ b/st2common/st2common/models/db/rbac.py
@@ -62,24 +62,26 @@ class UserRoleAssignmentDB(stormbase.StormFoundationDB):
     Attribute:
         user: A reference to the user name to which the role is assigned.
         role: A reference to the role name which is assigned to the user.
+        source: Path to the metadata for this user role assignment.
         description: Optional assigment description.
-        metata: Optional metadata for this assignment.
     """
     user = me.StringField(required=True)
-    role = me.StringField(required=True, unique_with='user')
+    role = me.StringField(required=True, unique_with=['user', 'source'])
+    source = me.StringField(required=True, unique_with=['user', 'role'])
     description = me.StringField()
     # True if this is assigned created on authentication based on the remote groups provided by
     # the auth backends.
     # Remote assignments are special in a way that they are not manipulated with when running
     # st2-apply-rbac-auth-definitions tool.
     is_remote = me.BooleanField(default=False)
-    metadata = me.DictField(default={})
 
     meta = {
         'indexes': [
             {'fields': ['user']},
             {'fields': ['role']},
+            {'fields': ['source']},
             {'fields': ['is_remote']},
+            {'fields': ['user', 'role']},
         ]
     }
 
@@ -112,15 +114,16 @@ class GroupToRoleMappingDB(stormbase.StormFoundationDB):
     Attribute:
         group: Name of the remote auth backend group.
         roles: A reference to the local RBAC role names.
+        source: Path to the metadata for this group to role mapping.
         description: Optional description for this mapping.
         metata: Optional metadata for this mapping.
     """
     group = me.StringField(required=True, unique=True)
     roles = me.ListField(field=me.StringField())
+    source = me.StringField()
     description = me.StringField()
     enabled = me.BooleanField(required=True, default=True,
                               help_text='A flag indicating whether the mapping is enabled.')
-    metadata = me.DictField(default={})
 
 
 # Specialized access objects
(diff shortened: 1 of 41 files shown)


st2common/st2common/models/db/rbac.py
```diff
@@ -63,6 +63,7 @@ class UserRoleAssignmentDB(stormbase.StormFoundationDB):
         user: A reference to the user name to which the role is assigned.
         role: A reference to the role name which is assigned to the user.
         description: Optional assigment description.
+        metata: Optional metadata for this assignment.
```
krusolu: spelling of metadata:

Kami: Thanks, we will actually re-work this a bit.

Spoke to @m4dcoder about it on Slack today and we will actually get rid of metadata, add `source` as a top-level field and add composite index on `user, role, source`.

st2common/tests/unit/services/test_rbac.py
```diff
@@ -203,6 +205,47 @@ def test_grant_and_revoke_role(self):
...
+            source='assignments/%s_2.yaml' % user_db.name)
+
+        role_dbs_2 = rbac_services.get_roles_for_user(user_db=user_db)
+        self.assertItemsEqual(role_dbs_2, [self.roles['custom_role_1']])
+
+        role_dbs = user_db.get_roles()
+        self.assertItemsEqual(role_dbs_2, [self.roles['custom_role_1']])
+
+        # Revoke previously assigned role, should have no roles again
+        rbac_services.revoke_role_from_user(role_db=self.roles['custom_role_1'], user_db=user_db)
```
Kami: I'm a bit confused about this one - just wondering how much sense `revoke_role_from_user` even makes now.

In this case it makes sense to remove all the assignments for a particular user, but not sure about the syncer case. If user is assigned role X from two different assignments and one of those assignments files is deleted, this role should still be there, right? I thought that's that part we are solving.

st2common/st2common/rbac/syncer.py
```diff
@@ -276,57 +282,71 @@ def _sync_user_role_assignments(self, user_db, role_assignment_dbs, role_assignm
...
-        for role_db in role_dbs_to_assign:
-            if role_db.name in role_assignment_api.roles:
-                description = getattr(role_assignment_api, 'description', None)
-            else:
-                description = None
-            assignment_db = rbac_services.assign_role_to_user(role_db=role_db, user_db=user_db,
-                                                              description=description)
+
+        for role in roles_to_create:
+            role_db = list(Role.query(name=role[0]))[0]
```
Kami: This will throw if a role user is to be granted doesn't exist in the database.

That's the use case we supported before - granting roles which don't (yet) exist in the database.

Kami: Actually, looking at the code above - we didn't support that use case, but we didn't throw, now it explodes if referenced role doesn't exist.

Will make a change to throw a more explicit message if a referenced role doesn't exist.

Kami: d8d4dc595da9b49e15aadc73ed378532ea80431b, feeee0718859fd54eb45fa1698f198c74033ad0d.

Kami:
I confirmed end to end that it works as expected for remote mapping files (multiple overlapping role assignments work correctly), but I still have some concerns and of course more end to end testing would be good.

PR #3769, opened 2017-09-28
Add support for glob based event type filtering, fix whitelist for /v1/stream API endpoint

Kami:
This pull request adds support for glob based event type / name filtering to the `/v1/stream` API endpoint.

In addition to that, it fixes a bug caught by @nzlosh - changes I added for streaming support accidentally filtered out `st2.announcement__errbot` events.

If we decide to officially support and maintain errbot we also need to add automated tests for it which will catch issues like that.

Note: This iteration over a list and glob based filtering might not be the most performant - if this indeed turn out to be a bottleneck at some point in the future, we can always optimize it.

Thanks to @nzlosh for catching and reporting this.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -69,7 +69,7 @@ Fixed
   ``st2.announcement__errbot`` and other event names starting with ``st2.announcement__*`` prefix
   are not filtered out. #3769 (bug fix)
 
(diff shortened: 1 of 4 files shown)


Kami:
Note: If we ever decide to do v2.4.2 we can include this change as well, but that's not a big priority.

PR #3773, opened 2017-10-03
New alias execution "match and execute" API endpoint, fix response format for the "alias match" API endpoint

Kami:
This pull request adds new `POST /v1/aliasexecution/match_and_execute` API endpoint.

TL;DR - this API endpoint will allow us to make chat bot plugins much simpler - simple thin wrappers around this API endpoint which just mostly take care of formatting the response correctly for the chat medium in question.

The endpoint takes command string as an input parameter and then tries to match it against aliases and if a match is found, it schedules an execution.

Same end result could already be achieved by using separate "match" and "execute" API endpoints which were contributed by @tonybaloney a while ago, but the idea is to use this API endpoint inside chat bot plugins and it's more efficient to make a single StackStorm API request for each chat message instead of making two.

## TODO

- [x] Tests
- [x] Upgrade notes entry - https://github.com/StackStorm/st2docs/pull/622


Kami:
On a related note - https://github.com/StackStorm/st2/pull/3773/files#diff-a8a854167a1bc5886ec8ccbc185ce896R91

There is no need for this API endpoint to return an array since code ensures it always only returns a single result. This means the API endpoint should really return a dictionary, but I'm not sure if I want to change it at this point.

I don't think we have any official consumers of this endpoint yet, but it would still mean following the standard deprecation process which adds quite some overhead.

st2common/st2common/util/actionalias_matching.py
```diff
@@ -95,3 +97,27 @@ def match_command_to_alias(command, aliases):
...
             results.append((alias, format_string[0], format_string[1]))
     return results
+
+
+def get_matching_alias(command):
+    """
+    Find a matching ActionAliasDB object (if any) for the provided command.
+    """
+    # 1. Get aliases
+    action_alias_dbs = ActionAlias.query(enabled=True)
```
Kami: If it turns out that retrieving all the aliases on each API endpoint is too slow, we can add some logic to cache it in memory inside the API process (controller).

nzlosh:
The errbot plugin uses the match/help api endpoints but changing the code to handle dict instead of array is trivial.

This is a great idea and will move even more logic out of the bot plugin into the core. :+1:

Kami:
> On a related note - https://github.com/StackStorm/st2/pull/3773/files#diff-a8a854167a1bc5886ec8ccbc185ce896R91

There is no need for this API endpoint to return an array since code ensures it always only returns a single result. This means the API endpoint should really return a dictionary, but I'm not sure if I want to change it at this point.

I don't think we have any official consumers of this endpoint yet, but it would still mean following the standard deprecation process which adds quite some overhead.

I will actually go ahead and make that change.

I checked the code and tests - the OpenAPI definition was correct (expecting an object / dict), but we had no test cases for this scenario so we didn't catch this issue :/

Kami:
There we go - ecc580be5497fccf30e65b652abab6510325125c, e4240390c0ec71cc7061746b456893b1f90ebf6c.

This also makes it very obvious why it's important to write test cases for all the various possible scenarios and edge cases and not just some - we had no test case for positive (success) scenario.

In addition to that, it also exposes an issue with our openapi schema which I mentioned in the past - we have no strict mode enabled by default so even though we have definition for a particular API response, API can return invalid data and we won't throw.

Kami:
@nzlosh Here we go - https://github.com/StackStorm/st2docs/pull/622.

It would be the best if you could update your code to handle both scenarios (either an array with a single item or directly a dict) to make upgrade less painful, before you migrate to the new "match and execute" API endpoint.

@tonybaloney ^ Do you happen to utilize action alias match API endpoint anywhere? If you do, this change would affect you.

nzlosh:
OK thanks for the warning!

PR #3776, opened 2017-10-04
Adjust 'resultstracker.query_interval' for more balanced CPU utilization

armab:
In the upcoming release `CPU` utilization for entire StackStorm system could be considerably higher.

- More CPU resources are now allocated by `st2stream` service (because of the new streaming feature for every action)
- Adjusted `st2resultstracker` `st2.conf` settings (https://github.com/StackStorm/st2/pull/3756/files) which improves delay/latency for Mistral workflow status, but eats more CPU by default
- More overhead by `st2rulesengine` I couldn't diagnose yet (https://stackstorm.slack.com/archives/C029K4CBU/p1507114952000035?thread_ts=1507113470.000011&cid=C029K4CBU)

With the https://docs.stackstorm.com/install/system_requirements.html we recommend running StackStorm under `m4.xlarge` instance for prod use.

We're running ourselves `st2cicd` server under better-CPU `c4.xlarge` and it takes all the CPU *just* on running `4` simultaneous e2e workflows which is so minimal usage (not talking about tens of workflows or more running at the same time).

So I think we need to adjust something so the StackStorm will run more or less smoothly with minimal number of workflows under the instance type we recommend for prod.

---

This PR provides more balanced setting for `resultstracker.query_interval` (`5s`) which initially was `20s` and was changed to much CPU-aggressive `1s` in #3756.

But even with such adjustment, we have a problem with CPU profile and need to research more improvements, keeping in mind recent changes and new features for the upcoming `v2.5.0`.


armab:
@m4dcoder See @lakshmi-kannan's PR https://github.com/StackStorm/st2/pull/3413 which initially introduced the `resultstracker.query_interval` and set it default to `20s` as a response to increased CPU usage by the `st2resultstracker` service.
We originally spot it on `st2cicd` server. With the recent `1s` change the problem is back (FYI we already had `1s` hardcoded before).

As said, mistral workflows latency overhead and CPU overusage are both UX issues.

The change introduced recently https://github.com/StackStorm/st2/pull/3756 with `query_interval` changed from `20s` ->`1s` resulted in really aggressive CPU usage.

Since we suggest in https://docs.stackstorm.com/install/system_requirements.html  `m4.xlarge` as production-friendly instance, we should test some workflows on that CPU profile. With the current `st2.conf` settings the CPU goes nuts. I think the expected is to have everything running smoothly on recommended instance with the default StackStorm installation.

Saying that, I think that making `resultstracker.query_interval` `5s`: https://github.com/StackStorm/st2/pull/3776 is a balanced approach between two evils (`20s` vs `1s`): CPU overusage <> mistral workflows timing.

m4dcoder:
Please make sure there's ample information in docs for users to configure this.

armab:
Thanks @m4dcoder, I'll add the requested information.

Will "_Fine Tuning_" section under the https://docs.stackstorm.com/mistral.html workflow doc work?

Or maybe it's better to gather some advices/tips like that under "_Advanced Topics_" of docs?

cc @LindsayHill the documentation master, WDYT?

LindsayHill:
@armab I'm thinking maybe a brief mention of it on /mistral page, with a cross-reference to https://docs.stackstorm.com/troubleshooting/mistral.html#troubleshooting-mistral-workflow-completion-latency for more details?

armab:
@LindsayHill I didn't know it's already documented. Thanks for pointing!
Yes, cross-referencing sounds good to allow users to find it a bit easier.

armab:
@m4dcoder Addressed your concern in docs: https://github.com/StackStorm/st2docs/pull/630

Also there was already section about the `resultstracker.query_interval`.

PR #3786, opened 2017-10-16
Update "st2 apikey load" to update an existing entry if one already exists

### LINKED ISSUE #3748: request: option to skip duplicates when importing apikeys
It would be neat if I could import apikeys without being concerned that a key might already exist. That way using configuration management tools to deploy stackstorm idempotently would be less of a bother.

current result:
```
$ st2 apikey load apikeys.yaml
ERROR: 409 Client Error: Conflict
MESSAGE: Tried to save duplicate unique keys (E11000 duplicate key error collection: st2.api_key_d_b index: key_hash_1 dup key: { : "<snipped>" }) for url: http://127.0.0.1:9101/v1/apikeys
```

suggestion:
```
$ st2 apikey load apikeys.yaml --skip-duplicates
key 'xyz' already exists - skipped!
```

thanks for your consideration

Kami:
This pull request fixes a bug / non-consistent behavior reported in #3748.

It updates ``st2 apikey load`` command so it doesn't throw an updates an existing entry if input file contains `id` attribute for an item and item already exists on the server.

This way behavior is also consistent with ``st2 key load`` command.

See https://gist.github.com/Kami/bc7567c6b2b8a62f7680f64921cbf7fc for example output.

Resolves #3748

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -79,6 +79,12 @@ Fixed
   Reported by Carlos.
 * Fix action-alias execute response to show execution id and matching action-alias #3231 (bug fix)
   Reported by Carlos.
+* Fix ``st2 apikey load`` command to update an existing entry if items in input file contain ``id``
+  attribute and item already exists on the server. This way the behavior is consistent with
+  ``st2 key load`` command and the command is idempotent if each item contains ``id`` attribute.
+  #3748 #3786
+
+  Reported by Christopher Baklid.
 
 2.4.1 - September 12, 2017
 --------------------------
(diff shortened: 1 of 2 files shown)


Kami:
It's also worth noting that API operations exposed for `/apikeys` and `/keys` API endpoints are not the same.

datastore exposes get, set, delete while api key one supports get, create, update, delete so the actual CLI code is not 1:1 mapping and we can't simply just call update as we can do in datastore case.

Kami:
Merged.

Will work on st2tests tests separately since there is really no other way to unit test this.

Issue #3783, opened 2017-10-09
API documentation for PUT /api/v1/keys/{name} is missing TTL


nmaludy:
The API documentation for `PUT /api/v1/keys/{name}` [link](https://api.stackstorm.com/api/v1/keys/#/key_value_pair_controller.put) is missing the `ttl` parameter that's available in the CLI:
```
[root@stackstorm ~]# st2 key set --help
usage: st2 key set [-h] [-t TOKEN] [--api-key API_KEY] [-j] [-y] [-l TTL] [-e]
                   [-s SCOPE] [-u USER]
                   name value

Set an existing key value pair.

positional arguments:
  name                  Name of the key value pair.
  value                 Value paired with the key.

optional arguments:
  -h, --help            show this help message and exit
  -t TOKEN, --token TOKEN
                        Access token for user authentication. Get
                        ST2_AUTH_TOKEN from the environment variables by
                        default.
  --api-key API_KEY     Api Key for user authentication. Get ST2_API_KEY from
                        the environment variables by default.
  -j, --json            Prints output in JSON format.
  -y, --yaml            Prints output in YAML format.
  -l TTL, --ttl TTL     TTL (in seconds) for this value.
  -e, --encrypt         Encrypt value before saving the value.
  -s SCOPE, --scope SCOPE
                        Specify the scope under which you want to place the
                        item.
  -u USER, --user USER  User for user scoped items (admin only).
```

nmaludy:
This can be closed. It's fixed by #3787

PR #3787, opened 2017-10-16
Fix for #3783 - TTL parameter for PUT /api/v1/keys/{name}

nmaludy:
This is a fix for #3783 where the TTL parameter for PUT /api/v1/keys/{name} was missing.

I added the parameter to `st2common/st2common/openapi.yaml.j2` and updated the main definition using `make .generate-api-spec`.

Is there a way, in the future, to auto-generate `st2common/st2common/openapi.yaml.j2` or does this need to be updated manually?

### DIFF
--- a/st2common/st2common/openapi.yaml
+++ b/st2common/st2common/openapi.yaml
@@ -4244,6 +4244,9 @@ definitions:
       scope:
         description: "Scope the item is under. Example: 'user'."
         type: string
+      ttl:
+        description: TTL (in seconds) for this value.
+        type: integer
       user:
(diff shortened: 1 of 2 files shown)


Kami:
Thanks.

As you have figured out in #3788, without auto-generation stuff it's a bit of a losing battle (can / will get out of date and broken), but we can still merge it.

nmaludy:
Not sure why the Travis build failed, doesn't look like anything related to my change... seems like a RabbitMQ failure?

Thoughts?

Kami:
Yeah, looks like a temporary issue.

PR #3790, opened 2017-10-17
Remove unneeded RunnerContainerService class abstraction

Kami:
This pull request removes `RunnerContainerService` class abstraction because there is simply no need for it. This class provided two methods which were actually just wrappers around existing methods inside `st2common.content.utils`.

Even though this class was actually not used in most places, it was copy and pasted around and caused cycling import issues since it was imported inside `st2common` which made it harder to make st2common standalone and not depend on any other `st2*` packages.


Kami:
Also removed cycling import in `st2common.garbage_collection.inquiries`. This also reduced code duplication since we could just re-use `invoke_post_run` function there - 98cc1d8743db25816bed8f6105b8d7a3e681460d, 4e28eafdedd505eb8c667c2565c64dd2e49f8a4c.

/cc @Mierdin - just a heads up. If i'm missing something, please let me know.

PR #3794, opened 2017-10-18
Run tests on PyPy on Travis

Kami:
This pull request updates Travis config so we now also run tests under PyPy there.

I've set `matrix.fast_finish` and `matrix.allow_failures` property for PyPy tests so they don't block the whole build - whole build is considered as finish as soon as other, non PyPy tests finish (https://docs.travis-ci.com/user/customizing-the-build/#Fast-Finishing).

I've done this to speed things up, because also waiting on PyPy builds would slow the whole build down for another 8-10 minutes which is unacceptable right now.

Down the road, we do need to finish a way to speed it up (perhaps buy more parallel contains) and make it part of the build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ dist: precise
 language: python
 python:
   - 2.7
+  - pypy
 
 branches:
   only:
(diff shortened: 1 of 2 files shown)


m4dcoder:
@Kami @armab You guys resolve your differences? Is this PR still applicable?

PR #3803, opened 2017-10-23
Make st2common package and Python runner standalone

Kami:
This pull request makes various changes to the Python runner to make it standalone. This means it's totally independent (all the arguments / parameters are passed to it) and it only depends on st2common.

This means it can be used inside the environments which don't have access to the database, message bus, etc. (e.g. serveless). A nice "side benefit" of this change is that it also makes testing (testing by the user during the action development and unit testing by StackStorm developers) easier - instead of loading things inside the runner class, we pass it to it (aka dependency injection).

Another nice "side benefit" is that very simple and short Python runner actions now finish a bit faster because we don't need connection to the database inside actions anymore. Not needing connection to the database inside the runner code is better because it means actions can now also run on servers which don't have access to the database.

One of the changes involved moving config loading and parsing outside runner class to the runner container (config is now passed to the runner via command line arguments). I think that's a better abstraction anyway and it means we can eventually also more easily add support for configs to the other runners.

Note: Those changes are happening last minute before v2.5.0 release will start, so they should go into v2.6.0 so we will have enough time to test them.


Kami:
Working on this also exposed some issues - we didn't have clear standards and rules for runners so some "bad practices" managed to slip it.

Runners should be self contained and all the information they need should be passed to them (auth token, parameters, config, etc.).

To be able to remove database dependency from the runner, I needed to update `DatastoreService` to utilize temporary auth token generated by the runner container which is valid for the duration of the action life time. That's how it worked originally, but then we received a contribution which adds support for `DatastoreService` to sensors. That PR changed it to generate and renew token inside the `DatastoreService`.

This breaks the abstractions so eventually we need to update that so for sensors, tokens are generated and managed by the SensorContainer and not by the `DatastoreService`.

Kami:
With changes in #3809 and this PR we are also now back in business Python runner performance wise:

```bash
usage: python_action_wrapper.py [-h] --pack PACK --file-path FILE_PATH
                                [--config CONFIG] [--parameters PARAMETERS]
                                [--user USER] [--parent-args PARENT_ARGS]
python_action_wrapper.py: error: argument --pack is required

real	0m0.195s
user	0m0.030s
sys	0m0.083s
```

Kami:
This pull request kinda served as a working branch for serverless stuff, but here is a quick recap and goal of this PR so it's easier to review and have some context.

This PR has three main goals:

1. Make ``st2common`` package standalone so it can be re-used without depending on any other st2 components. This makes it possible to be used in serverless environment and in the future to also be used by ``st2client`` and others.
2. Make Python runner standalone so it's only dependency is ``st2common`` package.
3. Make action runners Python packages so they can be installed using pip / setuptools. This doesn't conflict with the existing distribution mechanism, it's just an addition / another way to distribute the runners.

To get 2) to work, I needed to make some changes:

1. Move some of the functionality from the runner and process wrapper to the runner container service. Most notable one is retrieving pack config. Previously it was retrieved inside the wrapper and now it's retrieved inside the container and passed to the runner.
2. Bunch of refactoring in the `DatastoreService`. Previously it needed access to the database and generated API tokens by calling `create_token` function. This is no need for that and DB access since we already generate temporary auth token inside the runner container service. In the future we also need to refactor sensor wrapper to do the same - this will make it easier to run sensors on servers which don't have access to the DB and scale it out.

All of those changes are good overall - they provide better decoupling and API. It also makes it easier to potentially also add support for pack configs to other runners in the future.

Issue #3681, opened 2017-08-17
Jinja not rendered in rule trigger parameters


nmaludy:
I have a rule that i'm trying to define using a cron timer with values from the key/value store.

```
---
name: full_backup_cron
pack: backups
description: "Executes a full backup on a cron schedule."
enabled: true

trigger:
  type: "core.st2.CronTimer"
  # http://apscheduler.readthedocs.io/en/3.0/modules/triggers/cron.html#api
  parameters:
      timezone: "{{ st2kv.system.backups.full_backup_cron.timezone }}"
      day_of_week: "{{ st2kv.system.backups.full_backup_cron.day_of_week }}"
      hour: "{{ st2kv.system.backups.full_backup_cron.hour }}"
      minute: "{{ st2kv.system.backups.full_backup_cron.minute }}"
      second: "{{ st2kv.system.backups.full_backup_cron.second }}"
  
action:
  ref: "backups.full_backup"
```

When i try to register the rule i get the following error in `st2api` logs.

```
2017-08-16 21:28:18,008 83527792 AUDIT auth [-] Token with id "59946e2aa814c0066070ade8" is validated.
2017-08-16 21:28:18,020 83527792 ERROR router [-] Failed to call controller function "put" for operation "st2api.controllers.v1.rules:rule_controller.put": '{{ st2kv.system.backups.full_backup_cron.timezone }}'
Traceback (most recent call last):
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/router.py", line 426, in __call__
    resp = func(**kw)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2api/controllers/v1/rules.py", line 145, in put
    rule_db = RuleAPI.to_model(rule)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/models/api/rule.py", line 238, in to_model
    parameters=parameters)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/validators/api/reactor.py", line 106, in validate_trigger_parameters
    CronTrigger(**parameters)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/apscheduler/triggers/cron/__init__.py", line 50, in __init__
    self.timezone = astimezone(timezone)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/apscheduler/util.py", line 78, in astimezone
    return timezone(obj)
[...]

PR #3804, opened 2017-10-23
Allow Jinja variables in trigger with parameters

lakshmi-kannan:
Closes: https://github.com/StackStorm/st2/issues/3681

I am iffy about this change. Though the fix is simple (which is to just render the jinja values prior to registering the trigger), a user will lose visibility of the Jinja expression and just see rendered values in the rule. This can lead to confusions if users end up changing the jinja values later and expect the rule to automagically change. Plus, it can be confusing for users as to what kind of jinja expressions are allowed for triggers with parameters. But this confusion is not specific to this particular case. We have the same problem in a bunch of places where we allow Jinja expressions. 

While on this, I fixed the 500 that was thrown out when trigger parameters validation failed. A catch for CronTrigger is that the exception thrown is along of the lines of UnknownTimeZoneError class with message like 'pst'. So bubbling this exception all the way leads to UX issues where CLI reports something like ERROR: pst for url BLAH. So I fixed that by logging the actual exception and bubbling a generic message like "Cron trigger parameters do not match expected format" along with the actual validation error.

Once we agree on whether to actual merge this "feature", I'll add tests and documentation. I am unable to make up my mind. I am leaning towards not adding this feature myself. 

## TODO

* [x] Convince self 
* [x] Convince others 
* [x] Tests
* [ ] Documentation [We don't talk about triggers with parameters!]
* [ ] Changelog

### After

```
(virtualenv)vagrant@st2dev /m/s/s/st2 ❯❯❯ st2 rule create /opt/stackstorm/packs/examples/rules/backup.yaml
+-------------+----------------------------------------------------------+
| Property    | Value                                                    |
+-------------+----------------------------------------------------------+
| id          | 59ee08ffd9d7ed3b818877b6                                 |
| name        | full_backup_cron                                         |
| pack        | examples                                                 |
| description | Executes a full backup on a cron schedule.               |
| action      | {                                                        |
|             |     "ref": "backups.full_backup",                        |
|             |     "parameters": {}                                     |
|             | }                                                        |
| criteria    |                                                          |
| enabled     | True                                                     |
| ref         | examples.full_backup_cron                                |
| tags        |                                                          |
| trigger     | {                                                        |
|             |     "type": "core.st2.CronTimer",                        |
|             |     "ref": "core.3c45a9b4-e3bd-4f0a-9561-ade73e0c35ee",  |
|             |     "parameters": {                                      |
|             |         "timezone": "",                                  |
|             |         "second": "33",                                  |
|             |         "day_of_week": "4",                              |
|             |         "hour": "3",                                     |
|             |         "minute": "23"                                   |
|             |     }                                                    |
|             | }                                                        |
| type        | {                                                        |
|             |     "ref": "standard",                                   |
|             |     "parameters": {}                                     |
|             | }                                                        |
| uid         | rule:examples:full_backup_cron                           |
+-------------+----------------------------------------------------------+
(virtualenv)vagrant@st2dev /m/s/s/st2 ❯❯❯ cat /opt/stackstorm/packs/examples/rules/backup.yaml      issue_3681/jinja_rule_trigger_params ✭
---
name: full_backup_cron
pack: examples
description: "Executes a full backup on a cron schedule."
enabled: true

trigger:
  type: "core.st2.CronTimer"
  # http://apscheduler.readthedocs.io/en/3.0/modules/triggers/cron.html#api
  parameters:
      timezone: "{{ st2kv.system.backups.full_backup_cron.timezone }}"
      day_of_week: "{{ st2kv.system.backups.full_backup_cron.day_of_week }}"
      hour: "{{ st2kv.system.backups.full_backup_cron.hour }}"
      minute: "{{ st2kv.system.backups.full_backup_cron.minute }}"
      second: "{{ st2kv.system.backups.full_backup_cron.second }}"

action:
  ref: "backups.full_backup"
```


st2common/st2common/validators/api/reactor.py
```diff
@@ -103,7 +104,12 @@ def validate_trigger_parameters(trigger_type_ref, parameters):
         # Validate that the user provided parameters are valid. This is required since JSON schema
         # allows arbitrary strings, but not any arbitrary string is a valid CronTrigger argument
         # Note: Constructor throws ValueError on invalid parameters
-        CronTrigger(**parameters)
+        try:
+            CronTrigger(**parameters)
+        except:
+            msg = ('Cron trigger parameters do not match expected format.')
```
lakshmi-kannan: See description for reasons.

m4dcoder:
@lakshmi-kannan This PR is approved.  Is this still applicable?  If so, can you rebase, get CI to pass, and merge?

lakshmi-kannan:
@m4dcoder I have to write tests and docs etc. I'll pick this up.

st2common/st2common/models/api/rule.py
```diff
@@ -235,6 +241,17 @@ def to_model(cls, rule):
...
         parameters = trigger.get('parameters', {})
 
+        context = {}
+        context.update({
+            kv_constants.DATASTORE_PARENT_SCOPE: {
+                kv_constants.SYSTEM_SCOPE: kv_service.KeyValueLookup(
+                    scope=kv_constants.FULL_SYSTEM_SCOPE)
+            }
+        })
+        parameters = jinja_utils.render_values(mapping=parameters, context=context,
```
m4dcoder: By rending in the RuleAPI, will this render and expose secrets when getting the rule from the CLI?

m4dcoder: I prefer the rendering is done somewhere else. When getting the rule via API/CLI, users should see the expressions. 1) It won't confuse user with what's actually is defined for the rule 2) Secrets if any are not displayed.

lakshmi-kannan: @m4dcoder We already discussed this in the issue ^ and came to the decision that we want to render in the rule registration. We simply cannot render the parameters later because the rule needs to be registered with type and actual parameters. Rule matching happens with this info. 

I see the problems with secret however. Unfortunately, our rule model doesn't have a `secret` metadata flag for parameters. So we'd expose secrets. Good catch. To "fix" this, we'd have to update rule model. @dzimine @nmaludy This is getting a little hairy.

lakshmi-kannan:
Not working on this actively. Closing.

PR #3809, opened 2017-10-24
Fix performance regressions and overhead for Python runner actions

Kami:
In the past I spent quite a lot of time working on optimizing the run time of simple and short Python runner actions (there was a lot of overhead added by start up time, mostly by some very slow imports).

I managed to get it down to 0.2 seconds or so, but we recently moved some modules around and added some function definitions to some modules which introduced big performance overheads (run time is now back at ~2 seconds).

The reason for that was a broken test - the test which should ensure that we don't introduce performance regressions didn't work correctly because it didn't actually check that the script it tries to run exists (it broke when we moved wrapper script from `st2actions` to `st2common`).

As a first step, I of course fixed this test so things like that won't happen again - 69a63f55fca9c9636365b26016e83da5b119f7af.

I also managed to track down performance regressions we introduced and get run time down to around 0.5s - 9f7febad96a343cd032b9c651f78ff82d3e1630b. This is still not where we were originally, but with changes here and changes in #3803 (merged in time for v2.6), run time is down back to 0.2-0.3 seconds again.

Before:

```bash
(virtualenv)vagrant@vagrant-ubuntu-trusty-64:/data/stanley$ python /data/stanley/st2common/st2common/runners/python_action_wrapper.py --is-subprocess
usage: python_action_wrapper.py [-h] --pack PACK --file-path FILE_PATH
                                [--config CONFIG] [--parameters PARAMETERS]
                                [--user USER] [--parent-args PARENT_ARGS]
python_action_wrapper.py: error: argument --pack is required
(virtualenv)vagrant@vagrant-ubuntu-trusty-64:/data/stanley$ time python /data/stanley/st2common/st2common/runners/python_action_wrapper.py --is-subprocess
usage: python_action_wrapper.py [-h] --pack PACK --file-path FILE_PATH
                                [--config CONFIG] [--parameters PARAMETERS]
                                [--user USER] [--parent-args PARENT_ARGS]
python_action_wrapper.py: error: argument --pack is required

real	0m1.779s
user	0m0.371s
sys	0m0.691s
(virtualenv)vagrant@vagrant-ubuntu-trusty-64:/data/stanley$ python /data/stanley/st2common/st2common/runners/python_action_wrapper.py --is-subprocess
usage: python_action_wrapper.py [-h] --pack PACK --file-path FILE_PATH
                                [--config CONFIG] [--parameters PARAMETERS]
                                [--user USER] [--parent-args PARENT_ARGS]
python_action_wrapper.py: error: argument --pack is required
(virtualenv)vagrant@vagrant-ubuntu-trusty-64:/data/stanley$ time python /data/stanley/st2common/st2common/runners/python_action_wrapper.py --is-subprocess
usage: python_action_wrapper.py [-h] --pack PACK --file-path FILE_PATH
                                [--config CONFIG] [--parameters PARAMETERS]
                                [--user USER] [--parent-args PARENT_ARGS]
python_action_wrapper.py: error: argument --pack is required

real	0m1.779s
user	0m0.371s
sys	0m0.691s
```

After:

```bash
(virtualenv)vagrant@vagrant-ubuntu-trusty-64:/data/stanley$ time python /data/stanley/st2common/st2common/runners/python_action_wrapper.py --is-subprocess
usage: python_action_wrapper.py [-h] --pack PACK --file-path FILE_PATH
                                [--parameters PARAMETERS] [--user USER]
                                [--parent-args PARENT_ARGS]
python_action_wrapper.py: error: argument --pack is required

real	0m0.517s
user	0m0.070s
sys	0m0.225s
```

/cc @bigmstone @m4dcoder - just a heads up - some imports like eventlet, etc. can add a lot of overhead and that's why the imports were organized as they were before to avoid this overhead in python runner wrapper script

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -112,6 +112,8 @@ Fixed
 
   Note: This is a breaking change so if your code utilizes this API endpoint you need to update
   to treat response as a dictionary and not as an array with a single item. #377
+* Partially fix performance overhead and regression for short and simple Python runner actions.
+  Full / complete fix will be included in v2.6.0. #3809
 
 Changed
 ~~~~~~~
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,6 @@ argcomplete
 bcrypt
 eventlet==0.19.0
 flex==6.10.0
-git+https://github.com/Kami/entrypoints.git@dont_use_backports#egg=entrypoints
 git+https://github.com/Kami/logshipper.git@stackstorm_patched#egg=logshipper
 git+https://github.com/StackStorm/python-mistralclient.git#egg=python-mistralclient
 git+https://github.com/StackStorm/st2-auth-backend-flat-file.git@master#egg=st2-auth-backend-flat-file
--- a/st2actions/tests/integration/test_python_action_process_wrapper.py
+++ b/st2actions/tests/integration/test_python_action_process_wrapper.py
@@ -41,7 +41,8 @@ __all__ = [
 ]
 
 # Maximum limit for the process wrapper script execution time (in seconds)
-WRAPPER_PROCESS_RUN_TIME_UPPER_LIMIT = 0.20
+# TODO: Revert back to 0.20 once all performance fixes are back in
+WRAPPER_PROCESS_RUN_TIME_UPPER_LIMIT = 0.60
 
 ASSERTION_ERROR_MESSAGE = ("""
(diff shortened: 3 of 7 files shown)


koenbud:
Great to see this getting attention. In some environments a couple of seconds has a lot of impact. Thanks for the progress.

Kami:
@koenbud Yeah - I also fixed / improved the test so hopefully regressions like that won't happen in the future anymore.

PR #3810, opened 2017-10-24
st2 CLI Formatted Result Status and Timestamps

nmaludy:
# Problem

Previously when executing an action i was commonly thinking that my action, or more routinely my workflow, succeeded even though it had not. I realized this was caused by a workflow having a bunch of sub-tasks and each of those succeeding, but not realizing the main workflow had failed. It was hard to tell that the main workflow had failed and the failure definitely didn't "stand out". In all fairness the information was right in front of me in `status: failed`. 

To make this "pop" and i decided to colorize the the `status:` field in the result output on the CLI. As a result this also formats the `start_timestamp` and `end_timestamp` consistently with the format in the sub-task table.

### Action - Before

image

### Action - After

image

## Workflow - Before

image

## Workflow - After

image

## Workflow with succeeded tasks, but failed workflow - Before

image

## Workflow with succeeded tasks, but failed workflow - After

image


Kami:
Nice work :+1: 

I agree that we should utilize colors as much as possible in the default user-friendly CLI output (where it makes sense and adds clarify of course).

Kami:
(thinking out loud)

Since you are using existing code infrastructure for handling colors, I think it should also work with terminal emulators which don't support colors so we should be fine there.

Kami:
One place which might need updating is tests in https://github.com/stackstorm/st2tests.

Perhaps it makes sense to add global `no_colors = True` or similar CLI config option to make it a little less work for other generic st2tests tests which don't specifically test color functionality.

Kami:
Merged, thanks.

Will add changelog entry in a separate PR.

PR #3821, opened 2017-10-31
Allow disabling sandboxing on the python runner level

enykeev:


Kami:
LGTM, if it's backward compatible (should be) and non-breaking. I'm fine with this approach.

But there is actually more virtualenv related code in the runner which probably needs to be updated and we need some tests for it.

enykeev:
> But there is actually more virtualenv related code in the runner which probably needs to be updated and we need some tests for it.

There is, but from what I can say, the rest is fine. We still store pack dependencies separately from each other, we just use the same executable to run them.

Kami:
In this case we should perhaps call the parameter differently e.g. `use_python_binary_from_sandbox` or similar.

And adding a docstring which clarifies what this argument does would also be good.

enykeev:
docstring - sure, but I don't see the reason to be especially verbose with argument name. Why not rename timeout to `time_to_wait_before_we_consider_action_stuck`?

PR #3824, opened 2017-11-01
Use the correct class name in the log messages generated by Python runner actions

Kami:
Previously, we didn't correctly retrieve the name of the class of the action so it was always incorrectly logged as "ABCMeta" (this is the parent class of each action).

Before:

```yaml
result: 
  exit_code: 0
  result: false
  stderr: 'st2.actions.python.ABCMeta: DEBUG    test log message

    st2.actions.python.ABCMeta: DEBUG    value=10

    '
  stdout: ''
```

After:

```yaml
result: 
  exit_code: 0
  result: false
  stderr: 'st2.actions.python.PrimeCheckerAction: DEBUG    test log message

    st2.actions.python.PrimeCheckerAction: DEBUG    value=10

    '
  stdout: ''
```

## TODO

- [x] Changelog entry
- [x] Test


Kami:
Based on the feedback from a customer, I also decided to add new `log_level` runner-level parameter to Python runner.

With this parameter, user can control which log messages from Python runner actions are displayed and output to `stderr`. For backward compatibility reasons it defaults to `debug`.

The reason for adding it is that some actions depend on external libraries which output a lot of info under DEBUG log level and sometimes you just want to see info or above or similar.

PR #3825, opened 2017-11-03
Mask 'secret' properties in inquiry responses

### LINKED ISSUE #3822: Inquiry Responses Do Not Mask Secrets
# Problem

When responding to an inquiry with `st2 inquiry respond xxx` the response data displayed does not mask out properties marked as `secret: true`.

# Steps to reproduce

## Action Metadata
``` yaml
---
description: "Restarts a service on the system (with duo integrated 2FA)"
enabled: true
runner_type: "mistral-v2"
entry_point: workflows/service_restart_v3.yaml
name: service_restart_v3
pack: demo_inquiries
parameters:
  service:
    type: string
    description: "Name of the service to restart."
    required: true

```

## Workflow
``` yaml
version: '2.0'

demo_inquiries.service_restart_v3:
  description: Restarts a service on the system (with duo integrated 2FA)
  type: direct
  input:
    - service

  output:
    result: "{{ _.result }}"
    
  tasks:
    inquiry_2fa:
       action: core.ask
       input:
         route: developers
         schema:
           type: object
           properties:
             username:
               type: string
               description: "Please enter Duo username for restarting the '{{ _.service }}' service"
               required: true
             passcode:
               type: string
               description: "Please enter Duo passcode (generated on the mobile app) for restarting the '{{ _.service }}' service"
               required: true
               secret: true
       on-success:
         - duo_2fa

    duo_2fa:
      action: duo.auth_auth
      input:
        username: "{{ task('inquiry_2fa').result.response.username }}"
        factor: passcode
        passcode: "{{ task('inquiry_2fa').result.response.passcode }}"
      on-success:
        - service_restart
              
    service_restart:
      action: core.local_sudo
      input:
        cmd: "systemctl restart {{ _.service }}"
      publish:
        result: "{{ task('service_restart').result }}"
```

## Commands

``` shell
$ st2 run demo_inquiries.service_restart_v3 service=crond
.
id: 59f772a1032fb64ea8a99bd0
action.ref: demo_inquiries.service_restart_v3
parameters: 
  service: crond
status: pausing
start_timestamp: 2017-10-30T18:42:41.554471Z
end_timestamp: None
+--------------------------+----------------------+-------------+----------+-----------------+
| id                       | status               | task        | action   | start_timestamp |
+--------------------------+----------------------+-------------+----------+-----------------+
| 59f772a2032fb64ea8a99bd3 | running (2s elapsed) | inquiry_2fa | core.ask | Mon, 30 Oct     |
|                          |                      |             |          | 2017 18:42:42   |
|                          |                      |             |          | UTC             |
+--------------------------+----------------------+-------------+----------+-----------------+


# note the secret is masked out when receiving input, but not when displaying the response
$ st2 inquiry respond 59f772a2032fb64ea8a99bd3
passcode (secret): ******
username: testuser

 Response accepted. Successful response data to follow...
+----------+----------------------------+
| Property | Value                      |
+----------+----------------------------+
| id       | 59f772a2032fb64ea8a99bd3   |
| response | {                          |
|          |     "passcode": "123456",  |
|          |     "username": "testuser" |
|          | }                          |
+----------+----------------------------+
```

Mierdin:
This PR addresses some concerns about show secrets in Inquiry responses raised in #3822. Inquiry response schemas can declare certain values to be `secret`, but those values are still shown in the clear in a few places. This PR adds masking capabilities for these:

- Masks secret values in the inquiry response shown at the CLI immediately following a valid response.
- Adds functionality to the execution data model to mask values in the response (this is done because Inquiries still use the ActionExecution data model. This will be moved once Inquiries get their own DB model). This prevents secret values from being leaked via `st2 execution get <id>`, etc.

> NOTE: This doesn't handle masking secrets in nested objects in an Inquiry schema. Nested objects aren't currently supported in general for Inquiry schemas, and this works the same way.

Closes #3822


st2client/st2client/commands/inquiry.py
```diff
@@ -127,19 +127,29 @@ def __init__(self, resource, *args, **kwargs):
...
             instance.response = json.loads(args.response)
         else:
-            inquiry = self.get_resource_by_id(id=args.id, **kwargs)
             response = InteractiveForm(
                 inquiry.schema.get('properties')).initiate_dialog()
             instance.response = response
 
-        return self.manager.update(instance, **kwargs)
+        # Send the response to the API first, then sanitize secrets
+        instance = self.manager.update(instance, **kwargs)
```
Kami: Hm, I'm a bit confused - shouldn't the masking happen on the server (API) side otherwise it kinda defeats the purpose?

That's how we do it for other endpoints. In addition to that, we support `?show_secrets` query param which can be used by admin to get the response back unmasked.

Kami: Edit: I needed to expand the diff a bit to see more context.

The actual response and potential secrets are provided by the user and sent to the API (and not just returned to the user by the API) so the sanitization for `response` command is correct :)

And as far as get one and get all API endpoints go - we don't actually return the response provided by the user and which is stored as liveaction result as part of those APIs so we should be fine, right?

Mierdin: Actually I didn't think through that well enough. For some reason I "remembered" that we aren't storing the response so didn't need to worry about it.....but of course this is wrong because we obviously need to use the response data for later tasks in the workflow.

This level of masking on the client-side, however, isn't wasted work - should still be done. I just need to continue on the server side.

Mierdin:
@Kami I elected to make changes down at the model layer, rather than try to figure out all the ways that info might leak out. As a result, the values are masked when retrieving executions:

```
(virtualenv) vagrant@st2dev:~/st2$st2 execution get 5a02497632ed3540243001ab
id: 5a02497632ed3540243001ab
status: succeeded (0s elapsed)
parameters:
  route: developers
  schema:
    properties:
      secondfactor:
        description: Please enter second factor for authenticating to "foo" service
        required: true
        secret: true
        type: string
    type: object
result:
  response:
    secondfactor: '******'
  roles: []
  route: developers
  schema:
    properties:
      secondfactor:
        description: Please enter second factor for authenticating to "foo" service
        required: true
        secret: true
        type: string
    type: object
  ttl: 1440
  users: []
```

The way I did it also means that `?show_secrets` can be used to show/hide these as well.

st2common/st2common/util/secrets.py
```diff
@@ -58,3 +58,24 @@ def mask_secret_parameters(parameters, secret_parameters):
...
+
+    :param schema: Inquiry response schema
+    :type schema: ``dict``
+    """
+    result = copy.deepcopy(response)
+
+    for prop_name, prop_attrs in schema['properties'].items():
+        if prop_attrs.get('secret') is True:
+            if prop_name in response:
+                result[prop_name] = "******"
```
Kami: You could re-use `st2common.constants.secrets.MASKED_ATTRIBUTE_VALUE`

st2common/st2common/models/db/execution.py
```diff
@@ -110,6 +111,14 @@ def mask_secrets(self, value):
         if 'parameters' in liveaction:
             liveaction['parameters'] = mask_secret_parameters(parameters=liveaction['parameters'],
                                                               secret_parameters=secret_parameters)
+
+        # TODO(mierdin): This logic should be moved to the dedicated Inquiry
+        # data model once it exists.
+        if self.runner.get('name') == "inquirer":
+            response = result['result'].get('response')
+            schema = result['result'].get('schema')
```
Kami: Should probably do `schema = schema or {}` or similar here or in `mask_inquiry_response` since it seems like now it would explode in case schema is `None`.

(on a related note - is it even possible for schema to be empty and / or not have `properties` attribute?)

st2common/tests/unit/test_db_execution.py
```diff
@@ -0,0 +1,130 @@
...
+
+        Should properly mask these if the Inquiry is being retrieved
+        directly via `execution get` commands.
+
+        TODO(mierdin): Move this once Inquiries get their own data model
+        """
+
+        # Test Inquiry response masking is done properly within this model
+        masked = self.retrieved.mask_secrets(self.retrieved.to_serializable_dict())
+        self.assertEqual(masked['result']['response']['secondfactor'], "******")
```
Kami: Would be good to also test this in the API layer through API controller test case.

st2common/st2common/util/secrets.py
Mierdin: Good call, done in https://github.com/StackStorm/st2/pull/3825/commits/2e339b8252378c95c8848ae110a25dc09a076e7b

st2common/st2common/models/db/execution.py
Mierdin: Very good point. I tested this and in fact you *can* create an Inquiry with an empty schema. This results in bad behavior at respond time that I documented in https://github.com/StackStorm/st2/issues/3846 - I'll make sure to fix that in an upcoming release.

In the meantime, I changed this line to do an explicit key lookup for `schema` rather than using `.get()`. Unlike the `response` key, the `schema` key will always be there (enforced at the `inquirer` runner level), and this code shouldn't care what the contents of the schema are, it should only pass it to `mask_inquiry_response`.

st2common/tests/unit/test_db_execution.py
Mierdin: Added this test in https://github.com/StackStorm/st2/pull/3825/commits/881ecbaac3733ddbe6956f8079024283b9ed769e

st2common/st2common/models/db/execution.py
Mierdin: Still not working - looking at this now

Mierdin: Okay should be good now. In https://github.com/StackStorm/st2/pull/3825/commits/bf57136fc2d0dc4812a87fd1702beb6a864c4f19 I decided to revert back to `.get()` and default to an empty dict. In addition, I'm checking to ensure it's not empty before moving on to the masking function.

Kami:
LGTM

PR #3830, opened 2017-11-06
Upgrade various Python deps to the latest stable versions

Kami:
This pull request updates various Python libraries we use to latest stable versions which mostly include bug-fixes.

(Sadly we are still blocked on upgrading eventlet, mongoengine and requests due to some breaking API changes and performance issues)


lakshmi-kannan:
Is this a good candidate for a patch release or do you want to wait until next major release? cc: @humblearner

Issue #3823, opened 2017-11-01
get_value fails with " 403 Client Error: Forbidden\nMESSAGE: Decrypt option requires administrator access for url: " for encrypted keys


dpasupathi:
@kami:
Using data store action_service APIs tried to store and retrieve values in action context.
(similar to https://github.com/StackStorm/st2tests/blob/ddb6c8243fcacc03e12cb8739b81b68402705214/packs/fixtures/actions/pythonactions/datastore_test_action.py#L48)

My simple code snippet looks as follows:
self.action_service.set_value(name=lookup_key, value=passwd, local=False, encrypt=True)

user_kv = self.action_service.get_value(name=lookup_key, local=False, decrypt=decrypt)

When i run the action using "st2 run" with admin privilege  the retrieval failed with following error:
"""
Retrieving value from the datastore (name=switch.1.2.3.6.passwd)\nst2.actions.python.ABCMeta: ERROR    Exception retrieving value from datastore (name=switch.1.2.3.6.passwd): 403 Client Error: Forbidden\nMESSAGE: Decrypt option requires administrator access for url: http://127.0.0.1:9101/v1/keys/switch.1.2.3.6.passwd?scope=system&decrypt=true\nTraceback (most recent call last):\n  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/services/datastore.py\", line 98, in get_value\n    kvp = client.keys.get_by_id(id=name, params=params)\n
"""

But i run the same via CLI  "st2 key get" it succeeds.
```
user@ubuntu-bwc:/opt/stackstorm/packs/network_essentials/actions$ st2 key get switch.1.2.3.6.passwd --decrypt

+------------------+----------------------------+

| Property         | Value                      |

+------------------+----------------------------+

| name             | switch.1.2.3.6.passwd      |

| value            | password                   |

| secret           | True                       |

| encrypted        | False                      |

| scope            | st2kv.system               |

| expire_timestamp |                            |

+------------------+----------------------------+
```

***Environment:***

```
user@ubuntu-bwc:/opt/stackstorm/packs/network_essentials/actions$ st2 whoami

Currently logged in as admin

user@ubuntu-bwc:/opt/stackstorm/packs/network_essentials/actions$ st2 role list

+--------------------------+--------------+--------+--------------+

| id                       | name         | system | description  |

+--------------------------+--------------+--------+--------------+

| 59f25901ccad682d45c06810 | admin        | True   | admin        |

[...]

PR #3831, opened 2017-11-06
Add new "/v1/api/user" API endpoint which returns information about then authenticated user

Kami:
This pull request updates "root" API controller (`/`,` /v1`) to include metadata information about the authentication user in addition to metadata about the StackStorm installation.

Before:

```json
curl -X GET -H  'X-Auth-Token: foo' 'http://127.0.0.1:9101/v1/'

{
    "version": "2.6dev",
    "docs_url": "http://docs.stackstorm.com/latest",
}
```

After:

```json

curl -X GET -H  'X-Auth-Token: foo' 'http://127.0.0.1:9101/v1/'


{
    "version": "2.6dev",
    "docs_url": "http://docs.stackstorm.com/latest",
    "user": {
        "username": "testu",
        "rbac": {
            "enabled": true,
            "roles": [
                "admin",
                "rule_create"
            ]
        },
        "authenticated": true
    }
}
```

This should make troubleshooting issues like this (https://github.com/StackStorm/st2/issues/3823) where it's not immediately clear with which user action is authenticating easier.

Instead of adding this information to the root controller, another option would be to add new `/v1/whoami` controller (or similar) and include that information there.

What do others think?

Once we agree on the approach, I will add add a method to the `DataStoreService` class which will allow user to retrieve this metadata inside the actions and sensors.

## TODO

- [x] Decide on the approach
- [x] Changelog entry
- [x] Tests
- [x] Add a method to the datastore and action service for retrieving this metadata, update affected mock class used for tests


st2api/st2api/app.py
```diff
@@ -60,7 +60,7 @@ def setup_app(config={}):
 
     spec = spec_loader.load_spec('st2common', 'openapi.yaml.j2')
     transforms = {
-        '^/api/v1/': ['/', '/v1/'],
+        '^/api/v1/': ['/', '/v1/', '/v1'],
```
Kami: There was a small bug - we didn't correctly map `/v1/ to `/api/v1/` so hitting one of those paths would incorrectly return 404 instead of the root controller content.

Kami:
I decided to move this information outside the root controller to the new `GET /v1/whoami/` API endpoint. I also updated it to include more information including the method user used to authenticate (auth token / api key).

That's how the response looks right now:

```json
{
    "username": "testu",
    "rbac": {
        "is_admin": true,
        "enabled": true,
        "roles": [
            "admin",
            "rule_create"
        ]
    },
    "authentication": {
        "method": "authentication token",
        "location": "query"
    }
}
```

lakshmi-kannan:
I personally love /v1/whoami. I also think there is a feature request for st2 whoami somewhere. As part of this, if you could also introduce that CLI binding, it will complete the feature. @enykeev, @christopherjbaker: This could be a feature in UI too under "profile" section for the logged in user. Can you guys open a tracking issue in st2web?

Kami:
@lakshmi-kannan We actually already have `st2 whoami` CLI command - @Mierdin added it a while ago.

Right now that CLI command behaves a bit differently though - it returns username of the user which is currently logged in the CLI (aka token for which user is currently cached locally and will be used once the CLI makes a request to the API).

So in short - those are two different things, albeit related.

We should think if and how we can consolidate those two things though. I'm not totally sure if we can actually consolidate them since they serve two different purposes. Perhaps another `st2 api whoami` command, but this might be too confusing.

The idea behind the API endpoint was to make it useful when directly interacting with the API (not through the CLI - CLI is a higher level wrapper around the API which introduces token and credentials caching mechanism) and inside the action and sensor service (aka "I want to see with which user the sensor wrapper / action wrapper is currently authenticated").

Mierdin:
Wouldn't it make a little more sense to add a method to the `/auth` API for this? Seems strange to be on its own, amongst other functional APIs when we have a dedicated set of APIs for authentication-related purposes.

Also, I'm pretty -1 on having an API endpoint for `whoami` and a client command `whoami` that serve two different purposes. Way confusing. If we must have both, then the client command should be modified to show both pieces of information clearly. Continue to show the locally cached user, and then display all of the metadata about that user retrieved via this API. 

Finally, I don't think the `whoami` nomenclature is appropriate on the API side. From what I can see, this is a way to retrieve metadata about the currently authenticated user, so why not just `GET /auth/user`? The nature of REST means that identifying who is logged in is still a function of the client, not the server - so the simple `whoami` piece is still client-side - the metadata retrieval is only possible on the server.

Kami:
@Mierdin I'm fine with changing the path to `GET /auth/user` or similar.

It's also worth pointing out that this operation operates on the StackStorm API and not StackStorm Auth API. That's one of the reasons why I'm not sure if `/auth/user` is an appropriate path because all the `/auth` requests are routed to the auth api and not StackStorm API (two different services serving different purposes and the actual operation operates on StackStorm API).

Having said that - we still need to decide what to do with `st2 whoami` command. I'm not sure if we also want to add this new information there since right now it operates on the CLI level and not API level. In short those are two different layers of abstraction and it might be confusing to have all of that information in one place.

For now, I would prefer to rename it to `/auth/user` or similar and leave `whoami` CLI command as is. If needed, we can maybe, in the future, add `st2 auth user` or similar CLI command.

Kami:
We talked on Slack and decided to rename it to `/api/user` and in the future, we can come up with the plan on how to also display this information in the output of `st2 whoami` command.

The reason for that is that this operation operates on the StackStorm API level and not auth API level and this way we keep the existing abstraction and separation of concern.

Kami:
As discussed and agreed upon, I made the following changes:

* Renamed API endpoint from `/api/v1/whoami` to `/api/v1/user`
* Added new `get_user_info()` method to datastore service and wrappers to action and sensor service (this way it's easy for user to retrieve information about user account which is used for outgoing requests in the datastore service)
* Updated `st2 whoami` CLI command to utilize this new API endpoint instead.

Still need to update affected whoami tests though.

st2client/st2client/commands/auth.py
```diff
@@ -176,39 +177,37 @@ def __init__(self, resource, *args, **kwargs):
...
-            if self.app.client.debug:
-                raise
-        except Exception:
-            print("Unable to retrieve currently logged-in user")
             if self.app.client.debug:
                 raise
 
+        print('Currently logged in as "%s".' % (user_info['username']))
+        print('')
+        print('Authentication method: %s' % (user_info['authentication']['method']))
```
st2stanley: In subsequent pass, we could probably include expiry information for token. I am split because of the security concerns.

st2api/st2api/controllers/v1/user.py
```diff
@@ -0,0 +1,58 @@
...
+            role_dbs = get_roles_for_user(user_db=requester_user)
+            roles = [role_db.name for role_db in role_dbs]
+        else:
+            roles = []
+
+        data = {
+            'username': requester_user.name,
+            'authentication': {
+                'method': auth_info['method'],
+                'location': auth_info['location']
```
lakshmi-kannan: Not sure why this is useful?

st2client/st2client/client.py
```diff
@@ -162,6 +164,23 @@ def __init__(self, base_url=None, auth_url=None, api_url=None, stream_url=None,
...
+        Retrieve information about the current user which is authenticated against StackStorm API.
+
+        :rtype: ``dict``
+        """
+        url = '/user'
+        client = httpclient.HTTPClient(root=self.endpoints['api'], cacert=self.cacert,
+                                       debug=self.debug)
+        response = client.get(url=url, **kwargs)
+
+        if response.status_code != 200:
```
lakshmi-kannan: Use the constant as opposed to 200.

st2client/st2client/commands/auth.py
Kami: Yeah, we can format this info better as well. If you have any idea on the format, let me know.

st2client/st2client/client.py
Kami: Will do.

st2api/st2api/controllers/v1/user.py
Kami: In theory if you are connecting to the API yourself you know how the token is provided, but if you use CLI or similar this info might help you with debugging (is the token provider via header / query param / cookie)

st2client/st2client/client.py
Kami: Btw, I just follow the existing style (other code doesn't use constants), but fine.

st2client/st2client/commands/auth.py
Kami: There we go:

```bash
(virtualenv)vagrant@vagrant-ubuntu-trusty-64:/data/stanley$ python ./st2client/st2client/shell.py  whoami
Currently logged in as "testu".

Authentication method: authentication token
Authentication token expire time: 2017-11-14T10:26:35.724000Z

RBAC:
 - Enabled: False
 - Roles: 
```

We can also format the date nice aka more user-friendly.

PR #3833, opened 2017-11-07
Add a search operator to rules engine

ahubl-mz:
# Overview #

This PR adds a `search` operator to StackStorm's rules engine that searches through arrays and compares each element of the array to the child patterns of this operator. It takes a non-optional `condition` parameter that controls how individual items affect the overall match.

# Examples #

Many things are best explained via examples, so here's a few example usages for you:

#### At least one field should be a "Status" field with a new "Approved" status ####

```yaml
---
criteria:
  trigger.fields:
    type: search
    # Controls whether this criteria has to match any or all items of the list
    condition: any  # or all
    pattern:
      # Here our context is each item of the list
      # All of these patterns have to patch the item for the item to match
      # These are simply other operators applied to each item in the list
      item.field_name:
        type: "equals"
        pattern: "Status"
      item.to_value:
        type: "equals"
        pattern: "Approved"
```

This example criteria would search through a trigger's `fields` array for a dictionary:

```json
{
  "field_name": "Status",
  "to_value": "Approved"
}
```

Because the `search` condition was specified as `any`, the following payload would match:

```json
{
  "fields": [
    {
      "field_type": "Built-in",
      "field_name": "Status",
      "to_value": "Approved"
    }, {
      "field_type": "Custom",
      "field_name": "Signed off by",
      "to_value": "Stanley"
    }
  ]
}
```

If the `search` condition was specified as `all`, the above payload would not match.

#### All fields should be custom fields ####

```yaml
---
criteria:
  trigger.fields:
    type: search
    condition: all
    pattern:
      item.field_type:
        type: "equals"
        pattern: "Custom"
```

This example criteria would search through a trigger's `fields` array for a dictionary:

```json
{
  "field_type:" "Custom"
}
```

Because the `search` condition was specified as `all`, the following payload would match:

```json
{
  "fields": [
    {
      "field_type": "Custom",
      "field_name": "Status",
      "to_value": "Approved"
    }, {
      "field_type": "Custom",
      "field_name": "Signed off by",
      "to_value": "Stanley"
    }
  ]
}
```

But the original payload would not match:

```json
{
  "fields": [
    {
      "field_type": "Custom",
      "field_name": "Status",
      "to_value": "Approved"
    }, {
      "field_type": "Built-in",  // <-- must be Custom for search operator to match
      "field_name": "Signed off by",
      "to_value": "Stanley"
    }
  ]
}
```

# Extensibility #

The `search` condition is extensible if users need more conditions in the future:

* `count_gt` - Count of matched fields should be greater than...
* `count_gte` - Count of matched fields should be greater than or equal to...
* `count_lt` - Count of matched fields should be less than...
* `count_lte` - Count of matched fields should be less than or equal to...

However, we only need the `any` condition at this point, so we only implemented (and documented) the `any` and `all` conditions.

# Additional Considerations #

I needed to use the `PayloadLookup` class from `st2common.operators`, so I broke that out into its own module in `st2common.utils.payload` and imported it in `st2reactor.rules.filter` to keep reverse compatibility.

Note that this makes the rules engine a little bit more complex (it actually turns the rules engine into a recursive descent parser), and naive users could certainly create a rule that takes awhile to process. However, this negligible overhead (an additional `if` statement) to the rules engine unless it is used, and we think this potential complexity is worth "powering up" the rules engine.

# Tests #

Unit tests for the search operator itself, and for the additional logic in the `RuleFilter.filter` function are included - 100% coverage of all new code. I will need some guidance if you would like additional testing.

5dcd25c6c43e247f9a7bbdc1143091e705a73117 I have added unit tests for the `PayloadLookup` class as well since I broke it out into its own module.

The setup for the tests for the `search` operator are quite long and are separated into their own test case. I considered breaking them out into their own module, but I figured it was best to keep all of the operator tests in the same module.

#### Coverage ####

I tried to boost test coverage (including branch conditions) for any modules I modified. Here's an excerpt from my test coverage report:

```
Name                                              Stmts   Miss Branch BrPart  Cover
-----------------------------------------------------------------------------------
...
st2common/operators.py                              150      0     46      0   100%
...
st2common/util/payload.py                            15      0      4      0   100%
...
st2reactor/rules/filter.py                           98     18     30      3    80%
```

As you can see, I hit 100% for two of the three modules, and according to the detailed report, it seems that all of my additional code was tested.

Please note that this is only coverage from three different tests:

* `st2common/tests/unit/test_operators.py`
* `st2common/tests/unit/test_util_payload.py`
* `st2reactor/tests/unit/test_filter.py`

#### Additionally ####

* 37045ab12defe981f020c603637c9380eecb01f0 I sprinkled in a few small corrections to test failure messages
* 10035dd0af397895a9d13774e7864dfe24e6ab59 I added tests for the `get_operator` and `get_allowed_operators` functions from `st2common.operators`
* cb39323c8e1d50a1759c9f8ae16e7749e1dcd005 I added tests for all applicable operators where `criteria_pattern` is `None`

# Documentation #

The search operator function itself has an extensive docstring, complete with example usage.

I have not yet had time to write any patches for the StackStorm documentation, but I absolutely will do so once this PR is merged.

# Motivation #

We have a single webhook from an internal system that is more difficult to scale (we can have it dispatch multiple webhooks but it has to do a **lot** of filtering to decide which webhooks to dispatch). This new operator will allow us to search through a changelog in the webhook data for changed fields and do the filtering on the StackStorm side of things.

We could implement this via a workflow, but then we would need the workflow to dispatch an additional webhook back to our StackStorm instance, and there's a possibility that a poorly coded workflow could create an endless dispatch loop, or create a great deal of load on our StackStorm system. Implementing searching an array with arbitrary patterns is the cleanest approach to this problem.

This allows us to use the rules engine to filter the webhook data, which is a lot more lightweight than running an entire workflow. By reducing the number of webhooks from our internal system down to one and implementing this operator in StackStorm's rules engine, we are able to minimize the load on either system.

Finally, while StackStorm already has operators that work on arrays (the `contains` and `ncontains` operators), they are not flexible or extensible enough for our use case. StackStorm neatly sidesteps needing to iterate through dictionaries with the JSON-path notation, but without this operator, it simply does not support arrays to the level that we require.

# Final Notes #

I am hoping that this PR will be merged before the next point release (either 2.5.1 or 2.6), hopefully in the next few weeks. I am more than happy to work with you towards that goal. Please let me know if there's anything you need from me.

### DIFF
--- a/st2common/tests/unit/test_operators.py
+++ b/st2common/tests/unit/test_operators.py
@@ -19,6 +19,528 @@ from st2common import operators
 from st2common.util import date as date_utils
 
 
+def dict_strict_equal(d1, d2):
+    """
+    Compare two dictionaries, ensuring that all keys exist in both and that the
+    values for each key are the same.
+    """
+    if len(d1.keys()) != len(d2.keys()):
+        return False
+    for k1, v1 in d1.items():
+        if k1 not in d2 or d2[k1] != v1:
+            return False
+    return True
+
+
+def list_of_dicts_strict_equal(lofd1, lofd2):
+    """
+    Ensure that two unordered lists contain the same dicts
+    """
+    t2 = list(lofd2)  # mutable copy
+    if len(lofd1) != len(lofd2):
+        return False
+    for i1, el1 in enumerate(lofd1):
+        for i2, el2 in enumerate(t2):
+            if dict_strict_equal(el1, el2):
+                del t2[i2]
+                break
+        else:
+            return False
+    return not t2
+
+
+class DictsStrictEqualTest(unittest2.TestCase):
+    """
+    Tests dict_strict_equal
+
+    We should test our comparison functions, even if they're only used in our
+    other tests.
+    """
+    def test_empty_dicts(self):
+        self.assertTrue(dict_strict_equal({}, {}))
+
+    def test_simple_dicts(self):
+        self.assertTrue(dict_strict_equal(
+            {'a': 1},
+            {'a': 1},
+        ))
+
+        self.assertFalse(dict_strict_equal(
+            {'a': 1},
+            {'b': 1},
+        ))
+
+    def test_dict_with_different_lenths(self):
+        self.assertFalse(dict_strict_equal(
+            {'a': 2},
+            {'a': 1, 'b': 2},
+        ))
+
+    def test_complex_dicts(self):
+        self.assertFalse(dict_strict_equal(
+            {'a': 1, 'b': 1},
+            {'a': 1, 'b': 2},
+        ))
+
+        self.assertTrue(dict_strict_equal(
+            {'a': 1, 'b': 2},
+            {'a': 1, 'b': 2},
+        ))
+
+
+class ListOfDictsStrictEqualTest(unittest2.TestCase):
+    """
+    Tests list_of_dicts_strict_equal
+
+    We should test our comparison functions, even if they're only used in our
+    other tests.
+    """
+    def test_empty_lists(self):
+        self.assertTrue(list_of_dicts_strict_equal([], []))
+
+    def test_empty_dicts(self):
+        self.assertTrue(list_of_dicts_strict_equal([{}], [{}]))
+
+    def test_multiple_empty_dicts(self):
+        self.assertTrue(list_of_dicts_strict_equal([{}, {}], [{}, {}]))
+
+    def test_simple_dicts(self):
+        self.assertTrue(list_of_dicts_strict_equal([
+            {'a': 1},
+        ], [
+            {'a': 1},
+        ]))
+
+        self.assertFalse(list_of_dicts_strict_equal([
+            {'a': 1},
+        ], [
+            {'a': 2},
+        ]))
+
+    def test_lists_of_different_lengths(self):
+        self.assertFalse(list_of_dicts_strict_equal([
+            {'a': 1},
+        ], [
+            {'a': 1},
+            {'b': 2},
+        ]))
+
+        self.assertFalse(list_of_dicts_strict_equal([
+            {'a': 1},
+            {'b': 2},
+        ], [
+            {'a': 1},
+        ]))
+
+    def test_less_simple_dicts(self):
+        self.assertTrue(list_of_dicts_strict_equal([
+            {'a': 1},
+            {'b': 2},
+        ], [
+            {'a': 1},
+            {'b': 2},
+        ]))
+
+        self.assertTrue(list_of_dicts_strict_equal([
+            {'a': 1},
+            {'a': 1},
+        ], [
+            {'a': 1},
+            {'a': 1},
+        ]))
+
+        self.assertFalse(list_of_dicts_strict_equal([
+            {'a': 1},
+            {'a': 1},
+        ], [
+            {'a': 1},
+            {'b': 2},
+        ]))
+
+        self.assertFalse(list_of_dicts_strict_equal([
+            {'a': 1},
+            {'b': 2},
+        ], [
+            {'a': 1},
+            {'a': 1},
+        ]))
+
+
+class SearchOperatorTest(unittest2.TestCase):
+    # The search command extends the rules engine into being a recursive descent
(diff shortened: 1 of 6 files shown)


Kami:
Thanks for this contribution - we will have a look as soon as we get a chance.

Also, nice work on the PR - it should serve as a great example to others (good description, context, background, tests, etc.) :+1:

ahubl-mz:
Rebased off of `master`.

st2common/tests/unit/test_operators.py
```diff
@@ -19,16 +19,540 @@
 from st2common.util import date as date_utils
 
 
+def dict_strict_equal(d1, d2):
+    """
+    Compare two dictionaries, ensuring that all keys exist in both and that the
+    values for each key are the same.
+    """
+    if len(d1.keys()) != len(d2.keys()):
```
Kami: I believe `dict1 == dict2` should work in the same manner - aka deep strict comparison should be performed by default.

Would be good to confirm that's indeed the case because I would imagine if that's the case, Python implementations would optimize that use-case and make it fast and efficient :)

st2common/tests/unit/test_operators.py
```diff
@@ -19,16 +19,540 @@
...
+def list_of_dicts_strict_equal(lofd1, lofd2):
+    """
+    Ensure that two unordered lists contain the same dicts
+    """
+    t2 = list(lofd2)  # mutable copy
+    if len(lofd1) != len(lofd2):
+        return False
+    for i1, el1 in enumerate(lofd1):
+        for i2, el2 in enumerate(t2):
+            if dict_strict_equal(el1, el2):
```
Kami: I'm fine with this approach, but usually you need to be careful when dealing with mutable copies and iterating over them at the same time because the result might not always be what you expect.

So not a blocker, but I would prefer an approach where the list doesn't need to be mutated in place during iteration.

Kami: I just noticed those functions are only used in the tests so it's even less big of a deal.

Kami:
I had a look and overall it looks good to me - I don't see any blockers.

Only thing which is missing is some docs in st2docs (https://github.com/StackStorm/st2docs). Can you please add some? And please also document the potential "caveats" and performance issues in case user is not careful on what kind of payload this operator is used.

I will go ahead and merge it into master. I will tag it as v2.6.0 right now, but we will be discussing v2.5.1 / v2.6.0 release this week and if we decide to include it in v2.5.1, I will change the milestone.

Kami:
And thanks again for a high quality contribution - it was a joy to review it :)

st2common/tests/unit/test_operators.py
ahubl-mz: Yep, I don't like it either, so I made sure the semantics all worked out. I even wrote test cases for the comparison functions I use in the tests just to make sure they worked the way I expected them to work. 😆

ahubl-mz:
Yes, I will write some documentation later today for it and submit the PR for to st2docs either today or tomorrow. I'd like to get a coworker to review my docs patch before I push it out.

Thanks for the compliment, it was a joy to write! 😄

st2common/tests/unit/test_operators.py
ahubl-mz: You are correct - Python's `==` for dicts checks the key and values for equality, not just the `id()` or the `hash()` as I initially thought.

I'll fix this in another PR (should simply be able to remove the entire `dict_strict_equal` function, its test case, and convert `list_of_dicts_strict_equal` to use `==` instead).

ahubl-mz: PR #3839 fixes this.

ahubl-mz:
I've written the documentation in an internal fork of st2docs. I'm currently waiting on a step in our internal upstreaming process to push to that upstream repository.

ahubl-mz:
Documentation for the `search` operator is in [PR #666](https://github.com/StackStorm/st2docs/pull/666) ( 😈 ). Questions, comments, and feedback are all welcome.

PR #3835, opened 2017-11-08
[v2.5] Correctly scope token generated by DatastoreService to the user which ran the action

Kami:
This pull request fixes a bug reported in #3823.

The code incorrectly scoped the auth token which is available inside action service to Python runner actions to `action_service` user instead of to the user which ran the action.

This regression has been there for a quite a while since datastore service was refactored so it can also be used by sensors.

This change is only meant for v2.5.1 in case we decide to release it. It has been supersede by https://github.com/StackStorm/st2/pull/3803 for master / 2.6.0.

#3803 has a proper fix which is to use a temporary auth token which is generated by the action runner container for the duration of the action life-time and is correctly scoped to the user who ran the action (or stanley for actions kicked by the rules, etc.). That is in fact what we already did in the past at some point, but it got overwritten with some other changes in the past.

The temporary auth token has been designed especially for that purpose and has many benefits. One of the benefits is that this token is only active / valid for the duration of the action life-time. The token which is / was generated inside DatastoreService is valid until it expires (by default TTL is 24 hours) which is obviously not great.

## TODO

- [ ] Changelog entry

### DIFF
--- a/contrib/runners/python_runner/tests/unit/test_pythonrunner.py
+++ b/contrib/runners/python_runner/tests/unit/test_pythonrunner.py
@@ -568,6 +568,24 @@ class PythonRunnerTestCase(RunnerTestCase, CleanDbTestCase):
                        'No module named invalid')
         self.assertRaisesRegexp(Exception, expected_msg, wrapper._get_action_instance)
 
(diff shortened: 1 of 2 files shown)


Kami:
Merged directly into #3840 - closing.

PR #3839, opened 2017-11-08
Use Python's built-in '==' instead of rolling our own

ahubl-mz:
In PR #3833 I created a `dict_strict_equal` function to compare two dictionaries by key and value.

However, as @Kami [pointed out](https://github.com/StackStorm/st2/pull/3833#discussion_r149641102), I could just use Python's built-in `==` operator (I thought that just compared the results of `id()` for the two dictionaries. Python's built-in equality operator for dicts already does key and value comparisons properly.

This is the PR [that I said I would create](https://github.com/StackStorm/st2/pull/3833#discussion_r149822156) - it removes my `dict_strict_equal` function, the tests I created for it, and refactors `list_of_dicts_strict_equal` to use the built-in `==` instead.

I'm happy to answer any questions or respond to any comments or critiques.

### DIFF
--- a/st2common/tests/unit/test_operators.py
+++ b/st2common/tests/unit/test_operators.py
@@ -19,19 +19,6 @@ from st2common import operators
 from st2common.util import date as date_utils
 
 
-def dict_strict_equal(d1, d2):
-    """
-    Compare two dictionaries, ensuring that all keys exist in both and that the
-    values for each key are the same.
-    """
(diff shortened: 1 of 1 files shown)


Kami:
Thanks :+1: 

(only thing which I like more then adding code is removing code :)))

ahubl-mz:
Me too (especially when it's my derpy code to begin with). 😆

Issue #3815, opened 2017-10-27
st2 key load doesn't support secret values


nmaludy:
### Problem

Currently `st2 key load xxx.json` doesn't support loading `secret`/encrypted values. 

https://github.com/StackStorm/st2/blob/master/st2client/st2client/commands/keyvalue.py#L288-L300

Looks like the model in `st2client` needs to be updated: https://github.com/StackStorm/st2/blob/master/st2client/st2client/models/keyvalue.py

emptywee:
I second that.

Mierdin:
@nmaludy Is this ready to be closed?

PR #3853, opened 2017-11-11
Multiple improvements to 'st2 key load'

nmaludy:
This PR is a fix for #3815 along with a few other simple improvements detailed below:

- `st2 key load` now loads all properties of the KeyValuePair model, so things like `secret` can be specified (fixing #3815)
- `st2 key load` can now load keys from either JSON or YAML files.
- `st2 key load` can now support files containing a single KeyValuePair or an array of KeyValuePairs.


st2client/st2client/commands/keyvalue.py
```diff
@@ -269,37 +269,61 @@ def __init__(self, resource, *args, **kwargs):
...
+            # parse optional KeyValuePair properties
+            scope = item.get('scope', DEFAULT_SCOPE)
+            user = item.get('user', None)
+            secret = item.get('secret', False)
+            ttl = item.get('ttl', None)
+
+            # if the value is not a string, convert it to JSON
+            # all keys in the datastore must strings
+            if not isinstance(value, basestring):
+                value = json.dumps(value)
```
Kami: Perhaps we should do throw instead of doing this "magical" serialization? I think that would be better.

Kami:
Nice work and thanks for this contribution.

So far I only have one small comment, besides that, LGTM :+1:

st2client/tests/unit/test_keyvalue.py
```diff
@@ -0,0 +1,284 @@
...
+    'value': 'super cool value'
+}
+
+KEYVALUE_MISSING_VALUE = {
+    'id': 'kv_name',
+    'name': 'kv_name.'
+}
+
+
+class TestKeyValueBase(base.BaseCLITestCase):
```
Kami: Perhaps you could also utilize `CleanFilesTestCase` class and `to_delete_files` attribute of that class.

You can grep code for some of the examples :)

st2client/st2client/commands/keyvalue.py
nmaludy: So, i did this on purpose so we can potentially have the following data in the file:

``` yaml
- name: foo.list
  value:
    - a
    - b
    - c
- name: foo.dict
  value:
    a: 1
    b: 2
    c: 3
```

Without the conversion then the user would need to serialize their data into JSON in the file (ugly)

``` yaml
- name: foo.list
  value: '["a", "b", "c"]'
- name: foo.dict
  value: '{"a": 1, "b": 2, "c": 3}'
```

Thoughts?

Kami: For now I would prefer to throw because we don't have any "explicit" contract on serialization / de-serialization format and it could just cause confusion.

Aka we would just serialize it on the way in and not the way out so I would prefer to keep it simple and explicit.

But maybe if we some how explicitly document that and make it very obvious, I would be fine with that change.

Kami: So maybe would be the best to change this to throw now or perhaps add a log statement on what is going to happen when such value is encountered.

So this way this PR can be finalized and merged :)

nmaludy:
@Kami i've made the requested updates. I've adding a `-c/--convert` option to explicitly convert from objects to JSON strings. If this is not done and the user passes in an object, then an exception will be raised.

Thoughts?

Kami:
@nmaludy Thanks, LGTM :+1:

Issue #3697, opened 2017-08-23
Listing actions on CLI in YAML or JSON takes a long time


nmaludy:
# Problem
Came across a performance problem today when listing actions in the AWS pack using YAML and JSON formatting options. Below are command outputs (truncated) that show the timings for the various formatting options and their differences.

# Standard Formatting

```
$ time st2 action list -p aws | tail -n 1
+-----------------------------------------+------+-----------------------------------------+

real    0m11.901s
user    0m1.186s
sys     0m0.079s
```

# YAML Formatting

```
$ time st2 action list -p aws -y | tail -n 1


real    1m46.834s
user    1m34.807s
sys     0m1.106s
```

# JSON Formatting

```
 time st2 action list -p aws -j | tail -n1
]

real    1m46.032s
user    1m34.095s
sys     0m1.140s
```

# Observations

When running both the YAML and JSON formatting options the `st2` command was using 100% CPU in `top`. Memory was stable at around 68M throughout the processing.

armab:
Just a note, similar to https://github.com/StackStorm/st2/issues/3692 and https://github.com/StackStorm/st2/issues/3686 issues, I think that StackStorm _should_ be able to handle thousands of actions in pack(s) by design without a problem.

It's not the AWS pack guilty, but the root cause has to be fixed: we just need to identify the bottlenecks and optimize the entire StackStorm performance (which wasn't really a big focus before).

nmaludy:
Perfect. Yeah, i think this is another optimization opportunity on the `st2client` side. The `gunicorn` process spiked for about 10s then the rest of the time was spent in the `st2` process (`st2client`).

Kami:
Thanks for reporting this.

First we need to clarify, as you correctly determined it, that slowdown is due to parsing and mangling of JSON / YAML inside the CLI / client. API returns response in a reasonable time frame:

```bash
time curl "http://127.0.0.1:9101/v1/actions?pack=aws" > /dev/null

real	0m9.704s
user	0m0.009s
sys	0m0.005s
```

Now the question is how to solve that. For one, I don't think you ever want to non-programatically (e.g. when using CLI) see more than some reasonable number of items by default (e.g. 50 / 100 by once, similar limit as we use for other CLI commands).

I would propose to implement default limit for action list (100) and allowing user to filter / drill down and paginate through the pages, if needed.

So in short, I think that the best solution for CLI and WebUI is to implement pagination for actions API endpoint.

[...]

PR #3854, opened 2017-11-12
Improving JSON/YAML formatting performance on the CLI

nmaludy:
## Background 

Fix for #3697 . JSON/YAML formatting on the CLI were incredibly slow, especially compared to standard table formatting (seemed odd to me).

## Investigation

See timing results in #3697 . During code inspection and testing it was clear that the `jsonpath_rw` call was the culprit. 

##  Fix

This first attempt, i've replaced the `jsonpath_rw` calls with a very simple loop that iterates over a nested dictionary, extracting data based on the keys string as it goes. This loop is not anywhere close to as robust or full featured as `jsonpath_rw` and only supports simple key looks like `x.y.z`. However, these were the only types of lookups present in the unit tests and the only ones i could find when reading the code.

## Results

**Before**
```
$ time st2 action list -p aws -y | tail -n 1


real    1m46.834s
user    1m34.807s
sys     0m1.106s
```

**After**
```
$ time st2 action list -p aws -y | tail -n 1

real    0m10.953s
user    0m0.455s
sys     0m0.077s
```

## Note

Feedback is much appreciated here. Is everyone OK with removing the additional query functionality in order to improve performance? Should i look for another solution?


nmaludy:
Another piece of info:

I've also created an alternate implementation that uses `jmespath` instead of `jsonpath_rw`. This alternative implementation allows expressive queries, and performs almost as well as the hand-rolled-loop while still performing MUCH better than `jsonpath_rw`.

Here's the code in case we want to go that direction (simply replaces the `get_value()` function):

```python
import jmespath


def get_value(doc, key):
    if not key or not isinstance(doc, dict):
        raise ValueError()

    jmespath.search(key, doc)
```

nmaludy:
@Mierdin Totally understand. In my second comment i've proposed an alternate solution, which is mainly to switch to another query language [JMESPath](http://jmespath.org). Are you open to the alternate solution, or should i try to find a more optimized version of JSONPath?

Mierdin:
@nmaludy I was doing a bit of telecommuting so I didn't see your last comment before my review. Sorry for the IRL race condition. 😆 

The [library in use today](https://github.com/kennknowles/python-jsonpath-rw/tree/master/jsonpath_rw) doesn't seem to be that large or complicated - I wonder how difficult it would be to identify the performance issue there - might be a simple fix, and a better alternative to switching libraries at the st2 layer. Maybe you could post your numbers on the repo - the last commit was almost a year ago but hopefully they're still getting emails for the project.

st2client/st2client/utils/jsutil.py
```diff
@@ -13,19 +13,49 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from jsonpath_rw import parse
-
 
 def get_value(doc, key):
```
Kami: Can you please add some tests for this function? Or we have some already?

nmaludy: @Kami they already exist here : https://github.com/StackStorm/st2/blob/master/st2client/tests/unit/test_util_json.py

Kami: As per https://github.com/StackStorm/st2/pull/3854#pullrequestreview-75952944 it looks like that a good start should be, in any case, to add tests for all the various edge case which it looks like we don't have right now (dictionary key contains a dot, etc.) - it looks like we have just a couple of very basic / simple test cases.

This way we have a good starting ground and no matter the solution we go with (aka the implementation detail which uses jsonpath-rw right now) there will be a smaller chance for regression, at least for the scenarios we have tests for.

So in short, I would start with test cases for various common and edge cases for `get_value` and work there.

If we can be reasonably sure that there are no regressions and that all the tests still pass with your proposed implementation, I would be fine with merging that.

/cc @Mierdin what do you think? I think jsonpath-rw is very flexible, but if we have good test coverage for all the common and edge cases we want to support, I don't care about the actual implementation detail.

Perhaps one option would also be to use this simple / naive approach when only "foo.bar.baz" notation is used and for any other more advanced queries, simple fall back to jsonpath-rw. I would be fine with that as a good and pragmatic compromise for a common case.

Kami: And another future improvement / performance optimization would be to skip loading, parsing and dumping round-trip all together when user uses --json argument - https://github.com/StackStorm/st2/blob/master/st2client/st2client/formatters/doc.py#L34.

API already returns JSON and if user simply requests JSON output without any other flags such as `--k`, there is no need to go through the - parse, process, dump round-trip, it just provides no value and adds a lot of overhead.

nmaludy: I believe the reason that is not done currently is that all of the fields are sent from the API to the Client and then the client does the filtering.

By default, for every "command" / type there are certain fields that are shown called `display_attributes`: https://github.com/StackStorm/st2/blob/master/st2client/st2client/commands/action.py#L200

Maybe a better improvement would be to perform these filters on the server side using the MongoDB API? I'm not very familiar with Mongo but my SQL statement would look something like: `SELECT foo, bar, baz FROM model_table`. This would decrease the serialization requirements and the size of the payload sent between the server and client.

Mierdin: Back to the testing comment, I would love to see these tests do a little bit more. If we do end up with the "simple way" for common cases, with a fallback to jsonpath_rw, a "complicated case" example I would recommend is when IP addresses (with dotted decimal notation) are used as keys. Things like that, that would require more than a simple key.key.key lookup

nmaludy:
@Mierdin based on my profiling i did originally in #3697 it appears the root of the issue is in jsonpath_rw.parser. 

``` shell
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000   44.260   44.260 /opt/stackstorm/st2/lib/python2.7/site-packages/st2client/shell.py:351(main)
        1    0.001    0.001   44.141   44.141 /opt/stackstorm/st2/lib/python2.7/site-packages/st2client/shell.py:275(run)
        1    0.000    0.000   44.138   44.138 /opt/stackstorm/st2/lib/python2.7/site-packages/st2client/commands/resource.py:255(run_and_print)
        1    0.000    0.000   42.186   42.186 /opt/stackstorm/st2/lib/python2.7/site-packages/st2client/commands/__init__.py:86(print_output)
        1    0.000    0.000   42.183   42.183 /opt/stackstorm/st2/lib/python2.7/site-packages/st2client/commands/__init__.py:74(format_output)
        1    0.000    0.000   42.183   42.183 /opt/stackstorm/st2/lib/python2.7/site-packages/st2client/formatters/doc.py:54(format)
        1    0.003    0.003   42.147   42.147 /opt/stackstorm/st2/lib/python2.7/site-packages/st2client/formatters/doc.py:33(format)
      702    0.006    0.000   42.142    0.060 /opt/stackstorm/st2/lib/python2.7/site-packages/st2client/utils/jsutil.py:28(get_kvps)
     2106    0.009    0.000   42.136    0.020 /opt/stackstorm/st2/lib/python2.7/site-packages/st2client/utils/jsutil.py:19(get_value)
     2106    0.005    0.000   42.088    0.020 /opt/stackstorm/st2/lib/python2.7/site-packages/jsonpath_rw/parser.py:13(parse)
     2106    0.007    0.000   42.080    0.020 /opt/stackstorm/st2/lib/python2.7/site-packages/jsonpath_rw/parser.py:30(parse)
     2106    0.165    0.000   42.068    0.020 /opt/stackstorm/st2/lib/python2.7/site-packages/jsonpath_rw/parser.py:34(parse_token_stream)
     2106    0.268    0.000   38.902    0.018 /opt/stackstorm/st2/lib/python2.7/site-packages/ply/yacc.py:3214(yacc)
     2106    0.034    0.000   32.962    0.016 /opt/stackstorm/st2/lib/python2.7/site-packages/ply/yacc.py:2100(__init__)
     2106    5.421    0.003   29.708    0.014 /opt/stackstorm/st2/lib/python2.7/site-packages/ply/yacc.py:2533(lr_parse_table)
  2072304    6.498    0.000   12.183    0.000 /opt/stackstorm/st2/lib/python2.7/site-packages/ply/yacc.py:2163(lr0_goto)
     2106    1.055    0.001   10.773    0.005 /opt/stackstorm/st2/lib/python2.7/site-packages/ply/yacc.py:2198(lr0_items)
     2106    0.023    0.000    9.769    0.005 /opt/stackstorm/st2/lib/python2.7/site-packages/ply/yacc.py:2509(add_lalr_lookaheads)
     2106    2.694    0.001    4.872    0.002 /opt/stackstorm/st2/lib/python2.7/site-packages/ply/yacc.py:2371(compute_lookback_includes)
    92664    2.172    0.000    3.589    0.000 /opt/stackstorm/st2/lib/python2.7/site-packages/ply/yacc.py:2137(lr0_closure)
     2106    0.006    0.000    2.965    0.001 /opt/stackstorm/st2/lib/python2.7/site-packages/ply/yacc.py:323(parse)
     2106    0.073    0.000    2.960    0.001 /opt/stackstorm/st2/lib/python2.7/site-packages/ply/yacc.py:1001(parseopt_notrack)
     4212    0.014    0.000    2.826    0.001 /opt/stackstorm/st2/lib/python2.7/site-packages/jsonpath_rw/parser.py:177(token)
     4212    0.005    0.000    2.812    0.001 {next}
```

Digging down further a majority of the time is spent in the `ply.yacc` module. 

Looking at `jsonpath_rw.parser` https://github.com/kennknowles/python-jsonpath-rw/blob/master/jsonpath_rw/parser.py : It looks like i'd have to rewrite that whole to replace the library their using. Based on comments in that file, it appears that the `ply.yacc` module is actually writing temporary files to disk, so i'm sure this isn't helpful for a CPU-based task to incur disk I/O (just a hunch, i'd still need to measure to verify).

Thoughts?

nmaludy:
@Mierdin for your usecase of IP Address matching ( i think that's a great one). JMESPath fits the bill:

``` python
#!/usr/bin/env python

import jmespath

obj = {'address': {'192-168-1-1': {'name': 'aaa'}}}
query = 'address."192-168-1-1".name'

print jmespath.search(query, obj)
# output: aaa
```

There are a ton of extra options, an online evaluator and well documented (some things i feel JSON path is lacking). 

Main site and online evaluator: http://jmespath.org
Tutorial: http://jmespath.org/tutorial.html

Are you open to a change in implementation?

Mierdin:
@nmaludy Sure - my thoughts mirror that of @Kami:

>  if we have good test coverage for all the common and edge cases we want to support, I don't care about the actual implementation detail.

Primarily my concerns are consistent UX (i.e. if we're moving away from JSONpath, we should make sure we're doing it everywhere, so we're not supporting two different methods) and proper documentation including upgrade notes (this may/may not be a breaking change) and standard docs letting folks know where to find those resources you linked to for constructing proper queries.

Kami:
@nmaludy How is this PR going?

It would be great if we could go with the middle of the ground approach we discussed a while back here (https://github.com/StackStorm/st2/pull/3854#discussion_r150541003) and on Slack - use simple implementation you proposed for simple scenarios where jsonpath is not actually needed and for anything else more complex fall back to the existing  jsonpath approach.

I think this would be a great win for a common use case and speed things up significantly. And it should also be relatively simple and small change (+ more tests).

nmaludy:
@Kami i kind of got stalled out on the "use hand-rolled loop for simple case and jsonpath for others" approach and hadn't come back to it.

I got stuck trying to determine if the query was "simple" or "not simple" . Thinking through a little in my head i could check if the query contains only: letters [a-zA-Z], numbers [0-9], dashes [\-], underscores [_], and periods [\.]. If this is the case it should be a "simplistic" query we can use our hand-rolled loop for. Anything else will get passed through to jsonpath.

Regex: `^([a-zA-Z\-_]+[\.])*([a-zA-Z\-_]+)$`

https://regex101.com/r/SSc4I6/1/

This will match:
``` shell
aaa
aaa.bbb.ccc
````

But will not match:
``` shell
aaa.
aaa..bbb
#(*&$(*#
```

Thoughts?

Kami:
I think we should work. We just need to have enough test cases and also verify that regex doesn't actually slow things down a lot and make it worse.

nmaludy:
OK, not sure if i'll get to this before the end of the week (busy week here), but i should have some time this weekend to work on it.

Kami:
@nmaludy No rush, just ping me when you make some progress and I will have a look :)

Kami:
Any update on this? 

Would be great if we could get this finished and merged in time for the next release :)

st2client/st2client/utils/jsutil.py
nmaludy: @Mierdin working on this today (finally... sorry it took so long)

But it looks like the current implementation doesn't work for IP address queries. The `get_kvps` function sees the `.` in the key and splits it up into nested dictionaries. I'm going to spend a little time trying to fix it, but it basically boils down to that `get_kvps` expects "simple" expressions also where's the `.` in keys denote separators for nested dictionaries.
Example:
```

DOC_IP_ADDRESS = {
    'ips': {
        "192.168.1.1": {
            "hostname": "router.domain.tld"
        },
        "192.168.1.10": {
            "hostname": "server.domain.tld"
        }
    }
}

print jsutil.get_kvps(DOC_IP_ADDRESS,
                                         ['ips."192.168.1.1"', 'ips."192.168.1.10".hostname'])

# {'ips': {'"192': {'168': {'1': {'1"': {'hostname': 'router.domain.tld'},
#                               '10"': {'hostname': 'server.domain.tld'}}}}}}
```

nmaludy:
@Kami @Mierdin I've implemented the "fast path" logic we discussed in the review comments.
I've created the notion of a "simple expression" defined as:
```

# A simple expression is defined as a series of letters [a-zA-Z], numbers [0-9],
# dashes '-', and underscores '_' separated by one period '.'.
# A simple expression must not start with a period.
# A simple expression must not end with a period.
# A simple expression must not have more than one period in succession, ie. in
# between each period must be one or more of the valid non-period characters.
#
# Examples of valid "simple expressions":
#  abc
#  abc.def.ghi
#
# Examples of non-simple expressions:
#  .aaa
#  a..b
#  abc.
#  a(*
```

If the expression being passed in matches our definition of a "simple expression" given the regex compiled at the top of the file, then the "fast path" (naive "a.b.c" assumption) is used. Otherwise we fall back to the `jsonpath_rw` implementation.

As requested i've added additional unit tests to validate we're calling the "fast path" and "complex path" appropriately, along with an "ip address" test.

One note about IP addresses: The current logic in the `get_kvps()` function, when given an ip address query like: `ips."192.168.1.1"` breaks down that query by every period and returns a dict like:

```python
DOC_IP_ADDRESS = {
    'ips': {
        "192.168.1.1": {
            "hostname": "router.domain.tld"
        },
        "192.168.1.10": {
            "hostname": "server.domain.tld"
        }
    }
}

print jsutil.get_kvps(DOC_IP_ADDRESS,
                                         ['ips."192.168.1.1"',
                                          'ips."192.168.1.10".hostname'])
# {'ips': {'"192': {'168': {'1': {'1"': {'hostname': 'router.domain.tld'},
#                                '10"': {'hostname': 'server.domain.tld'}}}}}}
```

Instead I would expect it to output something like:
```
{'ips': {'192.168.1.1': {'hostname': 'router.domain.tld'},
        '192.168.1.10': {'hostname': 'server.domain.tld'}}}
```

At this point i don't want this to hold up this PR and instead create another issue to track this. Also, the behavior hasn't changed with this PR, it has always worked like this.

Kami:
@nmaludy Thanks for making those changes.

Did you happen to have a chance to check how much performance overhead does regex add? If I understand the code correctly, this regex expression will run `N * M` times where `N` is number of results / items returned by the API (we iterate over each item) and `M` is number of "keys" specified by the user, right?

If that's the case, I believe we could reduce this to just `M` aka `1` for a very common use case where a single key is provided by caching in memory if some `key` is simple or not. This should work fine because CLI is not a long running process and there is no danger of cache growing too large or similar because new process is created for each CLI invocation.

Kami:
Overall the changes look good to me (it would still be good to add some more test cases with more real-life like examples).

Only thing I would like to check is this - https://github.com/StackStorm/st2/pull/3854#issuecomment-350663407. It would be good to profile that code when operating on multiple documents (e.g. 100) and see how much overhead does calling re.match `100 * num_keys` add and if it's worth implementing simple memory cache for "is_simple_key".

nmaludy:
@Kami I don't have the numbers in front of me, but when I tested yesterday it was equal in runtime when using the regex and when bypassing and going straight to the simple method. Post numbers later once I'm at my laptop.

Kami:
From very simple and naive testing, the overhead is quite small so we are probably fine without caching:

```python
In [37]: timeit.timeit('for x in range(0,1000): SIMPLE_EXPRESSION_REGEX_CMPL.match("foo.bar.127.0.0.1")', setup='import re; SIMPLE_EXPRESSION_REGEX = "^([a-zA-Z0-9\-_]+\.)*([a-zA-Z0-9\-_]+)$"; SIMPLE_EXPRESSION_REGEX_CMPL = re.compile(SIMPLE_EXPRESSION_REGEX)', number=100)
Out[37]: 0.07767295837402344

In [38]: timeit.timeit('for x in range(0,10000): SIMPLE_EXPRESSION_REGEX_CMPL.match("foo.bar.127.0.0.1")', setup='import re; SIMPLE_EXPRESSION_REGEX = "^([a-zA-Z0-9\-_]+\.)*([a-zA-Z0-9\-_]+)$"; SIMPLE_EXPRESSION_REGEX_CMPL = re.compile(SIMPLE_EXPRESSION_REGEX)', number=100)
Out[38]: 0.6292898654937744
```

nmaludy:
@Kami @Mierdin here is some raw timing data:

### Default table formatting

```shell
(virtualenv) [root@stackstorm st2]# time st2 action list -p aws |tail -n 1
+-----------------------------------------+------+-----------------------------------------+

real    0m11.990s
user    0m1.185s
sys     0m0.110s
```

### With Regex Evaluation

```shell
(virtualenv) [root@stackstorm st2]# time st2 action list -p aws -y | tail -n 1


real    0m11.971s
user    0m1.346s
sys     0m0.096s
```

### Bypassing Regex and Going Directly To Simple path

```shell
(virtualenv) [root@stackstorm st2]# time st2 action list -p aws -y | tail -n 1


real    0m11.887s
user    0m1.229s
sys     0m0.104s
```

### Bypassing Regex And Going Directly To jsonpath_rw

```shell
(virtualenv) [root@stackstorm st2]# time st2 action list -p aws -y | tail -n 1


real    1m48.107s
user    1m35.157s
sys     0m2.263s
```

Looks pretty good and as @Kami found the regex doesn't add much overhead.

Thoughts? Anything further you would like to see?

Kami:
Merged.

Sorry for the delay and thanks again for working on this.

PR #3857, opened 2017-11-13
Fix "liveactions" vs "executions" manager naming mismatch & add example

vincent-legoll:
This is with:
```
Python 3.4.5 (default, May 29 2017, 15:17:55) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-11)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import st2client
>>> st2client.__version__
'2.5.0'
>>> from st2client.client import Client
>>> client = Client(base_url='http://127.0.0.1')
>>> print(dir(client))
['__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', 'actions', 'api_key', 'apikeys', 'cacert', 'debug', 'endpoints', 'inquiries', 'keys', 'liveactions', 'managers', 'packs', 'policies', 'policytypes', 'ruleenforcements', 'rules', 'runners', 'sensors', 'token', 'tokens', 'trace', 'triggerinstances', 'triggertypes']
>>> client.liveactions
<st2client.models.core.LiveActionResourceManager object at 0x7fb84dbc9898>
>>> client.executions
  File "<stdin>", line 1, in <module>
AttributeError: 'Client' object has no attribute 'executions'
```


Kami:
Thanks.

This is a good change, but the correct long term fix would actually be renaming live actions to executions. Live action is just an internal implementation detail and what should really be exposed to the users is executions.

vincent-legoll:
I agree with the long term goal, but wanted to help other users of the python API in the mean time.

What about opening a new issue for the API rename (maybe aliasing executions with liveactions during a release cycle or two). Then linking this issue with it and then merging the temporary fix...

So that the one fixing the API will be remined to also fix the associated doc...

Kami:
Yeah, I agree.

Will go ahead and merge this for now - thanks.

PR #3859, opened 2017-11-14
Make action runner modules fully blown Python packages

Kami:
This pull request makes all the action runners fully blown Python packages.

Before this change, to get serverless stuff to work, action runners were converted into single module Python package. I decided to go with single module approach because it's straight forward and fully backward compatible.

It sadly turned out that single module Python packages won't work well (there are some issues with data file installation) so I needed to change it into fully blown Python packages.

Fully blown Python packages are better and more flexible in any case and somewhere we want to be long term anyway.

For backward compatibility reasons I left aliased import import `__init__.py` so old "single module import notation" still works:

```python
# First approach is old and deprecated and only there for backward compatibility reasons
from python runner import PythonRunner
from python_runner.python_runner import PythonRunner
```

I would personally also like to get rid of this aliased stuff ASAP, but this would require some code changes and mean runners from StackStorm >= 2.6.0 are not compatible with older StackStorm versions, but this shouldn't be an issue anyway, because runners are always shipped as part of full StackStorm distribution.

And in longer term, we actually want to move away from the way we currently install runners to utilizing Python package based installation.


Kami:
Long term I would also propose renaming packages to `stackstorm-runner-<runner name>`.

This way we can avoid possible name clashes and conflicts with other Python packages.

Kami:
As far as name collisions and conflicts go - I would also propose renaming `inquirer` module name from `inquirer` to `inquirer_runner`. 

All other runner module names contain `_runner` suffix and this will make collisions less likely.

PR #3860, opened 2017-11-14
[WIP] Make elapsed time based off of "running" start timestamp

Mierdin:
Executions that are delayed for any reason, such as by policy, have an unintuitive elapsed time counter, since the start_timestamp is based off of the initial scheduled time, not when the execution actually started running. So if I simultaneously run 4 identical executions, and because of policy, they run in sequence, each elapsed time will seemingly build off of the previous one.

I'm playing around with different options here, to try to make this a little better.

### DIFF
--- a/st2actions/st2actions/worker.py
+++ b/st2actions/st2actions/worker.py
@@ -30,6 +30,7 @@ from st2common.transport.consumers import MessageHandler
 from st2common.transport.consumers import ActionsQueueConsumer
 from st2common.transport import utils as transport_utils
 from st2common.util import action_db as action_utils
+from st2common.util import date as date_utils
 from st2common.util import system_info
 from st2common.transport import queues
 
@@ -139,6 +140,7 @@ class ActionExecutionDispatcher(MessageHandler):
         # Update liveaction status to "running"
         liveaction_db = action_utils.update_liveaction_status(
             status=action_constants.LIVEACTION_STATUS_RUNNING,
+            running_start_timestamp=date_utils.get_datetime_utc_now,
             runner_info=runner_info,
             liveaction_id=liveaction_db.id)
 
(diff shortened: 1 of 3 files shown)


Kami:
Good catch and idea :+1: 

We already have "log" attribute on the execution which should contain execution state transition information - https://github.com/StackStorm/st2/pull/3860/files#diff-4fbb96d5b05c6afd6c08d4a6e34a3243L69 (this was added a while ago by @enykeev, grep for "_create_execution_log_entry").

Perhaps it might make sense to utilize that. Or maybe you have some specific performance concerns? In that case, I'm also fine with a new attribute since it's more efficient than introspecting `log` attribute.

On a related note, it looks like we really have no good docs for the execution model `log` attribute, we should add some docs about it to the model.

enykeev:
Well, I'd say that both parameters (time since scheduled vs time since running) are interesting in one way or the other. I see your point, but it seems to me that this PR will introduce a breaking change for no particular reason. Instead, we should note the behavior somewhere and deprecate elapsed time in favor of execution log.

Kami:
@enykeev For me personally, elapsed time is a very useful feature during debugging, etc. and I don't think we should remove / deprecate it.

I do agree that "log" complements it (and in some cases provides more info), but it's not something we can easily cram / display in output of "st2 execution list" result.

If anything, I would be OK with displaying both times, but not sure if we have enough space for that in "st2 execution list" output.

enykeev:
@Kami ok, so what execution times means to you then? A time between entering and leaving running state or between scheduling and the end of running?

m4dcoder:
@Kami and @enykeev since you guys were following this PR.  Is this still applicable?  It's WIP and @Mierdin is already gone.

Mierdin:
I'm going to close this for now as it's a bit stale. Feel free to pick up where I left off on the branch.

PR #3865, opened 2017-11-15
Improve Inquiry response functionality in the client

Mierdin:
In the next few weeks, I'd like to improve the "Inquiries" story by adding chatops capabilities, which means we will need a new action and action alias for responding to Inquiries in the [st2](https://github.com/stackstorm/stackstorm-st2) pack. To prepare for this, a few `st2` changes are needed first.

# Summary of Changes

## Move Respond Functionality into InquiryResourceManager

The main chatops-enabling purpose of this PR is to consolidate inquiry response functionality into a method in `st2client.models.core.InquiryResourceManager`. The actual `st2` client can still use this functionality, but by moving it into the resource manager, external entities like an `st2.inquiry.respond` action can use it as well.

## Remove Unnecessary and Unwanted Printing of Response Data

Currently the client prints the full response payload that was ultimately accepted by the API. This was a bad idea back when it was originally implemented, and as seen in #3825, it just leaked potentially sensitive response data with little to no benefit. So in this PR I omitted this output from showing at all - in it's place, a simple acknowledgement that the response was accepted by the API. If users, for some reason, wish to see the response data, they can always use the `--debug` flag for the `st2` client.

As a result, the client-side secrets masking introduced in #3825 was no longer needed, and I removed that as well. The server-side work in #3825 for the ActionExecution model is still relevant, and was not touched.


st2client/st2client/models/core.py
```diff
@@ -436,7 +436,21 @@ def resume(self, execution_id, **kwargs):
...
+
+    @add_auth_token_to_kwargs_from_env
+    def respond(self, inquiry_id, inquiry_response, **kwargs):
+        """
+        Update st2.inquiry.respond action
+        Update st2client respond command to use this?
+        """
+        url = '/%s/%s' % (self.resource.get_url_path_name(), inquiry_id)
+
+        resp = self.client.put(url, inquiry_response, **kwargs)
```
humblearner: minor s/resp/response to be consistent.

Mierdin: Ack. FWIW I did this because it was a little confusing with the `inquiry_response`, but an attentive eye should catch the difference anyways. Fixed in b15be5308

PR #3867, opened 2017-11-16
Add support for sudo password to the local and remote runners

Kami:
This pull request adds support for sudo password.

As an additional change, I added new ``--tail`` flag to ``st2 run`` and ``st2 execution re-run`` commands. When this flag is provided, new execution is automatically tailed once it's scheduled (it saves typing one command - previously you first need to do run then tail).

## TODO

- [x] More tests
- [x] Remote runner testing and implementation
- [x] Correctly intercept and handle "invalid sudo password" errors and throw a user-friendly exception
  - [x] Local runner
  - [x] Remote runner
- [x] Documentation

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -63,16 +63,16 @@ Changed
   ``AUDIT``. #3845
 * Mask values in an Inquiry response displayed to the user that were marked as "secret" in the
   inquiry's response schema. #3825
-* Added the ability of ``st2 key load`` to load keys from both JSON and YAML files.
-  Files can now contain a single KeyValuePair, or an array of KeyValuePairs.
-  Added the ability of ``st2 key load`` to load non-string values (hashes, arrays,
-  integers, booleans) and convert them to JSON before going into the datastore,
-  this conversion requires the user passing in the ``-c/--convert`` flag.
-  Updated ``st2 key load`` to load all properties of a key/value pair, now
-  secret values can be loaded. (improvement) #3815
+* Add the ability of ``st2 key load`` to load keys from both JSON and YAML files. Files can now
+  contain a single KeyValuePair, or an array of KeyValuePairs. (improvement) #3815
+* Add the ability of ``st2 key load`` to load non-string values (objects, arrays, integers,
+  booleans) and convert them to JSON before going into the datastore, this conversion requires the
+  user passing in the ``-c/--convert`` flag. (improvement) #3815
+* Updat ``st2 key load`` to load all properties of a key/value pair, now secret values can be
+  loaded. (improvement) #3815
 
   Contributed by Nick Maludy (Encore Technologies).
-  
+
 Fixed
 ~~~~~
 
@@ -212,7 +212,7 @@ Fixed
 Changed
 ~~~~~~~
 
-* Minor language and style tidy up of help strings and error messages #3782 
+* Minor language and style tidy up of help strings and error messages. #3782
 
 2.4.1 - September 12, 2017
 --------------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -71,7 +71,7 @@ Changed
 * Add the ability of ``st2 key load`` to load non-string values (objects, arrays, integers,
   booleans) and convert them to JSON before going into the datastore, this conversion requires the
   user passing in the ``-c/--convert`` flag. (improvement) #3815
-* Updat ``st2 key load`` to load all properties of a key/value pair, now secret values can be
+* Update ``st2 key load`` to load all properties of a key/value pair, now secret values can be
   loaded. (improvement) #3815
 
   Contributed by Nick Maludy (Encore Technologies).
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -35,6 +35,8 @@ Added
   common code inside a ``lib`` directory inside a pack (with an ``__init__.py`` inside ``lib``
   directory to declare it a python package). You can then import the common code in sensors and
   actions. Please refer to documentation for samples and guidelines. #3490
+* Add support for password protected sudo to the local and remote runner. Password can be provided
+  via the new ``sudo_password`` runner parameter. (new feature) #3867
 
 Changed
 ~~~~~~~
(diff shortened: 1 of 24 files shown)


st2common/st2common/models/system/action.py
```diff
@@ -30,6 +30,7 @@
 from st2common.constants.action import LIBS_DIR as ACTION_LIBS_DIR
 from st2common.util.secrets import get_secret_parameters
 from st2common.util.secrets import mask_secret_parameters
+from st2common.logging.formatters import MASKED_ATTRIBUTE_VALUE
```
Mierdin: Shouldn't this be from https://github.com/StackStorm/st2/blob/master/st2common/st2common/constants/secrets.py? Seems to me that we should only have `MASKED_ATTRIBUTE_VALUE` in one place.

st2common/tests/unit/test_shell_action_system_model.py
```diff
@@ -21,12 +21,17 @@
 
 from st2common.models.system.action import ShellCommandAction
 from st2common.models.system.action import ShellScriptAction
-
+from st2common.logging.formatters import MASKED_ATTRIBUTE_VALUE
```
Mierdin: Import from `secrets` here as well?

st2common/st2common/models/system/paramiko_script_action.py
```diff
@@ -49,7 +50,11 @@ def _format_command(self):
                 else:
                     command = quote_unix('cd %s && %s' % (cwd, script_path))
 
-            command = 'sudo -E -- bash -c %s' % (command)
+            sudo_arguments = ' '.join(self._get_common_sudo_arguments())
+            command = 'sudo %s -- bash -c %s' % (sudo_arguments, command)
+
+            if self.sudo_password:
+                command = 'echo -e %s | %s' % (quote_unix('%s\n' % (self.sudo_password)), command)
```
Mierdin: What about command history for bash? Would this not show up in cleartext in `~/.bash_history` by default?

lint-configs/python/.flake8
```diff
@@ -1,4 +1,4 @@
 [flake8]
 max-line-length = 100
-ignore = E128,E402
+ignore = E128,E402,E722
```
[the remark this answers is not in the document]
Kami: We have a bunch of them splashes over the code base and it breaks with new version of flake8.

CHANGELOG.rst
```diff
@@ -35,6 +35,11 @@ Added
   common code inside a ``lib`` directory inside a pack (with an ``__init__.py`` inside ``lib``
   directory to declare it a python package). You can then import the common code in sensors and
   actions. Please refer to documentation for samples and guidelines. #3490
+* Add support for password protected sudo to the local and remote runner. Password can be provided
+  via the new ``sudo_password`` runner parameter. (new feature) #3867
+* Add new ``--tail`` flag to the ``st2 run`` / ``st2 action execute`` and ``st2 execution re-run``
```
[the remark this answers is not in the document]
Kami: Done, updated the description.

CHANGELOG.rst
```diff
@@ -61,16 +66,16 @@ Changed
...
-  integers, booleans) and convert them to JSON before going into the datastore,
-  this conversion requires the user passing in the ``-c/--convert`` flag.
-  Updated ``st2 key load`` to load all properties of a key/value pair, now
-  secret values can be loaded. (improvement) #3815
+* Add the ability of ``st2 key load`` to load keys from both JSON and YAML files. Files can now
+  contain a single KeyValuePair, or an array of KeyValuePairs. (improvement) #3815
+* Add the ability of ``st2 key load`` to load non-string values (objects, arrays, integers,
+  booleans) and convert them to JSON before going into the datastore, this conversion requires the
+  user passing in the ``-c/--convert`` flag. (improvement) #3815
+* Updat ``st2 key load`` to load all properties of a key/value pair, now secret values can be
```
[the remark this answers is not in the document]
Kami: Fixed.

st2common/st2common/models/system/action.py
Kami: Yeah, it looks like we actually have the same constant in two places - we should try to consolidate that.

st2common/st2common/models/system/action.py
```diff
@@ -82,11 +84,27 @@ def get_full_command_string(self):
...
+        Get a command string which can be used inside the log messages (if provided, sudo password
+        is masked).
+
+        :rtype: ``password``
+        """
+        command_string = self.get_full_command_string()
+
+        if self.sudo_password:
+            # Mask sudo password
+            command_string = 'echo -e \'%s\n\' | %s' % (MASKED_ATTRIBUTE_VALUE, command_string)
```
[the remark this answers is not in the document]
Kami: That's correct - it doesn't include it for the local runner actions, but it does include it for remote runner actions (e.g. see https://github.com/StackStorm/st2/pull/3867/files#diff-c98f2920e8ffd90eae51578a09c4bcb5R71 for an example).

In local runner we use subprocess.Popen so doing `echo -e "foo\n" | sudo -S ....` won't work and we need to explicitly pass it to the command subprocess via stdin of an existing subprocess.

st2common/st2common/models/system/paramiko_script_action.py
Kami: That's correctly - sadly not much we can do it besides documenting this (it affects remote runner, but not local one).

Well, we could actually try to set some environment variables and try to disable bash history, but this it not a bullet-proof solution anyway.

It's similar with other actions which pass sensitive / secret stuff as command line arguments. We already have best practice docs which says sensitive data should be retrieved inside an action to reduce surface area and prevent exposure.

I will also document that limitation when using sudo password.

st2common/tests/unit/test_shell_action_system_model.py
Kami: Yep, will update it.

Kami: It looks like it's actually an in-direct import, which is kinda nasty.

I need to check why pylint didn't detect this - we have "__all__" in `st2common.logging.formatters` which doesn't include `MASKED_ATTRIBUTE_VALUE` so it should detect that.

Kami:
@Mierdin I verified and bash history is not actually saved for remote runner commands even though login shell is used because HISTFILE is not set.

Having said that, I will still push a small change to explicitly disable bash history for that shell in case `sudo_password` parameter is provided (as an additional safe guard).

PR #3868, opened 2017-11-17
Mask response fields for 'st2.inquiry.respond' executions

Mierdin:
https://github.com/StackStorm/st2/pull/3825 added capabilities for masking secret values in inquiry responses when being retrieved as executions.

However, this only applies to inquiries themselves, meaning those executions running under the inquirer runner. https://github.com/StackStorm-Exchange/stackstorm-st2/pull/7 introduces a new action on the `python-script` runner that will also leak these secrets.

So, this PR adds logic to address this. During execution of the `st2.inquiry.respond` action, we won't have a handle on the schema the inquiry intends to address, only the response data passed in via parameters. I **could** try to retrieve this inquiry via the API as a preliminary step, but I'd rather not cross that boundary, as it could cause operational headaches for RBAC. My current approach is simply to mask all response fields to be safe. Since the logic is located in the same place as the existing masking for `inquiries` themselves, the user can always use the `?show_secrets` query param to skip this masking.

# Examples

Using the new action I'm developing in https://github.com/StackStorm-Exchange/stackstorm-st2/pull/7, the parameters contained the full, unmasked response body. I'm planning on returning nothing from this action, so `result` will be empty.

Add in the work in this PR, and all of the response fields are masked properly:

```
(virtualenv) vagrant@st2dev:~/st2$ st2 execution list
s+----------------------------+----------------------------+--------------+------------------------+-------------------------------+-------------------------------+
| id                         | action.ref                 | context.user | status                 | start_timestamp               | end_timestamp                 |
+----------------------------+----------------------------+--------------+------------------------+-------------------------------+-------------------------------+
| + 5a0e710a32ed356babc4e599 | examples.mistral-ask-basic | stanley      | succeeded (6s elapsed) | Fri, 17 Nov 2017 05:18:02 UTC | Fri, 17 Nov 2017 05:18:08 UTC |
|   5a0e710d32ed356babc4e59e | st2.inquiry.respond        | stanley      | succeeded (1s elapsed) | Fri, 17 Nov 2017 05:18:05 UTC | Fri, 17 Nov 2017 05:18:06 UTC |
+----------------------------+----------------------------+--------------+------------------------+-------------------------------+-------------------------------+
(virtualenv) vagrant@st2dev:~/st2$ st2 execution get 5a0e710d32ed356babc4e59e
id: 5a0e710d32ed356babc4e59e
parameters:
  id: 5a0e710a32ed356babc4e59c
  response:
    secondfactor: '********'
result:
  exit_code: 0
  result: null
  stderr: ''
  stdout: ''
```


Mierdin:
RFR once https://github.com/StackStorm/st2/pull/3865 is merged (this builds on that)

st2common/st2common/models/db/execution.py
```diff
@@ -111,6 +111,19 @@ def mask_secrets(self, value):
             liveaction['parameters'] = mask_secret_parameters(parameters=liveaction['parameters'],
                                                               secret_parameters=secret_parameters)
 
+            if 'inquiry.respond' in liveaction.get('action', ''):
```
Kami: Maybe we should just do direct equal check for action name to be on the safe side?

I know it's very unlikely someone will name their action `packname.inquiry.responst.foo`, but still :)

Kami:
I'm fine with this approach :+1: 

Maybe it's not a bad idea to document this behavior somewhere so it's clear what is happening and why.

Mierdin:
@kami ack that, will add a TODO to https://github.com/StackStorm-Exchange/stackstorm-st2/pull/7 to add an st2docs PR prior to merging that which explains the new action as well as this behavior

st2common/st2common/models/db/execution.py
Mierdin: Ack, when I started tinkering with this, I couldn't remember if the pack name was included at this point so I did this to get a PoC running and neglected to change it back. Done in d032480ca42a1600584dc9694546ee8aa0767b86

Mierdin:
Docs updated in https://github.com/StackStorm/st2docs/pull/672

PR #3871, opened 2017-11-20
Fix edge case for workflows stuck in running state

m4dcoder:
There is an edge case where mistral receives connection error from the st2 API when requesting action execution. When this happens, the action execution is requested twice. The first action execution is requested, the record is written into the MongoDB, but it is not published to the action runners. The second action execution is scheduled and the mistral workflow reach completion. The mistral querier for the st2resultstracker will see the first action execution in a non-completed state and assume the workflow is still running. This patch identifies this specific edge case and ignores the duplicate action execution.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -74,6 +74,9 @@ Fixed
   executions in the database. (bug fix) #3760 #3802
 * Fix 'NameError: name 'cmd' is not defined' error when using ``linux.service`` with CentOS systems.
   #3843. Contributed by @shkadov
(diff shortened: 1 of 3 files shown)


nmaludy:
:100:

Kami:
:+1: 

Will this go into v2.5.1 or v2.6.0?

m4dcoder:
@Kami I will cherry pick into the v2.5 branch when this is merged into master. And so likely 2.5.2 unless we rebuild packages for v2.5.1. It's currently stuck in circleci...

Issue #3855, opened 2017-11-12
Inconsistent newlines for execution output


Mierdin:
I thought this was documented in another issue but I can't find it, so recording here. Sorry if duplicate.

For the last few months I've noticed that some executions respect newlines, and others do not (and in turn, record the literal `\n` in its place).

For long outputs like in our CI where the st2 bootstrap script is used, this makes troubleshooting **awful**.

It doesn't even seem to be specific to certain runners. For instance, both actions below use `python-script`, yet the `stderr` for the first execution renders newline characters, and the second does not.

```
vagrant@st2vagrant:~$ st2 run packs.setup_virtualenv packs=napalm
......................
id: 5a07a4f0c4da5f054cf2fe22
status: succeeded
parameters:
  packs:
  - napalm
result:
  exit_code: 0
  result: 'Successfuly set up virtualenv for the following packs: napalm'
  stderr: 'st2.actions.python.SetupVirtualEnvironmentAction: DEBUG    Setting up virtualenv for pack "napalm" (/opt/stackstorm/packs/napalm)

    st2.actions.python.SetupVirtualEnvironmentAction: INFO     Virtualenv path "/opt/stackstorm/virtualenvs/napalm" doesn''t exist

    st2.actions.python.SetupVirtualEnvironmentAction: DEBUG    Creating virtualenv for pack "napalm" in "/opt/stackstorm/virtualenvs/napalm"

    st2.actions.python.SetupVirtualEnvironmentAction: DEBUG    Creating virtualenv in "/opt/stackstorm/virtualenvs/napalm" using Python binary "/opt/stackstorm/st2/bin/python"

    st2.actions.python.SetupVirtualEnvironmentAction: DEBUG    Running command "/opt/stackstorm/st2/bin/virtualenv -p /opt/stackstorm/st2/bin/python --always-copy /opt/stackstorm/virtualenvs/napalm" to create virtualenv.

    st2.actions.python.SetupVirtualEnvironmentAction: DEBUG    Installing base requirements

    st2.actions.python.SetupVirtualEnvironmentAction: DEBUG    Installing requirement six>=1.9.0,<2.0 with command /opt/stackstorm/virtualenvs/napalm/bin/pip install six>=1.9.0,<2.0.

    st2.actions.python.SetupVirtualEnvironmentAction: DEBUG    Installing pack specific requirements from "/opt/stackstorm/packs/napalm/requirements.txt"

[...]

Issue #3602, opened 2017-07-23
Additional newlines printed with with `st2 execution get`


LindsayHill:
(original issue https://github.com/StackStorm/st2contrib/issues/556)

Looks like we get additional newlines printed when displaying the execution results. Seems to be specific to the CLI, as the JSON output looks OK.

```sh
vagrant@bwcvagrant:~$ cat localfile.txt
#### BLAH BLAH BLAH
###  & stuff...
#
vagrant@bwcvagrant:~$ st2 run core.local cmd='cat /home/vagrant/localfile.txt'
.
id: 597429e2bb0f185e19233cc5
status: succeeded
parameters:
  cmd: cat /home/vagrant/localfile.txt
result:
  failed: false
  return_code: 0
  stderr: ''
  stdout: '#### BLAH BLAH BLAH

    ###  & stuff...

    #'
  succeeded: true
vagrant@bwcvagrant:~$ st2 execution get 597429e2bb0f185e19233cc5 -j
{
    "id": "597429e2bb0f185e19233cc5",
    "parameters": {
        "cmd": "cat /home/vagrant/localfile.txt"
    },
    "result": {
        "failed": false,
        "return_code": 0,
        "stderr": "",
        "stdout": "#### BLAH BLAH BLAH\n###  & stuff...\n#",
        "succeeded": true
    },
    "status": "succeeded"
}
vagrant@bwcvagrant:~$ st2 --debug execution get 597429e2bb0f185e19233cc5
2017-07-23 04:46:16,468  DEBUG - Using cached token from file "/home/vagrant/.st2/token-st2admin"
# -------- begin 140262460773264 request ----------
curl -X GET -H  'Connection: keep-alive' -H  'Accept-Encoding: gzip, deflate' -H  'Accept: */*' -H  'User-Agent: python-requests/2.14.2' -H  'X-Auth-Token: b3f77ab505e24b4fbac96d13656aec48' 'http://127.0.0.1:9101/v1/executions/597429e2bb0f185e19233cc5?exclude_attributes=trigger_instance'
# -------- begin 140262460773264 response ----------
{
    "status": "succeeded",
    "start_timestamp": "2017-07-23T04:45:22.624375Z",
    "log": [
        {
            "status": "requested",
            "timestamp": "2017-07-23T04:45:22.637000Z"
        },
        {
            "status": "scheduled",
            "timestamp": "2017-07-23T04:45:22.720000Z"
        },
        {
            "status": "running",
[...]

PR #3511, opened 2017-06-23
Support unicode with line feed in CLI print 

lakshmi-kannan:
Fixes: https://github.com/StackStorm/st2/issues/3510

See https://mail.python.org/pipermail/python-list/2006-January/411909.html. 

TODO

- [x] Validate other prints (rule, sensor etc etc) actually use this util. 

Test cases without any code change:

```
======================================================================
ERROR: test_unicode_with_line_feed_like_chars (tests.unit.test_util_strutil.StrUtilTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/mnt/src/storm/st2/st2client/tests/unit/test_util_strutil.py", line 38, in test_unicode_with_line_feed_like_chars
    out_str = unescape(in_str)
  File "/mnt/src/storm/st2/st2client/st2client/utils/strutil.py", line 20, in unescape
    if isinstance(s, basestring) else s)
UnicodeEncodeError: 'ascii' codec can't encode character u'\u20ac' in position 0: ordinal not in range(128)
```

After the fix: 

```
test_unicode_with_line_feed_like_chars (tests.unit.test_util_strutil.StrUtilTestCase) ... ok
```

With the fix:

```
(virtualenv)vagrant@st2dev /m/s/s/st2 ❯❯❯ st2 action list --pack=cms                            issues/3510/unicode_support_action_list ✭ ✱
+------------------------+------+------------------------------+
| ref                    | pack | description                  |
+------------------------+------+------------------------------+
| cms.get_group_info     | cms  | 根据group_id查询group信息    |
| cms.get_pool_info      | cms  | 根据pool_id查询pool信息      |
| cms.get_server_info    | cms  | 根据pool_id查询server信息    |
| cms.get_site_info      | cms  | 根据group_id查询站点信息     |
| cms.get_vd_info        | cms  | 根据group_id查询虚拟目录信息 |
| cms.group_deletion     | cms  | 调用CMS接口删除Group         |
| cms.pool_deletion      | cms  | 调用CMS接口删除Group         |
| cms.site_deletion      | cms  | 调用CMS接口删除站点          |
| cms.ticket_association | cms  | 关联单子                     |
| cms.vd_deletion        | cms  | 调用CMS接口删除虚拟目录      |
+------------------------+------+------------------------------+
(virtualenv)vagrant@st2dev /m/s/s/st2 ❯❯❯
```

PR #3872, opened 2017-11-20
Fix bugs with newlines and unicode in execution formatter

Mierdin:
The execution formatter had some bugs that caused some problems with properly representing newline characters. This PR addresses two problems:

- Extra escape characters are added to the output of `yaml.safe_dump`. This means that newline characters `\n` are escaped, and are, as a result, not honored in the final client output. This means that complicated output like stack traces are output as a single line, which makes it wholly unreadable.
- `yaml.safe_dump` also adds a second `\n` to existing instances of `\n`, resulting in `\n\n`. Since we're printing this resulting value, this results in a lot of whitespace in output

# Explanation of Changes

As documented in #3855, the client has exhibited some strange behavior showing certain executions - in particular, Python stack traces:

```
~$ st2 run testpack.newline_test_error outcome="exception"
.
id: 5a1359b132ed3504e7a69b4f
status: failed
parameters:
  outcome: exception
result:
  exit_code: 1
  result: None
  stderr: "FOOBAR1\nTraceback (most recent call last):\n  File \"/home/vagrant/st2/st2common/st2common/runners/python_action_wrapper.py\", line 276, in <module>\n    obj.run()\n  File \"/home/vagrant/st2/st2common/st2common/runners/python_action_wrapper.py\", line 169, in run\n    output = action.run(**self._parameters)\n  File \"/opt/stackstorm/packs/testpack/actions/newline_test_error.py\", line 24, in run\n    raise Exception(\"This is an exception\")\nException: This is an exception\n"
  stdout: 'FOOBAR2

    '
```

The above example shows the entire stack trace on a single line, since the newline characters are escaped behind the scenes. This makes reading this stack trace incredibly hard to read. This is actually a VERY simple example - I've run into way worse examples during support for our own CI, such as when this comes up for tabular output, making tables 100% unreadable.

Please see [this comment](https://github.com/StackStorm/st2/issues/3855#issuecomment-345835804) for a detailed explanation of why this is happening.

The fix was to pass the resulting value for each attribute through `strutils.unescape()` once more. This stripped away the extra escape slashes, and newline characters are now rendered properly on output:

```
~$ st2 run testpack.newline_test_error outcome="exception"
.
id: 5a13599b32ed3504e7a69b4c
status: failed
parameters:
  outcome: exception
result:
  exit_code: 1
  result: None
  stderr: "FOOBAR1
  File "/home/vagrant/st2/st2common/st2common/runners/python_action_wrapper.py", line 276, in <module>
    obj.run()
  File "/home/vagrant/st2/st2common/st2common/runners/python_action_wrapper.py", line 169, in run
  File "/opt/stackstorm/packs/testpack/actions/newline_test_error.py", line 24, in run
    raise Exception("This is an exception")
Exception: This is an exception
"
  stdout: 'FOOBAR2
    '
```

In addition, I added a new function for removing duplicate newlines introduced by `yaml.safe_dump`. This addresses the concerns brought up in #3602 - note that in the first output above, `stdout` contained a line of whitespace at the end, and in the second output, this whitespace is gone.

Closes https://github.com/StackStorm/st2/issues/3855
Closes https://github.com/StackStorm/st2/issues/3602

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -74,6 +74,8 @@ Fixed
   executions in the database. (bug fix) #3760 #3802
 * Fix 'NameError: name 'cmd' is not defined' error when using ``linux.service`` with CentOS systems.
   #3843. Contributed by @shkadov
+* Fix bugs with newlines in execution formatter (client)
+  (bug fix) #3872
 
 2.5.0 - October 25, 2017
 ------------------------
--- a/st2client/st2client/formatters/execution.py
+++ b/st2client/st2client/formatters/execution.py
@@ -60,6 +60,7 @@ class ExecutionResult(formatters.Formatter):
                                                      width=sys.maxint,
                                                      indent=2)[len(attr) + 2:-1]
                     value = ('\n' if isinstance(value, dict) else '') + formatted_value
+                    value = strutil.dedupe_newlines(value)
 
                 # transform the value of our attribute so things like 'status'
                 # and 'timestamp' are formatted nicely
@@ -69,4 +70,4 @@ class ExecutionResult(formatters.Formatter):
 
                 output += ('\n' if output else '') + '%s: %s' % \
                     (DisplayColors.colorize(attr, DisplayColors.BLUE), value)
-        return output
+        return strutil.unescape(output)
--- a/st2client/st2client/formatters/execution.py
+++ b/st2client/st2client/formatters/execution.py
@@ -70,4 +70,5 @@ class ExecutionResult(formatters.Formatter):
 
                 output += ('\n' if output else '') + '%s: %s' % \
                     (DisplayColors.colorize(attr, DisplayColors.BLUE), value)
-        return strutil.unescape(output)
+
+        return strutil.unescape(output).decode('unicode_escape').encode('utf-8')
--- a/st2client/st2client/utils/strutil.py
+++ b/st2client/st2client/utils/strutil.py
@@ -27,6 +27,19 @@ def unescape(s):
     return s
 
 
+def dedupe_newlines(s):
+    """yaml.safe_dump converts single newlines to double.
+
+    Since we're printing this output and not loading it, we should
+    deduplicate them.
+    """
+
+    if isinstance(s, basestring):
+        s = s.replace('\n\n', '\n')
+
+    return s
+
+
 def strip_carriage_returns(s):
     if isinstance(s, basestring):
         s = s.replace('\\r', '')
--- a/st2client/tests/fixtures/execution_get_default.txt
+++ b/st2client/tests/fixtures/execution_get_default.txt
@@ -8,17 +8,11 @@ result:
     return_code: 0
     stderr: ''
     stdout: 'PING 127.0.0.1 (127.0.0.1) 56(84) bytes of data.
-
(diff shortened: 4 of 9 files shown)


Mierdin:
I added a test to address the unescape bug. Ensured it failed on `master` and resulted in a success in this branch - just making sure this regression is prevented in the future:

```
(virtualenv) vagrant@st2dev:~/st2$ nosetests -v st2client/tests/unit/test_formatters.py:TestExecutionResultFormatter.test_execution_unescape_newline
test_execution_unescape_newline (tests.unit.test_formatters.TestExecutionResultFormatter) ... FAIL

FAIL: test_execution_unescape_newline (tests.unit.test_formatters.TestExecutionResultFormatter)
  File "/home/vagrant/st2/virtualenv/local/lib/python2.7/site-packages/mock/mock.py", line 1305, in patched
    return func(*args, **keywargs)
  File "/home/vagrant/st2/st2client/tests/unit/test_formatters.py", line 120, in test_execution_unescape_newline
    self.assertEqual(content, FIXTURES['results']['execution_unescape_newline.txt'])
AssertionError: 'id: [155 chars]BAR1\\nTraceback (most recent call last):\\n "[54 chars]ue\n' != 'id: [155 chars]BAR1\nTraceback (most recent call last):\n "\n[50 chars]ue\n'
  id: 547e19561e2e2417d3dde123
  parameters: None
  result:
    localhost:
      failed: false
      return_code: 0
-     stderr: "FOOBAR1\nTraceback (most recent call last):\n "
+     stderr: "FOOBAR1
+ Traceback (most recent call last):
+  "
      stdout: 'FOOBAR2
-
        '
      succeeded: true
```

Mierdin:
I wanted to make sure this was also fixed for our internal purposes, where we run tests from the CLI's perspective by basically interacting with the `st2` client with `core.local`. In the cases where exceptions were thrown in a workflow executed this way, the output table was horrendous:

screenshot 2017-11-20 19 03 00

With this PR, the output is much more reasonable (though there's some kind of unicode thing going on there - this seems to be only specific to this corner case of executing st2 commands via core.local):

screenshot 2017-11-20 19 05 47

Kami:
LGTM, but probably also not a bad idea to run this on our CI server for a while since it's a behavior change to make sure everything is in order.

And good change, Python tracebacks look much better now :+1:

nmaludy:
@Mierdin just an FYI those are ANSI escape sequences for colors in the terminal https://en.wikipedia.org/wiki/ANSI_escape_code

Mierdin:
I reviewed the changes in https://github.com/StackStorm/st2/pull/3511, and after some discussion I thought it was useful to re-introduce the unicode encoding statement at the end of this formatting function. So that's done in https://github.com/StackStorm/st2/pull/3872/commits/8170226b3f02ce888b6462439751aaf90370c9d0

Mierdin:
I put together a simple test to show where things are at as of now. Here's the output for a workflow that runs my test action:

```
---
  version: '2.0'
  testpack.wftest:
    type: direct
    tasks:
      my_task:
        action: core.local
        input:
          cmd: 'source /home/vagrant/st2/virtualenv/bin/activate && st2 run testpack.newline_test_error outcome="exception"'
```

output:

screenshot 2017-11-27 10 02 42

Kami:
LGTM :+1: 

Let's merge it, deploy it to CI and see how it behaves there :)

PR #3881, opened 2017-11-24
Additional contexts for rendering live params

enykeev:
The feature is required for serverless plugin, but might be useful to further extend the variety of contexts during rendering.

### DIFF
--- a/st2common/st2common/util/param.py
+++ b/st2common/st2common/util/param.py
@@ -217,7 +217,7 @@ def _cast_params_from(params, context, schemas):
 
 
 def render_live_params(runner_parameters, action_parameters, params, action_context,
-        additional_contexts=None):
+        additional_contexts={}):
     '''
     Renders list of parameters. Ensures that there's no cyclic or missing dependencies. Returns a
     dict of plain rendered parameters.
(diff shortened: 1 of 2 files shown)


st2common/st2common/util/param.py
```diff
@@ -216,7 +216,8 @@ def _cast_params_from(params, context, schemas):
     return result
 
 
-def render_live_params(runner_parameters, action_parameters, params, action_context):
+def render_live_params(runner_parameters, action_parameters, params, action_context,
+        additional_contexts=None):
```
Kami: LGTM.

Let's at least document it needs to be a dict and at what point is added into the final context (e.g. it seems like it's added first and stuff from it could potentially be overwritten from other contexts).

Kami: And also at least a basic test case :)

st2common/st2common/util/param.py
```diff
@@ -216,7 +216,8 @@ def _cast_params_from(params, context, schemas):
     return result
 
 
-def render_live_params(runner_parameters, action_parameters, params, action_context):
+def render_live_params(runner_parameters, action_parameters, params, action_context,
+        additional_contexts={}):
```
Kami: I missed that earlier - we need to default this to `None` and do `additional_contexts = additional_contexts or {}` since it's mutable value and mutable default values cause issues and usually unexpected behavior.

PR #3884, opened 2017-11-29
Match and multiple execute

ahubl-mz:
# Overview #

This is the server-side implementation of [hubot-stackstorm PR #148](https://github.com/StackStorm/hubot-stackstorm/pull/148) with a few tweaks.

This PR refactors and re-implements the `match_and_execute` API endpoint to match a command to a single alias and execute the associated action multiple times.

That prose is kind of difficult to parse and grok, so here's some [pseudocode](https://stackoverflow.com/a/12076662):

```python
match = [command.match(alias) for alias in action_aliases if not alias.multiple_match]
if match:
    return {'results': [match[0].execute()]}
else:
    matches = [command.match(alias) for alias in action_aliases if alias.multiple_match]
    if matches:
        return {'results': [match.execute() for match in matches]}
```

This implementation will only ever match a *single* alias per command. Once it has matched that single alias, it will execute the alias action multiple times with each distinct captured group.

# Example #

#### An action alias that can match multiple times in a message ####

```yaml
---
name: "jira_get_issue"
pack: "jira"
action_ref: "jira.get_issue"
description: "Get a blurb from a Jira issue."
formats:
  # This display format can only match a command once
  - display: "jira issue <issue key>"
    representation:
      - "jira i(?:ssue)? (?P<issue_key>[A-Z][A-Z0-9]+-[0-9]+)"
  # This display format can match a command multiple times
  - display: "<issue key>"
    representation:
      - (?:^|\b)(?P<issue_key>[A-Z][A-Z0-9]+-[0-9]+)(?:\b|$)
    match_multiple: true
```

This action alias allows StackStorm to match a Jira issue key multiple times in one message, so the following Slack message would execute the associated action multiple times:

> Hey, I started working on SERVICEDESK-1234, but I got blocked by INFRASTRUCTURE-4321. I'm switching over to working on SERVICEDESK-1235 instead.

Specifically, with this change, the action would be dispatched three different times:

1. Once for the issue key `SERVICEDESK-1234`
2. Once for the issue key `INFRASTRUCTURE-4321`
3. Once for the issue key `SERVICEDESK-1235`

# Extensibility #

Our current actions simply respond with a blob of the the issue description in Slack, saving our users a round trip to Jira (which could itself include a login) to read the issue. However, the exact action is not limited to this.

# Reviewing Notes #

I needed to refactor quite a bit of code to support this. I tried to break up each refactor into its own commit, but some of them were a little noisy. I recommend GitHub's "split" mode for viewing the diffs, and look at the commits in the following order:

1. [`f5946e3`](https://github.com/StackStorm/st2/pull/3884/commits/f5946e34c9c023c81e8733e06c69ed872f832989#diff-8c5ac792de74fa5c41c1601685adde17R126) - You'll review the full commit later, but this link will take you to exactly what you need to look at now. Take a look at how I implemented single and multiple matching. This will give you context for the rest of the refactors.
2. [`7d6187f `](https://github.com/StackStorm/st2/pull/3884/commits/7d6187fef141b164b4a01a8b73e71a1c77ec4c57?diff=split) - Refactor `ActionAliasFormatParser` into multiple functions that mutate the object's state, to break out the regex matching into its own function ([`get_extracted_parameter_value`](https://github.com/StackStorm/st2/pull/3884/commits/7d6187fef141b164b4a01a8b73e71a1c77ec4c57#diff-ef06cf3802f64c52e0fab1e5b7cb8310R166)), and use `re.search()` instead of `re.match()` (explained below).
3. 1a70f14 - Add a function to `ActionAliasFormatParser` to match multiple times (using `re.finditer()`).
4. ae64b03 - Use a common `_post` function in `ActionAliasExecutionController` to do the heavy lifting, and use it in both `post()` and `match_and_execute()`.
5. 2f848d8 - Instead of returning a tuple and utilizing a decoding function, simply return a dictionary and use it directly (this makes the next commits easier to read).
6. 9fdbfbd - Make `list_format_strings_from_aliases` more consistent when an alias format doesn't have a representation.
7. 89f9f74 - There is a lot of noise in this commit, but the main change is the `query` function. I tweaked it to use keyword-only arguments for `offset`, `limit`, `order_by`, and `exclude_fields` (there were no uses of this function with positional arguments), and added support for using `Q` objects from `mongoengine`. I saw that `**filters` were cleaned up a bit, so I added a `_process_arg_filters` companion function to recursively process each filter in the tree of Q objects. Enabling Q objects in ORM queries allows me to query alias formats that don't have `match_multiple` keys OR where the value of the `match_multiple` key is `False`, as I do in the next commit.
8. f5946e3 - Now that you have the context for this seminal commit, check out the whole diff.
9. ac826be - A bonus test for the `normalise_alias_format_string`.
10. ea24f24 - Changelog update (and reword previous entry for clarity)

# Additional Considerations #

#### Zero reverse compatibility ####

As noted in [hubot-stackstorm PR #148](https://github.com/StackStorm/hubot-stackstorm/pull/148), this re-implementation is not reverse-compatible with previous uses. However, there are no official clients using this endpoint yet.

#### Keyword-only arguments in Python 2 ####

I ran into some issues with keyword-only arguments in some functions [[here](https://github.com/StackStorm/st2/pull/3884/commits/89f9f740c4caf802b4ccd36b8aca60ce840a142e#diff-734ce69988d4bfce994e3e76105db6c6R275) and commented on [here](https://github.com/StackStorm/st2/pull/3884/commits/89f9f740c4caf802b4ccd36b8aca60ce840a142e#diff-734ce69988d4bfce994e3e76105db6c6R308)], so I forced them to be keyword-only arguments by unpacking them from `**filters`. PEP-3102 enables explicit keyword-only parameters, so those function definitions can be tweaked once StackStorm supports Python 3+ (and stops supporting Python 2). 🤞 

#### `re.match` vs `re.search` ####

Note that in 7d6187f I switched the regex matching function from `re.match` [[here](https://github.com/StackStorm/st2/pull/3884/commits/7d6187fef141b164b4a01a8b73e71a1c77ec4c57#diff-ef06cf3802f64c52e0fab1e5b7cb8310L121)] to `re.search` [[here](https://github.com/StackStorm/st2/pull/3884/commits/7d6187fef141b164b4a01a8b73e71a1c77ec4c57#diff-ef06cf3802f64c52e0fab1e5b7cb8310R175)]. The `re.match` function only matches at the beginning of the string, whereas the `re.search` function matches anywhere in the string.

#### `QCombination` objects from `mongoengine` ####

The test for `QCombination` objects is a little wonky since the `QCombination` class is not "exported" in that module. The `QCombination` objects themselves might be a (fairly stable) implementation detail in `mongoengine`. I can remove [the check](https://github.com/StackStorm/st2/pull/3884/commits/89f9f740c4caf802b4ccd36b8aca60ce840a142e#diff-734ce69988d4bfce994e3e76105db6c6R394) altogether if necessary so instead of this:

```python
if isinstance(arg, visitor.Q):
    # ...
elif isinstance(arg, visitor.QCombination):
    # ...
else:
   raise TypeError(...)
```

we can just do this:

```python
if isinstance(arg, visitor.Q):
    # ...
else:
    # Assume some sort of object with `children` and `operation` attributes
    # here, and blow up if our implicit expectations are not met
    # ...
```

Since `.query` is an internal API anyway I can't imagine any users passing in anything other than Q objects, but I added the check in there to give a more useful error message.

# Tests #

I have added tests for almost all additional code. I did *not* add tests for:

* Non-`Q` objects in ORM queries
* `*args` to `query` besides my usage (which causes a surprisingly large amount of coverage in `_process_arg_filters`

I can add those tests in if you would like, but frankly I think they are of marginal utility.

#### Coverage ####

I am not able to run all of the unit tests on my local machine, but of the tests that I was able to run, I was able to exercise almost every line of changed or additional code. Here are the raw results for the relevant files:

```
Name                                                Stmts   Miss Branch BrPart  Cover
-------------------------------------------------------------------------------------
st2api/controllers/v1/actionalias.py                   96     16      4      2    82%
st2api/controllers/v1/aliasexecution.py               130     26     32      8    77%
st2common/models/db/__init__.py                       243     99     76     14    56%
st2common/models/utils/action_alias_utils.py          108      2     40      4    96%
st2common/util/actionalias_helpstring.py               30      2     20      1    94%
st2common/util/actionalias_matching.py                 55      0     24      0   100%
```

The "bonus" test for `normalise_alias_format_string` in `st2common/st2common/util/actionalias_matching.py`, along with the tests for my additional code, bumps the test coverage for that file up to 💯 %!

# Documentation #

I have not yet had the time to write documentation for this additional functionality, but I absolutely will do so once this PR is merged.

# Motivation #

This PR sets the stage for some of our changes to the [StackStorm Jira pack](https://github.com/StackStorm-Exchange/stackstorm-jira) <sup>1</sup>.

Our ultimate goal with this is to have a "JiraBot" listen in on Slack channels and when anybody mentions a Jira issue, the JiraBot pulls the issue summary from Jira and dumps it into the channel (with pretty formatting of course). Since a since Slack message can contain multiple issue keys, we want to be able to dump the summary for each one into Slack.

# Final Notes #

We are running hubot-stackstorm with my changes in [hubot-stackstorm PR #148](https://github.com/StackStorm/hubot-stackstorm/pull/148) to implement multiple dispatch until this is merged and released.

This is not a reverse compatible change, so it may not be appropriate to include in a point release (even though the endpoint has no official users yet).

However, if you are willing to include it in a point release, I'm hoping to get it into the next one (2.6, as of this writing). I am more than happy to work with you towards that goal. Please let me know if you need anything from me.

Thanks!

----------------

<sup>1</sup> This is upcoming - we're waiting on the release of StackStorm 2.5.1 to develop it since the webhook from Jira will require the `search` operator from my PR #3833.

### DIFF
--- a/st2api/st2api/controllers/v1/actionalias.py
+++ b/st2api/st2api/controllers/v1/actionalias.py
@@ -78,16 +78,18 @@ class ActionAliasController(resource.ContentPackResourceController):
         command = action_alias_match_api.command
 
         try:
-            match = get_matching_alias(command=command)
+            format_ = get_matching_alias(command=command)
         except ActionAliasAmbiguityException as e:
             LOG.exception('Command "%s" matched (%s) patterns.', e.command, len(e.matches))
             return abort(http_client.BAD_REQUEST, str(e))
 
         # Convert ActionAliasDB to API
-        action_alias_api = ActionAliasAPI.from_model(match[0])
-        match = [action_alias_api, match[1], match[2]]
-        result = self._match_tuple_to_dict(match=match)
-        return result
+        action_alias_api = ActionAliasAPI.from_model(format_['alias'])
+        return {
+            'actionalias': action_alias_api,
+            'display': format_['display'],
+            'representation': format_['representation'],
+        }
 
     def help(self, filter, pack, limit, offset, **kwargs):
         """
@@ -200,13 +202,6 @@ class ActionAliasController(resource.ContentPackResourceController):
         LOG.audit('Action alias deleted. ActionAlias.id=%s.' % (action_alias_db.id), extra=extra)
 
         return Response(status=http_client.NO_CONTENT)
-
-    def _match_tuple_to_dict(self, match):
-        return {
-            'actionalias': match[0],
-            'display': match[1],
-            'representation': match[2]
-        }
 
 
 action_alias_controller = ActionAliasController()
--- a/st2api/tests/unit/controllers/v1/test_alias_execution.py
+++ b/st2api/tests/unit/controllers/v1/test_alias_execution.py
@@ -28,8 +28,9 @@ FIXTURES_PACK = 'aliases'
 
 TEST_MODELS = {
     'aliases': ['alias1.yaml', 'alias2.yaml', 'alias_with_undefined_jinja_in_ack_format.yaml',
-                'alias4.yaml'],
-    'actions': ['action1.yaml'],
+                'alias4.yaml', 'alias_fixes1.yaml', 'alias_fixes2.yaml',
+                'alias_match_multiple.yaml'],
+    'actions': ['action1.yaml', 'action2.yaml'],
     'runners': ['runner1.yaml']
 }
 
@@ -93,7 +94,6 @@ class AliasExecutionTestCase(FunctionalTest):
     def test_execution_with_array_type_single_value(self, request):
         command = 'Lorem ipsum value1 dolor sit value2 amet.'
         post_resp = self._do_post(alias_execution=self.alias2, command=command)
-        self.assertEqual(post_resp.status_int, 201)
         expected_parameters = {'param1': 'value1', 'param3': ['value2']}
         self.assertEquals(request.call_args[0][0].parameters, expected_parameters)
 
@@ -151,7 +151,7 @@ class AliasExecutionTestCase(FunctionalTest):
 
     @mock.patch.object(action_service, 'request',
                        return_value=(None, EXECUTION))
-    def test_match_and_execute(self, mock_request):
+    def test_match_and_execute_doesnt_match(self, mock_request):
         base_data = {
             'source_channel': 'chat',
             'notification_route': 'hubot',
@@ -166,6 +166,15 @@ class AliasExecutionTestCase(FunctionalTest):
         self.assertEqual(str(resp.json['faultstring']),
                          "Command 'hello donny' matched no patterns")
 
+    @mock.patch.object(action_service, 'request',
+                       return_value=(None, EXECUTION))
+    def test_match_and_execute_matches_many(self, mock_request):
+        base_data = {
+            'source_channel': 'chat',
+            'notification_route': 'hubot',
+            'user': 'chat-user'
+        }
+
         # Command matches more than one pattern
         data = copy.deepcopy(base_data)
         data['command'] = 'Lorem ipsum banana dolor sit pineapple amet.'
@@ -175,16 +184,75 @@ class AliasExecutionTestCase(FunctionalTest):
                          "Command 'Lorem ipsum banana dolor sit pineapple amet.' "
                          "matched more than 1 pattern")
 
+    @mock.patch.object(action_service, 'request',
+                       return_value=(None, EXECUTION))
+    def test_match_and_execute_matches_one(self, mock_request):
+        base_data = {
+            'source_channel': 'chat',
+            'notification_route': 'hubot',
+            'user': 'chat-user'
+        }
+
         # Command matches - should result in action execution
         data = copy.deepcopy(base_data)
         data['command'] = 'run date on localhost'
         resp = self.app.post_json('/v1/aliasexecution/match_and_execute', data)
         self.assertEqual(resp.status_int, 201)
-        self.assertEqual(resp.json['execution']['id'], str(EXECUTION['id']))
-        self.assertEqual(resp.json['execution']['status'], EXECUTION['status'])
+        self.assertEqual(len(resp.json['results']), 1)
+        self.assertEqual(resp.json['results'][0]['execution']['id'], str(EXECUTION['id']))
+        self.assertEqual(resp.json['results'][0]['execution']['status'], EXECUTION['status'])
 
         expected_parameters = {'cmd': 'date', 'hosts': 'localhost'}
         self.assertEquals(mock_request.call_args[0][0].parameters, expected_parameters)
+
+    @mock.patch.object(action_service, 'request',
+                       return_value=(None, EXECUTION))
+    def test_match_and_execute_matches_one_multiple_match(self, mock_request):
+        base_data = {
+            'source_channel': 'chat',
+            'notification_route': 'hubot',
+            'user': 'chat-user'
+        }
+
+        # Command matches multiple times - should result in multiple action execution
+        data = copy.deepcopy(base_data)
+        data['command'] = 'JKROWLING-4 is a duplicate of JRRTOLKIEN-24 which is a duplicate of DRSEUSS-12'
+        resp = self.app.post_json('/v1/aliasexecution/match_and_execute', data)
+        self.assertEqual(resp.status_int, 201)
+        self.assertEqual(len(resp.json['results']), 2)
+        self.assertEqual(resp.json['results'][0]['execution']['id'], str(EXECUTION['id']))
+        self.assertEqual(resp.json['results'][0]['execution']['status'], EXECUTION['status'])
+        self.assertEqual(resp.json['results'][1]['execution']['id'], str(EXECUTION['id']))
+        self.assertEqual(resp.json['results'][1]['execution']['status'], EXECUTION['status'])
+
+        # The mock object only stores the parameters of the _last_ time it was called, so that's
+        # what we assert on. Luckily re.finditer() processes groups in order, so if this was the
+        # parameters to the mock object, we have _also_ called it with:
+        #
+        # {'issue_key': 'JRRTOLKIEN-24'}
+        #
+        # We've also already checked the results array
+        #
+        expected_parameters = {'issue_key': 'DRSEUSS-12'}
+        self.assertEquals(mock_request.call_args[0][0].parameters, expected_parameters)
+
+    @mock.patch.object(action_service, 'request',
+                       return_value=(None, EXECUTION))
+    def test_match_and_execute_matches_many_multiple_match(self, mock_request):
+        base_data = {
+            'source_channel': 'chat',
+            'notification_route': 'hubot',
+            'user': 'chat-user'
+        }
+
+        # Command matches multiple times - should result in multiple action execution
+        data = copy.deepcopy(base_data)
+        data['command'] = 'JKROWLING-4 fixes JRRTOLKIEN-24 which fixes DRSEUSS-12'
+        resp = self.app.post_json('/v1/aliasexecution/match_and_execute', data, expect_errors=True)
+        self.assertEqual(resp.status_int, 400)
+        self.assertEqual(str(resp.json['faultstring']),
+                         "Command '{command}' "
+                         "matched more than 1 (multi) pattern".format(command=data['command']))
 
     def _do_post(self, alias_execution, command, format_str=None, expect_errors=False,
                  show_secrets=False):
--- a/st2common/st2common/models/db/__init__.py
+++ b/st2common/st2common/models/db/__init__.py
@@ -20,6 +20,7 @@ import ssl as ssl_lib
 
 import six
 import mongoengine
+from mongoengine.queryset import visitor
 from pymongo import uri_parser
 from pymongo.errors import OperationFailure
 
@@ -271,10 +272,13 @@ class MongoDBAccess(object):
     def get_by_pack(self, value):
         return self.get(pack=value, raise_exception=True)
 
-    def get(self, exclude_fields=None, *args, **kwargs):
+    def get(self, *args, **kwargs):
+        exclude_fields = kwargs.pop('exclude_fields', None)
         raise_exception = kwargs.pop('raise_exception', False)
 
-        instances = self.model.objects(**kwargs)
+        args = self._process_arg_filters(args)
+
+        instances = self.model.objects(*args, **kwargs)
 
         if exclude_fields:
             instances = instances.exclude(*exclude_fields)
@@ -292,23 +296,34 @@ class MongoDBAccess(object):
         return self.query(*args, **kwargs)
 
     def count(self, *args, **kwargs):
-        result = self.model.objects(**kwargs).count()
+        result = self.model.objects(*args, **kwargs).count()
         log_query_and_profile_data_for_queryset(queryset=result)
         return result
 
-    def query(self, offset=0, limit=None, order_by=None, exclude_fields=None,
-              **filters):
+    # TODO: PEP-3102 introduced keyword-only arguments, so once we support Python 3+, we can change
+    #       this definition to have explicit keyword-only arguments:
+    #
+    #           def query(self, *args, offset=0, limit=None, order_by=None, exclude_fields=None,
+    #                     **filters):
+    def query(self, *args, **filters):
+        # Python 2: Pop keyword parameters that aren't actually filters off of the kwargs
+        offset = filters.pop('offset', 0)
+        limit = filters.pop('limit', None)
+        order_by = filters.pop('order_by', None)
+        exclude_fields = filters.pop('exclude_fields', None)
+
         order_by = order_by or []
         exclude_fields = exclude_fields or []
         eop = offset + int(limit) if limit else None
 
+        args = self._process_arg_filters(args)
         # Process the filters
         # Note: Both of those functions manipulate "filters" variable so the order in which they
         # are called matters
         filters, order_by = self._process_datetime_range_filters(filters=filters, order_by=order_by)
         filters = self._process_null_filters(filters=filters)
 
-        result = self.model.objects(**filters)
+        result = self.model.objects(*args, **filters)
 
         if exclude_fields:
             result = result.exclude(*exclude_fields)
@@ -342,11 +357,11 @@ class MongoDBAccess(object):
     def delete(self, instance):
         return instance.delete()
 
-    def delete_by_query(self, **query):
+    def delete_by_query(self, *args, **query):
         """
         Delete objects by query and return number of deleted objects.
         """
-        qs = self.model.objects.filter(**query)
+        qs = self.model.objects.filter(*args, **query)
         count = qs.delete()
         log_query_and_profile_data_for_queryset(queryset=qs)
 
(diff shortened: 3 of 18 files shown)


Kami:
@nzlosh ^ just a heads up.

I believe you were looking in utilizing new match and execute API endpoint. This PR changes the response format from a single item to a list because now a single API call and alias match can result in multiple action executions.

Kami:
Thanks for this PR, will have a more detailed look today.

ahubl-mz:
Not sure why the Travis tests aren't running. 😕

And I have zero visibility into the E2E tests - can somebody give me a traceback or a hint?

ahubl-mz:
@Kami Since every rebase means I have to update all of the links in the description, I'll wait to rebase onto `HEAD` until after your review.

nzlosh:
Thanks for the heads-up @Kami.  👍

st2api/st2api/controllers/v1/actionalias.py
```diff
@@ -78,16 +78,18 @@ def match(self, action_alias_match_api):
...
         except ActionAliasAmbiguityException as e:
             LOG.exception('Command "%s" matched (%s) patterns.', e.command, len(e.matches))
             return abort(http_client.BAD_REQUEST, str(e))
 
         # Convert ActionAliasDB to API
-        action_alias_api = ActionAliasAPI.from_model(match[0])
-        match = [action_alias_api, match[1], match[2]]
-        result = self._match_tuple_to_dict(match=match)
-        return result
+        action_alias_api = ActionAliasAPI.from_model(format_['alias'])
```
Kami: Does this change the response format of match API endpoint?

Because if it does, it's not backward compatible (match and match and execute are two different endpoints, match and execute is the recently added one which didn't have any official consumers yet).

Kami: Never mind, I see `_match_to_tuple` change below :D

st2api/tests/unit/controllers/v1/test_alias_execution.py
```diff
@@ -175,17 +190,76 @@ def test_match_and_execute(self, mock_request):
...
+    @mock.patch.object(action_service, 'request',
+                       return_value=(None, EXECUTION))
+    def test_match_and_execute_matches_many_multiple_match(self, mock_request):
+        base_data = {
+            'source_channel': 'chat',
+            'notification_route': 'hubot',
+            'user': 'chat-user'
+        }
+
+        # Command matches multiple times - should result in multiple action execution
```
Kami: I'm a bit confused about this comment and test. Why do we expect and error here? Is `match_multiple` set to false or something else is going on?

Edit: I see, it looks like the reason is there are two overlapping aliases, right?

st2common/st2common/models/db/__init__.py
```diff
@@ -271,10 +272,13 @@ def get_by_ref(self, value):
     def get_by_pack(self, value):
         return self.get(pack=value, raise_exception=True)
 
-    def get(self, exclude_fields=None, *args, **kwargs):
+    def get(self, *args, **kwargs):
```
Kami: I saw your comment on the PR about keyword args, but I'm not sure exactly what kind of issue you had in Python 2.5.

Combining keyword arguments + `**kwargs` should work fine, we use it in many places (and I really don't like `**kwargs` stuff in general since it makes code hard to read and understand what is going on, but sadly that's legacy in some places). Am I missing something?

st2common/st2common/models/db/__init__.py
```diff
@@ -292,23 +296,34 @@ def get_all(self, *args, **kwargs):
         return self.query(*args, **kwargs)
 
     def count(self, *args, **kwargs):
-        result = self.model.objects(**kwargs).count()
+        result = self.model.objects(*args, **kwargs).count()
```
Kami: Good catch :+1:

st2common/st2common/models/db/__init__.py
```diff
@@ -292,23 +296,34 @@ def get_all(self, *args, **kwargs):
...
         log_query_and_profile_data_for_queryset(queryset=result)
         return result
 
-    def query(self, offset=0, limit=None, order_by=None, exclude_fields=None,
-              **filters):
+    # TODO: PEP-3102 introduced keyword-only arguments, so once we support Python 3+, we can change
```
Kami: I would prefer to change pep8 rule or whatever the rule existing code was violating and keep it as it was.

And yeah correct long term solution is to get rid of `**filters` and replace it with explicit argument definition.

st2common/st2common/models/utils/action_alias_utils.py
```diff
@@ -117,36 +132,69 @@ def get_extracted_param_value(self):
...
+            result[pair[0]] = ''.join(pair[2:])
 
         if self._format and not (self._param_stream.strip() or any(result.values())):
             raise ParseException('No value supplied and no default value found.')
 
         return result
 
+    def get_extracted_param_value(self):
+        """
+        Match command against the format string and extract paramters from the command string.
```
Kami: Small type `paramters` -> `parameters`.

st2common/st2common/models/utils/action_alias_utils.py
```diff
@@ -117,36 +132,69 @@ def get_extracted_param_value(self):
...
+
+        :rtype: ``dict``
+        """
+        result = {}
+
+        # As there's a lot of questions about using regular expressions,
+        # I'll try to be thorough when documenting this code.
+
+        # 4. Matching the command against our regex to get the param values
```
Kami: Some in-line comments probably need to be updated :)

Kami:
I added some comments, mostly small things.

I will look into e2e test and gist the error. All the workflows, etc. are public, but the actual VM where tests run is not so it's quite a hassle to debug sometimes, even for us who have access to the VM.

Kami:
On a related note - do you have any plans to also work on the hubot changes on those st2 changes are in? :)

Kami:
Here is an output of a failing tests - https://gist.github.com/Kami/a02279bae9edde3cd97aac5af6bbeab5#file-1-bash-L510

It looks like some of the tests from here are failing - https://github.com/StackStorm/st2tests/tree/master/robotfm_tests/chatopsCI

Kami:
I believe this change is breaking it - https://github.com/StackStorm/st2/pull/3884/files#diff-f2911a69b88d56311f8945335e9457b7R151

It's a backward incompatible change.

It's OK to change "match_and_execute" API endpoint as you did in your change, but we need to keep "POST /v1/aliasexecution" aka endpoint which creates a new execution based on the format string and command intact and backward compatible.

This endpoint is used right now by hubot to schedule an execution based on the alias and format string which matched locally inside hubot.

Right now hubot performs the matching locally and sends both - matched format string and command to the API, but long-term goal is to utilize new match and execute endpoint (send every chat line there) and deprecate this endpoint which will then become obsolete.

So even though this endpoint will become mostly obsolete / unused in the future, we need to keep it fully backward compatible so it doesn't break all the existing installations.

st2common/st2common/models/utils/action_alias_utils.py
```diff
@@ -117,36 +132,69 @@ def get_extracted_param_value(self):
...
+        matched_stream = self._regex.search(self._param_stream)
+
+        return self.match_params_in_stream(matched_stream)
 
-def extract_parameters_for_action_alias_db(action_alias_db, format_str, param_stream):
+    def get_multiple_extracted_param_value(self):
+        """
+        Match command against the format string and extract paramters from the command string.
```
ahubl-mz: I really do not know how to spell "parameters" apparently. I'll fix this and any other misspellings of it.

st2api/tests/unit/controllers/v1/test_alias_execution.py
ahubl-mz: No, in the `test_match_and_execute_matches_many_multiple_match` test there are two aliases that matched. The intent of my code, and of the original code, was to only ever match *one* alias, never more than one.

However, that one matched alias could match the command multiple times, and cause multiple actions to be executed.

Pseudocode looks like:

```python
matched_aliases = []
for alias in all_aliases:
    for format in alias.formats:
        for representation in format.representations:
            if representation.match(command):
                matches.append(alias)

assert len(matched_aliases) == 1

alias = matched_aliases.first()

results = []
for match_parameters in alias.match(command):
    results.append(alias.action.execute(match_parameters))

return results
```

----------------

Now that we are returning an array of match results, we could also allow multiple aliases to match. That pseudocode would look like:

```python
matched_aliases = []
for alias in all_aliases:
    for format in alias.formats:
        for representation in format.representations:
            if representation.match(command):
                matched_aliases.append(alias)

results = []
for matched_alias in matched_aliases:
    for match_parameters in matched_alias.match(command):
        results.append(matched_alias.action.execute(match_parameters))

return results
```

Is matching multiple aliases acceptable?

st2common/st2common/models/utils/action_alias_utils.py
ahubl-mz: Not the only place I misspell it. I'll fix it.

st2api/st2api/controllers/v1/actionalias.py
ahubl-mz: The variable named `match` wasn't very clear the type of object it was matching (alias/format/representation?), so I renamed it to `format_` to make it more clear (and to avoid stomping on the `format` built-in).

st2api/tests/unit/controllers/v1/test_alias_execution.py
ahubl-mz: Just saw your edit: yep, that's why. Should I match multiple aliases since I'm changing the semantics of `match_and_execute` anyway?

st2common/st2common/models/db/__init__.py
ahubl-mz: Python 3 lets you put `*args` before any arguments that you want to be keyword-only arguments. Like this:

```python
def asdf(arg1, *args, kwarg=None, **kwargs):
    pass
```

But the exact same code will get you a syntax error in Python 2.7 (and presumably below).

```python
>>> def asdf(arg1, *args, kwarg=None, **kwargs):
  File "<stdin>", line 1
    def asdf(arg1, *args, kwarg=None, **kwargs):
                              ^
SyntaxError: invalid syntax
```

I have not checked Python 2.5 yet. Are some StackStorm users still running it with that?

st2common/st2common/models/db/__init__.py
ahubl-mz: Unfortunately that's not possible AFAIK. The syntax for keyword-only arguments in Python 3 causes a `SyntaxError` in Python 2.7. This is one of the changes in Python 3 that isn't backwards compatible.

It is, however, a good reason to start coding in Python 3. 😜

st2common/st2common/models/utils/action_alias_utils.py
ahubl-mz: Yep.

ahubl-mz:
@Kami Thanks for the review, I'll rebase onto `master`, fix the things you pointed out, and re-push.

st2common/st2common/models/utils/action_alias_utils.py
ahubl-mz: I'm not the only one!

* https://github.com/mz-techops/st2/blob/2f3315ab62a94a6bbb138c731b3db72678f8086e/contrib/runners/local_runner/tests/integration/test_localrunner.py#L356
* https://github.com/mz-techops/st2/blob/2f3315ab62a94a6bbb138c731b3db72678f8086e/st2common/tests/unit/test_resource_registrar.py#L165

(Fixing those too.)

st2common/st2common/models/db/__init__.py
ahubl-mz: Expanding a bit more on this change, `exclude_fields` is not a keyword-only argument in Python 2, so if I were to try to use `Q` objects, like so:

```python
ActionAlias.get(
    Q(formats__representations__match_multiple=None)
    | Q(formats__representations__match_multiple=False),
    enabled=True)
```

then within the original `.get` function, it would unpack to:

```python
exclude_fields = Q(
    QCombination(
        operation=QCombination.OR,
        children=(QNode(formats__representations__match_multiple=None),
                  QNode(formats__representations__match_multiple=False)))
args = (,)  # Empty tuple
kwargs = {}  # Empty dict
```

Now, a `QCombination` object is not a reasonable value for the `exclude_fields` variable, so code that used `exclude_fields` could explode later on.

The alternative that I have coded up makes `exclude_fields` an **implicit**, keyword-only argument in Python 2. This isn't ideal, but there's not really any reasonable way to make the `.get` function accept `*args` without confusing it with `exclude_fields`.

I briefly considered checking 

```python
if instanceof(exclude_fields, QCombination) and not args:
    args = exclude_fields
    exclude_fields = kwargs.pop('exclude_fields')
```

but that seemed like too much of a hack to me.

Only Python 3 will allow us to make `exclude_fields` an **explicit**, keyword-only argument.

st2common/st2common/models/db/__init__.py
Kami: Is it possible to always use keyword arguments when calling function then? I assume that's what we do right now and that's why we don't need `*args*.

st2common/st2common/models/utils/action_alias_utils.py
Kami: Haha, could have actually been me - probably miss-typed it once and then autocomplete took care of the rest :D

st2common/st2common/models/db/__init__.py
ahubl-mz: It's possible to force all arguments to be keyword-only, but I would recommend against that.

Right now the entire `MongoDBAccess` class is simply a wrapper around `mongoengine`, and `MongoDBAccess` tweaks filter arguments a bit to allow more expressiveness in queries. The `mongoengine` module is explicitly trying to copy Django's ORM API:

> It uses a simple declarative API, similar to the Django ORM.

I've developed a few websites with Django, which is where [`Q` objects came from](https://docs.djangoproject.com/en/1.11/topics/db/queries/#complex-lookups-with-q-objects).

Because our `MongoDBAccess` class is slightly closer to Django's ORM (we may want to look into upstreaming the `_process_arg_filters`, `_process_datetime_range_filters`, and `_process_null_filters` to `mongoengine`), it's (hopefully/presumably) easier for Python webdevs to contribute changes to StackStorm.

So I would keep `Q` objects as positional arguments and force pretty much every other argument into being keyword-only, simply because that's how Django and `mongoengine` do it.

And you are correct - (as far as I found) every use of `query` and `get` passes the `offset`, `limit`, `order_by`, and `exclude_fields` parameters as keyword arguments. However, passing them as keyword arguments hasn't been previously enforced. This change forces them to be.

st2common/st2common/models/utils/action_alias_utils.py
ahubl-mz: Done

st2common/st2common/models/utils/action_alias_utils.py
ahubl-mz: No worries! I fixed every occurrence I found.

ahubl-mz:
Rebased against `master`, fixed the `POST /aliasexecution` endpoint to keep reverse compatibility, fixed a few typos, moved and added a few comments around, updated links to commits in the PR description, and responded to a few comments on my commits.

I believe I've taken care of all concerns, but I'm happy to respond to any additional ones.

st2api/tests/unit/controllers/v1/test_alias_execution.py
Kami: I think it's fine the way it is right now :)

Kami:
Thanks for addressing the comments.

The changes look good to me overall, only thing I'm not happy about is usage of `*args` and `**kwargs` instead of explicit argument declaration so hopefully we can get rid of this in the near future :)

Kami:
@ahubl-mz Can you please push a dummy commit to trigger a Travis CI build?

It looks like it didn't like the rebase and I can't manually trigger a build for this PR from the Travis UI.

Kami:
On a related note - can you please add corresponding changelog entry / entries.

ahubl-mz:
I've pushed a dummy commit twice before now and Travis didn't ever pick it up. I've pushed a commit for the changelog (CircleCI is running it), but it still seems like Travis isn't picking it up (I added `[skip ci]` to the commit message for the changelog but Travis should still test commit ac826be).

Normally Travis gives you the option of rerunning a build (should be under "More Options" on the individual build pages), but if Travis isn't even receiving the webhook from GitHub that option probably doesn't work. I'm unsure if the issue is on the GitHub side of things or the Travis side of things. 😕

I'm working on setting up Travis for our fork of ST2 ([mz-techops/st2](https://github.com/mz-techops/st2)) so I can debug this issue.

Kami:
@ahubl-mz Sadly you can only re-run builds for local branches and not for PRs (aka branches which are not part of the origin remote) inside the Travis UI.

In the worst case, I will manually override the check and merge it.

I would imagine it's some wird issue with rebase or similar because I pushed commits to my other PR earlier today and Travis work fine.

ahubl-mz:
Yeah, for whatever reason PR #3885 worked just fine: https://travis-ci.org/StackStorm/st2/builds/309357890 ¯\\\_(ツ)\_/¯

Kami:
I will override the Travis CI check for this PR and go ahead and merge it.

Thanks again - we can sync up on how to proceed with the corresponding hubot change - if you will have a chance to do it in the near future.

ahubl-mz:
@Kami I should be able to work on it later this week. I'll put my ideas for hubot-stackstorm in that PR.

PR #3891, opened 2017-12-01
[WIP] Add MY_EXECUTION_VIEW Permission Type

cognifloyd:
MY_EXECUTION_VIEW is a limited form of EXECUTION_VIEW.
It allows a user to only view executions that they started.

This is important because, though ACTION_EXECUTE implies ACTION_VIEW,
it does not imply EXECUTION_VIEW. If a user had a role that gave them
ACTION_VIEW on a particular action, then they would get the implicit
EXECUTION_VIEW permission on all executions for that action even if they
did not start the action. The implicit permission from ACTION_EXECUTE
to ACTION_VIEW and from ACTION_VIEW to EXECUTION_VIEW make sense, but
adding an additional implicit permission from ACTION_EXECUTE to
EXECUTION_VIEW would, perhaps, open up more data than intended. This
resolves that by introducing a concept of execution ownership.

MY_EXECUTION_VIEW can be a global permission, or it can apply to a pack
or a particular action in a pack. As a global permission, a user can
view any execution that they started. As a pack-level permission, a user
can view any action executions when the associated action is part of a
particular pack and the execution was started by that same user. As an
action-level permission, the user is allowed to see the executions for a
particular action if they started that execution.

Example roles:
```yaml
---
name: "view_my_executions"
description: "Allow a user to view any of their own executions"
permission_grants:
  -
    permission_types:
      - "my_execution_view"
```
```yaml
---
name: "my_linux_actions"
description: "Allow a user to execute linux actions and view only the linux actions they started"
permission_grants:
  -
    resource_uid: "pack:linux"
    permission_types:
      - "action_execute"
      - "my_execution_view"
```
```yaml
---
name: "my_check_loadavg"
description: "Allow a user to execute check_loadavg actions and view only the executions they started"
permission_grants:
  -
    resource_uid: "action:linux:check_loadavg"
    permission_types:
      - "action_execute"
      - "my_execution_view"
```

NOTE: This only adds the MY_EXECUTION_VIEW permission. It does not make
it an implicit permission of any other permissions. It might make sense
to make ACTION_EXECUTE imply MY_EXECUTION_VIEW, but this commit does not
do that, leaving it as an opportunity for future development.

TODO:
- [ ] test it :)
- [ ] write unit tests

### DIFF
--- a/st2common/st2common/rbac/resolvers.py
+++ b/st2common/st2common/rbac/resolvers.py
@@ -682,22 +682,32 @@ class ExecutionPermissionsResolver(PermissionsResolver):
         action_uid = action['uid']
         action_pack_uid = pack_db.get_uid()
 
+        execution_user = resource_db['context']['user']
+        request_user = user_db['name']
+        my_execution = execution_user == request_user
+
+        if my_execution and permission_type == PermissionType.EXECUTION_VIEW:
+            self._user_has_global_permission(user_db=user_db,
+                                             permission_type=PermissionType.MY_EXECUTION_VIEW)
+
         # Note: "action_execute" also grants / implies "execution_re_run" and "execution_stop"
         if permission_type == PermissionType.EXECUTION_VIEW:
-            action_permission_type = PermissionType.ACTION_VIEW
+            action_permission_types = [PermissionType.ACTION_VIEW]
+            if my_execution:
+                action_permission_types.append(PermissionType.MY_EXECUTION_VIEW)
         elif permission_type in [PermissionType.EXECUTION_RE_RUN,
                                  PermissionType.EXECUTION_STOP]:
-            action_permission_type = PermissionType.ACTION_EXECUTE
+            action_permission_types = [PermissionType.ACTION_EXECUTE]
         elif permission_type == PermissionType.EXECUTION_ALL:
-            action_permission_type = PermissionType.ACTION_ALL
+            action_permission_types = [PermissionType.ACTION_ALL]
         elif permission_type == PermissionType.EXECUTION_VIEWS_FILTERS_LIST:
-            action_permission_type = PermissionType.EXECUTION_VIEWS_FILTERS_LIST
+            action_permission_types = [PermissionType.EXECUTION_VIEWS_FILTERS_LIST]
         else:
             raise ValueError('Invalid permission type: %s' % (permission_type))
 
         # Check grants on the pack of the action to which execution belongs to
         resource_types = [ResourceType.PACK]
-        permission_types = [PermissionType.ACTION_ALL, action_permission_type]
+        permission_types = [PermissionType.ACTION_ALL] + action_permission_types
         permission_grants = get_all_permission_grants_for_user(user_db=user_db,
                                                                resource_uid=action_pack_uid,
                                                                resource_types=resource_types,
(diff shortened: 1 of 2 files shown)


cognifloyd:
@Kami, what do you think of this? I haven't actually tested it, but I think it should be functional. It probably needs some tests as well.

We might also be able to do something like MY_EXECUTION_LIST, but that would probably be a much more invasive change, and might overlap with EXECUTION_VIEWS_FILTERS_LIST (which I don't really understand).

Kami:
Thanks for this contribution.

From a high level perspective, I'm fine with this change.

In fact, improved list and view permissions have been on our TODO for a while, but we haven't got to it yet, because most of our RBAC users have been happy with current implementation so we prioritized other features over that.

Kami:
As far as the code changes go - it would be great if you could finish them, get it working and add some tests and then we can review it :)

Kami:
Also, better permission name might be `execution_view_own` / `execution_view_self`.

This way it follows the existing conventions (resource name first, permission name / type after that).

Kami:
@cognifloyd How is this looking?

If you need any help or similar, let us know.

cognifloyd:
Sorry, I took a 2 week Christmas vacation to be with family, and then I got
the flu. I'll get back to this as soon as I can.

m4dcoder:
@cognifloyd We are trying to clear the PR queues. What's the progress of this PR? Thanks for your contribution. 👍

cognifloyd:
This overlaps slightly with `permission_isolation` option that is in #4013 
Using `execution_view_own` (Yeah, I'll rename `my_execution_view`) would probably only make sense when `permission_isolation==False`.

apolloliu:
Yeah, it will isolate resources including executions and rules through only setting `permission_isolation==True` in #4013. Sorry to tell you It's repeated for executionview. Perhaps I can split the configuration as two in terms of executions an rules respectively. But it still be wonderful that if the feature of execution view can take effect through single permission_type.

Kami:
Does those still make sense with #4054?

It seems like it's overlapping functionality and I prefer to have a single way of doing things to avoid the confusion.

In short, @cognifloyd it looks like that PR should also solve your issue?

cognifloyd:
@Kami Yes. This overlaps and is perhaps not needed. Though with #4054, I think we would need the opposite of `execution_view_own`. So, `execution_view_own` would have stepped into `isolation` on a per action or pack basis without isolation being the default. #4054 will flip that around so that things are isolated by default, so we'll need something to selectively open that up. I like starting from a locked-by-default point as that feels like a more secure starting point.

Maybe when `permission isolation` is enabled, you could have a `execution_view_without_isolation` (that's an ugly name, alternatives would be good) to allow certain users to see all the executions of certain actions or packs. For example, to allow someone who owns a service whose actions are accessible via a pack, but that same user is not a StackStorm admin, you can give that user (or group if you're using something like LDAP) access to view all of the executions for their service's pack.

I also wonder if there's a way with permission isolation for members of a group to see the executions/rules of their fellow group members. Then again, I don't use groups so I haven't looked into how groups work with RBAC in ST2. For example, if a "support" group can only run some actions relevant to their role, it would be helpful for that team to see all executions that any of their team members have run.

So, I'll close this for now, but it would be good to revisit and expand the RBAC permissions somewhat.

PR #3892, opened 2017-12-01
Retrieve live params first, include template-driven defaults

### LINKED ISSUE #3820: Jinja is not rendered in action default values prior to schema checking
# Problem

When trying to call an action with Jinja for its default value an error is raised complaining about a type mismatch.

## Action metadata

```yaml
---
description: Run a local linux command
enabled: true
runner_type: mistral-v2
entry_point: workflows/mistral-basic.yaml
name: mistral-basic
pack: examples
parameters:
  cmd:
    required: true
    type: string
  timeout:
    type: integer
    default: 60
  kv_test:
    type: integer
    default: "{{ st2kv.system.nick_test | int }}"
```

## Error - CLI

``` shell
[root@stackstorm st2]# st2 key set nick_test 123
+------------------+-----------+
| Property         | Value     |
+------------------+-----------+
| name             | nick_test |
| value            | 123       |
| expire_timestamp |           |
+------------------+-----------+

[root@stackstorm st2]# st2 run examples.mistral-basic cmd="ls"
MESSAGE: '{{ st2kv.system.nick_test | int }}' is not of type 'integer' for url: http://127.0.0.1:9101/v1/executions
```

## Error - Logs

``` shell
2017-10-30 16:03:59,072 107609136 ERROR actionexecutions [-] Unable to execute action. Parameter validation failed.
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2api/controllers/v1/actionexecutions.py", line 128, in _handle_schedule_execution
    pack=action_db.pack)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2api/controllers/v1/actionexecutions.py", line 177, in _schedule_execution
    liveaction_db, actionexecution_db = action_service.create_request(liveaction_db)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/services/action.py", line 87, in create_request
    allow_default_none=True)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/util/schema/__init__.py", line 293, in validate
    jsonschema.validate(instance=instance, schema=schema, cls=cls, *args, **kwargs)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/jsonschema/validators.py", line 541, in validate
    cls(schema, *args, **kwargs).validate(instance)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/jsonschema/validators.py", line 130, in validate
    raise error
ValidationError: u'{{ st2kv.system.nick_test | int }}' is not of type u'integer'

Failed validating u'type' in schema['properties'][u'kv_test']:
    {u'default': u'{{ st2kv.system.nick_test | int }}',
     u'type': u'integer'}

On instance[u'kv_test']:
    u'{{ st2kv.system.nick_test | int }}'
2017-10-30 16:03:59,072 107609136 ERROR router [-] Failed to call controller function "post" for operation "st2api.controllers.v1.actionexecutions:action_executions_controller.post": '{{ st2kv.system.nick_test | int }}' is not of type 'integer'
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/router.py", line 434, in __call__
    resp = func(**kw)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2api/controllers/v1/actionexecutions.py", line 575, in post
    show_secrets=show_secrets)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2api/controllers/v1/actionexecutions.py", line 134, in _handle_schedule_execution
    abort(http_client.BAD_REQUEST, re.sub("u'([^']*)'", r"'\1'", e.message))
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/router.py", line 53, in abort
    raise exc.status_map[status_code](message)
HTTPBadRequest: '{{ st2kv.system.nick_test | int }}' is not of type 'integer'
2017-10-30 16:03:59,073 107609136 INFO logging [-] 2c636a0c-d2ff-4405-9632-8268bd09d76b - 400 86 19.061ms (content_length=86,request_id='2c636a0c-d2ff-4405-9632-8268bd09d76b',runtime=19.061,remote_addr='127.0.0.1',status=400,method='POST',path='/v1/executions')
```

Mierdin:
# Problem

#3820 reported an issue that parameter defaults that use Jinja templates that return something other than string, simply don't work.

This is caused by two things:

- The code in the API controller that handles execution scheduling currently makes a call to `action_service.create_request` before the live parameters are retrieved, which is where any templates are rendered into a result value. So if a parameter is an `integer` but the default value is still a template (string) - even if that template is expected to return an int, naturally this will cause problems because it hasn't even rendered yet.
- The call to `param_utils.render_live_params` and the function calls within will render the templates, but will not actually use the resulting value, as the default values for parameters are not seen as "live". So even though there's a rendered value available to use, it's not placed in the params list because it wasn't passed in explicitly by the user, and the same problem occurs, as the unrendered template value is used (string).

# Proposed Solution

To fix the behavior observed in #3820, I propose the following changes:

- A simple re-order will address the first problem. Wait to call `action_service.request` until the live params are retrieved.
- After the existing behavior around live parameters, I suggest that we detect if a parameter's default is a template, and the resulting value of that template is actually being used (i.e. isn't being overridden by the user). In effect, we treat the resulting value as a live parameter (we're literally adding it to the same dict). If a default value is rendered and used, it shows up as if the user had passed that value in.

I think treating rendered default values as "live" parameters makes sense. Inherently, since we're using a template, it could easily be user-provided, even if the user provides it via `st2kv`. It's a bit of a paradigm change but I think it makes sense, personally.

One potential consequence to note is that the parameter schema will now have to validate against what the template's value **will be**, but IMO this should already be happening anyways. I would expect that if I'm getting a value from a template, that the datatype should be predictable, and if it's different for any reason, I would expect a validation error at runtime. We've already inherently moved this problem to runtime by using templates in params like this anyways - this just makes the resulting behavior more intuitive.

# Demo

I threw together a quick action to demonstrate the new logic:

```yaml
---
description: Run a local linux command
enabled: true
runner_type: local-shell-cmd
name: nmaludy-test
pack: testpack
parameters:
  kv_test:
    type: integer
    default: "{{ st2kv.system.nick_test | int }}"
```

I'll create the referenced key in the datastore and execute my action:

```
~$ st2 key set -s system nick_test 31415
+------------------+-----------+
| Property         | Value     |
+------------------+-----------+
| name             | nick_test |
| value            | 31415     |
| expire_timestamp |           |
+------------------+-----------+
~$ st2 run testpack.nmaludy-test cmd=date
.
id: 5a2667e632ed353390a8bc7a
status: succeeded
parameters:
  cmd: date
  kv_test: 31415
result:
  failed: false
  return_code: 0
  stderr: ''
  stdout: Tue Dec  5 09:33:27 UTC 2017
  succeeded: true
```

Note that `kv_test` appears in the resulting live parameters as if I had passed it in explicitly. While it's true, I didn't do this at the CLI, I believe the intention is the same - the default value is based on dynamic retrieval from the datastore via the template.

Note also that I can still override this with my own value - a "true" live param:

```
~$ st2 run testpack.nmaludy-test cmd=date kv_test=12345
.
id: 5a2667f832ed353390a8bc7d
status: succeeded
parameters:
  cmd: date
  kv_test: 12345
result:
  failed: false
  return_code: 0
  stderr: ''
  stdout: Tue Dec  5 09:33:44 UTC 2017
  succeeded: true
```

Closes #3820

### DIFF
--- a/st2common/st2common/util/param.py
+++ b/st2common/st2common/util/param.py
@@ -21,6 +21,7 @@ import networkx as nx
 from jinja2 import meta
 from st2common import log as logging
 from st2common.util.config_loader import get_config
+from st2common.util.jinja import is_jinja_expression
 from st2common.constants.action import ACTION_CONTEXT_KV_PREFIX
 from st2common.constants.pack import PACK_CONFIG_CONTEXT_KV_PREFIX
 from st2common.constants.keyvalue import DATASTORE_PARENT_SCOPE, SYSTEM_SCOPE, FULL_SYSTEM_SCOPE
@@ -207,12 +208,43 @@ def _cast_params_from(params, context, schemas):
     Pick a list of parameters from context and cast each of them according to the schemas provided
     '''
     result = {}
+
+    # First, cast only explicitly provided live parameters
     for name in params:
         param_schema = {}
         for schema in schemas:
             if name in schema:
                 param_schema = schema[name]
         result[name] = _cast(context[name], param_schema)
+
+    # Now, iterate over all parameters, and add any to the live set that satisfy ALL of the
+    # following criteria:
+    #
+    # - Have a default value that is a Jinja template
+    # - Are using the default value (i.e. not being overwritten by an actual live param)
+    #
+    # We do this because the execution API controller first determines live params before
+    # validating params against the schema. So, we want to make sure that if the default
+    # value is a template, it is rendered and added to the live params before this validation.
+    for schema in schemas:
+        for param_name, param_details in schema.items():
+
+            # Skip if the parameter doesn't have a default, or if the
+            # value in the context is identical to the default
+            if 'default' not in param_details or \
+                    param_details.get('default') == context[param_name]:
+                continue
+
+            # Skip if the default value isn't a Jinja expression
+            if not is_jinja_expression(param_details.get('default')):
+                continue
+
+            # Skip if the parameter is being overridden
+            if param_name in params:
+                continue
+
+            result[param_name] = _cast(context[param_name], param_details)
+
     return result
 
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -50,6 +50,9 @@ Added
 * Added flag `--auto-dict` to `st2 run` and `st2 execution re-run` commands. This flag must now
   be specified in order to automatically convert list items to dicts based on presence of colon
   (`:`) in all of the list items (new feature) #3909 
+* Jinja templates in default parameter values now render as live parameters, if no "real" live
+  parameter was provided. This allows the template to render pre-schema validation, resulting
+  in the intended value type. (improvement) #3892
 
 Changed
 ~~~~~~~
--- a/st2api/st2api/controllers/v1/actionexecutions.py
+++ b/st2api/st2api/controllers/v1/actionexecutions.py
@@ -174,8 +174,6 @@ class ActionExecutionsControllerMixin(BaseRestControllerMixin):
 
         # Schedule the action execution.
         liveaction_db = LiveActionAPI.to_model(liveaction)
-        liveaction_db, actionexecution_db = action_service.create_request(liveaction_db)
-
         action_db = action_utils.get_action_by_ref(liveaction_db.action)
         runnertype_db = action_utils.get_runnertype_by_name(action_db.runner_type['name'])
 
@@ -184,6 +182,10 @@ class ActionExecutionsControllerMixin(BaseRestControllerMixin):
                 runnertype_db.runner_parameters, action_db.parameters, liveaction_db.parameters,
                 liveaction_db.context)
         except param_exc.ParamException:
+
+            # We still need to create a request, so liveaction_db is assigned an ID
+            liveaction_db, actionexecution_db = action_service.create_request(liveaction_db)
+
             # By this point the execution is already in the DB therefore need to mark it failed.
             _, e, tb = sys.exc_info()
             action_service.update_status(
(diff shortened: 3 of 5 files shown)


nmaludy:
+100

st2common/st2common/util/param.py
```diff
@@ -207,12 +208,43 @@ def _cast_params_from(params, context, schemas):
...
+    # - Have a default value that is a Jinja template
+    # - Are using the default value (i.e. not being overwritten by an actual live param)
+    #
+    # We do this because the execution API controller first determines live params before
+    # validating params against the schema. So, we want to make sure that if the default
+    # value is a template, it is rendered and added to the live params before this validation.
+    for schema in schemas:
+        for param_name, param_details in schema.items():
+
+            # Skip if the parameter doesn't have a default, or if the
```
Kami: Is it possible to add some test cases just for this function for various scenarios which are possible here?

Kami:
Overall, from a high level perspective the change and approach looks good to me.

We had some discussion about this in the past (dynamic default parameter values) and I was strongly against it then.

My main reason was that is very hard to reason about - different values are available at different point in lifecycle of an execution (before any rendering is happen, after rendering happens, when execution is created, etc).

I'm still personally not a big fan of it (static > dynamic - a lot easier to reason about, develop, troubleshoot, see what is going on, etc.), but apparently other users see value in it and in a sense it's opt-in so I'm fine with it.

st2common/st2common/util/param.py
Kami: I approved it, but it would be great if we can add some more test cases for that, because just this for loop shows that at least 4 different scenarios are possible just in this code block.

Mierdin:
I agree with you actually. I far prefer predictable values than moving the problem to runtime. One of the reasons I go write Go code when Python starts grinding my gears again :)

In any case, my point was that IMO we crossed this boundary when we allowed templates in default values for actions in the first place. This PR just finishes that work by making the type cast portion work the way you'd expect.

st2common/st2common/util/param.py
Mierdin: Thanks for the approve, I'll add more test cases for this before merging.

PR #3893, opened 2017-12-01
Fix a bug with duplicated log messages in Python runner action stderr output

Kami:
There was a bug when all the log messages generated by a Python runner action would be duplicated when datastore service was used.

It looks like this issue was present for a very long time, but it only manifested itself when datastore service was also used inside an action (in that scenario two instances of the logger with the same name were created) so we didn't catch it earlier.

This PR fixes that by caching loggers by name and making sure there is always only a single instance of a logger with a particular name. Keep in mind that this function is called inside subprocess (python action wrapper) so there is no potential issue with this cache growing too large because cache is on per Python runner action process basis and not action runner basis.

Before:

```bash
actions.python.PrintConfigAction: DEBUG    test info
actions.python.PrintConfigAction: DEBUG    test info
actions.python.PrintConfigAction: INFO     test mesg
actions.python.PrintConfigAction: INFO     test mesg
actions.python.PrintConfigAction: DEBUG    Creating new Client object.
actions.python.PrintConfigAction: DEBUG    Creating new Client object.
actions.python.PrintConfigAction: DEBUG    Retrieving all the value from the datastore
actions.python.PrintConfigAction: DEBUG    Retrieving all the value from the datastore
actions.python.PrintConfigAction: DEBUG    test message
actions.python.PrintConfigAction: DEBUG    test message
actions.python.PrintConfigAction: INFO     test message
actions.python.PrintConfigAction: INFO     test message
```
After:

```bash
actions.python.PrintConfigAction: DEBUG    test info
actions.python.PrintConfigAction: INFO     test mesg
actions.python.PrintConfigAction: DEBUG    Creating new Client object.
actions.python.PrintConfigAction: DEBUG    Retrieving all the value from the datastore
actions.python.PrintConfigAction: DEBUG    test message
actions.python.PrintConfigAction: INFO     test message
```

### DIFF
--- a/st2common/st2common/runners/utils.py
+++ b/st2common/st2common/runners/utils.py
@@ -32,23 +32,34 @@ __all__ = [
 
 LOG = logging.getLogger(__name__)
 
+# Maps logger name to the actual logger instance
+# We re-use loggers for the same actions to make sure only a single instance exists for a
+# particular action. This way we avoid duplicate log messages, etc.
+LOGGERS = {}
+
 
 def get_logger_for_python_runner_action(action_name, log_level='debug'):
     """
     Set up a logger which logs all the messages with level DEBUG and above to stderr.
     """
-    level_name = log_level.upper()
-    log_level_constant = getattr(stdlib_logging, level_name, stdlib_logging.DEBUG)
     logger_name = 'actions.python.%s' % (action_name)
-    logger = logging.getLogger(logger_name)
 
-    console = stdlib_logging.StreamHandler()
-    console.setLevel(log_level_constant)
+    if logger_name not in LOGGERS:
+        level_name = log_level.upper()
+        log_level_constant = getattr(stdlib_logging, level_name, stdlib_logging.DEBUG)
+        logger = logging.getLogger(logger_name)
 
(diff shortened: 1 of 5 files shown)


st2common/st2common/runners/utils.py
```diff
@@ -32,23 +32,34 @@
...
 
-    formatter = stdlib_logging.Formatter('%(name)-12s: %(levelname)-8s %(message)s')
-    console.setFormatter(formatter)
-    logger.addHandler(console)
-    logger.setLevel(log_level_constant)
+        LOGGERS[logger_name] = logger
+    else:
+        logger = LOGGERS[logger_name]
 
     return logger
```
lakshmi-kannan: Will the shared logger work fine when running two actions of same kind simultaneously (mistral parallel for example)?

Kami: As described in the PR - it's per action process wrapper basis (aka per action invocation basis) so it's not an issue (two actions = two processes).

Even if it was shared among all the actions with the same name it also wouldn't be an issue. It would potentially only become an issue if it was per action runner process / service basis and there were many different actions (cache could grow kinda large in that scenario).

PR #3898, opened 2017-12-05
Disable bash history when sudo password is provided as a command line argument

Kami:
This pull request updates remote command and script runner to explicitly disable bash history when user provides `sudo_password` parameter which is passed to `sudo` binary as a command line (stdin) argument.

As mentioned here (https://github.com/StackStorm/st2/pull/3867#issuecomment-349273078), the way paramiko executes shell commands bash history is not enabled, but it's still better to be on the safe side and also try to explicitly disable it ourselves.

Disabling is done by setting `HISTFILE` (I set it to `/dev/null`, if this would become problematic in some installations where user doesn't have write access to `/dev/null` we can change that approach to try unseting this variable instead) and `HISTSIZE` environment variable and by explicitly calling `set +o history` command before any other commands which are executed as part of a paramiko session.

Even though we disable bash history, there are other ways in which user with the access to the box could intercept this password so that's just additional layer and actual limitation / potential security implications of supplying `sudo_password` parameter are documented in the docs.

### DIFF
--- a/st2common/st2common/models/system/action.py
+++ b/st2common/st2common/models/system/action.py
@@ -146,16 +146,16 @@ class ShellCommandAction(object):
         if self.env_vars:
             env_vars = copy.copy(self.env_vars)
 
-            # If sudo_passowrd is provided, explicitly disable bash history to make sure password
+            # If sudo_password is provided, explicitly disable bash history to make sure password
             # is not logged, because password is provided via command line
             if self.sudo and self.sudo_password:
                 env_vars['HISTFILE'] = '/dev/null'
                 env_vars['HISTSIZE'] = '0'
 
-            # Sort the dict to guarante consistent order
+            # Sort the dict to guarantee consistent order
             env_vars = collections.OrderedDict(sorted(env_vars.items()))
 
-            # Envrionment variables could contain spaces and open us to shell
+            # Environment variables could contain spaces and open us to shell
             # injection attacks. Always quote the key and the value.
             exports = ' '.join(
                 '%s=%s' % (quote_unix(k), quote_unix(v))
--- a/st2common/st2common/models/system/action.py
+++ b/st2common/st2common/models/system/action.py
@@ -143,11 +143,19 @@ class ShellCommandAction(object):
 
     def _get_env_vars_export_string(self):
         if self.env_vars:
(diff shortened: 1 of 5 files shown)


st2common/st2common/models/system/action.py
```diff
@@ -143,11 +144,22 @@ def _get_user_sudo_arguments(self, user):
 
     def _get_env_vars_export_string(self):
         if self.env_vars:
+            env_vars = copy.copy(self.env_vars)
+
+            # If sudo_passowrd is provided, explicitly disable bash history to make sure password
```
Mierdin: s/sudo_passowrd/sudo_password/

st2common/st2common/models/system/action.py
```diff
@@ -143,11 +144,22 @@ def _get_user_sudo_arguments(self, user):
...
+            # is not logged, because password is provided via command line
+            if self.sudo and self.sudo_password:
+                env_vars['HISTFILE'] = '/dev/null'
+                env_vars['HISTSIZE'] = '0'
+
+            # Sort the dict to guarante consistent order
```
Mierdin: s/guarante/guarantee/

st2common/st2common/models/system/paramiko_command_action.py
```diff
@@ -47,7 +47,8 @@ def get_full_command_string(self):
             command = 'sudo %s -- bash -c %s' % (sudo_arguments, command)
 
             if self.sudo_password:
-                command = 'echo -e %s | %s' % (quote_unix('%s\n' % (self.sudo_password)), command)
+                command = ('set +o history ; echo -e %s | %s' %
```
Mierdin: Now my understanding of the use of `+` here means you're enabling history, but since we're exporting `HISTFILE=/dev/null` it results in nothing being logged. Sort of a way of forcing the logging to happen on our terms (into the void). I have that right?

Kami: Afaik, it's turned on with `+o`, but I also additionally set HISTFILE and HISTSIZE just in case (you can also test `set +o history` locally, you will see this will be the last command saved in history until turning it back on with `-o`).

And I do agree, I would also expect it to work in the opposite manner, so the "API" is kinda confusing.

PR #3899, opened 2017-12-05
Fix lint violations from 3884

Kami:
For some reason Travis CI got stuck and didn't continue building PR #3884 so I overrided that check and merged it manually.

It turned out there were some small lint violations which this PR fixes.

### DIFF
--- a/st2api/st2api/controllers/v1/aliasexecution.py
+++ b/st2api/st2api/controllers/v1/aliasexecution.py
@@ -164,8 +164,10 @@ class ActionAliasExecutionController(BaseRestControllerMixin):
             if action_alias_db.ack:
                 try:
                     if 'format' in action_alias_db.ack:
+                        message = render({'alias': action_alias_db.ack['format']}, result)['alias']
+
                         result.update({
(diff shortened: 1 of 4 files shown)


ahubl-mz:
Whoops, thank you!

st2api/st2api/controllers/v1/aliasexecution.py
```diff
@@ -164,8 +164,10 @@ def _post(self, payload, requester_user, show_secrets=False, match_multiple=Fals
             if action_alias_db.ack:
                 try:
                     if 'format' in action_alias_db.ack:
+                        message = render({'alias': action_alias_db.ack['format']}, result)['alias']
+
                         result.update({
-                            'message': render({'alias': action_alias_db.ack['format']}, result)['alias']
+                            'message': message
```
ahubl-mz: Isn't this the same as `result['message'] = message` ?

Kami: Yeah, it is - I just followed the pattern existing code uses.

I needed to refactor it because of the line length violation.

PR #3900, opened 2017-12-06
Remove MongoDB database dependency from Python runner

Kami:
Not so long ago, we refactored a bunch of runners and runner container code to make runners fully standalone (they don't depend on MongoDB and RabbitMQ anymore).

We follow "dependency injection" approach and the required parameters are passed to runners by the action container service instead of being retrieved the the runner itself (that's a better abstraction anyway).

It turns out, https://github.com/StackStorm/st2/pull/3658 inadvertently added MongoDB dependency back to the Python runner.

This pull request updates the code to remove this dependency (common libs path is only retrieved if that feature is enabled in the config (off by default) and MongoDB related connection errors are ignored) so the Python runner works on AWS Lambda again.

Slightly better and long term solution is update the code so this pack_db object is retrieved inside the runner and pack_path passed to the runners by the action container service.

Or maybe even better, we don't store pack_path on the PackDB object and dynamically retrieve the path in the code by iterating over file system paths - we already do this in other places (I already originally had some concerns about that in the past - https://github.com/StackStorm/st2/pull/3658#discussion_r132112782).

In any case, we also need some tests which very that all the runners (starting with Python one) are fully standalone and we can be sure they work on lambda. If we had those tests, this issue would be caught faster and automatically.

And this is also a good example why we should aim to merge PRs faster, #3658 was lingering in limbo for a while before it was merged.

### DIFF
--- a/contrib/runners/python_runner/python_runner/python_runner.py
+++ b/contrib/runners/python_runner/python_runner/python_runner.py
@@ -26,7 +26,6 @@ from eventlet.green import subprocess
 from oslo_config import cfg
 
 from st2common import log as logging
-from st2common.persistence.pack import Pack
 from st2common.runners.base import ActionRunner
 from st2common.runners.base import get_metadata as get_runner_metadata
 from st2common.util.green.shell import run_command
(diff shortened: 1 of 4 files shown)


dzimine:
Thanks!

PR #3901, opened 2017-12-06
Match and multiple execute doc

ahubl-mz:
This PR is the companion API documentation for #3884.

I'm not super familiar with how Swagger all works, so please make sure everything looks fine, particularly my use `oneOf` to define a heterogeneous array. My intent here is that items of the `formats` array are either plain strings or objects that match the subschema.

The first commit is tweaking the injected header to make it clear what the actual command is. I tried to run `make .generate-api-spec make target`, which was incorrect.

The second commit is filling out the expected input and output JSON schemas.

### DIFF
--- a/st2common/st2common/openapi.yaml.j2
+++ b/st2common/st2common/openapi.yaml.j2
@@ -789,7 +789,7 @@ paths:
     post:
       operationId: st2api.controllers.v1.aliasexecution:action_alias_execution_controller.match_and_execute
       description: |
-        Create an execution based on action alias match (if any).
+        Create executions based on action alias match (if any).
       parameters:
         - name: input_api
           in: body
@@ -807,9 +807,16 @@ paths:
           description: User performing the operation.
       responses:
         '201':
-          description: Action alias execution created
+          description: Action alias executions created
           schema:
-            $ref: '#/definitions/AliasExecution'
+            type: object
+            properties:
+              results:
+                description: List of all matched and executed actions
+                type: array
+                items:
+                  type: object
+                  $ref: '#/definitions/AliasExecution'
           examples:
             application/json:
               ref: 'core.local'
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -4,22 +4,28 @@ sudo: required
 dist: precise
 language: python
 python:
-  - 2.7
+  - "2.7"
+  - "3.6"
```
armab: Any particular reason why Python 3 CI check is needed in this PR?

The reason I'm asking: adding more tasks into TravisCI will slow down the entire build time, since more jobs will go into the qeue (we verified that before). In short: we're using max tasks Travis gives us for free OSS, so there should be a good reason before we'll add new tasks into TravisCI.
So far as I know we don't have any plans about re-working entire platform for py3 support, but there could be smaller features like py3 runners in future.

Kami: I do believe it would be good to eventually also runs tests on Python 3, but @armab has valid point so it's better comment it out (for now).

st2common/st2common/openapi.yaml.j2
```diff
@@ -4115,9 +4122,124 @@ definitions:
...
+            type: string
+            description: Execution identifier
+          start_timestamp:
+            type: string
+            description: Execution started timestamp
+            format: datetime
+          status:
+            type: string
+            description: Execution status
+            enum: ['requested', 'scheduled', 'delayed', 'running', 'succeeded', 'failed', 'timeout', 'abandoned', 'canceling', 'canceled', 'pending', 'pausing', 'paused', 'resuming']
```
Kami: Should be able to utilize `{{ LIVEACTION_STATUSES }}` constant here (there should be same usage examples of that in this file).

Kami: Might also be able to reference existing `Execution` object definition?

.travis.yml
ahubl-mz: Nope, I'm playing around with Travis, but I realized that I'm playing around with Travis on this branch. I'll stop that (and fix this).

armab: If we can put that effort to run tests in Py3 and work on running StackStorm in py3 env, - I'm all  :100: for that to happen, but -1 if we just run the failing tests wasting the resources.

Anyways, if anyone wants to play with Python 3, - that would be a really good experiment as part of another PR :+1:

ahubl-mz: @armab Yeah, I wrote this PR, pushed this branch, and then forgot to switch branches when I started playing with Python 3 tests. Derp. 😆

I'll remove all of the Python 3 references, so this PR will only target the documentation as expected, and I'll put the Python 3 stuff into a different PR.

st2common/st2common/openapi.yaml.j2
ahubl-mz: Yes, I think I can simply reference `Execution` here.

PR #3902, opened 2017-12-07
Added capability to transform_array function to escape colons

Mierdin:
https://github.com/StackStorm/st2/pull/3670 introduced a cool new capability to split strings in an array parameter on a colon `:` character in order to specify a dictionary more easily.

However, this breaks use cases where the user wants to provide an array of strings that all contain `:` and doesn't want them split. For instance, some network interfaces have a colon in their name and it shouldn't be split up.

This PR builds on the work in #3670 and uses a negative lookbehind regular expression to ensure that the colon in question isn't preceded by a backslash: `\`. So, the user can place this behind the colon, and this conversion will be skipped.

### DIFF
--- a/st2client/tests/unit/test_command_actionrun.py
+++ b/st2client/tests/unit/test_command_actionrun.py
@@ -151,3 +151,27 @@ class ActionRunCommandTest(unittest2.TestCase):
             {'foo': '1', 'bar': '2'},
             {'hoge': 'A', 'fuga': 'B'}
         ])
+
+    def test_escape_colon(self):
+        runner = RunnerType()
+        runner.runner_parameters = {}
+
+        action = Action()
+        action.ref = 'test.action'
+        action.parameters = {
+            'param_array': {'type': 'array'},
+        }
+
+        subparser = mock.Mock()
+        command = ActionRunCommand(action, self, subparser, name='test')
+
+        mockarg = mock.Mock()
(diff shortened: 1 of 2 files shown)


Mierdin:
Some internal feedback seemed to indicate that this approach wasn't ideal for some users. Should ponder the approach further - maybe instead of allowing escape, provide a CLI option to `st2 run` to skip this conversion entirely.

Kami:
Good catch :+1: 

And yeah we have a lot of code which allows users to specify complex type in an "easier" manner, but in general that's quite hard and there are a lot of edge cases.

Are we targeting this for v2.5.1?

st2client/tests/unit/test_command_actionrun.py
```diff
@@ -151,3 +151,27 @@ def test_get_params_from_args_with_multiple_declarations(self):
...
+            'param_array': {'type': 'array'},
+        }
+
+        subparser = mock.Mock()
+        command = ActionRunCommand(action, self, subparser, name='test')
+
+        mockarg = mock.Mock()
+        mockarg.inherit_env = False
+        mockarg.parameters = [
+            'param_array=foo\:bar,foo2:bar2',
```
Kami: If we decide to go with this approach, some more test cases would be great (e.g. more than two items, etc.)

Mierdin:
Closing in favor of #3909

PR #3908, opened 2017-12-12
Throw more user-friendly error on if a cycle is found inside action parameters

Kami:
Previously only "Cyclic dependency found" error was thrown if a cycle was found (e.g. parameter is referencing itself or similar) which made it quite unfriendly and harder / slower to troubleshoot the issue.

With this change we now reference the actual variable where cycle is found to make the error more obvious and troubleshooting easier.

Before:

```
Cyclic dependency found
```

After:

```
Cyclic dependency found in the following variables: cyclic, morecyclic
```


Kami:
This would probably have saved @dzimine some time since it would immediately point him to variable "output".

PR #3909, opened 2017-12-12
Add flag to opt-in to dict conversion when colons are detected

Mierdin:
https://github.com/StackStorm/st2/pull/3670 introduced a cool new capability to split strings in an array parameter on a colon `:` character in order to specify a dictionary more easily.

However, this breaks a few corner cases, such as those where the user wants to provide an array of strings that all contain `:` and doesn't want them split. For instance, some network interfaces have a colon in their name and it shouldn't be split up.

This kicked off a long discussion internally, and we've decided to go for a more robust, exhaustive method of dealing with more complicated data types, likely to be implemented in the 2.6 release timeframe, perhaps 2.7.

Until that time, we need to take a stopgap measure to retain the functionality for those that need it, while getting things working again for some of the corner cases that don't want it. To that end, this PR implements a new flag `--auto-dict`, which you **MUST** specify to get the functionality introduced by #3670. The default functionality for `st2 run` will be to **NOT** interpret colons, but revert back to the way this was handled prior to #3670.

Note again that **this flag is temporary** and should not be relied upon long-term. Eventually, it will be replaced with more permanent, exhaustive handling of complex datatypes for action parameters.

# Usage

I put together a simple action that just returns its input values so we can see how the parameters have or haven't been affected by the use of this new flag. Note that without the flag, the default is to behave as it was before - it will not split the list items into dicts, but rather leave them as strings:

```
$ st2 run testpack.pythontest testparam=key1:value1,key2:value2
.
id: 5a30328a32ed350bde0a7ea5
status: succeeded
parameters:
  testparam:
  - key1:value1
  - key2:value2
result:
  exit_code: 0
  result:
  - key1:value1
  - key2:value2
  stderr: ''
  stdout: ''
```

However, this functionality can be opted into by using the `--auto-dict` flag:

```
$ st2 run testpack.pythontest testparam=key1:value1,key2:value2 --auto-dict
.
id: 5a30329232ed350bde0a7ea8
status: succeeded
parameters:
  testparam:
  - key1: value1
    key2: value2
result:
  exit_code: 0
  result:
  - key1: value1
    key2: value2
  stderr: ''
  stdout: ''
```

### DIFF
--- a/st2client/st2client/commands/action.py
+++ b/st2client/st2client/commands/action.py
@@ -303,6 +303,15 @@ class ActionRunCommandMixin(object):
         detail_arg_grp.add_argument('--tail', action='store_true',
                                     help='Automatically start tailing new execution.')
 
+        # Flag to opt-in to functionality introduced in PR #3670. More robust parsing
+        # of complex datatypes is planned for 2.6, so this flag will be deprecated soon
+        detail_arg_grp.add_argument('--auto-dict', action='store_true', dest='auto_dict',
+                                    default=False, help='Automatically convert list items to '
+                                    'dictionaries when colons are detected. '
+                                    '(NOTE - this parameter and its functionality will be '
+                                    'deprecated in the next release in favor of a more '
+                                    'robust conversion method)')
+
         return root_arg_grp
 
     def _print_execution_details(self, execution, args, **kwargs):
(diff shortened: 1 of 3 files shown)


samirsss:
@Mierdin quick question about the behavior from the Web ui? How will this get handled?

Mierdin:
@samirsss WebUI is unaffected, and is unrelated to this change as well as #3670

PR #3917, opened 2017-12-18
Update the port to be an integer.

NikosVlagoidis:
Fix the port to be an integer. Use the format function for string formatting.
ref #3667


LindsayHill:
Thanks @NikosVlagoidis. I've also added a tweak to make the trigger name match between the yaml & .py - they were not in sync before, which causes issues with dispatching triggers.

NikosVlagoidis:
@LindsayHill Yes, I think I correct it manually too but forget about it. Nice!

Issue #3233, opened 2017-02-17
[FEATURE] Add a Python 3 runner and pack wide virtualenv option


tonybaloney:
All of our dev effort into APIs, integrations and automation is going on Python 3.

It would be unreasonable to request StackStorm be upgraded to Python 2 so instead I'd like the option to use Python 3 packages and actions within my own packs. 
That is, 
1. Add a `python3` runner to compliment the `python` runner
2. Add the ability to configure in a pack which Python version is used with the virtualenv is created for the pack directory.

❤️

bigmstone:
Not a bad request. We had a recent debate on it too for just general pack support. One of the problems is OS support for the python version. (We still support el6 for instance). I'm curious what the wrapper would need to be 2/3 compliant. Let us get 2.2 out and I'll put some cycles towards investigating this.

Kami:
Yeah, from a high-level this sounds quite reasonable to me.

There are some edge cases (not all packages work with Python 3, etc.) but users who opt-in to use Python 3 would probably already be aware of that and we will still default to Python 2.

bjwschaap:
+1

LindsayHill:
I wonder if this is something we can do use the pluggable runners framework? Initially write a separate runner that people can install if they wish, which will use python3. Have to figure out all the virtualenv bits, but might be a way to start distributing an experimental runner?

punkrokk:
I’m very interested in seeing this and/or testing it.

etz69:
Hi, Is there any update on this ? Or is there a workaround/hack to use python3 for an extension ?

LindsayHill:
@etz69 as a workaround, you could have a shell-script action that launches a python3 script. But it won't be able to inherit from the ST2 Action class, or use pack config within the Python code, only in the parameters.

etz69:
@LindsayHill It worked fine. I packed in one directory under actions all the files (python 3 virtualenv, source etc..) and just called via bash by sourcing the right env and just passing the arguments.  Thanks !

LindsayHill:
Good to hear that it worked. It’s not ideal, and I would love to have a proper Python3 runner. But at least it gets you going

akrajewski:
Hello guys,

+1 for this feature, is there any update on it? 

[...]

PR #3922, opened 2018-01-01
Allow a pack to specify Python 3 and have corresponding virtualenvs use Python 3

tonybaloney:
As part of #3233 the pack (and/or) the user needs to be able to specify the minimum Python version required for the pack to operate.

In the case of Python 3, this requires another Python binary in the config.

This PR adds support for the `st2common.virtualenv` module to specify another Python binary path when building the env.

`st2.conf` now has a `python3_binary` setting to specify the default Python 3 binary path. This defaults to `/usr/bin/python3` in both Ubuntu and Redhat. For newer versions of ubuntu, `/usr/bin/python3` is a symlink to python3.5.

Python 3 has to be explicitly specified in the pack meta file. 

If the user does not have a Python 3 binary installed it will fail on Pack Installation, virtualenv will raise an error.

https://github.com/StackStorm/st2/pull/3922/files#diff-4572714cfd795436cbea16dff51e3231R18

The configuration is such:
```
[actionrunner]
...
# Python binary which will be used by Python actions.
python_binary = /data/stanley/virtualenv/bin/python
python3_binary = /data/stanley/virtualenv/bin/python
```

The configuration of Python 2 binaries would default to the one running the script, but since this won't be the case in 3, it defaults again to `/usr/bin/python3` https://github.com/StackStorm/st2/pull/3922/files#diff-f7f00f2d9fa85d6b39dc2def0f872561R222

pack.yaml now has a flag to specify the creation of the virtualenv using a Python 3 binary
```
system:
   python3: true
```

# TODO : 

- [x] write tests to check the integration
- [x] update pack.yaml meta template

# Notes:

The dev Makefile now explicitly specifies python 2.7 in the virtual env. Some modern distributions will default to 3.x
https://github.com/StackStorm/st2/pull/3922/files#diff-b67911656ef5d18c4ae36cb6741b7965R273

There is now a switch to change the default Python version from 2 to 3. This is not configurable since it would likely break everything! https://github.com/StackStorm/st2/pull/3922/files#diff-1672fdd913c3ff338b3d0f9285d2fa81R59 

Packs without meta files would not fail at the virtualenv stage, which is the same behaviour as before https://github.com/StackStorm/st2/pull/3922/files#diff-1672fdd913c3ff338b3d0f9285d2fa81R59


tonybaloney:
Just tested against a dev instance by:
1. Cloning the libcloud pack
2. Changing `pack.yaml` to have `python3: true` at the bottom
3. Installing the pack from git
4. Checking the Python version installed in the virtualenv

```
(virtualenv) ubuntu@ubuntu-xenial:~/st2$ /opt/stackstorm/virtualenvs/libcloud/bin/python --version
Python 3.5.2
```

tonybaloney:
Integration tested and added unit tests as well. This is now done.

tonybaloney:
@Kami It's all yours now 🎁

st2common/st2common/util/virtualenvs.py
```diff
@@ -73,6 +74,19 @@ def setup_pack_virtualenv(pack_name, update=False, logger=None, include_pip=True
...
+    try:
+        pack_meta = get_pack_metadata(pack_path)
+        has_pack_meta = True
+    except ValueError:
+        # Pack is missing meta file
+        has_pack_meta = False
+
+    if has_pack_meta:
+        logger.debug('Checking pack specific Python version.')
+        if 'python3' in pack_meta.keys():
```
Kami: A while ago we added a couple more fields to the pack model and pack.yaml file, notably `dependencies` and `system`.

Perhaps it would be better to utilize `system` field instead of a top-level one (e.g.` system.python3 = True` or similar).

We also need to decide and clarify what that attribute really means - does it means that pack supports just Python 3 or that it supports Python 3 in addition to Python 3? And once we decide what it means and represents, we should document it :)

st2common/st2common/util/virtualenvs.py
```diff
@@ -73,6 +74,19 @@ def setup_pack_virtualenv(pack_name, update=False, logger=None, include_pip=True
         msg = 'Pack "%s" is not installed. Looked in: %s' % (pack_name, search_paths)
         raise Exception(msg)
 
+    try:
+        pack_meta = get_pack_metadata(pack_path)
```
Kami: Perhaps, for dependency injection and testing reasons, this function should take `use_python3` as an argument instead of reading the metadata file itself.

Kami: Thinking out loud - I first thought this might slow down things quite a bit when running `st2ctl register --register-setup-virtualenvs` flag because we now also need to read and parse metadata file for each pack.

But after some more thought - we already need to do that, so another read and parse shouldn't add too much overhead since file should already be in filesystem cache. Future optimization would perhaps be to refactor the code a bit so we don't need to read and parse metadata file multiple times, but probably not needed right now.

Kami:
Nice work :+1:

st2common/st2common/util/virtualenvs.py
tonybaloney: Agreed on the documentation. Top-level attribute seems off also, where is the meta file model? I couldn't find it anywhere.
Currently, it means that the pack should use a Python3 virtualenv. So it supports Python 3 at least, but shouldn't fall back to 2

st2common/st2common/util/virtualenvs.py
tonybaloney: I assumed that the pack would still install if the meta-file wasn't in place, so I didn't want to alter the original behaviour

tonybaloney:
Also @Kami is the meta model typed anywhere? It seems a bit shonky to just hardcode the dictionary keys into the virtualenv logic.

tonybaloney:
What do you want me to change? Move the flag to system?

Kami:
> Also @Kami is the meta model typed anywhere? It seems a bit shonky to just hardcode the dictionary keys into the virtualenv logic.

It's not, but we can / should add a schema for it.

st2common/st2common/util/virtualenvs.py
Kami: Metadata is mandatory so it would fail at some other step, but that's not really important here and I'm fine with the current approach.

I was mostly just "thinking out loud" about the potential performance overhead from reading and parsing pack metadata multiple times during install phase :)

st2common/st2common/util/virtualenvs.py
Kami: We don't have a schema for `system` attribute yet, but you can add it and extend it here - https://github.com/StackStorm/st2/blob/master/st2common/st2common/models/api/pack.py#L135

For now, you can just add declaration for `system.python3` attribute there under `system` field.

Kami:
Hopefully my clarification on the comments helped so we can move this forward :)

Another thing which would be nice to have is ability to "force" which Python version to use. The use case would be pack declaring `system.python3 = True` in metadata, but it turns out pack doesn't actually work completely or in part under Python 3.

Not a blocker right now and another way to approach that would be to "fix" that in the upstream pack, but at least the prerequisite for us merging "system.python3 = True` change for each pack should be that we run tests for that pack under Python 3.

tonybaloney:
@Kami moved the configuration to `system.python3`. will raise issue to fix `st2-setup-tests` and a PR for docs

tonybaloney:
@Kami I'm done with this one now. The build should pass now I've run `make configgen` on a dev installation and pushed the updates back

Kami:
Also, just curious since I haven't done any Python 3 testing myself yet.

Actions are not fully self contained (same goes for sensors) and rely on some common StackStorm code which is only tested and works under Python 2.7.

I assume this could cause issues down the road so we should also add some basic StackStorm CI tests for a dummy / example python 3 pack which exercises all the "common" StackStorm code base, notably datastore functionality which is provided via action / sensor service.

Kami:
(from Slack)

> kami [1:14 PM] 
@anthonypjshaw https://github.com/StackStorm/st2/pull/3922#issuecomment-355541399 sorry for getting back to that, but i'm just worried this could cause bad experience for our users
> [1:14] 
since our code and packs are not fully tested under python 3
> [1:14] 
and using the flag could cause a bad experience, while i think explicit opt-in approach is better
> [1:14] 
we could still utilize this flag as "signaling" user that this pack also works with python 3
> [1:14] 
but it should be explicit opt-in
> [1:15] 
and eventually, when we also test more of the stackstorm code base and packs under python 3, we could perhaps do that
> [1:17] 
@anthonypjshaw if needed and / or wanted, and you agree on that approach, i can also take that over :slightly_smiling_face:

If you are fine with that approach, I can make those changes.

tonybaloney:
@Kami in all the scenarios I’m thinking of, it’s the pack author who chooses whether the pack and the actions require Python 3. This could be for a number of reasons, like a package that won’t install in Python 2 (there are more and more of those now) or they used the Python 3 standard library namespaces. Or they wanted to stick to the Python 3 string idioms. 
I would also assume the pack author is more aware and more technical than the one installing the pack. Agreed that the testing process needs to change- the pack unit tests need to run under Python 3 where it has been specified. Also, the subset of modules that are shared should also be run under 3 to check regressions. 
As an example I have packs where I’ve had to deal with Unicode inbound and encoding which I would clean up in a Python 3 action. That code would not be backward compatible with 2 and the user would not know that either.

Kami:
See https://github.com/StackStorm/st2/pull/3929#issuecomment-356072502 for somewhat related issue.

I don't feel comfortable merging this with a lot better and more complete test coverage. I'm just worried that without a complete end to end test coverage, it will just represent a liability for us and cause a bad experience and pain for us and users who will try to use it.

In general I think it's better user experience if something is not supported than having something somewhat supported, but then things break and there are many edge cases when trying to use it. That will just frustrate most of the "typical" users (there are of course more technical users who like challenges and fixing things like that, but those are far and between).

tonybaloney:
Considering the amount of effort and changes in better testing Python 3 in core libraries what is the new status of this PR ? @Kami  @LindsayHill

Kami:
Sorry, didn't have time to dig in yet and take it across the finish line.

Sadly this PR is just a small part of much bigger change to get it into "fully done" state - first I need to figure out how to finalize this PR so everyone will be happy, then documentation, then hooking it up against StackStorm Exchange CI to make sure tests correctly use Python 3 when needed, etc.

But as far as you original question goes - it will still take some time to get everything running under Python 3 and even when that is the case, we will probably still need change like that to allow users to run Python runner actions under different version of Python (e.g. 3 when platform is using 2 and also vice versa).

m4dcoder:
@Kami What's the plan for this PR?

brainstorm:
It would be rad to have it in place, today I was trying to have ansible 2.5.0rc1 and python3 but failed and found this PR... I think I'll just create a virtualenv with py3 manually for now, but it would be great to have it in, keep the good work @Kami!

Kami:
@brainstorm It's actually @tonybaloney who did the most work on this PR :)

And yeah, manually creating a virtualenv using python3 binary should work for now.

Last time we couldn't come up with a compromise on how to make this work in a robust and straight forward manner and get it merged.

Perhaps a good middle of the ground approach would be to just add `use_python3` (or similar) parameter to `pack install` action and CLI command. This way it would be opt-in and allow and easy way for user to switch back in case there are issues with Python 3.

What do others think?

Kami:
Closing - continuation in #4076.

Hopefully we can resolve some of those issues which were uncovered yesterday and get it shipped in v2.8.0.

PR #3923, opened 2018-01-03
Allow operator to change default Python runner log level in the config

### LINKED ISSUE #3889: Allow user to define default log level for (Python runner) actions
Right now we default log level for Python runner actions to `DEBUG` which means every log message with level debug or above is printed to stderr.

In a lot of scenarios this can be noisy so we should allow user to set `actionrunner.default_log_level` or similar st2 config option and override the default value.

Even better, I think we should default to it `INFO` (more reasonable default value) and allow user to override this in the config.

Note: This option can already be overridden on per-action (invocation) basis, but that's not always feasible and sometimes global override makes more sense.

Kami:
This pull request adds a change which allows operator to change default log level used by all Python runner actions by setting `actionrunner.python_runner_log_level` config option.

Keep in mind that this value can still be overriden on per action invocation basis by using `log_level` runner parameter.

Resolves #3889.


contrib/runners/python_runner/tests/unit/test_pythonrunner.py
```diff
@@ -667,6 +667,23 @@ def test_simple_action_log_messages_and_log_level_runner_param(self):
...
+        runner.entry_point = PASCAL_ROW_ACTION_PATH
+        runner.runner_parameters = {}
+        runner.pre_run()
+        (status, output, _) = runner.run({'row_index': 'e'})
+        self.assertEqual(status, LIVEACTION_STATUS_SUCCEEDED)
+        self.assertTrue(output is not None)
+        self.assertEqual(output['result'], [1, 2])
+
+        self.assertTrue(expected_msg_3 in output['stderr'])
+        self.assertFalse(expected_msg_4 in output['stderr'])
```
m4dcoder: Not really important. But I would probably do `self.assertTrue(expected_msg_4 not in output['stderr'])` so to be consistent and show the message not in the stderr output.  This caught me off guard on first pass in the review.

contrib/runners/python_runner/python_runner/python_runner.py
```diff
@@ -93,7 +94,7 @@ def __init__(self, runner_id, config=None, timeout=PYTHON_RUNNER_DEFAULT_ACTION_
         self._config = config
         self._timeout = timeout
         self._enable_common_pack_libs = cfg.CONF.packs.enable_common_libs or False
-        self._log_level = log_level
+        self._log_level = log_level or cfg.CONF.actionrunner.python_runner_log_level
```
m4dcoder: There's already a log_level in the python_runner's runner.yaml file at https://github.com/StackStorm/st2/blob/master/contrib/runners/python_runner/python_runner/runner.yaml#L16. How will this change impact the default in the runner parameter?

Kami: You are correct - now the default parameter value in runner.yaml is a bit redundant (one in the config has priority over the one in the runner metadata file).

The reason for that is that runner.yaml is not really meant to be modified by a user / operator while config is.

To reduce the confusion, I can also remove default value from the runner.yaml in the future (will leave it there for now for backward compatibility reasons).

contrib/runners/python_runner/tests/unit/test_pythonrunner.py
Kami: Yeah, this might be a bit of a personal preference, but I agree with you.

Kami:
@m4dcoder I believe I addressed all of your comments.

contrib/runners/python_runner/python_runner/python_runner.py
m4dcoder: I went back and look at the st2actions container.  https://github.com/StackStorm/st2/blob/master/st2actions/st2actions/container/base.py#L340 I don't even think the log_level is passed in anywhere. I don't even think there's a backward compat issue here. Anyway, I trust you'll remember to come back and remove default value from the runner.yaml.

PR #3924, opened 2018-01-03
Add step to skip immutable params to be overridden

humblearner:
Regression from: https://github.com/StackStorm/st2/pull/3892

```bash
$ st2 run linux.lsof hosts=localhost
MESSAGE: Override of immutable parameter(s) [u'cmd'] is unsupported. for url: http://127.0.0.1:9101/v1/executions
```


Kami:
Good catch - can you please also add a regression test case for it.

(if we had one before hand we would have caught that earlier :))

Kami:
:+1:

PR #3926, opened 2018-01-05
Enable real-time action output streaming by default

Kami:
As discussed in the past.

If needed or wanted, users can still disable it by updating st2.conf.

## TODO

- [ ] Update docs


armab:
Also CI looks right about complaining on missing `st2.conf` settings edit: https://travis-ci.org/StackStorm/st2/jobs/325390021#L838

PR #3929, opened 2018-01-06
Update st2client to support Python 3 natively

tonybaloney:
Leveraging six since it is already a dep

### DIFF
--- a/st2client/st2client/base.py
+++ b/st2client/st2client/base.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import absolute_import
 import os
 import pwd
 import json
@@ -111,7 +112,7 @@ class BaseCLIApp(object):
 
         # We skip automatic authentication for some commands such as auth
         try:
-            command_class_name = args.func.im_class.__name__
+            command_class_name = args.func.__self__.__class__.__name__
         except Exception:
             command_class_name = None
 
@@ -248,7 +249,7 @@ class BaseCLIApp(object):
             return None
 
         # Safety check for too permissive permissions
-        file_st_mode = oct(os.stat(cached_token_path).st_mode & 0777)
+        file_st_mode = oct(os.stat(cached_token_path).st_mode & 0o777)
         others_st_mode = int(file_st_mode[-1])
 
         if others_st_mode >= 4:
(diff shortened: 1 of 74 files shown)


Kami:
Thanks , nice work :+1:

Kami:
Yay, all green finally.

Will go ahead and merge it, thanks again :)

Kami:
I will also work on hooking this up to our CI (Travis to begin with) since without being it hooked up on CI and tests regularly running, it's basically the same as if we haven't done this change - it's just a question when it will get out of date and break, not if.

Kami:
@tonybaloney Did you ever get tests to work with tox?

It fails here:

```
    warning: no previously-included files matching '*.pyc' found anywhere in distribution
Downloading/unpacking idna>=2.0.0 (from requests[security]>=2.14.1,<2.15->st2client==2.6dev)
Downloading/unpacking pyOpenSSL>=0.14 (from requests[security]>=2.14.1,<2.15->st2client==2.6dev)
Downloading/unpacking cryptography>=1.3.4 (from requests[security]>=2.14.1,<2.15->st2client==2.6dev)
  Running setup.py (path:/data/stanley/.tox/py35/build/cryptography/setup.py) egg_info for package cryptography
    error in cryptography setup command: Invalid environment marker: python_version < '3'
    Complete output from command python setup.py egg_info:
    error in cryptography setup command: Invalid environment marker: python_version < '3'
```

Kami:
I managed to get it to work on Travis, but there are still Python 3 related build failures - https://travis-ci.org/StackStorm/st2/jobs/326457571.

I will try to fix those.

Kami:
I thought it will be a quick change and fix, but sadly it doesn't look like it - 319791a91968c439cff283d0ab98c696abdbc464, 0d2591a11369ef8bf15fdc79b7abc6e35b73cb6c, 7ce4f4273421b2579c6087512dc523d03890957c, etc.

I'm working on getting all the st2client tests to pass under Python 3, but it looks like it will take a bit longer than I originally thought.

Kami:
This also shows that there is no such thing as "free lunch" and supporting Python 3 through the whole platform is anything but trivial, especially if it's not taken into account (and continuously tested) early on.

Kami:
Also a test failure related to this - https://bugs.python.org/issue16308

Kami:
I decided to remove the changelog entry and not claim any Python 3 support by st2client.

There are simply too many issues and test failures. I'm worried that just by trying to get tests fixed and passed in a rushed manner, tests might indeed pass, but there will probably still be many, many issues and edge cases related to Python 3.

I already spent quite a long time trying to get things to pass and even though more tests now pass, there are still many issues which don't seem to be covered by tests.

So overall I think it's better to not claim any support, that have things in half broken state. Only time when we should claim support for Python 3 (even if just for st2client) is when we have full test coverage and everything working, but sadly this requires a lot more effort and work.

PR #3932, opened 2018-01-08
Changelog for #3929, update st2client code and tests so it works under Python 3, run st2client Python 3 tests on Travis

Kami:
This pull request adds changelog item for #3929.

In addition to that, it also includes some other changes:

* Import reorganization so it follows the existing systle
* Run st2client tests under Python 3 on Travis - if we don't have tests continuously running, it's the same as if we haven't done those changes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ env:
     - TASK=ci-unit NODE_INDEX=0 NODE_TOTAL=2
     - TASK=ci-unit NODE_INDEX=1 NODE_TOTAL=2
     - TASK=ci-integration
+    - TASK=ci-py3-tests
     - TASK="ci-checks ci-packs-tests"
 
 addons:
@@ -49,7 +50,7 @@ cache:
     - .tox/
 
 install:
-  - make requirements
+  - if [ ${TASK} = 'ci-py3-tests' ]; then pip install tox; else make requirements; fi
   - if [ ${TASK} = 'ci-unit' ] || [ ${TASK} = 'ci-integration' ]; then pip install codecov; sudo .circle/add-itest-user.sh; fi
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,16 +17,16 @@ env:
     - CACHE_NAME=JOB1
   matrix:
     include:
-        - python: 2.7
-          env: TASK=ci-unit NODE_INDEX=0 NODE_TOTAL=2
-        - python: 2.7
-          env: TASK=ci-unit NODE_INDEX=1 NODE_TOTAL=2
-        - python: 2.7
-          env: TASK=ci-integration
-        - python: 3.6
-          env: TASK=ci-st2client-py3-tests
-        - python: 2.7
-          env: TASK="ci-checks ci-packs-tests"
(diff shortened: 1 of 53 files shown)


.travis.yml
```diff
@@ -18,6 +18,7 @@ env:
     - TASK=ci-unit NODE_INDEX=0 NODE_TOTAL=2
     - TASK=ci-unit NODE_INDEX=1 NODE_TOTAL=2
     - TASK=ci-integration
+    - TASK=ci-py3-tests
```
armab: Can we name it more explicitly here to make clear that `st2client` tests are executed under `py3`?

Kami:
See https://github.com/StackStorm/st2/pull/3929#issuecomment-356072502 for details.

I decided to remove changelog entry and not claim any Python 3 support, since there are still many test cases failing and edge cases which don't work correctly under Python 3.

I will still leave some code changes and fixes I have done in place, because they do work in both Python 2 and 3, bit that's about as much as we can do right now without spending serious amount of resources on proper and correct Python 3 support.

tonybaloney:
I started down that rabbit hole but it seemed that even though the tox config was there it was never being used?

Kami:
Yeah - we don't use tox config anywhere.

We added it a while ago, but for consistency reasons we use make targets and not tox anywhere. This way we also have a better control for dependencies we want to cache across different steps, etc. which results in a faster build.

st2client/st2client/formatters/execution.py
```diff
@@ -29,6 +30,8 @@
 
 LOG = logging.getLogger(__name__)
 
+PLATFORM_MAXINT = 2 ** (struct.Struct('i').size * 8 - 1) - 1
```
tonybaloney: Python 3's int is just the original `long` type in 2, which has a near-infinite length. This could cause issues

Kami:
If you want it to also run on Travis, you need to do something like that - https://github.com/StackStorm/st2/pull/3932/commits/660eca4774f4280ebbd266096d04a1c6cc0158c9.

I removed it, because I didn't managed to get all the tests to pass. I believe I got it down to 3 failing now, some failures are related to dictionary ordering and those might be quite a pain to fix.

st2client/st2client/formatters/execution.py
Kami: This returns maxint as by the underlying C library so it should be fine aka the same as `sys.maxint` we used in the past.

Kami:
This took a lot more time then I originally hoped for, but finally have all the tests passing and running on Travis CI - https://travis-ci.org/StackStorm/st2/builds/326572524.

Some of the tests fixes are not ideal, but it should do for now.

.travis.yml
```diff
@@ -14,12 +12,18 @@ branches:
...
+    - env: TASK=ci-unit NODE_INDEX=0 NODE_TOTAL=2
+      python: 2.7
+    - env: TASK=ci-unit NODE_INDEX=1 NODE_TOTAL=2
+      python: 2.7
+    - env: TASK=ci-integration
+      python: 2.7
+    - env: TASK="ci-checks ci-packs-tests"
+      python: 2.7
+    - env: TASK=ci-st2client-py3-tests
+      python: 3.6
```
armab: Nice :+1:

PR #3937, opened 2018-01-10
Make execution deletion inside garbage collector more efficient

### LINKED ISSUE #3936: Optimize deleting old action execution data
Using `st2-purge-executions` or the `st2garbagecollector` process, all `ActionExecution` objects are fully loaded into memory, resulting in extremely high memory usage when cleaning up execution data in cases where actions have a lot of data as input and running the actions very frequently.

In our case we're fetching network measurements from a number of devices on a 5 minute interval, which generates a lot of data by a large amount of actions.

The troublesome line is the following: [https://github.com/StackStorm/st2/blob/6df54d7f3703dad890b83b7feed4b43584b2861a/st2common/st2common/garbage_collection/executions.py#L82](url).
Here, all `ActionExecution` objects matching the filters are unnecessarily loaded in to memory; only the ids are needed at [https://github.com/StackStorm/st2/blob/6df54d7f3703dad890b83b7feed4b43584b2861a/st2common/st2common/garbage_collection/executions.py#L108](url). This causes the deletion process to be slow and memory intensive in our case.

Loading more `ActionExecution` data into memory than the available system memory leads to the garbage collection never deleting everything, because all data is fetched before starting to delete data.

The process would run smoother and safer if only (necessary) data i.e. the ids of the `ActionExecution`s were fetched, rather than all of the data.

Kami:
This pull request makes execution deletion inside garbage collection more efficient.

Previously we retrieved and loaded whole execution objects in memory, which could get very expensive (memory wise) in situation where executions contain large result object.

This pull request fixes that to only retrieve id object since we don't need access to other fields.

Resolves #3936.


st2common/st2common/models/db/__init__.py
```diff
@@ -340,6 +340,8 @@ def query(self, *args, **filters):
         limit = filters.pop('limit', None)
         order_by = filters.pop('order_by', None)
         exclude_fields = filters.pop('exclude_fields', None)
+        only_fields = filters.pop('only_fields', None)
```
Kami: Because of the way our existing `query()` method works we can't simply chain the queryset filter methods - e.g. `query().only('foo').no_dereference()`, but we need to perform that chaining inside the query method.

That's important, because we already evaluate the queryset where we slice the result set a couple of lines below.

That's just more of a heads up / warning, because if you chained the methods like the examples I included above it wouldn't work correctly (it would have no effect) since result is already evaluated inside `query()` method. That "no effect" also wouldn't be visible unless you had a specific test case to catch that scenario.

^ @lakshmi-kannan @m4dcoder @bigmstone @enykeev  Just a heads up. This is not totally clear from the `query()` API (which mostly predates me), I only noticed it because I was using `--profile` flag to make sure `only()` works correctly.

Kami:
@m4dcoder It's not that easy to add a good test case for it - only option I have is mocking the method and checking it was called with the correct arguments.

It's not ideal, but I will add that.

Kami:
@m4dcoder test case in dccf90935729e8953782538c3a9502ec4c5a0cbd.

m4dcoder:
@Kami The problem with the current mock here is that we don't know what the expected result of the queries look like. The only fields is obvious. But how about no dereferencing? What does that output look like?  Can we create some mock Action Execution records in the DB and then query those in the unit tests?  I recall we have unit tests that write out those mock records.

Kami:
@m4dcoder I looked into test cases again and it's really hard to add any other test case which would provide value - we already have "integration" test cases which verify documents are indeed deleted from the database.

Having said that, I did add another test case for "only_fieds" filter which I believe should satisfy the requirements about a test case you asked for - 5026e91.

PR #3761, opened 2017-09-23
Add "limit=-1" support for the API and CLI and block negative values for limit query param

humblearner:
Closes #3735, Closes #3708 Partially Addresses: #3736

**Before:**
```
$ st2 execution list -n -1
ERROR: 500 Server Error: Internal Server Error
MESSAGE: Internal Server Error for url: http://127.0.0.1:9101/v1/executions/?limit=-1&exclude_attributes=result%2Ctrigger_instance&sort_desc=True&parent=null
```
```
$ st2 execution list -n -11
+--------------------------+------------+--------------+------------------------+-------------------------+------------------------+
| id                       | action.ref | context.user | status                 | start_timestamp         | end_timestamp          |
+--------------------------+------------+--------------+------------------------+-------------------------+------------------------+
| 59c4ad9a909a5049b27360ab | core.local | stanley      | succeeded (0s elapsed) | Fri, 22 Sep 2017        | Fri, 22 Sep 2017       |
|                          |            |              |                        | 06:28:42 UTC            | 06:28:42 UTC           |
| 59c4ad9c909a5049b27360ae | core.local | stanley      | succeeded (1s elapsed) | Fri, 22 Sep 2017        | Fri, 22 Sep 2017       |
|                          |            |              |                        | 06:28:44 UTC            | 06:28:45 UTC           |
| 59c4ada3909a5049b27360b1 | core.local | stanley      | succeeded (0s elapsed) | Fri, 22 Sep 2017        | Fri, 22 Sep 2017       |
|                          |            |              |                        | 06:28:51 UTC            | 06:28:51 UTC           |
| 59c4ada5909a5049b27360b4 | core.local | stanley      | succeeded (1s elapsed) | Fri, 22 Sep 2017        | Fri, 22 Sep 2017       |
|                          |            |              |                        | 06:28:53 UTC            | 06:28:54 UTC           |
+--------------------------+------------+--------------+------------------------+-------------------------+------------------------+
+----------------------------------------------------------------------------------------------------------------------------------+
| Note: Only first -11 action executions are displayed. Use -n/--last flag for more results.                                       |
+----------------------------------------------------------------------------------------------------------------------------------+

```

**After:**
```
$ st2 execution list -n -1
+--------------------------+------------+--------------+------------------------+-------------------------+------------------------+
| id                       | action.ref | context.user | status                 | start_timestamp         | end_timestamp          |
+--------------------------+------------+--------------+------------------------+-------------------------+------------------------+
[...]

PR #3939, opened 2018-01-11
[RBAC] Only allow limit=-1 to be used by admins, allow admins to override max_page_size

Kami:
This pull request adds some RBAC changes on top off #3761 we have agreed upon.

1. Only allow `limit=-1` (aka retrieve all the results at once) be used by RBAC admins
2. Allow RBAC admins to specify arbitrary large value for `?limit` parameter even if it's larger than `max_page_size` config option (in short, `max_page_size` limit only applies to non-admin users)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -40,9 +40,13 @@ Added
   (new feature) #3929 #3932
 
   Contributed by Anthony Shaw.
-* Add ``limit=-1`` support for the API to fetch full result set (CLI equivalent flag
+* Add ``?limit=-1`` support for the API to fetch full result set (CLI equivalent flag
   ``--last/-n``). Post error message for ``limit=0`` and fix corner case where negative values for
   limit query param were not handled correctly. #3761 #3708 #3735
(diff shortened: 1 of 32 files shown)


armab:
I just tried packages.
Not related to @Kami's PR, but `st2 action list` is missing `--limit` param in CLI, as well as some other actions cc @humblearner

Kami:
> Also just to double check: -1 should work for default non-enterprise StackStorm installation.

Yes, RBAC is disabled in default installation so it should work out of the box.

Kami:
@armab Good catch.

I'm going over test case to make sure we have "get all" test case for all the RBAC enabled API endpoints.

Kami:
@armab Pushed fixes and test cases, please see last couple of commits.

Kami:
Yeah, not all API endpoint support pagination right now (including `?sort` and `?offset` query parameter) and it would be to fix that and make it consistent across the board.

armab:
Just for reference, list of missing CLI: https://github.com/StackStorm/st2/issues/3940 as another issue.

Issue #3052, opened 2016-11-25
Pack Install should be Idempotent


armab:
Current `st2 pack install` always re-downloads and re-installs the pack, no matter what.

If pack is already installed - we shouldn't reinstall it again, unless some special flag or command specified. Re-downloading/re-installing by default takes time and in fact, it's `upgrade` or maybe `reinstall`.

Example story is configuration management tools, when you apply playbook/cookbook on a set of machines. Apart of deploying st2, it for example can install a bunch of st2 packs on the same machine. And it's normal to run playbook on the same machine several times. Re-downloading/Re-installing packs every time takes time and it's a waste of resources.

See related https://github.com/StackStorm/st2/pull/2982#discussion_r89614083

Kami:
I personally like it to keep it simple and idempotent (so just download and reinstall the pack).

Logic to determine if pack is already installed completely would be quite complex (would need to some how very that pack is correctly installed - checksum or similar, files are in place, etc. it's also more problematic because "install" consists of multiple steps which we can't easily make as part of a "transaction") and this introduce all kinda of edge cases probably.

If we can make the whole approach very robust, I would be fine with it though.

armab:
I don't think that current pack install is "really" Idempotent, even when pinning the version `st2 pack install github=0.0.4`.

We assume it is Idempotent in perfect "vacuum" conditions. But in real life any process involving re-running a bunch of logic could produce different results: 
- download failed
- upstream code could change (force push)
- virtualenv-related tasks failed
- pip pack dependencies changed and new versions installed
- `register pack` may re-enable previously disabled components (rules, sensors, etc), there are reported Issues
- machine is out of space which caused failed install
- etc

But the most annoying thing is that re-applying `pack install` for already installed packs takes time. For `10-20`+ packs (real-life deployments) this can take quite a lot.

If `apt-get/yum install` would work this way, we'd live in a much more painful world.

----


And yeah, I agree that it could be hard to do the right way, preserving the state and so on.
So it's not a super important thing and we can live without it, but it's definitely something that could be nice to have in future :+1:

nmaludy:
+1

PR #3942, opened 2018-01-13
Remove pyinotify dependency

### LINKED ISSUE #3934: Recommend dropping pyinotify as a dependency
st2 has a dependency on pyinotify, which is poorly maintained. Digging into it, it looks like no code in st2 references it, only the linux pack, which is no longer part of this code base.

Recommend moving to another, dropping it completely or using
https://github.com/chrisjbillington/inotify_simple

pyinotify is also a requirement in @Kami's custom build of logshipper git+https://github.com/Kami/logshipper.git@stackstorm_patched#egg=logshipper

enykeev:
Resolves #3934

### DIFF
--- a/fixed-requirements.txt
+++ b/fixed-requirements.txt
@@ -24,7 +24,6 @@ passlib==1.7.1
 lockfile==0.12.2
 python-gnupg==0.4.1
 jsonpath-rw==1.4.0
-pyinotify==0.9.6
 semver==2.7.7
 stevedore==1.27.1
 paramiko==2.3.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -27,7 +27,6 @@ passlib==1.7.1
 prettytable
 prompt-toolkit==1.0.15
 psutil
-pyinotify==0.9.6
 pymongo==3.5.1
 python-dateutil
 python-editor==1.0.3
--- a/st2actions/in-requirements.txt
+++ b/st2actions/in-requirements.txt
@@ -15,7 +15,6 @@ python-json-logger
 gitpython
 lockfile
 # needed by core "linux" pack - TODO: create virtualenv for linux pack on postinst
-pyinotify
 git+https://github.com/Kami/logshipper.git@stackstorm_patched#egg=logshipper
 # required by pack_mgmt/setup_virtualenv.py#L135
 virtualenv


Kami:
What's our plan for this?

Just removing this dependency without any other changes will break linux pack (file watch sensor) out of the box, so we shouldn't do that, imo.

Kami:
If we want to remove it "correctly", we need to update installer scripts / package post install script to correctly create virtual environment for those packages (IIRC, we don't do that right now).

We have also been going back and forth on this for a while now - at first, core and linux packages were not part of standard installation package and they were installed from Exchange, now they are.

In theory, it shouldn't matter how those core / default packages are installed, as long as it's transparent to the end user and packages are always installed no matter what installation method user uses.

/cc @LindsayHill

enykeev:
> Just removing this dependency without any other changes will break linux pack (file watch sensor) out of the box, so we shouldn't do that, imo.

I should have pointed out that the only reason for this PR was to check if the change triggers something in our CI process. 

> we need to update installer scripts / package post install script to correctly create virtual environment for those packages

yep, we need to. I don't really see why core or linux packs should be any different from other packs.

enykeev:
> we need to update installer scripts / package post install script to correctly create virtual environment for those packages

Actually, in light of recent discussion in regards to the future of pack management, I'd say that core and linux pack should be also installed via pack management routine i.e. via API.

armab:
> In short - I am for removing this dependency, but only if and when we found better way to mange installation of it (e.g. virtualenv creation for {core,linux} in post install or similar).
> package post install script to correctly create virtual environment for those packages (IIRC, we don't do that right now)

This will bring a regression in deb/rpm installation itself.
Adding an external dependency pulling something dynamic like that and doing compilation is a bad practice in deb/rpm pakaging. Instead of that, all external resources should be prepared during the package build stage, that's why we went from old deb/rpm to deb/rpm with `dh_virtualenv` with pre-installed pip dependencies on build stage.
Additionally, installing a package is not idempotent and potentially could fail in any step, see #3052 and relies on remote ref/resource.

> yep, we need to. I don't really see why core or linux packs should be any different from other packs.

Disagree.
Having core and linux packs located/shipped externally will add additional installation step and so complicate the story and UX.
Ideally if user has the "core" StackStorm system ready to work right after installing deb/rpm packages.

Add here air-gapped installation requirements, where users don't have immediate access to StackStorm exchange-like resources.
In short, the easy check is: installing `deb`/`rpm` shouldn't rely on any external resource, except of deb/rpm-like dependencies and shouldn't perform any type of compilation/execution, otherwise we'll return back to era with installation issues StackStorm had before.

m4dcoder:
@enykeev @armab @Kami Can you guys resolve your differences here?

blag:
Superceded by #4145.

armab:
:+1:

blag:
And #4145 was superseded by #4191.

PR #3945, opened 2018-01-17
Update publish_data.yaml

cityeremite:
{{ system.domain }}  =>>  {{ st2kv.system.domain }}    
After V2.2 ,this is not supported.


Kami:
Thanks :+1:

Issue #3856, opened 2017-11-13
latency issue with st2 execution list output


sibirajal:
Hello Team, 

I have tried to retrieve the execution stats from our ST2 instance and it took 27 seconds for retriving the 3593 records. In our case the mongodb,rabbitmq and postgresql are in seperate host. Also tried to get the execution stats with 50000 records and it failed with connection aborted error. 

The max page size limit is set to 100000 in our ST2 instance. Can you please take a look and investigate the reason for latency in retriving the results and connection error? 

time st2 execution list -n 30000 |wc -l
3593

real	0m27.943s
user	0m4.081s
sys	0m0.197s

st2 execution list -n 50000
ERROR: ('Connection aborted.', BadStatusLine("''",))

cat /etc/st2/st2.conf |grep max
max_page_size = 100000

st2 --version
st2 2.5.0

Kami:
Thanks for reporting this.

I believe it might be related to #3854.

CLI wasn't really designed to display 1000's of records at once (there is a lot of overhead in retrieving, parsing and formatting that many records inside the CLI).

If you really need to retrieve that many records at once, imo, the best way would be to directly hit the API (this way you avoid all the overhead in CLI).

And the connection error might be related to the actual result (HTTP response payload) size. Actual execution objects can be huge. The best way to avoid that is to implement pagination (e.g. retrieve 100 records at once).

Kami:
@sibirajal Can you please also clarify what you are trying to achieve / what's the expected end result / outcome?

It looks like you are trying to retrieve some statistics and there might be better ways to do that then listing all the executions :)

sibirajal:
@Kami Thanks for the response. 

I have been using the 'st2 execution list -n 10000' command to retrieve the execution results from our ST2 system for auditing purpose. Mostly we use that high numbers to pull the execution stats for last 30 days. 

Would it be possible to retrieve the same execution results using the API?

Kami:
@sibirajal Thanks.

Yes, it should be possible to retrieve that using the API - curl "http://URL/api/v1/executions". And depending on the fields you need, you can also exclude a bunch of other fields which contain a lot of data (result, trigger_instance, etc.) using `?exclude_attributes` query param filter (there are also other files which allow you to filter on status, rule, action, etc.)

Kami:
Discussed this offline - the user only needs IDs of executions from the last 30 days so the most efficient and simple solution for now is to simply directly query the database using MongoD REPL:

```bash
use st2();
[...]

PR #3953, opened 2018-01-19
"?include_attributes" query param filter for API endpoints

### LINKED ISSUE #3858: Add "only_attributes" query param filter to all the list API endpoints
Sometimes users only want to retrieve a specific field or set of field through the API and right now it's not possible to do that.

We already have "exclude_attributes" filter in the executions API endpoint, but we also need inverse of that (aka we already have "blacklist" filter, but we also need  a "whitelist" one).

Related to #3856.

Kami:
This pull request adds `?include_attributes` query param filter to some list / get all API endpoints, notably `GET /v1/actionexecutions`.

We already have `?exclude_attributes` API endpoint filter which allows user to exclude attributes which are returned in the response (handy when you want to filter out large fields such as `result` and `trigger_instance`) which works in "blacklist" manner.

`?include_attributes` is an inverse of that and works in "whitelist" manner - allows user to specify which attributes / fields to include.

This comes handy in many scenarios, e.g. when you want to perform some aggregation and statistics on top of action executions and you only care about the id and status field or similar. Previously, only way to efficiently do that was to directly query MongoDB via mongo REPL or similar which is obviously not ideal and not a good abstraction.

In addition to that change, I also made some changes and improvements to the API router code to handle casting of array query parameters inside the router and not inside the corresponding API controllers (more generic and centralized, that should happen there and it's responsibility of the router anyway).

Some examples:

```bash
(virtualenv)vagrant@vagrant-ubuntu-trusty-64:/data/stanley$ curl "http://127.0.0.1:9101/v1/executions?include_attributes=id,status&limit=2"
[
    {
        "status": "failed",
        "start_timestamp": "2018-01-19T13:36:12.446392Z",
        "id": "5a4e90d50640fd0bfbb366ee"
    },
    {
        "status": "failed",
        "start_timestamp": "2018-01-19T13:36:12.446683Z",
        "id": "5a4e90280640fd0bfbb366eb"
    }
]

virtualenv)vagrant@vagrant-ubuntu-trusty-64:/data/stanley$ curl "http://127.0.0.1:9101/v1/executions?include_attributes=id&limit=2"
[
    {
        "start_timestamp": "2018-01-19T13:36:25.658008Z",
        "id": "5a4e90d50640fd0bfbb366ee"
    },
    {
        "start_timestamp": "2018-01-19T13:36:25.658589Z",
        "id": "5a4e90280640fd0bfbb366eb"
    }
]
```

Resolves #3858 #3856


st2api/tests/unit/controllers/v1/test_actions.py
```diff
@@ -370,6 +370,36 @@ def test_get_all_exclude_attributes(self):
...
+        # Valid include attribute
+        resp = self.app.get('/v1/actions?include_attributes=entry_point')
+        self.assertEqual(resp.status_int, 200)
+        self.assertEqual(len(resp.json), 2, '/v1/actions did not return all actions.')
+        self.assertTrue(resp.json[0]['entry_point'])
+        self.assertTrue(resp.json[1]['entry_point'])
+
+        self.__do_delete(action_1_id)
+        self.__do_delete(action_2_id)
+
```
m4dcoder: Unit test for the exclude and include attributes mutual exclusion error and unit test for malformed include_attributes in the API call?

Kami: Good catch - forgot those, will add it.

Kami: Just had a look - unit test for malformed `include_attributes` is already there (see first test case / scenario in this method), added one for mutual exclusivity in 5d8d3d74a4e63e951f0e273e367716c98a591fbd, 1951ff13045a8b397c4fa782cd21fc674a9c310a.

PR #3951, opened 2018-01-18
Combination of #3950 #3949 #3948 #3947 and #3946

tonybaloney:
This is a combined branch of all the combined changes. 

If you want to review it all at once, use this one. If you want smaller pieecs, close this nd review the other 5.

This branch compileall (the stage in make tests) works under Python 3

PR #3954, opened 2018-01-22
Python 3 syntax fixes

Kami:
This pull request works on top of #3951, it:

* Fixes lint and other build failures
* Adds new `compilepy3` make target and run that on Travis

Resolves #3951 #3950 #3949 #3948 #3947 #3946


Kami:
I can't figure out why the integration test is failing and can't replicate it locally :/

Kami:
@tonybaloney 20830bc5f668417c99242c2164d3f68c552ea6e6 fixes the test failure.

I reverted the `range` change since there is really no need to it - we don't explicitly require a list and an iterable returned by `range` in Python 3 works just as fine.

It's the same story with some other `range()` and `list()` changes you have ,ade - they are not really necessary, but I will leave it as is for now.

st2common/st2common/jinja/filters/time.py
```diff
@@ -56,7 +57,7 @@ def _get_human_time(seconds):
         return '%s\u03BCs' % seconds  # Microseconds
 
     if isinstance(seconds, float):
-        seconds = long(round(seconds))  # Let's lose microseconds.
+        seconds = int(round(seconds))  # Let's lose microseconds.
```
m4dcoder: Is st2 going py3 only? Otherwise, should check for which py version and keep the long for py2.x.

st2common/st2common/services/trace.py
```diff
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import absolute_import, print_function
```
m4dcoder: Why only import the print_function here and not anywhere else where there're `print`?

st2common/st2common/jinja/filters/time.py
tonybaloney: Fixed in a3f1c0908

st2common/st2common/services/trace.py
tonybaloney: fixed in 7bc29d8a0

tonybaloney:
@Kami @m4dcoder I don't have permissions to push to this PR so you'll have to rebase those changes or patch them in.
https://github.com/StackStorm/st2/commit/a3f1c0908434a7a112fa8676eddf4dce7d0cb44d
https://github.com/StackStorm/st2/commit/7bc29d8a0b05de7a9570091620ab0dc8a65aa269

Kami:
@m4dcoder Thanks.

Merged in @tonybaloney's changes.

PR #3956, opened 2018-01-23
Add support for an array body type for non-st2 webhooks

### LINKED ISSUE #3955: Webhooks do not take array data as trigger.body
I have a webhook:
```
curl -X POST https://127.0.0.1/api/v1/webhooks/my/webhook?st2-api-key=apikey -H "Content-Type: application/json" --data '[{"data":"val"}]' -k
```

And I get this error:
```
TypeError: type object argument after ** must be a mapping, not list (_exception_data={},_exception_class='TypeError',_exception_message='type object argument after ** must be a mapping, not list')
2018-01-22 20:03:20,288 140714391321168 INFO logging [-] fcc8f336-09c3-4989-a29b-caeb398f1e8c - 500 46 8.433ms (content_length=46,request_id='fcc8f336-09c3-4989-a29b-caeb398f1e8c',runtime=8.433,remote_addr='127.0.0.1',status=500,method='POST',path='/v1/webhooks/my/webhook')
```

..I'm going to assume this means that a service cannot send list data to a webhook, is that correct?
If so, this is new for 2.5 as it worked in 2.4.  We have services that send lists..

Kami:
This pull request adds support for a body type being an array (in addition to object / dict) for custom, user- registered webhooks.

System (st2) webhooks still need to conform to a special scheme (contain trigger and payload attribute) and as such, they always need to be an object / dict.

Previously all of the API endpoints always took dictionary as an input object so I needed to make a change to the OpenAPI schema and router to support scenarios where body (input data) might not be an object.
 
## TODO

- [x] Tests

Resolves #3955.


st2common/st2common/openapi.yaml
```diff
@@ -4760,6 +4760,8 @@ definitions:
     type: object
   WebhookBody:
     description: Given support of Content-Type header, that could be anything.
+    x-api-model: st2common.models.api.webhook:WebhookBodyAPI
+    x-api-use-data-arg: True
```
Kami: I'm open to a better name of course.

It basically signals the router, that this API endpoint uses a special API model object where the actual input body / data is stored in the model `data` attribute.

With that, we can support arbitrary types for input data which are not an object.

Kami:
An alternative to this code change would be to only support objects, even for custom, user-defined webhook types.

That change would involve a small code change to throw a more user-friendly exception in case body is not an object and other part of the change would be documentation change where we would make it clear the body needs to be an object.

/cc @lakshmi-kannan @humblearner

PR #3957, opened 2018-01-23
Throw a more user-friendly exception on invalid API endpoint input type

Kami:
This pull request updates router code to throw a more user-friendly exception when invalid type is specified for the API endpoint input data. This only applies to API endpoints which don't explicitly define model schema such as `POST /v1/webhooks`.

Previously internal server error without any context was returned.

This pull request should be merged in any case, even if decide not to go with the approach implemented in #3956.

Before:

```bash
{
}
````

After:

```bash
{
    "faultstring": "Input body needs to be an object, got: array"
}
```

Related #3955 #3956

### DIFF
--- a/st2common/st2common/router.py
+++ b/st2common/st2common/router.py
@@ -37,6 +37,7 @@ from st2common.persistence.auth import User
 from st2common.rbac import resolvers
 from st2common.util import date as date_utils
 from st2common.util.jsonify import json_encode
+from st2common.util.jsonify import get_json_type_for_python_value
 from st2common.util.http import parse_content_type_header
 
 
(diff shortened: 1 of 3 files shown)


st2common/st2common/router.py
```diff
@@ -499,3 +499,16 @@ def as_wsgi(self, environ, start_response):
...
         resp = self(req)
         return resp(environ, start_response)
+
+    def _get_model_instance(self, model_cls, data):
+        try:
+            instance = model_cls(**data)
+        except TypeError as e:
+            # Throw a more user-friendly exception when input data is not an object
+            if 'type object argument after ** must be a mapping, not' in str(e):
+                msg = ('Input body needs to be an object, got: %s' % (type(data)))
```
enykeev: Strictly speaking, user provides JSON array, not <type 'list'>. He should not have to know python specifics.

Kami: True, but I'm not sure it's worth keeping such a type map inside the code, especially since the input could also be YAML.

Kami: ef93708d7969a49be5b4d6f362af0ec8670c7d4e

PR #3970, opened 2018-01-29
Fix race condition in execution output test case

Kami:
This pull request should fixes intermediate failures like the one here - https://travis-ci.org/StackStorm/st2/jobs/316638594.

The problem only manifests itself in the tests, because `get_listener` is a non-blocking method, so it's possible that `listener.listen` won't yet be called when we hit API endpoint and as such, we will miss out some messages.

This can't happen outside the tests, because listeners are pre-intialized.


armab:
I'd however re-run the Travis several times to make sure race in tests was really fixed.

PR #3971, opened 2018-01-29
Handling Large Parameters in Python Runner

### LINKED ISSUE #1598: Allow users to pass parameters which are large in size to Python actions
Currently parameters are passed to the Python action wrapper via command line arguments. This was done to provide better visibility and make Python actions easier to test and debug.

Both on Windows and Linux there is a limit on maximum size of the command line arguments (ARG_MAX, 8k on Windows, platform specific) which means users can't use really large values for parameters.

There are different ways around this limitation (e.g. store parameter value in the datastore instead), but instead of passing the arguments using command line arguments we could pass it to the action using stdin. This way users could pass parameters of (almost) arbitrary size to an action, although I don't think it makes much sense to pass arguments which are more than 4-8k or so in size.

I'm personally not convinced yet that's the best thing to do since switching to stdin decreases visibility, makes actions harder to test and if you want to pass really large value to an action as a parameter (> 8k) you should probably handle that in a different way anyway (e.g. store value in a datastore, upload it to object storage service or similar), but I'm open to feedback and I can be convinced otherwise :)

Note: This was originally reported by a user on the IRC.

bigmstone:
This PR fixes #1598. Since Linux kernel specifies an argument character limit on CLI arguments there are instances where we exceed this limit - especially when passing parameters between actions. This PR writes parameters to a tmp file if it exceeds the imposed limit and allows the wrapper to read from file when that limit is hit.

- [ ] Modify permissions to 600 for security
- [ ] Tests
- [ ] Check Corner Cases (Character limits close, but not exceeding limit)
- [x] Ensure files are deleted unless `debug` is `True`
- [ ] Ensure isolated tmp works (U16)


nmaludy:
+1

Kami:
Nice work :+1: 

As discussed on Slack, I like the idea, because it still preserves some visibility (relatively easy to debug and replicate using a file on disk).

My only concern is that there are potentially bigger security implications compared to using stdin or similar (you already mentioned some of them in the PR):

1. Need to make sure file is deleted as soon as it's read and parsed (unless debug mode is on)
2. We need to make sure file permissions are correctly, usually every user can read / write to /tmp, but yeah in this case we should use `600` and stanley / root should be the file owner

st2common/st2common/runners/utils.py
```diff
@@ -175,3 +178,12 @@ def invoke_post_run(liveaction_db, action_db=None):
...
     runner.post_run(liveaction_db.status, liveaction_db.result)
+
+
+def get_file_params_path(execution_id):
+    """Returns file path for file params
+
+    :param execution_id: Execution ID of action execution.
+    :type execution_id: ``str``
+    """
+    return os.path.join(PARAM_FILE_LOCATION, PARAM_FILE_PREPEND + execution_id)
```
Kami: To be on the safe side, we should probably use `tempfile.mkstemp` (we can still use suffix and prefix parameters of that function).

Kami: This should also take care of permissions for us, but yeah, we still need test cases for that https://docs.python.org/2/library/tempfile.html#tempfile.mkstemp

armab: +1 may secure us from the issues like https://cwe.mitre.org/data/definitions/377.html and others.

armab:
I'm not sure if it's good to keep the files, even with `debug` mode. It means file will exist with some potentially sensitive data like tokens, keys and so on.

Additionally, since `tempfile.mkstemp` doesn't handle file cleanup, my ask is to make sure the temporary files are cleaned up/deleted in 99% of the cases, ex. if user sends SIGnals like `INT` `QUIT` `HUP` `TERM` and others (to research), - service should handle those and do a proper files cleanup before exit (if they weren't cleaned before via read-parse stage).

^^ that's the corner case I'm worried about, apart of security issues @Kami noted.

Kami:
> Additionally, since tempfile.mkstemp doesn't handle file cleanup, my ask is to make sure the temporary files are cleaned up/deleted in 99% of the cases, ex. if user sends SIGnals like INT QUIT HUP TERM and others (to research), - service should handle those and do a proper files cleanup before exit (if they weren't cleaned before via read-parse stage).

There is no 100% approach aka we can't do much if SIGKILL is sent, but yeah, we should try "best effort" approach.

That's basically the downside of the file approach. Perhaps we should also prototype with stdin approach.

bigmstone:
Also potential something to add to garbage collection.

bigmstone:
I'm fine w/ stdin approach, but if we go that route I'd like to in-concert experiment with having forked wrappers waiting to execute actions to save on some of the forking/import timing.

enykeev:
> There is no 100% approach aka we can't do much if SIGKILL is sent, but yeah, we should try "best effort" approach.

I mostly kidding, but we can always encrypt the file with new key every time.

bigmstone:
Abandoned in favor of #3976

PR #3972, opened 2018-01-30
Re-gen dist_utils.py files, add missing dependency, various Python 3 compatibility fixes

Kami:
In previous PR we didn't include regenerated dist_utils.py files.

In addition to that, I noticed `st2tests/in-requirements.txt` is missing a required which is needed by tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,17 +10,17 @@ branches:
     - /^v[0-9]+\.[0-9]+$/
 
 env:
-  global:
-    - CACHE_NAME=JOB1
+  # global:
+  #  - CACHE_NAME=JOB1
 matrix:
   include:
-    - env: TASK=ci-unit NODE_INDEX=0 NODE_TOTAL=2
+    - env: TASK=ci-unit NODE_INDEX=0 NODE_TOTAL=2 CACHE_NAME=JOB1
       python: 2.7
-    - env: TASK=ci-unit NODE_INDEX=1 NODE_TOTAL=2
+    - env: TASK=ci-unit NODE_INDEX=1 NODE_TOTAL=2 CACHE_NAME=JOB1
       python: 2.7
-    - env: TASK=ci-integration
+    - env: TASK=ci-integration CACHE_NAME=JOB1
       python: 2.7
-    - env: TASK="ci-checks ci-packs-tests"
+    - env: TASK="ci-checks ci-packs-tests" CACHE_NAME=JOB1
       python: 2.7
     - env: TASK="compilepy3 ci-st2client-py3-tests" CACHE_NAME=py3
       python: 3.6
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ matrix:
       python: 2.7
     - env: TASK="ci-checks ci-packs-tests"
       python: 2.7
-    - env: TASK="compilepy3 ci-st2client-py3-tests"
+    - env: TASK="compilepy3 ci-st2client-py3-tests" CACHE_NAME=py3
       python: 3.6
 addons:
   apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,8 @@ branches:
     - /^v[0-9]+\.[0-9]+$/
 
 env:
-  # global:
-  #  - CACHE_NAME=JOB1
+  global:
+   - CACHE_NAME=JOB1
 matrix:
   include:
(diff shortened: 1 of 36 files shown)


tox.ini
```diff
@@ -28,13 +28,17 @@ commands =
...
 install_command = pip install -U --force-reinstall {opts} {packages}
 deps = -r{toxinidir}/requirements.txt
        -e{toxinidir}/st2client
 commands =
-    nosetests -sv st2client/tests/unit
+    nosetests --with-timer --rednose -sv st2client/tests/unit
+    nosetests --with-timer --rednose -sv st2api/tests/unit/controllers/v1/test_action_alias.py
+    nosetests --with-timer --rednose -sv st2api/tests/unit/controllers/v1/test_actions.py
+    nosetests --with-timer --rednose -sv st2api/tests/unit/controllers/v1/test_action_views.py
+    nosetests --with-timer --rednose -sv st2api/tests/unit/controllers/v1/test_action_views_rbac.py
```
armab: If we're adding more services here and test not only `st2client`, makes sense renaming `ci-st2client-py3-tests` in both Makefile and Travis.

Kami:
@armab We do already cache the dependencies - https://github.com/StackStorm/st2/blob/master/.travis.yml#L53 (I included that as part of my first PR to run tests under Python 3).

This folder contains virtualenv with all the dependencies for each tox environment.

armab:
@Kami Cool.
RE caching, let's please use `CACHE_NAME=py2` and `CACHE_NAME=py3` or similar instead of https://github.com/StackStorm/st2/blob/master/.travis.yml#L14 to avoid any possible conflicts between the 2 environments. Previous config was expecting shared caching env for all jobs.

Kami:
The end is near:

```bash
423 tests run in 188.199 seconds. 
12 FAILED, 84 errors, 2 skipped (325 tests passed)
```

Even when we get all of the API tests to pass, this won't be end game. It's just a start and even with tests passing it's likely there are some edge cases which are not reflected in the tests.

Kami:
Encountered a bigger problem - `keyczar` library we use doesn't support Python 3.

It looks like there is some kind of fork with Python 3 support (https://pypi.python.org/pypi/python3-keyczar/0.71rc0), but since this is a sensitive library which handles cryptography, I just don't want to go around and install some random fork without a proper audit.

Kami:
@armab 

> @Kami Cool.
RE caching, let's please use CACHE_NAME=py2 and CACHE_NAME=py3 or similar instead of https://github.com/StackStorm/st2/blob/master/.travis.yml#L14 to avoid any possible conflicts between the 2 environments. Previous config was expecting shared caching env for all jobs.

I don't think there should be any conflicts because for tox everything is under a `.tox` subdirectory.

Having said that, I will still look a bit into having two different caches, but if it causes too many issues (e.g. build times out because old cache is no longer valid and we can't build a new one, etc.), I will leave it as is.

Kami:
I believe this should do it - d4d60836435ebe6aaaccf7e590baf88763a3830f.

I re-used the same name for Python 2 tasks so the cache doesn't need to be fully rebuilt and just added a new one for Python 3 tasks.

.travis.yml
```diff
@@ -11,7 +11,7 @@ branches:
 
 env:
   global:
-    - CACHE_NAME=JOB1
+   - CACHE_NAME=JOB1
```
armab: missing space

Kami:
@tonybaloney I reversed the string -> bytes change you did for two module level constants - 16cd9a4c351cfea615264519bda97c3db699cb9b.

In other places we expect strings so it makes no sense to make those two constants bytes everywhere. We should just convert it to bytes where we need bytes (aka in the base test class which checks for secrets leak).

Kami:
Related issue for switching to cryptography library - #3974.

That change requires more work and shouldn't be part of that PR. I'm close to finishing this PR, got most of the tests passing, just a couple of "hard ones" like the one related to keyczar are still failing.

tonybaloney:
Oops. sorry @Kami all the dist_utils.py files I removed from my changeset as I wasn't sure why they had changed.

Kami:
To sum it up - right now I have all tests but the following passing:

* test_kvps.py which relies on keyczar - no easy out
* execution output (two test cases inside test_executions.py, now I added `unittest.skipIf(six.PY3`)) - some weird issue with iterator closed on read, couldn't figure it out (yet)
* test_validator_mistral.py - issue with got bytes not string, also couldn't figure it out (yet)

Kami:
@tonybaloney Maybe you will have more luck with the last two^ :)

PR #3976, opened 2018-01-30
Allow Large Params via stdin to wrapper

### LINKED ISSUE #1598: Allow users to pass parameters which are large in size to Python actions

bigmstone:
This PR fixes #1598. Since Linux kernel specifies an argument character limit on CLI arguments there are instances where we exceed this limit - especially when passing parameters between actions. This PR uses `stdin` to communicate parameters if they exceed the imposed limit.


bigmstone:
As of 9555ff5 I hate the implementation.

Kami:
I like it :+1: 

It's much simpler than the alternative and has less security implication.

Do you think we should still support old approach of providing parameters and only use this approach when len > max arg size?

I think so, even if the only reason if backward compatibility on upgrade in 2.7.0

st2common/st2common/runners/python_action_wrapper.py
```diff
@@ -261,12 +264,26 @@ def _get_action_instance(self):
...
+
+    if args.parameters:
+        LOG.debug('Getting parameters from argument')
+        args_parameters = args.parameters
+        args_parameters = json.loads(args_parameters) if args_parameters else {}
+        parameters.update(args_parameters)
+
+    if args.stdin_parameters:
+        LOG.debug('Getting parameters from stdin')
+        stdin_parameters = json.loads(input())
```
Kami: I wonder if there is some kind of timeout we could use here in case input is not provided in some reasonable time frame.

Shouldn't be an issue for action runner actions, but when you are debugging and running things manually.

contrib/runners/python_runner/python_runner/python_runner.py
```diff
@@ -212,15 +226,19 @@ def run(self, action_parameters):
         command_string = list2cmdline(args)
         LOG.debug('Running command: PATH=%s PYTHONPATH=%s %s' % (env['PATH'], env['PYTHONPATH'],
```
Kami: If stdin is used we should ideally print the whole command user could use to replicate, so something like `echo -e "serialized stdin params\n" | python ....`

PR #3979, opened 2018-02-02
Add version string to st2api module

apolloliu:
It occurred when installing:
```
Obtaining file:///opt/stack/github/st2/st2api
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/opt/stack/github/st2/st2api/setup.py", line 23, in <module>
        from st2api import __version__
    ImportError: cannot import name __version__
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /opt/stack/github/st2/st2api/
```

### DIFF
--- a/st2api/setup.py
+++ b/st2api/setup.py
@@ -20,18 +20,19 @@ from setuptools import setup, find_packages
 
 from dist_utils import fetch_requirements
 from dist_utils import apply_vagrant_workaround
-from st2api import __version__
+from dist_utils import get_version_string
 
 ST2_COMPONENT = 'st2api'
 BASE_DIR = os.path.dirname(os.path.abspath(__file__))
 REQUIREMENTS_FILE = os.path.join(BASE_DIR, 'requirements.txt')
+INIT_FILE = os.path.join(BASE_DIR, 'st2api/__init__.py')
 
 install_reqs, dep_links = fetch_requirements(REQUIREMENTS_FILE)
 
 apply_vagrant_workaround()
 setup(
     name=ST2_COMPONENT,
-    version=__version__,
+    version=get_version_string(INIT_FILE),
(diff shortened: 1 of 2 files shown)


Kami:
Thank you for your contribution.

We actually automatically populate `__version__` for all the st2 components based on the `__version__` value in `st2common`.

This happens as part of our CI / CD release workflow. We don't manually add `__version__` string in every component to avoid duplication because all of the components use the same version (we always release all the components at once so they have the same version).

Having said that - I do see an issue with not having `__version__` always present - you can't directly install components using pip from git repository. Something which is not officially supported right now anyway, but I agree it would be nice if you could do that.

So in short - thank for your contribution, but to make this a reality we need to decide on the approach and also correctly update / modify our release workflows.

apolloliu:
@Kami  Thanks for the response!

I'm curious how I should install every component (e.g. st2api). I have to install every component from my personal repo which is forked from st2 if I rewrite some codes in some components. I would rather not see this issue each time I install using `pip` or `python install`.

Kami:
@apolloliu Right now we officially only distribute `st2common` and `st2client` as Python packages.

Other components are distributed as part of our debian / rpm packages. They are also Python packages, but right now they are not meant to be used directly from git.

If you need access to Python packages for all components there are two ways to do it right now:

1. As you have done - fork the repo or similar and populate `__version__` for all the components
2. Download debian / rpm package, unpack them and you st2 components from there (those have __version__ populated)

Going forward - we could / should update our workflows so __version__ is already populated so people can install it directly from git if they wish so.

apolloliu:
@Kami Thank you pointing the two ways! It's really cool if updating the workflow of populating `version` because some developers want to install directly from git as soon as they modify the codes.

PR #3980, opened 2018-02-02
add utf8  supporting  for config file(yaml)

sumkire:
add utf8  supporting  for config file(yaml)


sumkire:
Already added.

st2client/tests/unit/test_config_parser.py
```diff
@@ -79,3 +81,12 @@ def test_parse(self):
...
         result = parser.parse()
         self.assertTrue(result['cli']['cache_token'], True)
+
+    def test_get_config_for_unicode_char(self):
+        parser = CLIConfigParser(config_file_path=CONFIG_FILE_PATH_UNICODE,
+                                 validate_config_exists=False)
+        config = parser.parse()
+        print parser.config_file_path
+        print config
+        self.assertEqual(config['credentials']['password'], u'测试')
```
cognifloyd: It looks like this `测试` doesn't match what's in st2client/tests/fixtures/test_unicode.ini: `密码`

Though I like your use of "test" and "password" (translation from google translate) in the tests. :)

Kami:
Fixed the test case in #3989 and merged it into master.

Thanks!

PR #3981, opened 2018-02-02
Fix cleanup db in launchdev

m4dcoder:
Fix relative virtualenv, python, and db cleanup file paths. Activate the virtualenv to execute the cleanup DB script otherwise the script will complain about non-existent module st2common.persistence.cleanup.

### DIFF
--- a/tools/launchdev.sh
+++ b/tools/launchdev.sh
@@ -59,6 +59,11 @@ function init(){
         ST2_REPO=${CURRENT_DIR}/${COMMAND_PATH}/..
     fi
 
+    VENV=${ST2_REPO}/virtualenv
+    PY=${VENV}/bin/python
+    echo "Using virtualenv: ${VENV}"
+    echo "Using python: ${PY}"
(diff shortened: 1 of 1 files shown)


tools/launchdev.sh
```diff
@@ -399,10 +404,10 @@ function st2stop(){
 
 function st2clean(){
     # clean mongo
-    ./virtualenv/bin/python \
-        ./st2common/bin/st2-cleanup-db \
-        --config-file $ST2_CONF
-    
+    . ${VENV}/bin/activate
+    python ${ST2_REPO}/st2common/bin/st2-cleanup-db --config-file $ST2_CONF
```
Kami: I'm a bit confused about this change - in the end, both approaches should have the same end result, but anyway.

m4dcoder: If I don't explicitly activate the python, the st2-cleanup-db script is complaining about missing some st2common modules...

PR #3988, opened 2018-02-06
Add support for advanced filters

enykeev:
screen shot 2018-02-06 at 12 44 04


st2api/st2api/controllers/resource.py
```diff
@@ -182,6 +182,16 @@ def _get_all(self, exclude_fields=None, include_fields=None,
             else:
                 filters['__'.join(v.split('.'))] = filter_value
 
+        if advanced_filters:
```
Kami: Can we please also add some test cases for this.

st2common/st2common/openapi.yaml
```diff
@@ -977,6 +977,11 @@ paths:
           description: Show secrets in plain text
           type: boolean
       x-parameters:
+        - name: filter
+          in: query
+          x-as: advanced_filters
```
Kami: Could this potentially have negative security implications?

Like injections or ability to retrieve data which shouldn't be allowed. Do we need a whitelist of keys and operators like we have for regular filters?

st2common/st2common/openapi.yaml.j2
```diff
@@ -973,6 +973,11 @@ paths:
           description: Show secrets in plain text
           type: boolean
       x-parameters:
+        - name: filter
+          in: query
+          x-as: advanced_filters
+          description: Advanced filter string. Overrides other filters.
```
Kami: Also, any particular reason this is a string not an object / array?

st2api/st2api/controllers/resource.py
enykeev: we absolutely will

st2common/st2common/openapi.yaml
enykeev: I could not think of anything that might be a security issue for execution document. An intruder may want to try and brute force the secrets, but I would not consider it a massive threat at the moment.

If you have a whitelist in mind, lmk.

We could also hide this feature behind rbac, although I still not believe it's dangerous enough.

st2common/st2common/openapi.yaml.j2
enykeev: Because it is much harder to express complex query logic in array or an object. Both require use of structural characters such as `{`, `[` or `,` that only create clutter and don't add anything useful to the query.

Because it being a GET request requires us to also make `filter` a query parameter which by itself is a string.

Because it helps us to keep all the logic in one place (in st2api) and avoid additional changes if we decide to change the syntax (and we may).

st2api/tests/unit/controllers/v1/test_executions_filters.py
```diff
@@ -194,6 +194,22 @@ def test_filters(self):
...
+                continue
+
+            value = self.fake_types[0]
+            for item in field.split('.'):
+                value = value[item]
+            response = self.app.get('/v1/executions?filter=%s:%s' % (field, value))
+            self.assertEqual(response.status_int, 200)
+            self.assertIsInstance(response.json, list)
+            self.assertGreater(len(response.json), 0)
+            self.assertGreater(int(response.headers['X-Total-Count']), 0)
```
m4dcoder: 1) Can you add another unit test that show the exception thrown if there are more than max split in the filter value?  For example `/v1/executions?filter=k1:v1,k2:v2`.
2) Also, is multiple filters supported?  For example `/v1/executions?filter=k1:v1&filter=k2:v2`

enykeev: Multiple filters are not supported. There's no way to express anything but OR logic this way and we've already been bitten by that in criteria so I don't want to make the same mistake again. Filter will always going to be a string user composes (semi-)manually so it makes sense to send it to API as a single string.

I'm going to make sure we output a readable exception for 1.

enykeev: Actually, I could not find a proper way to handle `/v1/executions?filter=k1:v1,k2:v2` case at the moment. My gut feeling tells me that we should treat `,` same as space (a delimiter), but at that point I simply don't know.

I've made advanced filter field less noticeable in UI and I'm think we should release it as is to get some early feedback and see where to go next.

Kami:
Going back to #4035 - does this also solve problem described in that issue?

Aka allowing user to specify a set of values and then performing filtering on those values using "OR" approach.

It looks like it doesn't, but it would be good to confirm.

And, yeah, if we add support for OR, then it would also make sense to think about AND. And for that, we probably need different syntax...

st2common/st2common/openapi.yaml.j2
Kami: I'm fine in this specific case, but I would argue that the input type doesn't matter that much.

That's why we have openapi for - to also cast and convert values as needed (same goes for boolean arguments, etc.)

st2api/tests/unit/controllers/v1/test_executions_filters.py
m4dcoder: If a user calls something like this /v1/executions?filter=k1:v1,k2:v2, then do we throw errors? unit test?

enykeev:
@Kami currently, it only supports AND filtering over every field on the document. This is most likely isn't enough on the long run and we have a solution for that in form of [pql](https://github.com/alonho/pql), but I see a value of shipping it earlier and iterate.

Kami:
I think we are all on similar both here - I'm fine with this change, for now, but eventually I would like us to better and more consistently implement better and more powerful filtering across the API.

Issue #3870, opened 2017-11-20
No Downtime Deployment of Packs


nmaludy:
## Background

One of the goals i have for our StackStorm deployment is to implement Continuous Delivery (CD) of our custom packs without having to schedule downtime of our StackStorm servers.

## Requirements

* Upgrades can be performed without taking downtime
* No race conditions exist that could cause errors of ongoing workings during a deployment
* External systems, that may be polling for status/completion of ongoing workflows, should not be affected
* Packs may need to deploy/update virtualenvs
* Pack configs may need to be loaded
* Key/value pairs may need to be added to the datastore
* Custom content may need to be installed on the box (deploy additional non-pack content)
* Custom scripts may need to be run on the box (install system level dependencies)

## Conclusion

This issue is to simply start the conversation and brainstorm ideas of how to implement such a feature/process.

cognifloyd:
I have one pack that depends on a package that introduces a dependency on pandas... Building that virtualenv takes forever. It would be marvelous to be able to update that pack and virtualenv without downtime.

I would guess that virtualenvs could be rebuilt by building them in one folder, and then moving the virtualenv into place once built. Or maybe, copy the existing virtualenv to a new folder, update it, then move it into place (moving the old out of the way at the same time). Perhaps something similar could be done for pack install.

Using https://github.com/EncoreTechnologies/ansible-stackstorm to install packs, it uninstalls and reinstalls the pack because there's no good way to idempotently install/update a pack. That will probably have to be addressed to make this work. This is issue #3052, but probably includes resolving these issues as well: #3191, #3313, #924, #3143, #3016, #3749, #3660, #3585, #3336

cognifloyd:
Or, if building a virtualenv in an alternate folder causes issues... ST2 could start using a chroot or a FS overlay (I'm thinking some container-like thing that overlays on top of the FS - I think systemd can do that on the fly iirc) for building virtualenvs and installing packs, only swapping them into place once complete.

nmaludy:
## In-place Deployments

From my previous experiences i believe this deployment strategy will be the hardest to get right technically. It will involve a lot of "transactional" type operations which are non-trivial to implement especially on the system level.

## Blue/Green Deployments

I think this has the best chance of being a workable solution. Basically, front the StackStorm instance with a load-balancer, stand up another StackStorm instance with the new code deployed and then flip a load-balancer over to the new instance.

[...]

PR #3997, opened 2018-02-13
Ability to specify a version of a pack resource (action) to use when running an action execution

Kami:
## Description

This pull request starts an implementation which allows user to specify a version (either a git revision / branch / tag) of a resource to use for a particular action execution.

Ability to use a specific version of a pack resource comes handy in many scenarios. Notably when running a distributed deployment and you want to ensure a specific version of pack resource is used.

That is something @enykeev and I and others have discussed in the past.

The code is far from finished and I just opened the PR so we can get everyone on the same page and agree on the implementation.

## Usage

User can specify which version of pack content to use by specifying `content_version` global runner parameter.

For example: `st2 run mypack.myaction arg=value content_version=v0.5.0`

## Limitations

1. Only packs which are git repositories are supported. That's expected because implementation heavily leverages git primitives and all the official StackStorm packs are git repositories.

Users can still use packs which are not git repositories, it simply means they can't leverage this functionality.

2. We only ensure a specific version of pack content (file from a pack git repository) is used. We don't handle any versioning of virtualenvs and pack dependencies for Python runner packs.

This would be simply too complex and add a lot of overhead for a little value.

Imo, the right way to handle that is containers. Virtualenvs are complex and Python packages can also depend on C libraries and system dependencies so just versioning virtualenv directories wouldn't solve the whole problem.

## Implementation

The implementation leverages `git worktree` functionality. This functionality was suggested by @enykeev and I agree it's a good idea.

git worktree simply makes a bare copy of a git repository directory and sets HEAD to the specified revision.

Before the action is executed a new git worktree directory is created for a specified revision and this directory is used when executing an action.

This directory is ephemeral (specific to an execution) and it's removed once the execution completes.

If we assume that git resources are immutable (that's indeed the case for commit revisions, but not for branches and tags) we could use a fixed directory for each revision (content_version). This way we would avoid directory churn since different executions with same `content_version` could re-use the same git worktree directory.

This approach would also be more efficient and faster because creating a new worktree directory for each action execution adds some overhead.

If we went with this approach we would probably still need some kind of garbage collection service which deletes pack git worktree directories which haven't been accessed for X days or similar.

## TODO

- [x] Decide on all the open questions / implementation
- [ ] Documentation
- [x] Finish code implementation


nmaludy:
Will help with #3870

cognifloyd:
You mention using containers instead of virtualenvs... I stopped using docker because it was too complex to duplicate my esoteric network setup from host into the container. Some of my actions need not only system packages, but system network setup (vpn, ssh sessions with local forwarded ports as a system service, access to listen to or use the broadcast address of a particular host link, access to particular usb hardware, ... I'm not using all of that in ST2 right now, but I would like to).

Kami:
@cognifloyd That's a valid point.

What I really just wanted to convey with the container possibility as a solution is that it's a complex problem and no solution is 100% and we are just trying to solve the "version of the code which is used for the execution" problem.

@nmaludy What do you think, will this help enough with #3870?

Kami:
Talked with @enykeev about the approach on Slack.

We decided to go with the "worktree per execution" approach for now. This way we get the most isolation and we simply don't have enough hard data to make a decision and everything else would be speculation.

If it turns out that it's too slow at some point in the future, we can change it and go with the worktree per revision / version approach (where the worktree for a particular version is shared among different executions).

CHANGELOG.rst
```diff
@@ -13,6 +13,17 @@ Added
 * Add support for utf-8 / unicode characters in the pack config files. (improvement) #3980 #3989
 
   Contributed by @sumkire.
+* Add support for new optional ``content_version`` runner parameter to the Python and Local Shell
+  Script runner. This parameter can contain a git commit hash / tag / branch from a pack git
+  repository and runner will ensure this revision of the pack content (Python action / local shell
+  script action) is used for a particular action execution.
```
m4dcoder: How come only python and local shell runners are supported?  From what i review of GitWorktreeActionRunner, it's overriding the entry point. So shouldn't this work for all actions with an entry point?  Please explain why this wouldn't work for other runners.

Kami: The feature makes only makes sense / applies to runners which use local script files (entry point) from disk. It allows users to use previous version of that entry_point script file based on the value of `content_version` value.

Those are mainly Python and local script runner.

It makes no sense for runners such as local command runner and remote command runner, http runner, etc. because in those cases there is no local script file and all the commands / parameters are provided during invocation.

m4dcoder: How about action chain and mistral? They have a WF definition file on disk.

Kami: Yeah, I was thinking about Mistral, but I believe that would also require work on the Mistral side - to make it use a different directory with workflow definition when `content_version` parameter is specified.

m4dcoder: It shouldn't require a lot of work since you already got this figured out. The WF definition for both Action Chain and Mistral are specified in the entry point.  Should just make those runners inherit from GitWorktreeActionRunner?  Need to write unit tests for them.

m4dcoder:
@Kami can you also add unit test(s) where the python action uses a python module local to the pack?  We have plenty of packs like this. If you change the location of the entry point, will it know to load the correct module as well?

Kami:
@m4dcoder Good point, will add it.

And yeah, in theory it should work as fine, because current working directory from where the entry_point / script is executed is added to PYTHONPATH implicitly and copy / version control the whole pack directory so it should just work out of the box.

Kami:
@m4dcoder Test case added in 4c02af35399fcb2bd4cc47018380952e6fc75afd.

As mentioned above, it indeed works correctly because the directory where the script which is executed is located gets automatically and implicitly added to `PYTHONPATH` (well, it doesn't really get added to PYTHONPATH, but Python import resolution code takes that into account).

The thing did get me thinking about other scenarios and I caught that - 16b88dd6e5c20a78758845104c1356f20c2a510b.

Will add test case for that shortly.

Kami:
Test case for pack commons libs path in 7d2818a2794f24445c681742b828354b0e22049c.

Kami:
As far as support for action chain and Mistral goes - I do agree that eventually it would be good to have support for that as well, but initial goal was only support for Python runner.

Only reason I also added support for local shell script runner is because a lot of code can be re-used between Python and Local shell script runner and the change was relatively straight forward. And not doing it for only a single runner resulted in a better code which is not "biased" against Python runner (better abstraction and separation of concerns).

The change for Mistral is more involved so I would like to postpone for the future.

Kami:
And to expand on that and to clarify further - I think that's the right model for releasing / shipping pretty much any kind of bigger "experimental" feature - in a small incremental manner.

Easier to change / adopt / etc things based on the feedback, compared to just throwing out one big change (agile! :joy:).

Having said that, in cases like that, it's also good that the initial implementation contains implementation for more than one runner to make sure it's not biased against a single runner which results in a better abstraction, etc.

Kami:
@lakshmi-kannan @bigmstone ^can you two please also have a look.

@m4dcoder and I have different ideas on how to proceed with this :)

enykeev:
I feel like the team is a little bit overzealous about using "request for change" without actualy mentioning a change they want to be made.

@bigmstone if you look closer at the code, you'll see that the only purpose of the worker is to dispatch a container with the liveaction and the actual work of reading the entrypoint code happening in the runner itself. Also, it seems to me that the runner knows better how to handle the specifics of its own execution. In case of python runner, for example, it should be able to also handle switching venv, a function the rest of the runners should not care about.

@m4dcoder If we're agreeing that it is runners responsibility to handle its own code versioning, then we should be able to do it slow and implement it runner by runner. I see no reason to block this PR simply on a basis that it didn't cover all the possible scenario. Following your logic, I don't see why cloudslang runner should be exempt. Do anyone of the team feel confident of implementing this change there?

bigmstone:
@enykeev if we're checking out a specific version of an `entry_point` in a repo then I'm not sure why which runner is used matters. It's all under `entry_point` no matter what the runner. As for handling corner cases like virtual environment that I have no problem being placed inside the runner itself. Am I missing something obvious here?

enykeev:
Ok, then I'm asking @StackStorm/team a counter-question: can someone explain to me in simple words the distinction between actionrunner worker, container and runner?

Kami:
@bigmstone Good question and @enykeev good explanation.

That's also why I said we should "formally" document each service and component responsibility - a lot of those decisions were made early on in the code, but there are not formally documented (in a code comment / docs / similar). Sometimes it's hard when building out new features, because you can always argue that X belongs in Y where Y can be pretty much any component.

In short, what @enykeev has said.

Entry point and other manipulation happens inside the runner class. Another reason is that some of the code is runner specific (e.g. making sure common libs path which only applies to Python runner correctly resolves to git worktree directory, etc.).

So while I could put the code in the action runner container (which I believe is what you mean when you said action runner), but I think it would result in a bad abstraction.

We would need to do `if runner == foo` and similar which I think is a code smell / anti pattern and usually means you need to do something differently (e.g. use inheritance or similar).

bigmstone:
Yes, container is where I envisioned this going. In hindsight making the distinction of `st2actionrunner` doesn't make any since sense the action runner, container, and runner all execute in that process.

I don't want to get to a point where we're branching to check runners, but checking out a version of `entrypoint` in the contain and doing runner specific stuff in the runner seems reasonable to me. I'm not going to die on that hill though. If we want to replicate calling the checkout code in each runner it's not a blocker to me - even though I think that's better suited in the container.

bigmstone:
@cognifloyd Agree - current implementation does this by having runner inherit from new class that can handle this. So net-new runners could just start by inheriting from this class form the onset.

Kami:
@bigmstone @m4dcoder I'm still missing "formal" approval of this PR so I can finally merge it into master :)

PR #3999, opened 2018-02-14
Add support for multiple action runners (runner modules) inside a single runner Python package

Kami:
This pull request adds support for multiple runners and runner modules inside a single action runner Python package.

We already partially introduced support for that in the past during runner refactor done by @bigmstone by allowing "runner.yaml" metadata file to be an array and contain multiple entries (one per runner).

The problem is that support was not fully implemented so we needed to resort to handling multiple runners inside a single runner module inside the code using if checks. That's not ideal and ability to utilize inheritance is better and results in a cleaner code.

I already partially introduced support for that when working on serverless stuff and making runners re-distributable and standalone Python packages. This pull request finishes what I started there.

I originally wanted to postpone this, but I need this code change to be able to continue work on #3997 (one of the changes I'm working in there only applies to local shell script, but not local shell command runner and inheritance makes code a lot more cleaner).

In addition to setting up ground work for multiple runner modules per Python package it also moves local and windows runners to that model.

Another benefit this change brings is that now we can put common code shared by local or remote runners in that runner package. Previously it needed to live inside `st2common` which was far from ideal.

## TODO

- [x] Move local runners to the new model
- [x] Move windows runners to the new model
- [x] Move remote runners to the new model
- [x] Remove now unneeded backward compatibility code from runner package `__init__.py` files


contrib/runners/local_runner/local_runner/__init__.py
```diff
@@ -19,4 +19,4 @@
 # This means you can either do "from runner_name import RunnerClass" (old way, don't do that)
 # or "from runner_name.runner_name import RunnerClass"
 from __future__ import absolute_import
-from .local_runner import *  # noqa
```
Kami: I will actually get rid of this backward compatibility stuff. It will make things nicer and we don't need it anymore, AFAIK.

Kami:
Just a heads up - once this PR is merged, you will need to re-run `make virtualenv` make target locally to re-create `virtualenv/bin/activat`e file with updated `PYTHONPATH` and then run `source virtualenv/bin/activate`.

PR #4000, opened 2018-02-15
Refactor mistral runner to support callback

m4dcoder:
Refactor mistral runner to support callback from mistral instead of relying on st2resultstracker. This reduces the unnecessary traffic and CPU time by querying the mistral API. Add a new PollingAsyncActionRunner for runner that need to add results tracking automatically. Mistral runner will not create a state entry anymore by default. Add a command to manually add a state entry for Mistral workflow execution to recover from any callback failures. Unit and integration tests are refactored accordingly.

### DIFF
--- a/contrib/examples/actions/workflows/tests/mistral-test-pause-resume.yaml
+++ b/contrib/examples/actions/workflows/tests/mistral-test-pause-resume.yaml
@@ -19,4 +19,4 @@ examples.mistral-test-pause-resume:
         task2:
             action: core.local
             input:
-                cmd: echo "<% $.var1 %>" 
+                cmd: 'sleep 10; echo "<% $.var1 %>"' 
--- a/contrib/examples/actions/workflows/tests/mistral-test-pause-resume.yaml
+++ b/contrib/examples/actions/workflows/tests/mistral-test-pause-resume.yaml
@@ -19,4 +19,4 @@ examples.mistral-test-pause-resume:
         task2:
             action: core.local
             input:
-                cmd: 'sleep 10; echo "<% $.var1 %>"' 
+                cmd: 'echo "<% $.var1 %>"' 
--- a/st2api/st2api/controllers/v1/actionexecutions.py
+++ b/st2api/st2api/controllers/v1/actionexecutions.py
@@ -606,16 +606,29 @@ class ActionExecutionsController(ActionExecutionsControllerMixin, ResourceContro
         if liveaction_db.status in action_constants.LIVEACTION_COMPLETED_STATES:
             abort(http_client.BAD_REQUEST, 'Execution is already in completed state.')
 
-        if (getattr(liveaction_api, 'result', None) is not None and
-                liveaction_api.status in [
-                    action_constants.LIVEACTION_STATUS_PAUSING,
-                    action_constants.LIVEACTION_STATUS_PAUSED,
-                    action_constants.LIVEACTION_STATUS_RESUMING]):
-            abort(http_client.BAD_REQUEST,
-                  'The result is not applicable for pausing and resuming execution.')
+        def update_status(liveaction_api, liveaction_db):
+            status = liveaction_api.status
+            result = getattr(liveaction_api, 'result', None)
+            liveaction_db = action_service.update_status(liveaction_db, status, result)
+            actionexecution_db = ActionExecution.get(liveaction__id=str(liveaction_db.id))
+            return (liveaction_db, actionexecution_db)
 
         try:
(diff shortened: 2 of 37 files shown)


nmaludy:
+1000000

cognifloyd:
The other half of this PR: StackStorm/st2mistral#35

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -606,37 +606,45 @@ def put(self, id, liveaction_api, requester_user, show_secrets=False):
...
-                actionexecution_db = ActionExecution.get(liveaction__id=str(liveaction_db.id))
+                liveaction_db, actionexecution_db = update_status(liveaction_api, liveaction_db)
         except runner_exc.InvalidActionRunnerOperationError as e:
             LOG.exception('Failed updating liveaction %s. %s', liveaction_db.id, str(e))
             abort(http_client.BAD_REQUEST, 'Failed updating execution. %s' % str(e))
         except runner_exc.UnexpectedActionExecutionStatusError as e:
             LOG.exception('Failed updating liveaction %s. %s', liveaction_db.id, str(e))
             abort(http_client.BAD_REQUEST, 'Failed updating execution. %s' % str(e))
         except Exception as e:
+            print str(e)
```
cognifloyd: WIP print statement?

m4dcoder: Yeah, i will clean that up

Kami:
How does this affect the deployment? Do we need to update the docs?

Since we are going from "we periodically poll mistral" (pull) vs "Mistral sends results on completion (push) it would be good to update the docs to clarify that. At the very least, it could affect firewall rules and similar (now Mistral needs to have outgoing access to StackStorm instance).

And in theory this also changes the failure and HA model since now st2api also needs to be deployed in HA manner (ideally multiple instances behind an HTTP load balancer) if HA is required / expected.

This was already a good idea before this change, but now it makes it even more important for HA deployments which also rely on / use Mistral.

Kami:
Eventually - for quality of service reasons, I think it would also make sense to have a separate HTTP service which exposes HTTP API to which Mistral posts results to.

Basically similar as we do with st2stream. The reason for that is general API and the one Mistral uses for sending back results have quite different usage patterns - general API tends to be more read heavy and Mistral uses it in more write heavy manner which could affect performance and QoS (e.g. regular API operations block Mistral or vice versa).

Kami:
Another question - how does Mistral authenticate against StackStorm API, API key?

If so, I guess this means another thing which potentially needs to be configured and documented...

Kami:
Also, from Slack - I'm slightly worried this could have negative QoS implications in some scenarios (depends on how users use the API and how many concurrent Mistral workflows they run) so I personally think it would be a good idea to keep the old result tracker way of doing polling there and allow user to select which method to use in the config (poll vs callback).

This way, once we ship this feature and if it turns out that it negatively affects some user deployments, it easy for user to switch back to the previous approach without needing to roll back to previous StackStorm version (think of it as a feature flag in a way).

m4dcoder:
@Kami Ok, I added a switch in the config (default to use callback). Please review again. Our CI will fail until our Mistral fork is merged with upstream.

nmaludy:
What about the scenario when a mistral task never leaves the "running" state and has to be killed manually by doing `mistral action-execution-update -s CANCELLED xxx` ?

I'm guessing that in this scenario the notifier will never send anything back to StackStorm to let it know the workflow was cancelled?

contrib/examples/actions/workflows/tests/mistral-test-pause-resume.yaml
```diff
@@ -19,4 +19,4 @@ examples.mistral-test-pause-resume:
         task2:
             action: core.local
             input:
-                cmd: echo "<% $.var1 %>" 
+                cmd: 'sleep 10; echo "<% $.var1 %>"'
```
bigmstone: Why's this needed now?

st2api/st2api/controllers/v1/actionexecutions.py
bigmstone: Bump

m4dcoder:
@bigmstone fixed the print statement and removed the sleep.  if i forgot to remove them after troubleshooting.

m4dcoder:
> What about the scenario when a mistral task never leaves the "running" state and has to be killed manually by doing mistral action-execution-update -s CANCELLED xxx ? I'm guessing that in this scenario the notifier will never send anything back to StackStorm to let it know the workflow was cancelled?

@nmaludy In mistral, when an action execution is updated and if it is associated with a task, it'll trigger the task handler to process the notification.

PR #4004, opened 2018-02-16
ST2 Metrics Collection

### LINKED ISSUE #2974: Instrument python runner to get more metrics around delayed executions
Simple Python actions take a few seconds to complete execution.
```
(virtualenv)vagrant@st2dev /m/s/s/st2 ❯❯❯ st2 run examples.isprime value=7                                  STORM-2935/st2_timers_list ✭ ✱ ◼
..
id: 580fb666d9d7ed14ab36f33f
status: succeeded
parameters:
  value: 7
```

```
(virtualenv)vagrant@st2dev /m/s/s/st2 ❯❯❯ st2 execution get 580fb666d9d7ed14ab36f33f -dj                    STORM-2935/st2_timers_list ✭ ✱ ◼
{
    "action": {
        "ref": "examples.isprime"
    },
    "context": {
        "user": "stanley"
    },
    "end_timestamp": "2016-10-25T19:45:44.263915Z",
    "id": "580fb666d9d7ed14ab36f33f",
    "liveaction": {
        "action": "examples.isprime",
        "action_is_workflow": false,
        "callback": {},
        "id": "580fb666d9d7ed14ab36f33e",
        "parameters": {
            "value": 7
        },
        "runner_info": {
            "hostname": "st2dev.stackstorm.net",
            "pid": 5224
        }
    },
    "parameters": {
        "value": 7
    },
     "start_timestamp": "2016-10-25T19:45:42.049665Z",
    "status": "succeeded"
}

It looks like the fork call takes a couple of seconds but I benchmarked the fork call and it doesn't appear that fork is expensive from a standalone tester (https://gist.github.com/lakshmi-kannan/eb03a3f92d72b6f391a2f18fc8211883).

### PR DESCRIPTION
**Pull request by bigmstone**
Fixes #2974

- [ ] Changelog Entry
- [ ] Tests
- [ ] Example Config
- [ ] Documentation Update
- [ ] Prometheus Exploration (Or abandoning)

### DIFF
--- a/contrib/runners/python_runner/python_runner/python_runner.py
+++ b/contrib/runners/python_runner/python_runner/python_runner.py
@@ -159,6 +159,10 @@ class PythonRunner(ActionRunner):
             # We only pass --log-level parameter if non default log level value is specified
             args.append('--log-level=%s' % (self._log_level))
 
+        if cfg.CONF.metrics.enable:
+            LOG.debug("Enabling metrics for wrapper")
+            args.append('--metrics')
+
         # We need to ensure all the st2 dependencies are also available to the
         # subprocess
         LOG.debug('Setting env.')
--- a/st2common/st2common/config.py
+++ b/st2common/st2common/config.py
@@ -381,6 +381,19 @@ def register_opts(ignore_errors=False):
     cli_opts = [debug, profile, use_debugger]
     do_register_cli_opts(cli_opts, ignore_errors=ignore_errors)
 
+    # Metrics Options stream options
+    metrics_opts = [
+        cfg.BoolOpt('enable', default=False,
+                    help='Enable metrics collection.'),
+        cfg.StrOpt('driver', default='statsd',
+                   help='Driver type for metrics collection.'),
+        cfg.StrOpt('host', default='127.0.0.1',
+                   help='Destination server to connect to if driver requires connection.'),
+        cfg.IntOpt('port', default=8125,
+                   help='Destination port to connect to if driver requires connection.'),
+    ]
+    do_register_opts(metrics_opts, group='metrics', ignore_errors=ignore_errors)
+
 
 def parse_args(args=None):
     register_opts()
--- a/st2common/st2common/metrics/drivers/prometheus.py
+++ b/st2common/st2common/metrics/drivers/prometheus.py
@@ -1,0 +1,30 @@
+"""
+"""
+from prometheus_client import Histogram, Counter
+
+from st2common.metrics.metrics import BaseMetricsDriver
+
+
+class PrometheusDriver(BaseMetricsDriver):
+    """ Base class for driver implementations for metric collection
+    """
+    def __init__(self):
+        pass
+
+    def time(self, key, time):
+        """ Timer metric
+        """
+        prometheus_histogram = Histogram(key)
+        prometheus_histogram.observe(time)
+
+    def inc_counter(self, key, amount=1):
+        """ Increment counter
+        """
+        prometheus_counter = Counter(key)
+        prometheus_counter.inc(amount)
+
+    def dec_counter(self, key, amount=1):
+        """ Decrement metric
+        """
+        prometheus_counter = Counter(key)
+        prometheus_counter.dec(amount)
--- a/st2common/st2common/metrics/metrics.py
+++ b/st2common/st2common/metrics/metrics.py
@@ -1,0 +1,41 @@
+"""
+"""
+from oslo_config import cfg
+
+
+class BaseMetricsDriver(object):
+    """ Base class for driver implementations for metric collection
+    """
+    def time(self, key, time):
+        """ Timer metric
+        """
+        pass
+
+    def inc_counter(self, key, amount=1):
+        """ Increment counter
+        """
+        pass
+
+    def dec_counter(self, key, amount=1):
+        """ Decrement metric
+        """
+        pass
+
+
+def _get_metrics_driver():
+    driver = cfg.CONF.metrics.driver
+
+    if driver == "prometheus":
+        from st2common.metrics.drivers.prometheus import PrometheusDriver
+        driver_instance = PrometheusDriver()
+    elif driver == "statsd":
+        from st2common.metrics.drivers.statsd_driver import StatsdDriver
+        driver_instance = StatsdDriver()
+    else:
+        driver_instance = None
+
+    return driver_instance
+
+
+if cfg.CONF.metrics.enable:
+    METRICS = _get_metrics_driver()
--- a/fixed-requirements.txt
+++ b/fixed-requirements.txt
@@ -42,3 +42,5 @@ routes==2.4.1
 flex==6.11.1
 webob==1.7.3
 prance==0.6.1
+prometheus_client==0.1.1
+statsd==3.2.2
--- a/requirements.txt
+++ b/requirements.txt
@@ -26,6 +26,7 @@ oslo.utils<=3.33.0,>=3.15.0
 paramiko==2.3.1
 passlib==1.7.1
 prettytable
+prometheus_client==0.1.1
 prompt-toolkit==1.0.15
 psutil
 pyinotify==0.9.6
@@ -44,6 +45,7 @@ routes==2.4.1
 semver==2.7.7
 six==1.10.0
 sseclient==0.0.18
+statsd==3.2.2
 stevedore==1.27.1
 tooz==1.57.4
 unittest2
--- a/st2common/in-requirements.txt
+++ b/st2common/in-requirements.txt
@@ -27,3 +27,5 @@ routes
 flex
 webob
 jsonpath-rw
+prometheus_client
+statsd
--- a/st2common/st2common/metrics/drivers/statsd_driver.py
+++ b/st2common/st2common/metrics/drivers/statsd_driver.py
@@ -1,0 +1,28 @@
+"""
+"""
+from oslo_config import cfg
+import statsd
+
+from st2common.metrics.metrics import BaseMetricsDriver
+
+
+class StatsdDriver(BaseMetricsDriver):
+    """ StatsD Implementation of the metrics driver
+    """
+    def __init__(self):
+        self._connection = statsd.StatsClient(cfg.CONF.metrics.host, cfg.CONF.metrics.port)
+
+    def time(self, key, time):
+        """ Timer metric
+        """
+        self._connection.timing(key, time)
+
+    def inc_counter(self, key, amount=1):
+        """ Increment counter
+        """
+        self._connection.incr(key, amount)
+
+    def dec_counter(self, key, amount=1):
+        """ Decrement metric
+        """
+        self._connection.decr(key, amount)
--- a/st2common/st2common/runners/python_action_wrapper.py
+++ b/st2common/st2common/runners/python_action_wrapper.py
@@ -122,7 +122,7 @@ class ActionService(object):
 
 class PythonActionWrapper(object):
     def __init__(self, pack, file_path, config=None, parameters=None, user=None, parent_args=None,
-                 log_level=PYTHON_RUNNER_DEFAULT_LOG_LEVEL):
+                 log_level=PYTHON_RUNNER_DEFAULT_LOG_LEVEL, metrics=False):
         """
         :param pack: Name of the pack this action belongs to.
         :type pack: ``str``
@@ -150,6 +150,7 @@ class PythonActionWrapper(object):
         self._user = user
         self._parent_args = parent_args or []
         self._log_level = log_level
+        self._metrics = None
 
         self._class_name = None
         self._logger = logging.getLogger('PythonActionWrapper')
@@ -162,12 +163,16 @@ class PythonActionWrapper(object):
 
         # Note: We can only set a default user value if one is not provided after parsing the
         # config
-        if not self._user:
+        if not self._user or metrics:
             # Note: We use late import to avoid performance overhead
             from oslo_config import cfg
+            from st2common.metrics.metrics import METRICS
+            self._metrics = METRICS
             self._user = cfg.CONF.system_user.user
 
     def run(self):
+        if self._metrics:
+            self._metrics.inc_counter("python_actions_in_progress")
         action = self._get_action_instance()
         output = action.run(**self._parameters)
 
@@ -211,6 +216,8 @@ class PythonActionWrapper(object):
         sys.stdout.write(print_output + '\n')
         sys.stdout.write(ACTION_OUTPUT_RESULT_DELIMITER)
         sys.stdout.flush()
+        if self._metrics:
+            self._metrics.dec_counter("python_actions_in_progress")
 
     def _get_action_instance(self):
         try:
@@ -258,6 +265,8 @@ if __name__ == '__main__':
                              ' JSON')
     parser.add_argument('--log-level', required=False, default=PYTHON_RUNNER_DEFAULT_LOG_LEVEL,
                         help='Log level for actions')
+    parser.add_argument('--metrics', required=False, action='store_true',
+                        help='Enable metrics')
     args = parser.parse_args()
 
     config = json.loads(args.config) if args.config else {}
@@ -274,6 +283,7 @@ if __name__ == '__main__':
                               parameters=parameters,
                               user=user,
                               parent_args=parent_args,
-                              log_level=log_level)
+                              log_level=log_level,
+                              metrics=args.metrics)
 
     obj.run()

### CODE REVIEW

**Comment by Kami** [comment]
Just some of my quick thoughts - if I worked on something like this, I would start in the following order:

1. Code instrumentation (make sure we have instrumentation in place and all the metrics we care about)
2. Expose those metrics via an HTTP endpoint - e.g. `/v1/health/metrics` or similar.
3. Third party tool integration (Prometheus or similar)

I think integration with 3rd party tools is important, but I think it's even more important to have a "generic" API endpoint which metrics which everyone can consume (this includes our WebUI, Prometheus or any other monitoring system a user might use).

On a related note, it would be good to identify / list somewhere metrics we should collect. Perhaps we can just keep a list in the description of this PR.

**Comment by bigmstone** [comment]
My original intent was to have an API endpoint where metrics were collected over a stream across our bus. When thinking through this that would require us to begin collecting that information somewhere - mongo isn't the right tool, postgres isn't the right tool. We'd need to rely on another service to begin collecting that information which is yet another dependency on st2. We can make this dependency optional but it also puts us into a position where we're opinionated on how to collect these metrics. This isn't desirable in environments where they're already collecting stats in a specific way.

One thing I'm still considering is having a new st2 service that will be able to aggregate the collection of these metrics. This is primarily to accommodate both streaming style collection (`statsd`) as well as scraping type collection (`prometheus`). We could potentially do running average/in-memory stats and provide that via a health API there, but it will be imporant to note to the end user that the data won't persist anywhere so if the service goes down for whatever reason the numbers won't be reliable. Also what's our HA story with this type of implementation? Consensus algo  running between them to constantly keep metrics in sync?

**Comment by enykeev** [comment]
I'm wondering if it is at all possible to expose this metrics some way outside of normal stackstorm communication channels (something along the lines of dtrace hooks) and leave gathering this metrics to external process.

File: st2common/st2common/metrics/metrics.py [hunk 1]
```diff
@@ -0,0 +1,165 @@
...
+        assert isinstance(key, str)
+        assert len(key) > 0
+        self.key = key
+        self._metrics = METRICS
+        self._start_time = None
+
+    def send_time(self, key=None):
+        """ Send current time from start time.
+        """
+        time_delta = datetime.now() - self._start_time
```
m4dcoder: How come this is not in UTC?

st2common/st2common/metrics/metrics.py
```diff
@@ -0,0 +1,165 @@
...
+    driver = cfg.CONF.metrics.driver
+
+    if driver == "prometheus" and cfg.CONF.metrics.enable:
+        from st2common.metrics.drivers.prometheus import PrometheusDriver
+        return PrometheusDriver()
+    elif driver == "statsd" and cfg.CONF.metrics.enable:
+        from st2common.metrics.drivers.statsd_driver import StatsdDriver
+        return StatsdDriver()
+
+    return BaseMetricsDriver()
```
m4dcoder: Can't we use stevedore to load the drivers?

st2common/st2common/util/loader.py
```diff
@@ -192,34 +192,42 @@ def register_plugin(plugin_base_class, plugin_abs_file_path):
...
-        RUNNER_MODULES_CACHE[package_name][module_name] = imp.load_source(load_name, module_path)
-    else:
-        LOG.info('Reusing runner module "%s" from cache.', module_path)
-
-    return RUNNER_MODULES_CACHE[package_name][module_name]
+    # This needs to be imported here to avoid other processes/runners that rely on this
+    # file but don't have oslo context.
+    from st2common.metrics.metrics import CounterWithTimer
+    from st2common.constants.metrics import METRICS_REGISTER_RUNNER
+    with CounterWithTimer(METRICS_REGISTER_RUNNER):
```
m4dcoder: Use a decorator?

st2common/st2common/constants/metrics.py
```diff
@@ -0,0 +1,22 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+METRICS_COUNTER_SUFFIX = "_counter"
+METRICS_TIMER_SUFFIX = "_timer"
+
+PYTHON_RUNNER_EXECUTION = "python_runner_execution"
+PYTHON_WRAPPER_EXECUTION = "python_wrapper_execution"
```
m4dcoder: Metrics for other runners?

contrib/runners/python_runner/python_runner/python_runner.py
```diff
@@ -114,138 +118,142 @@ def pre_run(self):
...
-        if self._enable_common_pack_libs:
-            try:
-                pack_common_libs_path = get_pack_common_libs_path_for_pack_ref(pack_ref=pack)
-            except Exception:
-                # There is no MongoDB connection available in Lambda and pack common lib
-                # functionality is not also mandatory for Lambda so we simply ignore those errors.
-                # Note: We should eventually refactor this code to make runner standalone and not
-                # depend on a db connection (as it was in the past) - this param should be passed
-                # to the runner by the action runner container
+        with CounterWithTimer(PYTHON_RUNNER_EXECUTION):
```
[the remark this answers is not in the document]
bigmstone: ACK, yes we can.

st2common/st2common/metrics/metrics.py
bigmstone: Good catch.

st2common/st2common/metrics/metrics.py
bigmstone: Perhaps, need to look into it.

st2common/st2common/util/loader.py
bigmstone: Ack

st2common/st2common/constants/metrics.py
bigmstone: Yes, I like the idea of doing generic runner metrics at the runner caller as you mentioned. Also need to incorporate the feedback from community and internal on what to test. This is progressing currently.

m4dcoder:
@bigmstone, What happens if there's failure during metrics collections or if there's failure with the statsd server?  Can you add some tests that shows errors will be handled gracefully and exception logged and won't affect execution of the action? Thanks

bigmstone:
@m4dcoder nothing. It's UDP so it serializes the packet and that is all. That said I should see if there's anything the package is set to raise and catch it/log it.

contrib/runners/python_runner/python_runner/python_runner.py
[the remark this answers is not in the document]
Kami: I'm personally not a big fan of decorators, at least when they are used in this manner (wrapping large blocks of code) - it adds another level of indentation and makes code harder to read (imo, cyclomatic complexity).

I'm fine when they are used to wrap function / method definitions. One way to reduce this level of indentation would be to move this code block in a utility method, then call that utility method inside the decorator. This way original code block is indented as it was before.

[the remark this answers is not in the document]
Kami: Edit: It looks like we are wrapping whole run method, in this case we should just add a decorator around the whole method :)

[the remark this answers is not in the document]
m4dcoder: @Kami LOL

contrib/runners/python_runner/python_runner/python_runner.py
```diff
@@ -119,7 +123,8 @@ def pre_run(self):
         if self._log_level == PYTHON_RUNNER_DEFAULT_LOG_LEVEL:
             self._log_level = cfg.CONF.actionrunner.python_runner_log_level
 
-    def run(self, action_parameters):
+    @CounterWithTimer(PYTHON_RUNNER_EXECUTION)
```
Kami: Thanks :+1:

st2common/setup.py
```diff
@@ -62,5 +62,10 @@
         'bin/st2-track-result',
         'bin/st2-validate-pack-config',
         'bin/st2-check-license'
-    ]
+    ],
+    entry_points={
+        'st2common.metrics.driver': [
+            'statsd = st2common.metrics.drivers.statsd_driver:StatsdDriver',
```
m4dcoder: Entry point for prometheus?  I assume you will add this once we finalize on the plugin implementation.

st2common/st2common/metrics/metrics.py
```diff
@@ -0,0 +1,162 @@
...
+        def wrapper(*args, **kw):
+            with self as counter_with_timer:
+                return func(*args, metrics_counter_with_timer=counter_with_timer, **kw)
+        return wrapper
+
+
+try:
+    METRICS = get_plugin_instance(PLUGIN_NAMESPACE, cfg.CONF.metrics.driver)()
+except (NoMatches, MultipleMatches, NoSuchOptError):
+    METRICS = BaseMetricsDriver()
```
m4dcoder: Seems wrong to put this block of code here. Maybe I don't understand so please clarify. The plugins depends on this module before the plugins are registered, so loading this module will always set METRICS to the BaseMetricsDriver.  Can we move the class definition for BaseMetricsDriver somewhere else and then have the statsd and prometheus plugins import that new file instead of this file?

bigmstone: AS we discussed in slack: This isn't used by the driver that imports base. It's loaded at runtime after the drivers have already be registered and the metrics constant will be loaded prior to the driver being registered and put into runtime memory.

st2common/setup.py
bigmstone: Yes, I need to do some more digging into the prometheus integration before I'm ready to register it my default.

st2common/st2common/metrics/metrics.py
m4dcoder: So if I do the following, METRICS is type of BaseMetricsDriver.  Although the plugins don't use this METRICS constant. The Timer, Counter, TimerWithCounter uses METRICS and so by default it's the BaseMetricsDriver. 

```
>>> from st2common.metrics import metrics
>>> metrics.METRICS
<st2common.metrics.metrics.BaseMetricsDriver object at 0x7fcfcf560bd0>
```

I tried to run st2common's setup.py to see if I can get METRICS to be loaded with the statsd driver. But I'm getting the following errors on setup.

```
Installed /home/vagrant/tmp/st2/virtualenv/lib/python2.7/site-packages/tooz-1.57.4-py2.7.egg
Searching for semver==2.7.7
Reading https://pypi.python.org/simple/semver/
Best match: semver 2.7.7
Downloading https://files.pythonhosted.org/packages/a3/30/dc388e483165014d03e34b9c4cd5e0d5d30794d396225d0cb78b59d86d95/semver-2.7.7.tar.gz#sha256=20ffbb50248a6141bb9eba907db0e47ee4a239ddb55fe0ada8696fc241495a9d
Processing semver-2.7.7.tar.gz
Writing /tmp/easy_install-VsR63w/semver-2.7.7/setup.cfg
Running semver-2.7.7/setup.py -q bdist_egg --dist-dir /tmp/easy_install-VsR63w/semver-2.7.7/egg-dist-tmp-GdRI5C
  File "setup.py", line 68, in <module>
    'statsd = st2common.metrics.drivers.statsd_driver:StatsdDriver',
  File "/usr/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/usr/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/setuptools/command/develop.py", line 34, in run
    self.install_for_development()
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/setuptools/command/develop.py", line 133, in install_for_development
    self.process_distribution(None, self.dist, not self.no_deps)
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/setuptools/command/easy_install.py", line 742, in process_distribution
    [requirement], self.local_index, self.easy_install
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 826, in resolve
    dist = best[req.key] = env.best_match(req, ws, installer)
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1092, in best_match
    return self.obtain(req, installer)
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1104, in obtain
    return installer(requirement)
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/setuptools/command/easy_install.py", line 665, in easy_install
    return self.install_item(spec, dist.location, tmpdir, deps)
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/setuptools/command/easy_install.py", line 695, in install_item
    dists = self.install_eggs(spec, download, tmpdir)
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/setuptools/command/easy_install.py", line 876, in install_eggs
    return self.build_and_install(setup_script, setup_base)
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/setuptools/command/easy_install.py", line 1115, in build_and_install
    self.run_setup(setup_script, setup_base, args)
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/setuptools/command/easy_install.py", line 1101, in run_setup
    run_setup(setup_script, args)
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/setuptools/sandbox.py", line 233, in run_setup
    with setup_context(setup_dir):
  File "/usr/lib/python2.7/contextlib.py", line 17, in __enter__
    return self.gen.next()
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/setuptools/sandbox.py", line 195, in setup_context
    yield
  File "/usr/lib/python2.7/contextlib.py", line 35, in __exit__
    self.gen.throw(type, value, traceback)
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/setuptools/sandbox.py", line 166, in save_modules
    saved_exc.resume()
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/setuptools/sandbox.py", line 141, in resume
    six.reraise(type, exc, self._tb)
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/setuptools/sandbox.py", line 154, in save_modules
    yield saved
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/setuptools/sandbox.py", line 194, in setup_context
    __import__('setuptools')
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/setuptools/__init__.py", line 11, in <module>
    from setuptools.extern.six.moves import filterfalse, map
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/setuptools/extern/__init__.py", line 1, in <module>
    from pkg_resources.extern import VendorImporter
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 71, in <module>
    __import__('pkg_resources.extern.packaging.requirements')
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/pkg_resources/extern/__init__.py", line 42, in load_module
    __import__(extant)
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/pkg_resources/_vendor/packaging/requirements.py", line 9, in <module>
    from pkg_resources.extern.pyparsing import stringStart, stringEnd, originalTextFor, ParseException
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/pkg_resources/extern/__init__.py", line 42, in load_module
    __import__(extant)
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/pkg_resources/_vendor/pyparsing.py", line 696, in <module>
    collections.MutableMapping.register(ParseResults)
  File "/home/vagrant/tmp/st2/virtualenv/lib/python2.7/abc.py", line 109, in register
    if issubclass(subclass, cls):
  File "/home/vagrant/tmp/st2/virtualenv/lib/python2.7/abc.py", line 180, in __subclasscheck__
    if issubclass(subclass, scls):
  File "/home/vagrant/tmp/st2/virtualenv/local/lib/python2.7/site-packages/typing.py", line 1249, in __subclasscheck__
    return super(GenericMeta, self).__subclasscheck__(cls)
TypeError: super() argument 1 must be type, not None
```

m4dcoder: So I try this differently. Using the get_available_plugins, the statsd driver is there and I get use get_plugin_instance to load the statsd driver. However, when I load the metrics module, the METRICS constant is assigned the BaseMetricsDriver. This looks wrong.  Isn't it suppose to be assigned the StatsdDriver?

```
>>> from st2common.util import loader
>>> loader.get_available_plugins('st2common.metrics.driver')
['statsd']
>>> loader.get_plugin_instance('st2common.metrics.driver', 'statsd')
<class 'st2common.metrics.drivers.statsd_driver.StatsdDriver'>
>>> from st2common.metrics import metrics
>>> metrics.METRICS
<st2common.metrics.metrics.BaseMetricsDriver object at 0x7fc7c2b84450>
```

st2common/st2common/metrics/drivers/prometheus.py
```diff
@@ -0,0 +1,48 @@
...
+class PrometheusDriver(BaseMetricsDriver):
+    """ Base class for driver implementations for metric collection
+    """
+    def __init__(self):
+        pass
+
+    def time(self, key, time):
+        """ Timer metric
+        """
+        prometheus_histogram = Histogram(  # pylint: disable=no-value-for-parameter
```
Kami: Just curious why that is needed - in theory, if prometheus_client is in PYTHONPATH, pylint should detect the properties (unless the code is doing some magic).

st2common/st2common/metrics/drivers/prometheus.py
```diff
@@ -0,0 +1,48 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
```
Kami: Maybe consistency in module naming? E.g. `statsd_driver` vs `prometheus`.

I assume you added `_driver` suffix because statsd name clashes 3rd party library. I'm fine going either way - adding `_driver` suffix to all the driver modules, or drop it everywhere and use `from __future__ import absolute_import` to avoid name clashes.

st2common/st2common/metrics/metrics.py
```diff
@@ -0,0 +1,162 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
```
Kami: Maybe call that module base for consistency with some other packages? In any case, not a blocker :)

st2common/st2common/metrics/metrics.py
```diff
@@ -0,0 +1,162 @@
...
+        self._start_time = None
+
+    def send_time(self, key=None):
+        """ Send current time from start time.
+        """
+        time_delta = get_datetime_utc_now() - self._start_time
+
+        if key:
+            assert isinstance(key, str)
+            self._metrics.time(key, time_delta.total_seconds())
```
Kami: (just thinking out loud)

I assume that in the context where we will be using that seconds are enough and there is no need for micro / nano seconds, although miliseconds might not be a bad idea.

st2common/st2common/metrics/metrics.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+class Timer(object):
+    """ Timer context manager for easily sending timer statistics.
+    """
+    def __init__(self, key):
+        assert isinstance(key, str)
+        assert len(key) > 0
+        self.key = key
+        self._metrics = METRICS
```
Kami: Am I missing something? I don't see this `METRICS` variable being defined anywhere.

st2common/st2common/metrics/metrics.py
Kami: I have similar concerns as @m4dcoder here - we need some kind of init phase or similar, otherwise if we do it directly top-level inside a module, the value will depend on the import order which is something we want to avoid where possible (ideally import should have no side affects).

st2common/st2common/util/loader.py
```diff
@@ -259,3 +259,21 @@ def load_meta_file(file_path):
...
+def get_available_plugins(namespace):
+    """Return names of the available / installed plugins for a given namespace.
+    """
+    from stevedore.extension import ExtensionManager
+
+    manager = ExtensionManager(namespace=namespace, invoke_on_load=False)
+    return manager.names()
+
+
+def get_plugin_instance(namespace, name):
```
Kami: Perhaps we should move and re-use the similarly named method we use from auth backends code here and re-use it in both places.

Also, right now the code just returns the driver class and not the instance :)

Kami:
Added some small comments.

Only real blocker I see is how we assign `METRICS` variable. Besides that, it's minor stuff.

I will approve it so you don't get blocked for too long again, but let's please address the metrics assignment stuff :)

Nice work :+1:

st2common/st2common/metrics/metrics.py
bigmstone: @m4dcoder likely you didn’t import config. Try/except block made it silently return the base.

bigmstone: @kami none of this, as far as I can tell at the moment, requires a specific import order. Just that the installation install st2common to register the plugin.

st2common/st2common/metrics/drivers/prometheus.py
bigmstone: Their library is magic and people complain about it and they don't care. :'(

st2common/st2common/metrics/drivers/prometheus.py
bigmstone: Agree.

st2common/setup.py
m4dcoder: Can we add a noop metrics driver here?

st2common/st2common/config.py
```diff
@@ -390,6 +390,19 @@ def register_opts(ignore_errors=False):
     cli_opts = [debug, profile, use_debugger]
     do_register_cli_opts(cli_opts, ignore_errors=ignore_errors)
 
+    # Metrics Options stream options
+    metrics_opts = [
+        cfg.BoolOpt('enable', default=False,
+                    help='Enable metrics collection.'),
+        cfg.StrOpt('driver', default='statsd',
```
m4dcoder: Can we make the noop metrics driver as the default?

st2common/st2common/metrics/base.py
```diff
@@ -0,0 +1,177 @@
...
+        return wrapper
+
+
+def metrics_initialize():
+    """Initialize metrics constant
+    """
+    try:
+        _METRICS = get_plugin_instance(PLUGIN_NAMESPACE, cfg.CONF.metrics.driver)()
+    except (NoMatches, MultipleMatches, NoSuchOptError):
+        _METRICS = BaseMetricsDriver()
```
m4dcoder: If we implement the noop metrics driver and make it a default, we don't need this try-catch block.

st2common/st2common/metrics/base.py
```diff
@@ -0,0 +1,177 @@
...
+
+    return _METRICS
+
+
+def get_driver():
+    """Return metrics driver instance
+    """
+    if not _METRICS:
```
m4dcoder: Why do we need to prefix this w/ an underscore?  We don't have this practice for constants.

st2common/st2common/service_setup.py
```diff
@@ -126,6 +127,9 @@ def setup(service, config, setup_db=True, register_mq_exchanges=True,
     if run_migrations:
         run_all_rbac_migrations()
 
+    if cfg.CONF.metrics.enable:
+        metrics_initialize()
```
m4dcoder: If we add a noop metrics driver and make that the default, do we still need this metrics enable config option?

st2common/st2common/util/loader.py
m4dcoder: I agree with @Kami here. At first glance, I thought this method is returning an instance. But since it's invoke_on_load=False, it's just returning the class.

st2common/st2common/metrics/base.py
```diff
@@ -0,0 +1,177 @@
...
+        @wraps(func)
+        def wrapper(*args, **kw):
+            with self as counter_with_timer:
+                return func(*args, metrics_counter_with_timer=counter_with_timer, **kw)
+        return wrapper
+
+
+def metrics_initialize():
+    """Initialize metrics constant
+    """
```
m4dcoder: I believe you need to reference the METRICS constant as global. See this example as reference https://github.com/StackStorm/st2/blob/39517c0fb80359c48324a5b6ba4088972fbcb9de/st2common/st2common/services/coordination.py.

st2common/st2common/metrics/base.py
bigmstone: It indicates the variable is private and shouldn't be accessed directly. Like all things python there's no runtime prevention of accessing it.

Makefile
```diff
@@ -275,6 +275,9 @@ requirements: virtualenv .sdist-requirements
 	# new version of requests) which we cant resolve at this moment
 	$(VIRTUALENV_DIR)/bin/pip install "prance==0.6.1"
 
+	# Install st2common to register metrics drivers
```
m4dcoder: How about s/register metrics drivers/register common plugins?

Makefile
```diff
@@ -275,6 +275,9 @@ requirements: virtualenv .sdist-requirements
...
+	(cd ${ROOT_DIR}/st2common; ${ROOT_DIR}/$(VIRTUALENV_DIR)/bin/python setup.py install)
```
m4dcoder: I'm ok with this.  I thought making the requirements already did this because the last time I tested this branch, I didn't have to explicitly run the setup.  Can you use `python setup.py develop` instead?

bigmstone: @m4dcoder I'm getting all kinda nasty side effects from doing this. I've tried develop and install trying to work around it. Found some strange WebOb bug that manifests itself only when `st2common` is installed and was able to resolve it by bumping the version. Now I'm investigating these config issues. Long story short just ignore my commits until I get the tests passing.

PR #4005, opened 2018-02-17
Small improvements in the Python runner

Kami:
This pull request implements small improvements in the Python runner mentioned in #3976:

1. When parameters are passed to the action wrapper process via stdin, time out and throw an exception if no input is received within 2 seconds.

Without that change, we would wait for the input indefinitely when running the wrapper script directly (e.g. for debugging purposes) and up to `timeout` runner parameter seconds when running action under action runner.

2. Update runner to print full command string which can be used to run python wrapper process when using stdin parameters

Before:

```bash
/opt/stackstorm/virtualenvs/examples/bin/python -u /data/stanley/st2common/st2common/runners/python_action_wrapper.py --pack=examples --file-path=/opt/stackstorm/packs/examples/actions/print_to_stdout_and_stderr.py --user=stanley "--parent-args=[\"--config-file\", \"conf/st2.dev.conf\"]" --stdin-parameters "--config={\"region\": \"us-east-1\"}"
```

After:

```bash
echo '{"parameters": {"count": 3, "sleep_delay": 0.5}}' | /opt/stackstorm/virtualenvs/examples/bin/python -u /data/stanley/st2common/st2common/runners/python_action_wrapper.py --pack=examples --file-path=/opt/stackstorm/packs/examples/actions/print_to_stdout_and_stderr.py --user=stanley "--parent-args=[\"--config-file\", \"conf/st2.dev.conf\"]" --stdin-parameters "--config={\"region\": \"us-east-1\"}"
```

## TODO

 - [x] Tests

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,15 +21,13 @@ Changed
 * Additional refactor which makes action runners fully standalone and re-distributable Python
   packages. Also add support for multiple runners (runner modules) inside a single Python package
   and consolidate Python packages from two to one for the following runners: local runners, remote
-  runners, windows runners. #3999
+  runners, windows runners. (improvement) #3999
 
 Fixed
 ~~~~~
-
-* Fix ``Argument list too long`` error when very large parameters are passed into the python
-  wrapper arguments. The fix utilizes ``stdin`` to pass parameters instead of cli argument
-  list.
-  (bug fix) #1598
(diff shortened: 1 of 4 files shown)


Kami:
@enykeev Yeah, sadly there is no nicer way of doing that which is also cross platform and cross Python 2.x and 3.x compatible, AFAIK.

PR #4007, opened 2018-02-19
Upgrade eventlet to the latest version (0.22.1)

### LINKED ISSUE #3968: Pinned eventlet version has outstanding bugs
pinned version 0.17 of eventlet has outstanding bugs on it's monkey patching of the ssl module.

e.g.
https://github.com/eventlet/eventlet/issues/371

```
# Note: 0.20.0 removed select.poll() on which some of our code and libraries we
# depend on rely
```

@Kami  committed this reversion in https://github.com/StackStorm/st2/commit/1ec43d294e6770e56ec8f9990c805cb9dffe98c5

What was the specific issue?

Kami:
This pull request upgrades `eventlet` to the latest stable version.

Some tests were failing with this version because of the removal of `select.poll()` method. To get it to pass, I added a work around which injects original `select.poll()` for tests.

This issue itself only affected tests.

We don't perform monkey patching in the subprocess which is used to execute Python runner actions - we tried to do monkey patching there in the past, but it negatively affected / didn't work with some 3rd party libraries so we decided to remove it and making it opt-in instead of being enabled by default.

Resolves #3968.


Kami:
It looks like CI fails with a lot of errors, even though unit tests pass locally.

It looks like some of issues are related to the change in the DNS resolving.

Kami:
@warrenvw @armab It looks like st2-packages build is failing with;

```
File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/eventlet/patcher.py", line 94, in inject
    module = __import__(module_name, {}, {}, module_name.split('.')[:-1])
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/eventlet/support/dns/rdtypes/IN/WKS.py", line 23, in <module>
    _proto_tcp = socket.getprotobyname('tcp')
socket.error: protocol not found
```

New version of eventlet uses `dnspython` (async DNS library) by default which consults `/etc/protocols` file. I SSHed into the container and it looks like that file doesn't exist.

It looks like we should probably ensure this file exists and install `netbase` package on Ubuntu (http://senthilthecoder.com/post/ubuntu-getprotobyname/), wdyt?

warrenvw:
Yes, install `netbase`.

armab:
@warrenvw :+1:

An addition to that, if minimal Docker image requires `netbase` apt package for st2 to function correctly, - we'll need to add it as a dependency in `st2-packages` for `st2`, as well as find the respective package name for `rpm` (if any).

warrenvw:
FWIW, seems we need `setup` rpm package, not `netcfg`:

```
[vagrant@st2el7 ~]$ rpm -q --whatprovides /etc/protocols
setup-2.8.71-7.el7.noarch
```

Kami:
Yay, looks like adding `netbase` as a dependency for `st2` package fixed the build - https://circleci.com/gh/StackStorm/st2/6769

Kami:
@m4dcoder It's the underlying git library we use for git related operations - it shells out and uses `.poll()` method.

As mentioned in the description - this only affects tests where we use `eventlet` monkey patching, but not actual deployments, because we don't use monkey patching in the Python action runner subprocess.

Kami:
And to clarify the test situation - there is actually no need for this and many other tests to use eventlet monkey patching, but the way the test runner works (multiple test files run in a single process) and the way eventlet monkey patching works (import side affect) it's sadly a side affect from some other test.

We had issues with this cross test side affects and pollution in the past, but sadly there is no easy way around it - running each test file in a separate process would be too slow and expensive.

PR #4013, opened 2018-02-27
Add advanced user filter

apolloliu:
Add configuration permission_isolation to allow isolate rule and execution
for users except system_user.


st2api/st2api/controllers/v1/ruleviews.py
```diff
@@ -85,15 +87,31 @@ class RuleViewController(resource.ContentPackResourceController):
...
+        rules = super(RuleViewController, self)._get_all(sort=sort,
+                                                         offset=offset,
+                                                         limit=limit,
+                                                         raw_filters=raw_filters,
+                                                         requester_user=requester_user)
         result = self._append_view_properties(rules.json)
         rules.json = result
         return rules
 
+    @staticmethod
```
enykeev: It makes sense to me to separate the function into a mixin to reduce repetition.

st2api/tests/unit/controllers/v1/test_executions_rbac.py
```diff
@@ -135,3 +138,36 @@ def test_get_all_limit_minus_one(self):
...
+        }
+
+        # User with admin role assignment
+        user_db = self.users['admin']
+        self.use_user(user_db)
+
+        resp = self._do_post(data)
+        self.assertEqual(resp.status_code, http_client.CREATED)
+
+        resp1 = self.app.get('/v1/actionexecutions?limit=-1')
```
enykeev: I wonder how this test would look like with positive limit. I have a feeling that it will queries the page first and filters users later which means the endpoint could potentially reduce the page of no records. Lets add a test and if it happens, let's deal with that before we merge.

st2api/st2api/controllers/v1/ruleviews.py
apolloliu: Could I add a new class named `BeseResourceIsolationHandlerMixin`?

st2api/tests/unit/controllers/v1/test_executions_rbac.py
apolloliu: I update my PR and there is a raw user filter when querying resources in mongo.  All entries will be suitable for every user, so I there a page of no records won't exist.

st2common/st2common/config.py
```diff
@@ -52,7 +52,12 @@ def register_opts(ignore_errors=False):
         cfg.BoolOpt('sync_remote_groups', default=False,
                     help=('True to synchronize remote groups returned by the auth backed for each '
                           'StackStorm user with local StackStorm roles based on the group to role '
-                          'mapping definition files.'))
+                          'mapping definition files.')),
+        cfg.BoolOpt('permission_isolation', default=False,
+                    help='Isolate Some resources for every user. These resources just include '
+                         'rules and executions temporarily. all resources can just be operated '
+                         'or obversed by respetive users execpt system_uer who has the highest '
+                         'authority.')
```
cognifloyd: Here's a slightly modified help text. With my changes, is it still accurate?
```
                    help='Isolate resources by user. For now, these resources only include '
                         'rules and executions. All resources can only be viewed or executed '
                         'by the owning user except the system_user who can view or run '
                         'everything.')
```
I always wished people would correct me more often when I was learning Portuguese as my second language. I'm happy to help or proofread any time. In this case, I would use the word "only" instead of "just" because the list of resources and actions are countable (there are two resources, and two actions).

On a technical note:
I'm not clear on what you mean by "operated or observed" - I think you're saying that you can only `view` your own rules and executions, and that you can only `re-run` your own executions? Or are you also saying that the user that created a rule is the only user who can trigger it?

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -26,6 +26,7 @@
 
 from st2api.controllers.base import BaseRestControllerMixin
 from st2api.controllers.resource import ResourceController
+from st2api.controllers.resource import BeseResourceIsolationHandlerMixin
```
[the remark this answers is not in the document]
apolloliu: I don't think this could suit the code style of whole project.

st2api/st2api/controllers/v1/ruleviews.py
apolloliu: @cognifloyd  resource_model_filter have been used in ContentPackResourceController which is super class of RuleViewController. So it's no need.

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -725,6 +727,22 @@ def _get_action_executions(self, exclude_fields=None, include_fields=None,
                                                                 raw_filters=raw_filters,
                                                                 requester_user=requester_user)
 
+    @staticmethod
+    def resource_model_filter(model, instances, requester_user=None,
```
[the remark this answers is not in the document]
apolloliu: Thanks

st2common/st2common/config.py
apolloliu: @cognifloyd Thanks for your significant pointing! I think I should use your description. What I mean in `operated or observed` refers to view or other operations( re-run exeution, trigger with a rule, etc.) from Web UI.

longfei-zhang:
I'm looking forward this pacth ! This feature is very helpful

armab:
Do we need a corresponding update in `st2docs` (https://github.com/StackStorm/st2docs/blob/master/docs/source/rbac.rst) once the change is merged?

apolloliu:
@armab Sure, I will update relevant docs once this PR is going to be merged.

Kami:
Thanks for contributing this change :+1: 

Something like this was on our long-term TODO for all the API endpoint, but we didn't get to it yet :)

I will look at the actual code changes shortly.

st2api/st2api/controllers/resource.py
```diff
@@ -355,7 +365,28 @@ def _validate_exclude_fields(self, exclude_fields):
...
+    """
+    Isolate resources for users except system_user.
+    """
+    @staticmethod
+    def resource_model_filter(model, instances, requester_user=None,
+                              offset=0, eop=0, **kwargs):
+        result = []
+        for instance in instances[offset:eop]:
+            item = model.from_model(instance, **kwargs)
+            if not cfg.CONF.rbac.permission_isolation:
```
Kami: We should probably also bail out early (first if check, most common scenario) if RBAC is not enabled aka `if not cfg.CONF.rbac.enable`.

Kami: And perhaps, to make this loop more efficient we could handle those most common scenarios (rbac is not enabled, permission isolation is not enabled) outside the loop - simply fall back to default `resource_model_filter` implementation.

This way we will perform this check for most common scenario only once instead of `N` times where `N` is number of loop iterations.

st2api/st2api/controllers/resource.py
```diff
@@ -355,7 +365,28 @@ def _validate_exclude_fields(self, exclude_fields):
...
+                result.append(item)
+            elif requester_user.name == cfg.CONF.system_user.user:
```
Kami: I think we should also allow users with admin permission to view all the resources.

In fact, that's how we already handle it right now for user scoped datastore items - regular users can only see items scoped to themselves, but admins can see all the datastore items or the items scoped to a particular user.

This also brings up another question - to be consistent with datastore API endpoint, we should probably allow admin user to see items for a particular user (e.g. by using `?user=foo` or `?scope_user=foo` query parameter or similar).

Kami: Edit: Never mind. Looking at the code below, it looks like that is already supported - https://github.com/StackStorm/st2/pull/4013/files#diff-9a7b5df7b28bc1b332b8a777a5930f9cR104

Can you please confirm.

st2api/tests/unit/controllers/v1/test_executions_rbac.py
```diff
@@ -135,3 +138,74 @@ def test_get_all_limit_minus_one(self):
 
         resp = self.app.get('/v1/actionexecutions?limit=-1')
         self.assertEqual(resp.status_code, http_client.OK)
+
+    def test_get_respective_actions(self):
```
Kami: Let's please also add test cases for the following scenarios:

* system user can view all
* admin user can view all
* any user can view all if rbac is disabled
* any user can view all if rbac is enabled and permission isolation is disabled

Kami:
I've added some comments.

Overall nice work, just some small things which need to be addressed :)

And as @armab pointed out, let's please add some documentation to st2docs which explains how to enable this feature, how it works and limitations (which endpoints are currently supported).

Kami:
@apolloliu How is this PR looking?

It would be great if we could get those comments addressed as soon as possible so we can include this change in v2.7.0 release.

We plan to start working on v2.7.0 either tomorrow or early next week.

st2api/st2api/controllers/resource.py
Kami: I started working on this and other comments.

There is another thing we should decide - we have a special `observer` system role which grants "view" permission on all the resources.

Should this role also grant "view" on all the resources even when permission isolation is enabled? I think so.

What do others think?

st2api/st2api/controllers/resource.py
apolloliu: @Kami  I should have discussed this with you. I assumed all items could only be seen by platform manager(system_admin). There is not a specific permission type for view and execute(create, update, get, delete) items belongs to a particular user, so I extend this permission limit on `admin`. Other common user has only permission of viewing.

st2api/st2api/controllers/resource.py
apolloliu: @Kami Thank you, I will check if rbac is enabled in the next commit. But I'm not sure I get you that this check only once.

st2api/tests/unit/controllers/v1/test_executions_rbac.py
apolloliu: will update in next commit.

Kami:
@apolloliu Sorry, didn't know if you will have a chance to work on it - #4054

Kami:
@apolloliu I will finish the test cases I was working on, push it to that PR and you can continue work from there if you want.

Please let me know so we won't both be working on the same thing and sorry again!

Kami:
Closing this in favor or #4054 which addresses some outstanding issues, adds more tests, etc.

Thanks again for your contribution and I hope we will be able to get #4054 merged in time for v2.7.0 release.

PR #4016, opened 2018-02-28
Add new --python3 flag to st2 pack install CLI command

Kami:
This pull request adds new `--python3` flag to `st2 pack install`` command.

When this flag is provided, virtual environment which is created for a pack uses python3 binary (based on the value of `actionrunner.python3` config option).

Underneath, it adds ``python3`` parameter to `packs.{install,setup_virtualenv}` actions.

This pull request builds on top of @tonybaloney's idea in #3922.

We couldn't come up to a consensus in that PR. I thought that approach in that PR is still a bit too invasive until we have CI, etc. in place for packs so we can verify end to end that pack also (or only) works with Python 3.

This pull request goes with a more incremental middle of the ground opt-in approach. If user wishes to use Python 3 for a particular pack they can opt-in. And if things don't go well (stuff doesn't work under Python 3), they can easily recreate virtualenv / reinstall pack with default Python 2 binary.

In the future, we can also build on top of this approach and perhaps utilize ``pack.yaml`` metadata attribute to signal if pack additionally / only works with Python 3 as suggested in #3922.

## TODO

- [x] Tests

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,14 @@ Added
 * Add support for utf-8 / unicode characters in the pack config files. (improvement) #3980 #3989
 
   Contributed by @sumkire.
+* Add new ``--python3`` flag to ``st2 pack install`` CLI command and ``python3`` parameter to
+  ``packs.{install,setup_virtualenv}`` actions. When the value of this parameter is True, it
+  uses ``python3`` binary when creating virtual environment for that pack (based on the value of
+  ``actionrunner.python3_binary`` config option).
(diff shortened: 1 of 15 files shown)


Kami:
@bigmstone Yeah that might change in the future when whole StackStorm platform supports and runs under Python 3 :)

Kami:
@tonybaloney Hopefully this PR will work as a good compromise until we have the whole platform to run on Python 3 :)

tonybaloney:
:+1:

PR #4023, opened 2018-03-06
Add database model to support coordinated write

m4dcoder:
Add a base model with a change revision field that is used as save condition in MongoDB for coordinated writes between multiple threads or processes. This is required to support coordinated update of workflow execution by parallel tasks running at multiple processes.

### DIFF
--- a/st2common/st2common/models/db/__init__.py
+++ b/st2common/st2common/models/db/__init__.py
@@ -30,7 +30,7 @@ from st2common import log as logging
 from st2common.util import isotime
 from st2common.models.db import stormbase
 from st2common.models.utils.profiling import log_query_and_profile_data_for_queryset
-from st2common.exceptions.db import StackStormDBObjectNotFoundError
+from st2common.exceptions import db as db_exc
 
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


st2common/st2common/models/db/__init__.py
```diff
@@ -498,6 +498,48 @@ def _process_datetime_range_filters(self, filters, order_by=None):
...
+        if not hasattr(instance, '_original_rev'):
+            return self.insert(instance)
+        else:
+            save_condition = {
+                'id': instance.id,
+                'rev': instance._original_rev
+            }
+
+            instance.rev = instance._original_rev + 1
+            instance.save(save_condition=save_condition)
```
enykeev: I wonder if we should catch the error here and reraise it as st2 conflict exception

m4dcoder: Yeah, I'll rewrap it as a st2 exception.

st2common/st2common/models/db/stormbase.py
```diff
@@ -272,3 +272,17 @@ def get_reference(self):
             ref = ResourceReference(pack=self.pack, name=self.name)
 
         return ref
+
+
+class ChangeRevisionFieldMixin(object):
```
Kami: Would be good to add some docstring to this class and clarify what it does and what kind of purpose it serves.

st2common/st2common/models/db/__init__.py
```diff
@@ -498,6 +498,42 @@ def _process_datetime_range_filters(self, filters, order_by=None):
         return filters, order_by_list
 
 
+class ChangeRevisionMongoDBAccess(MongoDBAccess):
+
```
Kami: Same here, docstring wouldn't hurt.

m4dcoder: ack

Issue #4025, opened 2018-03-06
[Feature] Create /etc/sysconfig/st2ctl to allow for customization of COMPONENTS and STANCONF


nmaludy:
Currently in the `st2ctl` script, `COMPONENTS` and `STANCONF` are hard coded.

In our HA setup i have to use a `sed` command to edit the file. Instead it would be great if there was an environment variable in `/etc/sysconfig` (rhel) and `/etc/defaults` (ubuntu) where these settings could be placed. Then, `st2ctl` can source this file if it exists and load in the customizations.

You can assign this to me and i'll knock it out when i have some free time.

nmaludy:
Closed by #4027

PR #4027, opened 2018-03-07
Fix for #4025 - st2ctl environment variables

nmaludy:
This is a fix for #4025 .

It adds the ability for `st2ctl` to load in environment variables from an `/etc/default/st2ctl` (ubuntu) or `/etc/sysconfig/st2ctl` file. This allows deployments to utilize selective component lists and non-default st2.conf  locations.

### DIFF
--- a/st2common/bin/st2ctl
+++ b/st2common/bin/st2ctl
@@ -14,13 +14,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-LOGFILE="/dev/null"
 COMPONENTS="st2actionrunner st2api st2stream st2auth st2garbagecollector st2notifier st2resultstracker st2rulesengine st2sensorcontainer st2chatops mistral"
-STANCONF="/etc/st2/st2.conf"
+ST2_CONFIG="/etc/st2/st2.conf"
 
 # Ensure global environment is sourced if exists
 # Does not happen consistently with all OSes we support.
 [ -r /etc/environment ] && . /etc/environment
+
+# load in environment to allow override of COMPONENTS and ST2_CONFIG above
+# Ubuntu/Debian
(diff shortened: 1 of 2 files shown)


st2common/bin/st2ctl
```diff
@@ -14,14 +14,19 @@
...
-STANCONF="/etc/st2/st2.conf"
+ST2_CONFIG="/etc/st2/st2.conf"
 
 # Ensure global environment is sourced if exists
 # Does not happen consistently with all OSes we support.
 [ -r /etc/environment ] && . /etc/environment
 
+# load in environment to allow override of COMPONENTS and ST2_CONFIG above
+# Ubuntu/Debian
+[ -r /etc/default/st2ctl ] && . /etc/default/st2ctl
```
bigmstone: I'd like it better if we used `source` here just be be more explicit, but this shebang is bash so it shouldn't matter - not a blocker, just a preference.

st2common/bin/st2ctl
```diff
@@ -14,14 +14,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-LOGFILE="/dev/null"
 COMPONENTS="st2actionrunner st2api st2stream st2auth st2garbagecollector st2notifier st2resultstracker st2rulesengine st2sensorcontainer st2chatops mistral"
-STANCONF="/etc/st2/st2.conf"
+ST2_CONFIG="/etc/st2/st2.conf"
```
bigmstone: I was going to complain about this since we use `STANCONF` in other places...but we also use `ST2_CONF` in other places and I prefer the name `ST2_CONF` so, YOLO?

nmaludy:
@bigmstone I just pushed a change that uses `source` instead of `.` and `ST2_CONF` instead of `ST2_CONFIG`.

PR #4033, opened 2018-03-13
Ensure parity between SensorType API spec and implementation

enykeev:
I'm currently working on Triggers app for st2web and I've noticed our /sensortypes endpoint doesn't confirm with its spec. The PR fixes that. It also allows us to filter sensortypes by triggers they produce, which in turn makes it easier to build trigger objects for the UI.


Kami:
Good catch although some tests would also be nice :)

PR #4041, opened 2018-03-19
Fix, tests for #4033

Kami:
This pull request fixes `?enabled` query parameter filter added in #4033. 

It was missing a transformation so it didn't actually work. In addition to that, it also adds tests for those filters.

A couple of things (most of them we already talked about in the past):
* Such changes are not much help without the corresponding test cases
* We should utilize correct types for query param filters and not just abuse `type: string` for everything. It's wrong and incorrect (e.g. enabled should have `type: boolean`). I already voiced my opinion about that in the past (especially for arrays), but we should do something about it.
* We should auto-generate tests for all the query param filters using openapi definition file. Previous point is a pre-requisite for this.


Kami:
I opened an issue for the second bullet point - #4042.

/cc @enykeev

Issue #4039, opened 2018-03-19
Invalid file path while trying to create an action through REST API


NikosVlagoidis:
https://api.stackstorm.com/api/v1/actions/#/actions_controller.post

While I was trying to create a new action which I previous deleted with the REST API so the file is in the filesystem the following error occurred.

Request:

```
{	
	"name": "test_action",
	"runner_type": "python-script",
	"data_files":[{
		"file_path": "/opt/stackstorm/packs.dev/my_pack/action",
		"content": "test_action.py"
	}]
}
```

Response: 

```
{
    "faultstring": "Invalid file path: /opt/stackstorm/packs.dev/my_pack/action"
}
```

LindsayHill:
I've only seen that used for Mistral & Action Chain workflow definition files, but it looks like it might work for other actions.

At a minimum, one problem you have is that your path looks wrong - I would expect it to end with `actions`. Looking at how Workflow Designer does it, it uses the same value as the entry point anyway - i.e. it's relative to the action itself. That also seems to use file_path as the full path to the filename - e.g. workflows/test_workflow.yaml, and then provides the content as content body, not file name.

NikosVlagoidis:
Do you mean you have to type the actual yaml file in the "content" field?

Kami:
Yes, the `content` attribute needs to contain the actual (text) file content.

And as @LindsayHill has pointed out, the `file_path` should probably be something like that - `/opt/stackstorm/packs.dev/my_pack/actions/test_action.py`.

NikosVlagoidis:
But the file already exists on the filesystem! Why can not I use the same?

Kami:
If action data files are still on disk and you only deleted / unregistered action metadata from the system, but the action data files and metadata files are still on disk, you should just be able to run `st2ctl reload --register-all` and action will be re-registered.

If you want to use API for that, I believe you can just leave `data_files` attribute empty.

NikosVlagoidis:
I have the same problem:

Request: 

```
{	"pack": "my_pack",
	"name": "test_action",
	"runner_type": "python-script",
	"data_files":[{
		"file_path": "/opt/stackstorm/packs.dev/my_pack/actions/test_action.py",
		"content": "import sys"
	}]
}
```

Response: 

```
{
    "faultstring": "Invalid file path: /opt/stackstorm/packs.dev/encode/actions/test_action.py"
}
```

The following request leads to interval error (notice the missing / in the start of the path):

Requests: 

```
[...]

PR #4046, opened 2018-03-20
Throw a more user-friendly error on invalid file path when resolving path to a pack resource

Kami:
For security reasons (to avoid directory traversal attacks, writing outside of the pack directory, etc.), we only allow users to write pack data files inside the pack directory.

The error we threw before was a bit ambiguous and not user-friendly so this PR fixes that.

Before:

```bash
Invalid file path: /opt/stackstorm/packs/dummy_pack_2/bar.py
```

After:

```
Invalid file path: /opt/stackstorm/packs/dummy_pack_2/bar.py. File path needs to be relative to the pack actions directory (opt/stackstorm/packs/test/actions). For example "my_action.py".
```

Related issue #4039.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -62,6 +62,8 @@ Changed
   This reduces the unnecessary traffic and CPU time by querying the mistral API. Included a command to
   manually add a state entry for Mistral workflow execution to recover from any callback failures.
   (improvement)
+* Throw a more user-friendly error when writing pack data files to disk and when an invalid file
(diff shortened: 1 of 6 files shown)


armab:
Do we auto-generate https://api.stackstorm.com/ somehow based on latest API definition?
I guess it's outdated

Kami:
IIRC, we re-generate it on master build, but I could be wrong - @enykeev should know for sure.

armab:
From S3, last update: `Nov 3, 2017` :disappointed:

Issue #4044, opened 2018-03-20
Windows Action Output Inconsistent


LindsayHill:
The output from running a `windows-cmd` or `windows-script` action has duplicate content, and is inconsistent with that used for other actions. 

For example, running `st2 run core.windows_cmd cmd='ping 10.0.1.127' host='x.x.x.x' password='xxxxx'` returns this:
```
$ st2 execution get 5ab01b573cac1c492f99067a
id: 5ab01b573cac1c492f99067a
status: succeeded (3s elapsed)
parameters:
  cmd: ping 10.0.1.127
  host: 10.0.0.104
  password: '********'
result:
  exit_code: 0
  result: "
Pinging 10.0.1.127 with 32 bytes of data:
Reply from 10.0.1.127: bytes=32 time<1ms TTL=64
Reply from 10.0.1.127: bytes=32 time<1ms TTL=64
Reply from 10.0.1.127: bytes=32 time<1ms TTL=64
Reply from 10.0.1.127: bytes=32 time<1ms TTL=64

Ping statistics for 10.0.1.127:
    Packets: Sent = 4, Received = 4, Lost = 0 (0% loss),
Approximate round trip times in milli-seconds:
    Minimum = 0ms, Maximum = 0ms, Average = 0ms
"
  stderr: ''
  stdout: "
Pinging 10.0.1.127 with 32 bytes of data:
Reply from 10.0.1.127: bytes=32 time<1ms TTL=64
Reply from 10.0.1.127: bytes=32 time<1ms TTL=64
Reply from 10.0.1.127: bytes=32 time<1ms TTL=64
Reply from 10.0.1.127: bytes=32 time<1ms TTL=64

Ping statistics for 10.0.1.127:
    Packets: Sent = 4, Received = 4, Lost = 0 (0% loss),
Approximate round trip times in milli-seconds:
    Minimum = 0ms, Maximum = 0ms, Average = 0ms
"
```

Note that the output is duplicated in the `stdout` and `result` field. It also does not return `succeeded` or `failed` fields. 

Expected behavior: return output formatted similarly to `core.local`, e.g.:
```shell
lhill@st2demo004:~$ st2 run core.local cmd=id
.
id: 5ab059da3cac1c492f990695
status: succeeded
parameters:
  cmd: id
result:
  failed: false
  return_code: 0
  stderr: ''
  stdout: uid=706(stanley) gid=706(stanley) groups=706(stanley),1001(sudoers)
  succeeded: true
[...]

PR #4047, opened 2018-03-20
Make windows runners output object be consistent with other runner output objects

Kami:
This pull request updates a result / output dictionary returned by the Windows runners to match local and remote runner output object.

The change is breaking / partially backward incompatible so it needs to be included in minor / major release (2.7.0) and we need to document that change in changelog and upgrade notes document.

## TODO

- [x] Update affected tests
- [x] New tests
- [x] Changelog entry
- [x] Upgrade notes entry - https://github.com/StackStorm/st2docs/pull/707

Related issue #4044.

### DIFF
--- a/contrib/runners/windows_runner/windows_runner/windows_command_runner.py
+++ b/contrib/runners/windows_runner/windows_runner/windows_command_runner.py
@@ -93,13 +93,12 @@ class WindowsCommandRunner(BaseWindowsRunner):
         if exit_code != 0:
             error = self._parse_winexe_error(stdout=stdout, stderr=stderr)
 
-        result = stdout
-
         output = {
             'stdout': stdout,
             'stderr': stderr,
-            'exit_code': exit_code,
-            'result': result
+            'return_code': exit_code,
+            'succeeded': (exit_code == 0),
+            'failed': (exit_code != 0)
         }
 
         if error:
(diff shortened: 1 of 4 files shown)


contrib/runners/windows_runner/windows_runner/windows_command_runner.py
```diff
@@ -93,13 +93,12 @@ def run(self, action_parameters):
...
             error = self._parse_winexe_error(stdout=stdout, stderr=stderr)
 
-        result = stdout
-
         output = {
             'stdout': stdout,
             'stderr': stderr,
-            'exit_code': exit_code,
-            'result': result
+            'return_code': exit_code,
```
Kami: I personally prefer exit code over return code, but I made the change for the consistency reasons.

Kami:
While working on this, I also noticed windows runner doesn't correctly handle / use "timeout" live action / execution status so I fixed that (windows runner probably predates "timeout" status).

contrib/runners/windows_runner/windows_runner/windows_command_runner.py
LindsayHill: Agreed - I think that consistency here is more helpful

LindsayHill:
Looking good so far. I think that you're right about some of this predating other changes. I suspect we updated the Python & Command runners, but didn't make the same changes to these runners. 

Assuming we move to pywinrm at some point in the near future, hopefully most of these changes will still be relevant.

Kami:
@LindsayHill this PR is ready for review now.

PR #4049, opened 2018-03-22
Upgrade various dependencies to latest stable versions

Kami:
Some more house keeping.


fixed-requirements.txt
```diff
@@ -19,24 +19,24 @@ jsonschema==2.6.0
...
+pymongo==3.6.1
 passlib==1.7.1
 lockfile==0.12.2
 python-gnupg==0.4.1
 jsonpath-rw==1.4.0
 pyinotify==0.9.6
-semver==2.7.7
+semver==2.7.9
 stevedore==1.28.0
 paramiko==2.4.0
```
armab: How about Paramiko?

I just got an alert that this package version is vulnerable, see:
https://snyk.io/vuln/SNYK-PYTHON-PARAMIKO-42086
and 
https://github.com/paramiko/paramiko/issues/1175

Kami: Good catch - thanks :+1: 

It looks like it only affects server mode which we don't use, but I will still look into the upgrade.

armab:
:+1:

PR #4052, opened 2018-03-23
Fix rules engine to correctly handle and render default action parameter values

### LINKED ISSUE #4050: Default parameter type validation fails for non-string types when workflows are triggerred using rules/aliases
**version: 2.6.0**
TL;DR
When an array from pack config is passed as default parameter to an action, the action fails with a type validation error when triggered via rules/aliases. But runs fine when triggered directly.

**Long Version**
I have a  workflow that takes an input of type array. In the pack config, I have a config item of type array.  In the action metadata  file, for the above parameter  I specified the default value from the pack config.

When this action is triggered via a rule or an alias, the action fails to execute due to a type validation error:
```
ValueError: Failed to cast value "{{ config_context.default_recipients }}" (type: unicode) for parameter "recipients" of type "array": malformed string. Perhaps the value is of an invalid type?
```
However, the actions runs fine when executed directly.

*config schema:*
```
---
  default_recipients:
    description: "Comma separated email addresses of the recipients to send the reports/notification"
    type: "array"
    required: true
```

*config:*
```
default_recipients: ["email1@address.com", "email2@address.com"]
```
 
Here are my actions:

*Metadata for Action that takes the array as param*
```
---
name: test_action
pack: test
runner_type: mistral-v2
description: Testing arrays
enabled: true
entry_point: workflows/test_action.yaml
parameters:
  recipients:
    required: true
    type: array
    description: list of email addresses
    default: "{{ config_context.default_recipients }}"

```

*Action definition*
```
version: '2.0'

test.test_action:
    input:
      - recipients
    tasks:
        t1:
            action: core.local
            input:
                cmd: echo "{{ _.recipients|join(', ') }}"

```

*The rule in question*
```
---
name: test_rule
pack: test
description: Rule to send notifications on completion of task.
enabled: true

trigger:
  type: core.st2.generic.notifytrigger
  parameters: {}
criteria:
  trigger.channel:
    pattern: "dummy"
    type: "equals"
action:
  ref: "test.test_action"

```

*Action that generates the notification (just for reference)*
```
---
description: Action that executes an arbitrary Linux command on the localhost.
enabled: true
entry_point: ''
name: test_notify_action
notify:
  on-complete:
    routes:
    - dummy
    message: '42'
parameters:
  cmd:
    description: Arbitrary Linux command to be executed on the remote host(s).
    required: true
    type: string
  sudo:
    immutable: true

```

I run the action `test_notify_action` to generate the notification that would trigger the action in question:
```
# st2 run test.test_notify_action cmd="echo 'tango'"

.
id: 5ab40ecf25acbf5b0b4c747e
status: succeeded
parameters: 
  cmd: echo 'tango'
result: 
  failed: false
  return_code: 0
  stderr: ''
  stdout: tango
  succeeded: true

```
It succeeds but fails to trigger the action.

*Corresponding Logs*

```
2018-03-23 01:45:12,995 140439250711088 INFO matcher [-] 1 rule(s) found to enforce for st2.generic.notifytrigger.
2018-03-23 01:45:12,999 140439250711088 INFO engine [-] Matched 1 rule(s) for trigger_instance 5ab40ed025acbf03cc58e55f (trigger=core.st2.generic.notifytrigger)
2018-03-23 01:45:13,002 140439250711088 INFO enforcer [-] Invoking action test.test_action for trigger_instance 5ab40ed025acbf03cc58e55f with params {}.
2018-03-23 01:45:13,063 140439250711088 ERROR enforcer [-] Failed kicking off execution for rule RuleDB(action=ActionExecutionSpecDB@140439245111760(ref="test.test_action", parameters="{}"), criteria={u'trigger.channel': {u'pattern': u'dummy', u'type': u'equals'}}, description="Rule to send notifications on completion of task.", enabled=True, id=5ab40ca425acbf2fbd8f2c68, name="test_rule", pack="test", ref="test.test_rule", tags=[], trigger="core.st2.generic.notifytrigger", type=RuleTypeSpecDB@140439245111888(ref="standard", parameters="{}"), uid="rule:test:test_rule").
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2reactor/rules/enforcer.py", line 74, in enforce
    execution_db = self._do_enforce()
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2reactor/rules/enforcer.py", line 114, in _do_enforce
    return RuleEnforcer._invoke_action(self.rule.action, params, context)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2reactor/rules/enforcer.py", line 161, in _invoke_action
    liveaction, execution = action_service.request(liveaction)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/services/action.py", line 156, in request
    liveaction, execution = create_request(liveaction)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/services/action.py", line 87, in create_request
    allow_default_none=True)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/util/schema/__init__.py", line 293, in validate
    jsonschema.validate(instance=instance, schema=schema, cls=cls, *args, **kwargs)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/jsonschema/validators.py", line 541, in validate
    cls(schema, *args, **kwargs).validate(instance)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/jsonschema/validators.py", line 130, in validate
    raise error
ValidationError: u'{{ config_context.default_recipients }}' is not of type u'array'

Failed validating u'type' in schema['properties'][u'recipients']:
    {u'default': u'{{ config_context.default_recipients }}',
     u'description': u'list of email addresses',
     u'required': True,
     u'type': u'array'}

On instance[u'recipients']:

...

```

when I run the action directly, it works fine.
```
# st2 run test.test_action
....
id: 5ab4104725acbf5b0b4c7484
action.ref: test.test_action
parameters: 
  recipients:
  - email1@address.com
  - email2@address.com
status: succeeded
result_task: t1
result: 
  failed: false
  return_code: 0
  stderr: ''
  stdout: email1@address.com, email2@address.com
  succeeded: true
start_timestamp: Thu, 22 Mar 2018 20:21:27 UTC
end_timestamp: Thu, 22 Mar 2018 20:21:34 UTC
+--------------------------+------------------------+------+------------+-------------------------------+
| id                       | status                 | task | action     | start_timestamp               |
+--------------------------+------------------------+------+------------+-------------------------------+
| 5ab4104825acbf5b0b4c7487 | succeeded (1s elapsed) | t1   | core.local | Thu, 22 Mar 2018 20:21:28 UTC |
+--------------------------+------------------------+------+------------+-------------------------------+

```

Kami:
User reported an issue in #4050 which shows that the default action parameters are not rendered when an execution is triggered via rule.

I checked the code and it looks like we skip whole "render_live_params" section which handles thing like rendering default parameter values, casting parameters, etc.

We only rendered some parameters based on the action parameters defined on the rule using `Jinja2BasedTransformer` class which uses values from trigger instance.

I'm not sure if I'm missing some context, but this doesn't look right to me.

I will update the code to use and call `render_live_params` in the same fashion when we do when scheduling an execution via the API. This fill happen after the first pass where we render rule action parameters using the trigger instance as the context.

I hope we have enough test cases for all the various scenarios and edge cases which would expose any issues related to this change - e.g. if there was a specific reason why we didn't render default parameter values there.

## TODO

- [x] Tests
- [x] Changelog entry

Resolves #4050.


Kami:
Exist tests seem to pass, also added test cases for all those various scenarios which we previously didn't handle correctly.

Kami:
I'm looking if we can also get rid of the existing data transformer abstraction and moving that to `render_live_params` by adding trigger instance to the context which is available in that function.

That would decrease the complexity and make the code simpler.

Kami:
I managed to get rid of it and simplified the code in - 18b8c6c06dfde67f89cc865d069144627cd88397.

I assume this function predates `render_live_params`, at least in some form and that's the reason why it was there. AFAIK, there is no need for it anymore because the same functionality can be handled by `render_live_params`.

All the tests still pass.

PR #4054, opened 2018-03-29
[RBAC] User resource permission isolation

Kami:
This pull request works on top of a great start in #4013.

The goal is to get this PR finished in time for v2.7.0.

It includes the following changes (most of them come from comments on that PR):

* [x] Skip permission isolation checks if RBAC is disabled
* [x] Allow admin users to view all the resource when permission isolation is enabled
* [x] Some code performance optimizations (perform check outside of the loop so it happens once instead of N times, etc.
* [x] More and better test cases for all the scenarios (user is admin, user is system user, etc.)
* [x] Implement resource isolation for rules API endpoint
* [x] Decide on how to handle special system "observer" role
* [x] Implement resource isolation for get one API endpoints
  * [x] Action Executions
  * [x] Rules 
* [x] Documentation change - https://github.com/StackStorm/st2docs/pull/714

### How to handle "observer" system role

Right now we have a special "observer" system role. This role grants user "view" permission on all the resources.

We should decide how to tie this into permission isolation - should this role also grant ability to view resources for all the users (aka user with this role can overcome permission isolation - same as for the admin user).


Kami:
@apolloliu and I decided it makes sense to leave observer role as it is - if permission isolation is enabled, user with observer role will still only be able to view their own resources.

Kami:
Another thing we need to decide on (and implement) is how to handle filtering for the "get one" API endpoints.

Right now we only do this "isolation filtering" for "get all" (list) API endpoints. This means user could still view resources from other users when permission isolation is enabled by using "get one" API endpoint.

This means that without also implementing it for "get one" (view one) API endpoints, it makes little sense and it might even provide false sense of security.

@apolloliu what do you think?

Kami:
Another thing we need to decide on (and implement) is how to handle filtering for the "get one" API endpoints.

Right now we only do this "isolation filtering" for "get all" (list) API endpoints. This means user could still view resources from other users when permission isolation is enabled by using "get one" API endpoint.

This means that without also implementing it for "get one" (view one) API endpoints, it makes little sense and it might even provide false sense of security.

@apolloliu what do you think?

apolloliu:
@Kami  What I thought all operations is based on WebUI, user can get the resources in their own view list. It makes great sense as far as API endpoints without doubt.

Kami:
@apolloliu Alright - in this case I will also add a check to the "get one" API endpoints.

If permission isolation is enabled and non-admin user tries to view a single item (for now execution, rule) via "get one" API endpoint, they will get access denied error back.

I think that makes sense and it's better than just get all which, by itself, gives false sense of security.

Kami:
Resource isolation for "get one" API endpoints is now also implemented - 83c54638a877e13b0c0c09165e40f61e55c42698, de24fc71d6c6b009416f628950cabbf1b0271191, 0af95302d620465544a490b00609a74c6fe4bf84, 7b24fb71e2abeff562303dceab3d67387c9c103e, 4623307f9cf0fc5888b9aa3ece855e2b5490c206.

The change took slightly longer than I would hope for, but it's a security sensitive code so we shouldn't rush it.

I still need to implement isolation for rules (just tests) and do some more smaller changes (more loggic, etc.).

/cc @apolloliu

Kami:
@m4dcoder Sure (and just working on documentation change which will make this more clear).

This pull request adds additional "filtering" on top of RBAC. If this feature is enabled, users can only view resources which are created / owned by them (context.user == currently authenticated user) when using "get all" and "get one" API endpoint. In short - all the users except admins and system user can't view resources which belong to other users.

Right now that feature is a behind a feature flag (disabled by default) and supported by `/v1/executions` and `/v1/rules` API endpoint.

In fact, adding something like this (and replacing `RESOURCE_LIST` permission with that) was on our long term RBAC road map, but we never good to it since it wasn't requested by the users too much.

Maybe the test cases will also help clear it up a bit - https://github.com/StackStorm/st2/pull/4054/files#diff-46185e3f660a54bffa53da7a21a2e073R181.

Kami:
Docs are now available in https://github.com/StackStorm/st2docs/pull/714.

CHANGELOG.rst
```diff
@@ -46,6 +46,13 @@ Added
   related methods on the SensorService class. (improvement) #3895 #4057 #4058
 
   Reported by @djh2020, @mxmader.
+* Add context field to rule model in which each rule has its own corresponding user. Besides, there
+  is a new RBAC configuration ``permission_isolation``. Whoever can only operate and observe their
+  own rules or executions except ``system_user`` and users with RBAC admin role when set to
+  ``True``. That means system_user has the most powerful permission to operate all resources
+  including rules or executions. (new feature) #4013
+
+  Contributed by Hanxi Liu (@apolloliu).
```
apolloliu: @Kami  Aha, I think you should add yourself into contributors as well  and change the PR number of line 53 into #4054. :smile:

apolloliu:
@Kami  We should inherit the same classes on ruleviews as rules did. You know it's very important to webUI. Besides, there are some significant issues we have to concern, for example, that the interval timer jobs which we take system_user (defaultly be stanley) to execute are unable to be obtained by a specific non-admin user.

st2api/st2api/controllers/resource.py
```diff
@@ -368,6 +400,61 @@ def _validate_exclude_fields(self, exclude_fields):
...
+        # RBAC or permission isolation is disabled, bail out
+        if not (cfg.CONF.rbac.enable and cfg.CONF.rbac.permission_isolation):
+            result = super(BaseResourceIsolationControllerMixin, self).resources_model_filter(
+                model=model, instances=instances, requester_user=requester_user,
+                offset=offset, eop=eop, **from_model_kwargs)
+
+            return result
+
+        result = []
+        for instance in instances[offset:eop]:
```
m4dcoder: Is this going to cause bottleneck in performance?

Kami: Any kind of filtering will definitely decrease the performance, but I think it should be fine since we only do it for non-system (non-stanley) users.

Kami: And to clarify further - we wanted something like that from the beginning but one of the reasons why I avoided implementing it (in addition to customers not requesting it) was that implementing it in a performance efficient manner would be complex and require caching, etc.

In this case, it's opt-in so it's not a big deal and we can sort out the performance (introduce caching, etc). later, if needed.

Kami:
@apolloliu Added support for rule views API endpoint - 91f979021ed451f6857034b0b1152f7c5fe515f7, e35ef6615ed2569cb4080b506071f2df1e9cec78.

Plan to merge it later today.

PR #4061, opened 2018-04-03
Allow descriptions in rule trigger and action fields

enykeev:
The `/ruleOverview` endpoint extends rule object with additional information to reduce the number of API calls client has to make. This PR enables client to edit this objects without additional changes.


Kami:
As discussed on Slack - so on PUT we ignore those fields and don't store them in the database, right?

enykeev:
exactly

PR #4067, opened 2018-04-05
Fix the encryption key path for the key value config in st2tests

m4dcoder:
The encryption key path that is overriding config in st2tests is relative. In certain unit tests, the path is not recognizable. This patch provides an absolute path of the key file instead.


Kami:
Good catch and change :+1: 

As mentioned on the Slack, it's usually a good idea to try to resolve it to a full absolute path everywhere where possible to avoid such issues :)

PR #4068, opened 2018-04-05
Refactor the callback module for post run to be more generic

m4dcoder:
Refactor the callback module for runner post run to be more generic. Currently, the callback takes a REST endpoint as input argument. The callback module is abstracted to take other input arguments. In this patch, the entire liveaction database object is passed to avoid overcrowding the method signature. Unfortunately, the liveaction object that the runner reference to is out dated by the time the runner makes the post run and there are more than one references of the object in the runner container. The runner container and runner is refactored to ensure references objects are used and updated.

### DIFF
--- a/st2actions/st2actions/container/base.py
+++ b/st2actions/st2actions/container/base.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 from __future__ import absolute_import
-import json
+
 import sys
 import traceback
 
@@ -34,6 +34,7 @@ from st2common.util.action_db import (update_liveaction_status, get_liveaction_b
 from st2common.util import param as param_utils
 from st2common.util.config_loader import ContentPackConfigLoader
 from st2common.metrics.base import CounterWithTimer, format_metrics_key
+from st2common.util import jsonify
 
 from st2common.runners.base import get_runner
 from st2common.runners.base import AsyncActionRunner, PollingAsyncActionRunner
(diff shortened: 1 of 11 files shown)


Kami:
Saw on Slack that this wont go into v2.7.0 so I will add it to the new v2.8.0 milestone.

m4dcoder:
@bigmstone I guess I'll go back to runnertype, not changing liveaction to live_action.  there were many inconsistencies in our code on runner_type, runnertype, runner_type_db...  i was trying to fix it one way.  what a mess...

m4dcoder:
@bigmstone nevermind, i will leave my change as is. i tried to change it back to runnertype but then again ran into situation where other places is referencing as runner_type and i don't want to touch these other part of the system.

st2actions/st2actions/container/base.py
```diff
@@ -327,28 +321,26 @@ def _get_rerun_reference(self, context):
...
-                            config=config)
+        runner = get_runner(
+            package_name=runner_type_db.runner_package,
+            module_name=runner_type_db.runner_module,
+            config=config)
 
         # TODO: Pass those arguments to the constructor instead of late
         # assignment, late assignment is awful
-        runner.runner_type_db = runnertype_db
+        runner.runner_type = runner_type_db
```
Kami: I really dislike not having a `_db` suffix since it's confusing.

If anything, I would to the reverse - add `_db` suffix to all the objects which represent DB models so it's intermediately clear what you are working with.

st2actions/tests/unit/test_runner_container.py
```diff
@@ -109,8 +109,8 @@ def test_pre_run_runner_is_disabled(self):
         runnertype_db = RunnerContainerTest.runnertype_db
         runner = get_runner(runnertype_db.runner_module, runnertype_db.runner_module)
 
-        runner.runner_type_db = runnertype_db
-        runner.runner_type_db.enabled = False
+        runner.runner_type = runnertype_db
```
Kami: Same here.

st2common/st2common/runners/base.py
```diff
@@ -172,26 +171,10 @@ def cancel(self):
...
+            callback_module = register_callback_module(self.callback['source'])
             callback_handler = callback_module.get_instance()
-
-            callback_handler.callback(
-                callback_url,
-                self.context,
-                status,
-                result
-            )
+            callback_handler.callback(self.liveaction)
```
Kami: I also prefer older approach here - better decoupling. I'm not a fan of passing whole live action since it contains tons of attributes which could also change in the future.

st2actions/st2actions/container/base.py
m4dcoder: I agree if we go back and make this consistent across the code base.  Going to take me some time before I get back to this.  Maybe we can give this task to the new guy?

st2common/st2common/runners/base.py
m4dcoder: The older approach assumes that the only way to do callback is via a webhook.  The new approach lets the handler plugin decide how to do the callback either via a queue, webhook, etc.

PR #4071, opened 2018-04-06
Add timeout and print extra info to unit tests for action chain pause and resume

m4dcoder:
Unable to reproduce unit test error in the action chain pause and resume. Increase the timeout for the core.local action in the unit tests and also print extra data during assert for troubleshooting.


Kami:
This timeout seems excessive, but will see how it goes on CI.

Kami:
@m4dcoder Same issue, it looks like timeout didn't change it, but at least now the output includes more info - https://travis-ci.org/StackStorm/st2/jobs/365508508#L3725

Edit: Looks like the link doesn't work anymore after re-run :/

Kami:
This failure happened again, now with more context - https://travis-ci.org/StackStorm/st2/jobs/378677516#L3732 (https://gist.github.com/Kami/94ca0e9b98417fe05fc19b7b715166c9)

@m4dcoder Any ideas?

PR #4072, opened 2018-04-09
Add the st2-track-result utility to setup

m4dcoder:
Add an entry for the st2-track-result command line utility into the setup script.

### DIFF
--- a/st2common/setup.py
+++ b/st2common/setup.py
@@ -59,6 +59,7 @@ setup(
         'bin/migrations/v2.1/st2-migrate-datastore-scopes.py',
         'bin/migrations/v2.1/st2-migrate-runners.sh',
         'bin/st2-self-check',
+        'bin/st2-track-result',
         'bin/st2-validate-pack-config',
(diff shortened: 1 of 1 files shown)


Kami:
Good catch, do we also need to have a corresponding st2-packages change or are we fine?

Also, do you consider this as a blocker for v2.7.0 or it's fine if we include it in v2.7.1?

Kami:
@m4dcoder Looks like the same live action race issue test failure - https://travis-ci.org/StackStorm/st2/jobs/364296590

Would be great if you could spend more time digging in when you get a chance, since afaik, that's the only travis intermediate failure we have at the moment.

Kami:
I need to make some other changes to the 2.7.0 release so I will also include this change in new changes which need to make it into v2.7.0.

Kami:
We indeed need corresponding st2-packages change - https://github.com/StackStorm/st2-packages/pull/547, https://github.com/StackStorm/st2-packages/pull/548.

PR #4077, opened 2018-04-10
Friendlier error message if git submodules are not checked out, check out submodules as part of make target

Kami:
Some of our unit tests now rely on git submodules being checked out.

This pull request includes two changes to make that more user (developer) friendly:

* Check out submodules as part of `requirements` make target. This way submodules will always be checked out for the tests which rely on `requirements` target. This change is not explicitly needed for CI systems because CI systems check out submodules by default, but that's not the case with older git versions when user (developer) checks out the git repo manually.

* Throw a more user-friendly exception is user tries to run a test which relies on git submodules, but submodules are not checked out.

Related docs PR https://github.com/StackStorm/st2docs/pull/720.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -275,6 +275,9 @@ requirements: virtualenv .sdist-requirements
 	# new version of requests) which we cant resolve at this moment
 	$(VIRTUALENV_DIR)/bin/pip install "prance==0.6.1"
 
+	# Some of the tests rely on submodule so we need to make sure submodules are check out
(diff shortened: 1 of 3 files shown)


Kami:
@armab Thank you for catching the make part - I didn't add it because CI's check it out by default, but I forgot about the developers (thought readme is enough, but it's not) and it's a good idea in any case to ensure consistent environment and git checkout across different environments / systems.

PR #4081, opened 2018-04-13
Include Python patch version and which Python executable is used when using st2 --version

Kami:
Small change to make the output a bit less confusing.

Before:

```bash
st2 2.8dev, on Python 2.7
```

After:

```bash
st2 2.8dev, on Python 2.7.6 (/data/stanley/virtualenv/bin/python)
```


armab:
I think this makes output even more confusing :)

PR #4085, opened 2018-04-17
Pass --no-download flag to virtualenv binary when we create pack virtualenv

Kami:
This pull request updates the code to pass `--no-download` flag to `virtualenv` binary when we create pack virtual environments.

This way version of pip which is on the system (should be 9.0.1) is used instead of downloading the latest stable version from PyPi (10.0.0).

I think that's a better and safer default behavior since it results in more reproducible pack virtualenvs - without this flag it's possible that virtualenvs which are created after new pip version is released will use this new version.

Having said that - we need to test this change well with the actual build packages to make sure that the correct (9.0.1) pip version is used in pack virtualenvs and not some other ancient pip version which be be installed on RHEL, etc. by default.

If for some reason user wants to change this behavior, they can pass `no_download=False` parameter to `packs.setup_virtualenv` action.

In addition to that, it also includes those changes:

- Update `st2-run-pack-tests` tool so it follows the same behavior and also uses pip 9.0.1 instead of 10.0.0.

PR which contains some details and context https://github.com/StackStorm/st2-packages/pull/549.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,18 @@ Changelog
 in development
 --------------
 
+Changed
+~~~~~~~
+
+* When creating a pack environment during the pack installation, we now pass ``--no-download`` flag
+  to the ``virtualenv`` binary. This way version of pip, wheel and distutils which is already
+  available on the system is used instead of downloading the latest stable versions from PyPi.
+
+  This results in more reproducible pack virtual environments and we also ensure pip 9.0 is used (
+  there are some known issues with pip 10.0).
+
+  If for some reason you want to revert to the old behavior, you can do that by passing
+  ``no_download=False`` parameter to the ``packs.setup_virtualenv`` action. #4085
 
 2.7.0 - April 12, 2018
 ----------------------
--- a/contrib/packs/actions/pack_mgmt/setup_virtualenv.py
+++ b/contrib/packs/actions/pack_mgmt/setup_virtualenv.py
@@ -74,7 +74,7 @@ class SetupVirtualEnvironmentAction(Action):
         if self.proxy_ca_bundle_path and not os.environ.get('proxy_ca_bundle_path', None):
             os.environ['no_proxy'] = self.no_proxy
 
-    def run(self, packs, update=False, python3=False):
+    def run(self, packs, update=False, python3=False, no_download=True):
         """
         :param packs: A list of packs to create the environment for.
         :type: packs: ``list``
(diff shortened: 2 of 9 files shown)


contrib/packs/actions/setup_virtualenv.yaml
```diff
@@ -19,6 +19,11 @@
       description: "Use Python 3 binary when creating a virtualenv for this pack."
       required: false
       default: false
+    no_download:
+      type: "boolean"
+      description: "Use version of pip which is already on the system instead of downloading the latest version from PyPi when creating a virtual environment."
+      required: false
+      default: false
```
armab: Are you sure it should be `false`?

Doesn't it mean we'll create a virtualenv with _upgraded_ & _latest_ pip on `st2 pack install`?

Kami: Yeah, sorry, should indeed default to true - good catch :)

armab:
Also the CHANGELOG is missing

contrib/packs/actions/pack_mgmt/setup_virtualenv.py
```diff
@@ -85,7 +85,8 @@ def run(self, packs, update=False, python3=False):
 
         for pack_name in packs:
             setup_pack_virtualenv(pack_name=pack_name, update=update, logger=self.logger,
-                                  proxy_config=self.proxy_config, use_python3=python3)
+                                  proxy_config=self.proxy_config, use_python3=python3,
+                                  no_download=True)
```
bigmstone: Don't you mean for this to use the `kwarg`?

Kami: Not sure I understand what you mean - I'm indeed passing this new argument to the function as a keyword argument :)

CHANGELOG.rst
```diff
@@ -4,6 +4,18 @@ Changelog
 in development
 --------------
 
+Changed
+~~~~~~~
+
+* When creating a pack environment during the pack installation, we now pass ``--no-download`` flag
+  to the ``virtualenv`` binary. This way version of pip, wheel and distutils which is already
+  available on the system is used instead of downloading the latest stable versions from PyPi.
```
armab: A small change to this would be saying:
> This way version of pip, wheel and distutils which is enforced by virtualenv version is used

Since virtualenv installs "pinned" version he associates with, according to https://virtualenv.pypa.io/en/stable/changes/

armab:
FYI the st2-packages build is back :heavy_check_mark:

PR #4104, opened 2018-05-03
Completing Initial Metrics Integration

### LINKED ISSUE #4102: Incorporate Metrics Collection Across ST2
Now that the metrics framework has been established it needs to be implemented across ST2.

Initial scope:
* Action execution total time by `st2.{pack}.{action}` name
* Number of action executions by `st2.{pack}.{action}` name
* Action execution total time in a specific state by `st2.{pack}.{action}.{state}` name
* Number of action executions in a specific state by `st2.{pack}.{action}.{state}` name
* Rule dispatch count by `st2.{pack}.{rule}` name
* Trigger count by `st2.{pack}.{trigger}` name
* Aggregate counts for the above I.E.: Total number of executions across all actions

bigmstone:
Closes #4102

- [x] Action execution total time by st2.{pack}.{action} name
- [x] Number of action executions by st2.{pack}.{action} name
- [x] Action execution total time in a specific state by st2.{pack}.{action}.{state} name
- [x] Number of action executions in a specific state by st2.{pack}.{action}.{state} name
- [x] Rule dispatch count by st2.{pack}.{rule} name
- [x] Trigger count by st2.{pack}.{trigger} name
- [x] Aggregate counts for the above I.E.: Total number of executions across all actions

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -279,6 +279,9 @@ requirements: virtualenv .sdist-requirements
 	# new version of requests) which we cant resolve at this moment
 	$(VIRTUALENV_DIR)/bin/pip install "prance==0.6.1"
 
+	# Install st2common to register metrics drivers
+	(cd ${ROOT_DIR}/st2common; ${ROOT_DIR}/$(VIRTUALENV_DIR)/bin/python setup.py develop)
+
 	# Some of the tests rely on submodule so we need to make sure submodules are check out
 	git submodule update --init --recursive
 
--- a/st2actions/st2actions/container/base.py
+++ b/st2actions/st2actions/container/base.py
@@ -119,7 +119,7 @@ class RunnerContainer(object):
                                                               action_parameters=action_params)
             extra = {'runner': runner, 'parameters': resolved_action_params}
             LOG.debug('Performing run for runner: %s' % (runner.runner_id), extra=extra)
-            with CounterWithTimer(key=format_metrics_key(action_db)):
+            with CounterWithTimer(key=format_metrics_key(action_db, key='action')):
                 (status, result, context) = runner.run(action_params)
 
             try:
--- a/st2actions/st2actions/container/base.py
+++ b/st2actions/st2actions/container/base.py
@@ -33,6 +33,7 @@ from st2common.util.action_db import (get_action_by_ref, get_runnertype_by_name)
 from st2common.util.action_db import (update_liveaction_status, get_liveaction_by_id)
 from st2common.util import param as param_utils
 from st2common.util.config_loader import ContentPackConfigLoader
+from st2common.metrics.metrics import CounterWithTimer, format_metrics_key
 
 from st2common.runners.base import get_runner
 from st2common.runners.base import AsyncActionRunner, PollingAsyncActionRunner
@@ -89,7 +90,9 @@ class RunnerContainer(object):
 
         return liveaction_db.result
 
-    def _do_run(self, runner, runnertype_db, action_db, liveaction_db):
+    @CounterWithTimer(key="st2.action.executions")
+    def _do_run(self, runner, runnertype_db, action_db, liveaction_db,
+                metrics_counter_with_timer=None):
         # Create a temporary auth token which will be available
         # for the duration of the action execution.
         runner.auth_token = self._create_auth_token(context=runner.context, action_db=action_db,
@@ -116,7 +119,8 @@ class RunnerContainer(object):
                                                               action_parameters=action_params)
             extra = {'runner': runner, 'parameters': resolved_action_params}
             LOG.debug('Performing run for runner: %s' % (runner.runner_id), extra=extra)
-            (status, result, context) = runner.run(action_params)
+            with CounterWithTimer(key=format_metrics_key(action_db)):
+                (status, result, context) = runner.run(action_params)
 
             try:
                 result = json.loads(result)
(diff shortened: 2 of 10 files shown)


nmaludy:
@bigmstone Sorry, don't want to throw this whole thing off but, isn't it possible for there to be an action, rule, and trigger all with the same name?

So, won't these potentially conflict?
- `st2.{pack}.{action}`
- `st2.{pack}.{rule}`
- `st2.{pack}.{trigger}`

st2common/st2common/metrics/base.py
```diff
@@ -24,11 +26,46 @@
...
+    from st2common.config import register_opts
+    register_opts()
+
+LOG = logging.getLogger(__name__)
 
 PLUGIN_NAMESPACE = 'st2common.metrics.driver'
 METRICS = None
 
 
+def format_metrics_key(action_db=None, liveaction_db=None, key=None):
```
Kami: I'm not a big fan of this function definition (it's a bit confusing and hard to understand). Maybe it would make it more clear if it's split into multiple functions (get_metrics_key_for_action_db, get_metrics_key_for_liveaction_db).

st2actions/st2actions/container/base.py
```diff
@@ -116,7 +119,8 @@ def _do_run(self, runner, runnertype_db, action_db, liveaction_db):
                                                               action_parameters=action_params)
             extra = {'runner': runner, 'parameters': resolved_action_params}
             LOG.debug('Performing run for runner: %s' % (runner.runner_id), extra=extra)
-            (status, result, context) = runner.run(action_params)
+            with CounterWithTimer(key=format_metrics_key(action_db, key='action')):
```
Kami: It would be good to use keyword arguments when passing action_db to format_metrics_key function for consistency reasons.

st2actions/st2actions/container/base.py
```diff
@@ -89,7 +90,9 @@ def dispatch(self, liveaction_db):
 
         return liveaction_db.result
 
-    def _do_run(self, runner, runnertype_db, action_db, liveaction_db):
+    @CounterWithTimer(key="st2.action.executions")
+    def _do_run(self, runner, runnertype_db, action_db, liveaction_db,
+                metrics_counter_with_timer=None):
```
Kami: Does the decorator inject `metrics_counter_with_timer` argument into the decorated function?

It looks like it's not used and if that's the case, it might perhaps make sense to add an argument to the decorator with which we can control where to and where not to inject argument to the function (it looks like there is no need to inject it here).

st2actions/st2actions/container/base.py
Kami: And it might even make sense to create another decorator class which inherits from `CounterWithTimer` called `CounterWithTimerForActionDB` and `CounterWithTimerForLiveActionDB` or similar which takes `action_db` directly and generates key for it (this way we don't need to call `format_metrics_key` here since that would happen inside the decorator).

st2common/st2common/util/action_db.py
```diff
@@ -193,6 +194,22 @@ def update_liveaction_status(status=None, result=None, context=None, end_timesta
                          'to unknown status string. Unknown status is "%s"',
                          liveaction_db, status)
 
+    if liveaction_db.status:
```
Kami: Maybe a comment would make it a bit easier to understand what is going on - "decrement the counter for current / previous" status here and below "increment the counter for current / new status" or similar.

bigmstone:
@Kami refactored the key generator. See if you hate it less now.

st2common/st2common/metrics/base.py
```diff
@@ -24,11 +26,49 @@
...
+    register_opts()
+
+LOG = logging.getLogger(__name__)
 
 PLUGIN_NAMESPACE = 'st2common.metrics.driver'
 METRICS = None
 
 
+def _format_metrics_key_for_action_db(action_db):
+    action_name = action_db.name
```
Kami: Yeah, much easier to understand / follow it now :+1:

st2common/st2common/metrics/base.py
```diff
@@ -24,11 +26,49 @@
...
+
+def format_metrics_key(action_db=None, liveaction_db=None, key=None):
+    """Return a string for usage as metrics key.
+    """
+    assert (action_db or key or liveaction_db), """Must supply one of key, action_db, or
+                                                 liveaction_db"""
+    metrics_key = 'st2'
+
+    if action_db:
+        metrics_key += _format_metrics_key_for_action_db(action_db)
```
Kami: One option would also be to return a string without a prefix from each of those utility functions and then join it using a separator (.), but current approach also works.

st2reactor/st2reactor/rules/worker.py
```diff
@@ -62,6 +63,12 @@ def process(self, pre_ack_response):
         if not trigger_instance:
             raise ValueError('No trigger_instance provided for processing.')
 
+        get_driver().inc_counter(
+            format_metrics_key(
+                key='trigger.%s' % trigger_instance.trigger
```
Kami: Minor thing - usually it's safer to wrap string formatting arguments inside parenthesis in case we ever decide to add more format string arguments.

st2common/st2common/services/action.py
```diff
@@ -132,6 +135,12 @@ def create_request(liveaction):
                 trace_service.get_trace_component_for_action_execution(execution, liveaction)
             ])
 
+    get_driver().inc_counter(
+        format_metrics_key(
+            action_db=action_db,
+            key='action.%s' % liveaction.status
```
Kami: Same here.

Makefile
```diff
@@ -279,6 +279,9 @@ requirements: virtualenv .sdist-requirements
 	# new version of requests) which we cant resolve at this moment
 	$(VIRTUALENV_DIR)/bin/pip install "prance==0.6.1"
 
+	# Install st2common to register metrics drivers
+	(cd ${ROOT_DIR}/st2common; ${ROOT_DIR}/$(VIRTUALENV_DIR)/bin/python setup.py develop)
```
Kami: Yeah, ideally we would fine a better way for that in the future...

PR #4111, opened 2018-05-04
Update all the entry points (services, scripts) to use /etc/st2/st2.conf as a default config file location

### LINKED ISSUE #3203: Update all the CLI scripts and tools to try to use st2.conf from the default production path by default
We should update all the command-line scripts and tools to try and use and load config from default production location (`/etc/st2/st2.conf`) by default on invocation.

Right now this doesn't happen so user needs to manually pass ``--config-file /etc/st2/st2.conf`` flag to each of those tools and scripts.

This should save everyone some typing and also improve the user experience - currently in some scenarios if ``--config-file`` argument is not provided logging is not set up which causes confusion.

Kami:
This pull request updates code to use `/etc/st2/st2.conf` as a default config file location for all the services and script entry points.

This will save users some typing and provide a better default user-experience.

Previously, if user ran some script (e.g. `st2-apply-rbac-definitions -v`) without providing ``--config-file`` CLI argument, a weird and non-user friendly exception was thrown. It meant user always needed to run scripts with that option (e.g. `st2-apply-rbac-definitions --config-file /etc/st2/st2.conf`).

This change is fully backward compatible since since it just uses this location as a default value and user can still override the location and use a config from a different path using ``--config-file`` CLI argument.

In addition to that, we should still try to explicitly provide ``--config-file`` argument everywhere where possible since that makes debugging easier and it's more explicit (we already do that in service manager setup files, wsgi entry points, etc.).

Resolves #3203.

## TODO

 - [x] Tests


nmaludy:
Now i have to remember not to pass in `--config-file /etc/st2/st2.conf` (joking).

Nice change! :+1:

PR #4113, opened 2018-05-05
Added issue template, update README

LindsayHill:
First draft of an issue template. New issues will be pre-populated with this content.

The goal is to (A) ensure users provide all relevant information, and (B) direct questions towards forum.stackstorm.com


Kami:
LGTM - maybe we should also mention the new forum somewhere and point people to use that instead of Github issues when it makes sense?

LindsayHill:
We mention the forum on our website & blog. I guess the other places here would be in our README, and the project description? I'll add something to the README fo this PR

LindsayHill:
I've tweaked the project description, and updated README to remove IRC link, add Forum badge & link, and update a few docs URLs to use HTTPS

LindsayHill:
5 updates from master & multiple CI re-runs later, and it's done...

PR #4116, opened 2018-05-08
Add CII Best Practices badge

dzimine:
We match and earn it. See here. 
https://bestpractices.coreinfrastructure.org/en/projects/1833


LindsayHill:
Maybe we combine with this PR since it's updating the same line in README? https://github.com/StackStorm/st2/pull/4113

dzimine:
Agreed, closing this.

PR #4118, opened 2018-05-09
Fix for a regression which broke sensors which rely on select.poll() functionality

### LINKED ISSUE #4094: Sensors have stopped working after upgrade to st2 2.7.1
Hi Stackstorm Dev Team,

Sensors have stopped working in st2 2.7.1 after upgrade from st2 2.6.0

It was giving the error : module "object" has no attribute "poll" whenever the sensor start to initialize. This error was seen in st2sensorcontainer.log

Even the default linux.FileWatchSensor threw errors during sensor initialization when we bring the st2sensorcontainer service

Kindly please look into this issue

Regards
Harish

Kami:
## Background

In v2.7.0 we upgraded eventlet to the latest stable version. That stable version removed `select.poll()` which was available in previous releases (but it was blocking) to make it harder for users to "accidentally" use blocking code.

That change broke some sensors which rely on `select.poll()` functionality.

Users would see an error similar to one below:

```bash
2018-05-09 07:57:29,425 ERROR [-] AttributeError
2018-05-09 07:57:29,425 ERROR [-] : 
2018-05-09 07:57:29,425 ERROR [-] 'module' object has no attribute 'poll'
2018-05-09 07:57:29,425 ERROR [-] 
```

## Proposed solution

This PR adds `select` module which includes blocking version of `poll()` back to `sys.modules` so it can be used by the sensors which need it.

The behavior will now be the same as in releases prior to v2.7.0. Blocking code is not great and ideally we could avoid it where we can, but sadly we have no control over 3rd party libraries and integrations which use that blocking code construct.

Keep in mind, that each sensor is running in a separate and dedicated process so blocking inside that process it not that big of deal (compared to blocking in a main service process which would be a no go).

Related issues #3968 #4007 #4094 

Resolves #4094

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,6 +25,11 @@ Fixed
 
   Note: This issue only affects users who utilize RBAC with remote LDAP groups to local RBAC
   roles synchronization feature enabled. (bug fix) #4103 #4105
+* Fix an issue with some sensors which rely on ``select.poll()`` (FileWatch, GithubSensor, etc.)
+  stopped working with StackStorm >= 2.7.0.
+
+  StackStorm v2.7.0 inadvertently introduced a change which broke a small set of
+  sensors which rely on ``select.poll()`` functionality. (bug fix)
 
 2.7.1 - April 20, 2018
 ----------------------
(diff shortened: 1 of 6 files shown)


schreyack:
I can confirm that these changes fix the issue for the git sensor.

Kami:
@schreyack Great - thanks for confirming :)

PR #4122, opened 2018-05-10
Fix for #4121 - Rendering nested lists in pack configs

### LINKED ISSUE #4121: The secret value in email config file seems not work
I use the following cmd to set a key in the datastore.
`$ st2 key set -e smtp_account.me.password "I'm Encrypted"`

and here are my email pack config /opt/stackstorm/configs/email.yaml

```
---
smtp_accounts:
  - name: name
    password: "{{ st2kv.system.smtp_account.me.password }}"
```

then run the email action but failed
```
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/runners/python_action_wrapper.py", line 278, in <module>
    obj.run()
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/runners/python_action_wrapper.py", line 171, in run
  File "/opt/stackstorm/packs/email/actions/send_email.py", line 49, in run
    s.login(account_data['username'], account_data['password'])
  File "/usr/lib64/python2.7/smtplib.py", line 621, in login
    raise SMTPAuthenticationError(code, resp)
smtplib.SMTPAuthenticationError: (460, 'ERR.LOGIN.PASSERR')
```

I added some log to see the password value in File "/opt/stackstorm/packs/email/actions/send_email.py", line 49.
It seems that we get
{{ st2kv.system.smtp_account.me.password }} as the password.

do i missed something?
FYI
If I run the send email action and put the {{ st2kv.system.smtp_account.me.password }} as the message
I can see the message's value is correct.

nmaludy:
In the ConfigLoader when traversing the config object structure, `dicts` were taken into account and recursed properly, however `lists` were not considered.

This PR recognizes lists in the recursion and takes steps to render the values within them, potentially recursing further into each item if necessary.

Closes #4121

### DIFF
--- a/st2common/st2common/util/config_loader.py
+++ b/st2common/st2common/util/config_loader.py
@@ -110,14 +110,31 @@ class ContentPackConfigLoader(object):
         """
         parent_keys = parent_keys or []
 
-        for config_item_key, config_item_value in six.iteritems(config):
-            schema_item = schema.get(config_item_key, {})
+        config_is_dict = isinstance(config, dict)
+        config_is_list = isinstance(config, list)
+        iterator = six.iteritems(config) if config_is_dict else enumerate(config)
+
+        for config_item_key, config_item_value in iterator:
+            if config_is_dict:
+                # different schema for each key/value pair
+                schema_item = schema.get(config_item_key, {})
+            if config_is_list:
+                # same schema is shared between every item in the list
+                schema_item = schema
+
             is_dictionary = isinstance(config_item_value, dict)
+            is_list = isinstance(config_item_value, list)
 
             # Inspect nested object properties
             if is_dictionary:
-                parent_keys += [config_item_key]
+                parent_keys += [str(config_item_key)]
                 self._assign_dynamic_config_values(schema=schema_item.get('properties', {}),
+                                                   config=config[config_item_key],
+                                                   parent_keys=parent_keys)
+            # Inspect nested list items
+            elif is_list:
+                parent_keys += [str(config_item_key)]
+                self._assign_dynamic_config_values(schema=schema_item.get('items', {}),
                                                    config=config[config_item_key],
                                                    parent_keys=parent_keys)
             else:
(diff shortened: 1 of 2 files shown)


nmaludy:
While i was at it, i also fixed #4123 so that a good (known) exception type is thrown, so a meaningful error can be shown to the user.

Closes #4123

st2common/st2common/util/config_loader.py
```diff
@@ -110,22 +110,42 @@ def _assign_dynamic_config_values(self, schema, config, parent_keys=None):
...
+        # config_item_key - if config_is_dict then this is the key in the dictionary
+        #                   if config_is_list then this is the index of them item
+        # config_item_value - the value of the key/index for the current item
+        for config_item_key, config_item_value in iterator:
+            if config_is_dict:
+                # different schema for each key/value pair
+                schema_item = schema.get(config_item_key, {})
+            if config_is_list:
+                # same schema is shared between every item in the list
+                schema_item = schema
```
Kami: Does this work as expected?

For arrays / lists you can specify optional schema for list items using `items` schema property (e.g. https://spacetelescope.github.io/understanding-json-schema/reference/array.html).

If I'm not mistaken, we should probably use `schema.get('items', {})` or similar here, but would need to double check and verify it.

Kami: Edit: Please ignore my comment.

Just now I see it's correctly handled a couple of lines below (https://github.com/StackStorm/st2/pull/4122/files#diff-1b5471e6a12ebd2f0617b57f18a3c425R140) :)

st2common/tests/unit/test_config_loader.py
```diff
@@ -328,6 +332,199 @@ def test_get_config_dynamic_config_item_render_fails_user_friendly_exception_is_
...
+    def test_get_config_dynamic_config_item_list(self):
+        pack_name = 'dummy_pack_schema_with_nested_object_7'
+        loader = ContentPackConfigLoader(pack_name=pack_name)
+
+        KeyValuePair.add_or_update(KeyValuePairDB(name='k0', value='v0'))
+        KeyValuePair.add_or_update(KeyValuePairDB(name='k1', value='v1'))
+
+        ####################
+        # values in list
+        values = {
```
Kami: Ideally we would also have some tests which verify schema is correctly handled for arrays items (aka schema which defines items.type or similar).

nmaludy: I agree. As a stop-gap i was trying to at least include some tests to cover various structural corner cases in the config.

Kami: I will merge this, but if you get a chance it would be great if you can open a PR with tests for that scenario :)

PR #4129, opened 2018-05-14
Add missing pip 10 fix to runners

enykeev:
We've apparently missed updating dist_utils.py files in runners the same way we do it in components. This resulted in #4069 fix not propagating there and as a result, serverless plugin being broken in pip 10 environments.

### DIFF
--- a/contrib/runners/action_chain_runner/dist_utils.py
+++ b/contrib/runners/action_chain_runner/dist_utils.py
@@ -17,14 +17,51 @@
 from __future__ import absolute_import
 import os
 import re
+import sys
 
-from pip.req import parse_requirements
+from distutils.version import StrictVersion
+
+GET_PIP = 'curl https://bootstrap.pypa.io/get-pip.py | python'
+
+try:
+    import pip
+    from pip import __version__ as pip_version
+except ImportError as e:
+    print('Failed to import pip: %s' % (str(e)))
+    print('')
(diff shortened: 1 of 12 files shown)


Kami:
Thanks, good catch :thumbsup:

Kami:
I will also cherry pick this into branch for v2.7.2 changes (it's not strictly needed, but it doesn't hurt either).

contrib/runners/action_chain_runner/dist_utils.py
```diff
@@ -17,16 +17,53 @@
 from __future__ import absolute_import
 import os
 import re
+import sys
 
-from pip.req import parse_requirements
+from distutils.version import StrictVersion
+
+GET_PIP = 'curl https://bootstrap.pypa.io/get-pip.py | python'
```
LindsayHill: This line makes me nervous. What happens if this is run on a system that does not have Internet access?

Kami: We just advise user to run that, they can also install pip using other means.

Kami: (and this only affects custom and advanced users which install StackStorm directly from sources and / or on an unsupported systems)

It doesn't affect users who install it using the official packages and official installer script.

LindsayHill: Ah good, it's not a problem then.

Issue #4139, opened 2018-05-22
Secret values on object in an array are not masked inside a pack config.


NikosVlagoidis:
##### ISSUE TYPE

 - Bug Report *
 - Feature Idea
 - Other

##### STACKSTORM VERSION

```
st2 2.8dev (0f1d9dc), on Python 2.7.6
```

##### OS / ENVIRONMENT / INSTALL METHOD

docker instalation

##### SUMMARY

When I mark a secret value on a pack config inside an array I get the values in plain text even if Ii don't make the request with parameter show_secret=true.

##### STEPS TO REPRODUCE

Pack config

```yaml
---
  instance:
    type: "array"
    required: false
    items:
      type: "object"
      properties:
          alias:
              description: "an alias for this instance"
              type: "string"
              required: true
              secret: false
          base_url:
              description: "Base URL for a service"
              type: "string"
              secret: false
              required: true
          secret:
              description: "The api secret key"
              type: "string"
              secret: true
              required: true
```

when  I make a request on : https://{{host}}/api/v1/configs/<pack_name>

```
{
    "values": {
        "instance": [
            {
                "alias": "alias1",
                "secret": "secret",
                "base_url": "10.10.10.1"
            },
            {
                "alias": "alias2",
                "secret": "secret23",
                "base_url": "10.10.10.2"
            }
        ]
    },
    "id": "5afadbb2b2c724049ee7e03c",
    "pack": "<pack_name>"
}
```

##### EXPECTED RESULTS

```
{
    "values": {
        "instance": [
            {
                "alias": "alias1",
                "secret": "****",
                "base_url": "10.10.10.1"
            },
            {
                "alias": "alias2",
                "secret": "*****",
                "base_url": "10.10.10.2"
            }
        ]
    },
    "id": "5afadbb2b2c724049ee7e03c",
    "pack": "<pack_name>"
}
```

##### ACTUAL RESULTS

```
{
    "values": {
        "instance": [
            {
                "alias": "alias1",
                "secret": "secret",
                "base_url": "10.10.10.1"
            },
            {
                "alias": "alias2",
                "secret": "secret23",
                "base_url": "10.10.10.2"
            }
        ]
    },
    "id": "5afadbb2b2c724049ee7e03c",
    "pack": "<pack_name>"
}
```

nmaludy:
Probably related to https://github.com/StackStorm/st2/pull/4122

Wonder if the secret masking process for the CLI / API needs the same modifications?

[...]

PR #4140, opened 2018-05-22
[WIP] Fix for #4139 - Secrets not being masked in pack config

nmaludy:
Fix for #4139 

This adds recursion into the secret masking process so that configs with nested `dicts` and `lists` will also have their secrets masked out for API calls.

**TODO**
- [] Unit tests
- [] Test various combinations of dicts/lists/other
- [] Find other edge cases

### DIFF
--- a/st2common/st2common/util/secrets.py
+++ b/st2common/st2common/util/secrets.py
@@ -27,16 +27,65 @@ from st2common.constants.secrets import MASKED_ATTRIBUTE_VALUE
 
 def get_secret_parameters(parameters):
     """
-    Filter the provided parameters dict and return a list of parameter names which are marked as
-    secret.
+    Filter the provided parameters dict and return a dict of parameters which are marked as
+    secret. Every key in the dict is the parameter name and values are the parameter type:
+
+    >>> d = get_secret_parameters(params)
+    >>> d
+    {
+        "param_a": "string",
+        "param_b": "boolean",
+        "param_c": "integer"
+    }
+
+    If a paramter is a dictionary or a list, then the value will be a nested dictionary
+    containing information about that sub-object:
+
+    >>> d = get_secret_parameters(params)
+    >>> d
+    {
+        "param_dict": {
+            "nested_a": "boolean",
+            "nested_b": "string",
+        },
+        "param_list": {
+            "nested_dict: {
+              "param_c": "integer"
+            }
+        }
+    }
+
+    Note: in JSON Schema, we're assuming lists contain the same data type for every element
+
 
     :param parameters: Dictionary with runner or action parameters schema specification.
     :type parameters: ``dict``
 
     :rtype ``list``
     """
-    secret_parameters = [parameter for parameter, options in
-                         six.iteritems(parameters) if options.get('secret', False)]
+
+    # determine if this parameters set is an object definition
+    # if it is, then drill in and grab the properties from the object itself
+    parameters_type = parameters.get('type')
+    if parameters_type == 'object':
+        parameters = parameters.get('properties', {})
+    elif parameters_type == 'array':
+        parameters = parameters.get('items', {})
+
+    # iterate over all of the parameters recursively
+    secret_parameters = {}
+    for parameter, options in six.iteritems(parameters):
+        # if parameter is a dict or a list, then we need to recurse into them
+        parameter_type = options.get('type')
+        if parameter_type == 'object':
+            sub_params = get_secret_parameters(options.get('properties', {}))
+            secret_parameters[parameter] = sub_params
+        elif parameter_type == 'array':
+            sub_params = get_secret_parameters(options.get('items', {}))
+            secret_parameters[parameter] = sub_params
+        elif options.get('secret', False):
+            # if this parameter is secret, then add it our secret parameters
+            secret_parameters[parameter] = parameter_type
 
     return secret_parameters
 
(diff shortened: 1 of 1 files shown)


nmaludy:
@Kami Here's an initial workup of a fix. Initial testing looks good (used the email pack schema as a reference https://github.com/StackStorm-Exchange/stackstorm-email/blob/master/config.schema.yaml).

I still need to develop test cases, but would appreciate comments early if you have time.

Kami:
Thanks, I will have a look tomorrow.

And yeah, I agree - we need a good and thorough test coverage for that.

Another thing we need to keep in mind is to try to make this function as performance efficient as possible (bail early if there are no secret params, only iterate over items which could contain secret params based on the schema, etc.).

If we don't do that and just blindly recursively iterate over nested objects, it has a potential to significantly slow API responses for all the endpoints with deeply nested objects (executions, etc.).

nmaludy:
An idea of how to speed this up (theoretically) we will need to measure to determine if it's actually faster:

In the current implementation we call `get_secret_parameters()` which fully traverses the entire JSON schema looking for the `secret` property. Next we call `mask_secret_parameters` which traverses the object, but only for the `secret_parameters` passed in.

To potentially speed this up, we could potentially skip traversing parts of the JSON schema if the input data is missing those properties. 

Example: if we have a schema with a nested dictionary

```
{
    'arg_string': {
        'description': 'Junk',
        'type': 'string',
    }
    'arg_optional_object': {
        'description': 'Mirror',
        'type': 'object',
        'properties': {
            'arg_nested_object': {
                'description': 'Mirror mirror',
                'type': 'object',
                'properties': {
                    'arg_double_nested_secret': {
                        'description': 'Deep, deep down',
                        'type': 'string',
                        'secret': True
                    }
                }
            },
            'arg_nested_secret': {
                'description': 'Deep down',
                'type': 'string',
                'secret': True
            }
        }
    }
}
```

And we pass in data:

```
{'arg_string': 'test'}
```

Then we can skip traversing down into the `arg_optional_object` schema since the input data doesn't have this property, saving us time/cycles.

nmaludy:
Additional thing to look at is the `copy.deepcopy()` when calling this recursively we can end up using a ton of memory. Need to figure out a memory-efficient way of doing `mask_secret_parameters()`

Kami:
As far as deep copy goes - yeah, ideally we would create a deep copy of the original object just once and then modify it in subsequent recursive calls (instead of creating a copy during each recursion which as you have said, is problematic).

And in general, I like to avoid recursion where possible because of issues like that, but here it should be fine, if we only create a deep copy once and manipulate that original object in place in subsequent recursive calls.

nmaludy:
@Kami Think this is ready for a review, all unit tests are passing on my local box and with the fix detailed below they should pass in travis.

FYI i ran into a "random failure" in the `st2client/tests/unit/test_inquiry.py` suite where random UUIds were getting generated and causing the `count()` function to return too many values in the CLI list output.

st2client/tests/unit/test_inquiry.py
```diff
@@ -32,6 +32,11 @@
 def _randomize_inquiry_id(inquiry):
     newinquiry = copy.deepcopy(inquiry)
     newinquiry['id'] = str(uuid.uuid4())
+    # ID can't have '1440' in it, otherwise our `count()` fails
+    # when inspecting the inquiry list output for test:
+    # test_list_inquiries_limit()
+    while '1440' in newinquiry['id']:
+        newinquiry['id'] = str(uuid.uuid4())
```
Kami: Thanks for fixing this intermediate test failure :+1:

st2common/st2common/util/secrets.py
```diff
@@ -27,36 +27,146 @@
...
+    # how we iterate depends on what data type was passed in
+    iterator = None
+    is_dict = isinstance(secret_parameters, dict)
+    is_list = isinstance(secret_parameters, list)
+    if is_dict:
+        iterator = six.iteritems(secret_parameters)
+    elif is_list:
+        iterator = enumerate(secret_parameters)
+    else:
+        return MASKED_ATTRIBUTE_VALUE
```
Kami: That's fine or we could also throw because an unsupported type was passed it :)

st2common/st2common/util/secrets.py
```diff
@@ -27,36 +27,146 @@
...
+
+    # only create a deep copy of parameters on the first call
+    # all other recursive calls pass back referneces to this result object
+    # so we can reuse it, saving memory and CPU cycles
+    if not result:
```
Kami: Maybe it's safer to do a more explicit check `if result is None`? Is there a possible scenario where empty dict / array could be returned during recursion?

st2common/tests/unit/test_util_secrets.py
```diff
@@ -0,0 +1,671 @@
...
+                },
+                {
+                    'arg_nested_secret': MASKED_ATTRIBUTE_VALUE
+                },
+                {
+                    'arg_nested_secret': MASKED_ATTRIBUTE_VALUE
+                }
+            ]
+        }
+        self.assertEqual(expected, result)
```
Kami: Nice work on the extensive tests :+1:

Kami:
Can you please add a corresponding changelog entry (and please try to be explicit and describe scenario where this could occur)?

Besides that, LGTM :+1:

Kami:
The fix for PY3 test failure is in my PR - #4134.

I hope to finish and merge it today.

nmaludy:
OK, i've updated the changelog and converted that one line to `if result is None`. All checks have passed, we should be good to go.

CHANGELOG.rst
```diff
@@ -42,6 +42,18 @@ Changed
 Fixed
 ~~~~~
 
+* Fixed a bug where secrets in pack configs weren't being masked. Recently we
```
Kami: Minor thing - we use 100 characters wide lines. Not a big deal and this can be changed in a subsequent PR.

PR #4146, opened 2018-05-25
Metrics Fixup

bigmstone:
Original metrics PR had some missing pieces. Chiefly incorrect naming generated by the metrics formatter. Wrote some tests to cover that from happening again. Also had to move decorator to the context manager for `_do_run` because we're accessing it by placing it into a dictionary and apparently that has some interesting side effects on decorators that I'll be interested to explore when I have some more time.

### DIFF
--- a/st2common/st2common/metrics/base.py
+++ b/st2common/st2common/metrics/base.py
@@ -40,7 +40,7 @@ def _strip_pack(action, pack):
     formatted_pack = "%s." % (pack)
 
     if formatted_pack in action:
-        return action.strip(formatted_pack)
+        return action.replace(formatted_pack, '')
 
     return action
 
--- a/st2common/st2common/metrics/base.py
+++ b/st2common/st2common/metrics/base.py
@@ -36,16 +36,25 @@ PLUGIN_NAMESPACE = 'st2common.metrics.driver'
 METRICS = None
 
 
+def _strip_pack(action, pack):
+    formatted_pack = "%s." % (pack)
+
+    if formatted_pack in action:
+        return action.strip(formatted_pack)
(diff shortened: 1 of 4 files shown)


st2common/st2common/metrics/base.py
```diff
@@ -36,16 +36,25 @@
...
+    action_name = _strip_pack(action_db.name, action_pack)
+    return [action_pack, action_name]
 
 
 def _format_metrics_key_for_liveaction_db(liveaction_db):
-    action_name = liveaction_db.action
     action_pack = liveaction_db.context.get('pack', 'unknown')
-    return '.%s.%s' % (action_pack, action_name)
+    action_name = _strip_pack(liveaction_db.action, action_pack)
+    return [action_pack, action_name]
```
Kami: Yeah, I like returning a list an joining in separator approach more :thumbsup:

st2common/tests/unit/test_metrics.py
```diff
@@ -337,3 +339,55 @@ def _get_tested(metrics_timer=None):
...
         )
+
+
+class TestFormatMetrics(unittest2.TestCase):
+    def test_format_metrics_liveaction_db_without_key(self):
+        pack = 'test'
+        action = 'lakface'
+
+        liveaction_db = MagicMock()
+        liveaction_db.context = {'pack': pack}
```
Kami: Would also be good to test edge cases aka when pack is not set and we default to unknown, etc.

Besides that, LGTM

bigmstone: Good call, forgot about that branch. Added.

PR #4154, opened 2018-05-31
Implement the Orchestra runner

m4dcoder:
The Orchestra runner and related workflow service is the new workflow engine for StackStorm. This new workflow engine and the new workflow DSL that accompanies the new engine will replace Mistral in StackStorm. For this beta release, Orchestra will support the following features.
- Run, pause, resume, and cancel of workflows and subworkflows.
- Sequential workflows and workflows with cycles, branches, forks, and joins.
- YAQL and Jinja expressions in workflow input, vars, task (action, input, when, publish, do), & output
- Multipoint inspection of syntax, semantics, context, and expressions on the workflow definition in a single pass.
- Improved runtime traceability and error handling.


Kami:
As this is a new feature, let's please make sure all the new and affected tests also run and pass under Python 3.

Kami:
@m4dcoder Here is an example of how we do it inside st2 repo using tox - https://github.com/StackStorm/st2/blob/master/tox.ini#L29.

Makefile
```diff
@@ -429,6 +429,26 @@ mistral-itests: requirements .mistral-itests
 	. $(VIRTUALENV_DIR)/bin/activate; nosetests $(NOSE_OPTS) -s -v --with-coverage \
 		--cover-inclusive --cover-html st2tests/integration/mistral || exit 1;
 
+.PHONY: orchestra-itests
```
bigmstone: What's the compelling reason to keep this separate from itests since this will be a formal part of st2? The requirement for st2 to be running?

st2actions/st2actions/workflows/workflows.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+class WorkflowDispatcher(consumers.MessageHandler):
+    message_type = wf_db_models.WorkflowExecutionDB
+
+    def __init__(self, connection, queues):
+        super(WorkflowDispatcher, self).__init__(connection, queues)
+
+    def get_queue_consumer(self, connection, queues):
+        # We want to use a special ActionsQueueConsumer which uses 2 dispatcher pools
```
bigmstone: Need updating now?

m4dcoder: That's a good question because the previous one was specifically for ActionRunners.  I caught this while troubleshooting the Xenial issue.  The use of the normal QueueConsumer seems to work as well.

Makefile
m4dcoder: There is a separate set of unit tests for orchestra that runs w/ the rest of st2 make tests.  The integration tests is a sanity check to make sure orchestra really does work.  I kept this separate because it was easy for me to bring up launchdev and then run all the orchestra tests.  We should at some point move this into the CI.  I do like the use of python to write the integration tests though cause it gave me an easy way to code for certain condition.

Issue #4153, opened 2018-05-30
Regression of #3820 - Jinja is not rendered in action default values


nmaludy:
##### ISSUE TYPE
 - Bug Report

##### STACKSTORM VERSION
```
# st2 --version
st2 2.7.2, on Python 2.7
```

##### OS / ENVIRONMENT / INSTALL METHOD
```
Install method = puppet-st2
OS = Red Hat Enterprise Linux Server release 7.5 (Maipo)
```

##### SUMMARY

`array` and `object` parameters are not being rendered in action metadata parameter defaults. `integer` parameters work fine.

###### 1) create a new action metadata file

`/opt/stackstorm/packs/default/actions/render_test.yaml`

```yaml
---
description: Run a local linux command
enabled: true
runner_type: mistral-v2
entry_point: workflows/render_test.yaml
name: render_test
pack: default
parameters:
  cmd:
    required: true
    type: string
  timeout:
    type: integer
    default: 60
  kv_array:
    type: array
    default: "{{ st2kv.system.kv_array | from_json_string }}"
  kv_object:
    type: object
    default: "{{ st2kv.system.kv_object | from_json_string }}"
```

###### 2) create a new workflow

`/opt/stackstorm/packs/default/actions/workflows/render_test.yaml`

```yaml
version: '2.0'

default.render_test:
  description: demo rendering failures
  type: direct
  input:
    - cmd
    - timeout
    - kv_array
    - kv_object
      
  tasks:
    task1:
      action: core.local
      input:
        cmd: "{{ _.cmd }}"
```

###### 3) assign values in the datastore

```sh
$ st2 key set kv_array '["a", "b", "c"]'
$ st2 key set kv_object '{"a": "value", "b": "value2", "c": "value3"}'
```

###### 4) register and run the action

```sh
$ st2 action create /opt/stackstorm/packs/default/actions/render_test.yaml
$ st2 run default.render_test cmd="ls"
```

##### EXPECTED RESULTS

Action to execute successfully with parameters:

```yaml
cmd: ls
timeout: 60
kv_array:
  - a
  - b
  - c
kv_object:
  a: value
  b: value2
  c: value3
```

##### ACTUAL RESULTS
```
$ st2 run default.render_test cmd="ls"
ERROR: 400 Client Error: Bad Request
MESSAGE: '{{ st2kv.system.kv_array | from_json_string }}' is not of type 'array' for url: http://127.0.0.1:9101/executions
```

[...]

PR #4149, opened 2018-05-28
Experimental and opt-in Python 3 support for Python runner actions

Kami:
This pull request adds experimental and opt-in support for using Python 3 for pack actions.

It continues work from other PRs - https://github.com/StackStorm/st2/pull/3922, https://github.com/StackStorm/st2/pull/4016, https://github.com/StackStorm/st2/pull/4075.

Our code changes in the past were a bit too optimistic and it turned out a lot more work is still needed to get that feature working.

The scope and target audience for this feature didn't change - it's still intended for advanced users and it's opt in - user needs to pass in ``--python3`` flag to ``st2 pack install <pack name>`` CLI command and Python 3 binary will be used when creating virtual environment for that pack and for all the subsequent pack action executions.

As mentioned before, this feature is mostly meant as an experimental stop gap for advanced users until the whole platform supports Python 3 (aka StackStorm itself can run on Python 3).

We don't have any specific ETA for Python 3 support for the whole platform, but we are slowly and incrementally working towards it - every time I touch some code, I try to also make it work under Python 3 and and make sure tests for that code also run under Python 3 on Travis.

Once full platform support is there, this feature will become obsolete.

## Included Changes

To be able to make that feature work, I needed to make the following changes.

1. Explicitly install `pyyaml` into virtual environment created for a pack. `pyyaml` is one of those libraries which has 2 different code bases for Python 2 and 3 (not ideal, but it's the most widely used yaml parsing / serializing library for Python).

The version which is installed on the system and used by StackStorm components / services uses Python 2, but when user opts in to use Python 3 for a pack, we need to install version which is compatible with Python 3 into that pack virtual environment.

2. Re-order and add new members to `PYTHONPATH` environment variable used python runner actions. If pack is using Python 3 and `/opt/stackstorm/virtualenvs/<pack name>/lib/python3.x/site-packages` directory exists, we add it to the beginning of `PYTHONPATH` to ensure pack uses `pyyaml` and other libraries from that pack virtual environment (and so it doesn't inherit and use system ones which don't work with Python 3).

We also explicitly add `/opt/stackstorm/pack/actions/lib/` directory to PYTHONPATH when Python 3 is used - that is done implicitly and not needed when using Python 2, but it's needed when using Python 3. 

### 1. Not using Python 3 (default behavior)

```bash
$ st2 run packs.setup_virtualenv packs=examples             
2018-05-28 13:57:14,306  WARNING - Auth API server is not available, skipping authentication.
....
[...]

PR #4155, opened 2018-05-31
Improvements to the rendering of default parameter values which contain Jinja expression and filters

Kami:
This pull request makes the following example work:

```yaml
---
description: Run a local linux command
enabled: true
runner_type: mistral-v2
entry_point: workflows/render_test.yaml
name: render_test
pack: default
parameters:
  cmd:
    required: true
    type: string
  timeout:
    type: integer
    default: 60
  kv_array:
    type: array
    default: "{{ st2kv.system.kv_array | from_json_string }}"
  kv_object:
    type: object
    default: "{{ st2kv.system.kv_object | from_json_string }}"
```

As mentioned in #4153, `from_json_string` filter cast call is not really needed since the value is already automatically cast based on the parameter type definition, but it doesn't hurt either.

This way both approaches are supported (one without a cast and one without explicit cast).

## TODO

- [x] Tests

### DIFF
--- a/st2common/st2common/jinja/filters/data.py
+++ b/st2common/st2common/jinja/filters/data.py
@@ -14,8 +14,10 @@
 # limitations under the License.
 
 from __future__ import absolute_import
+
 import json
 import yaml
+import six
 
 __all__ = [
     'from_json_string',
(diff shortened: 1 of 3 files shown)


Kami:
Circle CI build and all the checks were passing yesterday (same for #4149), but after syncing with latest master, Circle CI check started failing.

No idea why yet, it looks like services take long to start up or similar and "services listening" checks are failing. Everything looks fine locally though.

PR #4158, opened 2018-06-01
Fix sharing st2 logs when CircleCI task failed

armab:
When CircleCI Tests task failed it exits immediately, so the command to grab the st2 logs from Docker container didn't run.

Fix this behavior, by running the "grab the st2 logs" command with `when: always`.

This helps in debugging.


armab:
Thanks for merging :+1: 

I already forgot about CI issues from the last week :)

PR #4165, opened 2018-06-06
Move from keyczar to cryptography library for symmetric encryption / decryption

### LINKED ISSUE #3974: Switch from keyczar to cryptography library for cryptographic operations (Python 3 support)
Right now we use python-keyczar library for cryptographic operations.

The problem is that the library doesn't support Python 3. There is some unofficial fork / version on PyPi with Python 3 support, but since this is a library which handles critical and sensitive cryptographic operations it would be very irresponsible to just use some "random" non-audited fork.

Kami:
This pull request moves our code for handling symmetric AES encryption and decryption from `keyczar` to `cryptography` library.

The change is fully backward and forward compatible as it's an implementation detail so it should in no way affect end users (and they shouldn't notice it either).

### Background

When python-keyczar library was released, it was one of the best and most user-friendly crypto library with sane and safe defaults in the Python world.

Over the years the crypto world moved on (one of the notable moves forward was development of the cryptography library which was meant to be a more portable and safer / saner higher level replacement for PyCrypto, M2Crypto and other crypto libraries in the Python world).

Python-keyczar is still a relatively safe and good choice in the Python world, but one of the biggest problem is that it's not actively maintained / developed anymore and it doesn't support Python 3.

There are also some minor weakness in keyczar library, but nothing really major.

For us, the main reason for moving away from keyczar to cryptography is Python 3 support and active development and maintenance of `cryptography` library.

### Implementation

The main goal behind this change was to make our crypto code work under Python 3 and to make it fully backward compatible aka don't require any user interaction / migration.

To be able to do that, I needed to re-implement keyczar AES `Encrypt()` and `Decrypt()` methods, but use the primitives from the cryptography library.

I'm never a fan or proponent of "reimplementing" crypto stuff, but in this case it was mostly limited to simple stuff such as safely generating IV string (using safe and recommended `os.urandom`), generating HMAC signature for the message and serializing the encrypted message (and vice-versa for decryption).

Even though that code is relatively simple and actual hard part such as encryption + signing is outsourced to cryptography, please perform a very thorough review / audit of the code to make sure I'm not doing anything stupid / unsafe.

In addition to that, I also needed to implement the code which knows how to read and handle JSON key files generated by keyczar which contain AES key and HMAC key.

While working on that, I also made a small change / improvement - I changed the default size for the AES and HMAC keys we generate from 128 to 256 bits.

### Future Improvements

As mentioned above, re-implementing those two methods is not ideal.

In the future when we have a plan for a seamless migration in place, we should move to Fernet recipe in the cryptography library for handling symmetric encryption and decryption (https://cryptography.io/en/latest/fernet/).

Underneath, that recipe actually implements quite a similar algorithm to one used by keyczar (AES in CBC mode with SHA1 HMAC in keyczar case and AES in CBC mode with SHA256 HMAC in cryptography case), but it's better audited and maintained than our code (and that's always important when dealing with crypto / security sensitive code).

Keep in mind that the "seamless" migration in the future will at the very least require generating new secret key which will be used by cryptography fernet recipe.

With that change, we now have 90% or so of our unit tests running and passing under Python 3.6.

Resolves #3974

### DIFF
--- a/st2common/st2common/util/crypto.py
+++ b/st2common/st2common/util/crypto.py
@@ -60,7 +60,9 @@ __all__ = [
 
     # NOTE: Keyczar functions are here for testing reasons - they are only used by tests
     'keyczar_symmetric_encrypt',
-    'keyczar_symmetric_decrypt'
+    'keyczar_symmetric_decrypt',
+
+    'AESKey'
 ]
 
 # Keyczar related constants
@@ -84,11 +86,34 @@ class AESKey(object):
         self.aes_key_string = aes_key_string
         self.hmac_key_string = hmac_key_string
         self.hmac_key_size = hmac_key_size
-        self.mode = mode
-        self.size = size
+        self.mode = mode.upper()
+        self.size = int(size)
 
         self.hmac_key_bytes = Base64WSDecode(self.hmac_key_string)
         self.aes_key_bytes = Base64WSDecode(self.aes_key_string)
+
+    @classmethod
+    def generate(self, key_size=256):
+        aes_key_bytes = os.urandom(key_size / 8)
+        aes_key_string = Base64WSEncode(aes_key_bytes)
+
+        hmac_key_bytes = os.urandom(key_size / 8)
+        hmac_key_string = Base64WSEncode(hmac_key_bytes)
+
+        return AESKey(aes_key_string=aes_key_string, hmac_key_string=hmac_key_string,
+                      hmac_key_size=key_size, mode='CBC', size=key_size)
+
+    def __json__(self):
+        data = {
+            'hmacKey': {
+                'hmacKeyString': self.hmac_key_string,
+                'size': self.hmac_key_size
+            },
+            'aesKeyString': self.aes_key_string,
+            'mode': self.mode.upper(),
+            'size': int(self.size)
+        }
+        return json.dumps(data)
 
     def __repr__(self):
         return ('<AESKey hmac_key_size=%s,mode=%s,size=%s>' % (self.hmac_key_size, self.mode,
--- a/st2common/st2common/util/crypto.py
+++ b/st2common/st2common/util/crypto.py
@@ -124,7 +124,7 @@ class AESKey(object):
         return AESKey(aes_key_string=aes_key_string, hmac_key_string=hmac_key_string,
                       hmac_key_size=key_size, mode='CBC', size=key_size)
 
-    def __json__(self):
+    def to_json(self):
         """
         Return JSON representation of this key which is fully compatible with keyczar JSON key
         file format.
--- a/st2common/st2common/util/crypto.py
+++ b/st2common/st2common/util/crypto.py
@@ -13,40 +13,112 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""
+Module for handling symmetric encryption and decryption of short values (mostly used for encrypted
+datastore values aka secrets).
+
+Note: In the past this module used and relied on keyczar, but since keyczar doesn't support Python
+3, we moved to cryptography.
+
+symmetric_encrypt and symmetric_decrypt functions except values as returned by the AESKey.Encrypt
+and AESKey.Decrypt methods in keyczar. Those functions follow the same approach as keyczar methods,
+but they use and rely on primitives and methods from cryptography library.
+
+This was done to make the keyczar -> cryptography migration fully backward compatible.
+
+Eventually, we should  move to Fernet (https://cryptography.io/en/latest/fernet/) recipe for
+symmetric encryption / decryption, because it offers more robustness and safer defaults (SHA256
+instead of SHA1, etc.).
+"""
+
 from __future__ import absolute_import
+
+import os
+import json
 import binascii
+import base64
+
+from hashlib import sha1
+
+import six
+
+from cryptography.hazmat.primitives.ciphers import Cipher
+from cryptography.hazmat.primitives.ciphers import algorithms
+from cryptography.hazmat.primitives.ciphers import modes
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives import hmac
+from cryptography.hazmat.backends import default_backend
 
 __all__ = [
     'read_crypto_key',
+
     'symmetric_encrypt',
-    'symmetric_decrypt'
+    'symmetric_decrypt',
+
+    'cryptography_symmetric_encrypt',
+    'cryptography_symmetric_decrypt',
+
+    # NOTE: Keyczar functions are here for testing reasons - they are only used by tests
+    'keyczar_symmetric_encrypt',
+    'keyczar_symmetric_decrypt'
 ]
 
+# Keyczar related constants
+KEYCZAR_HEADER_SIZE = 5
+KEYCZAR_AES_BLOCK_SIZE = 16
+KEYCZAR_HLEN = sha1().digest_size
 
-def read_crypto_key(key_path, key_type=None):
+
+class AESKey(object):
     """
-    Return the crypto key given a path to key file and the key type.
+    Class representing AES key object.
+    """
 
-    :param key_path: Absolute path to file containing crypto key.
+    aes_key_string = None
+    hmac_key_string = None
+    hmac_key_size = None
+    mode = None
+    size = None
+
+    def __init__(self, aes_key_string, hmac_key_string, hmac_key_size, mode='CBC', size=128):
+        self.aes_key_string = aes_key_string
+        self.hmac_key_string = hmac_key_string
+        self.hmac_key_size = hmac_key_size
+        self.mode = mode
+        self.size = size
+
+        self.hmac_key_bytes = Base64WSDecode(self.hmac_key_string)
+        self.aes_key_bytes = Base64WSDecode(self.aes_key_string)
+
+    def __repr__(self):
+        return ('<AESKey hmac_key_size=%s,mode=%s,size=%s>' % (self.hmac_key_size, self.mode,
+                                                               self.size))
+
+
+def read_crypto_key(key_path):
+    """
+    Return the crypto key given a path to key file.
+
+    :param key_path: Absolute path to file containing crypto key in Keyczar JSON format.
     :type key_path: ``str``
 
-    :param key_type: Type of crypto key.
-    :type key_type: :class:`keyczar.keys.KeyType`
-
-    :rtype: ``str``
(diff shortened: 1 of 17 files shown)


test-requirements.txt
```diff
@@ -25,3 +25,4 @@ psutil==5.4.5
 webtest==2.0.25
 rstcheck>=3.3.0,<3.4
 pyrabbit
+python-keyczar
```
Kami: StackStorm doesn't depend on `python-keyczar` anymore, but we still needs it for tests where we check our code and output is fully compatible with keyczar one.

lakshmi-kannan:
Whoever wrote the original code with python-keyczar had right abstractions that changing libraries was almost easy :D. I like the changes.

test-requirements.txt
Kami: Yep, will do.

I thought I already added it, but probably to the other file.

Kami:
@bigmstone 

> I would almost prefer this be its own library w/ the work you put into building backwards compatibility for keyczar JSON compatibility and reading that into AESKey. +1

Yeah, I was actually thinking of making a new package, `cryptography-keyczar-adapter` or something along those lines.

It's worth nothing that keyczar offers a lot more functionality and APIs which we don't use and as such, it's not implemented here (asymmetric / public crypto, key rotation, etc). So the adapter is / would be specifically for symmetric AES crypto and reading keyczar compatible JSON key files :)

st2common/st2common/util/crypto.py
```diff
@@ -13,40 +13,291 @@
...
+        aes_key_bytes = os.urandom(int(key_size / 8))
+        aes_key_string = Base64WSEncode(aes_key_bytes)
+
+        hmac_key_bytes = os.urandom(int(key_size / 8))
+        hmac_key_string = Base64WSEncode(hmac_key_bytes)
+
+        return AESKey(aes_key_string=aes_key_string, hmac_key_string=hmac_key_string,
+                      hmac_key_size=key_size, mode='CBC', size=key_size)
+
+    def __json__(self):
```
blag: This is probably not a good idea at this point, since [special methods starting and ending with double underscores are reserved for the Python interpreter](https://bugs.python.org/issue27362#msg269002) (read: their names get mangled by CPython).

Is this for backwards compatibility with keyczar? That's really the only good reason I can think to use a dunder for this method.

I would rename this to something like `to_json` or `_to_json` if it needs to be private.

st2common/st2common/util/crypto.py
```diff
@@ -13,40 +13,291 @@
...
     'read_crypto_key',
+
     'symmetric_encrypt',
-    'symmetric_decrypt'
+    'symmetric_decrypt',
+
+    'cryptography_symmetric_encrypt',
+    'cryptography_symmetric_decrypt',
+
+    # NOTE: Keyczar functions are here for testing reasons - they are only used by tests
```
blag: You don't need to export these explicitly if you tweak everything that imports them to dig them out of the module:

```python
from st2common.util.crypto import keyczar_symmetric_encrypt

keyczar_symmetric_encrypt(...)
```

can be rewritten as:

```python
import st2common.util.crypto as crypto

crypto.keyczar_symmetric_encrypt(...)
```

st2common/tests/unit/test_crypto_utils.py
```diff
@@ -14,33 +14,227 @@
...
 from six.moves import range
+from cryptography.exceptions import InvalidSignature
+
+from st2common.util.crypto import KEYCZAR_HEADER_SIZE
+from st2common.util.crypto import AESKey
+from st2common.util.crypto import read_crypto_key
+from st2common.util.crypto import symmetric_encrypt
+from st2common.util.crypto import symmetric_decrypt
+from st2common.util.crypto import keyczar_symmetric_decrypt
+from st2common.util.crypto import keyczar_symmetric_encrypt
```
blag: Can be rewritten to:

```python
import st2common.util.crypto as crypto
keyczar_symmetric_encrypt = crypto.keyczar_symmetric_encrypt
```

so you don't have to export `keyczar_symmetric_encrypt`.

Same story with `keyczar_symmetric_decrypt`. See my companion comment in `st2common/utils/crypto.py`.

st2common/st2common/util/crypto.py
```diff
@@ -13,40 +13,291 @@
...
+    mode = None
+    size = None
 
-    :param key_type: Type of crypto key.
-    :type key_type: :class:`keyczar.keys.KeyType`
+    def __init__(self, aes_key_string, hmac_key_string, hmac_key_size, mode='CBC', size=128):
+        if mode not in 'CBC':
+            raise ValueError('Unsupported mode: %s' % (mode))
 
-    :rtype: ``str``
```
blag: I would raise a warning here instead of an error - UNLESS you are maintaining backwards compatibility.

Some users may want to import old (<128 bit) keys.

st2common/st2common/util/crypto.py
```diff
@@ -13,40 +13,291 @@
...
+        self.aes_key_bytes = Base64WSDecode(self.aes_key_string)
+
+    @classmethod
+    def generate(self, key_size=256):
+        """
+        Generate a new AES key with the corresponding HMAC key.
+
+        :rtype: :class:`AESKey`
+        """
+        if key_size < 128:
```
blag: You may also want to consider putting the minimum AES key size into its own constant to give the reader an idea of where the number 128 came from. Right now it's just a magic number to anybody who doesn't keep up on crypto recommendations.

st2common/st2common/util/crypto.py
```diff
@@ -13,40 +13,291 @@
...
+
+    The final encrypted string value consists of:
+
+    [message bytes][HMAC signature bytes for the message] where message consists of
+    [keyczar header plaintext][IV bytes][ciphertext bytes]
+
+    NOTE: Header itself is unused, but it's added so the format is compatible with keyczar format.
+
+    """
+    assert isinstance(encrypt_key, AESKey), 'encrypt_key needs to be AESKey class instance'
```
blag: These asserts look okay, just be aware that running Python with optimizations turned on bypasses all assertions made with the `assert` keyword. This has caused security issues in the past ([example: PySAML](https://github.com/IdentityPython/pysaml2/issues/451)).

[Assertions are meant to be used as debugging aids, not as a mechanism for error handling.](https://www.quora.com/How-do-I-use-Python-asserts-even-when-optimization-mode-is-used/answer/Dan-Bader-1)

It sucks, but you may want to use plain `if` statements for all of your assertions:

```python
if not isinstance(plaintext, (six.text_type, six.string_types, six.binary_types)):
    raise CustomErrorType(...)  # Not an AssertionError
```

More help: https://dbader.org/blog/python-assert-tutorial

st2common/st2common/util/crypto.py
```diff
@@ -57,28 +308,128 @@ def symmetric_encrypt(encrypt_key, plaintext):
     'Initialization Vector' per run and the IV is part of the output.
 
     :param encrypt_key: Symmetric AES key to use for encryption.
-    :type encrypt_key: :class:`keyczar.keys.AesKey`
+    :type encrypt_key: :class:`keyczar.keys.AESKey`
```
blag: This is close but not quite accurate. The class should be: `st2common.utils.crypto.AESKey`.

blag:
Definitely consider releasing it as its own package even if it only implements what we need from it. Other open source devs might be interested in it and implement what they need, to the point it becomes fairly complete. Heck, the `cryptography` developers themselves might be interested in absorbing it as an adapter.

I've appreciated keyczar but it's not really maintained anymore, so I welcome creating a backwards-compatible replacement/adapter for it.

st2common/st2common/util/crypto.py
```diff
@@ -13,40 +13,291 @@
...
+    aes_key_string = None
+    hmac_key_string = None
+    hmac_key_size = None
+    mode = None
+    size = None
 
-    :param key_type: Type of crypto key.
-    :type key_type: :class:`keyczar.keys.KeyType`
+    def __init__(self, aes_key_string, hmac_key_string, hmac_key_size, mode='CBC', size=128):
+        if mode not in 'CBC':
```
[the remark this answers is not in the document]
Kami: True, good catch :+1:

st2common/st2common/util/crypto.py
Kami: Yeah, I agree - I will change it to `to_json` (`__json__` is kinda a left over from the simplejson library,).

st2common/st2common/util/crypto.py
Kami: Importing each function separately is just a personal preference. I prefer it over module imports where module names are generic as this one.

As far as exporting all the public functions and using ``__all__`` goes - that's just a "safety guard" and good practice I tend to follow. It makes it easier to see which public methods / classes a module exports and in case someone does "from foo import *", it will only include "public" methods. I know asterik imports are a bad idea and we should have a lint check for that, but still.

st2common/tests/unit/test_crypto_utils.py
Kami: See my comment above - personal preference and I think it's easier to glance at the top of the module to see what functions / classes a particular module imports and uses.

st2common/st2common/util/crypto.py
Kami: We never supported smaller keys so I think there is no need for that (our tool always generated 128 bit keys).

st2common/st2common/util/crypto.py
Kami: Yeah, I used them for what they are intended - as debugging / developer aid aka to make sure programmer passes in a correct type (stuff which should get caught during tests and not during run time for which I would use exceptions).

For dynamic run time stuff I always use exceptions and not asserts.

st2common/st2common/util/crypto.py
Kami: Fixed.

st2common/st2common/util/crypto.py
Kami: Agreed, I will make a constant.

st2common/st2common/util/crypto.py
blag: Yep, I like to import each function separately too. Although I think you can still import individual functions even if they aren't "exported" with `__all__`, so you may not even need to export those functions at all. Whatever you think is best. 😃

Kami: Yep, __all__ just affects asterisk import (``*``) import behavior :)

PR #4171, opened 2018-06-08
Simplified issue template

LindsayHill:
I've simplified the issue template a little. Goal is still to have a template, but to get more useful content in the snippet displayed when issues are auto-posted to Slack.


cognifloyd:
:+1:

Kami:
Yep, much better :+1:

PR #4173, opened 2018-06-11
Fix config permissions on the config directory and the config file

### LINKED ISSUE #4144: Weak .st2 config file permissions allow reading st2 creds by other Linux users
`st2 login -w` creates `~/.st2/config` file with `r-o` permissions, meaning that unauthorized linux user can read that file and obtain st2 login credentials:
```
$ rm -rf ~/.st2
$ st2 login st2admin -p Ch@ngeMe -w

$ ls -la ~/.st2
total 16
drwxrwxr-x 2 vagrant vagrant 4096 May 24 14:59 .
drwxr-xr-x 8 vagrant vagrant 4096 May 24 14:59 ..
-rw-rw-r-- 1 vagrant vagrant   55 May 24 14:59 config
-rw------- 1 vagrant vagrant   77 May 24 14:59 token-st2admin
```

Additionally, `~/.st2/` dir permissions when created might be adjusted (currently `0755`).
Adding setgid for st2 config dir will help to better handle `sudo st2` cases when config or token file is created with `root` permissions (see https://github.com/StackStorm/packer-st2/issues/38 for example).

### Wrap Up
- `~/.st2/` dir when created by st2 should have `2770` permissions (currently `0755`)
- `~/.st2/config` when created by st2 should have `660` permissions (currently `664`)
- `~/.st2/token` when created by st2 should have `660` permissions (currently `640`)

### Related
https://github.com/StackStorm/st2-packages/issues/558 and https://github.com/StackStorm/packer-st2/issues/38

blag:
Fixes #4144.

This ensures that we create the config directory and file with the proper permissions.

It also tweaks the `st2 login` command to check the permissions on the config file (`~/.st2/config`) and directory (`~/.st2`) every time it is run, and then automatically fix permissions if they are too permissive. It does use the `warnings` module to warn the user about this, but does not give them a chance to opt out or turn the behavior off.

Automatically fixing the directory/file permissions assumes that there is no good reason for a process owned by another user or group to need read access to the config file. That is a very large assumption, so I'm inviting discussion about that.

The options are:

1. simply "warn the user" into fixing the config permissions, but not automatically fix it ourselves
2. use environment variables or command line flags to control warnings and automatic fixing behavior
3. warn the user and automatically fix it ourselves all the time
4. fix it automatically without warning

I have chosen option 3 here.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -46,7 +46,7 @@ Changed
 ~~~~~~~
 
 * Update st2 CLI to create the configuration directory and file, and authentication tokens with
-  secure permissions (eg: readable only to owner)
+  secure permissions (eg: readable only to owner) #4173
 * Refactor the callback module for the post run in runner to be more generic. (improvement)
 * Update various Python dependencies to the latest stable versions (gunicorn, gitpython,
   python-gnupg, tooz, flex). #4110
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -40,6 +40,8 @@ Added
 Changed
 ~~~~~~~
 
+* Update st2 CLI to create the configuration directory and file, and authentication tokens with
+  secure permissions (eg: readable only to owner)
 * Refactor the callback module for the post run in runner to be more generic. (improvement)
 * Update various Python dependencies to the latest stable versions (gunicorn, gitpython,
   python-gnupg, tooz, flex). #4110
--- a/st2client/st2client/base.py
+++ b/st2client/st2client/base.py
@@ -225,7 +225,10 @@ class BaseCLIApp(object):
         :rtype: ``str``
         """
         if not os.path.isdir(ST2_CONFIG_DIRECTORY):
-            os.makedirs(ST2_CONFIG_DIRECTORY)
+            os.makedirs(ST2_CONFIG_DIRECTORY, mode=0o2770)
+            # os.makedirs straight up ignores the setgid bit, so we have to set
+            # it manually
+            os.chmod(ST2_CONFIG_DIRECTORY, 0o2770)
 
         cached_token_path = self._get_cached_token_path_for_user(username=username)
 
@@ -253,7 +256,7 @@ class BaseCLIApp(object):
         file_st_mode = oct(os.stat(cached_token_path).st_mode & 0o777)
         others_st_mode = int(file_st_mode[-1])
 
-        if others_st_mode >= 4:
+        if others_st_mode >= 2:
             # Every user has access to this file which is dangerous
             message = ('Permissions (%s) for cached token file "%s" are to permissive. Please '
                        'restrict the permissions and make sure only your own user can read '
@@ -290,7 +293,10 @@ class BaseCLIApp(object):
         :type token_obj: ``object``
         """
         if not os.path.isdir(ST2_CONFIG_DIRECTORY):
-            os.makedirs(ST2_CONFIG_DIRECTORY)
+            os.makedirs(ST2_CONFIG_DIRECTORY, mode=0o2770)
+            # os.makedirs straight up ignores the setgid bit, so we have to set
+            # it manually
+            os.chmod(ST2_CONFIG_DIRECTORY, 0o2770)
 
         username = token_obj.user
         cached_token_path = self._get_cached_token_path_for_user(username=username)
@@ -326,9 +332,10 @@ class BaseCLIApp(object):
         # open + chmod are two operations which means that during a short time frame (between
         # open and chmod) when file can potentially be read by other users if the default
         # permissions used during create allow that.
-        fd = os.open(cached_token_path, os.O_WRONLY | os.O_CREAT, 0o600)
+        fd = os.open(cached_token_path, os.O_WRONLY | os.O_CREAT, 0o660)
         with os.fdopen(fd, 'w') as fp:
             fp.write(data)
+        os.chmod(cached_token_path, 0o660)
 
         self.LOG.debug('Token has been cached in "%s"' % (cached_token_path))
         return True
--- a/st2client/st2client/base.py
+++ b/st2client/st2client/base.py
@@ -264,12 +264,8 @@ class BaseCLIApp(object):
             # Every user has access to this file which is dangerous
             message = ('Permissions (%s) for cached token file "%s" are too permissive. Please '
                        'restrict the permissions and make sure only your own user can read '
-                       'from or write to the file.'
-                       '\n\n'
-                       'You can fix this by running:'
-                       '\n\n'
-                       '    chmod o-rw %s')
-            self.LOG.warn(message % (file_st_mode, cached_token_path, cached_token_path))
+                       'from or write to the file.' % (file_st_mode, cached_token_path))
+            self.LOG.warn(message)
 
         with open(cached_token_path) as fp:
             data = fp.read()
--- a/st2client/st2client/base.py
+++ b/st2client/st2client/base.py
@@ -153,24 +153,28 @@ class BaseCLIApp(object):
 
         return client
 
-    def _get_config_file_options(self, args):
+    def _get_config_file_options(self, args, validate_config_permissions=True):
         """
         Parse the config and return kwargs which can be passed to the Client
         constructor.
 
         :rtype: ``dict``
         """
-        rc_options = self._parse_config_file(args=args)
+        rc_options = self._parse_config_file(
+            args=args, validate_config_permissions=validate_config_permissions)
         result = {}
         for kwarg_name, (section, option) in six.iteritems(CONFIG_OPTION_TO_CLIENT_KWARGS_MAP):
             result[kwarg_name] = rc_options.get(section, {}).get(option, None)
 
         return result
 
-    def _parse_config_file(self, args):
+    def _parse_config_file(self, args, validate_config_permissions=True):
         config_file_path = self._get_config_file_path(args=args)
 
-        parser = CLIConfigParser(config_file_path=config_file_path, validate_config_exists=False)
+        parser = CLIConfigParser(config_file_path=config_file_path,
+                                 validate_config_exists=False,
+                                 validate_config_permissions=validate_config_permissions,
+                                 log=self.LOG)
         result = parser.parse()
         return result
 
@@ -258,10 +262,14 @@ class BaseCLIApp(object):
 
         if others_st_mode >= 2:
             # Every user has access to this file which is dangerous
-            message = ('Permissions (%s) for cached token file "%s" are to permissive. Please '
+            message = ('Permissions (%s) for cached token file "%s" are too permissive. Please '
                        'restrict the permissions and make sure only your own user can read '
-                       'from the file' % (file_st_mode, cached_token_path))
-            self.LOG.warn(message)
+                       'from or write to the file.'
+                       '\n\n'
+                       'You can fix this by running:'
+                       '\n\n'
+                       '    chmod o-rw %s')
+            self.LOG.warn(message % (file_st_mode, cached_token_path, cached_token_path))
 
         with open(cached_token_path) as fp:
             data = fp.read()
@@ -356,7 +364,7 @@ class BaseCLIApp(object):
         return result
 
     def _print_config(self, args):
-        config = self._parse_config_file(args=args)
+        config = self._parse_config_file(args=args, validate_config_permissions=False)
 
         for section, options in six.iteritems(config):
             print('[%s]' % (section))
--- a/st2client/st2client/commands/auth.py
+++ b/st2client/st2client/commands/auth.py
@@ -141,7 +141,6 @@ class LoginCommand(resource.ResourceCommand):
 
         with open(config_file, 'w') as cfg_file_out:
             config.write(cfg_file_out)
-        os.chmod(config_file, 0o660)
 
         return manager
 
--- a/st2client/st2client/commands/auth.py
+++ b/st2client/st2client/commands/auth.py
@@ -18,7 +18,6 @@ from __future__ import absolute_import
 import getpass
 import json
 import logging
-import os
 
 import requests
 from six.moves.configparser import ConfigParser
--- a/st2client/st2client/commands/auth.py
+++ b/st2client/st2client/commands/auth.py
@@ -18,6 +18,7 @@ from __future__ import absolute_import
 import getpass
 import json
 import logging
+import os
 
 import requests
 from six.moves.configparser import ConfigParser
@@ -140,6 +141,7 @@ class LoginCommand(resource.ResourceCommand):
 
         with open(config_file, 'w') as cfg_file_out:
             config.write(cfg_file_out)
+        os.chmod(config_file, 0o660)
 
         return manager
 
--- a/st2client/st2client/commands/auth.py
+++ b/st2client/st2client/commands/auth.py
@@ -18,6 +18,7 @@ from __future__ import absolute_import
 import getpass
 import json
 import logging
+import os
 
 import requests
 from six.moves.configparser import ConfigParser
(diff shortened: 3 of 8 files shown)


st2client/st2client/base.py
```diff
@@ -225,7 +225,7 @@ def _get_cached_auth_token(self, client, username, password):
         :rtype: ``str``
         """
         if not os.path.isdir(ST2_CONFIG_DIRECTORY):
-            os.makedirs(ST2_CONFIG_DIRECTORY)
+            os.makedirs(ST2_CONFIG_DIRECTORY, mode=0o770)
```
armab: Can we also set `setgid` for the dir? `2770`,

see https://github.com/StackStorm/st2/issues/4144 and https://github.com/StackStorm/packer-st2/issues/38 why

armab:
My vote goes in favor of:
> 1. simply "warn the user" into fixing the config permissions, but not automatically fix it ourselves

Create config dir and files with safe permissions and don't try to change them after that.

armab:
To fix the full story, need also to take care about the https://github.com/StackStorm/st2-packages/issues/558 (https://github.com/StackStorm/st2-packages/issues/520 might help with it)

Kami:
I agree with @armab, I also think 1) is better (user might have a (valid) use case to change the permissions after the fact and we probably shouldn't mess with the permissions after the initial set up, but should definitely warn in case of unsafe permissions).

st2client/st2client/config_parser.py
```diff
@@ -138,6 +139,20 @@ def parse(self):
             # Config doesn't exist, return the default values
             return CONFIG_DEFAULT_VALUES
 
+        config_dir_path = os.path.dirname(self.config_file_path)
+
+        if bool(os.stat(config_dir_path).st_mode ^ 0o770):
+            warnings.warn('Setting StackStorm config directory permissions '
```
Kami: Instead of using `warnings` module we should probably use ``logger.warn`` since we also use logging in other places for such things.

This way it uses the existing framework and setting log level, etc works as expected (e.g. user wants to suppress warnings and only see messages with level ERROR and above).

st2client/st2client/base.py
blag: Yes, I'll do this.

I assume we also want the `setuid` bit set, right? `0o6770`?

armab: I don't have any opinions about the `setuid`, since I didn't see that bit having effect on dir.
If you believe it can help in some use cases, - I'm good with adding it.

blag: Nope, I don't think it's helpful.

Fixed.

Kami:
Can you please add a changelog entry?

Besides that, LGTM :+1:

st2client/st2client/config_parser.py
```diff
@@ -138,6 +141,40 @@ def parse(self):
...
+                "You can fix this by running:"
+                "\n\n"
+                "chmod 770 {config_dir}".format(config_dir=config_dir_path))
+
+        # Make sure the setgid bit is set
+        if not bool(os.stat(config_dir_path).st_mode & 0o2000):
+            self.LOG.warn(
+                # TODO: Perfect place for an f-string
+                "The StackStorm configuration directory permissions are "
+                "insecure (SGID bit not set)."
```
armab: I don't think it's a best to warn user about the absence of SGID bit.
It's a workaround that'll help creating file names with the same group under the config dir, but not really that much security-related.

Besides of that minor comment, looks good to me too :+1:

st2client/st2client/base.py
armab: Still need to set `setguid` bit. Don't see it anywhere, except of tests setting that bit.

blag: Fixed, please re-review when you have time.

st2client/st2client/base.py
```diff
@@ -290,7 +293,10 @@ def _cache_auth_token(self, token_obj):
         :type token_obj: ``object``
         """
         if not os.path.isdir(ST2_CONFIG_DIRECTORY):
-            os.makedirs(ST2_CONFIG_DIRECTORY)
+            os.makedirs(ST2_CONFIG_DIRECTORY, mode=0o2770)
+            # os.makedirs straight up ignores the setgid bit, so we have to set
+            # it manually
+            os.chmod(ST2_CONFIG_DIRECTORY, 0o2770)
```
armab: Makes sense in dir conext :+1:

st2client/st2client/base.py
```diff
@@ -326,9 +332,10 @@ def _cache_auth_token(self, token_obj):
         # open + chmod are two operations which means that during a short time frame (between
         # open and chmod) when file can potentially be read by other users if the default
         # permissions used during create allow that.
-        fd = os.open(cached_token_path, os.O_WRONLY | os.O_CREAT, 0o600)
+        fd = os.open(cached_token_path, os.O_WRONLY | os.O_CREAT, 0o660)
         with os.fdopen(fd, 'w') as fp:
             fp.write(data)
+        os.chmod(cached_token_path, 0o660)
```
armab: Wouldn't this always try to chmod cached token file?
I think we previously agreed to do chmod operations only on initial file creation.

st2client/st2client/commands/auth.py
```diff
@@ -140,6 +141,7 @@ def run(self, args, **kwargs):
 
         with open(config_file, 'w') as cfg_file_out:
             config.write(cfg_file_out)
+        os.chmod(config_file, mode=0o660)
```
armab: Same here. I think we previously agreed to set file permissions only on creation and don't try chmod afterwards.

st2client/st2client/config_parser.py
```diff
@@ -138,6 +141,40 @@ def parse(self):
...
+        # Make sure the setgid bit is set on the directory
+        if not bool(os.stat(config_dir_path).st_mode & 0o2000):
+            self.LOG.info(
+                # TODO: Perfect place for an f-string
+                "The SGID bit is not set on the StackStorm configuration "
+                "directory."
+                "\n\n"
+                "You can fix this by running:"
+                "\n\n"
+                "chmod g+s {config_dir}".format(config_dir=config_dir_path))
```
armab: For the `LOG.info` message level, do you know/can you check where will it show while running `st2` commands?

st2client/st2client/base.py
blag: Yes it would, but as far as I can tell, it's only run when the file is initially created.

armab:
I tried the packages from this PR and got the error:
```
vagrant@stackstorm:~$ st2 login -w st2admin
Password: 
Failed to log in as st2admin: chmod() takes no keyword arguments
```

armab:
Also squashing all commits history in 1 during review makes it harder to follow the history and commenting/code diff.
If you prefer, - it's still possible to do during the merge as a last step.

Kami:
Can we get this resolved / finished today and merged in time for v2.8.0?

st2client/st2client/config_parser.py
blag: That depends greatly on the configuration file itself, but the `info` level is not output to stdout or stderr by default:

```python
>>> import logging
>>> logger = logging.getLogger("_")
>>> logger.info("Hello World!")
>>> 
```

blag:
> Also squashing all commits history in 1 during review makes it harder to follow the history, code diff from the most recent look and overall commenting.
> If you prefer, - it's still possible to do squash during the merge as a last step.

I just now read this - sorry! Will squash before I merge.

st2client/st2client/config_parser.py
armab: Thanks for verifying that :+1:

st2client/st2client/base.py
armab: Alright, thanks for explaining  :+1:

armab:
I can confirm that config dir and token file permissions are correct, however `~/.st2/config` with the actual username+password file still has `o+r` permissions.

See:
```
vagrant@ubuntu16:~$ rm -rf ~/.st2

vagrant@ubuntu16:~$ st2 login st2admin -w
Password: 
Logged in as st2admin

vagrant@ubuntu16:~$ ls -la ~/.st2
total 16
drwxrws--- 2 vagrant vagrant 4096 Jun 21 19:34 .
drwxr-xr-x 5 vagrant vagrant 4096 Jun 21 19:34 ..
-rw-rw-r-- 1 vagrant vagrant   55 Jun 21 19:34 config
-rw-rw---- 1 vagrant vagrant   77 Jun 21 19:34 token-st2admin

vagrant@ubuntu16:~$ st2 --version
st2 2.8dev (6d2af29), on Python 2.7.12
```

You can grab the `st2` package produced for each git push in this PR from the CircleCI -> Artifacts section to test the features while shipping.

armab:
Also curious if this "new" message comes from this PR or not when using st2 client:
```
vagrant@ubuntu16:~$ st2 --print-config
No handlers could be found for logger "st2client.config_parser"

[...]
```

blag:
@armab Thanks for the pointer to CircleCI > Artifacts!

I think I have fixed the issues you have pointed out:

```bash
stanley@ubuntu-bionic:~$ rm -rf ~/.st2
stanley@ubuntu-bionic:~$ st2 login st2admin -w 
Password: 
Logged in as st2admin
stanley@ubuntu-bionic:~$ ls -la ~/.st2 
total 16
drwxrws---  2 stanley stanley 4096 Jun 21 17:27 .
drwxr-xr-x 16 stanley stanley 4096 Jun 21 17:27 ..
-rw-rw----  1 stanley stanley   55 Jun 21 17:27 config
-rw-rw----  1 stanley stanley   77 Jun 21 17:27 token-st2admin
stanley@ubuntu-bionic:~$ 
```

```bash
stanley@ubuntu-bionic:~$ st2 --version 
st2 2.8dev (bc1eee8), on Python 2.7.15
stanley@ubuntu-bionic:~$ st2 --print-config 
[api]
url = None
[cli]
debug = False
timezone = UTC
cache_token = True
[stream]
url = None
[credentials]
username = st2admin
password = st2admin
api_key = None
[auth]
url = None
[general]
cacert = None
base_url = None
api_version = None
silence_ssl_warnings = False
stanley@ubuntu-bionic:~$ 
```

But double checking it doesn't hurt.

Kami:
I checked out this branch and confirmed that the chmod changes work as expected:

```
(virtualenv) vagrant@ubuntu-xenial:/data/stanley$ ls -la ~/.st2
total 16
drwxrws---  2 vagrant vagrant 4096 Jun 22 08:50 .
drwxr-xr-x 11 vagrant vagrant 4096 Jun 22 08:50 ..
-rw-rw----  1 vagrant vagrant   32 Jun 22 08:50 config
-rw-rw----  1 vagrant vagrant   77 Jun 22 08:50 token-testu
```

But as @armab pointed out, logging warning changes don't seem to work since the handler is not set up for st2client.config_parser. One option would be to pass logger object from `Client` to the `CLIConfigParser` class.

st2client/st2client/config_parser.py
```diff
@@ -120,6 +121,8 @@
 
 
 class CLIConfigParser(object):
+    LOG = logging.getLogger(__name__)  # logger instance to use
```
armab: Yes, looks like logging messages defined in `CLIConfigParser` didn't work. See previous comment with:
```
No handlers could be found for logger "st2client.config_parser"
```
and @Kami's hint https://github.com/StackStorm/st2/pull/4173#issuecomment-399372838 about passing logger object.

st2client/st2client/config_parser.py
```diff
@@ -138,6 +144,40 @@ def parse(self):
...
+        # Make sure the directory permissions == 0o770
+        if bool(os.stat(config_dir_path).st_mode & 0o777 ^ 0o770):
+            self.LOG.warn(
+                # TODO: Perfect place for an f-string
+                "The StackStorm configuration directory permissions are "
+                "insecure (too permissive)."
+                "\n\n"
+                "You can fix this by running:"
+                "\n\n"
+                "chmod 770 {config_dir}".format(config_dir=config_dir_path))
```
armab: Edge case, warning messages are duplicated:
```
$ st2 action list
2018-06-22 17:41:48,953  WARNING - The StackStorm configuration directory permissions are insecure (too permissive).

You can fix this by running:

chmod 770 /home/vagrant/.st2
2018-06-22 17:41:48,954  WARNING - The StackStorm configuration file permissions are insecure.

You can fix this by running:

chmod 660 /home/vagrant/.st2/config
2018-06-22 17:41:48,955  WARNING - The StackStorm configuration directory permissions are insecure (too permissive).

You can fix this by running:

chmod 770 /home/vagrant/.st2
2018-06-22 17:41:48,957  WARNING - The StackStorm configuration file permissions are insecure.

You can fix this by running:

chmod 660 /home/vagrant/.st2/config
2018-06-22 17:41:48,959  WARNING - Permissions (0664) for cached token file "/home/vagrant/.st2/token-st2admin" are to permissive. Please restrict the permissions and make sure only your own user can read from the file
```

To reproduce: 
```
chmod -R o+r ~/.st2
st2 action list
```

CHANGELOG.rst
```diff
@@ -45,6 +45,8 @@ Added
 Changed
 ~~~~~~~
 
+* Update st2 CLI to create the configuration directory and file, and authentication tokens with
+  secure permissions (eg: readable only to owner)
```
armab: NIT: To include the PR #number in the CHANGELOG.

st2client/st2client/base.py
```diff
@@ -253,12 +260,16 @@ def _get_cached_auth_token(self, client, username, password):
...
-        if others_st_mode >= 4:
+        if others_st_mode >= 2:
             # Every user has access to this file which is dangerous
-            message = ('Permissions (%s) for cached token file "%s" are to permissive. Please '
+            message = ('Permissions (%s) for cached token file "%s" are too permissive. Please '
                        'restrict the permissions and make sure only your own user can read '
-                       'from the file' % (file_st_mode, cached_token_path))
-            self.LOG.warn(message)
+                       'from or write to the file.'
+                       '\n\n'
```
armab: NIT: I think that's too much new lines here which makes it pretty inconsistent with existing stackstorm warning messages by taking a lot of space. Example:

```
vagrant@stackstorm:~$ st2 action list
2018-06-26 14:50:42,463  WARNING - The StackStorm configuration directory permissions are insecure (too permissive).

You can fix this by running:

    chmod 770 /home/vagrant/.st2

2018-06-26 14:50:42,463  WARNING - The StackStorm configuration file permissions are insecure.

You can fix this by running:

    chmod 660 /home/vagrant/.st2/config

2018-06-26 14:50:42,463  WARNING - Permissions (0664) for cached token file "/home/vagrant/.st2/token-st2admin" are too permissive. Please restrict the permissions and make sure only your own user can read from or write to the file.

You can fix this by running:

    chmod o-rw /home/vagrant/.st2/token-st2admin
```

Kami: Small nitpick - In some places in the log messages we use numerical permission levels and in other we use symbolic notation.

It's probably a good idea to unify it and use numerical / symbolic everywhere :)

Kami:
Probably a good idea to get this merged today so we can include it in v2.8.0.

Those non-blockers mentioned above can be fixed / improved later.

st2client/st2client/base.py
blag: I'll simply remove the help text. I just usually like to give users some idea of how to fix a problem, which requires readable formatting, but it might not be the best for log messages.

The octal vs symbolic issue is more than just a notation difference, it's a functional one. The octal notation sets the permissions for the user, group, and "others", but the symbolic notation here performs operations on the permissions for just "others" and leaves the bits for the user and group intact.

blag:
Warnings and tests are turned off in #4215 ([41769824b51aa19edb786a9e7e4b7f91b16f65d7](https://github.com/StackStorm/st2/pull/4215/commits/41769824b51aa19edb786a9e7e4b7f91b16f65d7) and [e16a59ac029926f63fa1eb507653f19fef18f4c7](https://github.com/StackStorm/st2/pull/4215/commits/e16a59ac029926f63fa1eb507653f19fef18f4c7)).

PR #4177, opened 2018-06-12
Adding st2kv function

bigmstone:
This PR adds st2kv custom expression function for orchestra rendering.

- [x] KVP Scope
- [x] Decrypt
- [x] Move implementation to st2common functions
~Refactor ST2API to use st2common implementation~ (Too messy)
- [x] Testing
- [x] Add context to get user executing st2kv
~Delayed rendering to avoid secret being recorded in WorkflowExecution~ Will happen in another PR

### DIFF
--- a/contrib/runners/orchestra_runner/functions/st2kv.py
+++ b/contrib/runners/orchestra_runner/functions/st2kv.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import six
+
+from st2common.util.keyvalue import get_key
+
+import logging
+
+LOG = logging.getLogger(__name__)
+
+
+def st2kv_(context, key, decrypt=False):
+    if not isinstance(key, six.string_types):
+        raise TypeError('Given key is not typeof string.')
+    if not isinstance(decrypt, bool):
+        raise TypeError('Decrypt parameter is not typeof bool.')
+
+    try:
+        user = context['__vars']['st2']['user']
+    except KeyError:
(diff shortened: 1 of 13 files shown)


st2api/st2api/controllers/v1/keyvalue.py
```diff
@@ -69,9 +69,6 @@ def get_one(self, name, requester_user, scope=FULL_SYSTEM_SCOPE, user=None, decr
             Handle:
                 GET /keys/key1
         """
-        if not scope:
-            scope = FULL_SYSTEM_SCOPE
-
```
m4dcoder: Are you sure about this change?

st2common/st2common/services/workflows.py
```diff
@@ -351,14 +363,18 @@ def handle_action_execution_completion(ac_ex_db):
 
 
 def deserialize_conductor(wf_ex_db):
+    context = copy.deepcopy(wf_ex_db.flow['contexts'][0].get('__vars', {}))
+    context.get('st2', {}).update({'action_id': str(wf_ex_db.action_execution)})
```
m4dcoder: I would prefer we add a column in WorkflowExecutionDB for context and just serialize/deserialize from there instead of rebuilding this here.

st2common/st2common/services/workflows.py
```diff
@@ -67,8 +67,20 @@ def request(wf_def, ac_ex_db):
...
+
+    context['st2'].update(ac_ex_db.context)
+
     # Instantiate the workflow conductor.
-    conductor = conducting.WorkflowConductor(wf_spec, **action_params)
+    conductor = conducting.WorkflowConductor(
+        wf_spec,
+        inputs=action_params,
+        context=context
+    )
```
m4dcoder: Just my style preference.  Can we do something like https://github.com/StackStorm/orchestra/blob/master/orchestra/tests/unit/conducting/test_workflow_conductor.py#L76?

st2api/st2api/controllers/v1/keyvalue.py
bigmstone: Left over from when I was experimenting w/ st2api. Removed.

contrib/runners/orchestra_runner/functions/st2kv.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def st2kv_(context, key, decrypt=False):
+    if not isinstance(key, six.string_types):
+        raise TypeError('Given key is not typeof string.')
+    if not isinstance(decrypt, bool):
+        raise TypeError('Decrypt parameter is not typeof bool.')
+
+    try:
+        user = context['__vars']['parent']['user']
```
m4dcoder: You will need to modify the orchestra runner to include the user (under runner.context.get('user', cfg.CONF.system_user.user)) who requested the workflow execution at https://github.com/StackStorm/st2/blob/master/contrib/runners/orchestra_runner/orchestra_runner/orchestra_runner.py#L59. Once it's asigned to the context in orchestra, it would be under `context['__vars']['st2']['user']`.

PR #4180, opened 2018-06-15
Initial commit to split timers to own process

lakshmi-kannan:
## What?
This PR splits out the timer portion (one that injects trigger instances based on user specified cron expressions in rules) into its own process. 

## Why? 

For the kubernetes (k8s) HA story, we are going to rely on a single timers engine container with failover handed natively by k8s. The scale requirements for timers aren't that rigorous and the timer doesn't do anything other than inject a trigger instance into rabbitmq. This makes the story for timers simple. This also allows us to scale rules engine horizontally without worrying about partitioning timers. Since rules engine don't modify state (other than add operational entries to DB), we can use k8s to set a scale number for rules engine and handle both failover and scaling with k8s primitives. So we make scaling rules engine story simpler too. 

In the future, we can decide to split timers into different partitions and go for a more complex HA model if needed. So this change will enable future scaling optimizations for timers. We are also thinking about adding the scheduling logic into this for workflow orchestrators. 

## TODO

* [x] Packaging changes https://github.com/StackStorm/st2-packages/pull/564
* [x] Production log configurations
* [x] st2ctl to start/stop 
* [x] Documentation (No special documentation needed)
* [x] Upgrade notes https://github.com/StackStorm/st2docs/pull/749
* [x] CI upgrade testing (Install 2.7.2, enable timer rule, upgrade to latest unstable, see if rule fires still)
* [x] Changelog 
* [x] st2docs change


conf/st2.conf.sample
```diff
@@ -312,11 +312,13 @@ ssh_key_file = /home/stanley/.ssh/stanley_rsa
...
-[timer]
-# Specify to enable Timer.
+# Old config group is timer and you can specify the exact same settings in [timer] section
+[timersengine]
+# Specify to enable Timer
 enable = True
 # Timezone pertaining to the location where st2 is run.
-local_timezone = America/Los_Angeles
+local_tz = America/Los_Angeles
+logging = st2reactor/conf/logging.timersengine.conf
```
armab: Probably `logging = conf/logging.garbagecollector.conf` fits this config

conf/st2.conf.sample
```diff
@@ -312,11 +312,13 @@ ssh_key_file = /home/stanley/.ssh/stanley_rsa
...
 
-[timer]
-# Specify to enable Timer.
+# Old config group is timer and you can specify the exact same settings in [timer] section
+[timersengine]
+# Specify to enable Timer
 enable = True
 # Timezone pertaining to the location where st2 is run.
-local_timezone = America/Los_Angeles
+local_tz = America/Los_Angeles
```
cognifloyd: `local_tz` vs `local_timezone` ?

lakshmi-kannan: It was auto-generated but I'll look into why the auto-gen code did this.

armab:
@lakshmi-kannan Since we're on `2.9`, is it possible to complete this PR so we'll start integrating with the new `st2timersegine`?

An additional request to include a description for the service in
https://docs.stackstorm.com/reference/ha.html#components

st2reactor/st2reactor/cmd/timersengine.py
```diff
@@ -0,0 +1,88 @@
...
+    try:
+        timer_thread = None
+        if cfg.CONF.timer.enable or cfg.CONF.timersengine.enable:
+            local_tz = cfg.CONF.timer.local_timezone or cfg.CONF.timersengine.local_timezone
+            timer = St2Timer(local_timezone=local_tz)
+            timer_thread = eventlet.spawn(_kickoff_timer, timer)
+            LOG.info(TIMER_ENABLED_LOG_LINE)
+            return timer_thread.wait()
+        else:
+            LOG.info(TIMER_DISABLED_LOG_LINE)
```
Kami: It looks like if timer engine is disabled service will just exit immediately on startup, right?

Just something to keep in mind / document for monitoring purposes (e.g. if timer engine service is disables, `st2timerengine` service won't be running and it will exit immediately on startup).

st2reactor/st2reactor/cmd/timersengine.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def _run_worker():
+    LOG.info('(PID=%s) TimerEngine started.', os.getpid())
+
+    timer = None
+
+    try:
+        timer_thread = None
+        if cfg.CONF.timer.enable or cfg.CONF.timersengine.enable:
```
Kami: It would be a bit config is one is False and other is True, so perhaps we should be more explicit and throw in such scenario? Or?

Kami:
Two small comments, besides that, LGTM.

st2reactor/st2reactor/cmd/timersengine.py
lakshmi-kannan: Correct. +1 to documenting it in monitoring docs.

st2reactor/st2reactor/cmd/timersengine.py
lakshmi-kannan: With default configs, you'll always have this behavior. So we have to actually detect if those variables are defined in /etc/st2/st2.conf. I think that's not really required. We have documented in upgrade notes. And when people upgrade, the configuration change diff will be shown to them too. So there are multiple checks and alerts to the user.

Kami:
I will take over that and try to get it finished and merged this week.

st2reactor/st2reactor/cmd/timersengine.py
Kami: One thing which is also config is that because we default one option to True, when user wants to disable the service they need to set **both** options to False.

I will document that (if not already).

Kami:
Pushed some fixes and test changes - 2422b8c74ff14557e2b495f043a96e1560311edb, 04d0d983e4d474c310300705c027ad3cb79b453c, 21dea06d7a5d91e082b3d4281c7d864a87837ce1.

As mentioned above, I'm still not a fan of this duplicated `enable` option which makes for a confusing behavior, but it is what is is now.

Kami:
Pushed some fixes and test changes - 2422b8c74ff14557e2b495f043a96e1560311edb, 04d0d983e4d474c310300705c027ad3cb79b453c, 21dea06d7a5d91e082b3d4281c7d864a87837ce1.

As mentioned above, I'm still not a fan of this duplicated `enable` option which makes for a confusing behavior, but it is what is is now.

Kami:
There is a chicken and the egg problem with our e2e tests - they depend on st2-packages changes. 

But for that changes to work, this PR needs to be merged first.

PR #4184, opened 2018-06-20
Add jinja filter to manipulate paths

reisingerf:
I was looking for a jinja filter that could get me the basename or dirname of a linux path string. I have outlined in this pull request how that could look like. Happy for any feedback!

PR #4187, opened 2018-06-21
Add jinja filter to manipulate paths (with changelog entry)

Kami:
This pull request adds entry for dictionary in `st2common.util.jinja.get_filters` and also adds a corresponding changelog entry on top of #4184.

I decided to base this new PR on top of the existing one so we can merge that today and include it in v2.8.0.

Resolves #4184.

### DIFF
--- a/st2common/st2common/util/jinja.py
+++ b/st2common/st2common/util/jinja.py
@@ -62,6 +62,7 @@ def get_filters():
     from st2common.jinja.filters import version
     from st2common.jinja.filters import json_escape
     from st2common.jinja.filters import jsonpath_query
+    from st2common.jinja.filters import path
 
     # IMPORTANT NOTE - these filters were recently duplicated in st2mistral so that
     # they are also available in Mistral workflows. Please ensure any additions you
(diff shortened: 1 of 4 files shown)


st2common/st2common/util/jinja.py
```diff
@@ -94,7 +95,10 @@ def get_filters():
         'use_none': use_none,
 
         'json_escape': json_escape.json_escape,
-        'jsonpath_query': jsonpath_query.jsonpath_query
+        'jsonpath_query': jsonpath_query.jsonpath_query,
+
+        'basename': path.basename,
+        'dirname': path.dirname
```
blag: Just a quick nit: Python ignores trailing commas in lists, tuples, and dictionary key/value...lists, so if we add a trailing comma here now, we don't have to see future diffs including silly things like:

```diff
-        'jsonpath_query': jsonpath_query.jsonpath_query
+        'jsonpath_query': jsonpath_query.jsonpath_query,
```

Kami: Heh, yeah, just a personal preference here as well :)

I know there are two sides in Python (and more so in JavaScript for a different reason) - trailing comma and not, I'm personally on the "not" side :D

PR #4194, opened 2018-06-22
Fix wrong information for /api/v1/apikeys/ doc

longfei-zhang:


longfei-zhang:
@LindsayHill  The CI failed, but I don't think it's due to my commit. What do you think?

PR #4196, opened 2018-06-23
Rename a misleading variable name

NikosVlagoidis:
Rename pack_uninstall_request variable to pack_config_values as it was confusing.

### DIFF
--- a/st2common/st2common/openapi.yaml
+++ b/st2common/st2common/openapi.yaml
@@ -8,7 +8,7 @@ info:
   version: "1.0.0"
   title: StackStorm API
   description: |
-    
+
     ## Welcome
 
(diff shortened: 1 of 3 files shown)


Kami:
Good catch, thanks :+1:

st2common/st2common/openapi.yaml
```diff
@@ -8,7 +8,7 @@ info:
   version: "1.0.0"
   title: StackStorm API
```
Kami: This file is automatically generated from `openapi.yaml.j2` one.

Please just make the change to the `.j2` file, run `make generate-api-spec` task and add and commit the generated file.

This will fix the tests / check which is currently failing and preventing us to merge this.

NikosVlagoidis: Ok thanks for the info. I updated my commit.

PR #4198, opened 2018-06-25
Add changelog entry for Orchestra

m4dcoder:
Add changelog entry for Orchestra to let user knows the feature is beta.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,7 @@ in development
 Added
 ~~~~~
 
+* Orchestra - new StackStorm-native workflow engine. This is currently in **beta**. (new feature)
 * Added metrics for collecting performance and health information about the various ST2 services
   and functions. (new feature) #4004 #2974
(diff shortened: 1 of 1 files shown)


CHANGELOG.rst
```diff
@@ -7,6 +7,7 @@ in development
 Added
 ~~~~~
 
+* Orchestra - new StackStorm-native workflow engine. This is currently in **beta**. (new feature)
```
cognifloyd: This doesn't say that it "will eventually replace ActionChain and Mistral." like the PR says

LindsayHill: This is deliberate, to avoid confusion from folks who just read the release notes, and might assume that Action Chains + Mistral workflows are no longer supported.

Better to wait until Orchestra gets to GA, and a deprecation timeline is established for ActionChain/Workflows.

m4dcoder: I updated the description above.

PR #4207, opened 2018-06-27
Fix race between task and action execution

m4dcoder:
Move the update of the task execution before the action execution is requested. Currently, the task execution is updated after the request and there are cases where the action execution completed very quickly and the completion of the action execution is processed first. When completion is processed, the task execution is also updated and this causees conflict with the previous task execution update.


Kami:
Is it possible to add test case for it?

m4dcoder:
@Kami I did think about it. I don't currently know how to write one correctly in the context of the workflow execution. I can dissect it and write unit test for it but then it wouldn't be accurate testing the logic in that function.

Kami:
I'll approve it for now, but it would be great if we can come up with some kind of test case in the near future.

PR #4208, opened 2018-06-27
Update Execution model in openapi

m4dcoder:
As part of the Orchestra changes, workflow_execution and task_execution attributes are added to the ActionExecution model. Add these two optional fields to the openapi spec.


Kami:
It looks like we some how missed merging that for v2.8.0.

In any case, I caught and fixed that in other PR (https://github.com/StackStorm/st2/pull/4227) so the fix is in master and v2.8 branch now.

I will go ahead and close this issue.

Issue #3367, opened 2017-04-19
st2client missing python dependencies


jwomack:
I needed to go through stack traces to find the following dependencies:
pytz
editor
jsonschema
prompt_toolkit

Fixed by:
sudo pip install pytz
sudo pip install editor
sudo pip install jsonschema
sudo pip install prompt_toolkit

LindsayHill:
How did you install st2client? Some of those dependencies - e.g. pytz - are covered in https://github.com/StackStorm/st2/blob/master/st2client/requirements.txt

cyberious:
I too am unable to get st2client installed, have tried both Python 2.7 or Python 3.6 using pip install and falling down a rabbit whole of dependency. Installing a OS X.

```
Traceback (most recent call last):
  File "/usr/local/bin/st2", line 7, in <module>
    from st2client.shell import main
  File "/usr/local/lib/python2.7/site-packages/st2client/shell.py", line 39, in <module>
    from st2client.commands import pack
  File "/usr/local/lib/python2.7/site-packages/st2client/commands/pack.py", line 18, in <module>
    import editor
ImportError: No module named editor
```
Finally got working after manually running `sudo pip install python-editor`

enykeev:
The dependency is [there](https://github.com/StackStorm/st2/blob/master/st2client/requirements.txt#L7) though. I would also like to know how you're installing st2client, it might that we missed something during, say, packaging....

cyberious:
Just using `pip install st2client`

cognifloyd:
The problem occurs when installing via pypi:
```pip install st2client --user```

The reason is that the requirements in [st2client/setup.py](https://github.com/StackStorm/st2/blob/v2.2.1/st2client/setup.py#L54-L61) do not match with [st2client/requirements.txt](https://github.com/StackStorm/st2/blob/v2.2.1/st2client/requirements.txt).

These are in requirements.txt, but not setup.py 

- `argcomplete`
- `pytz`
- `sseclient`
- `python-editor`
- `prompt-toolkit`

These version requirements don't match (requirements.txt vs setup.py):

- `requests[security]<2.12,>=2.11.1` vs `requests<3.0,>=2.7.0`

cognifloyd:
To avoid setup.py and requirements.txt getting out of sync, requirements.txt could be replaced with something simpler, as described here: https://caremad.io/posts/2013/07/setup-vs-requirement/

[...]

PR #4209, opened 2018-06-28
Make st2client setup.py requirements consistent

LindsayHill:
Installing st2client using `pip install st2client` has missing dependencies. It is unusable until you run `pip install argcomplete pytz jsonschema prompt_toolkit python-editor`.

This PR makes the setup.py requirements consistent with those in requirements.txt.

Fixes https://github.com/StackStorm/st2/issues/3367


bigmstone:
@Kami Is the reason you disabled pulling from `requirements.txt` in #2423 still valid or can we re-enable that? It'd be a better fix than trying to manually keep st2client in sync.

Kami:
We fixed `dist_utils.py` so it also works with pip 10.0.0 so I would be OK with reading from `requirements.txt` again if we can indeed verify it works correctly.

Kami:
I did some digging in and testing - we now install pip 9.0.1 inside st2 and pack virtual environments so I believe we should be fine.

It doesn't work with very ancient versions of pip, but I just checked and most other software also doesn't support pip < 9.0 so we should be fine.

I will push a change which updates it to use `requirements.txt` again.

@LindsayHill thanks for opening this PR and having us reevaluate this decision again.

I also tagged this issue with 2.9.0 and will wait with merging this we have a branch for v2.8.

Kami:
I pushed various changes which should resolve that - f9455e235e6304f7951431747f94bcb77824b2ad, 8c6659ccfdfcbb561ba772fd9ae1cab85e0e96df, 79d0d1399f87d8e5588cca0ed587f84673b1e73b.

It should work with pip >= 6.0.0, but we really shouldn't support anything < 9.0.0 so if a user complains, we should ask them to upgrade / use more recent version of pip.

LindsayHill:
Thanks for that @Kami. Totally fine for this to be 2.9.0. It's not a 2.8.0 blocker.

Kami:
I pushed some more changes / fixes to this branch and confirmed it indeed fixes issue reported in #3367 - https://github.com/StackStorm/st2/issues/3367#issuecomment-402174599.

PR #4211, opened 2018-06-28
Fixing runner versions

bigmstone:
Runners now get their version dynamically from `st2common`'s version variable.


Kami:
That works for me.

We just need to be careful if we ever publish runner packages to PyPi that we also publish st2common.

I assume we plan to merge this post v2.8.0, or?

Kami:
I wonder if that will break the wheel build changes you made in https://github.com/StackStorm/st2-packages/pull/572/.

I can merge it and see. If it breaks, I can revert and and we find a better approach for v2.9.0.

Kami:
@bigmstone It looks like the build is failing.

It's a chicken and the egg problem (st2common is not present), so this needs more work.

Kami:
Looking at the build failure and package build pipeline - we can probably make it work by adding st2common as a dependency for all the runner packages or maybe by injecting st2common as a dependency into every runner package during the build, similar as we do for st2client - https://github.com/StackStorm/st2-packages/blob/306b826fde833f1ee9e730974f2953a4fde0e270/packages/st2/Makefile#L110.

We just need to be careful it doesn't introduce some kind of cyclic dependency and weird issues down the road.

If possible, I would prefer first approach since it's more explicit and transparent compared to injecting a dependency during build time.

Kami:
@bigmstone https://circleci.com/gh/StackStorm/st2/8002?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

So as I assumed, the problem is indeed a cyclic dependency and this change - https://github.com/StackStorm/st2-packages/pull/572

We have st2 depending on `stackstorm-runner-orchestra` and runner itself depends on `st2common` which causes a cyclic dependency.

Perhaps the best way might indeed be to simply update those versions inside the release workflows.

Kami:
Superseded by #4239.

Issue #4186, opened 2018-06-21
`st2 inquiry` is not using `ST2_API_URL` environment variable


shusugmt:
##### SUMMARY
`st2 inquiry` always tries to connect to 127.0.0.1:9101 regardless of the value set in `ST2_API_URL`

##### ISSUE TYPE
 - Bug Report

##### STACKSTORM VERSION
`st2 2.7.2, on Python 2.7`

##### OS / ENVIRONMENT / INSTALL METHOD
- `pip install st2client` on CentOS 7

##### STEPS TO REPRODUCE

```
export ST2_API_URL=https://st2.example.com
st2 login hoge
st2 run core.ask
st2 inquiry list
```

##### EXPECTED RESULTS

Connect to the API endpoint specified by `$ST2_API_URL` and retrieves list of inquiries.

##### ACTUAL RESULTS

It tries to connect to 127.0.0.1:9101 and just fails (since our st2 instance is running on remote machine).

```
st2 inquiry list
ERROR: HTTPConnectionPool(host='127.0.0.1', port=9101): Max retries exceeded with url: /exp/inquiries/?limit=50 (Caused by NewConnectionError('<requests.packages.urllib3.connection.HTTPConnection object at 0x25239d0>: Failed to establish a new connection: [Errno 111] Connection refused',))
```

armab:
Thanks for reporting this.
Sounds indeed nasty if ST2_API_URL is not taken for inquiries.

LindsayHill:
The issue is that Inquiries is still marked as "experimental", so the URL for `st2 inquiries list` looks like `http://127.0.0.1:9101/exp/inquiries/?limit=50`, or for remote requests it would look like `https://my.st2.com/api/exp/inquiries/?limit=50`

In the [st2client code](https://github.com/StackStorm/st2/blob/master/st2client/st2client/client.py#L62), it always maps this to `ST2_BASE_URL:DEFAULT_API_PORT/exp`. Typically this would end up being `http://127.0.0.1:9101/exp`. If the config option `base_url` or environment option ST2_BASE_URL is set, this might end up being `https://my.st2.com:9101/exp`.

This is the code snippet:
```python
self.endpoints['exp'] = '%s:%s/%s' % (self.endpoints['base'], DEFAULT_API_PORT, 'exp')
```

It probably needs similar handling to the api_url:
```python
        if api_url:
            self.endpoints['api'] = api_url
        else:
            self.endpoints['api'] = os.environ.get(
[...]

PR #4212, opened 2018-06-28
[WIP] ST2_EXP_URL parameter

LindsayHill:
`st2 inquiry list` does not support `ST2_API_URL`. The broader problem is because it uses the `/exp` API, and st2client doesn't handle that.

This WIP code adds support for an `ST2_EXP_URL` parameter. If that's set to say `https://my.st2.com/api/exp`, then you can make remote `st2 inquiry list` calls.

Not sure if this is the right path here or not. Only done minimal testing so far.

Addresses https://github.com/StackStorm/st2/issues/4186


armab:
Having `EXP_URL` sounds pretty nasty, but it's a good start for discussion.

@Kami Guessing what's the context of `exp`, is it kind of "experimental" API versioning path which was added for Inquiries only?
Considering `/exp/` only adds functionality to `/v1/` and doesn't modify existing endpoints behavior in a breaking change (according to https://api.stackstorm.com/), can we merge `exp` into `v1` maybe?

LindsayHill:
I think there's a few issues/questions here:

1/ No way for users to use st2client to specify server where `/exp` API is hosted. 
2/ Inquiries is still using `/exp` API, but can probably graduate to `/v1`.
3/ Ongoing role of `/exp` vs `/v1`, when one should be used over the other, etc.

I think we need to address all of those. This PR is one way of dealing with the first one.

On the second one, I think we should do some cleanup work on inquiries in 2.9, and move it out of `/exp`. 

For the third one, I think that the reasoning behind using `/exp` here was because the nature of the API might change. Technically it would not break anything moving it to `/v1`, since it is (in this case) additive, but by leaving it behind `/exp`, we are explicitly stating "This API can, and probably will change in the short term". If we move it to `/v1`, users can have an expectation that there will be relatively few changes to that specific API in future, unless we go to `/v2`. So I think there is an ongoing role for an `/exp` API, but things should not stay there forever. We've probably left inquiries there for longer than we should have.

nmaludy:
Maybe a deeper root cause is the fact that things aren't really "based" off of the `base_url`.  Right now, when connecting from an external host, you need to specify several different URLs.

```
    client = st2client.client.Client(base_url="https://{}/".format(server),
                                     auth_url="https://{}/auth/v1".format(server),
                                     api_url="https://{}/api/v1".format(server))
```

Instead, if the code actually used the `base_url` as a base, then the `exp` URL might work as expected?

LindsayHill:
Proper fix here: https://github.com/StackStorm/st2/pull/4495

PR #4213, opened 2018-06-28
Fix user in Orchestra st2kv function

m4dcoder:
There are mix use of username nad user model in the st2kv function. Get the user object from the username in the Orchestra context, pass the username form the user object when determining scope, and use the user object on checking roles. Rewrite the unit tests, remove all the mocks, and use datastore functionality to test st2kv. Add a temporary symlink to the integration test module so make orchestra-itests will run the st2kv integration tests.


st2common/st2common/util/keyvalue.py
```diff
@@ -78,13 +79,14 @@ def get_key(key=None, user=None, scope=None, decrypt=False):
...
         raise TypeError('Given key is not typeof string.')
+
     if not isinstance(decrypt, bool):
         raise TypeError('Decrypt parameter is not typeof bool.')
 
     if not user:
         user = UserDB(cfg.CONF.system_user.user)
 
-    scope, key_id = _derive_scope_and_key(key, user, scope)
+    scope, key_id = _derive_scope_and_key(key, user.name, scope)
```
Kami: So it looks like we changed function signature - it looks like previously user was expected to be username string and not the actual `UserDB` object.

Would be good to document that via docstring and perhaps also add an assert guard.

Kami: Actually, looking at the code more, it seems like there is no need for the whole `UserDB` object so I would just prefer to pass in `user.name` instead of the whole user object.

st2common/tests/unit/test_util_keyvalue.py
```diff
@@ -141,7 +142,7 @@ def test_get_key(self, deseralize_key_value, KeyValuePair):
         KeyValuePair.get_by_scope_and_name().value = value
         deseralize_key_value.return_value = value
 
-        result = kv_utl.get_key(key=key, user=USER, decrypt=decrypt)
+        result = kv_utl.get_key(key=key, user=auth_db.UserDB(name=USER), decrypt=decrypt)
```
blag: Do you need `.name` here on `auth_db.UserDB(name=USER)`?

st2common/st2common/util/keyvalue.py
Kami: And if we decide to stick with `UserDB` object, I would prefer to call the argument `user_db` so it's consistent with other places and makes it immediately obvious what kind of argument it accepts.

Kami: I just had a look at the Slack discussion again and it looks like some other functions require actual `UserDB` object.

If that's the case, I'm fine with that change, I would just prefer to rename argument from `user` to `user_db` to make it immediately clear and obvious (other function also takes `user_db` argument).

m4dcoder: @Kami The get_key function is kind of a mess.
1. We see user being set to UserDB by default if user not pass at https://github.com/StackStorm/st2/blob/master/st2common/st2common/util/keyvalue.py#L85
2. Deriving scope is expecting a username string at https://github.com/StackStorm/st2/blob/master/st2common/st2common/util/keyvalue.py#L87
3. And RBAC role check requires a UserDB at https://github.com/StackStorm/st2/blob/master/st2common/st2common/util/keyvalue.py#L95

st2common/tests/unit/test_util_keyvalue.py
m4dcoder: Nope. This unit test is doing a shortcut and passing in a UserDB object.

PR #4219, opened 2018-07-01
Add filter actions by tag

NikosVlagoidis:
This PR adds the functionality to filter actions on `/api/v1/actions` endpoint by tags.
For example:
https://{host}/api/v1/actions?tags=test

Will return all actions that are tagged with the name test.


Kami:
Thanks, looks good to me.

It looks like the user can only filter by a single tag? And not multiple ones using "OR" approach?

If so, I'm fine with that since we don't really have a good existing convention for handling "OR" filters across the API (although we do have plans to work on that in the future).

Kami:
Please also add a corresponding changelog entry.

NikosVlagoidis:
Changelog entry added!

Kami:
Merged - thanks!

Issue #4220, opened 2018-07-02
Fix datastore default scoping inconsistencies between CLI and Mistral and Orchestra workflows


Kami:
Right now CLI defaults to "system" scope when setting, getting and listing values and Mistral and Orchestra default to "user" scope when retrieving a datastore value inside a workflow when a scope is not explicitly provided.

I first thought that's a regression / change of behavior in Orchestra, but it seems like it was like that in Mistral for a while now as well - https://github.com/StackStorm/st2mistral/blob/master/st2mistral/functions/stackstorm.py#L54

For example (all defaults to system scope by default):

```bash
st2 key list
set key set <name> <value>
st2 key get <key name>
```

When retrieving a datastore value inside Mistral / Orchestra workflow:

```yaml
cmd: "echo <% st2kv("keyname") %>"
```

Here it defaults to a user scope if scope is not explicitly provided (e.g. `system.keyname`, `user.keyname`).

It's confusing because if you do ``st2 key set foo bar`` and inside a workflow you do `<% st2kv("key") %>` you will get "key foo doesn't exist" error back.

I personally think defaulting to `system` scope (like in the CLI) make more sense since it's more commonly used one.

What do others think? The main problem is that this change would be backward incompatible since it changes the behavior.

We wouldn't have this problem if we only supported explicit notation (aka scope must be always provided).

It's also a bit weird that we default to user scope - if anything we should default to system scope for backward compatibility reasons when we only supported unscopped (system) values (I guess we botched that one when we added support for user scoped values).

lakshmi-kannan:
I absolutely think the default scope should be system scope unless the user explicitly says st2kv.user(). In mistral case IIRC, due to implementation, we could not have two functions like st2kv.system("foo") and st2kv.user("foo"). I think the problem was with the dots. I don't know if we can fix this in orchestra. I'd prefer a more explicit version with scope called out. I am also ok with st2kv_system("foo") and st2kv_user("foo"). We don't have to be compatible with mistral or action chain. We have to only think what would be a great UX and just drive our implementation to that.

PR #4221, opened 2018-07-02
Default to "all" scope when using "st2 key list" CLI command

Kami:
This pull request updates ``st2 key list`` CLI command to default to ``all`` scope when listing datastore values (previously it defaulted to ``system``).

This means that by default, both - user and system scoped datastore values will be displayed.

I think defaulting to ``all`` makes more sense and reduces some confusion. Most CLI arguments are there for additional filtering aka displaying less results and not more such as here where you currently need to pass ``--scope=all`` argument to the command to display all the values.

Partially related to #4220.


nmaludy:
Might want to add a `st2docs` PR since `all` scope is not documented: https://docs.stackstorm.com/datastore.html

Kami:
@nmaludy Thanks for the reminder, will add it.

Kami:
Docs at https://github.com/StackStorm/st2docs/pull/762.

PR #4223, opened 2018-07-03
Use yaml.safe_load instead of yaml.load

Kami:
This Tweet made me grep our code for `yaml.load` - https://twitter.com/anthonypjshaw/status/1014054816987803648

I was almost positive we already use `yaml.safe_load` everywhere, but it looks like a couple of `yaml.load` occurrences managed to creep into the tests.

It's not a big deal to do that in the tests. It's mostly problematic in places where we take user input, but it's still a good practice to use `yaml.safe_load` if there is no special reason / need to use `yaml.load`.

Thanks @tonybaloney


armab:
@tonybaloney https://github.com/StackStorm/st2/pull/4222 was first :stuck_out_tongue:

Kami:
@armab ah, lol, I didn't even see that one. Will close this one as a duplicate of #4222.

Or maybe Github just doesn't correctly handle nano second precision and I was actually the first one :trollface:

armab:
looks like it's 6 mins in diff between 2 PRs :)

Kami:
@armab No way - must be that we were routed to two different backend servers with local clocks out of sync :trollface:

armab:
Haha, that's a good one :+1:

PR #4224, opened 2018-07-03
Fix "proxy" authentication mode

Kami:
While looking into why proxy authentication mode is not working for a user (https://forum.stackstorm.com/t/how-to-enable-the-proxy-authentication-mode/198), I discovered we inadvertently broke it while moving to the OpenAPI.

This pull request fixes that.

## TODO

- [ ] Unit tests for /v1/tokens API controller (existing test was mocking API layer so we didn't actually test headers are read in correctly)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,12 @@ Changelog
 in development
 --------------
 
+* Fix ``proxy`` st2auth authentication mode which has been inadvertently broken when we moved to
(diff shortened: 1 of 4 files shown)


Kami:
EDIT: After more digging in - the existing code is indeed correct.

We want to read `remote_addr` and `remote_user` from environment which is passed to auth server by a proxy and not from a header. I will fix the documentation.

Issue #4235, opened 2018-07-10
Secret object parameters in the CLI and Web are no longer redacted


jjm:
##### SUMMARY

Following an upgrade to ST2 2.8.0, I found parameters with an type of `object` and marked as `secret: true` were not being redacted for both the WebUI and CLI.  

##### ISSUE TYPE
 - Bug Report

##### STACKSTORM VERSION
```
st2 2.8.0, on Python 2.7.6
```

##### OS / ENVIRONMENT / INSTALL METHOD
docker via st2-docker.

##### STEPS TO REPRODUCE
Have an action (e.g. `aws_boto3.boto3action`) that takes an object as a parameter and marked as `secret: true`.

When running the action with the following parameters :

```yaml
parameters:
  region:
    type: "string"
    description: "Region where action is performed"
    default: "eu-west-1"
  credentials:
    type: "object"
    description: "An AWS credentials object"
    secret: true
```

##### EXPECTED RESULTS
The secret parameter should be redacted, the following output is from `v2.7.x`:

```yaml
parameters:
  credentials: '********'
  region: eu-west-1
```

##### ACTUAL RESULTS

```yaml
id: 5b44b87390be8400f0b121b7
status: succeeded
parameters:
  credentials:
    Credentials:
      AccessKeyId: <deleted-non-redacted-key>
      Expiration: '2018-07-10 14:42:52'
      SecretAccessKey: <deleted-non-redacted-key>
      SessionToken: deleted-non-redacted-key
  region: eu-west-1
result:
  exit_code: 0
  result:
    allocation_ids: []
    no_association_ids: []
  stderr: ''
  stdout: ''
```

jjm:
A little testing and the following simpler action configuration:

```yaml
  credentials:
    type: "object"
    description: "Response from assume role"
    properties:
      AccessKeyId:
       type: string
       secret: true
      Expiration:
       type: string
      SecretAccessKey:
       type: string
       secret: true
      SessionToken:
       type: string
       secret: true
```

Produces the following output as expected:

```
id: 5b44c1ac90be8400f0b121c3
status: failed
parameters:
  credentials:
    AccessKeyId: '********'
    Expiration: '2018-07-10 14:42:52'
    SecretAccessKey: '********'
    SessionToken: '********'
  region: eu-west-1
```

Plus the following configuration:

```yaml
  credentials:
    type: "object"
    description: "Response from assume role"
    properties:
      Credentials:
        type: "object"
        properties:
[...]

PR #4236, opened 2018-07-10
Fix bug where objects and arrays marked as secret weren't being masked

nmaludy:
Most likely due to the re-work of secret masking here https://github.com/StackStorm/st2/pull/4140 `objects` and `arrays` that were marked with `secret: true` were not being masked. See https://github.com/StackStorm/st2/issues/4235 for details on the symptoms. 

This PR addresses this by looking for the `secret` property on `object` and `array` types. It then treats all data within the `object` or `array` as secret, returning a single masked value.

Example of the fixed version:

```
$ st2 run default.secret_objects secret_object='{"abc": "123"}' secret_array='[1, 2, 3]'
.
id: 5b44de76a814c004961c020e
status: succeeded
parameters: 
  secret_object: '********'
  secret_array: '********'
result: 
  failed: false
  return_code: 0
  stderr: ''
  stdout: ''
  succeeded: true
```

FYI, i create a unit test for this edge case so we can verify that it is not exposed in the future.

### DIFF
--- a/st2common/st2common/util/secrets.py
+++ b/st2common/st2common/util/secrets.py
@@ -66,6 +66,17 @@ def get_secret_parameters(parameters):
 
     secret_parameters = {}
     parameters_type = parameters.get('type')
+    # If the parameter itself is secret, then skip all processing below it
+    # and return the type of this parameter.
+    #
+    # **This causes the _full_ object / array tree to be secret (no children will be shown).**
+    #
+    # **Important** that we do this check first, so in case this parameter
+    # is an `object` or `array`, and the user wants the full thing
+    # to be secret, that it is marked as secret.
+    if parameters.get('secret', False):
+        return parameters_type
+
     iterator = None
     if parameters_type == 'object':
         # if this is an object, then iterate over the properties within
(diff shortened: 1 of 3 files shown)


nmaludy:
Closes https://github.com/StackStorm/st2/issues/4235

nmaludy:
@Kami i was able to find and fix the issue, please let me know if you have any comments

Kami:
@nmaludy Thanks, will have a look.

st2common/tests/unit/test_util_secrets.py
```diff
@@ -669,3 +701,39 @@ def test_mask_secret_parameters_nested_array_with_object(self):
             ]
         }
         self.assertEqual(expected, result)
+
+    def test_mask_secret_array(self):
```
Kami: It also wouldn't hurt to add a test case for scenarios where we have an object which is marked "secret: true" at top level, but it also has a sub-schema with different items of which some have "secret: true" attribute. Same goes for array.

I assume in such scenarios, if top level "secret: true" attribute is present, this top level attribute should have a precedence (and we could even ignore / skip iterating over child object / array attributes).

st2common/st2common/util/secrets.py
```diff
@@ -66,6 +66,17 @@ def get_secret_parameters(parameters):
...
+    # If the parameter itself is secret, then skip all processing below it
+    # and return the type of this parameter.
+    #
+    # **This causes the _full_ object / array tree to be secret (no children will be shown).**
+    #
+    # **Important** that we do this check first, so in case this parameter
+    # is an `object` or `array`, and the user wants the full thing
+    # to be secret, that it is marked as secret.
+    if parameters.get('secret', False):
+        return parameters_type
```
Kami: Great :+1: 

That's also a more common scenario and it's much more performant :)

Kami:
Thanks for good work on the tests :+1:

PR #4239, opened 2018-07-12
Update runner packages to read version from runner package local __init__.py file

Kami:
This pull request updates all the runner packages to they read their version from the runner package local `__init__.py` file.

Sadly we can't go with approach in #4211 due to a cyclic dependency issue (which right now only exists for orchestra runner, but once we move to dynamic stevedore loading will exist for every runner package if we make runners depend directly on st2common) - https://github.com/StackStorm/st2/pull/4211#issuecomment-404437539.

This way it's consistent with `st2client`.

Second part of this change is updating st2cd release workflow so it also updates version for all the runner packages (in the same place where we set version for st2common and st2client).

### DIFF
--- a/contrib/runners/action_chain_runner/setup.py
+++ b/contrib/runners/action_chain_runner/setup.py
@@ -23,6 +23,8 @@ from setuptools import find_packages
 from dist_utils import fetch_requirements
 from dist_utils import apply_vagrant_workaround
 
+from action_chain_runner import __version__
+
 BASE_DIR = os.path.dirname(os.path.abspath(__file__))
 REQUIREMENTS_FILE = os.path.join(BASE_DIR, 'requirements.txt')
 
@@ -31,7 +33,7 @@ install_reqs, dep_links = fetch_requirements(REQUIREMENTS_FILE)
 apply_vagrant_workaround()
 setup(
     name='stackstorm-runner-action-chain',
-    version='2.5.0',
+    version=__version__,
     description=('Action-Chain workflow action runner for StackStorm event-driven '
                  'automation platform'),
     author='StackStorm',
--- a/contrib/runners/action_chain_runner/action_chain_runner/__init__.py
+++ b/contrib/runners/action_chain_runner/action_chain_runner/__init__.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 # Licensed to the StackStorm, Inc ('StackStorm') under one or more
 # contributor license agreements.  See the NOTICE file distributed with
 # this work for additional information regarding copyright ownership.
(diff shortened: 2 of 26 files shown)


contrib/runners/action_chain_runner/setup.py
```diff
@@ -23,6 +23,8 @@
 from dist_utils import fetch_requirements
 from dist_utils import apply_vagrant_workaround
 
+from action_chain_runner import __version__
+
 BASE_DIR = os.path.dirname(os.path.abspath(__file__))
 REQUIREMENTS_FILE = os.path.join(BASE_DIR, 'requirements.txt')
```
armab: Since import doesn't work, can the version be read from the `../../../../../st2common/__init__.py` via similar "file" approach?
How worse or more hacky that could be, comparing to hardcoding same version number in all the runners?

Kami: This would kinda break the whole package approach - it would rely on st2common to be in a specific path and always available with a runner package.

And it's not really hard coding - it's being updated via release workflows. It's exactly the same approach we follow for st2common and st2client. 

And this also allows us to separately version and release runner packages in the future (if we ever need to do that).

armab: I understand the purpose of st2common and st2client being versioned separately, but don't get the reason (yet) for runners which are part of the core and are in st2 repo code.

Any future plans to do extract these runners from the core and distribute separately?

Kami: No plans for that right, but with package approach we do get more flexibility which allows us to do things like that in the future.

I believe @bigmstone did talk something about moving runners to standalone repos, but I personally don't want to do that any time soon yet (too much overhead).

Kami: And also, with those changes they are not really versioned separately / interdependently - as mentioned above, it follows the same approach as other components and we bump / set the version during the release process.

armab: I just feel it's a bit of mess to change the version in 10 files with the same number without having plans to extract these parts from the core repo.
If there are no other options to brainstorm, - let it be :+1: 

Would just like @bigmstone or @m4dcoder to jump in and review the PR, as someone more familiar with st2 core and plans in context of runners.

Kami:
@bigmstone Thanks.

Versions don't need to match, but right now everything is coupled so there is no reason for versions not to match :)

The change you did for orchestra runner doesn't specify a version. Going forward with rest of the runners, we could handle it in the same way as we handle `st2client` and inject runner dependencies with a version which matches StackStorm version.

Since runner packages are not published on PyPi (unlike st2client), this is not strictly needed, but it shouldn't hurt either.

armab:
I like @bigmstone point that, in fact, version for runners don't need to match and question why are we incrementing them, practically. That definitely makes more sense to me, - incrementing version when the content really changed, not just for the sake of versioning.

Anyways, not blocking. Hopefully it will come up with something good in the future :man_shrugging:

PR #4242, opened 2018-07-13
Allow user to force terminal size used by the CLI formatters by setting environment variable

Kami:
Right now, we try various approaches to obtain terminal size used by the CLI for table formatting purposes.

If none of that approaches work, we fall back to a default hard coded size (10 lines, 20 columns).

This pull request improves on that in two ways:

1. It uses a more reasonable default size. `20` columns for a total terminal width is simply a bad default. It means pretty much every table cell / column will wrap at couple of characters which will make output unreadable when we are unable to retrieve a terminal size.
2. Allows user to force terminal size using `ST2_CLI_FORCE_TERMINAL_SIZE=<lines,columns>` environment variable.

We have experienced this "issue" a lot in our end to end tests for a long time. Since those test commands don't run using shell, we fail to obtain a terminal size and we fall back to a really bad default which results in an unreadable output like that:

Before:

screenshot from 2018-07-13 14-07-45

After:

screenshot from 2018-07-13 14-08-06

Keep in mind that those actions run on a remote host which is later destroyed so we can't simply re-retrieve the output which means we are stuck with troubleshooting / debugging from this very hard to read output.

## TODO

- [x] Changelog entry

### DIFF
--- a/st2client/st2client/utils/terminal.py
+++ b/st2client/st2client/utils/terminal.py
@@ -27,7 +27,7 @@ __all__ = [
 ]
 
 
-def get_terminal_size(default=(80, 20)):
+def get_terminal_size(default=(80, 200)):
     """
     :return: (lines, cols)
     """
--- a/st2client/st2client/utils/terminal.py
+++ b/st2client/st2client/utils/terminal.py
@@ -26,55 +26,53 @@ DEFAULT_TERMINAL_SIZE_LINES = 80
 DEFAULT_TERMINAL_SIZE_COLUMNS = 150
 
 __all__ = [
-    'get_terminal_size'
+    'get_terminal_size_columns'
 ]
 
 
-def get_terminal_size(default=(DEFAULT_TERMINAL_SIZE_LINES, DEFAULT_TERMINAL_SIZE_COLUMNS)):
+def get_terminal_size_columns(default=DEFAULT_TERMINAL_SIZE_COLUMNS):
     """
-    Try to retrieve a default terminal size using various system specific approaches.
+    Try to retrieve COLUMNS value of terminal size using various system specific approaches.
 
     If terminal size can't be retrieved, default value is returned.
 
-    NOTE 1: LINES and COLUMNS environment variables are checked first, if those values are not set /
-    available, other methods are tried.
+    NOTE 1: COLUMNS environment variable is checked first, if the value is not set / available,
+            other methods are tried.
 
-    NOTE 2: This method requires both environment variables to be specified together.
-
-    :return: (lines, cols)
+    :rtype: ``int``
+    :return: columns
     """
-    # 1. Try LINES and COLUMNS environment variables first like in upstream Python 3 method -
+    # 1. Try COLUMNS environment variable first like in upstream Python 3 method -
     # https://github.com/python/cpython/blob/master/Lib/shutil.py#L1203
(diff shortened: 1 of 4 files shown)


st2client/st2client/utils/terminal.py
```diff
@@ -26,10 +27,20 @@
 ]
 
 
-def get_terminal_size(default=(80, 20)):
+def get_terminal_size(default=(80, 150)):
     """
     :return: (lines, cols)
     """
+    # Allow user to force terminal size using a environment variables
+    # E.g. ST2_CLI_FORCE_TERMINAL_SIZE=80,200 # lines, columns
```
armab: +1 for changing the default limits, but to control the terminal size, we can rely on existing logic, without implementing new `ST2_CLI_FORCE_TERMINAL_SIZE`: 

https://github.com/StackStorm/st2/blob/56f9cec9b9cfc973d4e03c062b9c9905d6fd34db/st2client/st2client/utils/terminal.py#L74-L78

Kami: Yes, it will still rely on those two environment variables if other method fails, but with this new environment variable there is also a way to override those values (e.g. in case some other method return a value, but user wants to use a different one without changing xterm settings, etc.).

Kami:
@arma and I discussed this some more and decided to go with this approach - 471b5ef.

Now we check `LINES` and `COLUMNS` environment variables first so there is no need for additional environment variables.

This way it's also consistent with upstream Python 3 implementation of `get_terminal_size` (https://github.com/python/cpython/blob/master/Lib/shutil.py#L1203).

Keep in mind that we need to maintain our own implementation because this function is not available in Python 2.

st2client/st2client/utils/terminal.py
```diff
@@ -14,32 +14,57 @@
...
+
     :return: (lines, cols)
     """
+    # Try LINES and COLUMNS environment variables first like in upstream Python 3 method -
+    # https://github.com/python/cpython/blob/master/Lib/shutil.py#L1203
+    # This way it's consistent with upstream implementation. In the past, our implementation
+    # checked those variables at the end as a fall back.
+    try:
+        lines = os.environ['LINES']
+        columns = os.environ['COLUMNS']
```
armab: :+1: on approach, but I think original implementation is safer on corner cases:
https://github.com/python/cpython/blob/master/Lib/shutil.py#L1223-L1231

For example, when only `LINES` or only `COLUMNS` can be overridden by the user.

Maybe we can directly use `shutil`?

Kami: Yeah, agreed - our implementation requires both variables to be set (but we actually only use columns and ignore lines all together).

Original implementation allows you to specify those independently. I was thinking of making that change as well, but it's a lot more involved than just doing what I do right now (and previous code also requires both to be specified).

st2client/st2client/utils/terminal.py
```diff
@@ -14,58 +14,80 @@
...
 from __future__ import absolute_import
+
 import os
 import struct
 import subprocess
 import sys
 
 from st2client.utils.color import format_status
 
+DEFAULT_TERMINAL_SIZE_LINES = 80
```
armab: Can remove as never needed

warrenvw:
Thanks. I guess we won't need https://github.com/StackStorm/st2-docker/pull/102 anymore... 👍

PR #4246, opened 2018-07-16
Pin orchestra to v0.1 release

m4dcoder:
Pin the orchestra dependency to v0.1 release so changes to orchestra will not break future v2.8.x patch releases.


Kami:
Thanks :+1:

PR #4247, opened 2018-07-17
Sync changelog with v2.8.1 release

Kami:
The title says it all.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,19 @@ Added
 * Add new ``?tags``, query param filter to the ``/v1/actions`` API endpoint. This query parameter
   allows users to filter out actions based on the tag name . By default, when no filter values are
   provided, all actions are returned. (new feature) #4219
+
+Changed
+~~~~~~~
+
+Fixed
+~~~~~
+
+2.8.1 - July 18, 2018
+---------------------
+
+Added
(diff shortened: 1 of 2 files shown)


CHANGELOG.rst
```diff
@@ -20,6 +20,19 @@ Added
...
+~~~~~~~
+
+Fixed
+~~~~~
+
+2.8.1 - July 18, 2018
+---------------------
+
+Added
+-----
```
blag: Use `~~~~~` here?

Kami: Yep, good catch, will fix it.

Kami: @blag Pushed a fix.

contrib/runners/http_runner/http_runner/runner.yaml
```diff
@@ -24,6 +24,7 @@
     password:
       description: Password required by basic authentication.
       type: string
+      secret: true
```
armab: The title still doesn't say it all :)

armab: As I understand we forgot to include this in master, while having the fix in `v2.8`

Kami: That's correct.

PR #4248, opened 2018-07-17
Allow more general case while searching hubot PID

sergiik:
hubot binary allows passing arguments


LindsayHill:
Can you provide an example of what that `ps ax` output looks like?

sergiik:
1903 ?        Ssl    0:34 node /opt/stackstorm/chatops/node_modules/.bin/coffee /opt/stackstorm/chatops/node_modules/.bin/hubot  st2chatops

In this case there are two spaces between 'hubot'  and st2chatops. That's because no parameters have been passed by startup script https://github.com/StackStorm/st2chatops/blob/master/bin/hubot
But in general case params can be passed.

Usage hubot [options]

Available options:
  -a, --adapter ADAPTER   The Adapter to use
  -c, --create PATH       Create a deployable hubot
  -d, --disable-httpd     Disable the HTTP server
  -h, --help              Display the help information
  -l, --alias ALIAS       Enable replacing the robot's name with alias
  -n, --name NAME         The name of the robot in chat
  -r, --require PATH      Alternative scripts path
  -t, --config-check      Test hubot's config to make sure it won't fail at startup
  -v, --version           Displays the version of hubot installed

PR #4045, opened 2018-03-20
"st2 pack install" command shouldn't need access to pack index

Kami:
A while ago we added a change which displays number of artifacts which are to be registered when running "st2 pack install" command (#3675).

That change inadvertently introduced a regression and dependency of "st2 pack install" command on access to index.stackstorm.org pack index.

This pull request includes two changes / fixes:

* Don't try to access pack index when installing a local pack (pack path starts with "file://")
* Ignore errors related to index access in "st2 pack install" command. This CLI command could run from any host and that host doesn't necessary need access to pack index - only the actual StackStorm instance where action runner runs needs access to index to be able to install a pack.

Now when installing a local pack or remote pack and server where "st2 pack install" is running can't access index, user experience is slightly degraded (graceful degradation) and no "resource counts" are displayed, but installation still works.

Resolves #3771 #3772.

Issue #3771, opened 2017-10-03
st2 pack install file:///<path> may still try to pull from Github


LindsayHill:
If I clone an existing Exchange pack repo, then modify it locally, and install using a local file path, it still pulls the remote version. This is unexpected - if I use `st2 pack install file:///<path>`, I expect it to use the local content.

Steps to reproduce:
1/ `git clone` some Exchange pack
2/ Edit the pack - e.g. change version in `pack.yaml`
3/ Install pack using `st2 pack install file:///home/lhill/mypack`
4/ Observe that installed pack has the Github version, not the local version.

Example:
```bash
vagrant@bwcvagrant:~$ st2 --version
st2 2.4.1
vagrant@bwcvagrant:~$ st2 pack search tesla
+-------+---------------------+---------+--------------+
| name  | description         | version | author       |
+-------+---------------------+---------+--------------+
| tesla | car horn automation | 0.2.1   | Anthony Shaw |
+-------+---------------------+---------+--------------+
vagrant@bwcvagrant:~$ git clone https://github.com/StackStorm-Exchange/stackstorm-tesla
Cloning into 'stackstorm-tesla'...
remote: Counting objects: 102, done.
remote: Total 102 (delta 0), reused 0 (delta 0), pack-reused 102
Receiving objects: 100% (102/102), 33.27 KiB | 0 bytes/s, done.
Resolving deltas: 100% (48/48), done.
Checking connectivity... done.
vagrant@bwcvagrant:~$ cd stackstorm-tesla/
vagrant@bwcvagrant:~/stackstorm-tesla$ ls
actions  aliases  CHANGES.md  circle.yml  config.schema.yaml  icon.png  pack.yaml  README.md  requirements.txt  tesla.yaml.example
vagrant@bwcvagrant:~/stackstorm-tesla$ vim pack.yaml
vagrant@bwcvagrant:~/stackstorm-tesla$ cat ~/stackstorm-tesla/pack.yaml
---
ref: tesla
name : tesla
description : car horn automation
keywords:
  - tesla
version : 0.3.1
author : Anthony Shaw
email : anthony.shaw@dimensiondata.com
vagrant@bwcvagrant:~/stackstorm-tesla$ st2 pack install file:///home/vagrant/stackstorm-tesla

	[ succeeded ] download pack
	[ succeeded ] make a prerun
	[ succeeded ] install pack dependencies
	[ succeeded ] register pack

+-------------+---------------------+
| Property    | Value               |
[...]

PR #4256, opened 2018-07-20
Add a new standalone st2-pack-install CLI command for installing a pack

### LINKED ISSUE #3912: Add utility script which just setup virtualenvs for packs without DB/MQ connection
## Background

In docker-ized environment, there are some use cases that we would like to put all packs, configs and virtualenvs contents baked on the image. Usually `st2ctl reload --register-setup-virtualenvs` can be used to setup virtualenvs for packs, but this script cannot be used in `docker build` step because it requires connection to MongoDB and RabbitMQ which does not exist in the image build process.

## Suggested solution

I've made a sample script:
https://github.com/shusugmt/st2-docker-pack-prebuild-example/blob/master/setup-pack-virtualenv.py

It would be great if st2 could ship a similar script in `/opt/stackstorm/st2/bin` or somewhere appropriate.

Before I wrote this script, I've come up with a fairly quick solution that manually activates st2 virtualenv in the build step and then run `pip install` for each pack, but apparently there are some [platform managed dependencies for all packes](https://github.com/StackStorm/st2/blob/master/st2common/st2common/constants/pack.py#L84-L88), so I thought it is better to follow more _st2-ish_ way.

Kami:
This pull request adds a new independent and standalone `st2-pack-install` CLI command for installing a pack and setting up pack virtual environment.

End result is the same as running `st2 pack install <foo>` minus the content registration step.

This command is fully standalone and only requires Python, pip, `st2common` PyPi package and git binary to be installed on the system where it's used. It doesn't need / require database (MongoDB) and message bus access (RabbitMQ).

This command is designed to be used in environments where all the resources (packs) are baked into the base VM / container image which is then deployed.

Keep in mind that the actual content still needs to be registered with StackStorm at some later point when a cluster is deployed and when MongoDB and RabbitMQ are up (`st2ctl reload --register-all` which is also an idempotent operation).

NOTE: This change required some long needed refactoring - moved a bunch of code outside of the action which makes it re-usable elsewhere and also easier to test.

Usage:

```bash
./st2common/bin/st2-pack-install libcloud
./st2common/bin/st2-pack-install doesnt-exist
./st2common/bin/st2-pack-install libcloud==9.9.9
./st2common/bin/st2-pack-install libcloud --debug
./st2common/bin/st2-pack-install libcloud --debug
./st2common/bin/st2-pack-install libcloud xml=9.9.9 csv
./st2common/bin/st2-pack-install libcloud doesntexist csv xml=9.9.9
```

Resolves #3912.


armab:
It'll make `st2-pack-install` consistent with `st2 pack install`, which already supports that kind of approach.

We want it to be as similar as possible in behavior, syntax, input with the original command, just don't interact with the DB/MQ. So from user's side original: 
`st2 pack install github ansible yolo=1.2.3 https://github.com/some/pack`
can be easily transformed to:
`st2-pack-install github ansible yolo=1.2.3 https://github.com/some/pack`
when only pack download + setup virtualenv is needed.

Kami:
I'll wait a bit and also let others chime on this one.

If anything, I would prefer to remove installing multiple packs at once support from `st2 pack install` as well (as mentioned above, I was never a fan of it) :P

Kami:
@nmaludy The big difference between those tools and StackStorm is that all those tools support and use transactions for those operations which makes error handling much easier and straight forward (either all succeed or all fail, no middle ground).

That's not the case for StackStorm "st2 pack install" command. It doesn't support or use transactions so installing each pack is done in exactly the same manner if user ran the same command each time with a different pack.

The problem is that handling and dealing with such errors is much harder and more complex in multi pack install scenario without transactions - aka how to proceed / rectify if it fails hall way through.

And not to mention that current UX for multi pack install in `st2 pack install` command is very bad anyway. We don't really handle or report scenarios where installation of a particular pack fails or similar. 

```bash
2018-07-20 13:39:06,826  WARNING - Auth API server is not available, skipping authentication.

For the "libcloud, xmlx=9.9.9, csv" packs, the following content will be registered:

rules     |  0
sensors   |  0
aliases   |  0
actions   |  32
triggers  |  0

Installation may take a while for packs with many items.

	[ succeeded ] download pack
	[ succeeded ] make a prerun
	[ succeeded ] install pack dependencies
	[ succeeded ] register pack

+----------+-------------------------------------------------+---------+------------------+
| name     | description                                     | version | author           |
+----------+-------------------------------------------------+---------+------------------+
| csv      | st2 content pack containing CSV integrations    | 0.4.2   | StackStorm, Inc. |
| libcloud | st2 content pack containing libcloud            | 0.4.3   | StackStorm, Inc. |
|          | integrations                                    |         |                  |
+----------+-------------------------------------------------+---------+------------------+
```

So if anything, I'd rather have a fully working "install single pack per invocation" mode than broken (or at the very least kinda working which is not enough and gives user false sense of success) "install multiple packs in a single command invocation" mode.

Kami:
I discussed this with @armab at length on Slack and we finally came up with a compromise. I will add support for installing multiple packs to this new command, but I won't really document it and installing single pack at once will be the preferred approach.

Either @armab and I will also file an issue for improvements in the existing `packs.install` action / `st2 pack install` CLI command so the end user experience and actual behavior on failures will be better and more consistent.

Kami:
Tests in https://github.com/StackStorm/st2tests/pull/138.

armab:
I'm testing this, but would be nice to have @warrenvw in a feedback & testing loop as well.

I'm wondering if `st2-pack-install` would be a clear enough naming. For example, @shusugmt in #3912 showcased his `setup-pack-virtualenv.py` script.
My thinking is to avoid confusion with `st2 pack install` which does "full install", opposing to `st2-pack-install` script has a bit different functionality: pack download + setup virtualenv.

@warrenvw @shusugmt WDYT?

Kami:
@armab I think the name is reasonable, and it's really hard to be 100% correct and explicit.

"setup-virtualenv" is also not a fully accurate name since setup virtualenv is just one of the two steps the script performs (1. download the pack, 2. setup pack virtualenv).

So if we wanted to be really explicit and correct, the right name would be `st2-download-pack-and-setup-virtualenv`.

But that's bordering on a bit too verbose :)

shusugmt:
Script name won't matter.

Can it support **ONLY** setup virtualenvs and skip download via option? There are cases that user will just `COPY` the pack dir from docker context in `docker build`

Kami:
@shusugmt I was wondering if we need such flexibility while working on it, but I opted for a simpler behavior.

If we do need it, we could have three scripts:

1. `st2-pack-download` - downloads the pack
2. `st2-pack-setup-virtualenv` - sets up pack virtual environment
3. `st2-pack-install` - does both steps 1) and 2)

/cc @armab

armab:
@shusugmt Good point.

From the tests: 
```
st2 pack install file:///opt/stackstorm/packs/bitcoin

	[ succeeded ] download pack
	[ succeeded ] make a prerun
	[ succeeded ] install pack dependencies
	[ succeeded ] register pack

+-------------+--------------------------+
| Property    | Value                    |
+-------------+--------------------------+
| name        | bitcoin                  |
| description | bitcoin integration pack |
| version     | 0.1.0                    |
| author      | StackStorm, Inc.         |
+-------------+--------------------------+
```
already works, it's re-using existing dir as expected with no download, see https://github.com/StackStorm/st2/issues/3771 https://github.com/StackStorm/st2/pull/4045.

As said before, ideally to have the same behavior between 2 scripts, so `st2-pack-install` should support that as well, which cover what you're saying natively.

Currently `st2 pack install file:///opt/stackstorm/packs/github` works, while `sudo /opt/stackstorm/st2/bin/st2-pack-install file:///opt/stackstorm/packs/github` fails.

shusugmt:
> As said before, ideally to have the same behavior between 2 scripts

This also seems to make `st2 packs install` command independent from `packs` pack easily afterwards.

\# In k8s environment especially, the dependency to `packs` pack requires to have some tricky initialization job like [this](https://github.com/StackStorm/st2-docker/blob/master/runtime/kubernetes-1ppc/init-st2.yml), and I'd like to take that out completely.

shusugmt:
> 1. st2-pack-download - downloads the pack
> 2. st2-pack-setup-virtualenv - sets up pack virtual environment
> 3. st2-pack-install - does both steps 1) and 2)

+1 This approach looks clean and clear.

armab:
@shusugmt Under tricky initialization do you mean registering content or another init job?

shusugmt:
@armab Both. This discussion is somewhat out of scope of the original PR, but I'll clarify. 

1. The usual st2+k8s setup seems to use RWX PV for `/opt/stackstorm/packs`. This will empty the dir so we somehow need to populate the files especially `packs` pack content since the core command `st2 pack install` depends on it.

2. Even if we use *bake-the-packs* approach and have `/opt/stackstorm/virtualenvs` pre-populated for `packs` pack, we still somehow need to execute `st2-register-content` so that all the dependent actions in `packs` pack are registered to mongodb.

armab:
In our K8s work @warrenvw figured that option `1` is no go in a real distributed clustered environment. We can solve it by adding another layer like NFS for storting/sycing pack content, but that has its drawbacks.


Talking about the `2` point and the requirement to register content, - yes, I already feel the pain in K8s work we're doing.
I believe and agree that all content that comes with the `core` stackstorm code should be registered by default on st2 init time. This way stackstorm starts in a minimally working state with no need to run anything else.

And while I'm all in favor of moving more into stackstorm startup init and make user's life easier, having additional steps of registering content for custom packs is inevitable.

---

Hopefully @Kami forgives us for offtopic.

Kami:
@armab I wasn't able to replicate the pack group issue locally.

It works fine and uses `content.pack_group` config option (which defaults to `st2packs` if not explicitly provided in the config).

```bash
(virtualenv) vagrant@ubuntu-xenial:/data/stanley$ ./st2common/bin/st2-pack-install bitcoin=0.1.0
2018-07-30 09:03:31,111 INFO [-] Installing pack "bitcoin=0.1.0"
2018-07-30 09:03:32,491 INFO [-] Successfuly installed pack "bitcoin=0.1.0"
2018-07-30 09:03:32,491 INFO [-] Setting up virtualenv for pack "bitcoin"
2018-07-30 09:03:35,689 INFO [-] Successfuly set up virtualenv for pack "bitcoin=0.1.0"
(virtualenv) vagrant@ubuntu-xenial:/data/stanley$ ls -la /opt/stackstorm/packs/
total 32
drwxr-xr-x 8 vagrant vagrant  4096 Jul 30 09:03 .
drwxr-xr-x 5 vagrant vagrant  4096 Jun 22 10:00 ..
drwxrwxr-x 4 vagrant st2packs 4096 Jul 30 09:03 bitcoin
```

Kami:
@armab Pushed a fix for `st2-pack-install bitcon=0.1.0` (ace321d28).

I can't replicate your other two issues though.

1. Group ownership stuff seems to work fine (see my comment above)

```bash
(virtualenv) vagrant@ubuntu-xenial:/data/stanley$ ./st2common/bin/st2-pack-install bitcoin=0.1.0
2018-07-30 09:03:31,111 INFO [-] Installing pack "bitcoin=0.1.0"
2018-07-30 09:03:32,491 INFO [-] Successfuly installed pack "bitcoin=0.1.0"
2018-07-30 09:03:32,491 INFO [-] Setting up virtualenv for pack "bitcoin"
2018-07-30 09:03:35,689 INFO [-] Successfully set up virtualenv for pack "bitcoin=0.1.0"
(virtualenv) vagrant@ubuntu-xenial:/data/stanley$ ls -la /opt/stackstorm/packs/
total 32
drwxr-xr-x 8 vagrant vagrant  4096 Jul 30 09:03 .
drwxr-xr-x 5 vagrant vagrant  4096 Jun 22 10:00 ..
drwxrwxr-x 4 vagrant st2packs 4096 Jul 30 09:03 bitcoin
```

2. `file://`` syntax also works fine as long as the input directory is a git repository (which is also the case for `st2 pack install`

```bash
(virtualenv) vagrant@ubuntu-xenial:/data/stanley$ ./st2common/bin/st2-pack-install file:///opt/stackstorm/packs/bitcoin
2018-07-30 09:11:19,309 INFO [-] Installing pack "file:///opt/stackstorm/packs/bitcoin"
2018-07-30 09:11:19,373 INFO [-] Successfuly installed pack "bitcoin"
2018-07-30 09:11:19,373 INFO [-] Setting up virtualenv for pack "bitcoin"
2018-07-30 09:11:21,424 INFO [-] Successful set up virtualenv for pack "bitcoin"
```

Both of those things should work out of the box since they use exactly the same code path as `st2 pack install.

Kami:
NOTE: End to end tests will keep failing until I merge this - https://github.com/StackStorm/st2tests/pull/139.

I plan to merge st2tests PR once I get this PR approved and once it's merged into master.

armab:
Try installing the packages produced from CircleCI. I guess that brings some diff.
That's what I'm getting: 

```
sudo /opt/stackstorm/st2/bin/st2-pack-install bitcoin=0.1.0
2018-07-30 14:56:38,490 INFO [-] Installing pack "bitcoin=0.1.0"
2018-07-30 14:56:40,167 INFO [-] Successfuly installed pack "bitcoin=0.1.0"
2018-07-30 14:56:40,167 INFO [-] Setting up virtualenv for pack "bitcoin=0.1.0"
  File "/opt/stackstorm/st2/bin/st2-pack-install", line 42, in <module>
    sys.exit(install_pack.main(sys.argv[1:]))
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/cmd/install_pack.py", line 81, in main
    no_download=True)
  File "/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/util/virtualenvs.py", line 70, in setup_pack_virtualenv
    raise ValueError('Invalid pack name "%s"' % (pack_name))
ValueError: Invalid pack name "bitcoin=0.1.0"
```

from latest package in this PR:
```
cat package.meta 
[server]
version = 2.9dev
git_sha = ac123a0
circle_build_url = https://circleci.com/gh/StackStorm/st2/8259
```

Kami:
@armab Sorry, my bad. I accidentally removed the fix in a subsequent commit.

Fix is not back 48a243d48c9eb2a1165ff4cadf26313c6a3659c1in.

armab:
Can confirm that pack-install script now works well :+1: 
The group ownership is still an issue.

@Kami Per https://github.com/StackStorm/st2/pull/4256#pullrequestreview-139535670 comment check the virtualenvs dir, not the packs dir: 

```
ls -la /opt/stackstorm/virtualenvs/
total 28
drwxrwxr-x  7 root st2packs 4096 Jul 30 16:06 .
drwxr-xr-x 11 root root     4096 Jul 17 16:47 ..
drwxr-xr-x  6 root root     4096 Jul 30 15:58 ansible
drwxr-xr-x  6 root root     4096 Jul 30 16:03 bitcoin
drwxr-xr-x  6 root root     4096 Jul 30 16:03 github
drwxrwxr-x  6 root st2packs 4096 Jul 30 16:06 hue
drwxrwxr-x  6 root st2packs 4096 Jul 18 17:45 st2

```

Kami:
@armab I assume that's because `st2actionrunnerservice` which runs packs.setup_virtualenv action runs under root:st2packs (and nothing to do with the actual code).

We don't explicitly change or set permissions for pack virtualenv directory, it just inherits whatever user and group it runs under.

I can make the change and add some code which sets the group based on the `pack_group` config value inside setup_virtualenv code, but I'm not a fan of because of multiple reasons:

1. It will require script to run as root / user which runs it will need to have sudo access
2. There is an edge case and possible inconsistency right now between the group `st2actionrunner` process runs under and the value of `content.pack_group` config option.

Kami:
@armab I pushed a commit (5513b7d73f137e68290aa06c3068875535b23728) which I believe is a good compromise and results in the same outcome no matter how you run the command.

Previously, virtualenv directory owner group would only be set correctly when using action runner action (`st2 pack install`), but not when using ``st2ctl reload --register-setup-virtualenvs`` and ``st2-pack-setup-virtualenv`` command.

As mentioned above, I'm not a fan of those commands requiring sudo access. That's why the "change directory owner" step doesn't fail if the directory group ownership change fails due to one reason or another (defined group doesn't exist, no sudo access, etc.).

Keep in mind that this "ignore errors on change directory owner" behavior already existed before my changes, but I assume it was not done intentionally, but now that's done intentionally / as part of the design.

So in short, this fixes inconsistencies between the commands (good catch, this issue existed for the virtualenv directory when using ``st2ctl reload --register-setup-virutalenvs`` for a long time already) and also also makes sudo optional.

armab:
In a normal environment, when the StackStorm is shipped via packages you couldn't run `st2-pack-install`, because `/opt/stackstorm/packs` dir is owned by `root:st2packs` with no write access for others.

So anyways, it was already a requirement to run `st2-pack-install` as a privileged user.

PR #4257, opened 2018-07-20
Fix make tests

m4dcoder:
The st2client.egg-info is included in the list of components and that caused errors when making requirements and running pylints on the components. The make tests command would abort at the pylint step and stop running unit tests. The wildcard function to identify the list of st2 components is replaced with a shell command to run ls and grep the entries that start with st2 but not end with .egg-info.

The itests step requires ci-prepare-integration so the currently pytests fail on itests. This patch reduce the scope of pytests to only run unit tests but expand the scope of make tests to cover specific steps from pytests, ci-unit, and ci-integration.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -396,10 +396,10 @@ tests: pytests
 pytests: compile requirements .flake8 .pylint .pytests-coverage
 
 .PHONY: .pytests
-.pytests: compile .configgen .generate-api-spec .unit-tests .itests clean
+.pytests: compile .configgen .generate-api-spec .unit-tests clean
 
 .PHONY: .pytests-coverage
-.pytests-coverage: .unit-tests-coverage-html .itests-coverage-html clean
+.pytests-coverage: .unit-tests-coverage-html clean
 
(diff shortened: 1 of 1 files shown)


Kami:
LGTM, although I'm curious where did you encounter this issue (since I never encountered it before)?

Since `.egg-info` should really only be created when building st2client PyPi package (and since those steps run inside `st2client/` directory, `egg-info` should also be created inside `st2client` directory and not in the repo root).

m4dcoder:
I started encountering this a few weeks ago. I just didn't have time to dig into it until now.  I don't recall how st2client.egg-info is created here.

Kami:
@m4dcoder Yeah, the thing is that this shouldn't happen - it looks like someone ran `python st2client/setup.py bdist` or similar command in repo root.

In theory I'm fine with this change, but I also don't see much need for it, since the same issue would arise if you created a random directory which name starts with `st2` inside the repo root (and we have similar wildcard list commands in other make files and scripts).

PR #4259, opened 2018-07-23
New core.inject_trigger action for injecting a trigger into the system

### LINKED ISSUE #4231: Feature Idea: core.inject_trigger
##### SUMMARY
I would like to inject/dispatch triggers in the middle of workflows so that rules can trigger additional workflows when those trigger points are reached. Then, custom triggers registered in a pack can serve as well known synchronization points.

##### ISSUE TYPE
 - Feature Idea

##### EXPECTED RESULTS
trigger `my_42_trigger` registered in `example` pack with the following action in a workflow to dispatch it.

```
dispatch_action_in_workflow:
  action: core.inject_trigger
  input:
    trigger: example.my_42_trigger
    parameters:
      answer_to: "{{ task('life').result.stdout }}"
      the_universe: "{{ _.and_everything }}"
```

#### EXAMPLE USE CASE
I have a data-processing workflow that gets triggered by a sensor when a new server comes online. The workflow gathers data about the server, processes it (using custom actions), and then dispatches/injects the additional triggers once each piece of data is available. This way, the sub-workflows can start as soon as the data they require is available, and the data-processing workflow can be much simpler without having to keep a centralized list of everything that should start once the data is ready.

#### IMPLEMENTATION THOUGHTS
This could be, as @nmaludy suggested in slack, an action that uses webhooks to kick off the various triggers/workflows.

If the action were in core, however, it could skip the webhook processing, and just inject the trigger. This might work something like the sensor service's dispatch method.

I looked at putting this in the `st2` pack, but the st2 client lib doesn't have the ability to kick off triggers, so it makes more sense to make this part of the `core` pack. Possible names include `core.inject_trigger` or `core.dispatch_trigger`. "dispatch" is nice because it mirrors the terminology used in sensors.

Looking at what might be useful in implementing this:
- [st2common.transport.reactor.TriggerDispatcher](https://github.com/StackStorm/st2/blob/a376219/st2common/st2common/transport/reactor.py#L75) Used by [actions API](https://github.com/StackStorm/st2/blob/6261a2a39c23b7c9e43e3dd43050b445bfd8e458/st2api/st2api/controllers/v1/actions.py) and [webhooks API](https://github.com/StackStorm/st2/blob/7c1e386f957a33cabd0142dc67440498b655503d/st2api/st2api/controllers/v1/webhooks.py)
- [st2reactor.container.utils.create_trigger_instance](https://github.com/StackStorm/st2/blob/f71f9b1/st2reactor/st2reactor/container/utils.py#L29)

Kami:
This pull request adds new ``core.inject_trigger`` action for injecting a trigger into the system.

## Background / Context

A lot of times inside the workflow, user needs to inject a trigger into the system to trigger another workflow / action execution or similar.

Right now it's possible to do that, but it's quite cumbersome because webhooks API endpoint needs to be used (see https://forum.stackstorm.com/t/is-there-a-way-to-trigger-event-in-workflow-like-inquiry-and-consume-it-in-rules/199/8 for example and details).

This pull request makes injecting a trigger as simple as running a new action inside the workflow.

It's also worth noting (as mentioned in my comment on the forum) that in some scenarios existing `core.st2.generic.actiontrigger` can be used and there is no need for injecting a custom trigger into the system (albeit explicitly injecting a trigger inside the workflow is more transparent and explicit).

## Implementation Details

Right now the prototype action in the PR directly connects to the message bus and the database.

This is not a problem right now because Python runner actions already need access to the database and message bus for other reasons, but it could be a problem in the future if we allow to run Python runner actions on servers which don't have access to the message bus and the database (well, that's actually already the case for serverless work).

To make this feature more robust and future proof, I will update the implementation to utilize "webhooks" API endpoint. This way it will be consistent with the datastore service which utilizes the API for managing datastore items.

## Note on the Implementation Details

The way trigger instance stuff works right now means we only create `TriggerInstanceDB` object inside the rules engine.

This means that the webhook API is async and there is no way to know if the actual TriggerInstanceDB object was created (the creation could fail due to the user providing an invalid trigger reference, invalid payload which doesn't validate against the schema, etc.).

That's a feature / limitation we discussed in the past already, but we don't have a good solution for it yet.

Resolves #4231.


Kami:
Pushed changes for the action to utilize API client - b0d2046e4ef4c42c00dcceac1252561a4f6b8a56.

Now it's consistent with the datastore service and means the action doesn't depend on the message bus and database access anymore.

Kami:
As a "bonus" and to provide a better user-experience in lieu of the asynchronous nature of the webhook API, I pushed a change which updates webhooks code to utilize the same trigger dispatching service / code which is utilized by the sensors.

This way we follow the same code path and can optionally also perform trigger payload validation and throw if a trigger doesn't exist in the db or if the trigger payload validation fails.

This provides a much better user-experience in the error scenarios - right now (as mentioned above), if trigger referenced in a webhook doesn't exist in the DB, this will only be visible in the `st2rulesengine` service log file and rule enforcement object and nowhere else (webhooks API will asynchronously return success and `TriggerInstanceDB` object will never be created).

In addition to that, I propose setting default value for ``system.validate_trigger_payload`` config option to ``True``. This means we will perform trigger payload validation by default for triggers which contain a schema.

I will of course document this behavior in the upgrade notes and also document how to disable this option if wanted / needed (which is really a bad idea, but OK).

Kami:
As discussed on slack with @LindsayHill, I enabled trigger parameters and payload schema validation for custom triggers (previously we only done it for system triggers).

We added that functionality in v2.2 so it's long overdue to finally enable it by default. I will document the change in behavior in the upgrade notes.

While working on that, I found that existing webhook tests have a lot of issues, the biggest two beginning:

1. Tests use triggers which don't conform to the payload schema
2. Tests mock too many methods and also invalid mocks are used, but this was never detected since most things were mocked and we weren't testing code paths we should have been testing

I fixed both of those issues.

Kami:
Upgrade notes change at - https://github.com/StackStorm/st2docs/pull/768.

If you are still wondering why the previous approach wasn't great, see examples in the st2docs PR (as mentioned above, previously it was very hard to troubleshoot such issues due to the async nature of the webhooks API endpoint and due to not performing any validation on webhook ingress).

st2api/st2api/controllers/v1/webhooks.py
```diff
@@ -154,10 +158,15 @@ def post(self, hook, webhook_body_api, headers, requester_user):
...
             payload['body'] = body
+
             # Dispatch trigger instance for each of the trigger found
-            for trigger in triggers:
-                self._trigger_dispatcher.dispatch(trigger, payload=payload,
-                    trace_context=trace_context)
+            for trigger_dict in triggers:
+                # TODO: Instead of dispatching the whole dict we should just
+                # dispatch TriggerDB.ref or similar
+                self._trigger_dispatcher_service.dispatch_with_context(trigger=trigger_dict,
```
m4dcoder: nit pick: move `trigger=trigger_dict` to a newline so it's consistent with styling on the other parameters below.

conf/st2.conf.sample
```diff
@@ -301,10 +301,10 @@ facility = local7
 [system]
 # Enable debug mode.
 debug = False
-# True to validate parameters for non-system trigger types when creatinga rule. By default, only parameters for system triggers are validated
-validate_trigger_parameters = False
+# True to validate parameters for non-system trigger types when creatinga rule. By default, only parameters for system triggers are validated.
+validate_trigger_parameters = True
 # True to validate payload for non-system trigger types when dispatching a trigger inside the sensor. By default, only payload for system triggers is validated.
-validate_trigger_payload = False
+validate_trigger_payload = True
```
m4dcoder: There's not much description on making the validation default. Is it correct to assume this is because we want to validate the parameters and payloads when a trigger is injected manually?

st2api/st2api/controllers/v1/webhooks.py
Kami: Yeah, I actually hate the dict stuff - didn't even know we support it here (since other dispatch function doesn't really document it can also be a dict) - I assumed it always need to be either a trigger_type_ref or trigger_ref.

Sadly that's not the case, but something we should change and improve in the future.

conf/st2.conf.sample
Kami: Correct - parameters when creating a trigger (so when creating a rule) and payload when a trigger is dispatched (either via webhook or sensor).

We've already done this in the past, but just for system triggers.

Documentation upgrade notes change is available at https://github.com/StackStorm/st2docs/pull/768.

Kami:
@m4dcoder Added unit tests in 3c5c2334717e6d66c98b297adc8d31451e5f8618.

Kami:
End to end tests in https://github.com/StackStorm/st2tests/pull/140.

contrib/core/tests/test_action_inject_trigger.py
```diff
@@ -0,0 +1,66 @@
...
+        action.run(trigger='dummy_pack.trigger1')
+        mock_api_client.webhooks.post_generic_webhook.assert_called_with(
+            trigger='dummy_pack.trigger1',
+            payload={},
+            trace_tag=None
+        )
+
+        mock_api_client.webhooks.post_generic_webhook.reset()
+
+        # 2. With payload
```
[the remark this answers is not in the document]
Kami: Agreed, I usually also do that, will make that change.

PR #4263, opened 2018-07-24
[WIP] Try adding back MongoDB changes to see if it helps with the tests performance

Kami:
Adding back MongoDB performance optimization changes which were (accidentally?) reverted quite a long time ago to see if it's they help with the tests performance and if it's still worth having those changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,8 +61,8 @@ install:
   - if [ "${TASK}" = 'ci-unit' ] || [ "${TASK}" = 'ci-integration' ]; then pip install codecov; fi
   - if [ "${TASK}" = 'ci-unit' ] || [ "${TASK}" = 'ci-integration' ] || [ "${TASK}" = 'compilepy3 ci-py3-unit' ] || [ "${TASK}" = 'ci-py3-integration' ]; then sudo .circle/add-itest-user.sh; fi
   # Note: We use ramdisk for better performance
-  - if [ ${TASK} = 'ci-unit' ] || [ ${TASK} = 'ci-integration' ] || [ ${TASK} = 'ci-py3-integration' ]; then sudo -E ./scripts/travis/create-and-mount-ramdisk.sh; fi
-  - if [ ${TASK} = 'ci-unit' ] || [ ${TASK} = 'ci-integration' ] || [ ${TASK} = 'ci-py3-integration' ]; then ./scripts/travis/install-and-run-mongodb.sh; fi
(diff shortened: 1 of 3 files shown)


Kami:
Even if it turns out the performance benefits are not large enough, we still want the `--notablescan` change which was inadvertently removed / disabled.

That flag is very handy because it makes tests fail if some MongoDB query doesn't use an index and results in a full table scan.

It also turns out that while this feature was disabled, we "regressed" / added some new features which don't correctly utilize indexes so tests are failing with that flag enabled.

I will look into fixing the code once we get baseline performance numbers.

.travis.yml
```diff
@@ -12,6 +12,7 @@ branches:
 env:
   global:
     - CACHE_NAME=JOB1
+    - MONGODB_VERSION=3.4.16
```
armab: Talking about version pinning, I believe what we have now `/mongodb-org/3.4`  in repo with `major.minor` would be overall a more preferable approach.
I think that in `mongodb` context pinning patch version instead of giving a better control would, in fact, work against us.

.travis.yml
```diff
@@ -60,10 +60,14 @@ install:
   - if [ "${TASK}" = 'compilepy3 ci-py3-unit' ] || [ "${TASK}" = 'ci-py3-integration' ]; then pip install "tox==3.0.0"; else make requirements; fi
   - if [ "${TASK}" = 'ci-unit' ] || [ "${TASK}" = 'ci-integration' ]; then pip install codecov; fi
   - if [ "${TASK}" = 'ci-unit' ] || [ "${TASK}" = 'ci-integration' ] || [ "${TASK}" = 'compilepy3 ci-py3-unit' ] || [ "${TASK}" = 'ci-py3-integration' ]; then sudo .circle/add-itest-user.sh; fi
+  # Note: We use ramdisk for better performance
+  - sudo -E ./scripts/travis/create-and-mount-ramdisk.sh
+  - ./scripts/travis/install-and-run-mongodb.sh
 
 # Let's enable rabbitmqadmin
 # See https://github.com/messagebus/lapine/wiki/Testing-on-Travis.
 before_script:
```
armab: For reference of editing existing mongo conf, this looks like a good hint: 
http://georgeshank.com/how-to-enable-a-mongodb-replica-set-on-travis/

```
- echo "foo = bar" | sudo tee -a /etc/mongodb.conf
- sudo service mongodb restart
```

which overall is more clean.

Kami:
Closing.

Will open a new PR which sets those options in the config file as discussed above with @armab.

PR #4270, opened 2018-07-25
Submit coverage reports to codecov.io on master merge

Kami:
This pull request fixes a regression introduced in https://github.com/StackStorm/st2/pull/4147 and updates the code so we again submit coverage reports to coveralls.io on merge to master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,7 @@ before_install:
 install:
   - if [ "${TASK}" = 'compilepy3 ci-py3-unit' ] || [ "${TASK}" = 'ci-py3-integration' ]; then pip install "tox==3.0.0"; else make requirements; fi
   # NOTE: We need eventlet installed so coverage can be correctly combined. This is needed because we are covering code which utilizes eventlet.
-  # Without eventlet being available to to coverage command it will fail with "Couldn't trace with concurrency=eventlet, the module isn't installed."
+  # Without eventlet being available to the coverage command it will fail with "Couldn't trace with concurrency=eventlet, the module isn't installed."
   - if [ "${TASK}" = 'ci-unit' ] || [ "${TASK}" = 'ci-integration' ] && [ "${ENABLE_COVERAGE}" = 'yes' ]; then pip install eventlet ; pip install -e "git+https://github.com/Kami/codecov-python.git@better_error_output#egg=codecov"; fi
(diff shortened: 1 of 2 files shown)


Kami:
Alright, coveralls coverage submission is working again. Fix is in 7c40ecd5a69282feb18ee959fe2211eea6c9e166.

The issue is related to our code utilizing eventlet which makes coverage use `eventlet` concurrency coverage mechanism. This means that eventlet library is also needs to be available for `coverage combine` step for it to work.

That means we also need to install `eventlet` in addition to `codecov` when coverage is enabled. Keep in mind that those libraries are installed outside of tests virtual environment and that's why we need to install them again.

Related PR which allowed me to discover the root cause - https://github.com/codecov/codecov-python/pull/153 (codecov has a bug which simply swallows any command output which makes such issues hard / impossible to troubleshoot).

Kami:
We are back in business - https://codecov.io/github/StackStorm/st2?branch=master.

Full coverage will be available once this branch is merged into master. I only enabled partial coverage for this branch / PR so I could test my fix is working.

I think the issue with combing unit + integration test coverage might still exist, but to get around that, it would require a lot more work and not something I will do in this PR.

We would need to run unit and integration tests in the same target. Right now we run it twice - one for unit and once for integration tests and we submit those reports separately to codecov and I believe codecov doesn't automatically combine them because they are part of different builds.

.travis.yml
```diff
@@ -64,14 +66,16 @@ before_install:
   - sudo pip install --upgrade "virtualenv==15.1.0"
 
 install:
-  - "echo ${ENABLE_COVERAGE}"
   - if [ "${TASK}" = 'compilepy3 ci-py3-unit' ] || [ "${TASK}" = 'ci-py3-integration' ]; then pip install "tox==3.0.0"; else make requirements; fi
-  - if [ "${TASK}" = 'ci-unit' ] || [ "${TASK}" = 'ci-integration' ] && [ "${ENABLE_COVERAGE}" = 'yes' ]; then pip install codecov; fi
+  # NOTE: We need eventlet installed so coverage can be correctly combined. This is needed because we are covering code which utilizes eventlet.
+  # Without eventlet being available to the coverage command it will fail with "Couldn't trace with concurrency=eventlet, the module isn't installed."
+  - if [ "${TASK}" = 'ci-unit' ] || [ "${TASK}" = 'ci-integration' ] && [ "${ENABLE_COVERAGE}" = 'yes' ]; then pip install eventlet ; pip install -e "git+https://github.com/Kami/codecov-python.git@better_error_output#egg=codecov"; fi
```
armab: `https://github.com/Kami/` :roll_eyes:

PR #4271, opened 2018-07-25
Remove use of async in args for python 3.7 support

j3p0uk:
In 3.7 'async' becomes a reserved keyword.  Rename the variable.

https://docs.python.org/3/whatsnew/3.7.html

### DIFF
--- a/st2client/st2client/commands/action.py
+++ b/st2client/st2client/commands/action.py
@@ -260,7 +260,7 @@ class ActionRunCommandMixin(object):
             return
 
         execution = self.run(args, **kwargs)
-        if args.async:
+        if args.action_async:
             self.print_output('To get the results, execute:\n st2 execution get %s' %
                               (execution.id), six.text_type)
             self.print_output('\nTo view output in real-time, execute:\n st2 execution '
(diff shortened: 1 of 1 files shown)


Kami:
Thanks for the contribution.

Our Python 3 tests run under Python 3.6 and that's why I assume they didn't catch it. In addition to that, we don't officially support Python 3 yet and once we will, we will only support Python 3.5 - 3.6 to begin with.

Having said that, this change doesn't break any other thing and it's a good future proofing so it's still good to merge it :+1:

j3p0uk:
Given the client libraries are incorporated in the codebase, could that restrict the integration of 3rd party applications with Stackstorm, where they only want to make API calls against ST2 using the client library?

Does that suggest a separate package just for the client that could support more versions?

Kami:
@j3p0uk Yeah that was actually our long term plan (decouple Python client library and the actual CLI tool, but we sadly didn't get to it yet).

Kami:
In the future, I would also want us to run tests under Python 3.7 in addition to Python 3.6, but that's sadly not feasible right now.

Our build is already quite slow and adding 2 new tasks for Python 3.7 would slow everything down for 20 minutes or so and that's sadly not acceptable.

j3p0uk:
I know that feeling well!  Travis CI doesn't seem to support python 3.7 at the moment either, I guess I'm just being too eager! :)

PR #4272, opened 2018-07-25
More coverage improvements

Kami:
Some small coverage / build related improvements:

1. Move my fork with bugfix for codecov-python library to StackStorm organization
2. Run "make play" during build. This prints out various environment information which helps with troubleshooting various build related issues.
3. Add back various MongoDB performance optimizations and re-enable "notablescan" option which will warn us if some query is not using an index.
4. Fix issues with missing indexes and queries which resulted in full table scans instead of using the index

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,7 +69,7 @@ install:
   - if [ "${TASK}" = 'compilepy3 ci-py3-unit' ] || [ "${TASK}" = 'ci-py3-integration' ]; then pip install "tox==3.0.0"; else make requirements; fi
   # NOTE: We need eventlet installed so coverage can be correctly combined. This is needed because we are covering code which utilizes eventlet.
   # Without eventlet being available to the coverage command it will fail with "Couldn't trace with concurrency=eventlet, the module isn't installed."
-  - if [ "${TASK}" = 'ci-unit' ] || [ "${TASK}" = 'ci-integration' ] && [ "${ENABLE_COVERAGE}" = 'yes' ]; then pip install eventlet ; pip install -e "git+https://github.com/Kami/codecov-python.git@better_error_output#egg=codecov"; fi
+  - if [ "${TASK}" = 'ci-unit' ] || [ "${TASK}" = 'ci-integration' ] && [ "${ENABLE_COVERAGE}" = 'yes' ]; then pip install eventlet ; pip install -e "git+https://github.com/StackStorm/codecov-python.git@better_error_output#egg=codecov"; fi
   - if [ "${TASK}" = 'ci-unit' ] || [ "${TASK}" = 'ci-integration' ] || [ "${TASK}" = 'compilepy3 ci-py3-unit' ] || [ "${TASK}" = 'ci-py3-integration' ]; then sudo .circle/add-itest-user.sh; fi
 
 # Let's enable rabbitmqadmin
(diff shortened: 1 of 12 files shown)


Kami:
As mentioned in the comment above, this pull request also includes "notablescan" MongoDB option which was accidentally / inadvertently removed a while back.

This will now again cause MongoDB queries which don't use an index and result in a full table scan to fail (and as such, also some tests to fail).

Since this feature has been disabled for a while now there were some regressions (missing indexes) which I fixed in 75fe2c1d2b1b09a150f83b5bea4f1e2d3c3a7653, c18911fe647745a20b3541c3cee4a0561b0fcc73, 17ad1d919b32379e56472e5162cb6816df9942c9 etc.

.travis.yml
```diff
@@ -67,22 +67,24 @@ before_install:
 
 install:
   - if [ "${TASK}" = 'compilepy3 ci-py3-unit' ] || [ "${TASK}" = 'ci-py3-integration' ]; then pip install "tox==3.0.0"; else make requirements; fi
-  # NOTE: We need eventlet installed so coverage can be correctly combined. This is needed because we are covering code which utilizes eventlet.
-  # Without eventlet being available to the coverage command it will fail with "Couldn't trace with concurrency=eventlet, the module isn't installed."
-  - if [ "${TASK}" = 'ci-unit' ] || [ "${TASK}" = 'ci-integration' ] && [ "${ENABLE_COVERAGE}" = 'yes' ]; then pip install eventlet ; pip install -e "git+https://github.com/Kami/codecov-python.git@better_error_output#egg=codecov"; fi
+  - if [ "${TASK}" = 'ci-unit' ] || [ "${TASK}" = 'ci-integration' ] && [ "${ENABLE_COVERAGE}" = 'yes' ]; then scripts/travis/install-coverage-dependencies.sh; fi
```
armab: Instead of that, you can have one wrapper script `scripts/codecov.sh`  which will run in `after_success` section and include codecov installation + codecov coverage reporting execution.

Will make it even cleaner.

Kami: Good point, I agree, no need to install codecov here.

Will make that change :)

.travis.yml
```diff
@@ -67,25 +67,31 @@ before_install:
...
+  # Print out various environment variables info
+  - make play
 
 script:
   - make ${TASK}
+  # NOTE: We only generate and submit coverage report for master and version branches
+  # NOTE: We put this here and not after_success so build is marked as failed if this step fails
+  # See https://docs.travis-ci.com/user/customizing-the-build/#breaking-the-build
+  # https://github.com/travis-ci/travis-ci/issues/758#issuecomment-266756853
+  - if [ ${TASK} = 'ci-unit' ] || [ ${TASK} = 'ci-integration' ] && [ "${ENABLE_COVERAGE}" = 'yes' ]; then ./scripts/travis/scripts/travis/submit-codecov-coverage.sh; fi
```
armab: BTW `/scripts/travis/scripts/travis/...` path is wrong here

Kami:
Travis offers a maximum of 5 concurrent builds for open-source projects and we had 6. This meant we needed to wait on one build to finish when job / build 6 can start.

I tried combining unit tests builds (e457751) and it looks like it now finishes faster by 5 minutes or so.

Before: https://travis-ci.org/StackStorm/st2/builds/408158415 - 1hr 5 min
After: https://travis-ci.org/StackStorm/st2/builds/408207945 - 1 hr

So it looks like having two separate builds and splitting unit tests into two chunks offers no benefits in Travis. Overhead of running "make requirements", etc. is too large and it in facts slow things down by 5 minutes or so.

PR #4274, opened 2018-07-25
Fix typo in travis.yml

Kami:
Fixes typo in the script path in travis config file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,6 +47,12 @@ PYTHON_TARGET := 2.7
 REQUIREMENTS := test-requirements.txt requirements.txt
 PIP_OPTIONS := $(ST2_PIP_OPTIONS)
 
+PYLINT_CONCURRENCY := $(PYLINT_CONCURRENCY)
+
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -47,6 +47,12 @@ PYTHON_TARGET := 2.7
 REQUIREMENTS := test-requirements.txt requirements.txt
 PIP_OPTIONS := $(ST2_PIP_OPTIONS)
 
+PYLINT_CONCURRENCY := $(PYLINT_CONCURRENCY)
```
blag: Why this line? I thought environment variables were considered normal variables within makefiles?

Kami: You are correct, I believe it's not needed, but I've done it because we also do that a couple of lines above.

Having said that, I can remove it.

Kami:
Confirmed the change is working on merge to master - https://codecov.io/gh/StackStorm/st2/commit/f92ef089fef5e7c4d770d68baf04ba4c5feaa036

PR #4275, opened 2018-07-26
More Travis CI changes

Kami:
This pull request includes more Travis changes:

1. Disable ``--with-timer`` flag on Travis (it's still used by default locally). This flag adds overhead of 2 or so minutes to each test job.
2. Various random test fixes and speedups (some of the fixed tests which were fixed never worked since they were missing a negative assert so we never failed when we should have).


Kami:
Those changes speed up the Travis build for another 4-5 minutes.

Before https://travis-ci.org/StackStorm/st2/builds/408267503, after https://travis-ci.org/StackStorm/st2/builds/408470518.

Kami:
Dug in a bit more and fixed the root cause ON_TRAVIS workaround is not needed anymore - ed3f88b, 
8955d6b.

PR #4277, opened 2018-07-27
Fail Travis build if a build job (make task) takes longer than the specified amount of seconds

Kami:
This pull request adds a script which exits with non-zero and fails Travis build if a build job (particular make task) takes longer than the specified amount of seconds.

This will hopefully allow us to catch some basic performance regressions in the code and tests which have previously sneaked into the master way more often than we would want them to.

I used current build times + 30-60 seconds as reasonable thresholds to begin with. It's possible that sometimes the build might intermediately fail due to to the unpredictable behavior in some (integration) tests (relying on the network, etc.)

In such scenarios, user should first try to re-run the build before fixing / looking into the tests.

Thresholds should only be increased if there is a very good and valid reason for it - e.g. a lot of new tests were added or similar.

Keep in mind that those thresholds apply just to the actual make task (e.g. `ci-unit`, etc.) and not to the whole build which includes other steps such as `make requirements`. Other steps timing can be very variable (e.g. `make requirements` take much longer if cache is purged or if a new required is added) so including that in the thresholds wouldn't be reliable and it would cause too many false positives.


.travis.yml
```diff
@@ -27,25 +27,23 @@ matrix:
...
-    - env: TASK=ci-unit
+    # NOTE: Thresholds only apply to the make target and not to the whole build
+    # job which also includes "make requirements" and other steps
+    # "make requirements" can take substantially lower if the cache is purged
+    # and this would cause too many intermediate failures / false positives
+    - env: TASK=ci-unit CACHE_NAME=py2 COMMAND_THRESHOLD=670
       python: 2.7
       name: "Unit Tests (Python 2.7)"
-    - env: TASK=ci-integration
+    - env: TASK=ci-integration CACHE_NAME=py2 COMMAND_THRESHOLD=670
```
m4dcoder: How did we arrive with the value of 670, 270, 310, etc?

Kami: I took at look at the current build time averages for each task and added 30-60 seconds (so we have some grace because build times can very quite a bit depending on how the infrastructure is utilized).

PR #4279, opened 2018-07-31
Fix registered paths

enykeev:


Kami:
LGTM, probably also good to add a test case :)

Kami:
EDIT: Looks like I already added back in November - ef65d43301e500885da9d4ea95d3fbf3b9439398.

So the only test we are missing is the one which checks that invalid paths are not registered, but that's probably a bit hard to add.

PR #4281, opened 2018-07-31
Add new ?output_type and ?existing_only query param filter to /v1/api/executions/<id>/output API endpoint

Kami:
This pull request adds new ``?output_type=[raw|event_source]`` and ``?existing_only=[true|false]`` query parameter filters to the ``/v1/api/executions/<id>/output`` API endpoint.

With this query parameters, user can control two things:

1. What type of output to receive

* ``raw`` for raw output data produced by the execution (default). This works well when user just wants to curl this endpoint, embed the output directly in a webpage or similar.
* ``event_source`` - for event source compatible format (new line delimited JSON). This comes handy when user is also interested in metadata such as timestamp and output type.

NOTE: ``event_source`` format can result in a slow response for actions which produced a lot of data because it means we need to format and JSON serialize every ActionExecutionOutputDB object. Sadly there is no easy way around it.

2. If they want the API endpoint to block or not.

It defaults to ``?existing_only=false`` which means the API endpoint will block until the execution finishes and write any new data when it arrives to the open connection.

Keep in mind that this only applies to executions which are still in running state.

This functionality was added to allow for a better integration with WebUI. Before this change, there was no way to retrieve output which execution produced so far if user wasn't listening to the stream API endpoint from the beginning when the execution started.

## TODO

- [x] Tests
- [ ] Update docs
- [ ] Add upgrade notes entry

### DIFF
--- a/st2api/st2api/controllers/v1/actionexecutions.py
+++ b/st2api/st2api/controllers/v1/actionexecutions.py
@@ -295,11 +295,6 @@ class ActionExecutionOutputController(ActionExecutionsControllerMixin, ResourceC
     }
     exclude_fields = []
 
-    CLOSE_STREAM_LIVEACTION_STATES = action_constants.LIVEACTION_COMPLETED_STATES + [
-        action_constants.LIVEACTION_STATUS_PAUSING,
-        action_constants.LIVEACTION_STATUS_RESUMING
-    ]
-
     def get_one(self, id, output_type='all', output_format='raw', existing_only=False,
                 requester_user=None):
         # Special case for id == "last"
--- a/st2api/st2api/controllers/v1/actionexecutions.py
+++ b/st2api/st2api/controllers/v1/actionexecutions.py
@@ -17,7 +17,6 @@ import copy
 import re
 import sys
 import traceback
-import itertools
 
 import six
 import jsonschema
@@ -39,7 +38,6 @@ from st2common.models.api.action import LiveActionAPI
 from st2common.models.api.action import LiveActionCreateAPI
 from st2common.models.api.base import cast_argument_value
 from st2common.models.api.execution import ActionExecutionAPI
-from st2common.models.api.execution import ActionExecutionOutputAPI
 from st2common.models.db.auth import UserDB
 from st2common.persistence.liveaction import LiveAction
 from st2common.persistence.execution import ActionExecution
@@ -54,12 +52,10 @@ from st2common.util import isotime
 from st2common.util import action_db as action_utils
 from st2common.util import param as param_utils
 from st2common.util.jsonify import try_loads
-from st2common.util.jsonify import json_encode
 from st2common.rbac.types import PermissionType
 from st2common.rbac import utils as rbac_utils
 from st2common.rbac.utils import assert_user_has_resource_db_permission
 from st2common.rbac.utils import assert_user_is_admin_if_user_query_param_is_provided
-from st2common.stream.listener import get_listener
 
 __all__ = [
     'ActionExecutionsController'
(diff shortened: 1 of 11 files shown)


Kami:
@enykeev Here you go - this should solve both of the issues / deficiencies you were heaving.

Having said that, I'm not a big fan of the `event_source` format - it will be very slow for executions which produce(d) a lot of data (we need to call `json_encode` on each output line...).

Sadly there is no easy way around that if we follow "event_source" specification.

That's why I suggest approach I already suggested on Slack - only support `raw` format which means we don't need to do any serialization and include timestamp of the last output object in the response headers or similar.

The reason why we need to include timestamp or similar is to allow client to avoid a race initially when retrieving running execution output - first client needs to request the existing output and then listen to any new output on the event stream API endpoint. Without having an access to a timestamp or similar this means the same output line could either be printed twice or zero time depending on the race time window.

Kami:
To add some hard data to my comment above.

Here is an example timing info for execution which produces 3200 short lines:

1. raw - `real	0m0.845s`
2. event_source - `real	0m2.527s`

So the overhead isn't all that large for relatively small executions, but it could probably get problematic with execution which produce more longer lines.

enykeev:
> Having said that, I'm not a big fan of the event_source format - it will be very slow for executions which produce(d) a lot of data (we need to call json_encode on each output line...).

- We can aggregate existing data into a single message.
- We can continue output the data as plain strings that are just wrapped with eventsource package and avoid json formatting altogether.
```
event: eventname
data: <plainstring>
[\n]
```

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -369,7 +389,15 @@ def format(gen):
             return gen
 
         def make_response():
-            app_iter = itertools.chain(existing_output_iter(), new_output_iter())
+            if existing_only:
+                # Only return existing output produced by now and return immediately. Don't wait
+                # for execution to finish
+                print('laaaaa')
+                print(existing_only)
```
enykeev: 🙄

st2common/st2common/openapi.yaml
```diff
@@ -1160,10 +1160,26 @@ paths:
           required: true
         - name: output_type
           in: query
-          description: Type of output to retrieve. If not provided, all output type is returned.
-          type: array
-          items:
-            type: string
+          description: Type of output to retrieve (stdout, stderr). If not provided, all output type is returned.
+          type: enum
```
enykeev: there's no `type: enum` in the spec. It's `type: string` with `enum` as property.

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -321,14 +323,30 @@ def get_one(self, id, output_type=None, requester_user=None):
...
+                return output_db_or_api.data
+            elif output_format == 'event_source':
+                if not isinstance(output_db_or_api, ActionExecutionOutputAPI):
+                    data = ActionExecutionOutputAPI.from_model(output_db_or_api)
+                else:
+                    data = output_db_or_api
+
+                event = 'st2.execution.output__create'
+                result = 'event: %s\ndata: %s\n\n' % (event, json_encode(data, indent=None))
+                return result
```
enykeev: We can add a third option that does not serialize json. I believe that you can also send timestamp as `id` property of eventsource package.

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -369,7 +389,15 @@ def format(gen):
...
+            if existing_only:
+                # Only return existing output produced by now and return immediately. Don't wait
+                # for execution to finish
+                print('laaaaa')
+                print(existing_only)
+                app_iter = existing_output_iter()
+            else:
+                app_iter = itertools.chain(existing_output_iter(), new_output_iter())
+
             res = Response(content_type='text/plain', app_iter=app_iter)
```
enykeev: Eventsource also needs certain headers: `Cache-Control: no-cache` and `Content-Type: text/event-stream`

enykeev: optionally, we can also look for `Last-Event-ID` request header and only output results after a certain timestamp.

st2common/st2common/openapi.yaml
Kami: Thanks, will fix that. I would assume it would error our on invalid type.

st2api/st2api/controllers/v1/actionexecutions.py
Kami: As discussed yesterday, we can just leave it like that for now and if turns out that it's too slow, I can change it.

What do you think?

armab:
> This also fixes separation of concern and moves streaming behavior from st2api to st2stream service. This way we are not mixing short running and long running connections in a single service (st2api).

That sounds really good :100:

enykeev:
The problem with this approach I myself did not anticipate is the fact that according to [the spec](https://www.w3.org/TR/2009/WD-eventsource-20090421/#processing-model), when event source endpoint "completes loading (i.e. the entire HTTP response body is received or the connection itself closes), the user agent must reset the connection." In other words, as soon as all the data from `/stream/executions/{id}/output` is loaded and server closes the connection, the client is required to reestablish the connection anew and in our case will receive entire result again.

There's multiple ways we can approach that one, but it seems to me that the best one would be to just emit another event at the end that would tell the stream is over successfully and no more data is ever coming via this channel again. Something like ```
event: st2.execution.output__close
data: ''
```

File: st2stream/st2stream/controllers/v1/executions.py [hunk 5]
```diff
@@ -0,0 +1,141 @@
...
+
+        def format_output_object(output_db_or_api):
+            if isinstance(output_db_or_api, ActionExecutionOutputDB):
+                data = ActionExecutionOutputAPI.from_model(output_db_or_api)
+            elif isinstance(output_db_or_api, ActionExecutionOutputAPI):
+                data = output_db_or_api
+            else:
+                raise ValueError('Unsupported format: %s' % (type(output_db_or_api)))
+
+            event = 'st2.execution.output__create'
```
enykeev: btw, can we rename it to something more appropriate? The verb `create` does not really represent what's happening here, IMHO.

Kami: I used this name for consistency since it's the same name which is used for those events in the `/v1/stream` endpoint.

Kami:
While working on this and testing it under Python 3 I encountered an issue with message bus message deserialization failing under Python 3 if message contained non-ascii data with the following error:

```bash
2018-08-02 09:45:42,044 ERROR [-] Can't decode message body: DecodeError(UnicodeDecodeError('ascii', b'[b\xd2\xb4\x96-tuu\xb6\x87\xbf', 2, 3, 'ordinal not in range(128)'),) (type:'application/x-python-serialize' encoding:'binary' raw:'b\'\\x80\\x02cst2common.models.db.execution\\nActionExecutionDB\\nq\\x01)\\x81q\\x02}q\\x03(U\\x0f_changed_fieldsq\\x04]q\\x05U\\x0f_fields_orderedq\\x06(U\\x02idq\\x07U\\x07triggerq\\x08U\\x0ctrigger_typeq\\tU\\x10trigger_instanceq\\nU\\x04ruleq\\x0bU\\x06actionq\\x0cU\\x06runnerq\\rU\\nliveactionq\\x0eU\\x12workflow_executionq\\x0fU\\x0etask_executionq\\x10U\\x06statusq\\x11U\\x0fstart_timestampq\\x12U\\rend_timestampq\\x13U\\nparametersq\\x14U\\x06resultq\\x15U\\x07contextq\\x16U\\x06parentq\\x17U\\x08childrenq\\x18U\\x03logq\\x19U\\x07web_urlq\\x1atq\\x1bU\\x0c_initialisedq\\x1c\\x88U\\x0f_dynamic_fieldsq\\x1dcbson.son\\nSON\\nq\\x1e)\\x81q\\x1f}q U\\n_SON__keysq!]sbU\\x08_createdq"\\x89U\\x05_dataq#h\\x1e)\\x81q$(U\\x03_idq%cbson.objectid\\nObjectId\\nq&)\\x81q\\\'U\\x0c[b\\xd2\\xb4\\x96-tuu\\xb6\\x87\\xbfq(bh\\x08}q)h\\t}q*h\\n}q+h\\x0b}q,h\\x0c}q-(X\\x0b\\x00\\x00\\x00runner_typeq.X\\x0f\\x00\\x00\\x00local-shell-cmdq/X\\x0b\\x00\\x00\\x00descriptionq0XA\\x00\\x00\\x00Action that executes an arbitrary Linux command on the localhost.q1X\\n\\x00\\x00\\x00parametersq2}q3(X\\x03\\x00\\x00\\x00cmdq4}q5(X\\x08\\x00\\x00\\x00requiredq6\\x88X\\x04\\x00\\x00\\x00typeq7X\\x06\\x00\\x00\\x00stringq8X\\x0b\\x00\\x00\\x00descriptionq9X9\\x00\\x00\\x00Arbitrary Linux command to be executed on the
```

The issue is related to `str_to_bytes` function in kombu (https://github.com/celery/kombu/blob/3.0/kombu/utils/encoding.py#L47) not passing encoding argument to `encode()` method which means it defaults to ``ascii`` and fails.

I've pushed a workaround for that in 35823e1c65d4133e31393dd2663bb85e8b89b656 (we now explicitly pass `utf-8` encoding to `encode()`).

Moving from pickle to JSON is on our road map and I already started to work on it in the past, but it's sadly a lot more involved because our code relies on actual object instances being passed around :/

Kami:
Docs changes in https://github.com/StackStorm/st2docs/pull/779.

PR #4290, opened 2018-08-03
Relocate orquesta logic from notifier to the workflow engine

m4dcoder:
Implement queue consumer that can handle variable message types. Relocate handling of action execution completion to workflow engine. Add policy application to the workflow engine for workflow tasks.

### DIFF
--- a/st2actions/st2actions/notifier/notifier.py
+++ b/st2actions/st2actions/notifier/notifier.py
@@ -29,10 +29,9 @@ from st2common.models.api.trace import TraceContext
 from st2common.models.db.execution import ActionExecutionDB
 from st2common.persistence.action import Action
 from st2common.persistence.liveaction import LiveAction
-from st2common.persistence.policy import Policy
-from st2common import policies
 from st2common.models.system.common import ResourceReference
 from st2common.persistence.execution import ActionExecution
+from st2common.services import policies as policy_service
 from st2common.services import trace as trace_service
 from st2common.transport import consumers
 from st2common.transport import utils as transport_utils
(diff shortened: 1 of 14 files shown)


st2common/st2common/services/policies.py
```diff
@@ -0,0 +1,69 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import absolute_import
+
+from st2common.constants import action as ac_const
```
Kami: I noticed you started using short hand variable names for various imports in your recent changes (ac_ex_db -> execution_db, etc.).

Would be good to use consistent name across the code base, since in most of the old / existing code base we use different aliases names for the imports and variables.

st2actions/st2actions/notifier/notifier.py
```diff
@@ -79,29 +76,27 @@ def __init__(self, connection, queues, trigger_dispatcher=None):
...
-        liveaction_id = execution_db.liveaction['id']
-        liveaction_db = LiveAction.get_by_id(liveaction_id)
-        self._apply_post_run_policies(liveaction_db=liveaction_db)
+        # Get the corresponding liveaction record.
+        liveaction_db = LiveAction.get_by_id(execution_db.liveaction['id'])
+
+        # If the action execution is executed under an orquesta workflow, policies for the
+        # action execution will be applied by the workflow engine. A policy may affect the
+        # final state of the action execution thereby impacting the state of the workflow.
+        if 'orquesta' not in execution_db.context:
```
Kami: I assume there is no better way to detect if it's an orquesta worklow.

Perhaps it would still be better to refactor it in utility function since it's used in multiple places - `is_execution_orquesta_workflow` or similar.

Kami:
Added a comment on aliased module import and variable naming consistency, besides that, LGTM :+1:

st2actions/st2actions/notifier/notifier.py
m4dcoder: There's no better way to detect if the action execution is executed as a task of an orquesta workflow.  I can add a helper method.

st2common/st2common/services/policies.py
m4dcoder: My plan is to start making similar changes across the code base. :P

Kami: Noooo :P

m4dcoder: Seriously though, i plan to clean up some existing code base.

Kami: I'm not a fan of it - there is really no need, unless the code is not using good variable names (which might be the case in some of the old existing code).

I don't like the short hand notation, especially since 90% of the existing code already uses variable names such as `execution_db`, `execution_api`, etc.

m4dcoder: I can probably live with execution_db, etc as is. But the individual import of action constants are just unsightly not to mention unnecessary as are other module import styles.

PR #4291, opened 2018-08-03
Upgrade mongoengine to the latest version (0.15.3)

Kami:
This pull request updates `mongoengine` to the latest version (0.15.3) from our fork (https://github.com/MongoEngine/mongoengine/pull/1833) which contains a work-around for a large regression in memory usage introduced in mongoengine v0.13.0.

I confirmed reverting that commit fixes an issue with increased memory usage.

Using our fork with a reverted commit:

```bash
(virtualenv) vagrant@ubuntu-xenial:/data/stanley$ python test_mongoengine.py 
Python: 2.7.12 (default, Dec  4 2017, 14:50:18) 
[GCC 5.4.0 20160609] on Linux-4.4.0-130-generic-x86_64-with-Ubuntu-16.04-xenial
mongoengine: 0.15.3
sleeping...

(virtualenv) vagrant@ubuntu-xenial:/data/stanley$ ps -eo rss,pid,euser,args:100 --sort %mem | grep -i 'test_mongoengine' |grep -v grep | awk '{printf $1/1024 "MB"; $1=""; print }'
18.3281MB 12696 vagrant python test_mongoengine.py
```

Using upstream version:

```bash
Python: 2.7.12 (default, Dec  4 2017, 14:50:18) 
[GCC 5.4.0 20160609] on Linux-4.4.0-130-generic-x86_64-with-Ubuntu-16.04-xenial
mongoengine: 0.15.3
sleeping...

(virtualenv) vagrant@ubuntu-xenial:/data/stanley$ ps -eo rss,pid,euser,args:100 --sort %mem | grep -i 'test_mongoengine' |grep -v grep | awk '{printf $1/1024 "MB"; $1=""; print }'
70.6875MB 6348 vagrant python test_mongoengine.py
```

Latest version will allow us to support MongoDB 3.6.

When / if mongoengine merges a patch for that issue upstream, we should move away from our fork.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -33,7 +33,7 @@ prometheus_client==0.1.1
 prompt-toolkit==1.0.15
 psutil==5.4.5
 pyinotify==0.9.6
-pymongo==3.6.1
+pymongo==3.7.1
 pyrabbit
 python-dateutil
 python-editor==1.0.3
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,10 +3,11 @@ RandomWords
 apscheduler==3.5.1
 argcomplete
 bcrypt
-cryptography==2.2.2
+cryptography==2.3.0
 eventlet==0.23.0
 flex==6.13.1
 git+https://github.com/Kami/logshipper.git@stackstorm_patched#egg=logshipper
+git+https://github.com/StackStorm/mongoengine.git@stackstorm_patched#egg=mongoengine
 git+https://github.com/StackStorm/orquesta.git@master#egg=orquesta
 git+https://github.com/StackStorm/python-mistralclient.git#egg=python-mistralclient
 git+https://github.com/StackStorm/st2-auth-backend-flat-file.git@master#egg=st2-auth-backend-flat-file
@@ -20,7 +21,6 @@ jsonschema==2.6.0
 kombu==4.2.1
 lockfile==0.12.2
 mock==2.0.0
-mongoengine==0.12.0
 networkx==1.11
 nose
 nose-timer<0.8,>=0.7.2
--- a/st2api/in-requirements.txt
+++ b/st2api/in-requirements.txt
@@ -2,7 +2,10 @@
 eventlet
 jsonschema
 kombu
-mongoengine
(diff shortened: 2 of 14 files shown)


st2api/in-requirements.txt
```diff
@@ -2,7 +2,10 @@
 eventlet
 jsonschema
 kombu
-mongoengine
+# Our fork contains a fix / revert of a change which was introduced in v0.13.0
+# and introduces a large regression in memory usage
+# See https://github.com/MongoEngine/mongoengine/pull/1833
```
armab: :+1:

requirements.txt
```diff
@@ -3,10 +3,11 @@ RandomWords
 apscheduler==3.5.1
 argcomplete
 bcrypt
-cryptography==2.2.2
+cryptography==2.3.0
 eventlet==0.23.0
 flex==6.13.1
 git+https://github.com/Kami/logshipper.git@stackstorm_patched#egg=logshipper
+git+https://github.com/StackStorm/mongoengine.git@stackstorm_patched#egg=mongoengine
```
m4dcoder: I'm ok with using a fork only if our patch is going to merge upstream in the near term. But if this is indefinitely, this is harder to agree because we will have to maintain this fork moving forward.

Kami: The problem is that I'm not really aware of a good upstream fix (besides reverting that change).

I wish there was a better way, but it doesn't look like there is one and the performance regression is sadly too large for us to simply ignore it.

"Good" news is that mongoengine has relatively slow release cadence so keeping our fork up to date shouldn't be too bad (keep in mind that we haven't upgraded mongoengine in our repo for 1 year because of that regression).

Kami:
Finally managed to also get unit tests on Travis to also run under MongoDB 3.6 - https://travis-ci.org/StackStorm/st2/builds/412583451

I will go ahead and merge this into master.

Kami:
It looks like tests run much slower under MongoDB 3.6.

It might be related to this - https://jira.mongodb.org/browse/SERVER-34604

armab:
A lot of changes added afterwards is not what I reviewed really.

And timing for Travis longest task is back returned to almost half-an-hour :/

Issue #4296, opened 2018-08-07
Creating your first pack docs issue


shreyas-shinde:
##### SUMMARY

Documentation issue.  
Here https://docs.stackstorm.com/reference/packs.html#creating-your-first-pack it says using runner_type: "local-shell-cmd" . It should be runner_type: "local-shell-script" else it gives error  
raise ValueError('entry_point is only valid for local-shell-script runner')

##### ISSUE TYPE
Pick one below and delete the rest:
 - Other: Documentation

armab:
Our docs are Open Source too: https://github.com/StackStorm/st2docs/blob/master/docs/source/reference/packs.rst
Contributions are very welcome!

PR #4298, opened 2018-08-07
Fix hello_st2 greet action

LindsayHill:
Somehow our example `hello_st2.greet` action had the wrong runner type, and no-one noticed for years.

Fixes https://github.com/StackStorm/st2/issues/4296

### DIFF
--- a/contrib/hello_st2/actions/greet.yaml
+++ b/contrib/hello_st2/actions/greet.yaml
@@ -1,7 +1,7 @@
 ---
 name: greet
 pack: hello_st2
-runner_type: "local-shell-cmd"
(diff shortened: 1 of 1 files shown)


Kami:
We should hook all the packs inside `contrib/` directory up to the same checks we use for exchange repos (if we don't already do that).

We have a "register resources" step in those checks and that should have caught issues like that.

EDIT: I just had a look and it's a run time error so that won't work.

One option is to change json schema so entry_point is not allowed when runner_type == local-shell-cmd (not even sure that's possible with jsonschema though - you can make two attributes mutually exclusive but not sure if you can do that depending on the attribute value).

PR #4300, opened 2018-08-09
[WIP] Better support for ?include_attributes and ?exclude_attributes API query param filters

Kami:
This pull request improves handling of the `include_attributes` and `exclude_attributes` API query param filters.

With those query parameters user can specify which attributes it wants to receive in the API response. Among other things, it also controls which fields we retrieve from the database.

Because of the way our API models and responses work, we convert database document to the API model before serializing that to JSON and returning it as a response. API model performs object validation and sets the default values for the fields which are not provided.

That means, for example, if user only requests `name` field it may get many other fields back, depending on the default values specified on the API model jsonschema (and those fields will use default values from the API model schema which is confusing and invalid).

The problem is related to the issue (deficiency in our API model handling) I've already mentioned a long time ago. We are abusing API model for two purposes - input and output. We really should have two models (one for input and one for output) and this way we would avoid issues like that.

Sadly that would require a lot more work so I'm going with the middle of the ground approach to begin with.

As per discussion with @enykeev, I decided to perform this filtering inside the router. This way it happens very last, just before serializing the response to JSON and returning it back to the user.

In the first attempt, I tried to do it inside the API model, but it really doesn't belong there.  API model represents data which has been validated and it's complete and valid (based on the model jsonschema). API model with some fields missing could potentially be invalid, hence that data shouldn't be represented by the API model.

Current approach is not the most efficient, because we still convert each database document to the API model before filtering / processing the response, but it should be a start. Future optimization would involve skipping the API model layer all together when either `?exclude_attributes` or `?include_attributes` query param filters is provided. That involves more work and refactoring though.

## TODO

- [ ] Tests
- [ ] Add support for those filters to all "get_all" / list API endpoints?

### DIFF
--- a/1
+++ b/1
@@ -1,4 +1,0 @@
-2018-07-23 08:15:07,950 INFO [-] Installing pack "libcloude"
-2018-07-23 08:15:07,953 DEBUG [-] Starting new HTTPS connection (1): index.stackstorm.org
-2018-07-23 08:15:08,099 DEBUG [-] https://index.stackstorm.org:443 "GET /v1/index.json HTTP/1.1" 200 73045
-2018-07-23 08:15:08,156 ERROR [-] Failed to install pack "libcloude": No record of the "libcloude" pack in the index.
--- a/1
+++ b/1
@@ -1,0 +1,4 @@
+2018-07-23 08:15:07,950 INFO [-] Installing pack "libcloude"
+2018-07-23 08:15:07,953 DEBUG [-] Starting new HTTPS connection (1): index.stackstorm.org
+2018-07-23 08:15:08,099 DEBUG [-] https://index.stackstorm.org:443 "GET /v1/index.json HTTP/1.1" 200 73045
+2018-07-23 08:15:08,156 ERROR [-] Failed to install pack "libcloude": No record of the "libcloude" pack in the index.
--- a/st2api/st2api/controllers/resource.py
+++ b/st2api/st2api/controllers/resource.py
@@ -134,6 +134,11 @@ class ResourceController(object):
         include_fields = include_fields or []
         query_options = query_options if query_options else self.query_options
 
+        if exclude_fields and include_fields:
+            msg = ('exclude_fields and include_fields arguments are mutually exclusive. '
+                   'You need to provide either one or another, but not both.')
+            raise ValueError(msg)
+
         # TODO: Why do we use comma delimited string, user can just specify
         # multiple values using ?sort=foo&sort=bar and we get a list back
         sort = sort.split(',') if sort else []
@@ -201,11 +206,6 @@ class ResourceController(object):
                     filters['__'.join(path)] = v
                 except LookUpError as e:
                     raise ValueError(str(e))
-
-        if exclude_fields and include_fields:
-            msg = ('exclude_fields and include_fields arguments are mutually exclusive. '
-                   'You need to provide either one or another, but not both.')
-            raise ValueError(msg)
 
         instances = self.access.query(exclude_fields=exclude_fields, only_fields=include_fields,
                                       **filters)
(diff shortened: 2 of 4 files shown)


Kami:
Another thing to keep in mind - due to how we handle our resource models (actions, sensors, rules, triggers, etc.), fields which form a `ref` attribute (aka resource primary key / unique identifier) need to be always present in the response.

For resource models those fields are `name` and `pack` and for other models, that field is `id`. I personally think that's not an issue and is desirable since "get_all" / list response which doesn't contain item's primary key is of no use / makes no sense.

Kami:
@enykeev I'm still working on adding support for those query param filters to all the resource API endpoints and improved tests, but the core functionality is ready for review.

Main changes are pretty straight forward and include:

1. New `_process_response` method which handles attribute exclusion before sending response back to the client - https://github.com/StackStorm/st2/pull/4300/files#diff-364c1d17f394113df1634d1423d1a3c4R578
2. Skip response validation if response schema is available and those filters are provided - https://github.com/StackStorm/st2/pull/4300/files#diff-364c1d17f394113df1634d1423d1a3c4R540. It makes no sense to perform response validation when those filters are provided because response will be partial and will likely fail validation (this should also speed things up a bit).
3. Removal of old "include_reference" related code which we don't need anymore (we haven't need it for a long time now) - https://github.com/StackStorm/st2/pull/4300/files#diff-364c1d17f394113df1634d1423d1a3c4R540

Kami:
On a related note - it's actually a good idea that we do this parameter filtering at the very and don't skip the DB model and API model layer.

Secrets masking happens inside the DB and API layer which means skipping it would be a bad idea. Right now the function which handles filtering (including, exclusion) gets passed in an object which has already been masked so there is no way to "fish out" secret parameter.

I also plan to spend a good amount of time next week writing test cases for all the various edge cases to make sure this functionality doesn't negatively affect the security (notable secret parameters and related stuff).

Kami:
I'm working on the test cases as discussed before.

The following get all / list endpoints now support those filters:

1. `/v1/actions`
2. `/v1/actions/views/overview`
3. `/v1/rules`
4. `/v1/rules/views`
5. `/v1/runnertypes`
6. `/v1/sensortypes`
7.  `/v1/ruleenforcements`
8. `/v1/ruleenforcements/views`
9. `/v1/traces`
10. `/v1/triggertypes`
11. `/v1/triggerinstances`
12. `/v1/actionalias`
13. `/v1/executions`
14. `/v1/inquiries`
15. `/v1/packs`
16. `/v1/policytypes`
14. `/v1/policies`

So pretty much all of the content pack resource API endpoints.

As far as the tests go - I added a base test class which exercises various exclude and include attributes scenarios based on the some meta attributes on the API controller test class (get_all_path, controller_cls, include_attribute_field_name, exclude_attribute_field_name).

This allows us to avoid a lot of copy and paste and duplicate code and still exercise the various scenarios.

In addition to those common scenarios, controller specific scenarios are exercised directly on that controller test class (e.g. making sure that secrets are still correctly masked when using those filters for the executions controller, etc).

Sadly we still have a lot of "legacy" test cases which insert and delete test data inside the test case method instead of utilizing the fixtures and inserting data inside `setUp` / `setUpClass` methods. This means that the base test class is not as nice as I would want it to be.

I hope to address this in the future, but definitely not in the PR since it's already turning into a big change (sadly no way around that, we need good test coverage).

Another thing which I already pointed out and discussed with @enykeev, but I want to point out again is that some of the model have mandatory attributes which are always included in the response. Those are usually primary keys which means that for most content pack resources that is `pack` and `name` (which form a ref) and for non content pack resources that is `id`.

In addition to that, we have some models which have additional mandatory attributes which can't be excluded (e.g. `trigger` for rules and `action.parameters`, and runner.runner_parameters` for execution objects, etc.). Those are either needed because they form a compound primary key or because of some other functionality (e.g. to be able to mask secrets correctly).

/cc @enykeev

Kami:
On a related note - this pull request also exposed a lot of issues with the current handling of `exclude_attribute` and lack of good tests for some other API endpoints.

PR #4302, opened 2018-08-10
Tweak runner descriptions to render properly in our ReStructuredText documentation

blag:
Proper fix for StackStorm/st2docs#780.


LindsayHill:
Will this make the output display a little oddly when users do `st2 runner get <blah>`?

Or when a user does something like `st2 run core.local -h`, and sees that output there? Or where it renders those parameters in st2web?

Would we instead be better off doing some post-processing in the st2docs script that generates the runner docs?

PR #4305, opened 2018-08-16
Action and runner output schema

bigmstone:
Actions and runners currently have arbitrary output which is not modeled in any deterministic way. This PR introduces a required runner output schema and option action output schema to allow more deterministic usage inside workflows, and programatic linking between output of one action and input of another in a workflow context.

- [x] Output Schema Model
- [x] Output Schema Enforcement
- [x] Output Schema Requirement for Runners
- [x] Update Runners with Output Schema
- [x] Update API to Get Combined Schema for Action


contrib/runners/python_runner/python_runner/runner.yaml
```diff
@@ -4,6 +4,25 @@
   enabled: true
   name: python-script
   runner_module: python_runner
+  output_schema:
+    result:
+      anyOf:
+        - type: "object"
```
m4dcoder: You don't really need to double or single quote strings in YAML.

bigmstone: I do for `null` and I'd prefer it be consistent.

contrib/runners/cloudslang_runner/cloudslang_runner/runner.yaml
```diff
@@ -14,3 +14,8 @@
       description: Action timeout in seconds. Action will get killed if it doesn't
         finish in timeout seconds.
       type: integer
+  output_schema:
+    stdout:
+      type: "string"
+    stdout:
```
blag: Copy/paste error? I imagine that the second key should be `stderr`.

bigmstone: Good catch.

contrib/runners/action_chain_runner/action_chain_runner/runner.yaml
```diff
@@ -12,3 +12,8 @@
       default: []
       description: List of tasks to skip notifications for.
       type: array
+  output_schema:
+    published:
+      type: "object"
```
Kami: Do we need to add `additionalProperties: false` here?

contrib/runners/announcement_runner/announcement_runner/announcement_runner.py
```diff
@@ -60,7 +60,13 @@ def run(self, action_parameters):
...
                                   payload=action_parameters,
                                   trace_context=trace_context)
-        return (LIVEACTION_STATUS_SUCCEEDED, action_parameters, None)
+
+        result = {
+            "output": action_parameters
+        }
+        result.update(action_parameters)
+
+        return (LIVEACTION_STATUS_SUCCEEDED, result, None)
```
Kami: I see, so the output format changes here - will we add upgrade notes entry for it?

contrib/runners/local_runner/local_runner/runner.yaml
```diff
@@ -35,6 +35,31 @@
...
       type: integer
+  output_schema:
+    succeeded:
+      type: boolean
+    failed:
+      type: boolean
+    return_code:
+      type: integer
+    stderr:
+      anyOf:
```
Kami: Hm, `stderr` and `stdout` should always be strings, right? Or am I missing something?

contrib/runners/python_runner/python_runner/runner.yaml
Kami: Another small nit pick - I'm not a fan of our random ordering of attributes in runner.yaml files. It's all over the place. I believe that happened when we converted json to yaml automatically and didn't sort the ordering.

contrib/runners/windows_runner/windows_runner/runner.yaml
```diff
@@ -28,6 +28,31 @@
       description: Username used to log-in.
       required: true
       type: string
+  output_schema:
+    stdout:
```
Kami: Same here, shouldn't it always be string?

st2actions/st2actions/container/base.py
```diff
@@ -127,6 +128,40 @@ def _do_run(self, runner):
...
+                    "properties": runner.runner_type.output_schema,
+                    "additionalProperties": False
+                }
+
+            LOG.info("Action Result: %s", result)
+
+            schema.validate(result, runner_schema, cls=schema.get_validator('custom'))
+
+            if runner.action.output_schema:
+                output_schema = copy.deepcopy(runner.runner_type.output_schema)
```
Kami: Maybe now that we have `fast_deepcopy` function it might make sense to use that. Or perhaps since it's a small dict it's not a big deal.

st2actions/st2actions/container/base.py
```diff
@@ -127,6 +128,40 @@ def _do_run(self, runner):
...
+                    "additionalProperties": True
+                }
+            else:
+                runner_schema = {
+                    "type": "object",
+                    "properties": runner.runner_type.output_schema,
+                    "additionalProperties": False
+                }
+
+            LOG.info("Action Result: %s", result)
```
Kami: Should be careful with logging whole result under `.info` - it could contain potentially sensitive information and that's why, IIRC, we only log it under DEBUG in some scenarios.

st2actions/st2actions/container/base.py
```diff
@@ -127,6 +128,40 @@ def _do_run(self, runner):
...
+            if runner.action.output_schema:
+                output_schema = copy.deepcopy(runner.runner_type.output_schema)
+                output_schema.update(runner.action.output_schema)
+                action_schema = {
+                    "type": "object",
+                    "properties": output_schema,
+                    "additionalProperties": False
+                }
+                LOG.debug('Validating action output: %s', action_schema)
+                schema.validate(result, action_schema, cls=schema.get_validator('custom'))
```
Kami: Ah, I see - so we do two passes - first for runner schema and then for optional action schema.

Kami: I think it would make sense / make it more user-friendly if we captured original exception and re-throw a more user-friendly one.

E.g. "Failed validation output for action "foo" against local runner schema: blah blah".

st2client/st2client/formatters/execution.py
```diff
@@ -33,6 +34,16 @@
 PLATFORM_MAXINT = 2 ** (struct.Struct('i').size * 8 - 1) - 1
 
 
+def _print_bordered(text):
+    lines = text.split('\n')
+    width = max(len(s) for s in lines) + 2
+    res = ['\n+' + '-' * width + '+']
+    for s in lines:
+        res.append('| ' + (s + ' ' * width)[:width - 2] + ' |')
+    res.append('+' + '-' * width + '+')
```
Kami:
```
        t
       .#.
      .###.
     .#%##%.
    .%##%###.
   .##%###%##. 
  .#%###%##%##.
        #
        #
```

Kami:
Added a couple of questions and comments - nothing really major.

st2actions/st2actions/container/base.py
Kami: I didn't see any tests for those scenarios (action output failing runner schema and action output failing action output schema).

Did I miss it, or we don't have any yet? :trollface:

contrib/runners/local_runner/local_runner/runner.yaml
bigmstone: You would think, but the way our jsonify works is it traverses any keys and tries to render JSON. So any valid JSON is rendered. For instances if `stdout` is `123` you get the functional equivalent of this:

```
In [1]: import json

In [2]: type(json.loads('123'))
Out[2]: int
```

st2actions/st2actions/container/base.py
bigmstone: I'll use it for consistency now that it exists. Plus never know. People do crazy things sometimes. That 1000 param output action will exist one day.

st2actions/st2actions/container/base.py
bigmstone: Ack

contrib/runners/windows_runner/windows_runner/runner.yaml
bigmstone: Same as above. https://github.com/StackStorm/st2/pull/4305#discussion_r216313660

contrib/runners/action_chain_runner/action_chain_runner/runner.yaml
bigmstone: Actually yeah this might be a good place to add it. I enforce that on actions but not runners because I couldn't refactor all runners w/o some more time investment and that transition needs to happen over a couple versions since output will change. Moving `additionalProperties` here will help w/ it defaulting to `false` for now if nothing is provided.

contrib/runners/announcement_runner/announcement_runner/announcement_runner.py
bigmstone: Yes, but it should also be backwards compatible so long as the user isn't iterating over all keys.

contrib/runners/local_runner/local_runner/runner.yaml
Kami: Ah, that's a bit nasty, I thought this only happens for result not also for stdout and stderr.

Kami: EDIT: Had a look again and notice that's just for the local runner, Python runner is "fine".

Still a bit nasty, but yeah, sadly an existing behavior.

contrib/runners/windows_runner/windows_runner/runner.yaml
bigmstone: If I'm reading this correctly it's all output for every runner.
https://github.com/StackStorm/st2/blob/fd9c6f59887efe588ecbce562ac0d80b491edd6a/st2actions/st2actions/container/base.py#L129

contrib/runners/local_runner/local_runner/runner.yaml
bigmstone: If I'm reading this correctly it's all output for every runner.
https://github.com/StackStorm/st2/blob/fd9c6f59887efe588ecbce562ac0d80b491edd6a/st2actions/st2actions/container/base.py#L129

Kami: Doesn't seem so for Python runner:

```bash
  stderr: '4321'
  stdout: '1234'

kami ~/w/stackstorm/st2 (git:master)$ cat contrib/examples/actions/pythonactions/print_python_version.py
import sys

class PrintPythonVersionAction(Action):

    def run(self):
        sys.stdout.write('1234')
        sys.stderr.write('4321')
```

And looking and `try_json` source, it doesn't seem to do any type casting.

contrib/runners/action_chain_runner/action_chain_runner/runner.yaml
bigmstone: Never mind. Implementing this here it here led to some weirdness w/ needing to understand some jsonschema semantics are going on. Either way the net result is that additional params is set to false in the container.

contrib/runners/announcement_runner/announcement_runner/runner.yaml
```diff
@@ -16,3 +16,6 @@
       maxLength: 255
       minLength: 1
       type: string
+  output_schema:
+    unmodeled:
+      type: boolean
```
m4dcoder: Why do we need to have this unmodeled property here?  If there's no output schema or if output schema is empty, isn't it unmodeled by default?

contrib/runners/orquesta_runner/orquesta_runner/runner.yaml
```diff
@@ -4,3 +4,17 @@
   enabled: true
   name: orquesta
   runner_module: orquesta_runner
+  output_schema:
+    errors:
```
m4dcoder: This should only be an array.

contrib/runners/orquesta_runner/orquesta_runner/runner.yaml
```diff
@@ -4,3 +4,17 @@
...
+      anyOf:
+        - type: "object"
+        - type: "array"
+    output:
```
m4dcoder: This should only be types of `object` or `null`.

st2actions/st2actions/container/base.py
m4dcoder: What if the runner is asynchronous like orquesta and mistral?  Where is the schema check when these runners post result back to st2? Orquesta handles and updates output for the live action and action execution at https://github.com/StackStorm/st2/blob/master/st2common/st2common/services/workflows.py#L829.  Mistral posts updates over API via https://github.com/StackStorm/st2mistral/blob/master/st2mistral/notifiers/stackstorm_notifier.py to PUT of action executions at https://github.com/StackStorm/st2/blob/master/st2api/st2api/controllers/v1/actionexecutions.py#L547.

m4dcoder: I believe the common denominator for both is update_status in action service.  So I would 1) refactor the output schema check into a method at action service 2) have the container base call this new method 3) add an output schema check in update_status of action service when result is updated 4) refactor update_execution_records in workflow service to also use the action_service.update_status instead of the update_liveaction_status and update_execution.

m4dcoder:
@bigmstone Also, I don't see much integration tests here such as orquesta integration tests, e2e, etc.?  If I missed them, please point me to them.

bigmstone:
@Kami @m4dcoder take a look at the new location it's implemented. If you guys don't see any negative/undesired side effects from locating it there I'm going to wrap some more tests around it and call it done.

st2common/st2common/util/action_output.py
```diff
@@ -0,0 +1,85 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
```
m4dcoder: I think over time, the team didn't make the st2 software architecture very clean.  I would call them in util modules if these functions are used only by a single st2 component.  If it's a functionality that is required by other st2 components, i kinda prefer them to be part of the service modules.  Having said that, I prefer the main function for validating output to be part of the action service.  In this case, the API and the workflow service would also need to access this.

m4dcoder: @bigmstone making the point short, can you move these functions to `st2common/services/action.py`?

m4dcoder: At first, I was thinking that you can implement the action output validation methods in action service and then put the check at https://github.com/StackStorm/st2/blob/master/st2common/st2common/services/action.py#L186. But then looking at what you just pointed me since the validation is going to be called by the action db util, I guess I'm ok with where you put them.  To not block this, just make sure it errors appropriately for orquesta and mistral.

contrib/runners/action_chain_runner/action_chain_runner/runner.yaml
```diff
@@ -12,3 +12,10 @@
       default: []
       description: List of tasks to skip notifications for.
       type: array
+  output_schema:
```
m4dcoder: To be honest, I still think we should restrict where the action output schema gets applied to separately for each runner.  Having said that, for the sake of progressing this, I don't think `output_path` should be under `output_schema`, that's just confusing.  Also why is `output_path` a list?  I don't see any of the runners below having more than one output_path.

```
output_path: published
output_schema:
  published:
    type: object
  ...
```

st2tests/st2tests/fixtures/generic/runners/actionchain.yaml
```diff
@@ -4,3 +4,6 @@ enabled: true
 name: action-chain
 runner_module: action_chain_runner
 runner_parameters: {}
+output_schema:
```
m4dcoder: Again, I prefer the absence of an output_schema over explicitly saying unmodeled here.  Also, even if I agree, this is the same problem with output_path, i prefer the unmodeled to be outside of the output_schema so not to be confused with the actual parameters.

st2tests/st2tests/fixtures/generic/actions/action-with-output-schema.yaml
```diff
@@ -0,0 +1,19 @@
...
+pack: core
+parameters:
+  cmd:
+    required: true
+  sudo:
+    immutable: true
+    type: boolean
+runner_type: "local-shell-cmd"
+output_schema:
+  stdout:
```
m4dcoder: Since there's an `output_path` now, why do we need to put stdout here again?  Couldn't we simplify this to the following?

```
output_schema:
  type: object
  parameters:
    bob:
      type: string
```

Kami:
I believe there were some issues with everyone trying to get on the same page so I will also add my high level feedback / opinion on this feature.

From what I mentioned in the past, I believe action output schema really only makes sense for actions which utilize runners which support returning different data types (object, array, etc.) such as Python runner, workflow runner and perhaps also http runner.

In local and remote runner there is really no good concept of a result and they work with strings (stdout and stderr). In most scenarios, result from those runners is not directly usable in other actions and workflows and needs to be formatted / processed by another (Python runner) action or workflow.

So having said that, I think action output schema makes little sense for runners which only work with strings. You could argue we could support regexes or something, but I think that's an overkill and doesn't really provide value, just complexity - and in the end that's action output schema functionality and not action output transformation functionality.

So to tie back to the implementation details.

1. Runner output schema - it's fine as it is. I really don't see it providing any value to the end user and it's just there to guard against "developer" errors. If runner output schema fails validation on an end user deployment, this means we (developers) introduced a bug or similar. So having said that, ideally CI should catch all those issues.

2. Action output schema - Basically what I wrote above.

As far as "notation" goes (I believe we had a hard time find consensus on that) - I think it can be simplified if we agreed what's the result field for a particular runner - e.g. "output" / "result". So schema would apply directly to that result field.

This way user doesn't need to duplicate keys such as output, result, etc. and this would reduce one level of nesting.

Again, I think this makes more sense if you view the feature as I do - it only applies and makes sense for runners which can return different result types. And in general schema mostly makes sense on objects - there is little you could do with a string. Granted, you could use a regex or something, but that's complicated, especially if you consider how local and remote runner output usually looks like (hard / impossible to write schema for it and usually not useful without some post processing / formatting).

st2tests/st2tests/fixtures/generic/runners/actionchain.yaml
bigmstone: Agree. I've added that in most recent commit.

st2tests/st2tests/fixtures/generic/actions/action-with-output-schema.yaml
bigmstone: Correct, but the test that used this was removed. I'll fix this fixture when I add the test back in.

contrib/runners/orquesta_runner/tests/unit/test_output_schema.py
```diff
@@ -0,0 +1,170 @@
...
+        wf_svc.handle_action_execution_completion(tk3_ac_ex_db)
+        wf_ex_db = wf_db_access.WorkflowExecution.get_by_id(wf_ex_db.id)
+        lv_ac_db = lv_db_access.LiveAction.get_by_id(str(lv_ac_db.id))
+        ac_ex_db = ex_db_access.ActionExecution.get_by_id(str(ac_ex_db.id))
+
+        with self.assertRaises(jsonschema.ValidationError):
+            output_schema._validate_runner(
+                FAIL_SCHEMA,
+                lv_ac_db.result
+            )
```
m4dcoder: This unit test looks wrong. If the output should fail schema validation, the lv_ac_db should be in a failed state with appropriate error message in the lv_ac_db.result['errors'].

contrib/runners/orquesta_runner/tests/unit/test_output_schema.py
```diff
@@ -0,0 +1,170 @@
...
+        tk3_ex_db = wf_db_access.TaskExecution.query(**query_filters)[0]
+        tk3_ac_ex_db = ex_db_access.ActionExecution.query(task_execution=str(tk3_ex_db.id))[0]
+        wf_svc.handle_action_execution_completion(tk3_ac_ex_db)
+        wf_ex_db = wf_db_access.WorkflowExecution.get_by_id(wf_ex_db.id)
+        lv_ac_db = lv_db_access.LiveAction.get_by_id(str(lv_ac_db.id))
+        ac_ex_db = ex_db_access.ActionExecution.get_by_id(str(ac_ex_db.id))
+
+        config = self.loader(os.path.join(BASE_DIR, '../../runner.yaml'))
+
+        output_schema._validate_runner(config[0]['output_schema'], lv_ac_db.result)
```
m4dcoder: This unit test looks wrong. The output schema should have been validated by the workflow service on completing the workflow. Why do we need to explicitly validate runner output here?

contrib/runners/orquesta_runner/tests/unit/test_output_schema.py
bigmstone: Failing w/ errors is still adhears to the output schema. But I will look through and change it to a succeeding WF.

contrib/runners/orquesta_runner/tests/unit/test_output_schema.py
bigmstone: Yes, it should be checked. This is an explicit test for the output so if for some reason the parameter gets change in config or some bug causes it to not be checked correctly this should catch output schema of Orquesta against an actual output.

contrib/runners/orquesta_runner/tests/unit/test_output_schema.py
bigmstone: I see what you're saying, but no the workflow can succeed from the Workflow engine's perspective but when it's handed over for DB write (where schema is checked) then it should raise the schema error and rewrite the result to the appropriate output. The workflow shows as failed. This test takes a valid output from orquesta's perspective but uses a different schema and confirms that it will raise when the output doesn't adhere to the schema. Probably better test would be to change the output of the WF result to a non-orquesta output instead of the schema itself.

Kami:
It looks like we are missing docs for this functionality (if I missed the docs, please point me to them).

A user asked about that functionality today. I wanted to link them to the docs and it looks like we don't have any (either that or my search failed me).

If we indeed don't have any docs, let's please add some to the actions section. Every feature should be treated as if it doesn't exist unless it has full test and docs coverage.

PR #4307, opened 2018-08-16
Improvements to the st2-inject-trigger-instances script

Kami:
Various improvements to the `st2-trigger-instances` script which is used to benchmark StackStorm (rules engine).

1. Specify duration in seconds
2. Reduce `random.expovariate` sleep by 55% to allow script to actually achieve rate requested by the user (that is, when requested rate is smaller than the max possible rate on that configuration). Previously actual achieved rate was always 50% or so less than the requested one, due to the overhead of `dispatcher.dispatch` call. (`eventlet.sleep(delta)` would only work correctly if duration of `dispatcher.dispatch` was zero aka no op). Keep in mind that the actual achieves rate might still be a bit less or a bit more than the requested one, but it will be much closer than before.
3. Add ``--max-throughput`` option which allows user to achieve maximum possible injection rate.

### DIFF
--- a/tools/st2-inject-trigger-instances.py
+++ b/tools/st2-inject-trigger-instances.py
@@ -62,6 +62,8 @@ def _inject_instances(trigger, rate_per_trigger, duration, payload=None, max_thr
         dispatcher.dispatch(trigger, payload)
 
         if rate_per_trigger:
+            # NOTE: We decrease sleep delay for 56% to take into account overhead / delay because
+            # of the call to dispatchet.dispatch method.
             delta = random.expovariate(rate_per_trigger)
             eventlet.sleep(delta * 0.56)
 
@@ -75,7 +77,7 @@ def _inject_instances(trigger, rate_per_trigger, duration, payload=None, max_thr
 
     # NOTE: Due the way this script works (allows user to specify a rate, actual rate will always
     # be a bit lower than the requested one)
-    if rate_per_trigger and (actual_rate < (rate_per_trigger * 0.60)):
+    if rate_per_trigger and (actual_rate < rate_per_trigger):
         print('')
         print('Warning, requested rate was %s triggers / second, but only achieved %s '
               'triggers / second' % (rate_per_trigger, actual_rate))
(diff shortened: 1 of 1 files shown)


tools/st2-inject-trigger-instances.py
```diff
@@ -58,12 +60,27 @@ def _inject_instances(trigger, rate_per_trigger, duration, payload={}):
     while elapsed < duration:
         # print('Dispatching trigger %s at time %s', trigger, date_utils.get_datetime_utc_now())
         dispatcher.dispatch(trigger, payload)
-        delta = random.expovariate(rate_per_trigger)
-        eventlet.sleep(delta)
-        elapsed = (date_utils.get_datetime_utc_now() - start).seconds / 60.0
+
+        if rate_per_trigger:
+            delta = random.expovariate(rate_per_trigger)
+            eventlet.sleep(delta * 0.56)
```
lakshmi-kannan: What is this 0.56?

tools/st2-inject-trigger-instances.py
```diff
@@ -58,12 +60,27 @@ def _inject_instances(trigger, rate_per_trigger, duration, payload={}):
...
 
-    print('%s: Emitted %d triggers in %d seconds' % (trigger, count, elapsed))
+    actual_rate = (count / elapsed)
+
+    print('%s: Emitted %d triggers in %d seconds (actual rate=%s triggers / second)' %
+          (trigger, count, elapsed, actual_rate))
+
+    # NOTE: Due the way this script works (allows user to specify a rate, actual rate will always
+    # be a bit lower than the requested one)
+    if rate_per_trigger and (actual_rate < (rate_per_trigger * 0.60)):
```
lakshmi-kannan: 0.60 seems like a magic number :).

tools/st2-inject-trigger-instances.py
Kami: Yeah, it is a bit of "magic".

For more accurate result, I would need to time ``dispatcher.dispatch`` call and subtract that from delta.

This "magic" number is what I got from my own testing (ran script multiple times, average rate achieved was usually 40-60% smaller than the desired one).

lakshmi-kannan: I'd remove all this magic numbers and treat the user supplied rate as approximate. Do we really require a precise rate?

Issue #4286, opened 2018-08-02
Multiple Inquiries in a workflow lead to pausing state frequently


wyjuly:
##### SUMMARY

This is related to #4135, #4137, #4138 and #4028

A single workflow:

workflow
	- action1
	- core.ask1
	- action2
	- core.ask2
	- action3
	
When core.ask1 comes the flow's status is changed to paused successfully. Respond to the inquiry and the flow status is changed to running and action2 is executed.
When core.ask2 comes the flow's status is changed to pausing. I test this many times and this issue is reproduced frequently even waiting for more than one hour. The flow is always pausing and core.ask2 status is pending.

It cannot be resumed in pausing state so I have to update in db.action_execution_d_b and change pausing to paused.
After that responding to the inquiry lead to:
ERROR: 500 Server Error: Internal Server Error
MESSAGE: Internal Server Error for url: http://127.0.0.1:9101/exp/inquiries/5b62842175ac351947192909

Finally I have to use st2 execution resume command to resume it. I also saw this in 2.6.

##### ISSUE TYPE
 - Bug Report

##### STACKSTORM VERSION
st2 2.7.2, on Python 2.7

##### OS / ENVIRONMENT / INSTALL METHOD
RHEL7.2 Stackstorm upgraded from 2.6

dbillor:
Experiencing same issue have to manually resume to fix it. Very frustrating

m4dcoder:
Is this written as action chain or as mistral workflows?

dbillor:
on my side mistral

m4dcoder:
Here is what I tested on Mistral and I'm not able to reproduce.

The following is the workflow I mocked that runs the core.ask, core.noop, core.ask, and core.noop sequentially.
```
version: '2.0'

sandbox.mistral_consecutive_inquiries:
    tasks:
        task1:
            action: core.ask
            input:
                schema:
                    type: object
                    properties:
                        approved:
                            type: boolean
                            description: "Continue?"
                            required: True
            on-success:
                - task2

        task2:
            action: core.noop
            on-success:
                - task3

        task3:
            action: core.ask
            input:
                schema:
                    type: object
                    properties:
                        approved:
                            type: boolean
                            description: "Really?"
                            required: True
            on-success:
                - task4

        task4:
            action: core.noop
```

Here's the execution result and the inquiry responses.

```
ubuntu@cadmus:~/st2$ st2 run sandbox.mistral_consecutive_inquiries -a
[...]

Issue #4138, opened 2018-05-22
core.ask in double nested mistral workflow sometimes fails to set the status of root workflow to paused


shusugmt:
##### ISSUE TYPE
 - Bug Report

##### STACKSTORM VERSION
```
st2 2.7.2, on Python 2.7
```

##### OS / ENVIRONMENT / INSTALL METHOD
- st2-docker

##### SUMMARY
This is related to #4135 and #4137 
Assume we have following mistral workflow:

```
root_workflow
  - nested_workflow
    - action1
    - action2
    - core.ask
    - action3
```

In `core.ask`, it will set the status of *parent* workflow = `nested_workflow` in this case, to `paused`. As a result, the root_workflow also changes to `paused` state.

But sometimes the root_workflow remains in `pausing` state. As far as I can see, it is just random. Some executions gets into paused, some won't and stuck in `pausing` state.  And it turns out that this seems to be the root cause of #4135

dbillor:
Experiencing same issue.  Have to manually resume and track workflows down.  Hard to reproduce as it is transient.

shusugmt:
This is solved by #4309

Issue #4135, opened 2018-05-21
Inquiries should not be responded when parent workflow is still in pausing state


shusugmt:
##### ISSUE TYPE
 - Bug Report

##### STACKSTORM VERSION
```
st2 2.7.2, on Python 2.7
```

##### OS / ENVIRONMENT / INSTALL METHOD
- st2-docker

##### SUMMARY

I'm facing the issue that in the mistral workflow there is `core.ask` action, and when I respond that with `st2 inquiry respond` *too fast*, the parent workflow state is sometimes still in pausing state (that means it is still transitioning from running to pause), and it fails to resume the parent workflow saying `ERROR: 500 Server Error: Internal Server Error` but the inquiry itself is already in success state and you can't respond it again to resume the parent workflow.

The workaround is, wait 10 sec before you respond. Also if you face the above error, you can just `st2 execution resume` to resume the parent workflow, but this apparently is not the desired situation.

Relevant code: https://github.com/StackStorm/st2/blob/master/st2api/st2api/controllers/exp/inquiries.py#L174
https://github.com/StackStorm/st2/blob/master/st2common/st2common/services/action.py#L320

Here's the log: you can see that LiveAction id `5b02dcfc069c4301d1f654ca` was in pausing state when the inquiry respond was trying to resume it.

```2018-05-21 14:53:31,108 139685573276848 INFO logging [-] 128c4d50-cf1d-433a-bd07-3c26cdfa5b2a - PUT /exp/inquiries/5b02dd2e069c4301d1f654ea with query={} (remote_addr='127.0.0.1',method='PUT',request_id='128c4d50-cf1d-433a-bd07-3c26cdfa5b2a',query={},path='/exp/inquiries/5b02dd2e069c4301d1f654ea')
2018-05-21 14:53:31,110 139685573276848 AUDIT auth [-] Token with id "5b02cee5069c4301cdba1881" is validated.
2018-05-21 14:53:31,189 139685573276848 INFO loader [-] Reusing runner module "/opt/stackstorm/runners/inquirer_runner/inquirer_runner/inquirer_runner.py" from cache.
2018-05-21 14:53:31,198 139685573276848 INFO loader [-] Reusing callback module "/opt/stackstorm/runners/mistral_v2/callback/mistral_v2.py" from cache.
[...]

Issue #4092, opened 2018-04-19
Responding to inquiry doesn't resume the flow


wyjuly:
Hi there,

Have you ever seen this before: 
I have a flow like
```
main_flow:
  subflow_1:
    subflow_1_2:
      subflow_1_2_3:
        inquiry
```
The inquiry is created, and the status of main_flow, subflow_1, subflow_1_2 and subflow_1_2_3 have been changed to paused, it's great.
But after responding to the inquiry successfully, only main_flow's status is changed to 'running', subflow_1, subflow_1_2 and subflow_1_2_3 are still in 'paused' status.

I cannot do anything but updating the mongodb, changed the status to  resuming and succeeded for subflow_1, subflow_1_2 and subflow_1_2_3, but the main_flow still cannot be resumed, it always show running and won't move forward.

Any suggestion for such scenario? Thanks a lot.
Jason

deepa2603:
I am also facing the same issue in mistral workflow. 

In mistral:

Action 1:
  on-failure: inquiry
Inquiry:
   on-success: A
A:
  echo a

After getting the response, inquiry is not resuming the workflow.

LindsayHill:
See #4309

PR #4309, opened 2018-08-17
 Add inquiry support for orquesta workflows

m4dcoder:
Move inquiry respond related functions to service module. Add inquiry support for orquesta workflows.  Fix a number of related bugs such as removing the pending status from completed states, fix duplicate task execution on subworkflow completion, race condition with resume on inquiry response, and race condition between processing a pending execution and pausing parent.


LindsayHill:
Looks like this resolves 
https://github.com/StackStorm/st2/issues/4286, 
https://github.com/StackStorm/st2/issues/4135, 
https://github.com/StackStorm/st2/issues/4092

m4dcoder:
@bigmstone this is covered in orquesta-itests

shusugmt:
I've tested this patch by continuously running the test mistral workflow for whole 3 days over 600 times in total and it seems the problem is gone. It never stuck in `pausing` state.

Tested environment:

```
root@10dc28221679:/# dpkg -l|grep st2
ii  st2                             2.9dev-41                                  amd64        StackStorm Event-driven automation
ii  st2chatops                      2.9dev-2                                   amd64        St2Chatops - StackStorm ChatOps.
ii  st2mistral                      2.9dev-5                                   amd64        st2 Mistral workflow service.
ii  st2web                          2.9dev-10                                  amd64        St2Web - StackStorm Web UI.
```

shusugmt:
This also can close #4138 

@LindsayHill #4135 originally addresses a different issue, so I'd like to keep that open. The original issue is to make `st2 inquiry respond` don't do anything (just show an error message and exit) if the root workflow is in `pausing` state. This still can be possible even after this patch is merged and the `pausing` bug is gone.

wyjuly:
Thanks a lot @m4dcoder 
So how can we test this fixing? In the future release 2.8.2 right?

wyjuly:
@m4dcoder  Got it and thanks again.

Kami:
@m4dcoder Nice work on the tests :+1: 

A couple of days ago, I started to notice more intermediate failurrs of mistral inquiry e2e tests. It seems to correlate with the time when this PR was merged. I wonder if it's related.

I would appreciate if you could have a look - here is an example failure https://gist.github.com/Kami/216b3a13abcf7d13d462b12e2cd019b1 (cicd will be rebuilt soon so that execution will be gone and that's why I gist part of it).

Kami:
I believe this should fix the intermediate failure issue - https://github.com/StackStorm/st2tests/commit/504565bac7830c7c97a7fff231a4156ba49b317a

The whole test itself and the way we check for status is quite fragile and not robust, but I just want to get it a bit more reliable without spending the whole day refactoring and improving those tests.

m4dcoder:
@Kami thank you for fixing st2tests! I appreciate your efficiency.

PR #4310, opened 2018-08-17
Additional code metrics instrumentation

Kami:
This pull request adds some additional metrics instrumentation to various services.

## Rules Engine

1. Track how many rules (trigger instances) are processed by rules engine and how long each processing took (counter + timer).
2. Track how long it took to process each unique trigger instance by the rules engine (timer, counter makes no sense since each TriggerInstance is unique and will always only be processed once).

## st2api, st2auth, st2stream

* Number of requests / second + request processing timing info
* Counters for various request and response related info:
  * request method
  * request path
  * response status code

## Other Changes

* [x] EchoMetricsDriver which prints out all the metrics information to console which makes debugging easier

## TODO

- [x] We need some documentation on exposed metrics and explanation what each one represents


st2reactor/st2reactor/rules/worker.py
```diff
@@ -87,7 +90,11 @@ def process(self, pre_ack_response):
 
             container_utils.update_trigger_instance_status(
                 trigger_instance, trigger_constants.TRIGGER_INSTANCE_PROCESSING)
-            self.rules_engine.handle_trigger_instance(trigger_instance)
+
+            with CounterWithTimer(key="st2.rule.processed"):
```
Kami: @bigmstone I'm open to a better name, something which would also make it consistent with action runner metric names.

I couldn't come up with anything better :/

Kami:
@bigmstone While working on that, I noticed we are missing "Gauge" metric type (so we can measure also total number of requests and similar and not just requests / second).

I will add that in this PR.

st2reactor/st2reactor/rules/worker.py
bigmstone: We should probably create a function to generate these namings in a standard way. Have a few input params so it's not so subjective, but like you I don't have many strong convictions here. I think it mainly just matters that it's consistent.

dzimine:
may be: 
* do we already have number of actions / sec - scheduled, executed, action queue size, time per action? 
* on rules - it's interesting to add color to rule processing: 
    * triggers instances, rule evaluations (per trigger), rule invocations (per trigger)

Issue #4312, opened 2018-08-21
Emit st2.generic.actiontrigger	when action/wf is requested/scheduled/running


shusugmt:
##### SUMMARY

It is better for uses to have choice that `st2.generic.actiontrigger` be emitted not only on action complete, but when an action or workflow is transitioned into requested/scheduled/running state.

##### ISSUE TYPE
 - Feature Idea

##### Detail / Background / Motivation

While I was trying to implement a Slack notification at both start and end of the workflow, I noticed that `st2.generic.actiontrigger` is only emitted when the workflow reaches [LIVEACTION_COMPLETED_STATES](https://github.com/StackStorm/st2/blob/master/st2common/st2common/constants/action.py#L112-L118), which makes me hard to implement a notification for the **start** of the workflow.

My idea is to use `st2.generic.actiontrigger` and write a rule that hook `state=running|succeeded` by criteria, then run a Slack sendMessage action. I think this is the best approach and design because the target workflow that will be notified on start/end is completely separated from any of the notification code. So notification part is like an *aspect* and we can *weave* that into any workflow you like by writing a rule. I can clearly keep a *separation of responsibility* principle. Some examples:
- All those configuration parameters needed for notification like tokens, channels, etc does not exist in the workflow code itself
- The notification rules can even be placed in a different pack and separated from the release cycle of those core workflows (which are tend to be more important than notifications since they do real business logic)
- If you want to *weave notification*, then you don't need to modify the target workflow. All you need is to write a rule. This makes us easy to add notification to the exchange pack workflows.

This is fairly easy to implement, but an impact to performance needs to be considered. So I think it is better to have a new config parameter and make this feature opt-in, like [this](https://github.com/StackStorm/st2/blob/master/conf/st2.conf.sample#L4-L6).

Kami:
Thanks for the feedback - I think that's a reasonable future request (and putting it behind a feature flag / config option is also a good idea).

cognifloyd:
I believe this is complete with #4315 and #4330.

Kami:
Yep - thank you.

PR #4315, opened 2018-08-24
Add option to specify when to emit st2.generic.actiontrigger

shusugmt:
As per motivation addressed in #4312 

- Introduced new config parameter `action_sensor.emit_when` which specifies the states when to emit `st2.generic.actiontrigger`
- If not specified, it defaults to `LIVEACTION_COMPLETED_STATES ` that is the exact same behavior as before. Meaning this change will not affect any existing deployments.
- If any state is given, it overrides the default. So users can even choose only to make actiontrigger emitted at for e.g. cancelled. If users want to completely disable this trigger then they can use existing `action_sensor.enable` parameter set to False.
- This PR also includes a bug fix that `action_sensor.enable` parameter was not honored at all.


nmaludy:
+100 !!!

I was thinking the other day it would be great if there were more triggers in the system for "internal" events like this!

st2actions/st2actions/notifier/notifier.py
```diff
@@ -231,10 +226,18 @@ def _get_trace_context(self, execution_id):
         return None
 
     def _post_generic_trigger(self, liveaction_db=None, execution_db=None):
-        if not ACTION_SENSOR_ENABLED:
+        if not cfg.CONF.action_sensor.enable:
             LOG.debug('Action trigger is disabled, skipping trigger dispatch...')
             return
 
+        target_statuses = cfg.CONF.action_sensor.emit_when
```
Kami: Would be a good idea to log a message under DEBUG if trigger is not emitted due to the status not being in ``emit_when` `list.

st2common/st2common/config.py
```diff
@@ -388,6 +388,10 @@ def register_opts(ignore_errors=False):
         cfg.BoolOpt(
             'enable', default=True,
             help='Whether to enable or disable the ability to post a trigger on action.'),
+        cfg.ListOpt(
+            'emit_when', default=[],
```
Kami: For explicitness sake, you could perhaps default this option to `LIVEACTION_COMPLETED_STATES`.

This way you could also simple the code here a bit - https://github.com/StackStorm/st2/pull/4315/files#diff-70b51c2693ac263742fc28ad30ebc932R234.

Kami:
Nice work, thank you :+1: 

I've added some comments, nothing major.

st2actions/st2actions/notifier/notifier.py
```diff
@@ -79,22 +78,18 @@ def process(self, execution_db):
         extra = {'execution': execution_db}
         LOG.debug('Processing action execution "%s".', execution_id, extra=extra)
 
-        if execution_db.status not in LIVEACTION_COMPLETED_STATES:
```
Kami: Not a blocker, but I found previous code a bit easier to read / follow ("return early" vs "nested if", one level less of nesting).

Was there any particular reason why you changed it?

st2common/st2common/config.py
shusugmt: > For explicitness sake, you could perhaps default this option to LIVEACTION_COMPLETED_STATES.
If we specify the list here explicitly, then we need to follow `LIVEACTION_COMPLETED_STATES` changes that may happen in the future. So I thought this way is better but I'm OK with either option. What do you think?

> This way you could also simple the code here a bit
Anyways that test case still needs to be there because users might have chance to specify a null or a blank list.

st2actions/st2actions/notifier/notifier.py
shusugmt: This change is inevitable since:

1. I thought the order of other function calls is important here and shouldn't be changed (to keep this: `apply_post_run_policies` -> `_post_notify_triggers` -> `_post_generic_trigger`)
2. We always need to call `_post_generic_trigger` regardless of the status. So If we want to do a "early return" then I need to move that call before `if execution_db.status not in LIVEACTION_COMPLETED_STATES:` line but that will break the order (`_post_generic_trigger` will be called before `_post_notify_triggers`)

st2actions/st2actions/notifier/notifier.py
shusugmt: Will do. Thanks for the comments!

shusugmt:
@Kami PR updated. Please take a look.

st2actions/st2actions/notifier/notifier.py
```diff
@@ -231,11 +226,25 @@ def _get_trace_context(self, execution_id):
...
     def _post_generic_trigger(self, liveaction_db=None, execution_db=None):
-        if not ACTION_SENSOR_ENABLED:
+        if not cfg.CONF.action_sensor.enable:
             LOG.debug('Action trigger is disabled, skipping trigger dispatch...')
             return
 
         execution_id = str(execution_db.id)
+        extra = {'execution': execution_db}
+
+        target_statuses = cfg.CONF.action_sensor.emit_when
```
Kami: Small style thing - I would do `target_statuses = cfg.CONF.action_sensor.emit_when or LIVEACTION_COMPLETED_STATES` this way we could get rid of else clause.

Besides that, LGTM :+1:

st2common/st2common/config.py
Kami: I will merge this into master and open a small PR with changelog entry and this change here which will allow us to slightly simplify the if statement.

Kami: And another reason why I prefer to use a default value here is because it's more explicit.

If this option contain no default value here, user needs to dig through the code and documentation to figure out what a default value of `None` indicates.

Kami:
Merged, thanks!

Changelog and default value change in #4330.

PR #4320, opened 2018-08-28
Add API endpoint and CLI command for orquesta workflow inspection

m4dcoder:
Add `/v1/workflows/inspect` endpoint in the API and a POST method to inspect workflow definition and return errors found. Add a `st2 workflow inspect` command that uses the API endpoint for workflow inspection.

### DIFF
--- a/contrib/runners/orquesta_runner/tests/unit/test_error_handling.py
+++ b/contrib/runners/orquesta_runner/tests/unit/test_error_handling.py
@@ -100,6 +100,43 @@ class OrquestaErrorHandlingTest(st2tests.DbTestCase):
         return sorted(errors, key=lambda x: x.get('task_id', None))
 
     def test_fail_inspection(self):
+        expected_errors = [
+            {
+                'type': 'content',
+                'message': 'The action "std.noop" is not registered in the database.',
+                'schema_path': 'properties.tasks.patternProperties.^\w+$.properties.action',
+                'spec_path': 'tasks.task3.action'
+            },
+            {
+                'type': 'context',
+                'language': 'yaql',
+                'expression': '<% ctx().foobar %>',
+                'message': 'Variable "foobar" is referenced before assignment.',
+                'schema_path': 'properties.tasks.patternProperties.^\w+$.properties.input',
+                'spec_path': 'tasks.task1.input',
+            },
+            {
+                'type': 'expression',
+                'language': 'yaql',
+                'expression': '<% <% succeeded() %>',
+                'message': (
+                    'Parse error: unexpected \'<\' at '
+                    'position 0 of expression \'<% succeeded()\''
+                ),
+                'schema_path': (
+                    'properties.tasks.patternProperties.^\w+$.'
+                    'properties.next.items.properties.when'
+                ),
+                'spec_path': 'tasks.task2.next[0].when'
+            },
+            {
+                'type': 'syntax',
+                'message': '[{\'cmd\': \'echo <% ctx().macro %>\'}] is not of type \'object\'',
+                'schema_path': 'properties.tasks.patternProperties.^\w+$.properties.input.type',
+                'spec_path': 'tasks.task2.input'
+            }
+        ]
+
         wf_meta = base.get_wf_fixture_meta_data(TEST_PACK_PATH, 'fail-inspection.yaml')
         lv_ac_db = lv_db_models.LiveActionDB(action=wf_meta['name'])
         lv_ac_db, ac_ex_db = ac_svc.request(lv_ac_db)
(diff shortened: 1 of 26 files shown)


nmaludy:
Is this a validation or something different?

m4dcoder:
@nmaludy yes, this is for validation

st2common/st2common/router.py
```diff
@@ -365,13 +365,15 @@ def __call__(self, req):
...
                 # NOTE: HACK: Workaround for eventlet wsgi server which sets Content-Type to
                 # text/plain if Content-Type is not provided in the request.
-                # All ouf our API endpoints except /exp/validation/mistral expect application/json
-                # so we explicitly set it to that if not provided (set to text/plain by the base
-                # http server) and if it's not /exp/validation/mistral API endpoint
+                # All ouf our API endpoints except /v1/workflows/inspection and
+                # /exp/validation/mistral expect application/json so we explicitly set it to that
+                # if not provided (set to text/plain by the base http server) and if it's not
+                # /v1/workflows/inspection and /exp/validation/mistral API endpoints.
                 if not self.is_gunicorn and content_type == 'text/plain':
```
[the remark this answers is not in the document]
Kami: An alternative would be to update our code to support "arbitrary content types" (really just `application/json` for everything else and `text/plain` for this endpoint).

This would require us, at the very least, to update `openapi.yaml.j2` file and define expected / valid content type for each API controller and making necessary changes in the router.

This wouldn't be too much work and worth thinking about it while we are already working on this (so we can get rid of this ugly hack).

contrib/runners/orquesta_runner/tests/unit/test_error_handling.py
```diff
@@ -100,19 +100,51 @@ def sort_wf_runtime_errors(self, errors):
         return sorted(errors, key=lambda x: x.get('task_id', None))
 
     def test_fail_inspection(self):
+        expected_errors = [
```
Kami: Overall, great improvement over Mistral :+1:

contrib/runners/orquesta_runner/tests/unit/test_error_handling.py
```diff
@@ -100,19 +100,51 @@ def sort_wf_runtime_errors(self, errors):
...
+                'message': 'Variable "foobar" is referenced before assignment.',
+                'schema_path': 'properties.tasks.patternProperties.^\w+$.properties.input',
+                'spec_path': 'tasks.task1.input',
+            },
+            {
+                'type': 'expression',
+                'language': 'yaql',
+                'expression': '<% <% succeeded() %>',
+                'message': (
+                    'Parse error: unexpected \'<\' at '
```
Kami: I know this question belongs to Orquesta more than here, but do we also have access to he line number where the error occurred? And can we include that in the error?

I see there is path which can help the user, but line number would make troubleshooting even easier.

st2client/st2client/commands/workflow.py
```diff
@@ -0,0 +1,111 @@
...
+            content = f.read()
+
+        return content
+
+    def run(self, args, **kwargs):
+        wf_def_file = args.file
+
+        # If file path not provided, try the action reference. This only works if the command
+        # is executed locally where the content is stored.
+        if not wf_def_file:
```
Kami: Yeah, I was wondering how much sense this option makes...

st2common/st2common/router.py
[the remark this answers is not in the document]
m4dcoder: I'm more incline to keep the hack as is. If this doesn't work, I probably would fix what you suggested and update all the API controllers. But since this hack already exists and i'm not deviating, I prefer to keep this change. If we want to improve the API in general, that seems like a separate PR to me.

contrib/runners/orquesta_runner/tests/unit/test_error_handling.py
m4dcoder: Line number is definitely nice to have. But there are multiple types of inspection here. As @bigmstone stated, some of these categories (i.e. contents, context, expressions, etc.) are related to content of the workflow definition and not jsonschema or YAML specific. The line numbers need to be determine by ourselves in those cases. Since we already know the path from inspection, we need a function that can map dot notation to line number. I did a brief search but can't identify any library that does this. And I already had this chat with @enykeev. He's ok without line numbers. I think he can already resolve dot notation to line number currently in st2flow.

st2client/st2client/commands/workflow.py
m4dcoder: Yeah, this option is limited. This option is helpful for users that are developing directly on a VM with st2 installed.

PR #4322, opened 2018-08-30
Travis: Manually add MongoDB repo

LindsayHill:
Travis have made changes to the APT add-on, and the safe listed APT sources.

https://blog.travis-ci.com/2018-08-29-addressing-reported-mitm-rce

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,12 +62,8 @@ matrix:
 addons:
   apt:
     sources:
-      - sourceline: 'deb [arch=amd64] http://repo.mongodb.org/apt/ubuntu precise/mongodb-org/3.4 multiverse'
-        key_url: 'https://www.mongodb.org/static/pgp/server-3.4.asc'
       - sourceline: 'ppa:git-core/ppa'
     packages:
-      - mongodb-org-server
-      - mongodb-org-shell
       - git
 
 services:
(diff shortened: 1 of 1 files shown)


Kami:
Based on the tests failure it looks like git package hasn't been installed / upgraded so still issues with APT addon.

One of the reasons we use apt addon instead of doing that (we did actually do that in the past) is so it works with container based infrastructure.

So we should check if we can still rely on apt addon, and if not, then I'm fine with this change. And hopefully Travis will eventually support sudo in container infra (as mentioned in the blog post) and then apt addons won't be relevant / needed anymore.

.travis.yml
```diff
@@ -98,6 +88,13 @@ install:
 # Let's enable rabbitmqadmin
 # See https://github.com/messagebus/lapine/wiki/Testing-on-Travis.
 before_script:
+  # key_url no longer works for APT addon
+  # 1. Install MongoDB 3.4 and latest version of git
+  - sudo add-apt-repository ppa:git-core/ppa
+  - curl https://www.mongodb.org/static/pgp/server-3.4.asc | sudo apt-key add -
+  - echo "deb [arch=amd64] http://repo.mongodb.org/apt/ubuntu precise/mongodb-org/3.4 multiverse" | sudo tee -a /etc/apt/sources.list
+  - sudo apt-get update -qq
+  - sudo apt-get install mongodb-org-server mongodb-org-shell git -y
```
armab: Ideally to move this section at early point before-install where also another apt-get update happens

Kami:
And thanks to @LindsayHill for starting this PR :+1:

.travis.yml
```diff
@@ -88,6 +78,12 @@ cache:
 before_install:
   # Work around for Travis timeout issues, see https://github.com/travis-ci/travis-ci/issues/9112
   - sudo apt-get update --option Acquire::Retries=100 --option Acquire::http::Timeout="60"
+  # 1. Install MongoDB 3.4 and latest version of git
+  - sudo add-apt-repository -y ppa:git-core/ppa
+  - curl https://www.mongodb.org/static/pgp/server-3.4.asc | sudo apt-key add -
+  - echo "deb [arch=amd64] http://repo.mongodb.org/apt/ubuntu precise/mongodb-org/3.4 multiverse" | sudo tee -a /etc/apt/sources.list
+  - sudo apt-get update -qq
```
armab: So again, why are we doing 2 `apt-get update`s in `before_install` section?
Could this be merged into one?

Kami: That was an oversight - I will remove the second apt-get update and move first one a couple of lines below.

PR #4323, opened 2018-08-30
Improve orquesta workflow metrics

Kami:
This pull request fixes / improves `st2.orquesta.action.executions` metric so we don't increase the counter and time the operation for non-orquesta workflow executions.

Without that change, we have a lot of noise and outliers for non-orquesta executions which finish immediately since handler function performs a no-op / returns early on non-orquesta execution objects.


st2actions/st2actions/workflows/workflows.py
```diff
@@ -54,6 +54,10 @@ def handle_workflow_execution_with_instrumentation(wf_ex_db):
                 return self.handle_workflow_execution(wf_ex_db=wf_ex_db)
 
         def handle_action_execution_with_instrumentation(ac_ex_db):
+            # Ignore non orquesta workflow executions
+            if 'orquesta' not in ac_ex_db.context:
```
Kami: To clarify, same thing already happens here - https://github.com/StackStorm/st2/pull/4323/files#diff-b528f98088d2afaa8a99f531b71fdd82R164.

I intentionally used a wrapper / decorator function for instrumentation purposes. Adding instrumentation directly to the handler method would result in more complexity and code which is harder to read and manage (whole chunk of code would be indented by one additional level).

st2actions/st2actions/workflows/workflows.py
```diff
@@ -203,6 +211,13 @@ def handle_action_execution(self, ac_ex_db):
         # Process completion of the action execution.
         wf_svc.handle_action_execution_completion(ac_ex_db)
 
+    def _is_orquesta_execution(self, ac_ex_db):
+        """
+        Return True if particular execution object represents an orquesta workflow, False
```
m4dcoder: This description is wrong. When "orquesta" is in the context, this means the action execution belongs to a task in an orquesta workflow. This doesn't mean the action execution is an orquesta workflow.

Kami: Fixed.

Kami:
@m4dcoder 

> To be honest, I'm not sure I like how this with instrumentation is coded here. I may change it after this PR.

Because of the way the code works, we can't directly use decorator for the execution handler because of the miscounts so "wrapper" method is a reasonable approach. "wrapper" method is really a decorator, it just doesn't have full decorator interface.

Another approach would be to add a new and explicit decorator just to be used for `handle_action_execution` which I'm not a fan of. I don't think it's any different / better than existing wrapper approach.

And yet another approach would be moving the context manager inside the method, but I'm not a fan of wrapping already log and nested code blocks with another level of indentation.

m4dcoder:
@Kami I'm going to approve this for now so you can move forward. So don't get offended when I come back and refactor it to use decorators and address the _is_orquesta_execution issue.

PR #4328, opened 2018-09-03
"st2 execution tail" CLI command improvements

### LINKED ISSUE #4267: st2 execution tail CLI command should support nested (workflow) executions
Right now "st2 execution tail" CLI command only supports / works correctly with simple (non-workflow, non-nested) executions.

We should add support for more complex workflow executions to it.

This will require us to implement better filtering and tracking of child executions inside the CLI commands (same thing will need to be done on the WebUI side when adding support for streaming output there).

Kami:
This pull request includes various improvements to the ``st2 execution tail`` CLI command:

1. Support for Orquesta workflows
2. Support for arbitrary nested workflows
3. Correct filtering - previously if user was tailing execution "X" and other executions were also running, data from other executions would incorrectly be included in the output

## TODO

- [x] Tests

Resolves #4267.

### DIFF
--- a/st2client/st2client/commands/action.py
+++ b/st2client/st2client/commands/action.py
@@ -1563,6 +1563,10 @@ class ActionExecutionTailCommand(resource.ResourceCommand):
             result['parent_execution_id'] = context.get('parent', {}).get('execution_id', None)
             result['execution_id'] = event['id']
             result['task_name'] = context.get('mistral', {}).get('task_name', 'unknown')
+        elif 'orquesta' in context:
+            result['parent_execution_id'] = context.get('parent', {}).get('execution_id', None)
+            result['execution_id'] = event['id']
+            result['task_name'] = context.get('orquesta', {}).get('task_name', 'unknown')
         else:
             # Action chain workflow
             result['parent_execution_id'] = context.get('parent', {}).get('execution_id', None)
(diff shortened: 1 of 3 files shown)


st2client/st2client/commands/action.py
```diff
@@ -1505,6 +1520,16 @@ def tail_execution(cls, execution_manager, stream_manager, execution, output_typ
                 # An execution is considered a child execution if it has parent execution id
                 is_child_execution = bool(task_parent_execution_id)
 
+                # Ignore executions which are not part of the execution we are tailing
+                if is_child_execution:
```
Kami: Ideally we could eventually perform this smart "belongs to parent execution X" filtering on the event stream API side.

It would reduce workload and traffic on the client and it could also be used by the WebUI. For now this should work though.

st2client/st2client/commands/action.py
```diff
@@ -1489,13 +1492,27 @@ def tail_execution(cls, execution_manager, stream_manager, execution, output_typ
             print('Execution %s has completed (status=%s).' % (execution_id, execution.status))
             return
 
-        events = ['st2.execution__update', 'st2.execution.output__create']
+        # We keep track of all the workflow executions which could contain children
```
Kami: Not a huge fan of all the code inside this function, but at least we now have test cases for all the various scenarios and edge cases.

Hopefully in the future we can come up with a more simplified code (perhaps by moving some of that filtering to the server as mentioned inside a comment).

PR #4330, opened 2018-09-05
Changelog entry for #4315

Kami:
This pull request adds a changelog entry for #4315.

In addition to that, it defaults ``action_sensor.emit_when`` config option to live action completed states. 

This way the default value is explicit and user doesn't need to dig through the code to figure out what happens if no value is specified.


st2common/st2common/config.py
```diff
@@ -389,9 +390,8 @@ def register_opts(ignore_errors=False):
             'enable', default=True,
             help='Whether to enable or disable the ability to post a trigger on action.'),
         cfg.ListOpt(
-            'emit_when', default=[],
-            help='List of execution statuses that a trigger will be emitted. '
-                 'If not specified, it defaults to action completion.'),
+            'emit_when', default=LIVEACTION_COMPLETED_STATES,
```
shusugmt: Now I understand this is reasonable change. Thanks!

Issue #4030, opened 2018-03-09
st2api being bogged down by MongoEngine


nmaludy:
Working recently with a bunch of large workflows and seeing the `st2api` service using 100% CPU.

I turned on a wsgi profiler (https://medium.com/@maxmaxmaxmax/measuring-performance-of-python-based-apps-using-gunicorn-and-cprofile-ee4027b2be41) and am seeing profile stacktraces like so:

```yamll
2018-03-08 19:59:43,333 100167760 ERROR wsgi_profiler [-] 
[4525] [INFO] [POST] URI /v1/actionexecutions
2018-03-08 19:59:43,333 100167760 ERROR wsgi_profiler [-] [4525] [INFO]          21554881 function calls (20361663 primitive calls) in 60.828 seconds

   Ordered by: internal time, cumulative time
   List reduced from 1272 to 30 due to restriction <30>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
  3888837    4.593    0.000    4.651    0.000 {isinstance}
    45770    3.272    0.000   25.510    0.001 /opt/stackstorm/st2/lib/python2.7/site-packages/mongoengine/base/document.py:36(__init__)
   374817    2.745    0.000    8.031    0.000 /opt/stackstorm/st2/lib/python2.7/site-packages/mongoengine/base/fields.py:123(__set__)
874662/830081    2.720    0.000   12.477    0.000 /opt/stackstorm/st2/lib/python2.7/site-packages/mongoengine/base/document.py:144(__setattr__)
53587/1799    2.446    0.000   11.942    0.007 /opt/stackstorm/st2/lib/python2.7/site-packages/mongoengine/dereference.py:163(_attach_objects)
1452191/1255348    2.137    0.000    6.932    0.000 {getattr}
  1052620    2.072    0.000    4.654    0.000 /opt/stackstorm/st2/lib/python2.7/site-packages/mongoengine/base/datastructures.py:382(__getitem__)
   874913    1.789    0.000    2.500    0.000 /opt/stackstorm/st2/lib/python2.7/site-packages/mongoengine/common.py:5(_import_class)
879819/506309    1.747    0.000   12.198    0.000 {setattr}
   640806    1.711    0.000    5.291    0.000 /opt/stackstorm/st2/lib/python2.7/site-packages/mongoengine/base/datastructures.py:396(get)
  1901626    1.603    0.000    1.603    0.000 {method 'get' of 'dict' objects}
[...]

PR #4331, opened 2018-09-05
Optimize performance of POST /v1/executions API endpoint

Kami:
This pull request optimizes performance of ``POST /v1/executions`` API endpoints by doing various things:

- [x] Removing unnecessary / duplicated queries (from 14 to 3 queries which are all essential now)
- [x] Perform partial document updates instead of updating whole documents where possible
- [x] Skip double validation where it's not needed
- [x] Use ujson round-trip when creating a deep copy of dictionaries we perform mongodb special character escaping on. I've done testing with various dictionary sizes, including actual executions and the speed up can be 10x compared to `copy.deepcopy()`. It has some edge cases with special types, notably `datetime`, but I believe I found a good work-around for that.

We are trading a bit of code readability and maintainability for performance, but that's needed here because the code was / is very wasteful and the performance is really bad. And this bad performance has cascading effect on Mistral and everything else which schedules executions through the API.

Related issue #4030.

### DIFF
--- a/st2common/st2common/services/executions.py
+++ b/st2common/st2common/services/executions.py
@@ -146,8 +146,8 @@ def create_execution_object(liveaction, action_db=None, runnertype_db=None, publ
     execution.web_url = _get_web_url_for_execution(str(execution.id))
 
     # NOTE: User input data is already validate as part of the API request,
-    # other data is set by us. Skipping validation here makes operation 10%-305 faster
-    execution = ActionExecution.add_or_update(execution, publish=publish)
+    # other data is set by us. Skipping validation here makes operation 10%-30% faster
+    execution = ActionExecution.add_or_update(execution, publish=publish, validate=False)
 
     if parent and str(execution.id) not in parent.children:
         values = {}
--- a/st2common/st2common/services/executions.py
+++ b/st2common/st2common/services/executions.py
@@ -144,7 +144,10 @@ def create_execution_object(liveaction, action_db=None, runnertype_db=None, publ
     # TODO: Do 100% research this is fully safe and unique in distributed setups
     execution.id = ObjectId()
     execution.web_url = _get_web_url_for_execution(str(execution.id))
-    execution = ActionExecution.add_or_update(execution, publish=publish)
+
+    # NOTE: User input data is already validate as part of the API request,
+    # other data is set by us. Skipping validation here makes operation 10%-305 faster
+    execution = ActionExecution.add_or_update(execution, publish=publish, validate=False)
 
     if parent and str(execution.id) not in parent.children:
         values = {}
--- a/st2common/st2common/services/executions.py
+++ b/st2common/st2common/services/executions.py
@@ -139,6 +139,7 @@ def create_execution_object(liveaction, action_db=None, runnertype_db=None, publ
 
     attrs['log'] = [_create_execution_log_entry(liveaction['status'])]
 
+    # TODO: This object initialization takes 20-30or so ms
     execution = ActionExecutionDB(**attrs)
     # TODO: Do 100% research this is fully safe and unique in distributed setups
     execution.id = ObjectId()
@@ -154,15 +155,15 @@ def create_execution_object(liveaction, action_db=None, runnertype_db=None, publ
 
 
 def _get_parent_execution(child_liveaction_db):
-    parent_context = child_liveaction_db.context.get('parent', None)
+    parent_execution_id = child_liveaction_db.context.get('parent', {}).get('execution_id', None)
 
-    if parent_context:
-        parent_id = parent_context['execution_id']
+    if parent_execution_id:
         try:
-            return ActionExecution.get_by_id(parent_id)
+            return ActionExecution.get_by_id(parent_execution_id)
         except:
-            LOG.exception('No valid execution object found in db for id: %s' % parent_id)
+            LOG.exception('No valid execution object found in db for id: %s' % parent_execution_id)
             return None
+
     return None
 
 
--- a/st2common/st2common/services/executions.py
+++ b/st2common/st2common/services/executions.py
@@ -99,14 +99,17 @@ def _create_execution_log_entry(status):
     }
 
 
-def create_execution_object(liveaction, publish=True):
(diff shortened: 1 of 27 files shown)


bigmstone:
@Kami I presume we will try to get this in for 2.9?

Kami:
@bigmstone Yep, added v2.9.0 milestone.

Kami:
On a related note - 2.9.0 release will also contain mongoengine upgrade which contains this change - https://github.com/MongoEngine/mongoengine/pull/1630.

This should help a bit with performance when instantiating large mongoengine Document objects (I confirmed it locally).

As discussed on Slack - there is still a larger issue with our reliance on mongoengine and our related message publishing framework in some places. Only way to improve performance in those places will be total refactor all of that code (that will be very hard due to coupling, etc.) and sadly nothing we can do in short term.

st2common/st2common/services/executions.py
```diff
@@ -135,18 +140,15 @@ def create_execution_object(liveaction, publish=True):
...
     execution.web_url = _get_web_url_for_execution(str(execution.id))
     execution = ActionExecution.add_or_update(execution, publish=publish)
 
-    if parent:
-        if str(execution.id) not in parent.children:
-            parent.children.append(str(execution.id))
-            ActionExecution.add_or_update(parent)
+    if parent and str(execution.id) not in parent.children:
+        values = {}
+        values['push__children'] = str(execution.id)
```
Kami: This should be more efficient and unlike previous approach (full document update), it's also atomic.

Kami:
I believe I've reached the end of "simple" optimizations.

I'm not happy because performance is still not great for larger documents. Main issue is related to how our code + mongoengine handle large documents with dictionaries (there is tons of recursive back and forth conversion of data going on, including our escaping of special fields, etc.).

Due to how our code is structured and coupled around mongoengine document abstraction and our publish framework, this will be a very hard change.

We will need to find an incremental approach which will allow us to move away from mongoengine for performance critical code paths (e.g. schedule execution, update execution). This will likely mean using pymongo directly so we can get rid of very wasteful and costly mongoengine data conversions.

As mentioned above, this will be very hard because most of the code relies on our current mongoengine abstractions and we also publish whole MongoDB model instances on the message bus - that's perhaps the biggest offender and think which will be the hardest to change / improve since all the services depend on current "message format"

@m4dcoder This will also negatively affect Orquesta. No matter how fast we make Orquesta, saving executions, etc. through our abstractions will be the bottle neck if we don't do something about it.

@StackStorm/team I will do more research and digging in, but we should schedule a short recap and meeting next week so we can start drawing up long term and short term plans for that because current performance for some operations is simply unacceptable.

st2common/st2common/services/executions.py
```diff
@@ -134,33 +139,34 @@ def create_execution_object(liveaction, publish=True):
...
+    execution.id = ObjectId()
     execution.web_url = _get_web_url_for_execution(str(execution.id))
-    execution = ActionExecution.add_or_update(execution, publish=publish)
 
-    if parent:
-        if str(execution.id) not in parent.children:
-            parent.children.append(str(execution.id))
-            ActionExecution.add_or_update(parent)
+    # NOTE: User input data is already validate as part of the API request,
+    # other data is set by us. Skipping validation here makes operation 10%-305 faster
```
bigmstone: 10%-30% presumably

st2common/st2common/util/action_db.py
```diff
@@ -229,6 +229,8 @@ def update_liveaction_status(status=None, result=None, context=None, end_timesta
     if runner_info:
         liveaction_db.runner_info = runner_info
 
+    # TODO: This is not efficient, perfon direct partial update and only update
```
bigmstone: perform?

st2common/st2common/util/mongoescape.py
```diff
@@ -71,14 +73,20 @@ def _translate_chars(field, translation):
 def escape_chars(field):
     # TODO: This is slow on large dicts, figure out if we can safely manipulate original value
     # instead of creating the copy
-    value = copy.deepcopy(field)
+    if not isinstance(field, dict):
+        return field
+
+    value = ujson.loads(ujson.dumps(field))
```
Kami: It's quite a difference, especially when working with large documents.

```bash
Timer unit: 1e-06 s

Total time: 0.330095 s
File: /data/stanley/st2common/st2common/util/mongoescape.py
Function: escape_chars_1 at line 73

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    73                                           @profile
    74                                           def escape_chars_1(field):
    75                                               # TODO: This is slow on large dicts, figure out if we can safely manipulate original value
    76                                               # instead of creating the copy
    77         4         11.0      2.8      0.0      if not isinstance(field, dict):
    78                                                   return field
    79                                           
    80         4     280657.0  70164.2     85.0      value = copy.deepcopy(field)
    81                                           
    82                                               #value = ujson.loads(ujson.dumps(field))
    83         4      49427.0  12356.8     15.0      return _translate_chars(value, ESCAPE_TRANSLATION)



Timer unit: 1e-06 s

Total time: 0.052098 s
File: /data/stanley/st2common/st2common/util/mongoescape.py
Function: escape_chars_1 at line 73

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
    73                                           @profile
    74                                           def escape_chars_1(field):
    75                                               # TODO: This is slow on large dicts, figure out if we can safely manipulate original value
    76                                               # instead of creating the copy
    77         4          6.0      1.5      0.0      if not isinstance(field, dict):
    78                                                   return field
    79                                           
    80         4       7437.0   1859.2     14.3      value = ujson.loads(ujson.dumps(field))
    81         4      44655.0  11163.8     85.7      return _translate_chars(value, ESCAPE_TRANSLATION)

Total time: 0.052882 s
File: test_escape.py
Function: escape1 at line 132
```


Problem is there are some edge cases, so I'm not sure yet if I can get it to fully work - 5af09e82891aa93d35d584ae988ed9050aba7874, 4917d47b1f7fc181a5d5944286a208330d4a7698, 67886184b511cb11b5f0c4f239be937c21124ef6.

PR #4332, opened 2018-09-06
Use version strings instead of floats in examples

blag:
The spec already accepts both [YAML strings and floats in the `version` key](https://github.com/StackStorm/orquesta/blob/master/orquesta/specs/native/v1/base.py#L21), but it's probably a good idea to use version strings in our examples (which are also used in our documentation) because it's problematic to compare strings and floats in Python:

```python
>>> '1.1' > 1.0  # Fine
True
>>> '0.9' > 1.0
True
>>> '1.0.1' > 2.0
True
>>> '1.1.0' > 2.0
True
```

and we can't use floats everywhere because we can't convert bugfix releases to floats:

```python
>>> float('1.0.1')
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for float(): 1.0.1
```


blag:
@m4dcoder We don't need to change anything in Orquesta - I've confirmed it already accepts version strings as well as unquoted floats (see my link in the summary). This change is simply changing our examples to follow what is probably best practices. And YAML is smart enough to interpret `1.0.1` as a string, so everything works as expected.

I can add a test that specifically tests version strings if you like, but I'm not too worried about this coverage.

m4dcoder:
So if the schema in orquesta supports float and string and there hasn't been any problem with the versioning and executing any orquesta workflow, what's the purpose of this change?

blag:
The thrust of this PR is subtle - it's simply about training users. Versions are fundamentally strings, not floats, and can only be compared as floats when neither version has a bugfix number. So this is nudging our users to use and think about versions as strings. Furthermore, Python can natively correctly compare version strings, but cannot reliably correctly compare version strings to floats.

Our [current code](orquesta/specs/native/v1/base.py) does try to convert the version to a float, which works - for now - because the version can be thought of as a float. And if/when we do ever need to add a bugfix version to our accepted versions, we can adjust the code as necessary. So I don't think there's a need to change any of the Orquesta validation code right now, but I do think it would make our lives slightly easier to demonstrate versions as strings in our example workflow YAML.

And just to be clear: this isn't a huge deal if it doesn't get merged, it's just a tiny amount of easy polish.

blag:
@m4dcoder I fixed everything you brought up in your review. Please re-review at some point.

PR #4333, opened 2018-09-07
Port YAQL/Jinja functions to orquesta workflows

m4dcoder:
Port the Jinja filters in st2common to orquesta. Instead of making another copy of the functions like st2mistral, reuse the existing set of functions. The json, yaml, regex, path, time, and version related functions are setup and configured. A new task function is implemented that is similar to the task function in mistral. Unit and integration tests are covered to ensure the functions are setup and configured properly.


nmaludy:
@m4dcoder Looking at this PR it appears to solve a major issue we had with Mistral in that we had to duplicate all of our expression function code over into the Mistral code base. Correct me if i'm wrong, but it appears that we can simply implement a new expression function in `st2common.expressions.functions` and add it to the list in `contrib/runners/orquesta_runner/setup.py`?


Also, dumb question... but could we use something like the stevedore to automatically load all of the expressions at runtime instead of needing to maintain a list in `contrib/runners/orquesta_runner/setup.py`?

m4dcoder:
@nmaludy My goal for this PR is to make sure existing functions are back ported for orquesta. As for dynamic loading of functions, I'm not ready to tackle this problem yet because this means allowing users to write their own functions. This involves other topics to discuss such as location, part of pack, etc.

requirements.txt
```diff
@@ -8,7 +8,7 @@ eventlet==0.23.0
 flex==6.13.2
 git+https://github.com/Kami/logshipper.git@stackstorm_patched#egg=logshipper
 git+https://github.com/StackStorm/mongoengine.git@stackstorm_patched#egg=mongoengine
-git+https://github.com/StackStorm/orquesta.git@master#egg=orquesta
+git+https://github.com/StackStorm/orquesta.git@9cba10037dcf66dd44778d9ccb769039edc81237#egg=orquesta
```
bigmstone: Is this intentional? Can we use a release tag?

m4dcoder: @bigmstone it's intentional per @Kami's request. if we don't add/change the revision, orquesta won't get updated in circleci.  we'll do a release for orquesta and use the release tag when we cut the release branch for st2.

tox.ini
```diff
@@ -55,7 +55,7 @@ commands =
...
+setenv = PYTHONPATH = {toxinidir}/external:{toxinidir}/st2common:{toxinidir}/st2auth:{toxinidir}/st2api:{toxinidir}/st2actions:{toxinidir}/st2exporter:{toxinidir}/st2reactor:{toxinidir}/st2tests:{toxinidir}/contrib/runners/action_chain_runner:{toxinidir}/contrib/runners/local_runner:{toxinidir}/contrib/runners/windows_runner:{toxinidir}/contrib/runners/python_runner:{toxinidir}/contrib/runners/http_runner:{toxinidir}/contrib/runners/noop_runner:{toxinidir}/contrib/runners/announcement_runner:{toxinidir}/contrib/runners/remote_runner:{toxinidir}/contrib/runners/remote_runner:{toxinidir}/contrib/runners/mistral_v2:{toxinidir}/contrib/runners/orquesta_runner:{toxinidir}/contrib/runners/inquirer_runner:{toxinidir}/contrib/runners/http_runner:{toxinidir}/contrib/runners/cloudslang_runner:{toxinidir}/contrib/runners/winrm_runner
```
bigmstone: I don't love this solution. Requires us to append/remove for ever runner change. I suppose there's a reason we don't `make requirements` for py3 tests. Perhaps @Kami can provide some insight there, but I'd prefer this loop over runners and run setup so we don't have to keep it in sync.

m4dcoder: This line change isn't the fix.  I had to loop thru the setup in scripts/travis/install-requirements.sh to install the functions properly.  I change the PYTHONPATH here at first because it was wrong for orquesta runner.  Yeah, @Kami need to provide what/why py3 is setup this way.

m4dcoder:
@bigmstone @Kami I'm still getting errors on py27 and py36 that is mongo related, saying `no query solutions` on unrelated tests.  I'm fixing indexing on the workflow model at https://github.com/StackStorm/st2/pull/4339.  but I'm not sure if it's related and I tried to fix it in this PR but it didn't really fix it in travis.  So if you guys can help take a look, I really appreciate it.

contrib/runners/orquesta_runner/setup.py
```diff
@@ -53,6 +53,34 @@
...
+            'version_more_than = st2common.expressions.functions.version:version_more_than',
+            'version_less_than = st2common.expressions.functions.version:version_less_than',
+            'version_equal = st2common.expressions.functions.version:version_equal',
+            'version_match = st2common.expressions.functions.version:version_match',
+            'version_bump_major = st2common.expressions.functions.version:version_bump_major',
+            'version_bump_minor = st2common.expressions.functions.version:version_bump_minor',
+            'version_bump_patch = st2common.expressions.functions.version:version_bump_patch',
+            'version_strip_patch = st2common.expressions.functions.version:version_strip_patch',
+            'yaml_dump = st2common.expressions.functions.data:to_yaml_string',
+            'yaml_parse = st2common.expressions.functions.data:from_yaml_string'
```
Kami: We should figure out if we can some how speed this up - st2api and other services already take a while to start up / initialize and a chunk of that time appears to be this dynamic entry point loading and parsing :/

Kami:
I can't reproduce the test issue locally and I use the same mongod.conf as Travis when running test isolation.

It looks like the issue might be related to some cross-test pollution or similar (some other test dropping the indexes / similar) because it only happens when running tests in bulk on Travis.

Kami:
@m4dcoder it looks like https://github.com/StackStorm/st2/pull/4333/commits/83f5f08eeda5067ce7420e9c6be37e275d5fb428 helped - https://travis-ci.org/StackStorm/st2/builds/427584853.

It looks like indeed some kind of cross test pollution issue since the issue doesn't manifest itself when running test case in isolation (aka by itself).

contrib/runners/orquesta_runner/setup.py
m4dcoder: Acknowledge. My plan is to look for alternative way to dynamically load these functions instead of using entry points. This will be post v2.9 though.

m4dcoder:
@Kami thank you for the fix, bro. I owe you one.

PR #4335, opened 2018-09-11
NGINX: Use 308 redirect rather than 301

LindsayHill:
Our default NGINX configuration redirects `http://` requests to `https://`. This uses a 301 redirect. When clients see this, they will either switch PUT/POST requests to GET, or they may continue with PUT/POST, but with no payload.

This change uses [HTTP 308](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/308) instead. The request method and body will be unchanged.

This will reduce confusion for people using custom code that targets `http://<ST2>/`, rather than `https://<ST2>`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -107,6 +107,8 @@ Changed
 * Improve ``st2 execution tail`` CLI command so it also supports Orquesta workflows and arbitrarily
   nested workflows. Also fix the command so it doesn't include data from other unrelated running
   executions. (improvement) #4328
+* Change default NGINX configuration to use HTTP 308 redirect, rather than 301, for plaintext requests.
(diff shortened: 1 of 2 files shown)


LindsayHill:
@armab would appreciate your input on this. No rush, does not need to be 2.9, could be 3.0.

armab:
According to https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.3.2 the issue with `301` should be something rarely seen nowadays:

>       Note: When automatically redirecting a POST request after
>      receiving a 301 status code, some existing HTTP/1.0 user agents
>      will erroneously change it into a GET request.


But still, good to have this improvement to be on the safe side :+1:

PR #4337, opened 2018-09-11
Various v2.9.0 changes

Kami:
Trying to get https://github.com/StackStorm/st2web/pull/605 finished so we can include at least one directly user-visible performance improvement.

## Changes

1. Some improvements in the `GET /v1/executions` API endpoint - sometimes `parameters` field can be None
2. `GET/v1/actions/overview` API endpoint - performance improvements
   1. Remove log statement which logged under INFO every action object which caused unnecessary and massive performance overhead
   2. Reduce number of queries from `N * 2` to `2`


st2common/st2common/models/db/execution.py
```diff
@@ -110,7 +110,7 @@ def mask_secrets(self, value):
         parameters.update(value.get('runner', {}).get('runner_parameters', {}))
 
         secret_parameters = get_secret_parameters(parameters=parameters)
-        result['parameters'] = mask_secret_parameters(parameters=result['parameters'],
+        result['parameters'] = mask_secret_parameters(parameters=result.get('parameters', {}),
```
Kami: Noticed this issue while testing WebUI changes.

st2api/st2api/controllers/v1/action_views.py
```diff
@@ -88,7 +88,6 @@ def _get_one(action_id, requester_user):
                 GET /actions/views/parameters/1
         """
         action_db = LookupUtils._get_action_by_id(action_id)
-        LOG.info('Found action: %s, runner: %s', action_db, action_db.runner_type['name'])
```
Kami: This reduces duration of this request for 10 seconds where there are many actions in the database.

st2api/st2api/controllers/v1/action_views.py
```diff
@@ -144,6 +143,9 @@ def get_all(self, exclude_attributes=None, include_attributes=None, sort=None, o
         result = []
         for item in resp.json:
             action_api = ActionAPI(**item)
+            # TODO: This is extreme inefficient and results in N * 2 queries where N is number of
```
Kami: This is simply horrible, with aws and some other packs API endpoint takes 30+ seconds to return (well, now 10 seconds less with logging fix, but still too long).

Kami:
Changes in bcb63f58cb9530c5ce4b73203992d19e4f8cb222 and 7d070d81c6f29d7b6952ba1c6f7c4cb06b9ab6a1 substantially improve performance of `GET /v1/actions/views/overview` API endpoint.

It changes the code to perform 1 database query instead of `n * 2` where `n` is number of actions in the database. This + the logging change now makes this API endpoint return in ~6-7 seconds instead of ~35 seconds on installations with a lot of actions.

Having said that, 6 is still too long, imo.

The main problem is that in the new WebUI "Rules" tab / app now depends on this API endpoint.

@enykeev Can we some how get rid of request to this API endpoint all together in the new rules tab? Or can we replace it with a call to some other (faster) API endpoint?

Kami:
Adding to my comment above - another option to provide a better WebUI rules tab user experience would be delay loading action parameters and display rules immediately on the page once rules are loaded and full in action parameters loaded once action data is loaded from that API endpoint.

This would provide a much better user-experience - current approach with page being "white" aka not-rendered / empty until actions overview API endpoint loads is very not friendly for installations with a lot of actions.

/cc @enykeev

PR #4340, opened 2018-09-12
Add log entry in workflow result on action execution failure

m4dcoder:
Add unit test to ensure that the error and result of an action execution failure for a task is logged in the orquesta workflow result.

### DIFF
--- a/st2common/in-requirements.txt
+++ b/st2common/in-requirements.txt
@@ -12,7 +12,7 @@ kombu
 # See https://github.com/MongoEngine/mongoengine/pull/1833
 git+https://github.com/StackStorm/mongoengine.git@stackstorm_patched#egg=mongoengine
 networkx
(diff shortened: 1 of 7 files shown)


st2common/in-requirements.txt
```diff
@@ -12,7 +12,7 @@ kombu
 # See https://github.com/MongoEngine/mongoengine/pull/1833
 git+https://github.com/StackStorm/mongoengine.git@stackstorm_patched#egg=mongoengine
 networkx
-git+https://github.com/StackStorm/orquesta.git@41692880f1733399ab36cf265857e00d0c7eec2c#egg=orquesta
+git+https://github.com/StackStorm/orquesta.git@99ef681b500791ebe8945abab53ddba834c4559d#egg=orquesta
```
Kami: Thanks for pinning it to a specific revision each time it's updated :+1: 

This should help with CI and local failures which occur due to caching and dependency not being updated.

PR #4353, opened 2018-09-21
[WIP] Fix sensor issues when pack is using Python 3 virtualenv but installation is using Python 2

Kami:
This pull request fixes an issue with sensors not running when a particular pack is using Python 3 virtual environment, but installation is running under Python 2.

(Note: This was a stop gap solution and in the future, we plan support full Python 3 deployments because mixing two versions is quite fragile and has a lot of edge cases)


Kami:
After some more though - better approach might simply be to throw in such scenario - we should only support running actions but not sensors in mixed environments (installation uses Python 2, pack virtualenv uses Python 3).

Getting sensors working is a lot more complicated and requires a lot more glue code. I will open a new PR which does that.

Kami:
Superseded by https://github.com/StackStorm/st2/pull/4354.

PR #4357, opened 2018-09-22
Fixing openapi definition

bigmstone:
Duplicate entry prevents openapi docs generation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -192,7 +192,7 @@ lint-api-spec: requirements .lint-api-spec
 	@echo
 	@echo "================== Lint API spec ===================="
 	@echo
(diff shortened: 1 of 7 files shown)


Kami:
Danke :+1:

Kami:
Perhaps we should also add a lint check / update `.generate-api-spec` Make task to throw on such issues?

Kami:
Pushed a change 76462afdf2bf9f21780bde39687dfaf4bf80f6a8 which updates spec loader and `generate-api-spec` make task to throw if input YAML file contains duplicated key.

Per YAML v1.2 spec, duplicated keys are not valid and this will allow us to detect such issues earlier (during st2 repo build).

PR #4369, opened 2018-10-02
Fix race when parallel inquiries set the execution status

m4dcoder:
When a workflow is paused and there are multiple requests to resume the workflow, there is a race and one of the request fails because the workflow is already in an active state. Instead of erroring, log the situation and there is no need to resume since the workflow is already active.


Kami:
Good catch :+1: 

Will also cherry pick it into v2.9 branch for inclusion in v2.9.1.

PR #4373, opened 2018-10-02
Added Mongo X509 auth support

aduca85:
Changed some of the bootstrap scripts to include the authentication_mechanism parameter used by pymongo to authenticate. By using authentication_mechanism = 'MONGODB-X509' in st2.conf you can now use X509 certificates to login to Mongo.


Kami:
Thank you for your contribution.

Change looks good to me, but it would also be good to add some test cases for it (I believe we already have some tests for MongoDB connection constructor).

Kami:

I fixed lint, re-generated sample config and added tests in #4391. That PR should get merged some time this week.

PR #4378, opened 2018-10-05
core.http - allow all basic http methods

emptywee:
Per https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol#Request_methods - allow all basic HTTP methods


emptywee:
Oops wrong branch.

PR #4379, opened 2018-10-05
Allow all basic HTTP methods in core.http

emptywee:
Per https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol#Request_methods - allow all methods listed, except CONNECT.


emptywee:
PURGE method is used by nginx api to clear cached content from nginx servers when it's used as a caching load balancer/reverse proxy.

Kami:
Thanks :+1: 

LGTM - can you please add  a changelog entry?

emptywee:
Of course. Though I am not sure under which section it should go: added/changed or fixed? :) I placed it under "changed" for now. Let me know if it's better to move it elsewhere.

Kami:
@emptywee Changed should be fine :)

PR #4380, opened 2018-10-08
Fix warning about invalid format in pip list

nmaludy:
In recent versions of `pip`, the `legacy` format has been removed causing the following warning when running `st2-run-pack-tests`:

```
Installing dependencies from st2 repository...

Usage:   
  pip list [options]

option --format: invalid choice: 'legacy' (choose from 'columns', 'freeze', 'json')
Installing global pack test dependencies...
Installing pack-specific dependencies...
Installing pack-specific test dependencies..
```

This PR fixes the issue by using the `freeze` format instead of `legacy`.


Kami:
Is ``freeze`` also supported in older versions of pip where we used ``legacy``?

Kami:
@laixintao Thanks for confirming.

I need to wait with merging this PR (and other PRs) a bit until we sort out our CI issue - we spun up a new instance, but webhooks were not set up correctly so end to end tests are not running on Github commit / push.

PR #4387, opened 2018-10-11
Update paramiko from vulnerable version

bigmstone:
Current paramiko version has known vulnerability. This PR upgrades to non-vulnerable version.


Kami:
Can you please also add a changelog entry.

PR #4392, opened 2018-10-16
Add initial list of StackStorm Adopters (part 1/2)

armab:
The idea is to have a dedicated `ADOPTERS.md` document which will list bigger orgs using StackStorm.

1) First phase is to gather a list of Adopters based on already existing StackStorm Thought Leaders https://stackstorm.com/stackstorm-thought-leaders/ with some minimal description about every use case.

2) Second phase is to iterate by adding big adopters to the list as we find them as well as talk with old ones. Probably needs some prior communication with every org about their StackStorm use.

Ex:
- https://github.com/coredns/coredns/blob/master/ADOPTERS.md
- https://github.com/linkerd/linkerd/blob/master/ADOPTERS.md
- https://github.com/fluent/fluentd/blob/master/ADOPTERS.md
- https://github.com/rook/rook/blob/master/ADOPTERS.md

This is something that's recommended by the "guidelines" and won't hurt us anyways.
@dzimine @lakshmi-kannan @LindsayHill I'd like you to pay attention to this PR as someone like you will be able to maintain this list best.


armab:
Thanks for review and edits @LindsayHill!

PR #4396, opened 2018-10-17
Use ?include_attributes API filter for CLI list and get operations

Kami:
This pull request updates ``st2 execution {list,get}`` command to use new ``?include_attributes`` API query param filter and only retrieve attributes which are displayed / needed by the CLI.

This will speed up the CLI in a similar manner as it speed up the WebUI.

Previously we used ``?exclude_attributes`` which didn't give us as fain grain control and we will also deprecate it in favor of include attributes in the future.

## TODO

- [x] Update other most commonly used CLI command to utilize this API query param filter
  - [x] st2 action list
  - [x] st2 rule list
  - [x] st2 trigger-instance list
  - [x] st2 sensor list


st2client/tests/unit/test_commands.py
```diff
@@ -246,34 +246,51 @@ def test_command_delete_failed(self):
...
+        result = ResourceViewCommand._get_include_attributes(args=args)
+        self.assertEqual(set(result), set(['id', 'status']))
+
+        args = cls(attr=['trigger_instance'])
+        result = ResourceViewCommand._get_include_attributes(args=args)
+        self.assertEqual(set(result), set(['trigger_instance']))
+
+        args = cls(attr=['all'])
+        result = ResourceViewCommand._get_include_attributes(args=args)
+        self.assertEqual(result, None)
```
lakshmi-kannan: Can  you add a test case for a field that doesn't exist in the dict? Also a test case for some existing and some not existing in the same call?

st2client/st2client/utils/types.py
```diff
@@ -0,0 +1,84 @@
...
+
+from __future__ import absolute_import
+import collections
+
+__all__ = [
+    'OrderedSet'
+]
+
+
+class OrderedSet(collections.MutableSet):
```
lakshmi-kannan: Wish there were associated tests for this.

Kami: Will add them.

In fact, that's copied from st2common (because of our unresolved dependencies between st2common and st2client).

st2client/tests/unit/test_commands.py
Kami: That would be an integration / end to end test.

Will add it to st2tests.

But yeah, in case user specifies an nonexistent field, API will return 400 error and that will be propagated to the CLI (we already have tests for that for st2api).

Kami: @lakshmi-kannan Added end to end tests in - https://github.com/StackStorm/st2tests/pull/146.

st2client/st2client/utils/types.py
Kami: Here we go - a387c03.

PR #4397, opened 2018-10-17
Add delay for task scheduling

bigmstone:
On execution creation there should be a way to delay the task by a specified amount of time. This PR provides a mechanism to specify in the the delay in milliseconds.


blag:
Closing and reopening to kick CI.

st2common/st2common/util/execution_queue_db.py
```diff
@@ -0,0 +1,94 @@
...
+        item from the queue and pop it off.
+    """
+    query = {
+        "start_timestamp__lte": date.get_datetime_utc_now(),
+        "order_by": [
+            "-priority",
+            "start_timestamp",
+            "delay",
+        ]
+    }
```
m4dcoder: This query doesn't seem to calculate if a delay wait is over.

st2common/st2common/util/execution_queue_db.py
```diff
@@ -0,0 +1,94 @@
...
+        delay or 0
+    )
+    execution_request.delay = delay
+    execution_request.priority = priority
+    execution_request.affinity = affinity
+
+    return execution_request
+
+
+@coordination.Locker('st2schedulerqueue')
```
m4dcoder: I prefer something like `@coordination.lock('st2schedulerqueue')`. The literal definition of `Locker` is not correct here.

st2common/st2common/util/execution_queue_db.py
bigmstone: I need to take delay out of that query. It was way more efficient to store the query start time as live action start + delay since I can make the one query based on anything with start timer <= current time.

st2actions/st2actions/scheduler/queue_handler.py
```diff
@@ -23,42 +23,42 @@
...
+
+    def loop(self):
+        LOG.debug('Entering scheduler loop')
+        while self._shutdown is not True:
+            # TODO: Make config setting
+            eventlet.greenthread.sleep(0.25)
+            try:
+                execution = exdb.pop_next_execution()
+                if execution:
+                    eventlet.spawn(self._handle_execution, execution)
```
Kami: It looks like you already call `loop` in a greenthread so there is probably no need to also use it here.

st2actions/st2actions/scheduler/queuer.py
```diff
@@ -0,0 +1,76 @@
...
+        execution_request = exdb.create_execution_request_from_liveaction(
+            liveaction_db,
+            delay=liveaction_db.delay
+        )
+        ExecutionQueue.add_or_update(execution_request)
+
+
+def get_queuer():
+    with Connection(transport_utils.get_messaging_urls()) as conn:
+        return ActionExecutionQueuer(conn, [ACTIONSCHEDULER_REQUEST_QUEUE])
```
Kami: Do we still have scheduler listening on this queue?

If we do, there is a "race".

Well, not really a race per say since we use round-robin message distribution, but it means that behavior will be undefined and unpredictable because messages will be round robined between those two services.

So we probably need to use a different queue (so queuer and and scheduler both get a copy of the message) and also use locking so they don't both try to process it at the same time - it looks like delay stuff should be processed first, similar as we do with policies.

st2actions/st2actions/scheduler/queuer.py
```diff
@@ -0,0 +1,76 @@
...
+    'ActionExecutionQueuer',
+    'get_queuer'
+]
+
+
+LOG = logging.getLogger(__name__)
+
+
+class ActionExecutionQueuer(consumers.MessageHandler):
+    message_type = LiveActionDB
```
Kami: Let's please add a docstring which clarifies what that service / handler does.

I'm also still thinking about a better name - "queuer" is a bit vague and also hard to type.

Maybe "ActionExecutionDelayHandler" or something along those lines to make it more obvious what it's purpose is just by looking at the name.

st2common/st2common/models/db/execution_queue.py
```diff
@@ -0,0 +1,56 @@
...
+
+__all__ = [
+    'ExecutionQueueDB',
+]
+
+
+LOG = logging.getLogger(__name__)
+
+
+class ExecutionQueueDB(stormbase.StormFoundationDB):
```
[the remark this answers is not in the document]
Kami: +1

st2common/st2common/models/db/execution_queue.py
```diff
@@ -0,0 +1,56 @@
...
+    RESOURCE_TYPE = ResourceType.EXECUTION_REQUEST
+    UID_FIELDS = ['id']
+    liveaction = stormbase.EscapedDictField(required=True)
```
Kami: Do we need access to the whole live action object or we can just store a reference?

I guess storing the whole object shouldn't be a big deal since this object shouldn't contain `result` attribute yet and potentially only `parameters` field could contain a lot of data.

st2common/st2common/util/execution_queue_db.py
```diff
@@ -0,0 +1,94 @@
...
+            "-priority",
+            "start_timestamp",
+            "delay",
+        ]
+    }
+
+    next_execution = ExecutionQueue.query(**query)
+    if next_execution:
+        next_execution = next_execution[0]
+        ExecutionQueue.delete(next_execution)
```
Kami: How do we handle "transactions" here - aka ideally execution would only be deleted after it's processed.

Otherwise, if the code which processes it dies / similar during the processing (or just after popping it and before changing the status) it would mean we would lose that execution, right?

Kami:
On a related note - we definitely need some integration and end to end tests for that (it won't be the must fun thing to write, but still).

Kami:
On a relate note - once this feature is implemented, we should update / improve ``retry`` policy to utilize this status and functionality to make more more robust and restart safe (see #3630).

st2common/st2common/util/execution_queue_db.py
Kami: I agree - storing absolute time is better, imo.

This way we can utilize it for any executions which need to be delayed and executed at some later point (also for retry policy - https://github.com/StackStorm/st2/pull/4397#issuecomment-432176089, etc.)

st2common/st2common/util/execution_queue_db.py
```diff
@@ -0,0 +1,94 @@
...
+        "order_by": [
+            "-priority",
+            "start_timestamp",
+            "delay",
+        ]
+    }
+
+    next_execution = ExecutionQueue.query(**query)
+    if next_execution:
+        next_execution = next_execution[0]
```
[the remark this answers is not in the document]
Kami: Or we could also simply return multiple executions which are to be run and queurer (better name TBD) handles that scenario (e.g. to avoid busy waiting we add short sleep / delay between each scheduling or similar).

st2common/st2common/util/execution_queue_db.py
m4dcoder: Yeah. My point here is that this query as it stands right now is not actually comparing current time, start time, and delay. As far as I can tell, the query find any records older then current time and pick the first one.  Where's the delay calculation?

st2common/st2common/util/execution_queue_db.py
m4dcoder: @Kami is right.

st2common/st2common/util/execution_queue_db.py
bigmstone: The only thing that needs to happen here is removal of delay, as I said. This query will sort by priority first, then by timestamp. So any available executions will be executed first by priority then by first added. Sorting by delay was left over from my originally trying to keep start_timestamp the same as the liveaction, but I abandoned that because it's impossible to query for that efficiently.

st2actions/st2actions/scheduler/queuer.py
bigmstone: Good catch

bigmstone: Technically it wouldn't be a "problem" since scheduler ignores things that are outside of delayed/requested states. It's just super inefficient so should be taken care of.

st2common/st2common/util/execution_queue_db.py
```diff
@@ -0,0 +1,94 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
```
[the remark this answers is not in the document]
bigmstone: I'm fine with moving it, I'm just following the established pattern currently there. Admittedly the pop function goes a bit above the others, but the others don't.

st2common/st2common/models/db/execution_queue.py
```diff
@@ -0,0 +1,56 @@
...
+    start_timestamp = ComplexDateTimeField(
+        default=date_utils.get_datetime_utc_now,
+        help_text='The timestamp when the liveaction was created.')
+    delay = me.IntField()
+    priority = me.IntField()
+    affinity = me.StringField()
```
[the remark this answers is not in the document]
bigmstone: Priority will be (is, just needs API) working in this PR, but I'll remove affinity.

st2actions/st2actions/scheduler/queuer.py
bigmstone: It's not a delay handler though. All executions funnel through it, delayed or not. Perhaps `ActionExecutionSchedulerEntrypoint` or something. I'm really struggling coming up with a better name. 😂

st2common/st2common/models/db/execution_queue.py
bigmstone: Ay, will move to just live action ID.

st2common/st2common/util/execution_queue_db.py
bigmstone: Yep, this is true of the entries into this collection as well. I briefly mentioned this in slack this morning. Easiest is to delete it after scheduling and on entry change status only after put into the collection since the scheduler will ignore things that are not in `delayed` or `requested` state. The only real way to mitigate orphaned  LiveActions is to periodically ensure all `requested` LiveActions exist in the execution queue collection.

st2common/st2common/util/execution_queue_db.py
m4dcoder: Yeah, this is kinda confusing at first glance. I see you calculated the delay into the start time when you created this record.  To make this clear, I would rename `start_timestamp` to `scheduled_start_timestamp` so not to be confused with actual start timestamp of the live action and you already say you will rename `delay` from the table and from the query.

st2common/st2common/models/db/execution_queue.py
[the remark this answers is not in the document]
m4dcoder: Also, when I run a new action execution but I didn't specify a priority, what's the default priority?  Can I assume st2 will run the action execution immediately, which means the priority is high?  Also having priority means there is a finite capacity. At what point will st2 start putting action execution into the scheduling queue here? What if there's a constant stream of action execution with default priority but not enough capacity?  When will lower priority execution in the scheduling queue be run?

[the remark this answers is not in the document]
m4dcoder: Having said that, the current scope of the PR is "delay".  Delay is straightforward problem to solve. I'm hesitant adding "priority" into this PR here without a good understanding and solution for the problem.

st2common/st2common/util/execution_queue_db.py
bigmstone: Yeah, I agree. Am planning to change it. Left over artifact from how I was calculating it prior.

st2common/st2common/models/db/execution_queue.py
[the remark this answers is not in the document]
bigmstone: I'll leave it out to have the discussion in another PR, but these have pretty well-established answers. At least in situations I've worked in with the past. You just select your finite number of priorities and have a probability distribution from which priority to choose from on scheduling. I'm sure there are other ways, but I can't think of another that wouldn't require shared state. As said though I'll leave it out for now.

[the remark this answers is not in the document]
m4dcoder: Thanks! I think this will help progress this PR quicker. The priority problem is still yours. Just need to explain it a bit in more details to us what these well established answers are and how it will be implemented in st2.

st2actions/st2actions/policies/concurrency.py
```diff
@@ -54,7 +54,7 @@ def _apply_before(self, target):
             LOG.debug('There are %s instances of %s in scheduled or running status. '
                       'Threshold of %s is not reached. Action execution will be scheduled.',
                       count, target.action, self._policy_ref)
-            status = action_constants.LIVEACTION_STATUS_SCHEDULED
+            status = action_constants.LIVEACTION_STATUS_REQUESTED
```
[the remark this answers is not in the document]
bigmstone: At this point it is not scheduled, as it would have been previously. The scheduler will move it into scheduled state if the rest of its checks pass. Eventually priority, delay, etc.

st2actions/st2actions/policies/concurrency.py
```diff
@@ -69,14 +69,22 @@ def _apply_before(self, target):
...
         target = super(ConcurrencyApplicator, self).apply_before(target=target)
 
-        # Exit if target not in schedulable state.
-        if target.status != action_constants.LIVEACTION_STATUS_REQUESTED:
-            LOG.debug('The live action is not schedulable therefore the policy '
+        valid_states = [
+            action_constants.LIVEACTION_STATUS_REQUESTED,
+            action_constants.LIVEACTION_STATUS_DELAYED,
+            action_constants.LIVEACTION_STATUS_POLICY_DELAYED,
+        ]
```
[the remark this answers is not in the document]
bigmstone: Policy delay happens when a policy delays it. Delay happens if there's an explicit delay. This is a function of how policies are applied, manipulating the liveaction directly instead of returning the desired status.

st2actions/st2actions/scheduler/entrypoint.py
```diff
@@ -61,28 +78,20 @@ def process(self, request):
...
-
-        # Update liveaction status to "scheduled".
-        if liveaction_db.status == action_constants.LIVEACTION_STATUS_REQUESTED:
+        if liveaction_db.delay > 0:
             liveaction_db = action_service.update_status(
-                liveaction_db, action_constants.LIVEACTION_STATUS_SCHEDULED, publish=False)
+                liveaction_db,
+                action_constants.LIVEACTION_STATUS_DELAYED,
+                publish=False
+            )
```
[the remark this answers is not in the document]
bigmstone: Looking at the rest of the entrypoint code, it will add the liveaction to the scheduling collection. It is not scheduled at this point. Only processed to await scheduling by the handler. If there is a delay it will add the delay to the scheduled start time and set the liveaction to `delayed` until the scheduler updates it to `scheduled`

st2actions/st2actions/policies/concurrency.py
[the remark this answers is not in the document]
m4dcoder: Can you point me to where the scheduler move it into scheduled state?  I don't seem to find the change in the PR.

st2actions/st2actions/policies/concurrency.py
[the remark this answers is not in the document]
m4dcoder: I mean add more comments in the code.  Provide description in the comments how/source of the LiveAction reaching the scheduler in specific status.

st2actions/st2actions/scheduler/entrypoint.py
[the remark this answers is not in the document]
m4dcoder: Yeah, I'm not finding the line where the liveaction is set to scheduled.

st2actions/st2actions/policies/concurrency.py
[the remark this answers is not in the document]
bigmstone: I didn't make the change. It's current functionality: https://github.com/StackStorm/st2/blob/87d991e5fc2cf383a25c89a6a563161bd86dd284/st2actions/st2actions/policies/concurrency.py#L57

st2actions/st2actions/scheduler/entrypoint.py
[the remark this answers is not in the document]
bigmstone: It happens in the pre_run_policies see: https://github.com/StackStorm/st2/pull/4397#discussion_r232081923

st2common/st2common/persistence/execution_queue.py
```diff
@@ -0,0 +1,27 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import absolute_import
+from st2common.models.db.execution_queue import EXECUTION_QUEUE_ACCESS
+from st2common.persistence import base as persistence
+
+
+class ExecutionQueue(persistence.Access):
```
m4dcoder: Please rename to ActionExecutionSchedulingQueue to be consistent with the DB model.

st2actions/st2actions/scheduler/entrypoint.py
```diff
@@ -61,28 +78,27 @@ def process(self, request):
...
 
-        # Publish the "scheduled" status here manually. Otherwise, there could be a
-        # race condition with the update of the action_execution_db if the execution
-        # of the liveaction completes first.
-        LiveAction.publish_status(liveaction_db)
+        execution_request = _create_execution_request_from_liveaction(
+            liveaction_db,
+            delay=liveaction_db.delay
+        )
+        ExecutionQueue.add_or_update(execution_request, publish=False)
```
m4dcoder: I totally miss this because I thought this was indented under `if liveaction_db.delay > 0`.  So does this means that every single action execution will be queued into MongoDB?  I thought we agree only to push it into the queue if there's a delay. If there's no delay, then just set it to `scheduled` and pass it to the policy engine.

bigmstone: No, that's was never the case. Nor do I really think it should be. Given where we want the scheduler to go all executions should go into queue to be scheduled.

m4dcoder: So this was my original thought on the subject but then @lakshmi-kannan disagrees.  https://github.com/StackStorm/discussions/issues/316

Then you, @Kami, and I talked over this in a design discussion. And I thought we agree to push only executions with special scheduling conditions into the table.  I thought this was an appropriate compromise.  Maybe we didn't clarify.  https://stackstorm.slack.com/archives/C81BGM60H/p1539793063000200

Anyway, I'm cool with pushing all the executions to a table because that was my original thought. But I prefer this gets a clear from @Kami and @lakshmi-kannan.  Also, if this is the case, we need to process the scheduling in bulk otherwise, the performance and latency will hit us.

bigmstone: Current implementation schedules in bulk any available executions ready for scheduling on every round of the loop.

st2actions/st2actions/scheduler/handler.py
```diff
@@ -0,0 +1,163 @@
...
+
+class ExecutionQueueHandler(object):
+    def __init__(self):
+        self.message_type = LiveActionDB
+        self._shutdown = False
+
+    def loop(self):
+        LOG.debug('Entering scheduler loop')
+        while self._shutdown is not True:
+            eventlet.greenthread.sleep(0.25)
```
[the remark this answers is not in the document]
m4dcoder: Since all executions is pushed to the table, I would decrease the value to 0.1 here.  We don't want to hit users with unnecessary latency when there is no delays defined.

Makefile
```diff
@@ -53,7 +53,7 @@ ifndef PYLINT_CONCURRENCY
 	PYLINT_CONCURRENCY := 1
 endif
 
-NOSE_OPTS := --rednose --immediate --with-parallel
+NOSE_OPTS := --rednose --immediate --with-parallel -x
```
Kami: What does the -x flag do?

st2actions/st2actions/cmd/scheduler.py
```diff
@@ -8,7 +8,10 @@
 import signal
 import sys
 
-from st2actions.scheduler import scheduler
+from st2actions.scheduler import (
+    handler as scheduler_handler,
+    entrypoint as scheduler_entrypoint
+)
```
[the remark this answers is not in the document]
Kami: Yeah, let's please use a consistent import style - I use one style (which was used since the beginning), then @m4dcoder started to use another and now yet another :P

st2actions/st2actions/scheduler/handler.py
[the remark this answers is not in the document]
Kami: Agreed with @m4dcoder comment.

st2actions/st2actions/scheduler/handler.py
```diff
@@ -0,0 +1,164 @@
...
+            try:
+                with coordination.lock('st2shcedulerqueue'):
```
Kami: This is not the most ideal lock since it's quite high level and limits concurrent processing of executions.

For now though, it should probably work, but eventually we should figure out if we can make it more fine granular so we have multiple scheduler services processing set of executions which are pending to be processed.

st2actions/st2actions/scheduler/handler.py
```diff
@@ -0,0 +1,164 @@
...
+
+        LOG.info("Liveaction Status Pre-Run: %s", liveaction_db.status)
+
+        if liveaction_db.status is action_constants.LIVEACTION_STATUS_POLICY_DELAYED:
+            liveaction_db = action_service.update_status(
+                liveaction_db, action_constants.LIVEACTION_STATUS_DELAYED, publish=False
+            )
+            execution.scheduled_start_timestamp = date.append_milliseconds_to_time(
+                date.get_datetime_utc_now(),
+                500
```
Kami: I see what we you did here.

Eventually we can utilize similar approach for retry policy.

st2actions/st2actions/scheduler/handler.py
```diff
@@ -0,0 +1,164 @@
...
+        if liveaction_db:
+            if not self._exit_if_not_runnable(liveaction_db, execution):
+                self._update_to_scheduled(liveaction_db, execution)
+
+    @staticmethod
+    def _apply_pre_run(liveaction_db, execution):
+        # Apply policies defined for the action.
+        liveaction_db = policy_service.apply_pre_run_policies(liveaction_db)
+
+        LOG.info("Liveaction Status Pre-Run: %s", liveaction_db.status)
```
Kami: As @m4dcoder said above - we need to include more context to make those log messages more useful in production (at least the live action id, etc.).

st2actions/st2actions/scheduler/handler.py
```diff
@@ -0,0 +1,164 @@
...
+            raise
+
+        liveaction_db = self._apply_pre_run(liveaction_db, execution)
+
+        if liveaction_db:
+            if not self._exit_if_not_runnable(liveaction_db, execution):
+                self._update_to_scheduled(liveaction_db, execution)
+
+    @staticmethod
+    def _apply_pre_run(liveaction_db, execution):
```
Kami: Would be good to document the overall high lever behavior in the code somewhere - now this function set's execution status, etc.

st2actions/st2actions/scheduler/handler.py
```diff
@@ -0,0 +1,164 @@
...
+            return None
+
+        if (liveaction_db.status in action_constants.LIVEACTION_COMPLETED_STATES or
+                liveaction_db.status in action_constants.LIVEACTION_CANCEL_STATES):
+            ExecutionQueue.delete(execution)
+            return None
+
+        return liveaction_db
+
+    def _exit_if_not_runnable(self, liveaction_db, execution):
```
Kami: I would rename it to ``_is_liveaction_runnable`` or something along those lines - ``exit`` makes it the process exists if it's not runnable or similar, but the function really just returns a boolean.

tools/launchdev.sh
```diff
@@ -59,12 +61,10 @@ function init(){
         ST2_REPO=${CURRENT_DIR}/${COMMAND_PATH}/..
     fi
 
-    VIRTUALENV=${VIRTUALENV:-${ST2_REPO}/virtualenv}
```
Kami: I think this merge conflict resolution didn't work out since this file is missing some changes from master.

st2common/st2common/models/db/execution_queue.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+LOG = logging.getLogger(__name__)
+
+
+class ActionExecutionSchedulingQueueDB(stormbase.StormFoundationDB,
+        stormbase.ChangeRevisionFieldMixin):
+    RESOURCE_TYPE = ResourceType.EXECUTION_REQUEST
+    UID_FIELDS = ['id']
+    liveaction = me.StringField(required=True)
```
Kami: Minor thing - for explicitness sake I would prefer this to be called ``liveaction_id`` or similar.

Now it's not clear if it contains id reference to liveaction or something else.

st2common/st2common/models/db/execution_queue.py
```diff
@@ -0,0 +1,55 @@
...
+    scheduled_start_timestamp = ComplexDateTimeField(
+        default=date_utils.get_datetime_utc_now,
+        help_text='The timestamp when the liveaction was created.')
```
Kami: Should probably say "The timestamp when this liveaction was scheduled to be started" or similar?

st2actions/st2actions/scheduler/handler.py
```diff
@@ -0,0 +1,215 @@
...
+            "limit": 1,
+            "order_by": [
+                "scheduled_start_timestamp",
+            ]
+        }
+
+        execution = ExecutionQueue.query(**query).first()
+
+        if execution:
+            execution.handling = True
```
Kami: Unless I'm missing something - it looks like there could still be race if we run multiple ``st2scheduler`` processes since we don't perform any locking here.

Or am I missing something?

Kami: EDIT: Never mind I see it's using ``ChangeRevisionMongoDBAccess`` which should prevent that.

Would still be good to add a comment to this function on how coordination is handled since it's not immediately obvious from the code.

st2common/tests/unit/test_db.py
```diff
@@ -36,6 +36,17 @@
...
+
+
+# Previous test disrupts state for tests in this module. We reload the import
+# to avoid this. This try/except block is for python 3 support.
+try:
+    from imp import reload
+except:
+    from importlib import reload
+
+reload(mock_liveaction)
```
Kami: Yeah, that's nasty :/

Every tests which use that mock should have ``.wait_all()`` call in ``tearDown()`` to  ensure mock liveaction has processed all the pending work.

st2tests/st2tests/base.py
```diff
@@ -269,6 +270,33 @@ class DbTestCase(BaseDbTestCase):
     register_packs = False
     register_pack_configs = False
 
+    @staticmethod
+    def _wait_on_status(liveaction, status):
```
Kami: :+1:

st2common/st2common/models/db/execution.py
```diff
@@ -71,6 +71,7 @@ class ActionExecutionDB(stormbase.StormFoundationDB):
     parent = me.StringField()
     children = me.ListField(field=me.StringField())
     log = me.ListField(field=me.DictField())
+    delay = me.IntField()
```
Kami: We default it to ``0`` on another model below so we should also do the same here.

st2common/st2common/models/db/execution.py
```diff
@@ -178,6 +179,7 @@ class ActionExecutionOutputDB(stormbase.StormFoundationDB):
     runner_ref = me.StringField(required=True)
     timestamp = ComplexDateTimeField(required=True, default=date_utils.get_datetime_utc_now)
     output_type = me.StringField(required=True, default='output')
+    delay = me.IntField(required=True, default=0)
```
Kami: Actually, looking at the code, I don't think we even need this attribute on this model - ``ActionExecutionOutputDB``.

st2common/st2common/services/coordination.py
```diff
@@ -180,3 +180,46 @@ def get_coordinator():
...
+#         LOG.debug("Attempting to secure lock for: %s", self._name)
+#         if not self._lock.acquire():
+#             LOG.info("Could not secure lock for %s. Retrying.", self._name)
+#             self._timeout -= 1
+#             eventlet.sleep(.25)
+#             self.__enter__()
+
+#     def __exit__(self, *_args, **_kwargs):
+#         LOG.debug("Releasing lock for: %s", self._name)
+#         self._lock.release()
```
m4dcoder: Are these still needed?

Kami:
As mentioned and described on Slack, I pushed some changes:

1. Fix "missing index" test case failure issue - df0dd0cd7089a650b17fcb6e9f8af05d287267f5

This one took a while to track down. It was related to a missing index because a test method called ``reset()` which drops the collections and deletes the indexes. This would cause any test which runs afterwards and doesn't recreate indexes to fail.

You could say we should re-create indexes for each test method run in every test file. This would indeed be the most robust approach, but it's not a feasible one - creating indexes is very expensive (especially if we create them for all the models before each test method run) and would slow tests down by at least 50-100% (I tried that in the past).

So right now we only create indexes once per test file and in some scenarios, like this one, also before each test method run (but again, if that's not needed we try to avoid it due to the overhead).

1. Base test class refactor - 68682ad7b6b82cf2c59cac4f9e1f10b087563ead

I moved ``_wait_on_status`` and other execution test case related utility method on new base test class. 

This way we can only ensure index for ExecutionQueueDB model for test cases which use those methods where that is needed. That speeds things up compared to ensuring index for that model for each test case which uses ``DbTestCase`` class.

Issue #4398, opened 2018-10-17
Optimize bootstrap operations using bulk inserts and caches


lakshmi-kannan:
Right now, content registration can be slow since we pretty much register one unit at a time. We could potentially speed this up by using mongo bulk operations. This might help address  https://github.com/StackStorm/st2/issues/3686

Also, in cases like https://github.com/StackStorm/st2/pull/4388#discussion_r225938200, we could potentially cache dependent content upfront so we don't do one read query per write query.

PR #4399, opened 2018-10-18
Instrument registration and AMQP operations

lakshmi-kannan:
Opening to address https://github.com/StackStorm/st2/issues/4398. 

E.g.

```
2018-10-18 18:49:25,288 INFO [-] st2.lak-dev.mongo.query.trigger_type: 0.15800000ms
2018-10-18 18:49:25,291 INFO [-] st2.lak-dev.mongo.query.trigger: 0.07600000ms
2018-10-18 18:49:25,296 INFO [-] st2.lak-dev.mongo.add_or_update.trigger: 2.00600000ms
2018-10-18 18:49:25,303 INFO [-] st2.lak-dev.amqp.pool_publisher.publish_with_retries.st2.trigger: 5.95600000ms
2018-10-18 18:49:25,303 INFO [-] st2.lak-dev.amqp.publish.update: 6.41800000ms
2018-10-18 18:49:25,314 INFO [-] st2.lak-dev.mongo.query.trigger_type: 0.78900000ms
2018-10-18 18:49:25,324 INFO [-] st2.lak-dev.mongo.query.trigger: 0.40500000ms
2018-10-18 18:49:25,334 INFO [-] st2.lak-dev.mongo.add_or_update.trigger: 2.63300000ms
```


st2common/st2common/content/bootstrap.py
```diff
@@ -329,14 +330,22 @@ def register_content():
...
     if register_all:
-        register_triggers()
-        register_sensors()
-        register_runners()
-        register_actions()
-        register_rules()
-        register_aliases()
-        register_policies()
-        register_configs()
+        with Timer(key='st2.register.triggers'):
```
Kami: Looks like we only time those operations when ``register_all` is used and not when using a specific flag to only register a subset of resources.

I think it would be better to move this instrumentation inside ``register_<resource>`` functions.

This way it should work "out of the box" for all the approaches (using `--register-all` or ``--register-actions --register-rules``, etc.).

st2common/st2common/models/db/__init__.py
```diff
@@ -441,28 +445,35 @@ def distinct(self, *args, **kwargs):
         return result
 
     def aggregate(self, *args, **kwargs):
-        return self.model.objects(**kwargs)._collection.aggregate(*args, **kwargs)
+        with Timer(key='mongo.aggregate.' + self.model_name):
```
Kami: That's fine, for a bigger picture, I guess, but it won't give us that much insight since operations can and do vary vastly based on the filters and kwargs used.

lakshmi-kannan: Right, I think that's when your profiler would help. This is just to get a sense of cost of all the db operations and how many we are doing.

st2common/st2common/content/bootstrap.py
lakshmi-kannan: Good point. I'll fix this.

st2common/st2common/models/db/__init__.py
lakshmi-kannan: As you pointed out, I removed the DB instrumentation since we already have the profiler.

Kami:
LGTM.

If instrumentation for each publish call adds to much overhead or noise, we can remove it in the future.

Issue #4408, opened 2018-10-23
pip package missing cryptography dependency


koliyo:
##### SUMMARY

Installed using pip package

```
pip3 install -U st2client
```

Got error for missing dependency package

```
$ st2
Traceback (most recent call last):
  File "/usr/local/bin/st2", line 7, in <module>
    from st2client.shell import main
  File "/usr/local/lib/python3.7/site-packages/st2client/shell.py", line 27, in <module>
    from cryptography.utils import CryptographyDeprecationWarning
ImportError: cannot import name 'CryptographyDeprecationWarning' from 'cryptography.utils' (/usr/local/lib/python3.7/site-packages/cryptography/utils.py)
```

Manually install dep
```
pip3 install -U cryptography
```

Now it works

```
$ st2 --version
st2 2.9.1, on Python 3.7.0
```

##### ISSUE TYPE
Pick one below and delete the rest:
 - Bug Report

##### STACKSTORM VERSION
Paste the output of `st2 --version`:

st2 2.9.1, on Python 3.7.0

##### STEPS TO REPRODUCE

```yaml

```

Kami:
Thanks for reporting this.

I can confirm this issue.

We made some changes in the past which should prevent such issues from appearing, but it looks like those weren't enough.

I will push a fix shortly and I will also look into adding test cases which will allow us to automatically catch such regression in the future.

LindsayHill:
Fixed by #4409

PR #4409, opened 2018-10-23
Fix st2client PyPi install bug, add tests for it

Kami:
This pull request fixes issue reported in #4408.

We had this happen multiple times in the past so we added some changes which should have prevented that from happening again (#4209), but it looks like that wasn't enough.

This pull request ensures this won't happen again by adding tests. Tests create a new virtual environment, installs st2client into it and verifies it works.


Kami:
@bigmstone Do you think we should do v2.9.2 with this change or it's not worth it?

PR #4411, opened 2018-10-26
Workaround for intermediate failures

Kami:
Just testing a change to see if it will help with a lot of intermediate failures of concurrency tests.


Kami:
I will merge that and observe how Travis behaves after merge (I restarted the build 5 times and those tests always passed, but will see).

End to end tests are an unrelated failure on which I'm working atm.

m4dcoder:
@Kami Good catch. If this becomes a problem again, I'll pick it up.

Kami:
Will keep an eye on it.

On an related note - if that's indeed an issue, we also need to keep an eye on other tests which utilize non-blocking publisher. "Safe" thing to do would be to add ``wait_all`` to ``setUp`` / ``tearDown`` method for all the test classes which utilize non blocking publisher.

This way we can ensure there are no races and cross test case pollution when non blocking publisher is used.

PR #4412, opened 2018-10-29
Add option to notify on orquesta task completion

m4dcoder:
Add a runner option named notify to allow user to specify notification on task completion.

### DIFF
--- a/contrib/runners/orquesta_runner/orquesta_runner/runner.yaml
+++ b/contrib/runners/orquesta_runner/orquesta_runner/runner.yaml
@@ -4,6 +4,16 @@
   enabled: true
   name: orquesta
   runner_module: orquesta_runner
+  runner_parameters:
+    notify:
+      type: "array"
+      description: List of tasks to trigger notifications for.
+      uniqueItems: true
+      items:
+        type: "string"
+        minLength: 1
+        pattern: "^\\w+$"
+      default: []
   output_key: output
   output_schema:
     errors:
--- a/contrib/runners/orquesta_runner/tests/unit/test_notify.py
+++ b/contrib/runners/orquesta_runner/tests/unit/test_notify.py
@@ -1,0 +1,331 @@
+# Licensed to the StackStorm, Inc ('StackStorm') under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import absolute_import
+
+import jsonschema
+import mock
+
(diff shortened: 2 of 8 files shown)


contrib/runners/orquesta_runner/tests/unit/test_notify.py
```diff
@@ -0,0 +1,261 @@
...
+    'publish_state',
+    mock.MagicMock(side_effect=mock_lv_ac_xport.MockLiveActionPublisher.publish_state))
+@mock.patch.object(
+    wf_ex_xport.WorkflowExecutionPublisher,
+    'publish_create',
+    mock.MagicMock(side_effect=mock_wf_ex_xport.MockWorkflowExecutionPublisher.publish_create))
+@mock.patch.object(
+    wf_ex_xport.WorkflowExecutionPublisher,
+    'publish_state',
+    mock.MagicMock(side_effect=mock_wf_ex_xport.MockWorkflowExecutionPublisher.publish_state))
```
Kami: That's a lot of mocking.

Are all those needed and there is no way around that? (in the past while fixing some test issues I noticed a lot of tests have copied mocks from other places and a lot were actually not needed.

contrib/runners/orquesta_runner/orquesta_runner/runner.yaml
```diff
@@ -4,6 +4,11 @@
   enabled: true
   name: orquesta
   runner_module: orquesta_runner
+  runner_parameters:
+    notify:
+      type: array
```
Kami: Probably also not a bad idea to specify type for each array item aka ``items: type: string``.

contrib/runners/orquesta_runner/orquesta_runner/runner.yaml
```diff
@@ -4,6 +4,11 @@
   enabled: true
   name: orquesta
   runner_module: orquesta_runner
+  runner_parameters:
+    notify:
```
Kami: For action chain and mistral we also support ``skip_notify`` runner parameter, right?

It seems like a good idea to also support that here for consistency reasons, or?

m4dcoder: Nope, it's not a good idea. The skip notify and notify conflicts. `skip_notify` assumes notification is default on all task completion.  `notify` is the opposite.  @LindsayHill agree with me here that we should not opt-in folks into task notification.  it's a real hassle to have to explicitly turn off notification for tasks.  it's more common to don't want notification.

contrib/runners/orquesta_runner/tests/unit/test_notify.py
m4dcoder: @Kami it is the right set of mocking. You should take a closer look at how I write unit tests for orquesta.  i'm only mocking MQ/messaging here. I'm trying NOT to mock any actual functionality. Once a live action is requested, the unit tests will actually run the workflow and the actions to completion and test rather the notification is called or not.

contrib/runners/orquesta_runner/orquesta_runner/runner.yaml
Kami: I agree about that behavior as well - I was never a fan of opt-out, should have been opt-in.

Having said that, we just need to be explicit about that "behavior change" in the docs / upgrade notes (well, it's not really a behavior change, but inconsistency between 3 runners).

Kami: Let's please document that behavior somewhere, besides that, LGTM :+1:

m4dcoder: Yep. I will document this.

PR #4382, opened 2018-10-09
Adding st2scheduler

bigmstone:
This PR adds the initial break out of the scheduler. This PR will only encompass breaking out the service. Further enhancements will happen in subsequent PRs.

PR #4415, opened 2018-10-29
Improve st2.conf migration for the new services

armab:
The initial problem is that adding new st2 service requires us to include upgrade instructions like https://docs.stackstorm.com/upgrade_notes.html#st2-v2-9 and force users to handle `st2.conf` diffs.

The way to improve this story is to make default `st2.conf` logging settings work in prod deployments and so transition, migration or absense of specific directives in st2.conf doesn't fail startup.
This is an issue in infra (https://github.com/StackStorm/ops-infra/pull/295), configuration management requiring version-to-version migrations and overall bad. Having st2 core handling it right would be a better path.

Ideally ST2 core should have safe defaults that even no `st2.conf` config would allow us to start the services with no issues.

---

As mentioned in https://github.com/StackStorm/st2/pull/4382/files#r224189566, try to use `/etc/st2/` path as default setting for the new st2 services logging config and see what happens and what other changes may be needed.

### Services
- [x] st2api
- [x] st2stream
- [x] st2sensorcontainer
- [x] st2rulesengine
- [x] st2actionrunner
- [x] st2resultstracker
- [x] st2notifier
- [x] st2exporter
- [x] st2garbagecollector
- [x] st2timersengine
- [x] st2auth
- [x] st2workflowengine
- [x] st2scheduler


armab:
Much wow, I didn't expect this to pass the unit tests.

Kami:
Nice work and a good change :+1: 

We just need to confirm services run "out of the box" on a dev setup (I will try / verify that).

Do we need a changelog entry for this?

PR #4419, opened 2018-10-30
Fix st2timersengine screen session typo in launchdev

m4dcoder:
Fix the name of the screen session for st2-timersengine otherwise launchdev complains that st2timersengine did not start.


lakshmi-kannan:
Good catch

lakshmi-kannan:
Travis build is failing. Other than that, +1 from me. I'll approve.

PR #4424, opened 2018-11-05
Various Python 3 related fixes and changes

Kami:
I uncovered some more Python 3 related issues while working on Bionic pacakges.

This pull request includes the following changes / fixes:

1. Add missing ``st2garbagecollector`` service to launchdev script
2. Fix launchdev script so it actually uses `VIRTUALENV` environment variable - this allows people to use it with a Python 3 virtual environment
3. Fix ``st2-pack-install`` so it works under Python 3
4. Make sure we run pack tests on Travis under Python 3
5. Correctly pass ``timezone`` kwarg to ``BlockingScheduler`` class constructor

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -829,6 +829,7 @@ ci-py3-unit:
 	@echo "==================== ci-py3-unit ===================="
 	@echo
 	NOSE_WITH_TIMER=$(NOSE_WITH_TIMER) tox -e py36-unit -vv
+	NOSE_WITH_TIMER=$(NOSE_WITH_TIMER) tox -e py36-packs -vv
 
 .PHONY: ci-py3-integration
 ci-py3-integration: requirements .ci-prepare-integration .ci-py3-integration
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ else
 	VIRTUALENV_ST2CLIENT_DIR ?= virtualenv-st2client
 endif
 
-PYTHON_VERSION = python2.7
+PYTHON_VERSION ?= python2.7
 
 BINARIES := bin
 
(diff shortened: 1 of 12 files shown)


Kami:
And for people who want to test Python 3 stuff locally, they can now do this:

```bash
source virtualenv/bin/activate # this is needed to ensure latest version of virtualenv binary is used by Python 3
PYTHON_VERSION=python3.6 VIRTUALENV_DIR=virtualenv-py3 make requirements
source virtualenv-py3/bin/activate
./tools/launchdev.start
```

This will create a new virtual environment using Python 3.6 binary and start all the services using this new virtual environment.

Keep in mind that this command requires Python 3.6, python3-virtualenv, etc. be installed on the system.

Makefile
```diff
@@ -13,7 +13,7 @@ else
 	VIRTUALENV_ST2CLIENT_DIR ?= virtualenv-st2client
 endif
 
-PYTHON_VERSION = python2.7
+PYTHON_VERSION ?= python2.7
```
lakshmi-kannan: Wow. Things you learn :)

st2common/st2common/config.py
```diff
@@ -567,6 +567,48 @@ def register_opts(ignore_errors=False):
...
+
+    do_register_opts(timer_logging_opts, group='timer', ignore_errors=ignore_errors)
+    do_register_opts(timers_engine_logging_opts, group='timersengine', ignore_errors=ignore_errors)
+
+    # NOTE: We default old style deprecated "timer" options to None so our code
+    # works correclty and "timersengine" has precedence over "timers"
+    # NOTE: "timer" section will be removed in v3.1
+    timer_opts = [
+        cfg.StrOpt(
+            'local_timezone', default=None,
```
lakshmi-kannan: I read the original code and default=True for both timers and timersengine was pretty dumb on my part. I now get what you mean. Good catch.

PR #4426, opened 2018-11-05
Add tests for unicode input to Orquesta workflows

m4dcoder:
Add tests to ensure input with unicode data is processed and output correctly in Orquesta workflows.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -82,6 +82,7 @@ Fixed
   existing ``st2 pack show -j``.
 
   Reported by Nick Maludy (improvement) #4260
+* Fix string operations on unicode data in Orquest workflows. (bug fix)
 
 2.9.1 - October 03, 2018
 ------------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -82,7 +82,8 @@ Fixed
   existing ``st2 pack show -j``.
 
   Reported by Nick Maludy (improvement) #4260
-* Fix string operations on unicode data in Orquest workflows. (bug fix)
+* Fix string operations on unicode data in Orquest workflows, associated with PR
+  https://github.com/StackStorm/orquesta/pull/98. (bug fix)
 
 2.9.1 - October 03, 2018
 ------------------------
--- a/st2tests/integration/orquesta/test_wiring_data_flow.py
+++ b/st2tests/integration/orquesta/test_wiring_data_flow.py
@@ -57,6 +57,22 @@ class WiringTest(base.TestWorkflowExecution):
         self.assertEqual(ex.status, ac_const.LIVEACTION_STATUS_SUCCEEDED)
         self.assertDictEqual(ex.result, expected_result)
 
+    def test_data_flow_unicode_concat_with_ascii(self):
+        wf_name = 'examples.orquesta-sequential'
+        wf_input = {'name': '薩諾斯'}
+
+        expected_output = {
+            'greeting': '%s, All your base are belong to us!' % wf_input['name'].decode('utf-8')
+        }
+
+        expected_result = {'output': expected_output}
+
+        ex = self._execute_workflow(wf_name, wf_input)
+        ex = self._wait_for_completion(ex)
+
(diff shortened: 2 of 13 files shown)


st2tests/integration/orquesta/test_wiring_data_flow.py
```diff
@@ -57,6 +57,22 @@ def test_data_flow_unicode(self):
         self.assertEqual(ex.status, ac_const.LIVEACTION_STATUS_SUCCEEDED)
         self.assertDictEqual(ex.result, expected_result)
 
+    def test_data_flow_unicode_concat_with_ascii(self):
+        wf_name = 'examples.orquesta-sequential'
+        wf_input = {'name': '薩諾斯'}
+
+        expected_output = {
+            'greeting': '%s, All your base are belong to us!' % wf_input['name'].decode('utf-8')
```
Kami: Do those tests pass under Python 3?

Since, afaik, unicode object which is default for strings has no decode method (you can only encode it into a byte string or similar).

Kami: While looking at this PR I noticed we don't actually run those tests under Python 3 (an oversight).

We need to update tox integration targets to also run tests under ``st2tests/integration/orquesta/`` and deal with the potential failures.

On a related note - is there a reason why those tests live here and not under ``contrib/runners/orquesta/integration/``? I assume it's because it's not pure Orquesta runner tests and because they also test more of the wiring?

Kami: To begin with, I would just add ``nosetests --rednose --immediate -sv st2tests/integration/orquesta/`` to the end of ``py36-integration`` tox target in tox.ini.

We can deal with Mistral later.

I ran those tests manually and it looks like all but one pass (the one which fails is the same issue as with those newly tests - will need to update it to make it also pass under Python 3 by adding if check or similar).

m4dcoder: Those tests didn't live under `contrib/runners/orquesta/integration` because it started out under st2tests and then @bigmstone started adding tests under the new location.  I plan to move them to `contrib/runners/orquesta/integration` when I'm done with my deliverables for v3 release.  It's something that should be done but not a very high priority.

m4dcoder: I'll run py36 and see what breaks and fix.

m4dcoder: @Kami The orquesta integration tests require st2 to be running in the background. Is it a problem to add it to tox py36-integration?

Kami: There is a `ci-py3-integration` make target which is used on Travis which does that.

Personally, I usually manually launch services using ``./tools/launchdev.sh start`` and then use ``tox -epy36-integration`` since it's a bit faster

m4dcoder:
@Kami This should address your concerns. Please let me know of any other. Thanks.

m4dcoder:
@Kami the `-e` is causing some failure to the st2workflowengine service in the packages which is causing the circle-ci failures. I'm going to come back to this in a separate PR.

tox.ini
```diff
@@ -94,4 +94,5 @@ commands =
     nosetests --rednose --immediate -sv contrib/runners/local_runner/tests/integration/
     nosetests --rednose --immediate -sv contrib/runners/mistral_v2/tests/integration/
     nosetests --rednose --immediate -sv contrib/runners/orquesta_runner/tests/integration/
+	nosetests --rednose --immediate -sv st2tests/integration/orquesta/
```
Kami: Extra whitespace, but it's not end of the world :P

CHANGELOG.rst
```diff
@@ -82,6 +82,7 @@ Fixed
   existing ``st2 pack show -j``.
 
   Reported by Nick Maludy (improvement) #4260
+* Fix string operations on unicode data in Orquest workflows. (bug fix)
```
Kami: Please also reference Github pull request number here.

tox.ini
m4dcoder: Done

PR #4427, opened 2018-11-06
Stop raising Python exception when execution result is disabled.

nzlosh:
The ChatOps pack is used to format execution results.  When the action-alias metadata is configured to disable execution result being returned to the chat backend, an exception is raised in `format_execution_result` which causes the action to be marked as failed.

The failed task and back track are unexpected and confusing because the action is following the requested behaviour of the user.  It seems more logical to have the action succeed even when execution_results are disabled in the action-alias metadata and simply not post the result back to the chat backend in this case.

This PR implements this behaviour with the following changes:

- Stop raising Python exception when execution result is disabled in action-alias.
- Port post_result to orquesta for conditional branching to decide to post the execution result or not.
- Bumped version to 1.0.1

### DIFF
--- a/contrib/chatops/CHANGES.md
+++ b/contrib/chatops/CHANGES.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## [1.0.1] - 2018-11-06
+### Added
+- `note` to `post_result` output to inform if the action-alias execution result was enabled or disabled.
+
+### Changed
+- Ported `post_result` action-chain to orquesta for conditional branching.
+
 ## 0.3.0
 
 - Introduce the following new actions:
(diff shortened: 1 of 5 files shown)


nzlosh:
The chatops pack is deployed with the st2 core, I figured since it works with v3.0dev, it'd save porting it twice.  If Orquesta is too fresh, I can port to Mistral.

LindsayHill:
Definitely don't want to do it with Mistral. There are people who deployed ST2 without Mistral. But with Orquesta, that should be OK, since it ships as part of core.

blag:
When this makes it into a full ST2 release, Orquesta will be GA, so this should work "out of the box".

Thanks for the contribution!

PR #4428, opened 2018-11-07
More Python 3 fixes

Kami:
Python 3 related fixes for various issues which were discovered by new Ubuntu Bionic packages which use Python 3.

* Fix st2-run-pack-tests so it works under Python 3
* Fix various code so it works under Python 3
* Disable threaded mode for flask dev server in example sensor since it blocks the event loop (I noticed this via other sensor which also uses flask dev server which is used by self check)
* Update ``st2ctl`` and ``st2-self-check`` to take into account that Mistral is not present on Ubuntu Bionic. Per our discussion with Slack, I decided to go with a simple distro check. We can adjust / improve that in the future, when we drop support for Mistral in other distros.

### DIFF
--- a/st2common/bin/st2ctl
+++ b/st2common/bin/st2ctl
@@ -17,7 +17,7 @@
 COMPONENTS="st2actionrunner st2api st2stream st2auth st2garbagecollector st2notifier st2resultstracker st2rulesengine st2sensorcontainer st2chatops st2timersengine st2workflowengine st2scheduler mistral"
 ST2_CONF="/etc/st2/st2.conf"
 
-# Ubuntu Bionic packages skip without Mistral
+# Ubuntu Bionic packages ship without Mistral
 SUBTYPE=$(lsb_release -a 2>&1 | grep Codename | grep -v "LSB" | awk '{print $2}')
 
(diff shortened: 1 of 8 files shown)


st2common/bin/st2ctl
```diff
@@ -17,6 +17,13 @@
 COMPONENTS="st2actionrunner st2api st2stream st2auth st2garbagecollector st2notifier st2resultstracker st2rulesengine st2sensorcontainer st2chatops st2timersengine st2workflowengine st2scheduler mistral"
 ST2_CONF="/etc/st2/st2.conf"
 
+# Ubuntu Bionic packages ship without Mistral
+SUBTYPE=$(lsb_release -a 2>&1 | grep Codename | grep -v "LSB" | awk '{print $2}')
```
armab: There are 2 things here:
- `lsb_release --codename --short` should output `bionic`
- `lsb_release` is not available under EL systems

How about simple `py` version check in `/opt/stackstorm/st2/bin`. If it's `py2` - work as expected, if `py3` - skip mistral? This will also work long-term for other py3 distros we'll add in future (EL8?).

Kami: 05964b4, again, I'm fine either way.

PR #4435, opened 2018-11-09
[WIP] Add test for using action_context in Jinja expressions within action metadata

blag:
This PR exposes an issue where Orquesta does not expose the `action_context` to an action parameter that uses a Jinja expression: `Failed to render parameter "immutable_default_user": 'dict object' has no attribute 'user'`

That feature is used in [`st2cd.create_vm_role`](https://github.com/StackStorm/st2cd/blob/master/actions/create_vm_role.meta.yaml#L43) (ignore the weird indentation - that's from the YAML file itself):

```yaml
---
  name: "create_vm_role"
  runner_type: "mistral-v2"
  description: "Create a VM, add DNS, bootstrap puppet"
  enabled: true
  entry_point: "workflows/create_vm_role.yaml"
  parameters:
    ...
    creator:
      type: "string"
      default: "{{ 'api_user' in action_context and action_context.api_user or action_context.user }}"
immutable: true
```

The WF is currently successfully run by Mistral, but my Orquesta migration is running into this issue.

[Link](https://travis-ci.org/StackStorm/st2/jobs/453126614#L5996) to the exact error in the Travis log.

### DIFF
--- a/contrib/examples/actions/orquesta-action-context.yaml
+++ b/contrib/examples/actions/orquesta-action-context.yaml
@@ -1,0 +1,7 @@
+---
+name: orquesta-action-context
+description: >
+  A sample workflow demonstrating use of st2 runtime context in action metadata
+runner_type: orquesta
+entry_point: workflows/orquesta-action-context.yaml
+enabled: true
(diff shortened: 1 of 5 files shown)


cognifloyd:
I was curious when I saw this PR in slack. I haven't used `action_context` before.
If anyone else is curious, `action_context` is documented here:
https://docs.stackstorm.com/actions.html#parameters-in-actions

blag:
Fixed in #4443.

PR #4439, opened 2018-11-12
Update Circle CI config so we also build Bionic packages

Kami:
This pull request updates Circle CI config so we also build Ubuntu Bionic packages for each PR here.

We need this so I can hook up Bionic e2e tests for each PR.


Kami:
For some reason install_os_packages Circle CI step was getting stuck on "apt-get update" "reading package list..." and I needed to apply this workaround to get it to work - https://github.com/StackStorm/st2-packages/commit/e20cecd50784691c6805e3896d50254cfd7140da

Tried to re-run it many times and it always got stuck / timed out on that step without the workaround :/

Kami:
@arma I simply can't get Circle CI build to pass (and Circle CI is driving me crazy).

Install os packages step on Bionic hangs on "Reading packages lists..." - https://circleci.com/gh/StackStorm/st2/tree/build_bionic_packages

I tried many different workarounds and it passed only once, but during that build, EL7 failed with some other random error so the whole build was marked as failed.

I pretty much tried everything I can, yet I still can't get it to pass. Build gets stuck on install os packages even if I remove "apt-get update" step all together (which makes it seem it then gets stuck on gdebi step).

It's also worth noting that exactly the same step (docker compose test) passed many times before on st2-packages PR and master with Bionic changes.

Any ideas?

Kami:
@armab It looks like this workaround worked - https://github.com/StackStorm/st2-packages/commit/0e4e9e547bc6fff2a5322fb179881b028b6f0481

The build passed and we will see going forward. It looks like gdebi tries to use IPv6 or similar on Bionic by default and that's why it times out. It's just weird though it doesn't happen on st2-packages with the same Bionic image (I guess it could also be related to Circle CI networking or similar).

armab:
@Kami glad it resolved :+1: 

I suspect everything we're trying to do could be only fixed by CircleCI as they clearly have issues, per your discussion with their support.

Talking about st2-packages succeeded while st2 failed, - this could be some canary or sharding thing when CircleCI runs these 2 builds in different environments.

PR #4440, opened 2018-11-13
[WIP] Add new API endpoint which allows user to write arbitrary files to the pack directory on disk

Kami:
This pull request adds new ``POST /api/v1/packs/<pack ref>/files`` API endpoint which allows user with the correct RBAC permission (``PACK_MODIFY``) to write an arbitrary file in the pack directory on disk.

We already have ``data_files`` attribute in the POST and PUT (create, update) action API endpoint which allows user to write arbitrary action data files to the pack actions directory.

The difference between the existing API endpoints and the new one is that the existing one only allows user to write non-metadata action related files to disk (e.g. scripts, workflow definitions, etc.).

New endpoint allows user to write arbitrary files to the pack directory on disk, including action metadata files.

Keep in mind that if the user writes raw YAML action metadata file to the pack directory on disk, this change won't be reflected in the actions API and database until content in the database is synced with the content on disk.

The change utilizes existing code for handling and writing data files, so it should have no negative security implications (we only allow user to write inside a pack sub-directory, etc.).

Here is an example usage taking into account the context on how it will be used inside st2flow - https://gist.github.com/Kami/5226ca5f153fda4ec919abef0d34c9a6.

## Other changes

* Ensure we always sort value of ``PackDB.files`` attribute so the order is stable and doesn't change during register runs if the content on disk doesn't change.

## TODO

- [ ] Tests

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,7 +19,7 @@ Added
 
   Contributed by @aduca85 #4373
 * Add new ``POST /api/v1/packs/{ref_or_id}/files`` API endpoint which allows user to write
-  arbitrary files into the pack directory. (new feature)
+  arbitrary files into the pack directory. (new feature) #4440
 
 Changed
 ~~~~~~~
@@ -54,8 +54,8 @@ Changed
   argument (same as before). (improvement) #4396
 * Update various internal dependencies to latest stable versions (greenlet, pymongo, pytz,
   stevedore, tooz). #4410
-
-* Improve ``st2.conf`` migration for the new services by using prod-friendly logging settings by default #4415
+* Improve ``st2.conf`` migration for the new services by using prod-friendly logging settings by
+  default. #4415
 
 Fixed
 ~~~~~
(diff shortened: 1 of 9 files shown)


nmaludy:
@kami This looks cool! Curious what the use case is?

Kami:
@nmaludy It servers a pretty specific use case - st2flow will use it to save action metadata YAML file on disk.

As mention in the description - similar thing can already be achieved using the existing action create / update API and ``data_files`` attribute, only difference is that the existing API doesn't allow you to write action metadata file (only actual workflow definition files and other arbitrary script / data files).

Having said that, we don't really advertise any of those APIs because they have many limitations / caveats.

Both of those APIs are designed for single server deployments and shouldn't be used in any serious distributed HA deployment (unless you are using NFS for pack content, but this has it's own issues and it's not really recommended for serious distributed deployments in 2018).

In a distributed deployment, content shouldn't be modified directly on a single server, but "modify in git" and "deploy to all the servers" distributed workflow should be used (or a similar workflow with Kubernetes where content is baked into base images).

Kami:
Talked to @enykeev and it looks like we might not need this change for now.

I will close this PR (re-open it if it's needed at some later point) and cherry pick other useful changes in a new branch.

PR #4443, opened 2018-11-14
Fix context related bugs in orquesta

m4dcoder:
Fix access to st2 context in various locations of the workflow definition.  Fix access to api_user in the action context from the task action metadata.

### DIFF
--- a/st2common/st2common/services/workflows.py
+++ b/st2common/st2common/services/workflows.py
@@ -497,6 +497,7 @@ def request_action_execution(wf_ex_db, task_ex_db, st2_ctx, ac_ex_req):
 
     # Set context for the action execution.
     ac_ex_ctx = {
+        'user': st2_ctx.get('user'),
         'parent': st2_ctx,
         'orquesta': {
             'workflow_execution_id': str(wf_ex_db.id),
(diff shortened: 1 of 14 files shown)


st2common/st2common/services/workflows.py
```diff
@@ -496,6 +496,8 @@ def request_action_execution(wf_ex_db, task_ex_db, st2_ctx, ac_ex_req):
 
     # Set context for the action execution.
     ac_ex_ctx = {
+        'api_user': st2_ctx.get('api_user'),
```
cognifloyd: Is there a test to make sure `st2_ctx.api_user` is available if it's in the root st2 context (assuming it gets added in the root st2 context)? Maybe its added somewhere else, but I don't see it here:
https://github.com/StackStorm/st2/blob/f4bb39a90b17d66678cfdc9c65407dc4d2b8cdc7/st2common/st2common/services/workflows.py#L819-L821

blag:
Closes #4435.

Kami:
I assume we need a changelog entry for this change?

m4dcoder:
@Kami @blag @cognifloyd Your individual concerns are addressed.

Issue #4442, opened 2018-11-13
Feature Request - Add get_children method to st2client for liveactions


trstruth:
##### SUMMARY
While writing a custom web application that displays data from my stackstorm executions, I came across the use case of retrieving the children of a given execution id.  It's currently possible to retrieve similar data by invoking `client.liveactions.get_all({'parent': <execution-id>})` but this doesn't allow for controlling the depth to which children are retrieved (which the `GET/api/v1/executions/{id}/children` does).  By adding a method `get_children(self, execution_id, **kwargs)` to `LiveActionResourceManager`, we can make a request to the api endpoint mentioned above to expose the desired functionality in the python client.

I'm opening this issue in hopes of confirming that this is the correct way of going about doing this in the eyes of the maintainers, in which case I will open a PR to implement it.

##### ISSUE TYPE
 - Feature Idea

Kami:
Thanks for this feature request.

It sounds reasonable and useful to me, so PR is more than welcome :)

lakshmi-kannan:
Fixed in https://github.com/StackStorm/st2/pull/4444. Closing.

PR #4444, opened 2018-11-15
Add get_children method to st2client for liveactions

trstruth:
See #4442

This is my first PR for stackstorm, I tried to make this PR look like the other active PRS, but if I am missing anything or doing something wrong please let me know!

### DIFF
--- a/st2client/st2client/models/core.py
+++ b/st2client/st2client/models/core.py
@@ -450,8 +450,7 @@ class LiveActionResourceManager(ResourceManager):
         response = self.client.get(url=url, params=params, **kwargs)
         if response.status_code != http_client.OK:
             self.handle_error(response)
-        return [self.resource.deserialize(item)
-                for item in response.json()]
+        return [self.resource.deserialize(item) for item in response.json()]
 
 
 class InquiryResourceManager(ResourceManager):
--- a/st2client/st2client/models/core.py
+++ b/st2client/st2client/models/core.py
@@ -436,6 +436,23 @@ class LiveActionResourceManager(ResourceManager):
 
         return self.resource.deserialize(response.json())
 
+    @add_auth_token_to_kwargs_from_env
+    def get_children(self, execution_id, **kwargs):
+        url = '/%s/%s/children' % (self.resource.get_url_path_name(), execution_id)
+
+        depth = kwargs.pop('depth', -1)
+
+        params = kwargs.pop('params', {})
+
+        if depth:
+            params['depth'] = depth
+
+        response = self.client.get(url=url, params=params, **kwargs)
+        if response.status_code != http_client.OK:
+            self.handle_error(response)
+        return [self.resource.deserialize(item)
+                for item in response.json()]
+
 
 class InquiryResourceManager(ResourceManager):
 
--- a/st2client/tests/unit/test_client_liveaction.py
+++ b/st2client/tests/unit/test_client_liveaction.py
@@ -169,20 +169,18 @@ class TestLiveActionResourceManager(unittest2.TestCase):
         httpclient.HTTPClient.put.assert_called_with(endpoint, data)
 
     @mock.patch.object(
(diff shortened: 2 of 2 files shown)


st2client/st2client/models/core.py
```diff
@@ -436,6 +436,22 @@ def resume(self, execution_id, **kwargs):
...
+
+        params = kwargs.pop('params', {})
+
+        if depth:
+            params['depth'] = depth
+
+        response = self.client.get(url=url, params=params, **kwargs)
+        if response.status_code != http_client.OK:
+            self.handle_error(response)
+        return [self.resource.deserialize(item) for item in response.json()]
```
lakshmi-kannan: deserialize could potentially through. So a list comprehension may not be the best approach here. Since this is a client facing tool, we should try to be more precise about the error and not raise just the underlying exception all the way. 

Maybe it's best to build the list manually and on exception in a single item, raise an exception with a message like "Invalid resource format from server for item %s  in response %s" % (item, response.json()) might be better?

lakshmi-kannan:
Thanks for following up with a PR. Also, nice you got a nice fancy number 4444 ;). 

I have a higher level concern which is should we be exposing LiveAction model to the user? I understand today the HTTP client exposes it but LiveAction is supposed to be an internal model that's a light weight "reference" to the larger execution model. Can you describe why you want to consume this? Also, is your custom web open source? Anything that you are willing to contribute back to the community?

lakshmi-kannan:
(BTW I cleared the cache and re-ran the travis build. We have some issues with travis caches and randomly failing tests that we haven't gotten to bottom of.)

trstruth:
@lakshmi-kannan Thanks 4 reviewing!  I'm also confused as to why liveactions are exposed in st2client as opposed to executions.  Reading through the [st2client README](https://github.com/StackStorm/st2/blob/master/st2client/README.rst), we see

> The resource managers for the models are instantiated under the client as triggers, rules, actions, liveactions (for executions), and keys.

This seemed confusing to me because, as you point out, liveactions seem to differ from executions in various ways.  Regardless, the LiveActionResourceManager constructs a reference to the api/v1/executions endpoint, and I've been using it to get execution data.  I assumed that using the name liveaction was a side effect from some early st2 design decision and had just become the norm.  I don't know where all st2client.liveactions is used - my guess is that renaming it st2client.executions it would be a significant refactor effort.  Any more thoughts?

The web app is an internal proof of concept at the moment, my team and I are exploring ways to make a "generic output adapter/parser".  Meaning for some arbitrary json execution result object, render a meaningful react element as opposed to just pretty printing some the stringified json.  It's still in the very early stages but if we manage to get anywhere I'd be happy to introduce some version of it to the community.  Maybe I'd take the adapter and incorporate it into a fork of st2web - any other ideas around how y'all might like to see a contribution like that?

The travis build still seems to be failing, but as far as I can tell not for any reasons related to my changes.

st2client/st2client/models/core.py
trstruth: Does manually constructing the list differ from just wrapping the comp in a try?
```suggestion
        try:
            return [self.resource.deserialize(item) for item in response.json()]
        except JSONDecodeError as e:
            raise JSONDecodeError("Invalid resource format from server for item %s in response %s" % (item, response.json()))
```
A similar list comp appears elsewhere in the same file, the `get_by_id()` and `get_property()` methods of `ResourceManager` implement them.  Should they be updated to catch this error as well?

Kami:
Yeah, ``liveactions`` stuff is a left-over from the early days :)

One thing we could do it add ``st2client.executions`` alias for ``st2client.liveactions`` and deprecate ``st2client.liveactions`` key and remove it in a future version.

Besides that, I don't really see any blocker so I will go ahead and merge it once tests pass.

I will look into making that ``st2client.executions`` alias in a separate PR.

PR #4447, opened 2018-11-15
Fix Exception messages

LindsayHill:
We had a few instances where we did not properly format our Exception messages, leading to a literal `%s` in the raised Exception.

I used `pylint -s n --disable=all -e W0715` against all our code to identify all places where this happened.


Kami:
Good catch.

I prefer `%` over `,` anyway (less error prone).

PR #4455, opened 2018-11-26
More scheduler changes

Kami:
Some more changes on top off #4397:

* Use a module level constant and use a larger delay when trying to re-schedule policy delayed executions. Trying to re-schedule it that often (500 ms) could potentially cause a lot of unnecessary load on the scheduler service and a slightly longer delay is not a big deal when we are re-scheduling policy delayed executions. This may also help with race in the tests.


st2actions/tests/unit/policies/test_concurrency.py
```diff
@@ -88,12 +88,18 @@ def setUpClass(cls):
         loader.save_fixtures_to_db(fixtures_pack=PACK,
                                    fixtures_dict=TEST_FIXTURES)
 
+    # NOTE: This monkey patch needs to happen again here because during tests for some reason this
+    # method gets unpatcked (test doing reload() or similar)
```
cognifloyd: s/unpatcked/unpatched/

st2actions/tests/unit/policies/test_concurrency.py
```diff
@@ -88,12 +88,18 @@ def setUpClass(cls):
...
+    @mock.patch('st2actions.container.base.get_runner', mock.Mock(return_value=runner.get_runner()))
     def setUp(self):
         super(ConcurrencyPolicyTestCase, self).setUp()
 
         # Wait for all threads to finish processing so there is no cross test polution
         MockLiveActionPublisherNonBlocking.wait_all()
 
+    # NOTE: This monkey patch needs to happen again here because during tests for some reason this
+    # method gets unpatcked (test doing reload() or similar)
```
cognifloyd: s/unpatcked/unpatched/

st2actions/tests/unit/policies/test_concurrency.py
Kami: Thanks, will fix those typos 

As far as the actual change goes, it's quite nasty. We've had those tests intermediy failing for a long time without being able to track down the root cause.

The problem is the asynchronous nature of the tests and that they only fail every now and then (usually on master where coverage which changes test timing a bit is enabled).

I'm still not 100% sure this will fix all the issues, but will see.

PR #4456, opened 2018-11-26
Add test to cover scheduler race condition.

bigmstone:
This test covers the case when two schedulers pick up the same execution
and ensures the task is not picked up twice.

### DIFF
--- a/st2actions/tests/unit/test_scheduler.py
+++ b/st2actions/tests/unit/test_scheduler.py
@@ -35,6 +35,7 @@ from st2common.models.db.liveaction import LiveActionDB
 from st2common.persistence.execution_queue import ActionExecutionSchedulingQueue
 from st2common.persistence.liveaction import LiveAction
 from st2common.services import executions as execution_service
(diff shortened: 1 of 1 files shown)


st2actions/tests/unit/test_scheduler.py
```diff
@@ -231,3 +232,30 @@ def test_processing_when_task_completed(self, mock_execution_queue_delete, mock_
...
+
+        with mock.patch(
+            'st2actions.scheduler.handler.ActionExecutionSchedulingQueue.add_or_update',
+            side_effect=db_exc.StackStormDBObjectWriteConflictError(schedule_q_db)
+        ):
+            schedule_q_db = self.scheduling_queue._get_next_execution()
+            self.assertIsNone(schedule_q_db)
+
+        mocked_logger.info.assert_called_once()
+        call_args = mocked_logger.info.call_args_list[0][0]
```
Kami: That works for now :+1: 

It's not the most robust test, but writing an end to end test to simulate that would be very hard.

PR #4457, opened 2018-11-26
Remove the delay parameter on action execution reruns

m4dcoder:
On rerunning a delayed action execution, the delay parameter is removed before the action execution is rerun.

### DIFF
--- a/st2api/st2api/controllers/v1/actionexecutions.py
+++ b/st2api/st2api/controllers/v1/actionexecutions.py
@@ -415,9 +415,14 @@ class ActionExecutionReRunController(ActionExecutionsControllerMixin, ResourceCo
             raise ValueError('Task option is only supported for Mistral workflows.')
 
         # Merge in any parameters provided by the user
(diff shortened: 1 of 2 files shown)


Kami:
LGTM, but let's document this somewhere - perhaps in the API docs (openapi.yaml.j2).

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -415,9 +415,14 @@ def post(self, spec_api, id, requester_user, no_merge=False, show_secrets=False)
             raise ValueError('Task option is only supported for Mistral workflows.')
 
         # Merge in any parameters provided by the user
+        old_parameters = getattr(existing_execution, 'parameters', {})
+        old_parameters.pop('delay', None)
```
Kami: Should we also log a message here?

Kami:
Looks like a test is failing.

m4dcoder:
This is not required because `delay` is not in the list of parameters.

PR #4458, opened 2018-11-27
Add support for task delay in orquesta workflow

m4dcoder:
Add support for task delay in the orquesta workflow definition and integrate with the scheduler to delay action execution when specified.


st2common/st2common/models/db/workflow.py
```diff
@@ -63,6 +63,7 @@ class TaskExecutionDB(stormbase.StormFoundationDB, stormbase.ChangeRevisionField
     task_name = me.StringField(required=True)
     task_id = me.StringField(required=True)
     task_spec = stormbase.EscapedDictField()
+    delay = me.IntField(min_value=0)
```
Kami: Probably a good idea to also set the same ``min_value`` on ExecutionDB and LiveActionDB models.

st2common/st2common/services/workflows.py
```diff
@@ -714,11 +743,22 @@ def refresh_conductor(wf_ex_id):
...
+        copy.copy(ac_const.LIVEACTION_COMPLETED_STATES) +
+        [ac_const.LIVEACTION_STATUS_PAUSED, ac_const.LIVEACTION_STATUS_PENDING]
+    )
 
     if ac_ex_status not in states_to_process:
         return
 
+    # Replace/Simplify specific states.
+    states_to_collapse = {
+        ac_const.LIVEACTION_STATUS_DELAYED: ac_const.LIVEACTION_STATUS_DELAYED
```
Kami: A bit confused on what this line is supposed to do.

Should one of the values be policy delayed, or?

m4dcoder: Yeah, one of this is suppose to be policy delayed.  Thanks for catching this. I worked on this patch before the scheduler PR was merged.

PR #4459, opened 2018-11-27
Adding delay to re-run

bigmstone:
This commit adds necessary changes to st2client and st2api to accommodate delay
on re-run of actions.

### DIFF
--- a/st2api/st2api/controllers/v1/actionexecutions.py
+++ b/st2api/st2api/controllers/v1/actionexecutions.py
@@ -406,6 +406,11 @@ class ActionExecutionReRunController(ActionExecutionsControllerMixin, ResourceCo
         if list(set(spec_api.reset) - set(spec_api.tasks)):
             raise ValueError('List of tasks to reset does not match the tasks to rerun.')
 
+        delay = None
+
+        if hasattr(spec_api, "delay") and isinstance(spec_api.delay, int):
+            delay = spec_api.delay
+
         no_merge = cast_argument_value(value_type=bool, value=no_merge)
         existing_execution = self._get_one_by_id(id=id, exclude_fields=self.exclude_fields,
                                                  requester_user=requester_user,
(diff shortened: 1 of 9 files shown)


st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -457,7 +463,7 @@ class ActionExecutionsController(BaseResourceIsolationControllerMixin,
                                  ActionExecutionsControllerMixin, ResourceController):
     """
         Implements the RESTful web endpoint that handles
-        the lifecycle of ActionExecutions in the system.
+        the lifecycle of ActionExecutios in the system.
```
m4dcoder: typo?

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -406,6 +406,11 @@ def post(self, spec_api, id, requester_user, no_merge=False, show_secrets=False)
         if list(set(spec_api.reset) - set(spec_api.tasks)):
             raise ValueError('List of tasks to reset does not match the tasks to rerun.')
 
+        delay = None
+
+        if hasattr(spec_api, "delay") and isinstance(spec_api.delay, int):
```
Kami: I would just do ``delay = getattr(spec_api, 'delay', None)``, but that also works.

st2common/st2common/models/db/execution.py
```diff
@@ -179,7 +179,7 @@ class ActionExecutionOutputDB(stormbase.StormFoundationDB):
     runner_ref = me.StringField(required=True)
     timestamp = ComplexDateTimeField(required=True, default=date_utils.get_datetime_utc_now)
     output_type = me.StringField(required=True, default='output')
-    delay = me.IntField(required=True, default=0)
+    delay = me.IntField()
```
Kami: Any particular reason why you removed that? I guess required with a default value doesn't make much sense / difference.

bigmstone: Making it consistent with other delay references in other collections. No other reason.

PR #4460, opened 2018-11-28
Fix race and random failures with unit tests for concurrency policy

m4dcoder:
Fix race and random failures with unit tests for concurrency policy by using the scheduling only mock publisher and handling the processing of the scheduling queue manually. Use the noop coordinator driver to prevent blocking in the process. The lock mechanism should be unit tested elsewhere. The separation of concerns should improve the robustness of these tests.

### DIFF
--- a/st2actions/tests/unit/policies/test_concurrency.py
+++ b/st2actions/tests/unit/policies/test_concurrency.py
@@ -15,29 +15,30 @@
 
 from __future__ import absolute_import
 
-import eventlet
 import mock
 from mock import call
 from six.moves import range
 
 import st2common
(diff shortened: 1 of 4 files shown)


m4dcoder:
@Kami @bigmstone This PR still needs some more work. When running `make ci-unit`, the coordinator driver is back to normal with `zake`. I'm tempted to make the noop driver the default for coordinator.

m4dcoder:
@Kami @bigmstone Ok the last commit should fix.

Kami:
Nice, much better 👍

Kami:
I will also cherry pick this in v2.9 branch since it's causing me issues there.

Issue #4451, opened 2018-11-21
(Question) Why does DatastoreService inhibit from handling non system-scope value?


userlocalhost:
##### SUMMARY
I want to handle user-scope encrypted value in the datastore from action through DatastoreService methods.

An use case is to get user defined values which are different at each users like account information to access to each integrated system from an action.

But a past patch(https://github.com/StackStorm/st2/pull/2731) [inhibits from getting non system-scope value](https://github.com/StackStorm/st2/blob/master/st2common/st2common/services/datastore.py#L115-L116).

There is no description and reference for that patch. So could you please tell me what kind of problem this patch tries to solve for, and whether that problem still exists?

Thank you.

##### ISSUE TYPE
 - Other

##### STACKSTORM VERSION
`3.0dev`

Kami:
Right now user-scoped datastore values can only be used and accessed for actions which are scheduled via the StackStorm API.

The reason for that is that those are the only actions which have a user context attached to them. If an action is ran via the API, user is set to the user which is authenticated against the API.

If action is ran via rule or if you want to use user-scoped value inside a sensor, that won't work because there is a no concept of "user" in those scenarios (sensors and rule triggered actions run under StackStorm system user ``stanley``).

Can you please share in more details your use case aka where you want to use user-scoped datastore values?

userlocalhost:
Thank you for your reply. I'm glad to hear the important points to use user-scoped value, especially the case that an action is run from rule by system user.

> Can you please share in more details your use case aka where you want to use user-scoped datastore values?

Yes. In our environment, the StackStorm integrates many infrastructure services which hss own authentication system. And StackStorm set account, named "st2", for each integrated system to the pack's configuration.

In accordance with the growth of the number of integrated systems and users, the next operation problems were happened.

### 1. It's hard to track who manipulates each integrated systems.
  There are a lot of operation logs by "st2" user for each integrated system. For example, once a wrong configured VM instance was found in vSphere, it's hard to find out who create it. We always dig into execution history of StackStorm. It takes much operation cost because the execution time of StackStorm and the time logged at integrated system doesn't match exactly. 

### 2. It makes a loophole of permissions.
  By using a common account for an integrated systems, all users can access them even though some systems don't permit to them. That prevents StackStorm from prevailing in our company.

---
[...]

PR #4463, opened 2018-12-03
Changed to be able to read user-scope datastore value via '{{ st2kv.user }}' notation.

userlocalhost:
## Background
The change of (https://github.com/StackStorm/st2/pull/3199) and [document of Datastore](https://docs.stackstorm.com/datastore.html#scoping-datastore-items) say StackStorm supports the I/F to get datastore value via {{ st2kv.system }} and {{ st2kv.user }}.

### Related Issue
* https://github.com/StackStorm/st2/issues/4451#issuecomment-442612125

## Problem
StackStorm didn't have an implementation to accept `{{ st2kv.user }}` expression.

## Change
This implements the processing to accept `{{ st2kv.user }}` notation and render it to user-scope value in datastore which is associated with executed user.


st2common/st2common/util/param.py
```diff
@@ -81,6 +82,13 @@ def _create_graph(action_context, config):
     '''
     G = nx.DiGraph()
     system_keyvalue_context = {SYSTEM_SCOPE: KeyValueLookup(scope=FULL_SYSTEM_SCOPE)}
+
+    # If both 'user' and 'api_user' are specified, this prioritize 'api_user'
+    user = action_context['user'] if 'user' in action_context else None
+    user = action_context['api_user'] if 'api_user' in action_context else user
+    if user:
+        system_keyvalue_context[USER_SCOPE] = UserKeyValueLookup(scope=FULL_USER_SCOPE, user=user)
```
Kami: Hm, is this desired?

I'm fine with this approach, but we should document this behavior (something along the lines that ``st2kv.user.foo`` will only work for actions which have user context attached to and and are executed via the API or similar).

Kami: Besides that, I'm fine with this change.

Nice work :+1:

userlocalhost: Thank you very much for your review. I got it. I'll do that soon!

userlocalhost:
Thank you again for your review and telling me advices to brush up.

> Also, what will happen if we are unable to read user from the context and action author references st2kv.user in a default value?

I've taken care of this problem by the idea that system user fills in for the one.

userlocalhost:
I fixed the points that you mentioned, would you please review this patch, @Kami?

userlocalhost:
Thank you very much!

st2common/st2common/util/param.py
Kami: @userlocalhost Can you please also open st2docs PR which documents this behavior?

Kami:
Merged, thanks.

Please also open a corresponding st2docs PR when you get a chance :)

userlocalhost:
I think we don't have to amend st2docs because the explanation of this feature has already been described at [the Datastore page in st2docs](https://github.com/StackStorm/st2docs/blob/fe748d1400b09ca0c085cbab4095e84429676004/docs/source/datastore.rst#scoping-datastore-items).

This is an excerpt from it.
```
Also, you can refer to user variables in actions or workflows. The Jinja syntax to do so is {{st2kv.user.date_cmd}}.

Note that the notion of st2kv.user is available only when actions or workflows are run manually. The notion of st2kv.user is non-existent when actions or workflows are kicked off via rules. So the use of user scoped variables is limited to manual execution of actions or workflows.
```

I think this description is good enough to explain how to use this and what limitations are.

PR #4474, opened 2018-12-14
Add new "python_versions" attribute to the pack metadata file

Kami:
This pull request adds new ``python_version`` attribute to the pack metadata file.

This attribute uses semver notation to specify which Python versions pack works with (e.g. ``>= 2.7.0, <= 3.7.0``).

The goal is to utilize this attribute for various things:

1. On pack installation, check if to be installed pack is compatible with Python version under which StackStorm is running
2. Advertise Python 2.x / Python 3.x compatibility on pack exchange
3. Utilize this attribute in the pack CI - only runs tests and checks for Python version pack is compatible with and consider failures as fatal

Once we agree on the attribute "design", I / we will go and manually populate this field for various most commonly used pack based on manual and automated testing (https://github.com/StackStorm-Exchange/ci/pull/61).

Related PR: https://github.com/StackStorm-Exchange/ci/pull/61

## Open Questions

I went with semver approach because we already use that for ``stackstorm_version`` attribute (consistency).

After more thought, I think it's too complicated - we could make it simpler by only advertising general python 2 and python 3 compatibility by adding ``{"python_versions": "python2": true, "python3": false}`` or similar attribute to pack metadata.

This attribute is different from ``stackstorm_version`` one - in StackStorm version case, you might want to specify different minor and patch versions and this should rarely be the case with Python version.

What do others think?


jeking3:
Use semver - it's standard.

lakshmi-kannan:
Although semver is standard and which is what we use for st2 version, the python version really doesn't need the level of details. We are really interested in the python major versions. So something like "python_major_versions": [2, 3] is an option too.

Kami:
Another thing worth noting - packs will only work with with Python 2.x and Python 3.x versions supported by the platform (aka StackStorm packages and components).

This means it makes no sense for pack to also specify Python version beyond the major one (aka also specifying minor and patch version).

Minor and patch versions will always be dictated by the platform.

Right now that is 2.7.x and 3.6.x. That means that packs which declare support for Python 2 are tested and work with Python 2.7.x and same goes for Python 3.

Kami:
I've pushed necessary changes which add new ``python_versions`` attribute which is a list (as discussed above).

Valid values for that field are "2", "3: and combination of those two. I made it a string, but if other think it makes more sense, I can also make it an integer (I didn't make it an integer because we will never do number like comparisons on it).

st2common/st2common/models/api/pack.py
```diff
@@ -100,6 +100,22 @@ class PackAPI(BaseAPI):
                                '">=1.8.0, <2.2.0"',
                 'pattern': ST2_VERSION_REGEX,
             },
+            'python_versions': {
+                'type': 'array',
+                'description': ('Major Python versiosn supported by this pack. E.g. '
```
jeking3: Typo "versiosn".

st2common/st2common/models/api/pack.py
```diff
@@ -100,6 +100,22 @@ class PackAPI(BaseAPI):
...
+                                '"2" for Python 2.7.x and "3" for Python 3.6.x'),
```
jeking3: This will change over time, is it wise to put the minor versions in here?  I could see this getting stale.

st2common/st2common/models/api/pack.py
```diff
@@ -100,6 +100,22 @@ class PackAPI(BaseAPI):
...
+                    'enum': [
+                        '2',
+                        '3'
+                    ]
+                },
+                'minItems': 1,
+                'maxItems': 2,
+                'uniqueItems': True,
+                'additionalItems': True
+            },
```
jeking3: What is the default for packs that do not specify it?

st2common/st2common/models/api/pack.py
```diff
@@ -100,6 +100,22 @@ class PackAPI(BaseAPI):
...
+                'description': ('Major Python versiosn supported by this pack. E.g. '
+                                '"2" for Python 2.7.x and "3" for Python 3.6.x'),
+                'items': {
+                    'type': 'string',
+                    'enum': [
+                        '2',
+                        '3'
+                    ]
+                },
+                'minItems': 1,
```
jeking3: Without a default and minItems = 1 will this break all existing packs?

st2common/st2common/models/api/pack.py
Kami: Thanks, will fix it.

st2common/st2common/models/api/pack.py
Kami: As mentioned here (https://github.com/StackStorm/st2/pull/4474#issuecomment-452314335), the minor and patch version itself are dictated by the StackStorm platform and the packages we ship.

We could also store it here, but I don't think it would serve any good purpose - the information itself would be duplicated and provide no additional value, just potential confusion.

The idea is if you are using official StackStorm packages under Python 2 and pack metadata says it works under Python 2, it should work. Same goes for Python 3.

st2common/st2common/models/api/pack.py
Kami: We don't specify ``required=True`` which means it's an optional attribute. If it's not provided, we don't perform any checks during pack install and register phase (we just assume it works with the version under which StackStorm is currently running).

That's done for backward compatibility reasons.

st2common/st2common/models/api/pack.py
Kami: Nope - min items only applies to this attribute when it's provided.

We also have various test cases which cover that scenarios (we have a lot of pack test fixtures without that attribute).

Kami:
@jeking3 Thanks for taking a look and your feedback.

I pushed a fix for the typo and I believe I addressed all of your comments. If I missed something, please let me know.

st2common/st2common/models/api/pack.py
blag: Why `maxItems` here? This means we have to update this code whenever a new major version of Python is released. Which, granted, doesn't happen all that often, but if packs want to specify a bunch of different values, I don't see how that is a problem.

st2common/st2common/models/api/pack.py
jeking3: My point was that the string on line 106 will become incorrect over time.

Kami:
@blag Re maxItems - that's correct, but it doesn't happen often.

And if a new version of Python is released this means we will likely need to update our code and definitely update our packages since we bundle the virtualenv with Python interpreter inside the packages.

In short - if a new version for Python is released and some user wants to use it, we will need to update our code (packs CI, etc.) anyway and at that time we will update maxItems as well.

armab:
@Kami Is this feature documented somewhere?
Couldn't find any references in st2docs.

Kami:
@armab Not yet - feel free to create an issue and assign it for me. I will go to it in time for v3.0.0.

contrib/packs/tests/test_action_download.py
```diff
@@ -241,41 +241,141 @@ def test_download_pack_stackstorm_version_identifier_check(self):
...
         # Version is satisfied
-        st2common.util.pack_management.CURRENT_STACKSTROM_VERSION = '2.0.0'
+        st2common.util.pack_management.CURRENT_STACKSTORM_VERSION = '2.0.0'
 
         result = action.run(packs=['test3'], abs_repo_base=self.repo_base)
         self.assertEqual(result['test3'], 'Success.')
 
         # Pack requires a version which is not satisfied by current StackStorm version
-        st2common.util.pack_management.CURRENT_STACKSTROM_VERSION = '2.2.0'
+        st2common.util.pack_management.CURRENT_STACKSTORM_VERSION = '2.2.0'
```
armab: There is definitely a 100 and 1 way to misspell StormStack :smiley:

armab:
@Kami Sounds good, here we go https://github.com/StackStorm/st2docs/issues/870

PR #4476, opened 2018-12-17
Fix for ?scope=all and ?user=<username> query parameter handling in /v1/keys API endpoint

Kami:
``GET /v1/keys`` API endpoint (get all / view keys) API endpoint incorrectly handled ``?scope=all`` and ``?user=<username>`` query parameter.

Enterprise deployments with RBAC enabled are not affected because there only admin users are allowed to use ``?scope=all`` query parameter.

This pull request fixes that.

### DIFF
--- a/st2api/tests/unit/controllers/v1/test_kvps.py
+++ b/st2api/tests/unit/controllers/v1/test_kvps.py
@@ -15,6 +15,14 @@
 
 from tests import FunctionalTest
 
+from st2common.models.db.auth import UserDB
+
+from six.moves import http_client
+
+__all__ = [
+    'KeyValuePairControllerTestCase'
+]
+
 KVP = {
     'name': 'keystone_endpoint',
     'value': 'http://127.0.0.1:5000/v3'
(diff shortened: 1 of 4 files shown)


st2api/tests/unit/controllers/v1/test_kvps.py
```diff
@@ -77,6 +85,154 @@ def test_get_one(self):
         self.assertEqual(self.__get_kvp_id(get_resp), kvp_id)
         self.__do_delete(kvp_id)
 
+    def test_get_all_all_scope(self):
```
lakshmi-kannan: Took a look at the tests. Good set.

Kami:
Once this PR is reviewed and merged, I will create branch which cherry pick those changes in v2.9 and v2.10 branches.

Kami:
Related branch for v2.9.2 and v2.10.0 release - https://github.com/StackStorm/st2/pull/4479, https://github.com/StackStorm/st2/pull/4480.

PR #4485, opened 2019-01-02
Fix kombu release version

LindsayHill:
The Kombu team had a release issue, and accidentally published the wrong code as 4.2.2 to PyPi. They have fixed this, and pulled 4.2.2. 

https://github.com/celery/kombu/issues/966

### DIFF
--- a/fixed-requirements.txt
+++ b/fixed-requirements.txt
@@ -5,7 +5,7 @@ greenlet==0.4.15
 # depend on rely
 eventlet==0.24.1
 gunicorn==19.9.0
-kombu==4.2.2
+kombu==4.2.2.post1
 # Note: amqp is used by kombu
 amqp==2.3.2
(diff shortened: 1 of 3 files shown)


fixed-requirements.txt
```diff
@@ -5,7 +5,7 @@ greenlet==0.4.15
 # depend on rely
 eventlet==0.24.1
 gunicorn==19.9.0
-kombu==4.2.2
+kombu==4.2.2.post1
```
armab: holy

Kami: Yeah, it's part of a Python versioning standard - https://www.python.org/dev/peps/pep-0440/#post-releases

I'm fine with it since it's considered a "final" and not a dev / beta release. But yeah, I would definitely prefer them to actually bump a version to 4.2.3 or similar.

PR #4491, opened 2019-01-07
Fix a bug with trigger payload validation in linux.FileWatchSensor sensor

### LINKED ISSUE #4486: [st2reactor] Issue about validation fail  before dispatching trigger,Trigger type with reference "linux.cd875f82-58d6-47ff-9eff-11bd81493bf7" doesn't exist in the database
wen I  test sensor linux.FileWatchSensor: 

```shell
root@6206c3910cd6:/opt/stackstorm/packs/linux/sensors# st2 sensor get linux.FileWatchSensor 
| id            | 5c2d8324d28ef50172db13e9                                     |
| uid           | sensor_type:linux:FileWatchSensor                            |
| ref           | linux.FileWatchSensor                                        |
| pack          | linux                                                        |
| name          | FileWatchSensor                                              |
| entry_point   | sensors.file_watch_sensor.FileWatchSensor                    |
| artifact_uri  | file:///opt/stackstorm/packs/linux/sensors/file_watch_sensor |
|               | .py                                                          |
| trigger_types | [                                                            |
|               |     "linux.file_watch.line"                                  |
|               | ]                                                            |
| description   | Sensor which monitors files for new lines                    |
| metadata_file | sensors/file_watch_sensor.yaml                               |
```

I use rule as follow:

```shell
root@6206c3910cd6:/opt/stackstorm/packs/examples/rules# cat sample_rule_file_watch.yaml 
---
name: sample_rule_file_watch
pack: "examples"
description: Sample rule custom trigger type - add a file to be watched by file_watch_sensor in linux pack.
enabled: true

trigger:
  parameters:
    file_path: /tmp/st2_test
  type: linux.file_watch.line

criteria: {}

action:
  parameters:
    cmd: echo "{{trigger}}"
  ref: core.local
```

I execute cmd, like this:
   `echo >> liucong /tmp/st2_test`

and err log in `/var/log/st2/st2sensorcontainer.log` as follows:

```
2019-01-03 03:39:07,046 139683133441840 INFO tail [-] Tailing /tmp/st2_test
2019-01-03 03:39:07,047 139683133441840 INFO file_watch_sensor [-] Added file "/tmp/st2_test"
2019-01-03 03:41:00,852 139683181345520 WARNING trigger_dispatcher [-] Failed to validate payload ({'file_name': u'st2_test', 'line': u'liucong', 'file_path': u'/tmp/st2_test'}) for trigger "linux.cd875f82-58d6-47ff-9eff-11bd81493bf7": Trigger type with reference "linux.cd875f82-58d6-47ff-9eff-11bd81493bf7" doesn't exist in the database
```

Kami:
This pull request fixes an issue with file watch sensor not working correctly since the trigger payload validation has been enabled by default a couple of releases ago.

The root cause was that ``validate_trigger_payload()`` method only worked with TriggerType reference, but not with Trigger reference.

FileWatch sensor is a bit special since it's the only sensor which works with actual triggers and not trigger types (aka issue didn't affect other sensors).

Resolves #4486 

## TODO

- [x] Test case


contrib/linux/sensors/file_watch_sensor.py
```diff
@@ -42,6 +44,9 @@ def add_trigger(self, trigger):
         if not self._trigger:
             raise Exception('Trigger %s did not contain a ref.' % trigger)
 
+        # Wait a bit to avoid initialization race in logshipper library
+        eventlet.sleep(1.0)
```
Kami: There is an edge case in logshipper library.

If a trigger is added and ``add_file`` is called before the tailer is initialized, an exception will be thrown and the sensor will die / crash (this can happen because ``add_trigger`` method is called asynchronously and it could be called before ``run()`` and tailer fully initialized).

bigmstone: :(

Kami: Yeah, it's not ideal.

I could make it 100% deterministic by shuffling around the code (set ``initialized`` or similar variable to True inside ``run()`` and only proceed inside ``add_handler`` when that variable is True).

I think that's too complicated though since sleep seems to work just as fine (and it's one line of code).

It also only affects a single sensor right now so there is no point in handling that inside sensor container / base sensor class code.

PR #4497, opened 2019-01-15
Added include_attributes and exclude_attributes to GET /api/v1/executions/{id}

nmaludy:
Previously `GET /api/v1/executions/{id}` did not have the `include_attributes` or `exclude_attributes` query parameters, meaning it couldn't filter the output of large executions. This PR adds that ability.

@Kami are any other modifications necessary?

**TODO**
- [x] Tests

### DIFF
--- a/st2common/st2common/router.py
+++ b/st2common/st2common/router.py
@@ -666,7 +666,7 @@ class Router(object):
                 result.append(item)
         elif isinstance(data, dict):
             # get_one response
-            result = process_item(item)
+            result = process_item(data)
         else:
             raise ValueError('Unsupported type: %s' % (type(data)))
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,8 +15,8 @@ Added
   (``pack.yaml``). With this attribute pack declares which major Python versions it supports and
   works with (e.g. ``2`` and ``3``).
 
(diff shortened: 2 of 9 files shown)


st2common/st2common/router.py
```diff
@@ -669,7 +669,7 @@ def process_item(item):
                 result.append(item)
         elif isinstance(data, dict):
             # get_one response
-            result = process_item(item)
+            result = process_item(data)
```
Kami: Good catch. It looks we don't have test case for this scenario? :)

Kami: Besides that and test case for handling of those attributes it looks good to me :+1: 

I believe we already have a pattern for include and exclude attributes test cases. We might not just have one for "get one" API endpoints yet.

nmaludy:
*Please don't merge this yet*

I found a least one bug in `st2client` where it's using `?include_attributes` already, but doesn't ask for of the attributes that it needs.

So far i've found the following is broken:
- `st2 execution get xxx`

nmaludy:
Latest commit should address the `st2client` issues

Kami:
It looks like some CLI tests are failing. I didn't have time to dig in yet (if it's related to this change or not).

Kami:
I pushed a small change and I believe this should now work correctly.

Kami:
The diff got really large and messy so I needed to rebase this on top of master and force push to get it sorted out.

PR #4500, opened 2019-01-16
Update requests and virtualenv dependency

Kami:
This pull request updates ``requests`` and ``virtualenv`` dependency to the latest version.

We have been blocked on those upgrades for a long time for various reasons (library issues, RHEL 6 issue, etc.).

Actual upgrade will involve upgrading those libraries in many repos including (all pack repos, Makefiles, etc.).


LindsayHill:
_most_ Exchange packs have been updated to either use newer versions of `requests()`, or to not specify a version. So they should be OK now

Kami:
I will also open a separate PR for updating ``pip`` dependency - we pin pip to 9.x in many other repos and places as well so I will keep that change separate.

armab:
Addressed in https://github.com/StackStorm/st2/pull/4712

PR #4504, opened 2019-01-19
Various fixes so our code also works with Travis Xenial VM image

Kami:
This pull request updates travis config to use ubuntu xenial instead of (deprecated a long time ago) precise.

I'm just experimenting to see if this will affect tons of intermediate build failures we have been seeing in the last couple of days.


Kami:
As mentioned on Slack - we need to figure out a path going forward with Travis CI.

It has simply been too slow and unstable for us for the last half a year or so.

> [success] 10.60% tests.unit.test_scheduler.ActionExecutionSchedulingQueueItemDBTest.test_no_processing_of_non_requested_actions: 194.7348s

The same test case takes 14 seconds locally. We haven't done anything major in the code which should affect build times in such a negative way. Build times on Travis have been slowly growing from 15 minutes (spent a lot of time getting it to that number a half a year or so ago) to 50+ minutes now.

We don't have much visibility on what is going on Travis, but I would imagine they changed VM size to some tiny one for old VM / sudo required builds or something similar.

Kami:
Alright, I made changes to ``./scripts/travis/prepare-integration.sh`` so the integration build now passes.

There are still issues with intermediate failures and slow build times for unit tests which seem to be related to Travis environment (aka nothing we can do at the moment).

Kami:
@armab Yeah, I saw it, but I haven't got yet to figuring out why it's slower (still dealing with integration test failure).

One of the problems also is that precise has been deprecated by Travis quite a long time ago, luckily they haven't fully removed it yet though.

I will also test it with Trusty.

Kami:
@armab I agree.

As long as it's still supported / available on Travis I'm fine with using Precise if the builds indeed finish faster (as you have pointed out, we would still need to wait for the longest build).

There are still issues with using old version of various components and dependencies (e.g. OpenSSL which ships in Precise won't be supported in newer version of ``cryptography`` library anymore, etc).

Good news also is that now our code and Travis config also supports Xenial (just need to uncomment and change a couple of lines in Travis config) so if we are ever forced to switch to it, the switch will be easy and fast.

On a related note - I haven't been able to figure out why unit tests are slower but integration ones are faster on Xenial. VMs should have the same amount of resources available to them (RAM and CPU).

PR #4507, opened 2019-01-22
Add amqp package to st2common

bigmstone:
This PR adds the AMQP package to st2common. In a recent update to
py-amqp (https://github.com/celery/py-amqp/pull/214) there was an
additional check added that uses a rabbit capability that is not
available in the rabbit version shipped in RHEL 6. Even though we pin
the correct version of py-amqp in our fixed-requirements.txt the lack of
any package directly requiring it allowed it to be skipped when building
final requirements.txt. Adding the requirement to st2common will make
the correct pinned version show up in requirements.txt.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,6 @@
 # Don't edit this file. It's generated automatically!
 RandomWords
+amqp==2.3.2
 apscheduler==3.5.3
 argcomplete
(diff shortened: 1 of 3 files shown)


Kami:
Good catch re RHEL 6 issue :+1:

PR #4510, opened 2019-01-22
Upgrade pyyaml to the latest version

Kami:
This pull request updates pyyaml to the latest version.

Keep in mind that the security issue which has been addressed with a safer default doesn't affect any of our code since we already use ``yaml.safe_load`` everywhere, but it's still a good idea to upgrade to avoid pack content or similar using potentially unsafe ``yaml.load``.

Sadly ``flex`` library depends on the old version of pyyaml so I needed to fork it so we can upgrade pyyaml in this repo. I tried to submit the change upstream, but it looks like the repo is not maintained anymore - https://github.com/pipermerriam/flex.


blag:
Upstream PR is here: https://github.com/pipermerriam/flex/pull/213

Given that, I don't think we need our own fork anymore. 😄

blag:
Tweaked the PR to use PyYAML between `v4.2b` and `v5.1` (inclusive) and the current (`v6.13.2`) or the very next version (`v6.13.3`).

That way, when PyYAML 4.2b gets yanked from PyPI it doesn't break our build, and we don't have to revisit this when the flex PR gets merged and released to PyPI.

blag:
Yep, exactly.

PR #4512, opened 2019-01-23
Remove redundant action-aliases for pack install

### LINKED ISSUE #4511: Matching action alias for 'pack install xxx' returns an error
##### SUMMARY

Noticed by `@ravi` on Slack.

When running `st2 action-alias match 'pack install vsphere'` multiple action-aliases are matched, causing an error.

This is confusing in two ways: 
1) Running that action-alias should not be ambiguous
2) The help message for `st2 action-alias match` says it returns a list of matching aliases, but instead it either returns one or it errors.

##### ISSUE TYPE
 - Bug Report

##### STACKSTORM VERSION
```shell
$ st2 --version
st2 2.10.1, on Python 2.7.5
```

```shell
# OS
$ cat /etc/redhat-release 

# Install method
puppet-st2
```

##### STEPS TO REPRODUCE
```shell
$ st2 action-alias match 'pack install vsphere'
```

##### EXPECTED RESULTS
```shell
$ st2 action-alias match 'pack install vsphere'
+--------------+-----------------------------------+
| name         | description                       |
+--------------+-----------------------------------+
| pack_install | Install/upgrade StackStorm packs. |
+--------------+-----------------------------------+
```

##### ACTUAL RESULTS
```shell
$ st2 action-alias match 'pack install vsphere'
MESSAGE: Command 'pack install vsphere' matched more than 1 pattern for url: http://127.0.0.1:9101/v1/actionalias/match
```

This is also confusing because `st2 action-alias match --help` says it should return a list of matching aliases, when instead it either returns a single alias or it errors out if multiple are found.

```shell
$ st2 action-alias match --help
usage: st2 action-alias match [-h] [-t TOKEN] [--api-key API_KEY] [-j] [-y]
                              [-a ATTR [ATTR ...]] [-w WIDTH [WIDTH ...]]
                              command

Get the list of action aliases that match the command text.
```

##### INVESTIGATION

It appears that the action-alias `packs.pack_install` has redundant patterns defined: https://github.com/StackStorm/st2/blob/master/contrib/packs/aliases/pack_install.yaml#L7-L12

I think the simplest fix is to remove the redundant pattern and change the display text for the one that is left behind.

nmaludy:
Closes #4511 

This removes the redundant `pack install {{ packs }}` action-alias and instead replaces it with a single pattern.

After this fix, it is now possible to run the following:

```
$ st2 action-alias match 'pack install vsphere'
$ st2 action-alias execute -u st2admin 'pack install vsphere'
```

Also fixed the help text on `st2 action-alias match` to not imply that a list of matches is returned when in fact only a single match is returned.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,6 +23,15 @@ Changed
 
 * Changed the ``inquiries`` API path from ``/exp`` to ``/api/v1`` #4495
 
+Fixed
+~~~~~
+
+* Fixed the ``packs.pack_install`` / ``!pack install {{ packs }}`` action-alias to not have
+  redundant patterns. Previously this prevented it from being executed via
+  ``st2 action-alias execute 'pack install xxx'``. #4511
+
+  Contributed by Nick Maludy (Encore Technologies)
+  
 2.10.0 - December 13, 2018
 --------------------------
 
--- a/contrib/packs/aliases/pack_install.yaml
+++ b/contrib/packs/aliases/pack_install.yaml
@@ -4,10 +4,7 @@ action_ref: "packs.install"
 pack: "packs"
 description: "Install/upgrade StackStorm packs."
 formats:
-  - display: "pack install <pack>[,<pack>]"
-    representation:
-      - "pack install {{ packs }}"
-  - display: "pack install <gitUrl>[,<gitUrl>]"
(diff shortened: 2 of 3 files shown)


Kami:
Good catch :+1: 

This indeed looks like a "bug" (conflicting alias defined) in the action alias file.

Kami:
Merged, thanks.

Issue #4515, opened 2019-01-24
PIP version restrictions on packs test virtualenv


pkaramol:
This is more of a question on an issue preventing our packs development rather than a bug.

We are using the `st2-run-pack-tests.sh` script to run some tests on our packs.

Given that we are using a private pip repo (nexus) and we recently upgraded it, the later seems to have issues with older versions of `pip`.

What are the implications of loosening [this](https://github.com/StackStorm/st2/blob/master/st2common/bin/st2-run-pack-tests#L201) restriction and allowing for `pip==19.0.1` which is the latest?

Kami:
Thanks for opening this issue.

The plan is to get those dependencies updated in #4500 in the near future.

In the past we had a lot of issues with newer pip versions and older distributions (RHEL 6) so we needed to pin it to an older version to ensure it works correctly across all the various distributions we support.

pkaramol:
So more or less it is an issue of the OS/distro?

In your opinion if we explicitly update pip to latest (19.0.1) would it create st2-pack related issues for the tests run on the virtualenv of the script?

pkaramol:
Ok thx for the feedback, closing this.

PR #4516, opened 2019-01-24
Update pip to the latest version (19.x)

Kami:
We had issues in the past with versions of pip >= 10.x breaking our code and build process on some distros (RHEL 6, etc.) so we pinned pip to 9.x to avoid those issues.

It has been a long time since that so we should re-evaluate if we can upgrade pip to the latest version.

Related PRs / issues:

* https://github.com/StackStorm/st2/pull/4500
* https://github.com/StackStorm/st2/issues/4515


Kami:
@armab Yeah I think it will be a lot if "try and see / fix".

IIRC, we might also need to update our dh-virtualenv fork, need to check upstream changes (I did do it not too long ago).

armab:
Complements https://github.com/StackStorm/st2/pull/4712
`st2common/bin/st2-run-pack-tests` is the file I missed in previous PRs.

PR #4519, opened 2019-01-24
Don't submit metrics for any of the "get one" API endpoints

Kami:
This pull request updates instrumentation middleware so we don't submit any request metrics for all the "get one" API endpoints.

Previously we manually excluded some of the "most offending" API endpoints (get a single execution, etc.), but it's the safest to simply skip submitting metrics for all of "get one" API endpoints.

Submitting metrics for those endpoints results in too many unique metrics which is an anti-pattern and doesn't scale.


Kami:
I pushed another small (unrelated) fix for ``sys.prefix`` assertion.

In some scenarios (e.g. when ``launchdev.sh`` was started outside of tools directory), ``sys.prefix`` will be set to a relative path and assertion will trigger because ``get_python_lib`` function always returns an absolute path.

PR #4525, opened 2019-01-30
Update launchdev script to launch many schedulers and workflow engines

m4dcoder:
Add arguments to the launchdev scripts to launch more than one schedulers and workflow engines. By default, only one scheduler and one workflow engine is launched.


Kami:
I like that we still launch only one by default, because it makes per service debugging, etc. easier :+1:

Issue #4513, opened 2019-01-23
st2api encrypt of £ (and other utf-8 chars) fails with python stacktrace 


dswebbthg:
##### SUMMARY

Provide a quick summary of your bug or feature request.

##### ISSUE TYPE
Pick one below and delete the rest:
 - Bug Report

##### STACKSTORM VERSION
Paste the output of `st2 --version`:
root@stackstorm-ha-st2client-68fdcb57f-xnxkv:/opt/stackstorm# st2 --version
st2 3.0dev (9577ff944), on Python 3.6.7

##### OS / ENVIRONMENT / INSTALL METHOD
kubernetes using ubuntu 18.04 using helm

##### STEPS TO REPRODUCE

root@stackstorm-ha-st2client-68fdcb57f-xnxkv:/opt/stackstorm# st2 key set --scope=system --encrypt test '£'
ERROR: 400 Client Error: Bad Request
MESSAGE: The length of the provided data is not a multiple of the block length. for url: http://stackstorm-ha-st2api:9101/keys/test

During handling of the above exception, another exception occurred:
#### error log from the st2api container
2019-01-23 16:06:14,301 ERROR [-] Failed to call controller function "put" for operation "st2api.controllers.v1.keyvalue:key_value_pair_controller.put": The length of the provided data is not a multiple of the block length.
Traceback (most recent call last):
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2api/controllers/v1/keyvalue.py", line 301, in put
    kvp_db = KeyValuePairAPI.to_model(kvp)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2common/models/api/keyvalue.py", line 175, in to_model
    value = symmetric_encrypt(KeyValuePairAPI.crypto_key, value)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2common/util/crypto.py", line 186, in symmetric_encrypt
    return cryptography_symmetric_encrypt(encrypt_key=encrypt_key, plaintext=plaintext)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2common/util/crypto.py", line 237, in cryptography_symmetric_encrypt
    ciphertext_bytes = encryptor.update(data) + encryptor.finalize()
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/cryptography/hazmat/primitives/ciphers/base.py", line 159, in finalize
    data = self._ctx.finalize()
[...]

PR #4527, opened 2019-01-30
Encode plaintext before padding it

nickbaum:
WARNING: I'm no crypto expert, so please review with care.

I ran into this same issue when trying to encrypt utf-8 text:

https://github.com/StackStorm/st2/issues/4513

I believe the problem is that the plaintext is padded before it's encoded, and the encoding changes the length of the data.

Moving the encoding before the padding seems to resolve the issue.

### DIFF
--- a/st2common/st2common/util/crypto.py
+++ b/st2common/st2common/util/crypto.py
@@ -216,6 +216,10 @@ def cryptography_symmetric_encrypt(encrypt_key, plaintext):
     assert isinstance(aes_key_bytes, six.binary_type)
     assert isinstance(hmac_key_bytes, six.binary_type)
 
+    # Convert data to bytes
+    if isinstance(plaintext, (six.text_type, six.string_types)):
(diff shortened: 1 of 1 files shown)


nickbaum:
Hmm, nevermind. I believe this fails pull request works on Python 2 but not on Python 3, because pkcs5_pad can't concatenate str and bytes on Python 3:

https://github.com/StackStorm/st2/blob/b100848f937397aaef3ed9ce7a108b356381f849/st2common/st2common/util/crypto.py#L370

Any thoughts on how to address this?

Kami:
Thanks again for reporting this issue and working on it.

You were indeed on a right track. We need to convert data to bytes before we pad it - #4528 (and also update pad function so it handles bytes correctly).

Fix will be included in a future v2.10.2 release.

nickbaum:
🙌🏻Thanks for the quick follow up.

PR #4528, opened 2019-01-31
Fix "cryptography_symmetric_encrypt" function and make sure it handles unicode values correctly

### LINKED ISSUE #4513: st2api encrypt of £ (and other utf-8 chars) fails with python stacktrace

Kami:
This pull request fixes an issue reported in #4513.

We didn't correctly handle unicode values so encryption didn't work for unicode values which were not of a correct AES block length (aka when padding was needed).

This pull request fixes that by making sure we convert data to bytes before we pass it to the padding function. It also includes a change in pad function so it handles bytes correctly.

Thanks to @dswebbthg and @nickbaum for reporting and working on this issue.

Resolves #4513.

### DIFF
--- a/st2common/tests/unit/test_crypto_utils.py
+++ b/st2common/tests/unit/test_crypto_utils.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 # Licensed to the StackStorm, Inc ('StackStorm') under one or more
 # contributor license agreements.  See the NOTICE file distributed with
 # this work for additional information regarding copyright ownership.
@@ -52,6 +53,31 @@ class CryptoUtilsTestCase(TestCase):
     def setUpClass(cls):
         super(CryptoUtilsTestCase, cls).setUpClass()
         CryptoUtilsTestCase.test_crypto_key = AESKey.generate()
+
+    def test_symmetric_encrypt_decrypt_short_string_needs_to_be_padded(self):
+        original = u'a'
+        crypto = symmetric_encrypt(CryptoUtilsTestCase.test_crypto_key, original)
+        plain = symmetric_decrypt(CryptoUtilsTestCase.test_crypto_key, crypto)
+        self.assertEqual(plain, original)
+
+    def test_symmetric_encrypt_decrypt_utf8_character(self):
+        values = [
+            u'£',
+            u'£££',
+            u'££££££',
+            u'č š hello đ č p ž Ž',
+            u'hello 💩',
+            u'💩💩💩💩💩'
+            u'💩💩💩',
+            u'💩😁'
+        ]
+
+        for index, original in enumerate(values):
+            crypto = symmetric_encrypt(CryptoUtilsTestCase.test_crypto_key, original)
+            plain = symmetric_decrypt(CryptoUtilsTestCase.test_crypto_key, crypto)
+            self.assertEqual(plain, original)
+
+        self.assertEqual(index, (len(values) - 1))
 
     def test_symmetric_encrypt_decrypt(self):
         original = 'secret'
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -31,7 +31,11 @@ Fixed
   ``st2 action-alias execute 'pack install xxx'``. #4511
 
   Contributed by Nick Maludy (Encore Technologies)
-  
+* Fix datastore value encryption and make sure it also works correctly for unicode (non-ascii)
+  values.
+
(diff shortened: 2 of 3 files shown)


st2common/tests/unit/test_crypto_utils.py
```diff
@@ -53,6 +54,31 @@ def setUpClass(cls):
...
 
+    def test_symmetric_encrypt_decrypt_short_string_needs_to_be_padded(self):
+        original = u'a'
+        crypto = symmetric_encrypt(CryptoUtilsTestCase.test_crypto_key, original)
+        plain = symmetric_decrypt(CryptoUtilsTestCase.test_crypto_key, crypto)
+        self.assertEqual(plain, original)
+
+    def test_symmetric_encrypt_decrypt_utf8_character(self):
+        values = [
+            u'£',
```
blag: I don't think the `u` syntax is supported in Python 3. Better to use [`six.u()`](https://pythonhosted.org/six/#six.u) so we can run our tests with Python 3.

blag:
```python
    if isinstance(plaintext, (six.text_type, six.string_types)):
        # Convert data to bytes
        data = plaintext.encode('utf-8')
```

This means that if I encrypt a string, I would get back bytes when I decrypt it, right? That's a little weird, but it's the previous behavior.

st2common/tests/unit/test_crypto_utils.py
Kami: It is - `u` as in prefix for strings which is also a default (e.g. `u'foo`).

Those tests also run under Python 2.x and 3.6 on Travis and I also ran them under both versions locally.

Kami:
@blag When you decrypt a string you always get back unicode - https://github.com/StackStorm/st2/blob/35de7b64027e47849deb038e813ea399faa18104/st2common/st2common/util/crypto.py#L302.

Also see tests - they perform encrypt and decrypt round-trip and decrypted value is the same as the original one (unicode string).

st2common/tests/unit/test_crypto_utils.py
blag: Awesome, good to know.

PR #4531, opened 2019-02-04
Fix a bug with "{{ config_text.foo }}" not working for executions scheduled via Orquesta workflow

Kami:
This pull request fixes a bug with ``{{ confix_context.<config item name> }}`` notation not working for action parameter default values for action executions which are scheduled via Orquesta (aka part of an Orquesta workflow).

### Background and Details

I was working on some st2community pack changes and improvements and while testing and developing the workflow I noticed "{{ config_context }}" notation doesn't work for executions scheduled via Orquesta (e.g. https://github.com/StackStorm/stackstorm-st2community/pull/1/files#diff-3c8d6de4c078da34b194a061826b147dR13), but it works just fine if those actions are run outside of the Orquesta workflow context.

After some digging in, I found orquesta runner doesn't add ``pack`` value to the execution context so ``get_config`` method will always return an empty dict because pack will be set to ``None``.

### TODO

- [x] Tests

### DIFF
--- a/st2common/st2common/util/config_loader.py
+++ b/st2common/st2common/util/config_loader.py
@@ -224,7 +224,7 @@ class ContentPackConfigLoader(object):
 def get_config(pack, user):
     """Returns config for given pack and user.
     """
-    LOG.debug('Attempting to get config')
+    LOG.debug('Attempting to get config for pack "%s" and user "%s"' % (pack, user))
     if pack and user:
         LOG.debug('Pack and user found. Loading config.')
         config_loader = ContentPackConfigLoader(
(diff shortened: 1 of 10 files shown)


st2common/st2common/util/config_loader.py
```diff
@@ -224,7 +224,7 @@ def _get_datastore_value_for_expression(self, key, value, config_schema_item=Non
 def get_config(pack, user):
     """Returns config for given pack and user.
     """
-    LOG.debug('Attempting to get config')
+    LOG.debug('Attempting to get config for pack "%s" and user "%s"' % (pack, user))
```
Kami: Originally this log message was not all that useful because it didn't include pack and username so you weren't able to spot ``pack`` is ``None`` just by looking at the log messages.

PR #4534, opened 2019-02-05
UTF-8 encode both body and subject of email

johandahlberg:
It seems that perhaps it is necessary to change the charset to UTF-8 in two places to get the body correctly encoded (exactly why this is so is a little bit unclear to me, I solved this by trail and error). I also added that the subject should be utf-8 encoded, to make sure that non-ascii characters are allowed there.

### DIFF
--- a/contrib/core/actions/send_mail/send_mail
+++ b/contrib/core/actions/send_mail/send_mail
@@ -52,7 +52,7 @@ if [[ -z $trimmed && $SEND_EMPTY_BODY -eq 1 ]] || [[ -n $trimmed ]]; then
 cat <<EOF
 TO: ${TO}
 FROM: ${FROM}
-SUBJECT: =?UTF-8?q?${SUBJECT}?=
+SUBJECT: =?UTF-8?B?$(echo ${SUBJECT} | base64)?=
 EOF
 
 if [[ -n $ATTACHMENTS ]]; then
--- a/contrib/core/actions/send_mail/send_mail
+++ b/contrib/core/actions/send_mail/send_mail
@@ -52,7 +52,7 @@ if [[ -z $trimmed && $SEND_EMPTY_BODY -eq 1 ]] || [[ -n $trimmed ]]; then
 cat <<EOF
 TO: ${TO}
 FROM: ${FROM}
-SUBJECT: ${SUBJECT}
+SUBJECT: =?UTF-8?q?${SUBJECT}?=
 EOF
 
 if [[ -n $ATTACHMENTS ]]; then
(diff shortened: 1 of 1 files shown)


Kami:
Thanks. I will have a look and merge those changes directly into my branch.

contrib/core/actions/send_mail/send_mail
```diff
@@ -83,7 +83,7 @@ EOF
 
   cat <<EOF
 --${boundary}
-Content-Type: ${CONTENT_TYPE}
+Content-Type: ${CONTENT_TYPE}; charset=UTF-8
```
Kami: I didn't add it here because I thought we don't need it here. This section refers to attachments which should alrady be base64 encoded so it should only contain ascii characters.

Having said that, if it works fine and has no negative side affects, I'm fine with keeping it here.

contrib/core/actions/send_mail/send_mail
```diff
@@ -52,7 +52,7 @@ if [[ -z $trimmed && $SEND_EMPTY_BODY -eq 1 ]] || [[ -n $trimmed ]]; then
 cat <<EOF
 TO: ${TO}
 FROM: ${FROM}
-SUBJECT: ${SUBJECT}
+SUBJECT: =?UTF-8?q?${SUBJECT}?=
```
Kami: I assume this works fine with all content types - text/html (default) and text/plain?

contrib/core/actions/send_mail/send_mail
johandahlberg: I agree that it is a bit odd, but when I did not include this, I got garbled non-ascii characters in the e-mail. Not sure why though.

contrib/core/actions/send_mail/send_mail
johandahlberg: If I understand it correctly based on this, https://tools.ietf.org/html/rfc1342, the encoding of the header is separate from the encoding of the message (which is why this needs to be specified separately). I've tested `text/plain` and 'text/html' and both seem to work fine.

Kami: It looks like we might want to base64 encode the subject and change `q` to `B`.

`q` specifies subject is encoded using ``quoted-printable`` encoding which is not the case so it probably won't work correctly.

johandahlberg: You're right. I've updated the PR here.

contrib/core/actions/send_mail/send_mail
Kami: Edit: I was wrong, this part also refers to the html version of the body. The section above refers to the attachments so it's correct to have charset here as well.

Kami:
LGTM, thanks :+1:

PR #4536, opened 2019-02-05
Add some metric instrumentation to the notifer service code

Kami:
This pull request adds some metrics instrumentation to the ``st2notifier`` service code.

We don't have any now so it would be good to have some. Especially because we noticed some "interesting" CPU spikes in the notifier service on our CICD server (could be related to blocking Jinja template rendering which maybe we can offload to a thread pool to avoid blocking main thread and reduce the main process CPU usage).

## TODO

- [x] Document those new metrics in the docs


Kami:
Docs at https://github.com/StackStorm/st2docs/pull/846.

Kami:
@armab Yeah that one is a bit of an exception, we usually don't need such granular metrics, but this one is a bit special.

I hope we can remove it once we get a better picture (it shouldn't cause any overhead and it's not documented as a public exposed metric either).

PR #4537, opened 2019-02-06
Move lock for concurrency policies into scheduler

m4dcoder:
Move the lock for coordinating concurrency policies into the scheduler. With the current approach, when there are more than one schedulers, there is a race in scheduling that results in failure to enforce the concurrency accurately.


CHANGELOG.rst
```diff
@@ -22,6 +22,7 @@ Changed
 ~~~~~~~
 
 * Changed the ``inquiries`` API path from ``/exp`` to ``/api/v1`` #4495
+* Moved the lock from concurrency policies into the scheduler. #4481 (bug fix)
```
Kami: Please also clarify in the changelog entry what bug it fixes. Otherwise if people go over the changelog they will have no idea what this change doesn't and if it affects them or not.

st2common/st2common/constants/policy.py
```diff
@@ -0,0 +1,25 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+__all__ = [
+    'POLICY_TYPES_REQUIRING_LOCK'
+]
+
+# Concurrency policies require scheduler to acquire a distributed lock to prevent race
+# in scheduling when there are multiple scheduler instances.
+POLICY_TYPES_REQUIRING_LOCK = [
```
Kami: :+1:

st2common/st2common/services/policies.py
```diff
@@ -24,6 +24,20 @@
...
+def has_policies(lv_ac_db, policy_types=None):
+    query_params = {
+        'resource_ref': lv_ac_db.action,
+        'enabled': True
+    }
+
+    if policy_types:
+        query_params['policy_type__in'] = policy_types
+
+    policy_dbs = pc_db_access.Policy.query(**query_params)
```
Kami: Adding ``.count()`` to the end would probably be a bit more efficient since the count will be calculated and returned server side and means we don't need to evaluate and load the whole result set in memory like we do if we use ``len()``.

Not a huge issue here since those documents are not large, but still an easy change :)

Kami:
Thanks for working on this change, LGTM :+1: 

On a related note - would it some how be possible for us to write end to end / integration tests which actually try to emulate the race and verify it's not there (probably quite hard to do end to end wise)?

Maybe spawn two scheduler process as part of an integration test?

st2common/st2common/services/policies.py
m4dcoder: Fixed

CHANGELOG.rst
m4dcoder: Fixed

st2common/st2common/services/coordination.py
```diff
@@ -180,46 +180,3 @@ def get_coordinator():
         COORDINATOR = coordinator_setup()
 
     return COORDINATOR
-
```
Kami: I assume this commented out code will be removed?

m4dcoder: Yes, it will be removed.

PR #4538, opened 2019-02-06
Don't include log messages with log level "AUDIT" in regular service log file

### LINKED ISSUE #4502: Ability to filter out AUDIT logs
##### SUMMARY

Currently, the log files `/var/log/st2/st2*.log` contain all of the AUDIT log information as well as the INFO and DEBUG information. These AUDIT logs are duplicated in the `/var/log/st2/st2*.audit.log` files.

In our case the `/var/log/st2/st2scheduler.log` and `/var/log/st2/st2scheduler.log` are both 5GB in size (10GB total) in less than 24h of runtime.

We would like the ability to disable sending AUDIT logs to the `/var/log/st2/st2*.log` files since the information is duplicated anyways.

##### ISSUE TYPE
 - Feature Idea

##### STACKSTORM VERSION
```
$ st2 --version
st2 2.10.1, on Python 2.7.5
```

```
$ uname -a
Linux nor1devssd01.dev.encore.tech 3.10.0-957.1.3.el7.x86_64 #1 SMP Thu Nov 15 17:36:42 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

$ cat /etc/redhat-release 

Install method = puppet-st2
```

##### STEPS TO REPRODUCE
```
$ grep "AUDIT action" st2scheduler.log | wc -l
1447
```

##### EXPECTED RESULTS
```
$ grep "AUDIT action" st2scheduler.log | wc -l
0
```

##### ACTUAL RESULTS

I think the "chattyness" of st2scheduler could be turned down as well. Not sure the entire action and results need to be printed every time a scheduling change occurs? (maybe it does?)

Kami:
This pull request resolves an issue with log messages with log level AUDIT ending up in regular service log file (``<service>.log``) in addition to the service audit log file (``<service>.audit.log``).

## Background

We implemented ``AUDIT`` log level by adding a new log level constant which contains the highest log level number (``CRITICAL`` + 10).

Default Python log level filtering works using "include log messages with log level <set> and above" approach. This means messages with ``AUDIT`` log level will always be included in the logs because ``AUDIT`` is the highest level.

In most cases that's not desired. That's especially true for production deployment where default log level is ``INFO`` and we have dedicated log file with audit log messages. We don't want audit log messages to also end up in the regular service log file where they just cause noise.

Implementation wise, I needed to implement it using a log filter - we exclude log messages with level ``AUDIT`` if log level is set to ``INFO`` or higher, but not to ``AUDIT``. I couldn't make a simpler change like setting ``AUDIT`` log level constant to lower than ``INFO``, because then we would have the opposite problem (non ``AUDIT`` messages would also end up in the audit log file).

NOTE: To aid with the debugging, ``AUDIT`` log messages are also included in a regular service log file if log level is set to ``DEBUG`` of if ``system.debug`` config option is set to ``True``. When debug is enabled there are many more log messages coming from other places so ``AUDIT`` should be the least of your concerns :)

Resolves #4502 (thanks to @nmaludy for reporting it).

## TODO

- [x] Integration tests

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,12 +23,14 @@ Changed
 
 * Changed the ``inquiries`` API path from ``/exp`` to ``/api/v1``. #4495
 * Update logging code so we exclude log messages with log level ``AUDIT`` from a default service
-  log file. Log messages with level ``AUDIT`` are already logged in a dedicated
-  ``<service>.audit.log`` file so there is no need for them to be duplicated in regular service log
-  file.
+  log file (e.g. ``st2api.log``). Log messages with level ``AUDIT`` are already logged in a
+  dedicated service audit log file (e.g. ``st2api.audit.log``) so there is no need for them to also
+  be duplicated and included in regular service log file.
 
   NOTE: To aid with debugging, audit log messages are also included in a regular log file when log
-  level is set to DEBUG. (improvement)
+  level is set to ``DEBUG`` or ``system.debug`` config option is set to ``True``.
+
+  Reported by Nick Maludy. (improvement) #4538 #4502
 
 Fixed
 ~~~~~
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -21,7 +21,14 @@ Added
 Changed
 ~~~~~~~
 
-* Changed the ``inquiries`` API path from ``/exp`` to ``/api/v1`` #4495
+* Changed the ``inquiries`` API path from ``/exp`` to ``/api/v1``. #4495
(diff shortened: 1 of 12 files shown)


Kami:
Added integration tests in 11111ffd5f126e252537114c5ce550f64d783659.

Kami:
Randomly found out just now while debugging some other things that this doesn't work correctly on production gunicorn deployments.

I believe issue is related to another change, I'm digging in.

Kami:
^Second time is the charm - #4621.

The code was correct, but we were using incorrect logging config for API services with packages so those log messages weren't filtered out by default.

PR #4540, opened 2019-02-08
Fix accidental regression in pylint Makefile check

Kami:
It looks like a couple of lines were accidentely removed in dc2617040d637fb4857b16ac1c40d9ef60c71830 and since then we didn't perform any pylint checks on StackStorm components and runners code :man_facepalming:.


Kami:
Per discussion with @LindsayHill on Slack - we should also enable some other non-error checks like "literal-comparison", etc.

The problem is that right now we only enable "error" checks (there is too much noise and false positives with many non fatal error messages).

If we want to enable a  single "literal-comparison" check which is part of warning / refactoring category we need to enable all other warnings and refactor messages as well and then explicitly disable all that we are not interested in.

This will take a wile (exclude list will be long) and I will do it in a different PR in the future.

m4dcoder:
LGTM but I guess you need to get all the CIs to pass here.

PR #4541, opened 2019-02-08
Support for SSL client cert authentication and server certificate validation when connecting to RabbitMQ

Kami:
This pull request adds two of the features we were missing for fully secure and verified connections to RabbitMQ message bus:

* Support for client based certificate authentication
* Support for verifying server certificate on the client side with the provided CA cert (bundle)

Both of those features were already supported for database (MongoDB) connection, but not for RabbitMQ.

Keep in mind that secure SSL / TLS connections to RabbitMQ were already possible (without verifying server certificate), but that wasn't documented anywhere. https://github.com/StackStorm/st2docs/pull/848 fixed that.

Nice thing is that we can use same naming for those options for ``database`` and ``messaging`` section (``ssl_keyfile``, ``ssl_certfile``, ``ssl_cert_reqes``, ``ssl_ca_certs``).

Example config entries:

```ini
[messaging]
url = amqp://guest:guest@127.0.0.1:5671/
ssl = True
ssl_keyfile = /home/vagrant/client/private_key.pem
ssl_certfile = /home/vagrant/client/client_certificate.pem
ssl_ca_certs = /home/vagrant/testca/ca_certificate_bundle.pem
ssl_cert_reqs = required
```

Implementation wise, I needed to refactor some code so we now retrieve a properly configured ``Connection`` object in a single centralized place.

## TODO

- [x] Regenerate sample config
- [x] Update documentation - https://github.com/StackStorm/st2docs/pull/850
- [x] Tests


st2common/st2common/transport/publishers.py
```diff
@@ -92,13 +109,14 @@ def get_publisher(self, urls):
...
-    def __init__(self, urls, exchange):
+    def __init__(self, exchange, urls=None):
+        urls = urls or transport_utils.get_messaging_urls()
         self._publisher = SharedPoolPublishers().get_publisher(urls=urls)
         self._exchange = exchange
 
     def publish_create(self, payload):
         with Timer(key='amqp.publish.create'):
-            self._publisher.publish(payload, self._exchange, CREATE_RK)
+            self._publisher.publgish(payload, self._exchange, CREATE_RK)
```
armab: Typo

Kami:
@armab @warrenvw Do we already handle CA authority and SSL certificate generation and management somewhere in our kubernetes-ha work?

If we do, it would also be good to eventually utilize it for MongoDB and RabbitMQ connections.

st2common/st2common/transport/announcement.py
```diff
@@ -20,16 +20,20 @@
...
 
+__all__ = [
+    'AnnouncementPublisher',
+    'AnnouncementDispatcher'
+]
+
 
 class AnnouncementPublisher(object):
-    def __init__(self, urls):
+    def __init__(self, urls=None):
```
Kami: Connection retrieval and URL handling is now centralized in a single place so there is really no need for this ``url`` argument here anymore.

To make the code nicer, I could simply remove it from all those classes.

armab:
@Kami No, we don't auto-generate SSL/CA in K8s/HA work, leaving that part for user consideration/configuration.

tools/config_gen.py
```diff
@@ -41,7 +41,7 @@
 
 SKIP_GROUPS = ['api_pecan', 'rbac', 'results_tracker']
 
-# We group auth options together to nake it a bit more clear what applies where
```
armab: ROFL :smiley:

Kami:
I pushed changes so we now also expose SSL listener when running RabbitMQ on Travis (67fc03f8ed2b672c44c290d8dcc0b26ffe8ded31, 3b7744939e4b7adbb69762b1ef0ac7f0befbbcb1, 922cb23c2d3480849819ea6eb94d7191793c12ff) and added some basic tests which verify that those parameters work as expected (88a6d575f179aba684857b46af2da419579902cf).

In the future, we should do the same for MongoDB (we don't have any integration tests for SSL related parameters, just unit ones which are lacking).

NOTE: At the moment we only run those tests on Travis, because RabbitMQ SSL listener is not set up on local vagrant dev VMs.

PR #4542, opened 2019-02-08
Fix CLI 'st2 apikey load' idempotence and API '/api/v1/apikeys' not honoring 'ID' for the new record creation

armab:
We've had this problem before and it seems a regression:
- https://github.com/StackStorm/st2/pull/3786
- https://github.com/StackStorm/st2/pull/3748

### Reproduce
```
$ st2 apikey list --detail --show-secrets --yaml > apikeys.yaml
$ st2 apikey delete ...
$ st2 apikey load apikeys.yaml
$ st2 apikey load apikeys.yaml
```

### Description
At a high level PR fixes `st2api key load` that wasn't idempotent and failing to import same record twice with the following error:
```
HTTPError: 409 Client Error: Conflict
MESSAGE: Tried to save duplicate unique keys (E11000 duplicate key error collection: st2.api_key_d_b index: key_hash_1 dup key: { : "903b7f0761094265969761443f4f25fc87c780b6248c38ce8dbaadfa52a2b53fd72381ac141072cb53ae78ff02e16fdfc0e40001472b9f25585a7b2864c501db" }) for url: http://127.0.0.1:9101/v1/apikeys
```

### Investigation
Looking at `st2api key load` behavior, we do several requests to guarantee command idempotence:
1) Try to check if ST2 API key ID exists
2) If inexistent, - create a new record with needed ID

However result of request `2.` goes wrong: 
```
# -------- begin 140602007925392 request ----------
curl -X POST -H  'Connection: keep-alive' -H  'Accept-Encoding: gzip, deflate' -H  'Accept: */*' -H  'User-Agent: python-requests/2.14.2' -H  'content-type: application/json' -H  'X-Auth-Token: 80fdee5a251c40ac9599afa889dbdce7' -H  'Content-Length: 231' --data-binary '{"key_hash": "903b7f0761094265969761443f4f25fc87c780b6248c38ce8dbaadfa52a2b53fd72381ac141072cb53ae78ff02e16fdfc0e40001472b9f25585a7b2864c501db", "metadata": {}, "enabled": true, "user": "st2admin", "id": "58e3f3330c0517062a3fda43"}' http://127.0.0.1:9101/v1/apikeys
# -------- begin 140602007925392 response ----------
{
    "uid": "api_key:903b7f0761094265969761443f4f25fc87c780b6248c38ce8dbaadfa52a2b53fd72381ac141072cb53ae78ff02e16fdfc0e40001472b9f25585a7b2864c501db",
    "created_at": "2019-02-08T19:50:15.500100Z",
    "enabled": true,
    "user": "st2admin",
    "key": null,
    "id": "5c5ddd776cb8de530e0a1391",
    "metadata": {}
}
# -------- end 140602007925392 response ------------

```
Instead of creating a record with the requested `ID`, it creates a new one with random `ID` and so next import `st2 apikey load` for the same data fails.

This leads to a root cause bug fixed in this PR: API `POST /api/v1/apikeys` wasn't creating new record with the provided `ID`: https://api.stackstorm.com/api/v1/apikeys/#/api_key_controller.post

----

Was spot during the https://github.com/StackStorm/stackstorm-ha/pull/45 work.


Kami:
Good catch :+1: 

Can we please also add a test case for it to avoid regression again in the future? :)

armab:
Right, that's the plan per `status:missing tests` tag.

PR #4543, opened 2019-02-11
Refactor scheduler process to exit properly

### LINKED ISSUE #4539: Executions stuck in requested state
##### SUMMARY

Earlier this week we noticed that our instance had a bunch of executions stuck in the `requested` state and none were ever being executed. Restarting the host resolved the problem.

##### ISSUE TYPE
 - Bug Report

##### STACKSTORM VERSION
```
$ st2 --version
st2 2.10.1, on Python 2.7.5
```

```
$ cat /etc/redhat-release

# install method: puppet-st2

# single host installed "all in one" with NO MongoDB clustering/replica-set
# the MongoDB instance is a single instance that lives on the same host as StackStorm
```

##### STEPS TO REPRODUCE
We've been unable to come up with steps that reproduce the issue consistently. It has been noticed, usually over night, that cron rules were not producing the expected output (emails and tickets). These executions were stuck in the `requested` state.

##### EXPECTED RESULTS
Executions run when invoked.

##### Logs
As stated above, restarted the host fixed the issue. Unfortunately the engineer who found the issue didn't debug any further and try restarting individual services or pin point which service was the issue.

Post-mortem, we found the following in the `/var/log/st2/st2scheduler.log`

```

2019-02-04 14:48:45,229 140516293681552 INFO scheduler [-] (PID=8536) Scheduler started.
2019-02-04 14:48:45,418 140516293681552 INFO consumers [-] Starting SchedulerEntrypoint...
2019-02-04 14:48:45,429 140516109106224 INFO mixins [-] Connected to amqp://xxx:**@127.0.0.1:5672//
2019-02-04 14:51:05,599 140516198248688 ERROR traceback [-] Traceback (most recent call last):

2019-02-04 14:51:05,599 140516198248688 ERROR traceback [-]   File "/opt/stackstorm/st2/lib/python2.7/site-packages/eventlet/hubs/poll.py", line 109, in wait

2019-02-04 14:51:05,600 140516198248688 ERROR traceback [-]     listener.cb(fileno)

2019-02-04 14:51:05,600 140516198248688 ERROR traceback [-]   File "/opt/stackstorm/st2/lib/python2.7/site-packages/eventlet/greenthread.py", line 219, in main

2019-02-04 14:51:05,600 140516198248688 ERROR traceback [-]     result = function(*args, **kwargs)

2019-02-04 14:51:05,600 140516198248688 ERROR traceback [-]   File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2actions/scheduler/handler.py", line 67, in run

2019-02-04 14:51:05,601 140516198248688 ERROR traceback [-]     execution_queue_item_db = self._get_next_execution()

2019-02-04 14:51:05,601 140516198248688 ERROR traceback [-]   File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2actions/scheduler/handler.py", line 125, in _get_next_execution

2019-02-04 14:51:05,601 140516198248688 ERROR traceback [-]     execution_queue_item_db = ActionExecutionSchedulingQueue.query(**query).first()

2019-02-04 14:51:05,601 140516198248688 ERROR traceback [-]   File "/opt/stackstorm/st2/lib/python2.7/site-packages/mongoengine/queryset/base.py", line 294, in first

2019-02-04 14:51:05,602 140516198248688 ERROR traceback [-]     result = queryset[0]

2019-02-04 14:51:05,602 140516198248688 ERROR traceback [-]   File "/opt/stackstorm/st2/lib/python2.7/site-packages/mongoengine/queryset/base.py", line 194, in __getitem__

2019-02-04 14:51:05,602 140516198248688 ERROR traceback [-]     queryset._cursor[key],

2019-02-04 14:51:05,602 140516198248688 ERROR traceback [-]   File "/opt/stackstorm/st2/lib/python2.7/site-packages/pymongo/cursor.py", line 610, in __getitem__

2019-02-04 14:51:05,603 140516198248688 ERROR traceback [-]     for doc in clone:

2019-02-04 14:51:05,603 140516198248688 ERROR traceback [-]   File "/opt/stackstorm/st2/lib/python2.7/site-packages/pymongo/cursor.py", line 1189, in next

2019-02-04 14:51:05,603 140516198248688 ERROR traceback [-]     if len(self.__data) or self._refresh():

2019-02-04 14:51:05,603 140516198248688 ERROR traceback [-]   File "/opt/stackstorm/st2/lib/python2.7/site-packages/pymongo/cursor.py", line 1104, in _refresh

2019-02-04 14:51:05,603 140516198248688 ERROR traceback [-]     self.__send_message(q)

2019-02-04 14:51:05,604 140516198248688 ERROR traceback [-]   File "/opt/stackstorm/st2/lib/python2.7/site-packages/pymongo/cursor.py", line 982, in __send_message

2019-02-04 14:51:05,604 140516198248688 ERROR traceback [-]     helpers._check_command_response(first)

2019-02-04 14:51:05,604 140516198248688 ERROR traceback [-]   File "/opt/stackstorm/st2/lib/python2.7/site-packages/pymongo/helpers.py", line 132, in _check_command_response

2019-02-04 14:51:05,604 140516198248688 ERROR traceback [-]     raise NotMasterError(errmsg, response)

2019-02-04 14:51:05,604 140516198248688 ERROR traceback [-] NotMasterError: interrupted at shutdown

2019-02-04 14:51:05,605 140516198248688 ERROR hub [-] Removing descriptor: 9

2019-02-05 03:28:30,413 140516198248688 INFO misc [-] Re-opening log file "/var/log/st2/st2scheduler.log" with mode "a"

2019-02-05 03:28:30,422 140516198248688 INFO misc [-] Re-opening log file "/var/log/st2/st2scheduler.audit.log" with mode "a"
```

It appears that the `st2scheduler` encountered an exception and the greenthread died and caused the service to no longer function.

In the next log file (after rotation) you can see when we fixed the issue:

```
2019-02-05 03:28:31,993 140516198248688 INFO misc [-] Re-opening log file "/var/log/st2/st2scheduler.log" with mode "a"

2019-02-05 03:28:31,996 140516198248688 INFO misc [-] Re-opening log file "/var/log/st2/st2scheduler.audit.log" with mode "a"

2019-02-05 10:29:53,767 140516293681552 INFO scheduler [-] (PID=8536) Scheduler stopped.
2019-02-05 10:29:53,767 140516293681552 INFO consumers [-] Shutting down SchedulerEntrypoint...
2019-02-05 10:31:30,168 139946447700368 INFO __init__ [-] Connecting to database "st2" @ "127.0.0.1:27017" as user "xxx".
2019-02-05 10:31:30,769 139946447700368 INFO __init__ [-] Successfully connected to database "st2" @ "127.0.0.1:27017" as user "xxx".
2019-02-05 10:31:46,988 139946447700368 AUDIT runnersregistrar [-] RunnerType updated. RunnerType RunnerTypeDB(d
```

m4dcoder:
Add retries in the scheduler handler to temporarily handle DB connection failures. Refactor how threads exit for the process to return proper code.  Fixes #4539


st2actions/st2actions/scheduler/handler.py
```diff
@@ -60,25 +62,37 @@ def __init__(self):
...
 
     def cleanup(self):
-        LOG.debug('Starting scheduler garbage collection')
+        LOG.debug('Starting scheduler garbage collection...')
 
         while not self._shutdown:
             eventlet.greenthread.sleep(cfg.CONF.scheduler.gc_interval)
             self._handle_garbage_collection()
 
+    @retrying.retry(
```
Kami: I'm personally still not too sure about this retry here. It's seems like a one off - aka we don't do it in other similar services.

Kami: Besides that, the linking looks good to me, let's please just add some test cases for it.

Kami:
I added some test cases in 442c57e1f.

It wasn't totally trivial to write them because we need to emulate async nature of those services and throw an exception in a specific place (``process()`` method).

If we don't emulate that async nature and exception will be thrown inside ``start()`` and not async after blocking ``wait()`` is called, it will look like the service is working and exiting correctly, but it's actually not.

Those tests, ``test_service_exits_correctly_on_fatal_exception_in_entrypoint_process`` specifically uncover an issue with the code, aka the change doesn't fix the whole issue.

Problem lies in this line of code - https://github.com/StackStorm/st2/pull/4543/files#diff-4d1c13310fc6ebda8f5e5d2ec414c2f9R71.

Doing ``handler.wait() or entrypoint.wait()`` means the process will only exit if ``handler.wait()`` throws. If only entrypoint throws, we will still be waiting on ``handler.wait()`` and service will still be running with one of the process being dead (entrypoint specifically).

In short - the same original issue still exists.

You can also replicate the issue manually in the same manner by adding ``raise Exception('')`` inside ``SchedulerEntrypoint.process()`` method, running the scheduler manually and scheduling an action execution.

Again, it's important you add it there. If you add it inside start(), run() or similar the exception will be correctly propagated and process will exit because that happens before the blocking ``handler.wait() or entrypoint.wait()``line is called.

Kami:
As far as the correct and best solution goes, I'm not too sure about that yet.

There are multiple approaches, but all of them will require us to refactor cmd entry point and use a non-standard code there.

Something along the lines of:

```python
        handler_wait_thread = eventlet.spawn(handler.wait)
        entrypoint_wait_thread = eventlet.spawn(entrypoint.wait)
        handler_wait_thread.link(exit_on_error)
        handler_wait_thread.link(exit_on_error)

# Sleep until the service exits on exit_on_error
while True:
    eventlet.sleep(0.1)
```

Should work. This means we need to perform logging and all the clean up (calling shutdown, teardown, etc.) in that ``exit_on_error`` function though (and we also need to handle CTRL+C, etc.).

On a related note - we really need to be careful with such code pattern - blocking wait on two independent green threads inside the service entry point.

If we use it in other places, it will exhibit the same issue and edge cases so we need to audit the code for such patterns and fix it if we also find it in some other place.

st2actions/st2actions/scheduler/handler.py
m4dcoder: I think having the retries there is ok.  For a single server install w/o any complex service management, this prevents temporary hiccups with MongoDB connection.  If users want to fail fast, they can reconfigure the retries.  If you're worry about consistency, we can revisit.  After this issue, our service pattern need an overhaul.

m4dcoder:
Yeah.  That's why I commented that the handler.wait needs to be evaluated first since entrypoint is more durable.  I agree, eventlets give us very little option to wait on multiple threads.  Either we use `link` throughout to signal all other threads to exit or we find another option which signals us if any thread exits (i.e. gevent.wait with count=1)

Kami:
Let's please add a changelog entry. Besides that, LGTM :+1:

PR #4544, opened 2019-02-11
[DONT MERGE] Benchmark st2 build during performance trial

warrenvw:
Test build performance with the following changes:

 - [x] Use `docker_layer_caching` instead of `reusable: true`.
 - [x] Remove https://discuss.circleci.com/t/setup-remote-docker-connection-failures/26434 since that link is no longer of any value. The pertinent links now return 404.
 - [x] Use `resource_class: small`.


.circleci/config.yml
```diff
@@ -125,10 +124,7 @@ jobs:
     steps:
       - checkout
       - setup_remote_docker:
-          reusable: true    # default - false
-          exclusive: true   # default - true
-          # Temporary workaround for Circle CI issue
-          # https://discuss.circleci.com/t/setup-remote-docker-connection-failures/26434
+          docker_layer_caching: false     # default - false
           version: 18.05.0-ce
```
armab:
```suggestion
```
That comment relates to this line of code which I guess we can try to remove now.

PR #4546, opened 2019-02-12
[WIP] Use Xenial image on Travis

Kami:
Just testing again if we can figure out why tests run slower on Travis when using Xenial image.

Problem with using Precise is that it runs ancient versions of software such as OpenSSL, RabbitMQ, etc. and we can't upgrade and as such, we can't run some tests such as RabbitMQ SSL ones, etc.


Kami:
Closing, for now.

Will re-open when it becomes relevant again (aka when we are forced to switch to Xenial from Precise).

PR #4547, opened 2019-02-13
Added ability to pass in encrypted key/values into the CLI and API

### LINKED ISSUE #4545: Feature Request - Ability to handle encrypted key/values
##### SUMMARY

Currently, when setting Key/Value pairs with `st2 key set` or `st2 key load` (and equivalent API calls) the data must be in a decrypted state. This means that if data is stored in an encrypted state we must decrypt it, then send it to the st2 CLI/API which then encrypts it again. Having this decrypt process fall on the responsibility of the end user requires them to implement this decrypt functionality.

Our use case here is that we have a `keys.yaml` file that contains all of our datastore key/values that contains encrypted values. This is exactly like grabbing a dump of `st2 key list`. When we provision a new instance of StackStorm or do a release we load this keys file in via `st2 key load`. Right now we have to decrypt all of the encrypted data, then pass that decrypted file in to `st2 key load`. We have to implement this decrypt code on our side and would like to allow StackStorm to handle this crypto work.

I propose allowing `st2 key set` and `st2 key load` (or equivalent API calls) to accept already encrypted values. The `st2client` or `st2api` would handle the decryption/encryption functions instead of the end user. 

##### ISSUE TYPE
 - Feature Idea

##### PROPOSAL - st2client

We could implement this on the `st2client` side for `st2 key set` by passing a `--decrypt` flag:

```shell
$ st2 key set -e --decrypt test.mysecretvalue ABCD123
```

On the `st2 key load` side we would add a new `decrypt: true` attribute for keys that need their values decrypted.

```yaml
- name: test.mysecretvalue
  value; ABCD123
  secret: true
  decrypt: true
```

##### PROPOSAL - st2api

We could handle the decryption within `st2client` if we like and pass the data to the API like we do today.

Or we could pass along this `decrypt` parameter to the API and allow `st2api` to decrypt and re-encrypt the data (or potentially pass through).

```
PUT/api/v1/keys/test.mysecretvalue

{
  "name": "test.mysecretvalue",
  "value": "ABCD123",
  "secret": true,
  "decrypt": true
} 
```

nmaludy:
Closes #4545 

This PR adds the ability to pass in encrypted key/value pairs on the CLI and via the API. The values remain encrypted and then are decrypted prior to being stored in the database.

To signify to the CLI that these values must be decrypted this adds a `-d/--debug` flag to `st2 key set`, example:

```shell
st2 key set -e -d myencryptedvalue ABC123
```

A new metadata parameter is now allowed in `st2 key load` data files called `decrypt: true` this signifies that the value for a given key is encrypted and needs to be decrypted before being processed.

```yaml
- name: myencryptedvalue
  value: ABC123
  decrypt: true
  secret: true
```

Finally the API endpoint `PUT /api/v1/keys/{name}` accepts a new parameter `decrypt` that again signifies that the value is encrypted and must be decrypted:

```shell
PUT /api/v1/keys/myencryptedvalue
{
  "name": "myencryptedvalue",
  "value": "ABC123",
  "decrypt": true,
  "secret": true
}
```

The decrypting operation is handled within the API so that `st2client` doesn't rely on `st2common`. This also makes it a consistent experience from CLI and API.

**TODO**
- [x] st2docs PR once this design is approved - https://github.com/StackStorm/st2docs/pull/869


armab:
FYI this comes from the https://github.com/StackStorm/puppet-st2/pull/250 PR where @nmaludy had no choice but to copy entire python cryptography code for st2 decrypt.

This PR solves it by adding new option to CLI/API.

Kami:
Thanks for contributing this change.

Could we avoid the decrypt -> encrypt round-trip and simply tell API the value is encrypted and to store it as-is?

The limitation of course is that the value needs to be encrypted with the same key which is used by the StackStorm installation. But that's also the case with this approach.

I guess the only downside of that approach would be that there is no "integrity check" on write - user would only find out that encrypted value is corrupted or not encrypted with the same key which is used on this particular installation on access / read time.

nmaludy:
@Kami right, i intentionally decrypt the value prior to saving to the datastore, so that the key validity is checked on write rather than read (at some point in the future).

Kami:
Alright, in this case I'm find with this approach :+1: 

It's almost always better to do integrity checks early on, on write :)

nmaludy:
I'm thinking about renaming the flag from --decrypt to --preencrypted

Does that seem more clear?

Kami:
Yeah, ``--decrypt`` is a bit config. Perhaps ``encrypted`` which signals that a value is already encrypted?

Kami:
Thanks for updating the PR.

How do you think this should tie into RBAC? It would probably be safer if only admins can pass pre-encrypted values to the PUT API to avoid potential secrets leakage?

In theory, we could allow every user to use this functionality since we only use decrypt on the API side to ensure encrypted value is encrypted with the correct key and not corrupted, but then we need to be careful that we never return plain-text decrypted version back to the user inside PUT response.

Kami:
Per discussion with @nmaludy on Slack, I made the following changes:

1. Change option name from ``decrypt`` to ``pre_encrypted``

After some more thought, ``pre_encrypted`` is much better name and doesn't cause any confusion with ``decrypt`` which does something totally different.

It signals the API that the value is already encrypted and should be used and stored as-is.

We do actually perform decryption on the API side, but that's an implementation detail and only done to ensure the integrity of the value (to ensure it has been encrypted with the correct key and that it's not corrupted).

Besides, that, we treat and store value as-is.

2. Make sure ``pre_encrypted`` always implies ``secret=True``

If we didn't do that, any user could read / decrypt any pre-encrypted value which is a big no-no and a security issue.

3. Only allow RBAC admins to use ``pre_encrypted=True``

This is an additional safe guard.

Primary use case for this feature is migration / restore of datastore values from one StackStorm instance which uses the same crypto key to another using ``st2 key load`` command.

I still need to work on the test coverage, especially around the security edge cases to ensure there is no accidental secrets leakage.

Kami:
Added test cases, and made the changes so ``st2 key list -n 100 -j > data.json ; st2 key load data.json`` now also works out of the box for encrypted values without needing to manipulate the format in ``data.json``.

I approved the PR, but since it's touches a security critical code, I also requested review from @m4dcoder.

st2client/st2client/commands/keyvalue.py
```diff
@@ -168,6 +168,10 @@ def __init__(self, resource, *args, **kwargs):
         self.parser.add_argument('-e', '--encrypt', dest='secret',
                                  action='store_true',
                                  help='Encrypt value before saving.')
+        self.parser.add_argument('--pre-encrypted', dest='pre_encrypted',
```
m4dcoder: Why not just simply name the argument `--encrypted`?

st2client/st2client/commands/keyvalue.py
```diff
@@ -168,6 +168,10 @@ def __init__(self, resource, *args, **kwargs):
...
+                                 action='store_true',
+                                 help=('Value provided is already encrypted with the instance '
+                                 'crypto key and should be stored as-is.'))
```
m4dcoder: Can we make the args `--encrypt` and `--encrypted/--pre-encrypted` mutually exclusive?  Meaning these args cannot both be present at the same time.

st2common/st2common/openapi.yaml.j2
```diff
@@ -4885,6 +4885,9 @@ definitions:
       secret:
         description: Encrypt value before saving the value.
         type: boolean
+      decrypt:
```
m4dcoder: This is a surprise.  How come we are adding decrypt attribute in this API spec?

Kami: You are commenting on an old diff, attribute is now named ``pre_encrypted``.

st2client/st2client/commands/keyvalue.py
Kami: Yeah, I was thinking about it, but I thought it might be a bit confusing in combination with ``--encrypt`` argument.

I'n fine either way though.

st2client/st2client/commands/keyvalue.py
Kami: Good point, will add it.

st2client/st2client/commands/keyvalue.py
m4dcoder: Adding mutually exclusive to the args will minimize the confusion.

Kami:
Made CLI flags mutually exclusive - 970b3eb56e90d7920acda8a3bd8980c6cdd8e2ff and changed the argument and CLI flag name from ``--pre-encrypted`` to ``--encrypted`` - 88607e3de9187da97a946935eb680ca711f30363.

st2common/st2common/models/api/keyvalue.py
```diff
@@ -165,13 +166,32 @@ def to_model(cls, kvp):
...
+
+            cls._verif_key_is_set_up(name=name)
+
+            try:
+                symmetric_decrypt(KeyValuePairAPI.crypto_key, value)
+            except Exception:
+                msg = ('Failed to verify the integrity of the provided value for key "%s". Ensure '
+                       'that the value is encrypted with the correct key and not corrupted.' %
+                       (name))
+                raise ValueError(msg)
```
m4dcoder: I may have missed it but I don't see a unit test that shows failure in checking integrity.

Kami: Good catch - I did plan to add that one, but forgot to.

Will add it.

Kami: Added in 17427bf.

PR #4548, opened 2019-02-13
Register services in service registry during the service bootstrap phase

### LINKED ISSUE #4015: Coordinator errors if session expires
In standing up our HA setup i came across an interesting corner case.

We chose `Consul` as our coordination backend for `tooz` and after a short while it appears that the auth session stored in the `coordinator` expired. When this happens and the coordination is then used to try and obtain a lock an exception is thrown. The exception below is seen in `/var/log/st2/st2api.log` when trying to add a key to the k/v store `st2 key set testkey testvalue`.

In order to restore functionality the st2api service has to be restarted.

```shell
2018-02-27 21:29:56,704 87179824 ERROR router [-] Failed to call controller function "put" for operation "st2api.controllers.v1.keyvalue:key_value_pair_controller.put": 500 rpc error making call: invalid session "2e20ab1d-4c3c-f224-02e9-f8261024ac1b"
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/router.py", line 450, in __call__
    resp = func(**kw)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2api/controllers/v1/keyvalue.py", line 192, in put
    with self._coordinator.get_lock(lock_name):
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/tooz/locking.py", line 52, in __enter__
    acquired = self.acquire(*args, **kwargs)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/tooz/drivers/consul.py", line 67, in acquire
    return _acquire()
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/tenacity/__init__.py", line 214, in wrapped_f
    return self.call(f, *args, **kw)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/tenacity/__init__.py", line 295, in call
    start_time=start_time)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/tenacity/__init__.py", line 252, in iter
    return fut.result()
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/concurrent/futures/_base.py", line 455, in result
    return self.__get_result()
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/tenacity/__init__.py", line 298, in call
    result = fn(*args, **kwargs)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/tooz/drivers/consul.py", line 58, in _acquire
    acquire=self._session_id)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/consul/base.py", line 605, in put
    CB.json(), '/v1/kv/%s' % key, params=params, data=value)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/consul/std.py", line 28, in put
    cert=self.cert)))
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/consul/base.py", line 207, in cb
    CB.__status(response, allow_404=allow_404)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/consul/base.py", line 164, in __status
    raise ConsulException("%d %s" % (response.code, response.body))
ConsulException: 500 rpc error making call: invalid session "2e20ab1d-4c3c-f224-02e9-f8261024ac1b"
2018-02-27 21:29:56,708 87179824 ERROR error_handling [-] API call failed: 500 rpc error making call: invalid session "2e20ab1d-4c3c-f224-02e9-f8261024ac1b"
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/middleware/error_handling.py", line 46, in __call__
    return self.app(environ, start_response)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/middleware/streaming.py", line 47, in __call__
    return self.app(environ, start_response)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/router.py", line 499, in as_wsgi
    resp = self(req)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/router.py", line 454, in __call__
    raise e
ConsulException: 500 rpc error making call: invalid session "2e20ab1d-4c3c-f224-02e9-f8261024ac1b" (_exception_data={},_exception_class='ConsulException',_exception_message='500 rpc error making call: invalid session "2e20ab1d-4c3c-f224-02e9-f8261024ac1b"')
```

This may be a bug in `tooz` but wanted to start the conversation here. Should StackStorm be refreshing/retrying the coordinator if an error occurs?

Kami:
This pull request updates service bootstrap phase (``st2common.service_setup.common_setup`` function) to register every StackStorm service in the service registry.

### Background and Context

Registering each service in a service segistry will allow us to have a better overview of the active and running services. This will allow us to implement things such as more dynamic (utilization and capability based) action execution scheduling / routing and more.

The goal was so it also works outside of the Kubernetes context in a traditional HA environment. It's also worth noting that Kubernetes has it's own service registry concept, but that's based on a slightly higher level Kubernetes view of a service (two related concepts, but not exactly the same).

### Implementation Details

Implementation utilizes [group membership](https://docs.openstack.org/tooz/queens/user/tutorial/group_membership.html) primitives from the [OpenStack tooz library](https://docs.openstack.org/tooz/latest/).

There were multiple reasons (as discussed before) for utilizing tooz library:

1. Consistency

We already use it in other places for distributed locking.

2. Support for multiple drivers / adapters

We will officially probably only support a single backend (e.g. ZooKeeper or etcd), but people will still be able to user other tooz supported backends at their own risk and this will also help with testing.

Keep in mind that only the following tooz drivers support group membership primitives which we utilize: zookeeper, redis, etcd3, etcd3gw, memcached.

To be able to implement it, I needed to change ``coordinator_setup`` and ``get_coordinator`` function signature so we can pass ``start_heartbeat`` argument to it. This way coordination service automatically starts periodic heartbeats which tell the backend that a particular service (member) is still alive.

In addition to that, I added ``tools/list_group_members.py`` script which allows user to list active services.

### Terminology

OpenStack tooz library exposes a couple of different terms. Here I describe how they fit into our implementation.

1. Group

Each service is part of a single group. In our case, group represents a service type (e.g. ``api``, ``actionrunner``, ``scheduler``, etc.). To view all the running services of a particular type, we just need to list all the members in a particular group.

2. Member

A service is represented as a member which is part of a particular group.

3. Capabilities

Optional metadata associated with a service (service name, listen host / port, service type, hostname, pid and other info). In the future, action runners will also utilize this to advertise their capabilities (e.g. linux, windows, etc.).

### Example Output

Example output of ``tools/list_group_members.py`` script when all the services are running using launchdev.sh script (to test it locally, I used the redis backend).

```bash
$ python ./tools/list_group_members.py --config-file conf/st2.dev.conf 
Available groups (13):
 - actionrunner
 - garbagecollector
 - workflow_engine
 - resultstracker
 - stream
 - timer_engine
 - auth
 - sensorcontainer
 - api
 - scheduler
 - rulesengine
 - notifier

Members in group "actionrunner" (1):
 - localhost_21837 (capabilities={'type': 'passive', 'hostname': 'localhost', 'pid': 21837, 'name': 'actionrunner'})
Members in group "garbagecollector" (1):
 - localhost_21841 (capabilities={'type': 'passive', 'hostname': 'localhost', 'pid': 21841, 'name': 'garbagecollector'})
Members in group "workflow_engine" (1):
 - localhost_21834 (capabilities={'type': 'passive', 'hostname': 'localhost', 'pid': 21834, 'name': 'workflowengine'})
Members in group "resultstracker" (1):
 - localhost_21871 (capabilities={'type': 'passive', 'hostname': 'localhost', 'pid': 21871, 'name': 'resultstracker'})
Members in group "stream" (1):
 - localhost_21930 (capabilities={'listen_host': '127.0.0.1', 'name': 'stream', 'listen_port': 9102, 'hostname': 'localhost', 'pid': 21930, 'type': 'active'})
Members in group "timer_engine" (1):
 - localhost_21853 (capabilities={'type': 'passive', 'hostname': 'localhost', 'pid': 21853, 'name': 'timerengine'})
Members in group "auth" (1):
 - localhost_21927 (capabilities={'listen_host': '127.0.0.1', 'name': 'auth', 'listen_port': 9100, 'listen_ssl': False, 'hostname': 'localhost', 'pid': 21927, 'type': 'active'})
Members in group "sensorcontainer" (1):
 - localhost_21849 (capabilities={'type': 'passive', 'hostname': 'localhost', 'pid': 21849, 'name': 'sensorcontainer'})
Members in group "api" (1):
 - localhost_21925 (capabilities={'listen_host': '127.0.0.1', 'name': 'api', 'listen_port': 9101, 'hostname': 'localhost', 'pid': 21925, 'type': 'active'})
Members in group "scheduler" (1):
 - localhost_21846 (capabilities={'type': 'passive', 'hostname': 'localhost', 'pid': 21846, 'name': 'scheduler'})
Members in group "rulesengine" (1):
 - localhost_21850 (capabilities={'type': 'passive', 'hostname': 'localhost', 'pid': 21850, 'name': 'rulesengine'})
Members in group "notifier" (1):
 - localhost_21870 (capabilities={'type': 'passive', 'hostname': 'localhost', 'pid': 21870, 'name': 'notifier'})
```

## TODO

- [x] Changelog entry
- [x] Tests

Resolves #4015.


tools/list_group_members.py
```diff
@@ -0,0 +1,72 @@
...
+if __name__ == '__main__':
+    cli_opts = [
+        cfg.StrOpt('group-id', default=None,
+                   help='If provided, only list members for that group.'),
+
+    ]
+    do_register_cli_opts(cli_opts)
+    config.parse_args()
+
+    main(group_id=cfg.CONF.group_id)
```
m4dcoder: Should we add an admin only API endpoint and a command in the st2 CLI to list services, members and capabilities?

st2common/st2common/service_setup.py
```diff
@@ -165,9 +173,39 @@ def setup(service, config, setup_db=True, register_mq_exchanges=True,
 
     metrics_initialize()
 
+    # Register service in the service registry
+    if service_registry:
```
m4dcoder: Should we add a config check here?  If user specifies a backend with no service registry support, we log and warn users and start st2 w/o service registry?

tools/list_group_members.py
Kami: That's a good idea - I was thinking about it.

I will add it and also add a corresponding CLI command.

nmaludy:
Should the heartbeat be started for the coordinator/locking driver instances? I believe in some implementations that if locks don't get a heartbeat then they are released.

Kami:
@nmlaudy I think you are correct, good catch :+1: 

I will also make that change, it will likely require test changes though. I already have some issues with this change (tests are getting stuck, even though I correctly shut down the coordinator everywhere, could also be a bug in a test driver we use).

Kami:
@m4dcoder This PR is now ready for a review.

I've added and fixed tests (there was an issue with tests hanging / timing out) and new CLI commands ``st2 service-registry group list``, ``st2 service-registry member list [--group-id=<group id>]``).

Adding CLI commands was a bit PITA since those commands don't follow common "pack resource" pattern and in general that CLI and http client related code is not the prettiest of things.

Kami:
On a related note^ - I propose changing the following RBAC CLI commands in a separate PR in the future:

* ``st2 role list`` -> ``st2 rbac role list``
* ```st2 role-assignment list`` -> ``st2 rbac role-assignment list``

This way it will be immediately obvious those commands refer to RBAC. In the past, we didn't have a pattern established yet for double nested commands so we went with a simple top-level approach which is somewhat confusing.

Kami:
Just a heads up - I also pushed a change so we start heartbeat process also in places where we use coordination backend only for locking - 98b77d1.

This is a safer default since it means it will work correctly with backends which are timeout based.

nmaludy:
Closes #4015

st2api/st2api/controllers/v1/service_registry.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class ServiceRegistryGroupsController(object):
+    def get_all(self, requester_user):
+        rbac_utils.assert_user_is_admin(user_db=requester_user)
+
+        coordinator = coordination.get_coordinator()
+
+        group_ids = list(coordinator.get_groups().get())
+        group_ids = [group_id_.decode('utf-8') for group_id_ in group_ids]
```
m4dcoder: Is the underscore at the end of `group_id_` mean something?

st2client/st2client/models/service_registry.py
```diff
@@ -0,0 +1,48 @@
...
+    'ServiceRegistry',
+
+    'ServiceRegistryGroup',
+    'ServiceRegistryMember'
+]
+
+
+class ServiceRegistry(core.Resource):
+    _alias = 'service-registry'
+    _display_name = 'service registry'
```
m4dcoder: Please capitalize first letters.

st2common/st2common/service_setup.py
m4dcoder: Did you address this?  Or is this a bad idea?

st2common/st2common/services/coordination.py
```diff
@@ -51,16 +55,30 @@ def heartbeat(self):
         return True
 
 
+class NoOpAsyncResult(object):
```
m4dcoder: Can you provide a short description why we need to wrap result with this class?

st2common/st2common/services/coordination.py
```diff
@@ -161,22 +192,44 @@ def coordinator_setup():
...
+    """
     global COORDINATOR
 
     if not configured():
         LOG.warn('Coordination backend is not configured. Code paths which use coordination '
                  'service will use best effort approach and race conditions are possible.')
 
+    if not use_cache:
+        coordinator = coordinator_setup(start_heart=start_heart)
+        return coordinator
```
m4dcoder: return in one line?  `return coordinator_setup(start_heart=start_heart)`

st2common/st2common/service_setup.py
```diff
@@ -170,9 +180,57 @@ def setup(service, config, setup_db=True, register_mq_exchanges=True,
...
+
+    try:
+        coordinator.create_group(group_id).get()
+    except GroupAlreadyExist:
+        pass
+
+    # Include common capabilities such as hostname and process ID
+    proc_info = system_info.get_process_info()
+    capabilities['hostname'] = proc_info['hostname']
+    capabilities['pid'] = proc_info['pid']
```
m4dcoder: Doesn't get_member_id already included this info?  Maybe doesn't matter much, but seems like we're calling system_info.get_process_info twice.

tools/launchdev.sh
```diff
@@ -68,6 +68,7 @@ function init(){
     fi
 
     VIRTUALENV=${VIRTUALENV_DIR:-${ST2_REPO}/virtualenv}
+    VIRTUALENV=$(readlink -f ${VIRTUALENV})
```
m4dcoder: I assume this is to accommodate symlink in your dev env.  Will this break existing setup for us?

tools/launchdev.sh
```diff
@@ -77,6 +78,8 @@ function init(){
     if [ -z "$ST2_CONF" ]; then
         ST2_CONF=${ST2_REPO}/conf/st2.dev.conf
     fi
+
+    ST2_CONF=$(readlink -f ${ST2_CONF})
```
m4dcoder: I assume this is to accommodate symlink in your dev env. Will this break existing setup for us?

tools/launchdev.sh
Kami: It's so an absolute path is used everywhere (just resolves relative path to an absolute one). Everything should still work the same.

st2api/st2api/controllers/v1/service_registry.py
Kami: Yes, ``group_id`` variable is already defined in the method signature and I don't want to re-define / overwrite it in a list comprehension.

st2client/st2client/models/service_registry.py
Kami: This attribute is really just a place holder, it's not used anywhere.

The class (model) is just needed to get ``st2 service-registry foo bar`` command structure to work since our, not so ideal CLI code, depends on it.

I can remove unused attributes all together.

st2common/st2common/services/coordination.py
Kami: That's a personal preference - I can prefer return in a separate line.

st2common/st2common/services/coordination.py
Kami: It's to correctly mimic tooz API - most of the tooz methods return an async result (future).

I can add that as a docstring.

st2common/st2common/service_setup.py
Kami: We already log a warning when ``get_coordinator`` is called inside ``register_service_in_service_registry`` function.

st2common/st2common/service_setup.py
Kami: Nope, this function doesn't include member info. Just hostname and pid.

st2api/st2api/controllers/v1/service_registry.py
m4dcoder: Practice wise, we don't usually put an underscore at the end of a variable.  This is confusing and error prone. Can you rename `group_id_` to something else?  Maybe `item` just for this list comprehension.

st2client/st2client/models/service_registry.py
m4dcoder: Ok these attributes are still here.  So, placeholder or not, please capitalize `service registry` to `Service Registry`. This is so the values are consistent with other models.

st2common/st2common/services/coordination.py
m4dcoder: Ok. My suggestion here is because we are not doing anything with the variable `coordinator` so why not just return here.  This is also 1 line of code less.  What's the reasoning for your preference?

Kami:
@m4dcoder 

> Practice wise, we don't usually put an underscore at the end of a variable. This is confusing and error prone. Can you rename group_id_ to something else? Maybe item just for this list comprehension.

That's a relatively common pattern in Python world. Also when variable names would clash with built-ins such as ``id``, etc.

PR #4552, opened 2019-02-15
Update pyyaml and flex contraints

blag:
Update requirements files with updated PyYAML and flex constraints.

I also add a make target to run `make requirements` and check that it doesn't change any requirements files (eg: some files were changed but not checked into git), and integrated that into the Travis tests.

H/T: @m4dcoder

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,7 +132,7 @@ play:
 	@echo
 
 .PHONY: check
-check: requirements flake8 checklogs
+check: requirements flake8 checkrequirements checklogs
 
 # NOTE: We pass --no-deps to the script so we don't install all the
 # package dependencies which are already installed as part of "requirements"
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -6,7 +6,7 @@ argcomplete
 bcrypt
 cryptography==2.4.2
 eventlet==0.24.1
-flex>=6.13.2,<6.15.0
+flex
```
m4dcoder: In your other PR at https://github.com/StackStorm/st2/pull/4510, `flex` is pinned to `flex>=6.13.2,<6.15.0`.  How come the version specific dependencies are gone?

Makefile
```diff
@@ -132,7 +132,7 @@ play:
 	@echo
 
 .PHONY: check
-check: requirements flake8 checklogs
+check: requirements flake8 checkrequirements checklogs
```
m4dcoder: This is a good add. Thanks.

requirements.txt
Kami: We are missing an entry for flex in ``fixed-requirements.txt``.

I will go ahead and add it.

Kami: Should be good to go now - d29cd7baa.

Kami:
LGTM, thanks for adding new ``check-requirements`` make task :+1:

PR #4554, opened 2019-02-18
Add missing entry for flex to fixed-requirements.txt and regenerate requirements.txt

Kami:
Changes in https://github.com/StackStorm/st2/pull/4510 didn't add / correctly update flex in requirements.txt and re-generate final requirements.txt file so requirements file was out of date.

### DIFF
--- a/fixed-requirements.txt
+++ b/fixed-requirements.txt
@@ -42,6 +42,7 @@ tooz==1.63.1
 zake==0.2.2
 routes==2.4.1
 webob==1.8.4
+flex==6.14.0
 prance==0.9.0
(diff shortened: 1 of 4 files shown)


Kami:
@blag Just a heads up - https://github.com/StackStorm/st2/blob/master/CONTRIBUTING.rst#managing-python-dependencies

In short, ``fixed-requirements.txt`` is our dependency lock file and only file where we specify actual dependency versions. ``requirements.txt`` is auto generated and shouldn't be updated manually.

Kami:
I some how missed change in #4552.

Closing this as duplicate and will continue there.

PR #4563, opened 2019-02-21
Canceling workflow with subworkflow can lead to execution stuck in canceling state

jinpingh:
Fixes https://github.com/StackStorm/orquesta/issues/121
WIP
Canceling workflow with subworkflow can lead to execution stuck in canceling state

Since unit test calls `request_cancellation(lv_ac_db, requester)` with root `LiveAction`, when Workflow `request_cancellation` is called,  `LiveAction` status changed to 'canceling` but is marked as `canceled` in orquesta, and  `ac_svc.request_cancellation(root_lv_ac_db, None)` is not call to Cascading cancellation request to parent workflow. It is why it stack in `canceling` stage. Tried this area  to deal with this spacial case and failed other test cases.

Working in area around `services/action.py request_cancellation` and `service/workflows.py request_cancellation`

### DIFF
--- a/contrib/runners/orquesta_runner/tests/unit/test_cancel.py
+++ b/contrib/runners/orquesta_runner/tests/unit/test_cancel.py
@@ -236,3 +236,33 @@ class OrquestaRunnerCancelTest(st2tests.ExecutionDbTestCase):
         # Assert the main workflow is canceling.
         lv_ac_db = lv_db_access.LiveAction.get_by_id(str(lv_ac_db.id))
         self.assertEqual(lv_ac_db.status, ac_const.LIVEACTION_STATUS_CANCELED)
+
+    def test_cancel_workflow_cascade_down_to_subworkflow_while_still_launching(self):
+       wf_meta = base.get_wf_fixture_meta_data(TEST_PACK_PATH, 'subworkflow.yaml')
+       lv_ac_db = lv_db_models.LiveActionDB(action=wf_meta['name'])
+       lv_ac_db, ac_ex_db = ac_svc.request(lv_ac_db)
+       lv_ac_db = lv_db_access.LiveAction.get_by_id(str(lv_ac_db.id))
+       self.assertEqual(lv_ac_db.status, ac_const.LIVEACTION_STATUS_RUNNING, lv_ac_db.result)
+
+       # Identify the records for the subworkflow.
+       wf_ex_dbs = wf_db_access.WorkflowExecution.query(action_execution=str(ac_ex_db.id))
+       self.assertEqual(len(wf_ex_dbs), 1)
+
+       tk_ex_dbs = wf_db_access.TaskExecution.query(workflow_execution=str(wf_ex_dbs[0].id))
+       self.assertEqual(len(tk_ex_dbs), 1)
+
+       tk_ac_ex_dbs = ex_db_access.ActionExecution.query(task_execution=str(tk_ex_dbs[0].id))
+       self.assertEqual(len(tk_ac_ex_dbs), 1)
+
+       tk_lv_ac_db = lv_db_access.LiveAction.get_by_id(tk_ac_ex_dbs[0].liveaction['id'])
+       self.assertEqual(tk_lv_ac_db.status, ac_const.LIVEACTION_STATUS_RUNNING)
+
+       # Cancel the main workflow.
+       requester = cfg.CONF.system_user.user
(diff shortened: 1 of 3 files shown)


st2common/st2common/services/action.py
```diff
@@ -260,6 +260,12 @@ def request_cancellation(liveaction, requester):
 
     # Run cancelation sequence for liveaction that is in running state or
     # if the liveaction is operating under a workflow.
+    if 'parent' not in liveaction.context and action_constants.LIVEACTION_STATUS_RUNNING:
```
m4dcoder: I can't tell why parent not in liveaction.context is related to the issue we're fixing here.  Can you explain?

st2common/st2common/services/workflows.py
```diff
@@ -388,12 +388,15 @@ def request_cancellation(ac_ex_db):
     root_ac_ex_db = ac_svc.get_root_execution(ac_ex_db)
 
     if root_ac_ex_db != ac_ex_db and root_ac_ex_db.status not in ac_const.LIVEACTION_CANCEL_STATES:
-        LOG.info('[%s] Cascading cancelation request to parent workflow.', wf_ac_ex_id)
+        LOG.info('[%s] Cascading cancellation request to parent workflow.', wf_ac_ex_id)
+        root_lv_ac_db = lv_db_access.LiveAction.get(id=root_ac_ex_db.liveaction['id'])
+        ac_svc.request_cancellation(root_lv_ac_db, None)
+    elif root_ac_ex_db == ac_ex_db and root_ac_ex_db.status == ac_const.LIVEACTION_STATUS_CANCELING:
```
m4dcoder: This is a good start, allowing users to cancel the workflow again even if it's in a canceling state.  Can you focus on this workaround first?  Currently, when this issue happens, the workflow and sub workflow is stuck in cancelling.  If we let users wait a bit and then allow them to run `st2 execution cancel` again, that would already improve the situation a bit.

m4dcoder: Actually revisiting this, I don't think adding the elif here for canceling will work.  This will just call the main workflow and ask it to cancel again.  The main workflow will then cascade the cancellation down back to this subworkflow.  And this may end up looping indefinitely.

st2common/st2common/services/workflows.py
```diff
@@ -388,12 +386,13 @@ def request_cancellation(ac_ex_db):
     root_ac_ex_db = ac_svc.get_root_execution(ac_ex_db)
 
     if root_ac_ex_db != ac_ex_db and root_ac_ex_db.status not in ac_const.LIVEACTION_CANCEL_STATES:
-        LOG.info('[%s] Cascading cancelation request to parent workflow.', wf_ac_ex_id)
+        LOG.info('[%s] Cascading cancellation request to parent workflow.', wf_ac_ex_id)
+        conductor.request_workflow_state(states.CANCELED)
```
m4dcoder: Why do we need to call request_workflow_state here?

contrib/runners/orquesta_runner/tests/unit/test_cancel.py
```diff
@@ -236,3 +236,33 @@ def test_cancel_subworkflow_cascade_up_to_workflow_with_other_subworkflows(self)
...
+       self.assertEqual(tk_lv_ac_db.status, ac_const.LIVEACTION_STATUS_RUNNING)
+
+       # Cancel the main workflow.
+       requester = cfg.CONF.system_user.user
+       lv_ac_db, ac_ex_db = ac_svc.request_cancellation(lv_ac_db, requester)
+       self.assertEqual(lv_ac_db.status, ac_const.LIVEACTION_STATUS_CANCELING)
+
+       # Re-run cancel the main workflow in case of workflow is stuck in canceling.
+       lv_ac_db, ac_ex_db = ac_svc.request_cancellation(lv_ac_db, requester)
+       self.assertEqual(lv_ac_db.status, ac_const.LIVEACTION_STATUS_CANCELED)
```
m4dcoder: I don't understand the test here.  tk_lv_ac_db is still running and the test didn't handle it's completion.  How come rerunning the cancellation will put lv_ac_db to canceled?

st2common/st2common/services/workflows.py
jinpingh: It is always called before (see line 380) even `ac_svc.request_cancellation` is not called to do the job. So moved line 380 to here only for the case if action cancellation is called.

m4dcoder:
Please see 
https://github.com/StackStorm/orquesta/issues/121#issuecomment-467970377 for why this PR is closed.

PR #4565, opened 2019-02-25
WinRM runner - Upload PowerShell scripts

### LINKED ISSUE #4514: Feature Request - Execute large WinRM scripts
##### SUMMARY

Provide the ability to execute large WinRM scripts using the `winrm-ps-script` runner.

##### ISSUE TYPE
 - Feature Idea

##### STEPS TO REPRODUCE
Execute a script larger than 8192 bytes: https://support.microsoft.com/en-us/help/830473/command-prompt-cmd-exe-command-line-string-limitation

##### EXPECTED RESULTS
Script runs successfully

##### ACTUAL RESULTS
Exits with `Command is too long`.

##### INVESTIGATION

This is a limitation of WinRM and the current runner limitation. Currently the runner simply takes the script and executes it as one long command on the command line. It does this because WinRM does not support file transfers out of the box. 

To accomplish file transfers in WinRM you can check out this thread from the `pywinrm` library: https://github.com/diyan/pywinrm/issues/18

nmaludy:
Closes #4514 

Improved the way that the ``winrm-ps-script`` runner sends scripts to the target Windows host.

Previously the script was read from the local filesystem and serialized as one long
command executed on the command line. This failed when the script was longer than either
2047 or 8191 bytes (depending on Windows version) as the Windows command line uses this
as its maximum length. To overcome this, the ``winrm-ps-script`` runner now uploads the
script into a temporary directory on the target host, then executes the script.

**NOTES** 
- This method could still fail if the parameters passed in to the action are larger than the maximum allowed Windows command line length.
- `pywinrm` is really slow and inefficient. Uploading large scripts can take a decent amount of time (many seconds). Looking at the implementation and you can see that it takes the "easy way out" on implementing many of its features (comparing to the Ruby and Go implementations of WinRM). At some point in the future, if this becomes an issue, we'll need to replace `pywinrm` with something more efficient.

**TODO** 
- [x] Tests
- [x] Detect long PowerShell commands from the winrm_ps_cmd runner and upload a script instead


nmaludy:
This is starting to come together. I have a few more tests to write, but it can be reviewed at any time.

Kami:
Thanks for working on this :+1: 

I will look in more details next week when I'm back from vacation. Similar to the last time, it will probably take some time to get end to end tests working (although, now we have infra setup and end to end tests running continuously so it should be less work).

nmaludy:
@Kami i got all of my unit tests added, i'm just adding one test to the e2e that runs a "large" script and ensure it works.

Read for a review when you are.

nmaludy:
Added a test to `st2tests` that should get run in e2e: https://github.com/StackStorm/st2tests/pull/151

contrib/runners/winrm_runner/winrm_runner/winrm_base.py
```diff
@@ -224,22 +247,139 @@ def _translate_response(self, response):
...
+            LOG.debug("WinRM uploading local file: {}".format(src_path_or_data))
+            with open(src_path_or_data, 'r') as src_file:
+                src_data = src_file.read()
+        else:
+            LOG.debug("WinRM uploading data from a string")
+            src_data = src_path_or_data
+
+        # upload the data in chunks such that each chunk doesn't exceed the
+        # max command size of the windows command line
+        for i in range(0, len(src_data), WINRM_UPLOAD_CHUNK_SIZE_BYTES):
```
Kami: Ah, I see.

So the script / file is still uploaded via command line string and not some other dedicated file upload protocol. There does appear to be overhead in serialization and de-serialization, but I imagine that's not an issue unless the script is very large (in 10's of megabytes or more).

Kami:
Thanks I will look and test this today :+1:

Kami:
End to end test on Ubuntu Xenial uncovered Python 3 issue which I fixed. Besides that, everything looks good :+1: 

I will do some more testing once it's merged. Thanks again for your contribution and nice work :+1:

Issue #4160, opened 2018-06-01
Binding of oslo.config causes oslo.log to not be usable within a pack


tonybaloney:
##### ISSUE TYPE
 - Bug Report

##### STACKSTORM VERSION
```
2.7.2
```

##### OS / ENVIRONMENT / INSTALL METHOD

one-line install

##### SUMMARY

StackStorm requires a specific version of `oslo.config`. This version does not support `mutable` as an argument.

The newer versions of `oslo.config` do.

The issue is that when installing a newer version of `oslo.log`, it will depend upon this setting. So any packs which inadvertantly use `oslo.log` will crash.

##### STEPS TO REPRODUCE

1. Create a pack with the dependency `oslo.log>1.8.0`
2. Create an action with

```python
from oslo_log import log as logging
LOG = logging.getLogger(__name__)
```

##### EXPECTED RESULTS

No error :-)

##### ACTUAL RESULTS

```

Traceback (most recent call last):
  File "/opt/stackstorm/runners/python_runner/python_runner/python_action_wrapper.py", line 320, in <module>
    obj.run()
  File "/opt/stackstorm/runners/python_runner/python_runner/python_action_wrapper.py", line 178, in run
    action = self._get_action_instance()
  File "/opt/stackstorm/runners/python_runner/python_runner/python_action_wrapper.py", line 231, in _get_action_instance
    raise exc_cls(msg)
TypeError: Failed to load action class from file "/opt/stackstorm/packs/infoblox/actions/get_object.py" (action file most likely doesn't exist or contains invalid syntax): __init__() got an unexpected keyword argument 'mutable'

Traceback (most recent call last):
  File "/opt/stackstorm/runners/python_runner/python_runner/python_action_wrapper.py", line 224, in _get_action_instance
    actions_cls = action_loader.register_plugin(Action, self._file_path)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2common/util/loader.py", line 173, in register_plugin
    module = imp.load_source(module_name, plugin_abs_file_path)
  File "/opt/stackstorm/packs/infoblox/actions/get_object.py", line 1, in <module>
    from lib.action import InfobloxBaseAction
  File "/opt/stackstorm/packs/infoblox/actions/lib/action.py", line 2, in <module>
    from infoblox_client import connector
  File "/opt/stackstorm/virtualenvs/infoblox/lib/python2.7/site-packages/infoblox_client/connector.py", line 27, in <module>
[...]

PR #4571, opened 2019-03-01
Fixed a problem that libraries in the virtualenv of a pack won't be prioritize to be loaded over st2's ones at PythonRunner action.

userlocalhost:
## Summary
This Fixes the problem of #4160 that the module of specified version wouldn't be loaded when same name library is specified in the st2, even though a library were declared with specific version in the requirements.txt of a pack.

## Root cause
This is because that loading a library from st2 library path is prioritized over packs virtualenv library path of a pack.

## Solution
As is mentioned by Kami at https://github.com/StackStorm/st2/issues/4160#issuecomment-436015668, this patch puts priority on loading virtualenv library in the pack's action by updating sys.path before loading them.


Kami:
Thanks, the change looks good to me.

We should just keep an eye on it once it's merged to make sure it doesn't cause any inadvertent regressions "in the wild" which are not caught by unit, integration and end to end tests.

It's also worth noting that this may not work in some scenarios - e.g. pack depends on a newer version of library, buy Python action wrapper depends on an older versions and there is a breaking change in the newer version of the library.

The surface area is quite small because it only affects things used by Python action wrapper script, but still something to keep in mind.

I will add a changelog entry.

Can you please also open st2docs PR which documents this behavior (e.g. as a foot note or similar)?

userlocalhost:
Thank you for your review, m4dcoder and Kami.

> Can you please also open st2docs PR which documents this behavior (e.g. as a foot note or similar)?

I got it!.

Kami:
I fixed the Travis test issues.

I will go ahead, merge this into master and keep in eye on potential inadvertent regressions.

Thanks again for your contribution :+1:

PR #4575, opened 2019-03-04
Don't use "null" for Access-Control-Allow-Origin

anna-wix:
As per https://w3c.github.io/webappsec-cors-for-developers/#avoid-returning-access-control-allow-origin-null


bigmstone:
Thanks for this contribution. I'm going to close in favor of #4577 - which should get tests to pass as well as return a proper CORS Allow-Origin header instead of unsettling it.

Thanks again for bringing this to our attention.

PR #4577, opened 2019-03-05
Fix improper CORS return

bigmstone:
Prior to this commit if you sent a request from an origin not listed in `allowed_origins` we would respond with `null` for the `Access-Control-Allow-Origin` header. Per [mozilla's documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin#Directives) `null` should not be used as some clients will allow the request to go through. This PR returns the first of our allowed origins if the requesting origin is not a supported origin.

### DIFF
--- a/st2api/tests/unit/controllers/v1/test_base.py
+++ b/st2api/tests/unit/controllers/v1/test_base.py
@@ -51,8 +51,8 @@ class TestBase(FunctionalTest):
             'origin': 'http://xss'
         })
         self.assertEqual(response.status_int, 200)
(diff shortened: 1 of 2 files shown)


Kami:
LGTM, probably also a good idea to document this behavior in st2docs?

Quitten:
Hey, I am the reporter of this bug, I assigned the following CVE for it:
CVE-2019-9580
Thanks for fixing and fast cooperation :)

bigmstone:
@Quitten thanks again. The patch release should be out today. Will keep you posted.

PR #4578, opened 2019-03-05
Cherry pick CORS security patch from master

m4dcoder:
The fix will address the security bug where request from an origin not listed in ``allowed_origins`` will be responded with ``null`` for the ``Access-Control-Allow-Origin`` header. The patch returns the first of our allowed origins if the requesting origin is not a supported origin.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,13 @@ Changelog
 In development
 --------------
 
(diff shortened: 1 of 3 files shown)


bigmstone:
Closing. Branch name preventing tests to pass since it's testing against `master`

PR #4505, opened 2019-01-21
Fix (workaround) for very slow test runs and timeouts on Travis CI

Kami:
After spending a lot of time on Saturday and today I believe I managed to finally track down the issue with very slow tests on Travis.

It turns out that on Travis RabbitMQ throws a non-fatal ``Channel.open: (504) CHANNEL_ERROR - second 'channel.open' seen`` error when using our connection retry wrapper code which causes the code to "hang" and sleep for 10 seconds here https://github.com/StackStorm/st2/blob/master/st2common/st2common/transport/connection_retry_wrapper.py#L130 (the issue can be reproduced 100% of the time on Travis).

This happens very often so tests are up to 20 times and more slower on Travis than locally.

For example:

```bash

Local VM:

(virtualenv) vagrant@ubuntu-xenial:/data/stanley$ nosetests --with-timer --rednose --immediate -sv st2actions/tests/unit/test_scheduler.py
test_create_from_liveaction (tests.unit.test_scheduler.ActionExecutionSchedulingQueueItemDBTest) ... passed
test_failed_next_item (tests.unit.test_scheduler.ActionExecutionSchedulingQueueItemDBTest) ... passed
test_garbage_collection (tests.unit.test_scheduler.ActionExecutionSchedulingQueueItemDBTest) ... passed
test_next_execution (tests.unit.test_scheduler.ActionExecutionSchedulingQueueItemDBTest) ... passed
test_next_executions_empty (tests.unit.test_scheduler.ActionExecutionSchedulingQueueItemDBTest) ... passed
test_no_double_entries (tests.unit.test_scheduler.ActionExecutionSchedulingQueueItemDBTest) ... passed
test_no_processing_of_non_requested_actions (tests.unit.test_scheduler.ActionExecutionSchedulingQueueItemDBTest) ... passed
test_processing_when_task_completed (tests.unit.test_scheduler.ActionExecutionSchedulingQueueItemDBTest) ... passed

TEST RESULT OUTPUT:

[success] 19.82% tests.unit.test_scheduler.ActionExecutionSchedulingQueueItemDBTest.test_processing_when_task_completed: 1.3436s
[success] 19.79% tests.unit.test_scheduler.ActionExecutionSchedulingQueueItemDBTest.test_no_processing_of_non_requested_actions: 1.3413s
[success] 14.81% tests.unit.test_scheduler.ActionExecutionSchedulingQueueItemDBTest.test_failed_next_item: 1.0039s
[...]

PR #4580, opened 2019-03-06
V2.9.3

bigmstone:
2.9.3 patch for CORS security fix.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,17 @@
 Changelog
 =========
 
+In development
+--------------
+
+Fixed
+~~~~~
+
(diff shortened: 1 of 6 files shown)


Kami:
To get the build to work I needed to remove lint api spec target - cec76ec.

There were some weird Travis build errors due to prance trying to install newer version of ``requests`` and that failed with permission error which I couldn't figure out how to fix (without forking and patching ``prance``).

I also tried to remove Travis cache, etc. without success.

If we will encounter same issues in master in the future we might have no choice but to temporary work prance until we can switch to newer version of ``requests``.

Kami:
I also cherry picked fix from https://github.com/StackStorm/st2/pull/4505 to avoid Travis unit tests from running 50+ minutes instead of ~15.

PR #4581, opened 2019-03-06
Add some additional asserts for origin header

Kami:
Just some additional test cases for fix in #4577 (to make sure there are no issues hiding in some edge cases).


Kami:
I intentionally left the first one outside of the loop to be on the safe side and make sure check is always executed - I've seen it happen before that for some reason or another (typo or similar) loop is not ran and that's obviously not good.

In cases like that where it's critical for the loop to run, I usually add additional check to the end after the loop and verify the iteration count aka verify loop did indeed run.

PR #4582, opened 2019-03-07
Update changelog for v2.10.3

m4dcoder:
Add changelog entry from the v2.10.3 patch release.


Kami:
Probably also a good idea to add changelog section for v2.9.3.

Kami:
@m4dcoder It looks like this Orquesta test randomly fails every now and then - https://travis-ci.org/StackStorm/st2/jobs/502930431

It would be good if you can have a look when you get a chance.

m4dcoder:
@Kami Ack. I will take a look at the test.

PR #4586, opened 2019-03-08
Fix random failing orquesta unit tests

m4dcoder:
The task function related unit tests are randomly failing because the route ID changes depending on the order of action execution. The workflows are rewritten to be more deterministic.


blag:
I'm curious how `task(..., route=>N)` works with a task that has `join: all`, but this LGTM.

m4dcoder:
@blag so non-join task with multiple references (aka split) is given a new route ID for each reference and then pass them to subsequent tasks.  For a join task (regardless of integer or all), it is not this split use case and will use the default route that get passed along.  in the normal use case of the task function, user don't have to pass the route ID.  the route ID is figured out internally by the context being passed into it.  in the unit/integration test, i want to be able to test whether each split got the right task along its route.  therefore, i explicitly pass the route info to retrieve the individual task9 to reach it's output.

Kami:
:+1:

PR #4587, opened 2019-03-08
Add cache control header

bigmstone:
This PR adds cache control header so that clients will always refresh their data from the API instead of favoring cached version. This was encountered in ST2Flow 2.0 development, loading workflows from the API.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,6 +18,9 @@ Added
    For backward compatibility reasons, if pack metadata file doesn't contain that attribute, it's
    assumed it only works with Python 2. (new feature) #4474
 
+* Adding ``Cache-Control`` header to all API responses so clients will favor
+  refresh from API instead of using cached version.
+
 Changed
 ~~~~~~~
 
--- a/st2api/st2api/app.py
+++ b/st2api/st2api/app.py
@@ -20,6 +20,7 @@ from st2common import log as logging
 from st2common.middleware.streaming import StreamingMiddleware
 from st2common.middleware.error_handling import ErrorHandlingMiddleware
 from st2common.middleware.cors import CorsMiddleware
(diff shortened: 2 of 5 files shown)


m4dcoder:
Will this impact performance?

bigmstone:
That's a broad question. For our CLI clients, other service level integrations, no. They don't use any caching. Even for our Web-UI this will not cause much of a change since most of our APIs fall under client-side policies that attempt to refresh first. So, broadly speaking, no. This will just be a formalization of what already happens. With the side benefit of making the Flow/Web interactions work correctly.

Kami:
I'm personally not a huge fan of this change since it basically breaks having any kind of proxy infront of st2api.

It looks like it only affects "file level" API endpoints. If that's the case, I would prefer to only add those headers there.

Or even better, if possible, send correct header based on the file modification time / content. In fact, that's what we already do for pack views endpoints (including the icon one since loading all the icons on each refresh is inefficient).

If you are busy with other things, point me at the problematic API endpoints and st2flow code where this issue is encountered and I can take a look at utilizing ETag header, if possible.

Kami:
Per our Slack discussion - I believe this change can be removed.

Yesterday I confirmed it works as expected without it. We correctly send ``ETag`` and ``Last-Modified`` header for all the "file" API endpoint responses (and new definition of workflow gets loaded after re-load on save in flow).

In your scenario it must probably be some environment specific thing - perhaps type of shared folder you use or similar (e.g. filesystem is mounted with nomtime and noatime option or similar).

I use NFS and it works fine, file modified time gets correctly updated once workflow definition is changed and saved.

The reason why I'm not a big fan of this change is because I specifically spent time with Kiril in the past adding those headers to avoid unnecessary load on the API when there are no changes (e.g. loading the same pack icon on each reload / load is inefficient).

PR #4592, opened 2019-03-13
Make sure we don't log sensitive query params in the st2api log file

### LINKED ISSUE #4589: The api key in the st2api log is not obfuscated
##### SUMMARY
The user found in clean API key in query request (for the load balancer health check)
```GET /api/v1/?st2-api-key=foo HTTP/1.1```

##### ISSUE TYPE
 - Bug Report
 
##### STACKSTORM VERSION
st2 2.10.3, on Python 2.7.12

Kami:
Our logging middleware didn't mask / remove auth token and api key values if those were provided via query parameters and not headers.

This pull request fixes that.

Keep in mind that providing it as header is still preferred, especially when integrating with 3rd party services, because a lot of services log actual full URL with a query string which could mean leaking those values there. I will add a note about that to the docs.

Resolves #4589.

Also, as commented on #4589, I believe that issue actually talks about token being logged in a loadbalancer log or similar because the log line doesn't look anything like st2api log line. In that scenario we can't do anything about that and that's why I mentioned above, using headers is preferred (and I will also document that).

## TODO

- [x] Tests
- [x] Documentation change

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -51,6 +51,7 @@ Fixed
   StackStorm/orquesta#112
 * Make sure we don't log auth token and api key inside st2api log file if those values are provided
   via query parameter and not header (``?x-auth-token=foo``, ``?st2-api-key=bar``). (bug fix) #4592
+  #4589
 
 2.10.3 - March 06, 2019
 -----------------------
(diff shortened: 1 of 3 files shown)


Kami:
I forgot I already added a note about that to the docs in the past so I only made a small formatting change - https://github.com/StackStorm/st2docs/pull/866.

PR #4595, opened 2019-03-13
Acquire lock in workflow engine before handling action execution completion

m4dcoder:
Acquire a distributed lock, keyed on workflow execution ID, in the workflow engine before handling the action execution completion. This avoids the performance hits from retrying write conflicts for workflows with a lot of parallelism such as a with items task with many items and no concurrency. The coordination service in st2 needs to be setup for the solution to take into effect.

### DIFF
--- a/st2tests/integration/orquesta/test_performance.py
+++ b/st2tests/integration/orquesta/test_performance.py
@@ -42,3 +42,21 @@ class WiringTest(base.TestWorkflowExecution):
             self.assertEqual(e.status, ac_const.LIVEACTION_STATUS_SUCCEEDED)
             self.assertIn('output', e.result)
             self.assertIn('vm_id', e.result['output'])
+
+    def test_with_items_load(self):
+        wf_name = 'examples.orquesta-with-items-concurrency'
+
+        num_items = 10
+        concurrency = 10
+        members = [str(i).zfill(5) for i in range(0, num_items)]
+        wf_input = {'members': members, 'concurrency': concurrency}
+
+        message = '%s, resistance is futile!'
+        expected_output = {'items': [message % i for i in members]}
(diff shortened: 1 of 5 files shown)


Kami:
:+1:

st2tests/integration/orquesta/test_performance.py
```diff
@@ -42,3 +42,21 @@ def test_concurrent_load(self):
...
+        concurrency = 10
+        members = [str(i).zfill(5) for i in range(0, num_items)]
+        wf_input = {'members': members, 'concurrency': concurrency}
+
+        message = '%s, resistance is futile!'
+        expected_output = {'items': [message % i for i in members]}
+        expected_result = {'output': expected_output}
+
+        ex = self._execute_workflow(wf_name, wf_input)
+        ex = self._wait_for_completion(ex)
```
Kami: Can we also add some kind of assert on timing (aka how long it took this to complete). Or would this be too fragile and cause too many false positives due to the nature of the CI environment?

m4dcoder: We currently don't setup coordination service for the orquesta-itests.  This is a test that I setup to manually verify that the fix works and i kept it here for future reference.  Also at 10 items, with or w/o the lock doesn't make a huge difference in timing (both under 10 seconds).  I only see any real difference in timing at 25-50 items and this will inflate the CI run time by 30+ seconds.  I think it is ok as is.  But if you think we should configure coordination service and add 30+ seconds to the CI runtime, I'm ok to add it.

Kami: Yeah, I thought so (since we use zake for coordination backend).

I'm fine with that for now. Eventually when we get continuous automated per commit performance testing up and running we should hook it up though.

PR #4597, opened 2019-03-14
Treat any statsd metric driver related errors as non-fatal (just log the errors)

Kami:
This pull request updates our statsd driver to treat any exception which may be thrown by the driver as non-fatal.

Metrics driver related error should never break or degrade application performance and this pull request ensures that is indeed the case.

### Background / Context

One of our users reported an issue with statd driving breaking / hanging the application when a metric backend server goes down. Initially I wasn't able to replicate it (statsd uses UDP so it should be fire and forget and there should be no exceptions).

Today the user provided log files. It turns out the problem lies in DNS resolution errors which can happen when establishing a connection to the backend server.

If a hostname and not IP address is used for metric backend and DNS resolution fails, statsd will propagate this exception all the way up and this will break the application since we don't catch and ignore / log those exceptions.

I wasn't able to reproduce the error because we use IP and not the hostname everywhere and even if backend server goes down, UDP send won't result in any exception due to the nature of the protocol.

### Proposed Solution

In this solution I simply use a decorator which just logs any exception thrown by the driver and doesn't propagate it.

Eventually we could perhaps move that logic to the statsd client library, but it's still safer to have such safe guard in our code. As mentioned above, we don't want any such error / exception to break or degrade application performance.

## TODO

- [x] Tests
- [x] Changelog entry

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -63,6 +63,13 @@ Fixed
   (bug fix) #4570
 * Add missing default config location (``/etc/st2/st2.conf``) to the following services:
   ``st2actionrunner``, ``st2scheduler``, ``st2workflowengine``. (bug fix) #4596
+* Update statsd metrics driver so any exception thrown by statsd library is treated as non fatal.
+
+  Previously there was an edge case if user used a hostname instead of an IP address for metrics
+  backend server address. In such scenario, if hostname DNS resolution failed, statsd driver would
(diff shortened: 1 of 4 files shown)


Kami:
Added tests.

I also needed to refactor the code a bit so now we only catch and ignore socket.error, IOError and OSError. We still want actual AssertionErrors to propagate since this would indicate a programmer error during development (and those exceptions should also be caught during development).

armab:
:+1: Cool stuff!

PR #4607, opened 2019-03-21
Temporary disable service registry service registration to work around issue in stackstorm-ha

Kami:
This pull request adds a config option which specifies if a service should be registered in a service registry (and in any case, it's a good idea to have a flag so people can opt-out, if they have a valid reason for it, once it will default to ``True``).

To work around bug / issue in - https://github.com/StackStorm/stackstorm-ha/issues/57 it defaults to ``False``.

I forgot that the etcd ``tooz`` driver we use is old and doesn't support group functionality.

I opened a PR for etcd3gw driver group functionality support in tooz quite a long time ago. I will look into updating version ``tooz`` version we use and once that's done, I will flip that config value to ``True``.

### DIFF
--- a/conf/st2.conf.sample
+++ b/conf/st2.conf.sample
@@ -97,6 +97,8 @@ packs_base_paths = None
 [coordination]
 # Endpoint for the coordination server.
 url = None
+# True to register StackStorm services in a service registry.
+service_registry = False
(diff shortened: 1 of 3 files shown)


armab:
Thanks for being on top of it! :+1:

PR #4608, opened 2019-03-21
[WIP] [DONT MERGE] Upgrade tooz, re-enable service registry registration by default

Kami:
This pull request updates tooz dependency to our fork - https://github.com/openstack/tooz/compare/master...StackStorm:etcd3_driver_fixes?expand=1.

Latest version does contain some group related functionality in ``etcd3`` and ``etcd3+http`` driver, but it's partial and broken.

I've fixed various issues in our fork and I also plan to submit changes upstream (but it might take a while to get them merged and new version releases) so we need to use our fork for now.

There are some more issues I need to fix there.

NOTE: For this to work in stackstorm-ha, we need to ensure etcd >= 3.0.0 is installed and used. In addition to that, we need to use ``etcd+http://localhost`` driver there.

### DIFF
--- a/st2common/in-requirements.txt
+++ b/st2common/in-requirements.txt
@@ -19,6 +19,8 @@ requests
 retrying
 semver
 six
+# Need our fork with various bug fixes in etcd drivers
+git+https://github.com/StackStorm/tooz.git@etcd3_driver_fixes#egg=tooz
 tooz
 # Required by tooz - on new versions of tooz, all the backend dependencies need
 # to be installed manually
--- a/fixed-requirements.txt
+++ b/fixed-requirements.txt
@@ -38,7 +38,7 @@ virtualenv==15.1.0
 sseclient==0.0.19
 python-editor==1.0.3
 prompt-toolkit==1.0.15
-tooz==1.63.1
+tooz==1.64.2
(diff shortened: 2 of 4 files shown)


Kami:
It looks like group membership functionality is quite badly broken in those two drivers and will need more work.

armab:
Tracking & cross-linking corresponding issue in `stackstorm-ha` for switching `etcd2` -> `etcd3`: https://github.com/StackStorm/stackstorm-ha/issues/58. We'll try to migrate to v3 in next few days.

st2common/in-requirements.txt
```diff
@@ -19,6 +19,8 @@ requests
 retrying
 semver
 six
+# Need our fork with various bug fixes in etcd drivers
+git+https://github.com/StackStorm/tooz.git@etcd3_driver_fixes#egg=tooz
```
armab: Per https://github.com/StackStorm/stackstorm-ha/pull/61#pullrequestreview-224058430, looks like we'll also need `etcd3gw` to make `etcd3+http` entrypoint (https://docs.openstack.org/tooz/latest/user/drivers.html#etcd3-gateway) work out of the box

PR #4610, opened 2019-03-25
Upgrade dependencies to latest stable versions

Kami:
Just some house keeping for v3.0.0.

Latest version of ``flake8`` also contains some new checks which identified some new lint issues (some of those are good checks, aka you should always use ``==`` and not ``is`` comparison operator when comparing strings).


Kami:
NOTE: We may also need to update kombu here - https://github.com/StackStorm/mistral/pull/41

Kami:
For now I needed to revert sseclient upgrade due to the issue reported in https://github.com/btubbs/sseclient/pull/26.

PR #4611, opened 2019-03-27
Fix Missing parameters in linux.wait_for_ssh

LindsayHill:
If you specified username via CLI, `st2client` would reject it due to a missing `type:` value.


Kami:
Good catch :+1: 

Ideally, if possible, we would also update our action parameters schema and make ``type`` required attribute.

LindsayHill:
I've been wondering about that. Problem is that everywhere else we assume that default type is string. It's only st2client that has issues, in certain scenarios. Medium chance of breaking things for quite a few people that have made assumption of default `string`.

See #4612 for some additional info

PR #4622, opened 2019-04-08
st2-self-check should ignore any pack actions which name doesn't start with "test_"

Kami:
This pull request fixes bug in ``st2-self-check`` inadvertently introduced in this PR - https://github.com/StackStorm/st2tests/pull/150.

It updates ``st2-self-check`` code so it only tries to run actions which name starts with ``test_``.

(Per my comment in the other PR, I assume we ended up with this bug / regression due to lack of understanding of the difference between various test repos and packs we have - we need better developer documentation on that).


Kami:
With https://github.com/StackStorm/st2tests/pull/153 which fixes the root cause, this change is not technically needed anymore, but I don't think it hurts either and it's an additional safe guard.

Kami:
I noticed yet another "unrelated" self check issue which I fixed in bb8be32.

We didn't skip winrm runner tests by default so self check would fail (same as other Windows runners, those tests don't work out of the box and shouldn't run by default).

st2common/bin/st2-self-check
```diff
@@ -119,18 +119,27 @@ rm -R st2tests/
...
 for TEST in $TEST_ACTION_LIST
 do
     # If -w option is not set, skip Windows related tests
-    # because smbclient is not installed by default.
     if [ ${RUN_WINDOWS_TESTS} = "false" ] && [ ${TEST} = "tests.test_windows_runners" ]; then
         echo "Skipping ${TEST}..."
         continue
     fi
 
+    if [ ${RUN_WINDOWS_TESTS} = "false" ] && [ ${TEST} = "tests.test_winrm_runners" ]; then
```
Kami: Just a heads up - that's something we missed while working on original WinRM runner PRs.

Those tests only works on servers which have access to a configured Windows host which is not the case for default StackStorm installation and ``st2-self-check`` script.

PR #4623, opened 2019-04-08
Fix erroneous carriage return character (\r) in the output of remote shell script and command actions

Kami:
Ladies and gentlemen, long time mystery of "erroneous" carriage return characters sometimes found in stdout and stderr of remote runner actions has finally been solved.

While working on and fixing a totally unrelated issue in https://github.com/StackStorm/st2tests/pull/154, I found our that sometimes ``\r`` character ends up in action stdout even though the underlying script only prints ``\n`` and not ``\r\n``.

It turns out when pty is used, default terminal option is ``onlcr`` which translates ``\n`` into ``\r\n`` (see https://stackoverflow.com/questions/7812142/how-to-toggle-cr-lf-in-gnu-screen and https://linux.die.net/man/1/stty for details).

This is not desired in our remote actions output (in fact, we had a lot of users report such issues in Slack and similar).

It looks like we already tried to resolve that all the way back in 2015 (eca3751bea8e66784f5e6152cb0d87ff1125db24), but our solution / workaround for it wasn't complete. We only removed those characters from the end of the string, but we didn't do anything with ``\r\n`` in the middle of the string.

## TODO

- [x] Tests
- [x] Changelog entry


Kami:
NOTE: We already have end to end test for it in st2tests repo, which wasn't failing due to the bug in our assertion code in st2tests repo which I fixed here - https://github.com/StackStorm/st2tests/pull/154.

Kami:
@m4dcoder Yeah there are unit tests in 56dfa1c9813ae10195f5dfa2bcd10ef500ad8cb0 and existing end to end tests in st2tests repo (which were not failing due to the bug in the assert function in st2tests repo, but I fixed that there).

st2common/tests/unit/test_misc_utils.py
```diff
@@ -94,3 +95,44 @@ def test_fast_deepcopy_success(self):
...
+            'foo',
+            'foo\nbar\nbaz',
+            'foo\r\nbar\r\nbaz',
+        ]
+
+        for input_str, expected_output in zip(input_strs, expected):
+            output = sanitize_output(input_str, uses_pty=False)
+            self.assertEqual(expected_output, output)
+
+        # 2. pty is used, \r\n should be replaced with \n
```
blag: Should this be split out into its own test function?

Kami: I'm fine either way :)

Kami: I decided to split it into two methods - 3524e05c2.

When to split the tests and when not to is not always totally clear - when a test needs to be isolated and potentially also requires on ``setUp`` and ``tearDown`` method to run then it's clear that such tests need to be separate methods, but in scenarios like that it's less clear / needed.

In general, I do agree with you though - I prefer separate methods since it provides more user-friendly test runner output (and unless setUp / tearDown add a huge overhead then there are really no drawbacks).

PR #4632, opened 2019-04-12
Trstruth/st2client user filter

trstruth:
the `st2 execution list` cli command doesn't expose a `--user` filter, despite the fact that st2client accepts a `user` kwarg for `st2client.client.executions.get_all()`.  This changes that.

image
image


Kami:
Merged, thanks.

PR #4634, opened 2019-04-12
Fixed a problem not to decrypt user scope data store value in Jinja expression

### LINKED ISSUE #4633: It's fail to decrypt user scope data store value in Jinja expression
##### SUMMARY

When I get user scope value from datastore in Jinja expression then decrypt it using `decrypt_kv` filter like below, st2api returns HTTP 400 responce as below.

##### ISSUE TYPE
Bug Report

##### STACKSTORM VERSION
`st2 3.0dev, on Python 2.7.12`

custom install in line with [Run From Sources](https://docs.stackstorm.com/development/sources.html)

##### STEPS TO REPRODUCE
Runs an action that gets and decrypts value in Jinja expression as below.
```
(virtualenv) ubuntu@ubuntu0:~/st2$ st2 key get -s user -d value
+------------------+------------+
| Property         | Value      |
+------------------+------------+
| name             | value      |
| value            | aabb       |
| secret           | True       |
| encrypted        | False      |
| scope            | st2kv.user |
| expire_timestamp |            |
+------------------+------------+
(virtualenv) ubuntu@ubuntu0:~/st2$ st2 action get default.test_kv
| id            | 5caf026a987b613c65c31229                                     |
| uid           | action:default:test_kv                                       |
| ref           | default.test_kv                                              |
| pack          | default                                                      |
| name          | test_kv                                                      |
| description   | Action that executes an arbitrary Linux command on the       |
|               | localhost.                                                   |
| entry_point   |                                                              |
| runner_type   | local-shell-cmd                                              |
| parameters    | {                                                            |
|               |     "cmd": {                                                 |
|               |         "default": "echo {{ st2kv.user.value | decrypt_kv    |
|               | }}",                                                         |
|               |         "required": true,                                    |
|               |         "type": "string",                                    |
|               |         "description": "Arbitrary Linux command to be        |
|               | executed on the local host."                                 |
|               |     },                                                       |
|               |     "sudo": {                                                |
|               |         "immutable": true                                    |
|               |     }                                                        |
|               | }                                                            |
| metadata_file |                                                              |
| notify        |                                                              |
| output_schema |                                                              |
| tags          |                                                              |
(virtualenv) ubuntu@ubuntu0:~/st2$ st2 run default.test_kv
MESSAGE: Failed to render parameter "cmd": ciphertext needs to either be a string/unicode or bytes for url: http://127.0.0.1:9101/v1/executions

(virtualenv) ubuntu@ubuntu0:~/st2$ 
```

##### EXPECTED RESULTS
It works fine to get and decrypt system scope value in the same way as above. I expect it also can do it in the case of user scope value.
```Diff
--- a/test_kv_user.yaml
+++ b/test_kv_system.yaml
@@ -8,7 +8,7 @@ parameters:
     description: Arbitrary Linux command to be executed on the local host.
     required: true
     type: string
-    default: "echo {{ st2kv.user.value | decrypt_kv }}"
+    default: "echo {{ st2kv.system.value | decrypt_kv }}"
   sudo:
     immutable: true
```
スクリーンショット 2019-04-12 11 45 56

##### ACTUAL RESULTS
* [This](https://gist.github.com/userlocalhost/4563f91ea6002cf5d893ccfae2369bf2) is an output of same command with `--debug` option).
* [This](https://gist.github.com/userlocalhost/49126a169ecff38fcb976668c216aeb2) is an extract of st2api.log when that action run.

userlocalhost:
## Summary
This fixes #4633.

## Changing Point
This changes the implementation of custom Jinja filter `decrypt_kv` to be able to handle user scope value.

### DIFF
--- a/st2common/tests/unit/test_jinja_render_crypto_filters.py
+++ b/st2common/tests/unit/test_jinja_render_crypto_filters.py
@@ -17,25 +17,35 @@ from __future__ import absolute_import
 from oslo_config import cfg
 
 from st2tests.base import CleanDbTestCase
-from st2common.constants.keyvalue import FULL_SYSTEM_SCOPE, SYSTEM_SCOPE, DATASTORE_PARENT_SCOPE
+from st2common.constants.keyvalue import (
+    FULL_SYSTEM_SCOPE,
+    SYSTEM_SCOPE,
+    DATASTORE_PARENT_SCOPE,
+    FULL_USER_SCOPE,
+    USER_SCOPE,
+)
 from st2common.models.db.keyvalue import KeyValuePairDB
 from st2common.persistence.keyvalue import KeyValuePair
-from st2common.services.keyvalues import KeyValueLookup
+from st2common.services.keyvalues import KeyValueLookup, UserKeyValueLookup
 from st2common.util import jinja as jinja_utils
 from st2common.util.crypto import read_crypto_key, symmetric_encrypt
 
 
 class JinjaUtilsDecryptTestCase(CleanDbTestCase):
+    def setUp(self):
+        super(JinjaUtilsDecryptTestCase, self).setUp()
 
-    def test_filter_decrypt_kv(self):
-        secret = 'Build a wall'
         crypto_key_path = cfg.CONF.keyvalue.encryption_key_path
         crypto_key = read_crypto_key(key_path=crypto_key_path)
-        secret_value = symmetric_encrypt(encrypt_key=crypto_key, plaintext=secret)
-        KeyValuePair.add_or_update(KeyValuePairDB(name='k8', value=secret_value,
+
+        self.secret = 'Build a wall'
+        self.secret_value = symmetric_encrypt(encrypt_key=crypto_key, plaintext=self.secret)
+        self.env = jinja_utils.get_jinja_environment()
+
+    def test_filter_decrypt_kv(self):
+        KeyValuePair.add_or_update(KeyValuePairDB(name='k8', value=self.secret_value,
                                                   scope=FULL_SYSTEM_SCOPE,
                                                   secret=True))
-        env = jinja_utils.get_jinja_environment()
 
         context = {}
         context.update({SYSTEM_SCOPE: KeyValueLookup(scope=SYSTEM_SCOPE)})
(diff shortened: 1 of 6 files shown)


st2common/tests/unit/test_jinja_render_crypto_filters.py
```diff
@@ -46,5 +56,22 @@ def test_filter_decrypt_kv(self):
         })
 
         template = '{{st2kv.system.k8 | decrypt_kv}}'
-        actual = env.from_string(template).render(context)
-        self.assertEqual(actual, secret)
+        actual = self.env.from_string(template).render(context)
+        self.assertEqual(actual, self.secret)
+
+    def test_filter_decrypt_kv_with_user_scope_value(self):
```
Kami: It would also be a good idea to add "end to end" test which verifies that it works for the following scenarios:

1. API user is provided (if action is scheduled via API, API user should be used)
2. User is not provided - it should default to stanley (if execution is scheduled outside of the API context, no user context is available so stanley is used by default)

Kami:
After some more digging in - it looks like this an "addition" to this change - https://github.com/StackStorm/st2/pull/4463 and we already have some tests for value of the user in the context there.

I will go ahead and merge this as is, but going forward it also wouldn't hurt to add some additional end to end tests for this functionality.

Kami:
While testing and reviewing this, I noticed there are more issues with this functionality.

1. User scoped datastore items are not supped in action-chain workflows.

I've fixed that in 42fe11b50aa515d46d4e92bed28da4bbcf2d2ce2. I also updated an action which verifies this functionality - 8e5702924f6add6d66f018943162edfda4940b7f.

2. Non-friendly error is throw if ``decrypt_kv`` Jinja function is used on non-existing datastore item

Previously ``Invalid or malformed ciphertext (too short)`` error was throw in such scenario which is rather confusing and not user-friendly.

Now in such scenario we throw `Referenced Datastore item "st2kv.system.my_cmd" doesn't exist` error.

I'm still working on test cases.

Kami:
I've pushed a small refactor and tests for additional fixes I made - ba3c6a50a99074e88b7c5f03a8bc6691fdeb6fd2.

userlocalhost:
> I will go ahead and merge this as is, but going forward it also wouldn't hurt to add some additional end to end tests for this functionality.

Thank you very much for your help!

And sorry for late to reply. It took time to seek the place where is appropriate to extend for doing 'end to end' tests. At last, I reached the test of `st2api/tests/unit/controllers/v1/test_executions.py` which might be suitable to confirm this change from end to end.

Could you please give me a couple of days to write it.

Kami:
@userlocalhost 

Thank you for the original contribution.

And yes, that's the right place for that test. I will probably go ahead and merge this and you feel free to open a new PR for that tests so take your time - no rush.

PR #4637, opened 2019-04-15
Update dependencies to latest versions

Kami:
Just some house keeping before the release.

I noticed we didn't pin ``bcrypt`` dependency so I also fixed that so we have reproducible builds.


Kami:
Per @LindsayHill - will also push a change to st2sdk so dependency versions there match versions defined here.

Kami:
@LindsayHill in-requirements.txt only contain a list of requirements without version numbers. version numbers are specified in ``fixed-requirements.txt`` - https://github.com/StackStorm/st2/blob/master/CONTRIBUTING.rst#managing-python-dependencies

LindsayHill:
OK then

PR #4640, opened 2019-04-18
Added end-to-end tests to coverage the past correction of decrypt_kv filter

userlocalhost:
This is a remaining work of #4634. This adds end-to-end tests for it.

The tests in this patch call `/v1/executions` endpoint in the following conditions.

* without user specification (auth.enable = False)
* with specific user (via X-Auth-Token and St2-Api-Key)
* with specific user, but the user-scope value corresponding to the specified key is not registered

### DIFF
--- a/st2api/tests/unit/controllers/v1/test_executions_auth.py
+++ b/st2api/tests/unit/controllers/v1/test_executions_auth.py
@@ -25,12 +25,19 @@ except ImportError:
     import json
 
 import st2common.validators.api.action as action_validator
-from st2common.util import date as date_utils
+from st2common.constants.keyvalue import FULL_USER_SCOPE
+from st2common.models.api.keyvalue import KeyValuePairAPI
 from st2common.models.db.auth import TokenDB
 from st2common.models.db.auth import UserDB
+from st2common.models.db.auth import ApiKeyDB
+from st2common.models.db.keyvalue import KeyValuePairDB
 from st2common.persistence.auth import Token
 from st2common.persistence.auth import User
+from st2common.persistence.auth import ApiKey
+from st2common.persistence.keyvalue import KeyValuePair
 from st2common.transport.publishers import PoolPublisher
+from st2common.util import crypto as crypto_utils
+from st2common.util import date as date_utils
 from st2tests.api import SUPER_SECRET_PARAMETER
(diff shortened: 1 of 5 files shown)


st2api/tests/unit/controllers/v1/test_executions_auth.py
```diff
@@ -145,3 +198,30 @@ def test_post_with_st2_context_in_headers(self):
...
         self.assertEqual(resp.json['context']['parent'], context['parent'])
+
+    @mock.patch.object(ApiKey, 'get', mock.Mock(return_value=TEST_APIKEY))
+    @mock.patch.object(User, 'get_by_name', mock.Mock(return_value=TEST_USER))
+    def test_template_encrypted_params_with_apikey(self):
+        resp = self._do_post(LIVE_ACTION_DEFAULT_ENCRYPT, headers={
+            'St2-Api-key': 'secret_key'
+        })
+        self.assertEqual(resp.status_int, 201)
+        self.assertEqual(resp.json['parameters']['encrypted_param'], 'foo')
```
Kami: I guess we should probably document this - "behavior". Not related to your change directly, but those tests exposed this "potential issue" which could bite users which are not aware of this behavior.

Something along the lines of "If encrypted system datastore parameter is used for action parameter default value the unecrypted version of the value will be visible to any user which has permission to execute that action. If the raw value shouldn't be exposed you should use alternative means like retrieve encrypted value from the datastore inside the workflow."

An "alternative" would also be to modify the behavior on our side - don't expose raw value to users. Only allow admins to see them when using show secrets.

EDIT: After some more thought - this should probably already work as expected as long as the parameter is also marked as ``secret: true``. Can you please add another test case for that? Same as the existing ones, but with the parameters additionally being marked as ``secret: true``.

Kami: I've added a test case for it here - 58326dfdb.

I will also push a corresponding change. As noted in the commit message and here - using ``decrypt_kv`` on parameter which is not marked as secret could be dangerous and rarely makes sense.

In the future, we could also add a "lint" rule for that to ``st2-register-content`` which would warn user if such parameter is found.

Kami:
Merged. Thanks again for working on this!

PR #4643, opened 2019-04-23
Update orquesta version to v0.5

m4dcoder:
Update orquesta dependency in requirements.txt to v0.5.


Kami:
Can we add a changelog entry for it?

PR #4645, opened 2019-04-24
Throw a more user-friendly error on permission denied when trying to to write a pack data file

Kami:
This pull request fixes a small issue I identified while testing latest st2flow packages.

Right now if pack permissions are not configured correctly and st2api doesn't have permission to write to the pack directory / sub-directory, Internal Server Error bubbles up all the way to st2flow.

This pull request fixes so we now throw a more user-friendly (non internal server error) error message.

Screenshot from 2019-04-24 12-30-25

From st2api.log:

```bash
...
IOError: [Errno 13] Permission denied: u'/opt/stackstorm/packs/fixtures/actions/aaa' (_exception_data={},_exception_class='IOError',_exception_message=u"[Errno 13] Permission denied: u'/opt/stackstorm/packs/fixtures/actions/aaa'")
...
````

### DIFF
--- a/st2common/st2common/util/pack_management.py
+++ b/st2common/st2common/util/pack_management.py
@@ -290,7 +290,7 @@ def apply_pack_owner_group(pack_path):
 
 def apply_pack_permissions(pack_path):
     """
-    Recursively apply permission 770 to pack and its contents.
+    Recursively apply permission 775 to pack and its contents.
     """
     # These mask is same as mode = 775
     mode = stat.S_IRWXU | stat.S_IRWXG | stat.S_IROTH | stat.S_IXOTH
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by Kami** [comment]
And the actual root cause for that error was fixed in this commit - 1342f74d9 (``st2-self-check`` didn't set correct permissions).

I will cherry pick this into v3.0 branch once it's approved.

File: st2common/st2common/util/pack_management.py [hunk 1]
```diff
@@ -290,7 +290,7 @@ def apply_pack_owner_group(pack_path):
 
 def apply_pack_permissions(pack_path):
     """
-    Recursively apply permission 770 to pack and its contents.
+    Recursively apply permission 775 to pack and its contents.
```
Kami: This docstring was incorrect.

Issue #4649, opened 2019-04-26
Shell script runners don't correctly serialize "False" value for boolean parameters


Kami:
I found out this issue while working on the release.

Currently, the script runner doesn't include "False" values for boolean parameter (https://github.com/StackStorm/st2/blob/master/st2common/st2common/models/system/action.py#L277) which means it won't work correctly in scenarios where boolean script parameter is not the last parameter.

For example, from one of our actions:

Script file:

Action metadata file:

```bash
---
name: st2_chg_ver_for_st2
description: Change the version for the st2 repo
enabled: true
runner_type: remote-shell-script
entry_point: st2_chg_ver_for_st2.sh
parameters:
    project:
        type: string
        description: Project name for st2
        default: st2
        position: 0
    version:
        type: string
        description: Version to update to. Should include the patch e.g. 0.1.0
        required: true
        position: 1
    fork:
        type: string
        description: Fork to use
        default: StackStorm
        position: 2
    branch:
        type: string
        description: Branch to update
        default: master
        position: 3
    update_mistral:
        type: boolean
        description: If true, update mistralclient version in requirements.txt.
        default: false
        position: 4
    update_changelog:
        type: boolean
        description: If true, update version in changelog.
        default: false
        position: 5
    local_repo:
        type: string
        description: Location where to clone the repo. Programmatically determined if not provided.
        position: 6
    dir:
        immutable: true
        default: /home/stanley/
    sudo:
        immutable: true
        default: false
    cmd:
        immutable: true
        default: ""
    kwarg_op:
        immutable: true
        default: "--"
```

```bash
...
#!/bin/bash
set -e

PROJECT=$1
VERSION=$2
FORK=$3
BRANCH=$4
UPDATE_MISTRAL=$5
UPDATE_CHANGELOG=$6
LOCAL_REPO=$7
GIT_REPO="git@github.com:${FORK}/${PROJECT}.git"
CWD=`pwd`
PUSH=0
....
```

In this case, if ``update_mistral`` and ``update_changelog`` parameters default to ``False``, script will get incorrectly called with the following arguments ``st2_chg_ver_for_st2.sh st2 3.1dev StackStorm master st2_1556296602_602`` which is incorrect since argument number 4 will now be ``local_repo`` instead of ``update_mistral``.

[...]

PR #4653, opened 2019-04-29
Add tests for render_final_params and local and remote script runner command construction

Kami:
This pull request adds unit tests for ``render_final_params`` function which we were lacking (we did have various integration and end to end tests, but not unit ones).

It also adds more command construction unit tests for local and remote shell script runners.

## Background / Context

I started to work on this issue - https://github.com/StackStorm/st2/issues/4649#issuecomment-487611846.

First step was adding unit tests which were lacking which supposed to fail before I fix the bug. But after digging in and adding unit tests, I simply couldn't reproduce the original issue. 

Tests pass and the code itself looks correct. I'm still not sure why this issue happened on CICD (it could be consistently reproduced and there are logs), but for now we at least have some more unit tests we were lacking before.

I will do some more digging in, but for the time being, I can't replicate it locally using latest master (it could be related to an issue in older version, we did make some param util changes and fix in v3.0.0 albeit unrelated ones).


Kami:
I also confirmed it on new cicd node which runs v3.0.0 and it works fine there. So likely issue related to v2.10.x.

Once this PR is merged, I will close #4649 as can't reproduce / already fixed.

PR #4656, opened 2019-05-02
Revert part of #4609, fixes #4650

### LINKED ISSUE #4650: 3.0 - `source_channel` no longer working for Mistral workflows
##### SUMMARY

In ST2 3.0, Mistral workflows no longer able to identify `source_channel` that alias was called from. 

##### ISSUE TYPE
 - Bug Report

##### STACKSTORM VERSION

`st2 3.0.0, on Python 2.7.12`

one-liner, on Ubuntu 16.04. 

System upgraded from 2.10.4 to 3.0.0.

##### STEPS TO REPRODUCE

Create a Mistral workflow that uses `source_channel`, and call it via a Chat alias.

Mistral workflow [here](https://github.com/StackStorm/st2_demos/blob/master/actions/workflows/multicap.yaml#L20).

Alias [here](https://github.com/StackStorm/st2_demos/blob/master/aliases/multicap.yaml).

Key lines:
```yaml
input:
    channel: "<% env().get('__actions').get('st2.action').st2_context.parent.source_channel %>"
    message: "Setting up environment and starting captures now...{~}"
on-success:
```

##### EXPECTED RESULTS

Workflow should post back to the original channel with my message 

> Setting up environment and starting captures now...

##### ACTUAL RESULTS

Workflow fails, because it can't get the value of source_channel. Error similar to

```
"Failed to run task [error=Can not evaluate YAQL expression [expression=env().get('__actions').get('st2.action').st2_context.parent.source_channel, error=u'source_channel', data={}], wf=st2_demos.multicap.main, task=notify_slack]:
```

This stems from https://github.com/StackStorm/st2/pull/4609/files#diff-f2911a69b88d56311f8945335e9457b7L148, which is part of PR #4609. If I add that removed line back to `aliasexecution.py`, it works as expected.

Looks like this data may be available in the `notify` dict.

/cc @blag @m4dcoder @emptywee

blag:
One line fix for `source_channel`.

In the future, it would be nice to simply use one `source_context` variable to store all of the information from the chat provider, but this fixes previous behavior.

Reverts part of #4609.
Fixes #4650.


LindsayHill:
Please also add a Changelog entry

Kami:
Yeah, let's please also add a test case for it, so we avoid such regressions in the future.

PR #4657, opened 2019-05-02
Add flake8 checks for license and copyright headers

m4dcoder:
Add custom flake8 extension to check that the Apache 2.0 license is included in code files.
Add Apache 2.0 license header to files that does not have it. Fix Apache 2.0 license header if it
doesn't match the header that is expected.


m4dcoder:
@Kami @bigmstone I'm going to fix this first and then add an extension to check copyright header.

Makefile
```diff
@@ -436,7 +436,7 @@ requirements: virtualenv .sdist-requirements install-runners
 	(cd ${ROOT_DIR}/st2common; ${ROOT_DIR}/$(VIRTUALENV_DIR)/bin/python setup.py develop --no-deps)
 
 	# Some of the tests rely on submodule so we need to make sure submodules are check out
-	git submodule update --init --recursive
+	git submodule update --recursive --remote
```
Kami: Just curious why this change is needed.

test-requirements.txt
```diff
@@ -2,6 +2,7 @@
 coverage==4.5.2
 pep8==1.7.1
 flake8==3.7.7
+hacking
```
Kami: Probably not a bad idea to pin it to a specific version to avoid surprises when a new version if published / similar. This has bitten us many times already in the past, same for ``flake8-copyright``.

Although checking ``hacking`` docs (https://pypi.org/project/hacking/) it looks like ``hacking`` itself installs specific versions of ``pep8``, etc. so there could be potential conflicts down the road.

m4dcoder: @Kami Pinned both flake8-copyright and hacking in test-requirements.txt.  Why did we pin a specific version of pep8 (pep8==1.7.1)?

Kami: New versions (> 2.0.0, IIRC), only work with Python 3.

PR #4661, opened 2019-05-05
Correctly use port specified in the SSH config file in the remote SSH runner

### LINKED ISSUE #4660: Paramiko SSH runner does not override ssh port from ssh config
##### SUMMARY
Paramiko SSH runner does not override ssh port from ssh config. It always use 22 port, I can override it only via runner parameters.

##### ISSUE TYPE
 - Bug Report

##### STACKSTORM VERSION
st2 3.0dev (f7a680a), on Python 2.7.12

- Centos 7, Python 2.7
- k8s (openshift)

##### STEPS TO REPRODUCE
**Config section**
```ini
    [ssh_runner]
    remote_dir = /tmp
    use_ssh_config = True
    ssh_config_file_path = ~/.ssh/config
```
**ssh config**
```
Host *
StrictHostKeyChecking no
ServerAliveInterval 60
ServerAliveCountMax 60
UserKnownHostsFile=/dev/null
ForwardAgent yes
Port 2200
HostName %h
```

**Action** (instance is action parameter)
```yaml
chain:
  - name: disk_use
    ref: core.remote
    parameters:
      cmd: df -Ph
      hosts: "{{ instance }}"
    publish:
      result: |
        {% for host, result in disk_use.items() %}
        {{ result.stdout }}
        {% endfor %}

```

##### EXPECTED RESULTS
Will connect to server via SSH port *2200* 

##### ACTUAL RESULTS
```json
"traceback": "Traceback (most recent call last):\n  File \"/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/runners/parallel_ssh.py\", line 261, in _connect\n    client.connect()\n  File \"/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/runners/paramiko_ssh.py\", line 144, in connect\n    self.client = self._connect(host=self.hostname, socket=self.bastion_socket)\n  File \"/opt/stackstorm/st2/local/lib/python2.7/site-packages/st2common/runners/paramiko_ssh.py\", line 702, in _connect\n    client.connect(**conninfo)\n  File \"/opt/stackstorm/st2/local/lib/python2.7/site-packages/paramiko/client.py\", line 362, in connect\n    raise NoValidConnectionsError(errors)\nNoValidConnectionsError: [Errno None] Unable to connect to port 22 on <IP>\n", 
```

Kami:
This pull request updates / fixes remote command and script runner to correctly use SSH port specified in the SSH config file if a non-default port is specified in the SSH config file and if SSH config file functionality is enabled (``ssh_runner.use_ssh_config`` st2.conf config option is set to ``True``).

NOTE: Due to the way of how we specify and handle default parameter values in the runner metadata file, this change is not 100% ideal, but it should cover most scenarios (Current code make it impossible to distinguish if a default parameter value is used or if user specified a value which matches a default value. Changing this would be more involved and also involves other parameters).

Resolves #4660.


Kami:
I made a change so we don't include a default value for the ``port`` parameter in ``runner.yaml`` file anymore.

This is the same approach we use for other default values which can also be read from the config (username, port, etc.) and it's more robust since it means it covers and works correctly in all the scenarios.

contrib/runners/remote_runner/remote_runner/runner.yaml
```diff
@@ -6,8 +6,7 @@
   runner_parameters:
     bastion_host:
       description: 'The host SSH connections will be proxied through. Note: This connection
-        is made using the same parameters as the final connection, and is only used
```
Kami: This description was old and out of date since we don't use fabric runner for a long time anymore.

PR #4664, opened 2019-05-07
Update lint configs subtree

Kami:
This pull request updates lint-configs/ git subtree with change in https://github.com/StackStorm/lint-configs/pull/1.

#4657 committed lint-config change directly to this repo instead of updating lint-configs git repo so it broke other repos which depends on requirements.txt files from StackStorm/st2 (private repos, CI repos, etc.).


Kami:
Will see if this works.

We may need two flake8 configs - one for core StackStorm repos (the ones which require Copyright headerr) and one for other repos (e.g. StackStorm Exchange ones).

Kami:
I also enabled "H" checks so missing Apache 2.0 license header errors are reported for projects with ASF 2.0 license.

Kami:
I will merge this for now, but we should still keep an eye on other repos and fix things there, if necessary.

We may also need to split flake8 config into two or three versions - one for open source projects, other one for private / proprietary ones and StackStorm exchange ones. Will see.

Kami:
Superseded by #4672.

PR #4666, opened 2019-05-08
Fix PYTHONPATH handling for Python runner actions using --python3 flag

### LINKED ISSUE #4658: st2 python3 interpreter is falling back to importing a python2 lib that is incompatible with python3
##### SUMMARY

 One of my libraries is trying to import `cassandra.cluster`.  That library is trying to import `from concurrent.futures import ThreadPoolExecutor, FIRST_COMPLETED, wait as wait_futures` - which, as I understand, is built into python3 (and a separate library for compatibility with python3)

##### ISSUE TYPE
 - Bug Report

##### STACKSTORM VERSION
```
# st2 --version
st2 3.0.0, on Python 2.7.6
```
Docker container running Ubuntu 14.04

##### STEPS TO REPRODUCE
Create a python runner in your python3-specific pack.  Inside the runner import cassandra libs and just create an object.
```
from cassandra.cluster import Cluster
        cluster = Cluster()
```

##### EXPECTED RESULTS
I expect the library to import and the object to initialize

##### ACTUAL RESULTS
st2 python3 falls back to python2 to import the lib and it throws an exception similar to
```
  File \"/opt/stackstorm/packs/ostk_common/actions/lib/ostkdbs.py\", line 2, in <module>
    from cassandra.cluster import Cluster
  File \"/opt/stackstorm/virtualenvs/ostk_common/lib/python3.5/site-packages/cassandra/cluster.py\", line 23, in <module>
    from concurrent.futures import ThreadPoolExecutor, FIRST_COMPLETED, wait as wait_futures
  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/concurrent/futures/__init__.py\", line 8, in <module>
    from concurrent.futures._base import (FIRST_COMPLETED,
  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/concurrent/futures/_base.py\", line 414
    raise exception_type, self._exception, self._traceback
                        ^
SyntaxError: invalid syntax
```

Kami:
This pull request fixes a bug which would cause some packs to break when running Python 3 actions on installations where StackStorm components run under Python 2.

## Background, Context

Some pack dependencies or action code depend on packages / modules which are available in Python 3 standard library, but they are also available in Python 2 site-packages directory (usually as an installable backport).

We didn't correctly add base Python 3 library directory (e.g. ``/usr/lib/python3.6``) to ``PYTHONPATH`` for such actions so things broke because it tried to use incompatible package / module from Python 2 instead from Python 3.

This pull request fixes that.

## TODO

- [x] End to end tests (I will add a test pack which contains an action which relies on such package)

Resolves #4658.


st2common/st2common/util/sandboxing.py
```diff
@@ -160,8 +160,30 @@ def get_sandbox_python_path_for_python_action(pack, inherit_from_parent=True,
...
                                                        virtualenv_directories[0],
                                                        'site-packages')
-        sandbox_python_path = (python3_lib_directory + ':' + python3_site_packages_directory + ':' +
-                               pack_actions_lib_paths + ':' + sandbox_python_path)
+
+        # Work around to make sure we also add system lib dir to PYTHONPATH and not just virtualenv
+        # one
+        # NOTE: abc.py is always available in base lib directory which is symlinked to virtualenv
+        # lib directory
+        abc_module_path = os.path.join(python3_lib_directory, 'abc.py')
```
Kami: I'm not too happy about this approach, but from the research I did, I couldn't come up with a better approach.

I will also look into it more when adding end to end tests and testing it under various Python 3 versions on all the supported versions to verify it indeed covers all the scenarios and works correctly on all the platforms.

oatsgithub:
We're still experiencing the same issue with StackStorm HA. We have 4 internal packs and 2 of them are failing because StackStorm tries to execute their actions in Python 2 (we don't have any Python 2 code).

requirements.txt:
```
hvac
python-consul
requests
```

All our packs have these libraries but only 2 of them try to run in Python 2.

Kami:
@oatsgithub Please open a new issue for that and include all the relevant information we need to be able to reproduce the issue (StackStorm versions, packs and actions in question, action runner debug level logs, etc.).

blag:
This is a great candidate for a bigfix release of v3.1.

PR #4668, opened 2019-05-08
isotime formatter: assume incoming epoch timestamp being in UTC

emptywee:
When the platform where st2 is running is configured with non UTC timezone executions retrieved from st2api have `log` entries with incorrectly formatted timestamp as shown on this screenshot:

image

After debugging it, turns out that api model is receiving them as unix timestamps, parsing them as local time producing naive datetime object and further assuming it's in UTC timezone, resulting in wrong timestamps provided via API calls.

This PR fixes that, but I am not sure if anything else may get affected. Maybe a proper way would be giving non-formatted timestamp and let API consumers deal with time zones?

@Kami please, take a look and chime in. Others welcome too.


Kami:
Thanks.

The change look good to me, but in such scenarios I would actually start with some unit tests (I will look into adding some) :)

Kami:
I was able to confirm the issue.

It's indeed related to how we handle time objects. For precision reasons we store timestamps as number of microseconds since the UNIX epoch and the format function didn't handle that correctly if a server time was not set to UTC (this is usually not a good practice, but anyway) - it assumed it was in local time while it was actually in UTC.

I will also push tests to this PR.

Thanks again and a good catch :+1:

Kami:
Merged, thanks!

emptywee:
Good deal! Thanks for looking into this with me :)

PR #4672, opened 2019-05-10
 Replace hacking with flake8 plugin to check license header

m4dcoder:
Replace hacking module and use the custom flake8 plugin in the StackStorm/lint-configs repo to check the Apache 2.0 license header.


Makefile
```diff
@@ -251,6 +251,7 @@ flake8: requirements .flake8
 	@echo
 	@echo "==================== flake ===================="
 	@echo
+	. $(VIRTUALENV_DIR)/bin/activate; cd ./lint-configs/python/flake8_plugins; python setup.py develop
```
Kami: That works for now although this usually results in a lot of issues if package is updated or similar so if we will have issues with it, we should move it to a separate repo and install it from git using using pip.

m4dcoder: @Kami Yeah, this is temporary until we setup the flake8 plugins repo and upload the package into pypi.

PR #4675, opened 2019-05-13
Fix a possible shell command injection in the linux.service action

Kami:
This pull request fixes a possible shell command injection in ``linux.service`` action.

## Background,  Context

The code didn't escape ``service`` and ``action`` parameter passed to ``subprocess.Popen`` and used ``shell=True``.

This pull request fixes the action to use ``shell=False`` and passes command to ``subprocess.Popen`` as a list of arguments instead of as a string (in this scenario, values are automatically escaped by ``subprocess.Popen``).

This issue was reported to us by James Robinson (Netskope and Veracode).


contrib/linux/actions/service.py
```diff
@@ -20,20 +20,25 @@
...
 
+from st2common.util.shell import quote_unix
+
 distro = platform.linux_distribution()[0]
 
-args = {'act': sys.argv[1], 'service': sys.argv[2]}
+if len(sys.argv) < 3:
+    raise ValueError('Usage: service.py <action> <service>')
+
+args = {'act': quote_unix(sys.argv[1]), 'service': quote_unix(sys.argv[2])}
```
Kami: This is now not strictly needed since ``subprocess.Popen`` already takes case of that when a list of args and ``shell=False`` is used.

PR #4683, opened 2019-05-16
Update st2rc.sample.ini

punkrokk:
I've added in the [stream] stanza and examples for remote st2client functionality. 
Please let me know anissue should be created. 

I couldn’t use the —tail option without the stream stanza

Fixes: #4684 

I'm not sure if this page needs to be updated - https://github.com/StackStorm/st2/tree/master/st2client

I don't know what a reactor and some other terms I'm not aware of are there, but it may make sense. cc @LindsayHill

### DIFF
--- a/conf/st2rc.sample.ini
+++ b/conf/st2rc.sample.ini
@@ -22,15 +22,15 @@ password = testpassword
 
 [api]
 url = http://127.0.0.1:9101/v1
-# or from remote
-# url = http://127.0.0.1/api/v1
+# or for a remote instance
+# url = https://example.com/api/v1
 
 [auth]
 url = http://127.0.0.1:9100/
 # or from remote:
-#url = https://10.0.1.16/auth
+#url = https://example.com/auth/v1
 
 [stream]
 url = http://127.0.0.1:9102
 #or from remote:
-# url = http://10.10.10.10:/api/v1/stream
+# url = https:/example.com/stream/v1
--- a/conf/st2rc.sample.ini
+++ b/conf/st2rc.sample.ini
@@ -22,6 +22,15 @@ password = testpassword
 
 [api]
 url = http://127.0.0.1:9101/v1
+# or from remote
+# url = http://127.0.0.1/api/v1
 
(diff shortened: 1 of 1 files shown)


conf/st2rc.sample.ini
```diff
@@ -22,6 +22,15 @@ password = testpassword
 
 [api]
 url = http://127.0.0.1:9101/v1
+# or from remote
+# url = http://127.0.0.1/api/v1
```
Kami: I assume here it should be some other external domain / IP and not, 127.0.0.1, right?

Perhaps we should use example.com?

Kami:
Thanks for the contribution :+1: 

I updated the code to use ``example.com`` for remote deployment and also updated the URLs to use correct paths.

conf/st2rc.sample.ini
```diff
@@ -22,6 +22,15 @@ password = testpassword
 
 [api]
 url = http://127.0.0.1:9101/v1
+# or for a remote instance
+# url = https://example.com/api/v1
 
 [auth]
 url = http://127.0.0.1:9100/
+# or for a remote instance
```
Kami: I'll actually confirm it works correctly with the paths I used.

it's a bit annoying those URLs are not consistent, but making them consistent might require code changes :)

Kami: Confirmed those URLs work correctly with the default nginx config so the same consistent suffix can be used everywhere.

punkrokk:
Thanks all. Good call on example.com. 

Any thoughts about the reactor stuff? Is that legacy terminology?

Kami:
I'm not sure what you mean / refer to with reactor.

punkrokk:
Never mind. I saw it somewhere. But can’t find it again.

punkrokk:
Basically I saw a reactor referenced in the read me and I had no idea what it was

Kami:
@punkrokk Ah, I see.

It looks like readme is out of date and needs some cleanup :)

PR #4690, opened 2019-05-22
Limit database index field names to 65 characters

### LINKED ISSUE #4688: MongoDB collection/index name used by ST2 is too long for AWS DocumentDB
##### SUMMARY

While migrating from ST2 v2.9 to ST2 v3.0, we encountered the following problem: one MongoDB index name used by ST2, namely `st2.action_execution_scheduling_queue_item_d_b.scheduled_start_timestamp` (C.F. [st2/st2common/st2common/models/db/execution_queue.py](https://github.com/StackStorm/st2/blob/master/st2common/st2common/models/db/execution_queue.py)) **is too long for AWS DocumentDB**. It _does_ work with an **official** MongoDB server.

AWS DocumentDB _should_ be compatible/honor the official limits of MongoDB (here, 127 characters for fully qualified index names), but I contacted the AWS DocumentDB support which confirmed me that it was **a known issue in DocumentDB and that they didn't have an estimated time of resolution for this problem**. They suggested me to **shorten the index name** if I can, but I think it's currently not customizable (which is okay, this should not be needed).

I realize that **this is not a StackStorm bug _per se_**, however AWS DocumentDB is a big player and it might be interesting for you to shorten this index name so ST2 users/customers can use DocumentDB with ST2 V3+.

##### ISSUE TYPE
 - Bug Report

##### STACKSTORM VERSION
st2 3.0.0, on Python 2.7.12

N/A

##### STEPS TO REPRODUCE
* Install StackStorm V3.0+
* Configure an AWS DocumentDB as the MongoDB server
* Run a command that will try to create the DB, e.g. `st2ctl reload --register-all`

##### EXPECTED RESULTS
`st2ctl reload --register-all` should successfully create the DB and work.

##### ACTUAL RESULTS
`st2ctl reload --register-all` exits with an error because it can't create the index name `st2.action_execution_scheduling_queue_item_d_b.scheduled_start_timestamp`.

Kami:
This pull request limits full database index names to 65 characters.

### Background, Context

For the most part we don't explicitly specify index name when we define index in the code. This means MongoDB automatically infer the name from db, collection and field name. For example ``st2.action_execution_d_b.status.action.ref.start_timestamp`` (``<db name>.<collection name>.<all the indexed field names>``.

This pull request explicitly specifies shorter index names for some indexes which don't work with AWS DocumentDB due to the limit (MongoDB itself claims it supports index names up to 128 characters, but that doesn't work with DocumentDB).

We don't officially support AWS DocumentDB, but it's a relatively easy change so it doesn't hurt us to include it.

In addition to that, there is a bigger long term question about MongoDB 3.6 compatibility once we switch to MongoDB 4.0 by default (which is not supported by AWS due to the MongoDB licensing change). Right now our code works with 3.6 and 4.0 and we don't really on any specific 4.0 features, but this could change in the future.

Here is an example of the offending index name - ``st2.action_execution_scheduling_queue_item_d_b.original_start_timestamp``.

As you can see, the collection name itself is also very long, but that's very hard / impossible to change that at the moment, because we don't have a migration framework in place for changing collection names. So only option we are left with, is shortening index names (which is fine).

I've added a test case so we will find out about such issues earlier in the future (during the development process).

Resolves #4688.

### DIFF
--- a/st2common/tests/unit/test_db.py
+++ b/st2common/tests/unit/test_db.py
@@ -38,6 +38,9 @@ from st2common.persistence.rule import Rule
 from st2common.persistence.trigger import TriggerType, Trigger, TriggerInstance
 from st2tests import DbTestCase
 
+from unittest2 import TestCase
+from st2tests.base import ALL_MODELS
+
 
 __all__ = [
     'DbConnectionTestCase',
@@ -49,6 +52,40 @@ __all__ = [
 
 SKIP_DELETE = False
 DUMMY_DESCRIPTION = 'Sample Description.'
+
+
+class DbIndexNameTestCase(TestCase):
+    """
+    Test which verifies that model index name are not longer than the specified limit.
+    """
+    LIMIT = 70
+
+    def test_index_name_length(self):
+        db_name = 'st2'
+
+        for model in ALL_MODELS:
+            collection_name = model._get_collection_name()
+            model_indexes = model._meta['index_specs']
+
+            for index_specs in model_indexes:
+                index_name = index_specs.get('name', None)
+                if index_name:
+                    # Custom index name defined by the developer
+                    index_field_name = index_name
+                else:
+                    # No explicit index name specified, one is auto-generated using
+                    # <db name>.<collection name>.<index field names> schema
+                    index_fields = dict(index_specs['fields']).keys()
+                    index_field_name = '.'.join(index_fields)
+
+                index_name = '%s.%s.%s' % (db_name, collection_name, index_field_name)
+                print index_name
+                continue
+
+                if len(index_name) > self.LIMIT:
+                    self.fail('Index name "%s" for model "%s" is longer than %s characters. '
+                              'Please manually define name for this index so it\'s shorter than '
+                              'that' % (index_name, model.__name__, self.LIMIT))
 
 
 class DbConnectionTestCase(DbTestCase):
--- a/st2common/tests/unit/test_db.py
+++ b/st2common/tests/unit/test_db.py
@@ -58,7 +58,7 @@ class DbIndexNameTestCase(TestCase):
     """
     Test which verifies that model index name are not longer than the specified limit.
     """
-    LIMIT = 70
+    LIMIT = 65
 
     def test_index_name_length(self):
         db_name = 'st2'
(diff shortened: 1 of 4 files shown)


GuiTeK:
> In addition to that, there is a bigger long term question about MongoDB 3.6 compatibility once we switch to MongoDB 4.0 by default (which is not supported by AWS due to the MongoDB licensing change). Right now our code works with 3.6 and 4.0 and we don't really on any specific 4.0 features, but this could change in the future.

Indeed, because if StackStorm uses MongoDB 4.0 features, we won't be able to use any cloud MongoDB solutions anymore (AWS DocumentDB, Azure CosmosDB, probably others). This is an important thing to consider because the alternative, maintaining a high-availability MongoDB cluster, has a non-negligible cost.

Thanks again for taking care of this issue so quickly!

Kami:
> Indeed, because if StackStorm uses MongoDB 4.0 features, we won't be able to use any cloud MongoDB solutions anymore (AWS DocumentDB, Azure CosmosDB, probably others). This is an important thing to consider because the alternative, maintaining a high-availability MongoDB cluster, has a non-negligible cost.

Correct and I agree.

While we may not have a choice, but to use MongoDB 4.0 on Ubuntu Bionic and other distros where 4.0 is available in standard repos, we can probably still keep our code compatible with MongoDB 3.6 (will see how this develops though, especially if pymongo and mongoengine add functionality which only works with 4.0).

Kami:
Note: I confirmed that this change has no negative implications as far as backward compatibility goes.

Renaming the index will simply cause old index to be dropped and a new one created during service bootstrap phase without any breaking or negative consequences.

st2common/tests/unit/test_db.py
```diff
@@ -51,6 +54,38 @@
 DUMMY_DESCRIPTION = 'Sample Description.'
 
 
+class DbIndexNameTestCase(TestCase):
+    """
+    Test which verifies that model index name are not longer than the specified limit.
+    """
+    LIMIT = 65
```
m4dcoder: Minor nit-pick. Make `LIMIT` lowercase if used in the class or make it a constant at top of module.  This is more consistent with our coding.

PR #4692, opened 2019-05-22
Refactor the orquesta st2kv function

### LINKED ISSUE #4678: Orquesta and st2kv() bug/regression
##### SUMMARY

Orquesta st2kv() behavior changed drastically, making it hard to migrate from Mistral engine.
1. It now throws exception if the specified key doesn't exist in the datastore (previously it would return null). Exceptions seem to be impossible to handle within YAQL/Jinja expressions.
2. Even if the key exists, but has an empty string value, it still throws an exception. See https://github.com/StackStorm/st2/blob/master/contrib/runners/orquesta_runner/orquesta_functions/st2kv.py#L48 - this is not the best option chosen for the if condition, in my humble opinion.

##### ISSUE TYPE
 - Bug Report

##### STACKSTORM VERSION
Paste the output of `st2 --version`:
```
$ st2 --version
st2 3.0.0, on Python 2.7.10
```

CentOS 6, custom install.

##### STEPS TO REPRODUCE

Meta:
```yaml
---
name: wf_orchestra_test
pack: playground
description: Test
runner_type: orquesta
entry_point: workflows/wf_orchestra_test.yaml
enabled: true
parameters:
  name:
    required: true
    type: string
```
Workflow:
```yaml
---
version: '1.0'
input: 
  - name
tasks:
  task1:
    action: core.noop
    next:
      - when: <% succeeded() %>
        publish:
          - val: <% st2kv('system.' + ctx().name) %>
```

##### EXPECTED RESULTS

`null` published to the `val` variable

##### ACTUAL RESULTS

```
$ st2 run playground.wf_orchestra_test name=ttt
.
id: 5cdc88702db7aa204ea689d9
action.ref: playground.wf_orchestra_test
parameters: 
  name: ttt
status: failed
start_timestamp: Wed, 15 May 2019 16:45:20 CDT
end_timestamp: Wed, 15 May 2019 16:45:21 CDT
result: 
  errors:
  - message: 'YaqlEvaluationException: Unable to evaluate expression ''<% st2kv(''system.'' + ctx().name) %>''. ExpressionEvaluationException: Key system.ttt does not exist in StackStorm datastore.'
    route: 0
    task_id: task1
    task_transition_id: noop__t0
    type: error
  output: null

```

And to illustrate item 2:

```
$ st2 key set ttt ''
+------------------+--------------+
| Property         | Value        |
+------------------+--------------+
| name             | ttt          |
| value            |              |
| scope            | st2kv.system |
| expire_timestamp |              |
+------------------+--------------+
$ st2 key get ttt
+------------------+--------------+
| Property         | Value        |
+------------------+--------------+
| name             | ttt          |
| value            |              |
| secret           | False        |
| encrypted        | False        |
| scope            | st2kv.system |
| expire_timestamp |              |
+------------------+--------------+
$ st2 run playground.wf_orchestra_test name=ttt
.
id: 5cdc88942db7aa204ea689dd
action.ref: playground.wf_orchestra_test
parameters: 
  name: ttt
status: failed
start_timestamp: Wed, 15 May 2019 16:45:56 CDT
end_timestamp: Wed, 15 May 2019 16:45:57 CDT
result: 
  errors:
  - message: 'YaqlEvaluationException: Unable to evaluate expression ''<% st2kv(''system.'' + ctx().name) %>''. ExpressionEvaluationException: Key system.ttt does not exist in StackStorm datastore.'
    route: 0
    task_id: task1
    task_transition_id: noop__t0
    type: error
  output: null
```

m4dcoder:
Fix the orquesta st2kv to return empty string and null values correctly and allow st2kv function to return a supplied default value for nonexistent key. Closes #4678

### DIFF
--- a/contrib/runners/orquesta_runner/orquesta_functions/st2kv.py
+++ b/contrib/runners/orquesta_runner/orquesta_functions/st2kv.py
@@ -45,9 +45,13 @@ def st2kv_(context, key, **kwargs):
         raise Exception('Failed to retrieve User object for user "%s" % (username)' %
                         (six.text_type(e)))
 
-    try:
-        kvp = kvp_util.get_key(key=key, user_db=user_db, decrypt=decrypt)
-    except Exception as e:
-        raise exc.ExpressionEvaluationException(str(e))
+    has_default = 'default' in kwargs
+    default_value = kwargs.get('default')
 
-    return kvp
+    try:
+        return kvp_util.get_key(key=key, user_db=user_db, decrypt=decrypt)
+    except Exception as e:
+        if not has_default:
+            raise exc.ExpressionEvaluationException(str(e))
+        else:
+            return default_value
--- a/contrib/runners/orquesta_runner/orquesta_functions/st2kv.py
+++ b/contrib/runners/orquesta_runner/orquesta_functions/st2kv.py
@@ -43,11 +43,9 @@ def st2kv_(context, key, decrypt=False):
         raise Exception('Failed to retrieve User object for user "%s" % (username)' %
                         (six.text_type(e)))
 
-    kvp = kvp_util.get_key(key=key, user_db=user_db, decrypt=decrypt)
-
-    if not kvp:
-        raise exc.ExpressionEvaluationException(
-            'Key %s does not exist in StackStorm datastore.' % key
-        )
+    try:
+        kvp = kvp_util.get_key(key=key, user_db=user_db, decrypt=decrypt)
+    except Exception as e:
+        raise exc.ExpressionEvaluationException(str(e))
 
     return kvp
--- a/contrib/runners/orquesta_runner/orquesta_functions/st2kv.py
+++ b/contrib/runners/orquesta_runner/orquesta_functions/st2kv.py
@@ -25,9 +25,11 @@ from st2common.util import keyvalue as kvp_util
 LOG = logging.getLogger(__name__)
 
 
-def st2kv_(context, key, decrypt=False):
+def st2kv_(context, key, **kwargs):
     if not isinstance(key, six.string_types):
         raise TypeError('Given key is not typeof string.')
+
+    decrypt = kwargs.get('decrypt', False)
 
     if not isinstance(decrypt, bool):
         raise TypeError('Decrypt parameter is not typeof bool.')
--- a/contrib/runners/orquesta_runner/orquesta_functions/st2kv.py
+++ b/contrib/runners/orquesta_runner/orquesta_functions/st2kv.py
@@ -18,6 +18,7 @@ import six
 
 from orquesta import exceptions as exc
 
+from st2common.exceptions import db as db_exc
 from st2common.persistence import auth as auth_db_access
 from st2common.util import keyvalue as kvp_util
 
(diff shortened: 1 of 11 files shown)


contrib/runners/orquesta_runner/orquesta_functions/st2kv.py
```diff
@@ -25,10 +25,12 @@
 LOG = logging.getLogger(__name__)
 
 
-def st2kv_(context, key, decrypt=False):
+def st2kv_(context, key, **kwargs):
```
Kami: Can we explicitly declare keyword arguments here? Not a fan of ``**kwargs`` since it doesn't make it immediately obvious which arguments a function takes.

I assume you did that to being able to distinguish if a default value is specified or not. Another way to handle that would be to set a default value of this keyword argument to some special value.

st2common/st2common/persistence/keyvalue.py
```diff
@@ -107,6 +109,11 @@ def get_by_scope_and_name(cls, scope, name):
         :rtype: :class:`KeyValuePairDB` or ``None``
         """
         query_result = cls.impl.query(scope=scope, name=name)
+
+        if not query_result:
```
Kami: Wonder how this change affects other code like datastore service, etc. which uses this function.

Also looking at the code below, we used to return ``None`` if value didn't exist so just need to make sure all the code which uses it has also been updated and it has no negative side affects.

st2common/st2common/util/keyvalue.py
```diff
@@ -110,12 +110,11 @@ def get_key(key=None, user_db=None, scope=None, decrypt=False):
...
+    kvp = KeyValuePair.get_by_scope_and_name(scope, key_id)
 
-    if value:
-        return deserialize_key_value(
-            value.value,
-            decrypt
-        )
+    # Decrypt in deserialize_key_value cannot handle NoneType.
+    if kvp.value is None:
+        return kvp.value
```
Kami: Wonder in which scenarios we could end up here since we only support strings for datastore values aka when ``None`` can be set / returned for a datastore value attribute.

contrib/runners/orquesta_runner/orquesta_functions/st2kv.py
m4dcoder: Yes. I did this to distinguish whether default value is actually provided and not to be confused if someone wants default to be `None`.  Use of kwargs is more elegant in this case IMO. But if this is what's blocking, then I can assign some bogus value to default.  I'll call it `KAMI_DONT_WANT_KWARGS`.  :P

st2common/st2common/persistence/keyvalue.py
m4dcoder: Yep, it's handled in the LookUp classes.

st2common/st2common/util/keyvalue.py
m4dcoder: That may be true from normal usage.  But nothing is preventing a None value from being written into the KeyValuePairDB record.

m4dcoder:
@Kami TBH, the keyvalue/datastore utils/services are very inconsistent.  Most of our services modules are just functions.  But the keyvalue service have separate classes in them.  And then we have datastore service which is confusing.  The functions in the keyvalue util should probably be in the service module.  They would be more consistent with functions of other services.  And then low level DB query should return error when record not found.  The utils/service can abstract that and decide whether to throw exception.

contrib/runners/orquesta_runner/orquesta_functions/st2kv.py
m4dcoder: @Kami Looks like I won't be able to change to an explicit arg for `default`.  When I set the default to some constant value, there may be a bug in YAQL that tries to make the parameter strongly typed.  When I try to use the function in YAQL expression, YAQL returns error saying it cannot recognize the signature.  If I set `default=None`, there's o problem.  I also try this with Jinja and there's no problem setting `default="SOME_CONSTANTS"`.  So at this point, you'll have to live with using kwargs here.

PR #4720, opened 2019-06-22
Fix expression function to_yaml_string to handle mongodb base types

### LINKED ISSUE #4700: Orquesta - Unable to serialize list objects
##### SUMMARY

When running an Orquesta workflow and trying to use the `to_yaml_string` filter with a list object, an error is raised even though the object is valid JSON.

##### ISSUE TYPE
 - Bug Report

##### STACKSTORM VERSION

```
$ st2 --version
st2 3.0.0, on Python 2.7.5
```

```
OS = Centos 7.6
Install = puppet-st2
```

##### STEPS TO REPRODUCE
`/opt/stackstorm/packs/default/actions/orquesta_list_yaml_error.yaml`
```yaml
---
name: orquesta_list_yaml_error
description: A basic sequential workflow.
runner_type: orquesta
entry_point: workflows/orquesta_list_yaml_error.yaml
enabled: true
parameters: {}
```

`/opt/stackstorm/packs/default/actions/workflows/orquesta_list_yaml_error.yaml`
```yaml
version: 1.0
  
description: A basic sequential workflow.

vars:
  - blah:
      - a
      - b
      - c

tasks:
  demo:
    action: core.local
    input:
      cmd: "echo '{{ ctx().blah | to_yaml_string }}'"
```

##### EXPECTED RESULTS

Workflow runs successfully

##### ACTUAL RESULTS

```
$ st2 run default.orquesta_list_yaml_error
.
id: 5cf5df6b9387ef5a16b63f71
action.ref: default.orquesta_list_yaml_error
parameters: None
status: failed
start_timestamp: Tue, 04 Jun 2019 03:03:07 UTC
end_timestamp: Tue, 04 Jun 2019 03:03:07 UTC
result: 
  errors:
  - message: 'JinjaEvaluationException: Unable to evaluate expression ''{{ ctx().blah | to_yaml_string }}''. RepresenterError: (''cannot represent an object'', [u''a'', u''b'', u''c''])'
    route: 0
    task_id: demo                                                                                                            
    type: error                                                                                                              
  output: null 
```

m4dcoder:
Refactor the filter to_yaml_string used in YAQL/Jinja expressions to be able to handle mongodb base types such as BaseDict and BaseList. The filter will call a helper function to recursively convert BaseDict and BaseList to python dict and list respectively. Fixes #4700.

### DIFF
--- a/st2common/st2common/util/db.py
+++ b/st2common/st2common/util/db.py
@@ -19,16 +19,16 @@ import six
 
 
 def mongodb_to_python_types(value):
+    # Convert MongoDB BaseDict and BaseList types to python dict and list types.
     if isinstance(value, mongoengine.base.datastructures.BaseDict):
         value = dict(value)
-
-    if isinstance(value, mongoengine.base.datastructures.BaseList):
+    elif isinstance(value, mongoengine.base.datastructures.BaseList):
         value = list(value)
 
+    # Recursively traverse the dict and list to convert values.
     if isinstance(value, dict):
         value = {k: mongodb_to_python_types(v) for k, v in six.iteritems(value)}
-
-    if isinstance(value, list):
+    elif isinstance(value, list):
         value = [mongodb_to_python_types(v) for v in value]
 
     return value
--- a/st2common/st2common/util/db.py
+++ b/st2common/st2common/util/db.py
@@ -1,0 +1,34 @@
+# Copyright 2019 Extreme Networks, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import absolute_import
+
+import mongoengine
+import six
+
+
+def mongodb_to_python_types(value):
+    if isinstance(value, mongoengine.base.datastructures.BaseDict):
+        value = dict(value)
+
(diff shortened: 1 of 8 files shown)


st2common/st2common/util/db.py
```diff
@@ -0,0 +1,34 @@
...
+    if isinstance(value, mongoengine.base.datastructures.BaseList):
+        value = list(value)
+
+    if isinstance(value, dict):
+        value = {k: mongodb_to_python_types(v) for k, v in six.iteritems(value)}
+
+    if isinstance(value, list):
+        value = [mongodb_to_python_types(v) for v in value]
+
+    return value
```
jinpingh: `value` passed is only one type of instance. Can it be checked with `if..else`.

m4dcoder: I'm not sure what you mean.  `value` can be any number of types.  The function recursively converts any BaseDict to dict and BaseList to list and if it's a dict/list then drill down to the values and convert those as well.  Anything other type, it will just leave the value as is.

jinpingh:
```
if isinstance(value, mongoengine.base.datastructures.BaseDict):
        value = dict(value)
elif isinstance(value, mongoengine.base.datastructures.BaseList):
        value = list(value)
elif isinstance(value, dict):
        value = {k: mongodb_to_python_types(v) for k, v in six.iteritems(value)}
else isinstance(value, list):
        value = [mongodb_to_python_types(v) for v in value]
```

m4dcoder: Oh I see what you mean. Yes, I will change.

m4dcoder: I still have to split it to two if-else block.  For example, after the convert from BaseDict to become dict, it needs to be traversed.  

```
if isinstance(value, mongoengine.base.datastructures.BaseDict):
        value = dict(value)
elif isinstance(value, mongoengine.base.datastructures.BaseList):
        value = list(value)

if isinstance(value, dict):
        value = {k: mongodb_to_python_types(v) for k, v in six.iteritems(value)}
else isinstance(value, list):
        value = [mongodb_to_python_types(v) for v in value]
```

PR #4732, opened 2019-06-30
Update dig.py error handling

punkrokk:
I have added an exception to the [linux.dig action script](https://github.com/StackStorm/st2/blob/master/contrib/linux/actions/dig.py) to handle the situation where the `dig` command is not installed, which, it seems on CentOS/RHEL gets the user this error: 

```
{
  "stdout": "",
  "result": "None",
  "stderr": "Traceback (most recent call last):
  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/python_runner/python_action_wrapper.py\", line 334, in <module>
    obj.run()
  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/python_runner/python_action_wrapper.py\", line 193, in run
  File \"/opt/stackstorm/packs/linux/actions/dig.py\", line 30, in run
    stdout=subprocess.PIPE)
  File \"/usr/lib64/python2.7/subprocess.py\", line 711, in __init__
    errread, errwrite)
  File \"/usr/lib64/python2.7/subprocess.py\", line 1327, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
",
  "exit_code": 1
}
```

This doesn't seem to be an issue on Ubuntu. It appears the `dnsutils` is installed. At least on the vagrant st2 dev environment. As you can see there is now a better error message if we get an `OSError` exception, suggesting that dig is either not on the PATH or installed. 

I have also update the linux pack readme to include the dig action. 

Since this is a core pack and installed by default, I opened the below PR to install bind-utils by default in EL7 distros. 
https://github.com/StackStorm/st2-packages/pull/619


contrib/linux/actions/dig.py
```diff
@@ -39,11 +39,17 @@ def run(self, rand, count, nameserver, hostname, queryopts):
...
-                                             .communicate()[0]
-                                             .split('\n'))
+
+        try:
+            result_list = filter(None, subprocess.Popen(cmd_args,
+                                                        stderr=subprocess.PIPE,
+                                                        stdout=subprocess.PIPE)
+                                 .communicate()[0]
+                                 .split('\n'))
+        except OSError as e:
```
blag: I'm not a huge fan of treating all `OSError`s as `ENOENT` errors.

Python 3 has an explicit [`FileNotFoundError`](https://docs.python.org/3/library/exceptions.html#FileNotFoundError), but for now we also need to support Python 2.

There is a way to check the exception for the type of error:

```python
import errno

# ...

except OSError as e:
    if e.errno == errno.ENOENT:
        # ...
    else:
        raise e
```

Can you tweak this to only emit the error message on an `ENOENT` `OSError`?

Also, please add a `NOTE:` comment so we can revisit this once we only support Python 3.

punkrokk: @blag We'll isn't that a little bit of python lore. Yes I will update this.

punkrokk: @blag pushed the updates you requested.

punkrokk:
@blag updated based on your requests.

contrib/linux/actions/dig.py
```diff
@@ -39,11 +41,25 @@ def run(self, rand, count, nameserver, hostname, queryopts):
...
+                                 .communicate()[0]
+                                 .split('\n'))
+
+        # NOTE: Python3 supports the FileNotFoundError, the errono.ENOENT is for py2 compat
+        # for Python3:
+        # except FileNotFoundError as e:
+
+        except OSError as e:
+            if e.errno == errno.ENOENT:
+                return False, "Can\'t find dig installed in our path (Usually /usr/bin/dig). " \
```
blag: For exceptions that are easily resolved, like this, I usually try to give some instructions for fixing it right in the error message. For this, I would probably add something like:

> Can't find dig installed in the path (usually /usr/bin/dig). If dig isn't installed, you can install it with 'sudo yum install bind-utils' or 'sudo apt install dnsutils'.

punkrokk: @blag  @armab all set.

blag:
Thanks for the contribution!

PR #4733, opened 2019-07-05
Fix the garbage collection for orquesta workflow executions

m4dcoder:
Fix logic in identifying orphaned workflow executions which causes long running workflow executions to be canceled. Disable this garbage collection feature for orphaned workflow executions by default.

### DIFF
--- a/conf/st2.conf.sample
+++ b/conf/st2.conf.sample
@@ -371,6 +371,6 @@ retry_stop_max_msec = 60000
 retry_max_jitter_msec = 1000
 # Interval inbetween retries.
 retry_wait_fixed_msec = 1000
-# Max seconds to allow workflow execution be idled before it is identified as orphaned and cancelled by the garbage collector.
-gc_max_idle_sec = 900
+# Max seconds to allow workflow execution be idled before it is identified as orphaned and cancelled by the garbage collector. A value of zero means the feature is disabled. This is disabled by default.
+gc_max_idle_sec = 0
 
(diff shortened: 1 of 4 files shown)


m4dcoder:
Please note that there are existing unit tests covering this feature @ https://github.com/StackStorm/st2/blob/master/st2common/tests/unit/services/test_workflow_identify_orphans.py.  Once the logic in the if statement was changed, these tests failed. Adding the additional filter `id__ne` when querying for ActionExecution fixes the issue.

conf/st2.conf.sample
```diff
@@ -371,6 +371,6 @@ retry_stop_max_msec = 60000
 retry_max_jitter_msec = 1000
 # Interval inbetween retries.
 retry_wait_fixed_msec = 1000
-# Max seconds to allow workflow execution be idled before it is identified as orphaned and cancelled by the garbage collector.
-gc_max_idle_sec = 900
+# Max seconds to allow workflow execution be idled before it is identified as orphaned and cancelled by the garbage collector. A value of zero means the feature is disabled. This is disabled by default.
```
armab: Looks like a better/safer default :+1:

Kami:
Can you please point me to an explicit test case which checks that nothing is deleted / marked as canceled when config option is set to ``0``?

Aka in addition to testing the default value, also having an explicit test case which tests with ``0``.

m4dcoder:
@Kami Unit tests added to address your concerns.  Note, the garbage collector service has zero unit tests for other gc routines.  I only added tests to cover the issue here.

armab:
Merging to have it in master/packages ASAP

PR #4734, opened 2019-07-05
Fix the garbage collection for orquesta workflow executions (v3.1)

m4dcoder:
Cherry pick from https://github.com/StackStorm/st2/pull/4733

Fix logic in identifying orphaned workflow executions which causes long running workflow executions to be canceled. Disable this garbage collection feature for orphaned workflow executions by default.


armab:
Merging, to have new packages deployed ASAP

PR #4736, opened 2019-07-08
Allow cancel to proceed if workflow execution is not found or completed

### LINKED ISSUE #4735: Orquesta workflow execution is stuck with canceling status
##### SUMMARY

If user cancels an orquesta workflow execution immediately after it was launched, there is a possibility where the workflow execution will be stuck in a canceling status. The cancelation failed because the workflow execution record has not been created yet.

##### ISSUE TYPE
 - Bug Report

##### STACKSTORM VERSION
st2 v2.8 -> v3.1

This affects all distros and installs.

##### STEPS TO REPRODUCE
Launch an orquesta workflow execution with one or more nested subworkflows. Cancel the workflow execution almostly immediately after it was launched.

##### EXPECTED RESULTS
Workflow execution should be in canceled state.

##### ACTUAL RESULTS
Workflow execution is stuck in canceling state.

m4dcoder:
Allow the workflow execution cancelation to proceed even if the workflow execution is not found or completed otherwise the method will raise an exception and the workflow execution will be stuck with a canceling status. Fixes #4735

### DIFF
--- a/contrib/runners/orquesta_runner/orquesta_runner/orquesta_runner.py
+++ b/contrib/runners/orquesta_runner/orquesta_runner/orquesta_runner.py
@@ -14,6 +14,8 @@
 
 from __future__ import absolute_import
 
+import sys
+import traceback
 import uuid
 
 import six
@@ -199,6 +201,8 @@ class OrquestaRunner(runners.AsyncActionRunner):
         return wf_ex_cancelable or ac_ex_cancelable
 
     def cancel(self):
+        result = None
+
         # Try to cancel the target workflow execution.
         try:
             wf_svc.request_cancellation(self.execution)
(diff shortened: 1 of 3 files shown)


contrib/runners/orquesta_runner/orquesta_runner/orquesta_runner.py
```diff
@@ -198,8 +199,20 @@ def task_cancelable(ac_ex):
...
+        try:
+            wf_svc.request_cancellation(self.execution)
+        # If workflow execution is not found because the action execution is cancelled
+        # before the workflow execution is created or if the workflow execution is
+        # already completed, then ignore the exception and proceed with cancellation.
+        except (wf_svc_exc.WorkflowExecutionNotFoundException,
+                wf_svc_exc.WorkflowExecutionIsCompletedException):
+            pass
+        # If there is an unknown exception, then log and rethrow the exception.
+        except Exception as e:
```
jinpingh: Not sure if we can add test case for  `Unable to cancel workflow execution`?

m4dcoder: @jinpingh Good catch. I changed how the unknown exception is handled and added unit test.

PR #4743, opened 2019-07-16
Install pack with the latest tag

### LINKED ISSUE #4384: Efficient pack development at scale
##### SUMMARY

Currently stackstorm-exchange packs have a development workflow that requires the version to get bumped with every pull request.  As part of a discussion in a pull request for the vSphere pack, some limitations of the current methodologies came up which will hinder scaling development.  As the number of developers in the community grows, packs will need a better way to accumulate many changes between releases.

##### ISSUE TYPE
 - Best Practices

StackStorm-Exchange

##### DISCUSSION
In the vSphere pack (and likely other packs) today, every pull request requires a version bump.  If two changes are committed to master and the version is not bumped, one cannot determine which version that they might have, since the master branch always has the version of the latest release in it, and there is no staging branch.

This will not scale, and a solution is needed to enable for expected growth in the community.  The boost community for example solves this by maintaining a master branch which is always deemed to be the latest release, and a "develop" branch to which all pull requests are targeted.  This allows each boost library to accumulate multiple changes in a holding area ("develop") until the maintainer deems them sound, good, etc.  The maintainer then merges "develop" into "master" (this should be a simple fast-forward), and then the maintainer updates the version and performs release activities (a final CI build, tagging the release, and bumping any version info to whatever will be next.)

To translate this mechanism to a pack repository, one would:

1. Create a "develop" branch off "master".
2. Change github to direct all PRs to "develop".
3. Now multiple developers can submit as many changes as they need.
4. Maintainer decides it is time for a release, and fast-forwards master to develop, then commits a change to pack.yaml with the version bump, then tags the release and performs any other release related activities.

This has many benefits:

1. Allows a pack to accumulate multiple changes between releases.
2. Avoids collisions on pack.yaml if every PR would normally need to update it with a version bump.
3. Puts the decision of what is a patch/minor/major bump on the maintainer, not the submitter.

jinpingh:
First part of Fixes StackStorm/discussions#354.
If pack version is not specified, install the latest version from pack index.json (https://index.stackstorm.org/v1/index.json), otherwise install from master branch as today.
Note: Direct github/repo install is not considered an Exchange index hit, therefore it is not covert by this fix.

Closes #4384

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,12 @@ Changelog
 in development
 --------------
 
+Changed
+~~~~~~~
+
+* Install pack with the latest tag version if it exists when branch is not specialized.
+  (improvement) #4743
+
 Fixed
 ~~~~~
 
--- a/contrib/packs/tests/test_action_download.py
+++ b/contrib/packs/tests/test_action_download.py
@@ -81,6 +81,17 @@ def mock_is_dir_func(path):
     return original_is_dir_func(path)
 
 
+def mock_get_gitref(repo, ref):
+    """
+    Mock get_gitref function which return mocked object if ref passed is
+    PACK_INDEX['test']['version']
+    """
+    if ref == 'v%s' % PACK_INDEX['test']['version']:
+        return mock.MagicMock(hexsha=PACK_INDEX['test']['version'])
+    else:
+        return None
+
+
 @mock.patch.object(pack_service, 'fetch_pack_index', mock.MagicMock(return_value=(PACK_INDEX, {})))
 class DownloadGitRepoActionTestCase(BaseActionTestCase):
     action_cls = DownloadGitRepoAction
@@ -572,3 +583,28 @@ class DownloadGitRepoActionTestCase(BaseActionTestCase):
         destination_path = os.path.join(self.repo_base, 'test4')
         self.assertTrue(os.path.exists(destination_path))
         self.assertTrue(os.path.exists(os.path.join(destination_path, 'pack.yaml')))
+
+    @mock.patch('st2common.util.pack_management.get_gitref', mock_get_gitref)
+    def test_run_pack_download_with_tag(self):
+        action = self.get_action_instance()
+        result = action.run(packs=['test'], abs_repo_base=self.repo_base)
+        temp_dir = hashlib.md5(PACK_INDEX['test']['repo_url'].encode()).hexdigest()
+
+        self.assertEqual(result, {'test': 'Success.'})
+        self.clone_from.assert_called_once_with(PACK_INDEX['test']['repo_url'],
+                                                os.path.join(os.path.expanduser('~'), temp_dir))
+        self.assertTrue(os.path.isfile(os.path.join(self.repo_base, 'test/pack.yaml')))
+
+        # Check repo.git.checkout is called three times
+        self.assertEqual(self.repo_instance.git.checkout.call_count, 3)
+
+        # Check repo.git.checkout called with latest tag or branch
+        self.assertEqual(PACK_INDEX['test']['version'],
+                         self.repo_instance.git.checkout.call_args_list[1][0][0])
+
+        # Check repo.git.checkout called with head
+        self.assertEqual(self.repo_instance.head.reference,
+                         self.repo_instance.git.checkout.call_args_list[2][0][0])
+
+        self.repo_instance.git.branch.assert_called_with(
+            '-f', self.repo_instance.head.reference, PACK_INDEX['test']['version'])
--- a/contrib/packs/tests/test_action_download.py
+++ b/contrib/packs/tests/test_action_download.py
@@ -505,6 +505,9 @@ class DownloadGitRepoActionTestCase(BaseActionTestCase):
         ]
 
         for default_branch, ref in edge_cases:
+            if not ref:
+                ref = PACK_INDEX['test']['version']
+
             self.repo_instance.git = mock.MagicMock(
                 branch=(lambda *args: default_branch),
                 checkout=(lambda *args: True)
@@ -529,10 +532,7 @@ class DownloadGitRepoActionTestCase(BaseActionTestCase):
 
             action = self.get_action_instance()
 
-            if ref:
-                packs = ['test=%s' % (ref)]
-            else:
-                packs = ['test']
+            packs = ['test=%s' % (ref)]
 
             result = action.run(packs=packs, abs_repo_base=self.repo_base)
             self.assertEqual(result, {'test': 'Success.'})
--- a/contrib/packs/tests/test_action_download.py
+++ b/contrib/packs/tests/test_action_download.py
@@ -505,9 +505,6 @@ class DownloadGitRepoActionTestCase(BaseActionTestCase):
         ]
 
         for default_branch, ref in edge_cases:
-            if not ref:
-                ref = PACK_INDEX['test']['version']
-
             self.repo_instance.git = mock.MagicMock(
                 branch=(lambda *args: default_branch),
                 checkout=(lambda *args: True)
@@ -523,7 +520,7 @@ class DownloadGitRepoActionTestCase(BaseActionTestCase):
 
             # Fool _get_gitref into working when its ref == our ref
             def fake_commit(arg_ref):
-                if arg_ref == ref:
+                if not ref or arg_ref == ref:
                     return gitref
                 else:
                     raise BadName()
@@ -532,7 +529,10 @@ class DownloadGitRepoActionTestCase(BaseActionTestCase):
 
             action = self.get_action_instance()
 
-            packs = ['test=%s' % (ref)]
+            if ref:
+                packs = ['test=%s' % (ref)]
+            else:
+                packs = ['test']
 
             result = action.run(packs=packs, abs_repo_base=self.repo_base)
             self.assertEqual(result, {'test': 'Success.'})
--- a/contrib/packs/tests/test_action_download.py
+++ b/contrib/packs/tests/test_action_download.py
@@ -86,8 +86,13 @@ def mock_get_gitref(repo, ref):
     Mock get_gitref function which return mocked object if ref passed is
     PACK_INDEX['test']['version']
     """
-    if ref == 'v%s' % PACK_INDEX['test']['version']:
-        return mock.MagicMock(hexsha=PACK_INDEX['test']['version'])
+    if PACK_INDEX['test']['version'] in ref:
+        if ref[0] == 'v':
+            return mock.MagicMock(hexsha=PACK_INDEX['test']['version'])
+        else:
+            return None
+    elif ref:
+        return mock.MagicMock(hexsha="abcDef")
     else:
         return None
 
--- a/st2common/st2common/util/pack_management.py
+++ b/st2common/st2common/util/pack_management.py
@@ -370,7 +370,7 @@ def get_repo_url(pack, proxy_config=None):
         if not pack:
             raise Exception('No record of the "%s" pack in the index.' % (name_or_url))
 
-        return (pack['repo_url'], version)
+        return (pack['repo_url'], version or pack['version'])
     else:
         return (eval_repo_url(name_or_url), version)
 


m4dcoder:
Have you actually test this change on a pack at GitHub with a tag?

CHANGELOG.rst
```diff
@@ -7,6 +7,9 @@ in development
 Fixed
```
armab: If this PR enhancement is not a bugfix, but improvement or change, it should come under different category.

See `CHANGELOG.rst` history for other examples.

armab:
Closes https://github.com/StackStorm/st2/issues/4384 and specific corner case https://github.com/StackStorm/st2/issues/4384#issuecomment-428629248

jinpingh:
> Have you actually test this change on a pack at GitHub with a tag?

Yes.

m4dcoder:
@jinpingh Can you post the output here for test against a real pack?  Thanks.

st2common/st2common/util/pack_management.py
```diff
@@ -249,6 +249,12 @@ def clone_repo(temp_dir, repo_url, verify_ssl=True, ref='master'):
         repo.git.checkout('-b', short_branch, branch)
         branch = repo.head.reference
     else:
+        # Checkout out the latest tag if tag exists when branch or version is not specified.
+        if not ref and repo.tags:
+            branch = repo.git.describe('--abbrev=0', '--tags')
```
armab: This is not a valid source for truth for getting a "latest" tag.

It should rely first on https://index.stackstorm.org/v1/index.json, - the latest version in package index that associated with the pack.

armab: Here is a pointer for more native fix: 
https://github.com/StackStorm/st2/blob/c70b2dbfd311f34ea98a25f9ac0b5a4e15141f12/st2common/st2common/util/pack_management.py#L108
Getting the correct version there would be better.

CHANGELOG.rst
```diff
@@ -4,6 +4,12 @@ Changelog
 in development
 --------------
 
+Added
+~~~~~
+
+* Install pack with the latest tag version if it exists when branch is not specialied.
```
armab: Typo in _specialied_

jinpingh:
> Have you actually test this change on a pack at GitHub with a tag? If so, can you post the output here? Thanks.

```
(virtualenv) vagrant@ubuntu-xenial:~/st2$ st2 pack install https://github.com/StackStorm-Exchange/stackstorm-jira

        [ succeeded ] download pack
        [ succeeded ] make a prerun
        [ succeeded ] install pack dependencies
        [ succeeded ] register pack

+-------------+-------------------+
| Property    | Value             |
+-------------+-------------------+
| name        | jira              |
| description | Jira integrations |
| version     | 0.12.1            |
| author      | StackStorm, Inc.  |
+-------------+-------------------+
(virtualenv) vagrant@ubuntu-xenial:~/st2$ cd /opt/stackstorm/packs/jira
(virtualenv) vagrant@ubuntu-xenial:/opt/stackstorm/packs/jira$ git status
HEAD detached at v0.12.1
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

        modified:   .circleci/config.yml
        modified:   .gitignore
        modified:   actions/add_field_value.py
        modified:   actions/add_field_value.yaml
        modified:   actions/assign_issue.yaml
        modified:   actions/attach_file_to_issue.py
        modified:   actions/attach_file_to_issue.yaml
        modified:   actions/attach_files_to_issue.py
        modified:   actions/attach_files_to_issue.yaml
        modified:   actions/get_issue_attachments.py
        modified:   actions/get_issue_attachments.yaml
        modified:   actions/get_issue_comments.py
        modified:   actions/get_issue_comments.yaml
        modified:   actions/run.py
        modified:   actions/transition_issue_by_name.yaml
        modified:   sensors/jira_sensor_for_apiv2.yaml
        modified:   tests/__init__.py
        modified:   tests/fixtures/blank.yaml
        modified:   tests/fixtures/full_auth_passwd.yaml
        modified:   tests/lib/__init__.py
        modified:   tests/lib/actions.py
        modified:   tests/test_action_run.py
        modified:   tests/test_sensor_jira_sensor.py
        modified:   tests/test_sensor_jira_sensor_for_apiv2.py

no changes added to commit (use "git add" and/or "git commit -a")
(virtualenv) vagrant@ubuntu-xenial:/opt/stackstorm/packs/jira$
```

st2common/st2common/util/pack_management.py
```diff
@@ -249,6 +250,13 @@ def clone_repo(temp_dir, repo_url, verify_ssl=True, ref='master'):
         repo.git.checkout('-b', short_branch, branch)
         branch = repo.head.reference
     else:
+        # Checkout out the latest tag when branch or version is not specified.
+        if not ref:
+            branch = get_pack_version_from_url(repo_url)
```
armab: There is no need in new function.

Please drill deeper in code based on https://github.com/StackStorm/st2/pull/4743#discussion_r304049691

Take a look at `get_repo_url` function which is called earlier that requests pack index. If we obtain/set the version there no other changes will be needed.

1) https://github.com/StackStorm/st2/blob/c70b2dbfd311f34ea98a25f9ac0b5a4e15141f12/st2common/st2common/util/pack_management.py#L108
2) https://github.com/StackStorm/st2/blob/c70b2dbfd311f34ea98a25f9ac0b5a4e15141f12/st2common/st2common/util/pack_management.py#L360
3) https://github.com/StackStorm/st2/blob/c70b2dbfd311f34ea98a25f9ac0b5a4e15141f12/st2common/st2common/util/pack_management.py#L374-L379
4) https://github.com/StackStorm/st2/blob/c70b2dbfd311f34ea98a25f9ac0b5a4e15141f12/st2common/st2common/util/pack_management.py#L153-L155

jinpingh: @armab From my understanding, when pack is installed with pack name  `st2 pack install jira`, `get_pack_from_index` is called. If pack is installed with repo URL `st2 pack install https://github.com/StackStorm-Exchange/stackstorm-jira`  `get_pack_from_index` is not called and clone_repo passes `ref=None`, at this case, pack name is unknown.  Therefore, need new function to get pack version.

jinpingh: By change query pack version from index file, it fixed test failure for st2cd.st2_e2e_tests_test_quickstart, but cause tests.test_run_pack_tests_tool failed.  Need to update index file for pack `csv` version from `0.1.0` to `0.4.4`.

armab: > From my understanding, when pack is installed with pack name st2 pack install jira, get_pack_from_index is called. If pack is installed with repo URL st2 pack install https://github.com/StackStorm-Exchange/stackstorm-jira get_pack_from_index is not called and clone_repo passes ref=None, at this case, pack name is unknown. Therefore, need new function to get pack version.

This is correct behavior.

We want to install latest stable pack version (instead of `master`) only for packs in Exchange which we control and know its consistency.
If pack is installed via URL, git, file://, - keep the previous/default behavior (install from `master` or default repo branch).

In other words, when user installs from his git repo or local folder it's expected to use current/default branch and not try to checkout any other release from git tag.

CHANGELOG.rst
```diff
@@ -4,6 +4,12 @@ Changelog
 in development
 --------------
 
+Added
```
armab: I think the PR belongs to `Changed` category as we change the default logic here and it's not a brand new feature or functionality.

CHANGELOG.rst
```diff
@@ -4,6 +4,12 @@ Changelog
...
+~~~~~
+
+* Install pack with the latest tag version if it exists when branch is not specialized.
+  (new feature) #354
```
armab: What's the reference URL behind the #354?

https://github.com/StackStorm/st2/pull/354 doesn't look relevant.

jinpingh: > What's the reference URL behind the #354?
> 
> `https://github.com/StackStorm/st2/pull/354` doesn't look relevant.

https://github.com/StackStorm/discussions/issues/354

armab: Shortcuts like `#354` are transformed into #354 links automatically in Github referring to same repo issues or PRs.
Private discussion link that OpenSource community can't access doesn't look appropriate for the changelog, please include current PR reference instead.

See https://help.github.com/en/articles/autolinked-references-and-urls

st2common/st2common/util/pack_management.py
armab: @jinpingh According to our discussions here and in Slack, the external exchange request you're doing shouldn't happen in git clone function. It's not appropriate place for that logic.

Following my comment that stable version should be utilized on pack installations relying on StackStorm Exchange index (eg. `st2 pack install csv`) and more code pointers above, having this code: 
https://github.com/StackStorm/st2/blob/c0458f1a06c2a81cac81697817c422ffb7631047/st2common/st2common/util/pack_management.py#L360-L375

Your first and easiest change would be 
```diff
-        return (pack['repo_url'], version)
+        return (pack['repo_url'], version or pack['version'])
```

Only that will make `st2 pack install csv` working as desired and pass correct version from Exchange Index (latest stable tag).

Let's please start with that.

jinpingh: @armab From my understanding, the different with my changes and your suggestion is:
For example `st2 pack install https://github.com/StackStorm-Exchange/stackstorm-jira`  or `st2 pack install jira`
My changes installed latest version of jira, that is `0.12.0`. 
```
(virtualenv) vagrant@ubuntu-xenial:/opt/stackstorm/packs/jira$ git status:
HEAD detached at v0.12.1
```
With your suggestion, it installed latest code from `master`
```
(virtualenv) vagrant@ubuntu-xenial:/opt/stackstorm/packs/jira$ git status
On branch master
```

From https://github.com/StackStorm/discussions/issues/351, requirement is `Install latest version tag, not from master.`  This is why i want to call a meeting to confirm that we are in the same page. Please let me know I misunderstand the requirement.    Thanks!

contrib/packs/tests/test_action_download.py
```diff
@@ -505,6 +505,9 @@ def side_effect(ref):
         ]
 
         for default_branch, ref in edge_cases:
+            if not ref:
+                ref = PACK_INDEX['test']['version']
```
armab: Any insight around this test modification?

Why it changes the initial `edge_cases` scenarios list on fly that expected to be immutable?

jinpingh: From my understanding, there are two StackStorm exchanges packs are used for `packs` test.
https://github.com/StackStorm-Exchange/stackstorm-test
https://github.com/StackStorm-Exchange/stackstorm-test2
They have the same pack name as `test` and is not included in `index.json` file.

This is why we have to simulate the way to get version from index.json, here is `PACK_INDEX`

armab: I don't think this helps. If we even place `9.9.9` as a pack version, - it will succeed as that test block tries to verify a very different thing.

---

Alright, what happens in case of reverting the code change in st2 core, will this new test fail?
If added test always succeeds irrespective of code change we focus on, how this case is intended to help?

Dig in deeper. We need a dedicated test case that will indeed cover new pack install behavior.

jinpingh: This is an existing test case that new code break.

armab: Understood.

With this change to tests we just removed `st2 pack install test` run with no params (because it failed) replacing it with `st2 pack install test=0.4.0` run, according to 
https://github.com/StackStorm/st2/blob/429e986ddfd4bc507eb98cf6b3660ff36fd696b1/contrib/packs/tests/test_action_download.py#L535-L538

Just adding `ref=0.4.0` is not helpful there.  It's not different to the following edge case with `ref=1.2.3`;
https://github.com/StackStorm/st2/blob/429e986ddfd4bc507eb98cf6b3660ff36fd696b1/contrib/packs/tests/test_action_download.py#L496-L497

Explore other ways to fix existing tests and please include the test case that will cover new behavior in full.

jinpingh: Initially,  searched index.json file and did not find `test` pack. This is why i was thinking the root cause for failure because of that. 

After do more investigation, found index is patched at https://github.com/StackStorm/st2/blob/c0458f1a06c2a81cac81697817c422ffb7631047/contrib/packs/tests/test_action_download.py#L84

Actually code comparison failed at https://github.com/StackStorm/st2/blob/c0458f1a06c2a81cac81697817c422ffb7631047/contrib/packs/tests/test_action_download.py#L523 for test case ('master', None) and ('default-branch', None). Therefore next line `return gitref` is not called. Finally it failed at `0.4.0" is not a valid version, hash, tag or branch in https://github.com/StackStorm-Exchange/stackstorm-test.`

Before fix, both `arg_ref` and `ref` are None. With fix, `arg_ref`=`0.4.0` and still is `ref=None`

armab: Good stuff! :+1:

jinpingh: Good catch! Thanks!

contrib/packs/tests/test_action_download.py
```diff
@@ -81,6 +81,17 @@ def mock_is_dir_func(path):
...
 
+def mock_get_gitref(repo, ref):
+    """
+    Mock get_gitref function which return mocked object if ref passed is
+    PACK_INDEX['test']['version']
+    """
+    if ref == 'v%s' % PACK_INDEX['test']['version']:
+        return mock.MagicMock(hexsha=PACK_INDEX['test']['version'])
+    else:
+        return None
```
armab: What if code flow reaches this path?
With `None` as a returned value it'll error in the `pack_management.py` code.

Try it out and provide something better for this case.

Except of that, everything looks pretty good :+1:

jinpingh: get_gitref is called few times during download. It do return None sometime if reference is not found. For this test case, it is called twice. First time ref passed is `0.4.0` and it return None, and second time `v0.4.0` is passed and gitref is returned.

armab: Imagine regression introduced in the code and https://github.com/StackStorm/st2/blob/219c9a263feaa87e9696a5a9a2ccde3b3e6423ad/st2common/st2common/util/pack_management.py#L373
got reverted to its previous state.

Taken reverted code, try to run the tests we now have to understand what I'm meaning.

PR #4750, opened 2019-07-28
Fix dist_utils.py so it doesn't rely on internal pip API

Kami:
This pull request fixes ``dist_utils.py`` module so it doesn't rely on internal pip functionality and so it works with latest pip version.

## Background, Context

While working on another project which has code similar to our ``dist_utils.fetch_requirements`` I noticed it fails with newer versions of pip with the following error:

```bash
    Failed to import parse_requirements from pip: cannot import name pep425tags
    Using pip: 19.2.1
```

In our case, the error itself could manifest itself when installing ``st2client`` using a specific version of pip.

Main problem is that the code in dist_utils.py depends on internal pip functionality which can change across the pip versions (and it already changed in the past and that caused us quite some trouble back then).

## Proposed Solution

This solution re-implements ``fetch_requirements`` so it doesn't depend on the pip anymore.

Internal pip ``parse_requirements`` does a lot of more stuff, but we actually only care about a simple use case - we only want a list of simple and URL requirements back and that's trivial to implement ourselves.

### DIFF
--- a/contrib/runners/action_chain_runner/dist_utils.py
+++ b/contrib/runners/action_chain_runner/dist_utils.py
@@ -27,31 +27,18 @@ PY3 = sys.version_info[0] == 3
 if PY3:
     text_type = str
 else:
-    text_type = unicode
+    text_type = unicode  # NOQA
 
 GET_PIP = 'curl https://bootstrap.pypa.io/get-pip.py | python'
 
 try:
     import pip
-    from pip import __version__ as pip_version
 except ImportError as e:
     print('Failed to import pip: %s' % (text_type(e)))
     print('')
     print('Download pip:\n%s' % (GET_PIP))
     sys.exit(1)
 
-try:
-    # pip < 10.0
-    from pip.req import parse_requirements
-except ImportError:
-    # pip >= 10.0
-
-    try:
-        from pip._internal.req.req_file import parse_requirements
-    except ImportError as e:
-        print('Failed to import parse_requirements from pip: %s' % (text_type(e)))
-        print('Using pip: %s' % (str(pip_version)))
-        sys.exit(1)
 
 __all__ = [
     'check_pip_version',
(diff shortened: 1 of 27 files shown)


contrib/runners/action_chain_runner/dist_utils.py
```diff
@@ -80,12 +68,43 @@ def fetch_requirements(requirements_file_path):
...
     links = []
     reqs = []
-    for req in parse_requirements(requirements_file_path, session=False):
-        # Note: req.url was used before 9.0.0 and req.link is used in all the recent versions
-        link = getattr(req, 'link', getattr(req, 'url', None))
-        if link:
-            links.append(str(link))
-        reqs.append(str(req.req))
+
+    def _get_link(line):
```
armab: I'm wondering how error-prone is that code block?
Did you take the rules from upstream or it's something else you consulted for code logic?

Kami: I don't check pip code itself, but I compared the results (which is the only thing I care about at this point - the function output).

If you check the tests, you will see we have tests which compare the result of our version and pip version to ensure they are the same.

It's likely there are still some edge cases somewhere, but it's not a big deal at this point since for our current ``requirements.txt``, the end result is the same.

punkrokk:
Thank you! This has been a `thing` in parts of my world at times.

punkrokk:
@blag is the v3.1.1 tag correct?

contrib/runners/action_chain_runner/dist_utils.py
blag: This code did not support environment markers from pip, so I had to add (hacky) support for them in #4895.

@Kami Why can the `dist_utils.py` module not use any third party dependencies?

```python
# NOTE: This script can't rely on any 3rd party dependency so we need to use this code here
```

Instead of importing from pip directly (which we should never have been doing in the first place, IMO), *if* we can use third party dependencies, we can utilize the [`packaging` package](https://packaging.pypa.io/en/latest/requirements/). It has [an entire section](https://packaging.pypa.io/en/latest/requirements/) devoted to reading and parsing `requirements.txt` files, and that would be kept up-to-date with what pip and setuptools support, and would lower our maintenance burden of this code.

blag:
@punkrokk I originally created the `v3.1.1` to track PRs that were good candidates for a `v3.1.1` bugfix release, but @m4dcoder and @armab disagreed on their usefulness, so I deleted the tag itself. That removed it from the list of tags on this PR, but apparently just deleting the tag doesn't create a "GitHub PR changelog" entry for removing the tag from this PR.

The reason I created a tag for bugfix versions is that you can only assign PRs to one milestone. Which kinda sorta makes some sense, but I do think we should have some way of tracking PRs that would be good candidates for the next bugfix release/s. At this point, however, that's NMFP.

nmaludy:
This just bit us.

Need to copy this to other places where we copy/paste `dist_utils.py` such as: https://github.com/StackStorm/st2-auth-backend-flat-file

PR #4751, opened 2019-07-29
Include requirements.txt files for all the components Python packages

Kami:
This pull request updates all the component Python packages so we include ``requirements.txt`` files.

This is needed to make all the components valid Python packages so they can be re-used outside of the StackStorm/st2 context.

Before that change, only ``st2common`` and ``st2client`` + all the runners were valid Python packages.

In addition to that, we also need to include ``__version__`` attribute as part of all the Python package ``__init__.py`` file. This will be handled by StackStorm/st2cd workflow.

NOTE: This is similar to the change I made in the past to make action runners valid Python packages.

## TODO

- [x] Corresponding change in StackStorm/st2cd - https://github.com/StackStorm/st2cd/pull/414
- [x] Tests


Kami:
As a "bonus", I also fixed intermediate test failure / race which has been happening for a long time - https://gist.github.com/Kami/83ee1d430a2366b3656fa2d28286f810.

This race happens because ``kill_func()`` is async and it means ``run_command()`` could return before that function returns and before ``process.returncode`` is set. As such, we need to set the return code twice to account for both scenarios (async ``kill_func()`` and non ``async process.kill()``).

I confirmed the fix locally by running the offending test 100 times - it didn't fail even a single time. Without that change, it fails quite often.

Kami:
@armab I initially missed this in #4750 - c919ff9b66cbb6b6fdd9699c3a9405fd206fa024.

Now ``pip`` is only needed for ``check_pip_version`` function so we only perform that check when ``check_pip_version`` is called.

While at it, I also added test cases for other functions in that file - e2ae1a6d1bc7a242d2d7ddd0ba9a84373d28e51b.

I included this change as part of this PR since it's also related to components Python packages and distribution.

Makefile
```diff
@@ -159,6 +162,38 @@ check-requirements: requirements
...
+	sed -i '/PYTHONPATH=/d' $(VIRTUALENV_COMPONENTS_DIR)/bin/activate
+	sed -i '/export PYTHONPATH/d' $(VIRTUALENV_COMPONENTS_DIR)/bin/activate
+
+	echo '_OLD_PYTHONPATH=$$PYTHONPATH' >> $(VIRTUALENV_COMPONENTS_DIR)/bin/activate
+	echo 'PYTHONPATH=${ROOT_DIR}:$(COMPONENT_PYTHONPATH)' >> $(VIRTUALENV_COMPONENTS_DIR)/bin/activate
+	echo 'export PYTHONPATH' >> $(VIRTUALENV_ST2CLIENT_DIR)/bin/activate
+	touch $(VIRTUALENV_COMPONENTS_DIR)/bin/activate
+	chmod +x $(VIRTUALENV_COMPONENTS_DIR)/bin/activate
+	
+	@for component in $(COMPONENTS_WITHOUT_ST2TESTS); do \
```
armab: The only complain that iterating over all services includes st2client, which duplicates `.st2client-install-check` that already ran during the `make ci-checks`

Kami:
@armab Correct, but st2client checks are a bit more extensive since we also install dependencies there (aka use ``python setup.py install``).

So in a sense, tests which run as part of this next task are a subset of the existing st2client tests.

Kami:
Per discussion with @arma, I moved some of the more extensive checks to a nightly (cron) build - e7df5c0543721f06f3ae297ae686181bff5fc1ce.

Any additional nightly checks now run as part of the ``Lint Checks, Packs Tests (Python 2.7)`` job during the nightly build.

Sadly Travis CI doesn't allow infrastructure as code (aka via travis config) approach to configure nightly builds so I needed to configure it in their WebUI.

Screenshot from 2019-07-30 17-41-10

In the future, we can utilize the same pattern to offload any additional and slower checks to a nightly build (so we speed up master branch merge and PR builds).

This doesn't mean we can just offload everything to the nightly build though (as much as possible should run as part of the every PR build). We should still practice diligence and always look for ways to speed up PR and master builds :)

Kami:
Confirmed nightly build is working correctly by "emulating" one (fdf5e96747659ea89bee2752dd0fffb6a3a7b98a) - https://travis-ci.org/StackStorm/st2/builds/565579951, https://travis-ci.org/StackStorm/st2/jobs/565579954#L3504.

armab:
FYI this broke the package build for `3.2.0` release as automation missed updating `__init__.py` files with the stable version. 

Fixed via https://github.com/StackStorm/st2cd/pull/435

PR #4765, opened 2019-08-13
Implement SSO plugin for user authentication in StackStorm

m4dcoder:
Add SSO endpoints to the st2 API spec. Add API controllers for the SSO redirect and handle Idp callback. Implement scaffolding for the SSO backends. Add unit tests to capture not implemented errors, unknown exception, and also SSO verification errors in the API controllers.

### DIFF
--- a/conf/st2.conf.sample
+++ b/conf/st2.conf.sample
@@ -71,7 +71,7 @@ sso = False
 mode = standalone
 # Specify to enable debug mode.
 debug = False
-# Single Sign On backend to use when SSO is enabled. Available backends: noop, saml2.
+# Single Sign On backend to use when SSO is enabled. Available backends: noop.
 sso_backend = noop
 
 # Standalone mode options - options below only apply when auth service is running in the standalone
--- a/conf/st2.conf.sample
+++ b/conf/st2.conf.sample
@@ -53,6 +53,8 @@ port = 9101
 # Common option - options below apply in both scenarios - when auth service is running as a WSGI
 # service (e.g. under Apache or Nginx) and when it's running in the standalone mode.
 
+# JSON serialized arguments which are passed to the SSO backend.
+sso_backend_kwargs = None
 # Enable authentication middleware.
 enable = True
 # Path to the logging config.
@@ -63,10 +65,14 @@ api_url = None
 service_token_ttl = 86400
 # Access token ttl in seconds.
 token_ttl = 86400
+# Enable Single Sign On for GUI if true.
+sso = False
 # Authentication mode (proxy,standalone)
 mode = standalone
 # Specify to enable debug mode.
 debug = False
+# Single Sign On backend to use when SSO is enabled. Available backends: noop, saml2.
+sso_backend = noop
 
 # Standalone mode options - options below only apply when auth service is running in the standalone
 # mode.
--- a/conf/st2.conf.sample
+++ b/conf/st2.conf.sample
@@ -71,7 +71,7 @@ sso = False
 mode = standalone
 # Specify to enable debug mode.
 debug = False
-# Single Sign On backend to use when SSO is enabled. Available backends: noop.
+# Single Sign On backend to use when SSO is enabled. Available backends: noop, saml2.
 sso_backend = noop
 
 # Standalone mode options - options below only apply when auth service is running in the standalone
--- a/st2auth/st2auth/app.py
+++ b/st2auth/st2auth/app.py
@@ -26,6 +26,8 @@ from st2common.router import Router
 from st2common.constants.system import VERSION_STRING
 from st2common.service_setup import setup as common_setup
 from st2common.util import spec_loader
+from st2common.util.monkey_patch import monkey_patch
+from st2common.util.monkey_patch import use_select_poll_workaround
 from st2auth import config as st2auth_config
 from st2auth.validation import validate_auth_backend_is_correctly_configured
 
@@ -60,6 +62,10 @@ def setup_app(config=None):
                      service_registry=True,
                      capabilities=capabilities,
                      config_args=config.get('config_args', None))
+
+        # pysaml2 uses subprocess communicate which calls communicate_with_poll
+        if cfg.CONF.auth.sso and cfg.CONF.auth.sso_backend == 'saml2':
+            use_select_poll_workaround(nose_only=False)
 
     # Additional pre-run time checks
     validate_auth_backend_is_correctly_configured()
--- a/st2auth/st2auth/app.py
+++ b/st2auth/st2auth/app.py
@@ -26,7 +26,6 @@ from st2common.router import Router
 from st2common.constants.system import VERSION_STRING
 from st2common.service_setup import setup as common_setup
 from st2common.util import spec_loader
-from st2common.util.monkey_patch import monkey_patch
 from st2common.util.monkey_patch import use_select_poll_workaround
 from st2auth import config as st2auth_config
 from st2auth.validation import validate_auth_backend_is_correctly_configured
--- a/st2auth/st2auth/controllers/v1/sso.py
+++ b/st2auth/st2auth/controllers/v1/sso.py
@@ -65,9 +65,8 @@ class IdentityProviderCallbackController(object):
 class SingleSignOnController(object):
     callback = IdentityProviderCallbackController()
 
-    def get(self, **kwargs):
+    def get(self, referer):
         try:
-            referer = kwargs.get('referer')
             response = router.Response(status=http_client.TEMPORARY_REDIRECT)
             response.location = SSO_BACKEND.get_request_redirect_url(referer)
             return response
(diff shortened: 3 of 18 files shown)


st2auth/st2auth/app.py
```diff
@@ -63,6 +64,10 @@ def setup_app(config={}):
                      capabilities=capabilities,
                      config_args=config.get('config_args', None))
 
+        # pysaml2 uses subprocess communicate which calls communicate_with_poll
+        if cfg.CONF.auth.sso and cfg.CONF.auth.sso_backend == 'saml2':
+            use_select_poll_workaround(nose_only=False)
```
Kami: I would have thought it's a pure Python implementation. Just curious, do yoy happen to know when it needs to spawn subprocesses?

st2auth/st2auth/controllers/v1/sso.py
```diff
@@ -0,0 +1,129 @@
...
+        except auth_exc.SSOVerificationError as e:
+            return process_failure_response(http_client.UNAUTHORIZED, e)
+        except Exception as e:
+            raise e
+
+
+class SingleSignOnController(object):
+    callback = IdentityProviderCallbackController()
+
+    def get(self, **kwargs):
```
Kami: Can we explicitly specify supported arguments instead of using ``**kwargs``? Based on the method body, it seems like only ``referer`` is supported.

st2auth/st2auth/controllers/v1/sso.py
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    body = CALLBACK_SUCCESS_RESPONSE_BODY % referer
+    resp = router.Response(body=body)
+    resp.headers['Content-Type'] = 'text/html'
+
+    resp.set_cookie(
+        'st2-auth-token',
+        value=quote(json.dumps(token_json)),
+        expires=datetime.timedelta(seconds=60),
```
Kami: Just to confirm I understand the flow correctly - st2auth sets a token with a low expire time since this token is directly read by the browser and written into local storage, right?

st2auth/st2auth/sso/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+    return driver_loader.get_available_backends(namespace=BACKENDS_NAMESPACE)
+
+
+def get_backend_instance(name):
+    sso_backend_cls = driver_loader.get_backend_driver(namespace=BACKENDS_NAMESPACE, name=name)
+
+    kwargs = {}
+    sso_backend_kwargs = cfg.CONF.auth.sso_backend_kwargs
+
+    if sso_backend_kwargs:
```
Kami: We use this pattern in multiple places (auth backend, sso backend, rbac backend) so it might make sense to refactor it in a new utility method.

st2auth/st2auth/app.py
m4dcoder: The pysaml2 module is running xmlsec on a subprocess at https://github.com/IdentityPython/pysaml2/blob/master/src/saml2/sigver.py#L891. This step happens during callback from the Idp and the pysaml2 is verifying/validating the signature of the response.

st2auth/st2auth/controllers/v1/sso.py
m4dcoder: Sure

st2auth/st2auth/controllers/v1/sso.py
m4dcoder: When SSO is successful (response is verified), st2auth will generate a st2 auth token with normal expiration time. In order to pass the st2 auth token from the server side to the client browser, the st2 auth token is written as a cookie with quick expiration time (60 seconds which is set here). The html page sent back to the browser in `CALLBACK_SUCCESS_RESPONSE_BODY` will read the st2 auth token from the cookie and then write it to the local storage.  After 60 seconds, the cookie should be expired and removed by the browser.

st2auth/st2auth/sso/__init__.py
m4dcoder: There is some similarity in `get_backend_instance` between the auth backend and sso backend.  But the rbac backend's `get_backend_instance` is different.  auth and sso version returns an instance and rbac returns a class.  I rather tackle this separately.

st2auth/st2auth/controllers/v1/sso.py
Kami: Yep, thanks, that's what I thought, just wanted to confirm I understand the flow and code correctly.

conf/st2.conf.sample
```diff
@@ -59,10 +61,14 @@ api_url = None
 service_token_ttl = 86400
 # Access token ttl in seconds.
 token_ttl = 86400
+# Enable Single Sign On for GUI if true.
+sso = False
 # Authentication mode (proxy,standalone)
 mode = standalone
 # Specify to enable debug mode.
 debug = False
+# Single Sign On backend to use when SSO is enabled. Available backends: noop.
```
armab: > Available backends: noop.

This auto-generated part looks non-friendly. In reality, what SSO backends will be available if sso package is installed?

m4dcoder: The only option is to either manually hard code the list or more this list altogether. The other `saml2` backend is not installed here and `make configgen` here will just keep throwing error.

st2auth/st2auth/controllers/v1/sso.py
```diff
@@ -0,0 +1,123 @@
...
+            var v = document.cookie.match('(^|;) ?' + name + '=([^;]*)(;|$)');
+            return v ? v[2] : null;
+        }
+
+        data = JSON.parse(window.localStorage.getItem('st2Session'));
+        data['token'] = JSON.parse(decodeURIComponent(getCookie('st2-auth-token')));
+        window.localStorage.setItem('st2Session', JSON.stringify(data));
+        window.location.replace("%s");
+    </script>
+    <body/>
```
armab: looks like broken closing HTML tag and it's not opening anywhere?

st2auth/st2auth/controllers/v1/sso.py
```diff
@@ -0,0 +1,123 @@
...
+        function getCookie(name) {
+            var v = document.cookie.match('(^|;) ?' + name + '=([^;]*)(;|$)');
+            return v ? v[2] : null;
+        }
+
+        data = JSON.parse(window.localStorage.getItem('st2Session'));
+        data['token'] = JSON.parse(decodeURIComponent(getCookie('st2-auth-token')));
+        window.localStorage.setItem('st2Session', JSON.stringify(data));
+        window.location.replace("%s");
+    </script>
```
armab: Would be nice to get @bigmstone to review this scripting part or someone with frontend experience.
Maybe there are any other alternatives available.

st2auth/st2auth/controllers/v1/sso.py
Kami: It's a self closing empty tag so that HTML should still be valid (similar to ```` for example).

m4dcoder: I have no issue testing this with `<body />`.  this is a valid html element, just means there's no text in between `<body></body>`.  Am i missing something here?

conf/st2.conf.sample
Kami: Either that or have a separate sample config for RBAC installations.

st2auth/st2auth/controllers/v1/sso.py
m4dcoder: Yeah, @bigmstone is added for review here.  I talked about this in the demo.  I talked about this in weekly meeting.

st2auth/st2auth/controllers/v1/sso.py
armab: I just checked
https://stackoverflow.com/questions/3558119/are-non-void-self-closing-tags-valid-in-html5
and 
http://xahlee.info/js/html5_non-closing_tag.html
Looks like very little point in it.

m4dcoder: @armab I removed `<body />` so we don't have to debate on this trivial subject.

conf/st2.conf.sample
m4dcoder: I manually hard coded this with `noop, saml2`.  It's simpler.

m4dcoder:
@armab Per your previous request, I added a GET endpoint here for the EWC UI to query whether SSO is enabled instead of putting the configuration option at the web app. This is the commit. https://github.com/StackStorm/st2/pull/4765/commits/abcee847f81041bf0044f0f6baf580e6dbb58ad0

conf/st2.conf.sample
armab: :+1:

armab:
@m4dcoder I saw last push to this PR was 2 days ago.
Considering difficulties with productizing this feature, I assume we don't plan this functionality for the upcoming `v3.2`? Please correct me as I'm removing it from the `3.2` milestone.

PR #4759, opened 2019-08-06
fixing enterprise ldap issue with execution view

jdmeyer3:
Closes #4758

The error caused in #4758 is because the ldap rbac is expecting the document returned from the ActionExecutionDB to contain action.pack and action.uid, but they are currently not being returned. 

Here is the traceback I got from the rbac
```
Traceback (most recent call last):
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2rbac_enterprise_backend/utils.py", line 120, in assert_user_has_resource_db_permission
    permission_type=permission_type)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2rbac_enterprise_backend/utils.py", line 287, in user_has_resource_db_permission
    permission_type=permission_type)
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2rbac_enterprise_backend/resolvers.py", line 673, in user_has_resource_db_permission
    pack_db = PackDB(ref=action['pack'])
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/mongoengine/base/datastructures.py", line 54, in __getitem__
    value = super(BaseDict, self).__getitem__(key)
KeyError: 'pack'
```
 
and here is the action object that is returned when it gets the key error
```
[...]

PR #4766, opened 2019-08-14
Fix rbac execution_view unable to verify the pack or uid (v3.1)

m4dcoder:
Fix rbac with execution view where the rbac is unable to verify the pack or uid of the execution because it was not returned from the action execution db. This would result in an internal server error when trying to view the results of a single execution. This fix is cherry picked from selected commits at PR https://github.com/StackStorm/st2/pull/4759.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,16 @@ Changelog
 in development
 --------------
 
+3.1.1 - August 14, 2019
+-----------------------
+
+Fixed
+~~~~~
+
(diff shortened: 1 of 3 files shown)


Kami:
Next time inside the PR description please also reference the original PR from which you are cherry picking the changes.

And also make it clear in the PR title that's targeted for the v3.1 branch.

m4dcoder:
@Kami Updated title and description. Thanks for the reminder.

PR #4767, opened 2019-08-16
Update various Python dependencies to latest stable versions

Kami:
This pull request updates various dependencies to latest stable versions.

I went over the changeslogs and changed things mostly include bug fixes and small improvements.

mongoengine v0.18.x includes change / performance improvement which should be of a particular interest to us - https://github.com/MongoEngine/mongoengine/pull/2049.

I tested / verified it locally and confirmed it increases "save" performance for very large objects (that's very large executions in our case). It's not really drastic, but around 5-30% performance increase depending on the execution size.

It's not as fast as directly inserting execution into database using pymongo and avoiding mongoengine layer all together (something we discussed in the past), but it comes mostly "for free".


Kami:
There seems to be regression in the latest version of pymongo and mongoengine (and potentially combination of both) related to unicode keys which causes test failure :/

I need to dig in and look into a fix.

Kami:
It turns out the issue is indeed related to pymongo and mongoengine - https://github.com/MongoEngine/mongoengine/pull/2147

I wasted quite a lot of time tracking it down (since it depends on Python version and default encoding used by Python).

In the mean time, I will revert ``pymongo`` upgrade and in the future, I will try to get that fix finished and merged into upstream mongoengine library.

CHANGELOG.rst
```diff
@@ -17,6 +17,8 @@ Changed
 * Install pack with the latest tag version if it exists when branch is not specialized.
   (improvement) #4743
 * Implement "continue" engine command to orquesta workflow. (improvement) #4740
+* Update various internal dependencies to latest stable versions (apscheduler, eventlet,
+  kombu, amqp, pyyaml, mongoengine, python-gnupg, paramiko, tooz, webob, bcrypt). #4767
```
armab: Makes sense to include a small notes that mongo led to perf improvements, considering you've tested it.

PR #4768, opened 2019-08-16
Fix copy/paste typo

blag:
Presumably, `core.local_sudo` only runs commands on `localhost`, and _not_ any remote host(s). The `cmd` parameter description should reflect that.


blag:
Hat tip to @NetDevLazg for noticing and pointing it out to me.

PR #4772, opened 2019-08-20
Move from Precise to Xenial image on Travis CI

Kami:
Travis CI builds recently (since Friday) started failing again and hanging / timing out on "travis_apt_get_update" step (e.g. https://travis-ci.org/StackStorm/st2/jobs/574431900).

We didn't change anything so it's likely upstream issue with Travis (likely similar to the one we've encountered in the past).

I'm just testing to see if I can work around the issue since it's a blocker for us and it's very annoying - even re-runs don't help most of the times because it keeps hanging / failing on that step.


Kami:
It looks like my changes / workaround is working.

I moved away from apt addon to manually running apt commands. This way we have more control and can specify increased timeout to work around what appear to be upstream Travis / networking issue.

This is obviously not ideal since apt-get related steps will take longer time now, but at least the build will continue and not get stuck on "travis_apt_get_update" script step.

In fact, this issue appears to be very similar to the issue we've encountered in the past (which was indeed related to upstream Travis / GCE networking setup).

Kami:
It appears that issue only affects precise builds, but not xenial ones.

So that issue may be the final nail in the coffin for precise. The only reason we still used it was because tests finish faster on it compared to xenial, but precise is ancient and it's really time to move on.

Kami:
It appears that this is now passing, but I still need to confirm that all the tests which need to run actually run (because of the chmod "workaround" we need to do on Xenial).

Sadly as discussed many times before, test timing is indeed quite slower on Xenial (verging on 30-50% slower which makes me sad) :/

PR #4771, opened 2019-08-20
Fix amqp version conflict between Kombu and ST2

GuiTeK:
C.F. https://github.com/StackStorm/st2/issues/4770

PR #4689, opened 2019-05-21
Update the requests library version to latest (2.22.0) in requirements

VineeshJain:
This PR updates the `requests` package version to the latest (2.22.0) - Closes #4680

Issue #4680, opened 2019-05-16
Dependency Upgrade: Requests


bigmstone:
Requests has fallen behind on dependency upgrades because it needs to be upgraded in multiple places.

PR #4774, opened 2019-08-21
Fix dependency conflicts

### LINKED ISSUE #4770: Currently breaking CIs: AMQP/Kombu dependency conflict
## SUMMARY

The ST2 requirements.txt currently on master (from commit https://github.com/StackStorm/st2/commit/a9f72ae54f41ead4dd7c20a0bf53be5bed0c6127) requires `amqp==2.5.0` and `kombu==4.6.4`. The problem is, `kombu==4.6.4` requires `amqp==2.5.1`, so there is a **dependency conflict**.

Pip fails with the following error:
```
ERROR: kombu 4.6.4 has requirement amqp<3.0,>=2.5.1, but you'll have amqp 2.5.0 which is incompatible.
```

When using a **recent version of `pip`** (e.g. 19.2.2), this will cause pip to **fail to install the requirements**. However, if you try with an old version (e.g. 8.1.1), it will work. I don't know at which specific version of Pip it stops working.

I think ST2 requirements.txt should require `amqp==2.5.1` instead of `amqp==2.5.0`. I made the PR: https://github.com/StackStorm/st2/pull/4771

**Note**: this might be considered urgent because as https://github.com/StackStorm-Exchange/ci uses the master branch of ST2 and not a specific version, **it is currently breaking CIs** (depending on what version of Pip is used by the CI).

### STACKSTORM VERSION

Current master branch, 26ffc2f4fe16b7a5b6fe46b98e32daf88e473dde

N/A

It's easier to reproduce with Docker.

1. Start a container with the following command: `docker run -it ubuntu:16.04 /bin/bash`
2. Run the following commands inside the Docker container:
```
apt-get update
apt-get install python3 python3-pip python3-dev git build-essential libffi-dev libssl-dev nano
pip3 install virtualenv # To have a newer version of pip easily
virtualenv st2-amqp-test
cd st2-amqp-test
source ./bin/activate
# Create the st2/requirements.txt file
pip install -r requirements.txt # This will fail
```

## Expected Results

Should correctly install all the dependencies when doing `pip install -r requirements.txt`.

## Actual Results

Pip fails with the following error:
```
ERROR: kombu 4.6.4 has requirement amqp<3.0,>=2.5.1, but you'll have amqp 2.5.0 which is incompatible.
```

blag:
This PR backs off on some of the updates to dependencies, and makes `make requirements` run successfully, on both Linux and Mac OS X (I tested both). This should also hopefully fix the weekly pack Circle CI runs.

I also tweak the requirements header to include directions on how to actually update dependencies, since it's definitely not a straightforward process.

On top of that, I also make sure the make targets actually fail, so (hopefully) future updates that break things won't pass CI.

Closes #4770 and #4771.
Reverts parts of #4689 and #4767.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -44,7 +44,7 @@ Fixed
 * Update ``dist_utils`` module which is bundled with ``st2client`` and other Python packages so it
   doesn't depend on internal pip API and so it works with latest pip version. (bug fix) #4750
 * Fix dependency conflicts in pack CI runs: downgrade requests dependency back to 0.21.0, update
-  internal dependencies (amqp, pyyaml, prance, six) (bugfix) #4774
+  internal dependencies and test expectations (amqp, pyyaml, prance, six) (bugfix) #4774
 
 3.1.0 - June 27, 2019
 ---------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,7 +14,6 @@ Added
 Changed
 ~~~~~~~
 
-* Improved instructions in requirements.txt and dist_utils.py comment headers (improvement) #4774
 * Install pack with the latest tag version if it exists when branch is not specialized.
   (improvement) #4743
 * Implement "continue" engine command to orquesta workflow. (improvement) #4740
@@ -23,6 +22,8 @@ Changed
 
   Latest version of mongoengine should show some performance improvements (5-20%) when
   writing very large executions (executions with large results) to the database. #4767
+* Improved development instructions in requirements.txt and dist_utils.py comment headers
+  (improvement) #4774
 
 Fixed
 ~~~~~
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,6 +14,7 @@ Added
 Changed
 ~~~~~~~
 
+* Improved instructions in requirements.txt and dist_utils.py comment headers (improvement) #4774
 * Install pack with the latest tag version if it exists when branch is not specialized.
   (improvement) #4743
 * Implement "continue" engine command to orquesta workflow. (improvement) #4740
@@ -42,6 +43,8 @@ Fixed
   Contributed by JP Bourget (@punkrokk Syncurity) #4732
 * Update ``dist_utils`` module which is bundled with ``st2client`` and other Python packages so it
   doesn't depend on internal pip API and so it works with latest pip version. (bug fix) #4750
+* Fix dependency conflicts in pack CI runs: downgrade requests dependency back to 0.21.0, update
+  internal dependencies (amqp, pyyaml, prance, six) (bugfix) #4774
 
 3.1.0 - June 27, 2019
 ---------------------
--- a/Makefile
+++ b/Makefile
@@ -478,7 +478,7 @@ requirements: virtualenv .sdist-requirements install-runners
 	# Note: We install prance here and not as part of any component
 	# requirements.txt because it has a conflict with our dependency (requires
 	# new version of requests) which we cant resolve at this moment
-	$(VIRTUALENV_DIR)/bin/pip install "prance==0.6.1"
+	$(VIRTUALENV_DIR)/bin/pip install "prance==0.15.0"
 
 	# Install st2common to register metrics drivers
 	# NOTE: We pass --no-deps to the script so we don't install all the
--- a/Makefile
+++ b/Makefile
@@ -455,7 +455,7 @@ requirements: virtualenv .sdist-requirements install-runners
 	done
 
 	# Fix for Travis CI race
-	$(VIRTUALENV_DIR)/bin/pip install "six==1.11.0"
+	$(VIRTUALENV_DIR)/bin/pip install "six==1.12.0"
 
 	# Fix for Travis CI caching issue
 	$(VIRTUALENV_DIR)/bin/pip uninstall -y "pytz" || echo "not installed"
--- a/Makefile
+++ b/Makefile
@@ -920,7 +920,7 @@ debs:
 	# Copy over shared dist utils module which is needed by setup.py
 	@for component in $(COMPONENTS_WITH_RUNNERS); do\
 		cp -f ./scripts/dist_utils.py $$component/dist_utils.py;\
-		sed -i -e '1s;^;# NOTE: This file is auto-generated - DO NOT EDIT MANUALLY\n;' $$component/dist_utils.py;\
+		scripts/write-headers.sh $$component/dist_utils.py || break;\
 	done
 
 	# Copy over CHANGELOG.RST, CONTRIBUTING.RST and LICENSE file to each component directory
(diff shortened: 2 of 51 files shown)


m4dcoder:
@blag I'm not sure I understand the PR here.  The title says backoff some of the dependency changes but the commits here are upgrading the dependencies. Can you provide some more context?

blag:
Sure, so the originating context here is that a few of the weekly pack CI runs have been failing and I took it upon myself to fix them if I could.

One of them was the stackstorm-git pack, which [failed](https://circleci.com/gh/StackStorm-Exchange/stackstorm-typeform/95) with this error:

> `error: amqp 2.5.0 is installed but amqp<3.0,>=2.5.1 is required by set(['kombu'])`

@Kami had updated the dependency version of amqp to 2.5.0 in #4767, which also updated kombu to [v4.6.4](https://github.com/StackStorm/st2/pull/4767/files#diff-a5f1fcca145ad8adf14b1a41a811b878), which in turn [requires `amqp>=2.5.1,<3.0`](https://github.com/celery/kombu/blob/master/requirements/default.txt#L1).

So #4767 broke anything that tried to install the dependencies in `requirements.txt` (like the pack CI), because pip could not satisfy two opposing constraints.

So I upgraded the required amqp version to 2.5.1, and then needed to run `make requirements` to put that change into each ST2 component `requirements.txt`, but that failed due to further issues.

At some point I ran across the error where pip had installed [PyYAML v5.1.2](https://github.com/StackStorm/st2/pull/4767/files#diff-a5f1fcca145ad8adf14b1a41a811b878), but the pack constants still expected `pyyaml>=3.12,<4.0`, so I upgraded that. PyYAML versioning has been a bit chaotic in the past few months, but that's a whole 'nuther story altogether.

The requests library upgrade was good, except that prance [v0.6.1 required `requests~=2.17`](https://github.com/jfinkhaeuser/prance/blob/v0.6.1/setup.py#L63) and we [updated to use version 2.22.0](https://github.com/StackStorm/st2/pull/4689), so that package install failed until I upgraded it to version 0.15.0 and downgraded requests back to 2.21.0, since [prance still pins requests to that version](https://github.com/jfinkhaeuser/prance/blob/master/setup.py#L82).

The next failure I came upon was where `six==1.12.0` was installed in the `fixed-requirements.txt`, but the Makefile explicitly installed version 1.11.0, so I tweaked the makefile to re-install six version 1.12.0.

I wasn't done with dealing with that section of the Makefile though, because immediately after that it uninstalls pytz, python-dateutil, and orquesta, and then runs `pip install` on `requirements.txt` and `test-requirements.txt`. When pip tries to install logshipper, it expects python-dateutil to already have been installed, so pip fails. The only clue I had there was the comment "Fix for Travis CI caching issue", so I threw the pip uninstalls behind a conditional that checks if it's running on Travis. If it is, it uninstalls them, and if it isn't (eg: if I'm just running `make requirements` locally), it leaves the packages installed so installing logshipper succeeds.

I also added instructions on how to update the `dist_utils.py` files and the `requirements.txt` files that we generate.

I've updated the title of this PR to more accurately reflect what it's attempting to achieve.

blag:
It's not just the weekly pack CI runs that are broken, it's also pack PR CI jobs that are broken:
https://github.com/StackStorm-Exchange/stackstorm-zabbix/pull/35
https://circleci.com/gh/StackStorm-Exchange/stackstorm-zabbix/189

armab:
:+1: The good, - it closes https://github.com/StackStorm/st2/issues/4770
:-1: The bad - it re-opens/reverts https://github.com/StackStorm/st2/issues/4680 for which we advertised in [Roadmap](https://docs.stackstorm.com/latest/roadmap.html#release-history) that we support latest requests.

I'm also unsure why do we need Mac OS X build and what consequences it brings. It was discussed several times before that we don't support any OSX-related builds and make sure tries to do that don't affect any st2 code.
If the rabbit hole is OS X related, I'd better go with more minimalistic idea behind the https://github.com/StackStorm/st2/pull/4771 fix.

blag:
I would rather have a roadmap bullet point slip than break all pack CIs.

ST2 still doesn't build on OS X (and likely never will), but at least we can run `make requirements` on it and have it generate the requirements files.

armab:
From the rabbit hole you mentioned in https://github.com/StackStorm/st2/pull/4774#issuecomment-523294302, how the resulting code would differ in this PR without OS X requirements, eg. if making requirements on OS we officially support?

blag:
The only pieces that are Mac OS X support are:

* the new [`scripts/write-headers.sh` script](https://github.com/StackStorm/st2/blob/4a0e8a11ef47da735c7c858520c883d10b9ef8fc/scripts/write-headers.sh)
* [the code that now calls that `write-headers.sh`](https://github.com/StackStorm/st2/pull/4774/files#diff-b67911656ef5d18c4ae36cb6741b7965)

And even with this change, we still only officially support building on Linux. The build still ultimately fails on OS X, since the [Python inotify package](https://pypi.org/project/inotify/) doesn't successfully install on anything but Linux.

I did make sure that `make requirements` successfully completed on Linux, and that it didn't make any changes from when I ran it on OS X.

m4dcoder:
@blag I appreciate the effort you put into getting all of this to work.  I'm OK w/ the fixes here.  The reason with my delayed response here is because the changes are a mishmash of dependency fixes.  I have to reread your explanation several times.  I came to the conclusion here that this fixes a number of dependency issues which plagues exchange CI.  It'll be unproductive to unnecessary block this.  Maybe we do need a better strategy to manage dependency.  But until we identify a solution, I think it will be helpful in the future for all of us to be very succinct about grouping changes and keeping in scope.  For example, OSX is not a supported distro.  For efficiency sake, the recommended platform for developers should be on Linux.  I see little value in maintaining whatever compatibility or functionality on OSX.

m4dcoder:
@armab I'm Ok w/ the fixes here as explained above.  I'll let you do the honor to approve this PR since you're still working on some minor details with @blag.

blag:
Resolved all requested changes.

CHANGELOG.rst
```diff
@@ -14,6 +14,7 @@ Added
 Changed
 ~~~~~~~
 
+* Improved instructions in requirements.txt and dist_utils.py comment headers (improvement) #4774
```
armab: Let's please follow one consistent way of composing changelog: from top to bottom when newer records go to the end like you did for another "fix" changelog, and not vice versa.

armab:
```suggestion
* Improved development instructions in requirements.txt and dist_utils.py comment headers (improvement) #4774
```

CHANGELOG.rst
```diff
@@ -42,6 +43,8 @@ Fixed
   Contributed by JP Bourget (@punkrokk Syncurity) #4732
 * Update ``dist_utils`` module which is bundled with ``st2client`` and other Python packages so it
   doesn't depend on internal pip API and so it works with latest pip version. (bug fix) #4750
+* Fix dependency conflicts in pack CI runs: downgrade requests dependency back to 0.21.0, update
+  internal dependencies and test expectations (amqp, pyyaml, prance, six) (bugfix) #4774
```
armab: :+1:

PR #4788, opened 2019-09-12
Added mask of rule action secret parameters

### LINKED ISSUE #4784: Secret action field show as plaint text in rule view
## SUMMARY

Secret fields are showing as plain text when we see rule information.

### STACKSTORM VERSION

st2 3.1.0, on Python 2.7.5

OS: Oracle Linux 7.5
Custom HA installation with ST2 RPM without changes. All services separate as a single VM.
st2.conf with mask_secrets = True 

Create a RULE with HTTP ACTION and set the password field with anything. When click to see RULE details you can see the password as plain text.

## Expected Results

As a secret parameter of action, that does not need to show in rule information

## Actual Results

We can see all secret field as plain text

Screen Shot 2019-08-30 at 3 46 24 PM

jeansfelix:
Closes #4784


Kami:
Thanks for the contribution.

The change looks good to me, but we would need some tests for it before we can merge it (including model level tests and API level tests).

We will likely also need to update mandatory include fields for rule API endpoints to ensure secrets masking works correctly when ``?include_attributes`` and ``?exclude_attributes`` API query param filters are used.

Nicodemos305:
Dear @Kami , the tests on SO Ubuntu and Centos failed but I couldn't see the error log, would you help me?

blag:
Merging updates from `master` into this branch fixed CI.

armab:
@nicholasamorim Please add a Changelog for this fix to make sure this PR is :100:  ready for final review and merge.

nicholasamorim:
@blag I'm not the PR author :D I've reported a similar issue, though, #4802. I guess this PR closes that too.

armab:
@nicholasamorim Fixed, sorry. Confused by Github autocomplete for the first 3 letters in @

Nicodemos305:
@armab Done!
👍

Kami:
I had a look at it and this needed a bit more work, so I tried to finish it.

I made the following changes:

1. Used existing DB abstraction to retrieve the corresponding DB model and only retrieved the fields we need - bf421a304cc7a410693d5ae90203cf8f97c27510.

2. Made sure those two API endpoints also support ``?show_secrets=true`` query parameters for RBAC admins - 70f960bbfe3758d756e1ac1c861c96b49989bf3d, 0343096689a78af373fa51b5cba01c4f78b21324. All the API endpoints which support secrets masking also need to support that query parameter.

3. Added some test cases which verify secrets masking works correctly - 3400a76aaef258b94e046e4d4be3689e9e064fd6.

Unless I missed something, this is now ready to be merged.

---

My original comment with regards to performance impact / overhead still stands, but that's probably fine for now.

In the future, we can implement some kind of caching / similar to avoid this action retrieval overhead on get / list all API operation.

Kami:
Merged into master via #4807.

PR #4791, opened 2019-09-17
GET entrypoint of a workflow from st2client

trstruth:
This PR implements a method in the st2client module which fetches the entrypoint of a given action ref/id.  This is essentially a wrapper of [this](https://api.stackstorm.com/api/v1/actions/#/entry_point_controller.get_one) endpoint.

We use the st2client package fairly extensively to facilitate tight integration between some of our internal python services and st2.  This PR is one of multiple which extend st2client to wrap more of the existing st2api endpoints which are already implemented.

Random thought - is there a way we can "sync" the api and the client?  ie ensure that all of the api endpoints have a corresponding client method implemented?


nmaludy:
On the "sync" stuff, i know i've brought it up before. I thought about generating the client from the swagger spec.

However, there is some "custom" code that gets added on the client side for some of the stuff to make the human output "pretty".

Would still be nice if there was a way for the end-client CLI to get all of the new features/flags and endpoints, even if it's basic support at firt.

st2client/tests/unit/test_client_actions.py
```diff
@@ -0,0 +1,88 @@
...
+    @classmethod
+    def setUpClass(cls):
+        super(TestActionResourceManager, cls).setUpClass()
+        cls.client = client.Client()
+
+    @mock.patch.object(
+        httpclient.HTTPClient, 'get',
+        mock.MagicMock(return_value=base.FakeResponse(json.dumps(ENTRYPOINT), 200, 'OK')))
+    def test_get_action_entry_point_by_ref(self):
+        self.client.actions.get_entrypoint(EXECUTION['action']['ref'])
```
m4dcoder: Please check that the return value of `self.client.actions.get_entrypoint` is the text of the fake response.

st2client/tests/unit/test_client_actions.py
```diff
@@ -0,0 +1,88 @@
...
+
+        endpoint = '/actions/views/entry_point/%s' % EXECUTION['action']['ref']
+
+        httpclient.HTTPClient.get.assert_called_with(endpoint)
+
+    @mock.patch.object(
+        httpclient.HTTPClient, 'get',
+        mock.MagicMock(return_value=base.FakeResponse(json.dumps(ENTRYPOINT), 200, 'OK')))
+    def test_get_action_entry_point_by_id(self):
+        self.client.actions.get_entrypoint(EXECUTION['id'])
```
m4dcoder: Please check that the return value of `self.client.actions.get_entrypoint` is the text of the fake response.

st2client/tests/unit/test_client_actions.py
```diff
@@ -0,0 +1,88 @@
...
+
+        endpoint = '/actions/views/entry_point/%s' % EXECUTION['id']
+
+        httpclient.HTTPClient.get.assert_called_with(endpoint)
```
m4dcoder: Please add a unit tests where the status code from the API call is not OK/200.

trstruth: Pushed a commit for this, but not sure if I implemented it correctly, please let me know if I should take a different approach.

m4dcoder: Can you use `assertRaisesRegexp` and check that the error message is actually Not Found returned in the FakeResponse?  Search for `assertRaisesRegexp` in st2 and you will see plenty of examples.

trstruth: @m4dcoder sorry for the delay, I've pushed a commit to use that method.  Please let me know if it's not correct.

m4dcoder:
@trstruth You also need to rebase this branch.  There is a merge conflict.

trstruth:
@m4dcoder I've resolved the conflict, let's see if the ci builds pass this time

PR #4792, opened 2019-09-19
Don't call eventlet directly inside the st2reactor code

Kami:
This pull request updates code in ``st2reactor`` package so all the concurrency related operations are routed through our ``st2common.util.concurrency`` abstraction.

This way that code should work with either eventlet or gevent (e.g. ``eventlet`` doesn't work in combination with the ``grpc`` library).

This is similar as the change in #4713, but for the ``st2reactor`` package.


st2common/st2common/config.py
```diff
@@ -374,6 +374,10 @@ def register_opts(ignore_errors=False):
             'virtualenv_opts', default=['--system-site-packages'],
             help='List of virtualenv options to be passsed to "virtualenv" command that '
                  'creates pack virtualenv.'),
+        cfg.ListOpt(
+            'pip_opts', default=[],
+            help='List of pip options to be passed to "pip install" command when installing pack '
+                 'dependencies into pack virtual environment.'),
```
m4dcoder: What do these pip related changes have to do with the changes to not directly call eventlet?

st2common/st2common/util/concurrency.py
```diff
@@ -111,3 +113,21 @@ def kill(green_thread, *args, **kwargs):
         return green_thread.kill(*args, **kwargs)
     else:
         raise ValueError('Unsupported concurrency library')
+
+
+def sleep(*args, **kwargs):
+    if CONCURRENCY_LIBRARY == 'eventlet':
```
m4dcoder: Seems like there should be a better way to handle the eventlet vs. gevent driver than to use all these if/else across different methods in this module.

st2reactor/st2reactor/container/process_container.py
```diff
@@ -322,27 +313,34 @@ def _spawn_sensor_process(self, sensor):
...
+            metadata = {
+                'service': 'sensors_container',
+                'sensor_path': sensor['file_path'],
+                'sensor_class': sensor['class_name']
+            }
+            temporary_token = create_token(username='sensors_container', ttl=ttl, metadata=metadata,
+                                           service=True)
+
+            env[API_URL_ENV_VARIABLE_NAME] = get_full_public_api_url()
+            env[AUTH_TOKEN_ENV_VARIABLE_NAME] = temporary_token.token
```
m4dcoder: What do these token related changes have to do with the concurrency change?

st2common/st2common/util/concurrency.py
Kami: I don't think we get much with doing it differently.

We could abstract it some more, in some driver / plugin approach but that would be massive amount of work for very little value.

st2common/st2common/config.py
Kami: Ideally I would open separate PR for each of those set of changes, but I kinda wanted to avoid that.

PR #4794, opened 2019-09-27
Pip already respects

guzzijones:
pip already respects https_proxy and http_proxy;  Further if these ARE set via command line pip then no_proxy setting is ignored by pip and a local pypi repo is not usable.

### DIFF
--- a/st2common/st2common/util/virtualenvs.py
+++ b/st2common/st2common/util/virtualenvs.py
@@ -241,14 +241,6 @@ def install_requirements(virtualenv_path, requirements_file_path, proxy_config=N
 
     if proxy_config:
         cert = proxy_config.get('proxy_ca_bundle_path', None)
-        https_proxy = proxy_config.get('https_proxy', None)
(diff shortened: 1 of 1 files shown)


guzzijones:
This currently makes setting no_proxy have no affect.  Can someone please review this?

Kami:
Thanks for the contribution.

One of the reasons we added those options is because setting ``https_proxy`` and ``http_proxy`` environment variable for ``st2actionrunner`` service / process is not totally trivial. And it also means you can't use different proxies for different ``st2 pack`` command invocations.

Depending on the deployment model and service manager used, user might need to edit service manager config file directly to set those environment variables and that's not really user-friendly.

So because of that reason, I think it would still be better to leave those options in place and modify / fix them if needed (e.g. https://github.com/StackStorm/st2/blob/master/contrib/packs/actions/pack_mgmt/setup_virtualenv.py#L50 here and in other places).

Kami:
Closing this since it doesn't appear to be relevant.

If it's still relevant, please open a new PR - thanks!

Issue #4798, opened 2019-10-02
Writes to Mongodb in Actionrunner are slow and cpu intensive with large documents


jdmeyer3:
## SUMMARY

When an action return ~4MB of data the actionrunner writing to MongoDB takes ~20 seconds for each document. During the writes, the CPU utilization spikes to 100%.

### STACKSTORM VERSION

St2 v3.1.0

OS: CentOS 7.7.1908 Kernel 3.10.0-957.el7.x86_64
Kubernetes v1.14.1
Docker 19.03.2
Base Docker Image: CentOS 7.6.1810 (custom image)

## Steps to reproduce the problem

In a *Python 3* pack, create a python action with the following code

```python
#!/usr/bin/env python3

import json

import requests
import datetime
from st2common.runners.base_action import Action


class RequestLargeDataset(Action):
    """
    Retrieves ~4MB of json data
    """

    def run(self, **kwargs):
        dataset = requests.get('https://data.ct.gov/api/views/rybz-nyjw/rows.json?accessType=DOWNLOAD')
        data = json.loads(dataset.text)
        return data

```
## Expected Results

Low CPU utilization and the result returning within a relatively short time (<5 seconds) 

## Actual Results

CPU spiked to 100% for 20-30 seconds 

![image](https://user-images.githubusercontent.com/25443446/66070634-ebce9d00-e50e-11e9-8bd5-e9bc7c22ba78.png)

adding some logs around `st2actions.container.base.py:296` with
```python

    def _update_status(self, liveaction_id, status, result, context):
        try:
            LOG.error("updating liveaction")
            LOG.debug('Setting status: %s for liveaction: %s', status, liveaction_id)
            liveaction_db, state_changed = self._update_live_action_db(
                liveaction_id, status, result, context)
            LOG.error("done updating liveaction")
        except Exception as e:
            LOG.exception(
                'Cannot update liveaction '
                '(id: %s, status: %s, result: %s).' % (
                    liveaction_id, status, result)
            )
            raise e

        try:
            LOG.error('updating execution')
            executions.update_execution(liveaction_db, publish=state_changed)
            LOG.error('done updating execution')
            extra = {'liveaction_db': liveaction_db}
            LOG.debug('Updated liveaction after run', extra=extra)
        except Exception as e:
            LOG.exception(
                'Cannot update action execution for liveaction '
                '(id: %s, status: %s, result: %s).' % (
                    liveaction_id, status, result)
            )
            raise e

        return liveaction_db
```

[...]

PR #4803, opened 2019-10-17
fixing subprocess to use system buffer instead of being unbuffered

jdmeyer3:
While testing large datasets in actions, I came across massive CPU spikes for a relatively small dataset (~8MB) being returned from the action. While investigating the issue I found two problems that were consuming the CPU. The first issue was documented in #4798. The second issue was with how the action was created. When the subprocess is created without the buffer size, it defaults to zero (unbuffered). According to the subprocess docs, 
>"If you experience performance issues, it is recommended that you try to enable buffering by setting bufsize to either -1 or a large enough positive value (such as 4096)." https://docs.python.org/2/library/subprocess.html.

 The bufsize can be a discussion whether to use the system default or a large value. After making this change on my local machine I noticed a performance increase minus the MongoEngine issue which still persists.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -28,6 +28,12 @@ Changed
   writing very large executions (executions with large results) to the database. #4767
 * Improved development instructions in requirements.txt and dist_utils.py comment headers
   (improvement) #4774
+* Add new ``actionrunner.stream_output_buffer_size`` config option and default it to ``-1``
+  (previously default value was ``0``). This should result in a better performance and smaller
+  CPU utilization for Python runner actions which produce a lot of output.
+  (improvement)
+
+  Reported and contributed by Joshua Meyer (@jdmeyer3) #4803
 
 Fixed
 ~~~~~
--- a/conf/st2.conf.sample
+++ b/conf/st2.conf.sample
@@ -20,6 +20,8 @@ python3_prefix = None
 virtualenv_binary = /usr/bin/virtualenv
 # Python 3 binary which will be used by Python actions for packs which use Python 3 virtual environment.
 python3_binary = /usr/bin/python3
+# Buffer size to use for real time action output streaming. 0 means unbuffered 1 means line buffered, -1 means system default, which usually means fully buffered and any other positive value means use a buffer of (approximately) that size
+stream_output_buffer_size = -1
 # List of virtualenv options to be passsed to "virtualenv" command that creates pack virtualenv.
 virtualenv_opts = --system-site-packages # comma separated list allowed here.
 # True to store and stream action output (stdout and stderr) in real-time.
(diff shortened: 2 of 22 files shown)


blag:
How does this impact the live execution results?

armab:
I'm wondering if this change may affect the real-time live output from actions like `st2 execution tail`?

Performance wise, you may also benefit from disabling "live output"
https://github.com/StackStorm/st2/blob/a643ba7a32f8908237c715b87a223c08743df578/conf/st2.conf.sample#L25-L26
which also takes a lot of CPU cycles.

jdmeyer3:
@blag At a `-1` value for buffering, it would depend on your system. Some systems will fully buffer while others will buffer 4KB - 8KB of memory before flushing. If the buff value is hardcoded to `4096` then the live output should be able to display the information in 4K increments. This is to avoid having the CPU process the output from the execution one byte at a time and instead buffer to reduce the compute overhead. I can change it to a hardcoded buffer value instead of the system default if that works better for you.

Kami:
Yeah, IIRC, in the past we've seen some issues with real time output hanging on some distros and that's why we used unbuffered output.

I'm fine with this change as long as we document it and make it configurable via config option in st2.conf (e.g. ``actionrunner.stream_output_buffer_size``).

This way users who would potentially have issues can change that value in the config instead of needing to directly edit the code.

I can make that change and merge that into master.

Kami:
I pushed the following changes:

1. Add new ``actionrunner.stream_output_buffer_size`` config option and default it to ``-1`` per @jdmeyer3 observations. 2f96d5db92ac270932cfa648acdda049e8b1cfe0

2. Only use ``bufsize=-1`` for Python runner actions. Changing the default for all the ``st2common.util.green.shell.run_command()`` function calls would affect a lot of other code as well so it's better to keep the change small and more contained. 25ea404641d80b5e49243fbf6635ddea81407b64, 2f96d5db92ac270932cfa648acdda049e8b1cfe0.

2. Add some basic tests for it. 711436fc167e4e209e0f5fda0b490b6f7e964c29

3. Document this change in upgrade notes - https://github.com/StackStorm/st2docs/pull/941.

We should still keep an eye on potential regressions on other distributions, but I think now that the value can be controlled via config option, it's safe enough to be merged into master.

If users experience issues, they can change the config option value to revert back to the old behavior.

Kami:
@jdmeyer3^ let me know if you are OK with the changes above.

Kami:
There were some weird / random Travis test failures.

I need to push some debugging changes (0b76bd505e0801bdf3e403c795dcbb84d5e0bf79) which will hopefully provide us with more information and context which will allow us to troubleshoot such issues in the future.

jdmeyer3:
@Kami Ya, those changes are fine with me.

Kami:
Yay, all green!

Thanks again for the contribution. I will go ahead and merge it into master.


@jdmeyer3  When you get a chance, can you please test the version / changes from master and confirm it's working as expected?

PR #4809, opened 2019-10-31
Fix a bug with authentication API endpoint returning internal server error under gunicorn if auth.api_url config option was not set

Kami:
This pull request fixes a bug with ``POST /auth/v1/tokens``API endpoint returning internal server error when running under gunicorn without ``auth.api_url`` config option being set.

NOTE: First I couldn't reproduce the issue, but it turns out it only happens when using gunicorn deployment.

When running st2auh directly, the WSGI implementation simply casts all the "None" header values to empty strings, but that's not the case for gunicorn.

Originally reported here - https://forum.stackstorm.com/t/ldap-error-handling-api-call-failed-expected-string-or-buffer/893.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -65,6 +65,11 @@ Fixed
   ``GET /v1/rules`` and ``GET /v1/rules/<ref>`` API endpoint. (bug fix) #4788 #4807
 
   Contributed by @Nicodemos305 and @jeansfelix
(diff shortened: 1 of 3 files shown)


Kami:
@armab oops, done - b8b874fcadc471d2db88011c9ef9e912a6eaa87f,

I already had it ready locally, I was just waiting to open a PR first so I get a PR number which I can reference in the changelog :)

PR #4819, opened 2019-11-19
Update pip dependencies, fix conflicting dependency version, ensure there are no conflicting dependencies

### LINKED ISSUE #4818: Master currently broken because of dependency conflict between requests and cryptography
## SUMMARY

The ST2 requirements.txt of `st2common` requires `cryptography==2.7` and `requests[security]<2.22.0,>=2.21.0`. The problem is, `requests[security]<2.22.0,>=2.21.0` requires `cryptography>=2.8`, so there is a **dependency conflict**.

It's the same kind of problem as https://github.com/StackStorm/st2/issues/4770.

Pip fails with the following error:
```
ERROR: pyopenssl 19.1.0 has requirement cryptography>=2.8, but you'll have cryptography 2.7 which is incompatible.
```

When using a **recent version of `pip`** (e.g. 19.3.1), this will cause pip to **fail to install the requirements**. However, if you try with an old version (e.g. 9.0.3), it will work. I don't know at which specific version of Pip it stops working.

**Note**: this might be considered urgent because as https://github.com/StackStorm-Exchange/ci uses the master branch of ST2, **it is currently breaking CIs** (depending on what version of Pip is used by the CI, and maybe the OS?).

### STACKSTORM VERSION

Current master branch, bf173256a88a00f2cf64ec918f92cd67c5b6c927

Ubuntu 16.04

It's easier to reproduce with Docker.

1. Start a container with the following command: `docker run -it ubuntu:16.04 /bin/bash`
2. Run the following commands inside the Docker container:
```
apt-get update
apt-get install python3 python3-pip python3-dev git build-essential libffi-dev libssl-dev nano
pip3 install virtualenv # To have a newer version of pip easily
virtualenv st2-requests-test
cd st2-requests-test
source ./bin/activate
# Create the st2/requirements.txt file
pip install -r requirements.txt # This will fail
```

## Expected Results

Should correctly install all the dependencies when doing `pip install -r requirements.txt`.

## Actual Results

Pip fails with the following error:
```
ERROR: pyopenssl 19.1.0 has requirement cryptography>=2.8, but you'll have cryptography 2.7 which is incompatible.
```

Kami:
This pull request includes the following changes:

1. Update ``cryptography`` dependency to v2.8 since version of ``requests`` we use depends on that version. As per https://github.com/StackStorm/st2/issues/4818#issuecomment-555545997, that is not actually a fatal issue since pip just prints a warning and doesn't fail on install, but it's still good to get it sorted out.
2. Update various dependencies to latest versions (jinja2, requests and pip, setuptools and pbr for tests)
3. Make sure versions in ``test-requirements.txt`` are in sync with versions in ``requirements.txt``
4. Run ``pip-conflict-checker`` as part of ``make requirements`` step. This command will fail and exit with non-zero if there are any conflicting dependency versions.

## TODO

- [x] Changelog entries

Resolves #4818.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -441,7 +441,7 @@ requirements: virtualenv .sdist-requirements install-runners
 	$(VIRTUALENV_DIR)/bin/pip --version
 	$(VIRTUALENV_DIR)/bin/pip install --upgrade "pip==19.3.1"
 	$(VIRTUALENV_DIR)/bin/pip install --upgrade "setuptools==41.0.1"  # Required for packs.install in dev envs
-	$(VIRTUALENV_DIR)/bin/pip install --upgrade "pbr==5.4.0"  # workaround for pbr issue
+	$(VIRTUALENV_DIR)/bin/pip install --upgrade "pbr==5.4.3"  # workaround for pbr issue
 
 	# Generate all requirements to support current CI pipeline.
 	$(VIRTUALENV_DIR)/bin/python scripts/fixate-requirements.py --skip=virtualenv,virtualenv-osx -s st2*/in-requirements.txt contrib/runners/*/in-requirements.txt -f fixed-requirements.txt -o requirements.txt
--- a/Makefile
+++ b/Makefile
@@ -446,6 +446,9 @@ requirements: virtualenv .sdist-requirements install-runners
 	# Generate all requirements to support current CI pipeline.
 	$(VIRTUALENV_DIR)/bin/python scripts/fixate-requirements.py --skip=virtualenv,virtualenv-osx -s st2*/in-requirements.txt contrib/runners/*/in-requirements.txt -f fixed-requirements.txt -o requirements.txt
 
+	# Remove any *.egg-info files which polute PYTHONPATH
+	rm -rf *.egg-info*
+
 	# Generate finall requirements.txt file for each component
 	@for component in $(COMPONENTS_WITH_RUNNERS); do\
 		echo "==========================================================="; \
--- a/Makefile
+++ b/Makefile
@@ -150,7 +150,6 @@ check: check-requirements flake8 checklogs
 # make targets. This speeds up the build
 .PHONY: install-runners
 install-runners:
-
 	@echo ""
 	@echo "================== INSTALL RUNNERS ===================="
 	@echo ""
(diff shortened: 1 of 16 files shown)


Makefile
```diff
@@ -440,12 +439,16 @@ requirements: virtualenv .sdist-requirements install-runners
...
-	$(VIRTUALENV_DIR)/bin/pip install --upgrade "virtualenv==16.6.0" # Required for packs.install in dev envs
+	$(VIRTUALENV_DIR)/bin/pip install --upgrade "pip==19.3.1"
+	$(VIRTUALENV_DIR)/bin/pip install --upgrade "setuptools==41.0.1"  # Required for packs.install in dev envs
+	$(VIRTUALENV_DIR)/bin/pip install --upgrade "pbr==5.4.0"  # workaround for pbr issue
 
 	# Generate all requirements to support current CI pipeline.
 	$(VIRTUALENV_DIR)/bin/python scripts/fixate-requirements.py --skip=virtualenv,virtualenv-osx -s st2*/in-requirements.txt contrib/runners/*/in-requirements.txt -f fixed-requirements.txt -o requirements.txt
 
+	# Remove any *.egg-info files which polute PYTHONPATH
+	rm -rf *.egg-info*
```
Kami: We should have done that before.

We add repo root to ``PYTHONPATH`` in virtualenv we create which means if you run ``python setup.py install`` or similar in ``st2client/`` directory, it will create egg-info directory for that Python package which will mask actual ``st2client/`` directory in ``PYTHONPATH``.

Kami:
I confirmed conflict checker tool works correctly. Here is an example if we use ``kombu==4.6.6`` and ``amqp==2.5.1``.

```bash
# Verify there are no conflicting dependencies
virtualenv/bin/pipconflictchecker
--------------------------------------------------
 Conflicts Detected
--------------------------------------------------
 - amqp(2.5.1) kombu(>=2.5.2,<2.6)
Makefile:437: recipe for target 'requirements' failed
```

It will return an error since that version of kombu request ``amqp >= 2.5.2``.

test-requirements.txt
```diff
@@ -19,16 +19,17 @@ nose-timer==0.7.5
...
 rstcheck>=3.3.1,<3.4
-tox==3.13.2
+tox==3.14.1
 pyrabbit
 # Since StackStorm v2.8.0 we now use cryptography instead of keyczar, but we still have some tests
 # which utilize keyczar and ensure new cryptography code is fully compatible with keyczar code
 # (those tests only run under Python 2.7 since keyczar doesn't support Python 3.x).
 # See https://github.com/StackStorm/st2/pull/4165
 python-keyczar
+git+https://github.com/Kami/pip-conflict-checker.git@fix_pip_issue#egg=pip-conflict-checker
```
Kami: Just FYI - that's the change we need - https://github.com/ambitioninc/pip-conflict-checker/pull/14.

I can also push that fork to StackStorm org and use that version until it's merged upstream.

Kami: Here we go - febde4622.

armab:
+  :100:

PR #4829, opened 2019-12-04
#4795 fix 'core.sendmail' for subject line wrapping

### LINKED ISSUE #4795: core.sendmail action UTF-8/base64 encoding doesn't handle longer subject lines properly
## SUMMARY

When using the ``core.sendmail`` action if the subject line is too long (approx 56 but apparently really depends on the line length of the base64 encoded text) then the encoding apparently line-wraps which in turn causes the formatting of the subject and the body of the message to get mangled.

### STACKSTORM VERSION

Paste the output of ``st2 --version``: ``st2 3.1.0, on Python 2.7.5``

RPM install on CentOS 7.7

`st2 run core.sendmail from="st2 notifier <XXXXXX@plexxi.com>" to=XXXXXXXXX@plexxi.com  body="<html><body><h2>Test message</h2>How does this end up getting formatted?<p>new line here</p></body></html>" subject="This is a subject line that is too long - it will get mangled"`

## Expected Results

HTML message with subject line: `This is a subject line that is too long - it will get mangled` and HTML formatted body

## Actual Results

Actual email message headers/body look like this:

>
>

>TO: XXXXXXXXX@plexxi.com
>FROM: st2 notifier <XXXXXX@plexxi.com>
>SUBJECT: =?UTF-8?B?>VGhpcyBpcyBhIHN1YmplY3QgbGluZSB0aGF0IGlzIHRvbyBsb25nIC0gaXQgd2lsbCBnZXQgbWFu
>Message-Id: <20190930180507.XXXXXXXX@XXXXXXXX.plexxi.com>
>Date: Mon, 30 Sep 2019 14:05:07 -0400 (EDT)
>
>Z2xlZAo=?=
>Content-Type: text/html; charset=UTF-8
>
><html><body><h2>Test message</h2>How does this end up getting formatted?<p>new line >here</p></body></html>
><br><br>
>This message was generated by StackStorm action send_mail running on XXXXXX.plexxi.com`

As you can see, the subject line apparently wrapper and flowed onto another line which resulted in a message that looked like this:

st2sendmail_bugreport

Shortening the subject line by a couple of characters causes everything to work properly and the HTML message is formatted as expected.

stevemuskiewicz:
Fixes #4795: add `-w 0` to base64 encoding of subject line to avoid line wrapping

Credit to @guzzijones for finding the fix here https://superuser.com/questions/1225134/why-does-the-base64-of-a-string-contain-n

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -85,6 +85,9 @@ Fixed
 * Update ``st2 execution get`` command to always include ``context.user``, ``start_timestamp`` and
   ``end_timestamp`` attributes. (improvement) #4739
 
+  Contributed by @stevemuskiewicz and @guzzijones
+* Fixed ``core.sendmail`` base64 encoding of longer subject lines (bug fix) #4795
+
 3.1.0 - June 27, 2019
 ---------------------
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -85,8 +85,9 @@ Fixed
 * Update ``st2 execution get`` command to always include ``context.user``, ``start_timestamp`` and
   ``end_timestamp`` attributes. (improvement) #4739
 
-  Contributed by @stevemuskiewicz and @guzzijones
 * Fixed ``core.sendmail`` base64 encoding of longer subject lines (bug fix) #4795
+
+  Contributed by @stevemuskiewicz and @guzzijones
 
 3.1.0 - June 27, 2019
 ---------------------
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -85,6 +85,9 @@ Fixed
 * Update ``st2 execution get`` command to always include ``context.user``, ``start_timestamp`` and
   ``end_timestamp`` attributes. (improvement) #4739
 
+  Contributed by @stevemuskiewicz and @guzzijones
```
armab: This line needs to be moved not before, but after the actual changelog record to be part of the same block.

See how it looks like https://github.com/StackStorm/st2/blob/0fffbdf00cf4d547d254bb4319d6bb5fcb42e976/CHANGELOG.rst#fixed rendered.

stevemuskiewicz: Wups, sorry about that.  I'll fix

PR #4831, opened 2019-12-09
Update rule comparison operators to work with mixed operator types (bytes and strings)

Kami:
This pull request updates all the rule criteria comparison operators which also work with strings (a lot of operators work on multiple types), to try to convert both operators to unicode / string type if either or both of them are of a type bytes.

In ideal world, user would always know a type of each input, but that's rarely the case. Depending on how data is retrieved or defined (e.g. webhook body, etc.) the actual type might be bytes or a string / unicode.

This means that pretty much the only option we are left with for rule comparison operators to work as expected when mixed types come into play is to try to convert both of the operators to the same common type aka unicode / string.

NOTE: This issue / problem only affects Python 3.x deployments.


Kami:
That's correct, this issue only affects Python 3. Tests I've added would already pass under Python 2.x without the changes I added.

Having said that, in ideal world, we (and our users) would always be very specific and explicit about the types so we wouldn't need to do this "best effort" type of conversion, but this breaks down in real life for various reasons:

1. Python is dynamically typed
2. Our platform relies on dynamic types and lose conversion between types in many scenarios (for better or worse)
3. Jinja expressions / variables only support string / unicode type
4. There are many ways for data to come into the system (webhooks, sensors, etc.) and we can't always enforce that this data conforms to a specific schema / type

As far as the forum post goes - yes, I believe this could be a fix for that issue, but I haven't seen any logs yet and that was just my best guess when I checked the operators code (and something which immediately stood out to me). It could also be that there are some other issues as well.

PR #4833, opened 2019-12-16
Add support for orquesta task retry

m4dcoder:
Update orquesta requirements to commit with the task retry feature. Add orquesta integration tests to cover task retry.

### DIFF
--- a/contrib/examples/actions/workflows/orquesta-task-retry-exhausted.yaml
+++ b/contrib/examples/actions/workflows/orquesta-task-retry-exhausted.yaml
@@ -1,9 +1,11 @@
 version: 1.0
 
 description: >
-  A sample workflow that illustrates the use of task retry to rerun task on error. In this example,
-  the workflow will loop until the file /tmp/done exists. A parallel task will wait for some time
-  before creating the file. When completed, /tmp/done will be deleted.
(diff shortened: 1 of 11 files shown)


contrib/examples/actions/workflows/orquesta-task-retry-exhausted.yaml
```diff
@@ -0,0 +1,35 @@
+version: 1.0
+
+description: >
+  A sample workflow that illustrates the use of task retry to rerun task on error. In this example,
+  the workflow will loop until the file /tmp/done exists. A parallel task will wait for some time
+  before creating the file. When completed, /tmp/done will be deleted.
```
Kami: Description should probably say that this workflow will demonstrate what happens when maximum retry count is reached / exhausted.

m4dcoder: Done

Kami:
As per Slack comment - it would be nice if we could soon deploy this feature to our CI/CD server and also start utilizing it in our various ci/cd workflows.

This are a lot of tasks / actions which would benefit from retries because of various network and other intermediate issues (similar as we did with retry policy in the past).

In addition to helping us reduce restart failed execution workload, it would also be good "at scale" test for this future.

PR #4834, opened 2019-12-16
Fix SSL support for MongoDB and RabbitMQ under Python 3.x

### LINKED ISSUE #4832: Connection to mongo db over SSL doesn't work
## SUMMARY

Unable to connect Stackstorm to Mongo DB over SSL.
Stackstorm can't function correctly when connection to MongoDB is encrypted by SSL,
There is exception of "aximum recursion depth exceeded while calling a Python object"  in st2api.log and st2auth.log. In this conditions stackstorm is in non functional state at all.

### STACKSTORM VERSION

st2 3.1.0, on Python 3.6.9
:

OS: Both the stackstorm and moongodb on Ubuntu 18.04 LTS but installed in diffrent VMs in the same subnet.

All Stackstorm components are installed on one VM but Mongo DB, RAbitMQ on the second VM.

Without SSL all works fine, but after moving a connection to SSL, Stackstorm can't connect to Mongo.
1. Put ssl=true in st2.conf

2. Other parameters related to ssl ,don't really matter . Tested in all combinations of these parameters with and without client certificate authentication and so on.

3. mongo client can connect to monog over ssl without any problem.  Simple python script using mongoengine module can connect to mongo over ssl. So it seems as not a problem of underlying infrastructure because all other tools and components can connect to mongo wih SSL but staskstorm can't/

## Expected Results

Staskstoem should work with mongo over SSL. That a basic requirements .
## Actual Results

Stackstorm starting up but in the st2api. log and st2auth.log we can see every few seconds error 

2019-12-15 08:08:45,891 140651884441824 INFO (unknown file) [-] Connecting to database "somedatabase" @ "mongodb-host:27017" as user "someuser".
2019-12-15 08:08:45,892 140651884441824 WARNING (unknown file) [-] Retry on ConnectionError - Cannot connect to database default :
maximum recursion depth exceeded while calling a Python object

it seems like some bug in __init__py file in _db_connect function in line number 125 where the connection attempt goes to endless recursion or something likes this.

Kami:
This pull request fixes #4832  SSL support for MongoD and RabbitMQ when running under Python 3.6.

## Background / Details

When running under Python 3 and enabling SSL support for MongoDB, st2api and st2auth would fail with a cryptic error:

```bash
2019-12-15 08:08:45,891 140651884441824 INFO (unknown file) [-] Connecting to database "somedatabase" @ "mongodb-host:27017" as user "someuser".
2019-12-15 08:08:45,892 140651884441824 WARNING (unknown file) [-] Retry on ConnectionError - Cannot connect to database default :
maximum recursion depth exceeded while calling a Python object
```

After some more digging in and manual instrumentation, I managed to track down the original exception / root cause - https://gist.github.com/Kami/ea8e63cdc539fd879fff41271969d650.

The root cause was an SSL error which happened because ``st2api`` and ``st2auth`` didn't perform eventlet monkey patching early enough. They performed monkey patching after some other module (likely mongoengine or pymongo) already imported ``ssl`` which has undefined behavior and won't work.

## Proposed Fix

This pull request fixes the issue by making sure we perform eventlet monkey patching as early as possible (before any other modules are imported) inside ``st2api`` and ``st2auth`` service.

In addition to that, I made another change by setting ``serverSelectionTimeoutMS`` MongoClient option to 5 seconds.

It defaults to 30 seconds which means if there is a SSL related connection error (e.g. handshake failed), exception won't be raised until 30 seconds has passed.

Keep in mind though that in such scenarios, only "connection closed" exception will be returned to the user and for the actual root cause, user will need to check the mongo server logs (that's the limitation of the client / server and nothing we can do).

This will also provide a better user-experience because previously in many scenarios our code would wait 30 seconds for this timeout to be reached before propagating the connection error.

## Affected Components / Services

Based on my digging it, this issue only affected ``st2api`` and ``st2auth`` because other services already performed eventlet monkey patching as early as possible.

This also explains why user reported that ``st2-register-content`` worked in #4832 (we intentionally don't perform any monkey patching there so it works fine).

Another thing worth keeping in mind is that we have two entry points for ``st2api`` and ``st2auth`` - WSGI one (for production gunicorn deployments) and the direct eventlet WSGI server one (for local testing and development).

Both entry points were affected so I needed to fix both.

## Configuration

For completeness, here is the st2.conf and mongod.conf configuration I used:

st2.conf:

```ini
...
[database]
ssl = true
host = 127.0.0.1
ssl_cert_reqs = none
username = stackstorm
password = ...
...
```

mongod.conf:

```yaml
...
net:
  port: 27017
  bindIp: 127.0.0.1
  ssl:
    mode: requireSSL
    PEMKeyFile: /home/ubuntu/mongodb.pem
    CAFile: /home/ubuntu/rootCA.pem
    allowConnectionsWithoutCertificates: true  # NOTE: If this option is not set, client also needs to specify client server otherwise connection / auth will fail
...
```

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -47,6 +47,10 @@ Changed
   function must be called separately. (improvement)
 * Update various internal dependencies to latest stable versions (cryptography, jinja2, requests,
   apscheduler, eventlet, amqp, kombu, semver, six) #4819 (improvement)
+* Improve MongoDB connection timeout related code. Connection and server selection timeout is now
+  set to 3 seconds. Previously a default value of 30 seconds was used which means that for many
+  connection related errors, our code would first wait for this timeout to be reached (30 seconds)
+  before returning error to the end user. #4384
 
 Fixed
 ~~~~~
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -50,7 +50,7 @@ Changed
 * Improve MongoDB connection timeout related code. Connection and server selection timeout is now
   set to 3 seconds. Previously a default value of 30 seconds was used which means that for many
   connection related errors, our code would first wait for this timeout to be reached (30 seconds)
-  before returning error to the end user. #4384
+  before returning error to the end user. #4834
 
 Fixed
 ~~~~~
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -99,7 +99,7 @@ Fixed
   result in cryptic "maximum recursion depth exceeded while calling a Python object" error on
   connection failure.
 
-  NOTE: This issue only affected installations using Python 3. (bug fix) #4382 #4384
+  NOTE: This issue only affected installations using Python 3. (bug fix) #4832 #4834
 
   Reported by @alexku7.
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -97,7 +97,9 @@ Fixed
   of a type bytes and the other is of a type unicode / string. (bug fix) #4831
 * Fix SSL connection support for MongoDB and RabbitMQ which wouldn't work under Python 3 and would
   result in cryptic "maximum recursion depth exceeded while calling a Python object" error on
-  connection failure. (bug fix) #4382 #4384
+  connection failure.
+
+  NOTE: This issue only affected installations using Python 3. (bug fix) #4382 #4384
 
   Reported by @alexku7.
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -91,6 +91,11 @@ Fixed
 * Update all the various rule criteria comparison operators which also work with strings (equals,
   icontains, nequals, etc.) to work correctly on Python 3 deployments if one of the operators is
   of a type bytes and the other is of a type unicode / string. (bug fix) #4831
+* Fix SSL connection support for MongoDB and RabbitMQ which wouldn't work under Python 3 and would
+  result in cryptic "maximum recursion depth exceeded while calling a Python object" error on
+  connection failure. (bug fix) #4382 #4384
+
+  Reported by @alexku7.
 
 3.1.0 - June 27, 2019
 ---------------------
--- a/st2common/st2common/models/db/__init__.py
+++ b/st2common/st2common/models/db/__init__.py
@@ -25,6 +25,7 @@ from mongoengine.queryset import visitor
 from pymongo import uri_parser
 from pymongo.errors import OperationFailure
 from pymongo.errors import ConnectionFailure
+from pymongo.errors import ServerSelectionTimeoutError
 
 from st2common import log as logging
 from st2common.util import isotime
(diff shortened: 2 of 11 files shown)


st2common/st2common/models/db/__init__.py
```diff
@@ -122,9 +123,13 @@ def _db_connect(db_name, db_host, db_port, username=None, password=None,
                                  authentication_mechanism=authentication_mechanism,
                                  ssl_match_hostname=ssl_match_hostname)
 
+    # NOTE: We intentionally set "serverSelectionTimeoutMS" to 5 seconds. By default it's set to
+    # 30 seconds, which means it will block up to 30 seconds and fail if there are any SSL related
+    # errors
     connection = mongoengine.connection.connect(db_name, host=db_host,
                                                 port=db_port, tz_aware=True,
                                                 username=username, password=password,
+                                                serverSelectionTimeoutMS=5000,
```
Kami: For "just in case", I will also make this configurable via st2.conf.

CHANGELOG.rst
```diff
@@ -91,6 +95,11 @@ Fixed
 * Update all the various rule criteria comparison operators which also work with strings (equals,
   icontains, nequals, etc.) to work correctly on Python 3 deployments if one of the operators is
   of a type bytes and the other is of a type unicode / string. (bug fix) #4831
+* Fix SSL connection support for MongoDB and RabbitMQ which wouldn't work under Python 3 and would
+  result in cryptic "maximum recursion depth exceeded while calling a Python object" error on
+  connection failure. (bug fix) #4382 #4384
```
armab:
```suggestion
  connection failure. (bug fix) #4832 #4834
```

CHANGELOG.rst
```diff
@@ -47,6 +47,10 @@ Changed
   function must be called separately. (improvement)
 * Update various internal dependencies to latest stable versions (cryptography, jinja2, requests,
   apscheduler, eventlet, amqp, kombu, semver, six) #4819 (improvement)
+* Improve MongoDB connection timeout related code. Connection and server selection timeout is now
+  set to 3 seconds. Previously a default value of 30 seconds was used which means that for many
+  connection related errors, our code would first wait for this timeout to be reached (30 seconds)
+  before returning error to the end user. #4384
```
armab:
```suggestion
  before returning error to the end user. #4834
```

CHANGELOG.rst
```diff
@@ -91,6 +95,13 @@ Fixed
 * Update all the various rule criteria comparison operators which also work with strings (equals,
   icontains, nequals, etc.) to work correctly on Python 3 deployments if one of the operators is
   of a type bytes and the other is of a type unicode / string. (bug fix) #4831
+* Fix SSL connection support for MongoDB and RabbitMQ which wouldn't work under Python 3 and would
+  result in cryptic "maximum recursion depth exceeded while calling a Python object" error on
+  connection failure.
+
+  NOTE: This issue only affected installations using Python 3. (bug fix) #4382 #4384
```
armab:
```suggestion
  NOTE: This issue only affected installations using Python 3. (bug fix) #4832 #4834
```

PR #4835, opened 2019-12-16
Update pymongo to the latest version

Kami:
This pull request updates pymongo to the latest version.

Issue which has prevent us from upgrading in the past (https://jira.mongodb.org/browse/PYTHON-1966) has been resolved in the latest release.

NOTE: Some of the SSL / TLS related constructor arguments have been renamed, but the old names still work so I skipped renaming those arguments for now - https://api.mongodb.com/python/current/changelog.html#changes-in-version-3-9-0.

### DIFF
--- a/fixed-requirements.txt
+++ b/fixed-requirements.txt
@@ -18,7 +18,7 @@ requests[security]==2.22.0
 apscheduler==3.6.3
 gitpython==2.1.11
 jsonschema==2.6.0
-pymongo==3.7.2
+pymongo==3.10.0
 mongoengine==0.18.2
 passlib==1.7.1
 lockfile==0.12.2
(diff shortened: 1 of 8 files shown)


fixed-requirements.txt
```diff
@@ -18,7 +18,7 @@ requests[security]==2.22.0
 apscheduler==3.6.3
 gitpython==2.1.11
 jsonschema==2.6.0
-pymongo==3.7.2
+pymongo==3.10.0
```
armab: As I remember `requirements.txt` needs to be generated too based on this

Kami: Yep, need to add generated files, was waiting on tests to pass first :)

PR #4837, opened 2019-12-19
[WIP] [DONT MERGE] [EXPERIMENT] Use more efficient mongoengine fork

Kami:
This is just a quick experiment to see if we could use a more efficient fork of mongoengine (https://github.com/closeio/mongoengine) which reduces a lot of conversion overhead which is involved in regular mongoengine (which results in very slow operations with large documents).

The problem is that even though a lot of similar projects claim almost full mongoengine API compatibility, that doesn't really help us much since our DB abstraction is based heavily around mongoengine and also utilizes various low level APIs and functionality.

In addition to that, this fork is also based around older version of mongoengine.

So this means there will likely be no quick and easy wins for us. If that doesn't work out, probably the most reasonable approach is to try to incrementally optimize hot code paths (especially the ones which workflow results are saved and retrieved) one by one and use pymongo directly in those code paths.

### DIFF
--- a/fixed-requirements.txt
+++ b/fixed-requirements.txt
@@ -18,7 +18,7 @@ requests[security]==2.22.0
 apscheduler==3.6.3
 gitpython==2.1.11
 jsonschema==2.6.0
-git+https://github.com/closeio/mongoengine.git@master#egg=mongoengine
+git+https://github.com/StackStorm/mongoengine.git@master#egg=st2_patched_mallard
 pymongo==3.10.0
 passlib==1.7.1
 lockfile==0.12.2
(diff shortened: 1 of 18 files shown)


Kami:
I managed to get a lot of unit tests to pass, but there are still many failures (mostly in st2api) related to various compatibility issues.

So not sure how much sense it makes to pursue this further.

Even if we get all the tests to pass, we would likely need to run that code for at least a month on our CI/CD server before we could decide it's safe to release a beta version with those changes (since there are likely more edge cases hiding in there).

My other concern is that this fork uses quite an old version of mongoengine and has diverged quite a bit from latest versions of mongoengine.

Kami:
Superseded by #4846.

PR #4839, opened 2019-12-23
WIP - Better support for working with config objects from pack configs

zthart:
Hey all, wanted to get this change in front of some eyeballs here to get some feedback.

We've (@punkrokk) had the need in many of our stackstorm packs to configure them such that they can be run against multiple instances or tenants of the service/API they're built against. Our sensors have typically be written to iterate over all of the tenant configurations in our pack configs, but actions tend to be run against only one tenant, and thus take a tenant name as a parameter in the action at runtime. 

Typically our pack config schema looks similar to this:

```yaml
---
service:
  type: "object"
  required: true
  patternProperties:
    "^\\w+":
      "$ref": "#/properties/credentials"
  additionalProperties: false
credentials:
  type: "object"
  properties:
    username:
      type: "string"
    password:
      type: "string"
      secret: true
```

Which results in a config file like:

```yaml
---
service:
  instance_one:
    username: {{ st2kv.service.instance_one.username }}
    password: {{ st2kv.service.instance_one.password | decrypt_kv }}
  instance_two:
    username: {{ ... }}
    password: {{ ... }}
...
```

We've been taking "credential names" or tenant names as parameters into our action, so that if an action should be run against `instance_one`, the parameter passed into the action would be `tenant=instance_one`. This has, however, resulted in a lot of reused boilerplate across all of our packs that support this kind of multi-tenancy for fetching all tenants, ensuring the provided tenant exists, and constructing the client object in our `lib` code for the requested tenant. We wanted to provide a simpler way to fetch entire objects of configuration details from the pack config, so we started first looking at the `PackConfigDict`. The simple implementation here was to just run a YAQL expression against the pack config to fetch the instance configuration, if present. This seemed pretty intuitive to us, since the query string could be parameterized to take in the instance name (since we have control over the pack schema against which the query is run, we're not worried about things moving around without us knowing), and we'd fetch the entire config object in a couple of lines plus some error handling, and send the config parameters to our client constructor in far less duplicated code across packs than before.

We wanted to get this up to get some feedback on it to see if we're on-base with where a change like this should exist, if it's enough of a change, or if there would be some consideration for changing pack schema to support the idea of a "tenant" object more natively. We posted some discussion about this in the community slack, but wanted to get the change more formalized here. 

(We also have [a variant on this](https://github.com/syncurity-exchange/st2/pull/2) that avoids the performance overhead of YAQL, but is a very naive solution - the idea could be fleshed out to be a more proper search on the dictionary for a key without a YAQL query, though - see https://github.com/syncurity-exchange/st2/pull/2)


m4dcoder:
Just so that we are on the same page, can you expand on your example and illustrate how the YAQL expression(s) will be like and how it will pass appropriate instance_one/two username and password?

punkrokk:
@zthart

zthart:
Sorry for being away from this for a while, here's some more details on the example.

Given all the configurations above, say we have an action whose configuration looks something like

```yaml
---

name: "do_something"
runner_type: "python-script"
description: "Provided a tenant, perform an action"
enabled: true
entry_point: "do_something.py"
parameters:
  tenant:
    type: "string"
    description: "Name of the desired tenant for which this action should be run"
    required: true
  thing:
    type: "string"
    description: "Thing to do "
    required: true
```

In a pack with many actions, we'd abstract our tenant selection into a base action class that might look like so:

```python
from client_lib import Client  # Example client we want to construct

class BaseThingAction(Action):
    def __init__(self, config):
        super(BaseThingAction, self).__init__(config=config)
        self.client = None

    def get_client(tenant_name):
        if self.client is None:
            # Query from config (None if tenant_name is not present)
            params = self.config.query('$.service.%s' % tenant_name)

            if params is None:
                raise RuntimeError('Tenant %s is invalid!' % tenant_name)

            # Let's make an assumption that our config file params match exactly our 
            # Client's __init__ function params
            self.client = Client(**params)
        else:
            return self.client
```

Then, in our child action class

```python
from base import BaseThingAction  # our base action

class DoSomethingAction(BaseThingAction):
    def run(self, tenant, thing):
        try:
            client = self.get_client(tenant)
        except RuntimeError as e:
            return False, str(e)

        result = client.do_something(thing)
        # Any validation of result happens here
        return True, result
```

Let me know if this clears things up @m4dcoder @punkrokk

m4dcoder:
Looking at your example, do you see use cases where you will require more complex YAQL queries?  We already have similar methods for accessing nested dicts using dot notation.  Examples are at https://github.com/StackStorm/st2/blob/master/st2client/st2client/utils/jsutil.py.  You may find similar methods under st2common/util.  If not, copy the methods in jsutil.py over to st2common/util.

punkrokk:
@m4dcoder So I just dug into this on behalf of @zthart and I. 

Here is how we are currently using nested config stanzas:

```

from cisco_amp import CiscoAMP

__all__ = [
    'BaseAmpAction'
]


class BaseAmpAction(Action):
    def __init__(self, config):
        super(BaseAmpAction, self).__init__(config=config)

        self.tenants = []
        self.instances = {}
        self.hosts = {}
        credentials = self.config.get('cisco_amp')

        for tenant, creds in credentials.items():
            plugin_user = creds.get('plugin_user', None)
            plugin_pass = creds.get('plugin_pass', None)
            target_host = creds.get('target_host', None)

            self.instances[tenant] = CiscoAMP(plugin_user, plugin_pass, target_host,
                                              logger=self.logger)
            self.hosts[tenant] = target_host

            self.tenants.append(tenant)
```

And our action logic for `using` `tenant.cred`:

```
    def run(self, connector_hostname, domain_name, tenant):

        if tenant not in self.tenants:
            response = {
                "response": "Tenant not found.  Must be one of: " + str(self.tenants)
            }
            return False, response
```

@zthart @m4dcoder I'm trying to remember why we didn't just go with this process in the first place. Is this approach good enough for the desire?

m4dcoder:
@punkrokk @zthart, If the use case here is to just access nested config either traversing the dict or with a dot notation, there's no need to implement YAQL.  I'll let you guys hash this out on your own.

punkrokk:
@m4dcoder Thanks for your input here. I'm going to close this for now, and circle back if needed.

PR #4841, opened 2020-01-12
New example for an Orquesta workflow

winem:
This is an example for an Orquesta workflow using the core.ask pack to query multiple parameters. I will also create a PR in the st2docs repository to include this into the documentation.


winem:
As far as I can tell the failed Travis CI build is not caused by the new example and rather an issue with the infrastructure. Please let me know if I'm wrong and there is anything I can do to fix the issue.

contrib/examples/actions/orquesta-ask-multiple-parameters.yaml
```diff
@@ -0,0 +1,7 @@
+---
+name: orquestar_ask_multiple_parameters
```
armab:
```suggestion
name: orquesta_ask_multiple_parameters
```

armab:
@winem Thanks for the contribution! This looks like a useful example :+1: 


@m4dcoder Can you please take a look if that workflow looks good to you or there are any recommendations?

contrib/examples/actions/orquesta-ask-multiple-parameters.yaml
winem: Fixed.

contrib/examples/actions/workflows/orquesta-ask-multiple-parameters.yaml
```diff
@@ -0,0 +1,56 @@
...
+            description: "Continue?"
+            required: True
+          department_id:
+            type: number
+            description: "Your department ID:"
+            required: True
+          extra_output:
+            type: string
+            description: "Your message to echo next if you approve to continue:"
+            default: "None"
```
m4dcoder: This is minor. But you can just use `None` as in NoneType instead of literal `"None"` which is a string.

contrib/examples/actions/workflows/orquesta-ask-multiple-parameters.yaml
```diff
@@ -0,0 +1,56 @@
...
+      ttl: 60
+    next:
+      - when: <% task(get_approval).result.response.extra_output != 'None' and task(get_approval).result.response.approved = true %>
```
m4dcoder: If you use NoneType, I believe the check would be `task(get_approval).result.response.extra_output != null`

contrib/examples/actions/workflows/orquesta-ask-multiple-parameters.yaml
```diff
@@ -0,0 +1,56 @@
...
+        publish: 
+          - custom_output: <% task(get_approval).result.response.extra_output %>
+          - approver_department_id: <% task(get_approval).result.response.department_id %>
+        do: echo_extra_message
+      - when: <% task(get_approval).result.response.approved = true %>
```
m4dcoder: This condition will be triggered regardless if extra_output is None.  So in the case where extra_output is not None, both the `echo_extra_message` and `finish` tasks will be triggered simultaneously.  Is this your intention?  If not, then the condition here should be `<% task(get_approval).result.response.extra_output = 'None' and task(get_approval).result.response.approved = true %>`

contrib/examples/actions/workflows/orquesta-ask-multiple-parameters.yaml
winem: I will change that as I actually prefer it. I just didn't know from the docs if that's supported, thanks.

contrib/examples/actions/workflows/orquesta-ask-multiple-parameters.yaml
winem: I will test this latest tomorrow and come back with a working solution.

contrib/examples/actions/workflows/orquesta-ask-multiple-parameters.yaml
winem: Ok, I may have a misunderstanding here. So, my assumption was that the different when conditions are evaluated in their order. As soon as one condition matches the evaluation of the remaining ones will be skipped. 

So my idea was: 
- if we have the approval AND a custom message -> echo_extra_message which will trigger the finishing action (the remaining two conditions won't be checked)
- if stackstorm proceeds to the 2nd condition we know that we got the approval but no extra message (otherwise the condition above would have matched)

Is this assumption wrong? I'll also do some more tests tomorrow to see what I missed during my recent tests.

m4dcoder: Yes, your assumption here is not accurate. The transition is evaluated in order but will be triggered if the condition is met. So in this example, multiple task transitions can be triggered.

winem: Ok, got it. Somehow, I get another error when I do not provide any extra_output: 
```      "message": "YaqlEvaluationException: Unable to resolve key 'extra_output' in expression '<% task(get_approval).result.response.extra_output != 'None' and task(get_approval).result.response.approved = true %>' from context.",```

This makes sense because the response sent by the browser contains just `"response":{"approved":true,"department_id":41411}`. So the default is not part of the response object. 

I also tried `extra_output in task(get_approval.result.response` but this seems not to be supported. I will check after lunch how to proceed here. Might be a YAQL expression that's needed here.

winem: Ok, was able to reproduce the issue and see the behaviour you described. PR will be updated shortly.

contrib/examples/actions/workflows/orquesta-ask-multiple-parameters.yaml
winem: This test was not successful. Only number, string and booleans seem to be supported. Used .containsKey() for this check.

winem: This test failed. Only boolean, string and names are supported types. I used .containsKey() for this check in the updated code.

contrib/examples/actions/workflows/orquesta-ask-multiple-parameters.yaml
winem: This failed but I'm not sure if this is intended or a bug. 

`extra_output` does not appear in the response object (e.g. `"response":{"approved":true,"department_id":41411}`) even if a default is defined. So any check on `task(get_approval).result.response.extra_output` fails with `"YaqlEvaluationException: Unable to resolve key 'extra_output' in expression ...`

PR #4843, opened 2020-01-15
fix transitive dep on more-itertools for python 2.7

cognifloyd:
Without this change I could not successfully run `make requirements` or `make lint` on python 2.7.

Something is pulling in more-itertools, but the metadata for more-itertools is confusing pip so it is selecting a version that is python 3 only when it should be installing more-itertools 5.0.0 or less on python 2.7.

This takes the shotgun approach to fixing that. It includes the dep for all st2 packages. If anyone cares to evaluate the tree of deps to only include this where needed, be my guest.

Note that this includes a fix for the fixate-requirements script to allow including the python_version marker.

### DIFF
--- a/st2tests/requirements.txt
+++ b/st2tests/requirements.txt
@@ -7,6 +7,7 @@
 # update the component requirements.txt
 RandomWords
 mock==2.0.0
+more-itertools<=5.0.0 ; python_version < "3"
 nose
 nose-parallel==0.3.1
 nose-timer==0.7.5
--- a/fixed-requirements.txt
+++ b/fixed-requirements.txt
@@ -46,6 +46,8 @@ webob==1.8.5
 flex==6.14.0
 prance==0.9.0
 pywinrm==0.3.0
+# transitive-dep fixing for py2
+more-itertools<=5.0.0 ; python_version < '3'
 # test requirements below
 nose-timer==0.7.5
 nose-parallel==0.3.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,6 +28,7 @@ kombu==4.6.6
 lockfile==0.12.2
 mock==2.0.0
 mongoengine==0.18.2
+more-itertools<=5.0.0 ; python_version < "3"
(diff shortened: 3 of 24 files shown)


armab:
Cool, thanks for the efforts!

@Kami might have more ideas as we discussed it recently.

st2tests/requirements.txt
```diff
@@ -7,6 +7,7 @@
 # update the component requirements.txt
 RandomWords
 mock==2.0.0
+more-itertools<=5.0.0; python_version < "3"
```
Kami: IIRC, for that notation, a more recent version of distutils / setuptools needs to be available during package create and install time, right?

Kami:
I'm fine with those changes as long as they don't break out package build process (I believe they shouldn't as long as a more recent version of setuptools and distutils is used there).

cognifloyd:
I just updated the commits to use ` ; ` instead of just `; ` (space before and after) because I came across  an [odd bug](https://github.com/pypa/setuptools/issues/1770) that suggested the space was required in some cases.

cognifloyd:
As far as setuptools versioning:
We are installing setuptools-41.0.1 as part of the dev environment.
I don't see any current setuptools or distutils version requirements.

setuptools-20.2 added [PEP 508](https://www.python.org/dev/peps/pep-0508) environment marker support, and had many fixes through 20.10 (see [here](https://setuptools.readthedocs.io/en/latest/history.html))
CentOS 7 comes with 0.9.8
CentOS 6 comes with 0.6.10
Ubuntu 14.04 comes with 3.3
Ubuntu 16.04 comes with 20.7
Ubuntu 18.04 comes with 39.0
I'm not sure where to check which version is used during the package build process.

I'm not sure how to check the version of distutils. Isn't that part of stdlib?

cognifloyd:
It looks like the packaging docker containers are installing setuptools, but no version is pinned.
https://github.com/StackStorm/st2packaging-dockerfiles/tree/master/packagingbuild

PR #4845, opened 2020-01-20
Fix the amqp connection for WorkflowExecutionHandler to pass the SSL params

mtatsuma:
To pass the SSL params, get_connection() should be used instead.

Related commit: dd4465a266c11e2dfd2fcf97896562ff8a9e8605

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -105,6 +105,9 @@ Fixed
   NOTE: This issue only affected installations using Python 3. (bug fix) #4832 #4834
 
   Reported by @alexku7.
+* Fix the amqp connection setup for WorkflowExecutionHandler to pass SSL params. (bug fix) #4845
+
+  Contributed by Tatsuma Matsuki (@mtatsuma)
 
 3.1.0 - June 27, 2019
 ---------------------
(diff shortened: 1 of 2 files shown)


Kami:
Thanks for the contribution and good catch.

The fix indeed looks correct :+1:

CHANGELOG.rst
```diff
@@ -104,6 +104,7 @@ Fixed
   NOTE: This issue only affected installations using Python 3. (bug fix) #4832 #4834
 
   Reported by @alexku7.
+* Fix the amqp connection setup for WorkflowExecutionHandler to pass SSL params.
```
armab: Please include PR number as well for the fix to keep it in history.
If you want, you're very welcome to also include your name as a contributor for the fix.

See other Changelog records as example.

mtatsuma: I added it.

armab: :+1:

mtatsuma:
I rebased and pushed again to pass the checks!

blag:
Based on the linked commit, this looks like a replacement that Kami missed in the original commit.

Since he's also already approved it previously, and it passes CI, I'm just going to merge it in even though technically it should get a re-review due to the rebase.

PR #4846, opened 2020-01-20
Optimize storage (serialization and de-serilization) of very large dictionaries inside MongoDB

Kami:
This pull request builds on top of various research I did in other PRs and issues in the past (#4837, #4838, etc).

The goal is to speed up serialization and de-serialization of very large dictionaries (e.g.large results produced by various actions) inside MongoDB.

Right now we store those dictionaries as native MongoDB dictionaries, but this means we need to escape the keys, since they can't contain ``$`` and ``.`` which are special characters.

This escaping is inefficient and wasteful with large (and nested) dictionaries.

## Proposed Implementation / Improvement

This pull request tries to optimize that by serializing those dictionaries as JSON instead.

Keep in mind that this is fine, since we always treat those values as opaque strings anyway.

In fact, a future improvement would be to have a result database object per runner, this way we could utilize a simple "string" type for the actual result, but that's a much more involved change.

Some actual numbers with synthetic dataset are available in other issues mentioned above. Actual numbers with production real-life data are to follow once we deploy this to CI/CD server.

The change is fully backward compatible, but it's behind a feature flag (opt-in) until we are really sure after running it in production for a while with diverse datasets it doesn't make some corner / edge case worse or similar.

Related optimization I talked about to @m4dcoder is getting rid of using ``EscapedDict`` DB field type in scenarios where keys can't contain special characters (e.g. potentially various workflow related models).

## TODO

- [x] Update all the affected models to use this field type
- [x] Tests
- [x] Deploy it on CI/CD for a month, observe performance, memory usage, etc
- [x] Add optional migration script for affected models - future
- [x] Document change in upgrade notes and point users to the optional migration script - future
- [x] Add some docs for developers on the new DB format
- [x] Update execution list / get command to display execution run duration which includes database write time not just action run time
- [x] Document the change


st2common/st2common/fields.py
```diff
@@ -114,3 +123,95 @@ def to_mongo(self, value):
...
+    and EscapedDynamicField format and the new one.
+
+    On retrieval, if an old format is detected it's correctly un-serialized and on insertion, we
+    always insert data in a new format.
+    """
+
+    def to_mongo(self, value, use_db_field=True, fields=None):
+        if not cfg.CONF.database.use_json_dict_field:
+            value = mongoescape.escape_chars(value)
+            return super(stormbase.EscapedDynamicField, self).to_mongo(
```
m4dcoder: This bugs out because JSONDictEscapedFieldCompatibilityField is incompatible with EscapedDynamicField.

st2common/st2common/models/api/execution.py
```diff
@@ -152,6 +152,10 @@ class ActionExecutionAPI(BaseAPI):
     @classmethod
     def from_model(cls, model, mask_secrets=False):
         doc = cls._from_model(model, mask_secrets=mask_secrets)
+
+        import json
+        doc['result'] = json.loads(doc['result'])
```
m4dcoder: I have to add this to get the result in JSONDictEscapedFieldCompatibilityField with ujson/cjson backend to work properly.  This is because _from_model will call to_mongo which dumps the result in JSON to string.  The CLI output will be displayed as a string.

guzzijones:
I have taken a look at this. +1 from me.  Saving the result as a string instead of a mongo object/dict is ideal.  There is no current use case to index nested dicts from workflow transitions while running.

Kami:
After some discussions on Slack, I decided to try to attack this problem again and wrap up this change - slow insertion of large executions problem (which is fairly common in StackStorm world) still hasn't gone anyway and it's very much alive as it was 3 years ago.

I already performed a lot of benchmarkinh, research and prototyping in the past, but since quite a lot of time has passed since then (and data is scattered across various different issues and PRs), I decided to start from scratch with micro benchmarks which compare using different approach for storing ``LiveActionDB.result`` field value which is the main offender in our case.

Good thing about micro benchmarks is that it's also very easy to add a new one or use a different fixture file, etc. to see how it affects the numbers (and they are also fully reproducible by everyone).

First to refresh everyone's memory (including mine).

The two issues which contribute to slow insertion / saving of executions with large results:

1) mongoengine inefficency when converting large dictionaries to native pymongo SON types.
2) Escaping in our ``EscapedDictField`` which is performed on every single key in a dictionary (we escape ``$``, ``.`` - this is needed if you want to store dictionaries which keys contain those values which are special in MongoDB since they act as operators and separators).

And one very important invariant on the result field itself - we use dictionary type for execution result field, but for the purpose of database itself, that field is an opaque binary blob - we don't perform any querying on it's values.

That's very important because it allows us to treat it as a string / binary blog which makes things easier (in fact, one of the option I already debated in the past is storing result in a separate collection as a string value and then add a FK to LiveAction collection).

## Proposed Approach

The proposed change to mitigate / fix this issue is still the same one as I proposed and researched more than a year ago - storing result as a dictionary serialized to a JSON string / blog. Only difference is that now I use ``orjson`` instead of ``ujson`` since micro benchmark shows it's 5-15% faster than ``ujson`` on this specific dataset.

A lot of other approaches were also brought up in the past - things such as removing and replacing mongo all together, using a different ORM library, etc. 

Yes, many of those would likely result in a much better performance if they were executed and implemented correctly (if!), but most of them are not even close to realistic - they are much more risky (much larger change, harder to make it backward compatible, etc.) and simply not realistic since they would require many man month hours worth of work (and that's an ideal scenario, less ideal and more realistic scenario where various edge cases crop up, compatibility issues, etc. is closer to 6 months).

This means that this approach is one of the few possible realistic approaches we have left.

This approach exploits the invariant that ``result`` is for all purposes, opaque binary blob to the database itself and if we use a binary field, we don't need to run escaping code on it + mongoengine doesn't need to perform slow conversion.

## Initial Results

Initial results from micro benchmarks are very promising.

### Save

Screenshot from 2021-02-18 21-48-49

### Read

Screenshot from 2021-02-18 21-46-32

For a very large execution (8 MB), we see almost ~10x improvement in insertion speed (5s vs 500ms) and ~5x improvement in read speed (800ms vs 150ms).

The good news is that this change also scales down - aka it's not slower for smaller results - in fact it's also a bit faster for those scenarios as well.

Those numbers are from running micro benchmarks on my computer. They also run on continuously CI and everyone who wants to run it can easily run them themselves and also easily add a different dataset (e.g. some real life execution result data).

## Things to experiment with

I still want to experiment with compressing actual json blob using ``zstandard``.

The goal here is not to save number of bytes on disk (those were never really a big issue), but to save number of bytes which need to be sent over the network when talking to the database - this could perhaps help and reduce overall operation time even more for very large executions, but I doubt it will be worth the effort in the end. Will see what data will show us.

## TODO

- [ ] Also add micro benchmark just for the dict key escaping code
- [ ] Add micro benchmark which include zstandard compression
- [ ] Finish up the code
- [ ] Add more backward and forward compatibility tests

Kami:
I added micro benchmarks just for the escaping and unescape function and also unit tests for the new fields.

Running existing tests did uncover some issues - in some places, we run a set query for updating a live action and we pass dict for the value (which also kinda brakes the our ORM layer / DB abstraction and shows why such changes are hard - no abstraction is perfect and there are always leaks).

Sadly this won't out of the box, I will need to think and experiment some more on how to best approach this.

Technically, I think should be able to make it work, as long as I add some additional code to the places which use ``field__set`` notation to make sure the value we pass is already serialized as JSON instead of the raw dict - sadly ``to_mongo`` is not called of set queries.

Kami:
For now, I'm leaning towards adding a header to the serialized field value even though we will likely just support single serialization library and format (``orjson``) to make it a bit more future-proof in case we ever need to change the serialization format.

Something along the lines of ``b"<compression algorithm (if any):><serialization format>:<serialized data>``. 

So the value would look like something like this: ``n:or:{"foo": "bar"}`` where ``o`` stands for ``orjson`` and ``n`` for no compression (and ``z`` could be zstandard).

This adds some overhead, but it's really just 4 bytes so it's negligible (especially if we ever support compression).

m4dcoder:
Can we be opinionated here and support a specific method of serialization? There's too much options to support (i.e. orjson, ujson, cjson) and I rather json be json w/o any prefix that only st2 understands.

blag:
100% agree with @m4dcoder here.

If we keep it as just JSON then we can read it with other tools if/when things go awry (I accidentally created a forkbomb/loopbomb in ST2 last night and had to clean up RabbitMQ and MongoDB manually because browser tabs were crashing st2web when trying to load executions with large results).

If we slap our own header on it, then it's no longer just JSON, and we need to develop all of our own tools to work with it. I'd rather we not burden ourselves with that.

Kami:
@m4dcoder @blag 

> Can we be opinionated here and support a specific method of serialization? There's too much options to support (i.e. orjson, ujson, cjson) and I rather json be json w/o any prefix that only st2 understands.

Yeah, those options were originally added a while ago before we were Python 3 only. Now we are Python 3 only and based on micro benchmarks, ``orjson`` is faster than all the alternatives (and also more compliant) so there is no need to support different JSON backends / libraries.

As far as the general header or not goes, I'm still debating it. I do see the value in raw JSON (can be parsed by other tools, etc.), but then again the whole db structures should be opaque to the end user and the whole point of ORM abstraction (aka our DB models) is to hide those implementation details away.

One reason why I think having a header would be good is forward compatibility and also ability to add compression - in situations with very large results (thinks many MBs to many 10s of MBs - not unheard of), compression would save a lot of space. Although you could argue that compression should be done on the whole DB / collection level (which I believe Mongo support).

Having said that - for now, I will try to get all the tests to pass with the current approach (no header, plain JSON) and then think about the header issue again.

I did make good progress on tests, but some are still failing - those are mostly the ones related to the code which is not using DB layer / models correctly :/

guzzijones:
I say just zstd compressed json. We need the compression for large objects.

Kami:
Good news - I now have all API unit tests passing :tada: :champagne: 

It's still far from being finished though, but it's as far as we've been on this issue - we will of course need more end to end testing, especially with existing installations to make sure all the edge cases are handled transparently.

While trying to get tests to pass, I did encounter various abstraction leaks which made my life harder. 

I needed to add a "one off" to base ``API.from_model()`` method to handle this new field type. The approach is not great, but I don't have many options to make it better.

That whole code is somewhat a mystery to me - there really shouldn't be a need to call ``unescape_chars`` on the whole document, we would only need to do it in fields which utilize ``EscapedDynamicField`` / ``EscapedDictField``. Likely we never encoutered this issue because we utilize those fields in most places.

Maybe the reason why we do this "shoot and forget" approach and call unescape on the whole document is because API layer is not directly aware of underlying DB model field types and that was the easiest approach.

But as mentioned before - escaping and unescaping is slow! Especially for large documents / nested dicts since we need to recursively process them.

Technically I could probably refactor it to make it a better more generic, but that would open another mine field and can of worms I want to avoid. At this point, I just want to make everything work correctly (with tests, of course) and in the future we can think about improving some of those things.

For now, I only updated execution and live action ``result`` field to use this new field type since those are the biggest "offenders", but there is really not reason we couldn't use it for other fields which can be large dictionaries and we don't query on (e.g. perhaps parameters and context fields).

In ideal world, if there were no abstraction leaks this should simply mean just changing the field type, but as mentioned before, we do have a couple of abstraction leaks which means we will need to update a bit more code (mostly API

P.S. We also had some tests which passed string value for the dict ``result`` field and that somewhat magically worked. I updated them to use a dictionary instead which is the correct thing to do.

Kami:
@guzzijones We first need more data to decide if compression makes sense - it will definitely make things slower, the question is by how much and at what point (result size) it starts to make sense.

And if we do plan to support compression, I think it should be a config option (disabled by default) which can be enabled by the users for all the JSONDict fields.

In that case, we do need to use some kind of header for the stored binary field value.

Yes, raw JSON is probably the nicest and easiest to work with, but we should make it more future proof and this is needed if we plan to support optional compression.

As mentioned above - if we didn't use any kind of ORM + our DB models layers and we directly worked with raw data from (py)mongo, then yes, I would strongly be against using this more complex field value serialization scheme.

But since we already do, it shouldn't matter - only way anyone should really interact with database values is via DB models - there is simply no guarantee (and never was) things will work as expected if you start manually manipulating the documents since a lot of field types depend on specific validation, etc. logic before they are saved to the database.

Yes, things may work to some extent if you start to manually poking around, but I don't think that's something we should support and people who do that do it at their own risk.

In short, I don't think using more complex serialization scheme for ``JSONDictField`` is any different than using ``ComplexDateTimeField`` or any other existing field type which includes custom serialization / deserialization logic. That's the whole point of that fields abstraction. So either we should fully embrace it or get rid of it all together (not realistic).

/cc @blag @m4dcoder

Kami:
@guzzijones Also, did you ever considered enabling compression on Mongo (whole database / collection) level? That may give even better results as far as database size reduction goes.

guzzijones:
I agree on using a header. Agree on compression as optional and stored in
the header. Great work so far.

Kami:
So far I only posted results for more isolated micro benchmarks scenarios, but here are some results of everything end to end.

I ran a Python runner actions which loads 8 MB json fixture file in memory, parses it and returns parsed dictionary as an action result.

The difference is around 1 second for this PR and around 15+ for master (aka massive difference).

To be able to time everything end to end, I needed to add some additional instrumentation to the action runner. We can't really on ``start_timestamp`` and ``end_timestamp`` fields.

Those represent values when runner starts to run an action and when it finished to run it, but before the actual result is saved and available in the database (that operation itself isn't directly timed anywhere).

As the main bottle neck is storing large execution object in the database and not actual runner running the action, in both cases based on the ``start_timestamp`` and ``end_timestamp`` values, the actual Python runner action execution time is the same (sub second), but overall operation time to execution being available in the database is vastly different - ~1s vs ~15+s.

Same behavior can be observed / "measured" using ``run --async`` + ``execution tail`` polling approach (you will see execution will be in running state much longer with master version of the code).

So we should also think how we can improve that since current implementation will give incorrect and deceiving picture to the users and operators - running an action and using execution list to poll until it has finished / data is available in database for querying will take a long time even though duration based on start_timestamp and end_timestamp will be short.

I'm not sure how to best handle that.

One approach would be to insert another entry into ``execution__log`` field as an atomic operation (add single item to an array field) where the ``status`` value would be ``result_wrriten_to_db`` or similar.

Kami:
Re "execution full duration" issue. I added this change which should provide us with better visibility into that - 89617ff25.

As mentioned above - there are a couple of approaches possible, but I think this one is the most reasonable one. It also should involve very little overhead since it's two atomic operations on very small fields.

In reality, this change will likely be neede much less with improvements in this PR (since now the actual database writes will take much less time), but we should still have this data available.

Kami:
@m4dcoder My main goal at this point is to get all the tests passing + this PR finished, but once that's done we should probably also evaluate ``WorkflowExecutionDB`` model for similar improvements.

I don't use tons of Orquesta workflows, but from my quick check of the model, we use same slow and inefficient field type so as long as those fields can also store large dicts (which we don't need to directly query on), then they should exhibit the same performance issue and this fix should directly apply to them as well.

Kami:
I pushed a change which also implements approach with header which should make everything more flexible and future proof - 42f70e71e.

I made ability to turn header on or off with a constructor argument for testing and benchmarking purposes only, so we can see how much overhead using a header (+ compression) vs simple / naive approach adds. In real world, this option will never be exposed to the end user.

Here are some results from micro benchmarks:

Screenshot from 2021-02-19 23-59-19

Screenshot from 2021-02-19 23-59-40

It does add some overhead, but not tons (and in some scenarios it's even faster - probably because it results in sending less data to DB and even though compression takes CPU cycles, that can be offset by sending less data to mongodb).

Actual results of course very much depend on the result field value and how compressible that data is (that dictates how much work compression algorithm needs to do).

So in the end, it would probably depend on the end users datasets - for users with very large well compressible results, turning it on could results in large space savings by trading it for a slightly longer duration of the whole save / read operation.

m4dcoder:
@Kami 
- How does this (w/ headers) impact backward compatibility (i.e. upgrading in production from an earlier st2 version)?
- I prefer an opinionated approach than to try and please everyone with different options because the code ends up having too much branching and harder to maintain over time. Can we automatically compress if data is greater than specific size?

blag:
I still think supporting multiple serialization libraries is overkill. They should theoretically all serialize to the same format anyway, and the abstraction layer adds complexity into a critical component that doesn't strictly need the flexibility. I don't see a distinct benefit from supporting multiple serialization libraries. Probably all users will use the default option, and the only people who will switch are those that find better performance by switching to the fastest library. So, if StackStorm is opinionated, and switches over to using orjson and just orjson, I don't think many users will know or care.

If the concern is storage compression, MongoDB already supports that on collections:
https://www.mongodb.com/blog/post/new-compression-options-mongodb-30

> In MongoDB 3.0, WiredTiger provides three compression options for collections:
> 
> * No compression
> * Snappy (enabled by default) – very good compression, efficient use of resources
> * zlib (similar to gzip) – excellent compression, but more resource intensive

And [as of MongoDB 4.2, zstd compression was added](https://docs.mongodb.com/manual/core/wiredtiger/#compression).

The same options are available for network compression:
https://docs.mongodb.com/manual/reference/program/mongod/#cmdoption-mongod-networkmessagecompressors

Note that MongoDB versions 3.6 and 4.0 already enabled network compression by default, using the Snappy compression library.

Now, I'm not sure that StackStorm installs MongoDB with storage compression turned on, and if we don't install the Python Snappy package (which requires the `libsnappy` development headers and the `libsnappy` library from the Deb/RPM package), then the MongoDB library may be smart about it and disable network compression. I would be 100% in favor of turning those features on in the StackStorm bootstrap/one-line install script though!

I still strongly think we should delegate compression to MongoDB itself and to our Mongo libraries, and keep our database formats as simple as possible. I don't think the application layer is the right place to compress data. Leave that to the database.

So if we let the database handle compression for us, and switch to using orjson, then there's no reason to create our own not-quite-JSON format.

I'm somewhat fine with storing JSON-as-a-string (meaning just JSON, not header+JSON), but if we do that, then we lose out on the promise that what we read back will be parse-able JSON [I'm trying really hard to avoid the term "type safety" here]. That is a good benefit that storing JSON-as-JSON gives us. If we go that route, then we should make sure our code can gracefully handle JSON parsing issues, and _those_ changes could get a bit involved.

Kami:
I agree on the compression for data at rest - it's likely better to enable it on the whole database level (but more research needs to be done to see how it actually impacts things).

As far as transport level compression - this also needs to be benchmarked, etc. to ensure it gives expected results (aka making sure pymongo does compression in fast an efficient manner).

As far as header goes, I think we will likely need a tie breaker here :)

Overall, I'm with you on simple formats (that's also why I like HTTP 1 over HTTP 2 - simple text vs binary, etc.), but I also really don't see much drawback in using a header approach, only advantages. As mentioned above, if we ever want to change the format, having a header would make that much easier for us doing it in a backward and forward compatible manner.

Even without header, we could handle some changes transparently, but that would be less robust and slower (since we may need to deserialize the whole data to determine which format it's using).

> I'm somewhat fine with storing JSON-as-a-string (meaning just JSON, not header+JSON), but if we do that, then we lose out on the promise that what we read back will be parse-able JSON [I'm trying really hard to avoid the term "type safety" here]. That is a good benefit that storing JSON-as-JSON gives us. If we go that route, then we should make sure our code can gracefully handle JSON parsing issues, and those changes could get a bit involved.

Kami:
And just to do another quick summary on the backward compatibility of this change @m4dcoder asked about on Slack.

The change is fully backward compatible - the de-serialization code supports both formats (existing one and the new one). This is handled transparently when using the model and reading data from the database.

Once the model is written to the database, new format is used. This means the change will only affect new executions and any existing executions which are being updated.

This means that the ``result`` field for existing executions won't be updated. This is totally fine because most important things are new executions - those cause the most load on the system due to slow read / writes (when reading and writing them to the database).

Technically, we could also try to migrate existing data to the new format when reading executions from the database even if they are not being updated (think read-repair), but this really wouldn't get us much beyond additional code complexity, etc. (we don't have any good object versions and data migration story in place so that would require more code).

In addition to that, a lot of installations have garbage collection enabled for executions and other objects which makes not updating old executions even less of an issue.

This does mean the code will need to support both formats for probably quite a long time, but we would need to do that regardless and that code is not that complex.

Technically, in the future, we could still provide a CLI tool which migrates existing executions (reads the whole model + re-writes it to the database), but that's not a priority right now.

Kami:
Since all unit tests are passing now (:tada:) and that's about as much as I can do until 3.4.0 is out so I can also test with integration tests + deploy changes to ci/cd for more extensive real-life testing, I decided to also look at the workflow related models.

I identified the following fields which I believe could benefit from those changes. Really any dict field should benefit from it, but the large the data it stores, the more it benefits).

Only limitation for that field is that we can't directly query on the field value since data is stored in a binary format and dictionary can only contain simple types.

I believe both of those should already be true for most of those affected fields. And even if we needed to query on some field value, instead of querying directly on the dict attribute, we should make that field top-level model attribute on query that instead - we already do that in some places and that's much more performant.

1. ``WorkflowExecutionDB``

- spec
- graph
- context
state
output

2. ``TaskExecutionDB``

- result
- context
- task_spec

Good news is that I updated those locally, run some Orquesta workflow and everything seems to be working fine.

I also tested it end to end using a very simple Orquesta workflow which calls python runner action which returns a larger result (4 MB) and passes some data around (will push that action shortly).

Here are the results:

1. Master (no optimizations anywhere) - total workflow run duration: 35 seconds
2. Only ``ExecutionDB`` + ``LiveActionDB`` model optimizations, without workflow ones - total run duration: 26 seconds
3. ``ExecutionDB``, ``LiveActionDB``, ``WorkflowExecutionDB``, ``TaskExecutionDB`` models include those optimization - total run duration: 12 seconds

In short, yes, we really do also need those changes for workflow related models in case we also want to substantially speed up Orquesta workflows which pass large datasets around.

And as mentioned before and expected based on the much shorter duration - those optimizations don't just speed up everything, but they result in much lower CPU utilization for all the related service (action runner, workflow engine, api) when running the same workflows / actions and with the same amount of resources, they will be able to provide a better overall throughput (executions / second).

/cc @m4dcoder

Kami:
While testing all the various edge cases again, I found a very subtle bug which none of the existing tests detected.

mongoengine will only update field value in the db if it has detected that it has changed on the retrieved instance. It implements that using custom ``__set__`` method (https://github.com/MongoEngine/mongoengine/blob/master/mongoengine/base/fields.py#L140).

This means it will correctly detect if we do something like this:

```python
model_db.result = new_dict_result
...
```

But not something like:

```python
model_db.result["foo"] = "bar"
```

It won't detect that if we work with a plain dictionary since ``__set__`` on the parent field won't be called.

To be able to correctly handle this "partial dict value update" scenario as well, I will need to introduce a wrapper class for raw dict, which will always be returned by ``to_python()`` method from that field.

In fact, that's very similar to what existing ``DictField`` uses, it uses ``mongoengine.base.datastructures.BaseDict`` data structure which is a light wrapper around dict which signals changes to the model instance if there have been any dict updates.

This approach is not 100% ideal since it will add some overhead (which should be that bad unless we are not updating tons of dictionary keys in a tight), but it's basically the same approach mongoengine already uses.

Only way we could avoid doing that would be to always do "full dict updates", so something like:

```python
new_dict_result = model_db.result
new_dict_result["foo"] = "bar"
model_db.result = new_dict_result
...
```

This is of course not ideal since we would need to audit all of the code base. So for now, we will need to leave with this wrapper class.

I'm still researching if I can handle this differently with less code.

Kami:
I did more research and I think that for v1, the best approach is to only use this field type for fields which are more or less for all purposes already ``immutable`` - so fields such as the ``result`` field (which are also the primary "offenders").

This allows us to postpone implementing very complex field tracking logic (which would likely also require some changes to mongoengine as well).

Yes, it would be nice if we could also use this field type for other fields which can contain large values, but that can be researched again in the future.

nmaludy:
@Kami would it also make sense for the input/parameters field as well? Again, immutable and can contain large values.

Kami:
@nmaludy Yes, pretty much any large dict field would benefit from those changes as long as we take recent "immutable and updated at the end" finding and limitation into account.

I'm still prototyping and researching to see if I can get dict wrapper class and attribute class to work with reasonable amount of code and without (too many) changes to the mongoengine.

If I can't, then likely we will only want to use it for ``Execution``, ``LiveAction`` and ``TaskExecution`` ``result`` field and ``WorkflowExecution`` ``output`` field.

Those attributes already operate in that manner - they are immutable and we update them "at the end" (aka when execution has finished and no updates will be performed to that object anymore).

In any case, I will also make sure those limitations are very clearly and explicitly document in the code - with current "no access / changes tracking" implementation, that field would be a bad candidate for any dict field which can updated very frequently on an existing model.

Kami:
OK, good news, I managed to get dict state tracking to work with very little code changes (mostly with piggybacking on the existing mongoengine code).

This means we should be able to utilize this field for any document field which can store a large dictionary and we don't need to perform direct database queries on. Immutability is still a nice to have, but not a mandatory property anymore.

Same with the "update at the end" - this is not a mandatory property anymore since tracking dict changes allow us to only write the field to the database when updating an existing document in case it has changed (without state tracking, we would need to write it each time that's why "only updated at the end" property would be mandatory to avoid overhead of that).

Nice thing also is that all of that works in exactly the same manner as an existing mongoengine ``DictField`` field so there are no surprised or discrepancies there.

m4dcoder:
@Kami Thanks for the time and effort spent in this PR and the diligence in running tests and explaining the details here. Looking at the performance and resource utilization improvement, I say go for it.

The only concern I have is backward compatibility and your comments that we may have to support this in the code for some time. I don't understand why migration script is not an option here. 

If users want to take advantage, they have to upgrade to v3.5 and if they want to upgrade, they have to run the migration script to update existing data in the database to the new format. This won't be the first time that users have to run migration script in between version.

I want to emphasize here to avoid inflating the st2 codebase and we should strive to reduce the amount of options and branching for maintainability reasons.

Kami:
@m4dcoder Technically if we could ensure that all the services are stopped when migration script is running (aka nothing is reading those values), then a migration script would indeed be enough, but we don't really want to instruct users to stop everything while migration script is running (wouldn't work for most HA deployments).

So the code still needs to support both formats (at the very least after the upgrade until the migration script has finished), but as mentioned above, I'm not even a huge fan of a migration script approach.

There isn't much additional code we need to maintain to support both formats (because those field types we are replacing are still used for a bunch of other fields).

If anything, I would be more for something like "read-repair" approach or something which doesn't put additional burden on the operator like the migration script does.

In theory, we could have some kind of service or similar responsible for running migrations (something we have already discussed in the past), but we don't have tons of changes like that on the models so we never went implementing one (and as mentioned above, that wouldn't really change the premise that the code needs to support all formats until the migration has finished).

m4dcoder:
@Kami 

I look forward to the improvement from this change.

But just to be clear, I'm only OK with the change if we are moving forward and NOT providing a bunch of different configurable options to opt out and continue using the old json lib or old field types. If we move forward, then we move completely.

On HA deployment, I don't recall that we support rolling upgrade. How do users upgrade in HA deployment today? It is expected there are no executions running (no services running to be exact in the docs) during single node or multi-node upgrade. We documented on upgrades to run any required migration scripts in between versions. So are we changing our approach here?

> There isn't much additional code we need to maintain to support both formats (because those field types we are replacing are still used for a bunch of other fields).

Can you identify these bunch of other fields that will continue to use old field types? Why wouldn't we move them to the new field type?

Kami:
@m4dcoder 

> But just to be clear, I'm only OK with the change if we are NOT providing a bunch of different configurable options to opt out and continue using the old json lib or old field types. If we move forward, then we move completely.

Yes, I'm 100% with you on that - we won't provide any user-visible config options for that - the options we expose right now are internal and only used for benchmarking and testing purposes (so we can compare and evaluate different approaches).

Only option we may provide at some point in the future is per-field level compression, if we ever reach a consensus on that in the future that is.

> On HA deployment, I don't recall that we support rolling upgrade. How do users upgrade in HA deployment today? It is expected there are no executions running (no services running to be exact in the docs) during single node or multi-node upgrade. We documented on upgrades to run any required migration scripts in between versions. So are we changing our approach here?

Yeah, that's a long known "issue" / limitation with the migration script approach.

Only way we could do it without visible end-user impact would be to instruct the users to stop all the services and only start them back up once the migration completes. That's of course not something we want to tell our users though (and downtime is unacceptable).

In the past we didn't really treat migrations as seriously as we should and most of them were on a much smaller scale - users could have a lot of executions (+ trigger instances + workflows) with very large result and other fields which means actual migration could easily take hours.

So in theory, if it was a smaller / shorter migration, we could probably look away and say "stop all the services" (that's kinda what we did in the past), but I don't really think that's a good answer.

> Can you identify these bunch of other fields that will continue to use old field types? Why wouldn't we move them to the new field type?

Pretty much all other values which are dictionaries which utilize ``DictField`` and ``DynamicField`` - there are many of those - parameters, context, runner_info, etc.

Technically, we could indeed utilize this new type for all the fields which are dictionaries on which we don't perform direct queries on (which are majority of them, IIRC, we do perform queries on some dict field values, but those are far and between).

Initial benchmarks showed the performance improvements also scales down to small fields (at least it's not slower), but if we do decide to do that, I do want to add more benchmarks for small field types.

armab:
@m4dcoder The K8s upgrade for StackStorm HA is a rolling upgrade when the new image for each st2 service is deployed replacing the previous one, while traffic switched from the old to new via load-balancer leading to node replacement with zero downtime.
Docker deployment is simpler, but it's somewhat similar. We don't have mechanisms for automated scripted upgrades in any other deployment methods as well or any framework around that.

So yes, avoiding manual DB or Scripted migrations would be preferred to follow the cloud-native way when possible. I think that makes the user experience better as well.

m4dcoder:
> Yeah, that's a long known "issue" / limitation with the migration script approach.
Only way we could do it without visible end-user impact would be to instruct the users to stop all the services and only start them back up once the migration completes. That's of course not something we want to tell our users though (and downtime is unacceptable).
In the past we didn't really treat migrations as seriously as we should and most of them were on a much smaller scale - users could have a lot of executions (+ trigger instances + workflows) with very large result and other fields which means actual migration could easily take hours.
So in theory, if it was a smaller / shorter migration, we could probably look away and say "stop all the services" (that's kinda what we did in the past), but I don't really think that's a good answer.

I agree we should strive to minimize impact on upgrade. But I don't think we'll be serious with supporting rolling upgrades any time soon either because proper support requires platform level feature such as triggering a maintenance on action runners and draining action executions before proper shut down and similar to other st2 components.  Given this and that not everyone is running HA, I think keeping with the current upgrade approach with migration script is acceptable. So as you've mentioned above, please still provide a migration script. Ideally, we'll just migrate data to the new format and we don't need to keep extra code to do this read-repair scenario.  But let's not keep debating. I'm OK if migration script is provided and users who want to migrate during upgrade can do that. Thanks.

Kami:
@m4dcoder I assume you typed your comment before seeing @armab's comment.

But yeah, I do still plan to provide a migration script, but I don't plan to document it as a mandatory step, just something operator can run if they wish to to convert old format to the new one (with a note that it may take a long time depending on their environment).

m4dcoder:
> Technically, we could indeed utilize this new type for all the fields which are dictionaries on which we don't perform direct queries on (which are majority of them, IIRC, we do perform queries on some dict field values, but those are far and between).
Initial benchmarks showed the performance improvements also scales down to small fields (at least it's not slower), but if we do decide to do that, I do want to add more benchmarks for small field types.

@Kami If there's no degradation, my vote is to move to the new types instead of supporting both types.  Yeah, I type my comment before I saw Eugene's.

@armab If rolling upgrades is already supported, then move forward with the read-repair approach. I'm curious how the rolling upgrades work.  Does the action runners know to drain action executions that it is running before shut down?

Kami:
@m4dcoder Yeah, there should be no degradation, only change is that data in DB is now stored in a binary format.

This means it's not as easily inspectable / modified using mongo shell, but asdescribed above, this is really an implementation detail and official way for interacting with data is using DB models (aka to the end user the experience will be exactly the same as before).

And another benefit this new field adds is that it's also much more space efficient - DynamicField and DictField both have a much larger overhead storage wise - I assume that's because MongoDB uses some kind of rich internal structure and it also needs to build an index for dict fields in case they are queried upon.

armab:
@m4dcoder https://github.com/StackStorm/st2/issues/4716 would answer your question. There is a whole Github project around HA issues: https://github.com/orgs/StackStorm/projects/8 which is just the tip of the visible iceberg.

There are 500+ issues in StackStorm repository and we can keep saying "X is not supported" and "You're on your own" which we always did. But eventually, that doesn't help us nor our users or bigger adopters who expect more stability from the platform.

With that, what Tomaz suggests would make the user experience :+1: in this particular situation. And the price is not that high!

m4dcoder:
@armab Exactly, I know what the answer is and when you say we support rolling upgrades in k8s I'm curious if we have a different definition.

Kami:
Btw, once the change is finally ready and merged into master, I think we should also do a "beta release" with a friendly user.

Someone who runs StackStorm at scale and is willing to deploy that release and see how it affects their deployment.

That user should also have server + StackStorm process monitoring in place so we can also see how it manifests in those metrics (server CPU and memory utilization, etc.).

Any possible candidates?

Kami:
New micro-benchmark task takes almost 30 minutes on CI vs couple of minutes locally (my local desktop is quite beefy).

I assume that's because github actions containers have much less resources dedicated to them. I will update it so it only runs on a nightly basis instead.

Having said that, this does show that the actual field change will likely bring even larger improvements to deployments on smaller / less powerful servers (at least as far as absolute values goes).

---

On a related note, originally micro-benchmarks only benchmarked ``EscapedDynamicField``, but I updated them to also benchmark ``EscapedDictField``.

We kinda use those field interchangeably even though they are not exactly the same (but we can replace both of them with the new field type).

It turns out that ``EscapedDictField`` is even slower. With new field types we se up to ~10x faster DB writes times compared to ``EscapedDynamicField`` (600 ms 6 seconds) vs and up to ~20x for ``EscapedDictField`` (600 ns vs 12 seconds).

Screenshot from 2021-02-24 12-46-02

Kami:
I'm also working on a change for ``st2 execution get`` to include "log" attribute by default.

As described before, current "end_timestamp" and duration output is deceiving and not accurate since it doesn't include time it takes to write results to the database (and execution is not completed until results are written to the database).

This means it's possible (and quite likely with large payloads) to have a scenario where duration and end_timestamp show 1 seconds, but in reality, complete execution and duration took 10 seconds (rest of it was spend writing result to the database).

This will be less of an issue with this new field type, but we should still include this value. Technically, we could also change when and how ``end_timestamp`` is populated, but that would require more work.

Kami:
In the light of the various performance improvements changes, I decided to take another good look at our old friend aka action registration.

There is a long standing issue where it's slow when registering packs with many actions such as the ``aws`` one.

As discussed in the past, having a single pack with that many actions is not a good idea in any case, but if we can get squeeze some additional performance out of this code without too much work, we should still do it.

I made the following change:

- Use new field type for ``ActionDB.parameters`` field - should be safe since we don't query on this field value, but I may still revert and punt those changes for a future PR (just to reduce the scope)
- Only retrieve 3 fields when retrying action from the database to determine if it exists so we can perform update instead of insert (aka the whole function behaves as upsert). We only need access to the ID field in that case (and pack + ref due to our ActionDB model abstraction). Retrieving all the fields is slow and unnecessary (especially for actions with many parameters).
- Utilize runner DB cache - should be safe to do it in this context. This reduces number of get runner by ref queries from M where M is number of actions (3.5k in AWS pack) to N where N is number of different runner types used by aws actions (should be < 3 or probably even just 1). IIRC, only reason why we didn't utilize it in the past is because of the test - so this may require some work to get tests to pass.
- Use ``fast_deepcopy`` everywhere in JSON schema validation code. This should be safe since schema code only utilizes simple native JSON type. And this should get even faster (couple of % for the overall resource registration operation) once my other orjson PR is merged where we also utilize ``orjson`` for ``fast_deepcopy()``.

All of that brings around 15-20% improvement in the registration time for that pack.

It's not massive, but most rest of the CPU cycles is spent on actual JSON schema + other validation code and I don't think there are any relatively easy and quick wins there.

I will push those changes shortly.

Kami:
Here is the change - 053bd9383d7bd07baabe0347d116802437620ddd.

I also updated the comment to include another optimization - using C version of the YAML safe loader when loading / parsing resource YAML metadata - this offers a nice speed up.

Kami:
I pushed a small change to exclude logs from "st2.st2common.bootstrap.* module from nosetests log capture (https://nose.readthedocs.io/en/latest/plugins/logcapture.html)" by default when running tests - 4793ba375.

Each test loads tons of resources as part of setUp / setUpClass which results in ton of noise if a test failurs which makes it harder and slower to debug failures (there are pages and pages of noise from loading fixture files into db).

With that change, we exclude resource registrar related log messages from nosetests log capture.

Kami:
After some more digging into our slow end to end tests, I believe this change should also help with our end to end tests and CI/CD instance which is very slow at this point (up takes u 10 seconds to load, etc.).

I'm not too surprised though - a lot of those end to end workflows pass a lot of data around and return large results and we run 4 of the concurrency on a PR...

st2common/st2common/constants/meta.py
```diff
@@ -14,12 +14,24 @@
...
+
 __all__ = [
     'ALLOWED_EXTS',
     'PARSER_FUNCS'
 ]
 
+
+# NOTE: We utilize CSafeLoader if available since it uses C extensions and is faster.
+def yaml_safe_load(stream):
+    return yaml.load(stream, Loader=YamlSafeLoader)
```
m4dcoder, line 33: Using yaml.load will raise an alert when we run security scan / static analysis on st2 codebase because yaml.load is not safe to load text from untrusted source.

m4dcoder (changes requested):
Can you address the security concern on using yaml.load?

st2common/st2common/constants/meta.py
Kami: @m4dcoder There is no security concern  here - it's using ``YamlSafeLoader`` (which is either a native Python implementation or C extension one).

And yes, I'm aware of various static analysis tools - they all have limitations and usually detect many false positives.

This line represents no security risks since it uses ``YamlSafeLoader`` which is the same thing used internally by ``yaml.safe_load()`` - but we can't use it directly since we want to use faster C implementation.

I'm not saying that security analysis tools are bad, but results should always be taken with grain of salt and manually audited for false positives, etc. (same as the openapi spec one - even though we didn't use safe load there it didn't represent security risk since it loaded a static file which is not controlled by the end user - but yeah, I still agree that we should use safe load there just to be on the safe side).

Kami: If you want, I can add a code comment and make it explicit, but I kinda assumed it wasn't necessary since it's using SafeLoader class (and as pointed in orquests PR, there are even tests for that scenario).

blag: Yeah, documenting that in a comment would be good for the next person who investigates this issue.

Is there a way to turn off that security check for that one line? This might be one of those rare times that that is acceptable.

Kami: Will add a comment.

As far as turning off the check -goes  yeah, that will need to be handled on the tool side (aka nothing StackStorm/st2 specific).

Technically, if it's some commonly used (open source) tool, I would even be fine with adding some tool specific pragma to that line itself, but otherwise it will need to be handled on the tool side - most such tools have internal database for marking and handling false positives, etc.

And honestly, if that was the only false positive returned by this tool on StackStorm/st2 codebase, I would be vastly surprised :)

Over the years, I managed and set up many such tools and there were always, tons and tons of false positives, even with more advanced tools which use AST based rules + ML (I would assume this one does very simple string / regex based searches for common vulnerabilities based on the pattern above, but I could also be wrong).

Kami: Added a comment in 167ca3f2a91cd20f38b32ab204855a1e86785c67.

m4dcoder: I reviewed pyyaml and safe_load is just calling load but pass in SafeLoader which is the same as what we coded here. I think it's good to put comment describing its use with SafeLoader and that it's the same as calling safe_load. This will help the person who is going to touch this code next and unintentionally revert to default loader. If I didn't reviewed the pyyaml code, a casual contributor would not know that this is not insecure use because it's not using safe_load. The static analysis tool has limitation I agree. In those cases, we need to document while this is a false positive when we identified it so we don't keep circling back to this. Thanks.

m4dcoder (commented):
Tomaz has addressed my concerns with yaml.load and will add comments in the code to document the use of yaml.load instead of yaml.safe_load. Thanks.

Kami:
This one is now also ready for a final review - everything is passing, including end to end tests :tada:

It was a long time coming, but I think we are finally there.

I decided to skip migration script for now to reduce the scope. I'm really not a big fan of our current migration scripts approach (bunch of copy and paste, no good framework and tests for it, etc.).

I may work on the migration script in the future, but if I do work on it, I want to make it more robust than existing scripts and ensure it has good test coverage, etc.

As mentioned above, the code does and will support both approaches for the foreseeable future so there is no big time pressure for a migration script - and the primary benefit of those changes is for new executions - well, yes it migration would also speed up retrieval for old executions, but due to the nature of the data, that's not a big deal.

st2common/st2common/persistence/base.py
```diff
@@ -357,12 +361,19 @@ def _dispatch_trigger(cls, operation, trigger, payload):
 
 class ContentPackResource(Access):
     @classmethod
-    def get_by_ref(cls, ref):
+    def get_by_ref(cls, ref, only_fields: Optional[List[str]] = None):
+        """
+        :param: only_field: Optional lists if fields to retrieve. If not specified, it defaults to
```
cognifloyd, line 366:
```suggestion
        :param: only_fields: Optional list of fields to retrieve. If not specified, it defaults to
```

PS it is so refreshing to see some type hints in the StackStorm code.

st2common/st2common/util/action_db.py
```diff
@@ -158,17 +161,20 @@ def get_action_by_id(action_id):
...
 
-def get_action_by_ref(ref):
+def get_action_by_ref(ref, only_fields: Optional[List[str]] = None):
     """
     Returns the action object from db given a string ref.
 
     :param ref: Reference to the trigger type db object.
     :type ref: ``str``
 
+    :param: only_field: Optional lists if fields to retrieve. If not specified, it defaults to all
```
cognifloyd, line 171:
```suggestion
    :param: only_fields: Optional list of fields to retrieve. If not specified, it defaults to all
```

.github/workflows/ci.yaml
```diff
@@ -40,6 +40,10 @@ jobs:
           - name: 'Unit Tests'
             task: 'ci-unit'
             python-version: '3.6'
+          # This job is slow so we only run in on a daily basis
```
cognifloyd, line 43: Github actions works much better if we use more workflows because you can only restart jobs at the Workflow level, not per job.

So, when we add the benchmark job, it should be in a separate Workflow file.

cognifloyd (commented):
Only minor nitpicks.

Woohoo! This looks superb.

Kami:
I added some additional metrics which should provide better operational visibility into action runner.

In addition to that, I also added some developer docs - https://github.com/StackStorm/st2docs/pull/1060.

Kami:
In the light of https://github.com/StackStorm/st2web/issues/867, I decided to also try and tackle server side component of that Web UI issue.

Good news is that with those changes, st2api and st2action runner will breeze through large executions without much sweat, but those large results will still freeze the whole Web UI.

To improve on that situation I implemented two changes (still WIP, need to polish it, add tests, changelog entry, etc.):

1) Add ``result_size`` field to the ``ActionExecutionDB`` model. This field contains size for the serialized execution result in number of bytes. It's only available for the new field storage format (which is totally fine) and is populated lazily in ``to_mongo()`` so we don't need to serialize data twice to be able to access the size.

2) Add new API endpoint which allows user to retrieve or download "raw" action execution result (``/v1/executions/<id>/result/raw[?download=true]``)

My JavaScript is very rusty these days, but hopefully implementing WebUI part won't be too hard so I will try to tackle it over weekend.

The plan basically is that instead of trying to render the execution result in the code highlighter widget for executions with large results (we will use same sane default value for definition of "large" based on benchmarking etc, but user will also be able to override it in st2web config), we will display something along the lines of "This execution result is too large to be rendered here. Click <here> to view it raw or <here> to download the result as JSON file".

st2api/st2api/controllers/v1/actionexecutions.py
```diff
@@ -365,6 +366,61 @@ def get(self, id, attribute, requester_user):
...
+            # For backward compatibility we also support old non JSON field storage format
+            string_data = orjson.dumps(result["result"], option=orjson.OPT_INDENT_2)
+        else:
+            # For new JSON storage format we just use raw value since it's already JSON serialized
+            # string
+            string_data = result["result"]
+            # TODO: Should we re-serialize for pretty print purposes. It adds around 80ms overhead
+            # for large executions so it's not terrible - e.g. 30ms for using raw results vs 100ms
+            # for reserializing it with pretty indent
+            # string_data = orjson.dumps(orjson.loads(result["result"]), option=orjson.OPT_INDENT_2)
```
cognifloyd, line 409: Maybe add another query param, `pretty=1` (default 0) so people can opt into pretty formatting without copying it into some other service.

Kami: Yeah, that's what I ended up doing in the end (``pretty_format`` query parameter).

.github/workflows/ci.yaml
Kami: ack. I will likely add it in the future to run once a week or something since the job is very slow on CI (~30 minutes).

contrib/examples/actions/workflows/orquesta-data-flow-large-data.yaml
```diff
@@ -0,0 +1,37 @@
+version: 1.0
+
+description: A basic workflow which passes large data around.
+
+input:
+  - file_path
+  - b1: <% ctx().file_path %>
```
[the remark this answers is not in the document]
Kami: I actually just used an existing sample workflow for reference :D

st2client/st2client/formatters/execution.py
```diff
@@ -21,6 +21,11 @@
 
 import yaml
 
+try:
+    from yaml import CSafeDumper as YamlSafeDumper
```
amanda11, line 25: Could we possibly add a comment to explain why we may not have CSafeDumper, and need this catch/exception?

amanda11 (commented):
LGTM - just added a minor comment.
Would like someone else to review as well though, if possible.

st2client/st2client/formatters/execution.py
Kami: Funny enough, I just added similar change recently (a521bde4bd4f6ffb47410fff560e34763c39da04).

Technically as long as people are using official packages it should be there. But if they do from source install, etc it may not be.

Agreed it's a good ideal and will add a comment as well :)

Kami: 83627682d

contrib/core/tests/test_action_sendmail.py
```diff
@@ -167,18 +158,11 @@ def test_sendmail_utf8_subject_and_body(self):
...
-                "This message was generated by StackStorm action "
-                "send_mail running on %s" % (HOSTNAME)
-            )
-        else:
-            expected_body = (
-                "Hello there \\U0001f603\\U0001f603.\n\n"
-                "This message was generated by StackStorm action "
-                "send_mail running on %s" % (HOSTNAME)
-            )
+        expected_body = (
```
amanda11, line 161: It looks like we've kept the code from the if - which if I read it was only if we were running on python2, so should we not have kept the else logic?

contrib/core/tests/test_action_sendmail.py
```diff
@@ -126,20 +125,12 @@ def test_sendmail_utf8_subject_and_body(self):
...
-                "send_mail running on %s" % (HOSTNAME)
-            )
-        else:
-            expected_body = (
-                "Hello there \\U0001f603\\U0001f603.\n"
-                "<br><br>\n"
-                "This message was generated by StackStorm action "
-                "send_mail running on %s" % (HOSTNAME)
-            )
+        expected_body = (
```
amanda11, line 128: It looks like we've kept the code from the if - which if I read it was only if we were running on python2, so should we not have kept the else logic?

st2common/st2common/openapi.yaml
```diff
@@ -1346,6 +1346,50 @@ paths:
...
+          in: query
+          description: True to force result file download instead of returning it as a response
+          type: boolean
+          default: false
+        - name: compress
+          in: query
+          description: True to compress the response body
+          type: boolean
+          default: false
+        - name: pretty_format
```
amanda11, line 1372: Out of interest, is the UI going to use the pretty format or the non-pretty format, and does this affect how the response appears in the UI at all?

amanda11 (commented):
Sorry - realised I hadn't seen the full PR changes on last review.

Just a couple of other queries...

contrib/core/tests/test_action_sendmail.py
Kami: The version of the used library has also been changed / upgraded so now the behavior is different / more correct.

contrib/core/tests/test_action_sendmail.py
Kami: Same as above.

st2common/st2common/openapi.yaml
Kami: Yeah, UI will use pretty format, but keep in mind that this is only used when displaying results for large executions - we don't display it directly in the UI anymore (since that would freeze the browser. etc.), but we just display link to the API endpoint which returns raw result.

Here is a WIP st2web change - https://github.com/StackStorm/st2web/pull/868.

amanda11 (approved):
LGTM

Makefile
```diff
@@ -63,14 +63,17 @@ ifndef PYLINT_CONCURRENCY
 	PYLINT_CONCURRENCY := 1
 endif
 
-NOSE_OPTS := --rednose --immediate --with-parallel --nocapture
+# NOTE: We exclude resourceregistrar DEBUG level log messages since those are very noisy (we
+# loaded resources for every tests) which makes tests hard to troubleshoot on failure due to
+# pages and pages and pages of noise.
+NOSE_OPTS := --rednose --immediate --with-parallel --nocapture --logging-filter=-st2.st2common.bootstrap
```
blag, line 69: The minus sign in front of `st2.st2common.bootstrap` is unexpected.

Can you add a link to the nose documentation, and/or a quote from that explaining what it is and what it is doing?

```suggestion
# The minus in front of st2.st2common.bootstrap filters out logging statements from that module.
# See https://nose.readthedocs.io/en/latest/usage.html#cmdoption-logging-filter
NOSE_OPTS := --rednose --immediate --with-parallel --nocapture --logging-filter=-st2.st2common.bootstrap
```

st2client/st2client/commands/action.py
```diff
@@ -1462,6 +1486,13 @@ def __init__(self, resource, *args, **kwargs):
         self.parser.add_argument(
             "id", help=("ID of the %s." % resource.get_display_name().lower())
         )
+        self.parser.add_argument(
+            "-x",
```
blag, line 1490: Can we get a long form argument as well?

```suggestion
            ["-x", "--exclude-result"],
```

test-requirements.txt
```diff
@@ -35,3 +35,4 @@ pyrabbit
 pip-tools<5.4
 pytest==5.4.3
 pytest-benchmark==3.2.3
+zstandard==0.15.1
```
blag, line 38: Where did this come from? I'd add a comment noting where it's used, or what depends on it.

blag (commented):
LGTM, but there's a few small improvements that could be done.

test-requirements.txt
Kami: It's test required for micro benchmarks, will add a test.

st2client/st2client/commands/action.py
Kami: Yep

Makefile
Kami: This output has been bugging me for a long time (it just causes tons of noise for every test and makes easier issues harder to spot).

But yeah, it's nose specific syntax, will add a comment.

Kami:
@blag Thanks for the review.

I will go ahead and merge this since it's already as massive PR, but I plan to open another small PR on top of that in the near future.

---

Having said that, 'm a bit OCD and that extra query on get in st2web change (https://github.com/StackStorm/st2web/pull/868) was not making me happy, so I identified another approach I could use to avoid additional query in st2web to get execution result when result size < threshold.

Basically, I will modify API and add ``?max_result_size=X`` (or similar, tbd) query parameter to ``/v1/executions/<id>`` API endpoint.

Instead of performing a simple "get one by id" query, we will perform a more complex query (something along the lines of ``id == Yo and (result_size < X | not result_size exists)`` and only return result with the execution object in case the query returns any data.

Good news is that this query is fast since first filter already filter on the PK so it should add negligible overhead (funny enough, based on the testing I did, sometimes it even returns faster than just a plan get by ID query, no idea why), but it will allow us to simplify st2web change and also avoid one extra API operation on st2web side.

jegesh:
I'm looking to patch this PR into my current st2 installation. Any documentation I should refer to? Tips? I'm running v3.4 on Ubuntu 18 and st2 slows to a crawl when we run a few nested workflows, even though I set `WORKERS=30` in the config file (and I can see 30 worker processes in `htop`), so fixing this will really be a godsend for us.

Kami:
@jegesh this has been merged and will be included in the next release which should be out soon. In the mean time you can use dev unstable package for testing if you feel adventurous.

PR #4847, opened 2020-01-20
Fix to st2client install check

m4dcoder:
The make .st2client-install-check command is failing because latest more-itertools no longer supports python 2.7. To workaround, pin more-itertools to an older version.  The zipp requirement is pinned to >=0.5,<=1.0.0 because the latest 2.0.0 version is registered as 0.0.0 and causing install check to fail.

Closes #4843

### DIFF
--- a/fixed-requirements.txt
+++ b/fixed-requirements.txt
@@ -57,3 +57,4 @@ ujson==1.35
 python-dateutil==2.8.0
 bcrypt==3.1.7
 jinja2==2.10.3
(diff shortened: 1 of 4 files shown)


armab:
Looks like Travis is still not happy about some of st2client dependencies, according on https://travis-ci.org/StackStorm/st2/jobs/639637822

m4dcoder:
@armab This is a new and different error.

PR #4848, opened 2020-01-20
Notify #development channel in community Slack about TravisCI failures

armab:
Post TravisCI build failures on master or nightly builds to `#development` channel in `stackstorm-community` Slack


armab:
As master Travis build should fail because of issue described in https://github.com/StackStorm/st2/pull/4847, we'll check if notification is sent to new `#development` Slack channel.

PR #4854, opened 2020-01-29
Convert escaped dict fields in workflow db models to dict fields

m4dcoder:
To improve database performance, convert stormbase.EscapedDictField to mongoengine.DictField and stormbase.EscapedDynamicField to mongoengine.DynamicField in workflow and task execution DB models. An escaped field requires additional step to read from and write to the database.


CHANGELOG.rst
```diff
@@ -54,6 +54,8 @@ Changed
   before returning error to the end user. #4834
 * Upgrade ``pymongo`` to the latest stable version (``3.10.0.``). #4835 (improvement)
 * Remove `.scrutinizer.yml` config file. No longer used.
+* Convert escaped dict and dynamic fields in workflow db models to normal dict and dynamic fields.
```
armab: Please include a note that it's performance optimization, which is not clear from the current changelog description.
IMO it's important to communicate perf. enhancements explicitly as they're really "high valued" improvements our users will be happy to find.

PR #4861, opened 2020-02-11
#4804 Handle action parameters of type array correctly

### LINKED ISSUE #4804: Shell script action not serializing array properly.
## SUMMARY

According to the documentation (https://docs.stackstorm.com/actions.html#update-argument-parsing-in-the-script), arrays should be passed to bash script actions as a comma-delimited string.  

> array - Serialized as a comma delimited string (e.g. foo,bar,baz).

But it's coming through as a python list serialized into a string.

`[u'foo', u'bar', u'baz']`

### STACKSTORM VERSION

Paste the output of ``st2 --version``:

```
stanley@st2-dev:~$ st2 --version
st2 3.1.0, on Python 2.7.12
```

Post what OS you are running this on, along with any other relevant information.

Ubuntu 16.04 LTS installed via apt packaging.

**Example Action YAML File**

```
---
name: "test_bash_array"
runner_type: "remote-shell-script"
description: "Tests an array passed to a bash script."
enabled: true
entry_point: "test_bash_array.sh"
parameters:
    test_array:
        type: "array"
        description: "Test array in bash"
        position: 0
```

**Example Action Bash Script**

```
#!/usr/bin/env bash

TEST_ARRAY=$1

IFS=","; for i in ${TEST_ARRAY}; do echo "$i"; done

exit 0
```

**Invoking Action**

```
test_array:
  - foo
  - bar
  - baz
```

## Expected Results

What did you expect to happen when running the steps above?

Given the above bash script, I'd expect

```
foo
bar
baz
```

## Actual Results

What happened? What output did you get?

Instead I'm getting

```
[u'foo'
 u'bar'
 u'baz']
```

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

winem:
This closes #4804 and actually affects more than just remote shell runner actions.

In the beginning this looked like a simple issue with the serialization and parsing of the parameter but it actually looks like this did not work at all. 

So I address 2 issues with this PR: 
1. `     elif argument_type == 'list': ` actually did not match when the type was `array`. This means that the final else matched and the parameter was just transformed to a unicode string. Although according to the documentation, `array` is the correct value (See https://docs.stackstorm.com/actions.html#action-metadata) I see some potential how this could break something in existing environments where the unicode string is already handled by the script and users did not notice that this is actually a bug. 
**Question 1: Do we want to address this and if yes, how?** This change should at least be mentioned in the changelog once it's released. 
**Question 2: Is there any reason (i.e. backward compatibility) I am not aware of to check if the parameter type is either `list` or `array` and not just `array`?**
2. Its required to map the values and cast them to strings to handle the case where not all the values of the array are either strings or integers. `join` on an array with strings and integers or other values for example would fail terribly. 

The solution provided with this PR is the pragmatic approach. It's actually a trade-off and may apply unnecessary casting operations when applying str() to the item. 

## Alternative approach
The alternative would look like this: 
```
    elif argument_type == 'list' or argument_type == 'array':
        # Lists are serialized a comma delimited string (foo,bar,baz)
        argument_value_types = set([type(value) for value in argument_value])
        if len(argument_value_types) == 1 and int in argument_value_types or str in argument_value_types:
            # join arrays that hold just strings or integers
            argument_value = ','.join(argument_value) if argument_value else ''
        else:
            # join arrays with items of any other type and transform these items to strings
            argument_value = ','.join(map(str, argument_value)) if argument_value else ''
```
Here we check the type of all the items in an array first and just join the items if they are either just strings or just integers. In all other cases, str() would be applied to all the items in the array.

But I think this just shifts the overhead to another place and is not worth it. I did not do any performance tests with real large arrays yet but I doubt that this matters here. Any input by more experienced experts is appreciated!


winem:
I will take a look at the failed test. I was actually reviewing the current tests already and was wondering why that tests did not fail in the past. So, let's see... Investigation ongoing.

winem:
So the failed test was caused by: https://github.com/StackStorm/st2/blob/cdfeeafc597cff44556a7b61a9e90426194313e0/st2common/tests/unit/test_action_db_utils.py#L466 where the used type was also 'list' and not 'array'. 

So, since I don't believe in coincidence this seems to confirm the concerns I mentioned in the PR: just checking the type for `array` instead of `list` will break existing stuff and their might be historical reasons why st2 should support both types.

I pushed a commit that checks if the type is in `[ 'array', 'list' ]` and assume that the test will be fixed by this commit. In the end, I'm open to adapt the test to use `type: array` as well but I would like to get some feedback from the "old hands".

st2common/st2common/util/action_db.py
```diff
@@ -283,9 +283,9 @@ def serialize_positional_argument(argument_type, argument_value):
             argument_value = '1' if bool(argument_value) else '0'
         else:
             argument_value = ''
-    elif argument_type == 'list':
+    elif argument_type in [ 'array', 'list' ]:
```
blag:
```suggestion
    elif argument_type in ['array', 'list']:
```

This should fix the Travis linting build.

blag:
I reran the failing Travis test and it passed. Solving the intermittent failures on Travis is out of the scope of this PR.

At this point the only failing Travis build is the linting run, so fixing that as I've suggested should solve it. @winem If you're still interested in merging this, just make that fix and I'll approve it and merge it in! 😃

st2common/st2common/util/action_db.py
winem: Done. Unfortunately the build still fails but this looks like it ran into some timeout. Let me know if anything else is needed.

winem:
Hi @blag, do you have any opinion on the `array` vs. `list` vs. `support both` topic? Or maybe even some insights from the history of stackstorm?

blag:
I reran all of the failing tests - everything passes now!

At this point I think we'll need to support both `list` and `array`. I'm skeptical that `argument_type` is ever `list` in production, but I don't think it would hurt to support both.

However, this PR also needs at least a unit test for this function so nobody breaks it in the future. You should be able to extend [the test for `get_args`](https://github.com/StackStorm/st2/blob/cdfeeafc597cff44556a7b61a9e90426194313e0/st2common/tests/unit/test_action_db_utils.py#L329) to include tests for arrays.

winem:
Thanks for re-running the tests and and the feedback! 

I just added the tests. Let's just wait for the unit tests to be done.

blag:
Sorry I should have mentioned this before: please add a changelog entry for this fix.

winem:
No worries, @blag. Could have though of it as well. Anyway, done. :+1:

winem:
Hi @m4dcoder,

to be honest, I can't follow, so I will just provide as much details as possible and hope that it helps. 

Without the change from this PR an array would be passed to the shell command as a unicode string if the parameter type is array (as the documentation recommends).
So the parameter to the script was 
```
[u'foo'
 u'bar'
 u'baz']
```
instead of the expected `foo,bar,baz`. 

What this PR does is to handle parameters of both types, list and array, the same way and pass the list values as comma separated argument to the shell command runner. 

I added unit tests which use the `params.actionarray` in addition to `params.actionlist` to ensure that we test both types of list parameters and have the same tests for both. 

So I think I can just come up with tests that prove that the solution works but I can't provide one that reproduces the old behavior and would fail expectedly. 

Please let me know if I miss anything or got something wrong.

cognifloyd:
This is primarily an issue with documented vs actual functionality.

[In the docs](https://docs.stackstorm.com/actions.html#action-metadata):
> `parameters` - A dictionary of parameters and optional metadata describing type and default. The metadata is structured data following the JSON Schema specification draft 4. The common parameter types allowed are `string`, `boolean`, `number` (...), `object`, `integer` (...) and `array`.

Which references the JSON Schema Spec 4. [This section](https://tools.ietf.org/html/draft-zyp-json-schema-04#section-3.5) is particularly relevant:
> JSON Schema defines seven primitive types for JSON values:
>  - `array`  A JSON array.
>  - `boolean`  A JSON boolean.
>  - `integer`  A JSON number without a fraction or exponent part.
>  - `number`  Any JSON number.  Number includes integer.
>  - `null`  The JSON null value.
>  - `object`  A JSON object.
>  - `string`  A JSON string.

So, `array` is called out in the docs as "the" way to get a array/list parameter into an action, and cites an external spec as the source for the naming convention. However, only `list` was implemented, not `array`. So, this PR adds proper `array` support but maintains backwards compatibility for `list`.

The one thing I see that could probably still use another test is this line:
https://github.com/StackStorm/st2/blob/0725178b038c31222dae49c21d393fb340d3d6ce/st2common/st2common/util/action_db.py#L288
That is the implementation of @winem's second point:
> 2. Its required to map the values and cast them to strings to handle the case where not all the values of the array are either strings or integers. join on an array with strings and integers or other values for example would fail terribly.

Maybe this needs one or more tests that demonstrate a list that includes something that is neither int nor string to show when the casting to string is necessary.

winem:
Hi @cognifloyd, thank you for the extensive answer. Now I got it and totally agree that this kind of test is missing.

I added a test for arrays with values of different types and might have found another issue. But this may also be a matter of expectation. So when someone enters a complex number as value it will be interpreted as complex number first and st2 would just pass the result.

This is probably obvious for developers but may cause issues in the rare case that a user wants to pass a complex number as parameter to a script. Do we wanna add this to the documentation?

st2common/tests/unit/test_action_db_utils.py
```diff
@@ -412,10 +418,31 @@ def test_get_args(self):
...
+            '',
             ''
         ]
         pos_args, named_args = action_db_utils.get_args(params, ActionDBUtilsTestCase.action_db)
         self.assertListEqual(pos_args, expected_pos_args, 'Positional args not parsed correctly.')
+
+        # Test arrays and lists with values of different types
+        params = {
+            'actionarray': [ None, False, 1, 4.2e1, '1e3', 'foo' ],
+            'actionlist': [ None, False, 1, 73e-2, '1e2', 'bar' ]
```
cognifloyd, line 430: Silence flake8:
```suggestion
            'actionarray': [None, False, 1, 4.2e1, '1e3', 'foo'],
            'actionlist': [None, False, 1, 73e-2, '1e2', 'bar']
```

armab:
@cognifloyd Thanks for the review and feedback, that was helpful! :+1: 

@m4dcoder this PR is up to you for the final decision.

PR #4862, opened 2020-02-11
Switch travis to use Bionic, Mongo 4.0, and RabbitMQ 3.8

cognifloyd:
Switches travis to use Bionic, Mongo 4.0, and RabbitMQ 3.8

- The commits transition first to xenial and then to bionic.
- I enabled the rabbitmq tests that were disabled until we moved to xenial.
- Those ssl rabbitmq tests required a newer version of rabbitmq, so we use 3.8 (the latest).
- The python3.6 tests were partially using python2.7 which causes the tests to fail in bionic.
- Mongo 3.4 is not available in Bionic, so 4.0 is required.
- Also, the test thresholds were increased to account for increased test times due to the mongo update and the python 3.6 fix.

closes #4772 (replaced by and partially included in this PR)
closes #4863 (included in this PR)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -951,6 +951,12 @@ ci-py3-unit:
 	@echo "==================== ci-py3-unit ===================="
 	@echo
 	NOSE_WITH_TIMER=$(NOSE_WITH_TIMER) tox -e py36-unit -vv
+
+.PHONY: ci-py3-packs-tests
+ci-py3-packs-tests:
+	@echo
+	@echo "==================== ci-py3-packs-tests ===================="
+	@echo
 	NOSE_WITH_TIMER=$(NOSE_WITH_TIMER) tox -e py36-packs -vv
 
 .PHONY: ci-py3-unit-nightly
--- a/Makefile
+++ b/Makefile
@@ -672,7 +672,7 @@ endif
 		echo "-----------------------------------------------------------"; \
 		. $(VIRTUALENV_DIR)/bin/activate; \
 		    COVERAGE_FILE=.coverage.integration.$$(echo $$component | tr '/' '.') \
-		    nosetests $(NOSE_OPTS) -s -v --exe $(NOSE_COVERAGE_FLAGS) \
+		    nosetests $(NOSE_OPTS) -s -v $(NOSE_COVERAGE_FLAGS) \
 		    $(NOSE_COVERAGE_PACKAGES) \
 		    $$component/tests/integration || exit 1; \
 		echo "-----------------------------------------------------------"; \
--- a/scripts/travis/install-requirements.sh
+++ b/scripts/travis/install-requirements.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-if [ "${TASK}" = 'compilepy3 ci-py3-unit' ] || [ "${TASK}" = 'ci-py3-integration' ]; then
+if [[ "${TASK}" = *'ci-py3'* ]]; then
     pip install "tox==3.8.6"
 
     # cleanup any invalid python2 cache
@@ -22,15 +22,21 @@ if [ "${TASK}" = 'compilepy3 ci-py3-unit' ] || [ "${TASK}" = 'ci-py3-integration
     # NOTE: We create the environment and install the dependencies first. This
     # means that the subsequent tox build / test command has a stable run time
     # since it doesn't depend on dependencies being installed.
-    if [ "${TASK}" = 'compilepy3 ci-py3-unit' ]; then
-        TOX_TASK="py36-unit"
+    build_tox_env() {
+        tox -e $1 --notest
+    }
+
+    if [[ "${TASK}" = *'ci-py3-unit'* ]]; then
+        build_tox_env py36-unit
     fi
 
-    if [ "${TASK}" = 'ci-py3-integration' ]; then
-        TOX_TASK="py36-integration"
+    if [[ "${TASK}" = *'ci-py3-packs-tests'* ]]; then
+        build_tox_env py36-packs
     fi
 
-    tox -e ${TOX_TASK} --notest
+    if [[ "${TASK}" = *'ci-py3-integration'* ]]; then
+        build_tox_env py36-integration
+    fi
 else
     make requirements
 fi
--- a/scripts/travis/install-requirements.sh
+++ b/scripts/travis/install-requirements.sh
@@ -3,6 +3,11 @@
 if [ "${TASK}" = 'compilepy3 ci-py3-unit' ] || [ "${TASK}" = 'ci-py3-integration' ]; then
     pip install "tox==3.8.6"
 
+    # cleanup any invalid python2 cache
+    test -d virtualenv/lib/${PYTHON_VERSION} || rm -rf virtualenv/*
+    # rebuild virtualenv if necessary
+    test -f virtualenv/bin/activate || virtualenv --python=${PYTHON_VERSION} virtualenv --no-download
+
     # Install runners
     . virtualenv/bin/activate
 
--- a/scripts/travis/permissions-workaround.sh
+++ b/scripts/travis/permissions-workaround.sh
@@ -9,6 +9,6 @@ fi
 # stanley needs to work with some fixtures during the tests
 # this can't be the travis user because 'stanley' is the hardcoded user in the tests
 # o=other; X=only set execute bit if user execute bit is set (eg on dirs)
-chmod -R o+rwX ${TRAVIS_BUILD_DIR}/st2tests/st2tests/fixtures/packs
+chmod -R o+rX ${TRAVIS_BUILD_DIR}/st2tests/st2tests/fixtures/packs
 # rabbitmq user needs access to these certs
-chmod -R o+rwX ${TRAVIS_BUILD_DIR}/st2tests/st2tests/fixtures/ssl_certs
+chmod -R o+rX ${TRAVIS_BUILD_DIR}/st2tests/st2tests/fixtures/ssl_certs
--- a/scripts/travis/permissions-workaround.sh
+++ b/scripts/travis/permissions-workaround.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+set -e
+
+if [ "$(whoami)" != 'root' ]; then
+    echo 'Please run with sudo'
+    exit 2
+fi
+
+# stanley needs to work with some fixtures during the tests
+# this can't be the travis user because 'stanley' is the hardcoded user in the tests
+# o=other; X=only set execute bit if user execute bit is set (eg on dirs)
+chmod -R o+rwX ${TRAVIS_BUILD_DIR}/StackStorm/st2/st2tests/st2tests/fixtures/packs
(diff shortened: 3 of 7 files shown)


scripts/travis/permissions-workaround.sh
```diff
@@ -0,0 +1,19 @@
...
+
+UBUNTU_VERSION=`lsb_release -a 2>&1 | grep Codename | grep -v "LSB" | awk '{print $2}'`
+
+# Workaround for Travis on Ubuntu Xenial so local runner integration tests work
+# when executing them under user "stanley" (by default Travis checks out the
+# code and runs tests under a different system user).
+# NOTE: We need to pass "--exe" flag to nosetests when using this workaround.
+if [ "${UBUNTU_VERSION}" == "xenial" ]; then
+  echo "Applying workaround for stanley user permissions issue to /home/travis on Xenial"
+  chmod 777 -R /home/travis
```
blag: Using octal notation here forces all files to be executable, which then forces you to modify a million files adding the `--exe` flag to all `nosetests` invocations.

Instead, I would use the more human-friendly `chmod` syntax to make only the permissions changes that you want to:

```bash
chmod ugo+r -R /home/travis
```

Also consider that you can change the owner of files to `stanley` instead of changing permissions:

```bash
chown -R stanley /home/travis/...
```

Or better yet, we should figure out if we can get Travis to run setup scripts as the `stanley` user.

Makefile
```diff
@@ -1011,6 +1011,10 @@ ci-unit-nightly:
 .ci-prepare-integration:
 	sudo -E ./scripts/travis/prepare-integration.sh
 
+.PHONY: .ci-travis-permissions-workaround
```
blag: Instead of adding another make target for specifically running on Travis, could you simply add this to the `scripts/travis/prepare-integration.sh` script called from the `.ci-prepare-integration` target?

st2common/st2common/util/virtualenvs.py
```diff
@@ -199,14 +199,14 @@ def create_virtualenv(virtualenv_path, logger=None, include_pip=True, include_se
...
-        exit_code, _, stderr = run_command(cmd=cmd)
+        exit_code, stdout, stderr = run_command(cmd=cmd)
     except OSError as e:
         raise Exception('Error executing command %s. %s.' % (' '.join(cmd),
                                                              six.text_type(e)))
 
     if exit_code != 0:
-        raise Exception('Failed to create virtualenv in "%s": %s' %
-                        (virtualenv_path, stderr))
+        raise Exception('Failed to create virtualenv in "%s":\n%s\n|||||\n%s' %
```
blag: I assume that is only temporary while you debug the Travis Bionic environment.

If this is meant to be permanent, then I am not a huge fan of this output syntax - the format itself isn't very readable (I recommend using `.format()`), and it doesn't produce very readable output.

st2common/st2common/util/virtualenvs.py
```diff
@@ -169,7 +169,7 @@ def create_virtualenv(virtualenv_path, logger=None, include_pip=True, include_se
     logger.debug('Creating virtualenv in "%s" using Python binary "%s"' %
                  (virtualenv_path, python_binary))
 
-    cmd = [virtualenv_binary]
+    cmd = [virtualenv_binary, '-v']
```
blag: It's odd that you have to modify StackStorm's internals just to get it to run properly on Travis. It just seems like that is "fixing" the problem from the wrong end.

Instead of making StackStorm run better on Travis (which implies that it _doesn't_ run as well on non-Travis machines, like users' machines, which should be a warning flag!), I think the ultimate solution will be correcting the Travis configuration to properly run StackStorm.

It's best to have your test environment resemble a normal production environment as much as humanly possible. Otherwise you end up with software that runs tests well but doesn't actually run user workloads well.

st2common/tests/integration/test_rabbitmq_ssl_listener.py
```diff
@@ -58,14 +55,19 @@ def test_non_ssl_connection_on_ssl_listener_port_failure(self):
...
+        epxected_msg_4 = '[Errno 104] ECONNRESET'
 
         try:
             connection.connect()
         except Exception as e:
             self.assertFalse(connection.connected)
             self.assertTrue(isinstance(e, (IOError, socket.error)))
-            self.assertTrue(expected_msg_1 in six.text_type(e) or expected_msg_2 in
-                            six.text_type(e))
+            self.assertTrue(expected_msg_1 in six.text_type(e) or
```
blag: Not a huge fan of this. It's fine for now, but if gets more complicated I think a good combination of a list comprehension and set intersection might be a better way to implement this.

st2common/st2common/util/virtualenvs.py
cognifloyd: yes. This is just for extremely verbose debugging. :)

st2common/st2common/util/virtualenvs.py
cognifloyd: This was just for increasing the debugging output as well.

st2common/tests/integration/test_rabbitmq_ssl_listener.py
cognifloyd: seems like every platform has a slightly different error message. I had to add one for xenial, and then bionic had yet another. :/

Makefile
cognifloyd: I will go back and look at why kami added this to see if that is enough. I cherry-picked this change from #4772

scripts/travis/permissions-workaround.sh
cognifloyd: yes. this is very ugly. switching to `ugo+r` would be much nicer.

st2common/st2common/util/virtualenvs.py
blag: :+1: Resolving this. I'll have another look once this PR is finished.

st2common/tests/integration/test_rabbitmq_ssl_listener.py
blag: Yeah, it's fine for now, it's just something we should keep in mind in the future.

Makefile
cognifloyd: Seems it was for packs and unit tests as well, not just integration tests (which had the fix first).

scripts/travis/permissions-workaround.sh
cognifloyd: I used `a+rwX`. Seems to work. Maybe figuring out how to run the tests as stanley would be good in a different PR. I'm updating mongo/rabbitmq repos/versions to make the tests work in the new environment, but I'm avoiding making any other changes.

blag: Why are you touching the executable bit at all though? If you use the `+` notation, you don't have to specify it and `chmod` will leave the executable bit alone.

I'm also not a huge fan of changing the permissions for the owner, the group, and all others. That's very likely not how ST2 is run in production systems, and it shouldn't be how it's run during tests.

Would it be simpler to just `chown` the ST2 code to the `stanley` user?

Makefile
blag: Are you referring to tests for packs in StackStorm-Exchange? Because they exclusively use CircleCI, not Travis, and I don't believe ST2 used in those tests runs this make target.

Maybe we should just rename that script, update any references to it, and update the make target name as well. 😆

cognifloyd: https://github.com/StackStorm/st2/compare/8c55ab37bc1dd546558aafa5c9f277c2430f2c7c..d19dc8dbda30a4cedadb89a699cf064396e0e405

make targets:
ci-py3-unitci-packs-test

cognifloyd: These make targets are used in the 3rd and 4th travis environments:
- Lint Checks, Packs Tests (Python 2.7)
- Unit Tests, Pack Tests (Python 3.6)

Plus the 2nd and 5th travis environments which already had the workaround:
- Integration Tests (Python 2.7)
- Integration Tests (Python 3.6)

Leaving the 1st environment as the only one not using the workaround:
- Unit Tests (Python 2.7 MongoDB 4.0)

cognifloyd:
Mongo 4.0 significantly changes the test runtimes. 2x was too much of a
jump, so this is approximately 60s more than the last successful run
time. That comes out to (test time used to calculate in parentheses):

24% increase for "Unit Tests (Python 2.7 MongoDB 4.0)" (812s)
37% increase for "Integration Tests (Python 2.7)" (903s)
95% increase for "Lint Checks, Packs Tests (Python 2.7)" (490s)
71% increase for "Unit Tests, Pack Tests (Python 3.6)" (1101s)
252% increase for "Integration Tests (Python 3.6)" (1031s)

Note that this PR also fixes the Python 3.6 integration tests to
actually run in python 3.6 instead of 2.7, so that might account for
some of the significant test time increase.

When we move to python 3.7 that will probably decrease the py3 times.

cognifloyd:
This is ready. I updated the PR description to better cover the updates required in moving to Bionic.

armab:
Thanks for the testing and timing breakdown! :+1: 
According to stats you provided, we'd likely go with the https://github.com/StackStorm/st2/pull/4863 Xenial + Mongo 3.4 for the st2 tests.

scripts/travis/install-requirements.sh
```diff
@@ -3,6 +3,11 @@
 if [ "${TASK}" = 'compilepy3 ci-py3-unit' ] || [ "${TASK}" = 'ci-py3-integration' ]; then
     pip install "tox==3.8.6"
 
+	# cleanup any invalid python2 cache
```
blag: Whitespace issues here - using tabs when the rest of the file uses spaces.

scripts/travis/prepare-integration.sh
```diff
@@ -38,8 +38,7 @@ chmod 777 logs/*
 # Workaround for Travis on Ubuntu Xenial so local runner integration tests work
 # when executing them under user "stanley" (by default Travis checks out the
 # code and runs tests under a different system user).
-# NOTE: We need to pass "--exe" flag to nosetests when using this workaround.
-if [ "${UBUNTU_VERSION}" == "xenial" ]; then
-  echo "Applying workaround for stanley user permissions issue to /home/travis on Xenial"
-  chmod 777 -R /home/travis
+if [ "${UBUNTU_VERSION}" == "xenial" ] || [ "${UBUNTU_VERSION}" == "bionic" ]; then
```
blag: If you duplicated this logic into the new `scripts/travis/permissions-workaround.sh`, then I would remove this code here, and simply call that script here. Alternatively, you could use make magic to build both targets.

scripts/travis/permissions-workaround.sh
```diff
@@ -0,0 +1,18 @@
...
+fi
+
+UBUNTU_VERSION=`lsb_release -a 2>&1 | grep Codename | grep -v "LSB" | awk '{print $2}'`
+
+# Workaround for Travis on Ubuntu Xenial/Bionic so local unit and packs tests work
+# when executing them under user "stanley" (by default Travis checks out the
+# code and runs tests under a different system user).
+if [ "${UBUNTU_VERSION}" == "xenial" ] || [ "${UBUNTU_VERSION}" == "bionic" ]; then
+  echo "Applying workaround for stanley user permissions issue to /home/travis on Xenial/Bionic"
+  chmod -R a+rwX /home/travis
```
blag: Would it make more sense to change the ownership of some of the files in `/home/travis` to `stanley` instead of trying to `chmod` them to the appropriate settings?

And even if it still makes sense to use `chmod`, we still should not be setting all of the bits. We should be as selective as possible.

PR #4863, opened 2020-02-12
Switch Travis from Precise to Xenial infra

cognifloyd:
Try to upgrade travis to xenial from precise.
Ensures that mongo is 3.4 not 4.0. mongo 3.4 is not available in bionic so #4862 probably won't work until dependence on 3.4 is gone.
#4772 tried to update and ran into a many issues esp slowness. It also missed downgrading mongodb.

closes #4772
closes #4862

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,8 +58,8 @@ addons:
       #- sourceline: 'deb [arch=amd64] https://dl.bintray.com/rabbitmq/debian precise rabbitmq-server-v3.6.x'
       #  key_url: 'https://github.com/rabbitmq/signing-keys/releases/download/2.0/rabbitmq-release-signing-key.asc'
     packages:
-      - mongodb-org-server
-      - mongodb-org-shell
+      - mongodb-org-server=3.4.*
+      - mongodb-org-shell=3.4.*
       - erlang
       - rabbitmq-server
       - libffi-dev
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,20 +35,6 @@ jobs:
     - env: TASK=ci-unit CACHE_NAME=py2 COMMAND_THRESHOLD=700
       python: 2.7
       name: "Unit Tests (Python 2.7 MongoDB 3.4)"
-    #- env: TASK=ci-unit CACHE_NAME=py2 COMMAND_THRESHOLD=700
-      #python: 2.7
-      #name: "Unit Tests (Python 2.7 MongoDB 3.6)"
-      #addons:
-      #  apt:
-      #    sources:
-      #      - mongodb-upstart
-      #      - sourceline: 'deb [arch=amd64] http://repo.mongodb.org/apt/ubuntu precise/mongodb-org/3.6 multiverse'
-      #        key_url: 'https://www.mongodb.org/static/pgp/server-3.6.asc'
-      #      - sourceline: 'ppa:git-core/ppa'
-      #    packages:
-      #      - mongodb-org-server
-      #      - mongodb-org-shell
-      #      - git
     - env: TASK=ci-integration CACHE_NAME=py2 COMMAND_THRESHOLD=700
       python: 2.7
       name: "Integration Tests (Python 2.7)"
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,6 +85,7 @@ install:
   - cp conf/st2.dev.conf "${ST2_CONF}" ; sed -i -e "s/stanley/travis/" "${ST2_CONF}"
   - sudo scripts/travis/add-itest-user-key.sh
   - sudo .circle/add-itest-user.sh
+  - if [[ "${TASK}" = *'-packs-tests'* ]] || [[ "${TASK}" = *'-integration'* ]]; then sudo scripts/travis/permissions-workaround.sh; fi
 
 # Let's enable rabbitmqadmin
 # See https://github.com/messagebus/lapine/wiki/Testing-on-Travis.
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 os: linux
-# NOTE: We use precise because tests finish faster than on Xenial
-dist: precise
+dist: xenial
 language: python
 
 branches:
@@ -51,21 +50,18 @@ jobs:
 addons:
   apt:
     sources:
-      - mongodb-upstart
-      - sourceline: 'deb [arch=amd64] http://repo.mongodb.org/apt/ubuntu precise/mongodb-org/3.4 multiverse'
+      - sourceline: 'deb [arch=amd64] http://repo.mongodb.org/apt/ubuntu xenial/mongodb-org/3.4 multiverse'
         key_url: 'https://www.mongodb.org/static/pgp/server-3.4.asc'
       # NOTE: Precise repo doesn't contain Erlang 20.x, latest version is 19.x so we need to use RabbitMQ 3.7.6
       #- sourceline: 'deb [arch=amd64] http://packages.erlang-solutions.com/ubuntu precise contrib'
       #  key_url: 'https://packages.erlang-solutions.com/ubuntu/erlang_solutions.asc'
       #- sourceline: 'deb [arch=amd64] https://dl.bintray.com/rabbitmq/debian precise rabbitmq-server-v3.6.x'
       #  key_url: 'https://github.com/rabbitmq/signing-keys/releases/download/2.0/rabbitmq-release-signing-key.asc'
-      - sourceline: 'ppa:git-core/ppa'
     packages:
       - mongodb-org-server
       - mongodb-org-shell
       - erlang
       - rabbitmq-server
-      - git
       - libffi-dev
 
 cache:
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,13 +42,13 @@ jobs:
     - env: TASK=ci-integration CACHE_NAME=py2 COMMAND_THRESHOLD=700
       python: 2.7
       name: "Integration Tests (Python 2.7)"
-    - env: TASK="ci-checks ci-packs-tests" CACHE_NAME=py2 COMMAND_THRESHOLD=280
+    - env: TASK="ci-checks ci-packs-tests" CACHE_NAME=py2 COMMAND_THRESHOLD=430
       python: 2.7
       name: "Lint Checks, Packs Tests (Python 2.7)"
-    - env: TASK="compilepy3 ci-py3-unit ci-py3-packs-tests" CACHE_NAME=py3 PYTHON_VERSION=python3.6 COMMAND_THRESHOLD=680
+    - env: TASK="compilepy3 ci-py3-unit ci-py3-packs-tests" CACHE_NAME=py3 PYTHON_VERSION=python3.6 COMMAND_THRESHOLD=750
       python: 3.6
       name: "Unit Tests, Pack Tests (Python 3.6)"
-    - env: TASK="ci-py3-integration" CACHE_NAME=py3 PYTHON_VERSION=python3.6 COMMAND_THRESHOLD=310
+    - env: TASK="ci-py3-integration" CACHE_NAME=py3 PYTHON_VERSION=python3.6 COMMAND_THRESHOLD=770
       python: 3.6
       name: "Integration Tests (Python 3.6)"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,15 +52,9 @@ addons:
     sources:
       - sourceline: 'deb [arch=amd64] http://repo.mongodb.org/apt/ubuntu xenial/mongodb-org/3.4 multiverse'
         key_url: 'https://www.mongodb.org/static/pgp/server-3.4.asc'
-      # NOTE: Precise repo doesn't contain Erlang 20.x, latest version is 19.x so we need to use RabbitMQ 3.7.6
-      #- sourceline: 'deb [arch=amd64] http://packages.erlang-solutions.com/ubuntu precise contrib'
-      #  key_url: 'https://packages.erlang-solutions.com/ubuntu/erlang_solutions.asc'
-      #- sourceline: 'deb [arch=amd64] https://dl.bintray.com/rabbitmq/debian precise rabbitmq-server-v3.6.x'
-      #  key_url: 'https://github.com/rabbitmq/signing-keys/releases/download/2.0/rabbitmq-release-signing-key.asc'
     packages:
       - mongodb-org-server=3.4.*
       - mongodb-org-shell=3.4.*
-      - erlang
       - rabbitmq-server
       - libffi-dev
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,11 @@ jobs:
     # job which also includes "make requirements" and other steps
     # "make requirements" can take substantially lower if the cache is purged
     # and this would cause too many intermediate failures / false positives
+    # NOTE: TASK is inspected in commands below and other travis scripts.
+    # For example, ci-py3* targets trigger an alternate virtualenv build method.
+    # If you rename or reorder make targets in TASK, you may need to adjust:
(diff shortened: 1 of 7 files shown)


cognifloyd:
This is a simpler version of #4772 (I cherry-picked the nosetests workarounds). It looks like travis has resolved some of the other issues that Kami was encountering with xenial.

Tests timing comparison:
- Tests run in comparable or faster time than the latest master build (w/ coverage): https://travis-ci.org/StackStorm/st2/builds/649088426
- Tests run in comparable time (some slightly faster some slightly slower) compared with the most recently successful PR (w/o coverage): https://travis-ci.org/StackStorm/st2/builds/648999898

cognifloyd:
It looks like mongodb 4.0 was the reason for the slowdown in #4772 because #4862 is hitting similar slowdowns.

blag:
At some point we should flip the linter checks to be run under Python 3. That will automatically turn on some additional checks in the linters themselves to keep our code forward compatible with Python 3 linting.

Not saying that has to happen in this PR though.

scripts/travis/install-requirements.sh
```diff
@@ -3,6 +3,11 @@
 if [ "${TASK}" = 'compilepy3 ci-py3-unit' ] || [ "${TASK}" = 'ci-py3-integration' ]; then
     pip install "tox==3.8.6"
 
+	# cleanup any invalid python2 cache
```
blag: Whitespace issue here - use of tabs, when the rest of the diff uses spaces.

scripts/travis/prepare-integration.sh
```diff
@@ -38,8 +38,7 @@ chmod 777 logs/*
 # Workaround for Travis on Ubuntu Xenial so local runner integration tests work
 # when executing them under user "stanley" (by default Travis checks out the
 # code and runs tests under a different system user).
-# NOTE: We need to pass "--exe" flag to nosetests when using this workaround.
 if [ "${UBUNTU_VERSION}" == "xenial" ]; then
   echo "Applying workaround for stanley user permissions issue to /home/travis on Xenial"
-  chmod 777 -R /home/travis
+  chmod -R a+rwX /home/travis
```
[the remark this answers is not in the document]
cognifloyd: K. I will DRY this out a bit.

scripts/travis/permissions-workaround.sh
```diff
@@ -0,0 +1,18 @@
...
+fi
+
+UBUNTU_VERSION=`lsb_release -a 2>&1 | grep Codename | grep -v "LSB" | awk '{print $2}'`
+
+# Workaround for Travis on Ubuntu Xenial so local runner integration tests work
+# when executing them under user "stanley" (by default Travis checks out the
+# code and runs tests under a different system user).
+if [ "${UBUNTU_VERSION}" == "xenial" ]; then
+  echo "Applying workaround for stanley user permissions issue to /home/travis on Xenial"
+  chmod -R a+rwX /home/travis
```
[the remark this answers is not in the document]
cognifloyd: `a+rwX` is better than `777` as it was before. We can change this to only manipulate the other permissions: `o+rwX`.
`r` because we're having issues accessing from another user.
`w` because python will want to write pyc files and other things might need temp files as well.
`X` only sets the execute bit when a file is already executable for the user. Primarily that helps with directories (for listing contents) and actually executable files.

We can probably have a more explicit list of directories to recurse into, but I'm not sure what all is used, so we'd have to experiment with that.

I think changing the ownership of files in a home directory is generally a bad idea.

Another thought: What about changing the st2.test.conf to run things as travis instead of as stanley?

scripts/travis/install-requirements.sh
cognifloyd: oops. will fix

scripts/travis/install-requirements.sh
```diff
@@ -3,6 +3,11 @@
...
+	test -d virtualenv/lib/${PYTHON_VERSION} || rm -rf virtualenv/*
+	# rebuild virtualenv if necessary
+    test -f virtualenv/bin/activate || virtualenv --python=${PYTHON_VERSION} --no-site-packages virtualenv --no-download
```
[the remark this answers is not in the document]
cognifloyd: I will add a commit that drops `--no-site-packages`

scripts/travis/permissions-workaround.sh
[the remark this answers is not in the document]
blag: Awesome, thanks for the explanation there. That's something we should throw in a comment, so when people (or me) run across this in the future they don't have to dig around for why we're setting things they way they are. :smile:

cognifloyd:
Yes. I want to fix the issues metioned by @blag but I can't do that until the other underlying issue(s) are fixed. So, I'm figuring out the fix and just happen to be using this PR (since it can't be merged till after this is fixed anyway). Once I figure out what fixes are required, then I can submit a separate PR.

scripts/travis/permissions-workaround.sh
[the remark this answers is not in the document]
cognifloyd: #4869 included a simplified permissions fix, so I've removed all the permissions stuff from this PR :) (except for one commit where I removed some --exe flags).

cognifloyd:
I had to re-add some permissions workarounds, but I added the permissions to a more selective set of files under /home/travis (instead of the whole thing).

At least for the rabbitmq stuff, we could probably copy the ssl certs to /etc instead of using them in the fixtures dir, but I'm not sure that would make much of an impact because of all the other fixtures that stanley needs to access (as far as I understand).

I also got rid of some of the legacy stuff in `.travis.yml`, clarified the make targets used in each travis job, and took a stab at updating the COMMAND_THRESHOLD numbers. In the latest build, the ci-py3-integration tests are even longer than what I put, so I could amend that commit with a new number.

scripts/travis/install-requirements.sh
```diff
@@ -17,15 +22,21 @@ if [ "${TASK}" = 'compilepy3 ci-py3-unit' ] || [ "${TASK}" = 'ci-py3-integration
     # NOTE: We create the environment and install the dependencies first. This
     # means that the subsequent tox build / test command has a stable run time
     # since it doesn't depend on dependencies being installed.
-    if [ "${TASK}" = 'compilepy3 ci-py3-unit' ]; then
-        TOX_TASK="py36-unit"
+    build_tox_env() {
```
armab: Please revert the code logic as it was before without using this function.
It increases level of indirection, it's arguable if new version is easier comparing to previous one.

Additionally, with increased amount of diff it's harder to navigate what was the _actual_ change.

scripts/travis/install-requirements.sh
```diff
@@ -1,8 +1,13 @@
 #!/usr/bin/env bash
 
-if [ "${TASK}" = 'compilepy3 ci-py3-unit' ] || [ "${TASK}" = 'ci-py3-integration' ]; then
+if [[ "${TASK}" = *'ci-py3'* ]]; then
```
armab: Please revert to more explicit match as there is a chance to include irrelevant tasks in future.

scripts/travis/install-requirements.sh
```diff
@@ -1,8 +1,13 @@
...
     pip install "tox==3.8.6"
 
+    # cleanup any invalid python2 cache
+    test -d virtualenv/lib/${PYTHON_VERSION} || rm -rf virtualenv/*
```
armab: Can you please provide more context around this cleanup?

Are we removing something that should be cached by Travis and saved build time before?

scripts/travis/install-requirements.sh
cognifloyd: And I want to make sure that it catches all py3 tasks here as they all run in tox. I don't want to have to update this line repeatedly if we move the jobs around (as @blag suggested we need to do in another PR with the lint checks moving to python3).

scripts/travis/install-requirements.sh
cognifloyd: Do the commit messages not help?

.travis.yml
```diff
@@ -34,54 +31,41 @@ matrix:
...
+      #- mongodb-upstart
+      - sourceline: 'deb [arch=amd64] http://repo.mongodb.org/apt/ubuntu xenial/mongodb-org/3.4 multiverse'
         key_url: 'https://www.mongodb.org/static/pgp/server-3.4.asc'
-      # NOTE: Precise repo doesn't contain Erlang 20.x, latest version is 19.x so we need to use RabbitMQ 3.7.6
-      #- sourceline: 'deb [arch=amd64] http://packages.erlang-solutions.com/ubuntu precise contrib'
-      #  key_url: 'https://packages.erlang-solutions.com/ubuntu/erlang_solutions.asc'
-      #- sourceline: 'deb [arch=amd64] https://dl.bintray.com/rabbitmq/debian precise rabbitmq-server-v3.6.x'
-      #  key_url: 'https://github.com/rabbitmq/signing-keys/releases/download/2.0/rabbitmq-release-signing-key.asc'
-      - sourceline: 'ppa:git-core/ppa'
+      - sourceline: 'deb [arch=amd64] http://packages.erlang-solutions.com/ubuntu xenial contrib'
```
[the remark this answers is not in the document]
cognifloyd: ok. I'll drop more of the commented stuff.

.travis.yml
```diff
@@ -34,54 +31,41 @@ matrix:
...
+        key_url: 'https://packages.erlang-solutions.com/ubuntu/erlang_solutions.asc'
+      - sourceline: 'deb [arch=amd64] https://dl.bintray.com/rabbitmq/debian xenial rabbitmq-server-v3.8.x'
+        key_url: 'https://github.com/rabbitmq/signing-keys/releases/download/2.0/rabbitmq-release-signing-key.asc'
+      #git is v2.21.0 in xenial
+      #- sourceline: 'ppa:git-core/ppa'
     packages:
-      - mongodb-org-server
-      - mongodb-org-shell
+      - mongodb-org-server=3.4.24
+      - mongodb-org-shell=3.4.24
```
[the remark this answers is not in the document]
cognifloyd: I thought you couldn't specify a version with apt unless it was the full version. So I looked up the latest version and put that here.

.travis.yml
```diff
@@ -34,54 +31,41 @@ matrix:
...
       python: 3.6
       name: "Integration Tests (Python 3.6)"
 
 addons:
   apt:
     sources:
-      - mongodb-upstart
-      - sourceline: 'deb [arch=amd64] http://repo.mongodb.org/apt/ubuntu precise/mongodb-org/3.4 multiverse'
+      # xenial has systemd not upstart
+      #- mongodb-upstart
```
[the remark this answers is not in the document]
cognifloyd: ok I'll drop more of the comments.

.travis.yml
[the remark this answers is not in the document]
cognifloyd: I tried to reenable the xenial tests per @Kami's comments. He was trying to update rabbitmq so I did too. I can try leaving the version at whatever is in xenial to see if the ssl tests still pass.

scripts/travis/install-requirements.sh
cognifloyd: Changing the targeted tasks here makes it very clear that this is expected to run for all py3 tasks and that it is because they are py3 that we're targeting them separately, not just because they're a random selection of tasks.

scripts/travis/install-requirements.sh
cognifloyd: I discovered that the python3 builds sometimes ran with a cached python2 virtualenv, invalidating the tests. I added this fix in c15112b81514c79be3fbb2cc17ae1e3bbda34898.

The makefile only checks to see if the activate script is present, not if it is using the right version of python, so if the cache does inadvertently get corrupted (ie as I experienced with a cache that included python2 instead of python3) I want to catch that and force rebuilding the virtualenv.

This is somewhat related to the change that catches `*ci-py3*` tasks. I want to make sure this runs for all python3 tasks.

scripts/travis/install-requirements.sh
cognifloyd: I was confused several times by how the packs-tests were being run when the target task was clearly just for unit tests. Plus, because the packs-tests were hidden within the unit tests make target, the tox environment for the packs-tests was not actually getting built here as expected. So this change makes it clear that each of these tasks has a tox environment that needs to be built.

blag: @armab Has a good point here. Since it's a one-line function, this is a bit of indirection that isn't really useful.

scripts/travis/install-requirements.sh
blag: If you're going to keep this in, it needs to be commented in somewhere, like `jobs.include`. And you can even include a link to this file in those comments so people can check for themselves.

We should definitely leave breadcrumbs for the poor soul who has to update this in the future.

scripts/travis/install-requirements.sh
blag: Same here - add comments to this effect so the next person who reads this understands why it's important.

scripts/travis/install-requirements.sh
cognifloyd: Yes. It was not the nicest thing to go grepping through to find out where the task names had been hard coded so I could change one of them. I'll add a comment to .travis.yml on where the task names are used.

scripts/travis/install-requirements.sh
cognifloyd: k. I was avoiding repeating the `tox -e ...` multiple times. Because a task can have multiple tox environments I needed a way to either loop over them, or just call it. I can un-DRY this here and repeat the tox calls I guess.

.travis.yml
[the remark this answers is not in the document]
cognifloyd: This was why I went with the full version. https://askubuntu.com/a/916601
I didin't know I could use `.*` but the apt-get man page does say that that is allowed.
I thought it was odd to require only exact versions. Cool that that is not the case.

I will adjust the version.

scripts/travis/install-requirements.sh
blag: Yeah, this both increases complexity (via indirection) and increases the number of lines in the file for very little benefit. Sorry if my comments elsewhere about the DRY-principle steered you this way, but in this case it's probably fine to "unindirect" the call to `tox`.

cognifloyd:
There we go. I heavily rebased and squashed commits. I also included comments and fixes to address reviews. I just pushed one last edit that adjusts only 3 of the thresholds instead of all 5.

Once CI returns green, I think this is ready to merge. :)

scripts/travis/install-requirements.sh
cognifloyd: added a comment.

scripts/travis/install-requirements.sh
cognifloyd: commented in relevant places

scripts/travis/install-requirements.sh
cognifloyd: removed the function and added comments explaining this.

.travis.yml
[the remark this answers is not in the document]
cognifloyd: OK. default from travis works just fine :)

cognifloyd:
added a changelog

CHANGELOG.rst
```diff
@@ -109,6 +111,8 @@ Fixed
 
   Contributed by Tatsuma Matsuki (@mtatsuma)
 
+* Fix python3 tests on travis to ensure test virtualenvs are actually python3. #4863
```
armab: As mentioned in other PR, changes to builds/CI/CD/Makefiles don't need Changelog record as it doesn't change the st2 core platform functionality.

cognifloyd: So, neither of these entries need to be in the changelog? Or just skip the "Fix python3 ..." one?

cognifloyd: I dropped both changelog entries.

PR #4869, opened 2020-02-19
Fix master builds

cognifloyd:
Fixes includes:

Update gitpython requirement (Thanks to @blag for tracking down the bug.)

- gitpython requires gitdb2, but that has removed some things in newer
  versions, and 2.11.1 had an open ended dep. So, update gitpython to a
  newer version that has more specific version requirements for gitdb2.
- This should fix these errors:
  https://github.com/gitpython-developers/GitPython/issues/983
  https://travis-ci.org/StackStorm/st2/jobs/652088774#L9416
  https://circleci.com/gh/StackStorm-Exchange/stackstorm-sumologic/155

Drop --no-site-packages from virtualenv calls

Make travis use the travis user instead of stanley, and add a restricted permissions workaround for the root user.

Update requests which had a conflicting dep on idna (primary said `idna>=2.5,<2.9` but security extra wanted any `idna>=2.0.0`. idna 2.9 was released a couple of days ago, and requests 2.23.0 with fixed deps was released today.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,9 @@ virtualenv-components
 virtualenv-components-osx
 .venv-st2devbox
 
+# generated travis conf
+conf/st2.travis.conf
+
 # Installer logs
 pip-log.txt
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,8 @@ env:
     # We only run tests with "--with-timer" flag on master and not for PRs since it adds 1-2
     # # minutes of overhead to each build.
     - NOSE_TIME=$([ "${TRAVIS_PULL_REQUEST}" = "false" ] && [ "${IS_NIGHTLY_BUILD}" = "no" ] && echo "yes" || echo "no")
+    # Travis-specific st2.conf (with travis user instead of stanley)
+    - ST2_CONF=conf/st2.travis.conf
 matrix:
   include:
     # NOTE: We combine builds because Travis offers a maximum of 5 concurrent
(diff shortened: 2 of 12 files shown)


cognifloyd:
This won't fix all of the build errors on master. There are other issues as well, this is just the only one with a clear fix so far.

cognifloyd:
Other issues include:
- something installing wrong idna version (`.st2client-install-check` in the Lint Checks build; can't find what changed here)
- build pack virtualenv fails (in 2.7 Integation Tests build. Maybe due to permissions issue? Build has no details)

Once we figure those issues out, they should probably go in separate PRs. Can we merge this fix even without resolving those other issues?

m4dcoder:
@cognifloyd If these issues are causing the errors in the CI above, then you have to fix them before we can merge.

blag:
NAK - investigating, will edit this comment in a bit

cognifloyd:
I just cherry-picked the `--no-site-packages` fix from #4863. 
Plus I'm trying to resolve the permissions issue in a nicer way than the xenial workaround to see if that fixes the 2.7 integration tests.
Still stumped on the idna issue.

cognifloyd:
It's GREEN! :green_heart:

armab:
It's arguable pip requirements fix needs _that_ amount of changes. I'd prefer to not mix different changes in one PR. This keeps history clear and straightforward.

Additionally, per previous comment we're missing CHANGELOG record which is required for pip requirements update for st2 core.

Even though, many thanks for fixing master! :heavy_check_mark:

cognifloyd:
@blag asked me to get rid of the 777 permissions workaround affecting the entire home directory.
I'm not sure why precise started having permissions issues, but I needed to adjust something, so I moved some of the permissions changes from #4863 to this PR to resolve that, only I made sure to do something that would selectively change permissions only where needed.

So, no, this was not just a pip dep update.

Changelog PR coming soon.

PR #4875, opened 2020-02-24
Rollback field changes on workflow and task execution collection

m4dcoder:
Rollback fields from DictField to EscapedDictField because in various use cases, result may contain dictionary which IP address as key. DictField does not allow "." in the key name. The execution result may be published into the context and into workflow output and then passed as input into subworkflow. Appropriate fields are rolled back to EscapedDictField to avoid breaking existing user workflows.


blag:
I'm glad we caught this before any users found it. 👍

blag:
Rolls back part of #4854.

blag:
Related: #4892.

PR #4892, opened 2020-03-21
Rollback workflow DB change on errors field

m4dcoder:
Rollback workflow DB change on errors field because there are cases where action execution result is included in the detail of the error log and the action execution result can contain dictionary with key that has dot notation such as IP address. Also identified and fix mongoescape bug on handling list of dicts in dynamic field and also multiple iteration on unescape.

### DIFF
--- a/st2common/st2common/util/mongoescape.py
+++ b/st2common/st2common/util/mongoescape.py
@@ -49,8 +49,9 @@ def _translate_chars_in_list(field, translation):
 
 
 def _translate_chars_in_key(key, translation):
-    for k, v in [(k, v) for k, v in six.iteritems(translation) if k in key]:
-        key = key.replace(k, v)
+    for k, v in six.iteritems(translation):
+        if k in key:
+            key = key.replace(k, v)
 
     return key
 
--- a/st2common/st2common/util/mongoescape.py
+++ b/st2common/st2common/util/mongoescape.py
@@ -19,70 +19,62 @@ from six.moves import zip
 
 from st2common.util.ujson import fast_deepcopy
 
+# Note: Because of old rule escaping code, two different characters can be translated back to dot
+RULE_CRITERIA_UNESCAPED = ['.']
+RULE_CRITERIA_ESCAPED = [u'\u2024']
+RULE_CRITERIA_ESCAPE_TRANSLATION = dict(list(zip(RULE_CRITERIA_UNESCAPED, RULE_CRITERIA_ESCAPED)))
+RULE_CRITERIA_UNESCAPE_TRANSLATION = dict(list(zip(RULE_CRITERIA_ESCAPED, RULE_CRITERIA_UNESCAPED)))
+
 # http://docs.mongodb.org/manual/faq/developers/#faq-dollar-sign-escaping
 UNESCAPED = ['.', '$']
 ESCAPED = [u'\uFF0E', u'\uFF04']
 ESCAPE_TRANSLATION = dict(list(zip(UNESCAPED, ESCAPED)))
-UNESCAPE_TRANSLATION = dict(list(zip(ESCAPED, UNESCAPED)))
-
-# Note: Because of old rule escaping code, two different characters can be translated back to dot
-RULE_CRITERIA_UNESCAPED = ['.']
-RULE_CRITERIA_ESCAPED = [u'\u2024']
-RULE_CRITERIA_ESCAPE_TRANSLATION = dict(list(zip(RULE_CRITERIA_UNESCAPED,
-                                            RULE_CRITERIA_ESCAPED)))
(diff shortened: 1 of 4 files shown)


blag:
Rolls back part of #4854.

blag:
It seems like we could have more escaping and unescaping tests in `st2common/tests/unit/test_mongoescape.py`, but if you feel that your tests adequately cover all of the usecases, then that's fine.

The other thing to consider is that the code in `test_mongoescape.py` would be easier to debug than all of this.

m4dcoder:
I can add my test case to st2common/tests/unit/test_mongoescape.py.  But i'll keep my specific test for the workflow engine.  The basic tests in test_mongoescape.py didn't catch the use case for the workflow engine and definitely didn't catch the bug in the unescape where it's translating twice but not passing the translated value from one _translate_chars to the second _translate_chars.  Also, the second _translate_chars is actually redundant add adds performance overhead.

blag:
The second call isn't redundant, it just very much looks like it is. The `_translate_chars` function mutates its `field` parameter, which is why calling it twice like that didn't break the tests.

Still very weird and overdue for a refactoring like this.

I believe this code is also an excellent candidate to be implemented with [single dispatch](https://docs.python.org/3/library/functools.html#functools.singledispatch):

```python
from functools import singledispatch

# ...


@singledispatch
def _translate_chars(field, translation):
    return field


@_translate_chars.register(list)
def _translate_chars_in_list(field, translation):
    return [_translate_chars(value, translation) for value in field]


def _translate_chars_in_key(key, translation):
    for k, v in [(k, v) for k, v in six.iteritems(translation) if k in key]:
        key = key.replace(k, v)
    return key


@_translate_chars.register(dict)
def _translate_chars_in_dict:
    return {
        _translate_chars_in_key(k, translation): _translate_chars(v, translation)
        for k, v in six.iteritems(field)
    }


def escape_chars(field):
    return _translate_chars(fast_deepcopy(field), ESCAPE_TRANSLATION)


def unescape_chars(field):
    return _translate_chars(fast_deepcopy(field), UNESCAPE_TRANSLATION)
```

blag:
@m4dcoder For your consideration: #4894.

st2common/st2common/util/mongoescape.py
```diff
@@ -19,70 +19,62 @@
...
-        oldkey = work_item[0]
-        value = work_item[1]
-        work_field = work_item[2]
-        newkey = oldkey
 
-        for t_k, t_v in six.iteritems(translation):
-            if t_k in newkey:
-                newkey = newkey.replace(t_k, t_v)
+def _translate_chars_in_key(key, translation):
+    for k, v in [(k, v) for k, v in six.iteritems(translation) if k in key]:
```
nmaludy: If we're worried about performance, i think this might be able to be improved.

Currently the `[(k, v) for k, v in six.iteritems(translation) if k in key]` iterates over `key` either 1x (for escape) or 3x for unescape. Then, for every character it iterates over `key` again, so potentially another 1x or 3x times. `O(n^3)`. Also, every time a replace is done a new copy of a `key` is made.

Seems like we could do some things to make this better?
* Google lead me here: https://stackoverflow.com/questions/10017147/removing-a-list-of-characters-in-string
* Also, maybe if we simply changed `for k, v in [(k, v) for k, v in six.iteritems(translation) if k in key]:` to `for k, v in six.iteritems(translation):` that might help?

m4dcoder: @nmaludy i'm not sure if you missed the if condition for the inner loop. Anyway, went back to the original use of the for loop and if.

```
def _translate_chars_in_key(key, translation):
    for k, v in six.iteritems(translation):
        if k in key:
            key = key.replace(k, v)

    return key
```

m4dcoder: The original method will be definitely worst.  I consolidated two calls of _translate_chars into one in the unescape (and also fixed a bug there).  The list/dict comprehension is going to be less logic on deleting keys and append new dict entry.  A ran a number of orquesta integration tests between master and this branch and I am seeing completion time reduced with the change.

m4dcoder: Also, the string.translate method doesn't work here because the translation is a 1:1 character translation.  The translation here is 1:many or many:1 characters.

blag:
As of commit `dbf78d4`, this PR should be good to go. It is the fastest of the four options (`for` loop vs. list comprehension and explicit switch vs. single dispatch decorator).

st2common/st2common/util/mongoescape.py
blag: See my benchmarks [here](https://github.com/StackStorm/st2/pull/4894#issuecomment-603580440).

This PR (now) implements the fastest versions of the options presented so far (eg: the explicit `for` loop and explicit switch on `field` type). Resolving.

PR #4894, opened 2020-03-24
Refactor fix-workflow-db to use singledispatch

blag:
Refactor the `_translate_chars` to use single dispatch with the `functools.singledispatch` decorator on Python 3 or the `singledispatch` package on Python 2.

Issue #4681, opened 2019-05-16
PIP Upgrade


bigmstone:
When PIP upgraded to 10 it caused a lot of issues. We pinned to 9.x. Needs to be reassessed now.

armab:
For the history: https://github.com/StackStorm/st2-packages/pull/549

armab:
Because `st2` repo during these months already incorporated env fixes related to py3, pip and dist_utils,
the remaining work just took 1 fix in `st2-packages` and another 10 PRs to update pip in different repos.

We're now on latest pip `v19` instead of old `v9`.

Issue #3596, opened 2017-07-21
EL6/EL7 failure 'SSL3_GET_SERVER_CERTIFICATE' after recent pip upgrade [staging-unstable]


armab:
This affects `staging-unstable` packages for now. Recent massive pip versions upgrade (#3538) broke `EL6`/`EL7` builds with e2e tests with the following error:

```
[root@st2-ent-pkg-unstable-el7 stanley]# st2 run core.http url=https://google.com/
.
id: 5971d6acd549c538fa717614
status: failed
parameters: 
  url: https://www.google.com
result: 
  error: '("bad handshake: Error([(''SSL routines'', ''SSL3_GET_SERVER_CERTIFICATE'', ''certificate verify failed'')],)",)'
  traceback: "  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/st2actions/container/base.py\", line 105, in _do_run\n    (status, result, context) = runner.run(action_params)\n  File \"/opt/stackstorm/runners/http_runner/http_runner.py\", line 93, in run\n    result = client.run()\n  File \"/opt/stackstorm/runners/http_runner/http_runner.py\", line 217, in run\n    verify=self.verify\n  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/requests/api.py\", line 58, in request\n    return session.request(method=method, url=url, **kwargs)\n  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/requests/sessions.py\", line 502, in request\n    resp = self.send(prep, **send_kwargs)\n  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/requests/sessions.py\", line 612, in send\n    r = adapter.send(request, **kwargs)\n  File \"/opt/stackstorm/st2/lib/python2.7/site-packages/requests/adapters.py\", line 514, in send\n    raise SSLError(e, request=request)\n"
```

After reverting back to revision before #3538 upgrade everything is OK:
```
[root@st2-ent-pkg-unstable-el7 stanley]# st2 run core.http url=https://google.com/
.
id: 5971d94ed549c543310da786
status: failed
parameters: 
  url: https://google.com/
result: 
[...]

PR #4895, opened 2020-03-24
Various requirements fixups

blag:
# Background #

I wanted to specify `python_version < '3.4'` for the `singledispatch` dependency in #4894, but the `fixate-requirements.py` script does not support what pip calls "markers". Markers are environment specifications that instruct pip when to install, or more importantly, when _not_ to install a dependency.

Here's what a Python version marker would look like for #4894:

```
singledispatch==3.4.0.3 ; python_version < '3.4'
```

That marker would only install the `singledispatch` dependency if installing for Python < v3.4. The `singledispatch` package is a backport of the same feature for Python 3.4+, available in the standard library in the `functools` package. So installing it for Python 3.4+ is not necessary.

# Description #

This PR:

* [x] Upgrades `fixate-requirements.py` to write out dependency markers if present in `fixed-requirements.txt` or `in-requirements.txt`.
* [x] Tweaks some `Makefile` output to be more readble
* [x] ~Adjusts some `Makefile` dependencies so multiple `make` invocations aren't constantly ping-ponging between different versions of pip, setuptools, etc. (read: speeds up the build)~
* [x] Alphabetizes `fixed-requirements.txt` (in two groups) so it's easier to visually search the file for a dependency by name
* [x] (unrelated) Fixed backtick syntax for ReStructured Text (eg: not Markdown)

# Future Work #

I ran `make requirements` after these changes to ensure that the behavior of the script was no different than before. And I will be adding the environment marker to #4894 if this is merged in before that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -328,9 +328,9 @@ flake8: requirements .flake8
 	touch $(VIRTUALENV_ST2CLIENT_DIR)/bin/activate
 	chmod +x $(VIRTUALENV_ST2CLIENT_DIR)/bin/activate
 
-	$(VIRTUALENV_ST2CLIENT_DIR)/bin/pip install --upgrade "pip==19.3.1"
+	$(VIRTUALENV_ST2CLIENT_DIR)/bin/pip install --upgrade "pip>=19.3.1"
 	# NOTE We need to upgrade setuptools to avoid bug with dependency resolving in old versions
-	$(VIRTUALENV_ST2CLIENT_DIR)/bin/pip install --upgrade "setuptools==41.0.1"
+	$(VIRTUALENV_ST2CLIENT_DIR)/bin/pip install --upgrade "setuptools>=42"
 	$(VIRTUALENV_ST2CLIENT_DIR)/bin/activate; cd st2client ; ../$(VIRTUALENV_ST2CLIENT_DIR)/bin/python setup.py install ; cd ..
 	$(VIRTUALENV_ST2CLIENT_DIR)/bin/st2 --version
 	$(VIRTUALENV_ST2CLIENT_DIR)/bin/python -c "import st2client"
--- a/Makefile
+++ b/Makefile
@@ -505,7 +505,9 @@ requirements: virtualenv .sdist-requirements install-runners
 	git submodule update --recursive --remote
 
 	# Verify there are no conflicting dependencies
-	$(VIRTUALENV_DIR)/bin/pipconflictchecker
+	cat st2*/requirements.txt contrib/runners/*/requirements.txt | sort -u > req.txt && \
+	$(VIRTUALENV_DIR)/bin/pip-compile req.txt; \
+	if [[ -e req.txt ]]; then rm req.txt; fi
 
 .PHONY: virtualenv
 	# Note: We always want to update virtualenv/bin/activate file to make sure
--- a/Makefile
+++ b/Makefile
@@ -446,10 +446,10 @@ requirements: virtualenv .sdist-requirements install-runners
 	# .st2client-install-check target and .travis.yml to match
 	# Make sure we use latest version of pip
 	$(VIRTUALENV_DIR)/bin/pip --version
-	$(VIRTUALENV_DIR)/bin/pip install --upgrade "pip==19.3.1"
+	$(VIRTUALENV_DIR)/bin/pip install --upgrade "pip>=19.3.1"
 	# setuptools >= 41.0.1 is required for packs.install in dev envs
 	# setuptools >= 42     is required so setup.py install respects dependencies' python_requires
-	$(VIRTUALENV_DIR)/bin/pip install --upgrade "setuptools==41.0.1"  # Required for packs.install in dev envs
+	$(VIRTUALENV_DIR)/bin/pip install --upgrade "setuptools>=42"
 	$(VIRTUALENV_DIR)/bin/pip install --upgrade "pbr==5.4.3"  # workaround for pbr issue
 
 	# Generate all requirements to support current CI pipeline.
--- a/Makefile
+++ b/Makefile
@@ -453,9 +453,10 @@ requirements: virtualenv .sdist-requirements install-runners
 	@for component in $(COMPONENTS_WITH_RUNNERS); do\
 		echo "==========================================================="; \
 		echo "Generating requirements.txt for" $$component; \
-		echo "==========================================================="; \
 		$(VIRTUALENV_DIR)/bin/python scripts/fixate-requirements.py --skip=virtualenv,virtualenv-osx -s $$component/in-requirements.txt -f fixed-requirements.txt -o $$component/requirements.txt; \
 	done
+
+	@echo "==========================================================="
 
 	# Fix for Travis CI race
 	$(VIRTUALENV_DIR)/bin/pip install "six==1.12.0"
--- a/Makefile
+++ b/Makefile
@@ -439,8 +439,8 @@ requirements: virtualenv .sdist-requirements install-runners
 	@echo
 	# Make sure we use latest version of pip which is 19
 	$(VIRTUALENV_DIR)/bin/pip --version
-	$(VIRTUALENV_DIR)/bin/pip install --upgrade "pip==19.3.1"
-	$(VIRTUALENV_DIR)/bin/pip install --upgrade "setuptools==41.0.1"  # Required for packs.install in dev envs
+	$(VIRTUALENV_DIR)/bin/pip install --upgrade "pip>=19.3.1"
+	$(VIRTUALENV_DIR)/bin/pip install --upgrade "setuptools>=41.0.1"  # Required for packs.install in dev envs
 	$(VIRTUALENV_DIR)/bin/pip install --upgrade "pbr==5.4.3"  # workaround for pbr issue
 
 	# Generate all requirements to support current CI pipeline.
--- a/Makefile
+++ b/Makefile
@@ -439,8 +439,8 @@ requirements: virtualenv .sdist-requirements install-runners
 	@echo
 	# Make sure we use latest version of pip which is 19
 	$(VIRTUALENV_DIR)/bin/pip --version
-	$(VIRTUALENV_DIR)/bin/pip install --upgrade "pip>=19.3.1"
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -439,8 +439,8 @@ requirements: virtualenv .sdist-requirements install-runners
 	@echo
 	# Make sure we use latest version of pip which is 19
 	$(VIRTUALENV_DIR)/bin/pip --version
-	$(VIRTUALENV_DIR)/bin/pip install --upgrade "pip==19.3.1"
-	$(VIRTUALENV_DIR)/bin/pip install --upgrade "setuptools==41.0.1"  # Required for packs.install in dev envs
+	$(VIRTUALENV_DIR)/bin/pip install --upgrade "pip>=19.3.1"
```
armab: Are both pip and setuptools version change required for this change?

We pinned both pip and setuptools in many places for the reasons based on experience of bad surprises.
If there is a motivation to upgrade it, - please pin to specific version and this should be changed in many repos across the org, incl. packaging and build Dockerfiles.

blag: Fixed.

blag:
Note that the Travis tests are actually passing. The GitHub test status is not getting updated with the new results.

Makefile
punkrokk: I’d be a fan up upgrading pip and setup tools right after release to work out bugs, I agree with @arma they are pinned for a reason, but also setup tools v20.0.x is way better.

armab: I agree. Let's create an issue in `stackstorm/st2` to track upgrading both pip + virtualenvs across the repos in next release and as another series of PRs.

blag:
There is a list of markers in [PEP 508](https://www.python.org/dev/peps/pep-0508/#environment-markers):


| Marker | Python equivalent | Sample values |
| ------- | ------------------ | -------------- |
| `os_name` | `os.name` | `posix`, `java` |
| `sys_platform` | `sys.platform` | `linux`, `linux2`, `darwin`, `java1.8.0_51` (note that "linux" is from Python3 and "linux2" from Python2) |
| `platform_machine` | `platform.machine()` | `x86_64` |
| `platform_python_implementation` | `platform.python_implementation()` | `CPython`, `Jython` |
| `platform_release` | platform.release()` | `3.14.1-x86_64-linode39`, `14.5.0`, `1.8.0_51` |
| `platform_system` | platform.system()` | `Linux`, `Windows`, `Java` |
| `platform_version` | `platform.version()` | `#1 SMP Fri Apr 25 13:07:35 EDT 2014`,<br>`Java HotSpot(TM) 64-Bit Server VM`,<br>`25.51-b03, Oracle Corporation`,<br>`Darwin Kernel Version 14.5.0: Wed Jul 29 02:18:53 PDT 2015; root:xnu-2782.40.9~2/RELEASE_X86_64` |
| `python_version` | `'.'.join(platform.python_version_tuple()[:2]` | `3.4`, `2.7` |
| `python_full_version` | `platform.python_version()` | `3.4.0`, `3.5.0b1` |
| `implementation_name` | `sys.implementation.name` |` cpython` |
| `implementation_version` | see definition below | `3.4.0`, `3.5.0b1` |
| `extra` | An error except when defined by the context interpreting the specification. | `test` |

The `implementation_version` marker variable is derived from `sys.implementation.version`:

```python
def format_full_version(info):
    version = '{0.major}.{0.minor}.{0.micro}'.format(info)
    kind = info.releaselevel
    if kind != 'final':
        version += kind[0] + str(info.serial)
    return version

if hasattr(sys, 'implementation'):
    implementation_version = format_full_version(sys.implementation.version)
else:
    implementation_version = "0"
```

blag:
With #4894 not being merged this isn't strictly necessary. And with ST2 v3.3 dropping support for Python 2.7, we won't have to worry as much about Python versions when specifying our dependencies.

However, this PR does support a fairly new part of the `requirements.txt` standard, and it could still be useful in the future.

For instance, while StackStorm doesn't run on macOS, we could at least build the `st2` client on macOS, and we could at least help `make requirements` run successfully on macOS by making the [`pyinotify` dependency](https://github.com/StackStorm/st2/blob/b844a8110f062f7f8e2ccc976db172eb1e8c20e8/requirements.txt#L43) only for Linux.

Currently when you try to run `make requirements` on macOS you get:

```bash
Collecting pyinotify
  Using cached https://files.pythonhosted.org/packages/e3/c0/fd5b18dde17c1249658521f69598f3252f11d9d7a980c5be8619970646e1/pyinotify-0.9.6.tar.gz
    ERROR: Command errored out with exit status 1:
     command: .../st2/venv/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/var/folders/ls/_2q7p19s2n1c3xdnphq4btzr0000gn/T/pip-install-ctf42xh_/pyinotify/setup.py'"'"'; __file__='"'"'/private/var/folders/ls/_2q7p19s2n1c3xdnphq4btzr0000gn/T/pip-install-ctf42xh_/pyinotify/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base pip-egg-info
         cwd: /private/var/folders/ls/_2q7p19s2n1c3xdnphq4btzr0000gn/T/pip-install-ctf42xh_/pyinotify/
    Complete output (1 lines):
    inotify is not available on macosx-10.9-x86_64
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

But if we marked `pyinotify` as being only for Linux:

```requirements.txt
pyinotify==0.9.6 ; platform_system == 'Linux'
```

Then we can at least run `make requirements` to completion:

```bash
Ignoring pyinotify: markers 'sys_platform == "Linux"' don't match your environment
```

blag:
Waiting on https://gitlab.com/python-devs/importlib_metadata/-/merge_requests/118 to stop breaking the build.

Makefile
```diff
@@ -328,9 +328,9 @@ flake8: requirements .flake8
 	touch $(VIRTUALENV_ST2CLIENT_DIR)/bin/activate
 	chmod +x $(VIRTUALENV_ST2CLIENT_DIR)/bin/activate
 
-	$(VIRTUALENV_ST2CLIENT_DIR)/bin/pip install --upgrade "pip==19.3.1"
+	$(VIRTUALENV_ST2CLIENT_DIR)/bin/pip install --upgrade "pip>=19.3.1"
```
nmaludy: Any reason we're doing >= here instead of pinning?

blag: Good question, the answer is: yes! Pinning these tools to specific versions has caused build failures a few times now, and some of our dependencies are relying on newer and newer versions of tools. So continuing to piecemeal pin versions is just going to cause us more breakage in the future.

Instead of fixing this once a month for the next year, I'm just relaxing the version criteria to allow us to use the most recent versions of those tools. Once we release v3.2 and totally drop support for Python 2.7, we can tighten these requirements back up, because by that point, we will probably stop seeing as much breakage.

nmaludy: I'm fine with this. I'll let other comment though, just in case there are any other concerns.

blag: Yep, I think @armab might have an opinion on this.

I'll also be adding comments and adjusting all of the pip and setuptools versions to match these requirements as well.

blag:
Expands on the work done in StackStorm/st2#4750.

Makefile
armab: :-1: for this approach.

As mentioned in the past, if we want to switch to specific version of pip, setuptools, etc, etc dependencies stackstorm relies on, - that should be done in dedicated PR and across of the other repos: there are a bunch of other places. Inconsistencies in versions result in hard to catch bugs and builds that have less repeatability.

Additionally, relaxing version requirements and relying automatically on latest version of these tools led us to more problems before which are bigger comparing to py2-related issues @blag was mentioning. This is rarely a good approach as it leads to bugs hard to spot due to upstream updates. When upstream updates the behavior, introduces breaking changes and we automatically rely on that, - you newer know what's going on before (suddenly) hitting an issue in production and diving deep into the rabbit hole.

The way that worked better, - update versions periodically _everywhere_. This means periodic maintenance: reviewing what was changed in the new versions of the tools we rely on, test them, try them and then change version interval across all the repositories.

@StackStorm/maintainers WDYT about it?

-----

For wondering, just a couple of issues that we hit in the past:
- https://github.com/StackStorm/st2-packages/pull/549
- https://github.com/StackStorm/st2/issues/4681
- https://github.com/StackStorm/st2/issues/3596
- https://github.com/StackStorm/st2/issues/3595
How pip upgrade should be done:
- https://github.com/StackStorm/st2/pull/4712 (references many other repos)

I bet there are much more examples from @Kami who mostly worked with upgrading upstream versions.

Eg. we can't blindly follow latest version which is just a matter of time to hit us again.

Makefile
armab: Pinning to pip version interval and avoiding relying on latest automatically was part of my review/feedback for this PR.
Moving this pip version change from `Makefile` to the `.travis.yml` is not a good way to address that feedback.

I'm wondering why this comment was marked as _Resolved_.

blag:
Pulled in a few commits from #4896.
Closes #4896.

Makefile
blag: I changed this back for you and marked it as resolved, then re-reverted my change and forgot to unresolve it. Sorry!

But, as per our Slack conversation, I consider pinning to exact pip and setuptools versions to be a code smell, and I'm against doing so until we have exhausted and documented every other possible workaround.

armab: As said before, if you're planning the change which finds an issue to reach the consensus, - do that as a proposal discussion to change the existing behavior. If TSC votes for the new change, - it's best to plan it for the next release and update consistently across all the repositories.

Before that happens, the building mechanisms should persist, especially before the release where any major changes or experiments should be discouraged and codebase freezed.

Makefile
blag: I've opened StackStorm/discussions#2 to discuss this further.

Makefile
blag: Proposal discussion is in StackStorm/discussions#2.

I'll revert this back to what it was pending the conclusion of that discussion.

PR #4897, opened 2020-03-27
Don't test or build Ubuntu 14 packages

blag:
Waiting on StackStorm/st2cd#428 and StackStorm/st2ci#171 to be merged first.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -107,7 +107,7 @@ jobs:
 
   # Build & Test st2 packages
   packages:
-    parallelism: 6
+    parallelism: 5
     # 4CPUs & 8GB RAM CircleCI machine
     # sadly, it doesn't work with 'setup_remote_docker'
     resource_class: large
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -232,7 +232,7 @@ jobs:
       - image: circleci/ruby:2.4
     working_directory: /tmp/deploy
     environment:
-      - DISTROS: "trusty xenial bionic el6 el7"
+      - DISTROS: "xenial bionic el6 el7"
```
armab: This uncovered oversight in previous EL8 work.

Deployment should include `el8` as well:
```suggestion
      - DISTROS: "xenial bionic el6 el7 el8"
```

blag: I noticed that, but that's outside the scope of this PR.

armab: cc @punkrokk

punkrokk: Is there another PR then or do I need to make one?

armab: There is no PR for this kind of fix yet

blag: Captured in #4899.

PR #4900, opened 2020-04-03
Pin build dependencies to single version and update instructions

blag:
Follow up on #4895 to pin pip and setuptools dependencies to specific versions.

I picked the current latest versions because it's less work in the future.

I believe we will need to update these versions in many other repositories as well. This is simply the first one.

This PR fixes a bug in the `write-header.sh` script so it now correctly copies the "magic comment" lines before injecting its comment header. Editors should now respect the encoding of the generated `dist_utils.py` files. Additionally, I tweaked some of the behavior so it doesn't print lines that begin with `# //` (a C++-esque comment within a Python comment), and included instructions for re-generating all of the subpackage `dist_utils.py` files, since I was not aware that this needed to be done when I changed the master `dist_utils.py` script in #4895 and future developers may not know either. Ideally we would make this a CI lint check, but I don't think that's hugely necessary at this point.


.travis.yml
```diff
@@ -82,7 +82,7 @@ cache:
 before_install:
   # If you update these versions, make sure you update the versions in the .st2client-install-check
   # and the requirements targets in the Makefile to match
-  - pip install --upgrade "pip>=19.3.1"
+  - pip install --upgrade "pip==20.0.2"
```
armab, line 85: For the best results it's best to relax patch version by using interval.
This way we still benefit from the important bugfixes and security updates, excluding breaking changes coming in.

blag: This is fine, but we need the `Makefile` and the `travis.yml` file to have the exact same version specifier, or else the build "ping pongs" between installing one version in Travis, and then another during the build, and then another in Travis, and then another during the build, which slows the build down.

blag: I will revert this to an interval pending the conclusion of StackStorm/discussions#2.

PR #4904, opened 2020-04-08
try to support tooz in EL8/py3

punkrokk:
Problem: Tooz < 2.2.0 doesn't support consul heartbeat. @nmaludy Got it committed to tooz: https://github.com/openstack/tooz/commit/6e750b6921c42f3c2ffed2e135be4d9c65e96f36#diff-ac4db250c2e1f5b2d95e610996274b96.

I have tested this and updated fixed-requirements.txt as follows: 

```
tooz==1.66.1; python_version <= '2.7'
tooz==2.3.0; python_version >= '3.6'
```

The question is - will this build correctly? It seems to me (@blag ) that the make requirements script doesn't put both lines in the generated requirements files. 

Proof that it works: 
image

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -60,7 +60,7 @@ Changed
 * Use ``pip-compile`` from ``pip-tools`` instead of ``pip-conflict-checker`` (improvement) #4896
 * Refactor how inbound criteria for join task in orquesta workflow is evaluated to count by
   task completion instead of task transition. (improvement)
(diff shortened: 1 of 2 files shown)


armab:
Looks like CI is not happy with that requirements definition

armab (changes requested):
Considering we're Python 3 only now which unblocks this PR, please update the code to include the latest `tooz`.

PR #4911, opened 2020-04-17
[bugfix] fix string %s issue with unicode

misterpah:
configuration: 
1PPC

issue:
command crash when running `st2ctl reload --register-setup-virtualenvs` due to unicode and string (byte) string-replacement process for LOG.debug function.

commit message :
It is possible for shell to return unicode.
It's perfectly fine. Nothing wrong with that.
Issue happen when doing substitution.
A unicode cannot do substitution in python2's string (byte).
The solution is to separate it.

Signed-off-by: misterpah <misterpah@gmail.com>


st2common/st2common/util/pack_management.py
```diff
@@ -339,8 +339,8 @@ def apply_pack_owner_group(pack_path):
 
         if exit_code != 0:
             # Non fatal, but we still log it
-            LOG.debug('Failed to change owner group on directory "%s" to "%s": %s' %
-                      (pack_path, pack_group, stderr))
+            LOG.debug('Failed to change owner group on directory "{}" to "{}":'.format(pack_path, pack_group))
```
nmaludy, line 342: probably best to leave the logging format the same, in this case keeping `stderr` on the same line instead of another debug line.

nmaludy (changes requested):

st2common/st2common/util/pack_management.py
```diff
@@ -339,8 +339,8 @@ def apply_pack_owner_group(pack_path):
...
+            LOG.debug(stderr)
```
nmaludy, line 343:
```suggestion
            LOG.debug('Failed to change owner group on directory "{}" to "{}": {}'
                              .format(pack_path, pack_group, stderr))
```

nmaludy (approved):

PR #4912, opened 2020-04-18
Allow user to retrieve more than one resource using "st2 resource get" CLI command

Kami:
This pull request addresses one of my biggest minor annoyances with StackStorm CLI.

A lot of times when I'm debugging / troubleshooting things, I need to view multiple resources at once to see what is going on.

This means that currently I need to run more than one command to achieve that.

For example:

```bash
st2 execution get id1
st2 execution get id2
st2 execution get id3
```

This slows things down and means more typing (yes, I can use a bash alias or function, but that's not ideal and it's not available out of the box).

With this change, user can now view more than one resource using the get command.

The same pattern is already available with ``st2 execution cancel`` command (I added it a while ago).

For example:

```bash
st2 execution get id1 id2 id3
```

Example output:

```bash
st2 action get packs.unload packs.show
+---------------+--------------------------------------+
| Property      | Value                                |
+---------------+--------------------------------------+
| id            | 5e9ae3535550f016857e19f9             |
| uid           | action:packs:unload                  |
| ref           | packs.unload                         |
| pack          | packs                                |
| name          | unload                               |
| description   | Unregisters all content from a pack. |
| enabled       | True                                 |
| entry_point   | pack_mgmt/unload.py                  |
| runner_type   | python-script                        |
| parameters    | {                                    |
|               |     "packs": {                       |
|               |         "items": {                   |
|               |             "type": "string"         |
|               |         },                           |
|               |         "required": true,            |
|               |         "type": "array"              |
|               |     }                                |
|               | }                                    |
| metadata_file | actions/unload.yaml                  |
| notify        |                                      |
| output_schema |                                      |
| tags          |                                      |
+---------------+--------------------------------------+
| id            | 5e9ae3535550f016857e19f7                                     |
| uid           | action:packs:show                                            |
| ref           | packs.show                                                   |
| pack          | packs                                                        |
| name          | show                                                         |
| description   | Get detailed information about pack from the remote          |
|               | StackStorm exchange index.                                   |
| entry_point   | pack_mgmt/show_remote.py                                     |
| runner_type   | python-script                                                |
| parameters    | {                                                            |
|               |     "pack": {                                                |
|               |         "required": true,                                    |
|               |         "type": "string",                                    |
|               |         "description": "Name of pack to lookup"              |
|               |     }                                                        |
|               | }                                                            |
| metadata_file | actions/show.yaml                                            |
| notify        |                                                              |
| output_schema |                                                              |
| tags          |                                                              |
```

## Open Questions / To Decide

Currently when a single source id is provided, the command behaves exactly in the same manner as it did before - it's fully backward compatible.

When retrieving multiple resources, if one of the provided resources is not found, we simply print that, but we don't immediately exit with non-zero status code and print other resources which were found.

As far as using the existing command name goes - I think that's better than adding yet another command (e.g. ``st2 <resource> get-more`` or similar) - the whole idea is that it's simple and it works out of the box so any other change which would make it less simple (e.g. new argument or filter to list command, etc.) is a no go.

## TODO

- [x] (Once agreed on the approach) Implement it for other non-resource "get" commands
- [x] (Once agreed on the approach) Add tests

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -257,6 +257,22 @@ Improvements
 
   Contributed by @cognifloyd.
 
+* Update majority of the "resource get" CLI commands (e.g. ``st2 execution get``,
+  ``st2 action get``, ``st2 rule get``, ``st2 pack get``, ``st2 apikey get``, ``st2 trace get``,
+  ``st2 key get``, ``st2 webhook get``, ``st2  timer get``, etc.) so they allow for retrieval
+  and printing of information for multiple resources using the following notation:
+  ``st2 <resource> get <id 1> <id 2> <id n>``, e.g. ``st2 action.get pack.show packs.get
+  packs.delete``
+
+  This change is fully backward compatible when retrieving only a single resource (aka single
+  id is passed to the command).
+
+  When retrieving a single source the command will throw and exit with non-zero if a resource is
+  not found, but when retrieving multiple resources, command will just print an error and
+  continue with printing the details of any other found resources. (new feature) #4912
+
+  Contributed by @Kami.
+
 Fixed
 ~~~~~
 
--- a/st2actions/st2actions/notifier/config.py
+++ b/st2actions/st2actions/notifier/config.py
@@ -21,8 +21,6 @@ import st2common.config as common_config
 from st2common.constants.system import VERSION_STRING
 from st2common.constants.system import DEFAULT_CONFIG_FILE_PATH
 
-common_config.register_opts()
-
 CONF = cfg.CONF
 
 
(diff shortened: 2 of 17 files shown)


Kami:
I've pushed a change which adds this support for all the "resource get" commands.

In addition to that, while working on this change, I cleaned up the code a bit - removed some overriden duplicated methods in sub-classes which are the same as in parent classes, made sure we don't re-hrow ``ResourceNotFoundError`` twice in some situations (no need to re-throw it, we can just propagate the original error), fixed inconsistency for "resource not found" errors - most commands would throw ``ResourceNotFoundError``, but a couple would throw / propagate original ``OperationFailureException``.

Kami:
I tried to also add some end to end tests, but sadly I don't have many hours to deal with and wait on end to end tests.

So I added some more unit tests which verify this new notation works for all the commands - 986989f570a26646f92ceafcfc447dcc7dc6f531.

Kami:
OK, this is now synced with latest master.

Resolving merge conflicts was too painful due to black changes, so I just manually re-applied my old changes on top of latest master.

cognifloyd (approved):

Issue #4715, opened 2019-06-20
SSH connect timeout is too long (60s)


tclh123:
The st2 ssh connect timeout in st2 ssh_runner is 60s and it is not configurable.
It is too long - the ansibile default is [10s](https://docs.ansible.com/ansible/2.3/intro_configuration.html#timeout).

We better make it configurable and give it a more suitable default value.

armab:
Thanks for the report.

To make sure we're talking about the same timeout, how did you identify it's `60s`?

tclh123:
Hi @armab 
It's hardcoded here https://github.com/StackStorm/st2/blob/master/st2common/st2common/runners/paramiko_ssh.py#L84.

I make my own runner to bypass this problem.
By the way, I have another doubt - the `parallel` option is set to False and immutable, so how can users use this runner in parallel? It's weird.

armab:
Thanks for more info @tclh123, indeed configurable connect timeout is something that could be improved.
I think it makes sense to expose it via `st2.conf` settings and https://github.com/StackStorm/st2/blob/0b833228e7d63d7c0feaa96d32f3d3defaf3ab1e/conf/st2.conf.sample#L293-L303 looks like an ideal place.

We accept PRs/contributions so you're motivated to enhance that, - we're open to it.

winem:
Just a short notice on this topic to avoid redundant work: 

I am working on a fix. Would call it code-complete and it just needs a bit more of testing. PR will follow very soon.

PR #4913, opened 2020-04-18
Feature request #4715: make ssh connect timeout configurable

winem:
This PR adds support for `ssh_connect_timeout` as new configuration parameter in the `ssh_runner` group. It was suggested in #4715 and follows the _fail-fast principle_ when the SSH connection can not even be established.

The runner will fail if the TCP connect to the target host does not succeed within the specified timeout. (see: http://docs.paramiko.org/en/stable/api/client.html -> SSHClient -> connect -> timeout). 

This was tested with a high number of test cases with different defaults, ssh_connect_timeouts in the st2.conf and a variation of action timeouts.


winem:
Oh sorry, I'll close this one. It was outdated and the update from the master makes the history and overview of changes kinda messy. Will create a clean one in a few seconds.

PR #4914, opened 2020-04-18
Feature request #4715: configurable ssh connect timeout

winem:
This PR adds support for ssh_connect_timeout as new configuration parameter in the ssh_runner group. It was suggested in #4715 and follows the fail-fast principle when the SSH connection can not even be established.

The runner will fail if the TCP connect to the target host does not succeed within the specified timeout. (see: http://docs.paramiko.org/en/stable/api/client.html -> SSHClient -> connect -> timeout).

This was tested with a high number of test cases with different defaults, ssh_connect_timeouts in the st2.conf and a variation of action timeouts.

This one makes #4913 which I just closed obsolete.


CHANGELOG.rst
```diff
@@ -22,7 +22,9 @@ Fixed
 
 Added
 ~~~~~
-
+* Add support for a configurable connect timeout for SSH connections as requested in #4715
+  by adding the new configuration parameter ``ssh_connect_timeout`` to the ``ssh_runner``
+  group in st2.conf.
```
armab, line 27: Please move the record up to `In Development` section of Changelog and also include PR # number.
If you want, it's also good to add a credit in Changelog line.

winem:
I thought about the unit tests as well and back then I just knew how to test it manually. It's probably a bit tricky but I think I just got an idea for a proper unit test that can differentiate between failures due to breaching the `timeout` and the `ssh_connect_timeout`. I'll work on it over the weekend.

CHANGELOG.rst
winem: Thanks for the hint and done.

winem:
Marked as draft because a merge of the master branch into my feature branch went wrong. Will provide the proper solution before end of this week. 

In the meantime any input regarding my previous comment and the unit tests is highly appreciated!

winem:
> Are there any unit tests we can add for this new PR enhancement?

Hi @armab, I thought about the tests a bit more and am not sure what a proper implementation would look like. I know that some people use google.com:81 in tests to run a timeout but I think that every request to foreign infrastructure should be avoided (and is just the wrong way). 

Only thing that seems to be an option is to use Mocks side_effects and raise an exception. In this case we would see the exception but it's just a mocked one and does not actually prove that the action failed after the ssh_connect_timeout which would have to be lower than the action timeout in this case.

So I ran out of ideas but am all in for ideas and every input is appreciated. Please let me know if you have any ideas or any hints for something I could look into. 

Cheers,
Marcel

CHANGELOG.rst
```diff
@@ -3,6 +3,12 @@ Changelog
 
 in development
 --------------
+Added
```
armab, line 6: Looks like we have 2 `Added` blocks snow.
Please move the Changelog record under the existing `Added` section.

winem: Done.

CHANGELOG.rst
```diff
@@ -10,6 +10,10 @@ Added
   to ensure update to the models require schema to be regenerated. (new feature)
 * Improved st2sensor service logging message when a sensor will not be loaded when assigned to a
   different partition (@punkrokk)
+* Add support for a configurable connect timeout for SSH connections as requested in #4715
+  by adding the new configuration parameter ``ssh_connect_timeout`` to the ``ssh_runner``
+  group in st2.conf.
+  This option was requested by Harry Lee (@tclh123) and contributed by Marcel Weinberg (@winem).
```
armab, line 16: Best way to verify your changelog format is by comparing it with the records from previous releases.

```suggestion
* Add support for a configurable connect timeout for SSH connections as requested in #4715
  by adding the new configuration parameter ``ssh_connect_timeout`` to the ``ssh_runner``
  group in st2.conf. (new feature) #4914
  
  This option was requested by Harry Lee (@tclh123) and contributed by Marcel Weinberg (@winem).
```

It looks like we got several misleading changelog records in this release which show consistency issues cc @punkrokk.

winem: Done. Sorry, will pay more attention to the details next time.

armab: looks like there is still a diff.

winem: Done. Forgot to push again.

armab: Thanks! :+1:

winem:
Anytime! What do you think about the (missing) tests? Do you see any chance to cover this feature by a test?

armab:
Currently I can't help much here, but I really appreciate the effort trying to add the test scenarios. If nothing worked, - I'm good to merge it as is.

PR #4915, opened 2020-04-21
Add st2-generate-schemas to autogen schemas for the content models

m4dcoder:
Add the script st2-generate-schemas to autogen schemas for the content models such as pack, action, alias, policy, and rule. The schemas will be used in developer workflow to help with user generated contents.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -271,6 +271,16 @@ configgen: requirements .configgen
 	echo "" >> conf/st2.conf.sample
 	. $(VIRTUALENV_DIR)/bin/activate; python ./tools/config_gen.py >> conf/st2.conf.sample;
 
+.PHONY: schemasgen
+schemasgen: requirements .schemasgen
+
+.PHONY: .schemasgen
+.schemasgen:
+	@echo
+	@echo "================== content model schemas gen ===================="
+	@echo
+	. $(VIRTUALENV_DIR)/bin/activate; python ./st2common/bin/st2-generate-schemas;
+
 .PHONY: .pylint
 .pylint:
 	@echo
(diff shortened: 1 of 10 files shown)


contrib/schemas/rule.json
```diff
@@ -0,0 +1,105 @@
...
+        "description": {
+            "type": "string"
+        }, 
+        "name": {
+            "required": true, 
+            "type": "string"
+        }, 
+        "ref": {
+            "type": "string", 
+            "description": "System computed user friendly reference for the action.                                 Provided value will be overridden by computed value."
```
blag, line 33: Weird spaces added here.

contrib/schemas/alias.json
```diff
@@ -0,0 +1,115 @@
...
+            "description": "Possible parameter format."
+        }, 
+        "action_ref": {
+            "required": true, 
+            "type": "string", 
+            "description": "Reference to the aliased action."
+        }, 
+        "ref": {
+            "type": "string", 
+            "description": "System computed user friendly reference for the alias.                                 Provided value will be overridden by computed value."
```
blag, line 104: Weird spaces here.

Makefile
```diff
@@ -1052,7 +1062,18 @@ ci-py3-integration: requirements .ci-prepare-integration .ci-py3-integration
...
+	cp contrib/schemas/rule.json /tmp/rule.json.upstream
+	make .schemasgen
+	diff contrib/schemas/pack.json /tmp/pack.json.upstream || (echo "contrib/schemas/pack.json hasn't been re-generated and committed. Please run \"make schemasgen\" and include and commit the generated file." && exit 1)
+	diff contrib/schemas/action.json /tmp/action.json.upstream || (echo "contrib/schemas/pack.json hasn't been re-generated and committed. Please run \"make schemasgen\" and include and commit the generated file." && exit 1)
+	diff contrib/schemas/alias.json /tmp/alias.json.upstream || (echo "contrib/schemas/pack.json hasn't been re-generated and committed. Please run \"make schemasgen\" and include and commit the generated file." && exit 1)
+	diff contrib/schemas/policy.json /tmp/policy.json.upstream || (echo "contrib/schemas/pack.json hasn't been re-generated and committed. Please run \"make schemasgen\" and include and commit the generated file." && exit 1)
+	diff contrib/schemas/rule.json /tmp/rule.json.upstream || (echo "contrib/schemas/pack.json hasn't been re-generated and committed. Please run \"make schemasgen\" and include and commit the generated file." && exit 1)
```
blag, line 1076: :+1: :+1:

contrib/schemas/rule.json
m4dcoder: Good catch. No idea. Must be from the formatting in the content models.  I'll fix it.

blag: Awesome, thanks! Also, I think the description here was copy/pasted from `st2common/st2common/models/api/action.py`, and the `ref` value wasn't changed. I think this should be changed to:

> System computed user friendly reference for the **rule**. Provided value will be overridden by computed value.

(Emphasis mine)

blag: I'm not 100% sure about that though. It could be referring to the action the rule references.

contrib/schemas/alias.json
m4dcoder: Good catch. Fixed

contrib/schemas/rule.json
m4dcoder: Good catch. Fixed.

armab:
> @punkrokk punkrokk locked and limited conversation to collaborators 20 days ago 

Please avoid using administrative buttons without the serious need to do so.
This is an Open Source project in the first place.

punkrokk:
Definitely not intentional. Might have hit it in the github iPhone app

Issue #4918, opened 2020-04-23
Typo in #4874 breaks pack filtering by tags via the API


mweinberg-cm:
## SUMMARY

Filtering packs by tags via the API fails due to a typo in #4874. 

### STACKSTORM VERSION

Paste the output of ``st2 --version``: 
```
st2 3.2dev (c8534cae7), on Python 3.6.9
```

##### OS, environment, install method

One-line install on Ubuntu 18.04.3 LTS. 

## Steps to reproduce the problem

curl -H "X-Auth-Token: 9c1f9268b5fb4f8aaadfea1fa4bb992e" -k https://localhost/api/v1/actions?tags=anytag

## Expected Results

Either an empty list or an action should be returned.

## Actual Results

```
root@st2-dev-01:~/.st2# curl -H "X-Auth-Token: xxx" -k https://localhost/api/v1/actions?tags=marvel-project
{
    "faultstring": "Internal Server Error"
}
```

Will provide a PR to fix this.

armab:
Fixed via https://github.com/StackStorm/st2/pull/4919

Thanks @winem for reporting and fixing!

BTW you can use https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword github magic words to automatically close an Issue once the linked PR was merged.

PR #4919, opened 2020-04-24
Fix typo in the supported API filters when filtering packs

winem:
This PR fixes the issue I found tonight when testing the tag filter as reported in #4918 and adds unit tests to test the API with actions filtered by the name or tag.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -65,6 +65,10 @@ Changed
 
 Fixed
 ~~~~~
+* Fix a typo that caused an internal server error when filtering actions by tags. Fixes #4918
+
+  Reported by @mweinberg-cm and contributed by Marcel Weinberg (@winem)
+
 * Fix the action query when filtering tags. The old implementation returned actions which have the
   provided name as action name and not as tag name. (bug fix) #4828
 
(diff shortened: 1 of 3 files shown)


winem:
Please don't merge it yet, I'm currently fixing the new unit tests to avoid that this feature breaks unnoticed again. Will drop another comment and remove the `WIP:` once the unit tests were successful.

armab:
:+1:  We definitely want to include this for `v3.2`.
@winem Just converted this PR to draft to make sure we don't merge it before you're ready.

See https://github.blog/2019-02-14-introducing-draft-pull-requests/

winem:
@armab thanks for the support. I think and hope that especially the unit tests will be useful in the future.

The tests run fine when I run them locally with nosetests and I can not really see the root cause for the failing travis tests in the code itself. It looks more like a infrastructure issue or an issue with external dependencies (services, dbs, ...). 

I'll remove the WIP flag now and would like to ask for a 2nd opinion regarding the failing tests.

armab:
@winem I just re-started CI jobs and they're green now.
Probably was a temporary failure from TravisCI.

Issue #4922, opened 2020-04-27
Logging API Breaks in Python 3.8


Dahfizz9897:
## SUMMARY

The Python logging module added an optional argument to the findCaller function, which breaks the way that we overwrite the functoin. Similar to this: https://twistedmatrix.com/trac/ticket/9668

### STACKSTORM VERSION

`st2 3.2dev, on Python 3.8.2`

##### OS, environment, install method

Installed from source on Ubuntu 20.04

## Steps to reproduce the problem

Install st2 from source on ubuntu 20.04, which has python 3.8, and run `./tools/launchdev.sh start`

## Expected Results

st2 starts

## Actual Results

```Registering sensors, runners, actions, rules, aliases, and policies...
Traceback (most recent call last):
  File "./st2common/bin/st2-register-content", line 22, in <module>
    sys.exit(content_loader.main(sys.argv[1:]))
  File "/root/st2/st2common/st2common/content/bootstrap.py", line 406, in main
    setup(argv)
  File "/root/st2/st2common/st2common/content/bootstrap.py", line 397, in setup
    common_setup(config=config, setup_db=True, register_mq_exchanges=True,
  File "/root/st2/st2common/st2common/script_setup.py", line 96, in setup
    db_setup()
  File "/root/st2/st2common/st2common/database_setup.py", line 56, in db_setup
    connection = db_init.db_setup_with_retry(**db_cfg)
  File "/root/st2/st2common/st2common/persistence/db_init.py", line 68, in db_setup_with_retry
    return db_func_with_retry(db_setup, db_name, db_host, db_port,
  File "/root/st2/st2common/st2common/persistence/db_init.py", line 58, in db_func_with_retry
    return retrying_obj.call(db_func, *args, **kwargs)
  File "/root/st2/virtualenv/lib/python3.8/site-packages/retrying.py", line 206, in call
    return attempt.get(self._wrap_exception)
  File "/root/st2/virtualenv/lib/python3.8/site-packages/retrying.py", line 247, in get
    six.reraise(self.value[0], self.value[1], self.value[2])
  File "/root/st2/virtualenv/lib/python3.8/site-packages/six.py", line 696, in reraise
    raise value
  File "/root/st2/virtualenv/lib/python3.8/site-packages/retrying.py", line 200, in call
    attempt = Attempt(fn(*args, **kwargs), attempt_number, False)
  File "/root/st2/st2common/st2common/models/db/__init__.py", line 164, in db_setup
[...]

PR #4923, opened 2020-04-27
Fixes Python Logging API Change

Dahfizz9897:
The logging module added a stacklevel=1 kwarg in python3.8 to the findCaller() function
We need to do the same since we override that method.

Closes https://github.com/StackStorm/st2/issues/4922


Dahfizz9897:
This fixes this issue: https://github.com/StackStorm/st2/issues/4922

blag (dismissed):
If the `stacklevel` parameter isn't actually used anywhere, it would probably be better to use `**kwargs` to capture it.

Dahfizz9897:
Updated, but I had to use `*args, **kwargs` as the only parameters because other places in the logging library already call this function with the stacklevel parameter without specifying its name, so it would still error without an `*args`. And the existing `stack_info` wasn't used.

nmaludy (changes requested):
Can you please add a CHANGELOG entry for this?

nmaludy (approved):

CHANGELOG.rst
```diff
@@ -98,6 +98,10 @@ Fixed
   interpolated properly. (bug fix)
 
   Contributed by @misterpah
+* Fixed a compatibility issue with the latest version of the ``logging`` library API
+  where the ``find_caller()`` function introduced some new variables. (bug fix)
```
armab, line 102:
```suggestion
  where the ``find_caller()`` function introduced some new variables. (bug fix) #4923
```

PR #4933, opened 2020-05-01
#4932 - Fixes for field escaping in orquesta workflows

### LINKED ISSUE #4932: Orquesta vars can no longer contain '.' in 3.2
## SUMMARY

In StackStorm 3.2, if an Orquesta workflow contains a `.` in any of its `vars` (ie context) then the workflow will fail to run with an error:
```shell
$ st2 run default.orquesta_vars_with_periods.
id: 5eac3faeca07c849e9b4f280
action.ref: default.orquesta_vars_with_periods
parameters: None
status: failed
start_timestamp: Fri, 01 May 2020 15:26:38 UTC
end_timestamp: Fri, 01 May 2020 15:26:38 UTC
result: 
  errors:
  - message: key 'with.period' must not contain '.'
  output: null
```

### STACKSTORM VERSION

```shell
$ st2 --version
st2 3.2.0, on Python 2.7.5
```

```shell
$ cat /etc/redhat-release 
Red Hat Enterprise Linux Server release 7.8 (Maipo)

# install method = puppet-st2
```

/opt/stackstorm/packs/default/actions/orquesta_vars_with_periods.yaml 
```yaml
---
description: "Demos bug with orquesta vars with periods"
enabled: true
runner_type: orquesta
entry_point: workflows/orquesta_vars_with_periods.yaml
name: orquesta_vars_with_periods
pack: default
parameters:
  demo:
    type: string
    default: "demo"
```

/opt/stackstorm/packs/default/actions/workflows/orquesta_vars_with_periods.yaml 
```
---
version: 1.0

description: "Demos bug with orquesta vars with periods"

input:
  - demo

vars:
  - myvar.with.period: 'this line is going to cause an error'

tasks:
  dispatch:
    action: core.noop
```

Register and run the action
```shell
$ st2 action create /opt/stackstorm/packs/default/actions/orquesta_vars_with_periods.yaml 
$ st2 run default.orquesta_vars_with_periods
.
id: 5eac47b9ca07c849e9b4f289
action.ref: default.orquesta_vars_with_periods
parameters: None
status: failed
start_timestamp: Fri, 01 May 2020 16:00:57 UTC
end_timestamp: Fri, 01 May 2020 16:00:58 UTC
result: 
  errors:
  - message: key 'myvar.with.period' must not contain '.'
  output: null
```

## Expected Results

Workflow runs without error

## Actual Results

Workflow errors trying to write state to the database

nmaludy:
Closes #4932 

This is a fix for the bug found in #4932 .

I had to revert 3x fields in the workflow DB model:

1. `WorkflowExecutionDB.spec` for any fields/dict keys that were statically defined in the workflow spec that include a `.` in their name:
Example:
```yaml
vars:
  - vars.key.with.periods: "vars.value.with.periods"
  - vars.nested.with.periods:
      nested.periods: "vars.nested.value.with.periods"

output:
  - wf.hostname.with.periods: "{{ ctx()['published.hosts'] }}"
  - wf.output.with.periods: "{{ ctx()['published.field'] }}"
```

2. `WorkflowExecutionDB.context` for any dicts/objects that is present in the "context", ie. from a published variable. This can occur, for example, when running `core.remote`. The resulting dictionary has hostnames/IPs as keys in the returned dict.
Example:
```yaml
  run:
    action: core.remote
    input:
      cmd: date
      hosts: "hostname.domain.tld,hostname2.domain.tld"
    next:
      - when: "{{ succeeded() }}"
        publish:
          - published.hosts: "{{ result().result }}"
          - published.field: "{{ result().result['hostname2.domain.tld']['stdout'] }}"
```

3. `TaskExecutionDB.task_spec` for any inputs passed into the task that may have a dict with a key contain a `.`.  
Example:
```

  bolt_plan_run:
    action: bolt.plan_run
    input:
      plan: "{{ ctx().plan }}"
      targets: "{{ ctx().targets }}"
      env:
        ST2KV_CONFIG:
          ssh.username: system.linux.username
          ssh.password: system.linux.password
```

I've added a single integration test case `examples.orequest-test-field-escaping` that invokes these three corner cases.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,18 @@ Changelog
 in development
 --------------
 
+Fixed
+~~~~~
+* Fixed a bug where persisting Orquesta to the MongoDB database returned an error
+  ``message: key 'myvar.with.period' must not contain '.'``. This happened anytime an
+  ``input``, ``output``, ``publish`` or context ``var`` contained a key with a ``.`` within
+  the name (such as with hostnames and IP addresses). This was a regression introduced by
+  trying to improve performance. Fixing this bug means we are sacrificing performance of
+  serialization/deserialization in favor of correctness for persisting workflows and
+  their state to the MongoDB database. (bug fix) #4932
+
+  Contributed by Nick Maludy (@nmaludy Encore Technologies)
+
 
 3.2.0 - April 27, 2020
 ----------------------
--- a/Makefile
+++ b/Makefile
@@ -495,8 +495,9 @@ distclean: clean
 
 .PHONY: .requirements
 .requirements: virtualenv
-	# Generate all requirements to support current CI pipeline.
+	# Print out pip version so we can see what version was restored from the Travis cache
 	$(VIRTUALENV_DIR)/bin/pip --version
+	# Generate all requirements to support current CI pipeline.
 	$(VIRTUALENV_DIR)/bin/python scripts/fixate-requirements.py --skip=virtualenv,virtualenv-osx -s st2*/in-requirements.txt contrib/runners/*/in-requirements.txt -f fixed-requirements.txt -o requirements.txt
 
 	# Remove any *.egg-info files which polute PYTHONPATH
--- a/Makefile
+++ b/Makefile
@@ -496,6 +496,7 @@ distclean: clean
 .PHONY: .requirements
 .requirements: virtualenv
 	# Generate all requirements to support current CI pipeline.
+	$(VIRTUALENV_DIR)/bin/pip --version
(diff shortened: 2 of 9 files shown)


winem:
LGTM!

And I successfully tested the fix on one of my dev nodes. I just added an additional variable `some.value` to the packs/actions/workflows/install.yaml

Without the fix:
Selection_107

With the fix applied:
Selection_109

nmaludy:
FYI build failures were caused by poisoned Travis cache where `pip==20.1` was installed and cached. Travis uses a hierarchical cache where if it can't find a cache for the current PR, it uses the cache from `master`. Since `pip==20.1` breaks our `fixate-requirements.py` this caused both the `master` build and this PRs build to fail.

nmaludy:
@armab Added a changelog. Thank you for noticing this, i was in such a rush to get it fixed i forgot!

Issue #4946, opened 2020-05-13
how to run linux.netstat action with arguments


scguoi:
**PLEASE READ**: If you are a paid Extreme Workflow (EWC) customer, please contant [Extreme
Network GTAC](https://gtacknowledge.extremenetworks.com/articles/How_To/How-to-contact-Extreme-Networks-Global-Technical-Assistance-Center-GTAC)
or send an email to support@stackstorm.com

For free community support, please use out forum at [https://forum.stackstorm.com](https://forum.stackstorm.com)

I try to run linux.netstat action with args, but result are error as follow.

$ st2 run linux.netstat hosts=172.31.114.68 args="-antp"
ERROR: 'type'

Sheshagiri:
looks like `type` attribute is missing for `netstat` action. I hope #4947 will fix that permanently. 

In the meantime, you can add the type attribute and re-register the Linux pack and that should fix the issue.
update `/opt/stackstorm/packs/linux/actions/netstat.yaml` to
```
---
  name: "netstat"
  runner_type: "remote-shell-cmd"
  description: "Run netstat"
  enabled: true
  entry_point: ""
  parameters:
    cmd:
      immutable: true
      default: "netstat {{args}}"
    args:
      description: "Command line arguments"
      default: " "
      type: "string"
``` 
and re-register the Linux pack using the following command
```bash
st2 pack register linux
```
This should resolve the issue for you.

PR #4947, opened 2020-05-13
fix netstat action in linux pack

Sheshagiri:
looks like the `type` attribute is missing for `netstat` action in Linux pack because of which the action in not usable.

```
$ st2 run linux.netstat hosts=172.31.114.68 args="-antp"
ERROR: 'type'
```

closes: #4946

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,10 @@ Added
 
 Fixed
 ~~~~~
+* Fixed a bug where `type` attribute was missing for netstat action in linux pack. Fixes #4946
+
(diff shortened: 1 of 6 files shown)


Sheshagiri:
> Good find!
> 
> Could you please add a CHANGELOG entry for this.

done 👍

armab:
@Sheshagiri Can you also try to fix the lint failures in https://travis-ci.org/github/StackStorm/st2/jobs/686593938#L2218-L2219. Looks like a small change, but if it's not, - you can start a new PR.

It's something new as master is green and I don't think failures are related to your change. Probably just an upstream update that broke it.

PR #4957, opened 2020-05-23
Pin pip consistently in one place across the targets, fix st2docs build

armab:
`st2docs` build which is based on `st2` makefile started to fail (https://circleci.com/gh/StackStorm/st2docs/2529) due to upstream pip updated to latest `20.1.1`, while we failed to pin it to `20.0.2`. Thanks to @winem for the find.

We should not pin `pip` in files like `.travis.yml`, but instead use single consistent way and not hack across the many places as other systems rely on st2 makefiles for the builds.

Related: https://github.com/StackStorm/discussions/issues/2

This change sets pip version in one Makefile place and fixes st2docs builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,7 +82,6 @@ cache:
 before_install:
   # If you update these versions, make sure you update the versions in the .st2client-install-check
   # and the requirements targets in the Makefile to match
-  - pip install --upgrade "pip==20.0.2"
   - sudo pip install --upgrade "virtualenv==16.6.0"
 
 install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,9 +80,8 @@ cache:
     #- .tox/
 
 before_install:
-  # If you update these versions, make sure you update the versions in the .st2client-install-check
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -80,10 +80,8 @@ cache:
     #- .tox/
 
 before_install:
-  # If you update these versions, make sure you update the versions in the .st2client-install-check
-  # and the requirements targets in the Makefile to match
-  - pip install --upgrade "pip==20.0.2"
-  - sudo pip install --upgrade "virtualenv==16.6.0"
```
armab, line 86: FYI pinning `virtualenv` here never really worked properly. In Travis environment even after updating virtualenv to latest:
```
virtualenv --version

16.1.0
```
for py2 env

and

```
virtualenv --version

16.0.0
```
for py3 env

armab: It turns out because different virtualenv binaries were used when using `sudo` and without.

armab:
There are a bunch of st2docs PRs which merging is blocked by this CI failure and I'd like us to fix this ASAP.

PR #4963, opened 2020-06-06
st2cliient - elapsed time in a more user-friendly format

### LINKED ISSUE #4944: print elapsed time in user friendly format instead of seconds on cli
This is not a bug but just a small user experience improvement. 

When we get the status of execution using the `st2` cli time is mentioned in seconds it would be nice to print time in a user-friendly format.
 
image

ex: `running (339s elapsed)`, could be: `running (5m39s elapsed)`

I'll try to make a pr with the changes if its okay.

winem:
This PR changes the output format of the elapsed time on `st2 execution get`. 

Old: `status: succeeded (76s elapsed)`
New: `status: succeeded (0:01:16 hours elapsed)` 

I changed the unit presentation from `s` to `hours` because `datetime.timedelta` represents days in the following format: `3 days, ...` 

So I think `1 day, 0:32:32 hours` seems more intuitive and consistent than `1 day, 0:32:32h`.

Other output formats would require some additional formatting.

fixes #4944


winem:
Hi @armab, I updated the PR to present the elapsed time as requested.

winem:
Oh jesus, sorry! This explains a lot... will fix this shortly.

CHANGELOG.rst
```diff
@@ -9,6 +9,14 @@ Added
 * Add make command to autogen JSON schema from the models of action, rule, etc. Add check
   to ensure update to the models require schema to be regenerated. (new feature)
 
+Changed
+~~~~~~~
+* Changed the output of elapsed time on `st2 execution get` to a more user-friendly format.
+  The new format is ``<days>d <hours>h<minutes>m<seconds>s``. Units without a value are omitted. 
+  Old: ``status: succeeded (92342s elapsed)``, ``status: succeeded (113s elapsed)``
+  New: ``stauts: succeeded (1d 1h39m2s elapsed)``, ``status: succeeded (1m53s elapsed)``
```
Sheshagiri, line 17: I guess there is no space now.

winem: Oh yes, fixed.

winem:
I fixed the multiple spaces before the = operator as reported by lint. That's something I'll keep in mind as I use this to format my code in other projects from time to time. So it shouldn't happen again on a st2 PR.

The PR description was updated as well.

I will provide a proper unit test in the next days as I'm working on st2 and other open source projects mostly in the evening after my usual job (which also explains my delayed reaction to the CI status from time to time. :) ).

winem:
I added a test but it looks like I got something wrong. My expectation was that the reported duration will be calculated from `end_timestamp - start_timestamp` but the duration is always 1 second. I will investigate this tomorrow.

winem:
Hey team, 

I definitely need some help here. :( The unit test keeps failing unless I use `mock.patch.object` to mock the response which should not be needed. Example for the working test: https://github.com/StackStorm/st2/blob/c5d8e63eae7ba69f2e39caac598161194c3899e3/st2client/tests/unit/test_formatters.py#L137

But as far as I can say (and @nmaludy agreed here on our recent slack discussion on this topic) it should not be required to use `mock.patch.object` to return the correct response. 

The cause for the failing test is that the test always get's the default executions object with the duration of 1 second and not the expected execution_with_hours_in_elapsed_time.json with a duration of 2h0m1s as per:
```
"start_timestamp": "2014-12-02T19:56:06.900000Z",
"end_timestamp": "2014-12-02T21:56:07.000000Z",
```

So the question is, why `argv = ['execution', 'get', DURATION_HOURS['id'], '--attr', 'status']` in my test on line 119 returns the start_timestamp and end_timestamp of `EXECUTION = FIXTURES['executions']['execution.json']` instead of the one from `DURATION_HOURS = FIXTURES['executions']['execution_with_hours_in_elapsed_time.json']`.

(I also wrote a summary in the slack #development channel last week, just in case it is helpful in any way: https://stackstorm-community.slack.com/archives/CSBELJ78A/p1592686455460800)

Any help or hint is highly appreciated. I'd also be fine to use `mock.patch.object` again if anyone can explain why it is needed here, so that I can learn from it for my next PR. 

Cheers,
Marcel

st2client/tests/unit/test_formatters.py
```diff
@@ -112,6 +115,12 @@ def test_execution_get_attributes(self):
         content = self._get_execution(argv)
         self.assertEqual(content, FIXTURES['results']['execution_get_attributes.txt'])
 
+    def test_execution_get_attribute_with_hours_in_elapsed_time(self):
+        argv = ['execution', 'get', DURATION_HOURS['id'], '--attr', 'status']
+        content = self._get_execution(argv)
```
cognifloyd, line 120: I think using `self._get_execution()` is the source of the error because `_get_execution()` has a mock so it always returns the json from `EXECUTION`, but you need it to return `DURATION_HOURS`.

Here's `_get_execution` with the mocked `EXECUTION` response:
```
    @mock.patch.object(
        httpclient.HTTPClient,
        "get",
        mock.MagicMock(
            return_value=base.FakeResponse(json.dumps(EXECUTION), 200, "OK", {})
        ),
    )
    def _get_execution(self, argv):
        self.assertEqual(self.shell.run(argv), 0)
        self._undo_console_redirect()
        with open(self.path, "r") as fd:
            content = fd.read()

        return content
```

That's why your example unit test that uses mock.MagicMock worked better than `self._get_execution()`.

cognifloyd (commented):
Helpful FYI (a warning of sorts): We just merged the PR that reformatted the code with `black`. So, this needs to be rebased or master needs to be merged (resolve any conflicts & reformat with black).

As far as the unit test goes, I left a comment. All of the methods are employing mocked http responses, albeit indirectly for those that use `self._get_execution`. So, if you don't want to mock, you'll need to come up with alternative test machinery.

PR #4966, opened 2020-06-12
remove deprecated ssl and duplicate proxy host header

punkrokk:
This PR removes:

- the duplicate host header in the Load balancer configs for each API endpoint
- the `ssl on` directive with is deprecated and redundant with ` listen       *:443 ssl;` directive

Tested


punkrokk:
@armab I added this to changelog

Issue #4975, opened 2020-06-26
Using boto3 in a python3 sensor gives a `RecursionError` when building boto sensor


punkrokk:
## SUMMARY

Initializing a boto3 client with a python3 sensor results in a `RecursionError`

### STACKSTORM VERSION

st2 3.2.0, on Python 3.6.8

##### OS, environment, install method

Post what OS you are running this on, along with any other relevant information/

CentOS 8, ansible

## Steps to reproduce the problem

Show how to reproduce the problem, using a minimal test-case. Make sure to include any content
```
    import boto3
    from botocore.exceptions import ClientError

    self.boto3_client = 'guardduty'

    client = boto3.client(self.boto3_client,
                              aws_access_key_id=self.key,
                              aws_secret_access_key=self.secret,
                              region_name=self.region)
```

## Expected Results
We can create the client

## Actual Results

```
File "/opt/stackstorm/packs/aws_guardduty/lib/base_guardduty_sensor.py", line 146, in make_client
    region_name=region)
  File "/opt/stackstorm/virtualenvs/aws_guardduty/lib/python3.6/site-packages/boto3/__init__.py", line 91, in client
    return _get_default_session().client(*args, **kwargs)
  File "/opt/stackstorm/virtualenvs/aws_guardduty/lib/python3.6/site-packages/boto3/session.py", line 263, in client
    aws_session_token=aws_session_token, config=config)
  File "/opt/stackstorm/virtualenvs/aws_guardduty/lib/python3.6/site-packages/botocore/session.py", line 835, in create_client
    client_config=config, api_version=api_version)
  File "/opt/stackstorm/virtualenvs/aws_guardduty/lib/python3.6/site-packages/botocore/client.py", line 85, in create_client
    verify, credentials, scoped_config, client_config, endpoint_bridge)
  File "/opt/stackstorm/virtualenvs/aws_guardduty/lib/python3.6/site-packages/botocore/client.py", line 287, in _get_client_args
    verify, credentials, scoped_config, client_config, endpoint_bridge)
  File "/opt/stackstorm/virtualenvs/aws_guardduty/lib/python3.6/site-packages/botocore/args.py", line 107, in get_client_args
    client_cert=new_config.client_cert)
  File "/opt/stackstorm/virtualenvs/aws_guardduty/lib/python3.6/site-packages/botocore/endpoint.py", line 298, in create_endpoint
    client_cert=client_cert,
[...]

PR #4976, opened 2020-06-26
Fixes: #4875 monkeypatch st2 sensor earlier

punkrokk:
Fixes: #4975 
Fixes: #4832

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -37,7 +37,6 @@ Fixed
 * Fixed a bug where a python3 sensor using ssl needs to be monkey patched earlier. #4975 ( See also
   #4832 and https://github.com/gevent/gevent/issues/1016) Contributed by @punkrokk
 
-=======
 Removed
 ~~~~~~~
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,6 +27,8 @@ Fixed
 * Fix a bug where passing an empty list to a with items task in a subworkflow causes
   the parent workflow to be stuck in running status. (bug fix) #4954
 
+* Fixed a bug in the example nginx HA template declared headers twice (bug fix) #4966
+  Contributed by @punkrokk
 
 3.2.0 - April 27, 2020
 ----------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -45,7 +45,6 @@ Removed
 
   Contributed by Amanda McGuinness (@amanda11 Ammeon Solutions)
   
-
 3.2.0 - April 27, 2020
 ----------------------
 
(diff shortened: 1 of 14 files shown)


punkrokk:
@alexku7 Could you test this branch out and tell me if it resolves #4832 for you?

guzzijones:
Yeah monley patching should always be the very first thing. Before imports.

st2reactor/st2reactor/container/sensor_wrapper.py
```diff
@@ -14,6 +14,18 @@
 
 from __future__ import absolute_import
 
+# Note: We need to perform monkey patching in the worker. If we do it in
+# the master process (gunicorn_config.py), it breaks tons of things
+# including shutdown
+# NOTE: It's important that we perform monkey patch as early as possible before any other modules
+# are important, otherwise SSL support for sensor clients won't work.
```
cognifloyd, line 21:
```suggestion
# are imported, otherwise SSL support for sensor clients won't work.
```

guzzijones:
It looks like these are all in the `cmd` folder so they are command scripts that are run by for services.  This way of monkey patching should work just fine. And yes it needs to be before any other imports.  If we have other cases of monkey patching we need to do the same. 
I don't see any of these scripts used as imports at first glance which is what I thought would be an issue.

punkrokk:
@Kami Do you have any suggestions for how to add a test case for this? 

The only thing I can think of is adding one that fails if we see a recursion error?

CHANGELOG.rst
```diff
@@ -34,13 +34,17 @@ Fixed
...
+  #4832 and https://github.com/gevent/gevent/issues/1016) Contributed by @punkrokk
+
 Removed
 ~~~~~~~
 
 * Removed ``CentOS 6``/``RHEL 6`` support #4984
 
   Contributed by Amanda McGuinness (@amanda11 Ammeon Solutions)
   
+
```
armab, line 47:
```suggestion
```

CHANGELOG.rst
```diff
@@ -34,13 +34,17 @@ Fixed
 
   Contributed by Nick Maludy (@nmaludy Encore Technologies)
 
+* Fixed a bug where a python3 sensor using ssl needs to be monkey patched earlier. #4975 ( See also
+  #4832 and https://github.com/gevent/gevent/issues/1016) Contributed by @punkrokk
```
armab, line 38:
```suggestion
* Fixed a bug where a python3 sensor using ssl needs to be monkey patched earlier. (bug fix) #4976
  See also #4832, #4975 and https://github.com/gevent/gevent/issues/1016
  
  Contributed by @punkrokk
```

punkrokk:
Forum thread: https://forum.stackstorm.com/t/centos-8-problems-with-rabbitmq-connections-when-sensor-dispatches-trigger/1328/12

PR #4981, opened 2020-06-30
ST2api Log Fix So That Querying a Non-Existent Key in the Datastore Will Not Print Traceback

### LINKED ISSUE #4979: When trying to get a key from the datastore, st2api logs an error but also prints the traceback
## SUMMARY

Provide a quick summary of your bug report.

### STACKSTORM VERSION

v3.2.0

Post what OS you are running this on, along with any other relevant information/

CentOS 7 / 8

```
def get_value(self, name, local=True, scope=SYSTEM_SCOPE, decrypt=False):
        """
        Retrieve a value from the datastore for the provided key.
        By default, value is retrieved from the namespace local to the pack/class. If you want to
        retrieve a global value from a datastore, pass local=False to this method.
        :param name: Key name.
        :type name: ``str``
        :param local: Retrieve value from a namespace local to the pack/class. Defaults to True.
        :type: local: ``bool``
        :param scope: Scope under which item is saved. Defaults to system scope.
        :type: local: ``str``
        :param decrypt: Return the decrypted value. Defaults to False.
        :type: local: ``bool``
        :rtype: ``str`` or ``None``
        """
        if scope != SYSTEM_SCOPE:
            raise ValueError('Scope %s is unsupported.' % scope)

        name = self._get_full_key_name(name=name, local=local)

        client = self.get_api_client()
        self._logger.debug('Retrieving value from the datastore (name=%s)', name)

        try:
            params = {'decrypt': str(decrypt).lower(), 'scope': scope}
            kvp = client.keys.get_by_id(id=name, params=params)
        except Exception as e:
            self._logger.exception(
                'Exception retrieving value from datastore (name=%s): %s',
                name,
                e
            )
            return None

        if kvp:
            return kvp.value
```

## Expected Results

I expect to get an ERROR, but not the traceback

## Actual Results

```
Jun 26 18:27:35 ip-10-15-1-121 gunicorn[184385]: 2020-06-26 18:27:35,134 ERROR [-] Failed to call controller function "get_one" for operation "st2api.controllers.v1.keyvalue:key_value_pair_controller.get_one": The key "aws_inspector.AWSInspector:us-east-2.last_start_time_" does not exist in the StackStorm datastore.
Jun 26 18:27:35 ip-10-15-1-121 gunicorn[184385]: Traceback (most recent call last):
Jun 26 18:27:35 ip-10-15-1-121 gunicorn[184385]:  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2common/router.py", line 515, in __call__
Jun 26 18:27:35 ip-10-15-1-121 gunicorn[184385]:    resp = func(**kw)
Jun 26 18:27:35 ip-10-15-1-121 gunicorn[184385]:  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2api/controllers/v1/keyvalue.py", line 124, in get_one
Jun 26 18:27:35 ip-10-15-1-121 gunicorn[184385]:    from_model_kwargs=from_model_kwargs
Jun 26 18:27:35 ip-10-15-1-121 gunicorn[184385]:  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2api/controllers/resource.py", line 402, in _get_one_by_scope_and_name
Jun 26 18:27:35 ip-10-15-1-121 gunicorn[184385]:    instance = self.access.get_by_scope_and_name(scope=scope, name=name)
Jun 26 18:27:35 ip-10-15-1-121 gunicorn[184385]:  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2common/persistence/keyvalue.py", line 115, in get_by_scope_and_name
Jun 26 18:27:35 ip-10-15-1-121 gunicorn[184385]:    raise StackStormDBObjectNotFoundError(msg % name)
Jun 26 18:27:35 ip-10-15-1-121 gunicorn[184385]: st2common.exceptions.db.StackStormDBObjectNotFoundError: The key "aws_inspector.AWSInspector:us-east-2.last_start_time_" does not exist in the StackStorm datastore.sod
```
Thanks!

saucetray:
ST2api.log will print a traceback when you attempt to query a non-existent key in the datastore. This should fix that.

Fixes #4979


amanda11:
Currently exceptions raised are re-thrown in router.py and then caught higher up.
In api.py at https://github.com/StackStorm/st2/blob/master/st2api/st2api/app.py#L82 defines the generic error handling middleware.
And then in st2common you've got this code to convert any exceptions that bubble-up into http response codes https://github.com/StackStorm/st2/blob/8f507d6c8d9483c8371e386fe2b7998596856fd7/st2common/st2common/middleware/error_handling.py#L38

One possible solution would be to keep the throw of StackStormDBObjectNotFoundError. Then add an explicit catch  next to https://github.com/StackStorm/st2/blob/master/st2common/st2common/router.py#L514-L519 for the StackStormDBObjectNotFoundError, still re-raise so you get the http response code handling as before. But rather than use a LOG.exception use a LOG that just logs the exception msg, rather than the full stack trace.
So you handle the permitted exceptions of objectnotfound by just keeping the message and not stack, and for non-expected exceptions you then log the full stack trace, so you don't loose it when you rethrow it?

guzzijones:
@amanda11 +1 for that solution

saucetray:
Oh. Okay, right. That sounds better.

st2common/st2common/router.py
```diff
@@ -513,6 +514,10 @@ def __init__(self, **entries):
 
         try:
             resp = func(**kw)
+        except StackStormDBObjectNotFoundError as e:
```
armab, line 517: Looks like with this change any DB query with DBObjectNotFound will result in a Warning now instead of more actionable log. It doesn't look like desired outcome to me as trying to fix one not found K/V issue we may introduce a regression.

We probably need only a warning when Key is not found in the datastore?

saucetray: Is a solution to this to add a new exception and have that raised @armab ?

armab: That would be a good idea to explore. For example, I see `TokenNotFoundError`, `ApiKeyNotFoundError` and similar examples in this file and that makes more sense.

punkrokk: I think the solution is to really add a new exception such as: `DataStoreKeyNotFoundError`. The two you suggest aren't the precise meaning of them.

saucetray: `
class StackStormDBDataStoreKeyNotFoundError(StackStormBaseException):
    pass
`
Maybe would be a good exception in https://github.com/StackStorm/st2/blob/master/st2common/st2common/exceptions/db.py#L21?

punkrokk: I was actually thinking that if you got the DBObjectNotFoundError while looking for a key, you would then raise a DataStoreKeyNotFoundError

armab: > The two you suggest aren't the precise meaning of them.

I hope that would be a good analogy and pointer where to look at and follow similar approach.

saucetray: I think this might be resolved.

st2common/st2common/exceptions/db.py
```diff
@@ -16,6 +16,10 @@
 from st2common.exceptions import StackStormBaseException
 
 
+class StackStormDBKeyNotFoundError(StackStormBaseException):
```
armab, line 19: `DataStoreKeyNotFoundError` that @punkrokk mentioned fit to be part of `exceptions/keyvalue.py`.
Same as `ApiKeyNotFoundError` is part `exceptions/auth.py`: https://github.com/StackStorm/st2/blob/8f507d6c8d9483c8371e386fe2b7998596856fd7/st2common/st2common/exceptions/auth.py#L55-L56
which is then mentioned in the `router.py`:
https://github.com/StackStorm/st2/blob/8f507d6c8d9483c8371e386fe2b7998596856fd7/st2common/st2common/router.py#L318-L335

Does that look more like a pattern now?

saucetray: Oh. I see! I apologize.

amanda11:
I think I've found a problem with making it extend base exception.

There are a number of calls made to get_by_scope_and_name that check if the DBObjectNotFoundError is returned and act on it. So I think by changing it to a new exception (that doesn't extend DbObjectNotFoundError) then we will break these, e.g.
in st2common/st2common/services/keyvalues.py and st2common/st2common/services/config.py 

So I think it might be better to change this new exception to extend DBObjectNotFoundError.

saucetray:
If this extends that exception, then shouldn’t it be put in db.py?

amanda11:
> If this extends that exception, then shouldn’t it be put in db.py or should it still be in keyvalue.py?

No - there's a few other examples, see trace.py and auth.py - both have exceptions in them that extend the DBObjectNotFoundError.

saucetray:
> > If this extends that exception, then shouldn’t it be put in db.py or should it still be in keyvalue.py?
> 
> 
> 
> No doesn't need to stay in db.py - there's a few other examples, see trace.py and auth.py - both have exceptions in them that extend the DBObjectNotFoundError.

Ok, let me then see what I come up with.

st2common/st2common/persistence/keyvalue.py
```diff
@@ -19,7 +19,7 @@
 from st2common.constants.triggers import KEY_VALUE_PAIR_UPDATE_TRIGGER
 from st2common.constants.triggers import KEY_VALUE_PAIR_VALUE_CHANGE_TRIGGER
 from st2common.constants.triggers import KEY_VALUE_PAIR_DELETE_TRIGGER
-from st2common.exceptions.db import StackStormDBObjectNotFoundError
+from st2common.exceptions.db import StackStormDBKeyNotFoundError
```
amanda11, line 22: I think I can see why the unit-tests are failing. Although we've changed the catch to be for the DataStoreKeyNotFoundError we are raising the StackStormDBKeyNotFoundError.

Can we swap from StackStormDBKeyNotFoundError to st2common.exceptions.keyvalue.DataStoreKeyNotFoundError.

st2common/st2common/middleware/error_handling.py
```diff
@@ -62,7 +62,7 @@ def __call__(self, environ, start_response):
             if isinstance(e, exc.HTTPException):
                 status_code = status
                 message = six.text_type(e)
-            elif isinstance(e, db_exceptions.StackStormDBObjectNotFoundError):
+            elif isinstance(e, db_exceptions.StackStormDBObjectNotFoundError) or isinstance(e, db_exceptions.StackStormDBKeyNotFoundError):
```
amanda11, line 65: As we extend DBObjetNotFoundError we  no-longer need this or on isinstance of StackStormDBKeyNotFoundError

st2common/st2common/exceptions/db.py
amanda11, line 19: We can remove the definition of StackStormDBKeyNotFoundError in the st2common.exceptions.db package as we've now got it in st2common.exceptions.keyvalue.DataStoreKeyNotFoundError exception instead. So you should be able to remove this definition.

st2common/st2common/middleware/error_handling.py
```diff
@@ -62,7 +62,7 @@ def __call__(self, environ, start_response):
             if isinstance(e, exc.HTTPException):
                 status_code = status
                 message = six.text_type(e)
-            elif isinstance(e, db_exceptions.StackStormDBObjectNotFoundError):
+            elif isinstance(e, db_exceptions.StackStormDBObjectNotFoundError)
```
amanda11, line 65: : missing here

saucetray: oh thats embarrassing

saucetray:
ermmm
`======================================================================
91961) FAIL: test_process_wrapper_exits_in_reasonable_timeframe (tests.integration.test_python_action_process_wrapper.PythonRunnerActionWrapperProcessTestCase)
9197----------------------------------------------------------------------
9198   Traceback (most recent call last):
9199    contrib/runners/python_runner/tests/integration/test_python_action_process_wrapper.py line 101 in test_process_wrapper_exits_in_reasonable_timeframe
9200      self.assertTrue(avg_run_time_seconds <= WRAPPER_PROCESS_RUN_TIME_UPPER_LIMIT, assertion_msg)
9201   AssertionError: False is not true : Python wrapper process script took more than 0.31 seconds to execute (0.31875). This most likely means
9202   that a direct or in-direct import of a module which takes a long time to load has been added (e.g.
9203   jsonschema, pecan, kombu, etc).
9204   
9205   Please review recently changed and added code for potential slow import issues and refactor /
9206   re-organize code if possible.`

amanda11:
> ermmm
> 
> ```
> 91961) FAIL: test_process_wrapper_exits_in_reasonable_timeframe (tests.integration.test_python_action_process_wrapper.PythonRunnerActionWrapperProcessTestCase)
> 9197----------------------------------------------------------------------
> 9198   Traceback (most recent call last):
> 9199    contrib/runners/python_runner/tests/integration/test_python_action_process_wrapper.py line 101 in test_process_wrapper_exits_in_reasonable_timeframe
> 9200      self.assertTrue(avg_run_time_seconds <= WRAPPER_PROCESS_RUN_TIME_UPPER_LIMIT, assertion_msg)
> 9201   AssertionError: False is not true : Python wrapper process script took more than 0.31 seconds to execute (0.31875). This most likely means
> 9202   that a direct or in-direct import of a module which takes a long time to load has been added (e.g.
> 9203   jsonschema, pecan, kombu, etc).
> 9204   
> 9205   Please review recently changed and added code for potential slow import issues and refactor /
> 9206   re-organize code if possible.
> ```

That's not you. It's a known test that occassionally runs slow.... We've not yet worked out how to resolve this one.

amanda11:
@saucetray There is however a lint error that needs fixing

". virtualenv/bin/activate; flake8 --config ./lint-configs/python/.flake8 st2actions st2api st2auth st2client st2common st2debug st2exporter st2reactor st2stream st2tests
st2common/st2common/exceptions/keyvalue.py:33:1: E302 expected 2 blank lines, found 1"

saucetray:
oh yeah I fixed that already I'm seeing if there's any other errors before I start this

st2common/st2common/exceptions/keyvalue.py
```diff
@@ -25,6 +27,10 @@ class CryptoKeyNotSetupException(StackStormBaseException):
     pass
 
 
+class DataStoreKeyNotFoundError(StackStormDBObjectNotFoundError):
```
armab, line 30: :+1:

CHANGELOG.rst
```diff
@@ -22,6 +22,11 @@ Changed
 
 Contributed by @punkrokk 
 
+* Added an enhancement where ST2api.log no longer reports the entire traceback when trying to get a datastore value
+  that does not exist. It now reports a simplified log for cleaner reading. Addresses and Fixes #4979
```
armab, line 26:
```suggestion
* Added an enhancement where ST2api.log no longer reports the entire traceback when trying to get a datastore value
  that does not exist. It now reports a simplified log for cleaner reading. Addresses and Fixes #4979. (improvement) #4981
```

st2common/st2common/router.py
```diff
@@ -513,6 +514,10 @@ def __init__(self, **entries):
 
         try:
             resp = func(**kw)
+        except DataStoreKeyNotFoundError as e:
+            LOG.warning('Failed to call controller function "%s" for operation "%s": %s, %s' %
```
armab, line 518: Can you please also show an example how the new Warning looks like in the logs?

saucetray: Ok. Will do.

saucetray:
```2020-07-31 18:58:15,737 140482139303256 WARNING (unknown file) [-] Failed to call controller function "get_one" for operation "st2api.controllers.v1.keyvalue:key_value_pair_controller.get_one": The key "sod.aws.inspector.mapping_dict" does not exist in the StackStorm datastore., The key "sod.aws.inspector.mapping_dict" does not exist in the StackStorm datastore.

**Comment by armab** [comment]
At least it shouldn't print some information 2 times in one message.

**Comment by amanda11** [comment]
LGTM

File: CHANGELOG.rst [hunk 9]
```diff
@@ -27,6 +27,11 @@ Changed
 
   Contributed by @punkrokk
 
+* Added an enhancement where ST2api.log no longer reports the entire traceback when trying to get a datastore value
+  that does not exist. It now reports a simplified log for cleaner reading. Addresses and Fixes #4979. (improvement) #4981
+
+Contributed by Justin Sostre (@saucetray)
```
armab, line 33: Please fix the indentation here. It will format this line as separated to the changelog list item.

CHANGELOG.rst
```diff
@@ -27,6 +27,11 @@ Changed
 
   Contributed by @punkrokk
 
+* Added an enhancement where ST2api.log no longer reports the entire traceback when trying to get a datastore value
+that does not exist. It now reports a simplified log for cleaner reading. Addresses and Fixes #4979. (improvement) #4981
+
+Contributed by Justin Sostre (@saucetray)
```
blag, line 33: I think this is what @armab meant:

```suggestion
* Added an enhancement where ST2api.log no longer reports the entire traceback when trying to get a datastore value
  that does not exist. It now reports a simplified log for cleaner reading. Addresses and Fixes #4979. (improvement) #4981

  Contributed by Justin Sostre (@saucetray)
```

This will fix the linting test failure as well.

Issue #2574, opened 2016-03-10
Make st2actions and st2tests packages available on PyPi


codyaray:
My current development workflow is basically code in PyCharm, rsync to my stackstorm host, and test there. (There's probably a better workflow using st2workroom or something that I haven't tried.) But anyway, it'd be nice to be able to write (and run) my tests locally for development. But I can't easily install these dependencies.

### Workaround: Build st2tests From Source

git clone https://github.com/StackStorm/st2
cd st2/st2tests
source ~/.virtualenvs/my-awesome-new-pack
python setup.py build
python setup.py install

### Issue Building st2actions from source on Mac OS X:

    Collecting pyinotify<=0.10,>=0.9.5 (from -r requirements.txt (line 13))
      Downloading pyinotify-0.9.6.tar.gz (60kB)
        100% |████████████████████████████████| 61kB 8.6MB/s
        Complete output from command python setup.py egg_info:
        inotify is not available on macosx-10.10-intel

        ----------------------------------------
    Command "python setup.py egg_info" failed with error code 1 in /private/tmp/pip-build-TfxaiK/pyinotify/

Kami:
Currently you can do that by using a StackStorm development VM (st2workroom).

In any case, we do plan to push individual components to PyPi in the near future. We have started that work when starting to work on the new debian / rpm packages. We are quite close to that at this point (each component has setup.py file, etc.), we just need to polish it and add some automation (StackStorm! :)) around publishing packages to PyPi.

StephenWeber:
This is a feature I would like, to be able to use ST2-provided Mock objects in my tests. Looks like, as of v3.2.0 and (current) v3.3dev there is no `setup.py` provided for the `st2tests` package.

I'm installing Stackstorm packages directly from GitHub, so I do not necessarily need the packages published to PyPI, so my needs would be satisfied by adding the file (and other changes like adding the version to its `__init__.py`

armab:
@StephenWeber Looks like you know best what exactly is needed there.
We’ll be happy to accept PR for that! 👍

PR #4982, opened 2020-07-01
Add version string to st2tests to make it installable

StephenWeber:
Prior to this change, this will fail:

    cd st2tests/st2tests
    pip install .

After this change that command successfully installs the `st2tests` package. This will also work for installing via GitHub as in:

    pip install -e git+https://github.com/StackStorm/st2.git@v3.2.0#egg=st2tests&subdirectory=st2tests

The original request in #2574 is to get st2tests onto PyPI, and I'm not sure if this will accomplish that request, but this is a good first step.


StephenWeber:
I see I need to sign the CLA, please wait while I seek approval to do so.

StephenWeber:
I'm blocked on running tests because of #3934, though I might see if I can get enough installed to just run the st2tests integration tests.

armab:
You can also use Vagrant (https://github.com/stackstorm/st2vagrant) or Docker for StackStorm Development as StackStorm comes with lots of requirements and infra, including DB/MQ/etc.

@StephenWeber BTW, do you think it would be easier for you to sign DCO or CLA, considering your work environment?
We might switch to DCO soon.

StephenWeber:
Thanks for the tip on testing!

> do you think it would be easier for you to sign DCO or CLA, considering your work environment?
> We may switch to DCO soon.

We have an internal process for CLA, I do not believe we have a process for DCO but I can ask.

That said, my request is in with Legal and I expect an answer soon.

StephenWeber:
I heard back from legal, and they did say that a DCO would be easier - they prefer to sign company-level CLAs. I haven't gotten confirmation back that I can sign the individual CLA but I expect that to be approved this week.

In the meantime, I suppose I could also branch off of v3.2 and provide a similar PR for that in case a patch ever is released for that? My deployment is using v3.2 for the time being.

StephenWeber:
Integration test error:

    Python wrapper process script took more than 0.31 seconds to execute (0.3175). This most likely means
    that a direct or in-direct import of a module which takes a long time to load has been added (e.g.
    jsonschema, pecan, kombu, etc).`

I know I'm adding a new project import, and it's not over by much (0.0175 seconds) - unfortunately passing tests do not print out the actual value so I don't know how much import time I'm adding to the system with this change..

StephenWeber:
Integration tests failed to complete in time, closing and reopening ticket to trigger a new build.

StephenWeber:
Now the integration tests are back to failing for `test_process_wrapper_exits_in_reasonable_timeframe`, I'll see about running these locally... but I probably need help...

PR #4983, opened 2020-07-02
Remove authentication headers from webhook payloads

potato:
Authentication headers (along with authentication cookies) shouldn't be stored in the payload of the trigger instance, since authentication data from `GET` query parameters are not stored either.

Not filtering in `_log_request` since on `DEBUG` loglevel those can be useful data.


st2api/tests/unit/controllers/v1/test_webhooks.py
```diff
@@ -319,6 +319,26 @@ def get_webhook_trigger(name, url):
...
+        return_value=True))
+    @mock.patch.object(WebhooksController, '_is_valid_hook', mock.MagicMock(
+        return_value=True))
+    @mock.patch.object(HooksHolder, 'get_triggers_for_hook', mock.MagicMock(
+        return_value=[DUMMY_TRIGGER_DICT]))
+    @mock.patch('st2common.transport.reactor.TriggerDispatcher.dispatch')
+    def test_authentication_headers_should_be_removed(self, dispatch_mock):
+        headers = {
+            'Content-Type': 'application/x-www-form-urlencoded',
+            'St2-Api-_Key': 'foobar',
```
armab, line 332:
```suggestion
            'St2-Api-Key': 'foobar',
```

Looks like a typo here?

potato: Nice catch, a typo indeed, my bad.

potato:
updated the `CHANGELOG.rst`

st2api/st2api/controllers/v1/webhooks.py
```diff
@@ -218,6 +220,10 @@ def _get_headers_as_dict(self, headers):
             headers_dict[key] = value
         return headers_dict
 
+    def _filter_authentication_headers(self, headers):
+        auth_headers = [HEADER_API_KEY_ATTRIBUTE_NAME, HEADER_ATTRIBUTE_NAME, 'Cookie']
```
armab, line 224: Filtering out `St2-Api-Key` and `X-Auth-Token` is great.
However `Cookie` may not necessarily contain auth data, but some useful params to use and filter as payload.

@StackStorm/maintainers any opinions?

potato: May not but could if I understand `st2common/st2common/router.py#L278` right. Another option is to parse cookies here or pass the `req.cookies` from the router to the controller.

armab: This fix is marked for the upcoming `v3.3.0` release.

@StackStorm/maintainers @StackStorm/contributors any opinions about this special case? WDYT?

nmaludy (changes requested):
I like this idea a lot. Probably good to list out what headers are going to be stripped out in the CHANGELOG and probably also in `st2docs`

nmaludy (approved):

armab:
Thanks @potato @knagy for the contribution and @nmaludy for review! :+1:

PR #4984, opened 2020-07-02
Remove EL6 support

amanda11:
EL6 is being deprecated. Remove running of builds in EL6.

Remove rhel-rpm in Makefiles, and keep the rpm for EL7 and above. I was a bit confused with the name rhel-rpm - it seemed to just use the el6 specs - and so I was presuming it was no-longer required - but would need confirmation.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -107,7 +107,7 @@ jobs:
 
   # Build & Test st2 packages
   packages:
-    parallelism: 5
+    parallelism: 4
     # 4CPUs & 8GB RAM CircleCI machine
     # sadly, it doesn't work with 'setup_remote_docker'
     resource_class: large
@@ -116,7 +116,7 @@ jobs:
       - image: circleci/python:2.7
     working_directory: ~/st2
     environment:
-      - DISTROS: "xenial bionic el6 el7 el8"
+      - DISTROS: "xenial bionic el7 el8"
       - ST2_PACKAGES_REPO: https://github.com/StackStorm/st2-packages
       - ST2_PACKAGES: "st2"
       - ST2_CHECKOUT: 0
@@ -232,7 +232,7 @@ jobs:
       - image: circleci/ruby:2.4
     working_directory: /tmp/deploy
     environment:
-      - DISTROS: "xenial bionic el6 el7 el8"
+      - DISTROS: "xenial bionic el7 el8"
     steps:
       - attach_workspace:
           at: .
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -37,7 +37,7 @@ Fixed
 Removed
 ~~~~~~~
 
-* Removed EL6 #4984
+* Removed ``CentOS 6``/``RHEL 6`` support #4984
 
   Contributed by Amanda McGuinness (@amanda11 Ammeon Solutions)
   
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,7 @@ in development
 
 Added
 ~~~~~
+* Removed EL6 support
 * Add make command to autogen JSON schema from the models of action, rule, etc. Add check
   to ensure update to the models require schema to be regenerated. (new feature)
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,8 +6,6 @@ in development
 
 Added
 ~~~~~
-* EL6 deprecation. Removed building of EL6 RPMs and associated jobs
-  Contributed by Amanda McGuinness (@amanda11 Ammeon Solutions)
 * Add make command to autogen JSON schema from the models of action, rule, etc. Add check
   to ensure update to the models require schema to be regenerated. (new feature)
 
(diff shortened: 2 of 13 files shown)


CHANGELOG.rst
```diff
@@ -6,6 +6,8 @@ in development
 
 Added
 ~~~~~
+* EL6 deprecation. Removed building of EL6 RPMs and associated jobs
+  Contributed by Amanda McGuinness (@amanda11 Ammeon Solutions)
```
armab, line 10: This fits into `Changed` sub-category under the Changelog. Please take a look at examples from the previous release. Also please try to format this changelog message based on those examples.

st2actions/bin/runners.sh
```diff
@@ -28,7 +28,7 @@ elif [ -z "$sv" ] && ( /sbin/start 2>&1 | grep -q "missing job name" ); then
   sv=upstart
   svbin=$UPSTARTCTL
 else
-  # Old debians, redhats and centos, amazon etc
+  # Old debians, amazon etc
   sv=sysv
```
armab, line 32: Good that you found this file!

`EL6` was the last platform with `sysv` init system which relies on `/etc/init.d/` service startup files. All OS systems we support right now have `systemd` only.

Please try to cleanup this file based on that information, I'll help where I can.

.circleci/config.yml
```diff
@@ -107,7 +107,7 @@ jobs:
 
   # Build & Test st2 packages
   packages:
-    parallelism: 5
+    parallelism: 4
```
armab, line 110: :+1:

st2actions/Makefile
```diff
@@ -27,13 +27,6 @@ rpm:
 	cp packaging/rpm/$(COMPONENTS).spec $(RPM_SPECS_DIR)/
 	cd $(RPM_SPECS_DIR) && rpmbuild --clean --rmsource -ba $(COMPONENTS).spec
 
-.PHONY: rhel-rpm
```
armab, line 30: Just FYI the rpm/deb logic in the Makefiles of `st2` repository is really outdated and unused code we're guilty to remove before.
All packaging happens in https://github.com/stackstorm/st2-packages

amanda11:
Given comment about init.d then there is a python file that checks whether it’s systemd , upstart or sysv and complains if it’s none of those.
Shall I update that as well to remove the allowance of sysv?


Sent from my iPhone

> On 3 Jul 2020, at 14:39, Eugen C. <notifications@github.com> wrote:
> 
> @armab commented on this pull request.
> 
> In st2actions/Makefile:
> 
> > @@ -27,13 +27,6 @@ rpm:
>  	cp packaging/rpm/$(COMPONENTS).spec $(RPM_SPECS_DIR)/
>  	cd $(RPM_SPECS_DIR) && rpmbuild --clean --rmsource -ba $(COMPONENTS).spec
>  
> -.PHONY: rhel-rpm
> Just FYI the rpm/deb logic in the Makefiles of st2 repository is really outdated and unused code we're guilty to remove before.
> All packaging happens in https://github.com/stackstorm/st2-packages
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or unsubscribe.

-- 
Ammeon Solutions puts the security of the client at a high priority. 
Therefore, we have put efforts into ensuring that the message is error and 
virus-free. Unfortunately, full security of the email cannot be ensured as, 
despite our efforts, the data included in emails could be infected, 
intercepted, or corrupted. Therefore, the recipient should check the email 
for threats with proper software, as the sender does not accept liability 
for any damage inflicted by viewing the content of this email.

CHANGELOG.rst
amanda11: Sorry - didn't see the different sub-sections.
I see in 3.2.0 there actually was a Removed sub-category with "Removed Ubuntu 14.04 from test matrix #4897" in it.
Would you prefer it in Changed or Removed?

armab: Yes absolutely, `Removed` is even better :+1:
Didn't know we had that section. Thanks for noticing that!

amanda11:
> Given comment about init.d then there is a python file that checks whether it’s systemd , upstart or sysv and complains if it’s none of those. Shall I update that as well to remove the allowance of sysv?

It was runners.sh that also mentioned sysv. So I've updated that - but I'm not sure if its used, as the st2-packages/runners.sh seems to be what is installed - but that still mentions sysv - so not sure if there's a separate issue raised for updating st2-packages/runners.sh to remove the sysv support.

armab:
Yeah, I will handle st2-packages separately in another PR, per https://github.com/StackStorm/discussions/issues/39

CHANGELOG.rst
```diff
@@ -33,8 +33,14 @@ Fixed
   up correctly, specifically when specifying a bastion host / jump box. (bug fix) #4973
 
   Contributed by Nick Maludy (@nmaludy Encore Technologies)
-  
 
+Removed
+~~~~~~~
+
+* Removed EL6 #4984
```
armab, line 40:
```suggestion
* Removed ``CentOS 6``/``RHEL 6`` support #4984
```

st2common/bin/st2ctl
```diff
@@ -115,10 +115,8 @@ function service_manager() {
     /sbin/initctl $action $svcname
   elif command -v service > /dev/null 2>&1; then
     service $svcname $action
-  elif [ -x /etc/init.d/${1} ]; then
-    /etc/init.d/$svcname $action
```
armab, line 119: good catch for `st2ctl` :+1:

PR #4992, opened 2020-07-14
Fix st2.action.file_writen typo

blag:
PR #1785 introduced a typo in the `st2.action.file_writen` trigger name that has gone apparently unnoticed for five years.

This PR simply fixes the typo.

Please note that this is not reverse compatible, and so it should be noted in the upgrade documentation. However, the fact that this hasn't ever been reported or fixed would seem to indicate that not many people use this trigger, so hopefully the impact is minimal.

A companion PR for st2docs is incoming.


blag:
Since this is a backwards-incompatible change, I would recommend against merging this into ST2 `v3.2.1`.

CHANGELOG.rst
```diff
@@ -35,6 +35,8 @@ Fixed
   up correctly, specifically when specifying a bastion host / jump box. (bug fix) #4973
 
   Contributed by Nick Maludy (@nmaludy Encore Technologies)
+* The built-in ``st2.action.file_writen`` trigger has been renamed to ``st2.action.file_written``
```
armab, line 38: Looks like such a rename would be a better fit the `Changed` category rather then `Fixed`?

PR #5002, opened 2020-07-21
Use upstream codecov instead of our own fork

nmaludy:
We've been seeing repeated build failures for codecov issues: https://travis-ci.org/github/StackStorm/st2/jobs/710376520#L27406

The fix is to upgrade to `codecov-python >= 2.1` per https://community.codecov.io/t/http-400-while-uploading-to-s3-with-python-codecov-from-travis/1428/6

However, we are maintaining our own fork of `codecov-python` and those changes haven't been merged into our fork.

Couple things here:

- It looks like we're maintaining our own fork so that we can get better error outputs: https://github.com/codecov/codecov-python/compare/master...StackStorm:better_error_output
- I've submitted a PR upstream to include these changes so we don't have to maintain our own fork: https://github.com/codecov/codecov-python/pull/288
- Once the PR above is merged we can merge this PR

Until we get the changes merged upstream, i've also submitted a PR against our own `codecov-python` fork to pull in the upstream changes so our tests can pass again without these errors: https://github.com/StackStorm/codecov-python/pull/2

**DECISION POINTS / OPTIONS**

1. If we don't mind having the "better error messages" shown in case of `codecov` failure, we can merge this PR and just use upstream
  - At some point in the future our "better error messages" PR will be merged and we'll take advantage of that automatically
2. We can let this PR sit until the upstream changes are merged and instead merge PR https://github.com/StackStorm/codecov-python/pull/2 so we can stop getting errors in our builds.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -55,6 +55,7 @@ Removed
 * Removed ``CentOS 6``/``RHEL 6`` support #4984
 
   Contributed by Amanda McGuinness (@amanda11 Ammeon Solutions)
+* Removed our fork of ``codecov-python`` and have switched back to the upstream version
   
 3.2.0 - April 27, 2020
 ----------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -55,7 +55,7 @@ Removed
 * Removed ``CentOS 6``/``RHEL 6`` support #4984
 
   Contributed by Amanda McGuinness (@amanda11 Ammeon Solutions)
-* Removed our fork of ``codecov-python`` and have switched back to the upstream version
+* Removed our fork of ``codecov-python`` for CI and have switched back to the upstream version (improvement) #5002
   
 3.2.0 - April 27, 2020
 ----------------------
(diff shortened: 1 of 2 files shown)


punkrokk:
Just restarted the failed job

CHANGELOG.rst
```diff
@@ -55,6 +55,7 @@ Removed
 * Removed ``CentOS 6``/``RHEL 6`` support #4984
 
   Contributed by Amanda McGuinness (@amanda11 Ammeon Solutions)
+* Removed our fork of ``codecov-python`` and have switched back to the upstream version
```
armab, line 58: Would be good to reference the actual PR #
```suggestion
* Removed our fork of ``codecov-python`` for CI and have switched back to the upstream version (improvement) #5002
```

Not sure though if this enhancement really fits the `Removed` Changelog section as that's not the primary category.
Eg it's not a deprecation of the previously supported OS or st2.conf setting removal or change in some core st2 functionality we need to highlight in the `Removed` section.

Issue #4762, opened 2019-08-08
Mistral Deprecation Game Plan


armab:
Mistral deprecation in favor of new workflow engine Orquesta was communicated for quite a long time.
StackStorm v3.2 will be latest version to ship Mistral with full removal in st2 `v3.3` per [Roadmap](https://docs.stackstorm.com/latest/roadmap.html).

> There is a helper `Mistral` -> `Orquesta` conversion tool available:
> https://github.com/EncoreTechnologies/orquestaconvert 
> Special thanks @nmaludy an EncoreTechnologies for the tool

### The Plan: https://github.com/orgs/StackStorm/projects/16
- [x] Identify & Migrate any missing mistral workflows in StackStorm Exchange
- [x] Remove from `st2ci` testing and `st2cd` release automation
  - [x] st2ci https://github.com/StackStorm/st2ci/pull/189 (@amanda11)
  - [x] st2cd https://github.com/StackStorm/st2cd/pull/440 (@amanda11)
- [x] `st2` code cleanup to remove the mistral integration https://github.com/StackStorm/st2/pull/5011 (@amanda11)
- [x] st2-packages
  - [x] Remove `deb`/`rpm` rules from `st2-packages` https://github.com/StackStorm/st2-packages/pull/656 (@amanda11)
  - [x] Remove mistral from scripted curl-bash installer https://github.com/StackStorm/st2-packages/pull/657 (@amanda11)
  - [ ] Later step: move `st2-packages` rpm/deb scriptlets into `st2` https://github.com/StackStorm/st2-packages/issues/662 (@armab)
- [x] Remove Mistral from [`st2tests`](https://github.com/stackstorm/st2tests) https://github.com/StackStorm/st2tests/pull/190 (@amanda11)
- [x] Close all Mistral issues
- [x] Remove mistral from Deployments
  - [x] ~`st2-docker`~
  - [x] ~`Chef`~ (deprecated)
  - [x] Ansible https://github.com/StackStorm/ansible-st2/issues/270 (@amanda11)
  - [x] Puppet https://github.com/StackStorm/puppet-st2/pull/313 (@nmaludy) 
- [x] Remove mistral from the launch_dev.sh script (https://github.com/StackStorm/st2/issues/5012)
- [x] Update docs
  - [x] https://github.com/StackStorm/st2docs/pull/1009 (@amanda11)
  - [x] https://github.com/StackStorm/st2docs/pull/1010 (@winem)
[...]

PR #5011, opened 2020-08-04
Initial mistral removal from st2

### LINKED ISSUE #5012: Remove mistral from the launch_dev.sh script
The[ launch_dev.sh](https://github.com/StackStorm/st2/blob/master/tools/launchdev.sh#L96) script has funtionality that needs to get removed so it doesn't try to spin up mistral things

amanda11:
Remove mistral and postgres from st2ci workflows as part of the Mistral deprecation

Addresses st2from StackStorm/st2#4762

Resolves #5012


amanda11:
Initial mistral removal, to see what happens on the CI runs.

amanda11:
Not ready for review, missed a whole bunch of changes!

armab:
FYI CircleCI build is green now as we merged the st2-packages change: https://github.com/StackStorm/st2-packages/pull/656

amanda11:
@armab Thanks for merge, the CircleCI tests are now passing.

I still need to work out why I've broken python3 - I would have expected that I'd break python2.7 unit-tests not python3. But I'm investigating this...

tox.ini
```diff
@@ -50,22 +50,9 @@ commands =
...
-setenv = PYTHONPATH = {toxinidir}/external:{toxinidir}/st2common:{toxinidir}/st2api:{toxinidir}/st2actions:{toxinidir}/st2exporter:{toxinidir}/st2reactor:{toxinidir}/st2tests:{toxinidir}/contrib/runners/action_chain_runner:{toxinidir}/contrib/runners/local_runner:{toxinidir}/contrib/runners/python_runner:{toxinidir}/contrib/runners/http_runner:{toxinidir}/contrib/runners/noop_runner:{toxinidir}/contrib/runners/announcement_runner:{toxinidir}/contrib/runners/remote_runner:{toxinidir}/contrib/runners/remote_runner:{toxinidir}/contrib/runners/mistral_v2:{toxinidir}/contrib/runners/orquesta_runner:{toxinidir}/contrib/runners/inquirer_runner:{toxinidir}/contrib/runners/http_runner:{toxinidir}/contrib/runners/winrm_runner
-         VIRTUALENV_DIR = {envdir}
-passenv = NOSE_WITH_TIMER TRAVIS
-install_command = pip install -U --force-reinstall {opts} {packages}
-deps = virtualenv
-       -r{toxinidir}/requirements.txt
-       -e{toxinidir}/st2client
-       -e{toxinidir}/st2common
-commands =
-    nosetests --rednose --immediate -sv contrib/runners/mistral_v2/tests/unit/
```
amanda11, line 64: As the unit tests nightly only seemed to do mistral - I removed them.

st2client/st2client/commands/action.py
```diff
@@ -1269,7 +1266,7 @@ def __init__(self, resource, *args, **kwargs):
                                  help='Name of the workflow tasks to re-run.')
         self.parser.add_argument('--no-reset', dest='no_reset', nargs='*',
                                  help='Name of the with-items tasks to not reset. This only '
-                                      'applies to Mistral workflows. By default, all iterations '
+                                      'applies to Orquesta workflows. By default, all iterations '
```
amanda11, line 1269: Option looks to have always been supported for Orquesta - it is in the documents and we had a unit-test, but the parser just referred to Mistral - so think it should have previously said both - so adjusted to Orquesta.

st2common/bin/st2-track-result
```diff
@@ -52,10 +52,6 @@ def add_result_tracker(exec_id):
     # Check runner type.
```
amanda11: If results tracker was only supported for mistral, then should this whole file go? Or should we leave it in with the runner_type != mistral-v2 so the logic is there if need to add for new runners?
What about the actual results tracker itself?

st2common/bin/st2-track-result
```diff
@@ -44,73 +44,5 @@ def setup():
     script_setup.setup(config, register_mq_exchanges=False)
 
 
-def add_result_tracker(exec_id):
```
m4dcoder, line 47: Can we keep the st2 result tracker and minimize changes to it?  st2 result tracker is a generic solution to pull results from a remote runner instead of having the runner push result into st2.  It is not specific to Mistral. Although Mistral is the only one that uses it.

st2common/bin/st2-track-result
```diff
@@ -44,73 +44,5 @@ def setup():
...
-
-    # Add result tracker for children workflows.
-    for child_exec_id in exec_db.children:
-        child_exec = ActionExecution.get(id=child_exec_id, raise_exception=True)
-        if child_exec.runner['name'] == 'mistral-v2':
-            LOG.info('Adding result tracker for children "%s"...', child_exec_id)
-            add_result_tracker(child_exec_id)
-
-
-def del_result_tracker(exec_id):
```
m4dcoder, line 93: See comment above.

st2actions/st2actions/container/base.py
```diff
@@ -249,7 +249,7 @@ def _clean_up_auth_token(self, runner, status):
         it doesn't throw.
         """
         # Deletion of the runner generated auth token is delayed until the token expires.
-        # Async actions such as Mistral workflows uses the auth token to launch other
+        # Async actions use the auth token to launch other
```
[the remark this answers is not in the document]
m4dcoder: Keep this as is because the concept of async actions is generic although Mistral is the only runner that requires this at the moment.

st2common/tests/unit/test_query_base.py
```diff
@@ -40,9 +40,9 @@ def test_fire_queries_doesnt_loop(self):
         mock_query_state_1 = QueryContext(
             uuid.uuid4().hex,
             uuid.uuid4().hex,
-            'mistral_v2',
+            'orquesta_runner',
```
m4dcoder, line 43: The orquesta runner doesn't use the Querier. Mistral is the only one that uses the querier.  I want to preserve the generic solution here.  Can we add a mock async runner in the test setup and use that here instead?

st2common/tests/unit/test_runners_base.py
```diff
@@ -35,23 +35,11 @@ def test_get_runner_failure_not_found(self):
         self.assertRaisesRegexp(ActionRunnerCreateError, expected_msg,
                                 get_runner, 'invalid-name-not-found')
 
-    def test_get_query_module_success(self):
```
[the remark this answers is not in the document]
m4dcoder: Yes, Mistral is the only one async runner that uses the query/callback.  Can we set up a mock async runner in the test setup and then use that here so we preserve the functions and tests for any async runner in the future?

st2common/bin/st2-track-result
amanda11: > Can we keep the st2 result tracker and minimize changes to it? st2 result tracker is a generic solution to pull results from a remote runner instead of having the runner push result into st2. It is not specific to Mistral. Although Mistral is the only one that uses it.

Hi @m4dcoder I've put back the st2-track-result back to where it was, but with an additional comment to explain why it's there, and to update when other runners support the result tracker.

st2common/tests/unit/test_query_base.py
amanda11: I'll look into getting that mock runner in this week, and update tests to use the mock runner.

st2common/tests/unit/test_runners_base.py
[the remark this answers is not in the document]
amanda11: @m4dcoder Is there a way to get it to load mock runners. At moment it looks like the unit-test will only find the real modules, not any in fixtures. I had a play with the fixturesloader, but it doesn't seem to load them at the right level - just in the db. 
So the stevedore doesn't load them. 

I could MagicMock at the stevedore level, but I wasn't sure if we wanted to test with that in.

Any thoughts? (For now I've put back the tests but commented out)

st2client/tests/unit/test_execution_tail_command.py
```diff
@@ -130,130 +129,6 @@
...
 }
 
-# Mock objects for Mistral workflow execution
-MOCK_LIVEACTION_4_RUNNING = {
-    'id': 'idfoo4',
-    'status': LIVEACTION_STATUS_RUNNING
-}
-
-MOCK_LIVEACTION_4_CHILD_1_RUNNING = {
-    'id': 'idmistralchild1',
```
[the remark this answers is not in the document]
amanda11: > That would be nice if you can take a look and give it a try. The tail output for orquesta workflow is different though. Thanks!

Taking a look at the output in the scheduler log then it looks like the children of an orquesta workflow have the same fields like task_name but in an orquesta context. There may be different other parameters, but out of the ones that are mocked in these replies they seem to be the same.
I also took a look at a tail of an orquesta log and the output looks the same.

If I'm missing other changes to make to these mock objects please let me know.

CHANGELOG.rst
```diff
@@ -78,6 +78,9 @@ Changed
 Removed
 ~~~~~~~
 
+* Removed ``Mistral``
```
armab, line 81: Minor recommendation: 
```suggestion
* Removed ``Mistral`` workflow engine (deprecation) #5011
```

at least, having a PR # number in the changelog is helpful.

st2tests/st2tests/fixtures/generic/workflows/wb_invalid_syntax.yaml
```diff
@@ -1,72 +0,0 @@
-version: "2.0"
-name: "examples.mistral-complex"
```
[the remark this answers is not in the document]
LindsayHill: I don't think you need to worry about ensuring there is an Orquesta equivalent as part of this change. It would be good to have those, but there is a decent set of Orquesta examples, so don't worry too much about it right now.

armab:
@m4dcoder looks like this mistral removal PR needs a second review round?

PR #5026, opened 2020-08-13
Deprecate mistral - remove mistral from the HA config and st2.*.conf samples

winem:
This PR goes along with https://github.com/StackStorm/st2docs/pull/1010/ which references some of the HA samples. 

I removed mistral relevant content from the HA samples as well as the st2.*.conf samples. 

Removing the mistral.dev folder requires updates on the launchdev.sh and db_cleanup.sh and I saw that there is another PR #5011 addressing this topic.


winem:
Oops, I missed that this is already covered by #5011. So I'll close this one. :)

PR #5030, opened 2020-08-31
Investigate restricting cryptography [WIP] 

amanda11:
Investigate if packing error is due to cryptography version


amanda11:
This didn't resolve the problem with the build - so deleting.

Issue #4972, opened 2020-06-24
EL7/U16 -> Mongo 4.0 Master Issue


punkrokk:
Since Mongo 3.4 has been deprecated, we need to following Issues resolved in order to update/support Mongo ~3.6~ 4.0 in ST2 v3.3.0

- [x] stackstorm/st2-packages#643 one line installer scripts
- [x] StackStorm/ansible-st2#264 ansible
- [x] StackStorm/puppet-st2#298 puppet
- [x] StackStorm/st2docs#988 st2docs
- [x] st2docs https://github.com/StackStorm/st2docs/blob/master/docs/source/install/__mongodb_note.rst https://github.com/StackStorm/st2docs/pull/994
- [x] update TravisCI and CircleCI to use mongodb `4.0` instead of `3.4`

armab:
RE st2docs, the following documentation page will need an update too:
https://github.com/StackStorm/st2docs/blob/master/docs/source/install/__mongodb_note.rst

amanda11:
@punkrokk I think I may have covered the st2docs change under this one: https://github.com/StackStorm/st2docs/pull/994/files (was amending a few other files when removing EL6 for functionality I knew we'd changed!)

amanda11:
@punkrokk In the st2 repo. In the circleci config file. When it runs the integration tests it points to the docker mongo:3.4. Does that need to be altered to point to mongo 4.0 as part of this issue?

armab:
For future reference, ideally not to mix several logically independent changes in one PR like EL6 deprecation _and_ MongoDB 4.0 _and_ EL8 link update _and_ something else.
It's always better to create several dedicated PRs for each change for easy review and change/history reasons.

punkrokk:
@armab Good call, I think I lost focus on what the actual issue was !!!

amanda11:
> For future reference, ideally not to mix several logically independent changes in one PR like EL6 deprecation _and_ MongoDB 4.0 _and_ EL8 link update _and_ something else.
> It's always better to create several dedicated PRs for each change for easy review and change/history reasons.

Point noted. Got carried away when noticing errors in docs. Will do separate PRs next time to make it easier to review/track.

amanda11:
@punkrokk I noticed that in st2 repo the travis and circleci still refer to mongo 3.4 for their runs on python 2.7, not mongo 4.0. Do we need another checkbox to resolve those?

punkrokk:
I think so. @armab?

[...]

PR #5032, opened 2020-09-03
Update ST2 CircleCI and Travis to use mongodb 4.0 instead of 3.4

amanda11:
Update CircleCI and Travis to use mongodb 4.0

Addresses CircleCI/Travis part of https://github.com/StackStorm/st2/issues/4972


armab (approved):
Thanks :+1:

That's also a good find about mongo 3.4 in CI.

PR #5034, opened 2020-09-03
remove py3-unit-nightly from Travis runs

amanda11:
py3-unit-nightly was removed as part of mistral deprecation, as it just ran the mistral UTs, but Travis run still called it. Remove call to it from travis runs to stop error on overnight runs.


amanda11:
Example of a nightly build where it attempts to run the nightly build - but it failed as tox environment didn't exist. But the Makefile target still existed - so the run nightly if exists attempted to run it.

armab (approved):
:+1:

Issue #4999, opened 2020-07-20
Add tests for all linux pack actions


blag:
The [contrib `linux` pack](https://github.com/StackStorm/st2/tree/master/contrib/linux) has a few broken actions (`check_loadavg`,  `check_processes`, `dig`: #4993, `diag_loadavg`, possibly more) at least on CentOS 8. It's really disheartening to spin up StackStorm and immediately run into broken actions.

We should add tests for those actions to the pack.

The [tests](https://github.com/StackStorm/st2/tree/master/contrib/core/tests) from the [`core`](https://github.com/StackStorm/st2/tree/master/contrib/core) pack can be used as a template for writing tests for the `linux` pack.

I will try to add tests for the `dig` action in my PR #4993, to serve as a concrete example, but adding tests would be a useful way to introduce somebody to developing StackStorm that would have an immediate impact.

blag:
Copying and pasting this from Slack:

For the `netstat` action there really isn’t much to do since it’s a straight `remote-shell-cmd` action, and it only has a single user-specified parameter, `args`. So this one probably doesn’t need a unit test.

Same story for the `cp`, `mv`, `netstat_grep`, `pkill`, `rm`, `rsync`, `scp`, and `vmstat` actions.

I’m a little confused why the `file_touch` action doesn’t use the touch command…that could probably be changed, but if you do I would rename the existing `file_touch` action to something like `file_init`.

For the service action, it’s a `remote-shell-script` but it’s actually a Python script, so you can probably import directly from it, mock up and monkeypatch the `sys` and `subprocess` modules in it, and then assert that the right command arguments are passed to `subprocess.call`.

`check_loadavg` needs some tests, I would mock up and monkeypatch `loadAvgFile` and `cpuInfoFile`, override `sys.stdout` and check its output. Once you’ve got some tests, it then needs to be updated to support both Python 2 and 3 (`except Exception:`).

`check_processes` needs tests - you don’t need to test every method, but I would test at least `byState`, `byPid`, and `byName`. First, choose a pid number and set the action instance's `myPid` property to it, then create a test fixture directory that contains directories with purely numeric names, a few with non-numeric names, and a name that matches your chosen `myPid` number (say: `3456`. Make sure you put `stat` and `cmdline` files in each pid directory:

```
proc_fixture/
proc_fixture/1234/
proc_fixture/1234/stat
proc_fixture/1234/cmdline
proc_fixture/2345/
proc_fixture/2345/stat
[...]

Issue #5004, opened 2020-07-23
ImportError: No module named st2common.util.shell when run linux.service


dove-young:
## SUMMARY

Provide a quick summary of your bug report.

```
st2 run linux.service service=mongod act=restart hosts=agentavtone-worker1
...
id: 5f191803f06137c07014deb3
action.ref: linux.service
context.user: st2admin
parameters:
  act: restart
  hosts: agentavtone-worker1
  service: mongod
status: failed
start_timestamp: Thu, 23 Jul 2020 04:54:27 UTC
end_timestamp: Thu, 23 Jul 2020 04:54:33 UTC
result:
  agentavtone-worker1:
    failed: true
    return_code: 1
    stderr: "0: 2 windows (created Thu Jul  9 00:06:22 2020)
1: 1 windows (created Tue Jul 21 20:58:59 2020)
Traceback (most recent call last):
  File "/tmp/5f191803f06137c07014deb2/service.py", line 23, in <module>
    from st2common.util.shell import quote_unix
ImportError: No module named st2common.util.shell"
    stdout: ''
    succeeded: false

```

### STACKSTORM VERSION

Paste the output of ``st2 --version``:

st2 3.2.0, on Python 2.7.5

##### OS, environment, install method

Post what OS you are running this on, along with any other relevant information/

- e.g. Docker, Vagrant, Kubernetes, etc. Describe how you installed ST2
- e.g. one-line install, custom install, etc -->

one-line install 

## Steps to reproduce the problem

Show how to reproduce the problem, using a minimal test-case. Make sure to include any content
(pack content - workflows, actions, etc.) which are needed to reproduce the problem.

## Expected Results

What did you expect to happen when running the steps above?

## Actual Results

What happened? What output did you get?

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

dove-young:
I've tried to create virtualenv for linux pack `st2 run packs.setup_virtualenv packs=linux`,  then I did these steps to debug

```
cd /opt/stackstorm/virtualenvs/linux
source bin/activate
python

>>> from st2common.util.shell import quote_unix
```

I found `oslo_config` is missing, then I installed it via 

```
source bin/activate
bin/pip install oslo_config
```

then import `st2common.util.shell` in python works

```
python

>>> from st2common.util.shell import quote_unix
```

[...]

PR #5035, opened 2020-09-04
Fix linux pack actions For CentOS 8

winem:
Part of #4999

This PR addresses the actions that fail on CentOS8 but work on other operating systems and distributions and writes the output of the load_diag actionchain to a json file which is parsable via jq or other tools and therefore hopefully a bit more useful to the users.

I did not add any tests yet. I would do this in another PR or give others the chance to add tests.


contrib/linux/actions/checks/check_processes.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
```
armab, line 1: Unsure if we could use python 2 like that, considering the fact that platform should be able to run in both environments.
Eg. there might be the cases when py2 available, while py3 is not present in the system and vice versa.

winem: Oh sorry, I should have mentioned that in the original post. So I saw that we solved the issue with `python2` in other places, too. After some testing on Ubuntu 14.04-18.04 and CentOS6-8 I saw that all of them have python2 available per default and I also remember a short slack discussion on this topic. 

So I think this is not really any risk but should be in the back of our mind when we deprecate Python2. This is just another topic that'll be way easier to handle once we support Python3 only.

Anyway, I'm all ears if you have any better idea to evaluate or maybe even a solution that truly supports both Python versions. :)

armab: Besides of VM there might be also Docker environment running without py2 installation.
We just can't add more `py2`-only things, considering active development and aim towards the [Python 2 Deprecation](https://github.com/StackStorm/discussions/issues/40), otherwise it'll just make the migration harder for us.

armab (changes requested):
Thanks for trying fixing this EL8 issue!

Per comments in this PR, we can't rely on a `python 2` env only considering active development and aim towards the Python 2 Deprecation.

There might be other ways to explore to fix the pack.

armab:
BTW is this an issue only on EL8 or in any python3 environment, including Ubuntu Bionic?

winem:
This issue applies to CentOS8 only. Ubuntu 18.04 has `python` in the path which is linked to one of the installed Python versions. I don't know for sure which one but I think it's the Python3.x 

That's why a shebang like '#!/usr/bin/env python` works on the supported Ubuntu distributions.

contrib/linux/actions/checks/check_processes.py
winem: So, can we live with these broken actions as "known issues" in py2-Environments and support Py3 only, here? 

For me, this would be fine. I mean, in the end it was one of the team who noticed the issue(s) and we did not get any complaints from users yet. I know this is not the best from the users perspective but handling it as known issue may work for the meantime.

armab:
Can you elaborate more on the root cause of the issue? Is just `#!/usr/bin/env python` failing to find the python binary in EL8 system?
If so, can we hardcode it and use the st2 py one, like `#!/opt/stackstorm/st2/bin/python`? This would also use the st2 virtualenv with all its installed dependencies.

winem:
> Can you elaborate more on the root cause of the issue? Is just `#!/usr/bin/env python` failing to find the python binary in EL8 system or something else?
> If so, can we hardcode it and use the st2 py one, like `#!/opt/stackstorm/st2/bin/python`? This would also use the st2 virtualenv with all its installed dependencies.

Exactly, that should to it and seems to be an optimal solution. I'll test this tonight!

CHANGELOG.rst
```diff
@@ -75,6 +75,9 @@ Fixed
 * Fixed a regression in the ``linux.dig`` action on Python 3. (bug fix) #4993
 
   Contributed by @blag
+* Fix actions from the contrib/linux pack that fail on CentOS-8 but work on other operating systems and distributions.
+
+  Reported by @blag and contributed by winem (bug fix) #4999
```
armab, line 80: A very small adjustment to the Changelog format:

```suggestion
* Fix actions from the contrib/linux pack that fail on CentOS-8 but work on other operating systems and distributions.  (bug fix) #4999

  Reported by @blag and contributed by @winem
```

winem: I actually had #4999 in mind when I wrote the changelog. But I don't mind to mention @dove-young as well although I don't tackle the issue with the linux.service action here.

winem:
> Can you elaborate more on the root cause of the issue? Is just `#!/usr/bin/env python` failing to find the python binary in EL8 system or something else?
> If so, can we hardcode it and use the st2 py one, like `#!/opt/stackstorm/st2/bin/python`? This would also use the st2 virtualenv with all its installed dependencies.

Thanks, this works like a charm and it's finally a solution that satisfies me! Latest commit includes the change.

CHANGELOG.rst
winem: I updated the changelog and @dove-young is mentioned now, too. Using the st2 py fixes the issue with the linux.service action as well. So all actions in the linux pack will be fixed with this PR now.

contrib/linux/actions/workflows/diag_loadavg.yaml
```diff
@@ -47,5 +43,5 @@
       name: "dump_results"
       ref: "core.local"
       parameters:
-        cmd: "echo \"ST2 Workflow\tdiag_loadavg:\t{{__results}}\" >> /tmp/diag_loadavg && echo 'Output written to file /tmp/diag_loadavg'"
+        cmd: "echo \"{{__results}}\" | python2 -c \"import json; import sys; from datetime import datetime; diag_data=eval(sys.stdin.read()); diag_data['timestamp'] = str(datetime.now()); print(json.dumps(diag_data))\" >> /tmp/diag_loadavg.json && echo 'Output written to file /tmp/diag_loadavg.json'"
```
armab, line 46: Looks like this is the only place with the hardcoded `python2` now.
Can we do something with it?

CHANGELOG.rst
armab: Thanks! I just was confused by the #5004 reference. Looks clear now :+1:

contrib/linux/actions/workflows/diag_loadavg.yaml
Kami: Yeah, any reason why we need to use ``python2`` here? I believe that line should also work fine with Python 3 (aka just using ``python`` system binary - should work if it's either python 2 or python 3).

winem: I don't see any issues regarding py3 and it should work as well (I can put that on my list again). 

Let's also tackle the `eval`. The safe alternative without a bunch of JSON processing on stdin is ast.literal_eval: https://docs.python.org/3/library/ast.html#ast.literal_eval

Would this be fine for everyone? Last resort would be some manual parsing & processing with the JSON library.

cognifloyd: `ast.literal_eval` is awesome (and quite safe)! go for it.

cognifloyd (commented):
It doesn't look like the `black` reformat will interfere with this PR much. (hooray!) Master still needs to be merged in though (looks pretty minor).

contrib/linux/actions/workflows/diag_loadavg.yaml
cognifloyd: Is this sufficient?
```suggestion
        cmd: "echo \"{{__results}}\" | python3 -c \"import json; import sys; from datetime import datetime; diag_data=eval(sys.stdin.read()); diag_data['timestamp'] = str(datetime.now()); print(json.dumps(diag_data))\" >> /tmp/diag_loadavg.json && echo 'Output written to file /tmp/diag_loadavg.json'"
```

cognifloyd:
I merged in master. Hopefully we can get this bugfix over the hump quickly. If not in 3.6.0, then in 3.7.0.

cognifloyd (approved):

armab (approved):
There's a need to update the pack version following the changes in this PR:
https://github.com/StackStorm/st2/blob/a82b1bedb74a69ea173e085fbefb4a532b75cdef/contrib/linux/pack.yaml#L19

otherwise looks good.

PR #5038, opened 2020-09-10
Fix for issue 4998 - Web Hook Rules check http headers in case sensitive manner

### LINKED ISSUE #4998: Web Hook Rules check http headers in case sensitive manner
## SUMMARY

The case used for the header name in trigger.headers[<headername>] in a web-hook rule is treated in a case sensitive manner. HTTP headers are case insensitive so the case of the name in the headers should not e relevant.

### STACKSTORM VERSION

3.2.0

Seen on one-line install and HA

See https://github.com/StackStorm/st2/issues/4995 for initial case.
1. Configure webhookrule with trigger.headers['X-GitHub-Event']
2. Send in header via curl of X-GitHub-Event to webhook
3. Rule doesn't match
4. Change rule to be trigger.headers['X-Github-Event'] - rule matches

## Expected Results

As http headers are case insensitive then it should not matter what case is used in the rule. Therefore no matter what case header is or case of rule then they should match.

## Actual Results

Only matched when rule defined as X-Github-Event

Rand01ph:
I have add a payload['headers_lower'] for webhook and add a unittest for this.
Fixes #4998


cognifloyd:
FYI: we just merged a change that reformats all the code with `black`. Please rebase on master (or merge master into your branch) and reformat your code with black as well. You can use `pre-commit install` to have `black` run on your commits. Thanks!

st2api/st2api/controllers/v1/webhooks.py
```diff
@@ -172,6 +172,7 @@ def post(self, hook, webhook_body_api, headers, requester_user):
             payload = {}
 
             payload["headers"] = headers
+            payload["headers_lower"] = dict((k.lower(), v) for k, v in headers.items())
```
cognifloyd, line 175:
```suggestion
            payload["headers_lower"] = {k.lower(): v for k, v in headers.items()}
```

cognifloyd (changes requested):
@Rand01ph could you please sign the CLA? (If you signed it before, you'll need to resign it now that it says "Linux Foundation" instead of one of the previous StackStorm stewards).

Also, I just resolved the conflicts, merged master into your branch, and reformatted with black.

I have one style suggestion.

I would like to merge your PR fairly quickly if you can help me at least with the CLA. Hopefully there's not much left before we can merge this.

armab:
Merged, thanks!

Issue #4842, opened 2020-01-14
st2client : Pack Install is hanging with HTTPS


harishgopalan:
Hello Stackstorm team,

We're noticing an issue with "st2 pack install file://$PWD" command with HTTPS endpoints.

For eg: if we set the below environment variables and try to install a pack , it works without any issues.

export ST2_API_URL=http://127.0.0.1:9101
export ST2_STREAM_URL=http://127.0.0.1:9102
export ST2_AUTH_URL=http://127.0.0.1:9100
export ST2_BASE_URL=http://127.0.0.1

But when we try to install a pack with the below environment variables wherein we route every request via Nginx reverse proxy/load balancer, the st2client fails to show the progress.

export ST2_BASE_URL=**https:**//127.0.0.1
export ST2_AUTH_URL=**https:**//127.0.0.1/auth
export ST2_API_URL=**https:**//127.0.0.1/api
export ST2_STREAM_URL=**https:**//127.0.0.1/stream
export ST2_CACERT=/etc/ssl/st2/st2.crt

Basically the pack install will show the progress like but the client won't exit and it would simply hang forever. But the pack install action chain would have really executed and completed without any issues.

![image](https://user-images.githubusercontent.com/2215990/72335082-6ac59b00-36e4-11ea-90a6-21dafb40e401.png)

Appreciate if any pointers to resolved this issue.

Regards
Harish

armab:
Try to run the CLI command with `--debug` flag on, see the API calls made, try to replicate the HTTP request via curl first.

Kami:
Does the same issue happen if you don't use SSL / TLS on the load balancer?

It appears it's something related to the streaming API endpoint from which client consumes the events.

It could be related to some load balancer keep alive or similar mechanism, but I think client should still handle that and we can't rely on server (aka stream API endpoint) closing the connection since that API endpoint is long running.

In any case, as @armab pointed out, gistting the whole output of the command when you include ``--debug`` flag would help.

harishgopalan:
Hi Kami,

No it doesn't happen with HTTP , it only hangs with HTTPS (Nginx : 443).

Below is the debug output. I suspect something to do with st2client streaming connection with st2stream where it is not properly getting closed.

Regards
Harish

[st2dev@st2-dev-1 ~ ]$ st2 --debug pack remove -d pack_install -t 1f3964cf767f4643a330269d5633dedc
# -------- begin 140659860837776 request ----------
[...]

PR #5042, opened 2020-09-16
Fix sse and nginx buffering

guzzijones:
# Environment:
`PYTHON_VERSION=python3.6 make requirements.txt`

setup nginx `st2.conf` listening on 8443 
```
server {
  listen       *:8443 ssl;
  client_max_body_size 5M;

  ssl on;

  ssl_certificate           /etc/ssl/st2/st2.crt;
  ssl_session_timeout       5m;

```

client call:
```
python -m st2client.shell --debug  --auth-url https://localhost:8443/auth --api-url https://localhost:8443/api --stream-url https://localhost:8443/stream --cacert=False --skip-config pack install /home/ubuntu/utilstack --python3
```
# What is the problem you're trying to solve with this patch?
https://github.com/StackStorm/st2/issues/4842

# What's the solution?
Solution is to watch the execution id status events and look for one of the LIVEACTION_STATUS_COMPLETED  events

# What's end event and end execution ID do?
tell the `generator` function in stream to stop sending events back to the client

# How is the cacert option has to do with this change? 
allows stream to use the --cacert option via command line.  Needed this to test the change using a self signed certificate.

# Proxy timeout? 
sets a reasonable timeout for connection and for read.  other endpoints have it as well.


guzzijones:
Well it turns out setting the timeout will cut off the stream going back to the client which causes other issues client side.  Going to have to keep digging here.

conf/nginx/st2.conf
```diff
@@ -85,6 +85,8 @@ server {
     proxy_set_header Host $host;
     proxy_set_header X-Real-IP $remote_addr;
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    proxy_read_timeout 10;
```
guzzijones, line 88:
```suggestion
    proxy_read_timeout 90;
```

guzzijones:
looks like circle and travis had issues.
afaik this PR should pass.  
Leaving it here for review.

guzzijones:
Looks like this section of code is missing unit testing. I will create a few.

m4dcoder (changes requested):
There's no description for the PR.  It looks like you're trying to fix a hang on your setup.  You have to provide more information in the description.  How is your environment setup?  What is the problem you're trying to solve with this patch?  What's the solution?  What's end event and end execution ID do?  How is the cacert option has to do with this change?  Proxy timeout?  There's a lot of dial that are touched but there's no comments and no description what these options are fixing.

guzzijones:
> Thank you for including detail explanation in the comments of this PR. This goes a long way to help reviewer understand the context of the change(s).
> 
> What happens if something is wrong and the expected end event and end execution ID never arrives? Will the client connection time out? What will the client output looks like in this case?

You would get the same UX as the current implementation of code without this change.  The client will hang on writing the results to stdout and you will need to kill the command with ctrl-c.

m4dcoder:
@guzzijones Thanks for the clarification. Ok with the fix as the events cover most completed statuses and failure case when there's a connection issue or other unforeseen errors.

m4dcoder (approved):
@guzzijones Thank you for contributing this fix.

armab (changes requested):
Just a reminder that we'll need a Changelog for this fix before merging.

guzzijones:
Ok will do.

CHANGELOG.rst
```diff
@@ -9,6 +9,11 @@ Changed
 * Improve the st2-self-check script to echo to stderr and exit if it isn't run with a
   ST2_AUTH_TOKEN or ST2_API_KEY environment variable. (improvement) #5068
 
+Fixed
+~~~~~~~~~
+* Fix nginx buffering long polling stream to client.  Instead of waiting for closed connection
+  wait for final event to be sent to client. fixes issue #4842
```
armab:
```suggestion
  wait for final event to be sent to client. (bug fix) #4842  #5042
```
to follow the existing format somewhat closely

guzzijones: done

armab (approved):

Issue #5041, opened 2020-09-15
Show & Log a py2 deprecation warning on st2 services startup


armab:
An addition to [Add python warning if install a pack that only supports python 2 #5037 PR](https://github.com/StackStorm/st2/pull/5037) and [Python 2 Deprecation Game Plan](https://github.com/StackStorm/discussions/issues/40), apart of warning for `pack install`, we might need to also warn users that's running StackStorm in py2 environment.

I'd suggest to show and log that warning for each st2 service during its initial startup, but open to other ideas.
Maybe also show for every `st2ctl` execution?

amanda11:
@armab  I think we can close this now?

armab:
@amanda11 Sure, thanks for the follow-up :+1: 
Feel free to manage/triage issues if you'll see anything like that in the future.

PR #5043, opened 2020-09-17
Add warning messages to logs if StackStorm running with python 2

amanda11:
Add warning messages in log files if StackStorm service is using python2

Warning message appears in log file, and also systemctl status (until journal will get wrapped).

Addresses part of #5041

Does not cover the reporting from st2ctl


amanda11:
Output from systemctl status on one of the stackstorm services just after starting:

`Sep 17 14:14:13 testsyslog st2workflowengine[14015]: 2020-09-17 14:14:13,852 WARNING [-] DEPRECATION WARNING. Support for python 2 will be removed in future StackStorm release(s). Please ensure that all packs used are python 3 compatible. Python 3 will already be used if you upgrade to a newer OS release`

Output from a log file on startup:
2020-09-17 14:14:06,917 139703825615920 WARNING service_setup [-] DEPRECATION WARNING. Support for python 2 will be removed in future StackStorm release(s). Please ensure that all packs used are python 3 compatible. Python 3 will already be used if you upgrade to a newer OS release.

Warning appeared in following log files:

- st2actionrunner.<pid>.log
- st2api.log
- st2auth.log
- st2garbagecollector.log
- st2notifier.log
- st2resultstracker.log
- st2rulesengine.log
- st2scheduler.log
- st2sensorcontainer.log
- st2stream.log
- st2timersengine.log
- st2workflowengine.log

Placement in the service_setup.py ensures that it goes to the log file as well as to systemctl status.

Does not cover the changes for reporting via st2ctl, will need to be covered in separate PR. Not added as would be in a different area of code.

st2common/st2common/constants/error_messages.py
```diff
@@ -32,3 +33,8 @@
 virtual environment with python2 binary:
 "st2 run packs.setup_virtualenv packs=%(pack)s python3=false"
 '''
+
+PYTHON2_DEPRECATION = 'DEPRECATION WARNING. Support for python 2 will be removed in future ' \
+                      'StackStorm release(s). Please ensure that all packs used are python ' \
+                      '3 compatible. Python 3 will already be used if you upgrade to a newer ' \
+                      ' OS release.'
```
armab, line 40: > Python 3 will already be used if you upgrade to a newer OS release.

I think this may sound a bit misleading due to generic message.

Maybe we could clarify it a bit? Something along the lines:
> Your StackStorm installation may be upgraded from python 2 to python 3 in the future platform releases. It's recommended to plan the manual migration to the python 3 native platform like Ubuntu 18.04 LTS or CentOS/RHEL 8.

WDYT? Feel free to adjust.

amanda11: Thanks for feedback, have made a change. What about following, or is too long?

`logs/st2workflowengine.log:2020-09-17 21:22:08,968 140485222490672 WARNING service_setup [-] DEPRECATION WARNING. Support for python 2 will be removed in future StackStorm releases. Please ensure that all packs used are python 3 compatible. Your StackStorm installation may be upgraded from python 2 to python 3 in future platform releases. It is recommended to plan the manual migration to a python 3 native platform, e.g. Ubuntu 18.04 LTS or CentOS/RHEL 8.`

st2common/st2common/service_setup.py
```diff
@@ -186,6 +187,9 @@ def setup(service, config, setup_db=True, register_mq_exchanges=True,
         register_service_in_service_registry(service=service, capabilities=capabilities,
                                              start_heart=True)
 
+    if sys.version_info[0] == 2:
+        LOG.warning(PYTHON2_DEPRECATION)
```
armab, line 191: Looking at the code, maybe it makes more sense to move this warning at earlier point, just after the
```
    LOG.debug('Using Python: %s (%s)' % (version, sys.executable))
```
?

amanda11: If you do that then it doesn't go to the log files. As it is before the logger is setup (I originally put it there until I realised that!)

st2common/st2common/constants/error_messages.py
armab: Looks good :+1:

st2common/st2common/service_setup.py
armab: Cool! So I've made same mistake :) 
Thanks for the explanation.

armab (approved):
LGTM :+1: 

Also it's nice to see that implementation for this is actually so easy & clean.

Issue #4938, opened 2020-05-06
[RFC]/[WIP] Add Python 2.7 deprecation warning


punkrokk:
At today's [TSC](https://github.com/StackStorm/st2/blob/master/GOVERNANCE.md) [meeting](https://github.com/StackStorm/discussions/issues/10), the subject of python 2.7 deprecation was discussed. Although the exact timeline for removing Python 2.7 support from is not yet known, the TSC agreed that it made sense to add a deprecation warning to installs that use Python 2.7. 

Things that need to be decided: 
- [ ] Do this by OS/Package or both
- [ ] When to show the `deprecation warning`
- [ ] other things I'm not thinking of

nmaludy:
Couple thoughts...
- Definitely python 3 only on new OSes
- I know there is a python3.6 install path in RHEL 7, so support for python3 only should be OK there as long as the user has the packages installed.
- Not sure about python3 on Ubuntu 16.04?
- The phase out time of rhel7/ubuntu16 might be longer than we want to keep python2 around? Especially considering rhel8 and ubuntu18 andare relatively new.

👍 for depreciation warnings for python 2 users. Probably want to show that at various parts of the cli and maybe install process... potentially web ui?

armab:
There are 2 types of deprecation warnings that might be useful:
- When using python2 OS (currently EL7 and U16)
- When using python2 packs
  - Output a warning during the `st2 pack install`
  - Print warnings in st2 logs

@blag Want to start the Python2 deprecation discussion and also the new project in Github so we could get everything organized?

cognifloyd:
Deprecation notices need to be in rpm, deb, and (at least for st2client) wheel.
Other possible places for deprecation warnings: the installers could print warnings, on startup (or even periodically?), all of the st2* packages could print and log a deprecation warning if running under py2.7. st2client could print a warning every time you auth or login.

Can we build rpm/deb variants so that people can select to install py3 before the official deprecation period is up on EL7/U16? Then, at the end of the period, the next release obsoletes both packages?

armab:
As we're getting closer to `v3.3.0` release, any takers to start adding the py2 deprecation warnings in various system parts?
As earlier we would start injecting it and communicating python 2 deprecation, - faster we can get rid of old python 2.

@StackStorm/maintainers and @StackStorm/contributors.

amanda11:
[...]

PR #5044, opened 2020-09-21
Add deprecation warning on st2ctl start if st2/python is python 2.x

amanda11:
Add warning messages in yellow to screen if perform st2ctl start (or restart) and /opt/stackstorm/st2/bin/python is python 2.x

Addresses part of #5041 and #4938

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,7 +27,7 @@ Added
   Python 2. (new feature) #5043
 
   Contributed by @amanda11
-* Added deprecation warning to st2ctl start,  if st2 python version is Python 2. (new feature) #5044
+* Added deprecation warning to st2ctl, if st2 python version is Python 2. (new feature) #5044
 
   Contributed by @amanda11
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,7 +27,7 @@ Added
   Python 2. (new feature) #5043
 
   Contributed by @amanda11
-* Added deprecation warning to st2ctl, if st2 python version is Python 2. (new feature) #5044
+* Added deprecation warning to st2ctl, if st2 python version is Python 2. (new feature) #5044 
 
   Contributed by @amanda11
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -27,6 +27,9 @@ Added
   Python 2. (new feature) #5043
 
   Contributed by @amanda11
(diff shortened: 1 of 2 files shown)


amanda11:
image

st2common/bin/st2ctl
```diff
@@ -87,6 +87,11 @@ function st2start() {
   for COM in ${COMPONENTS}; do
     service_manager ${COM} start
   done
+  PYTHON_VERSION=$(/opt/stackstorm/st2/bin/python --version 2>&1 | awk '{print $2'} | awk -F"." '{print $1}')
+  if [ "${PYTHON_VERSION}" = "2" ]; then
+    echo -e "\e[33mDeprecation warning: Support for python 2 will be removed in future StackStorm releases. Please ensure that all packs used are python 3 compatible. Your StackStorm installation may be upgraded from python 2 to python 3 in future platform releases. It is recommended to plan the manual migration to a python 3 native platform, e.g. Ubuntu 18.04 LTS or CentOS/RHEL 8. \e[0m\n"
```
armab, line 92: That looks nice! 

Can we include the Deprecation warning for the st2ctl reload/register actions as well to make it even more visible people can't ignore?

amanda11: That should be easy enough to add - will do.

nzlosh: Nothing major, but the second awk is redundant. A single call to `awk` can give the python major version, as follows

```
$ python --version 2>&1 | awk -F"[. ]" '{print $2}'
3
```

amanda11: Cool - will fix that up as well. And also try and remember the bit of ST2 I stole it from!

amanda11: @armab Added warning into st2ctl reload and st2ctl restart-component. (start and restart covered by logic already there). 
@nzlosh Thanks for the awk comment, much neater now, that's much neater (I'd stolen it from st2-self-check - but I won't patch that up under this PR).

CHANGELOG.rst
```diff
@@ -27,6 +27,9 @@ Added
   Python 2. (new feature) #5043
 
   Contributed by @amanda11
+* Added deprecation warning to st2ctl start,  if st2 python version is Python 2. (new feature) #5044
```
armab, line 30:
```suggestion
* Added deprecation warning to st2ctl, if st2 python version is Python 2. (new feature) #5044
```

armab (approved):
Looks good to me! :+1: 

Thanks @nzlosh for additional awk tips!

PR #5048, opened 2020-09-25
Update the st2 API validation on RBAC configuration

m4dcoder:
The RBAC backend will be named "default" instead of "enterprise" after it has been open sourced. Update the API validation to look for the "default" value instead of "enterprise". Minor update to comments related to the RBAC enterprise feature.


amanda11 (changes requested):
Looks good except I think each change needs an update to CHANGELOG.rst?

amanda11 (approved):

PR #5049, opened 2020-09-26
Run make configgen to update the sample st2 conf file

m4dcoder:
It seems the st2 configuration has changed. Update sample st2 conf file by running `make configgen`.


m4dcoder:
This looks like a side effect of moving to python3 for core development.

PR #5052, opened 2020-10-05
Enable redis server as coordination backend for platform development

m4dcoder:
A lot of workflow use cases that contain parallel branches, join task, and with items task require coordination service to be configured to be executed properly. For platform development, setup coordination service using redis server as backend by default to avoid confusion in troubleshooting and debugging workflows.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,15 +38,15 @@ jobs:
     # scripts/travis/run-nightly-make-task-if-exists.sh
     - name: "Unit Tests (Python 2.7 MongoDB 4.0)"
       python: 2.7
-      env: TASK=ci-unit CACHE_NAME=py2 COMMAND_THRESHOLD=700
+      env: TASK=ci-unit CACHE_NAME=py2 PYTHON_VERSION=python2.7 COMMAND_THRESHOLD=700
 
     - name: "Integration Tests (Python 2.7)"
       python: 2.7
-      env: TASK=ci-integration CACHE_NAME=py2 COMMAND_THRESHOLD=700
+      env: TASK=ci-integration CACHE_NAME=py2 PYTHON_VERSION=python2.7 COMMAND_THRESHOLD=700
 
     - name: "Lint Checks, Packs Tests (Python 3.6)"
       python: 3.6
-      env: TASK="ci-checks ci-packs-tests" CACHE_NAME=py3 COMMAND_THRESHOLD=430
+      env: TASK="ci-checks ci-packs-tests" CACHE_NAME=py3 PYTHON_VERSION=python3.6 COMMAND_THRESHOLD=430
 
     - name: "Unit Tests, Pack Tests (Python 3.6)"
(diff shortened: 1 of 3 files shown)


punkrokk:
I feel like this opens a can of worms. For example, how will we test that things continue to function where coordination is not present. If we are developing for the other? 

JP Bourget / @punkrokk

m4dcoder:
@punkrokk I'm not sure I know what you mean. Coordination service is really a requirement given that the number of workflow use cases that depends on it to work correctly. The only workflow use case that does not require coordination service are sequential workflows. For sequential workflows, the workflow engine does not invoke the coordination service whether it is configured or not. So configuring the coordination service has no bearing on testing these workflows.

amanda11:
I think in the documentation at the moment then it states that the coordination service is only required for the concurrency policy or HA.
I don't think it states it is needed for the other examples that have been listed, so do we also need a doc update?

armab:
Indeed coordination was really documented to be a requirement for Policies and HA only.

If the reality now that `[coordination]` is required for any normal Orquesta Workflow execution, - apart of just documentation update it should be added to the default StackStorm architecture and all the installers/deployment methods we have.

m4dcoder:
It really is required for any non-sequential workflows to function properly. The documentation should be updated.  We will save a lot of user headaches just including a coordination service backend as part of the StackStorm install.  For this PR here though, which is specifically to ensure development environment is setup properly at least, I'm proposing redis server as the backend.

guzzijones:
YES omg YES. I have been wanting this as well.  Orquesta workflows don't work correctly if they are non-sequential without a coordination backend.

punkrokk:
Ok, I see your point. I am also seeing situations where concurrency is enabled absent coordination and the scheduler service is very unhappy. (@m4dcoder )

armab (changes requested):
Per team feedback and as a follow-up,

The change here is part of the whole story of enabling `[coordination]` backend across the StackStorm fleet. From what I understand the following might be involved:
- documentation update signalizing that [coordination] is now required/enabled
- scripted installer changes to include Redis in default installation

That sounds like a major item for a release.
I believe we all care about the user experience making sure Orquesta workflows are executing fine in production, not only in dev env. 

@m4dcoder Do you think you can target next `v3.5.0` release with these major set of changes?

cognifloyd (approved):
Once merge conflicts are resolved, LGTM.

armab (commented):
I believe we'll need to include `redis` in main pip requirements https://github.com/StackStorm/st2/blob/master/requirements.txt

Besides of that, it's all good as the target is to ship the Redis as a default st2 coordination backend in v3.5.0 :+1:

punkrokk:
Eugen is correct. 

JP Bourget / @punkrokk

m4dcoder:
Replaced this PR with https://github.com/StackStorm/st2/pull/5226.

PR #5059, opened 2020-10-14
Fixed python 3 incompatibility in logging API implemenation of findCaller (v3.3 cherry pick)

### LINKED ISSUE #5057: ST2 logs writing "(unknown file)" in log prefix, instead of module name
## SUMMARY

In ST2 v3.2, the st2sensorcontainer.log, and in v3.3 all st2 logs have the `module` name replaced with `(unknown file)`

### STACKSTORM VERSION

v3.2, v3.3, EL8 and U18.04
Python 3.6

One Line and Ansible so far. 

Show how to reproduce the problem, using a minimal test-case. Make sure to include any content
(pack content - workflows, actions, etc.) which are needed to reproduce the problem.

## Expected Results

No `(unknown file)`, but the python module name. 

## Actual Results
```
2020-10-13 17:55:13,787 140460262501416 INFO (unknown file) [-] Sensor linux.FileWatchSensor started
2020-10-13 17:55:15,444 139725927337456 INFO (unknown file) [-] No config found for sensor "FileWatchSensor"
2020-10-13 17:55:15,446 139725927337456 INFO (unknown file) [-] Watcher started
2020-10-13 17:55:15,446 139725927337456 INFO (unknown file) [-] Running sensor initialization code
2020-10-13 17:55:15,454 139725927337456 INFO (unknown file) [-] Running sensor in passive mode
```

nmaludy:
Closes #5057

See that issue for description of what the problem was. The fix was simply to ensure our implementation handles both the python 2 version and returns a 3-tuple, along with the python 3 version which returns a 4-tuple and is based on the code here: https://github.com/python/cpython/blob/3.9/Lib/logging/__init__.py#L1502-L1536

I've tested this change both my a python 2.7 (centos7) and python 3.6 (centos8 and ubuntu18) box and the logging modules are returned properly.

### DIFF
--- a/st2common/st2common/log.py
+++ b/st2common/st2common/log.py
@@ -69,25 +69,62 @@ LOGGER_KEYS = [
 _srcfile = get_normalized_file_path(__file__)
 
 
-def find_caller(*args, **kwargs):
+def find_caller(stack_info=False, stacklevel=1):
     """
     Find the stack frame of the caller so that we can note the source file name, line number and
     function name.
 
     Note: This is based on logging/__init__.py:findCaller and modified so it takes into account
-    this file - https://hg.python.org/cpython/file/2.7/Lib/logging/__init__.py#l1233
+    this file:
+    https://github.com/python/cpython/blob/2.7/Lib/logging/__init__.py#L1240-L1259
+
+    The Python 3.x implementation adds in a new argument `stack_info` and `stacklevel`
+    and expects a 4-element tuple to be returned, rather than a 3-element tuple in
+    the python 2 implementation.
+    We derived our implementation from the Python 3.9 source code here:
+    https://github.com/python/cpython/blob/3.9/Lib/logging/__init__.py#L1376-L1404
+
+    We've made the appropriate changes so that we're python 2 and python 3 compatible depending
+    on what runtine we're working in.
     """
-    rv = '(unknown file)', 0, '(unknown function)'
+    if six.PY2:
+        rv = '(unknown file)', 0, '(unknown function)'
+    else: # python 3, has extra tuple element at the end for stack information
+        rv = '(unknown file)', 0, '(unknown function)', None
 
     try:
-        f = logging.currentframe().f_back
+        f = logging.currentframe()
+        #On some versions of IronPython, currentframe() returns None if
+        #IronPython isn't run with -X:Frames.
+        if f is not None:
+            f = f.f_back
+        orig_f = f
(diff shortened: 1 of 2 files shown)


m4dcoder (approved):

st2common/st2common/log.py
```diff
@@ -69,25 +70,64 @@
...
     Note: This is based on logging/__init__.py:findCaller and modified so it takes into account
-    this file - https://hg.python.org/cpython/file/2.7/Lib/logging/__init__.py#l1233
+    this file:
+    https://github.com/python/cpython/blob/2.7/Lib/logging/__init__.py#L1240-L1259
+
+    The Python 3.x implementation adds in a new argument `stack_info` and `stacklevel`
+    and expects a 4-element tuple to be returned, rather than a 3-element tuple in
+    the python 2 implementation.
+    We derived our implementation from the Python 3.9 source code here:
+    https://github.com/python/cpython/blob/3.9/Lib/logging/__init__.py#L1376-L1404
```
cognifloyd, line 86:
```suggestion
    https://github.com/python/cpython/blob/3.9/Lib/logging/__init__.py#L1502-L1536
```

amanda11:
I installed the package from this PR onto U18 and it did get rid of most of the unknown file messages in the logs. I'm still left with "unknown file" on exception in the st2sensorcontainer log:

`2020-10-15 08:22:43,653 139699041297216 ERROR (unknown file) [-]   File "/opt/stackstorm/packs/openstack/sensors/messaging_sensor.py", line 52, in poll`

nmaludy:
@amanda11 can you confirm if that is a new problem or one that already exists in 3.2?

amanda11:
> @amanda11 can you confirm if that is a new problem or one that already exists in 3.2?

I think it probably exists on 3.2 as 3.2 had lots of "unknown file" messages, but I'll confirm - I have to break a system to get the exception!

amanda11:
> @amanda11 can you confirm if that is a new problem or one that already exists in 3.2?

@nmaludy - confirmed it is python 3 only. I can raise an issue for it - it just seems to be the exceptions that are a problem now on 3.3.

amanda11 (approved):
LGTM - I willl raise a separate issue for the "unknown module" on exceptions in sensor container on py3 (if one already doesn't exist), after confirming that the problem also exists on 3.2.

cognifloyd (approved):
LGTM

PR #5062, opened 2020-10-22
Fix travis jobs after making python3 as default for development

m4dcoder:
The image that travis uses contain both python 2.7 and python 3.6. Explicitly add the PYTHON_VERSION environment variable to instruct which python executable to use. Fix python 2.7 jobs in travis that are failing.

### DIFF
--- a/test-requirements-py27.txt
+++ b/test-requirements-py27.txt
@@ -1,0 +1,35 @@
+# NOTE: codecov only supports coverage==4.5.2
+coverage==4.5.2
+pep8==1.7.1
+st2flake8==0.1.0
+astroid==1.6.5
+pylint==1.9.4
+pylint-plugin-utils>=0.4
+bandit==1.5.1
+ipython<6.0.0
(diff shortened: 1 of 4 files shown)


test-requirements-py27.txt
```diff
@@ -0,0 +1,35 @@
...
+# Required by st2client tests
+pyyaml==5.1.2
+RandomWords
+gunicorn==19.9.0
+psutil==5.6.6
+webtest==2.0.25
+rstcheck>=3.3.1,<3.4
+tox==3.14.1
+pyrabbit
+# Since StackStorm v2.8.0 we now use cryptography instead of keyczar, but we still have some tests
```
amanda11, line 30: A query - if keyczar tests only run on python2.7, then as these are split requirement files, should it get removed from the test-requirements.txt?

amanda11 (approved):
LGTM - just had a query if additional change was needed on test-requirements.txt due to comment on the py27 test requirements

PR #5064, opened 2020-10-27
Improve st2client homepage

blag:
This PR improves the PyPI description of st2client by doing the following:

* Removes the redundant `license` specifier argument to `setup()` (see the official documentation on it [here](https://packaging.python.org/specifications/core-metadata/#license), and why this is a problem [here](https://github.com/diyan/pywinrm/pull/284))
* Adds the content of `README.rst` as the long description for the st2client package on PyPI
* Adds additional URLs to the project, which I think will be rendered in the sidebar of the PyPI project page, giving the links more visibility for users looking at the st2client package on PyPI

### DIFF
--- a/st2client/setup.py
+++ b/st2client/setup.py
@@ -40,7 +40,6 @@ setup(
                  'automation platform.'),
     author='StackStorm',
     author_email='info@stackstorm.com',
-    license='Apache License (2.0)',
     url='https://stackstorm.com/',
     classifiers=[
         'Development Status :: 5 - Production/Stable',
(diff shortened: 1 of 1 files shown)


nmaludy (commented):
Should we set long_description_content_type to `text/x-rst`

https://packaging.python.org/guides/making-a-pypi-friendly-readme/

blag:
According to [the docs](https://packaging.python.org/specifications/core-metadata/#description-content-type), that's not strictly necessary:

> If a Description-Content-Type is not specified, then applications should attempt to render it as text/x-rst; charset=UTF-8 and fall back to text/plain if it is not valid rst.

nmaludy (approved):
Great work, thank you!

blag:
Thanks for the review!

st2client/setup.py
```diff
@@ -63,5 +67,16 @@
         'console_scripts': [
             'st2 = st2client.shell:main'
         ]
+    },
+    project_urls={
```
armab, line 71: That's a great list :+1:

blag: Thanks!

PR #5068, opened 2020-10-27
Better st2-self-check error messages

### LINKED ISSUE #2997: st2-self-check fails
```
¯\_(ツ)_/¯
```

blag:
Closes #2997

The `st2-self-check` script must be run as root and run with `ST2_AUTH_TOKEN` or `ST2_API_KEY` environment variables, but it does not currently check for them. This means that it will run "successfully" even if the workflows it kicks off do not complete successfully.

This PR redirects the root user error message to stderr, and adds a check and informative error message for `ST2_AUTH_TOKEN` and `ST2_API_KEY` environment variables.


armab (approved):
:+1: 

Thanks for fixing that! I remember hitting the same problem before.

PR #5086, opened 2020-11-19
Rbac integration

hnanchahal:
Integrate RBAC backend https://github.com/StackStorm/st2-rbac-backend into st2 core natively.


fixed-requirements.txt
```diff
@@ -10,6 +10,7 @@ cryptography==2.8
 eventlet==0.25.1
 flex==6.14.0
 gitpython==2.1.15
+git+https://github.com/StackStorm/st2-rbac-backend.git@master#egg=st2-rbac-backend
```
armab, line 13: I tested the dev packages from this PR and indeed rbac package wasn't picked up.

Let's try to include the rbac in `st2common/in-requirements.txt` instead of this file and generate the requirements.txt with `make requirements`.

That should help.

armab (changes requested):
Thanks for the contribution!

I left a comment below that'll help to integrate the rbac py package into st2.

fixed-requirements.txt
armab, line 13:
```suggestion
```
as the dependency is in `st2common/in-requirements.txt`

hnanchahal: Updated global requirements

armab:
As we're still waiting for TravisCI, wondering that there is also a need to test for `RBAC` presence in st2-packaging tests.
Additionally, looks like RBAC ships the `st2-apply-rbac-defintions` binary required to apply the rbac rules.

@hnanchahal Can you please open a PR against the st2-packages to include something along the lines there: 
https://github.com/StackStorm/st2-packages/commit/0bf20766f126fd234321604e723379d8b52e95df

This will ensure that st2-apply-rbac-defintions binary is shipped with the `st2` deb/rpm package and test it.

armab:
CI reports some real issues. Please check the https://travis-ci.org/github/StackStorm/st2/jobs/744939156#L1775-L1787 with the instructions. Similar to previous PR, this needs another make requirements to update all the requirements.

armab:
`st2-packages` still needs a change for enabling apply-rbac-defintions, described here in details: https://github.com/StackStorm/st2/pull/5086#issuecomment-731364977
@hnanchahal Can you please compose that PR to ensure we're shipping all the RBAC bits with the packages?

armab (changes requested):
To recap the blockers for this PR:

* st2-packages need a change for enabling apply-rbac-defintions, described here in details: https://github.com/StackStorm/st2/pull/5086#issuecomment-731364977
* Some Unit Tests related to RBAC should be fixed, according to CI failure: https://travis-ci.org/github/StackStorm/st2/jobs/745513542#L2227

fixed-requirements.txt
armab: OK this looks good now.

hnanchahal:
@armab I think the Unit test is failing because its not getting "default" from available_rbac_backends. Any pointers?

def validate_rbac_is_correctly_configured():
    """
    Function which verifies that RBAC is correctly set up and configured.
    """
    if not cfg.CONF.rbac.enable:
        return True

    from st2common.rbac.backends import get_available_backends
    available_rbac_backends = get_available_backends()

    # 1. Verify auth is enabled
    if not cfg.CONF.auth.enable:
        msg = ('Authentication is not enabled. RBAC only works when authentication is enabled. '
               'You can either enable authentication or disable RBAC.')
        raise ValueError(msg)

    # 2. Verify default backend is set
    if cfg.CONF.rbac.backend != 'default':
        msg = ('You have enabled RBAC, but RBAC backend is not set to "default". '
               'For RBAC to work, you need to install "st2-rbac-backend" package, set '
               '"rbac.backend" config option to "default" and restart st2api service.')
        raise ValueError(msg)

    # 2. Verify default RBAC backend is available
    if 'default' not in available_rbac_backends:
        msg = ('"default" RBAC backend is not available. Make sure '
               '"st2-rbac-backend" system packages are installed.')
        raise ValueError(msg)

armab:
@hnanchahal Right, the new state is that `RBAC` is now part of the st2 core and those tests need an adjustment to reflect the integration.

For instance, instructions like:
```
For RBAC to work, you need to install "st2-rbac-backend" package, set rbac.backend" config option to "default" and restart st2api service
```
recommends installing the `st2-rbac-backend` and restarting the services, while it's not needed anymore.

hnanchahal:
> @hnanchahal Right, the new state is that `RBAC` is now part of the st2 core and those tests need an adjustment to reflect the integration.
> 
> For instance, instructions like:

Understood that part. But since the rbac-backend package is being included in the requirements for st2common, why the driver_loader is failing to pick up rbac backend. 
I assume 
from st2common.rbac.backends import get_available_backends
    available_rbac_backends = get_available_backends()

is not returning the default rbac backend.

armab:
Are you referring to this https://travis-ci.org/github/StackStorm/st2/jobs/745513542#L2230-L2236 unit tests failure?
```
   AssertionError: ValueError not raised by validate_rbac_is_correctly_configured
```
The test enables rbac, auth and expects that it'll raise an error because previously RBAC wasn't available in the default installation.

https://github.com/StackStorm/st2/blob/24ad84353362e2fdee0b9965da909d2bd488c851/st2api/tests/unit/test_validation_utils.py#L54-L61

Because RBAC package is now part of the core, I think the unit test case there should not expect error message (ValueError)  anymore.

Let's start from the point you mentioned, like more explicitly check that `default` is now in `get_available_backends()` for RBAC in the Unit tests?

st2api/st2api/validation.py
```diff
@@ -38,13 +38,12 @@ def validate_rbac_is_correctly_configured():
 
     # 2. Verify default backend is set
     if cfg.CONF.rbac.backend != 'default':
-        msg = ('You have enabled RBAC, but RBAC backend is not set to "default". '
-               'For RBAC to work, you need to install "st2-rbac-backend" package, set '
-               '"rbac.backend" config option to "default" and restart st2api service.')
+        msg = ('You have enabled RBAC, but RBAC backend is not set to "default".')
```
armab, line 41:
```suggestion
        msg = ('You have enabled RBAC, but RBAC backend is not set to "default". '
                    'For RBAC to work, you need to set '
                    '"rbac.backend" config option to "default" and restart st2api service.')
```

armab (changes requested):
The Unit tests are passing now, - that's some good progress :+1: 

However linting is unhappy due to minor formatting issues. Check the instructions here: https://travis-ci.org/github/StackStorm/st2/jobs/745963664#L1642-L1646 about how to fix it.

I also provided other minor comments we'll need to fix before merging this PR.

armab:
@hnanchahal There is still a linting issue that needs a fix, see: https://travis-ci.org/github/StackStorm/st2/jobs/746190520#L1643

armab (approved):
Looks good now.

I've tested the produced packages myself and can confirm that RBAC worked.

armab:
@hnanchahal Thanks for the integration! 

We'll need st2docs update for the RBAC feature as a next step, see more details here: https://github.com/StackStorm/st2docs/issues/1038

PR #5089, opened 2020-11-23
Remove Python 2.7 from tests and builds

blag:
* [x] Remove Python 2.7 from `.travis.yml` and `Makefile`
* [x] Refactor `.travis.yml` to generate the build matrix implicitly (eg: `env.jobs` and `jobs`)
* [ ] Build packages with Python 3 instead of Python 2.7
* [ ] Change Python shebangs in scripts to use Python 3


blag:
I force-pushed this to make it even with `master` and GitHub assumed that meant I had merged the branch and closed the PR. Opened #5090 instead.

PR #5090, opened 2020-11-23
 Remove Python 2.7 from tests and builds

blag:
* [x] Remove Python 2.7 from `.travis.yml` and `Makefile`
* [x] Refactor `.travis.yml` to generate the build matrix implicitly (eg: `env.jobs` and `jobs`)
* [x] Build packages with Python 3 instead of Python 2.7
* [x] Change Python shebangs in scripts to use Python 3


nmaludy (approved):

Makefile
```diff
@@ -450,7 +441,7 @@ compilepy3:
 .st2common-circular-dependencies-check:
 	@echo "Checking st2common for circular dependencies"
 	find ${ROOT_DIR}/st2common/st2common/ -name \*.py -type f -print0 | xargs -0 cat | grep st2reactor ; test $$? -eq 1
-	find ${ROOT_DIR}/st2common/st2common/ \( -name \*.py ! -name runnersregistrar\.py -name \*.py ! -name compat\.py | -name inquiry\.py \) -type f -print0 | xargs -0 cat | grep st2actions ; test $$? -eq 1
+	find ${ROOT_DIR}/st2common/st2common/ \( -name \*.py ! -name runnersregistrar\.py -name \*.py ! -name compat\.py ! -name inquiry\.py \) -type f -print0 | xargs -0 cat | grep st2actions ; test $$? -eq 1
```
blag, line 444: @m4dcoder Is this change acceptable to you? I think this is what you were trying to do with [this commit](https://github.com/StackStorm/st2/pull/4309/commits/e88323eaf4c6bf013c3139cdc09d82b7f4706dac#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R268), but I want to double check with you.

Without this fix we get two errors: `find: missing )` and `Command not found '-name'`, which I think is caused by the `|` character terminating the `find` command prematurely.

I also noticed that `-name \*.py` is in the `find` parameters twice, but I don't think that would cause any issues.

CHANGELOG.rst
```diff
@@ -25,6 +25,10 @@ Changed
 
   Contributed by @hnanchahal
 
+* Updated Travis CI and CircleCI builds to only use Python 3
```
armab, line 28: Please don't forget to include the PR number #5090 also the type of the change.

Also is it (improvement) or (cleanup)? Maybe `Removed` section below would fit this better, considering it's breaking change nature.

Makefile
```diff
@@ -233,7 +230,7 @@ check-python-packages:
 	@echo ""
 	@echo "================== CHECK PYTHON PACKAGES ===================="
 	@echo ""
-	test -f $(VIRTUALENV_COMPONENTS_DIR)/bin/activate || virtualenv --python=$(PYTHON_VERSION) $(VIRTUALENV_COMPONENTS_DIR) --no-download --system-site-packages
+	test -f $(VIRTUALENV_COMPONENTS_DIR)/bin/activate || $(PYTHON_VERSION) -m venv $(VIRTUALENV_COMPONENTS_DIR) --system-site-packages
```
armab, line 233: I like the transition to native py3 `venv`, however we stick to `virtualenv` tool and its behavior in many other places, including packaging, as well as other CIs using this st2 Makefile. I suspect that this might provide problems and looks like there were already issues related to compatibility with the previous tool.

Can we keep the old `virtualenv` approach for this PR and do `venv` migration in another isolated PR?
So this PR would do 1 thing, and do it well: removing python2 only (as advertized).

armab (changes requested):

.circleci/config.yml
```diff
@@ -79,7 +79,7 @@ jobs:
   # Run st2 Lint Checks
   lint:
     docker:
-      - image: circleci/python:2.7
+      - image: circleci/python:3.6
```
armab, line 82: FYI CircleCI is only for deb/rpm packaging and Unit/Integration test instructions in CircleCI are not used.
We'll need to remove them to avoid confusion, but that's another story.

CHANGELOG.rst
```diff
@@ -37,6 +37,9 @@ Removed
 ~~~~~~~~
 * Removed check-licence script (cleanup) #5092
   Contributed by @kroustou
+* Updated Travis CI and CircleCI builds to only use Python 3 (cleanup) #5090
```
armab, line 40: I think the other significant change to mention is that py2 was used in the Makefile. Because other scripts and systems may rely on it it worth highlighting.

Maybe something along the lines?

```suggestion
* Updated Makefile and build system to only use Python 3 (cleanup) #5090
```

Makefile
blag: Done. :+1:

armab (approved):
LGTM :+1:

PR #5098, opened 2020-12-01
Convert CI to GitHub Actions

nmaludy:
## TODO List
- [x] TASK=ci-checks
- [x] TASK=compilepy3
- [x] TASK=ci-packs-tests
- [x] TASK=ci-unit
- [ ] TASK=ci-integration
- [x] caching
- [x] apt sources
- [x] install script
- [x] before script
- [x] script
- [x] notifications

## Status
Everything has been converted over except for Integration test and Slack notifications.

### Integration Tests
- The RabbitMQ "service" container that we use currently can not start up with our custom config and SSL certificates required in integration tests: st2common/tests/integration/test_rabbitmq_ssl_listener.py
  - To "fix" this we need to install/config/run RabbitMQ ourselves, but getting this working in GHA has proven to be difficult (unable to install rabbitqmq-server apt package successfully)
- Ran into issues with the python_action_process_wrapper tests (contrib/runners/python_runner/tests/integration/test_python_action_process_wrapper.py).
  - The specific test that is failing is: (test_stdin_params_timeout_no_stdin_data_provided)[https://github.com/StackStorm/st2/blob/master/contrib/runners/python_runner/tests/integration/test_python_action_process_wrapper.py#L127-L139]
  - It is failing because inside of python_action_wrapper.py it is always receiving an empty string on `stdin` rather than blocking on input and timing out here: https://github.com/StackStorm/st2/blob/master/contrib/runners/python_runner/python_runner/python_action_wrapper.py#L305-L315


tools/st2-setup-examples
```diff
@@ -23,7 +22,7 @@ echo -e "Bootstrapping virtualenv for ${PACK} pack..."
 if [ -f ${TESTSPATH}/${PACK}/requirements.txt ]; then
     echo "Creating virtualenv for ${PACK}..."
     mkdir -p /opt/stackstorm/virtualenvs
-    virtualenv -p ${PY27} --system-site-packages /opt/stackstorm/virtualenvs/${PACK}
+    virtualenv /opt/stackstorm/virtualenvs/${PACK}
```
amanda11, line 25: Looks great, had one query...

On this line - which version of python will this use? Should we have a -p python3, so that on distributions where python2.7 is still the default, that python3 is used? (Or point to /opt/stackstorm/st2/bin/python - which would be python3 on those distributions...)

Similar comment on the st2-setup-tests...

armab:
Great stuff! :tada: 

For cosmetic and readiness reasons, I'd suggest to play with the CI naming a bit:
WIP _Convert_CI_to_GitHub_Actions_by_nmaludy_·_Pull_Request_5098_·_StackStorm_st2_-_2020-12-03_14 38 55

How about:
- `CI / lint`
- `CI / unit`
- `CI / packs-tests`
- `CI / integration`

to make it easier, faster to read and navigate

armab:
Python2 CI/Makefiles removal by @blag was merged https://github.com/StackStorm/st2/pull/5090 with some additional changes.

@nmaludy looks like there are some diff conflicts to resolve. Can you help with that?
We want to merge this PR next as it'll help us a lot.

Makefile
```diff
@@ -523,6 +521,13 @@ requirements: virtualenv .requirements .sdist-requirements install-runners insta
 	@echo
 	@echo "==================== requirements ===================="
 	@echo
+	# Show pip installed packages before we start
+	$(VIRTUALENV_DIR)/bin/pip list
+
+	# Note: Use the verison of virtualenv pinned in fixed-requirements.txt so we
+	#       only have to update it one place when we change the version
+	$(VIRTUALENV_DIR)/bin/pip install --upgrade $(shell grep "^virtualenv" fixed-requirements.txt)
```
armab, line 529: nice stuff :+1:

st2common/tests/integration/test_rabbitmq_ssl_listener.py
```diff
@@ -32,15 +32,16 @@
...
 
 
-# NOTE: We only run those tests on Travis because at the moment, local vagrant dev VM doesn't
-# expose RabbitMQ SSL listener by default
-@unittest2.skipIf(not ON_TRAVIS, 'Skipping tests because not running on Travis')
+# NOTE: We only run those tests on the CI provider because at the moment, local
+#       vagrant dev VM doesn't expose RabbitMQ SSL listener by default
+@unittest2.skipIf(not ST2_CI,
+                  'Skipping tests because ST2_CI environment variable is not set to "true"')
 class RabbitMQTLSListenerTestCase(unittest2.TestCase):
```
armab, line 45: This part is not skipped in TravisCI for some reason, according to
https://travis-ci.org/github/StackStorm/st2/builds/749891682#L9616-L9839
and so Travis Integration tests are failing now.

armab (changes requested):
Looks like the only blocker here is that previously :heavy_check_mark: Integration Tests in Travis are failing now due to some ENV changes.

@nmaludy can you please fix that small issue and we're good to go with moving everything except of Integration tests to Github Actions :+1:  That'll already help us with the test environment a lot.

armab (approved):
:+1: Considering tests are green

Kami:
Nice work :+1: 

In the future we should see if we can try speeding up unit tests a bit since they take almost 12 minutes. I think the best option would be to try to split them into chunks and run them as part of different build jobs.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,188 @@
...
+          sudo .circle/add-itest-user.sh
+      - name: Permissions Workaround
+        if: "${{ env.TASK == 'ci-packs-tests' || env.TASK == 'ci-integration' }}"
+        run: |
+          echo "$ST2_CI_REPO_PATH"
+          sudo ST2_CI_REPO_PATH="${ST2_CI_REPO_PATH}" scripts/travis/permissions-workaround.sh
+      - name: Setup RabbitMQ (NOT WORKING YET)
+        if: "${{ env.TASK == 'ci-integration' }}"
+        run: |
+          # Use custom RabbitMQ config which enables SSL / TLS listener on port 5671 with test certs
```
Kami, line 142: I think eventually it would be nice to move this rabbitmq stuff inside a service Docker container which runs in the background for the duration of the build.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,188 @@
...
+    branches: ['*']
+    tags:
+      - v*
+  pull_request:
+    type: [opened, reopened, edited]
+  schedule:
+    # run every night at midnight
+    - cron:  '0 0 * * *'
+
+jobs:
```
Kami, line 14: I think it would also be good to utilize this action to skip redundant builds - https://github.com/fkirc/skip-duplicate-actions.

I've been using it quite a lot recently with various GHA workflows. In a sense, it's similar to Circle CI auto cancel old workflows, but in addition to that, it also allows you to skip a build if a build for a specific SHA and set of files already exist.

Having said that, we should just apply that check to non master and version branches. Something along the lines of https://github.com/apache/libcloud/blob/trunk/.github/workflows/main.yml#L31.

.github/workflows/ci.yaml
Kami: EDIT: Never mind, I see the comment here now - https://github.com/StackStorm/st2/pull/5098/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR42.

Kami:
EDIT: Now that I think of it, splitting unit tests into multiple jobs would require more work to get coverage to work.

Having said that, current coverage set up is already far from ideal and too complicated so changing it a bit would not be a bad thing.

One option would be to generate uniquely named coverage file as part of each test job, at the end, preserve it as  a build artifact and then have another job which runs at the end which combines the coverage data and submits it to codecov (well, codecov.io already supports partial coverage uploads and combining on the server side so that may not be strictly needed for codecov.io itself).

blag:
The `Makefile` is already partially orchestrated to run and even combine coverage results. For anybody looking into this, check out the [.coverage-combine target](https://github.com/StackStorm/st2/blob/7ad17fd4acf810080b1c158ede5403d8a5e6e21b/Makefile#L818) to get started.

Kami:
Yeah, that code is pretty complicated though.

If we only ever plan to view and submit coverage to codecov.io,  we could simplify the Makefile by removing all the coverage combining related stuff and simply rely on server side merging on codecov.io :)

blag:
@winem Can you push this PR to completion, or at least partial completion? If we only offload four of the five `TASK`s to GitHub Actions, that's will still be very useful for us. Travis CI latencies are terrible, so we still gain something even if we don't completely move off of their platform entirely.

winem:
@blag yes, I'll tackle the rabbitmq issue.

armab:
Thanks @winem.

Per discussion, the transient agreement was to move the Unit tests to Github Actions,
while keeping the Integration Tests in TravisCI. 

That would already help a lot.

armab:
To merge this PR we need to make sure:
* Unit Tests + Checks are running in GH Actions (done)
* Unit Tests in GH Actions are green: 90% done
* Only Integration Tests are running in TravisCI (done)
* Integration Tests in TravisCI are green (as before)

winem:
@armab I don't see any failing unit tests when I take a look at https://github.com/StackStorm/st2/runs/1704636828?check_suite_focus=true for example. All the exceptions that are shown are expected and the tests as such pass. 

Please enlighten me. Looks like I miss something here. I'm happy to tackle remaining tests over the weekend.

st2common/tests/integration/test_rabbitmq_ssl_listener.py
blag: I think these tests are intended to run on Travis CI. I think the Travis tests are failing because the RabbitMQ config file was changed.

blag:
🎉

armab:
That was a great team effort @nmaludy @blag @winem! :+1: 
Really happy to see this merged and part of our tests running in GH Actions.

PR #5099, opened 2020-12-03
Improve python2 pack warning when running on Python 3 environment

amanda11:
Update the python2 pack warning when on a python3 environment, to indicate python2 pack is now removed.

Part of https://github.com/StackStorm/discussions/issues/54


amanda11:
I've kept in the old warning for when running on a py2 environment, in case the U16 changes don't get in.

st2common/st2common/util/pack.py
```diff
@@ -43,9 +43,15 @@
...
 
 # Common format for python 2.7 warning
-PACK_PYTHON2_WARNING = "DEPRECATION WARNING: Pack %s only supports Python 2.x. " \
+if six.PY2:
+    PACK_PYTHON2_WARNING = "DEPRECATION WARNING: Pack %s only supports Python 2.x. " \
                        "Python 2 support will be dropped in future releases. " \
                        "Please consider updating your packs to work with Python 3.x"
+else:
+    PACK_PYTHON2_WARNING = "DEPRECATION WARNING: Pack %s only supports Python 2.x. " \
+                       "Python 2 support has been removed. " \
```
armab, line 52: Do you think it worth mentioning when the python 2 removal happened in the platform? Might give some users hints if they want to downgrade the st2 version.

```suggestion
                       "Python 2 support has been removed since st2 v3.4.0. " \
```

armab (approved):
:+1: 

Left one minor comment to consider. We'll also need a Changelog for this PR, as usual.

PR #5100, opened 2020-12-03
Remove --python3 option

amanda11:
Remove the --python3 option and associated python3_... config items.


amanda11:
Waiting for U16 to move to python3, and is also dependant on fixes in progress for cryptography and using python3 on ST tests.

st2client/st2client/commands/pack.py
```diff
@@ -189,10 +189,6 @@ def __init__(self, resource, *args, **kwargs):
                                  metavar='pack',
                                  help='Name of the %s in Exchange, or a git repo URL.' %
                                  resource.get_plural_display_name().lower())
-        self.parser.add_argument('--python3',
```
blag, line 192: Keep this for backwards-compatibility please.

st2client/st2client/commands/pack.py
```diff
@@ -210,7 +206,7 @@ def run(self, args, **kwargs):
         if not is_structured_output:
             self._get_content_counts_for_pack(args, **kwargs)
 
-        return self.manager.install(args.packs, python3=args.python3, force=args.force,
```
blag, line 213: Pass this through.

st2client/st2client/models/core.py
```diff
@@ -507,12 +507,11 @@ class AsyncRequest(Resource):
 
 class PackResourceManager(ResourceManager):
     @add_auth_token_to_kwargs_from_env
-    def install(self, packs, force=False, python3=False, skip_dependencies=False, **kwargs):
+    def install(self, packs, force=False, skip_dependencies=False, **kwargs):
```
blag, line 510: Keep the `python3` argument here, but emit a warning that it's now ignored, and then don't pass it in the `payload` (like you already do). This will keep backwards compatibility for the CLI.

blag (changes requested):
Keep backwards compatibility for the CLI, since we haven't been emitting deprecation warnings for that. We can emit a warning when it is passed in, but we don't need to pass it to the ST2 API.

armab:
@blag We announced that we're removing python2 completely and users had that notice and time to migrate. This means we're removing all the python2-related functionality.

I don't think keeping that old flag would help anymore, considering this behavior doesn't work and would just provide erroneous results. I think that'll bring confusion and false sense of a working feature.

We had enough deprecation Warnings about full Python2 removal via:
1) st2 startup logs
2) via `st2ctl cmd`
2) on pack install

This includes entire platform python2 feature set.

amanda11:
The cli did have a warning. St2 pack install had a warning in 3.3 if you installed a python 2 pack.

blag:
If my understanding is correct, then there's two warnings that should have been generated:

* [x] Warning about installing Python 2-only pack
* [ ] Warning about `--python3` flag being deprecated and removed, since ST2 is going Python 3 only (unless you use the `--force` flag)

We definitely warned about Python 2 packs, but did we also warn about the removal of the `--python3` flag? I don't see it, but if I missed it, please let me know.

Consider what people would have done in the past to ensure that their packs were always Python 3:

1. Their packs were Python 2 --> not good.
2. Update their packs to support Python 2 and Python 3 --> better.
3. Install their packs with `--python3` in their deployment scripts or whatever --> awesome.

So now their pack still has to be installed with the `--python3` flag, otherwise it gets installed with Python 2.

Since we defaulted to installing packs as Python 2 for so long, people _who are trying to do things right_ have been forced into using the `--python3` flag to ensure that their packs use Python 3.

If we up and remove this flag, the CLI will no longer recognize the flag and throw an error. That will cause those people's deployment scripts or whatever to immediately break. That's pretty rude of us - we are effectively penalizing people who have been doing things "right" and trying to switch to Python 3 for awhile.

Instead, we should at least allow a grace period where we still accept that (now superfluous) `--python3` flag, but we emit a warning about it. That will give our users, who have been doing things the Right Way (tm) so far, time to remove the use of that flag from their deployment scripts.

If I have been operating under a misconception please correct me, but as I understand it, we should not break things for our best users.

amanda11:
I think your understanding is correct, and I get your point as to why we should allow them to still use --python3 option with a warning.
I am happy to rework as per your suggestion, if @armab also happy.

armab:
@amanda11 Per https://stackstorm-community.slack.com/archives/CSBELJ78A/p1608157605269500 discussion, I'm fine with @blag to keep it for one more release if there are concerns that python 2 removal/deprecation communication wasn't enough for this change.

amanda11:
@blag @armab I've kept in the --python3 pack, but I only kept it in the pack/install.py. I didn't see any other print etc being done in the manager files, so I wasn't sure why we wanted to pass the argument through to the manager and do the prints in there, as all the other printing seems to be done in the commands/*.py files. 
If it's needed I will move it to the manager file.

Example output:
```
$ st2 pack install aws_boto3 --python3

For the "aws_boto3" pack, the following content will be registered:

actions   |  6
rules     |  0
sensors   |  0
aliases   |  0
triggers  |  0

Installation may take a while for packs with many items.

DEPRECATION WARNING: --python3 flag will be ignored, as ST2 now runs with python3 as the default on all OS


	[ succeeded ] init_task
	[ succeeded ] download_pack
	[ succeeded ] make_a_prerun
	[ succeeded ] get_pack_dependencies
	[ succeeded ] check_dependency_and_conflict_list
	[ succeeded ] install_pack_requirements
	[ succeeded ] get_pack_warnings
	[ succeeded ] register_pack

+-------------+-------------------------+
| Property    | Value                   |
+-------------+-------------------------+
| ref         | aws_boto3               |
| name        | aws_boto3               |
| description | AWS actions using boto3 |
| version     | 0.8.0                   |
| author      | StackStorm, Inc.        |
+-------------+-------------------------+
```

st2client/st2client/commands/pack.py
```diff
@@ -210,7 +210,10 @@ def run(self, args, **kwargs):
...
+        if args.python3:
+            print('\nDEPRECATION WARNING: --python3 flag will be ignored, as ST2 now runs with python3 as the default on all OS\n')
```
armab, line 214: Maybe something along the lines would be more precise?

>DEPRECATION WARNING: --python3 flag is ignored and will be removed in v3.5.0 as StackStorm now runs with python3 only

st2client/st2client/commands/pack.py
```diff
@@ -210,7 +210,10 @@ def run(self, args, **kwargs):
         if not is_structured_output:
             self._get_content_counts_for_pack(args, **kwargs)
 
-        return self.manager.install(args.packs, python3=args.python3, force=args.force,
+        if args.python3:
+            print('\nDEPRECATION WARNING: --python3 flag is ignored and will be removed in v3.5.0 as StackStorm now runs with python3 only\n')
```
blag, line 214: Any reason to not use [`warnings.warn`](https://docs.python.org/3/library/warnings.html#warnings.warn) here?

blag (changes requested):

blag:
Rerunning end-to-end tests with this to ensure that StackStorm/st2tests#196 didn't break anything.

amanda11:
@blag I fixed the lint problem as it also needed an import of warnings, but I think we need to do more as haven't registered what warnings should do. Output using warnings is as follows:
```
IInstallation may take a while for packs with many items.
/home/centos/st2/st2client/st2client/commands/pack.py:215: UserWarning: 
DEPRECATION WARNING: --python3 flag is ignored and will be removed in v3.5.0 as StackStorm now runs with python3 only

  warnings.warn('\nDEPRECATION WARNING: --python3 flag is ignored and will be removed '

	[ succeeded ] init_task
```

blag (approved):

Issue #4648, opened 2019-04-25
Feature Request: Trigger Instance for Policies


sf720d:
##### SUMMARY

When a policy cancels an action or a workflow, there's no way one can know the action/workflow was canceled _by the policy_. From the conversation we had in Slack, we'd like to have a trigger instance to be able to write a rule and do whatever is needed. 
In addition to the trigger instance, it's a good idea to have a message being published when a workflow/action is canceled by a policy so StackStorm users can have an idea of what caused the cancellation. 

##### ISSUE TYPE
 - Feature Idea

##### STACKSTORM VERSION
st2 2.10.3, on Python 2.7.5

LindsayHill:
Seems reasonable. I don't think anyone else is working on this, or planning to do so in the near future. If you need it sooner, you'll need to submit a PR.

> it's a good idea to have a message being published when a workflow/action is canceled by a policy

Where are you thinking about publishing that message to?

sf720d:
I was thinking it can be published on the same workflow that was cancelled by the policy. 
Something as simple as "Cancelled by policy" would be sufficient.

LindsayHill:
As an additional result field? Or change the current "canceled" status to something else?

sf720d:
As an additional result field.
Although changing the "canceled" status would work too and might be more efficient!?!

Issue #4655, opened 2019-05-01
Scheduler bounces LiveActions between delayed and policy-delayed


nmaludy:
##### SUMMARY

I noticed this when debugging another problem and wanted to say something in case this behavior wasn't intended.

When there are executions that are `delayed` by a policy, the `st2scheduler` is constantly flipping them between `policy-delayed` and `delayed` state. Each state change is accumulated in the `LiveAction` instance and causes fairly large objects, especially if executions are delayed for a long period of time (large backlog).

##### ISSUE TYPE
 - Bug Report
 - Feature Idea

##### STACKSTORM VERSION
```
$ st2 --version
st2 3.0, on Python 2.7.5
```

##### OS / ENVIRONMENT / INSTALL METHOD
```
CentOS 7.6
Puppet-st2
```

##### STEPS TO REPRODUCE
Create the following policy `/opt/stackstorm/packs/default/policies/core_local_concurrency.yaml
```yaml
pack: default
name: core.local.concurrency
description: Limits the concurrent executions for core.local
enabled: true
resource_ref: core.local
policy_type: action.concurrency
parameters:
  action: delay
  threshold: 10
```

Run the following script to kick off a bunch of `core.local` excutions

```
#!/bin/bash
for NUM in `seq -w 0 250`; do
  st2 action execute core.local cmd='sleep 120' timeout=180
done
```

Tail the `st2scheduler` logs
```
tail -f /var/log/st2/st2scheduler.log
```

##### ACTUAL RESULTS
```
2019-05-01 15:02:59,673 140648388594096 INFO handler [-] Scheduling liveaction: 5cc9ec569387ef586c0c82c3 (queue_item_id=5cc9ec569387ef58ba380e73) (queue_item_id='5cc9ec569387ef58ba380e73',liveaction_id='5cc9ec569387ef586c0c82c3')
2019-05-01 15:02:59,711 140648388594096 INFO policies [-] Applying policy "default.core.local.concurrency" (action.concurrency) for liveaction "5cc9ec569387ef586c0c82c3".
[...]

PR #5102, opened 2020-12-06
add policy related statuses

marceloapps:
This pull request is related to the issue https://github.com/StackStorm/st2/issues/4648.
The idea is to add cancel/delay statuses related to when policies afect an action execution.

Changes were made on the constants list, concurrency and concurrency_by_attrib, scheduler handler and tests.


marceloapps:
Hi guys, do I have to change something here? Didn't quite catch the problem with CI

armab:
Just updated your branch with the latest `master`. Let's see if that fixes CI issues.

For the review, I'd expect @m4dcoder to take a look at it, keeping in mind it's more of his area.

m4dcoder:
The status changes here is going to break a lot of places in st2. The functioning of scheduler and workflow engine are coupled to statuses. I can't tell you specifically where to fix because there are a lot of places where this value change is going to impact.  You'll have to keep iterating until the unit and integrate tests go green.  The other option is to search for all the occurrences of DELAYED and CANCELED status in the code and figure out whether this change impacts it or not and make appropriate changes.  Good luck.

m4dcoder (changes requested):
As commented above, this change is going to cascade and impact a lot of different places in the scheduler, action runner, and workflow engine. This PR is not close to ready.

marceloapps:
Ok, thanks for the review! I'll get to work then, and get back to you guys.

Kami:
I would have swear we had ``policy-delayed`` status in the past which got removed at some point due to some bug which I can't recall anymore.

EDIT: From changelog:

> * Remove policy-delayed status to avoid bouncing between delayed statuses. (bug fix) #4655

Some more context for removal is in that PR. If we want to add it back, we need to ensure that issue is fixed (aka no flipping between states - will likely require changes to the scheduler).

/cc @m4dcoder

cognifloyd:
So, you'll want to merge master into your branch, resolve the conflicts (sorry!), and reformat your code with black (I recommend using `pre-commit install` so it just happens whenever you commit).

cognifloyd:
What if we added a `reason` field (or something like that) so that whatever cancels or delays an action can document why. That way, you don't have to edit any of the status-based state machines in workflow engine and other places.

cognifloyd:
@marceloapps do you want to pursue this any further? Is it helpful to leave this open?

marceloapps:
We may close this issue for now

PR #5103, opened 2020-12-08
Remove st2-debug-info

amanda11:
Remove st2debug component, as it was only used for submit-debug-info tool


amanda11:
Associated PRs would need merging first before checks fail. But should probably review this PR first to ensure agree to remove the st2debug component, and then if agreed to then review the linked PRs referenced above.

armab:
Can you please resolve the git conflicts for this PR?
That'll also restart the CI which hopefully go green after https://github.com/StackStorm/st2-packages/pull/680.

armab:
Following the TravisCI https://travis-ci.org/github/StackStorm/st2/jobs/750849455#L1775 error, this change might need a `make .requirements` update.

armab (approved):
Looks good, thanks for removing that legacy! :+1:

PR #5104, opened 2020-12-09
Update __init__.py

anirudhbagri:
Add version number to st2tests package so that it can be downloaded independently. 

Closes #4982


armab:
Thanks for the PR.
Can you provide a description for initial PR message, what you're trying to achieve here?

anirudhbagri:
Hi @armab,

My bad, I missed the PR description.
I was working on creating a custom polling sensor and while writing tests for it, I thought of using the **st2tests** package. 

However, I was not able to install the package using the `pip` command. After debugging a bit, I can to know that it was because no version number was added to the st2tests package.

armab (changes requested):
Your change makes sense! Thanks for that :+1: 


Before we can merge it, we'll need the CD workflows updated to change this version on every release. Can you provide a PR to add `st2tests` here as well https://github.com/StackStorm/st2cd/blob/master/actions/st2_prep_release_for_st2.sh#L77-L85 ? Should be very easy to do.

Besides of that please add a `Changelog.rst` record for this `st2` PR.

anirudhbagri:
Added a [PR for adding st2tests](https://github.com/StackStorm/st2cd/pull/449) to `st2_prep_release`.

anirudhbagri:
@armab, I have updated this PR and added a new PR as you suggested.

armab (approved):
Looks good, thanks!

armab:
We have some issues with our e2e CI, but I'll merge this PR now as it shouldn't affect any app behavior.

PR #5105, opened 2020-12-14
Fix chardet as fix requests

amanda11:
We pin requests[security] but that has a check when running that chardet is < 3.1.0, but chardet is not listed as requirement with a pinned requirement.
Also fix typo in in-requirements.txt comments

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -17,6 +17,9 @@ Added
 
 Changed
 ~~~~~~~~~
+* Pin chardet requirement #5101
+  Contributed by @amanda11
+
 * Improve the st2-self-check script to echo to stderr and exit if it isn't run with a
   ST2_AUTH_TOKEN or ST2_API_KEY environment variable. (improvement) #5068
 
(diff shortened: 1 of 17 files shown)


CHANGELOG.rst
```diff
@@ -17,6 +17,9 @@ Added
 
 Changed
 ~~~~~~~~~
+* Pin chardet requirement #5101
```
armab, line 20: Can you please also include in the Changelog why this change was needed.
Following that, it might be a better fit for a Fixed section, rather then Changed.

amanda11: Thanks for feedback - fixed

armab (approved):
:+1:

blag (approved):

blag:
Should be fixed once I tweak StackStorm/st2ci#194, merge it in, and deploy it.

armab:
It makes sense to merge it as is, because it'll fix the other 3 OS builds in this round.
Let's just wait for TravisCI to :heavy_check_mark:

blag:
Travis tests passed. I'm going to merge this in even with the U16 end-to-end tests failing since we need to get our CI passing.

PR #5108, opened 2020-12-18
Remove st2resultstracker from st2ctl, the development environment and the st2actions setup

### LINKED ISSUE #5070: Remove the st2resulstracker service from st2ctl and development env
Mistral was deprecated and removed from the st2 in `v3.3.0` release. However there is a `st2resultstracker` service which is used for polling result from async runner. Mistral was the only async runner which means resultsracker is obsolete and could be disabled.

Team agreed that we should keep the service implementation in st2 codebase if we'll need it in future, but remove service from the associated packaging/startup logic, st2ctl , development env.

This should be easy & fun task, if anyone from @StackStorm/contributors wants to help removing that.

winem:
This removes the references to the st2resultstracker service while keeping the code base of the resultstracker. 

close #5070


winem:
It looks like we have a circular dependency between this PR and https://github.com/StackStorm/st2-packages/pull/683 now. 

The st2resultstracker is a service managed via the debian/rules file. So the post_install job fails. 

In my opinion, it would pass once https://github.com/StackStorm/st2-packages/pull/683 is merged. 

I could also try to revert some of the deletions I made as part of this PR and see what it takes to pass the tests but this would lead to an intransparent state of the resultstracker configurations and service definitions in the codebase. So, for now, I'd like to avoid it. 

Please let me know what you prefer or if you have another idea.

winem:
Agreed with @blag to keep this PR as it is and merge both PRs on monday. The Exchange pack CI builds do have a higher priority and we want them to succeed over the weekend. 

See https://stackstorm-community.slack.com/archives/CSBELJ78A/p1608330230290400

armab (commented):
Thanks for the PR! 

I left a comment in the `st2-packages` PR before we can merge it.
Besides of that, we'll need st2docs update https://github.com/StackStorm/st2docs/search?q=st2resultstracker to ensure the full service removal.

winem:
Hi, I addressed the comment in the other PR and updated both branches so that they're up-to-date with the corresponding master. I'll take care of st2docs as well.

armab:
Looks like https://travis-ci.org/github/StackStorm/st2/jobs/751261368#L1513-L1537 CI is not happy about the `st2.conf.sample` changes.

The PR needs a bit more work to make sure engine doesn't pick up `resultstracker` module automatically for st2.conf generation.

winem:
Hi @armab, now the unit and integration tests are failing. This feels like a rabbit hole and I'm no longer sure if I'm on the right path. Should I proceed by removing / fixing failing tests or do we want to keep the st2resultstracker configuration for the tests only? 
To be honest, I'm not sure if the latter option would actually work.

blag:
Disclaimer: I haven't looked into this in-depth yet.

I don't think we will be able to get this to pass tests unless we remove the code for resultstracker from st2.

How do we get tests to pass unless we remove [this line](https://github.com/StackStorm/st2/blob/master/st2common/st2common/query/base.py#L54)?

```python
        if not config_value and config_option in cfg.CONF.resultstracker:
            config_value = getattr(cfg.CONF.resultstracker, config_option)
```

If we're going to remove resultstracker, we should rip it out of the codebase. That might prove to be a rabbit hole, but I don't see another way.

winem:
I had some time to think about it now and I agree. The initial agreement (see https://github.com/StackStorm/st2/issues/5070) was to keep the code and drop the configs, setups, etc. but the experience I made with this PR so far let me doubt that it's worth it. It might even require a few "hacks". 
The idea is to have the code at hand if we want to use it for a similar implementation in the future. I'm wondering if a reference to this PR wouldn't work as well to have a source of the dropped code.

blag:
I don't want to keep dead code in the codebase. That's more code to read, more code to track, more code to update, more code that can fail tests (as we see here), and more code to make our coverage statistics look worse than they really are. I say we adopt the YAGNI (you ain't gonna need it) principle and just up and remove it here. If - and it's a big if - we need the code in the future, we can pull it back out of the diff for this PR.

@armab What do you think?

armab:
What the implementation team say here as a burden of maintaining the unit/integration tests, coverage and the dead codebase confusion makes sense to me.

As I recall Winson insisted to keep the st2resultstracker service code in-place.
@m4dcoder Can you bring more context if we still need that code here?

armab:
@winem Yes, please remove the unused code as it'll maintenance burden. We can look at the git history if we'll ever need it later.

winem:
Thanks for the feedback. I removed the `under discussion` label and will proceed on the actual task now.

winem:
Is the querier at all still relevant? To me, it looks like it's only used to check the state asynchronously. 

So I would also remove the whole st2common.query module: https://github.com/StackStorm/st2/tree/master/st2common/st2common/query 

Please let me know if I miss anything and it's also needed for other cases as well. I couldn't find any use of st2common.query.base that still seems to be relevant.

winem:
Ok, removing the whole query module causes many test failures to fail. See: https://travis-ci.org/github/StackStorm/st2/jobs/754918090#L1954 

So I'm sure I missed something and will check the st2 code again to get an idea how to proceed here. I'm happy about any guidance, sharing the task or contributions.

st2common/st2common/runners/base.py
```diff
@@ -227,12 +188,6 @@ def cancel(self):
             self.liveaction.context
         )
 
-    def post_run(self, status, result):
```
blag, line 230: Need to keep this function:
https://travis-ci.org/github/StackStorm/st2/jobs/755265427#L1555
https://app.circleci.com/pipelines/github/StackStorm/st2/872/workflows/c7ffafaa-d8b7-40b0-825e-61c36b404168/jobs/12202/parallel-runs/0/steps/0-110

blag (changes requested):

st2common/st2common/runners/base.py
winem: Thanks, will do. Will proceed on this PR on ~40 mins.

winem: Done.

armab:
Oh, nice! Travis unit/integration tests are now :heavy_check_mark: 

Looks like we're good here after fixing the `e2e` test global issues (irrelevant to this PR).

blag:
Working on fixing the end-to-end tests today.

winem:
I did a last review today and would keep it as it is - given that the E2E-tests pass as well. 

As a follow up to this topic we might discuss if the code for asynchronous runners (everything around the ActionExecutionState incl the db model, API, etc.) is still needed. But that's out of scope for this one. 

And tbh I'm super happy that I got this done. :)

st2actions/tests/unit/test_runner_container.py
```diff
@@ -330,87 +327,6 @@ def test_dispatch_override_default_action_params(self):
         self.assertTrue(result.get('action_params').get('actionint') == 20)
         self.assertTrue(result.get('action_params').get('actionstr') == 'foo')
 
-    def test_state_db_created_for_polling_async_actions(self):
```
amanda11, line 333: I'm not sure why (but there may be a good reason) we've removed the test on the Polling AsyncActionRunner. That class still exists in the st2common/runners/base.py - so if the code is kept, shouldn't we keep the test that checks the logic?

amanda11 (commented):
Just a query to confirm why a particular test was removed - see comments.

st2actions/tests/unit/test_runner_container.py
winem: Oh sorry, that was a mistake when I started to think whether we want to drop the whole async runner. I added the tests again and all tests passed the CI.

amanda11:
LGTM - but I'd appreciate another @blag / @armab sign off. I presume the agreement was to get rid of the resultstracker code plus the query/callback feature that was previously only used by mistral.

st2actions/tests/unit/test_runner_container.py
blag: Resolved with 34ffaf4.

blag (approved):

PR #5117, opened 2021-01-19
Update orquesta in requirements to v1.3.0

m4dcoder:
Update orquesta in requirements to v1.3.0 in preparation for StackStorm v3.4 release.


contrib/runners/orquesta_runner/in-requirements.txt
```diff
@@ -1 +1,2 @@
-git+https://github.com/StackStorm/orquesta.git@v1.2.0#egg=orquesta
+git+https://github.com/StackStorm/orquesta.git@update-v130#egg=orquesta
```
blag, line 1: I take this to mean that you are creating the `update-130` branch on Orquesta so that we can make any necessary changes to Orquesta for ST2 v3.4 on that branch instead of the `v1.2.0` branch. And the branch will change to `v1.3.0` by ST2's release automation during the release process.

Is that correct?

m4dcoder: Nope. I have this other PR in the Orquesta repo that's getting release for v1.3.0 release.  I want to get this PR going as well so use update-v130 branch name as placeholder.  The other PR is merged now and I tagged the v1.3.0 release so I updated the requirements here now.

blag (approved):
LGTM!

Issue #5060, opened 2020-10-15
Requirements issue cannot install package with aio-pika in requirements


dor-abu:
## SUMMARY

A few days ago packs that had [aio-pika](https://github.com/mosquito/aio-pika) in requirements failed in "install pack dependencies" phase.

We found out that [aio-pika](https://github.com/mosquito/aio-pika/blob/master/setup.py) uses [yarl ](https://github.com/aio-libs/yarl/) as a requirement without a fixed version.
[yarl ](https://github.com/aio-libs/yarl/blob/master/setup.py) pushed changes to master in recent days, and also uses multidict who also created changes in recent days.

We resolved the issue when we manually updated the version of pip to latest - 20.2.3.

### STACKSTORM VERSION

st2 3.1.0, on Python 3.6.9

##### OS, environment, install method

We uses Docker and custom docker-compose.

## Steps to reproduce the problem

Insert yarl, aio-pika, or multidict to requirements with pip version 19.1.1 in the virtual environment of the packages 

## Expected Results

We expect to pass the phase of "install pack dependencies".

## Actual Results

This error will be raised:

`Collecting yarl
  Using cached https://files.pythonhosted.org/packages/6d/0b/44006eb5b3b7898e54df4b7481839d22d15cf3fc4f0420394ad9e3bb214e/yarl-1.6.2.tar.gz
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Requirement already satisfied: idna>=2.0 in ./lib/python3.6/site-packages (from yarl) (2.10)
Collecting typing-extensions>=3.7.4; python_version < "3.8" (from yarl)
  Using cached https://files.pythonhosted.org/packages/60/7a/e881b5abb54db0e6e671ab088d079c57ce54e8a01a3ca443f561ccadb37e/typing_extensions-3.7.4.3-py3-none-any.whl
Collecting multidict>=4.0 (from yarl)
  Using cached https://files.pythonhosted.org/packages/d2/5a/e95b0f9ebacd42e094e229a9a0a9e44d02876abf64969d0cb07dadcf3c4a/multidict-5.0.0.tar.gz
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Building wheels for collected packages: yarl, multidict
  Building wheel for yarl (PEP 517) ... error
[...]

PR #5123, opened 2021-01-25
Update fixate-requirements.py to cope with PIP 20.3.3

amanda11:
Building upon #5060. Updates to fixate-requirements.py to see if can use newer version of pip for make requirements.


amanda11:
Need to resolve problem seen when try and building this version of ST2 using the dockerfiles from https://github.com/StackStorm/st2packaging-dockerfiles/pull/97:
```
ERROR: Command errored out with exit status 1:
[wheelhouse: st2]           command: /usr/bin/python3 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-wheel-njg1eihv/simplejson_b53c2b0962984f95ae8a5c6c5ab9b98d/setup.py'"'"'; __file__='"'"'/tmp/pip-wheel-njg1eihv/simplejson_b53c2b0962984f95ae8a5c6c5ab9b98d/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-w_idsqh0
[wheelhouse: st2]               cwd: /tmp/pip-wheel-njg1eihv/simplejson_b53c2b0962984f95ae8a5c6c5ab9b98d/
[wheelhouse: st2]          Complete output (6 lines):
[wheelhouse: st2]          Traceback (most recent call last):
[wheelhouse: st2]            File "<string>", line 1, in <module>
[wheelhouse: st2]            File "/tmp/pip-wheel-njg1eihv/simplejson_b53c2b0962984f95ae8a5c6c5ab9b98d/setup.py", line 39
[wheelhouse: st2]              except DistutilsPlatformError, x:
[wheelhouse: st2]                                           ^
[wheelhouse: st2]          SyntaxError: invalid syntax
```

amanda11:
Looks like if you upgrade the version of pip used when you do the build that it has a different way of determining dependencies, and it goes back through lots of old versions - if it gets to versions that aren't python3 compatible then it fails. So have to then fixate those requirements so the min is at least a py3 version.
See: https://github.com/pypa/pip/issues/8713

amanda11:
The build takes much longer with this pip, will need to probably update the version requirements to much higher ones to enable it to have less versions to compare when it works out its dependencies.

amanda11:
If you use the option to use PIP 20.3.4 with the legacy resolver, then it builds much quicker. But then get the opposite problem that you cannot list packageA directly if packageB lists packageA. If you do, and they don't have identical versions then it will fail.
e.g. if one doesn't specify a version number, and the other does -> it fails.
and if one specifies a version number, and the other specifies another - even if there is a package that meets both conditions, it will also fail.

Trying to see if remove the extra ones added during this PR if can now build a PR. It now gets past the wheelhouse, but fails on the package st2 step.

amanda11:
Managed to resolve all the 3rd party conflicts by removing the ones added, but now get odd complaints about st2 packages themselves:
```
[package: st2]      ERROR: Cannot install st2actions==3.4.dev0, st2api==3.4.dev0, st2auth==3.4.dev0, st2client==3.4.dev0, st2common==3.4.dev0, st2exporter==3.4.dev0, st2reactor==3.4.dev0, st2stream==3.4.dev0 and st2tests==3.3.dev0 because these package versions have conflicting dependencies.
[package: st2]      
[package: st2]      The conflict is caused by:
[package: st2]          st2actions 3.4.dev0 depends on st2client==3.4.dev0
[package: st2]          st2client 3.4.dev0 depends on st2client==3.4.dev0
[package: st2]          st2api 3.4.dev0 depends on st2client==3.4.dev0
[package: st2]          st2auth 3.4.dev0 depends on st2client==3.4.dev0
[package: st2]          st2common 3.4.dev0 depends on st2client==3.4.dev0
[package: st2]          st2exporter 3.4.dev0 depends on st2client==3.4.dev0
[package: st2]          st2reactor 3.4.dev0 depends on st2client==3.4.dev0
[package: st2]          st2stream 3.4.dev0 depends on st2client==3.4.dev0
[package: st2]          st2tests 3.3.dev0 depends on st2client==3.3.dev0
```

amanda11:
Packages built for Centos7 and passing the packagingbuild and test. 

Need to use with the docker containers from: https://github.com/StackStorm/st2packaging-dockerfiles/pull/98 and the rpm build changes from: https://github.com/StackStorm/st2-packages/pull/684

Need to verify can build all 4 variants, and perform some manual-testing with the built package to ensure that virtualenv version/pip are as expected...

test-requirements.txt
```diff
@@ -7,7 +7,7 @@ pylint==2.6.0
 pylint-plugin-utils>=0.4
 bandit==1.5.1
 ipython<6.0.0
-isort<=4.0.0
+isort>=4.2.5,<5
```
blag, line 10: Was there a reason you pinned this to <5.0.0?

According to [the changelog for `isort` v5.0.0](https://github.com/pycqa/isort/blob/master/CHANGELOG.md#500-penny---july-4-2020):

> isort now requires Python 3.6+ to run but continues to support formatting on ALL versions of python including Python 2 code.

Which should be fine for this repo now.

If we do need to keep it pinned to <5.0.0, that would also work fine, but I'd like to start documenting why we pin certain packages to certain versions.

If we keep that pinned to a maximum version, it [potentially complicates our life in the future with pylint](https://app.circleci.com/pipelines/github/StackStorm/st2-rbac-backend/551/workflows/5c2127ec-d717-4ef2-ad1a-e6866e7f102a/jobs/2081):

> ERROR: Cannot install -r /tmp/st2/test-requirements.txt (line 6) and isort<=4.0.0 because these package versions have conflicting dependencies.
> 
> The conflict is caused by:
>     The user requested isort<=4.0.0
>     pylint 2.6.0 depends on isort<6 and >=4.2.5
> 
> To fix this you could try to:
> 1. loosen the range of package versions you've specified
> 2. remove package versions to allow pip attempt to solve the dependency conflict

StackStorm/st2-rbac-backend#46 is being held up by this PR due to this, which is not a big deal, it just forces a resolution order as we slowly crawl towards the next major ST2 release.

amanda11: @blag I copied the version that was in st2-auth-ldap that was needed to get the CI working on st2-auth-ldap. And I believe that requirement was for pylint 2.6.0.
BUT that's a bit odd that the same pylint is now saying < 6 not < 5.
I was also seeing the new pip complain if you have one requirement saying packageA >=4.2.5   and packageB say >=4.25 and < 6. I will see if I can get it to pass with isort > 4.2.5, but I think with the new pip that will complain. But I'll double-check in case it was something else.

amanda11:
Changes not enough - as when used rpm to install ended up with pip 19.1.1.
Noticed I had lost change to use later virtualenv, but need to be careful as virtualenv 50.2.2 uses pip 20.3.1 according to: https://virtualenv.pypa.io/en/latest (which doesn't have the legacy resolver) 

Looking at virtualenv history the last version of virtualenv that used a pip 20.2 was version 20.2.0 which uses this version:
  "pip": "pip-20.2.4-py2.py3-none-any.whl",
        "setuptools": "setuptools-50.3.2-py3-none-any.whl",
        "wheel": "wheel-0.35.1-py2.py3-none-any.whl",

amanda11:
Checked on the new packages in the packagingteste, and confirmed that /opt/stackstorm/st2/bin/pip3 was version 20.3.3, and that if used that pip it was using virtualenv 20.2.2 and wheel 0.29.0

This doesn't match the virtualenv 20.2.2 which states it was using wheel 0.36.1

Not sure why the original upgrade on setuptools had got removed, so going to re-instate...

amanda11:
Latest versions on dockertest after install:
```
[root@1a6f3e7adfea /]#  /opt/stackstorm/st2/bin/pip3 --version
pip 20.3.3 from /opt/stackstorm/st2/lib/python3.6/site-packages/pip (python 3.6)
[root@1a6f3e7adfea /]#  /opt/stackstorm/st2/bin/virtualenv --version
virtualenv 20.2.2 from /opt/stackstorm/st2/lib/python3.6/site-packages/virtualenv/__init__.py

```

and in a virtualenv for a pack:
```
wheel-0.36.1.dist-info
setuptools-51.0.0.dist-info
pip-20.3.1.dist-info 
```

So virtualenvs are getting the version of pip that comes with virtualenv.

Going to try and up virtualenv to 20.4.0 which is latest, and would pull in pip 20.3.3

amanda11:
Installed centos7 rpm following the manual instructions on a CentOS 7 build and all installed correctly, and successfulled installed a dummy pack that had a requirements.txt contained aio-pika that failed before.

```
[centos@centos-7-test ~]$ /opt/stackstorm/st2/bin/pip3 --version
pip 20.3.3 from /opt/stackstorm/st2/lib/python3.6/site-packages/pip (python 3.6)
[centos@centos-7-test ~]$ /opt/stackstorm/st2/bin/virtualenv --version
virtualenv 20.4.0 from /opt/stackstorm/st2/lib/python3.6/site-packages/virtualenv/__init__.py
[centos@centos-7-test ~]$ ls /opt/stackstorm/virtualenvs/pack1/
bin/        .gitignore  lib/        lib64/      pyvenv.cfg  
[centos@centos-7-test ~]$ ls /opt/stackstorm/virtualenvs/pack1/lib*/python3.6/site-packages/
/opt/stackstorm/virtualenvs/pack1/lib64/python3.6/site-packages/:
multidict  multidict-5.1.0.dist-info  yarl  yarl-1.6.3.dist-info

/opt/stackstorm/virtualenvs/pack1/lib/python3.6/site-packages/:
aio_pika                  pkg_resources
aio_pika-6.7.1.dist-info  __pycache__
aiormq                    setuptools
aiormq-3.3.1.dist-info    setuptools-51.3.3.dist-info
_distutils_hack           setuptools-51.3.3.virtualenv
distutils-precedence.pth  six-1.15.0.dist-info
easy_install.py           six.py
idna                      typing_extensions-3.7.4.3.dist-info
idna-3.1.dist-info        typing_extensions.py
pamqp                     _virtualenv.pth
pamqp-2.3.0.dist-info     _virtualenv.py
pip                       wheel
pip-20.3.3.dist-info      wheel-0.36.2.dist-info
pip-20.3.3.virtualenv     wheel-0.36.2.virtualenv
```

amanda11:
Noticed a time difference between installing packs on pip 20.3.3.

I have a pack that has dependencies on 7 packs. On the new environment it took 119s to install_pack_requirements, and on an older system 8s. So we probably have to change the st2 core as well to use legacy resolver if pip is 20.x - or make that an option... to pack install.

cognifloyd:
This is just a note to let you know why merging master into this branch will be a bit less fun this time around and to commiserate on how hard it is to get pip updated...

scripts/fixate-requirements.py
```diff
@@ -138,19 +144,34 @@ def merge_source_requirements(sources):
...
+                parsedreq = parse_req_from_line(req.requirement, req.line_source)
+                if parsedreq.requirement:
+                    # Skip already added project name
+                    if parsedreq.requirement.name in projects:
+                        continue
+                    projects.add(parsedreq.requirement.name)
+                    merged_requirements.append(req)
+
+                    # Requirements lines like "vcs+proto://url"
+                elif parsedreq.link:
```
cognifloyd, line 158: Is black indenting this comment? I had to reread it a couple times to realize it was for the next line, not the current block.
```suggestion
                # Requirements lines like "vcs+proto://url"
                elif parsedreq.link:
```

scripts/fixate-requirements.py
```diff
@@ -138,19 +144,34 @@ def merge_source_requirements(sources):
...
-                raise RuntimeError("Unexpected requirement {0}".format(req))
+                if req.req:
+                    # Skip already added project name
+                    if req.name in projects:
+                        continue
+                    projects.add(req.name)
+                    merged_requirements.append(req)
+
+                    # Requirements lines like "vcs+proto://url"
+                elif req.link:
```
cognifloyd, line 171:
```suggestion
                # Requirements lines like "vcs+proto://url"
                elif req.link:
```

fixed-requirements.txt
```diff
@@ -46,9 +46,8 @@ six==1.13.0
 sseclient-py==1.7
 stevedore==1.30.1
 tooz==2.8.0
-# Note: virtualenv embeds the pip wheel, so pinning virtualenv also pins pip
-# virtualenv==20.0.18 has pip==20.0.2
-virtualenv==20.0.18
+# Note: Use version of virtualenv which uses a pip < 21 - 20.4.0 uses pip 20.3.3, wheel 0.36.2 and setuptools 51.3.3
```
cognifloyd, line 49:
```suggestion
# Note: virtualenv embeds wheels for pip, wheel, and setuptools. So pinning virtualenv pins those as well.
# virtualenv==20.4.0 (<21) has pip==20.3.3 wheel==0.36.2 setuptools==51.3.3
```

cognifloyd (commented):
Just a few comment suggestions. The code looks great.

cognifloyd:
CI failures are resolved on master. Merging should resolve them.

cognifloyd:
Weird. I wonder why GHA didn't check the last commit.

amanda11:
Related change is merged, but waiting for dockerfiles to be published: https://github.com/StackStorm/st2packaging-dockerfiles/pull/98
May need to wait for https://github.com/StackStorm/st2-packages/pull/684 to be merged

Kami (approved):

armab (approved):
:+1: 

Thanks a lot for diving deep here!

PR #5131, opened 2021-01-29
Pin PyOpenSSL<20 and prettytable<2 due to python2.7 conflicts

cognifloyd:
In order to install st2 v3.3 requirements on python 2.7, we need to pin `PyOpenSSL<20` and `prettytable<2`.

We require `requests==2.23.0` which needs `pyOpenSSL >= 0.14`.
https://github.com/psf/requests/blob/v2.23.0/setup.py#L105

We also pin `cryptography==2.8`, but `pyOpenSSL==20.0.0` was released Nov 27, 2020 and that requires `cryptography>=3.2`.
https://github.com/pyca/pyopenssl/blob/20.0.0/setup.py#L98

st2client requires prettytable. prettytable 2.0 was released in November as well, and it does not support python2. pip is still trying to install it though, so we need to prevent that.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -142,7 +142,7 @@ jobs:
             set -x
             git clone ${ST2_PACKAGES_REPO} ~/st2-packages
             cd ~/st2-packages
-            git checkout ${CIRCLE_BRANCH} || true
+            git checkout ${CIRCLE_BRANCH} || git checkout v3.3 || true
       - run:
           name: Initialize packages Build Environment
           command: |
--- a/fixed-requirements.txt
+++ b/fixed-requirements.txt
@@ -5,9 +5,6 @@ apscheduler==3.6.3
 # NOTE: 2.0 version breaks pymongo work with hosts
 dnspython>=1.16.0,<2.0.0
 cryptography==2.8
-# requests==2.23 (below) needs pyOpenSSL>=0.14, but pyOpenSSL 20 (Nov 2020)
-# requires cryptography>=3.2 which conflicts with our 2.8 requirement.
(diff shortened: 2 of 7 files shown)


cognifloyd:
Fixing this is required to fix the exchange pack tests on python2.7.

For the pack tests, we don't need to cut a release. Later, we'll probably need to release v3.3.1 with this fix. Anyone that needs python2.7 can't upgrade to v3.4, and this install is broken on 2.7.

cognifloyd:
I’m trying to fix the packaging build. At first it was trying to use the master branch of st2-packaging to build this 3.3 only change. That checked for an rbac command that is only installed with the rbac package and that isn’t installed on 3.3, so I fixed that here: https://github.com/StackStorm/st2/pull/5131/commits/97a031525ac8392c271a27967ebac000828fc794

Now, `el8` and `bionic` are passing on circleci, but both `el7` and `xenial` are not.
They are both complaining that pip is not installed during “Build the ${DISTRO} Packages”. Any ideas?
https://app.circleci.com/pipelines/github/StackStorm/st2/948/workflows/07219ab8-90a7-403e-a1f8-adc553d3d6de/jobs/12289/parallel-runs/0?filterBy=ALL

Oh, and Travis was passing fine. One of the builds errored this time, but re-running should fix it.

cognifloyd:
It turns out that packaging build images no longer have python2.7. So, since `xenial` and `centos7` needed python2.7 to build, we can no longer build the packages.

Re-adding python2.7 would take significant effort and probably isn't worth it. See the discussion here:
https://github.com/StackStorm/discussions/issues/40#issuecomment-769992629

PR #5134, opened 2021-01-29
Update versions of st2tests and isort

amanda11:
Update version of isort to be compatible with pylint.
Also update st2tests to be version 3.4dev


amanda11:
isort changes are required for 
version of st2tests wasn't updated at last release

Although changes are included in https://github.com/StackStorm/st2/pull/5123 do not want to block other PRs for that one, which is a large change.https://github.com/StackStorm/st2-rbac-backend/pull/46

armab (approved):

PR #5138, opened 2021-02-07
Fix "st2 action-alias execute" CLI command

Kami:
I tried to use StackStorm after quite a long time and in the first 5 minutes I encountered two exceptions / bugs (like in the good old days - guess I'm always the "lucky" one :D).

There are two bugs in ``st2 action-alias execute`` command so it doesn't work.

1) If ``user`` query param is not specified, it will fail with the error below due to API returning an error due to type mismatch (out of the box behavior with st2-docker).

```bash
# -------- begin 140132975715104 request ----------
curl -X POST -H  'User-Agent: python-requests/2.23.0' -H  'Accept-Encoding: gzip, deflate' -H  'Accept: */*' -H  'Connection: keep-alive' -H  'X-Auth-Token: XXXX' -H  'content-type: application/json' -H  'Content-Length: 61' --data-binary '{"command": "foo=bar", "user": null, "source_channel": "cli"}' http://st2api:9101/aliasexecution/match_and_execute
# -------- begin 140132975715104 response ----------
{
    "faultstring": "None is not of type 'string'"
}
```

2) Result for ``match_and_execute``  API operation is not correctly de-serialized in the client so client returns empty object and the code throws an exception. API returns a container object with a list so we need to handle that correctly.

```bash
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2client/shell.py", line 408, in run
    func(args)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2client/commands/action_alias.py", line 127, in run_and_print
    print("Matching Action-alias: '%s'" % execution.actionalias['ref'])
AttributeError: 'ActionAliasExecution' object has no attribute 'actionalias'
```

This PR fixes both issues.

## TODO

- [x] Add tests (no idea when I'll get a chance to do it, but for sure we need tests because I assume there are not otherwise they would have caught this bug)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -53,6 +53,11 @@ Fixed
 
 * StackStorm now explicitly decodes pack files as utf-8 instead of implicitly as ascii (bug fix) #5106, #5107
 
+* Fix broken ``st2 action-alias execute`` command and make sure it works
+  correctly. (bug fix) #5138
+
+  Contributed by @Kami.
+
 Removed
 ~~~~~~~~
 * Removed --python3 pack install option  #5100
--- a/st2api/tests/unit/controllers/v1/test_action_alias.py
+++ b/st2api/tests/unit/controllers/v1/test_action_alias.py
@@ -24,7 +24,8 @@ from st2tests.api import APIControllerWithIncludeAndExcludeFilterTestCase
 FIXTURES_PACK = 'aliases'
 
 TEST_MODELS = {
-    'aliases': ['alias1.yaml', 'alias2.yaml', 'alias_with_undefined_jinja_in_ack_format.yaml']
+    'aliases': ['alias1.yaml', 'alias2.yaml', 'alias_with_undefined_jinja_in_ack_format.yaml'],
+    'actions': ['action1.yaml'],
 }
 
 TEST_LOAD_MODELS = {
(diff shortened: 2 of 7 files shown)


Kami:
Adding direct ``st2client`` tests will be a bit of a pain so to begin with, I added in-direct test for the API endpoint itself (5df4e8e51 - similar to the ones in #5141) which verify that API endpoint indeed returns a list with specific key present.

Kami:
OK, I couldn't stand not having any tests for such a basic functionality so I also added direct tests in 19e220698978a5cfe8129d1f207d390d1a343c1a (still not ideal since response is mocked, but that's the pattern we use for CLI tests :/).

I also pushed a small related improvements which such make troubleshooting CLI tests easier - 75a114c81089fffc5d7550fc4a65f2ed33d734a1.

armab (approved):

cognifloyd (approved):
Wow. I didn't know about `st2 action-alias execute`. Awesome.
Fixes and tests lgtm

Kami:
I'm also working on related functionality - ``execute-and-format`` which will run the execution which matches an alias and also format and print the result - this will make testing ChatOps easier and faster since you will be able to do it quickly on command line and no need to do it via chat.

PR #5140, opened 2021-02-07
Fix st2client @key=file_path.json notation for specifying complex objects in a file

Kami:
I tried to use ``st2 run @param=file_path.json`` notation to specify complex JSON parameter value in a file.

It turns out it's not working correctly. Using ``--debug`` flag it shows it's not correctly converting bytes to unicode which means JSON serialization fails (actual value has``b""`` prefix which indicates that code is incorrectly calling ``str()`` on the value somewhere instead of using ``bytes.decode()``).

```bash
st2 --debug run scalyr.timeseries_queries @queries=1.json
...
ERROR: Object of type 'bytes' is not JSON serializable
....
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2client/utils/httpclient.py", line 98, in post
    response = requests.post(self.root + url, json.dumps(data), **kwargs)
  File "/usr/lib/python3.6/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/usr/lib/python3.6/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib/python3.6/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/usr/lib/python3.6/json/encoder.py", line 180, in default
    o.__class__.__name__)
TypeError: Object of type 'bytes' is not JSON serializable
```

This pull request fixes that.

## TODO

- [x] Add tests
- [x] Add tests and verify it works correctly with http runner (that one may expect bytes, but unlikely since the current code doesn't handle that anyway)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -53,6 +53,11 @@ Fixed
 
 * StackStorm now explicitly decodes pack files as utf-8 instead of implicitly as ascii (bug fix) #5106, #5107
 
+* Fix ``@parameter_name=/path/to/file/foo.json`` notation in the ``st2 run`` command which didn't
+  work correctly because it didn't convert read bytes to string / unicode type. (bug fix) #5140
+
+  Contributed by @kami.
+
 Removed
 ~~~~~~~~
 * Removed --python3 pack install option  #5100
--- a/st2client/st2client/commands/action.py
+++ b/st2client/st2client/commands/action.py
@@ -541,6 +541,12 @@ class ActionRunCommandMixin(object):
         action_ref_or_id = action.ref
 
         def read_file(file_path):
+            """
+            Read file content and return content as string / unicode.
+
+            NOTE: It's only mean to be used to read non-binary files since API right now doesn't
+            support passing binary data.
+            """
             if not os.path.exists(file_path):
                 raise ValueError('File "%s" doesn\'t exist' % (file_path))
 
--- a/st2client/st2client/commands/action.py
+++ b/st2client/st2client/commands/action.py
@@ -550,7 +550,7 @@ class ActionRunCommandMixin(object):
             with open(file_path, 'rb') as fp:
                 content = fp.read()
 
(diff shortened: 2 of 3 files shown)


Kami:
I've added tests for it (32d3c8df0) and verified the functionality was indeed never meant to be used with binary data, but only string data.

In theory, we could probably also make it support binary data, but this would likely also require changes in the http layer (client, API) and I don't think anyone requests something like that before.

armab (approved):
That's a good find.
I bet we have more dark rarely used corners like that.

cognifloyd (approved):
LGTM. I haven't looked at Travis to see why it's failing.

blag:
I can't restart the Travis-CI job for some HTTP 403 reason. I tried signing out of Travis and back in, same result. @Kami Can you try restarting the Travis build?

Kami:
Lint check is failing and it doesn't appear to be related to any of my changes:

```
 ************* Module st2common.models.db.execution
st2common/st2common/models/db/execution.py:101:39: E1101: Instance of 'ActionExecutionDB' has no 'id' member (no-member)
st2common/st2common/models/db/execution.py:101:39: E1101: Instance of 'ActionExecutionDB' has no 'id' member (no-member)
st2common/st2common/models/db/execution.py:101:39: E1101: Instance of 'ActionExecutionDB' has no 'id' member (no-member)
st2common/st2common/models/db/execution.py:101:39: E1101: Instance of 'ActionExecutionDB' has no 'id' member (no-member)
st2common/st2common/models/db/execution.py:101:39: E1101: Instance of 'ActionExecutionDB' has no 'id' member (no-member)
st2common/st2common/models/db/execution.py:101:39: E1101: Instance of 'ActionExecutionDB' has no 'id' member (no-member)
```

It seems more pylint related - e.g. our custom pylint plugin not running or similar.

PR #5141, opened 2021-02-07
Fix a bug in invalid action alias array parameter casting

Kami:
This pull request fixes a bug in action alias array parameter value casting when executing an alias.

The code would incorrectly assume value is always a string and try to cast it to an array (the value may very well be a string when executing command from chat, but even then is not always true).

This would not work if value is already a list (e.g. if parameter contains a default value or a default value is specified using ``immutable_parameters`` in action alias definition).

## Background, Context

I encountered this bug when trying to define a default value for a parameter in action alias which is a list.

Here is action alias definition in question:

```yaml
---
name: "test"
pack: "scalyr"
action_ref: "scalyr.timeseries_queries"
description: "Get server temps."
formats:
    - "temps"
result:
  format: |
    CPUs: 

    Core 0: {{ execution.result.result.results[0]["values"][0]|int }} C
    Core 1: {{ execution.result.result.results[1]["values"][0]|int }} C 
immutable_parameters:
  queries:
    -
      filter: "$source='tsdb' $serverHost='foo' metric = 'output' and instance = 'core_0_temp_celsius'"
      function: "mean(value)"
      startTime: "10 minutes"
      buckets: 1
    -
      filter: "$source='tsdb' $serverHost='foo' metric = 'output' and instance = 'core_1_temp_celsius'"
      function: "mean(value)"
      startTime: "10 minutes"
      buckets: 1
 ...
```

## TODO

- [x] Tests

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -56,6 +56,13 @@ Fixed
 
 * StackStorm now explicitly decodes pack files as utf-8 instead of implicitly as ascii (bug fix) #5106, #5107
 
+* Fix incorrect array parameter value casting when executing action via chatops or using
+  ``POST /aliasexecution/match_and_execute`` API endpoint. The code would incorrectly assume the
+  value is always a string, but that may not be the cast - they value could already be a list and
+  in this case we don't want any casting to be performed. (bug fix) #5141
+
+  Contributed by @Kami.
+
 Removed
 ~~~~~~~
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -54,7 +54,8 @@ Fixed
 
   Contributed by @guzzijones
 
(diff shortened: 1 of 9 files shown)


Kami:
Added tests.

It was quite painful. My old development environment was not working and getting a new one up also took much longer then it should (couldn't find any good up to date docs so I created one myself from scratch).

Kami:
On a related note - I assume the whole str to list casting code was added as a quick hack / workaround at some point in the past to make it easier to specify lists in a chat.

It's not ideal or complete since it doesn't handle more complex scenarios (e.g. lists of objects with ``key1=value,key2=value`` notation), but my main and only objective in this PR is to fix it so it works correctly for "already a list" scenario.

armab (approved):
:+1:

Issue #5144, opened 2021-02-09
Duplicated CI executions in GH Actions


armab:
Just noticed that we got duplicated test executions (for pull_request and push).  Something might be changed from the GH Actions side and that could be a really quick adjustment. So it's not a bug, but a feature that we want to disable.

![multiple-CircleCI-runs-_2021-02-09_14 27 03](https://user-images.githubusercontent.com/1533818/107377560-05a8c880-6ae3-11eb-9de3-406d9836952f.png)

Good first issue, help wanted from anyone willing to contribute.

armab:
Implemented in https://github.com/StackStorm/st2/pull/5150

PR #5150, opened 2021-02-12
Adjust GitHub Actions config to avoid double testing on PRs

blag:
Our current GitHub Actions configuration runs tests twice for each push - once when the branch is pushed, and once more when a pull request is opened, reopened, or edited.

This is inefficient and slows down development since people have to wait for more tests to finish.

This PR slightly tweaks the GHA config to only run tests on pushes to:

* `master` - ensure post-merge code is good
* `v*` branches - version branches that track minor versions (eg: `v3.4`)
* `v*` tags - version tags that track bugfix releases (eg: `v3.4.0`), typically are tagged on commits to version branches

I have included comments on the configuration to ensure that everybody who isn't familiar with GHA understands what the intent is.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,9 +2,11 @@ name: ci
 
 on:
   push:
-    branches: ['*']
+    branches:
+      - master  # <-- only on merges to master branch
+      - v*      #     and version branches, which only include minor versions (eg: v3.4)
     tags:
-      - v*
+      - v*      # <-- also version tags, which include bugfix releases (eg: v3.4.0)
   pull_request:
     type: [opened, reopened, edited]
   schedule:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,6 +12,10 @@ on:
       - v[0-9]+.[0-9]+.[0-9]+
   pull_request:
     type: [opened, reopened, edited]
+    branches:
+      # Only for PRs targeting those branches
+      - master
+      - v[0-9]+.[0-9]+
   schedule:
     # run every night at midnight
     - cron:  '0 0 * * *'
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,10 +3,10 @@ name: ci
 on:
   push:
     branches:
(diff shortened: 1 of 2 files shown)


cognifloyd (approved):
LGTM

.github/workflows/ci.yaml
```diff
@@ -2,9 +2,14 @@ name: ci
...
+      # only on merges to master branch
+      - master
+      # and version branches, which only include minor versions (eg: v3.4)
+      - v[0-9]+.[0-9]+
     tags:
-      - v*
+      # also version tags, which include bugfix releases (eg: v3.4.0)
+      - v[0-9]+.[0-9]+.[0-9]+
   pull_request:
     type: [opened, reopened, edited]
```
Kami, line 14: Technically we could / should probably also filter pull requests so builds are only triggered if it's a PR against master branch (and perhaps a version one?).

st2common/st2common/models/db/execution.py
```diff
@@ -98,7 +98,7 @@ class ActionExecutionDB(stormbase.StormFoundationDB):
 
     def get_uid(self):
         # TODO Construct od from non id field:
-        uid = [self.RESOURCE_TYPE, str(self.id)]
+        uid = [self.RESOURCE_TYPE, str(self.id)]  # pylint: disable=no-member
```
Kami, line 101: Please add a comment on why we need this pylint pragma and hopefully we can figure out root cause in the near future.

Kami:
I added branches filter to PR trigger and I will merge it once the build passes so I can sync and merge rest of my PRs.

st2common/st2common/models/db/execution.py
blag: I don't know the root cause. pylint just randomly decided to not like that line, so I dug through the code, and checked the mongoengine documentation to make sure that the object would have an `id` attribute by the time that code is evaluated.

amanda11:
LGTM

amanda11 (approved):

Kami (approved):

armab:
closes https://github.com/StackStorm/st2/issues/5144

PR #5152, opened 2021-02-13
Add support to the client (CLI, http client) for sending additional basic auth credentials (aka MFA)

Kami:
This pull request adds support for new ``credentials.basic_auth = username:password`` StackStorm CLI config option and ``-basic-auth`` CLI argument.

This argument allows client to use additional set of basic auth credentials when talking to the StackStorm API endpoints (api, auth, stream) - that is, in addition to the token / api key native StackStorm auth.

## Background, Context

I'm running a StackStorm instance which I want to put behind additional auth wall (think second factor auth). I could probably achieve that using SSO and handling multi factor auth on the SSO side, but I'm not using SSO nor want to use it (and even with SSO used, if there is a bug in StackStorm auth code for SSO, dual factor auth on SSO side would not help in this case while having additional auth in another layer like nginx would).

The easiest way to add second factor to any kind of application which runs behind some kind of http proxy (nginx, Apache, etc.) is to simply utilize existing basic auth primitives of the proxy - this PR does exactly that.

## Notes, Limitations

1. Current CLI and http client code is very messy and hard to work it. We don't have good abstractions for it (a lot of kwargs hacks all over the place) so it's a PITA to update it and work with it (something which has been a known issue for a long time already).

2. If this second factor auth is enabled and user wants to use webhook integrations, they will also need to send basic auth credentials with webhooks (which shouldn't be an issue since most 3rd party service support that and if they don't, they could also add exceptions in http proxy layer to not require it for webhooks API endpoint or similar).

3. Currently ``/v1/auth/tokens`` API endpoint also utilizes basic auth for authentication which means you should disable proxy based basic auth wall for ``/v1/auth/tokens`` API endpoint. This is of course not ideal since it means actually ``auth`` endpoint won't be behind MFA wall (aka user may be able to auth and obtain a valid st2 auth token without additional basic auth auth, but other operations and endpoints won't work if valid basic auth header is not provided), but all other endpoints will be.

In nginx proxy, this would look something like this:

```nginx
...
# Require additional basic auth credentials for all StackStorm API endpoints except /auth/tokens one
location / {
   auth_basic "off";
   auth_basic	"Auth required";
   auth_basic_user_file	/etc/nginx/htpasswd/foo.bar.example;
  ...
}
...

location /auth/tokens {
   auth_basic "off";
  ...
}
```

## TODO

- [x] Unit tests
- [x] Verify it works with all the endpoints (so far I verified all the basic one + event stream aka execution tail one)
- [x] Verify st2web (may require changes in case the current client doesn't simply send existing basic auth credentials browser has cached)


Kami:
I just tested it and it indeed appears it doesn't work with st2web XHR requests out of the box. With many other apps I didn't have such issues - XHR requests would simply re-use any basic auth credentials (aka Authorization header which browser has saved).

So I need to dig in some more to see what is going on. One option would likely be to add those credentials as additional option to st2web (via web config or via text input), but that's something I would like to avoid. Worst case, I case use "cookie hack" on the http proxy side which I believe should also work with XHR requests.

blag:
> If this second factor auth is enabled and user wants to use webhook integrations, they will also need to send basic auth credentials with webhooks (which shouldn't be an issue since most 3rd party service support that and if they don't, they could also add exceptions in http proxy layer to not require it for webhooks API endpoint or similar).

I haven't ever run across a third party service that supports basic auth for outgoing webhooks. Jira doesn't, GitHub doesn't. So allowing the base webhook endpoint, and all children, to bypass this would definitely be required.

I'm not hugely happy with this. I think putting our efforts into an SSO backend that handles MFA itself would be better.

Also, I'm afraid that this change would give users more undeserved confidence in running StackStorm exposed to the internet, which I don't think has been our security model/mindset so far. I'm already aware of one such instance, and I hope it's the only one.

st2client/st2client/client.py
```diff
@@ -129,115 +130,194 @@ def __init__(
 
         self.api_key = api_key
 
+        if basic_auth:
+            if len(basic_auth.split(":")) != 2:
```
winem, line 134: This prohibits passwords with a `:` in them. We should either document the password requirements somewhere or update the operator to >=.

Kami:
@blag 

> I haven't ever run across a third party service that supports basic auth for outgoing webhooks. Jira doesn't, GitHub doesn't. So allowing the base webhook endpoint, and all children, to bypass this would definitely be required.

I'm pretty sure Github and lot of other services support ``https://<credentials>@url`` notation for the webhook URLs and handle credentials correctly.

But if that's not possible in some situations, user can also handle that in nginx config level (similar as tokens auth endpoint) so that can be documented.

> I'm not hugely happy with this. I think putting our efforts into an SSO backend that handles MFA itself would be better.
> Also, I'm afraid that this change would give users more undeserved confidence in running StackStorm exposed to the internet, which I don't think has been our security model/mindset so far. I'm already aware of one such instance, and I hope it's the only one.

Security is all about layers and I think it's good to give users a choice of adding additional layer on top StackStorm auth (+optionally SSO with MFA).

SSO + MFA is definitely a good choice, but for some more security conscious users, it's better to have another layer somewhere outside of the app layer.

There were definitely cases in the past where it was possible to bypass login + mfa due to the bug in the app code.

st2client/st2client/client.py
Kami: Good catch, will use left split with limit of 1 (this should work as long as the username doesn't contain a colon).

st2client/st2client/client.py
```diff
@@ -129,115 +130,194 @@ def __init__(
...
+                raise ValueError(
+                    "basic_auth config options needs to be in the "
+                    "username:password notation"
+                )
+
+            self.basic_auth = tuple(basic_auth.split(":"))
```
blag, line 140: Why not use tuple unpacking?

```suggestion
            username, *password = basic_auth.split(":")
            password = ":".join(password)  # Put the password back together
            self.basic_auth = (username, password)
```

blag:
> I'm pretty sure Github and lot of other services support https://<credentials>@url notation for the webhook URLs and handle credentials correctly.

Oh, right, yes they do, nevermind.

st2client/st2client/client.py
Kami: That would also work, but per my comment above (https://github.com/StackStorm/st2/pull/5152#discussion_r618716121), splitting with ``maxsplit=1`` is even simple - 1 line vs 3 lines :)

st2client/st2client/client.py
Kami: e67d40e9d

st2client/st2client/client.py
Kami: e67d40e9d

st2client/st2client/models/core.py
```diff
@@ -679,8 +692,15 @@ def match(self, instance, **kwargs):
...
+        super(StreamManager, self).__init__(
+            resource=None,
+            endpoint=endpoint,
+            cacert=cacert,
+            debug=debug,
+            basic_auth=basic_auth,
+        )
         self._url = httpclient.get_url_without_trailing_slash(endpoint) + "/stream"
         self.debug = debug
         self.cacert = cacert
```
cognifloyd, line 706:
```suggestion
```

cognifloyd (approved):

nzlosh:
@Kami Would you mind rebasing the branch against master so we can merge it for the 3.7 release?

Kami:
@nmaludy I've added changelog entry and synced it up with latest master. Will go ahead and merge it once CI passes - thanks.

PR #5155, opened 2021-02-16
Add a debug contrib pack

blag:
This PR starts a contributed `debug` pack that is useful for debugging StackStorm itself.

Right now it only has a few actions:

* [x] `pythonpath` returns the `PYTHONPATH` and the value of `sys.path`
* [x] `python_version` returns the current Python version in a dictionary/object:

  ```jsonc
  {
    "version": "...",  // from Python's sys.version_info
    "details": {
      "major": ...,  // from Python's sys.version_info.major
      "minor": ...,  // from Python's sys.version_info.minor
      "micro": ...,  // from Python's sys.version_info.micro
      "releaselevel": ...,  // from Python's sys.version_info.releaselevel
      "serial": ...,  // from Python's sys.version_info.serial
    }
  }
  ```
* [x] `print_ctx` prints the passed-in context dictionary to `stdout`

I'm sure more ideas will spring up from time to time, and this pack is a good place to add them.

We will need to add some tests, and I think we will eventually want to install this pack by default to help users.

Edit: Updated description with additional actions


cognifloyd:
Just merge master to get the black config, and then you can reformat this with `black`.

Kami:
IIRC, we already have a bunch of actions like that in ``examples`` pack  (and even the one which does exactly the same thing as the one you added) so we should decide what to do - likely the best course of actions is to move all such actions into the new ``debug`` + rectify duplication (especially since examples pack is already quite large).

It may require some end to end tests to be updated though.

cognifloyd:
@blag, can you reformat with black and rebase on / merge in master?
I would love to see this in 3.6.0

cognifloyd (approved):

blag:
@Kami Yes, we have the `examples` pack, but I think we overestimate how useful that is for users, since there's examples for _everything_ in there. I think we should be really judicious about what we move into the `debug` pack, especially if we're going to install it by default.

contrib/debug/actions/pythonpath.py
```diff
@@ -0,0 +1,15 @@
+import os
```
amanda11, line 1: Should we have the copyright statement on this file?

amanda11 (commented):
LGTM - but just a query on copyright header statement...

blag:
I've added a few more actions if people would like to review them.

cognifloyd:
`debug.python_version` and `debug.print_ctx` look like good, generally useful utilities. Nice!

contrib/debug/actions/print_ctx.py
```diff
@@ -0,0 +1,33 @@
...
+
+try:
+    from st2common.runners.base_action import Action
+except ImportError:
+    Action = object
+
+
+class PrintCtxAction(Action):
+    def run(self, ctx=None, *args, **kwargs):
+        print(ctx)
```
cognifloyd, line 28: Should we prettyprint it for nested objects?

blag:
Possibly blocked by #5257.

contrib/debug/actions/print_ctx.py
blag: That's a good idea, but no. I want this action to do the absolute bare minimum so you know exactly what data you are feeding it.

That's a great idea for a different action though!

amanda11 (approved):
LGTM

PR #5158, opened 2021-02-19
Add error handler or log error instead of using "pass" on exception in st2

ashwini-orchestral:
Replace pass statement with explicit error logs. The pass/continue statement when run in production (non-debug) code will skip the error. We want to practice safe coding so exception handling is necessary.

### DIFF
--- a/contrib/linux/actions/checks/check_processes.py
+++ b/contrib/linux/actions/checks/check_processes.py
@@ -52,8 +52,8 @@ class CheckProcs(object):
                 cmdfh = open(self.procDir + "/" + p + "/cmdline")
                 cmd = cmdfh.readline()
                 pInfo[1] = cmd
-            except:
-                continue
+            except IOError:
+                print("Error: can't find file or read data.")
             finally:
                 cmdfh.close()
                 fh.close()
--- a/contrib/linux/sensors/file_watch_sensor.py
+++ b/contrib/linux/sensors/file_watch_sensor.py
@@ -44,8 +44,8 @@ class FileWatchSensor(Sensor):
 
             try:
                 self._tail.notifier.stop()
-            except Exception:
-                pass
+            except Exception as e:
+                print("Error: tail notifier is still running.", e)
 
     def add_trigger(self, trigger):
         file_path = trigger['parameters'].get('file_path', None)
--- a/contrib/runners/python_runner/python_runner/python_runner.py
+++ b/contrib/runners/python_runner/python_runner/python_runner.py
@@ -322,7 +322,8 @@ class PythonRunner(GitWorktreeActionRunner):
 
         if ACTION_OUTPUT_RESULT_DELIMITER in stdout:
             split = stdout.split(ACTION_OUTPUT_RESULT_DELIMITER)
-            assert len(split) == 3
+            if not len(split) == 3:
+                raise ValueError("Length should be 3")
             action_result = split[1].strip()
             stdout = split[0] + split[2]
         else:
--- a/contrib/runners/python_runner/python_runner/python_runner.py
+++ b/contrib/runners/python_runner/python_runner/python_runner.py
@@ -323,7 +323,7 @@ class PythonRunner(GitWorktreeActionRunner):
         if ACTION_OUTPUT_RESULT_DELIMITER in stdout:
             split = stdout.split(ACTION_OUTPUT_RESULT_DELIMITER)
             if not len(split) == 3:
-                raise ValueError("Length should be 3")
+                raise ValueError('The result length should be 3.')
             action_result = split[1].strip()
             stdout = split[0] + split[2]
         else:
--- a/st2actions/st2actions/cmd/scheduler.py
+++ b/st2actions/st2actions/cmd/scheduler.py
@@ -109,7 +109,7 @@ def _run_scheduler():
             handler.shutdown()
             entrypoint.shutdown()
         except:
-            pass
+            LOG.exception('Unable to shutdown scheduler.')
 
         return 1
 
--- a/st2common/st2common/content/utils.py
+++ b/st2common/st2common/content/utils.py
@@ -282,7 +282,7 @@ def get_pack_file_abs_path(pack_ref, file_path, resource_type=None, use_pack_cac
     result = os.path.join(*path_components)     # pylint: disable=E1120
 
     if normalized_file_path not in result:
-        raise ValueError("This is not normalized path to prevent directory traversal")
+        raise ValueError('This is not a normalized path to prevent directory traversal.')
 
     # Final safety check for common prefix to avoid traversal attack
     common_prefix = os.path.commonprefix([pack_base_path, result])
--- a/st2common/st2common/content/utils.py
+++ b/st2common/st2common/content/utils.py
@@ -281,7 +281,8 @@ def get_pack_file_abs_path(pack_ref, file_path, resource_type=None, use_pack_cac
     path_components.append(normalized_file_path)
     result = os.path.join(*path_components)     # pylint: disable=E1120
 
-    assert normalized_file_path in result
+    if normalized_file_path not in result:
+        raise ValueError("This is not normalized path to prevent directory traversal")
 
     # Final safety check for common prefix to avoid traversal attack
     common_prefix = os.path.commonprefix([pack_base_path, result])
--- a/st2common/st2common/models/api/keyvalue.py
+++ b/st2common/st2common/models/api/keyvalue.py
@@ -189,7 +189,7 @@ class KeyValuePairAPI(BaseAPI):
 
             # Additional safety check to ensure that the value hasn't been decrypted
             if not value == original_value:
-                raise ValueError("Value doesn't match with original value")
+                raise ValueError("The value doesn't match with original value.")
         elif secret:
             cls._verif_key_is_set_up(name=name)
 
--- a/st2common/st2common/models/api/keyvalue.py
+++ b/st2common/st2common/models/api/keyvalue.py
@@ -188,7 +188,8 @@ class KeyValuePairAPI(BaseAPI):
                 raise ValueError(msg)
 
             # Additional safety check to ensure that the value hasn't been decrypted
-            assert value == original_value
+            if not value == original_value:
+                raise ValueError("Value doesn't match with original value")
         elif secret:
             cls._verif_key_is_set_up(name=name)
 
--- a/st2common/st2common/models/system/action.py
+++ b/st2common/st2common/models/system/action.py
@@ -177,7 +177,8 @@ class ShellCommandAction(object):
 
         :rtype: ``str``
         """
-        assert isinstance(args, (list, tuple))
+        if not isinstance(args, (list, tuple)):
+            raise ValueError("Args needs to be a list and tuple")
 
         args = [quote_unix(arg) for arg in args]
         args = ' '.join(args)
--- a/st2common/st2common/models/system/action.py
+++ b/st2common/st2common/models/system/action.py
@@ -178,7 +178,7 @@ class ShellCommandAction(object):
         :rtype: ``str``
         """
         if not isinstance(args, (list, tuple)):
-            raise ValueError("Args needs to be a list and tuple")
+            raise ValueError('The args has a value that is not a list and tuple.')
 
         args = [quote_unix(arg) for arg in args]
         args = ' '.join(args)
--- a/st2common/st2common/rbac/types.py
+++ b/st2common/st2common/rbac/types.py
@@ -185,7 +185,7 @@ class PermissionType(Enum):
 
         split = permission_type.split('_')
         if not len(split) >= 2:
-            raise ValueError("Length should be greater than or equal to 2")
+            raise ValueError('The length should be greater than or equal to 2.')
 
         return '_'.join(split[:-1])
 
@@ -198,7 +198,7 @@ class PermissionType(Enum):
         """
         split = permission_type.split('_')
         if not len(split) >= 2:
-            raise ValueError("Length should be greater than or equal to 2")
+            raise ValueError('The length should be greater than or equal to 2.')
 
         # Special case for PACK_VIEWS_INDEX_HEALTH
         if permission_type == PermissionType.PACK_VIEWS_INDEX_HEALTH:
--- a/st2common/st2common/rbac/types.py
+++ b/st2common/st2common/rbac/types.py
@@ -184,7 +184,8 @@ class PermissionType(Enum):
             return ResourceType.EXECUTION
 
         split = permission_type.split('_')
-        assert len(split) >= 2
+        if not len(split) >= 2:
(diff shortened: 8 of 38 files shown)


m4dcoder (changes requested):
@ashwini-orchestral Looks like you included changes from another PR/commit into this PR. It is hard to tell what changes relate to this PR. Can you clean this up? You should branch from master and then apply the changes.

st2common/st2common/content/utils.py
```diff
@@ -281,7 +281,8 @@ def get_pack_file_abs_path(pack_ref, file_path, resource_type=None, use_pack_cac
     path_components.append(normalized_file_path)
     result = os.path.join(*path_components)     # pylint: disable=E1120
 
-    assert normalized_file_path in result
+    if normalized_file_path not in result:
+        raise ValueError('This is not a normalized path to prevent directory traversal.')
```
blag, line 285: It would be good to include exactly what the path value is here.

st2common/st2common/models/api/keyvalue.py
```diff
@@ -188,7 +188,8 @@ def to_model(cls, kvp):
                 raise ValueError(msg)
 
             # Additional safety check to ensure that the value hasn't been decrypted
-            assert value == original_value
+            if not value == original_value:
```
blag, line 191: This comparison is weird.

```suggestion
            if value != original_value:
```

contrib/runners/python_runner/python_runner/python_runner.py
```diff
@@ -322,7 +322,8 @@ def _get_output_values(self, exit_code, stdout, stderr, timed_out):
 
         if ACTION_OUTPUT_RESULT_DELIMITER in stdout:
             split = stdout.split(ACTION_OUTPUT_RESULT_DELIMITER)
-            assert len(split) == 3
+            if not len(split) == 3:
```
blag, line 325: Weird comparison here.

```suggestion
            if len(split) != 3:
```

st2common/st2common/models/system/action.py
```diff
@@ -177,7 +177,8 @@ def _get_command_string(self, cmd, args):
 
         :rtype: ``str``
         """
-        assert isinstance(args, (list, tuple))
+        if not isinstance(args, (list, tuple)):
+            raise ValueError('The args has a value that is not a list and tuple.')
```
blag, line 181: Grammar fix.

```suggestion
            raise ValueError('The args has a value that is not a list or a tuple.')
```

st2common/st2common/rbac/types.py
```diff
@@ -184,7 +184,8 @@ def get_resource_type(cls, permission_type):
             return ResourceType.EXECUTION
 
         split = permission_type.split('_')
-        assert len(split) >= 2
+        if not len(split) >= 2:
```
blag, line 187: Weird comparison here.

```suggestion
        if len(split) < 2:
```

st2common/st2common/rbac/types.py
```diff
@@ -184,7 +184,8 @@ def get_resource_type(cls, permission_type):
...
+            raise ValueError('The length should be greater than or equal to 2.')
```
blag, line 188: This isn't a good error message, because it hides what the actual issue is, and it doesn't emit what the actual erroneous value is.

The root issue here is that the `permission_type` does not include an underscore. That should be the error message, with the bad value of `permission_type` included in the exception message.

st2common/st2common/rbac/types.py
```diff
@@ -196,7 +197,8 @@ def get_permission_name(cls, permission_type):
         :rtype: ``str``
         """
         split = permission_type.split('_')
-        assert len(split) >= 2
+        if not len(split) >= 2:
+            raise ValueError('The length should be greater than or equal to 2.')
```
blag, line 201: Same issue here - needs a better comparison and a better error message that identifies the root issue.

st2common/st2common/runners/base.py
```diff
@@ -288,7 +288,8 @@ def pre_run(self):
                                                   entry_point=self.entry_point,
                                                   worktree_path=self.git_worktree_path)
 
-            assert(entry_point.startswith(self.git_worktree_path))
+            if not (entry_point.startswith(self.git_worktree_path)):
```
blag, line 291: Outer parentheses aren't necessary

```suggestion
            if not entry_point.startswith(self.git_worktree_path):
```

st2common/st2common/runners/base.py
```diff
@@ -288,7 +288,8 @@ def pre_run(self):
...
+                raise ValueError("The entry point value does not match with git worktree path.")
```
blag, line 292: Please include the entry point and the git worktree path so the user has some idea of what to do to correct the issue.

st2common/st2common/runners/base.py
```diff
@@ -375,8 +376,11 @@ def cleanup_git_worktree(self, worktree_path, pack_name, content_version):
         :rtype: ``bool``
         """
         # Safety check to make sure we don't remove something outside /tmp
-        assert(worktree_path.startswith('/tmp'))
-        assert(worktree_path.startswith('/tmp/%s' % (self.WORKTREE_DIRECTORY_PREFIX)))
+        if not (worktree_path.startswith('/tmp')):
+            raise ValueError('The worktree path does not match with /tmp.')
+        if not (worktree_path.startswith('/tmp/%s' % (self.WORKTREE_DIRECTORY_PREFIX))):
+            raise ValueError('The worktree path does not match with /tmp/"%s".' %
+                            (self.WORKTREE_DIRECTORY_PREFIX))
```
blag, line 383: Same issues here.

st2common/st2common/util/crypto.py
```diff
@@ -80,7 +80,8 @@
 
 DEFAULT_AES_KEY_SIZE = 256
 
-assert DEFAULT_AES_KEY_SIZE >= MINIMUM_AES_KEY_SIZE
+if not DEFAULT_AES_KEY_SIZE >= MINIMUM_AES_KEY_SIZE:
+    raise ValueError('Verify the key size :"%s".' % (DEFAULT_AES_KEY_SIZE))
```
blag, line 84: Needs better comparison, and an error message that indicates that the AES key size is smaller than the minimum key size.

st2common/st2common/util/crypto.py
```diff
@@ -206,15 +207,18 @@ def cryptography_symmetric_encrypt(encrypt_key, plaintext):
...
+    if not isinstance(plaintext, (six.text_type, six.string_types, six.binary_type)):
+        raise TypeError('Plaintext needs to either be a string/unicode or bytes.')
 
     aes_key_bytes = encrypt_key.aes_key_bytes
     hmac_key_bytes = encrypt_key.hmac_key_bytes
 
-    assert isinstance(aes_key_bytes, six.binary_type)
-    assert isinstance(hmac_key_bytes, six.binary_type)
+    if not isinstance(aes_key_bytes, six.binary_type):
+        raise TypeError('AESKey bytes does not match with binary type.')
```
blag, line 219:
```suggestion
        raise TypeError(f'AESKey is not bytes (it is {type(aes_key_bytes)}).')
```

st2common/st2common/util/crypto.py
```diff
@@ -206,15 +207,18 @@ def cryptography_symmetric_encrypt(encrypt_key, plaintext):
...
+    if not isinstance(hmac_key_bytes, six.binary_type):
+        raise TypeError('HMACKey bytes does not match with binary type.')
```
blag, line 221:
```suggestion
        raise TypeError(f'HMACKey is not bytes (it is {type(hmac_key_bytes)}).')
```

st2common/st2common/util/crypto.py
```diff
@@ -263,15 +267,18 @@ def cryptography_symmetric_decrypt(decrypt_key, ciphertext):
...
 
     aes_key_bytes = decrypt_key.aes_key_bytes
     hmac_key_bytes = decrypt_key.hmac_key_bytes
 
-    assert isinstance(aes_key_bytes, six.binary_type)
-    assert isinstance(hmac_key_bytes, six.binary_type)
+    if not isinstance(aes_key_bytes, six.binary_type):
+        raise TypeError('The aes key bytes does not match with binary type.')
+    if not isinstance(hmac_key_bytes, six.binary_type):
+        raise TypeError('The hmac key bytes does not match with binary type.')
```
blag, line 281: Same issues here - binary type can just be bytes, and include what type was given.

st2common/st2common/util/green/shell.py
```diff
@@ -91,7 +91,8 @@ def run_command(cmd, stdin=None, stdout=subprocess.PIPE, stderr=subprocess.PIPE,
     """
     LOG.debug('Entering st2common.util.green.run_command.')
 
-    assert isinstance(cmd, (list, tuple) + six.string_types)
+    if not isinstance(cmd, (list, tuple) + six.string_types):
+        raise TypeError('Command must be a type of list, tuple and string.')
```
blag, line 95:
```suggestion
        raise TypeError(f"Command must be a type of list, tuple, or string, not '{type(cmd)}'.")
```

st2common/st2common/util/jsonify.py
```diff
@@ -73,8 +73,8 @@ def json_loads(obj, keys=None):
     for key in keys:
         try:
             obj[key] = json.loads(obj[key])
-        except:
-            pass
+        except Exception as error:
+            print('Criteria pattern not valid JSON: "%s".', error)
```
blag, line 77:
```suggestion
            print(f'Criteria pattern not valid JSON: "{error}".')
```

st2common/st2common/util/sandboxing.py
```diff
@@ -117,7 +110,8 @@ def get_sandbox_python_path(inherit_from_parent=True, inherit_parent_virtualenv=
         site_packages_dir = get_python_lib()
 
         sys_prefix = os.path.abspath(sys.prefix)
-        assert sys_prefix in site_packages_dir
+        if sys_prefix not in site_packages_dir:
+            raise ValueError('The file with prefix sys is not found in the directory.')
```
blag, line 114: Please include the values in this error message.

st2common/st2common/util/shell.py
```diff
@@ -74,7 +74,8 @@ def run_command(cmd, stdin=None, stdout=subprocess.PIPE, stderr=subprocess.PIPE,
 
     :rtype: ``tuple`` (exit_code, stdout, stderr)
     """
-    assert isinstance(cmd, (list, tuple) + six.string_types)
+    if not isinstance(cmd, (list, tuple) + six.string_types):
+        raise TypeError('Command must be a type of list, tuple and string.')
```
blag, line 78:
```suggestion
        raise TypeError(f"Command must be a type of list, tuple, or string, not '{type(cmd)}'.")
```

st2reactor/st2reactor/garbage_collector/base.py
```diff
@@ -196,7 +196,8 @@ def _purge_action_executions(self):
         timestamp_str = isotime.format(dt=timestamp)
         LOG.info('Deleting action executions older than: %s' % (timestamp_str))
 
-        assert timestamp < utc_now
+        if not timestamp < utc_now:
```
blag, line 199: Weird comparison here.

```suggestion
        if timestamp > utc_now:
```

st2reactor/st2reactor/garbage_collector/base.py
```diff
@@ -196,7 +196,8 @@ def _purge_action_executions(self):
...
+            raise ValueError('Calculated timestamp violates the utc now.')
```
blag, line 200: This should be worded better.

```suggestion
            raise ValueError(f'Calculated timestamp ({timestamp}) is later than now in UTC ({utc_now}).')
```

st2reactor/st2reactor/garbage_collector/base.py
```diff
@@ -216,7 +217,8 @@ def _purge_action_executions_output(self):
         timestamp_str = isotime.format(dt=timestamp)
         LOG.info('Deleting action executions output objects older than: %s' % (timestamp_str))
 
-        assert timestamp < utc_now
+        if timestamp < utc_now:
+            raise ValueError('Calculated timestamp violates the utc now.')
```
blag, line 221: Bad comparison here, it's doing the exact opposite of what the previous assertion was doing.

The error message also needs to be tweaked - see my previous comment.

One of these is likely causing the CircleCI failures.

st2reactor/st2reactor/garbage_collector/base.py
```diff
@@ -239,7 +241,8 @@ def _purge_trigger_instances(self):
         timestamp_str = isotime.format(dt=timestamp)
         LOG.info('Deleting trigger instances older than: %s' % (timestamp_str))
 
-        assert timestamp < utc_now
+        if timestamp < utc_now:
+            raise ValueError('Calculated timestamp violates the utc now.')
```
blag, line 245: Bad comparison here, it's doing the exact opposite of what the previous assertion was doing.

The error message also needs to be tweaked - see my previous comment.

One of these is likely causing the CircleCI failures.

blag (changes requested):
This PR inadvertently changes the behavior of some of the code, and the error messages need improvement, but this is a good start to improving the codebase. Thank you!

contrib/linux/actions/checks/check_processes.py
```diff
@@ -52,8 +52,8 @@ def process(self, criteria):
                 cmdfh = open(self.procDir + "/" + p + "/cmdline")
                 cmd = cmdfh.readline()
                 pInfo[1] = cmd
-            except:
-                continue
+            except IOError:
+                print("Error: can't find file or read data.")
```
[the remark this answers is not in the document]
Kami: We should likely include original error message as well (aka ``str(e)``.

contrib/linux/sensors/file_watch_sensor.py
```diff
@@ -44,8 +44,8 @@ def cleanup(self):
 
             try:
                 self._tail.notifier.stop()
-            except Exception:
-                pass
+            except Exception as e:
+                print("Error: tail notifier is still running.", e)
```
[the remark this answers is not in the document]
Kami: Should also include original error message.

st2actions/st2actions/cmd/scheduler.py
```diff
@@ -109,7 +109,7 @@ def _run_scheduler():
             handler.shutdown()
             entrypoint.shutdown()
         except:
-            pass
+            LOG.exception('Unable to shutdown scheduler.')
```
Kami, line 112: This error was / is not fatal before so we shouldn't use log.exception which results in error log level. Should probably just use info log level + add original error message (or maybe even debug + exc_info=True).

m4dcoder:
Closing this PR since it's a duplicate of https://github.com/StackStorm/st2/pull/5163.

PR #5159, opened 2021-02-19
use yaml.safe_load instead of yaml.load

ashwini-orchestral:
It is not safe to call yaml.load with any data received from an untrusted source. It allows instantiation of arbitrary objects.  The function yaml.safe_load limits this ability to simple Python objects like integers or lists. For security reasons, yaml.safe_load is used.

### DIFF
--- a/st2common/st2common/util/spec_loader.py
+++ b/st2common/st2common/util/spec_loader.py
@@ -77,7 +77,7 @@ def load_spec(module_name, spec_file, allow_duplicate_keys=False):
 
     # 1. Check for duplicate keys
     if not allow_duplicate_keys:
-        yaml.load(spec_string, UniqueKeyLoader)
+        yaml.safe_load(spec_string)
 
     # 2. Generate actual spec
     spec = yaml.safe_load(spec_string)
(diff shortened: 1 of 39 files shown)


m4dcoder (changes requested):
@ashwini-orchestral Looks like you included changes from another PR/commit into this PR. It is hard to tell what changes relate to this PR. Can you clean this up? You should branch from master and then apply the changes.

st2common/st2common/util/spec_loader.py
```diff
@@ -77,7 +77,7 @@ def load_spec(module_name, spec_file, allow_duplicate_keys=False):
 
     # 1. Check for duplicate keys
     if not allow_duplicate_keys:
-        yaml.load(spec_string, UniqueKeyLoader)
```
Kami, line 80: This function is only used to load ``openapi.yaml.j2`` file which is controlled by us (developers), which means there should be no chance for "yaml type injection", but I do agree and still think it's better and safer to use ``yaml.safe_load``.

Kami:
@ashwini-orchestral As @m4dcoder said, please remove unrelated changes so we can review ``yaml.safe_load`` change in isolation.

Kami:
Superseded by #5162.

PR #5162, opened 2021-02-22
Refactor spec_loader to use yaml.load with SafeLoader

ashwini-orchestral:
It is not safe to call yaml.load with any data received from an untrusted source. It allows instantiation of arbitrary objects. The function yaml.safe_load limits this ability to simple Python objects like integers or lists. For security reasons, yaml.safe_load is used.


Kami (approved):
Just echoing my comment here (https://github.com/StackStorm/st2/pull/5159#discussion_r580630073), but otherwise, LGTM.

st2common/st2common/util/spec_loader.py
```diff
@@ -77,7 +77,7 @@ def load_spec(module_name, spec_file, allow_duplicate_keys=False):
 
     # 1. Check for duplicate keys
     if not allow_duplicate_keys:
-        yaml.load(spec_string, UniqueKeyLoader)
+        yaml.safe_load(spec_string)
```
Kami, line 80: I missed this earlier - we intentionally use ``yaml.load()`` and not ``yaml.safe_load()`` here since we use a custom loader which detects duplicate keys which ``yaml.safe_load()`` doesn't support.

So using safe_load would change and break the existing behavior and that's not what we want.

Per my comment on other PR - this function is only used to load trusted openapi.yaml file which is controlled by us so not using ``safe_load()`` should be fine.

I suggest reverting this change + adding a comment to that line of code clarifying why we don't use safe_load there.

Kami: EDIT: It looks like it's actually possible to achieve "detect duplicate keys" behavior with ``safe_load``, so we should do something like this instead - https://gist.github.com/danielpops/5a0726f2fb6288da749c4cd604276be8 (we of course can't just copy that code since it has no license which means we need to develop it ourselves or find similar snippet with an appropriate license).

Kami (changes requested):

st2common/st2common/util/spec_loader.py
blag: That gist looks super overcomplicated for what it's trying to achieve. I'm cooking a patch for PyYAML that should allow us to inherit from `SafeLoader` and customize it to disallow duplicate keys.

m4dcoder: @blag @Kami please let @ashwini-orchestral what she needs to do here. @blag if you are working on a patch, will you be including the change here in this PR?

blag: Two PRs to PyYAML to make this easier for us:
* yaml/pyyaml#502
* yaml/pyyaml#503

If one of those is accepted then it will allow this code to be much simpler.

Kami:
@m4dcoder Can you please add a small test case for it? Here is quick example of similar test case I added recently - https://github.com/StackStorm/st2/pull/4846/files#diff-4d4c6f566c2d72910eff03ac16f442387a53beb0aa7e9b0ad5944ec5ce602ab2R101.

Besides that, LGTM.

Kami (approved):
LGTM :+1:

st2common/tests/unit/test_spec_loader.py
```diff
@@ -0,0 +1,35 @@
...
+
+from __future__ import absolute_import
+
+import unittest2
+import yaml
+
+from st2common.util import spec_loader
+
+
+class SpecLoaderTest(unittest2.TestCase):
```
Kami, line 23: Also wouldn't hurt to add a test case which verifies SafeLoad is indeed used, but otherwise, LGTM.

cognifloyd:
Luckily, merging master into this PR should not have many conflicts. Cheers!

Also, we need to look at performance when we switch out `yaml.load` for `yaml.safe_load`. @m4dcoder just added CSafeLoader to orquesta, and ended up needing to pass CSafeLoader into yaml.load to get both the performance boost and the safety. See: https://github.com/StackStorm/orquesta/pull/230/files#diff-c2248ccb7408cc089d472f1dcb72a531c386cac3a9af7ad7de9bebc1d5910f3c (and the discussion on that PR as well).

PR #5163, opened 2021-02-22
Add error handler or log error instead of using "pass" on exception in st2

ashwini-orchestral:
Replace pass statement with explicit error logs. The pass/continue statement when run in production (non-debug) code will skip the error. We want to practice safe coding so exception handling is necessary.


contrib/linux/actions/checks/check_processes.py
```diff
@@ -52,7 +52,8 @@ def process(self, criteria):
                 cmdfh = open(self.procDir + "/" + p + "/cmdline")
                 cmd = cmdfh.readline()
                 pInfo[1] = cmd
-            except:
+            except IOError:
```
m4dcoder, line 55: Why specifically IOError? It is possible for other exception to occur in this block (i.e. index error, etc.) and this will skip the exception.

contrib/linux/sensors/file_watch_sensor.py
```diff
@@ -44,8 +44,8 @@ def cleanup(self):
 
             try:
                 self._tail.notifier.stop()
-            except Exception:
-                pass
+            except Exception as e:
+                print(f"Error: tail notifier is still running. {e}")
```
m4dcoder, line 48: Can you do a `self._logger.exception("Unable to stop the tail notifier.")` instead of print? This should include the exception by default.

st2common/st2common/util/jsonify.py
```diff
@@ -73,8 +73,8 @@ def json_loads(obj, keys=None):
     for key in keys:
         try:
             obj[key] = json.loads(obj[key])
-        except:
-            pass
+        except Exception as e:
+            print(f'Criteria pattern not valid JSON: "{e}".')
```
m4dcoder, line 77: log instead of print?

st2reactor/st2reactor/container/sensor_wrapper.py
```diff
@@ -181,8 +181,9 @@ def __init__(self, pack, file_path, class_name, trigger_types,
         # 1. Parse the config with inherited parent args
         try:
             config.parse_args(args=self._parent_args)
-        except Exception:
-            pass
+        except Exception as e:
+            print('Failed to parse config using parent args (parent_args=%s): "%s".' %
+                  (str(self._parent_args), six.text_type(e)))
```
m4dcoder, line 186: log instead of print?

m4dcoder (changes requested):

contrib/linux/actions/checks/check_processes.py
```diff
@@ -52,7 +56,8 @@ def process(self, criteria):
                 cmdfh = open(self.procDir + "/" + p + "/cmdline")
                 cmd = cmdfh.readline()
                 pInfo[1] = cmd
-            except:
+            except Exception as e:
+                LOG.exception("Error: can't find file or read data. %s", e)
```
m4dcoder, line 60: Log exception already includes the exception. You don't need to reprint `e`.

st2common/st2common/util/jsonify.py
```diff
@@ -73,8 +76,8 @@ def json_loads(obj, keys=None):
     for key in keys:
         try:
             obj[key] = json.loads(obj[key])
-        except:
-            pass
+        except Exception as e:
+            LOG.exception('Criteria pattern not valid JSON : "%s"', e)
```
m4dcoder, line 80: Log exception already includes the exception object so there's no need to reprint `e`.

st2reactor/st2reactor/container/sensor_wrapper.py
m4dcoder: See comments above on log exception.

st2actions/st2actions/cmd/scheduler.py
```diff
@@ -109,7 +109,7 @@ def _run_scheduler():
             handler.shutdown()
             entrypoint.shutdown()
         except:
-            pass
+            LOG.exception('Unable to shutdown scheduler.')
```
m4dcoder, line 112: Please see comment in the old PR https://github.com/StackStorm/st2/pull/5158 to address concerns from reviewer.

m4dcoder (changes requested):
Not quite there yet. Please review the old PR https://github.com/StackStorm/st2/pull/5158 for any concerns that haven't been addressed yet.

contrib/linux/sensors/file_watch_sensor.py
```diff
@@ -44,8 +44,8 @@ def cleanup(self):
 
             try:
                 self._tail.notifier.stop()
-            except Exception:
-                pass
+            except Exception as e:
+                self._logger.exception("Unable to stop the tail notifier : '%s'", e)
```
m4dcoder, line 48: Since you're using logger.exception, you don't need to pass e in the text message.

m4dcoder (changes requested):

m4dcoder (approved):

cognifloyd:
@ashwini-orchestral you got the short end of the stick. Sorry about that. This one needs to be updated to use black formatting too.

st2common/st2common/util/jsonify.py
```diff
@@ -67,8 +70,8 @@ def json_loads(obj, keys=None):
     for key in keys:
         try:
             obj[key] = json.loads(obj[key])
-        except:
-            pass
+        except Exception:
+            LOG.exception("Criteria pattern not valid JSON.")
```
Kami, line 74: I noticed this issue while testing latest dev build - this exception is not fatal and we shouldn't log anything (or use debug log level) since it will cause unnecessary noise, churn and confusion in the action runner.

The method is named ``try_loads`` and it uses best-effort approach where dictionary item not being JSON  does not represent an error.

I will make that change in another PR - remove log statement and add a comment on why we don't log anything (error is not fatal).

/cc @m4dcoder

Kami: b2ed03b54

PR #5164, opened 2021-02-24
Update keyvalue.py to be able to add new values without leaking them in the shell history.

wingiti:
Add possibility to add a value by using a prompt instead of a parameter. This prevents secrets from being documented/leaked in the shell history.


st2client/st2client/commands/keyvalue.py
```diff
@@ -188,10 +188,14 @@ def run(self, args, **kwargs):
         instance = KeyValuePair()
         instance.id = args.name  # TODO: refactor and get rid of id
         instance.name = args.name
-        instance.value = args.value
         instance.scope = args.scope
         instance.user = args.user
 
+        if not args.value:
+            instance.value = input("Please insert value for key: ")
```
Kami, line 195: There is also ``getpass.getpass()`` which displays entered value masked on the screen.

But if we go with that approach, we will likely need to ask for confirmation (aka input the secret value twice).

cognifloyd:
Please merge master into your branch, resolve the conflicts (Ouch! Sorry!), and reformat with black (I recommend running pre-commit install after you've merged master so that black reformatting happens automatically on commit).

armab:
The tests look :green_heart: again.

I expect @Kami to review it again, but @wingiti can you please also add the Changelog.rst record for this PR? We'll need it anyways.

Thanks!

st2client/st2client/commands/keyvalue.py
cognifloyd: In a future PR, I think we could add `getpass.getpass()` support, but only turn it on if encrypt=True. I think this PR is good enough to merge as is.

cognifloyd (approved):

wingiti:
@Kami do you want to approve this? Would be nice, otherwise I have to overwrite this all the time after a Stackstorm upgrade in my local installations.

amanda11 (commented):
@wingiti The lint checks are failing as there is trailing space on the CHANGELOG.rst.

Could you amend, and check if you have signed the CLA (there was a new version issued, so you may have to re-sign).

Many thanks.

PR #5165, opened 2021-02-24
[BUGFIX] Use pip 20.0.2 to build virtualenvs

cognifloyd:
We have already been using `pip==20.0.2` for testing since st2 v3.2.0. `virtualenv` embeds the wheel for `pip`, so we need to adjust the pinned version of virtualenv to update which version of pip is used to create virtualenvs (eg for packs).

This also clarifies the comment about why we're pinning virtualenv instead of pip directly.

### DIFF
--- a/fixed-requirements.txt
+++ b/fixed-requirements.txt
@@ -46,8 +46,9 @@ six==1.13.0
 sseclient-py==1.7
 stevedore==1.30.1
 tooz==2.8.0
-# Note: We use latest version of virtualenv which uses pip 19
-virtualenv==16.6.0
+# Note: virtualenv embeds the pip wheel, so pinning virtualenv also pins pip
+# virtualenv==20.0.18 has pip==20.0.2
+virtualenv==20.0.18
 webob==1.8.5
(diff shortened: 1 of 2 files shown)


cognifloyd:
Rebased on master. Let's see if tests pass now.

Kami (approved):
LGTM :+1:

fixed-requirements.txt
```diff
@@ -46,8 +46,9 @@ six==1.13.0
 sseclient-py==1.7
 stevedore==1.30.1
 tooz==2.8.0
-# Note: We use latest version of virtualenv which uses pip 19
-virtualenv==16.6.0
+# Note: virtualenv embeds the pip wheel, so pinning virtualenv also pins pip
+# virtualenv==20.0.18 has pip==20.0.2
+virtualenv==20.0.18
```
armab, line 51: :+1:

armab (commented):
Nice one :+1: 

@cognifloyd or @Kami can you also compose a follow-up PR to document this update in the Changelog?

Very important.

PR #5167, opened 2021-02-25
Add support to force recreation of packs virtualenvs

winem:
This PR adds support for `--register-setup-virtualenvs-force-recreation` to st2ctl. Use-case is, like we have it now with the upgrade to st2 v3.4, an upgrade of the Python version inside the virtual environments for example. 

I'll run some tests now but wanted to provide the PR already so that others who ran into this issue can give it a try, too.


st2common/bin/st2ctl
```diff
@@ -43,19 +43,20 @@ function print_usage() {
...
+    echo "  --register-triggers                           Register all triggers."
+    echo "  --register-sensors                            Register all sensors."
+    echo "  --register-rules                              Register all rules."
+    echo "  --register-runners                            Register all runners."
+    echo "  --register-actions                            Register all actions."
+    echo "  --register-aliases                            Register all aliases."
+    echo "  --register-policies                           Register all policies."
+    echo "  --register-configs                            Register all configuration files."
+    echo "  --register-setup-virtualenvs                  Create Python virtual environments for all the registered packs."
+    echo "  --register-setup-virtualenvs-force-recreation (Delete and re-)create Python virtual environments for all the registered packs."
```
armab, line 56: Thinking about the shorter version, how this would sound instead? :thinking: 

```diff
---register-setup-virtualenvs-force-recreation
+--register-setup-recreate-virtualenvs
```

winem: I like it especially because it's more readable. Done & pushed. 

I'll have to go to sleep now and will proceed with the testing tomorrow as well as looking into failing tests (if any).

m4dcoder:
Can we rename options to `--setup-venvs` or `--setup-venvs-reset`?  Does venvs need registering, is that why we have to prefix it with register?

st2common/bin/st2ctl
cognifloyd: OOH! The bike shed needs another coat of paint :trollface: 

Let's shorten that even farther:
```diff
-  --register-setup-virtualenvs-force-recreation
+  --register-recreate-virtualenvs
```

cognifloyd: I'm resisting the recommendation to drop "register", even though re/creating virtualenvs doesn't actually "register anything. Alas, keeping "register" keeps the args consistent.

st2common/bin/st2ctl
```diff
@@ -140,7 +141,7 @@ function reopen_component_log_files() {
 }
 
 function register_content() {
-  ALLOWED_REGISTER_FLAGS='--register-all --register-actions --register-aliases --register-runners --register-policies --register-rules --register-sensors --register-triggers --register-configs --register-setup-virtualenvs --register-fail-on-failure --register-no-fail-on-failure --verbose'
+  ALLOWED_REGISTER_FLAGS='--register-all --register-actions --register-aliases --register-runners --register-policies --register-rules --register-sensors --register-triggers --register-configs --register-setup-virtualenvs --register-setup-recreate-virtualenvs --register-fail-on-failure --register-no-fail-on-failure --verbose'
```
amanda11, line 144: Does the new option need to get added to the sudo flags?

st2common/st2common/content/bootstrap.py
```diff
@@ -83,7 +83,7 @@ def register_opts():
 register_opts()
 
 
-def setup_virtualenvs():
+def setup_virtualenvs(recreate_virtualenvs=False):
```
amanda11, line 86: I think the tests might be failing because thee isn't a setup-recreate-virtualenvs in the cfg.BoolOpt like there is above for setup-virtualenvs.

amanda11 (changes requested):
Thanks for PR.

The tests are failing, because I think of a missing code change to register the new BoolOpt on the cfg, I've highlighted where I think it needs to go.

I also failed to run the command on upgrade using the artifact from this PR, which I thnk is same problem. 

Also we probably should add a new test into ./st2common/tests/integration/test_register_content_script.py for the new parameter, similar to some of the ones already there for the current setup virtualenv.

armab:
The context is that a new flag is a nice to have addition that just mimicks the following construction, required for 3.4 upgrade only:
```
sudo rm -rf /opt/stackstorm/virtualenvs/*
sudo st2ctl reload --register-setup-virtualenvs
```
@m4dcoder I think at this point we're just adding a new flag instead of refactoring existing params which would be a breaking change and undesired as a last-minute release enhancement.

st2common/bin/st2ctl
Kami: Yeah, I like ``--register-recreate-virtualenvs`` since it's the shortest :D

IIRC, register prefix was added at some point in the past due to some oslo config limitations or similar, but I forget the details.

st2common/bin/st2ctl
Kami: Yeah, it does need to run using sudo.

st2common/bin/st2ctl
winem: Yes, there seems to be a limitation that requires the --register-xxx prefix. I can dig deeper into if needed but will focus on the failing tests and actual issues for now.

winem:
> The context is that a new flag is a nice to have addition that just mimicks the following construction, required for 3.4 upgrade only:
> 
> ```
> sudo rm -rf /opt/stackstorm/virtualenvs/*
> sudo st2ctl reload --register-setup-virtualenvs
> ```
> 
> @m4dcoder I think at this point refactoring existing params is undesired as a last-minute release enhancement and would be an overkill breaking change.

@armab we don't change any defaults so I don't see how this is could be a breaking change. Do I miss something?

cognifloyd:
@winem I think @armab was referring to @m4dcoder's request to rename the options:

> Can we rename options to `--setup-venvs` (replaces `--register-setup-virtualenvs`) and `--reset-venvs` (replaces `--register-setup-virtualenvs-force-recreation`)? Does venvs need registering, is that why we have to prefix it with register?

Renaming the options would be a breaking change. Merely adding a new option as this PR does, is not a breaking change, IMO.

Kami:
Yeah, I agree with @cognifloyd.

In the future, I maybe be OK with renaming it (if we can work around oslo.config limitations), but for this last minute change I'm for ``--register-recreate-virtualenvs`` since it's the shortest and consistent with the current naming scheme.

blag:
Try pinning `mail-parser` to `>=3.9.1,<3.10.0` in `contrib/core/requirements-tests.txt` and see if that fixes it.

That package just had a release like 4 hours ago:
https://pypi.org/project/mail-parser/#history

And they might have broken something we were relying on, or fixed a bug that we were relying on.

winem:
Oh saw that you just pinned it. So let's see  if that fixes the issue. 

@cognifloyd thanks for clarifying it. Now it makes sense.

amanda11 (approved):

amanda11:
Signed off my review comments, I've not had chance to try on an upgrade to check command works - but didn't want my comments to be a blocker for merging.

blag:
Travis needs payment. I'll be out for an hour and fix it after that.

winem:
@amanda11 your input was actually very helpful and is always appreciated. And I added an integration test. :)

blag:
Paid off Travis-CI, adding a commit to kick CI again.

armab:
:tada:

PR #5169, opened 2021-02-27
Fix path for pip dir which is used for CI cache to speed up the build

Kami:
I noticed "requirements" step takes ~80-90 seconds on CI.

After inspecting the output, I noticed it's not using cached pip dependencies, but downloading all of them from the internet. It turns out, we were not using a correct path for the pip cache directory (``~/.cache/pip``).

This pull request fixes this.

With this change the step now takes ~36 seconds (if cache is populated / unchanged).

Before:

Screenshot from 2021-02-27 22-26-23

After:

Screenshot from 2021-02-27 22-26-10

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -113,7 +113,7 @@ jobs:
       - uses: actions/cache@v2
         with:
           path: |
-            .cache/pip
+            ~/.cache/pip
             virtualenv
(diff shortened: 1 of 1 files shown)


Kami:
Just FYI - various code performance improvements in #4846 should also result in a couple of minutes faster unit tests task run on CI.

nmaludy (approved):

PR #5177, opened 2021-03-03
Python3.8 on Ubuntu 20.04 (requiring MongoDB 4.4)

nzlosh:
### Work in Progress
Add support for Python3.8 on Ubuntu 20.04.  To support this configuration, it implicitly includes MongoDB 4.4 unless we consider standalone instances of StackStorm can no longer host the mongodb instance locally for Ubuntu 20.04.

 - Bump eventlet/greenlet to latest to benefit from threading fix for python3.7+
 - Bump pymongo/mongoengine to support Mongo4.4 which is the only version packaged for Ubuntu 20.04.
 - Added Python3.8 into tox tests.
 
 https://github.com/StackStorm/discussions/issues/68


cognifloyd:
This is just a note to let you know why it is going to be more work than normal when you merge master into this branch. Sorry!

winem:
I'll use this PR to drop some findings and topics we have to address when I find them: 
- a GET on /v1/executions/123123123123123123123123/attribute/result returns an internal server error if the execution does not exist: 
```
{
}
```
It works fine for existing executions.

Kami:
Just to confirm - new version of pymongo still supports older versions of MongoDB, right?

On a related note, we should also add some upgrade notes to st2docs on how to upgrade MongoDB for existing installations.

nzlosh:
@Kami According to the pymongo documentation 3.11 supports mongodb from 3.6 to 4.4 https://docs.mongodb.com/drivers/pymongo/#mongodb-compatibility

Kami:
Just a general question on how are we going to approach testing things on Github actions:

* Do we plan to run all the unit + integration tests on 3.6 and  3.8?
* How about MongoDB? Do we plan to test against multiple versions?
* How about end to end tests + installer script? That also needs to be updated to support Ubuntu 20.04 + Python 3.8 so we can be sure all the end to end tests indeed pass under Python 3.8

In general, I do think we should test with all the versions we claim support, but that's going to inflate the build matrix a lot and slow down the build so we should take that into consideration.

armab:
The first and I guess the most significant step here was making sure st2 can run and is tested under python v3.8 (unit + integration). Congrats @nzlosh and @winem with that achievement!

Under that py 3.8 testing environment, I think it makes sense to set up the MongoDB 4.4 leaving everything else as before.

After merging Python 3.8 support for st2 (this PR) folks could then start working on Ubuntu 20.04 LTS packages (py3.8) and then the installer script. The e2e tests may come later after adding U20 platform. Once it's done, we can get rid of all the U16 legacy.

Looks like @nzlosh highlighted the plan here perfectly: https://github.com/StackStorm/discussions/issues/68 :v:

armab:
@Kami With all the complex matrix and number of jobs, any chance we could reduce the number of parallel runs from `3` to maybe `2`? `18` total jobs is really too long list and we have to target 10mins anyway to match the CircleCI timing. Looks like we're already fast here with `8mins`:
Use_python3 8_for_pre_commit_hooks _·_StackStorm_st2@72d5ed9_-_2021-04-02_14 06 30

Kami:
@armab Another option may be to only run tests on Python 3.8 (in addition to 3.6) on master run.

contrib/linux/actions/service.py
```diff
@@ -18,7 +18,7 @@
 import re
 import sys
 import os
-import platform
+import distro as platform
```
Kami, line 21: IIRC, that won't work since that ``service.py`` file is used for remote shell runner which means we copy file to a remote host first where it's executed (and there ``distro`` package may not be available).

In short - Python script which use ``remote-shell-script`` runner can't rely on any 3rd party dependencies since they run using system Python and not using a pack virtual environment.

/cc @blag

st2common/st2common/util/monkey_patch.py
```diff
@@ -31,7 +31,7 @@
 PARENT_ARGS_FLAG = "--parent-args="
 
 
-def monkey_patch(patch_thread=None):
+def monkey_patch(patch_thread=False):
```
Kami, line 34: Just curious how what ``patch_thread=False`` affects things?

I assume it will cause issues with blocking the event loop if we have some code which utilizes threading module.

tox.ini
```diff
@@ -90,3 +90,76 @@ commands =
     nosetests --rednose --immediate -sv contrib/runners/orquesta_runner/tests/integration/
     nosetests --rednose --immediate -sv st2tests/integration/orquesta/
     nosetests --rednose --immediate -sv contrib/runners/python_runner/tests/integration/
+
+# Python 3.8 tasks
```
Kami, line 94: IIRC, we don't use tox for anything anymore. 

If that's indeed the case, we should probably just remove tox.ini to reduce confusion and amount of code we need to maintain.

IIRC, we used it in the past when we still supported Python 2 for running tests under Python 3.


What do other ppl think?

contrib/linux/actions/service.py
Kami: It seems like one, probably not as robust but probably good alternative would be to just shell out to ``lsb_release`` in this module - https://github.com/mp15/irods/commit/fd8af0c9154d2fa7308c0562c18b2c236e748511.

Kami:
Before / soon after this is merged we will also need to update ``st2vagrantdev`` to install and use Python 3.8 otherwise things won't work correctly (pre commit hooks, etc.).

tox.ini
Kami: I just pushed a change which removes it - d345d6bb6.

If we ever plan to utilize it again in the future (e.g. if / when we switch to pytest), we can add it back then. Until that there is point for a code which is not being utilized to be left in the repo :)

contrib/linux/actions/service.py
Kami: This should do it - b6cb22d2f0184b621c123e94dda28a324eefb7ac.

Kami:
I also pushed a change to make pre commit hook Python version agnostic - aka it will just try to use whatever Python interpreter is available.

Kami:
Tests seem to be failing quite often with:

```bash
  error pulling image configuration: Get https://registry-1.docker.io/v2/bitnami/rabbitmq/blobs/sha256:27f32b8c06acec740cc67ac2f5d711d9b389944b1809a4df5205cf5b42e2de85: EOF
  Warning: Docker pull failed with exit code 1, back off 9.64 seconds before retry.
  /usr/bin/docker pull bitnami/rabbitmq:3.8
  Error response from daemon: Get https://registry-1.docker.io/v2/: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```

May be related to GHA / Docker Registry issues or it could also be related to us trying to run too many parallel builds at once which doesn't play along well with Docker Hub rate limits or similar.

Kami:
Just to recap my comment again from slack.

I don't think we can go with not monkey patching ``threading`` module.

If we do that and there is any code which utilizes threading (which I'm sure there is even though we don't do that directly in out code base since we have tons of dependencies), it will block the eventlet event loop which is bad (it will cause potential ordering issue, things to freeze, code not to run, etc.)

Sadly issues like that also won't be detected by our unit or integration tests since they weren't designed to detect those issues, but to run functional checks.

Technically even if something is blocking the main loop (either a code which is not monkey patched or some long running code) that will usually only manifest itself as code taking a longer time to run or similar, but rarely in some functional test failing.

We could probably write some check to detect such issues, but it would probably not be trivial and likely also result in false positives from some "normal" code which is just CPU intensive and takes a while to run.

Perhaps we could start by enabling this (https://eventlet.net/doc/modules/debug.html#eventlet.debug.hub_blocking_detection) for end to end tests or similar.

nzlosh:
I've reverted the `thread=False` change and unit tests are now blocking on my local setup.  I'll try pushing the change to github to see if there is a different behaviour there since both @winem and I have different results when running `make unit-tests` locally compared to github unit tests.

The whole eventlet monkey patching being untested, yet being a core component, feels like we're flying blind.  It would be reassuring to have some form of testing for this.  `hub_blocking_detection` looks like a good option, something being better than nothing.

Kami:
@nzlosh Were you able to pin point it down to a specific test causing the hanging? Or it only happens when running tests in a specific order?

If we can pin point it down to a specific test or combination, that would be a good start.

nzlosh:
It blocks during the setup of `st2api`.  The last output is
```
===========================================================
Running tests in st2api
-----------------------------------------------------------
```
Attempting the break out of the tests produces an exception but can't cancel out of the running process.
```
^CException in thread pymongo_server_monitor_thread:
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/root/st2/virtualenv/lib/python3.8/site-packages/pymongo/periodic_executor.py", line 125, in _run
    if not self._target():
  File "/root/st2/virtualenv/lib/python3.8/site-packages/pymongo/monitor.py", line 56, in target
    monitor._run()
  File "/root/st2/virtualenv/lib/python3.8/site-packages/pymongo/monitor.py", line 180, in _run
    self._server_description = self._check_server()
  File "/root/st2/virtualenv/lib/python3.8/site-packages/pymongo/monitor.py", line 217, in _check_server
    return self._check_once()
  File "/root/st2/virtualenv/lib/python3.8/site-packages/pymongo/monitor.py", line 254, in _check_once
    response, round_trip_time = self._check_with_socket(sock_info)
  File "/root/st2/virtualenv/lib/python3.8/site-packages/pymongo/monitor.py", line 274, in _check_with_socket
    response = IsMaster(conn._next_reply(), awaitable=True)
  File "/root/st2/virtualenv/lib/python3.8/site-packages/pymongo/pool.py", line 605, in _next_reply
    reply = self.receive_message(None)
  File "/root/st2/virtualenv/lib/python3.8/site-packages/pymongo/pool.py", line 726, in receive_message
    self._raise_connection_failure(error)
  File "/root/st2/virtualenv/lib/python3.8/site-packages/pymongo/pool.py", line 724, in receive_message
    return receive_message(self, request_id, self.max_message_size)
  File "/root/st2/virtualenv/lib/python3.8/site-packages/pymongo/network.py", line 195, in receive_message
    _receive_data_on_socket(sock_info, 16, deadline))
  File "/root/st2/virtualenv/lib/python3.8/site-packages/pymongo/network.py", line 285, in _receive_data_on_socket
    wait_for_read(sock_info, deadline)
  File "/root/st2/virtualenv/lib/python3.8/site-packages/pymongo/network.py", line 244, in wait_for_read
    readable = sock_info.socket_checker.select(
  File "/root/st2/virtualenv/lib/python3.8/site-packages/pymongo/socket_checker.py", line 61, in select
    res = self._poller.poll(timeout_)
KeyboardInterrupt
```
Looking into the pymongo code, this is a mongo health check that tests the health of mongo at regular intervals.

Kami:
OK so it looks like it's indeed pymongo + eventlet + Python 3.8 related and will require more digging in.

Could be a bug in eventlet, pymongo or similar.

Kami:
If we plan to support two Python versions, another change we should probably make is to store virtualenv inside ``virtualenv-<py version>`` directory and update Makefile accordingly.

Otherwise it's a pain to test and work with multiple Python versions locally.

Kami:
OK, looks like I'm making progress. :tada: 

My suspicious from last time turned out to be correct - the tests hanging issue when patching ``thread`` indeed seem to be related to us not performing monkey patching early enough in the life cycle before pymongo imports various things.

It likely worked fine in the past and with older Python versions due to different import ordering or similar.

So hopefully just changing and re ordering the tests imports will do the trick. The downside / problem is that it may require some duplication, but sadly we don't have much choice.

I will see if I can get it to work without adding monkey_patch call to the top of every test file since that would not be so ideal and require a bunch of duplication.

Kami:
Alright, I pushed a change / workaround which I think should fix the tests hanging - 569bc80ab.

Per the code comment, we don't have all that many options. Perhaps the most robust one would be to add monkey patching at the top of every test file, but that would result in tons of duplication and hard to maintain the code. Because of that, I decided to go with a middle of the ground workaround.

Keep in mind that this issue doesn't affect the services itself because there the entry point / start up code is correctly and we perform monkey patching as one of the first things we do.

We still need to decide how to handle the CI though. Having that many jobs seems a bit messy. And we can't really combine them much / reduce number of workers either since that would slow the CI down.

Kami:
Good news is that it indeed looks like it fixes it for unit tests, but we will likely need a similar change for integration tests.

Kami:
Once all the tests pass, I will merge this into master.

Unit and integration tests passing should be good indicator of things (probably) working end to end, but it's also not a guarantee since we don't run end to end tests under Python 3.8 + MongoDB 4.4 yet.

Once we add new distro with those versions and run end to end tests on it, we will be able to see if all the end to end tests pass.

Kami:
OK, it looks like reducing 1 parallel job run increases aggregate test duration for ~2 minutes.

This is not great, especially since I spent many hours / days over the last weeks speeding up the tests.

Having said that, it may be OK compromise for now until we come up with some approach to make the build matrix more manageable (e.g. only run tests under Python 3.8 on master builds, part of a nightly job or similar).

Kami:
Also, now that we use newer version of pymongo and MongoDB, we could also experiment / benchmark with zstandard compression for wire transport.

Kami:
st2client install check started failing because a new version of ``importlib-metadata`` has been released (https://pypi.org/project/importlib-metadata/#history) and one of our transitive dependencies explicitly specified ``importlib-metadata < 4.0``.

Will push a change which pins it to < 4.0 and hopefully that will work.

Kami:
While at it, I also decided to add new ``database.compressors`` config option which allows users to specify compression algorithms client supports / should use for network / transport level compression - 61d5520eb124ecaf55aa18203827f3a51ac2d9f8.

I also added ``zstandard`` dependency which means setting this value to ``zstd`` should work out of the box as long as the server is running MongoDB >= 4.2.

By default (same as before), no transport / network level compression is used. I did run some quick micro benchmarks locally with and without zstd compression and there is no perceived difference in the run time so I decided to leave it disabled by default.

We should perform more larger scale benchmarks and then decide if it perhaps makes sense to enable it by default at some point in the future. I also plan to add some micro benchmarks for that in the future which will cover various scenarios and document / data sizes.

In the mean time, I will open st2docs change with release notes + config change which tells user how to configure this option in to be released StackStorm v3.5.

Kami (approved):

blag:
> If we plan to support two Python versions, another change we should probably make is to store virtualenv inside virtualenv-<py version> directory and update Makefile accordingly.

Aren't virtualenvs at least partially created like this already? `virtualenv/lib/python3.6/site-packages/...` Is it possible to use multiple Python versions with a single virtualenv?

Kami:
@blag There are still other issues, for example, files in ``bin/`` will be symlinks to a specific python version, same goes for virtualenv activation script.

Kami:
Re network / transport level compression - interesting thing is that the existing micro benchmarks showed very little difference in operation duration when using zstd for network level compression, but when running unit tests with compression on they do finish a bit faster.
As mentioned above, I plan to add some compression specific micro benchmarks in the future. Existing ones are not really geared towards benchmarking transport level compression, but they should at least serve as  good starting point to catch any regression in terms of performance with compression enabled.

Kami:
I pushed some unit test optimizations for low hanging fruit and managed to save a minute or so from unit test run.

Kami:
Re MongoDB 4.4 support.

It looks like that we might need to do more work. I got this error locally:

```
  File "/home/vagrant/st2/virtualenv/lib/python3.8/site-packages/pymongo/mongo_client.py", line 1360, in _cmd
    return server.run_operation_with_response(
  File "/home/vagrant/st2/virtualenv/lib/python3.8/site-packages/pymongo/server.py", line 135, in run_operation_with_response
    _check_command_response(first, sock_info.max_wire_version)
  File "/home/vagrant/st2/virtualenv/lib/python3.8/site-packages/pymongo/helpers.py", line 164, in _check_command_response
    raise OperationFailure(errmsg, code, response, max_wire_version)
pymongo.errors.OperationFailure: Path collision at context, full error: {'ok': 0.0, 'errmsg': 'Path collision at context', 'code': 31250, 'codeName': 'Location31250'}
```

Didn't have time to dig in and track it down yet. It looks like it's related to change in MongoDB and we may need to modify some queries or similar - https://docs.mongodb.com/manual/release-notes/4.4-compatibility/#path-collision-restrictions.

Kami:
From Slack:

It's an issue with "overlapping" projection query.

That's the projection for that query (defualt executions list API endpoint): ``['end_timestamp', 'runner.runner_parameters', 'action.ref', 'parameters', 'id', 'context', 'start_timestamp', 'action.pack', 'action.parameters', 'action.uid', 'status', 'context.user']``.

It should either be just ``context`` or ``context.user`` depending on what we want (all the context fields or just one). Previously in older versions in scenarios like that, all values from context would be returned.

In short:

1. We need to fix this project query and audit rest of them
2. Update our code to throw early and a better error in case there are "overlapping" projections
3. Add tests

PR #5182, opened 2021-03-07
Warn on service startup if a non utf-8 encoding is used

Kami:
This pull request updates service start up code to log a warning if non utf-8 locale / encoding is used for StackStorm service process.

## Background, Context

As encountered recently (and bubbled up by the Xenial end to end tests), using a non utf-8 locale when running StackStorm under Python 3 and working with unicode data can cause all kinds of issues.

There isn't tons we can do to work around those issues in code, but at the very least, we should emit a warning if such scenario is detected and warn the users.


st2common/st2common/service_setup.py
```diff
@@ -111,17 +125,54 @@ def setup(
...
-    LOG.debug("Using Python: %s (%s)" % (version, sys.executable))
+
+    # We print locale related info to make it easier to troubleshoot issues where locale is not
+    # set correctly (e.g. using C / ascii, but services are trying to work with unicode data
+    # would result in things blowing up)
+
+    fs_encoding = sys.getfilesystemencoding()
+    default_encoding = sys.getdefaultencoding()
+    lang_env = os.environ.get("LANG", "notset")
+    pythonipencoding_env = os.environ.get("PYTHONIOENCODING", "notset")
```
cognifloyd, line 136:
```suggestion
    pythonioencoding_env = os.environ.get("PYTHONIOENCODING", "notset")
```

st2common/st2common/service_setup.py
```diff
@@ -111,17 +125,54 @@ def setup(
...
+            used_locale = "unable to retrieve locale"
+    except Exception as e:
+        language_code, encoding = "unknown", "unknown"
+        used_locale = "unable to retrieve locale: %s " % (str(e))
+
+    LOG.info("Using Python: %s (%s)" % (version, sys.executable))
+    LOG.info(
+        "Using fs encoding: %s, default encoding: %s, locale: %s, LANG env variable: %s, "
+        "PYTHONIOENCODING env variable: %s"
+        % (fs_encoding, default_encoding, lang_env, used_locale, pythonipencoding_env)
```
cognifloyd, line 153:
```suggestion
        % (fs_encoding, default_encoding, lang_env, used_locale, pythonioencoding_env)
```

cognifloyd (approved):
Var naming nit.

cognifloyd:
Kick the travis CI job. There's a race condition in some of the orquesta tests. It's looking for status "canceling" but it is already "canceled" at that point. Re-run and it'll probably pass.

PR #5183, opened 2021-03-08
Added web security headers for nginx configuration

ashwini-orchestral:
Added a few web security headers in the Nginx configuration.
For example - X-Frame-Options Header, Cache-control.
Disable server version information.


conf/nginx/st2.conf
```diff
@@ -135,5 +143,6 @@ server {
     sendfile on;
     tcp_nopush on;
     tcp_nodelay on;
+    server_tokens off;
```
armab, line 146: I think `server_tokens off;` would be sufficient for the `server` block and not needed to repeat for every possible `location`.
Can you try that?

conf/nginx/st2.conf
```diff
@@ -110,6 +116,7 @@ server {
     error_page 502 = @authError;
 
     rewrite ^/auth/(.*)  /$1 break;
+    server_tokens off;
```
armab, line 119: Sorry if I wasn't clear enough in my previous comment.

`server_tokens off;` would be sufficient for the `server` blocks and not needed to repeat for every possible `location` as they'll inherit it. This way we'll also avoid duplication.

Can you try that?

armab (changes requested):

conf/nginx/st2.conf
```diff
@@ -40,6 +40,10 @@ server {
 
   add_header              Front-End-Https on;
   add_header              X-Content-Type-Options nosniff;
+  add_header              X-Frame-Options DENY always;
+  add_header              Strict-Transport-Security "max-age=3153600;includeSubDomains";
+  add_header              Cache-Control "no-store; no-cache; max-age=0; must-revalidate;";
```
armab, line 45: While the security-focused headers are all good, I'm wondering what would be the user impact for the Cache-Control changes here? @ashwini-orchestral what are the specific reasons behind hardening cache-control this way?

WDYT @StackStorm/maintainers @StackStorm/contributors? Any downsides, any behavior/UX changes you could think of?

conf/nginx/st2.conf
armab: Thanks for the changes @ashwini-orchestral  :+1:

conf/nginx/st2.conf
Kami: Yeah, +1, I don't see much benefit security wise from cache control headers, it would just slow things down if we need to re-fetch all the static resources on each page (re)load...

But others (iframe, stric transport security, etc.) are indeed a good idea :+1: 

On that note, I believe we need to sync up multiple places (if they don't contain those headers already) - Docker, installer, etc. Or do those places use config from this repo?

Kami:
On a somewhat related note:

>   ssl_protocols             TLSv1 TLSv1.1 TLSv1.2;

TLS v1.0 and v1.1 are also both past it's prime time now so it would probably be a good idea to just support TLS v1.2 going forward.

I don't think it should negatively affect any installations (most everything supports TLS v1.2 these days), but we can document it in ugprade notes and on how to add support for TLS v1.1 back, if needed.

conf/nginx/st2.conf
armab: Right, we're pulling nginx config from the `st2` repo (this file).

armab:
@Kami Good point.
@nmaludy raised this before https://github.com/StackStorm/discussions/issues/44 and feels it's a good time to do it?

conf/nginx/st2.conf
armab: @ashwini-orchestral can you please remove the `Cache-Control` headers then?

Kami:
@armab Yeah, I think now is the right time.

I will also open docs and upgrade notes entry with that in case it affects someone, but I think that would be quite unlikely.

conf/nginx/st2.conf
```diff
@@ -24,12 +24,12 @@ server {
...
 server {
   listen       *:443 ssl;
-
+  server_tokens off;
   ssl_certificate           /etc/ssl/st2/st2.crt;
   ssl_certificate_key       /etc/ssl/st2/st2.key;
   ssl_session_cache         shared:SSL:10m;
   ssl_session_timeout       5m;
-  ssl_protocols             TLSv1 TLSv1.1 TLSv1.2;
+  ssl_protocols             TLSv1.2;
```
armab, line 32: cc @nmaludy @punkrokk

conf/nginx/st2.conf
armab, line 45:
```suggestion
```
@ashwini-orchestral can you please remove the `Cache-Control` headers per team feedback?

armab (changes requested):
@ashwini-orchestral Left one change request per team feedback and please also provide a Changelog record for this PR.

shital-orchestral:
Added web header settings for possible security issues, X-Frame-Options, Strict-Transport-Security, X-XSS-Protection and server-tokens.
Thanks,
shital-orchestral

m4dcoder (approved):
Per @ashwini-orchestral, she has tested this nginx configuration change on an st2 install and there's no obvious issue as a result. The nginx configuration gaps here were identified from an active vulnerability scan of st2. This change is required to harden st2 deployment.

m4dcoder:
@ashwini-orchestral Can you add an entry in the CHANGELOG?

CHANGELOG.rst
```diff
@@ -18,6 +18,15 @@ Changed
...
   Contributed by @Kami.
 
+* Default nginx config (``conf/nginx/st2.conf``) which is used by the installer and Docker
+  images has been updated to only support TLS v1.2 (support for TLS v1.0 and v1.1 has been
+  removed).
+
+  If someone still needs to support those versions, they can modify the nginx config
+  accordingly. #5183
+
+  Contributed by @Kami and @ashwini.
```
armab, line 28: It's not only about the TLS change.
Can we include a note about hardening nginx config with extra security headers?

That's where we started.

Kami (approved):

PR #5186, opened 2021-03-11
Use sudo -E to preserve env vars in CI

cognifloyd:
- make add-itest-user scripts more verbose
- use `sudo -E` to preserve env vars

Github Actions has this in their `Default: env_reset` in their sudoers config file. This means that HOME and other vars are set to match the sudo target user. Using `sudo -E` preserves the environment allowing `~` to resolve correctly.

I'm not sure why we didn't hit this earlier. Maybe Github changed their sudoers config? In any case, this should fix it.

### DIFF
--- a/.circle/add-itest-user.sh
+++ b/.circle/add-itest-user.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -e
+set -ex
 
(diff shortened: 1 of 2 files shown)


m4dcoder (approved):
LGTM

PR #5193, opened 2021-03-19
Fixed Apache range header DOS and proxy disclosure issue

ashwini-orchestral:
The default Nginx configuration can have many vulnerabilities, a fix for a few of them.
1. Disable unwanted HTTP methods - Allowing TRACE or DELETE is risky as it can allow a Cross-Site Tracking attack and potentially allow a hacker to steal the cookie information.
2. HTTP requests with a large byte range in the range header can trigger the crash so have to handle this by using this proxy_set_header Range "";

### DIFF
--- a/conf/nginx/st2.conf
+++ b/conf/nginx/st2.conf
@@ -15,7 +15,7 @@ server {
   if ($ssl_protocol = "") {
        return 308 https://$host$request_uri;
   }
-  if ($request_method !~ ^(GET|HEAD|POST|PUT|DELETE)$ ) {
+  if ($request_method !~ ^(GET|HEAD|POST|PUT|DELETE|OPTIONS)$ ) {
        return 405;
   }
   index  index.html;
@@ -28,7 +28,7 @@ server {
   listen       *:443 ssl;
   server_tokens off;
 
-  if ($request_method !~ ^(GET|HEAD|POST|PUT|DELETE)$ ) {
+  if ($request_method !~ ^(GET|HEAD|POST|PUT|DELETE|OPTIONS)$ ) {
        return 405;
   }
 
--- a/conf/nginx/st2.conf
+++ b/conf/nginx/st2.conf
@@ -15,7 +15,9 @@ server {
   if ($ssl_protocol = "") {
        return 308 https://$host$request_uri;
   }
-
+  if ($request_method !~ ^(GET|HEAD|POST)$ ) {
+       return 405;
+  }
   index  index.html;
 
   access_log /var/log/nginx/st2webui.access.log combined;
@@ -134,6 +136,8 @@ server {
   location / {
     root      /opt/stackstorm/static/webui/;
     index     index.html;
+
+    proxy_set_header Range "";
 
     sendfile on;
     tcp_nopush on;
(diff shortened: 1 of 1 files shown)


conf/nginx/st2.conf
```diff
@@ -15,7 +15,9 @@ server {
   if ($ssl_protocol = "") {
        return 308 https://$host$request_uri;
   }
-
+  if ($request_method !~ ^(GET|HEAD|POST)$ ) {
+       return 405;
+  }
```
m4dcoder, line 20: Is this section required below for the main API server on 443 ssl?

m4dcoder: If we also restrict this on the 443 server, note these
- There are API functons that use `PUT`. Ref: https://github.com/StackStorm/st2/blob/master/st2api/st2api/controllers/v1/actions.py#L156
- There are API functions that use `DELETE` to remove resources. How will this impact these functions? Ref: https://github.com/StackStorm/st2/blob/master/st2api/st2api/controllers/v1/actions.py#L208

conf/nginx/st2.conf
```diff
@@ -135,6 +137,8 @@ server {
     root      /opt/stackstorm/static/webui/;
     index     index.html;
 
+    proxy_set_header Range "";
```
m4dcoder, line 140: How about the other locations above (i.e. /auth/, /api/, etc.)?

m4dcoder (commented):
Got some questions. Please clarify. Thanks.

conf/nginx/st2.conf
Kami: Yeah, as @m4dcoder said this is likely untested and won't work - API operations also utilize PUT, POST, OPTIONS (for cors) operations.

conf/nginx/st2.conf
Kami: Is this more of a hypothetical / possible vulnerability or the code is actually vulnerable to this issue?

Because afaik, st2api doesn't handle ``Range`` header anywhere (so it should just be ignored). Or does it apply to nginx server static files?

conf/nginx/st2.conf
ashwini-orchestral: @m4dcoder - I have modified the code and the only TRACE will not be allowed.

conf/nginx/st2.conf
```diff
@@ -15,7 +15,9 @@ server {
   if ($ssl_protocol = "") {
        return 308 https://$host$request_uri;
   }
-
+  if ($request_method !~ ^(GET|HEAD|POST|PUT|DELETE)$ ) {
+       return 405;
+  }
```
armab, line 20:
```suggestion
```
Similar to what @m4dcoder was saying before, looks like the directive added is not effective for the HTTP server section here as it's just redirecting to the https one.

ashwini-orchestral: I have added a directive for HTTPS as well.
TRACE will be not be allowed in the HTTP method. Screenshots of testing for your reference.
trace_nginx_request
trace_nginx_response

I have added max_ranges 0 to disable the partial content responses on Nginx. If a range is set as a byte then it is more vulnerable. It is recommended to not set the range as a byte.

Kami: Thanks for the update.

As per my comment above - we also needs to support / whitelist ``OPTIONS`` that's used by st2web for CORS purposes.

Kami:
Thanks for the contribution.

I will add a changelog entry and merge it into master.

PR #5197, opened 2021-03-20
Implement coditional execution "result" field retrieval for the "/v1/executions/<id>" (get one execution) API endpoint

Kami:
As per my comment in #4846 (https://github.com/StackStorm/st2/pull/4846#issuecomment-803292658), I couldn't live with that one additional query on the st2web side to implement "only retrieve execution result if it's smaller than X" functionality even though that with the additional query overall end user experience and performance was already much, much better than previous status quo, but I'm kinda OCD and wanted to get rid of that additional API operation in st2web (https://github.com/StackStorm/st2web/pull/868).

So to do that, I implemented conditional "get one execution" API operation where the ``result`` field is only returned if it's smaller than the provided value.

Imo, this implementation is a win-win since it allow us to get rid of additional API operation on the st2web side, we can even simplify st2web code a lot and other clients can also utilize this new API operation / query parameter. In addition to that it's also a win-win for the server - in case client won't actually render the result if it's too large it also makes no sense for the server to return it (so this way we reduce load a bit on the server as well).

We also have fall back in st2web so it's fully backward compatible and even works with old executions which may not contain ``result_size`` field (in that case we calculate result size on the client which takes some time, but it's still very, very fast compared to actually rendering and displaying large results).

## Implementation Details

If you check the code and comments, you will see the operation is implemented by utilizing two MongoDB queries.

One to check if corresponding execution with result_size smaller than the specified threshold exists and one to then retrieve that execution (and only including the result field size is smaller than specified threshold).

I experimented with a couple of different approaches (always retrieve execution result and do filtering and result field removal in the Python code, etc.), but it turns out this two query approach is actually the fastest.

To some extent that's expected because the first query we run already performs first filtering on the primary key (object id) which is an index lookup and it's very fast and we also do a projection operation where only 2 tiny indexed fields are retrieved and returned.

I will make necessary changes to st2web PR (https://github.com/StackStorm/st2web/pull/868) to utilize this new query param filter shortly.


Kami:
Alright, I also updated the st2web changes in https://github.com/StackStorm/st2web/pull/868 (which are very simple now).

Kami:
@winem Thanks for the review.

I believe 0c80a0078 should address all the feedback.

mweinberg-cm:
LGTM now!

winem:
LGTM now!

winem (approved):

PR #5198, opened 2021-03-20
Speed up service start up time by reducing runner registration / extension loading time

Kami:
I was looking at st2api profiling data some more and noticed we spend more time in stevedore code then we should.

It turns on that we were not loading all the extensions just once, but X + 1 times where X is number of runners (stevedore extensions).

This is obviously not the fastest thing since loading extensions requires pkg resources scan for available extensions.

This pull request fixes that. It should now also result in much less DEBUG level noise on service start up time.

### DIFF
--- a/st2common/st2common/bootstrap/runnersregistrar.py
+++ b/st2common/st2common/bootstrap/runnersregistrar.py
@@ -42,7 +42,7 @@ def register_runners(experimental=False, fail_on_failure=True):
     manager = ExtensionManager(namespace=RUNNERS_NAMESPACE, invoke_on_load=False)
 
     # NOTE: We use ExtensionManager directly instead of DriverManager per extension since that is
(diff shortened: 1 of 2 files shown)


st2common/st2common/bootstrap/runnersregistrar.py
```diff
@@ -41,15 +40,13 @@ def register_runners(experimental=False, fail_on_failure=True):
     runner_count = 0
 
     manager = ExtensionManager(namespace=RUNNERS_NAMESPACE, invoke_on_load=False)
-    extension_names = manager.names()
 
-    for name in extension_names:
+    # NOTE: We use ExtensionManager directly instead of DriverManager per extension since that is
+    # much faster and allows us to reduce stevedore loading overead for each runner
```
amanda11, line 45: typo - overhead not overead

amanda11 (approved):
LGTM - just a minor typo

PR #5199, opened 2021-03-21
Add "--enable-profiler" flag to all the StackStorm services + CLI

Kami:
This pull request adds ``--enable-profiler`` flag to all the StackStorm services + CLI which can be used to profiler the code using cProf based profiler during development and troubleshooting.

Keep in mind that cProf offers only basic information and usually when drilling down (into a function) you need to use line profiler, but this may not working as expected our of the box due to our usage of eventlet (need to add some docs on that).


st2client/st2client/shell.py
```diff
@@ -65,7 +65,7 @@
 from st2client.utils.logging import LogLevelFilter, set_log_level_for_all_loggers
 from st2client.commands.auth import TokenCreateCommand
 from st2client.commands.auth import LoginCommand
-
+from st2client.utils.profiler import setup_regular_profiler
```
Kami, line 68: NOTE: st2client can't depend on st2common that's why we intentionally need to duplicate the code in this package.

blag:
Can we use a different profiler? Something like [Scalene](https://github.com/plasma-umass/scalene)?

Kami:
@blag We probably could for code which doesn't rely on eventlet.

For most rest of the code which relies on eventlet we probably can't use it (unless that profiler knows how to handle eventlet code and context switching).

cognifloyd (approved):

cognifloyd:
@Kami looks like all you have to do is re-sign the CLA (which we changed recently to say Linux Foundation instead of one of the previous StackStorm stewards).

PR #5200, opened 2021-03-22
Add sensor to list of schemas generated by st2-generate-schemas

m4dcoder:
The st2-generate-schemas is missing the sensor schema. Add SensorTypeAPI to the list. Update trigger_types of SensorTypeAPI to have items of type TriggerTypeAPI. These schemas will be used for validation of meta files during pack development. Add a new command line utility named `st2-validate-pack` to validate pack contents against schema. This command can be used by pack developers to validate pack during development.


contrib/schemas/sensor.json
```diff
@@ -0,0 +1,95 @@
+{
```
Kami, line 1: Can you please add ``README.md`` to ``contrib/schema/`` directory which describes what it contains and how to (re)generate those files using ``st2-generate-schemas``?

And let's please also add a note which says those files are auto-generated and should not be edited manually.

st2common/bin/st2-generate-schemas
```diff
@@ -26,13 +26,15 @@ from st2common.models.api import action as action_models
...
 content_models = {
     'pack': pack_models.PackAPI,
     'action': action_models.ActionAPI,
     'alias': action_models.ActionAliasAPI,
     'policy': policy_models.PolicyAPI,
     'rule': rule_models.RuleAPI,
+    'sensor': sensor_models.SensorTypeAPI,
 }
 
 def main():
```
Kami, line 40: I know that's an existing code, but still - https://github.com/StackStorm/st2/pull/5200/files#diff-2ac436102029c3245a263895b2bb3a5639f7d52c8396aefdeb0500deda58e23dR41.

Rather "interesting" way to get current file, ``__file_`` is much more commend and used elsewhere in the code base (``BASE_DIR = os.path.dirname(os.path.abspath(__file__))``).

Kami (approved):
As per my review comment, let's please add some docs to that directory, besides that, LGTM :+1:

st2common/bin/st2-generate-schemas
Kami: Another quick comment on an existing line - https://github.com/StackStorm/st2/blob/fcfed7c72afa728fc654bdd02a9a1623d7ed9434/st2common/bin/st2-generate-schemas#L48

I know ``name`` is controlled by us in this case, but as a general secure coding practice, we should still use ``os.path.join()`` to avoid potential undesired path manipulation (if the code ever gets changed so the name is an input or similar), so ``        schema_file = os.path.join(schemas_dir, "%s.json" % (name))``.

m4dcoder: Fixed.

contrib/schemas/sensor.json
m4dcoder: Added README. JSON files do not allow comments so the note to not edit manually is in the README.

st2common/bin/st2-generate-schemas
m4dcoder: Fixed.

Kami:
Thanks for adding the readme, etc. LGTM.

PR #5201, opened 2021-03-22
Upgrade paramiko

amanda11:
Problems reported with SSH keys when using core.remote to access remote systems. 

According to https://github.com/paramiko/paramiko/pull/1723 then older version of cryptography would re-compute key if not match, but newer version of cryptography requires it. Upgrade paramiko to the version that is compatible with newer cryptography.

From paramiko changelog: "[Bug] #1723: Fix incorrectly swapped order of p and q numbers when loading OpenSSH-format RSA private keys. At minimum this should address a slowdown when using such keys, and it also means Paramiko works with Cryptography 3.1 and above (which complains strenuously when this problem appears)."


Kami (approved):
Please add a changelog entry, besides that, LGTM.

blag:
CircleCI is failing: https://app.circleci.com/pipelines/github/StackStorm/st2/1318/workflows/42f00c84-12b6-4e80-a097-93bb492c7b58/jobs/12707

According to the [Quay.io Status page](https://status.quay.io/):

> ### Quay.io image pulls failing ###
> **Investigating** - Image pulls from quay.io are currently failing with the code "502 Bad Gateway". Engineers are currently investigating.
> Mar 22, 16:30 EDT

blag:
Is this something to pull into a potential 3.4.2 release?

cognifloyd (approved):

PR #5202, opened 2021-03-23
Enable ci-integration tests on GHA (incl fixing a few tests)

cognifloyd:
Move integration tests to GitHub Actions (GHA)

We also have to fix a few tests to make the transition.

List of commits:

- improve CI output
- improve GHA folding
- fix virtualenv lock files issue again
- Switch tests to check w/ CA that is present in GHA
- apt download cache
- use bitnami/rabbitmq service container in GHA
- bust python cache
- Use virtualenv from st2 virtualenv by default
- Installing virtualenv in --user on github
- Add workaround for failing stdin test
- fix check-dependency-conflicts by updating pip-tools

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -142,9 +142,12 @@ jobs:
           path: |
             ~/.cache/pip
             virtualenv
-          key: ${{ runner.os }}-${{ matrix.python-version }}-${{ hashFiles('requirements.txt', 'test-requirements.txt') }}
+          # TODO: maybe make the virtualenv a partial cache to exclude st2*?
+          # !virtualenv/lib/python*/site-packages/st2*
+          # !virtualenv/bin/st2*
+          key: ${{ runner.os }}-python-${{ matrix.python-version }}-${{ hashFiles('requirements.txt', 'test-requirements.txt') }}
           restore-keys: |
-            ${{ runner.os }}-${{ matrix.python }}-
+            ${{ runner.os }}-python-${{ matrix.python }}-
       - uses: actions/cache@v2
         with:
           path: |
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -114,6 +114,12 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: '${{ matrix.python-version }}'
+      - name: Get date components for use in cache-keys
+        id: date
+        run: |
+          echo "::set-output name=year::$(/bin/date -u "+%Y")"
+          echo "::set-output name=month::$(/bin/date -u "+%m")"
+          echo "::set-output name=week::$(/bin/date -u "+%U")"
       - uses: actions/cache@v2
         with:
           path: |
@@ -122,6 +128,16 @@ jobs:
           key: ${{ runner.os }}-${{ matrix.python-version }}-${{ hashFiles('requirements.txt', 'test-requirements.txt') }}
           restore-keys: |
             ${{ runner.os }}-${{ matrix.python }}-
+      - uses: actions/cache@v2
+        with:
+          path: |
+            /var/cache/apt/archives/*.deb
+            /var/cache/apt/archives/partial/*.deb
+            /var/cache/apt/*.bin
+          key: ${{ runner.os }}-apt-${{ steps.date.outputs.year }}-${{ steps.date.outputs.week }}
+          restore-keys: |
+            ${{ runner.os }}-apt-${{ steps.date.outputs.year }}-
+            ${{ runner.os }}-apt-
       - name: Install apt depedencies
         run: |
           # install dev dependencies for Python YAML and LDAP packages
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -44,28 +44,45 @@ jobs:
           # - name: 'Micro Benchmarks'
           #   task: 'micro-benchmarks'
           #   python-version: '3.6'
-          # Integration tests are not working yet, still done in Travis
-          # - name: 'Integration Tests'
-          #   task: 'ci-integration'
+          - name: 'Integration Tests'
+            task: 'ci-integration'
+            python-version: '3.6'
     services:
       mongo:
         image: mongo:4.0
         ports:
           - 27017:27017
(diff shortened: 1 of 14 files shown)


cognifloyd:
Historical note:
I built this in st2sandbox/st2#1 then squashed/reordered the 100+ commits into a few discreet commits in st2sandbox/st2#2. In squashing/reordering commits, I removed most of the history of places I added various forms of debug logging.

Once this is merged, a maintainer should disable StackStorm/st2 on travis-ci.com. A future PR can clean up the repo, removing travis config, and getting rid of anything else that is travis-specific.

cognifloyd:
failure in integration tests is one of the orquesta tests that have been flaky lately. Please restart GitHub Actions workflows.

cognifloyd:
Fixes a bug with the cache. All tests pass now.

.github/workflows/ci.yaml
```diff
@@ -149,34 +199,43 @@ jobs:
...
-          #   Travis-CI:      /home/travis/build/StackStorm/st2/st2tests/st2tests/fixtures/ssl_certs/ca/ca_certificate_bundle.pem
-          #   GitHub Actions: /home/runner/work/st2/st2/st2tests/st2tests/fixtures/ssl_certs/ca/ca_certificate_bundle.pem
-          sed -i 's|/home/travis/build/StackStorm|/home/runner/work/st2|g' scripts/travis/rabbitmq.config
-          # Now that we've manged the config file, install it
-          sudo cp scripts/travis/rabbitmq.config /etc/rabbitmq/rabbitmq.config
-          # Install rabbitmq_management RabbitMQ plugin
-          sudo service rabbitmq-server restart
-          sleep 5
-          sudo rabbitmq-plugins enable rabbitmq_management
+          sudo cp scripts/github/rabbitmq.conf /home/runner/rabbitmq_conf/custom.conf
```
Kami, line 211: In the future I think it would be better to move that code in ``scripts/ci/setup-rabbitmq.sh`` or similar so we can also run shell check, etc on those files.

tools/launchdev.sh
```diff
@@ -153,7 +154,9 @@ function st2start(){
         cp -Rp ./contrib/examples $PACKS_BASE_DIR
         # Clone st2tests in /tmp directory.
         pushd /tmp
-        git clone https://github.com/StackStorm/st2tests.git
+        echo Cloning https://github.com/StackStorm/st2tests.git
+        # -q = no progress reporting (better for CI). Errors will still print.
+        git clone -q https://github.com/StackStorm/st2tests.git
```
Kami, line 159: I guess we could even use ``--depth 1`` // swallow clone to speed things up, but then again that repo is not that bug.

scripts/travis/prepare-integration.sh
```diff
@@ -16,6 +16,9 @@ st2 --version
 # Clean up old screen log files
 rm -f logs/screen-*.log
 
+# ::group::/::endgroup:: is helpful github actions syntax to fold this section.
+echo ::group::launchdev.sh start -x
```
Kami, line 20: As per comment above since that script is now specific to gha, we should probably rename the directory and update the affected files :)

Kami:
Thanks for doing this, it's great to how all the tests now run on GHA.

As far as orquesta test failures go - yeah, it appears that some orquesta integration tests which were failing quite often on Travis are also failing quite often on GHA.

I already assumed before it's some kind of race condition and now it looks even more likely.

@m4dcoder Do you think setting up a coordination backend for integration tests would make a difference with those race conditions / occasional failures? Or it's likely some other root cause we should look at improving in tests / similar?

Here are some examples:

* https://travis-ci.com/github/StackStorm/st2/builds/220669792
* https://travis-ci.com/github/StackStorm/st2/builds/220020111
* https://travis-ci.com/github/StackStorm/st2/builds/220015532 - here they also hand and timeout, perhaps some deadlock?

armab:
Wohoo! Congrats @cognifloyd, great achievement here that helps everyone a lot! :metal:

armab (approved):
:heavy_check_mark: integration tests for the GH Actions is the best indicator for me :+1: 
Great work!

Quick notes:
`ci.yaml` will need a bit of cleanup from the irrelevant comments and notes. Can we also remove Travis.yml?

cognifloyd:
Agreed, we need to do quite a bit of cleanup to get rid of all the remnants of travis.

The scope of this PR is to do the bare minimum to get it working in GHA. I do not want to disable Travis in this PR to demonstrate that it works in both CI systems. Once merged, a Maintainer can go into the dashboard on travis-ci.com to disable running Travis on this repo. 

Also, I tried to document in `ci.yaml` why I made various decisions so that when we do cleanup that file, we don't inadvertently break things by going back to something that didn't work. That said, there is a lot of room for improvement with the ci files. I definitely want to see better organization with some scripts mixing out of the workflow definition and others just going away as we make them obsolete.

Please, let's save a lot of this cleanup for future PRs.

scripts/travis/prepare-integration.sh
cognifloyd: It's not very difficult to make this work across multiple CI providers (like I did in https://github.com/ansible-community/molecule/pull/2976 ), but I didn't see the point since this will only be running in GHA for now.

As I don't want to disable Travis until after this PR is merged, I minimized my changes to the travis scripts. Once we disable travis, though, there's a lot of code on the chopping block.

contrib/runners/python_runner/python_runner/python_action_wrapper.py
```diff
@@ -359,6 +359,12 @@ def _get_action_instance(self):
 
         stdin_data = sys.stdin.readline().strip()
 
+        if not stdin_data:
+            # This could indicate that parent process (e.g. process which runs the tests has
+            # incorrectly opened the stdin and that one is then inherited by the process which is
+            # spawning it which will cause issues)
+            raise ValueError("Received no valid parameters data from sys.stdin")
```
blag, line 366: Not a blocker, but noting this in the future: it would be a good idea to log what already has been read into `stdin_data`.

```suggestion
            raise ValueError(f"Received no valid parameters data from sys.stdin:\n{stdin_data}")
```

blag (approved):

PR #5203, opened 2021-03-23
Various CI code related cleanup

Kami:
This pull request includes various CI code related cleanup:

* Move inline code into script files so we can run shellcheck, etc on it and make it easier to maintain.
* Move CI agnostic script files to ``scripts/ci/`` directory.
* Move GHA specific CI code to ``scripts/github/`` directory.
* Remove unused Travis CI files.
* Hook up shell check lint check to CI, clean up various shell chek lint violations


Kami:
As mentioned on Slack, this is just a starting up.

In the future, we also need to clean up actual workflow itself and split into multiple ones since current structure is not the best (we should move to different workflows for different tasks instead of re-using the same steps for most of the tasks which results in running unnecessary steps for some tasks).

Kami:
I also pushed a change to install runners and fixate requirements in parallel (f2b388a52, e13cae508) that's safe to do and saves us 20 seconds total or so.

Not a ton, but adds up across all the jobs and builds.

Kami:
Problem with moving integration tests to Github Actions is that we can't retry only a single job on failure (like we could on Travis).

This is a big of an issue / annoying since integration tests have a bunch of race conditions which we should try to fix asap.

cognifloyd (commented):
Only a minor doc nit. Looks nice and clean otherwise.

cognifloyd (approved):

PR #5205, opened 2021-03-24
Workaround modified GitHub Actions PATH var

cognifloyd:
We relied on ~/.local/bin being in PATH, which was added in:
actions/virtual-environments#2982

But, then they reverted that because it wasn't cross-shell consistent:
actions/virtual-environments#2982

So, vars were getting loaded from /etc/profile.d, then they moved it
back into ~/.bashrc. But, by default GHA workflows do not source
~/.bashrc.

Who knows how else they are going to mess with the PATH, so just write
our own.


amanda11 (approved):
LGTM

PR #5206, opened 2021-03-25
Configure redis coordination backend for orquesta integration tests

Kami:
This pull request configures redis coordination backend for integration tests to see if this will help with various race conditions which cause test failures we see quite often with orquesta integration tests.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -44,25 +44,25 @@ jobs:
       fail-fast: false
       matrix:
         include:
-        #   - name: 'Lint Checks'
-        #     task: 'ci-checks'
-        #     python-version: '3.6'
-        #   - name: 'Compile'
-        #     task: 'ci-compile'
-        #     python-version: '3.6'
-        #   - name: 'Pack Tests'
-        #     task: 'ci-packs-tests'
-        #     python-version: '3.6'
-        #   - name: 'Unit Tests'
-        #     task: 'ci-unit'
-        #     python-version: '3.6'
+          - name: 'Lint Checks'
+            task: 'ci-checks'
+            python-version: '3.6'
+          - name: 'Compile'
+            task: 'ci-compile'
+            python-version: '3.6'
+          - name: 'Pack Tests'
+            task: 'ci-packs-tests'
+            python-version: '3.6'
+          - name: 'Unit Tests'
+            task: 'ci-unit'
+            python-version: '3.6'
+          - name: 'Integration Tests'
+            task: 'ci-integration'
+            python-version: '3.6'
           # This job is slow so we only run in on a daily basis
           # - name: 'Micro Benchmarks'
           #   task: 'micro-benchmarks'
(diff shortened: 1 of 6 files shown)


amanda11:
Is this duplicate of https://github.com/StackStorm/st2/pull/5052?

Kami:
@armab Thanks - I missed that one.

In short, yeah, it's a "kinda of a duplicate.

My PR just configures it for tests, whereas that one also configures is for local development (and as such, also requires changes / updates to st2vagrantde, etc.).

For the goal is to just make those integration tests more stable and fail less often due to the race conditions.

So in short, technically we could also merge this PR earlier and then sync it up once #5052 is ready.

cognifloyd (commented):
I experimented with redis here:
https://github.com/st2sandbox/st2/blob/gha-service/.github/workflows/test-with-services.yml#L143-L203
I was looking for how to reconfigure it on the fly.

maybe that would be helpful

.github/workflows/ci.yaml
```diff
@@ -82,6 +98,22 @@ jobs:
...
+      # https://github.community/t/conditional-services-in-a-job/135301/3
+      redis:
+        # Docker Hub image
+        image: redis
+        # Set health checks to wait until redis has started
+        options: >-
+          --health-cmd "redis-cli ping"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
```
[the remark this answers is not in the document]
Kami: Yeah, for now I just added it as a service container, but once I have some more time and once I tackle GHA workflow refactoring into multiple jobs, I will re-organize it so we only spin it up for integration tests.

At this point I don't think we get tons if we move it to a task and conditionally spin it up there since we need to refactor GHA workflow sooner or later since it's a bit messy and instead of those various ifs, we should utilize multiple jobs...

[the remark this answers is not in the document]
Kami: And done - added the container name (will push it once the build is done).

Kami:
First test run with coordination backend passed, but this doesn't really mean anything.

I will change the code to run it in a loop a couple of time to see if there are any failures.

Kami:
OK so after 6 runs it didn't fail so that's a good sign.

If there are no objections, I will go ahead and merge it. It doesn't mean it fixes the problem for 100%, but we will see over time over more test runs.

st2common/st2common/service_setup.py
```diff
@@ -147,6 +147,9 @@ def setup(
 
     LOG.info("Using logging config: %s", logging_config_path)
 
+    LOG.info("Using coordination url: %s", cfg.CONF.coordination.url)
```
cognifloyd, line 150: Would we need to obfuscate the url if there is a password in it?

Kami: @cognifloyd Good catch, indeed - will add tests for it and obfuscate it.

Kami: d9e6c5517

PR #5212, opened 2021-03-27
Update requirements.txt

chadpatt:
Proposing changing pywinrm from 0.3.0 to 0.4.1

I have noticed type errors with pywinrm 0.3.0 that seems fixed in the later versions.


blag (dismissed):
Looks like you skipped reading the header on this file:

```bash
# Don't edit this file. It's generated automatically!
# If you want to update global dependencies, modify fixed-requirements.txt
# and then run 'make requirements' to update requirements.txt for all
# components.
# If you want to update depdencies for a single component, modify the
# in-requirements.txt for that component and then run 'make requirements' to
# update the component requirements.txt
```

Don't worry, I've done the exact same thing too sometimes, and I'm a StackStorm maintainer!

We like to keep all of the st2 components using the same versions of dependencies or else it quickly becomes a zoo. So to fix this for all of our components, please:

1. Update the dependency in _only_ `fixed-requirements.txt`
2. Run `make requirements` (this will create a virtualenv and run a script to update `pywinrm` in the `requirements.txt` files for all components and the base `requirements.txt` file)

That should make the lint step of the CI tests happy.

armab:
Would be great to include the Changelog record for this PR, see `CHANGELOG.rst`

Kami:
This seems like a reasonable change as long as all the tests pass - I believe window end to end tests should now be back in action so they should cover this change as well :)

Kami:
It looks like winrm unit test and end to end tests failed (I didn't dig into end to end failure yet, but it could be related to this change).

It looks like this will need more work (fixing / adjusting code and tests if and where needed). It looks like some of the failures at least may be Python 3 related so those will hopefully be easy to fix, but dunno without digging in.

For reference - here is the changelog - https://github.com/diyan/pywinrm/blob/master/CHANGELOG.md

Kami:
I pushed a small change (bbb74916ab5f2ae335c82e02bc4ffdb522b04c41) which fixed unit tests and looking at the end to end tests failure, I also think / hope it will fix end to end tests.

In short, it looks like one of the places in pywinrm used to accept string in the past and not it accepts bytes.

chadpatt:
Thank you. Pywinrm changed and does exactly what you saw.

Kami:
It looks like all is green after that small change / fix so it be safe to merge the PR.

Kami (approved):

PR #5218, opened 2021-04-03
Reformat more code with black

Kami:
This pull request re-formats some more files with black which were missed during the first pass.

The missed files were Python script which don't contain ``.py`` file name extension (but contain a shebang which makes them a Python script).

The corresponding Makefile targets have also been updated.


winem (commented):
Is `$$component/bin` correct? This gives me plenty of
```
===========================================================
Running black on contrib/runners/winrm_runner
===========================================================
All done! ✨ 🍰 ✨
13 files left unchanged.
grep: contrib/runners/winrm_runner/bin: No such file or directory
No Path provided. Nothing to do 😴
```

The following components / runners don't contain a bin directory:
st2client
st2tests
contrib/runners/inquirer_runner
contrib/runners/noop_runner
contrib/runners/local_runner
contrib/runners/winrm_runner
contrib/runners/http_runner
contrib/runners/remote_runner
contrib/runners/orquesta_runner
contrib/runners/python_runner
contrib/runners/announcement_runner
contrib/runners/action_chain_runner

I'd guard it in an if statement and check if the folder exists before running the grep command. 
Another option would be to omit /bin at all since grep is executed with -r (recursive) but this would need some more testing to avoid that files out of scope are being checked. 
(I already modified the Makefile in my local dev env and it did not show any errors.)

Makefile
```diff
@@ -375,13 +375,15 @@ black: requirements .black-check
 		echo "Running black on" $$component; \
 		echo "==========================================================="; \
 		. $(VIRTUALENV_DIR)/bin/activate ; black --check --config pyproject.toml $$component/ || exit 1; \
+		. $(VIRTUALENV_DIR)/bin/activate ; black $$(grep -rl '^#!/.*python' $$component/bin) || exit 1; \
```
winem, line 378: This is one out of 4 places where grep fails due to the non-existing bin-directory. Example: `grep: contrib/runners/announcement_runner/bin: No such file or directory`

Kami:
@winem 

> Is $$component/bin correct? This gives me plenty of

Yeah, it's correct.

I actually had if there first, but it's redundant so I removed it (aka black doesn't fail if directory doesn't exist and works just fine).

Makefile
Kami: Same here (see comment below).

If you feel strongly about it, I can add if back (but do keep in mind that this makes the Makefile a bit more messy :)).

winem:
Ok, I just did some tests and don't think that there is an easy way to handle this case without seeing the grep error and *not* having the risk of masking / missing other issues.

Given that we have a Makefile here and if's come with more duplicate code I'm fine with the current solution. 

We may just echo a line like "Missing files / directories were intentionally skipped." at the end of black-check and black-format to avoid confusion if someone is running these targets locally and sees the message. But I don't feel strongly about it and approve the PR as is now. :)

winem (approved):

PR #5219, opened 2021-04-03
Improve GHA workflow

Kami:
This pull request attempts to improve our GHA workflow which works, but it's somewhat hacky and performs a bunch of steps which are not needed.

It tries to accomplish that by splitting workflow into multiple jobs - lint checks, unit-tests, integration-tests/

This way we can avoid starting up service containers and running a bunch of other steps which are not needed for lint checks, unit tests, etc. In addition to that, utilizing multiple jobs is more appropriate for what we are doing and trying to achieve in our workflow (before we were really just "abusing" single job and if statement for doing what should be done as part of multiple jobs).

Sadly this does mean we will need to duplicate some steps since Github actions doesn't support YAML anchors yet (at least that was the case last time I checked and also one big negative thing for GHA from my perspective). 

Hopefully in the future, GHA will add support for YAML anchors so we can reduce some duplication.


Kami:
Now the UI output is also much better / more appropriate.

Screenshot from 2021-04-03 18-26-46

Kami:
All green and passing. RFR.

Screenshot from 2021-04-03 20-50-44

armab (approved):
:+1:

I remember it's something @cognifloyd mentioned the other day.
Makes it easy to re-start the specific failed jobs instead of the entire workflow from the beginning.

Kami:
@armab 

> I remember it's something @cognifloyd mentioned the other day.
Makes it easy to re-start the specific failed jobs instead of the entire workflow from the beginning.

I believe that's still not possible with this approach. Hopefully that will change in the future though since it's a often requested functionality for GHA.

To be able to do that, we would need to split it further into multiple workflows - which should now be possible quite easily with those changes, but it would mean even more duplication (and even if GHA adds supports for YAML anchors, we couldn't directly utilize it to reduce duplication since workflows are defined in separate files).

If people are OK with some more duplication, I'm fine with splitting into into 3 workflows (lint checks, unit tests, integration tests).

armab:
Thanks for the explanation. The current improvement makes a lot of sense too.
I'd worry about re-running if we really need it as GH Actions were much stable comparing to previous TravisCI.

winem:
I'd vote for separate workflows for now. This means to duplicate some code but, if we keep them clean and structured, it should be easy to merge them back into a single one, once GHA offers the required functionality.

Kami:
Here is an example which splits it into three workflows - 5b4239c450f82898bf0500800df03041ea5d265a.

Tentatively, for now, I'm fine with this change so we can just re-run some of the workflows on failure (timeout, race, whatever), but in the future when / if Github adds support for re-running a subset of jobs and YAML anchors, I'll likely strongly advocate for switching back since that will allow us to reduce some of the duplication - we can't utilize YAML anchors across multiple files (well, we could build out custom pre-processing scripts which do that, but that's definitely something we should not do).

---

EDIT: Actually, I'm more in favor of a single workflow, even at this point since it makes it easier to view the outputs for all the jobs - with 3 workflows, you need to visit three separate URLs, but with a single one, you just need to visit one. Aka better / more friendly UX.

If workflows were not related - aka not something you would want to view for every single PR / change, then having separate workflows may indeed be fine (e.g. workflow for lint / build / tests and a separate workflow for "deploy").

PR #5220, opened 2021-04-03
Fix subprocess handling when killed on timeout

r0m4n-z:
**Problem:**
Running ST2 actions as subprocesses (`eventlet.green`) spawns zombies on timeout.

**Reproduce:**
1. Spawn process which execution is going to take more than command's timeout.
2. `process.kill()` will be called.
3. Spawned process becomes zombie.

Method `process.kill()` is never handled as well as `process.communicate()`, which calls `process.wait()` - all those methods check `process.returncode` against None and if `returncode` already set, it just will be returned.

Why do I suggest to remove line 170: it doesn't seem to be necessary to set returncode here to avoid race condition for some `kill_func`. Probably additional custom attribute could be used here on `process` to provide that information if it's really necessary, but I don't think ST2 should do that.

### DIFF
--- a/st2common/st2common/util/green/shell.py
+++ b/st2common/st2common/util/green/shell.py
@@ -154,6 +154,9 @@ def run_command(
             read_stderr_func, process.stderr, read_stderr_buffer
         )
 
+    # Special attribute we use to determine if the process timed out or not
+    process._timed_out = False
+
     def on_timeout_expired(timeout):
         global timed_out
 
@@ -165,6 +168,8 @@ def run_command(
             # Note: We explicitly set the returncode to indicate the timeout.
             LOG.debug("Command execution timeout reached.")
 
+            process._timed_out = True
+
             if kill_func:
                 LOG.debug("Calling kill_func.")
                 kill_func(process=process)
@@ -173,9 +178,7 @@ def run_command(
                 process.kill()
 
             process.wait()
-            # NOTE: It's imporant to set returncode here as well, since call to process.kill() sets
-            # it and overwrites it if we set it earlier.
-            process.returncode = TIMEOUT_EXIT_CODE
+            process._timed_out = True
 
             if read_stdout_func and read_stderr_func:
                 LOG.debug("Killing read_stdout_thread and read_stderr_thread")
@@ -202,7 +205,11 @@ def run_command(
         stdout, stderr = process.communicate()
 
     concurrency.cancel(timeout_thread)
-    exit_code = process.returncode
+
+    if getattr(process, "_timed_out", False):
+        exit_code = TIMEOUT_EXIT_CODE
+    else:
+        exit_code = process.returncode
 
     if read_stdout_func and read_stderr_func:
         # Wait on those green threads to finish reading from stdout and stderr before continuing
--- a/st2common/st2common/util/green/shell.py
+++ b/st2common/st2common/util/green/shell.py
@@ -165,6 +165,8 @@ def run_command(
     # Special attribute we use to determine if the process timed out or not
     process._timed_out = False
 
+    # TODO: Now that we support Python >= 3.6 we should utilize timeout argument for the
+    # communicate() method and handle killing the process + read threads there.
     def on_timeout_expired(timeout):
         global timed_out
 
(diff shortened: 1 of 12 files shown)


Kami:
Thanks for the contribution.

That change will likely take a while to review since I remember we had edge cases in that code before so we need to make sure it's reviewed and tested well (and we also need a regression test case for this issue).

st2common/st2common/util/green/shell.py
```diff
@@ -165,17 +165,14 @@ def on_timeout_expired(timeout):
             # Note: We explicitly set the returncode to indicate the timeout.
             LOG.debug("Command execution timeout reached.")
 
-            # NOTE: It's important we set returncode twice - here and below to avoid race in this
-            # function because "kill_func()" is async and "process.kill()" is not.
-            process.returncode = TIMEOUT_EXIT_CODE
```
Kami, line 170: I'm looking into this tonight.

I started by adding a test case and I was indeed able to reproduce the zombie process test issue with a local command runner like setup where we use a custom ``kill_func()``.

It seems that adding ``process.wait()`` after kill fixes the zombie process issue and we can leave that exit code assignment in place - I think that's important since if we remove it process will have a different exit code set which would break our timeout detection logic in some places.

Kami: Also, just curious where you encountered this issue?

Was it via local shell command action runner or some other code path?

Kami: And under which Python version you encountered this issue?

Since technically now that only support Python >= 3.6, we could also simplify that function a bit and utilize ``timeout`` parameter for ``communicate()`` method which we couldn't do then we still supported Python 2.7.

Kami:
I added a integration regression test cases (39e95b40c99fd3bc04344b86c3d529114cb08634) and made a change to utilize a different variable on the process object to signal if a timeout has occurred (a4e3c98b5e5e08e2fc826b40449ff7dc3c0707af) - this should indeed be safer since it won't affect any upstream or other code which may be basing behavior of the value of that attribute. 

In the future when I get a chance, I also plan to simplify that function and utilize timeout parameter of the ``communicate()`` method.

Thanks again for reporting this issue.

Kami (approved):

st2common/st2common/util/green/shell.py
r0m4n-z: @Kami 
> Also, just curious where you encountered this issue?

That was long-running Python actions, but next morning found zombies and started to investigate root cause of them. Finally I realized those processes just never receive kill signal, checked source of `kill`/`communicate`/`wait` and here we are :)


> It seems that adding process.wait() after kill fixes the zombie process issue and we can leave that exit code assignment in place - I think that's important since if we remove it process will have a different exit code set which would break our timeout detection logic in some places.

No, `wait()` won't work in both cases (eventlet and gevent) because this method checks `returncode` against `None` to proceed, otherwise it will just return existing `returncode`. Also it should not cause any mess with timeout detection logic becuse it will be set to `-9` right after `wait()`.


> And under which Python version you encountered this issue?

Python 3.6. I'll be able to look into using `communicate()` here tonight.

Kami: Using ``communicate()`` is actually not a blocker, just a potential future simplification :)

And thanks for answering those questions.

I think the change which was merged should cover all those scenarios and I believe we now also have integration tests for most of those situations (shell=True as is the case for local runner, shell=False for python action runner, etc. Only thing we are missing is tests for read_stdout_func + read_stdout_func which verify threads which call those functions are correctly stopped / killed on timeout.).

Can you please confirm the change which has been merged indeed fixes the issue (I believe it should since it's based on your changes)?

r0m4n-z: > Using communicate() is actually not a blocker, just a potential future simplification :)
I see :)

> Can you please confirm the change which has been merged indeed fixes the issue (I believe it should since it's based on your changes)?
Just tried to reproduce the issue and everything seem to be fine:
1. subprocess is getting killed on timeout
2. it lefts no zombie process after kill process
3. sure, subprocess might be tracked down within ST2 as timed out process

Thank you for helping with my PR

Kami: Great. Thanks for confirming and submitting this bug fix.

PR #5225, opened 2021-04-08
Fix decrypting pack config keys under additionalProperties

cognifloyd:
The pack config loader uses "secret: true" in a schema to trigger decrypting a key from the datastore. However, it does not look at the schema under additionalProperties, so those properties can't use encrypted keys.

This issue was first documented in https://github.com/StackStorm-Exchange/stackstorm-jira/pull/19#discussion_r242319464

This PR fixes that so that the pack config loader handles additionalProperties when defined.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -42,6 +42,11 @@ Added
 
   Contributed by @Kami.
 
+* Fix a bug in the pack config loader so that objects covered by an additionalProperties schema
+  can use encrypted datastore keys and have their default values applied correctly.
+
+  Contributed by @cognifloyd.
+
 Changed
 ~~~~~~~
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -43,7 +43,7 @@ Added
   Contributed by @Kami.
 
 * Fix a bug in the pack config loader so that objects covered by an additionalProperties schema
-  can use encrypted datastore keys and have their default values applied correctly.
+  can use encrypted datastore keys and have their default values applied correctly. #5225
 
   Contributed by @cognifloyd.
 
--- a/st2common/st2common/util/config_loader.py
+++ b/st2common/st2common/util/config_loader.py
@@ -145,7 +145,7 @@ class ContentPackConfigLoader(object):
             is_list = isinstance(config_item_value, list)
 
             # pass a copy of parent_keys so the loop doesn't add sibling keys
-            current_keys = parent_keys + [str(config_item_keys)]
+            current_keys = parent_keys + [str(config_item_key)]
 
             # Inspect nested object properties
             if is_dictionary:
--- a/st2common/st2common/util/config_loader.py
+++ b/st2common/st2common/util/config_loader.py
@@ -135,7 +135,11 @@ class ContentPackConfigLoader(object):
         for config_item_key, config_item_value in iterator:
             if config_is_dict:
                 # different schema for each key/value pair
-                schema_item = schema.get(config_item_key, {})
+                try:
+                    # do not use schema.get() as schema might be a defaultdict
+                    schema_item = schema[config_item_key]
+                except KeyError:
+                    schema_item = {}
             if config_is_list:
                 # same schema is shared between every item in the list
                 schema_item = schema
--- a/st2common/st2common/util/config_loader.py
+++ b/st2common/st2common/util/config_loader.py
@@ -144,9 +144,11 @@ class ContentPackConfigLoader(object):
             is_dictionary = isinstance(config_item_value, dict)
             is_list = isinstance(config_item_value, list)
 
+            # pass a copy of parent_keys so the loop doesn't add sibling keys
+            current_keys = parent_keys + [str(config_item_keys)]
+
             # Inspect nested object properties
             if is_dictionary:
-                parent_keys += [str(config_item_key)]
                 property_schema = self._get_object_property_schema(
                     schema_item,
                     additional_properties_keys=config_item_value.keys(),
@@ -154,15 +156,14 @@ class ContentPackConfigLoader(object):
                 self._assign_dynamic_config_values(
                     schema=property_schema,
                     config=config[config_item_key],
-                    parent_keys=parent_keys,
+                    parent_keys=current_keys,
                 )
             # Inspect nested list items
             elif is_list:
-                parent_keys += [str(config_item_key)]
                 self._assign_dynamic_config_values(
                     schema=schema_item.get("items", {}),
                     config=config[config_item_key],
-                    parent_keys=parent_keys,
+                    parent_keys=current_keys,
                 )
             else:
                 is_jinja_expression = jinja_utils.is_jinja_expression(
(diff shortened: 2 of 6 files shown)


cognifloyd:
I'm force-pushing to re-trigger tests to get the integration test to pass. I can't figure out why one of 3 chunks has been failing.

cognifloyd:
All GHA tests are passing. CircleCI tests are broken due to a pbr+logshipper issue (known based on conversation in slack).

Kami:
Nice work :+1: 

I will have a more detailed look later this week.

cognifloyd:
Weird. I'm patching a v3.2 StackStorm running python2.7 and this delta was insufficient. afaik, the behavior of defaultdict should be the same under python3, so I'm not sure why the unit test is passing. In any case, I will push a commit with the fix momentarily.

CHANGELOG.rst
```diff
@@ -42,6 +42,11 @@ Added
 
   Contributed by @Kami.
 
+* Fix a bug in the pack config loader so that objects covered by an additionalProperties schema
```
Kami, line 45: Please also add the PR reference here - aka #5225

st2common/st2common/util/config_loader.py
```diff
@@ -98,6 +100,20 @@ def _get_values_for_config(self, config_schema_db, config_db):
         config = self._assign_default_values(schema=schema_values, config=config)
         return config
 
+    @staticmethod
+    def _get_object_property_schema(object_schema, init_additional_properties=None):
```
Kami, line 104: Please add docstring which explains what this function + ``init_additional_properties`` argument does.

st2common/st2common/util/config_loader.py
```diff
@@ -98,6 +100,20 @@ def _get_values_for_config(self, config_schema_db, config_db):
...
+        additional_properties = object_schema.get("additionalProperties", {})
+        if additional_properties and isinstance(additional_properties, dict):
+            property_schema = defaultdict(lambda: additional_properties)
+        else:
+            property_schema = {}
+        if init_additional_properties:
+            # ensure that these keys are present in the object (vs just defaultdict)
+            for key in init_additional_properties:
+                property_schema.__missing__(key)
```
Kami, line 113: Could we instead use regular for loop here + ``property_schema[key] = additional_properties``? Or we would then also need to use copy.copy?

Just thinking out loud since that code is more standard and may be easier to understand than relying on internal ``__missing__`` method.

Kami: And I guess if we just always init all the keys, we would not even need to use ``defaultdict``? Or am I missing some scenario?

st2common/tests/unit/test_config_loader.py
```diff
@@ -518,6 +518,68 @@ def test_get_config_dynamic_config_item_nested_list(self):
...
+                    "host": "127.0.0.1",
+                    "port": 8181,
+                    # unencrypted in datastore
+                    "token": "{{st2kv.system.k0}}",
+                },
+                "prod": {
+                    "host": "127.1.2.7",
+                    "port": 8282,
+                    # encrypted in datastore
+                    "token": "{{st2kv.system.k1_encrypted}}",
```
Kami, line 548: If user wants that value in the config to actually be decrypted, they still need to use ``decrypt_kv`` Jinja filter, right?

st2common/st2common/util/config_loader.py
cognifloyd: Sure. I added this section when defaultdict wasn't enough for applying the default values.

Kami:
Overall LGTM, just had some questions / comments on internal implementation details.

CHANGELOG.rst
cognifloyd: done

st2common/st2common/util/config_loader.py
```diff
@@ -101,16 +99,19 @@ def _get_values_for_config(self, config_schema_db, config_db):
...
         if additional_properties and isinstance(additional_properties, dict):
-            property_schema = defaultdict(lambda: additional_properties)
-        else:
-            property_schema = {}
-        if init_additional_properties:
-            # ensure that these keys are present in the object (vs just defaultdict)
-            for key in init_additional_properties:
-                property_schema.__missing__(key)
+            # ensure that these keys are present in the object
+            for key in additional_properties_keys:
```
Kami, line 113: Thanks, if that works it should be much more straight forward to understand the code now - well, at least for me :)

st2common/tests/unit/test_config_loader.py
cognifloyd: No. Config is automatically decrypted when the schema lists the key as `secret: true`

st2common/st2common/util/config_loader.py
Kami: And re - using copy - I actually verified, if we didn't use copy with ``defaultdict`` version, we also don't need to use it here - it's returning by reference version in both scenarios.

And since I believe we indeed never manipulate those values, only read / access them, copy is probably not needed.

st2common/tests/unit/test_config_loader.py
cognifloyd: See `_get_datastore_value_for_expression()` in `config_loader.py`

st2common/st2common/util/config_loader.py
cognifloyd: That ended up much cleaner. Thanks!

st2common/st2common/util/config_loader.py
cognifloyd: done

st2common/st2common/util/config_loader.py
cognifloyd: Yes. We modify the `config` object, but we don't modify (and probably should not modify) the schema itself. So, references are perfect in this case.

st2common/tests/unit/test_config_loader.py
Kami: Thanks for clarification - I was just somewhat confused by the test case aka something didn't add up.

Kami: I pushed a fix for this test - 574034f7636350dbb3ed81c3489278fec3b38fe6.

That's the part which confused me a bit.

I assumed we don't do decryption because the value which was stored and the one which was retrieved was supposed to be the same (but the KVP model layer doesn't perform any encryption and we need to encrypt the value ourselves when storing it).

Kami:
Some of those integration tests still seem to be failing quite often due to race / similar, quite annoying.

We should probably modify that job step to try to retry up to 2-3 times on failure to avoid annoying and manual re-run of the complete workflow.

st2common/tests/unit/test_config_loader.py
cognifloyd: Ah! That's why it seemed to be working before, but wasn't. Thanks for the fix!

cognifloyd:
GHA is passing now. Hooray.
Circle failed. Could you restart https://app.circleci.com/pipelines/github/StackStorm/st2/1542/workflows/128ecfa9-04ff-4b05-b410-94541fb12be2/jobs/12968/parallel-runs/2

cognifloyd:
All tests are green. Quick, let's merge before anything else gets merged into master. :)

Kami (approved):

PR #5226, opened 2021-04-08
Make redis the default coordinator backend

m4dcoder:
The coordinator service is required in most cases for running workflows. Install redis-server and configure redis as the default coordinator backend used by tooz. This will avoid a number of workflow issues, especially for new StackStorm users.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -42,6 +42,8 @@ Added
 
   Contributed by @Kami.
 
+* Make redis the default coordinator backend.
+
 Changed
 ~~~~~~~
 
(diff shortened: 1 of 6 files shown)


m4dcoder:
This PR pairs with https://github.com/StackStorm/st2-packages/pull/698.

armab (approved):
:+1: 

Would be nice to identify if stackstorm docs would need an update around this.
I believe we'll also need to include the change in the v3.5.0 Roadmap.

m4dcoder:
@armab Sure, will update docs to make mention

Kami (commented):
LGTM. This means we can now also simplify github st2.conf.dev step a bit.

Kami:
I assume there will also be a corresponding change for ``st2vagrantdev`` and other places (e.g. docker files) in the future?

Kami:
https://github.com/StackStorm/st2vagrantdev/pull/9 - will merge it as soon as the tests / CI there pass.

Issue #5221, opened 2021-04-06
Notify parameter in orquesta workflow not working as expected


khushboobhatia01:
## SUMMARY

https://docs.stackstorm.com/chatops/notifications.html#how-do-i-setup-notifications-for-orquesta Notification triggers are still emitted when notify set to [].

### STACKSTORM VERSION

st2 3.5dev, on Python 3.6.9

##### OS, environment, install method
Linux,

Post what OS you are running this on, along with any other relevant information/

- e.g. Docker, Vagrant, Kubernetes, etc. Describe how you installed ST2
- e.g. one-line install, custom install, etc -->

## Steps to reproduce the problem

orquesta-basic-meta.yaml
```
name: orquesta-basic
pack: examples
description: Run a local linux command
runner_type: orquesta
entry_point: workflows/orquesta-basic.yaml
enabled: true
notify:
  on-complete:
    routes:
      - slack
    message: '"@channel: Action run by {{action_context.user}} succeeded. Cmd run was {{cmd}}."'
parameters:
  cmd:
    required: true
    type: string
  timeout:
    type: integer
    default: 60
  notify:
    type: array
    default: []
```

orquesta-basic.yaml
```
version: 1.0
description: A basic workflow that runs an arbitrary linux command.
input:
  - cmd
  - timeout
tasks:
  task1:
    action: examples.local-notify cmd=<% ctx(cmd) %>
    next:
      - when: <% succeeded() %>
        publish:
          - stdout: <% result().stdout %>
          - stderr: <% result().stderr %>
output:
  - stdout: <% ctx(stdout) %>
 ```

local-meta.yaml
```
description: Action that executes an arbitrary Linux command on the localhost.
enabled: true
entry_point: ''
name: local-notify
notify:
  on-complete:
    routes:
    - slack
    message: '"@channel: Action run by {{action_context.user}} succeeded. Cmd run was {{cmd}}."'
parameters:
  cmd:
    description: Arbitrary Linux command to be executed on the remote host(s).
    required: true
    type: string
  sudo:
    immutable: true
runner_type: "local-shell-cmd"
```

## Expected Results

Notification trigger should be emitted only once for the workflow and not for `local-notify` action since notify parameter is set to an empty list.

## Actual Results

What happened? What output did you get?
ST2 trace
```
+--------------------------+------------------+--------------------------------+-----------------------------+
| id                       | type             | ref                            | updated_at                  |
+--------------------------+------------------+--------------------------------+-----------------------------+
[...]

PR #5227, opened 2021-04-09
Add support to skip action notifications when executed under workflow context

khushboobhatia01:
This pull request resolves https://github.com/StackStorm/st2/issues/5221

### DIFF
--- a/contrib/runners/orquesta_runner/tests/unit/test_notify.py
+++ b/contrib/runners/orquesta_runner/tests/unit/test_notify.py
@@ -348,3 +348,29 @@ class OrquestaNotifyTest(st2tests.ExecutionDbTestCase):
         self.assertEqual(ac_ex_db.status, action_constants.LIVEACTION_STATUS_SUCCEEDED)
         self.assertTrue(notifier.Notifier._post_notify_triggers.called)
         notifier.Notifier._post_notify_triggers.reset_mock()
+
+
+    def test_no_notify_task_list_for_task_with_notify(self):
+        wf_meta = base.get_wf_fixture_meta_data(TEST_PACK_PATH, "subworkflow-with-notify-task.yaml")
+        lv_ac_db = lv_db_models.LiveActionDB(
+            action=wf_meta["name"]
+        )
+        lv_ac_db.notify = notify_api_models.NotificationsHelper.to_model(MOCK_NOTIFY)
+        lv_ac_db, ac_ex_db = action_service.request(lv_ac_db)
+
+        # Assert action execution is running.
+        lv_ac_db = lv_db_access.LiveAction.get_by_id(str(lv_ac_db.id))
+        self.assertEqual(lv_ac_db.status, action_constants.LIVEACTION_STATUS_RUNNING)
+        wf_ex_db = wf_db_access.WorkflowExecution.query(
+            action_execution=str(ac_ex_db.id)
+        )[0]
+        self.assertEqual(wf_ex_db.status, action_constants.LIVEACTION_STATUS_RUNNING)
+
+        # Assert task1 notify is not set.
+        query_filters = {"workflow_execution": str(wf_ex_db.id), "task_id": "task1"}
+        tk1_ex_db = wf_db_access.TaskExecution.query(**query_filters)[0]
+        tk1_ac_ex_db = ex_db_access.ActionExecution.query(
+            task_execution=str(tk1_ex_db.id)
+        )[0]
+        tk1_lv_ac_db = lv_db_access.LiveAction.get_by_id(tk1_ac_ex_db.liveaction["id"])
+        self.assertIsNone(tk1_lv_ac_db.notify)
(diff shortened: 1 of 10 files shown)


contrib/runners/orquesta_runner/tests/unit/test_notify.py
```diff
@@ -348,3 +348,29 @@ def test_cascade_notify_to_tasks(self):
         self.assertEqual(ac_ex_db.status, action_constants.LIVEACTION_STATUS_SUCCEEDED)
         self.assertTrue(notifier.Notifier._post_notify_triggers.called)
         notifier.Notifier._post_notify_triggers.reset_mock()
+
+
+    def test_no_notify_task_list_for_task_with_notify(self):
```
Kami, line 353: If we don't have test cases for various other scenarios already (multiple tasks in a workflow, notify set only for some of them, all of them, etc.), it would be good to add them.

If we already have them, please ignore my comment.

st2common/st2common/services/action.py
```diff
@@ -548,3 +551,17 @@ def _is_notify_empty(notify_db):
...
         return True
     return not (notify_db.on_complete or notify_db.on_success or notify_db.on_failure)
+
+
+def _is_notify_skipped(liveaction):
+    """
+    notification is skipped if action execution is under workflow context and
+    task is not specified under wf_ex_db.notify["tasks"].
+    """
+    if not workflow_service.is_action_execution_under_workflow_context(liveaction):
```
Kami, line 561: I believe this will only work / correctly detect Orquesta workflows, but not action-chain ones, right?

Does this issue / bug only affect Orquesta workflows or also action-chain ones? If it also affects action chain ones, we probably also need to handle that scenario here

Kami:
Thanks for the contribution.

The bug fix / change looks good to me overall, I just had some questions / comments as far as tests and action-chain workflows go.

st2common/st2common/services/action.py
```diff
@@ -548,3 +551,17 @@ def _is_notify_empty(notify_db):
...
+        return False
+    wf_ex_db = WorkflowExecution.get_by_id(liveaction.workflow_execution)
```
Kami, line 563: Just thinking out loud.

This means that we will now perform 2 additional DB queries for any workflow action when calling ``create_request``, right?

I guess that is probably not that big deal, but it may make sense to utilize "only_fields" so we only retrieve fields we need on those two models (``notify`` for ``WorkflowExecutionDB``, ``task_name`` for ``TaskExecution``).

---

EDIT: Yeah, ``WorkflowExecutionDB`` can also contain ``result`` and other large fields values so it is actually important we only retrieve fields we need.

Luckily my recent DB performance optimizations make this less of an issue (in the past, retrieving large fields such as results from the database could easily take many seconds or more), but we should still be careful and avoid retrieving unnecessary fields / data we don't need.

st2common/st2common/services/action.py
khushboobhatia01: Verified, issue exists for action-chain workflows as well. Modified the changes to fix same for action-chains. The fix isn't as direct as it for orquesta, as skip_notify parameter is available only in runner parameters. Please review and let me know if the changes look ok to you. Keeping this thread unresolved till then.

Kami: Thanks for also tackling this for action chain :+1:

khushboobhatia01: @Kami If the changes look ok to you, can you please approve and merge? Thanks :)

contrib/runners/orquesta_runner/tests/unit/test_notify.py
khushboobhatia01: Addressed.

st2common/st2common/services/action.py
khushboobhatia01: Addressed.

Kami (approved):
LGTM :+1: 

Thanks for working.

I will merge this into master and add changelog entry separately so I don't block and hold this change any longer.

Kami:
For some reason after merging this, those new tests are failing for me locally with this error - https://gist.github.com/Kami/52543866f5400e29537f00911654bfe1.

No idea (yet) why it happens locally, but not on CI. I likely won't have time to dig in before the weekend.

Kami:
OK, I see the issue - https://github.com/StackStorm/st2/pull/5227/files#diff-c86e8df3fb95a5e1b0806ce20a77ee1a92f721e933a1d230e5a302b1ebb10873R210.

This function needs some more love. Right now it will return as soon as execution has 1 or more children, but from the code that's not actually what we want - we want to wait for 2 children. So the code needs to be changed to handle that better otherwise it will be prone to timing / race conditions.

I will push a fix shortly.

Kami:
Here we go - 5856fa482.

I confirmed that with this change, the tests now consistently pass locally.

khushboobhatia01:
> Here we go - [5856fa4](https://github.com/StackStorm/st2/commit/5856fa482f3371696ca0574b4576722c34ee3e8d).
> 
> I confirmed that with this change, the tests now consistently pass locally.

Thanks for fixing the test case 👍

PR #5228, opened 2021-04-09
Drop unused dependencies

cognifloyd:
`prometheus_client` was part of a metrics driver that was dropped. #4310 (https://github.com/StackStorm/st2/pull/4310/commits/82772cc64026b83792a344561844c4dcb09946e7) #4504 (https://github.com/StackStorm/st2/pull/4504/commits/873c57f8331e29832e17904fd9dbb9494d53c53b) _unused since v2.9.0_

`zipp` and `more-itertools` were py2-only transitive deps. #4843 #4847 (afaict these were transitive deps via `wheel`) _added in v3.2.0; unused since v3.4.0 when we dropped python 2_

`python-gnupg` was dropped with the removal of st2debug. #5103 _unused since v3.4.0_

background: _I discovered this as part of an experiment using [pants](https://pantsbuild.org) in [st2sandbox/st2@pants](https://github.com/st2sandbox/st2/tree/pants) to replace the Makefile and related scripts._

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -238,6 +238,10 @@ Improvements
 
   Contributed by @ashwini-orchestral
 
+* Drop unused python dependencies: prometheus_client, python-gnupg, more-itertools, zipp. #5228
+
+  Contributed by @cognifloyd.
+
 Fixed
 ~~~~~
 
--- a/fixed-requirements.txt
+++ b/fixed-requirements.txt
@@ -32,7 +32,6 @@ prompt-toolkit==1.0.15
 pyinotify==0.9.6; platform_system=="Linux"
 pymongo==3.11.3
 python-editor==1.0.4
-python-gnupg==0.4.5
 python-keyczar==0.716
 pytz==2021.1
 pywinrm==0.3.0
@@ -55,14 +54,11 @@ zake==0.2.2
 bcrypt==3.2.0
 jinja2==2.11.3
 mock==4.0.3
-more-itertools==5.0.0
 nose-timer==0.7.5
 nose-parallel==0.3.1
 psutil==5.8.0
 python-dateutil==2.8.1
 python-statsd==2.1.0
-prometheus_client==0.1.1
 ujson==1.35
 orjson==3.5.1
 udatetime==0.0.16
-zipp>=0.5,<=1.0.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -30,7 +30,6 @@ kombu==4.6.6
 lockfile==0.12.2
 mock==4.0.3
 mongoengine==0.23.0
-more-itertools==5.0.0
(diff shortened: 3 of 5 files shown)


winem (approved):

amanda11 (approved):
LGTM - but perhaps add a CHANGELOG entry

blag (approved):

cognifloyd:
This is cleanup missed from previous PRs that removed the dependencies. I don't think it warrants a changelog entry.

armab:
@cognifloyd For context, what's the PR # and changelog entry you're referring to?

blag:
@armab That's a bit of a fishing expedition to ask somebody to go on, don't you think? And I don't understand how the results would matter.

I agree with @cognifloyd, I don't know that this is really a big enough change on its own to warrant a line in the changelog. This isn't going to change any user-facing code.

armab:
I believe having a Changelog record helps a lot for both users and developers and agree that having it would be nice for consistency with the engineering flow.
Looking deeper, I can't find the 3.5dev record indicating the requirements cleanup and wondering if it's 3.4dev or any other change?
In the past, we included Changelog for the changes like CI, Tests and other non-user facing items and don't see a problem with updating it here.

amanda11:
To make it simple the changelog could just say remove unused dependencies - rather than needing a full investigation to work out exactly when those dependencies were removed. It states the point of the change, but doens't require lots of effort for what is a nice clean up - which is always appreciated.

cognifloyd:
I think that's too much noise in the changelog for something like this.

I did all that research when I validated that these were not needed. Let me do it again to provide info. I'll update the original post with details as I find it.

cognifloyd:
`prometheus_client` unused since v2.9.0
The other 3 are not needed as of v3.4.0 when we dropped python2 and removed st2debug.

cognifloyd:
I don't think a changelog entry adds any value, so I'm not sure what to write.
All of you are maintainers, please feel free to add a changelog entry with whatever verbiage you think makes sense.

armab:
Something simple like this would be totally fine: 
```rst
* Cleanup unused ``pip`` dependencies (improvement). #5228

  Contributed by @cognifloyd.
```

Kami (approved):
Yay, the more stuff we can remove, the better :+1: 

But yeah, please still add a changelog entry for just in case.

Kami:
Eventually we could also still removing all the Python 2 / 3 compatibility boiler plate and also get rid of ``six``.

Kami:
And re changelog entry - IIRC, anything from StackStorm virtualenv is also available to pack virtualenv (but I could also be wrong on that since that behavior had some edge cases and has changed in the past) so in theory a pack could be depending on some of those dependencies without specifying them in ``requirements.txt``, but that's also very unlikely.

PR #5229, opened 2021-04-09
Add some more tests for run_command() function

Kami:
This PR adds some more tests for ``st2common.util.green.shell.run_command()`` function.

### DIFF
--- a/st2common/tests/fixtures/print_to_stdout_stderr_sleep.sh
+++ b/st2common/tests/fixtures/print_to_stdout_stderr_sleep.sh
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
 # Copyright 2021 The StackStorm Authors.
-# Copyright 2019 Extreme Networks, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/st2common/tests/fixtures/print_to_stdout_stderr_sleep.sh
+++ b/st2common/tests/fixtures/print_to_stdout_stderr_sleep.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+# Copyright 2021 The StackStorm Authors.
+# Copyright 2019 Extreme Networks, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+echo "pre sleep"
+>&2 echo "pre sleep stderr"
+sleep 1599
+echo "post sleep"
+>&2 echo "post sleep stderr"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -469,13 +469,13 @@ jobs:
           ./scripts/ci/print-versions.sh
       - name: make
         if: "${{ env.TASK == 'ci-integration' }}"
-        timeout-minutes: 8
+        timeout-minutes: 7
         # use: script -e -c to print colors
         run: |
           script -e -c "make ${TASK}" && exit 0
       - name: make (orquesta tests)
         if: "${{ env.TASK == 'ci-orquesta' }}"
-        timeout-minutes: 8
+        timeout-minutes: 10
         env:
           MAX_ATTEMPTS: 3
           RETRY_DELAY: 5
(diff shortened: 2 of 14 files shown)


st2common/tests/fixtures/print_to_stdout_stderr_sleep.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+# Copyright 2021 The StackStorm Authors.
+# Copyright 2019 Extreme Networks, Inc.
```
amanda11, line 3: Just a comment on copyright. If this is a new file then surely it should only have copyright from the date it was created?

Kami: I actually just followed example from existing files (they didn't seem to update Extreme line), but that's a good / interesting question.

Technically since it's a new file, Extreme doesn't hold a copyright to it so it could probably be removed (but would be good to double check with LF legal and keeping track of that may be a bit hard).

Also, a lot of projects yearly update copyright strings in all the files so they reflect start - current year (e.g. 2014-2021).

amanda11:
It was more I didn't think you could copyright a file with a date earlier than it was created.

st2common/tests/fixtures/print_to_stdout_stderr_sleep.sh
armab: That's right. I can confirm that for all the new files we just use the StackStorm Authors and don't need Extreme.

```
# Copyright 2021 The StackStorm Authors.
```

armab, line 3: That's right. I can confirm that for all the new files we just use the StackStorm Authors and don't need Extreme.


```suggestion
# Copyright 2021 The StackStorm Authors.
```

Kami: thanks for confirming :+1:

Kami:
Also pushed a change (d19ebac25) which I think should help with occasional unit test failure we see.

```bash
TEST RESULT OUTPUT:
1) ERROR: test_service_setup_default_st2_conf_config_is_used (tests.unit.test_service_setup.ServiceSetupTestCase)
    virtualenv/lib/python3.6/site-packages/mock/mock.py line 1346 in patched
      return func(*newargs, **newkeywargs)
    st2common/tests/unit/test_service_setup.py line 160 in test_service_setup_default_st2_conf_config_is_used
      run_migrations=False,
    st2common/st2common/service_setup.py line 120 in setup
      config.parse_args(config_args)
    st2common/st2common/config.py line 723 in parse_args
      register_opts()
    st2common/st2common/config.py line 176 in register_opts
      do_register_opts(webui_opts, "webui", ignore_errors)
    st2common/st2common/config.py line 33 in do_register_opts
      cfg.CONF.register_opts(opts, group=group)
    virtualenv/lib/python3.6/site-packages/oslo_config/cfg.py line 1820 in __inner
      result = f(self, *args, **kwargs)
    virtualenv/lib/python3.6/site-packages/oslo_config/cfg.py line 1977 in register_opts
      self.register_opt(opt, group, clear_cache=False)
    virtualenv/lib/python3.6/site-packages/oslo_config/cfg.py line 1824 in __inner
      return f(self, *args, **kwargs)
    virtualenv/lib/python3.6/site-packages/oslo_config/cfg.py line 1961 in register_opt
      return group._register_opt(opt, cli)
    virtualenv/lib/python3.6/site-packages/oslo_config/cfg.py line 1344 in _register_opt
      if _is_opt_registered(self._opts, opt):
    virtualenv/lib/python3.6/site-packages/oslo_config/cfg.py line 574 in _is_opt_registered
      raise DuplicateOptError(opt.name)
   DuplicateOptError: duplicate option: webui_base_url
```

Kami:
Per slack - we still see failures and timeouts with some Orquesta integration tests quite often.

Those failures and especially timeouts are rather annoying.

To try to mitigate and improve this situation, I will make the following change and see if it helps - isolate problematic tests and run them separate from other integration tests in a separate process. I will also set reasonable / low timeout for those tests and only try to retry those tests on failure / timeout - this should result in better UX, faster tests and less wasted time re-running CI builds...

Kami:
Once CI is passing I will merge this into master to see if it's working correctly.

If we still encounter a lot of failures going forward, I will split orquesta integration tests in a new separate workflow. This will allow us to retry just that workflow on timeout / failure. It's not ideal, but better than current status quo / alternatives (wasting tons of time on re-runs).

Kami:
Alright, orquesta integration tests now run as part of a separate workflow. This will allow us to save a bunch of time only retrying that workflow (over the last weeks I wasted many hours on manual re-runs, etc.).

Hopefully in the future we will also be able to track and and fix actual race conditions so we can merge everything back into a single workflow again (and get rid of some duplication).

/cc @m4coder - can you please dig into those tests when you get a chance (st2tests/integration/orquesta). I did dig in a bit a while back, but without much success. You can find a bunch of examples of failures and timeouts due to race conditions in Travis and GHA build run logs.

Kami:
I'll merge this since nightly CI is failing now daily and I think this should help with that (at the very least just Orquesta workflow will fail).

PR #5234, opened 2021-04-14
Unify and improve config option registration code

Kami:
This PR attempts to unify and make all the config registration code more consistent - aka so it exposes / utilizes the same API.

I only started to dig into since I started a bunch of test errors like that recently - https://gist.github.com/Kami/fcfa62f1ae7f0b84cd3981ecd6f5fe37.

It looks like there is some kind of race with config option registration which likely depends on the import ordering or similar. Either this issue has been there for a long time already (not so unlikely) or some recent code changes re-arranged the imports or similar which started to trigger this issue more often (I did quickly check git history without much success., but I could have also missed something)

Sadly most of the config modules itself have import time side affects (options are registered on import) which is bad. Fixing / improving that would likely be a very deep rabbit hole since tons of existing code and tests relies on this behavior.

So with this PR I'm really just trying to unify the code and gets tests to pass consistently. Keep in mind that registering the same options multiple times should not be and it's not fatal so if we ignore such errors it's OK (well, it's not prefect, but it's compromise - an alternative is much larger code change which could likely takes many days to even weeks).


Kami:
I'll merge this since it also contains a test fix (https://github.com/StackStorm/st2/pull/5227#issuecomment-819812421) and the change itself should be safe and fully backward compatible - this way we can get some "CI miles" on it as part of nightly job and see how it behaves over multiple build runs.

PR #5237, opened 2021-04-15
Update more dependencies to various stable versions

Kami:
This PR updates more dependencies to latest stable versions.

Only potentially more risky ones could be kombu / amqp, but I went over the changelog and integration + end to end tests should catch if there are any issues.


Kami:
OK, tests exposed an issue with kombu / pyaqmp upgrade.

`ssl_cert_reqs`` now defaults to cert.required which is a good and right thing, but our existing tests rely on that value defaulting to cert.none (which used to be true in older versions).

Problem also is that one of the tests is mixing connection parameters in URL + connection parameters via config options - and that doesn't work well / correctly when trying to use custom ca cert for ssl validation.

I think I can fix / change that test and document that behavior in docs and upgrade notes and we should be fine.

It seems like behavior when combining connection parameters in url + kwargs has changed.

nmaludy (approved):

winem (approved):

blag (approved):

blag:
Thanks for getting this done.

PR #5246, opened 2021-04-22
Update sample nginx config with information on how to enable response compression

Kami:
This pull request updates nginx config with example on how to update response compression for st2api + st2stream endpoints.

This should come handy in a lot of situations where StackStorm is accessed over the public internet (aka slower connection).

I tested it with my StackStorm deployment and everything seems to be working well. I think we should also apply this to our st2cicd instance (that + DB performance improvements will make a massive difference with slow and laggy UI).

## TODO

- [ ] Corresponding st2docs change


conf/nginx/st2.conf
```diff
@@ -36,6 +36,8 @@ server {
   ssl_certificate_key       /etc/ssl/st2/st2.key;
   ssl_session_cache         shared:SSL:10m;
   ssl_session_timeout       5m;
+  # If your version of OpenSSL supports it, you are strongly recommended to also (or only) use TLS v1.3
+  #ssl_protocols             TLSv1.2 TLSv1.3;
```
armab, line 40: What prevents us to stick with this new setting by default?
From the StackStorm supported OSes `U16/18` and `EL7/8` which one doesn't support `TLS v1.3`?

Kami: Ubuntu 16.04.

So once we remove support for it we can change the default to be TLSv1.2 + TLSv1.3 (and in the future just TLSv1.3).

Kami: @armab I checked online and it appears OpenSSL 1.1.1 which is needed for TLS v1.3 is also not available in CentOS 7 by default.

But I didn't verify it yet with out installer (e.g. if we add some epl repo which can pull down latest version).

armab, line 40: This could be removed now, considering https://github.com/StackStorm/st2/pull/5280 is merged
```suggestion
```

armab (approved):
LGTM!

PR #5248, opened 2021-04-22
Add support for setting SameSite and Secure attribute for auth cookie we set

Kami:
This pull request includes a small "security hardening" change.

It allows operator to configure value for ``SameSite`` attribute (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite, https://web.dev/samesite-cookies-explained/) which is set with the ``auth-token`` cookie we set in some situations (e.g. when authenticating via st2web and similar).

The value defaults to ``Lax`` which should work as a good secure default (defining it to Strict may break some in some situations, see the link above).

## TODO

- [x] Unit tests
- [x] Improve config option help string
- [x] Upgrade notes entry - https://github.com/StackStorm/st2docs/pull/1070
- [ ] Perhaps a st2docs page on various "best security practices" for deployments? (not necessary blocker for this PR)


Kami:
I noticed we also don't set ``secure`` flag for the cookie.

I will also add an option for that and default it to True since it's a best security practice.

In case someone doesn't run StackStorm over https (bad idea), they will need to set it to False. I will open st2docs upgrade notes entry which documents how to do that.

Also keep in mind that this cookie is pretty much only used when logging via token / api key in query parameters (which pretty much only means st2web for our official stuff).

conf/st2.conf.sample
```diff
@@ -32,6 +32,11 @@ workflows_pool_size = 40
 [api]
 # List of origins allowed for api, auth and stream
 allow_origin = http://127.0.0.1:3000 # comma separated list allowed here.
+# SameSite attribute value for the auth-token cookie we set on successful authentication from st2web. If you don't have a specific reason (e.g. supporting old browsers) you are recommended to set this value to strict.
```
cognifloyd, line 35:
```suggestion
# SameSite attribute value for the auth-token cookie we set on successful authentication from st2web. If you don't have a specific reason (e.g. supporting old browsers) we recommend you set this value to strict.
```

st2api/st2api/validation.py
```diff
@@ -15,10 +15,58 @@
...
+                cfg.CONF.api.auth_cookie_same_site,
+                type(cfg.CONF.api.auth_cookie_same_site),
+            )
+        )
+
+    # Now we try to make a dummy cookie to verify all the options are configured correctly. Some
+    # Options are mutually exclusive - e.g. SameSite none and Secure false.
+    try:
+        # NOTE: None and none don't mean the same thing - None implies not setting this attribute
+        # (backward compatibility) and none implies setting this attribute value to none
```
cognifloyd, line 47: wow. That's a very minute distinction - could we use "unset" instead of "None"?

st2common/st2common/config.py
```diff
@@ -373,6 +373,22 @@ def register_opts(ignore_errors=False):
...
+            "successful authentication via st2web. You should only set this to False if you have "
+            "a good reason to not run and access StackStorm behind https proxy.",
+        ),
+        cfg.StrOpt(
+            "auth_cookie_same_site",
+            default="lax",
+            choices=["strict", "lax", "none", "None"],
+            help="SameSite attribute value for the "
+            "auth-token cookie we set on successful authentication from st2web. If you "
+            "don't have a specific reason (e.g. supporting old browsers) you are recommended to "
```
cognifloyd, line 389:
```suggestion
            "don't have a specific reason (e.g. supporting old browsers) we recommend you "
```

cognifloyd (commented):
Ooh. nice.
I have a doc wording suggestion, and a question about `none` vs `None`.

st2api/st2api/validation.py
Kami: Yeah, that's how it's handled by the browser, I'm not a big fan of it either.

I'm fine for using a different more explicit value such as ``unset``.

Kami:
@cognifloyd I pushed a change which renames ``None`` to ``unset`` (which I also think it's better).

Hopefully CI and tests will pass since my local dev environment is totally toast and I don't have multiple hours to spend to try to fix it at this point.

cognifloyd (approved):
Awesome. I'm not merging right now since we're in code freeze for the 3.6 release.

Kami:
Now that v3.6.0 merge freeze is over, I will go ahead and merge it into master.

PR #5250, opened 2021-04-29
Secret masking in output_schema feature is added

mahesh-orch:
With this implementation, action responses marked as secret in output schema are masked in CLI and UI.


m4dcoder (changes requested):
This is a good start. But from the current PR, I don't see how this is wired to the API and CLI masking the result in the output on execution get. Can you write a unit test that can show this is working for GET method of action execution in the API?  Also, the secret masking of action execution output shouldn't affect action execution in a workflow, can you also add a test case for workflow execution where the output of an action execution of a task isn't masked?

mahesh-orch:
Thanks @m4dcoder for the comments. I have pushed the new changes and they are wired to the API and CLI masking the result in the output on execution get. Please review the changes.

mahesh-orch:
@m4dcoder I have moved the logic for masking secret parameters in output schema from https://github.com/StackStorm/st2/blob/master/st2common/st2common/models/api/execution.py to https://github.com/StackStorm/st2/blob/master/st2common/st2common/models/db/execution.py. Also added a test case in https://github.com/orchestral-st2/st2/blob/master/st2common/tests/unit/test_db_execution.py. Please review the added code. Thanks.

st2common/st2common/models/api/execution.py
```diff
@@ -152,9 +152,9 @@ class ActionExecutionAPI(BaseAPI):
     @classmethod
     def from_model(cls, model, mask_secrets=False):
         doc = cls._from_model(model, mask_secrets=mask_secrets)
-
+
```
cognifloyd, line 155:
```suggestion

```

st2common/st2common/models/api/execution.py
```diff
@@ -152,9 +152,9 @@ class ActionExecutionAPI(BaseAPI):
     @classmethod
     def from_model(cls, model, mask_secrets=False):
         doc = cls._from_model(model, mask_secrets=mask_secrets)
-
+        
         doc["result"] = ActionExecutionDB.result.parse_field_value(doc["result"])
-
+
```
cognifloyd, line 157:
```suggestion

```

cognifloyd (commented):
Nice! This is coming along, and I'm really looking forward to using it.

Some whitespace needs to be cleaned up.

@m4dcoder Can you enable running tests on this PR? This is @mahesh-orch's first PR on st2, so github is blocking running actions until someone with write access enables them.

st2common/st2common/models/db/execution.py
```diff
@@ -141,11 +152,27 @@ def mask_secrets(self, value):
                         p: "string" for p in liveaction["parameters"]["response"]
                     },
                 )
-
+        
+        output_schema_result = ActionExecutionDB.result.parse_field_value(
+            result["result"]
```
m4dcoder, line 157: The result["result"] here is specific to python action. For shell or workflow action, this is going to return key error. There's an output key defined in the action runner model that says which attribute the action is output to in the result (i.e. result for python, output for workflow, etc.). The output key is defined at https://github.com/StackStorm/st2/blob/master/st2common/st2common/models/db/runner.py#L70.  You can see how it is used to validate output schema at https://github.com/StackStorm/st2/blob/master/st2common/st2common/util/action_db.py#L247. The mask secret for output will also need to do this. This also means you will need unit tests for result of different action types.

st2common/st2common/models/db/execution.py
```diff
@@ -141,11 +152,27 @@ def mask_secrets(self, value):
...
+        )
+        
+        # accessing parameters marked secret as true in the output_schema
+        # and masking them for the result in action execution API
+        if output_schema_result:
```
m4dcoder, line 162: I suggest moving this block into a function called `mask_secret_output` and put it under the module https://github.com/StackStorm/st2/blob/master/st2common/st2common/util/output_schema.py. This will help whoever is making changes next can look at the output_schema functionalities in one place.

m4dcoder (changes requested):

st2common/st2common/models/db/execution.py
```diff
@@ -142,10 +154,22 @@ def mask_secrets(self, value):
...
+            result["result"]
+        )
+
+        # accessing parameters marked secret as true in the output_schema
+        # and masking them for the result in action execution API
+        if output_schema_result:
+            for key, spec in six.iteritems(result["action"]["output_schema"]):
+                if spec.get("secret", False):
+                    # TODO: Change to output_schema_result[output_key][key] = MASKED_ATTRIBUTE_VALUE
+                    output_schema_result["result"][key] = MASKED_ATTRIBUTE_VALUE
```
m4dcoder, line 167: The result["result"] here is specific to python action. For shell or workflow action, this is going to return key error. There's an output key defined in the action runner model that says which attribute the action is output to in the result (i.e. result for python, output for workflow, etc.). The output key is defined at https://github.com/StackStorm/st2/blob/master/st2common/st2common/models/db/runner.py#L70. You can see how it is used to validate output schema at https://github.com/StackStorm/st2/blob/master/st2common/st2common/util/action_db.py#L247. The mask secret for output will also need to do this. This also means you will need unit tests for result of different action types.

m4dcoder (changes requested):

mahesh-orch:
Hi @m4dcoder, thanks for the comments. With newly pushed code, KeyError with result["result"] issue has been resolved. Also, http-runner test has passed. mask_secret_output block has been shifted to output_schema.py. Added new test cases for the same. Please review the changes.

st2common/st2common/util/output_schema.py
```diff
@@ -51,6 +52,23 @@ def _validate_action(action_schema, result, output_key):
...
+        if (
+            value.get("action", {})
+            .get("output_schema", {})
+            .get(key)
+            .get("secret", False)
+        ):
+            if output_schema_result.get("result"):
+                output_schema_result["result"][key] = MASKED_ATTRIBUTE_VALUE
+            if output_schema_result.get("body"):
+                output_schema_result["body"][key] = MASKED_ATTRIBUTE_VALUE
```
m4dcoder, line 67: @mahesh-orch Instead of hard coding "result" and "body" and using if else, how come you are not using the output key from the runner model dynamically like https://github.com/StackStorm/st2/blob/master/st2common/st2common/util/action_db.py#L247?  I mentioned this in my last comment. What you've done here only works for python and http runner. Workflow action will fail. Please add a unit test for the Orquesta runner and you will see.

m4dcoder (changes requested):

mahesh-orch:
@m4dcoder I have put several if conditions before using output_key because some runners don't have output_key. I have to mention orquesta runner_type also in condition because it was masking secret parameters in output of the task in workflow. Please review the changes.

st2common/st2common/util/output_schema.py
```diff
@@ -51,6 +52,28 @@ def _validate_action(action_schema, result, output_key):
...
+        and value["action"]["runner_type"] != "orquesta"
+        and value["action"]["runner_type"] != "announcement"
+        and value["action"]["runner_type"] != "inquirer"
+        and value["action"]["runner_type"] != "noop"
+        and value["action"]["runner_type"] != "remote-shell-script"
+        and value["action"]["runner_type"] != "winrm-cmd"
+        and value["action"]["runner_type"] != "winrm-ps-cmd"
+        and value["action"]["runner_type"] != "winrm-ps-script"
+    ):
+        output_key = value["runner"]["output_key"]
```
cognifloyd, line 68: orquesta runner does have output_key:
https://github.com/StackStorm/st2/blob/c15d9ba6f8e706522b5d96912ae09c74975cfc7f/contrib/runners/orquesta_runner/orquesta_runner/runner.yaml#L17

But you are right that `output_key` is not required:
https://github.com/StackStorm/st2/blob/c15d9ba6f8e706522b5d96912ae09c74975cfc7f/st2common/st2common/models/api/action.py#L114-L118
I wonder about (in a different PR) making it required so that output masking would work with these other runners too.

So, maybe do this instead:
```python
def mask_secret_output(value, output_schema_result):
    output_key = value["runner"].get("output_key")
    if not output_key:
        return output_schema_result
    value = copy.deepcopy(value)
    for key, spec in value["actions"]["output_schema"].items():
```

Also we don't need to use six anymore as we don't support python2. Just do `.items()`.

cognifloyd:
> I have to mention orquesta runner_type also in condition because it was masking secret parameters in output of the task in workflow. 

What do you mean? Can you share an example of what was getting masked when masking the output of an orquesta workflow?

m4dcoder (changes requested):
@mahesh-orch

I have refactored your PR please review my changes. The if/else check on runner type is too fragile and again is bug prone if someone goes and change the schema for the runner type of add new runner types. Your change already introduced a bug for the orquesta runner (workflow).  I refactored it and also added a unit test for orquesta runner in st2actions/tests/unit/test_output_schema.py to catch that. There are other changes I did to simplify the PR. Also, I've fixed tests so CI succeeded now.

There are a still open items in that PR you should follow up on.  1) Please add a test for workflow. Add a task with secrets in the output schema, publish the result of the task to the workflow output and make sure the value is not masked in the task. You can add this test at https://github.com/StackStorm/st2/blob/master/contrib/runners/orquesta_runner/tests/unit/test_data_flow.py. 2) Please add a unit test in the API to make sure output with secrets is masked at https://github.com/StackStorm/st2/blob/master/st2api/tests/unit/controllers/v1/test_executions.py.

mahesh-orch:
> What do you mean? Can you share an example of what was getting masked when masking the output of an orquesta workflow?

@cognifloyd Thank you for the comments and inputs. Previous my changes without if condition to orquesta runner were causing masking the secret parameter in workflow task output but @m4dcoder had improved that piece of code and now it is working fine.

mahesh-orch:
@m4dcoder, @cognifloyd Hi, I am pushing code of unit test to show in workflow secret parameters in output_schema in action should not be masked in task output. I need help as unit test is failing at the task which contain python action.
Unit test is in https://github.com/StackStorm/st2/blob/master/contrib/runners/orquesta_runner/tests/unit/test_data_flow.py. Test is failing at L262 as tk3_ac_ex_db object having result with error named unrecognized arguments and status as failed. Thanks.

m4dcoder (approved):

Issue #5247, opened 2021-04-22
Fix projection queries for compatibility with Mongo DB 4.4


Kami:
``st2 execution list`` CLI command is failing out of the box when using MongoDB.

See two comments here for more info - https://github.com/StackStorm/st2/pull/5177#issuecomment-825200761, https://github.com/StackStorm/st2/pull/5177#issuecomment-825214833.

amanda11:
@Kami Is this issue now resolved, given the #5251 merge?

Kami:
Yep, sorry, forgot that the issue didn't auto close (forgot to reference it in the PR description).

PR #5251, opened 2021-05-01
Fix overlapping include field values for compatibility with Mongo DB 4.4

Kami:
This pull request includes a small fix needed for MongoDB 4.4 compatibility issue previously reported here - #5247.

Versions of MongoDB prior to 4.4 automatically and transparently handled overlapping values for projection queries (e.g. ``foo, bar, baz, bar.foo`` would get "translated" to ``foo, bar, baz``), but 4.4 throws if such values are provided for a projection query.

This pull request includes a small change which acts as a compatibility layer. We now perform "sanitization" of the include fields in the API side (aka if we see foo,bar,foo.a, we translate it to foo,bar).

In ideal world, we would throw a validation exception in such scenario, but this would change previous behavior and could break custom scripts or similar if someone is using overlapping values for ``?include_attributes`` query parameter value.

### DIFF
--- a/.github/workflows/microbenchmarks.yaml
+++ b/.github/workflows/microbenchmarks.yaml
@@ -3,12 +3,6 @@
 name: Micro Benchmarks
 
 on:
-  pull_request:
-    type: [opened, reopened, edited]
-    branches:
-      # Only for PRs targeting those branches
-      - master
-      - v[0-9]+.[0-9]+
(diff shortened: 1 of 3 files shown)


nzlosh (commented):
This looks good to me. :+1:

amanda11 (approved):
LGTM

PR #5252, opened 2021-05-02
Upgrade various deps to latest versions

Kami:
This PR upgrades more dependencies to the latest stable versions (gitpython, semver + dev / testing deps).

gitpython one was very outdated. I verified that installing packs from git still works correctly.


amanda11 (approved):
LGTM

PR #5253, opened 2021-05-02
Support special characters in CLI config parameters

### LINKED ISSUE #5180: Special characters in st2admin account causing st2 key failure
Just upgraded to version 3.4 and my keyvault is having problems. I believe it's due to my st2admin password containing special characters.

```
[root@stackstorm workflows]# st2 key list --scope=all
  File "/bin/st2", line 10, in <module>
    sys.exit(main())
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2client/shell.py", line 470, in main
    return Shell().run(argv)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2client/shell.py", line 385, in run
    config = self._parse_config_file(args=args, validate_config_permissions=False)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2client/base.py", line 183, in _parse_config_file
    result = parser.parse()
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2client/config_parser.py", line 197, in parse
    value = get_func(section, key)
  File "/usr/lib64/python3.6/configparser.py", line 800, in get
    d)
  File "/usr/lib64/python3.6/configparser.py", line 394, in before_get
    self._interpolate_some(parser, option, L, value, section, defaults, 1)
  File "/usr/lib64/python3.6/configparser.py", line 444, in _interpolate_some
    "found: %r" % (rest,))
configparser.InterpolationSyntaxError: '%' must be followed by '%' or '(', found: '%3C#V~Bvg%3E3t+'
```

This password above is what I used to install stackstorm. Or at least part of it. I've since changed the password via the documented htpasswd method, but the issue persists. Any tips? Left the password in for research purposes.

curl -sSL https://stackstorm.com/packages/install.sh | bash -s -- --user=st2admin --password='q7j/t%3C#V~Bvg%3E3t+'

winem:
This PR uses `ConfigParser.get(raw=True)` (see: https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.get) to read config parameters from the CLI without throwing an exception. 

Example of an exception with `raw=False` which is the default when setting a password for the admin user with the `%` in the password: 
```
20210502T203403+0000 Traceback (most recent call last):
20210502T203403+0000   File "/usr/bin/st2", line 10, in <module>
20210502T203403+0000     sys.exit(main())
20210502T203403+0000   File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2client/shell.py", line 470, in main
20210502T203404+0000     return Shell().run(argv)
20210502T203404+0000   File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2client/shell.py", line 385, in run
20210502T203404+0000     config = self._parse_config_file(args=args, validate_config_permissions=False)
20210502T203404+0000   File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2client/base.py", line 178, in _parse_config_file
20210502T203404+0000     result = parser.parse()
20210502T203404+0000   File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2client/config_parser.py", line 197, in parse
20210502T203404+0000     value = get_func(section, key)
20210502T203404+0000   File "/usr/lib/python3.6/configparser.py", line 800, in get
20210502T203404+0000     d)
20210502T203404+0000   File "/usr/lib/python3.6/configparser.py", line 394, in before_get
20210502T203404+0000     self._interpolate_some(parser, option, L, value, section, defaults, 1)
20210502T203404+0000   File "/usr/lib/python3.6/configparser.py", line 444, in _interpolate_some
20210502T203404+0000     "found: %r" % (rest,))
20210502T203404+0000 configparser.InterpolationSyntaxError: '%' must be followed by '%' or '(', found: '%3C#V~Bvg%3E3t+'
```
I came across this issue while investigating #5180 

I'll run some more tests once I can use the CircleCI artifacts for a quick setup with this change. This PR is WIP until I'm done with the tests.

close #5180


Kami:
So far things LGTM, but need unit tests, etc, yeah :)

And IIRC, ConfigParser supports % notation where you can reference other config section values. If we don't need this functionality, then "raw" should indeed work - we just need to make sure to document that in upgrade notes in case someone relied on that behavior (unlikely, but just in case).

Kami:
@winem 

> All tests passed, so I'd assume that we don't use the %notation to reference variables or further configuration values. And I'm actually not sure if that's something we should support. I see that this would allow some fancy configurations that may be useful but I also see a chance for (unintended) misconfiguration.

I didn't assume or mean that we use it (I assumed we don't), but technically a user could still be relying on this functionality (even though it's not explicitly documented in our docs, but it's pretty common and known functionality of Python config parser abstraction).

So we should still add upgrade notes entry for this change :)

armab (approved):
Thanks @winem ! :+1:

winem:
I did some more tests and added a comment to the tests to explain why we test for the % sign as well. 

Now it's good to be merged and no longer WIP.

PR #5255, opened 2021-05-07
Add migration script for migrating old execution field data to the new format

Kami:
This pull request adds a simple migration script which migrations data for executions which utilize old and slow + inefficient fielld type to the new type.

I tested it locally and it works fine (it's also a bit of a PITA to manually test, easiest is to just create MongoDB snapshot with old version of st2 and re-use that with st2 master).

## Implementation Details

#4846 introduced a new field types which is up to 10-30x faster when saving and retrieving objects with larger values.

In that PR I wasn't a massive fan of a migration script - yes I could add it, but I already spent many, many days on those performance optimization changes and I think migration script adds little value yet the effort is not totally trivial - yes the code changes mostly is, but things are not just a code change - it includes test cases, documentation, upgrade notes, etc and that takes quite some effort.

The migration script doesn't provide tons of value because it only works in old executions and most users only care about new and recent executions so if when viewing data for some old execution it takes longer to load it's not the end of the world.

I would be all of migration if that would actually provide more value - e.g. reduce maintenance burden / amount of code we need to maintain, but that is not the case and it likely won't change any time soon. We still utilize those two field types in many other places and migration script step is optional and manual and we can't really force users to run them and assume by the next release everyone has done it so we can just remove that code.

For now, I only implemented it for execution related objects. Technically we could also do it for workflow, trigger instance and rule related objects, but that would require even more work and some more "work around" to determine if specific object in database utilizes all type or now - it's possible to detect that, but it's not totally trivial and just re-writing all objects sames wasteful.

Complete contributions (with tests) for other models are also welcome (especially in this case where the change itself will likely save many 10's of thousands and likely even much more $ per month in terms of CPU utilization across all the large StackStorm installations so contributing such a small change seems worth trade off for such a big improvement :)).

## TODO

- [x] Unite tests (will be a bit of a PITA, but it's manageable)
- [x] Upgrade notes entry - https://github.com/StackStorm/st2/pull/5255


Kami:
I also pushed a change so we specify complete Python version for setup-python action and virtualenv cache.

Previously we didn't use complete version for virtualenv directory cache which means build started to fail once Github actions upgraded Python 3.8 from 3.8.9 to 3.8.10.

Example builds:

- https://github.com/StackStorm/st2/runs/2514019971?check_suite_focus=true (3.6.9 pre upgrade)
- https://github.com/StackStorm/st2/pull/5255/checks?check_run_id=2536076316 (3.6.10 post github actions upgrade - failing)

Technically, we could still only specify major + minor version, but we would still need cache per major + minor + patch version which means we would need another Github actions step which captured complete version post setup-python action which just adds additional complexity.

/cc @cognifloyd

Kami:
I also included eventlet bump in this PR - that security issue doesn't affect us directly since we don't utilize websocket functionality, but it's still a good idea to upgrade.

Kami:
OK, so I actually needed to revert eventlet upgrade since it breaks gunicorn eventlet worker - it seems there is a regression in 0.31.0 which breaks the worker (sadly no time to look in and contribute fix upstream).

guzzijones:
Kami is going to only pull the ids then iterate over to migrate each mongo entry. We noticed lots of ram used to query all action records . i will try again after the change

st2common/bin/migrations/v3.5/st2-migrate-db-dict-field-values
```diff
@@ -0,0 +1,306 @@
...
+        print("Migrating ActionExecutionDB with id %s" % (execution_id))
+
+        # This is a bit of a "hack", but it's the easiest way to tell mongoengine that a specific
+        # field has been updated and should be saved. If we don't do, nothing will be re-saved on
+        # .save() call due to mongoengine only trying to save what has changed to make it more
+        # efficient instead of always re-saving the whole object.
+        execution_db._mark_as_changed("result")
+        execution_db._mark_as_changed("result_size")
+
+        #  print(getattr(execution_db, "_changed_fields", []))
```
winem, line 107: Obsolete comment? Although it may remain as it may be useful for debugging in the future.

st2common/bin/migrations/v3.5/st2-migrate-db-dict-field-values
```diff
@@ -0,0 +1,306 @@
...
+                "Failed to migrate LiveActionDB with id %s, ignoring error (%s)"
+                % (liveaction_id, str(e))
+            )
+            # If liveaction for some reason doesn't exist (would likely represent corrupted data) we
+            # simply ignore that error since it's not fatal.
+            continue
+
+        liveaction_db._mark_as_changed("result")
+
+        # print(getattr(liveaction_db, "_changed_fields", []))
```
winem, line 126: Same here.

st2common/bin/migrations/v3.5/st2-migrate-db-dict-field-values
```diff
@@ -0,0 +1,306 @@
...
+        )
+        print("")
+
+
+def migrate_objects(display_prompt: bool = True) -> None:
+    print("StackStorm v3.5 database field data migration script\n")
+
+    if display_prompt:
+        input(
+            "You are stronly recommended to create database backup before running the "
```
winem, line 257: `strongly` misses its `g`.

st2common/bin/migrations/v3.5/st2-migrate-db-dict-field-values
```diff
@@ -0,0 +1,306 @@
...
+            "migration script.\n\nDepending on the number of the objects in the database, "
+            "migration may take multiple hours or more. You are recommended to start the the "
```
winem, line 259: Duplicate `the`

st2common/bin/migrations/v3.5/st2-migrate-db-dict-field-values
```diff
@@ -0,0 +1,306 @@
...
+
+def main():
+    if "--yes" in sys.argv:
+        sys.argv.remove("--yes")
+        display_prompt = False
+    else:
+        display_prompt = True
+
+    config.parse_args()
+    db_setup()
```
winem, line 291: Is it intended that exceptions while estabilishing the db connection aren't handled?

My first idea was that exceptions should be handled but it may be ok without  explicit exception handling since any error would immediately be visible to the user. So I don't have a strong opinion on this topic but catching exceptions and printing a message like the one below when an exception during migrate_objects is raised is probably a bit more user-friendly.

guzzijones:
`/opt/stackstorm/st2/bin/python st2-migrate-db-dict-field-values
Migrating affected database objects to utilize new field type
Migration execution objects
Found no ActionExecutionDB objects to migrate.
Migrating workflow objects
Will migrate 4 WorkflowExecutionDB objects

Migrating WorkflowExecutionDB with id 609437de5394ae4d4788ee83
WorkflowExecutionDB with id 609437de5394ae4d4788ee83 has been migrated

Migrating WorkflowExecutionDB with id 609438911ea56bf3eaa9d947
WorkflowExecutionDB with id 609438911ea56bf3eaa9d947 has been migrated

Migrating WorkflowExecutionDB with id 60943b567bfea2dde60a16c4
WorkflowExecutionDB with id 60943b567bfea2dde60a16c4 has been migrated

Migrating WorkflowExecutionDB with id 60943eb17bfea2dde60a16c7
WorkflowExecutionDB with id 60943eb17bfea2dde60a16c7 has been migrated

ABORTED: Objects migration aborted on first failure:
  File "st2-migrate-db-dict-field-values", line 239, in main
    migrate_objects()
  File "st2-migrate-db-dict-field-values", line 229, in migrate_objects
    migrate_workflow_objects()
  File "st2-migrate-db-dict-field-values", line 176, in migrate_workflow_objects
    print("Will migrate %s TaskExecutionDB objects" % (len(task_execution_dbs)))
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/mongoengine/queryset/queryset.py", line 62, in __len__
    list(self._iter_results())
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/mongoengine/queryset/queryset.py", line 110, in _iter_results
    self._populate_cache()
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/mongoengine/queryset/queryset.py", line 129, in _populate_cache
    self._result_cache.append(next(self))
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/mongoengine/queryset/base.py", line 1574, in __next__
    raw_doc = next(self._cursor)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/cursor.py", line 1207, in next
    if len(self.__data) or self._refresh():
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/cursor.py", line 1144, in _refresh
    self.__send_message(g)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/cursor.py", line 1001, in __send_message
    address=self.__address)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/mongo_client.py", line 1372, in _run_operation_with_response
    exhaust=exhaust)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/mongo_client.py", line 1471, in _retryable_read
    return func(session, server, sock_info, slave_ok)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/mongo_client.py", line 1366, in _cmd
    unpack_res)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/server.py", line 117, in run_operation_with_response
    reply = sock_info.receive_message(request_id)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/pool.py", line 726, in receive_message
    self._raise_connection_failure(error)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/pool.py", line 724, in receive_message
    return receive_message(self, request_id, self.max_message_size)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/network.py", line 214, in receive_message
    data = _receive_data_on_socket(sock_info, length - 16, deadline)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/network.py", line 280, in _receive_data_on_socket
    buf = bytearray(length)
MemoryError
`

guzzijones:
MemoryError. huh.  bet it ran out of memory

winem (changes requested):
Code looks good so far. Just had a few remarks on comments and exceptions on the db_setup may be catched. 

The only serious issue I couldn't test because I don't have enough data yet to run OoM is the memory issue reported by @guzzijones. This sounds like pagination is needed at least for the tasks (which is probably the db with the highest number of documents).

st2common/bin/migrations/v3.5/st2-migrate-db-dict-field-values
Kami: Yeah, I mostly left it there for debugging purposes, will add a comment.

Kami:
@winem I pushed a comment which I believe should address review feedback.

guzzijones:
```
Migrating trigger objects
ABORTED: Objects migration aborted on first failure: Cursor not found, cursor id: 131584450700, full error: {'ok': 0.0, 'errmsg': 'Cursor not found, cursor id: 131584450700', 'code': 43, 'codeName': 'CursorNotFound'}
  File "st2-migrate-db-dict-field-values", line 294, in main
    migrate_objects(display_prompt=display_prompt)
  File "st2-migrate-db-dict-field-values", line 271, in migrate_objects
    migrate_triggers()
  File "st2-migrate-db-dict-field-values", line 230, in migrate_triggers
    str(trigger_instance_db.id) for trigger_instance_db in trigger_instance_dbs
  File "st2-migrate-db-dict-field-values", line 230, in <listcomp>
    str(trigger_instance_db.id) for trigger_instance_db in trigger_instance_dbs
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/mongoengine/queryset/queryset.py", line 110, in _iter_results
    self._populate_cache()
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/mongoengine/queryset/queryset.py", line 129, in _populate_cache
    self._result_cache.append(next(self))
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/mongoengine/queryset/base.py", line 1574, in __next__
    raw_doc = next(self._cursor)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/cursor.py", line 1207, in next
    if len(self.__data) or self._refresh():
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/cursor.py", line 1144, in _refresh
    self.__send_message(g)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/cursor.py", line 1001, in __send_message
    address=self.__address)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/mongo_client.py", line 1372, in _run_operation_with_response
    exhaust=exhaust)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/mongo_client.py", line 1471, in _retryable_read
    return func(session, server, sock_info, slave_ok)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/mongo_client.py", line 1366, in _cmd
    unpack_res)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/server.py", line 135, in run_operation_with_response
    _check_command_response(first, sock_info.max_wire_version)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/helpers.py", line 162, in _check_command_response
    raise CursorNotFound(errmsg, code, response, max_wire_version)
pymongo.errors.CursorNotFound: Cursor not found, cursor id: 131584450700, full error: {'ok': 0.0, 'errmsg': 'Cursor not found, cursor id: 131584450700', 'code': 43, 'codeName': 'CursorNotFound'}
```

guzzijones:
Another issue I noticed is st2_migrate_db_dict_field_values.py just has one line in it and won't work when you run `python st2_mist2_migrate_db_dict_field_values.py` due to the dashes in st2-migrate-db-dict-field-values

I am not sure the purpose of this script.

Kami:
@guzzijones 

> Another issue I noticed is st2_migrate_db_dict_field_values.py just has one line in it and won't work when you run python st2_mist2_migrate_db_dict_field_values.py due to the dashes in st2-migrate-db-dict-field-values
>
> I am not sure the purpose of this script.

That's just a symlink for testing purposes (so we can import module with dashes in the name, alternative would be to use importlib).

It's only used by tests and nothing and nowhere else - it's not installed into ``/opt/stackstorm/st2/bin`` directory either.

guzzijones:
I still get a mongo crash when doing the triggers.
```
[2/4] Migrating WorkflowExecutionDB with id 609438911ea56bf3eaa9d947
WorkflowExecutionDB with id 609438911ea56bf3eaa9d947 has been migrated

[3/4] Migrating WorkflowExecutionDB with id 60943b567bfea2dde60a16c4
WorkflowExecutionDB with id 60943b567bfea2dde60a16c4 has been migrated

[4/4] Migrating WorkflowExecutionDB with id 60943eb17bfea2dde60a16c7
WorkflowExecutionDB with id 60943eb17bfea2dde60a16c7 has been migrated

Found no TaskExecutionDB objects to migrate.

Migrating trigger objects
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/pool.py", line 1180, in connect
    sock = _configured_socket(self.address, self.opts)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/pool.py", line 988, in _configured_socket
    sock = _create_connection(address, options)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/pool.py", line 972, in _create_connection
    raise err
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/pymongo/pool.py", line 965, in _create_connection
    sock.connect(sa)
ConnectionRefusedError: [Errno 111] Connection refused
```

guzzijones:
I would reccomend just pointing people to the purge page before running this.  The juice doesn't seem worth the squeeze on this one off migration script. 
https://docs.stackstorm.com/troubleshooting/purging_old_data.html?highlight=purge

guzzijones:
It is working now that I pruned 2021-02-05 and before
587845 triggers to migrate.

Kami:
> It is working now that I pruned 2021-02-05 and before
> 587845 triggers to migrate.

Great.

When I get a chance (hopefully some time this week), I will likely still implement two things:

* ``--from-ts`` and ``--to-ts`` flag - it will default to 30 days ago and "now". I already added first flag, just need to add second one. This way people will also be able to run it batches - e.g. a day at once or similar.
* Support for processing objects in batches of 10k items or so - this should help with the issue you saw with trigger instances. The query we perform takes a long time if there are a lot of items in the database, even if we just do ``.count()``, because it requires a full collection scan (and there isn't much else we can do to avoid that).

I will also update upgrade notes and add a note on purging old data.

And yes, as I mentioned in the pas, one of the reasons I wanted to avoid migration script is because it's not a totally trivial task if we want to correctly handle all the edge cases, etc. and (imo) it provides less value so additional effort is simply not worth it.

guzzijones:
I am trying to figure out why but I still am seeing chrome load up old executions with large json on the history page.

guzzijones:
I am going to rerun and a large json and see if it tries to load it.

guzzijones:
https://stackstorm-vf-dev.ifp.lmco.com/api/v1/executions/609bfafc35774f26bfb9f31c?max_result_size=102400

brought back a 3MB json and chrome is struggling to render the component to show the results still.

Kami:
@guzzijones Does that execution have ``result_size`` attribute in the collection document (you can use mongo shell to find out)?

Since that's what ``?max_result_size`` query parameter logic utilizes. As described in st2web PR, if that attribute is not present, we can't perform that filtering and full result will be returned (same as before).

Kami:
@guzzijones This should do the trick for you - 2a0969f64.

As you can see from the commit message and code, it include "one off" code change for your situation - once you confirm it works, I will remove it since that won't be needed anymore for other users.

I confirmed that change myself - ran mongo query to unset result size (``db.action_execution_d_b.update({}, {$unset: {result_size: 1}}, false, true)``) and then ran migration script.

guzzijones:
> @guzzijones Does that execution have `result_size` attribute in the collection document (you can use mongo shell to find out)?
> 
> Since that's what `?max_result_size` query parameter logic utilizes. As described in st2web PR, if that attribute is not present, we can't perform that filtering and full result will be returned (same as before).
> 
> EDIT: It could be that migration script doesn't set `result_size`, but I would need to double check. If that's the case, I will need to make a small change to the migration script.

This was a NEW execution I just ran so it should have the result_size attribute.

guzzijones:
This is a new execution i just ran with latest build of master.

> https://stackstorm-vf-dev.xxx.com/api/v1/executions/609bfafc35774f26bfb9f31c?max_result_size=102400
> 
> brought back a 3MB json and chrome is struggling to render the component to show the results still.

guzzijones:
st2-3.5dev-65.x86_64.rpm

Kami:
What happens if you manually run cURL command to retrieve execution details with ?max_result_size query param? And as per my comment above - inspect collection using mongo engine and make sure that execution contains ``result_size`` attribute.

I can't reproduce it locally - with latest master it works  fine for all the new (and migrated) executions.

guzzijones:
I will need to lookup how to query mongo directly.  Let me see if I can get the latest rpm installed, run a new workflow, and check the st2web result

Kami:
Something like this should work in mongo shell:

```bash
use st2;
db.action_execution_d_b.find({"_id": ObjectId("ID")}).pretty();
```

guzzijones:
```{ "_id" : ObjectId("609bfafc35774f26bfb9f31c"), "result_size" : 232 }```

it says the result size is 232?  This json is 3MB

**Comment by guzzijones** [comment]
fwiw this is a workflow not an action.

**Comment by Kami** [comment]
@guzzijones Ah, that does change everything - for workflow executions, I believe that actual execution result field will just contain what workflow explicitly publishes as output (but, I need to double check and confirm).

How does workflow work - does it publish the whole result / similar? A workflow which reproduces this issue would be ideal.

But it could also be that it works as expected and it's more an issue / feature request for st2web.

**Comment by guzzijones** [comment]
The problem is the input to the workflow is huge. And that is displayed via st2web.

**Comment by Kami** [comment]
@guzzijones If it's indeed related to input (workflow action parameter?) then this it has nothing to do with this functionality and migration script - this one only solves it for execution results (same for related st2web change).

Having said that, please still post all the context which is needed to reproduce it (workflow + action definitions + data / parameters which reproduces it).

But if it's indeed not related to the result attribute itself, then you will need to open a new issue (feature request) for it, likely against st2web repo.

**Comment by guzzijones** [comment]
yes it is a feature request against st2web.  Whatever you did for the result set we also need to do for input parameters.

**Comment by Kami** [comment]
That will probably be quite an involved change and likely also need to involve "support" on the server side if we want to implement it efficiently (aka adding ``parameters_size`` field or similar...).

So likely sadly nothing which will happen any time soon.

**Comment by Kami** [comment]
@guzzijones Also, have you re-ran the migration script yet with my temporary change (migrating executions without result_size field)?

Please let me know when you and finish running it so I can remove that temporary change.

**Comment by Kami** [comment]
@guzzijones Thanks for letting me know - will revert it.

**Comment by Kami** [comment]
@winem I believe all your feedback has been addressed and this is now ready for a final look.

File: st2common/bin/migrations/v3.5/st2-migrate-db-dict-field-values [hunk 6]
```diff
@@ -0,0 +1,460 @@
...
+from st2common.persistence.workflow import TaskExecution
+from st2common.persistence.trigger import TriggerInstance
+from st2common.exceptions.db import StackStormDBObjectNotFoundError
+from st2common.constants.action import LIVEACTION_COMPLETED_STATES
+from st2common.constants.triggers import TRIGGER_INSTANCE_COMPLETED_STATES
+
+# We retrieve and process up to this amount of object IDs in a single MongoDB query. This is done
+# to avoid potentially retrieving many 10s of millions of object ids in a single query
+# TODO: Implement batching --since-ts, --model=action,trigger,workflow flag
+BATCH_SIZE = 1000
```
winem, line 68: Looks like batching is not yet implemented and I guess this variable definition is obsolete? I'll approve the PR to unblock you and all those looking forward for the fix on the apt cache issue as this is just a minor issue.

winem (approved):

st2common/bin/migrations/v3.5/st2-migrate-db-dict-field-values
Kami: Yep, this is not needed anymore - won't implement it since --start-dt and --end-dt was implemented which can be used to achieve similar end result.

Will remove it.

PR #5262, opened 2021-05-10
Add redis configuration to st2.package.conf

m4dcoder:
Add redis configuration to st2.package.conf since redis is now included in the platform install and configured by default.


m4dcoder:
@armab The redis change is already documented in the changelog in an earlier PR.

armab:
@m4dcoder That's even better, thanks for the heads up :+1:

PR #5263, opened 2021-05-10
Remove duplicate proxy_set_header in nginx conf

m4dcoder:
There is a duplicate host header under the auth endpoint in the nginx config. Remove the duplicate proxy_set_header.


Kami:
Will this break diff patch application we have in st2 docker, I believe it will, but not 100%.

armab:
High chance, yeah

Issue #5265, opened 2021-05-12
SET Type is not JSON serializable


guzzijones:
## SUMMARY
`.toSet()` yaql is no longer serialized to json by the workflow engine.
Provide a quick summary of your bug report.
While running a workflow with large input json I stumbled upon this serialization error for the workflow engine.
```
st2workflowengine.log:2021-05-12 17:49:46,697 140687301693512 ERROR workflows [-] [609c121c04f287f5ebf421a5] {'type': 'error', 'message': 'TypeError: Type is not JSON serializable 
```

### STACKSTORM VERSION

st2 3.5dev (7af7122), on Python 3.6.8

##### OS, environment, install method

Centos 7

## Steps to reproduce the problem
Create a workflow with a yaql statement that outputs a set 
`.toSet()`

ie
```
version: 1.0
tasks:
  valid:
    action: core.noop
    next:
      - when: <%succeeded()%>
        publish:
          - valid: <% list(alist).toSet() %>
vars:
  - alist: [1,2,3,4]
```

## Expected Results
Workflow does not fail

## Actual Results
```
st2workflowengine.log:2021-05-12 17:49:46,697 140687301693512 ERROR workflows [-] [609c121c04f287f5ebf421a5] {'type': 'error', 'message': 'TypeError: Type is not JSON serializable  set', 'task_id': 'valid_emails', 'route': 1}
```
workflow fails

guzzijones:
It turns out orjson will not dump sets like the previous code did. 
 https://github.com/StackStorm/st2/blob/940219703f420df940938bfa342ba58c3a3ad85e/st2common/st2common/fields.py#L464

Kami:
Good catch - we should also start adding some tests for this functionality (aka serializing sets - it's weird we had no existing orquesta or similar test which would have caught that).

guzzijones:
I am not sure where to fix this?  It doesn't look like orjson can have a custom serializer for dict inputs.

guzzijones:
maybe this 'default' parameter: https://github.com/ijl/orjson#default I will try it

guzzijones:
yup that works. PR incoming

guzzijones:
I am going to add a unit test for this.

Kami:
Since this is a performance critical change, we also need a good set of micro benchmarks for it (can extend existing ones to cover new situations and various sizes with mixed types + sets).

[...]

PR #5267, opened 2021-05-13
Orjson serialize set to list

guzzijones:
orjson not serializing set from Yaql `.toSet()` function when publishing a variable.
Fixes issue #5625


Kami:
Thanks for the contribution.

Per https://github.com/StackStorm/st2/issues/5265#issuecomment-840741867, this is a performance critical change and will need more work.

Since none of the existing Orquesta and other tests caught this issue it looks like this functionality is not that widely used so we should understand how much overhead it adds (if any) and then decide how to proceed.

If it turns out it adds non-trivial amount of overhead, we should consider some other approach instead of using ``default`` for every single ``orjson.dumps`` call (aka only fall back to ``orjson.dumps`` with defaults if regular call without default raises an exception or similar) - especially, because, afaik, orquesta workflows are only place where ``set`` may be valid type, in other places it's always just native JSON types which doesn't include sets.

And to understand how much / if any overhead it adds, we need to add some micro-benchmarks at the very least.

There are already some examples you can use as a starting point. New micro benchmarks need to cover multiple scenarios so we can understand how it affects performance (small dict size, medium dict size, large dict with and having a set item in various places in the dict - e.g. deeply nested, top level attribute value, etc.).

st2common/tests/unit/test_db_fields.py
```diff
@@ -73,6 +73,16 @@ class ModelWithJSONDictFieldDB(stormbase.StormFoundationDB):
 
 
 class JSONDictFieldTestCase(unittest2.TestCase):
+    def test_set_to_mongo(self):
+        field = JSONDictField(use_header=False)
+        result = field.to_mongo({"test": {1, 2}})
```
Kami, line 78: Round trip test would also be good - aka to ensure that when we unserialize the value, we get a list back.

guzzijones:
Pretty sure it only runs this code if it bumps into a set. Can you point me
to the existing benchmark code?
The major issue is fixed with your zstd compression.
But alas I agree benchmarks will prove it.

Kami:
Can add it here - https://github.com/StackStorm/st2/blob/master/st2common/benchmarks/micro/test_json_serialization_and_deserialization.py#L32

Also, not sure what you mean with zstandard - that code is not actually used in prod.

It was one of the proposed approaches, but in the end we decided to go with a simpler approach which doesn't include field level compression (since MongoDB server already handles compression on the server aka storage size).

guzzijones:
The blob storage for funning workflows. That fixed most of my problems

Kami:
I've added a micro benchmark (fb2dce41e17a9ef503ccac2edcf283acbffa78ec) and results show that the overhead is indeed very small / negligible.

Having said that, since set is only valid in Orquesta workflow it still makes sense to only support it for Orquesta DB models.

Kami:
I added a round-trip test case and will ago ahead and merge it into master as-is.

I added a comment to the code and if it turns out if indeed adds more overhead in some other scenarios micro benchmarks don't cover, we can change the code then to only use ``default`` for Workflow related models.

Kami (approved):

Kami:
Merged, thanks again for catching this.

PR #5280, opened 2021-06-03
Update nginx config to support TLS v1.3 in addition to TLS v1.2

### LINKED ISSUE #5216: Enable TLS v1.3 support in the default nginx config
We should update nginx config to support TLS v1.3 in addition to TLS v1.2.

Since we still support Ubuntu 16.04, we can't do that easily for the installer (we could make it worth, but it's not the effort and complexity at this point), but we should at least do it for the docker setup (via nginx config patch) to begin with.

That image and nginx version should work just fine with TLS v1.3.

Kami:
This pull request updates production + sample nginx configs to also support TLS v1.3 in addition to TLS v1.2.

Keep in mind that TLS v1.3 will only be used if the server and client support it. On the server side, this means it will work out of the box on more recent distros where nginx version is >= v1.13 and nginx is compiled against OpenSSL v 1.1.1 which supports TLS v1.3.

Resolves #5216.


armab (approved):
Thanks a lot, @Kami :+1: 
and @cognifloyd, @punkrokk for assistance!

armab:
Thinking that this user-affecting change probably worth a small note in the upcoming Release Announcement

armab:
Do we also need a small remark in the https://docs.stackstorm.com/upgrade_notes.html ?

cognifloyd (approved):

punkrokk (approved):
Looks good. Regarding ciphers, anything < 256 should get removed. Not critical as I think browsers these days want the harder ciphers.

PR #5282, opened 2021-06-10
Remove xenial

amanda11:
Remove xenial from circleCI runs.


amanda11:
On purpose haven't added focal, as we need to re-structure circleci to do focal with 3.8.
This is blocking ST2 PR runs, so will stick to fixing runs with this PR.
Adding focal to be done under separate PR.

amanda11:
Closing - and using #5283 instead - which adds focal and removes xenial

PR #5283, opened 2021-06-10
Add focal and remove xenial

amanda11:
Update circleci to run packages for focal and remove xenial

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -107,7 +107,7 @@ jobs:
 
   # Build & Test st2 packages
   packages:
-    parallelism: 4
+    parallelism: 3
     # 4CPUs & 8GB RAM CircleCI machine
     # sadly, it doesn't work with 'setup_remote_docker'
     resource_class: large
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     # 'machine' executor runs Unit tests ~x1.5 faster, comparing to 'docker' executor
     # but the fastest is still ~x1.5-2 slower, comparing to Travis
     machine: true
-    parallelism: 4
+    parallelism: 3
```
blag, line 11:
```suggestion
    parallelism: 4
```

If we're adding focal at the same time that we're removing xenial, then let's leave this alone.

blag (changes requested):
Aside from one small, possibly necessary fix, this looks good. Thanks!

I'm assuming that it doesn't matter what version of Python our CircleCI jobs run in (currently using: `circleci/python:3.6` image), since they are packaged within Docker containers, and the Ubuntu Focal packaging container/s will be running Python 3.8.

.circleci/config.yml
amanda11: @blag Fixed...
Also yes my understanding is the same. It does the building and testing in the docker containers, and so the focal containers have python 3.8 so I believe this is all good. As the image is just what you run the essentially commands to pull up docker run in.

blag (approved):
👍

Issue #3356, opened 2017-04-16
Ports for api and authentication services can't be changed


hagay3:
Hi,
I'm trying to change the port for authentication module with no luck, the port remains on '9100' .
After the change I did also st2ctl restart.

I tried to edit this section, with new port 9109
```
[auth]
port = 9109
host = 0.0.0.0
use_ssl = True
```

Stackstrom process still listens on 9100 port
```
tcp        0      0 127.0.0.1:9100          0.0.0.0:*               LISTEN
tcp        0      0 127.0.0.1:9101          0.0.0.0:*               LISTEN
tcp        0      0 127.0.0.1:9102          0.0.0.0:*               LISTEN
```

I checked also the api and it's the same issue.

Kami:
To change listening ports on production installation (where installed using packages and services are managed by service manager and not running in a standalone mode), you need to change ports in the service manager files.

For example, for Ubuntu, that is in the upstart file - https://github.com/StackStorm/st2-packages/blob/master/packages/st2/debian/st2.st2api.upstart

Note to us: We should either document this or also utilize st2.conf when services are managed by a service manager.

hagay3:
I'm not sure what you meant by saying
"services are managed by service manager and not running in a standalone mode".
What is exactly standalone mode?

hagay3:
I still cant get how to change the port.
OS is Ubuntu 14.04

armab:
@hagay3 For Ubuntu14, edit the upstart init file located at: `/etc/init/st2auth.conf`.
You'll find the port definition there.

When you're done with modifying, don't forget to reload upstart configuration: `sudo initctl reload-configuration` and restart your service.

Hope that answers your question.

estee-tew:
https://stackstorm.slack.com/archives/community/p1492346619722875

hagay3:
Thanks @armab it did the magic.
I still think the one conf file should utilize the port as it does in many other products (mysql, cassandra etc..)

mbrannigan:
I was having this issue with CentOS 7 and fixed it this way:

1. Edit `/usr/lib/systemd/system/st2api.socket`
Change `ListenStream=127.0.0.1:9101` to `ListenStream=0.0.0.0:9101`

2. Reload systemd
`systemctl daemon-reload`

3. Reload st2api
`st2ctl restart-component st2api`

4. Check to see if it's re-bound
```ss -lnt | grep 9101
LISTEN     0      128          *:9101                     *:*
```

[...]

Issue #2676, opened 2016-05-06
st2api and st2auth always binds to 127.0.0.1


saahn:
I've set the `host` param in st2.conf for auth and api to `0.0.0.0`, but the config is ignored and the gunicorn process binds to 127.0.0.1:

```
[root@base-stackstorm2-138-onint ~]# ps -ef | grep gunicorn
st2      20711     1  9 22:50 ?        00:00:00 /opt/stackstorm/st2/bin/python /opt/stackstorm/st2/bin/gunicorn_pecan /opt/stackstorm/st2/lib/python2.7/site-packages/st2api/gunicorn_config.py -k eventlet -b 127.0.0.1:9101 --workers 1 --threads 1 --graceful-timeout 10 --timeout 30
st2      20723     1  9 22:50 ?        00:00:00 /opt/stackstorm/st2/bin/python /opt/stackstorm/st2/bin/gunicorn_pecan /opt/stackstorm/st2/lib/python2.7/site-packages/st2stream/gunicorn_config.py -k eventlet -b 127.0.0.1:9102 --workers 1 --threads 10 --graceful-timeout 10 --timeout 30
st2      20735     1  9 22:50 ?        00:00:00 /opt/stackstorm/st2/bin/python /opt/stackstorm/st2/bin/gunicorn_pecan /opt/stackstorm/st2/lib/python2.7/site-packages/st2auth/gunicorn_config.py -k eventlet -b 127.0.0.1:9100 --workers 1 --threads 1 --graceful-timeout 10 --timeout 30
...
```

I've tried modifying the hard-coded local addresses in both `https://github.com/StackStorm/st2/blob/master/st2api/st2api/gunicorn_config.py#L29` and `https://github.com/StackStorm/st2/blob/master/st2auth/st2auth/gunicorn_config.py#L29` 

as well as the init scripts `/etc/init.d/st2api`, `/etc/init.d/st2auth`, which was hard-coding the bind address in the DAEMON_ARGS variable: `DAEMON_ARGS="-k eventlet -b 127.0.0.1:9101 --workers 1 --threads 1 --graceful-timeout 10 --timeout 30 --pid ${PIDFILE}"`.

However, the auth and api endpoints are still binding to 127.0.0.1 after restarting the st2 processes.

I'm using st2 v1.4, and my environment was set up by following the [documentation](https://docs.stackstorm.com/install/deb.html).

My st2.conf looks like:

```
[exporter]
logging = /etc/st2/logging.exporter.conf

[...]

PR #5286, opened 2021-06-13
Add reload systemd to run unit generator for st2ctl

nzlosh:
This PR is part of https://github.com/StackStorm/st2-packages/pull/706 to use `st2.conf`to dynamically generate `st2api`, `st2auth` and `st2stream` socket files.

For this to work seamlessly for most users, an update to `st2ctl` is required to perform a `systemctl daemon-reload` which will regenerator the `.socket` files and pickup any modifications to `st2.conf`.

Fixes: https://github.com/StackStorm/st2/issues/3356 https://github.com/StackStorm/st2-packages/issues/686 https://github.com/StackStorm/st2/issues/2676


st2common/bin/st2ctl
```diff
@@ -108,6 +109,11 @@ function service_manager() {
   local svcname=$1 action=$2
   if [ -d /run/systemd/system ]; then
     # systemd is running
+    if [ -z $SYSTEMD_RELOADED ]; then
+      #Reload systemd to regenerate socket files from st2.conf
+      systemctl daemon-reload
+      SYSTEMD_RELOADED="yes"
```
armab, line 115: Considering `SYSTEMD_RELOADED` state is not saved in any file storage or passed externally, what's the idea here?

Will it run the `systemctl daemon-reload` on every new st2ctl run?

From that point, is the daemon-reload that triggers system-generator then?

armab:
Thanks, @nzlosh!

Great item for including in the v3.6.0 :+1:  Marked it as such.

st2common/bin/st2ctl
nzlosh: Yes, you've understood the logic correctly.  The `SYSTEMD_RELOADED` is to guard against reloading systemd for every daemon process `st2ctl` will restart for a single invocation.  The `systemctl daemon-reload` triggers the execution of the generators. This way the socket files are regenerated with each `st2ctl restart`.

While this may seem inefficient, because they likelihood of `st2.conf` being changed on each restart is low, the logic required to test `st2.conf` last modified time and compare it with the age the daemon process being restarted would have added unnecessary complexity.  The daemon-reload is ~200ms on my dev machine which is insignificant compare with the time it to restart all st2 daemon (12.848s on my dev machine).

For the sake of science here's the time it takes to restart each of the concerned daemons:
```
root@u1804:~# time systemctl restart st2api
real    0m4.014s
user    0m0.004s
sys     0m0.003s

root@u1804:~# time systemctl restart st2auth
real    0m1.879s
user    0m0.007s
sys     0m0.000s

root@u1804:~# time systemctl restart st2stream
real    0m1.483s
user    0m0.003s
sys     0m0.004s
```

cognifloyd (approved):

armab:
This PR comes with the https://github.com/StackStorm/st2-packages/pull/706 and should be merged only once the change in packaging is ready.

PR #5291, opened 2021-06-17
Add missing setup.py entry for the new migration script

Kami:
This PR adds missing setup.py scripts entry so the new migration script gets installed into virtualenv ``bin/`` directory.

https://github.com/StackStorm/discussions/issues/81#issuecomment-863227036

### DIFF
--- a/st2common/setup.py
+++ b/st2common/setup.py
@@ -63,6 +63,7 @@ setup(
         "bin/st2-pack-install",
         "bin/st2-pack-download",
         "bin/st2-pack-setup-virtualenv",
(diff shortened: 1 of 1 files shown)


amanda11:
@Kami Note, the python module also isn't present in the package at the moment.

Kami:
@amanda11 Python module is just a symlink to ``st2-migrate-db-dict-field-values`` and it's just needed for testing purposes (unit tests) and nothing else.

amanda11 (approved):
Verified that the migration script is now present in the RPM built from this PR.

Issue #5222, opened 2021-04-06
Race condition causing orquesta workflow status to change to running after it completed successfully


khushboobhatia01:
## SUMMARY

Race condition causing execution status to change to running after it completed successfully.
Action Execution logs:
```
log" : [ { "timestamp" : ISODate("2021-03-27T13:24:52.932Z"), "status" : "requested" }, { "timestamp" : ISODate("2021-03-27T13:24:54.938Z"), "status" : "scheduled" }, { "timestamp" : ISODate("2021-03-27T13:25:00.200Z"), "status" : "running" }, { "timestamp" : ISODate("2021-03-27T13:25:40.857Z"), "status" : "succeeded" }, { "timestamp" : ISODate("2021-03-27T13:25:43.470Z"), "status" : "running" } ]
```

Potential race condition causing issue when https://github.com/StackStorm/st2/blob/master/st2common/st2common/services/workflows.py#L980 completes before https://github.com/StackStorm/st2/blob/master/st2actions/st2actions/container/base.py#L380 is executed . When working with 1000s of executions, some delay is expected while running an action/updating records in DB which is leading to this race condition. We have been seeing this issue quite frequently now.

### STACKSTORM VERSION

Paste the output of ``st2 --version``:
st2 3.5dev, on Python 3.6.9

##### OS, environment, install method
Linux

Post what OS you are running this on, along with any other relevant information/

- e.g. Docker, Vagrant, Kubernetes, etc. Describe how you installed ST2
- e.g. one-line install, custom install, etc -->

## Steps to reproduce the problem

Show how to reproduce the problem, using a minimal test-case. Make sure to include any content
(pack content - workflows, actions, etc.) which are needed to reproduce the problem.

Added few sleep statements in the code to reproduce the race condition.

Add this before https://github.com/StackStorm/st2/blob/master/st2actions/st2actions/container/base.py#L380 `executions.update_execution(`
```
if liveaction_db.action_is_workflow==True:
         concurrency.sleep(seconds=120)
```

Add this before https://github.com/StackStorm/st2/blob/master/st2common/st2common/services/action.py#L80 `if parent_user:` (This is to avoid action execution to complete quickly which is expected when running large no. of executions)
```
concurrency.sleep(seconds=30)
```

Run any workflow. 
## Expected Results

Status transitioning should stop at succeeded even if the race condition occurs. Adding step to check if a transition is valid or not could help? 

## Actual Results

[...]

PR #5293, opened 2021-06-19
5222 - Skip action update if execution already in completed state.

khushboobhatia01:
Race condition causing orquesta workflow status to change to running after it completed successfully. To avoid this incorrect status update, skipping execution object update is execution is already in completed state.

More details about issue: https://github.com/StackStorm/st2/issues/5222

### DIFF
--- a/contrib/runners/orquesta_runner/tests/unit/test_error_handling.py
+++ b/contrib/runners/orquesta_runner/tests/unit/test_error_handling.py
@@ -998,11 +998,6 @@ class OrquestaErrorHandlingTest(st2tests.WorkflowTestCase):
         # Action execution result can contain dotted notation so ensure this is tested.
         result = {"127.0.0.1": {"hostname": "foobar"}}
 
-        tk1_ac_ex_db = ex_db_access.ActionExecution.query(
-            task_execution=str(tk1_ex_db.id)
-        )[0]
-        tk1_lv_ac_db = lv_db_access.LiveAction.get_by_id(tk1_ac_ex_db.liveaction["id"])
-        self.assertEqual(tk1_lv_ac_db.status, ac_const.LIVEACTION_STATUS_FAILED)
         self.assertDictEqual(tk1_lv_ac_db.result, result)
 
         # Manually handle action execution completion.
--- a/contrib/runners/orquesta_runner/tests/unit/test_error_handling.py
+++ b/contrib/runners/orquesta_runner/tests/unit/test_error_handling.py
@@ -993,6 +993,7 @@ class OrquestaErrorHandlingTest(st2tests.WorkflowTestCase):
         )[0]
         tk1_lv_ac_db = lv_db_access.LiveAction.get_by_id(tk1_ac_ex_db.liveaction["id"])
         self.assertEqual(tk1_lv_ac_db.context.get("user"), username)
+        self.assertEqual(tk1_lv_ac_db.status, ac_const.LIVEACTION_STATUS_FAILED)
 
         # Action execution result can contain dotted notation so ensure this is tested.
         result = {"127.0.0.1": {"hostname": "foobar"}}
--- a/contrib/runners/orquesta_runner/tests/unit/test_error_handling.py
+++ b/contrib/runners/orquesta_runner/tests/unit/test_error_handling.py
@@ -998,13 +998,6 @@ class OrquestaErrorHandlingTest(st2tests.WorkflowTestCase):
         # Action execution result can contain dotted notation so ensure this is tested.
         result = {"127.0.0.1": {"hostname": "foobar"}}
 
-        ac_svc.update_status(
-            tk1_lv_ac_db,
-            ac_const.LIVEACTION_STATUS_FAILED,
-            result=result,
-            publish=False,
-        )
-
         tk1_ac_ex_db = ex_db_access.ActionExecution.query(
             task_execution=str(tk1_ex_db.id)
         )[0]
--- a/contrib/runners/orquesta_runner/tests/unit/test_error_handling.py
+++ b/contrib/runners/orquesta_runner/tests/unit/test_error_handling.py
@@ -29,6 +29,7 @@ tests_config.parse_args()
 
 from tests.unit import base
 
+from local_runner import local_shell_command_runner
 from st2common.bootstrap import actionsregistrar
 from st2common.bootstrap import runnersregistrar
 from st2common.constants import action as ac_const
@@ -58,6 +59,12 @@ PACKS = [
     TEST_PACK_PATH,
     st2tests.fixturesloader.get_fixtures_packs_base_path() + "/core",
 ]
+
+RUNNER_RESULT_FAILED = (
+    ac_const.LIVEACTION_STATUS_FAILED,
+    {"127.0.0.1": {"hostname": "foobar"}},
+    {},
+)
 
 
 @mock.patch.object(
@@ -954,6 +961,11 @@ class OrquestaErrorHandlingTest(st2tests.WorkflowTestCase):
     @mock.patch.object(
         runners_utils, "invoke_post_run", mock.MagicMock(return_value=None)
     )
+    @mock.patch.object(
+        local_shell_command_runner.LocalShellCommandRunner,
+        "run",
+        mock.MagicMock(side_effect=[RUNNER_RESULT_FAILED]),
+    )
     def test_include_result_to_error_log(self):
         username = "stanley"
         wf_meta = base.get_wf_fixture_meta_data(TEST_PACK_PATH, "sequential.yaml")
@@ -981,9 +993,7 @@ class OrquestaErrorHandlingTest(st2tests.WorkflowTestCase):
         )[0]
         tk1_lv_ac_db = lv_db_access.LiveAction.get_by_id(tk1_ac_ex_db.liveaction["id"])
         self.assertEqual(tk1_lv_ac_db.context.get("user"), username)
-        self.assertEqual(tk1_lv_ac_db.status, ac_const.LIVEACTION_STATUS_SUCCEEDED)
 
-        # Manually override and fail the action execution and write some result.
         # Action execution result can contain dotted notation so ensure this is tested.
         result = {"127.0.0.1": {"hostname": "foobar"}}
 
--- a/contrib/runners/orquesta_runner/tests/unit/test_policies.py
+++ b/contrib/runners/orquesta_runner/tests/unit/test_policies.py
@@ -144,11 +144,11 @@ class OrquestaRunnerTest(st2tests.ExecutionDbTestCase):
             workflow_execution=str(wf_ex_db.id)
         )[0]
         t1_lv_ac_db = lv_db_access.LiveAction.query(task_execution=str(t1_ex_db.id))[0]
+        self.assertEqual(t1_lv_ac_db.status, ac_const.LIVEACTION_STATUS_FAILED)
         t1_ac_ex_db = ex_db_access.ActionExecution.query(
             task_execution=str(t1_ex_db.id)
         )[0]
 
-        t1_lv_ac_db = lv_db_access.LiveAction.query(task_execution=str(t1_ex_db.id))[0]
         t1_ac_ex_db = ex_db_access.ActionExecution.query(
             task_execution=str(t1_ex_db.id)
         )[0]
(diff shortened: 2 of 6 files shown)


m4dcoder (changes requested):
1. I don't completely recall but I think there may be some edge cases impacted by this. One test case I can think of is with items task in which one of the executions for an item failed, this leads to the workflow execution going into failed state. However, items that are still in progress are allowed to update result into the execution. I suggest getting CI jobs here to completely pass so we know it is not impacting other documented test cases.
2. Please make sure there's unit test associated with your change.
3. Have you tested this change locally on your workflows? Did it completely resolve?

khushboobhatia01:
> 1. I don't completely recall but I think there may be some edge cases impacted by this. One test case I can think of is with items task in which one of the executions for an item failed, this leads to the workflow execution going into failed state. However, items that are still in progress are allowed to update result into the execution. I suggest getting CI jobs here to completely pass so we know it is not impacting other documented test cases.
> 2. Please make sure there's unit test associated with your change.
> 3. Have you tested this change locally on your workflows? Did it completely resolve?

@m4dcoder 
>1: For failing unit test cases, statuses are being changed manually for testing a functionality which is not expected during normal execution. Can we directly change the action execution object status instead of using `update_execution()` in these cases?

>3: Yes, I've tested my changes locally with the race condition scenario and was able to resolve the issue. The status didn't change from `succeeded -> running` after the changes.

khushboobhatia01:
Please find response inline @m4dcoder

> 1. I don't completely recall but I think there may be some edge cases impacted by this. One test case I can think of is with items task in which one of the executions for an item failed, this leads to the workflow execution going into failed state. However, items that are still in progress are allowed to update result into the execution. I suggest getting CI jobs here to completely pass so we know it is not impacting other documented test cases.

----> All the test cases are fixed now. The one-off edge case you mentioned regarding with-items failure is verified with _test_with_items_failure_ test case which is running fine.
Screenshot 2021-07-14 at 6 54 21 PM


> 2. Please make sure there's unit test associated with your change.

----> Unit test case added _st2common/tests/unit/test_executions_util.py_


> 3. Have you tested this change locally on your workflows? Did it completely resolve?

----> Yes, I've verified the changes locally. I've setup an environment where I'm able to reproduce the race condition. 
Below screenshot is without my changes. You can see in the execution logs, status changed from running -> succeeded -> running.
Screenshot 2021-07-14 at 7 19 08 PM

With my changes, you can see the status not changing to running after it succeeded. Also, added screenshot of log line when execution update was skipped.
Screenshot 2021-07-14 at 7 04 37 PM
Screenshot 2021-07-14 at 7 05 37 PM

contrib/runners/orquesta_runner/tests/unit/test_rerun.py
```diff
@@ -216,10 +219,12 @@ def test_rerun_with_missing_workflow_execution_id(self):
...
         wf_db_access.WorkflowExecution.delete(wf_ex_db, publish=False)
 
-        # Manually delete the workflow_execution_id from context of the action execution.
+        # Manually delete the workflow_execution_id from context of the liveaction.
         lv_ac_db1.context.pop("workflow_execution")
         lv_ac_db1 = lv_db_access.LiveAction.add_or_update(lv_ac_db1, publish=False)
-        ac_ex_db1 = execution_service.update_execution(lv_ac_db1, publish=False)
+        # Manually delete the workflow_execution_id from context of the action execution.
+        ac_ex_db1.context.pop("workflow_execution")
+        ac_ex_db1 = ex_db_access.ActionExecution.add_or_update(ac_ex_db1, publish=False)
```
m4dcoder, line 227: Can you provide explanation here?  How come we cannot use execution_service.update_execution here?

contrib/runners/orquesta_runner/tests/unit/test_with_items.py
```diff
@@ -315,6 +328,11 @@ def test_with_items_concurrency(self):
         lv_ac_db = lv_db_access.LiveAction.get_by_id(str(lv_ac_db.id))
         self.assertEqual(lv_ac_db.status, action_constants.LIVEACTION_STATUS_SUCCEEDED)
 
+    @mock.patch.object(
+        local_shell_command_runner.LocalShellCommandRunner,
+        "run",
+        mock.MagicMock(return_value=RUNNER_RESULT_RUNNING),
+    )
```
m4dcoder, line 335: What happens to this test if we do not patch here?

st2common/st2common/services/executions.py
```diff
@@ -194,6 +194,15 @@ def update_execution(liveaction_db, publish=True, set_result_size=False):
                             on the "result_size" database field.
     """
     execution = ActionExecution.get(liveaction__id=str(liveaction_db.id))
+
+    # Skip execution object update when action is already in completed state.
+    if execution.status in action_constants.LIVEACTION_COMPLETED_STATES:
+        LOG.debug(
+            "[%s] Skipping execution update. Action execution state: %s"
```
m4dcoder, line 201: Can you also include the new status value that is being applied?

st2common/tests/unit/test_executions_util.py
```diff
@@ -171,6 +171,19 @@ def test_execution_update(self):
...
+        executions_util.create_execution_object(liveaction)
+        pre_update_status = liveaction.status
+        liveaction.status = "running"
+        executions_util.update_execution(liveaction)
+        execution = self._get_action_execution(
+            liveaction__id=str(liveaction.id), raise_exception=True
+        )
+        self.assertEqual(len(execution.log), 1)
+        # Check status is not updated if it's already in completed state.
+        self.assertEqual(execution.log[0]["status"], pre_update_status)
```
m4dcoder, line 185: Can you add a check to make sure that liveaction/action execution is already completed?

m4dcoder (changes requested):
Everything passes. But I have some more questions, specifically the patching in the unit tests.

contrib/runners/orquesta_runner/tests/unit/test_rerun.py
khushboobhatia01: By the time we come to execution_service.update_execution action is already in completed state [(#L212)](https://github.com/StackStorm/st2/pull/5293/files/3eac2c0092e3708bc1398e173f208b1f3e297e60#diff-456027f186e755cf87449fa1fa7373cdbd63e8faa091067b961987a6cdd0e37eL212) which is why popping of workflow id and and updating the execution object will not work with my changes. That's why I've manually updated the object to remove workflow id.

contrib/runners/orquesta_runner/tests/unit/test_with_items.py
khushboobhatia01: Below test case tries to cancel a task for which the task needs to be in running state. Before this patch, the status was being changed to RUNNING [here](https://github.com/StackStorm/st2/pull/5293/files#diff-4be2f900127f1b9ff8d34fcc0aeb2d585467d3a3c74ba881ebba07291d7c56f7R364). Which will not work with my changes if the task completed before reaching to this line. Therefore, the patch will ensure status being returned RUNNING.

contrib/runners/orquesta_runner/tests/unit/test_rerun.py
m4dcoder: Can you put what you describe here in the comment of the code changes?

m4dcoder (changes requested):
Overall I don't have any issue with the changes since all unit and integrate tests pass. I requested some minor cleanups. I will approve when cleanups are addressed and everything goes green.

contrib/runners/orquesta_runner/tests/unit/test_error_handling.py
```diff
@@ -981,9 +993,8 @@ def test_include_result_to_error_log(self):
         )[0]
         tk1_lv_ac_db = lv_db_access.LiveAction.get_by_id(tk1_ac_ex_db.liveaction["id"])
         self.assertEqual(tk1_lv_ac_db.context.get("user"), username)
-        self.assertEqual(tk1_lv_ac_db.status, ac_const.LIVEACTION_STATUS_SUCCEEDED)
+        self.assertEqual(tk1_lv_ac_db.status, ac_const.LIVEACTION_STATUS_FAILED)
```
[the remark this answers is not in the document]
khushboobhatia01: Addressed. Sorry I missed removing these in all the test cases.

contrib/runners/orquesta_runner/tests/unit/test_rerun.py
khushboobhatia01: Added a comment

khushboobhatia01:
> Overall I don't have any issue with the changes since all unit and integrate tests pass and the concern address from the change is valid. I requested some minor cleanups. I will approve when cleanups are addressed and everything goes green.

@m4dcoder All review comments addressed. Please approve and merge.

contrib/runners/orquesta_runner/tests/unit/test_with_items.py
```diff
@@ -580,12 +601,6 @@ def test_with_items_concurrency_pause_and_resume(self):
...
 
-        # Reset the action executions to running status.
-        for ac_ex in t1_ac_ex_dbs:
-            self.set_execution_status(
-                ac_ex.liveaction["id"], action_constants.LIVEACTION_STATUS_RUNNING
-            )
-
         t1_ac_ex_dbs = ex_db_access.ActionExecution.query(
             task_execution=str(t1_ex_db.id)
         )
```
m4dcoder, line 606: Since we are not updating the action executions manually, this query is redundant.

contrib/runners/orquesta_runner/tests/unit/test_with_items.py
```diff
@@ -491,12 +507,6 @@ def test_with_items_pause_and_resume(self):
...
```
m4dcoder, line 512: Since we are not updating the action executions manually, this query is redundant.

contrib/runners/orquesta_runner/tests/unit/test_with_items.py
```diff
@@ -417,12 +434,6 @@ def test_with_items_concurrency_cancellation(self):
...
```
m4dcoder, line 439: Since we are not updating the action executions manually, this query is redundant.

contrib/runners/orquesta_runner/tests/unit/test_with_items.py
```diff
@@ -340,12 +358,6 @@ def test_with_items_cancellation(self):
...
```
m4dcoder, line 363: Since we are not updating the action executions manually, this query is redundant.

contrib/runners/orquesta_runner/tests/unit/test_policies.py
```diff
@@ -136,13 +144,11 @@ def test_retry_policy_applied_on_workflow_failure(self):
...
         t1_ac_ex_db = ex_db_access.ActionExecution.query(
             task_execution=str(t1_ex_db.id)
         )[0]
 
-        # Manually set the status to fail.
-        ac_svc.update_status(t1_lv_ac_db, ac_const.LIVEACTION_STATUS_FAILED)
-        t1_lv_ac_db = lv_db_access.LiveAction.query(task_execution=str(t1_ex_db.id))[0]
         t1_ac_ex_db = ex_db_access.ActionExecution.query(
             task_execution=str(t1_ex_db.id)
         )[0]
```
m4dcoder, line 154: Since we are not updating the action executions manually, this query is redundant.

contrib/runners/orquesta_runner/tests/unit/test_error_handling.py
```diff
@@ -981,19 +993,11 @@ def test_include_result_to_error_log(self):
...
-        ac_svc.update_status(
-            tk1_lv_ac_db,
-            ac_const.LIVEACTION_STATUS_FAILED,
-            result=result,
-            publish=False,
-        )
-
         tk1_ac_ex_db = ex_db_access.ActionExecution.query(
             task_execution=str(tk1_ex_db.id)
         )[0]
```
m4dcoder, line 1003: Since we are not updating the action executions manually, this query is redundant.  Also, the LiveAction.get_by_id query below and the assert on the status.  Move the assert on the result up to where the status was checked?

m4dcoder (changes requested):
Almost there, still missing some cleanups. Can you check the unit tests that you've refactored and make sure there are no duplicate query/asserts on the live action and action execution? If there are asserts that are out of place after the cleanup, please move them closer to where they should be checked (i.e. move live action result check closer to where the status was checked).

m4dcoder (approved):
Good job!

PR #5298, opened 2021-07-03
Changed X-XSS-Protection to follow OWASP standards due to deprecation.

LiamRiddell:
Modified the `conf/nginx/st2.conf` to disable the `X-XSS-Protection` functionality in order to align with the OWASP standards. As most browser vendors have deprecated or removed this feature due to the fact it can introduce additional security issues.  Please view the following sources:

OWASP:
image

Mozilla Developer Network Web Docs:
image

References:
OWASP - https://owasp.org/www-project-secure-headers/#x-xss-protection
MDN - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection


Kami:
Seems reasonable to me although it would also be good to check the "legacy browsers" list and ensure it's indeed very old version with very little or no usage in real world.

LiamRiddell:
Agreed, I believe we could potentially use `Can I Use?` maintained service to get an idea of browser list. 
https://caniuse.com/mdn-http_headers_x-xss-protection

cognifloyd:
Do we need to replace X-XSS-Protection with CSP? https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP

cognifloyd:
@LiamRiddell Would you please add a changelog entry and replace `X-XSS-Protection` with `Content-Security-Policy`?

LiamRiddell:
@cognifloyd Hi Jacob, apologies. You want me to add a new changelog entry in the "CHANGELOG.rst" file in the root directory of the project? Best, Liam.

cognifloyd:
Correct

LiamRiddell:
> Correct

Do you want me to add it under the "in development" change log?

cognifloyd:
Yes. Then when we cut a release, that section gets renamed to the version number and we add a new in development section.

cognifloyd (approved):

cognifloyd:
I'm not sure why CircleCI isn't running this PR. I'm going to choose and reopen to trigger it (hopefully)

PR #5304, opened 2021-07-15
Modify API to remove related files on delete of actions/workflows

mahesh-orch:
This change is to remove related files of actions/workflows from disk when API called from CLI and web UI (https://github.com/StackStorm/st2web/pull/898). Currently when this delete API is invoked then this only de-register the actions from database and when user reloads the st2 then these actions get registered into database again.


nzlosh:
What issue does this PR relate to?

This is a big conceptual shift to StackStorm as a whole since the API does not create/modify or delete files in a pack because it is not authoritative, git is.  This change would break the fundamental premise of infrastructure as code and would have changes reverted on the next package upgrade creating a lot of confusion for the lay user.

amanda11:
I think this is related to adding delete functionality into the UI, so the reverse of the workflow composer + that creates a new workflow on disk and in the live pack.
I've not reviewed the code, but I believe that is the context...

However altering the current behaviour of the endpoint might not be the best course of action, as those already using the DELETE endpoint might rely on the fact it deletes from memory and not disk. So I wonder if it should have at least a new parameter to the call that the UI calls, that allows the default behaviour to be unchanged.

st2api/st2api/controllers/v1/actions.py
```diff
@@ -235,9 +236,20 @@ def delete(self, ref_or_id, requester_user):
             ref_or_id,
             action_db,
         )
+        
+        BASE_PATH = cfg.CONF.system.base_path
+        pack_name = action_db["pack"]
+        entry_point = action_db["entry_point"]
+        metadate_file = action_db["metadata_file"]
+        ACTION_PYTHON_FILE_PATH = os.path.join(BASE_PATH, "packs", pack_name, "actions", entry_point)
+        ACTION_METADATA_FILE_PATH = os.path.join(BASE_PATH, "packs", pack_name, metadate_file)
```
blag, line 245: Two nitpicks:

1. Not all action entrypoints are Python files.
2. Variable names in Python don't need to be uppercased.

```suggestion
        action_entrypoint_file_path = os.path.join(BASE_PATH, "packs", pack_name, "actions", entry_point)
        action_metadata_file_path = os.path.join(BASE_PATH, "packs", pack_name, metadate_file)
```

st2api/st2api/controllers/v1/actions.py
```diff
@@ -235,9 +236,20 @@ def delete(self, ref_or_id, requester_user):
...
 
         try:
             Action.delete(action_db)
+            if os.path.exists(ACTION_PYTHON_FILE_PATH):
+                os.remove(ACTION_PYTHON_FILE_PATH)
+            if os.path.exists(ACTION_METADATA_FILE_PATH):
+                os.remove(ACTION_METADATA_FILE_PATH)
```
blag, line 252:
```suggestion
            if os.path.exists(action_entrypoint_file_path):
                os.remove(action_entrypoint_file_path)
            if os.path.exists(action_metadata_file_path):
                os.remove(action_metadata_file_path)
```

blag (changes requested):
A few nitpicks that need to be cleaned up.

Also, this code will explode if two different worker processes try to remove the same pack at almost the same time, or if the process cannot remove the files for some reason (eg: permissions issues). Please wrap them in a try/catch block to ignore exceptions where the file is already deleted, and to write a log error if the process doesn't have permission to delete the file.

blag:
Just (hopefully) clarify a few points here.

> This is a big conceptual shift to StackStorm as a whole since the API does not create/modify or delete files in a pack because it is not authoritative, git is.

This is true, but removing the entire pack directory, `.git/` and all, is the ultimate authoritative action.

> I think this is related to adding delete functionality into the UI, so the reverse of the workflow composer + that creates a new workflow on disk and in the live pack.

As far as I'm aware, the workflow composer does not write the action metadata and workflow file out to disk, it only stores them in the database.

The overall problem here is that StackStorm allows workflows to be created three different ways:

1. Installed as part of a pack installed via infrastructure-as-code, like in a build pipeline, Dockerfile, Packer script, etc.
2. Installed as part of a pack installed by a ST2 worker, as part of the `packs.install` action.
3. Created manually in the workflow composer/designer.

File management is one of the integration points that the workflow composer/designer lacks.

amanda11:
@blag - agree with most of your points.

I am pretty sure it does write to disk on the composer. It certainly updates the files on edit - as have copied the modified files from the "live" file system to my gitrepo before.

There is logic in the modified file to write the data files out on the api create.

Rules - get written directly to database and not to file (by the UI), but I believe workflows get written to disk and to database - which is inconsistent!!!

m4dcoder:
This is related to delete of action/workflows in general. The current delete operation via API (invoked from CLI or UI related to https://github.com/StackStorm/st2web/pull/898) just deletes the database registration. If operator does a reload, the action/workflow that was deleted just gets registered to the database again. 

If user manages infra and packs with codes, I don't think this affects or changes that paradigm. They can still manage and load packs and actions thru git, puppet, chef, helm, or which ever way they automate their infra.

The target audiences for this change here are mostly action/workflow developers that is using a development instance of StackStorm to delete an action/workflow.

m4dcoder (changes requested):
@mahesh-orch

- Please update the description. The description should explain why you are making this change and how is the change going to work.
- Please make sure there are unit tests that cover the change.
- Can we add a confirmation prompt in the CLI to let user know that the operation will delete files and user can OK or back out of the operation?

mahesh-orch:
> 
> 
> A few nitpicks that need to be cleaned up.
> 
> Also, this code will explode if two different worker processes try to remove the same pack at almost the same time, or if the process cannot remove the files for some reason (eg: permissions issues). Please wrap them in a try/catch block to ignore exceptions where the file is already deleted, and to write a log error if the process doesn't have permission to delete the file.

Sure, I will do the suggested changes. Thanks.

mahesh-orch:
> 
> 
> @mahesh-orch
> 
>     * Please update the description. The description should explain why you are making this change and how is the change going to work.
> 
>     * Please make sure there are unit tests that cover the change.
> 
>     * Can we add a confirmation prompt in the CLI to let user know that the operation will delete files and user can OK or back out of the operation?

mahesh-orch:
@m4dcoder Sure, I will write the unit test to cover the made changes. And I will also work on asking for confirmation prompt in the CLI to let user know removal of files to proceed with users agreement or to back out the operation. Thanks

amanda11:
@m4dcoder if we are going to issue a warning on the CLI and print for confirmation, could we also add a --yes or similar so that you can confirm it on the command line - to make it easier for anyone scripting the cli calls.

mahesh-orch:
@blag I have added try except block for handling PermissionError while removing action files from disk. Also, code is added for logging errors. Please review. Thanks.

mahesh-orch:
@m4dcoder Added code to add a confirmation prompt in CLI to let user know that the operation will delete files on disk as well. If user says OK then to proceed otherwise to back out the operation. Please review. Thanks.

st2client/st2client/commands/resource.py
```diff
@@ -735,16 +736,29 @@ def __init__(self, resource, *args, **kwargs):
     def run(self, args, **kwargs):
         resource_id = getattr(args, self.pk_argument_name, None)
         instance = self.get_resource(resource_id, **kwargs)
-        self.manager.delete(instance, **kwargs)
+        if isinstance(instance, st2client.models.action.Action):
```
amanda11, line 739: Is it possible to have a cli argument like --yes so that it is possible to delete actions within having to give input. This makes it much easier for anyone using st2 cli in scripts.

amanda11 (changes requested):

st2api/st2api/controllers/v1/actions.py
```diff
@@ -235,9 +236,51 @@ def delete(self, ref_or_id, requester_user):
...
+        BASE_PATH = cfg.CONF.system.base_path
+        pack_name = action_db["pack"]
+        entry_point = action_db["entry_point"]
+        metadate_file = action_db["metadata_file"]
+        action_entrypoint_file_path = os.path.join(
+            BASE_PATH, "packs", pack_name, "actions", entry_point
+        )
+        action_metadata_file_path = os.path.join(
+            BASE_PATH, "packs", pack_name, metadate_file
+        )
```
m4dcoder, line 249: The pack base path can be overridden (ref: https://docs.stackstorm.com/packs.html#under-the-hood-pack-basics). Therefore, it is better to use get_pack_base_path @ https://github.com/StackStorm/st2/blob/master/st2common/st2common/content/utils.py#L135 to determine the pack file path.

st2api/st2api/controllers/v1/actions.py
```diff
@@ -235,9 +236,51 @@ def delete(self, ref_or_id, requester_user):
...
+                        'No permission to delete the "%s" file',
+                        action_metadata_file_path,
+                    )
+                except Exception as e:
+                    LOG.error(
+                        'Unable to delete "%s" file. Exception was "%s"',
+                        action_metadata_file_path,
+                        e,
+                    )
+
```
m4dcoder, line 283: Can you create a function at https://github.com/StackStorm/st2/blob/master/st2common/st2common/services/packs.py to delete the action entrypoint and action metadata files? Move the path calculation that you did above into this function as well.

st2client/st2client/commands/resource.py
m4dcoder: I agree, giving the option `-y` or `--yes` makes this scriptable.

m4dcoder: Also, you should move this changes to the ActionDeleteCommand at https://github.com/StackStorm/st2/blob/master/st2client/st2client/commands/action.py#L280 instead of using if isinstance to check if this resource is an Action.

m4dcoder (changes requested):
Where are the unit tests?

st2client/st2client/commands/resource.py
mahesh-orch: @amanda11 Sure, I will try to add functionality to have CLI argument --yes. Thanks.

mahesh-orch:
@amanda11 code is added to accept CLI argument `--yes` to `st2 action delete ref_or_id` command. `-y` argument is already used for yaml formatting. So, only `--yes` command has been added. Please review the changes. Thanks.

mahesh-orch:
@m4dcoder 
1. Code is added to use `get_pack_base_path` @ https://github.com/StackStorm/st2/blob/master/st2common/st2common/content/utils.py#L135 to determine the pack file path.
2. Function created at https://github.com/StackStorm/st2/blob/master/st2common/st2common/services/packs.py to delete the action entry_point and action metadata files. Also, moved path calculation code there.
3. Tried to move code added under ResourceDeleteCommand class at `/st2client/commands/resource.py` to ActionDeleteCommand class at https://github.com/StackStorm/st2/blob/master/st2client/st2client/commands/action.py#L280 but as ResourceDeleteCommand class is inherited by ActionDeleteCommand class via ContentPackResourceDeleteCommand class, it is complex to access method under ActionDeleteCommand class in ResourceDeleteCommand class.
4. Unit test added for added code at `/st2common/services/packs.py`
Please review the changes. Thanks.

st2common/st2common/services/packs.py
```diff
@@ -215,3 +218,43 @@ def search_pack_index(
...
+    if os.path.exists(action_metadata_file_path):
+        try:
+            os.remove(action_metadata_file_path)
+        except PermissionError:
+            LOG.error(
+                'No permission to delete the "%s" file',
+                action_metadata_file_path,
+            )
+        except Exception as e:
+            LOG.error(
```
amanda11, line 256: If we fail to delete the files from disk as we are catching the exceptions, does this not mean that the API is going to return success?
If we fail to delete from the database then the action appears to be to log.error, then in the API we return internal_server_error. But we can't do that if we suppress the exception here.


Would it be better to re-throw the exception so that in the api/controllers you can then return an internal server error?

amanda11 (changes requested):
The --yes looks good, but from what I read of the code if the delete to delete from filesystem then the API will still return success, and only failure if it fails to deelte from db? This would be confusing in my opinion.

mahesh-orch:
@amanda11 added code to return `internal_server_error` if not able to delete files from disk. If it failed to delete action file except `PermissionError` then also `internal_server_error` will be thrown with message file still exists. And if file doesn't exists it will give action not found message as there will be no registered action in database. This is because action delete API call is de-registering action from database first and then we are removing file from disk. Once the action gets de-registered from database and we try to invoke action delete API then it gives message that action is not found.
Please review the changes. Thanks.

mahesh-orch:
I couldn't figure out why CI / Integration Tests (Orquesta) tests are failing.

st2client/st2client/commands/resource.py
```diff
@@ -731,20 +732,47 @@ def __init__(self, resource, *args, **kwargs):
 
         self.parser.add_argument(argument, metavar=metavar, help=help)
 
+        self.parser.add_argument(
+            "--yes",
```
m4dcoder, line 736: Can you add an option for `-y`?

st2client/st2client/commands/resource.py
```diff
@@ -731,20 +732,47 @@ def __init__(self, resource, *args, **kwargs):
...
+        if args.yes:
+            self.manager.delete(instance, **kwargs)
+            print(
+                'Resource with id "%s" has been successfully deleted from database and disk.'
+                % (resource_id)
+            )
+        else:
+            if isinstance(instance, st2client.models.action.Action):
+                user_input = input(
+                    "It will delete action files on disk as well. Do you want to continue? (y/n): "
```
m4dcoder, line 754: Can you change the message to the following? 

```
The resource files on disk will be deleted. Do you want to continue? (y/n): 
```

st2common/tests/unit/services/test_packs.py
```diff
@@ -0,0 +1,57 @@
...
+        """
+        Test that the action files present in the pack and removed
+        on the call of delete_action_files_from_pack function.
+        """
+
+        self.assertTrue(os.path.exists(self.entry_point))
+        self.assertTrue(os.path.exists(self.metadata_file))
+        delete_action_files_from_pack(TEST_PACK, self.entry_point, self.metadata_file)
+        self.assertFalse(os.path.exists(self.entry_point))
+        self.assertFalse(os.path.exists(self.metadata_file))
```
m4dcoder, line 57: Please add unit tests to cover if path does not exists and cover the try/except blocks on permission and other errors for both entry point and metadata files.

st2common/st2common/services/packs.py
```diff
@@ -215,3 +221,68 @@ def search_pack_index(
...
+                'No permission to delete the "%s" file',
+                action_entrypoint_file_path,
+            )
+            msg = 'No permission to delete "{0}" file from disk'.format(
+                action_entrypoint_file_path
+            )
+            abort(
+                http_client.INTERNAL_SERVER_ERROR,
+                six.text_type(msg),
+            )
```
m4dcoder, line 249: Please throw an exception here instead of doing a http abort.  Let the API controller do the abort for the http request on exception.

st2common/st2common/services/packs.py
```diff
@@ -215,3 +221,68 @@ def search_pack_index(
...
+                'Unable to delete "%s" file. Exception was "%s"',
+                action_entrypoint_file_path,
+                e,
+            )
+            msg = (
+                'Delete operation unsuccessful. "{0}" file still exists on disk'.format(
+                    action_entrypoint_file_path
+                )
+            )
+            abort(http_client.INTERNAL_SERVER_ERROR, six.text_type(msg))
```
m4dcoder, line 261: Please throw an exception here instead of doing a http abort. Let the API controller do the abort for the http request on exception.

st2common/st2common/services/packs.py
```diff
@@ -215,3 +221,68 @@ def search_pack_index(
...
+                'No permission to delete the "%s" file',
+                action_metadata_file_path,
+            )
+            msg = 'No permission to delete "{0}" file from disk'.format(
+                action_metadata_file_path
+            )
+            abort(
+                http_client.INTERNAL_SERVER_ERROR,
+                six.text_type(msg),
+            )
```
m4dcoder, line 277: Please throw an exception here instead of doing a http abort. Let the API controller do the abort for the http request on exception.

st2common/st2common/services/packs.py
```diff
@@ -215,3 +221,68 @@ def search_pack_index(
...
+            LOG.error(
+                'Unable to delete "%s" file. Exception was "%s"',
+                action_metadata_file_path,
+                e,
+            )
+            if os.path.isfile(action_metadata_file_path):
+                msg = 'Delete operation unsuccessful. "{0}" file still exists on disk'.format(
+                    action_metadata_file_path
+                )
+                abort(http_client.INTERNAL_SERVER_ERROR, six.text_type(msg))
```
m4dcoder, line 288: Please throw an exception here instead of doing a http abort. Let the API controller do the abort for the http request on exception.

st2client/st2client/commands/resource.py
```diff
@@ -731,20 +732,47 @@ def __init__(self, resource, *args, **kwargs):
...
+                        % (resource_id)
+                    )
+                else:
+                    print("Action is not deleted.")
+            else:
+                self.manager.delete(instance, **kwargs)
+                print(
+                    'Resource with id "%s" has been successfully deleted.'
+                    % (resource_id)
+                )
```
m4dcoder, line 769: By putting this in the ResourceDeleteCommand, this will affect other pack contents (not just action) and expanded the scope for this PR. So will you be expanding the scope?  If so, please make sure pack contents such as rules, policies, etc. are going to be deleted from disk properly as well.

m4dcoder (changes requested):
Let's address all other comments before we revisit the orquesta CI failure. It may be a temporary failure.

st2client/st2client/commands/resource.py
mahesh-orch: > 
> 
> Can you add an option for `-y`?

`-y` argument is already used for yaml formatting at https://github.com/StackStorm/st2/blob/master/st2client/st2client/commands/resource.py#L163. So it is conflicting. If any other suitable option is there as an argument, I will add it.

mahesh-orch:
@m4dcoder 
I have added unit tests to cover if path does not exists and cover the try/except blocks on permission and other errors for both entry point and metadata files.
I have added code at said locations to throw an error instead of doing a http abort.
Please review the changes. Thanks.

st2client/st2client/commands/resource.py
mahesh-orch: @m4dcoder 
Moving code under ResourceDeleteCommand class at /st2client/commands/resource.py to ActionDeleteCommand class at https://github.com/StackStorm/st2/blob/master/st2client/st2client/commands/action.py#L280 causes problems like if we import code from action.py to resource.py it gives import errors for other modules, it is probably because of circular imports. It becomes complex to call method from child class to parent class. Also, there is code under ResourceDeleteCommand on which test cases are dependent.

As we are handling condition if isinstance(instance, Action) therefore as per my understanding we are not affecting other pack contents like rules, policies etc. and if the resource is not instance of Action then flow is exactly same as it was before making the changes.

But still if this code need to be moved, please suggest me another appropriate place to move code under ResourceDeleteCommand to that location. Thanks.

m4dcoder: We shouldn't be handling the condition for sub type using if isinstance(instance, Action) in a generic class. The ActionDeleteCommand is there to override logic for specific sub type. I advise that you revisit putting this under ActionDeleteCommand. Can you post the import errors here?

st2common/st2common/exceptions/packs.py
```diff
@@ -0,0 +1,25 @@
...
+from __future__ import absolute_import
+from st2common.exceptions import StackStormBaseException
+
+__all__ = [
+    "ResourceDiskFilesRemovalError",
+]
+
+
+class ResourceDiskFilesRemovalError(StackStormBaseException):
+    pass
```
m4dcoder, line 25: Can you move this custom exception to https://github.com/StackStorm/st2/blob/master/st2common/st2common/exceptions/content.py?

st2common/tests/unit/services/test_packs.py
```diff
@@ -0,0 +1,193 @@
...
+        metadata_file = os.path.join(TEST_PACK_PATH, "actions", "test_metadata.yaml")
+
+        remove.side_effect = PermissionError("No permission to delete file from disk")
+
+        # asserting metadata files exists
+        self.assertTrue(os.path.exists(metadata_file))
+
+        # asserting delete_action_files_from_pack function raises PermissionError
+        with self.assertRaises(PermissionError):
+            delete_action_files_from_pack(TEST_PACK, entry_point, metadata_file)
```
m4dcoder, line 163: How do we know this is the error for the metadata file?

st2common/tests/unit/services/test_packs.py
```diff
@@ -0,0 +1,193 @@
...
+        metadata_file = os.path.join(TEST_PACK_PATH, "actions", "test_metadata.yaml")
+
+        remove.side_effect = PermissionError("No permission to delete file from disk")
+
+        # asserting metadata files exists
+        self.assertTrue(os.path.exists(entry_point))
+
+        # asserting delete_action_files_from_pack function raises PermissionError
+        with self.assertRaises(PermissionError):
+            delete_action_files_from_pack(TEST_PACK, entry_point, metadata_file)
```
m4dcoder, line 148: How can we tell this is the error for the entry point file?

st2common/tests/unit/services/test_packs.py
```diff
@@ -0,0 +1,193 @@
...
+        metadata_file = os.path.join(TEST_PACK_PATH, "actions", "test_metadata.yaml")
+
+        remove.side_effect = Exception("Another exception occured")
+
+        # asserting metadata files exists
+        self.assertTrue(os.path.exists(entry_point))
+
+        # asserting delete_action_files_from_pack function raises exception
+        with self.assertRaises(Exception):
+            delete_action_files_from_pack(TEST_PACK, entry_point, metadata_file)
```
m4dcoder, line 178: How can we tell this is the error for the entry point file?

st2common/tests/unit/services/test_packs.py
```diff
@@ -0,0 +1,193 @@
...
+        metadata_file = os.path.join(TEST_PACK_PATH, "actions", "test_metadata.yaml")
+
+        remove.side_effect = Exception("Another exception occured")
+
+        # asserting metadata files exists
+        self.assertTrue(os.path.exists(metadata_file))
+
+        # asserting delete_action_files_from_pack function raises exception
+        with self.assertRaises(Exception):
+            delete_action_files_from_pack(TEST_PACK, entry_point, metadata_file)
```
m4dcoder, line 193: How do we know this is the error for the metadata file?

m4dcoder (changes requested):

st2client/st2client/commands/resource.py
m4dcoder: How about `-f` or `--force` to be consistent with `rm` command?

st2client/st2client/commands/resource.py
mahesh-orch: I realized the mistake I was doing. I was creating another function under ActionDeleteCommand and importing it in resource.py to use under ResourceDeleteCommand. Now I have shifted the whole code along with run and print_and_run methods required to delete action files under ActionDeleteCommand. Which is working fine.

mahesh-orch:
@m4dcoder 
I have shifted code from reource.py under ResourceDeleteCommand to action.py under ActionDeleteCommand.
Moved custom exception ResourceDiskFilesRemovalError to https://github.com/StackStorm/st2/blob/master/st2common/st2common/exceptions/content.py
Updated the unit tests under /st2common/tests/unit/services/test_packs.py
Please review the changes. Thanks.

st2client/st2client/commands/action.py
```diff
@@ -278,7 +279,37 @@ class ActionDisableCommand(resource.ContentPackResourceDisableCommand):
...
+                % (resource_id)
+            )
+        else:
+            user_input = input(
+                "The resource files on disk will be deleted. Do you want to continue? (y/n): "
+            )
+            if user_input.lower() == "y" or user_input.lower() == "yes":
+                self.manager.delete(instance, **kwargs)
+                print(
+                    'Resource with id "%s" has been successfully deleted from db and disk.'
```
m4dcoder, line 299: Minor fix request. Can you change `db` to `database` or vice versa so the message is the same as the one above? Maybe use a variable to dedupe this message?

st2client/st2client/commands/action.py
```diff
@@ -278,7 +279,37 @@ class ActionDisableCommand(resource.ContentPackResourceDisableCommand):
...
+                print("Action is not deleted.")
+
+    def run_and_print(self, args, **kwargs):
+        resource_id = getattr(args, self.pk_argument_name, None)
+
+        try:
+            self.run(args, **kwargs)
+        except ResourceNotFoundError:
+            self.print_not_found(resource_id)
+            raise OperationFailureException("Resource %s not found." % resource_id)
```
m4dcoder, line 312: Do we need to reraise exception here? Didn't we print not found already? If we have to reraise this, how come we can use reraise the original exception?

st2client/st2client/commands/resource.py
m4dcoder: Can we change to -f and --force?

m4dcoder (changes requested):
Still got some minor changes to take care of.

mahesh-orch:
@m4dcoder I have added `-f` and `--force` flags instead of `--yes` flag for action delete command to auto delete resource files. Also, done said minor changes. Please review the changes. Thanks.

st2client/st2client/commands/resource.py
```diff
@@ -731,6 +731,18 @@ def __init__(self, resource, *args, **kwargs):
...
+            action="store_true",
+            help="Auto yes flag to delete action files from disk.",
+        )
+
+        self.parser.add_argument(
+            "--force",
+            action="store_true",
+            help="Auto yes flag to delete action files from disk.",
+        )
+
```
m4dcoder, line 745: These CLI parser arguments need to be moved to the ActionDeleteCommand. By putting this in the generic ResourceDeleteCommand, these arguments are also showing up for other resources (i.e. `st2 rule delete -h`).  Please review other commands at https://github.com/StackStorm/st2/blob/master/st2client/st2client/commands/action.py for examples to see how to move these arguments to the ActionDeleteCommand.

m4dcoder (changes requested):

mahesh-orch:
@m4dcoder I have moved CLI parser arguments to ActionDeleteCommand. Please review changes. Thanks.

st2client/st2client/commands/action.py
```diff
@@ -277,8 +277,65 @@ class ActionDisableCommand(resource.ContentPackResourceDisableCommand):
...
+            **kwargs,
+        )
+
+        argument = self.pk_argument_name
+        metavar = self._get_metavar_for_argument(argument=self.pk_argument_name)
+        help = self._get_help_for_argument(
+            resource=resource, argument=self.pk_argument_name
+        )
+
+        self.parser.add_argument(argument, metavar=metavar, help=help)
```
m4dcoder, line 298: Why are lines 281, 292-298 here needed? If you change the base class that ActionDeleteCommand inherits from to resource.ResourceDeleteCommand, the __init__ can be simplified.

m4dcoder: Try the following instead

```
class ActionDeleteCommand(resource.ResourceDeleteCommand):
    def __init__(self, resource, *args, **kwargs):
        super(ActionDeleteCommand, self).__init__(resource, *args, **kwargs)

        self.parser.add_argument(
            "-f",
            action="store_true",
            help="Auto yes flag to delete action files from disk.",
        )

        self.parser.add_argument(
            "--force",
            action="store_true",
            help="Auto yes flag to delete action files from disk.",
        )
```

m4dcoder (changes requested):

mahesh-orch:
@m4dcoder As per your input I have simplified the constructor in ActionDeleteCommand. Only one different thing I did is I have inherited ContentPackResourceDeleteCommand class instead of ResourceDeleteCommand and it is working fine. Please review. Thanks.

st2client/st2client/commands/action.py
```diff
@@ -278,7 +278,48 @@ class ActionDisableCommand(resource.ContentPackResourceDisableCommand):
...
+            "-f",
+            action="store_true",
+            help="Auto yes flag to delete action files from disk.",
+        )
+
+        self.parser.add_argument(
+            "--force",
+            action="store_true",
+            help="Auto yes flag to delete action files from disk.",
+        )
```
m4dcoder, line 294: I noticed another thing here. Please collapse the two arguments into one. `-f` is just an abbreviation for `--force`.  You can follow this for example: https://github.com/StackStorm/st2/blob/master/st2client/st2client/commands/resource.py#L155. The end result should look like the following.
```
vagrant@axon:~/st2$ st2 action delete -h
usage: st2 action delete [-h] [-t TOKEN] [--api-key API_KEY] [-j] [-y] [-f]
                         ref-or-id

Delete an existing action.

positional arguments:
  ref-or-id             Reference or ID of the action.

optional arguments:
  -t TOKEN, --token TOKEN
                        Access token for user authentication. Get
                        ST2_AUTH_TOKEN from the environment variables by
                        default.
  --api-key API_KEY     Api Key for user authentication. Get ST2_API_KEY from
                        the environment variables by default.
  -j, --json            Print output in JSON format.
  -y, --yaml            Print output in YAML format.
  -f, --force           Auto yes flag to delete action files from disk.
```

m4dcoder (changes requested):

mahesh-orch:
@m4dcoder I have combined `-f` and `--force` arguments. `st2 action delete -h` command result is looking exactly same as you mentioned above.

m4dcoder (approved):

st2api/st2api/controllers/v1/actions.py
```diff
@@ -236,8 +237,17 @@ def delete(self, ref_or_id, requester_user):
...
+        metadata_file = action_db["metadata_file"]
+
         try:
             Action.delete(action_db)
+            delete_action_files_from_pack(
+                pack_name=pack_name,
+                entry_point=entry_point,
+                metadata_file=metadata_file,
+            )
         except Exception as e:
```
amanda11, line 251: As the Action.delete and delete_action_files_from disk are in the same catch exception. If I get an error on deleting the files but not on the database, doesn't this mean that I'm still going to get a LOG.error saying "Database delete encountered exception"?

This would be a bit confusing. So either we need to amend the text slightly to indicate it might be database or file, or have two catch blocks.

amanda11 (changes requested):
Minor comment about logging when delete from disk fails.

amanda11:
@m4dcoder One other point as this is a change in the way the existing delete API works, we will need to make sure that it gets documented as a breaking change in the documentation.
Can we get a CHANGELOG added to indicate its a breaking change, as I don't think we have a changelog entry yet for this PR?

We don't know whether anyone using the existing API is reliant on it not deleting the workflow on disk, so need to make this as clear as possible - and perhaps start the st2docs update for the upgrade notes now so it doesn't get forgotten?

The alternative would be to keep it backward compatible by introducing a new boolean parameter on the API call such as delete_on_disk, and the CLI and UI set that to true. And existing direct API users would then not be affected, but we'd keep this new behaviour for UI and CLI users.

But I get the feeling that others think that is too complicated, and I get that there is a balance between over-complicating code and backwards-compatability - so I am just mentioning it rather than putting as a blocking review comment.

m4dcoder:
@amanda11 Can we put this change out first? If there are complaints about this new behavior, we can add the `delete_on_disk` parameter as you suggested. I don't want to overthink here and add feature if no one is really going to use it.

amanda11:
@m4dcoder As mentioned I know there's a balance, so as long as we make it clear in CHANGELOG and in upgrade notes that its not backwards compatible. (It's one of those things if we don't make clear in changelog or upgrade notes - we might forget at release time).
 
So the things that are currently blocking my review are:

- CHANGELOG is missing
- the exception message when there is an error deleting from disk, will result in a second error stating there is an error deleting from db - when the delete db will have worked.

st2api/st2api/controllers/v1/actions.py
m4dcoder: @mahesh-orch Please address this.

st2client/st2client/commands/action.py
```diff
@@ -278,7 +278,44 @@ class ActionDisableCommand(resource.ContentPackResourceDisableCommand):
...
+                "The resource files on disk will be deleted. Do you want to continue? (y/n): "
+            )
+            if user_input.lower() == "y" or user_input.lower() == "yes":
+                self.manager.delete(instance, **kwargs)
+                print(resource_delete_msg)
+            else:
+                print("Action is not deleted.")
+
+    def run_and_print(self, args, **kwargs):
+        resource_id = getattr(args, self.pk_argument_name, None)
```
blag, line 313: [This line](https://github.com/StackStorm/st2/blob/5587123977142467fba20eaa22edc3dbea248fa8/st2client/st2client/commands/resource.py#L127) will guarantee that `self.pk_argument_name` will at least be `None`.

And [this line](https://github.com/StackStorm/st2/blob/5587123977142467fba20eaa22edc3dbea248fa8/st2client/st2client/commands/resource.py#L758) will guarantee that it will be `"ref_or_id"`.

So you don't need the third argument to `getattr()` here.

st2common/tests/unit/services/test_packs.py
```diff
@@ -0,0 +1,213 @@
...
+import mock
+import unittest2
+
+import st2tests
+
+from st2common.services.packs import delete_action_files_from_pack
+
+TEST_PACK = "dummy_pack_1"
+TEST_PACK_PATH = (
+    st2tests.fixturesloader.get_fixtures_packs_base_path() + "/" + TEST_PACK
```
blag, line 31: Why not use `pathlib.Path(...) / TEST_PACK` or at least `os.path.join()` here?

st2common/tests/unit/services/test_packs.py
blag: Please add an assert for the message of the mocked `PermissionError`.

st2api/st2api/controllers/v1/actions.py
cognifloyd: Could you put the try/except blocks one after the other instead of nesting them?

st2common/tests/unit/services/test_packs.py
blag: Please add an assert for the message of the mocked `PermissionError`.

blag (changes requested):
Still needs some work. Please use descriptive commit messages and minimize the number of commits made.

cognifloyd (changes requested):
Nice. This is a great addition.

Can you make the exception tests more specific (see @m4dcoder comments)

CHANGELOG.rst
```diff
@@ -282,6 +282,16 @@ Changed
 
 * Update orquesta to v1.4.0.
 
+* Modified action delete api. Action delete api removes related action/workflow files on disk
```
amanda11, line 285: This has been added to the changes made to 3.5.0. It should be in the "in development" section at the top.

amanda11 (changes requested):

mahesh-orch:
@blag 
* If file to delete doesn't exist then logging warning message at https://github.com/orchestral-st2/st2/blob/modify_action_delete_api/st2common/st2common/services/packs.py#L257 and L286
* In case of no permission to delete file logging full exception and returning proper error message via `abort()` at https://github.com/orchestral-st2/st2/blob/modify_action_delete_api/st2common/st2common/services/packs.py#L245 and L273 as `abort()` takes care of returning error message here.
* In case of any other exception logging full exception and returning generic message at https://github.com/orchestral-st2/st2/blob/modify_action_delete_api/st2common/st2common/services/packs.py#L255 and L284
* The logic in st2client/st2client/commands/action.py has cleaned up to reduce duplication as per your suggestion.
* Removed an unnecessary argument for `getattr()` in st2client/st2client/commands/action.py.
* Used `os.path.join(....)` to prepare path instead of concatenating strings in st2common/tests/unit/services/test_packs.py.
* Added assertions for the message of the mocked `PermissionError` in st2common/tests/unit/services/test_packs.py.
Please review the changes. Thanks.

mahesh-orch:
@cognifloyd The try/except blocks have been placed one after another instead of nesting as per your suggestion. I had made the changes according to @m4dcoder's suggestions. Docstrings are added in test classes to clear specificness of the unit tests in st2common/tests/unit/services/test_packs.py. Please review. Thanks.

st2api/st2api/controllers/v1/actions.py
cognifloyd: Thanks! that looks much better.

st2common/tests/unit/services/test_packs.py
cognifloyd: This looks resolved.

st2common/tests/unit/services/test_packs.py
cognifloyd: This one still needs to check either `PermissionError` or `ResourceDiskFilesRemovalError` or both. `Exception` is too broad.

st2common/tests/unit/services/test_packs.py
cognifloyd: This looks resolved.

st2common/tests/unit/services/test_packs.py
cognifloyd: This one still needs to check either `PermissionError` or `ResourceDiskFilesRemovalError` or both. `Exception` is too broad.

amanda11 (approved):
Thanks for the updates - my points have all been addressed. Much appreciated.

st2common/tests/unit/services/test_packs.py
m4dcoder: @mahesh-orch You missed this one I think

m4dcoder (approved):

st2common/tests/unit/services/test_packs.py
mahesh-orch: @cognifloyd I have added separate unit tests for `PermissionError`. As most probably this will be the exception will occur here and I'm raising same. But if any other exception occurs then I have used generalized `Exception` and raising `ResourceDiskFilesRemovalError`.

st2common/tests/unit/services/test_packs.py
mahesh-orch: @m4dcoder I checked in recent push. It has been updated.

st2common/tests/unit/services/test_packs.py
cognifloyd: Oh. I understand the test now. Cool.

cognifloyd (approved):

st2common/st2common/services/packs.py
```diff
@@ -215,3 +219,71 @@ def search_pack_index(
...
+                action_metadata_file_path
+            )
+            raise PermissionError(msg)
+        except Exception as e:
+            LOG.error(
+                'Could not delete "%s" file. Exception was "%s"',
+                action_metadata_file_path,
+                e,
+            )
+            if os.path.isfile(action_metadata_file_path):
```
blag, line 280: This check is probably redundant, since if `os.remove()` fails it will throw some sort of exception.

st2common/st2common/services/packs.py
```diff
@@ -215,3 +219,71 @@ def search_pack_index(
...
+                msg = 'The action file "{0}" could not be removed from disk, please check the logs or ask your StackStorm administrator to check and delete the actions files manually'.format(
```
blag, line 281: I'm kind of surprised our linting is not catching this long line.

st2common/st2common/services/packs.py
```diff
@@ -215,3 +219,71 @@ def search_pack_index(
...
+                action_entrypoint_file_path
+            )
+            raise PermissionError(msg)
+        except Exception as e:
+            LOG.error(
+                'Unable to delete "%s" file. Exception was "%s"',
+                action_entrypoint_file_path,
+                e,
+            )
+            msg = 'The action file "{0}" could not be removed from disk, please check the logs or ask your StackStorm administrator to check and delete the actions files manually'.format(
```
blag, line 252: I'm kind of surprised that our linting is not catching this long line.

blag (commented):
I would make some changes, but it's nothing serious.

mahesh-orch:
@blag I have done your suggested changes. Also, done similar changes in other files. Please review.

nzlosh:
This PR failed to provide sufficient options to remove a database entry without touching the files on disk

`--force` assumes yes to delete the file.  Yet, if you want to only remove the database entry and not touch the contents on disk, you must do it interactively.  The `force` option must take an argument of force yes/no so that non interactive shell commands to be run.

nzlosh:
Digging into this change a little more:

I triggered a corner case that isn't handled correctly. If there is a file access error, the database entry is deleted, leaving St2 in an inconsistent state.  This operation should be atomic, if the file was requested to be removed and it fails, the database entry should not be removed.
```
root@u1804:/opt/stackstorm/packs/mongodb# st2 action delete mongodb.user_actionx
The resource files on disk will be deleted. Do you want to continue? (y/n): y
MESSAGE: No permission to delete "/opt/stackstorm/packs/mongodb/actions/user_actionx.py" file from disk for url: http://127.0.0.1:9101/v1/actions/612cf0f24d30afa26ef56b88
```
Files `user_actionx.py` and `user_actionx.yaml` are still present in the pack.
```
root@u1804:/opt/stackstorm/packs/mongodb# ls actions/
collection_create.py    collection_delete.yaml  database_create.py    database_delete.yaml  lib              user_actionx.yaml  user_delete.py    user_update.yaml
collection_create.yaml  collection_list.py      database_create.yaml  database_list.py      __pycache__      user_create.py     user_delete.yaml  workflows
collection_delete.py    collection_list.yaml    database_delete.py    database_list.yaml    user_actionx.py  user_create.yaml   user_update.py
```
However, the action `user_actionx` is gone from the database.
```
root@u1804:/opt/stackstorm/packs/mongodb# st2 action list --pack=mongodb
+---------------------------+---------+--------------------------------------------------+
| ref                       | pack    | description                                      |
+---------------------------+---------+--------------------------------------------------+
| mongodb.collection_delete | mongodb | Delete a collection in a database.               |
| mongodb.collection_list   | mongodb | List the available collections for the database. |
| mongodb.database_create   | mongodb | Create a database on a mongo server              |
| mongodb.database_delete   | mongodb | List the available collections for the database. |
| mongodb.database_list     | mongodb | List the available databases in a mongo server   |
| mongodb.user_create       | mongodb | Create a mongodb user account                    |
| mongodb.user_delete       | mongodb | Delete an existing database user.                |
| mongodb.user_update       | mongodb | Update an existing database user account.        |
+---------------------------+---------+--------------------------------------------------+
```

cognifloyd:
The opposite applies for the default stackstorm-ha install where the packs filesystem is read-only, so deletes should only affect the db.
Maybe we need a system-wide setting in /etc/st2/st2.conf to enable/disable deleting files.

nzlosh:
I assume you mean for kubernetes.  I run StackStorm HA on VMs and the packs are not on read-only filesystems.  Regardless of which environment St2 is being run on, the code should be able to handle the situation in a generic and coherent way.

nzlosh:
@mahesh-orch @m4dcoder Please open a new PR to address the issues mentioned above.  As it stands, this change prevents removing just the database entry.

amanda11:
Could we get an issue raised and assign it to 3.6 milestone and 3.6 release project?
I think we can't release without a change as this will mean it's impossible to do a st2 delete without an error raised on the Kubernetes HA with it in the current state.

The idea of making it configurable in st2.conf could address most of the issues,and probably requires changes to both api and cli and doc.

PR #5306, opened 2021-07-16
Add st2client README checks to CI

### LINKED ISSUE #5191: Add a CI lint task to check st2client's README.md
We need to make sure that the st2client `README.rst` file is acceptable to PyPI, since any syntax errors in it will cause the `push_st2client` task of the `st2cd.st2_finalize_release` workflow to fail.

We can check the syntax using the same renderer that PyPI itself uses:

```bash
# Use the same README renderer that PyPI uses to catch syntax issues in the
# README.rst file  # st2client uses README.rst
# https://pypi.org/help/#description-content-type
# https://pypi.org/project/readme-renderer
# https://packaging.python.org/tutorials/packaging-projects/#description
echo "Checking README.rst syntax"
virtualenv venv-st2client-readme-checker
. venv-st2client-readme-checker/bin/activate
pip install --upgrade readme_renderer
python -m readme_renderer README.rst
deactivate
```

It would be nice if we could catch these errors before release, which means that we should create a step in our CI tooling to check it before any bad changes get merged.

amanda11:
Add lint checks to st2client README.rst so that catch before do pypi upload during release

Resolves #5191

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -482,7 +482,7 @@ flake8: requirements .flake8
 
 # Make task which verifies st2client README will parse pypi checks
 . PHONY: .st2client-pypi-check
-st2client-pypi-check:
+.st2client-pypi-check:
 	@echo
 	@echo "==================== st2client pypi check ===================="
 	@echo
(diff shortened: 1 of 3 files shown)


armab (approved):
Thanks for enhancement! :+1: 

fewer surprises during the release

PR #5307, opened 2021-07-20
Update sensor_wrapper.py

guzzijones:
syntax error in python3

### DIFF
--- a/st2reactor/st2reactor/container/sensor_wrapper.py
+++ b/st2reactor/st2reactor/container/sensor_wrapper.py
@@ -435,7 +435,6 @@ if __name__ == "__main__":
     if not isinstance(parent_args, list):
         raise TypeError(
             "Command line arguments passed to the parent process must be a list"
-            f" (was {type(parent_args)})."
         )
 
     obj = SensorWrapper(


st2reactor/st2reactor/container/sensor_wrapper.py
```diff
@@ -435,7 +435,6 @@ def _sanitize_trigger(self, trigger):
     if not isinstance(parent_args, list):
         raise TypeError(
             "Command line arguments passed to the parent process must be a list"
-            f" (was {type(parent_args)})."
```
cognifloyd, line 438: Why does this cause a syntax error?

guzzijones:
No idea but it def does. Try running that file you will see.

st2reactor/st2reactor/container/sensor_wrapper.py
amanda11: Also in the original PR change, it was requested that as part of the error it reports the type of the argument that isn't a list, so that it enables easier debugging. So even if it needs changing we still need that type in there.

It's odd - as I can run that block of code lines 435-438 perfectly fine.

Are we definitely sure its being run in a python 3.6 or python 3.8 environment?

guzzijones:
Ah I am betting the bin still has python in it. Instead of python 3

guzzijones:
Nope. I checked and the bin is correct. it has python3.  I have completely abandoned sensors anyhow due to them failing after 2 tries and not saving their state anywhere to be queried.  I am guessing it was something haywire on our installation.

PR #5309, opened 2021-07-22
bug fix for `output_value[output_key]` is not iterable

guzzijones:
if output_value[output_key] is not iterable and instead a boolean I receive the following error code from the st2api.
IE:
```
output_value: {'stdout': '', 'stderr': '', 'exit_code': 0, 'result': True}
spec: {'status': 'bool'}
key: output
output_key: result
```
ERROR:
```
MESSAGE: Internal Server Error for url: https://stackstorm-vf.ifp.lmco.com/api/executions/60f94708c3a6d1a338dc0ad7/children/?depth=-1
```

### DIFF
--- a/st2common/st2common/util/output_schema.py
+++ b/st2common/st2common/util/output_schema.py
@@ -69,8 +69,9 @@ def mask_secret_output(ac_ex, output_value):
         return output_value
 
     for key, spec in output_schema.items():
-        if key in output_value[output_key] and spec.get("secret", False):
-            output_value[output_key][key] = MASKED_ATTRIBUTE_VALUE
+        if not isinstance(output_value[output_key], bool):
+            if key in output_value[output_key] and spec.get("secret", False):
+                output_value[output_key][key] = MASKED_ATTRIBUTE_VALUE
 
     return output_value
 


st2common/st2common/util/output_schema.py
```diff
@@ -69,8 +69,9 @@ def mask_secret_output(ac_ex, output_value):
         return output_value
 
     for key, spec in output_schema.items():
-        if key in output_value[output_key] and spec.get("secret", False):
-            output_value[output_key][key] = MASKED_ATTRIBUTE_VALUE
+        if not isinstance(output_value[output_key], bool):
```
cognifloyd, line 72: What about this:

Add `from collections.abc import Iterable` and:
```suggestion
        if isinstance(output_value[output_key], Iterable):
```

cognifloyd:
This would probably also apply to int, float, and other types that are not `Collection`s.

guzzijones:
Works for me.

st2common/st2common/util/output_schema.py
```diff
@@ -69,8 +70,9 @@ def mask_secret_output(ac_ex, output_value):
         return output_value
 
     for key, spec in output_schema.items():
-        if key in output_value[output_key] and spec.get("secret", False):
-            output_value[output_key][key] = MASKED_ATTRIBUTE_VALUE
+        if isinstance(output_value[output_key], Collection):
```
cognifloyd, line 73: Hmm. On closer inspection, maybe this check should move up a few lines (lines 69-70) so that we don't have to check it for every entry in the output_schema.

```python
    if not output_value.get(output_key) or not isinstance(output_value[output_key], Collection):
        return output_value
```

And while you're in there, we could probably simplify the dict/list check on lines 60-61:
```python
    if not isinstance(output_value, Collection):
        return output_value
```

cognifloyd: Sorry for changing my suggestion. Thanks again!

guzzijones: I don't want to change anymore than I have.  It sounds like you have some idea of how to clean this up more so go ahead and take over if you would like.

m4dcoder (changes requested):
Can you add a unit test that reproduce this issue so we know this patch is going to fix?

cognifloyd:
Hmm. So there were two issues: 1) the bug you found. 2) your spec is not well formed.

Instead of this spec:
```
status: bool
```
You need this jsonschema spec:
```
status:
  type: bool
```

I'm looking at how to adjust the tests now.

cognifloyd:
K. I added #5319 based on this.
It turns out that output_schema_validation is actually quite limited.
Your output has to be an object and all properties in that object have to be objects as well. Then, you can use secret masking.
I left a FIXME comment about adjusting the schema to allow output_schema to describe more kinds of output (and then allow secret masking on those additional types as well).

guzzijones:
Where did we leave this?

cognifloyd:
I have too many irons in the fire. I can't work on this for now. I'll come back to it when I can.

guzzijones:
@guzzijones  add changelog entry.

cognifloyd (approved):
OK. I added a changelog entry that mentions that `output_schema` handling will change in a future release.
I also adjusted bailing out early if the output value is not an object or if a spec in output_schema is mal-formed.

I think this is ready to merge.

PR #5311, opened 2021-07-27
Update Makefile

blag:
This PR updates the `Makefile` to use Python 3's `venv` module, and it fixes a few oversights made when adding `black`.


armab (commented):
Thanks! The packaging is heavily relying on the `virtualenv` and so consistency is the first idea that came to my mind here.

But what is the motivation behind replacing `virtualenv` with `venv` at least for the Makefile?

blag:
I believe that the `virtualenv` executable is going to go away in future versions of official Python releases <sup>1</sup> and the `venv` module will be the primary way to create virtualenvs.

Now that we're only supporting ST2 on Python 3, we should always have the `venv` module, and we should migrate to that before it breaks our CI process and becomes an emergency.

<sup>1</sup> While I'm sure somebody in the Python community will create and maintain a `virtualenv` executable, I don't think this is something that we should rely on long term.

winem:
How do we want to proceed here? I see the point and am a big fan of tooling consistency but @blag has a very good point and I don't see any real risk in switching to venv. 

Is there anything else that may help to make a final decision?

armab (approved):
:+1: 

Forgot to follow up on this.

winem (approved):

cognifloyd (approved):

PR #5313, opened 2021-08-02
enhancement of error message for output schema validation failure

mahesh-orch:
Output schema fails in validation if metayaml file has a parameter of a specific type and python script doesn't have equivalent type of it for same parameter. For an example:
In metayaml file:
```
output_schema:
  osTestParam:
    description: test parameter
    type: string
    required: true
    secret: true
```
In python script:
```
class SampleAction(Action):
    def run(self):
        osTestParam = None
        return {"osParam": osTestParam}
```
If this action ran from st2 web UI then it gives `// Action produced no data` output in action output box and nothing else rather than this. It is needful to have an appropriate error message on web UI when such case occurs.


st2common/st2common/util/output_schema.py
```diff
@@ -77,6 +79,61 @@ def mask_secret_output(ac_ex, output_value):
...
+                msg = "Failed to validate action output. '{0}' is not of type '{1}' in entry point file.".format(
+                    param, action_schema[param]["type"]
+                )
+                LOG.exception("Failed to validate output. %s", msg)
+                result["exit_code"] = FAILURE_EXIT_CODE
+                result["result"] = "None"
+                result["stderr"] = msg
+                result["stdout"] = ""
+                return (result, status)
+
```
m4dcoder, line 136: This doesn't look right. How come the schema validation isn't able to pick up the type error and we have to manually check this with a bunch of if/elif statements? Can you research and provide a reasonable explanation here?

m4dcoder (changes requested):

st2common/st2common/util/output_schema.py
mahesh-orch: @m4dcoder As you said schema validation gives type error and gives message why output schema validation failed at CLI but when we run same action from web UI, it only gives `// Action produced no data` as output. As there is no error message, it isn't clear why action failed. After adding conditions for checking output schema property types in metadata and entry point files we can prepare appropriate result if any condition fails. It is giving proper error message at CLI and web UI as well.

mahesh-orch: I have observed that these changes are causing key error in st2 pack install workflow. Please give me inputs for better approach to resolve this issue.

m4dcoder: The if/else here is just wrong and heading in the wrong direction. The schema validation is doing the job of detecting the type error.  I would check the result with the error and message. Each runners have different format/structure for their result. I think the UI is expecting the error to be in a specific path in the result. Can you check this with the UI developer?  To fix the issue, you may have to make changes at https://github.com/StackStorm/st2/blob/master/st2common/st2common/util/output_schema.py#L95 to write the result correctly for the runner of the action.

m4dcoder (changes requested):

st2common/st2common/util/output_schema.py
m4dcoder: The if/else here is just wrong and heading in the wrong direction. The schema validation is doing the job of detecting the type error. I would check the result with the error and message. Each runners have different format/structure for their result. I think the UI is expecting the error to be in a specific path in the result. Can you check this with the UI developer? To fix the issue, you may have to make changes at https://github.com/StackStorm/st2/blob/master/st2common/st2common/util/output_schema.py#L95 to write the result correctly for the runner of the action.

m4dcoder (changes requested):

mahesh-orch:
@m4dcoder I have removed the several if/else statements. Have updated the result at https://github.com/StackStorm/st2/blob/master/st2common/st2common/util/output_schema.py#L95. It is giving proper error message in web UI as well. Also, added unit tests for changes. Had to make minor changes in existing unit tests in https://github.com/StackStorm/st2/blob/master/st2common/tests/unit/test_util_output_schema.py and https://github.com/StackStorm/st2/blob/master/contrib/runners/orquesta_runner/tests/unit/test_output_schema.py. Please review the changes. Thanks.

st2common/st2common/util/output_schema.py
```diff
@@ -93,8 +94,10 @@ def validate_output(runner_schema, action_schema, result, status, output_key):
         status = action_constants.LIVEACTION_STATUS_FAILED
         # include the error message and traceback to try and provide some hints.
         result = {
-            "error": str(ex),
-            "message": "Error validating output. See error output for more details.",
+            "exit_code": FAILURE_EXIT_CODE,
+            "result": "None",
+            "stderr": str(ex),
+            "stdout": "",
```
cognifloyd, line 100: validation is not a subprocess, so using `exit_code`, `stderr`, and `stdout` feels a little odd here.
Can we adjust something in st2web to pick up the error message instead?

cognifloyd (changes requested):
I really like adding tests for output_schema validation!

I'm not sure about forcing this into a exit_code/stderr though. That is semantically wrong, esp from an orquesta workflow. If we need to tweak the output here (maybe call it `validation_error`?), I think that's cool. But, I think st2web is going to need massaging to recognize the validation error messages.

contrib/runners/orquesta_runner/tests/unit/test_output_schema.py
```diff
@@ -197,9 +197,12 @@ def test_fail_incorrect_output_schema(self):
         self.assertEqual(ac_ex_db.status, ac_const.LIVEACTION_STATUS_FAILED)
 
         expected_result = {
-            "error": "Additional properties are not allowed",
-            "message": "Error validating output. See error output for more details.",
+            "stdout": "",
+            "stderr": "Additional properties are not allowed",
+            "exit_code": 1,
+            "result": "None",
```
m4dcoder, line 203: The structure of the expected result is different for each runner. What you have patched here is only valid for shell runner. I doubt this error will be caught for workflow execution. 1) Can you review each action runner (i.e. local shell, python, orquesta, etc.)? 2) Can you write a unit test for each action runner?

m4dcoder (changes requested):

mahesh-orch:
@cognifloyd , @m4dcoder Done research with help of UI developer and found that for showing action output in web UI the currently we are picking up `result.result`, `result.stdout`, `result.stderr` and `result.traceback` only. As you both suggested, we will do changes in UI to pick up `result.error`.

contrib/runners/orquesta_runner/tests/unit/test_output_schema.py
mahesh-orch: @m4dcoder As per recent research it is looking like this problem can be solved with minor changes in UI. Let me try with changes in UI if they don't work I will surely work on reviewing and unit tests on each runner.

PR #5319, opened 2021-08-04
Allow `output_schema` to validate non-object action output (int, bool, etc)

cognifloyd:
This PR turns `output_schema` into a full jsonschema which provides several benefits.
- validate action output even if it output is a boolean, integer, or other non-object type.
- mask entire output of an action (mark all of an action's output as secret)
- improve secret masking in action output for more object types

We ignore schemas that are not well-formed to avoid throwing errors. But, we now log a warning when a schema is invalid as an invalid schema prevents validation and secret masking from working.

If anyone is using `output_schema`, they will need to update metadata to keep using ouptut validation and secret masking.

For example, this schema:

```yaml
    output_schema:
      property1:
        type: bool
      property2:
        type: str
```
should be updated like this:
```yaml
    output_schema:
      type: object
      properties:
        property1:
          type: bool
        property2:
          type: str
      additionalProperties: false
```

<hr>

This builds on #5309 to refactor the output_schema validation and add missing tests.

Ultimately, this is a follow-up for #5250 which added secret-masking to the output_schema validation, but missed some of the possible data types for output_values.

Before this PR, `ouptut_schema` could only do output_validation and secret masking for properties of objects.

The runner schema required that output_schema be an object and `patternProperties` here allowed each property to be a schema:
https://github.com/StackStorm/st2/blob/b8402b4c730a4d20797d84e584a8ba20e6e2684a/st2common/st2common/models/api/action.py#L119-L125
And the same applies to action schema:
https://github.com/StackStorm/st2/blob/b8402b4c730a4d20797d84e584a8ba20e6e2684a/st2common/st2common/models/api/action.py#L231-L237
This is where the schema for each output_schema property schema is loaded:
https://github.com/StackStorm/st2/blob/b8402b4c730a4d20797d84e584a8ba20e6e2684a/st2common/st2common/util/schema/__init__.py#L89-L95
https://github.com/StackStorm/st2/blob/b8402b4c730a4d20797d84e584a8ba20e6e2684a/st2common/st2common/util/schema/__init__.py#L57-L59
And `action_output_schema.json` is the full json schema spec with our extensions (secret masking, etc)
https://github.com/StackStorm/st2/blob/b8402b4c730a4d20797d84e584a8ba20e6e2684a/st2common/st2common/util/schema/action_output_schema.json#L2-L4
https://github.com/StackStorm/st2/blob/b8402b4c730a4d20797d84e584a8ba20e6e2684a/st2common/st2common/util/schema/action_output_schema.json#L92-L99

Since `output_schema` used `patternProperties`, it could only validate objects - the only type that has properties. If your action or runner needed to output any other types (list, int, bool, ...), then output_validation and secret masking was unavailable. And, if your action provided an object whose keys needed to be masked (vs the property values), then that was also not possible.


st2common/st2common/util/output_schema.py
```diff
@@ -53,22 +54,39 @@ def _validate_action(action_schema, result, output_key):
...
 
 def mask_secret_output(ac_ex, output_value):
-    if not output_value:
-        return output_value
-
-    if not isinstance(output_value, dict) and not isinstance(output_value, list):
+    # We only support output_schema validation when the output_value is a JSON object.
+    # Invididual keys of that object can be marked secret, but the entire output
+    # object cannot be marked as secret.
+    # FIXME: Should we support non-objects under output_key?
```
m4dcoder, line 60: We should. Is this something we can include in this PR to fix?

st2common/st2common/util/output_schema.py
```diff
@@ -53,22 +54,39 @@ def _validate_action(action_schema, result, output_key):
...
+    #        Changing that will require changes in one or more of:
+    #          st2common/util/schema/action_output_schema.json
+    #          st2common/util/schema/__init__.py
+    #          st2common/models/api/action.py
+
+    if not output_value or not isinstance(output_value, Mapping):
```
m4dcoder, line 66: Are we sure list is not supported?  I thought json schema can be applied to a list. Is the output schema for the output value defined correctly?

m4dcoder (changes requested):
I got some questions.

st2common/st2common/util/output_schema.py
cognifloyd: I suppose, but that will increase the scope of this PR. I'll have to play with the schema to see how to allow more than just object.

st2common/st2common/util/output_schema.py
cognifloyd: Yes jsonschema can validate all JSON types, but `action_output_schema.json` currently only allows for object.

cognifloyd: Oh. I see where I misread the schema. Of course the schema itself is an object. And it accepts any of the types:
https://github.com/StackStorm/st2/blob/b8402b4c730a4d20797d84e584a8ba20e6e2684a/st2common/st2common/util/schema/action_output_schema.json#L136-L140

So, yes, output_schema already supports all simple types, so we need to support all of those types as well.

st2common/st2common/util/output_schema.py
cognifloyd: As in my other comment, yes, this should be in-scope for this PR. I misread the schema. I will push an update.

st2common/st2common/util/output_schema.py
cognifloyd: `output_schema` != `action_output_schema.json`
https://github.com/StackStorm/st2/blob/b8402b4c730a4d20797d84e584a8ba20e6e2684a/st2common/st2common/models/api/action.py#L231-L237

So, the output_schema begins with the premise of describing a list of properties. Thus the top-level `output_value[output_key]` object has to be an object in order to be validated.

So, `st2common/models/api/action.py` is the place that needs to change if we want the output_schema to describe any other types.

st2common/st2common/util/output_schema.py
cognifloyd: I added support for recursively masking output whatever form the spec is in.
But we still only support top-level output_value being objects, so I left my `isinstance(..., Mapping)` check in place.

What would I do if output_value is an array? Apply the output_schema to each of the (assumed) objects in that array?
That's not how _validate_runner() and _validate_action() use the output_schema. So, other places would need to learn to support non-object output.

cognifloyd:
@m4dcoder I added some comments where I think we'll need to change things to support validating non-object action output.
Could you take a look again.
Do we need to fix this in a backwards compatible way so that output_schemas that are defined for objects can be supported alongside full schema definitions?

guzzijones:
Here is the offending yaml for reference.  And yes it was malformed:

guzzijones:
```
---
name: "array_to_file_newline"
runner_type: "python-script"
description: "convert json to string and dump to file"
enabled: true
entry_point: "array_to_file_newline.py"
parameters:
  json: 
    type: array 
    description: "listjson to convert to newline string"
    required: true
  output_file:
    type: string
    description: "file name to output"
    required: true
output_schema:
  output:
    status: bool 

```

Here is the offending yaml file.  Yes it was malformed.

m4dcoder (dismissed):

guzzijones:
you can remove me from the reviewers.  I won't be able to add any value as this has gone well beyond my scope of understanding.

st2common/st2common/models/api/action.py
```diff
@@ -230,9 +233,19 @@ class ActionAPI(BaseAPI, APIUIDMixin):
...
+                        "type": "object",
+                        # using patternProperties like this implies that output_schema
+                        # defines the "properties" schema of an object where each key
+                        # is a property name.
+                        "patternProperties": {
+                            r"^\w+$": util_schema.get_action_output_schema()
+                        },
+                        "additionalProperties": False,
+                    },
+                ],
```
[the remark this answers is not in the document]
cognifloyd: OK.
1. I've changed both runner and action schemas
2. Both schemas now use use `"output_schema": util_schema.get_action_output_schema()`
3. It is simpler by sacrificing support for the legacy `output_schema` that only supported validation/masking of object properties.

To make these changes, I had to update everything that defined `output_schema` (including runner metadata and a variety of tests).

rush-skills (approved):

CHANGELOG.rst
```diff
@@ -25,6 +25,41 @@ Added
...
+
+  .. code-block:: yaml
+
+    output_schema:
+      property1:
+        type: bool
+      property2:
+        type: str
+
+  should be updated like this:
```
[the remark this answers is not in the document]
cognifloyd: I believe we can upgrade packs ahead of time and the malformed schema will just be ignored.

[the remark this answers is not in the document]
amanda11: I think to be honest they could upgrade the pack during the ST2 downtime using the st2-pack-install script - as that doesn't rely on ST2 being up. But we should test the recommend procedure.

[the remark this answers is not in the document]
amanda11: @cognifloyd I'll approve the PR, but if you can just update the changelog to make it clearer that only the new format is supported that would be great.
And then the upgrade procedure would be a st2docs PR anyway....

amanda11 (approved):

CHANGELOG.rst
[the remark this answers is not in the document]
cognifloyd: I added a `BREAKING CHANGE` bit to the changelog.

cognifloyd:
Documentation added in https://github.com/StackStorm/st2docs/pull/1137

PR #5321, opened 2021-08-06
Pack config: Process defaults and datastore values in patternProperties and additionalItems

cognifloyd:
Building on #5225 allowed pack config to get defaults populated and get datastore values injected when needed under `additionalProperties`

This PR does the same for `patternProperties` and `additionalItems`.

We're were not processing defaults or datastore values in pack config under:
- `patternProperties` on objects (discovered in https://github.com/StackStorm-Exchange/exchange-incubator/pull/166#discussion_r675093473)
- `additionalItems` on arrays (noticed while adding patternProperties support)

This makes sure we are applying defaults on nested items in arrays as well.

TODO:
- [x] tests


CHANGELOG.rst
```diff
@@ -185,6 +185,12 @@ Added
 
   Contributed by @nzlosh
 
+* Fix a bug in the pack config loader so that objects covered by an ``patternProperties`` schema
```
amanda11, line 188: Unfortunately, this needs moving into the 3.8.0 section.

st2common/st2common/util/config_loader.py
```diff
@@ -112,9 +114,56 @@ def _get_object_property_schema(object_schema, additional_properties_keys=None):
             # ensure that these keys are present in the object
             for key in additional_properties_keys:
                 property_schema[key] = additional_properties
-        property_schema.update(object_schema.get("properties", {}))
+
+        properties_schema = object_schema.get("properties", {})
+        property_schema.update(properties_schema)
+
+        potential_patterned_keys = set(additional_properties_keys) - set(
```
amanda11, line 121: Maybe a comment, it took me a while to double-check this. But I think your logic is ok.

additionalProperties only applies to keys that are not matched by properties or patternProperties. So patternProperties takes precedence. Therefore any keys that have their own schema AND match the patternProperties are going to have the pattern rule applied rather than the additionalProperty defined.

I can see this confusing some, so I wonder as well if we should have a debug log statement for any keys that do get their property updated.

st2common/st2common/util/config_loader.py
```diff
@@ -112,9 +114,56 @@ def _get_object_property_schema(object_schema, additional_properties_keys=None):
...
+            # update any matching key
+            for raw_pattern, pattern_schema in pattern_properties.items():
+                if not potential_patterned_keys:
+                    # nothing to check. Don't compile any more patterns
+                    break
+                pattern = re.compile(raw_pattern)
+                for key in list(potential_patterned_keys):
+                    if pattern.search(key):
+                        property_schema[key] = pattern_schema
+                        potential_patterned_keys.remove(key)
```
amanda11, line 137: Can we add into the tests a case where we have additionalProperties as a list and a patternProperties. Show that the additionalProperty that doesn't match the pattern has its schema applied, but where the additionalProperty matches the patternProperty that the patternProperty one is applied.
Is there a need to remove the key from the potential_patterned_keys set whilst we are traversing it? It's just a temporary set.

amanda11 (commented):
This is great. I've not had chance to check the arrayItems changes, but went through patternProperties. Just raised a couple of minor comments on the patternProperties.

CHANGELOG.rst
cognifloyd: Oops. Thanks. Fixed.

st2common/st2common/util/config_loader.py
cognifloyd: No, this is calculating which keys are not explicitly in `properties`. Then, patternProperties is only compared against this subset of keys.

I can see hope this code is confusing. I just added some comments and renamed a few variables to hopefully make this more grokkable. Could you take another look?

Do my changes clarify your question, or did I misunderstand?

st2common/st2common/util/config_loader.py
cognifloyd: Why would additionalProperties be a list? Is this yet another schema feature I didn't account for?

cognifloyd: We remove the key from the set so that we don't match a key more than once. The first found pattern wins.

cognifloyd:
I just refactored var naming a bit, so hopefully the arrayItems method will be easier to grok.

st2common/st2common/util/config_loader.py
amanda11: I was going by this: http://json-schema.org/understanding-json-schema/reference/object.html#additional-properties - where it has an example of additionalproperties that is a list.

st2common/st2common/util/config_loader.py
```diff
@@ -99,21 +100,105 @@ def _get_values_for_config(self, config_schema_db, config_db):
         return config
 
     @staticmethod
-    def _get_object_property_schema(object_schema, additional_properties_keys=None):
+    def _get_object_properties_schema(object_schema, objecy_keys=None):
```
amanda11, line 103: Is this a typo, should that be object_keys?

st2common/st2common/util/config_loader.py
amanda11: As discussed on Slack, correction from me - additioanProperties is object or boolean.

st2common/st2common/util/config_loader.py
cognifloyd: Ah yes. That's what I get for typing in my phone. I'll fix that.

st2common/st2common/util/config_loader.py
cognifloyd: Ok. I've pushed my refactor - the properties logic feels a lot cleaner now. Next chance I get I'll add tests that show the order of precedence of properties, patternProperties, and additionalProperties.

amanda11 (commented):
Thanks for update, the additionalProperties handling is really clear now, and looks good.
I haven't looked at additionalItems, but will try and do that when you've added the extra tests.

Thanks for the refactor.

st2tests/st2tests/fixtures/packs/dummy_pack_schema_with_additional_items_1/config.schema.yaml
```diff
@@ -0,0 +1,24 @@
+---
+  region:
+    type: "string"
+    required: false
+    default: "us-east-1"
+  profiles:
+    type: "array"
+    required: false
+    additionalItems:
```
amanda11, line 9: As additionaItems can be a dict or boolean, is it possible to add a test with additionalItems with the true/false behaviour test.

Also as an aside do we state which JSON Schema we use? As I think its renamed from additionalItems to items in more recent schema: https://json-schema.org/understanding-json-schema/reference/array.html#additional-items

cognifloyd:
K. I added a test showing the order of precedence in of properties, patternProperties, and additionalProperties in pack config.

cognifloyd:
I'm reviewing the schema to make sure I implemented patternProperties correctly.

We use draft 4 (and draft 3's `required`):
https://github.com/StackStorm/st2/blob/bb04575f8e56a4046546fed05ae651c041beeba6/st2common/st2common/util/schema/draft4.json#L2
https://github.com/StackStorm/st2/blob/9e336d817405cb5b4ab117b47bac716b1e1fa5f6/st2common/st2common/util/schema/__init__.py#L46-L50

In draft 4, patternProperties creates an all-of set of schemas for each property. Each property gets compared against all patterns. And it must match all of the schemas for the patterns it matches. So, we can't avoid compiling some of the patterns like I thought. 😥
https://datatracker.ietf.org/doc/html/draft-fge-json-schema-validation-00#section-8.3

So, I need to refactor a bit more to check all patterns against all keys not included in `properties`.

cognifloyd:
K. Refactored and ready for more reviews

amanda11 (approved):
this looks good to me.
The only thing I thought of was whether it was worth having additional additionalItems tests, e.g. are we covering the case where additionalItems is "true" and additionalIItems is "false".

I've approved though, if you don't think those are needed. As the logic looks good, and other than that tests look great.

Thanks for all the work.

cognifloyd:
With `additionalItems: true`, the schema for each of those additional items is effectively `{}`. So, there are:
- no defaults to interpolate
- no automatically decrypted secrets (because nothing has `secret: true` unda a `{}` schema.

But, I suppose there can still be jinja templates without auto-decryption. So, I just added a test for a template that uses `|decrypt_kv`

st2tests/st2tests/fixtures/packs/dummy_pack_schema_with_additional_items_1/config.schema.yaml
cognifloyd: done

rush-skills (approved):
LGTM Code-wise, didn't get a chance to set up and test.

PR #5322, opened 2021-08-07
Ignore more vim temp files

cognifloyd:
I use vim quite a bit, so I end up with a variety of vim temporary files.
I could use my user gitignore, but this could be helpful for other vim users as well, so I'm submitting a quick PR.


winem (approved):
Nice, very helpful.

PR #5328, opened 2021-08-13
Ensure correct error message when have error decrypting parameter and also have keys that convert

### LINKED ISSUE #5312: Decrypting K/V values in action parameters causing misleading errors if the keys don't exist and another K/V parameter needs to be converted to bool, array, or object
## SUMMARY

In an action's parameters, decrypting an encrypted value in the K/V store results in a misleading error message if that key either doesn't exist or is not encrypted

### STACKSTORM VERSION

3.5.0

RHEL 8, Python 3.6.8

st2 key set test_bool "true"

test_action.yaml
```
---
name: 'test_action'
runner_type: 'remote-shell-cmd'
description: 'Test action inputs from the k/v store'
enabled: true
entry_point: ''
pack: 'test_pack'
parameters:
    username:
        type: 'string'
        description: 'Username to login to the remote system'
        required: true
    password:
        type: 'string'
        description: 'Password to login to the remote system'
        required: true
    hosts:
        type: 'string'
        description: 'Comma separated list of hosts to run the command on'
        required: true
    cmd:
        immutable: true
        default: 'echo "TEST: {{ test_bool }}"'
    connect_timeout:
        type: 'integer'
        description: 'SSH connect timeout in seconds'
        default: 30
    test_bool:
        type: boolean
        description: "Test boolean variable"
        default: "{{ st2kv.system.test_bool }}"
    test_encrypt:
      	type: string
        description: "Test encrypted variable"
        required: false
        secret: true
        default: "{{ st2kv.system.test_encrypt | decrypt_kv }}"
```

st2 run test_pack.test_action username="root" password="pass" hosts="localhost"

## Expected Results

Prior to 3.5, if test_encrypt does not exist in the K/V store then a None value would be used for that parameter. If that can't be done then an error message about the test_encrypt parameter would be expected

## Actual Results

Running the action above will result in the following misleading error if test_encrypt either is not in the k/v store or if it exists but isn't encrypted:
```
MESSAGE: '{{ st2kv.system.test_bool }}' is not of type 'boolean' for url: http://127.0.0.1:9101/executions
```

amanda11:
When you have an action that takes two key/value pairs whereby:
1) One parameter is encrypted and the key/value isnt' set
2) Another parameter is of a type other than string, and key/value pair is set

When the error is reported for the parameter that can't be decrypted, on handling that error another exception is raised when it attempts to validate the parameter json in regard to the parameter in key/value store that needs to be converted. This means that the incorrect error message is reported.
See #5312 for an example case of when this occurs.

Closes #5312


st2common/st2common/services/action.py
```diff
@@ -55,7 +55,9 @@ def _get_immutable_params(parameters):
     return [k for k, v in six.iteritems(parameters) if v.get("immutable", False)]
 
 
-def create_request(liveaction, action_db=None, runnertype_db=None):
+def create_request(
+    liveaction, action_db=None, runnertype_db=None, validate_params=True
```
Kami, line 59: I guess docstring which explains what ``validate_params`` does would help.

st2common/tests/unit/services/test_action.py
```diff
@@ -624,3 +624,24 @@ def test_root_liveaction(self):
             child, expected_root = self._create_nested_executions(depth=i)
             actual_root = action_service.get_root_liveaction(child)
             self.assertEqual(expected_root["id"], actual_root["id"])
+
+    def test_invalid_json_request_validate(self):
+        parameters = {"hosts": "127.0.0.1", "cmd": "uname -a", "arg_default_value": 123}
+        liveaction = LiveActionDB(action=ACTION_REF, parameters=parameters)
+
+        self.assertRaises(
```
Kami, line 632: Probably a good idea to also assert on what exception is being thrown?

st2common/tests/unit/services/test_action.py
```diff
@@ -624,3 +624,24 @@ def test_root_liveaction(self):
...
+            action_service.create_request,
+            liveaction,
+        )
+
+    def test_invalid_json_request_skipvalidate(self):
+        parameters = {"hosts": "127.0.0.1", "cmd": "uname -a", "arg_default_value": 123}
+        liveaction = LiveActionDB(action=ACTION_REF, parameters=parameters)
+
+        self.assertRaises(
+            jsonschema.ValidationError,
```
Kami, line 643: Same here.

amanda11:
@Kami Thanks for review comments, I've addressed them now - and in process, realised I had one of the UT wrong! Ready for re-review when you get a chance.

winem (approved):
LGTM!

PR #5331, opened 2021-08-17
Fix running scripts/write-headers.sh on Darwin

cognifloyd:
I tried running scripts/write-headers.sh on Mac OSX but the sed behavior, or perhaps the escaping behavior, was slightly different, so the headers were malformed. In particular `\n` did not add a newline as expected. So, I massaged the HEADER message such that injecting that newline was no longer necessary.


amanda11 (approved):
LGTM

PR #5332, opened 2021-08-17
Make our custom pylint plugin use the AST

cognifloyd:
pylint plugins are supposed to inspect code without importing it to avoid side effects like importing eventlet.
In some of my manual pylint runs, I ran into a variety of really odd errors and hangs that were ultimately caused by importing the st2 modules from within the pylint plugin.

So, this PR modifies the plugin to use the astroid AST instead of using `__import__`.

~This is a draft while I compare run times with master branch. I'm curious if this will speed things up at all.~

- Fix pylint plugin to rely on ast instead of `__import__()`
- finish pylint api_models plugin so it uses AST only
- drop unused import

### DIFF
--- a/pylint_plugins/api_models.py
+++ b/pylint_plugins/api_models.py
@@ -36,6 +36,15 @@ def register(linter):
     pass
 
 
+def infer_copy_deepcopy(node):
+    if not (
+        isinstance(node.value, nodes.Call)
+        and node.value.func.as_string() == "copy.deepcopy"
+    ):
+        return
+    return node.value.args[0].infer()
+
+
 def transform(cls):
     if cls.name in CLASS_NAME_BLACKLIST:
         return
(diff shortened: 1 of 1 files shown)


cognifloyd:
If this change offers any speedups, the speedups are negligible.

CI master - https://github.com/StackStorm/st2/actions/runs/1140506567
  py3.6 - make time 3m 26s
  py3.8 - make time 2m 55s

CI unrelated PR - https://github.com/StackStorm/st2/actions/runs/1140762169
  py3.6 - make time 3m 14s
  py3.8 - make time 3m 35s

CI with pylint AST changes (this PR) - https://github.com/StackStorm/st2/actions/runs/1140785614
  py3.6 - make time 3m 10s
  py3.8 - make time 3m 0s

cognifloyd:
Rebased on master. @Kami Is anything left before we can merge this? Could you review (and hopefully approve)?

m4dcoder:
@cognifloyd How do you run pylint manually? I have not run into issue or have seen the errors you posted above. The changes here looks more complex than before.

cognifloyd:
Importing code that we are linting is not nice. It breaks the promise of pylint because it is no longer static code analysis, it is reflection.

I have been experimenting running pylint via another program called pants, because it promises to have aggressive caching of results to speed things up. It understands python imports, and so it can track and see that if you change a file b and file a imports file b, then it needs to rerun lint and test steps on both file a and file b, not just the one file you changed. That could really speed up PR CI. Plus, I had one company suggest they would give StackStorm an open source plan to use their remote execution services which would allow us to share cached results between CI and local, so that the tests are that much faster both in CI and locally.

However, because our plugin is using `__import__` there is no way to know statically that linting one file will require so many other transient dependencies, not because it uses them, but because of all of the import side effects.

cognifloyd:
I guess I failed in my goal of justifying this without referencing `pants`. I guess I'll close this until I'm ready to demonstrate `pants`.

cognifloyd:
OK. I'm adding a bunch of comments to this code to explain how it works and what it is doing.

winem (approved):
This looks good to me and I think it's a great improvement. But I'd love to see more reviews by other maintainers.

Issue #4641, opened 2019-04-19
bug: core pack: `core.inject_trigger`: `trigger` parameter masks `trigger` payload in rule setting


djhaskin987:
##### SUMMARY

*The action parameter for the `core.inject_trigger` action is masking the `trigger` variable available from within the jinja in a rule.*

I am trying to fire off a trigger from a rule using `core.inject_trigger`, and I want to pass along some of the information from the webhook trigger payload (which matches the rule) into the new trigger which I am creating. This doesn't work, because when I use, e.g. `{{ trigger.<thing> }}` within a rule, the `trigger` jinja variable seems to be masked by the `trigger` parameter in the `core.inject_trigger` rule. 

##### ISSUE TYPE

Bug Report

##### STACKSTORM VERSION
Paste the output of `st2 --version`:

st2 2.10.4, on Python 2.7.5

##### OS / ENVIRONMENT / INSTALL METHOD

Reproduced on basic one-server install

##### STEPS TO REPRODUCE

1. Create a rules file named whatever you want, and have it look like this:
```
---
name: "pass-along-trigger"
pack: "mypack"
description: "Process webhook POST that comes in."
enabled: true

trigger:
  type: "core.st2.webhook"
  parameters:
    url: "process-webhook"

criteria: {}

action:
  ref: "core.inject_trigger"
  parameters:
    trigger: "mypack.clean_trigger"
    payload:
      thing1: {{ trigger.thing1 | regex_replace("cleanthings", "up") }}
      thing2: {{ trigger.thing2 | regex_replace("cleanthings", "more") }}
  trace_tag: "cool-i-can-set-a-trace-tag-based-on-{{ trigger.webhook-content-stuff }}"
```
Now just register the rule and run it with `st2-rule-tester`.

##### EXPECTED RESULTS

I expect the stuff from the payload of the webhook trigger instance for `process-webhook` webhook to be available in the jinja of the rule.

##### ACTUAL RESULTS

Instead, when I use `trigger` in the jinja above stackstorm insists that the value of the jinja `trigger` variable is `"mypack.clean_trigger"`, which is the value of the `trigger` action parameter.

*The action parameter for the `core.inject_trigger` action is masking the `trigger` variable available from within the jinja in a rule.*

###### Suggested Fix

[...]

PR #5335, opened 2021-08-19
Trigger name collision workaround

mjtice:
This addresses the jinja trigger name collision noted in issue #4641


cognifloyd:
This looks like a good clean change.

@mjtice would you please sign the CLA https://github.com/StackStorm/st2/pull/5335#issuecomment-901859873 ? It would be great to merge this sooner rather than later.

cognifloyd:
It looks like maintainer edits aren't allowed - and I don't think github allows that for forks in organizations.

So, let's see if we can get this ready to merge together. Please:

1. add a changelog entry
2. merge in master

I'm not sure why GHA tests haven't run for the last commit (only circleci ran? weird). I hope tests will run once you do that, if not I might need to pull in your commits another way.

cognifloyd:
I just opened #5383 to include your changes. As soon as someone else approves it, we can get it merged.

PR #5338, opened 2021-08-23
Shutdown kombu consumer on worker shutdown 

khushboobhatia01:
As of now, only the disptacher is shutdown in the shutdown method. This change takes care of shutting down Kombu consumer thread. 

In HA environment this will ensure other healthy action runner pods to pick up requested actions instead of an unhealthy pod picking the action from queue and abandoning it.

In a non-HA environment, actions requested after worker has stopped will be in scheduled state and will be picked after new action runner comes up.


cognifloyd (approved):
Nice.
For other reviewers. Here are the docs for `should_stop`:
https://docs.celeryproject.org/projects/kombu/en/stable/reference/kombu.mixins.html#kombu.mixins.ConsumerMixin.should_stop

armab (changes requested):
Thanks @khushboobhatia01, appreciate the contribution based on running st2 in HA prod!

Could you please also add a `CHANGELOG.rst` record for this fix?

khushboobhatia01:
> Thanks @khushboobhatia01, appreciate the contribution based on running st2 in HA prod!
> 
> Could you please also add a `CHANGELOG.rst` record for this fix?

Yes, updated.

armab (approved):
Thanks for the Changelog! :+1:

khushboobhatia01:
> LGTM, but can you please also add a test case for it?

@Kami Added test case. Can you please merge? Thank you! :)

cognifloyd (commented):
An excellent test. It feels more like an integration test though. Could you move the new test to a new file under `st2actions/tests/integration`?

Thanks!

khushboobhatia01:
> An excellent test. It feels more like an integration test though. Could you move the new test to a new file under `st2actions/tests/integration`?
> 
> Thanks!

Done!

PR #5347, opened 2021-08-29
make http/https_proxy env var checking case insensitive (address issue #5137)

### LINKED ISSUE #5137: should check both 'http_proxy' and 'HTTP_PROXY' when get proxy info from environment variables
## SUMMARY

I just faced an embarrassed problem that command `st2 action list` keeps return 503. 

I tried `st2 --debug action list` and the result in line  `HTTP_PROXY`(**uppercase**) was empty. This mislead me that the proxy info was set(which is not).

I finally found that the root cause is a **lowercase** `http_proxy` env variable was set in the system.

So I think it maybe more friendly to check both 'http_proxy' and 'HTTP_PROXY' to get proxy info when using st2 command with `--debug` option.

related code:
https://github.com/StackStorm/st2/blob/5c4e5f8e93c7ed83c4aa4d196085c2912ae7b285/st2client/st2client/base.py#L404

### STACKSTORM VERSION

3.3

centos 7.6, custom install

1) set **lowercase** proxy environment variable, eg. `export http_proxy='127.0.0.1:8888'`
2) run `st2 --debug action list`

## Expected Results

proxy info should be printed

## Actual Results

proxy info did not show, value of HTTP_PROXY is empty.

Thanks!

S-T-A-R-L-O-R-D:
checking both 'http_proxy' and 'HTTP_PROXY' to get proxy info when using st2 command with --debug option.

Fixes #5137

### DIFF
--- a/st2client/st2client/base.py
+++ b/st2client/st2client/base.py
@@ -447,6 +447,12 @@ class BaseCLIApp(object):
         print("")
         print("Proxy settings:")
         print("---------------")
-        print("HTTP_PROXY: %s" % (os.environ.get("HTTP_PROXY", os.environ.get("http_proxy"))))
-        print("HTTPS_PROXY: %s" % (os.environ.get("HTTPS_PROXY", os.environ.get("https_proxy"))))
+        print(
+            "HTTP_PROXY: %s"
+            % (os.environ.get("http_proxy", os.environ.get("HTTP_PROXY", "")))
+        )
+        print(
+            "HTTPS_PROXY: %s"
+            % (os.environ.get("http_proxy", os.environ.get("HTTPS_PROXY", "")))
+        )
         print("")
--- a/st2client/st2client/base.py
+++ b/st2client/st2client/base.py
@@ -447,6 +447,6 @@ class BaseCLIApp(object):
         print("")
         print("Proxy settings:")
         print("---------------")
-        print("HTTP_PROXY: %s" % (os.environ.get("HTTP_PROXY", "")))
-        print("HTTPS_PROXY: %s" % (os.environ.get("HTTPS_PROXY", "")))
+        print("HTTP_PROXY: %s" % (os.environ.get("HTTP_PROXY", os.environ.get("http_proxy"))))
+        print("HTTPS_PROXY: %s" % (os.environ.get("HTTPS_PROXY", os.environ.get("https_proxy"))))
         print("")
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -256,6 +256,9 @@ Changed
 * Fixed ``__init__.py`` files to use double quotes to better align with black linting #5299
 
   Contributed by @blag.
+* Fixed ``st2client/st2client/base.py`` file to check for http_proxy and https_proxy environment variables for both lower and upper cases.
+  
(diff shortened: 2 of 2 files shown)


S-T-A-R-L-O-R-D:
I have agreed to the CLA license about 5 hours ago, but it still asking me to sign

st2client/st2client/base.py
```diff
@@ -445,6 +445,6 @@ def _print_client_settings(self, args):
         print("")
         print("Proxy settings:")
         print("---------------")
-        print("HTTP_PROXY: %s" % (os.environ.get("HTTP_PROXY", "")))
-        print("HTTPS_PROXY: %s" % (os.environ.get("HTTPS_PROXY", "")))
+        print("HTTP_PROXY: %s" % (os.environ.get("HTTP_PROXY", os.environ.get("http_proxy"))))
```
amanda11, line 448: This means that if neither are specified that None is outputted instead of "", could you use , "" on the inner os.environ.get.

Also, what happens if both are set. I think http_proxy in lowercase generally takes prioiryt, so we should probably change the order so that we check for the lowercase value, and only look for uppercase if its not set.

amanda11 (changes requested):
Thanks for the contribution. I've put a suggestion in for a change, and could you also:

- sign the CLA with the email address you are using for github (we did change the CLA so might need to re-sign)
- add an update to the changelog file

cognifloyd:
@S-T-A-R-L-O-R-D:
You need to sign the CLA with your zscaler address (the email used on the first commit).
Or amend the commit to use your gmail address (the email you've signed the CLA with).

S-T-A-R-L-O-R-D:
My zscaler Id was my corporate Id and I no longer have to acces to that Id.
If possible I would like to rewrite my history.

Thanks and regards,
Devesh Solanki

cognifloyd:
Rebased on master and corrected the author email.

cognifloyd (approved):

cognifloyd:
Rebased and reformatted with black.

rush-skills (approved):

st2client/st2client/base.py
```diff
@@ -447,6 +447,12 @@ def _print_client_settings(self, args):
...
         print("---------------")
-        print("HTTP_PROXY: %s" % (os.environ.get("HTTP_PROXY", "")))
-        print("HTTPS_PROXY: %s" % (os.environ.get("HTTPS_PROXY", "")))
+        print(
+            "HTTP_PROXY: %s"
+            % (os.environ.get("http_proxy", os.environ.get("HTTP_PROXY", "")))
+        )
+        print(
+            "HTTPS_PROXY: %s"
+            % (os.environ.get("http_proxy", os.environ.get("HTTPS_PROXY", "")))
```
wfgydbu, line 456: is this http_proxy or http**s**_proxy for HTTPS_PROXY ENV?

cognifloyd: Good catch. If you submit a PR to add the missing `s`, then I can merge it fairly quickly. Thanks!

wfgydbu: @cognifloyd  Done, please check: https://github.com/StackStorm/st2/pull/5678.

feel free to let me know if any parts need to be modified.

PR #5351, opened 2021-09-02
Bug fix in modified action delete API

mahesh-orch:
Currently if PermissionError (or any other exception) occurs during removing files from disk, action delete operation deregisters the action from database but action files remains in filesystem i.e. this operation leaving st2 in an inconsistent state. So, it is required to rollback the operation and register action to database again if such case occurs.

### DIFF
--- a/st2api/st2api/controllers/v1/actions.py
+++ b/st2api/st2api/controllers/v1/actions.py
@@ -258,9 +258,17 @@ class ActionsController(resource.ContentPackResourceController):
                 entry_point=entry_point,
                 metadata_file=metadata_file,
             )
+
+        except PermissionError as e:
+            LOG.error("No permission to delete resource files from disk.")
+            action_db.id = None
+            Action.add_or_update(action_db)
+            abort(http_client.FORBIDDEN, six.text_type(e))
+            return
+
         except Exception as e:
             LOG.error(
-                "Exception encountered during deleting resource files from disk."
+                "Exception encountered during deleting resource files from disk. "
                 "Exception was %s",
                 e,
             )
--- a/st2api/st2api/controllers/v1/actions.py
+++ b/st2api/st2api/controllers/v1/actions.py
@@ -264,6 +264,9 @@ class ActionsController(resource.ContentPackResourceController):
                 "Exception was %s",
                 e,
             )
+            action = ActionAPI.from_model(action_db)
+            action_db = ActionAPI.to_model(action)
+            Action.add_or_update(action_db)
             abort(http_client.INTERNAL_SERVER_ERROR, six.text_type(e))
             return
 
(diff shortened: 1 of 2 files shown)


st2api/st2api/controllers/v1/actions.py
```diff
@@ -264,6 +264,9 @@ def delete(self, ref_or_id, requester_user):
                 "Exception was %s",
                 e,
             )
+            action = ActionAPI.from_model(action_db)
+            action_db = ActionAPI.to_model(action)
+            Action.add_or_update(action_db)
```
m4dcoder, line 269: Can you modify the unit tests to cover this change?

st2api/st2api/controllers/v1/actions.py
```diff
@@ -264,6 +264,9 @@ def delete(self, ref_or_id, requester_user):
                 "Exception was %s",
                 e,
             )
+            action = ActionAPI.from_model(action_db)
+            action_db = ActionAPI.to_model(action)
```
m4dcoder, line 268: Why do we need to covert the models to ActionAPI and back to ActionDB here?

m4dcoder (changes requested):

mahesh-orch:
@m4dcoder I have polished code and added new unit test to cover the made changes. Please review.

st2api/tests/unit/controllers/v1/test_actions.py
```diff
@@ -626,6 +626,23 @@ def test_delete(self):
...
+    ):
+        msg = "No permission to delete action files from disk"
+        mock_delete_action_files_from_pack.side_effect = PermissionError(msg)
+        post_resp = self.__do_post(ACTION_1)
+        with mock.patch("st2api.controllers.v1.actions.Action.add_or_update"):
+            del_resp = self.__do_delete(
+                self.__get_action_id(post_resp), expect_errors=True
+            )
+            self.assertEqual(del_resp.status_int, 500)
+            self.assertEqual(del_resp.json["faultstring"], msg)
```
m4dcoder, line 644: Can you add a check to make sure that ACTION_1 is added back to the database?

m4dcoder (changes requested):

st2api/st2api/controllers/v1/actions.py
```diff
@@ -264,6 +264,8 @@ def delete(self, ref_or_id, requester_user):
                 "Exception was %s",
                 e,
             )
+            action_db.id = None
+            Action.add_or_update(action_db)
             abort(http_client.INTERNAL_SERVER_ERROR, six.text_type(e))
```
m4dcoder, line 269: Also, if the deletion is related to permission error, can you change the error code to 403 Forbidden here? Otherwise for any other error, return 500 is fine.

m4dcoder (changes requested):

mahesh-orch:
@m4dcoder 
* Unit test has been updated for checking action is getting re-registered to database in case of permission error. Also, separate unit test added for any other exception and action added to database. 
* Error code has changed to 403 (Forbidden) in case of permission error.
Please review. Thanks.

m4dcoder (approved):
LGTM

PR #5354, opened 2021-09-09
Implementation of RBAC for KeyValuePair

ashwini-orchestral:
Implemented RBAC functionality and unit tests for key-value pairs for existing and new permission types. Previously, RBAC feature for key value pairs are not yet implemented.
- Existing permission types - KEY_VALUE_VIEW, KEY_VALUE_SET, KEY_VALUE_DELETE
- New permission types - KEY_VALUE_LIST, KEY_VALUE_ALL

RBAC is enabled in the st2.conf file. Access to a key value pair is checked in the KeyValuePair API controller.
- SET and DELETE permissions will automatically grant LIST and VIEW permissions.
- By default, user has access to his/her own user scoped KVPs without requiring specific permission grant.
- A non-admin user can be explicitly granted permission to one or more system scoped KVPs.
- A non-admin user cannot access another user's KVPs.
- By default, admin and system_admin has ALL access to system scoped KVPs.
- Admin has full access to another user's KVPs (behavior in current version).

This change requires RBAC backend support @ PR https://github.com/StackStorm/st2-rbac-backend/pull/55.

### DIFF
--- a/st2api/st2api/controllers/v1/keyvalue.py
+++ b/st2api/st2api/controllers/v1/keyvalue.py
@@ -97,6 +97,8 @@ class KeyValuePairController(ResourceController):
         key_ref = name
         if scope in [USER_SCOPE, FULL_USER_SCOPE]:
             # Set user scope prefix for the provided user (or current user if user not provided)
+            # NOTE: It's very important raw_filters['prefix'] is set when requesting user scoped
+            # items to avoid information leakage (aka user1 retrieves items for user2)
             key_ref = get_key_reference(name=name, scope=FULL_USER_SCOPE, user=user)
 
         # Setup a kvp database object used for verifying permission
@@ -179,13 +181,11 @@ class KeyValuePairController(ResourceController):
             self._validate_scope(scope=scope)
             raw_filters["scope"] = scope
 
-        # Set prefix which will be used for user-scoped items.
-        # NOTE: It's very important raw_filters['prefix'] is set when requesting user scoped items
-        # to avoid information leakage (aka user1 retrieves items for user2)
-        is_admin = rbac_utils.user_is_admin(user_db=requester_user)
+        # Check if user is granted one of the system roles.
+        has_system_role = rbac_utils.user_has_system_role(user_db=requester_user)
 
         # Check that an admin user has permission to all system scoped items.
-        if is_admin and scope in [ALL_SCOPE, SYSTEM_SCOPE, FULL_SYSTEM_SCOPE]:
+        if has_system_role and scope in [ALL_SCOPE, SYSTEM_SCOPE, FULL_SYSTEM_SCOPE]:
             rbac_utils.assert_user_has_resource_db_permission(
                 user_db=requester_user,
                 resource_db=KeyValuePairDB(scope=FULL_SYSTEM_SCOPE),
@@ -201,6 +201,8 @@ class KeyValuePairController(ResourceController):
             )
 
         # Set user scope prefix for the provided user (or current user if user not provided)
+        # NOTE: It's very important raw_filters['prefix'] is set when requesting user scoped
+        # items to avoid information leakage (aka user1 retrieves items for user2)
         user_scope_prefix = get_key_reference(
             name=prefix or "", scope=FULL_USER_SCOPE, user=user
         )
@@ -213,8 +215,8 @@ class KeyValuePairController(ResourceController):
         kvp_apis_user = []
 
         if scope in [ALL_SCOPE, SYSTEM_SCOPE, FULL_SYSTEM_SCOPE]:
-            # If user is an admin, then retrieve all system scoped items
-            if is_admin:
+            # If user has system role, then retrieve all system scoped items
+            if has_system_role:
                 raw_filters["scope"] = FULL_SYSTEM_SCOPE
                 raw_filters["prefix"] = prefix
 
--- a/st2api/st2api/controllers/v1/keyvalue.py
+++ b/st2api/st2api/controllers/v1/keyvalue.py
@@ -314,9 +314,13 @@ class KeyValuePairController(ResourceController):
         )
 
         # Acquire a lock to avoid race condition between concurrent API calls
-        with self._coordinator.get_lock(self._get_lock_name_for_key(name=key_ref, scope=scope)):
+        with self._coordinator.get_lock(
+            self._get_lock_name_for_key(name=key_ref, scope=scope)
+        ):
             try:
-                existing_kvp_api = self._get_one_by_scope_and_name(scope=scope, name=key_ref)
+                existing_kvp_api = self._get_one_by_scope_and_name(
+                    scope=scope, name=key_ref
+                )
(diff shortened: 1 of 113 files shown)


m4dcoder (changes requested):

st2api/st2api/controllers/v1/keyvalue.py
```diff
@@ -105,9 +105,7 @@ def get_one(
 
         if is_admin and user_query_param_filter:
             # Retrieve values scoped to the provided user
-            user_scope_prefix = get_key_reference(
-                name=name, scope=USER_SCOPE, user=user
-            )
+            user_scope_prefix = get_key_reference(name=name, scope=scope, user=user)
```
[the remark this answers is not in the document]
ashwini-orchestral: @m4dcoder - Now the scope is dynamically passed, so when the scope is not passed then it will return FULL_SYSTEM_SCOPE. Check for scope==USER_SCOPE is already handled in the method.
Screenshot for your reference. Please let us know your suggestions.
keyvalue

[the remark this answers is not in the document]
m4dcoder: Where is scope==USER_SCOPE handled?  Can you give me the link on your branch?  Please don't screenshot. I need to trace the code.

m4dcoder (changes requested):
Where are the unit tests?

st2api/tests/unit/controllers/v1/test_kvps.py
```diff
@@ -296,6 +307,146 @@ def test_get_all_all_scope(self):
...
+        self.assertEqual(put_resp_3.status_int, 200)
+        put_resp_4 = self.__do_put(
+            "key27", {"name": "key27", "value": "val27", "scope": "st2kv.system"}
+        )
+        self.assertEqual(put_resp_4.status_int, 200)
+
+        resp = self.app.get("/v1/keys?scope=system")
+        result = sorted(resp.json, key=lambda k: k["name"])
+
+        # asserting the system scope kvps in the response
```
winem, line 439: The PR looks good to me but I'd add another assert to ensure that kvps the user is *not* allowed to see (`echo`, `new_action`)  aren't in the result.

st2api/st2api/controllers/v1/keyvalue.py
```diff
@@ -332,7 +358,20 @@ def put(self, kvp, name, requester_user, scope=FULL_SYSTEM_SCOPE):
...
+                if scope in [SYSTEM_SCOPE, FULL_SYSTEM_SCOPE]:
+                    if is_admin:
+                        kvp_db = KeyValuePair.add_or_update(kvp_db)
+                    else:
+                        for key in get_all_system_kvp_names_for_user(user):
+                            try:
+                                if key == name:
+                                    kvp_db = KeyValuePair.add_or_update(kvp_db)
+                            except Exception as e:
+                                LOG.error("Unable to get key %s: %s", key, str(e))
```
cognifloyd, line 370: Why is this in a try/except block? Won't the outer try/except block catch this?

Also, I think we need a `break`.
```suggestion
                            if key == name:
                                kvp_db = KeyValuePair.add_or_update(kvp_db)
                                break
```

If we do need the try/except block, then the error message needs to change:
```suggestion
                            try:
                                if key == name:
                                    kvp_db = KeyValuePair.add_or_update(kvp_db)
                            except Exception as e:
                                LOG.error("Unable to put key %s: %s", key, str(e))
```

st2api/st2api/controllers/v1/keyvalue.py
```diff
@@ -395,7 +478,19 @@ def delete(self, name, requester_user, scope=FULL_SYSTEM_SCOPE, user=None):
...
+                if scope in [SYSTEM_SCOPE, FULL_SYSTEM_SCOPE]:
+                    if is_admin:
+                        KeyValuePair.delete(kvp_db)
+                    else:
+                        for key in get_all_system_kvp_names_for_user(user):
+                            try:
+                                if key == name:
+                                    KeyValuePair.delete(kvp_db)
+                            except Exception as e:
+                                LOG.error("Unable to get key %s: %s", key, str(e))
```
cognifloyd, line 490: Why is this in a try/except block? Won't the outer try/except block catch this?

Also, I think we need a `break`.
```suggestion
                            if key == name:
                                kvp_db = KeyValuePair.add_or_update(kvp_db)
                                break
```

If we do need the try/except block, then the error message needs to change:
```suggestion
                            try:
                                if key == name:
                                    kvp_db = KeyValuePair.add_or_update(kvp_db)
                            except Exception as e:
                                LOG.error("Unable to delete key %s: %s", key, str(e))
```

st2common/st2common/services/keyvalues.py
```diff
@@ -256,3 +260,34 @@ def get_key_reference(scope, name, user=None):
         raise InvalidScopeException(
             'Scope "%s" is not valid. Allowed scopes are %s.' % (scope, ALLOWED_SCOPES)
         )
+
+
+def get_uids(user):
```
cognifloyd, line 265:
```suggestion
def get_uid(user):
```
Why is this `uids`?

cognifloyd (changes requested):

st2api/tests/unit/controllers/v1/test_kvps.py
m4dcoder: Refactored the unit tests so your comments here should be addressed.

cognifloyd (approved):
LGTM

st2common/st2common/services/keyvalues.py
m4dcoder: Because it is returning the specific uid that RBAC assign to a resource. (i.e. key_value_pair:st2kv.system:key1).  It's plural because the function is trying to return all (multiple) resources that have been explicitly granted for a user.

m4dcoder:
Moving this feature to v3.7.0 to give more time for folks to soak this in.

m4dcoder (approved):

m4dcoder (changes requested):
Blocking this temporary until post v3.6.0 release.

st2client/st2client/commands/resource.py
```diff
@@ -198,10 +197,12 @@ def get_resource_by_pk(self, pk, **kwargs):
         try:
             instance = self.manager.get_by_id(pk, **kwargs)
         except Exception as e:
-            traceback.print_exc()
+            # traceback.print_exc()
```
m4dcoder, line 200: @ashwini-orchestral Please remove this line instead of commenting it out. It's a bad practice and no one will come back and clean this up.

m4dcoder (changes requested):

m4dcoder (approved):

armab:
@ashwini-orchestral The PR needs a Changelog. Please add one.

@m4dcoder Additionally, do we need documentation changes for this new feature as well?

m4dcoder:
@armab Corresponding PR for docs at https://github.com/StackStorm/st2docs/pull/1092

PR #5359, opened 2021-09-16
Use Jinja sandboxed environment

amanda11:
Use Jinja sandboxed environment, so that any templates that attempt to access insecure code will be rejected.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -37,6 +37,10 @@ Changed
 
   Contributed by @khushboobhatia01
 
+* Move to using Jinja sandboxed environment #5359
+
+  Contributed by Amanda McGuinness (@amanda11 Ammeon Solutions)
+
 Fixed
 ~~~~~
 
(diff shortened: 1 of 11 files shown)


nzlosh (approved):

m4dcoder (changes requested):
- Do we need to update Jinja >= 2.11 in requirements?
- Please update orquesta version in requirements to use the commit with the similar fix (after https://github.com/StackStorm/orquesta/pull/240 is merged).
- Please add a changelog entry.
- Please update description of this PR on why this fix is needed.

amanda11:
> * Do we need to update Jinja >= 2.11 in requirements?
> * Please update orquesta version in requirements to use the commit with the similar fix (after [StackStorm/orquesta#240](https://github.com/StackStorm/orquesta/pull/240) is merged).
> * Please add a changelog entry.
> * Please update description of this PR on why this fix is needed.

@m4dcoder I've addressed your requirements, Jinja is already at 2.11 in the requirements on ST2 requirements.

m4dcoder (approved):

PR #5360, opened 2021-09-17
Action delete API rework - support for action file deletion

mahesh-orch:
* Action delete API modification to have option to delete action database entry without removing files from disk
* From CLI `st2 action delete <pack>.<action>` will delete database entry only
* From CLI `st2 action delete --remove-files <pack>.<action>` will delete database entry and files.
* If action DELETE method has additional argument as `{"remove_files": true}` in the body then it will remove related action files from disk along with database entry.
* `{"remove_files": false}` or no additional argument in action DELETE method body will remove only database entry.

### DIFF
--- a/st2api/st2api/controllers/v1/actions.py
+++ b/st2api/st2api/controllers/v1/actions.py
@@ -206,7 +206,7 @@ class ActionsController(resource.ContentPackResourceController):
 
         return action_api
 
-    def delete(self, ref_or_id, requester_user):
+    def delete(self, files_remove_request, ref_or_id, requester_user):
         """
         Delete an action.
 
@@ -240,6 +240,7 @@ class ActionsController(resource.ContentPackResourceController):
         pack_name = action_db["pack"]
         entry_point = action_db["entry_point"]
         metadata_file = action_db["metadata_file"]
+        remove_files = files_remove_request.remove_files
 
         try:
             Action.delete(action_db)
@@ -252,20 +253,30 @@ class ActionsController(resource.ContentPackResourceController):
             )
             abort(http_client.INTERNAL_SERVER_ERROR, six.text_type(e))
             return
-        try:
-            delete_action_files_from_pack(
-                pack_name=pack_name,
-                entry_point=entry_point,
-                metadata_file=metadata_file,
-            )
-        except Exception as e:
-            LOG.error(
-                "Exception encountered during deleting resource files from disk."
-                "Exception was %s",
-                e,
-            )
-            abort(http_client.INTERNAL_SERVER_ERROR, six.text_type(e))
-            return
+
+        if remove_files:
+            try:
+                delete_action_files_from_pack(
+                    pack_name=pack_name,
+                    entry_point=entry_point,
+                    metadata_file=metadata_file,
+                )
+            except PermissionError as e:
+                LOG.error("No permission to delete resource files from disk.")
+                action_db.id = None
+                Action.add_or_update(action_db)
+                abort(http_client.FORBIDDEN, six.text_type(e))
+                return
+            except Exception as e:
+                LOG.error(
+                    "Exception encountered during deleting resource files from disk. "
+                    "Exception was %s",
+                    e,
+                )
+                action_db.id = None
+                Action.add_or_update(action_db)
+                abort(http_client.INTERNAL_SERVER_ERROR, six.text_type(e))
+                return
 
         extra = {"action_db": action_db}
         LOG.audit("Action deleted. Action.id=%s" % (action_db.id), extra=extra)
--- a/st2api/st2api/controllers/v1/actions.py
+++ b/st2api/st2api/controllers/v1/actions.py
@@ -206,7 +206,7 @@ class ActionsController(resource.ContentPackResourceController):
 
         return action_api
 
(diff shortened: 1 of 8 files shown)


st2api/st2api/controllers/v1/actions.py
```diff
@@ -206,7 +206,7 @@ def put(self, action, ref_or_id, requester_user):
 
         return action_api
 
-    def delete(self, ref_or_id, requester_user):
+    def delete(self, files_remove_request, ref_or_id, requester_user):
```
m4dcoder, line 209: Can you rename `files_remove_request` to `options`?

st2common/st2common/openapi.yaml.j2
```diff
@@ -409,6 +409,11 @@ paths:
           description: Entity reference or id
           type: string
           required: true
+        - name: files_remove_request
```
m4dcoder, line 412: Can you rename `files_remove_request` to `options`?

st2common/st2common/openapi.yaml
```diff
@@ -413,6 +413,11 @@ paths:
           description: Entity reference or id
           type: string
           required: true
+        - name: files_remove_request
```
m4dcoder, line 416: Can you rename `files_remove_request` to `options`?

st2client/st2client/commands/action.py
```diff
@@ -282,31 +282,24 @@ def __init__(self, resource, *args, **kwargs):
...
         self.parser.add_argument(
-            "-f",
-            "--force",
+            "-rf",
+            "--remove-files",
             action="store_true",
-            dest="force",
-            help="Auto yes flag to delete action files from disk.",
+            dest="remove_files",
+            help="Remove action files from disk.",
```
m4dcoder, line 289: Do we need to add `default=False`?

st2client/st2client/commands/action.py
```diff
@@ -282,31 +282,24 @@ def __init__(self, resource, *args, **kwargs):
...
-                "The resource files on disk will be deleted. Do you want to continue? (y/n): "
-            )
-        if args.force or user_input.lower() == "y" or user_input.lower() == "yes":
-            self.manager.delete(instance, **kwargs)
-            print(msg)
+        if args.remove_files:
+            remove_files = True
         else:
-            print("Action is not deleted.")
+            remove_files = False
```
m4dcoder, line 299: Since the remove_files arg is already defined as boolean above, you can change this if/else block to `remove_files = args.remove_files`.

st2api/tests/unit/controllers/v1/test_actions.py
```diff
@@ -854,3 +856,12 @@ def __do_delete(self, action_id, expect_errors=False):
         return self.app.delete(
             "/v1/actions/%s" % action_id, expect_errors=expect_errors
         )
+
```
m4dcoder, line 859: - Update existing unit test `test_delete` where there's no payload and assert database record is removed and mocked delete_action_files_from_pack is not called
- Add new unit test `test_delete_remove_files_false` where remove_files=False in payload and assert database record is removed and mocked delete_action_files_from_pack is not called
- Add new unit test `test_delete_remove_files_true` where remove_files=True in payload and assert database record is removed and mocked delete_action_files_from_pack is called

m4dcoder (changes requested):

ashwini-orchestral:
@m4dcoder I have added unit tests, updated existing unit test and done other minor changes as suggested. Please review. Thanks.

mahesh-orch:
@m4dcoder I have added unit tests, updated existing unit test and done other minor changes as suggested. Please review. Thanks.

m4dcoder (approved):
LGTM

st2client/st2client/commands/action.py
```diff
@@ -282,31 +282,21 @@ def __init__(self, resource, *args, **kwargs):
         super(ActionDeleteCommand, self).__init__(resource, *args, **kwargs)
 
         self.parser.add_argument(
-            "-f",
-            "--force",
+            "-rf",
```
nzlosh, line 285: Single hyphen flags should be a single letter only.  Under Linux, it's possible to have flags grouped by a single hyphen which would make the `-rf` flag seem like two flags `-r -f` were being passed. `-r` should be sufficient for the `remove_files` flag unless it's already taken.

st2common/st2common/openapi.yaml.j2
```diff
@@ -5380,6 +5388,14 @@ definitions:
         items:
           type: string
     additionalProperties: false
+  ActionDeleteSchema:
+    type: object
+    properties:
+      remove_files:
+        type: boolean
+        description: Force delete action files from disk
```
amanda11, line 5396: Minor comment - should this be reworded as in ohter schema, as its not really a force - its just a flag.

amanda11 (commented):
Just a minor comment on wording in the schema, but @nzlosh point more important.

mahesh-orch:
@nzlosh I have changed CLI action delete optional argument `-rf` to `-r`.
@amanda11 I have updated the description of ActionDeleteSchema `remove-files` argument.
Please review. Thanks.

nzlosh (approved):

PR #5367, opened 2021-09-21
Implement proper workflow engine shutdown

khushboobhatia01:
This pull request implements graceful workflow engine shutdown. On shutdown, we stop consuming new requests and give existing active requests some time to complete before we kill the engine.

### DIFF
--- a/conf/st2.conf.sample
+++ b/conf/st2.conf.sample
@@ -345,6 +345,8 @@ webui_base_url = https://localhost
 gc_max_idle_sec = 0
 # Location of the logging configuration file.
 logging = /etc/st2/logging.workflowengine.conf
+# How much time to give to the request in progress to finish in seconds before killing workflow engine.
+request_shutdown_time = 0
 # Max jitter interval to smooth out retries.
 retry_max_jitter_msec = 1000
 # Max time to stop retrying.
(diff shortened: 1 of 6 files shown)


m4dcoder (changes requested):
If we are implementing proper shutdown, this is not sufficient.

1. On shutdown, the workflow engine should remove itself from the service registry.
2. If there are other workflow engines in the service registry, then do the shutdown sequence. The shutdown sequence should be something like stop accepting incoming requests and wait until this workflow engine completed processing any requests (not some arbitrary sleep).
3. Else if there are no other workflow engine in the service registry, then the workflow engine needs to stop accepting requests, pause all running workflow executions, and then proceed with rest of the shutdown sequence.
4. On workflow engine startup, any workflow executions paused by the shutdown should be resumed.

amanda11:
@khushboobhatia01 Have we agreed to go for a bigger fix, if so is it ok if i remove this PR from the 3.6 milestone?

khushboobhatia01:
Hi @amanda11, I think we can go ahead and remove the milestone tag. I've started working on the approach we've discussed and will close this MR.

amanda11:
> Hi @amanda11, I think we can go ahead and remove the milestone tag. I've started working on the approach we've discussed and will close this MR.

Thanks for update

PR #5368, opened 2021-09-22
Add RBAC permission types for key value pairs

m4dcoder:
Add RBAC permission types for key value pairs so the changes at https://github.com/StackStorm/st2-rbac-backend/pull/55 to introduce RBAC to KVP can pass CI.

### DIFF
--- a/st2common/st2common/rbac/types.py
+++ b/st2common/st2common/rbac/types.py
@@ -602,7 +602,7 @@ PERMISION_TYPE_TO_DESCRIPTION_MAP = {
     PermissionType.KEY_VALUE_SET: ("Ability to set a Key-Value Pair."),
     PermissionType.KEY_VALUE_DELETE: ("Ability to delete an existing Key-Value Pair."),
     PermissionType.KEY_VALUE_ALL: (
-        "Ability to perform all the support operations on a Key-Value Pair."
+        "Ability to perform all the supported operations on a Key-Value Pair."
     ),
     PermissionType.TRACE_LIST: ("Ability to list (view all) traces."),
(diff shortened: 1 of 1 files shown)


amanda11 (changes requested):
Shouldn't the new permission types be added either to the resource type they belong to, or to global permission array if they are global types?
At moment I don't think they are in either so can't be used.

m4dcoder:
I added to the resource type to permission type map. The global permission type dynamically applies the list permission.

st2common/st2common/rbac/types.py
```diff
@@ -593,9 +597,13 @@ class SystemRole(Enum):
     PermissionType.API_KEY_ALL: (
         "Ability to perform all the supported operations on an API Key."
     ),
+    PermissionType.KEY_VALUE_LIST: ("Ability to list (view all) Key-Value Pairs."),
     PermissionType.KEY_VALUE_VIEW: ("Ability to view Key-Value Pairs."),
     PermissionType.KEY_VALUE_SET: ("Ability to set a Key-Value Pair."),
     PermissionType.KEY_VALUE_DELETE: ("Ability to delete an existing Key-Value Pair."),
+    PermissionType.KEY_VALUE_ALL: (
+        "Ability to perform all the support operations on a Key-Value Pair."
```
amanda11, line 605: Probably should be "supported" instead of "support" - like the others.

amanda11 (approved):
LGTM - just a minor reword required - but approved so can be merged after that is fixed.

cognifloyd (approved):

PR #5376, opened 2021-10-01
Pin networkx to versions lower than 2.6 which doesn't support Python 3.6

nzlosh:
Just seen on CentOS7/8 during package build process:
```
[wheelhouse: st2] [13:58:59]      Processing /tmp/wheelhouse/networkx-2.6.3-py3-none-any.whl
[wheelhouse: st2] [13:58:59]        File was already downloaded /tmp/wheelhouse/networkx-2.6.3-py3-none-any.whl
[wheelhouse: st2] [13:58:59]      ERROR: Package 'networkx' requires a different Python: 3.6.8 not in '>=3.7'
[wheelhouse: st2] [13:58:59]      WARNING: You are using pip version 20.3.3; however, version 21.2.4 is available.
[wheelhouse: st2] [13:58:59]      You should consider upgrading via the '/usr/bin/python3 -m pip install --upgrade pip' command.
[wheelhouse: st2] [13:58:59]      make: *** [.stamp-wheelhouse] Error 1
ERROR: 1

Exited with code exit status 1
```
This PR pins networkx to versions lower than 2.6

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -92,7 +92,7 @@ jobs:
           # TODO: maybe make the virtualenv a partial cache to exclude st2*?
           # !virtualenv/lib/python*/site-packages/st2*
           # !virtualenv/bin/st2*
-          key: ${{ runner.os }}-v2-python-${{ matrix.python-version }}-${{ hashFiles('requirements.txt', 'test-requirements.txt') }}
+          key: ${{ runner.os }}-v3-python-${{ matrix.python-version }}-${{ hashFiles('requirements.txt', 'test-requirements.txt') }}
           restore-keys: |
             ${{ runner.os }}-v2-python-${{ matrix.python }}-
       - name: Cache APT Dependencies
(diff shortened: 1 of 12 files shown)


armab:
Do we need a Changelog for this?

Per https://app.circleci.com/pipelines/github/StackStorm/st2/2203/workflows/a41eaa05-3034-4f36-aa22-372ca0f83aea/jobs/13713  Ubuntu Bionic build is failing on:
```
[package: st2] [15:15:58]      ERROR: Cannot install -r ./requirements.txt (line 30) and -r ./requirements.txt (line 36) because these package versions have conflicting dependencies.
[package: st2] [15:15:58]      The conflict is caused by:
[package: st2] [15:15:58]          jsonpath-rw 1.4.0 depends on decorator
[package: st2] [15:15:58]          networkx 2.5.1 depends on decorator<5 and >=4.3
```
Or do you think it'll get fixed by the Orquesta https://github.com/StackStorm/orquesta/pull/242 change?

winem:
I can confirm that the Orquesta-PR should fix the failure on U18. 

networkx does no longer require "decorator" as stated in the release notes here: https://github.com/networkx/networkx/blob/v2.6/doc/release/release_2.6.rst

I also checked the code of networkx 2.6 and can confirm that the dependency was dropped (as expected).

cognifloyd:
Here's the issue with decorator:
networkx depends on `decorator<5`, so it should resolve `decorator-4.4.2`. On python3.6, it does just that. On python3.8, however, the wheel does not say that it supports python3.8. That was added 1 commit after the wheel for 4.4.2 was uploaded to pypi. In other words, the commit tagged 4.4.2 on github is one commit past what was used to build the wheel on pypi.
Let's try pinning decorator, a transient dep, to see if that helps pip figure out what to do.

cognifloyd:
Yeah! All CI is passing now

cognifloyd (approved):

cognifloyd:
Hmm. Maybe we should merge https://github.com/StackStorm/orquesta/pull/242 and update the orquesta commit in this PR as well.

winem (approved):
Thanks!

nzlosh:
Thanks for the help @cognifloyd and @winem

PR #5377, opened 2021-10-03
use python3 in scripts/populate-package-meta.sh

cognifloyd:
I noticed a python3 issue while reviewing st2-packaging logs (as `python` does not exist on newer distros), so here's the fix.


armab (approved):
:+1:

PR #5096, opened 2020-11-29
Refactor FileWatchSensor to remove logshipper

blag:
This PR refactors the `FileWatchSensor` from the `linux` pack, to remove it's dependency on `logshipper`, and use `watchdog` instead.

Upstream `logshipper` hasn't been touched in years, so we switched to our own fork. That hasn't been well maintained either, and the `Tail` implementation, which is all we use from it, is a bit of a mess on top of depending on `pyinotify` has been itself unmaintained for even longer.

This PR borrows some code from the previous sensor, but also breaks up the code into well structured classes that all have a single focus, possibly allowing this code to be reused elsewhere. Additionally, you can run the sensor Python script itself, which should help serve as an example of how to write a "good" sensor.

The `watchdog` project is now being actively maintained, has wide support for Linux, macOS, BSD, Windows, and a polling fallback implementation. The `pyinotify` package refuses to even install on macOS (not being able to run is understandable, but preventing installation is too limiting IMO).

I realize that it's a bit odd to make the `linux` depend less on Linux-specific features, but since `pyinotify` refused to even install on macOS, I couldn't do simple things like run `make requirements` without Linux running in a full Docker or VM environment.

This PR also fixes a bug in `logshipper` (has to do with lines that do not end in newlines - previous code only took the last character, new code takes the whole line), and also presents a workaround for another possible bug in the comments.

Currently being held up by #5095.

PR #5380, opened 2021-10-06
silence pylint warning about pika dep in tool

cognifloyd:
Every now and then, pylint gets upset about a `pika` import in `tools/direct_queue_publisher.py`. Here's a snippet from a GHA run in #5096 (NB: that PR does not touch any of these files, so fixing this in that PR doesn't make sense):

```
. virtualenv/bin/activate; pylint -j 4 -E --rcfile=./lint-configs/python/.pylintrc --load-plugins=pylint_plugins.api_models tools/*.py || exit 1;
************* Module tools.direct_queue_publisher
tools/direct_queue_publisher.py:21:0: E0401: Unable to import 'pika' (import-error)
make: *** [Makefile:351: .pylint] Error 1
```
https://github.com/StackStorm/st2/pull/5096/checks?check_run_id=3808812826#step:11:467

https://github.com/StackStorm/st2/blob/8496bb2407b969f0937431992172b98b545f6756/tools/direct_queue_publisher.py#L21

I think it's weird that we're not consistently getting this error because pylint should be running on all the tools files:

https://github.com/StackStorm/st2/blob/f65f62f06a27511356392b33cd944ce946f10f74/Makefile#L361

This tool dates back to April 2016 (f6897d26dd06651f3bad86da6497a169a739866f) with this in the commit message:

> * Exisiting tool queue_producer published to an exchange. This is for direct publishing to a queue.
> * Written on top of pika instead of kombu as this was simpler to write given code samples on rabbitmq tutorial site

Plus, from what I can tell, kombu doesn't offer a way to publish directly to a queue; it requires an exchange. So this tool has to use something other than kombu.

So, we can either delete this script or leave it around for possible future debugging. To leave it around, we should protect it such that pylint doesn't complain and anyone who tries to use it gets a helpful message about why it doesn't work. That's what I did in this PR.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -71,6 +71,9 @@ Changed
   This functionality should never be used in production, but only in development environments or
   similar when debugging code. #5199
 
+* Silence pylint about dev/debugging utility (tools/direct_queue_publisher.py) that uses pika because kombu
+  doesn't support what it does. If anyone uses that utility, they have to install pika manually. #5380
+
 Fixed
 ~~~~~
 


amanda11 (approved):
LGTM, but could you also add a changelog entry,

Issue #5310, opened 2021-07-27
Using a different python version for a pack is no longer working. actionrunner.python_binary


wingiti:
## SUMMARY
Using a different python version (in my case 3.7) for packs instead of the default system one (3.6) is not working.
Till Stackstorm 3.4 this worked like described in the docs:
https://docs.stackstorm.com/latest/packs.html?highlight=python_binary#python-versions-in-pack-python-virtual-environment

### STACKSTORM VERSION

Paste the output of ``st2 --version``:
st2 3.5.0, on Python 3.6.8

##### OS, environment, install method

I think it doesn't matter. I tried on our System RHEL 7 and the OVA image of Stackstorm 3.5.0

## Steps to reproduce the problem

Add a different python binary to st2.conf for packs:
```
[actionrunner]
python_binary = /usr/bin/python3.7
```

## Expected Results

The pack should be run using python 3.7 and finish successful.

## Actual Results

The execution stops and throws an error:
```
Traceback (most recent call last):
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/python_runner/python_action_wrapper.py", line 35, in <module>
    import orjson
ModuleNotFoundError: No module named 'orjson'
```

Using some debug output in my action I was able to verify that the binary mentioned in the config is used.

As further debug step I tried following config and installed Stackstorm requirements for python3.7:
```
[actionrunner]
python_binary = /usr/bin/python3.7
virtualenv_opts =  --system-site-packages
```

It resulted in a different error:
```
Traceback (most recent call last):
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/python_runner/python_action_wrapper.py", line 59, in <module>
    from st2common import log as logging
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2common/log.py", line 32, in <module>
    from st2common.logging.handlers import FormatNamedFileHandler
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2common/logging/handlers.py", line 24, in <module>
    from st2common.util import date as date_utils
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2common/util/date.py", line 24, in <module>
    import udatetime
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/udatetime/__init__.py", line 20, in <module>
    from udatetime.rfc3339 import (
ModuleNotFoundError: No module named 'udatetime.rfc3339'
```

[...]

PR #5388, opened 2021-10-12
Idea to fix issue #5310 - Using site packages of a different python version for actionrunner

wingiti:
idea how to use site-package of custom python version for action runners.


st2common/st2common/util/sandboxing.py
```diff
@@ -137,6 +137,26 @@ def get_sandbox_python_path_for_python_action(
 
     pack_base_path = get_pack_base_path(pack_name=pack)
     virtualenv_path = get_sandbox_virtualenv_path(pack=pack)
+    python_binary = cfg.CONF.actionrunner.python_binary
+
+    # Add the custom python's site-packages directory in front of the Python
+    # system site-packages if python_binary is not default.
+    if python_binary is not sys.executable and os.path.isfile(python_binary):
+        python_version = python_binary.rsplit("/", 1)[1]
```
cognifloyd, line 145: Can you give me an example of how this is supposed to work?
eg if someone sets `cfg.CONF.actionrunner.python_binary` to `/usr/local/bin/python` for some custom build of python (or whatever), then that path is not going to have a "version" in its string.

st2common/st2common/util/sandboxing.py
```diff
@@ -137,6 +137,26 @@ def get_sandbox_python_path_for_python_action(
 
     pack_base_path = get_pack_base_path(pack_name=pack)
     virtualenv_path = get_sandbox_virtualenv_path(pack=pack)
+    python_binary = cfg.CONF.actionrunner.python_binary
```
cognifloyd, line 140: OK. I see that this is already defined. I'll just leave this note here for the next time I review this so I don't have to look it up again.

https://github.com/StackStorm/st2/blob/f5c0c0fcac41e743460de976723dc1decb144578/st2common/st2common/config.py#L454-L458

cognifloyd (changes requested):
I have a note (for me to avoid future confusion) ... You can ignore that.

And I have a question on how this is supposed to work.

Then, we just need some tests to make sure using a custom python binary for python actions doesn't break with future python updates.

st2common/st2common/util/sandboxing.py
wingiti: Hi @cognifloyd, as mentioned in the title already this is an idea and I am open for better solutions. I just created it based on what I currently see on my system and found on the internet about python and site packages.

If someone just sets `/usr/local/bin/python` the variable name `python_version` might be misleading (I just thought about using different python binaries containing version digits, like python3.7, etc. when I chose the name) but should still to the job, if the site-packages are at the same directories as "usual". Maybe `binary_name` would be better...

What I have seen so far is that site-packages are usually located at:
`/usr/lib/<binary_name>/site-packages`
`/usr/lib/<binary_name>/dist-packages`
`/usr/local/lib/<binary_name>/site-packages`
`/usr/local/lib/<binary_name>/dist-packages`

That is the reason why I created `system_prefix_dirs` and `system_dir_names`. But it might be that there are linux distributions or custom configs out there using different locations. In this case the solution will not work.

The idea behind these lines was to fix the issue without additional config options as a user might think it should work out of the box when he sets the `python_binary` config option. This is what I thought as well... ;)

If you want to catch all possibilities, it might be better to implement a new config option like `python_binary_site-packages`. In this case a user needs to set up two config settings and needs to know where his site packages are located, but he has full control.

Third possibility would be to use both. Look for `python_binary_site-packages` and if not defined try the lines above.

cognifloyd: Instead of hard-coding the list of site-packages dirs, what if you looked for the dir relative to the `python_binary`? So `Path(python_binary).parent.parent / "lib" / "site-packages"` and variants for `lib*` directories and either `site-` or `dist-`

PR #5390, opened 2021-10-15
Explicitly mention cffi

amanda11:
With the introduction of cffi version 1.15.0 then on CentOS 7 and 8 builds, it started to look for libffi.so.8 - however on EL repos when libffi-devel is installed, libffi.so.6 is installed.
Fixed cffi to version < 1.15.0 so that build succeeds on EL 7/8.


armab:
Thanks a lot for fixing the master build :+1:

PR #5396, opened 2021-10-19
Deregister services from service registry on shutdown

khushboobhatia01:
Few services like worker, workflow engine, scheduler and notifier perform a shutdown sequence before exiting the process. This PR will ensure that the services deregister themselves from service registry on shutdown and then continue with shutdown sequence. 

(PR based on https://github.com/StackStorm/st2/discussions/5373 step 1)

### DIFF
--- a/st2common/tests/unit/test_service_setup.py
+++ b/st2common/tests/unit/test_service_setup.py
@@ -23,6 +23,7 @@ import mock
 from oslo_config.cfg import ConfigFilesNotFoundError
 
 from st2common import service_setup
+from st2common.services import coordination
 from st2common.transport.bootstrap_utils import register_exchanges
 from st2common.transport.bootstrap_utils import QUEUES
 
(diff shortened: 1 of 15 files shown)


st2common/tests/unit/test_service_setup.py
```diff
@@ -216,3 +217,23 @@ def mock_get_logging_config_path():
...
+        self.assertEqual(len(list(members.get())), 1)
+        service_setup.deregister_service(service)
+        self.assertEqual(len(list(members.get())), 0)
+
+    def test_deregister_service_when_service_registry_disables(self):
+        service = "api"
+        try:
+            service_setup.deregister_service(service)
+        except:
+            assert False, "service_setup.deregister_service raised exception"
```
m4dcoder, line 237: @khushboobhatia01 What happens if this error is raised on shutdown? I think it is ok to make this silent if there is nothing to be deregister.

m4dcoder (changes requested):
Minor concern on error handling during deregistration.

st2common/tests/unit/test_service_setup.py
khushboobhatia01: @m4dcoder We don't expect deregister() to raise an exception when service_registry is disabled or there's nothing to deregister. This test case would fail only when some unintended change is causing deregister to fail. All edge cases are handled in deregister(). Therefore it might be good to know if some new change is causing it to fail.

m4dcoder (approved):
LGTM. Thanks for your contribution and patience.

armab (commented):
@khushboobhatia01 Could you please also add Changelog for this PR before we merge it?

khushboobhatia01:
> @khushboobhatia01 Could you please also add Changelog for this PR before we merge it?

Done @armab. Thanks for reminding :)

armab:
Great stuff! Thanks @khushboobhatia01 for the work and @m4dcoder for review! :+1:

PR #5405, opened 2021-10-24
Add workflow system info in Workflow Execution context

khushboobhatia01:
This PR adds workflow engine system information to WorkflowExecution context.
Ref https://github.com/StackStorm/st2/discussions/5373

```
{
    "engine_info": {
        "hostname": "st2", 
        "pid": 1085262
    }, 
    "parent": {
        "user": "stanley", 
        "pack": "examples"
    }, 
    "st2": {
        "action_execution_id": "6174ce479195da5bbddf5443", 
        "api_url": "http://127.0.0.1:9101/v1", 
        "runner": "orquesta", 
        "workflow_execution_id": "6174ce48f3b5fca8fa795f97", 
        "user": "stanley", 
        "action": "examples.orquesta-sequential", 
        "pack": "examples"
    }
}
```

This information will be used to identify orphaned workflow execution.

### DIFF
--- a/st2common/st2common/services/workflows.py
+++ b/st2common/st2common/services/workflows.py
@@ -20,6 +20,8 @@ import datetime
 import retrying
 import six
 
+from functools import wraps
+
 from orquesta import conducting
 from orquesta import events
 from orquesta import exceptions as orquesta_exc
@@ -49,6 +51,7 @@ from st2common.services import executions as ex_svc
 from st2common.util import action_db as action_utils
 from st2common.util import date as date_utils
 from st2common.util import param as param_utils
+from st2common.util import system_info
 
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


m4dcoder (changes requested):
1. The action execution that corresponds to the workflow execution should have the system info recorded. Do we need to duplicate that here in the workflow execution document?
2. The workflow execution can be processed by 1 or more different workflow engine during its execution. I see the need to update the system info in the action execution whenever the workflow execution is being picked up again and processed. Can you visit the workflow service at https://github.com/StackStorm/st2/blob/master/st2common/st2common/services/workflows.py and review functions that start with `handle_*`? These functions are usually the entry point for workflow engine to rehydrate a workflow execution.
3. Can you explain how you will determine that a workflow execution is orphaned? The engine info here is not sufficient because this just tells us workflow engine "X" is the last process that worked on this workflow execution and it's not actually orphaned.

Kami:
I won't have time to perform a detailed review, but just two quick comments / questions:

1. Performance implications

It appears that depending on the object passed to the function, we will either perform 1 or 2 additional GET one and 1 (2? I don't know why we need that delete at the end in finally) additional UPDATE query each time ``handle_workflow_execution`` and ``handle_action_execution`` is called.

A while back I spent a substantial amount of time of profiling and optimizing performance of that code path (which is considered a hot code path since it's executed often and has major performance implication on how long running / scheduling executions and workflows takes).

And per @m4dcoder coder, that code can be executed multiple times (each time execution it's picked up by workflow engine) so this makes it even more problematic.

Even though we made a lot of performance improvements which should make those GET and UPDATE queries much faster now, we should still think about it some more and see if there is any way to reduce number of queries and make that code more efficient.

2. Code formatting / style

I don't think we should use a decorator.

It makes code harder to read and understand.

It should be a regular function / method (or two of them so we can get rid of *args and **kwargs stuff which also makes code hard to read and understand) which gets called in places where it's needed.

st2common/st2common/services/workflows.py
```diff
@@ -1561,3 +1564,38 @@ def identify_orphaned_workflows():
             continue
 
     return orphaned
+
+
+def add_system_info_to_action_context(func):
```
Kami, line 1569: Per my other comment, I think we should refactor this 1 or 2 separate functions and not use a decorator.

st2common/st2common/services/workflows.py
```diff
@@ -1561,3 +1564,38 @@ def identify_orphaned_workflows():
...
+
+        lv_ex.context.update({"engine_info": engine_info})
+        lv_db_access.LiveAction.add_or_update(lv_ex, publish=False)
+
+        try:
+            return func(*args, **kwargs)
+        except Exception as e:
+            raise e
+        finally:
+            lv_ex = lv_db_access.LiveAction.get_by_id(lv_ex.id)
```
Kami, line 1595: It's not clear why we need this code path so we should add a comment on why this delete is here and why it's needed.

Kami (changes requested):

st2common/st2common/services/workflows.py
khushboobhatia01: @Kami The intention here is to add engine information to the context at the start of handling and remove it once handling is complete.
Any action in running/pausing/cancelling state, having engine information about a deregistered service can be classified as orphaned workflow.

khushboobhatia01: Since the workflow execution can be processed by 1 or more different workflow engine during its execution, it's difficult to know which workflow engine is currently handling the execution.

cognifloyd:
@khushboobhatia01 what are your plans with this PR? Do you know when you'll be able to pick it up again and address some of the feedback here?

Issue #3459, opened 2017-06-08
ST2API doesn't parse percent encoded URIs


bigmstone:
Tracking StackStorm/st2web#357 here.

PR #5415, opened 2021-10-27
WIP: Labels

cognifloyd:
Experimenting with a label changing workflow. Do not merge.
The idea is to adjust the label colors and descriptions in a file that we check into the repo.

Current labels list is an export made with https://github.com/micnncim/label-exporter
Eventually, I'll work on updating the label colors and descriptions.

tests of labels workflow:
- :heavy_check_mark: the dry-run says it has nothing to change before there are any changes: https://github.com/StackStorm/st2/runs/4023190147?check_suite_focus=true#step:3:260

### DIFF
--- a/.github/labels.yml
+++ b/.github/labels.yml
@@ -142,136 +142,136 @@
 - color: 0075ca # blue (standard github)
   description: Improvements or additions to documentation
   name: documentation
-- color: FBCA04
+- color: 0075ca # blue
   description: ""
   name: API
-- color: FBCA04
+- color: 0075ca # blue
   description: ""
   name: API/v2
-- color: FBCA04
+- color: 0075ca # blue
   description: ""
   name: CLI
-- color: 16bdff
+- color: 0075ca # blue
   description: ""
   name: component:st2auth
-- color: 16bdff
+- color: 0075ca # blue
   description: ""
   name: component:st2client
-- color: 16bdff
+- color: 0075ca # blue
   description: ""
   name: component:st2web
-- color: 0F0000
+- color: 0075ca # blue
   description: ""
   name: migrations
-- color: 0000ff
+- color: 0075ca # blue
   description: ""
   name: runners
-- color: 16bdff
+- color: 0075ca # blue
   description: ""
   name: 'service: action runner'
-- color: 16bdff
+- color: 0075ca # blue
   description: ""
   name: 'service: api'
-- color: 16bdff
+- color: 0075ca # blue
   description: ""
   name: 'service: auth api'
-- color: 16bdff
+- color: 0075ca # blue
   description: ""
   name: 'service: rules engine'
-- color: 16bdff
+- color: 0075ca # blue
   description: ""
   name: 'service: scheduler'
-- color: 16bdff
+- color: 0075ca # blue
   description: ""
   name: 'service: workflow engine'
 
 # ##### area/python labels ##### #
-- color: E9F68C
+- color: 0075ca # blue
   description: ""
   name: python3
-- color: E9F68C
+- color: 0075ca # blue
   description: ""
   name: unicode
 
 # ##### area/pack_resources labels ##### #
-- color: 006B75
+- color: 003459 # blue
   description: ""
   name: action aliases
-- color: 006B75
+- color: 003459 # blue
   description: ""
   name: chatops
-- color: 006B75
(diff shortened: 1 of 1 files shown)


.github/labels.yml
```diff
@@ -0,0 +1,277 @@
...
+  description: ""
+  name: tests
+
+# ##### area/code_location labels ##### #
+- color: 0075ca # blue (standard github)
+  description: Improvements or additions to documentation
+  name: documentation
+- color: 0075ca # blue
+  description: ""
+  name: API
```
cognifloyd, line 147: I think we can use this color of blue for `area/code_location` and `area/python` labels. Here is how `#0075ca` looks in dark and light mode:
image

.github/labels.yml
```diff
@@ -0,0 +1,277 @@
...
+# ##### status labels ##### #
+- color: cfd3d7 # gray (standard github)
+  description: This issue or pull request already exists
+  name: duplicate # status:duplicate
+- color: e4e669 # yellow (standard github)
+  description: This doesn't seem right
+  name: invalid # status:invalid
+- color: ffffff # white (standard github)
+  description: This will not be worked on
+  name: wontfix # status:wontfix
```
cognifloyd, line 95: I think these three labels should have the `status:` prefix like the rest of the status labels.

.github/labels.yml
```diff
@@ -0,0 +1,277 @@
...
+# ##### type labels ##### #
+- color: b60205
+  description: ""
+  name: breaking change
+- color: d73a4a # red (standard github)
+  description: Something isn't working
+  name: bug
+- color: d73a4a
+  description: ""
+  name: bug fix
```
cognifloyd, line 49: Let's replace `bug fix` with just `bug`. If it's an issue labeled `bug`, then it describes a bug. If it's a PR labeled `bug` then it fixes a bug.
```suggestion
- color: d73a4a # red (standard github)
  description: Something isn't working
  name: bug
```

.github/labels.yml
```diff
@@ -0,0 +1,277 @@
...
+  description: ""
+  name: python3
+- color: 0075ca # blue
+  description: ""
+  name: unicode
+
+# ##### area/pack_resources labels ##### #
+- color: 003459 # blue
+  description: ""
+  name: action aliases
```
cognifloyd, line 199: I think we can use this color of blue for `area/pack_resources` labels. Here is how `#003459` looks in dark and light mode:
image

.github/labels.yml
```diff
@@ -0,0 +1,277 @@
...
+  description: ""
+  name: 'workflows: action chain'
+- color: 003459 # blue
+  description: ""
+  name: 'workflows: orquesta'
+
+# ##### area/external_services labels ##### #
+- color: 8336ff # blue
+  description: ""
+  name: database
```
cognifloyd, line 225: I think we can use this color of purple for `area/external_services` and `area/logging_and_performance` labels. Here is how `#8336ff` looks in dark and light mode:
image

.github/labels.yml
```diff
@@ -0,0 +1,277 @@
...
+  description: ""
+  name: metrics
+- color: 8336ff # purple
+  description: ""
+  name: performance
+
+# ##### area/install labels ##### #
+- color: 50209c # purple
+  description: ""
+  name: deployment
```
cognifloyd, line 250: I think we can use this color of purple for `area/install` labels. Here is how `#50209c` looks in dark and light mode:
image

.github/labels.yml
```diff
@@ -0,0 +1,277 @@
...
+  description: ""
+  name: st2-packages
+
+# ##### area/security labels ##### #
+- color: 63112b # red purple
+  description: ""
+  name: RBAC
+- color: 63112b # red purple
+  description: ""
+  name: security
```
cognifloyd, line 276: And I opted for a more red/purple `#63112b` for the `area/security` labels.
image

cognifloyd (commented):
@armab I'm slowly working through labels and colors to try and come up with some kind of systematic color scheme. All random colors means the colors have no meaning at all. I'm trying to give the colors common meaning.

I started by setting putting in the colors as they are now, and then resetting the standard github labels back to the github standard colors + descriptions. Then, I started grouping the labels and looking for a color to use for each group of labels.

What do you think of the grouping?
What about the colors?

PR #5428, opened 2021-11-08
Graceful shutdown of action runner.

khushboobhatia01:
This PR adds a wait on runner shutdown for all the ongoing executions to complete.

PR has dependency - https://github.com/StackStorm/st2/pull/5396

(Ref https://github.com/StackStorm/st2/discussions/5373)


st2actions/st2actions/worker.py
```diff
@@ -134,7 +138,21 @@ def process(self, liveaction):
...
+        coordinator = coordination.get_coordinator()
+        member_ids = []
+        service = "actionrunner"
+
+        try:
+            member_ids = list(coordinator.get_members(service.encode("utf-8")).get())
+        except GroupNotCreated:
+            pass
+
+        while member_ids and self._running_liveactions:
```
Kami, line 152: That sleep delay and code path in general is something which probably needs to be user configurable (and likely also have ability to fully disable that functionally via config option / feature flag).

And we may also want a hard timeout after which we give up - technically, depending on the service manager used (systemd, upstart) that may already be handled there (first send SIGTERM, if service still running after X seconds, send SIGKILL), but it's not a guarantee.

st2actions/st2actions/worker.py
```diff
@@ -134,7 +138,21 @@ def process(self, liveaction):
...
     def shutdown(self):
         super(ActionExecutionDispatcher, self).shutdown()
-        # Abandon running executions if incomplete
+
+        # Check if there are other runners in service registry
+        coordinator = coordination.get_coordinator()
+        member_ids = []
+        service = "actionrunner"
+
+        try:
```
Kami, line 147: Is there a potential for race condition where, do we need locking?

st2actions/st2actions/worker.py
Kami, line 152: Per my question / comment above - it appears we may need to utilize locking and "refresh" ``member_ids`` on each iteration - since that list could change during the time we sleep and wait.

Kami (changes requested):

st2actions/st2actions/worker.py
khushboobhatia01: Moved refresh inside the loop

khushboobhatia01: Addressed

st2actions/st2actions/worker.py
khushboobhatia01: Moved refresh inside the loop

khushboobhatia01:
@Kami Please review the updated changes.

conf/st2.conf.sample
```diff
@@ -18,6 +22,8 @@ pip_opts =  # comma separated list allowed here.
 python_binary = /usr/bin/python
 # Default log level to use for Python runner actions. Can be overriden on invocation basis using "log_level" runner parameter.
 python_runner_log_level = DEBUG
+# Time interval between subsequent queries to check running executions.
+sleep_delay = 2
```
cognifloyd, line 26: This config param could use a more descriptive name.

st2common/st2common/config.py
```diff
@@ -500,6 +500,28 @@ def register_opts(ignore_errors=False):
         dispatcher_pool_opts, group="actionrunner", ignore_errors=ignore_errors
     )
 
+    graceful_shutdown_opts = [
+        cfg.BoolOpt(
+            "graceful_shutdown",
+            default=False,
```
cognifloyd, line 506: I think we should enable this by default, shouldn't we? Maybe this should default to true when the coordination backend is configured.

st2common/st2common/config.py
```diff
@@ -500,6 +500,28 @@ def register_opts(ignore_errors=False):
...
+        ),
+        cfg.IntOpt(
+            "exit_timeout",
+            default=300,
+            help="How long to wait for process (in seconds) to exit after receiving shutdown signal.",
+        ),
+        cfg.IntOpt(
+            "sleep_delay",
+            default=2,
+            help="Time interval between subsequent queries to check running executions.",
```
cognifloyd, line 517: I think this option needs a more descriptive name. Maybe something with `interval` like `exit_still_active_check_interval` or `ready_to_exit_check_interval`?

cognifloyd (approved):
Thanks! I'm really looking forward to using this change in prod.

PR #5434, opened 2021-11-09
lower CircleCI parallelism

cognifloyd:
CircleCI is refusing to run our jobs now giving this error message:
```
Uh oh! When we created this build, your project was configured to use 4x parallelism but you only had 2 containers. If you haven't already, please add more containers or lower your parallelism.
Sorry! Your build didn't run because it needs more containers than your plan allows. Please decrease this project's build parallelism or add more containers to your plan.
```
Apparently they tweaked our plan so that we have 2 concurrent containers instead of 2 concurrent jobs.
Aargh.


cognifloyd:
Nevermind. This doesn't work. Looks like we have to switch to the CircleCI open source plan.

PR #5462, opened 2021-11-28
fix: timediff typo

blackstrip:
#### Problem:
When i use the webui to make some timediff critera comparison in rules, it always return the  ``TypeError: '>' not supported between instances of 'float' and 'str'``

#### Check:
I tried so many value in critera form, it's just not work.
Then i found in `operators.py`, the `_timediff` is returned `(utc_now - diff_target_utc).total_seconds()` as a float number, the `timedelta.total_seconds()` is always return float type, but the critera-comparison is str or int

#### Fix:
Just convert the critera value to float, it will be fine.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,14 @@ Changelog
 in development
 --------------
 
+Fixed
+~~~~~
+
+* Fix Type error for ``time_diff`` critera comparison. convert the timediff value as float to match
+  ``timedelta.total_seconds()`` return.
+
+  Contributed by @blackstrip
+
 Added
 ~~~~~
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,7 +8,7 @@ Fixed
 ~~~~~
 
 * Fix Type error for ``time_diff`` critera comparison. convert the timediff value as float to match
-  ``timedelta.total_seconds()`` return.
+  ``timedelta.total_seconds()`` return. #5462
 
   Contributed by @blackstrip
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,13 @@ Changelog
 in development
 --------------
 
+Fixed
+~~~~~
+
+* Fix Type error for ``time_diff`` critera comparison. convert the timediff value as float to match
+  ``timedelta.total_seconds()`` return.
+
+  Contributed by @blackstrip
 
 3.6.0 - October 29, 2021
 ------------------------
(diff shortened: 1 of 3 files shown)


armab:
@blackstrip Thanks for the contribution!

Could you please add the Changelog for this PR and, ideally some Unit test to cover the case.

Perhaps more details about how you caught this issue/error would help too for someone to review it.

blackstrip:
> @blackstrip Thanks for the contribution!
> 
> Could you please add the Changelog for this PR and, ideally some Unit test to cover the case.
> 
> Perhaps more details about how you caught this issue/error would help too for someone to review it.

Hi @armab , Thanks for the reply!
The PR is updated.

amanda11:
@blackstrip Thanks for contribution, the latest merge seems to have removed your CHANGELOG changes, could you update to master and try again with the changelog?

blackstrip:
@amanda11 Thanks for review, updated.

CHANGELOG.rst
```diff
@@ -4,6 +4,14 @@ Changelog
 in development
 --------------
 
+Fixed
+~~~~~
+
+* Fix Type error for ``time_diff`` critera comparison. convert the timediff value as float to match
+  ``timedelta.total_seconds()`` return.
```
amanda11, line 11: Minor comment - if you look at the other items in the change log, can you add #5462 to the end of the description of the change request. That way we can match changes to the PRs that introduced them, helps us find out changes made at a later point.

amanda11 (changes requested):
@blackstrip One small comment on changelog, and I wonder did you look to see if it was possible to write a UT for the fix as per @armab's suggestion?

CHANGELOG.rst
blackstrip: Ok, i'll make the update. and write UT too.

blackstrip (commented):
Updated. please check again. thx.

armab (approved):
Thanks for the fix! :+1:

armab:
@blackstrip Looks like the linting is failing https://github.com/StackStorm/st2/runs/4715712255?check_suite_focus=true#step:11:122
Can you please reformat the `unit/test_operators.py` with Black?

Additionally, new unit tests might need a bit more work per: https://github.com/StackStorm/st2/runs/4715712802

blackstrip:
> @blackstrip Looks like the linting is failing https://github.com/StackStorm/st2/runs/4715712255?check_suite_focus=true#step:11:122 Can you please reformat the `unit/test_operators.py` with Black?
> 
> Additionally, new unit tests might need a bit more work per CI failures: https://github.com/StackStorm/st2/runs/4715712802

Hi @armab , thx for review. the case is udpated, but maybe git-action failed? please can you re-check?

armab:
That worked.
Merged, thanks!

PR #5463, opened 2021-11-29
Workflow engine graceful shutdown

khushboobhatia01:
Please ref: https://github.com/StackStorm/st2/discussions/5373


khushboobhatia01:
Marking the PR as WIP until dependent MR is merged https://github.com/StackStorm/st2/pull/5396

st2actions/st2actions/workflows/workflows.py
```diff
@@ -85,6 +93,53 @@ def process(self, message):
...
+                    pass
+
+    def shutdown(self):
+        super(WorkflowExecutionHandler, self).shutdown()
+
+        if cfg.CONF.workflow_engine.graceful_shutdown:
+
+            coordinator = coordination.get_coordinator()
+            member_ids = []
+            service = "workflow_engine"
```
m4dcoder, line 112: Get the service name from the constant at https://github.com/StackStorm/st2/blob/master/st2actions/st2actions/cmd/workflow_engine.py#L40

st2actions/st2actions/workflows/workflows.py
```diff
@@ -44,6 +51,7 @@
 class WorkflowExecutionHandler(consumers.VariableMessageHandler):
     def __init__(self, connection, queues):
         super(WorkflowExecutionHandler, self).__init__(connection, queues)
+        self._running_workflows = list()
```
m4dcoder, line 54: This list is not sorted. I can see performance issue trying to search/delete from this list if there are a lot of workflows being processed. Can this be a dict or some other data structure w/ constant lookup time?  Also, I don't see any lock mechanism, this is not thread safe. Since the workflow engine at any given point will be processing multiple workflows and updating this list, it's going to cause issue w/o some sort of coordination.

st2actions/st2actions/workflows/workflows.py
```diff
@@ -85,6 +93,53 @@ def process(self, message):
...
+
+            running_workflows = set(self._running_workflows)
+
+            while running_workflows:
+                execution_id = running_workflows.pop()
+                try:
+                    ac_ex_db = ex_db_access.ActionExecution.get_by_id(str(execution_id))
+                    lv_ac_db = lv_db_access.LiveAction.get_by_id(
+                        ac_ex_db.liveaction["id"]
+                    )
```
m4dcoder, line 139: Can you create a common function get_liveaction_by_execution in https://github.com/StackStorm/st2/blob/master/st2common/st2common/services/action.py? I thought there may be some common function like this already but I couldn't find any. Please add appropriate unit test(s) as well.

st2actions/st2actions/workflows/workflows.py
```diff
@@ -85,6 +93,53 @@ def process(self, message):
...
+            else:
+                try:
+                    self._running_workflows.remove(str(message["parent"]))
+                except ValueError:
+                    pass
+
+    def shutdown(self):
+        super(WorkflowExecutionHandler, self).shutdown()
+
+        if cfg.CONF.workflow_engine.graceful_shutdown:
```
m4dcoder, line 108: This shouldn't be an option. The service should always attempt graceful shutdown.

st2actions/st2actions/workflows/workflows.py
```diff
@@ -85,6 +93,53 @@ def process(self, message):
...
+
+            coordinator = coordination.get_coordinator()
+            member_ids = []
+            service = "workflow_engine"
+            exit_timeout = cfg.CONF.workflow_engine.exit_timeout
+            sleep_delay = cfg.CONF.workflow_engine.sleep_delay
+            timeout = 0
+
+            while timeout < exit_timeout and self._running_workflows:
```
m4dcoder, line 117: Can you put some comments what this while loop achieves?  Some work is still needed so the mechanism is more robust. I also think support of some edge cases is missing (i.e. what happens if all engines shutdown at the same time. will some workflows in running state while other in pause state).

m4dcoder (changes requested):
Some work is still needed so the mechanism is more robust. When a workflow engine starts up, does it know to resume workflows that have been paused due to a shutdown?

khushboobhatia01:
> Some work is still needed so the mechanism is more robust. When a workflow engine starts up, does it know to resume workflows that have been paused due to a shutdown?

Added feature to resume.

st2actions/st2actions/workflows/workflows.py
```diff
@@ -85,6 +97,76 @@ def process(self, message):
...
+                except ValueError:
+                    pass
+
+    def shutdown(self):
+        super(WorkflowExecutionHandler, self).shutdown()
+        self._shutdown = True
+        if self._helper_thread:
+            concurrency.kill(self._helper_thread)
+
+        if cfg.CONF.workflow_engine.graceful_shutdown:
```
m4dcoder, line 117: This is not resolved yet. We don't need a `graceful_shutdown` config option. The workflow engine should alway graceful shutdown.

m4dcoder (changes requested):
@khushboobhatia01 The implementation here needs some more rework.

Here're some background on the workflow engine that may help you
- a workflow execution can be processed by any workflow engines
- the process function to handle workflow execution or handle action execution messages is short lived
- the workflow engine is multi-threaded and many `process` functions could be running at any given time
- if the workflow engine needs to pause workflows because there are no other workflow engines running, then the workflow engine should query the database for the list of running workflows

Given the above,
- For each workflow engine, we just need to track how many messages it is still processing. We don't need to track the specific workflow execution or task execution. We just need to increment a counter when invoking `process` and decrement the same counter when exiting `process`. On shutdown, if the counter is > 0, that means the workflow engine is processing message(s).
- Incrementing or decrementing the counter should be thread safe so need to thread lock before updating the counter.
- On shutdown, the workflow engine should initiate shutdown and stop receiving any more messages. Then it waits until the counter becomes zero.
- After the workflow engine stopped processing, if there are no other workflow engine running, then acquire a distributed lock (use coordinator), query the database for any active/running workflow execution, and pause the workflows. The start up sequence of workflow engine should use the same distributed lock when trying to resume paused workflows. This is to prevent workflow engines stepping over each other.

Please review and let me know if need more clarification.

khushboobhatia01:
> @khushboobhatia01 The implementation here needs some more rework.
> 
> Here're some background on the workflow engine that may help you
> 
> * a workflow execution can be processed by any workflow engines
> * the process function to handle workflow execution or handle action execution messages is short lived
> * the workflow engine is multi-threaded and many `process` functions could be running at any given time
> * if the workflow engine needs to pause workflows because there are no other workflow engines running, then the workflow engine should query the database for the list of running workflows
> 
> Given the above,
> 
> * For each workflow engine, we just need to track how many messages it is still processing. We don't need to track the specific workflow execution or task execution. We just need to increment a counter when invoking `process` and decrement the same counter when exiting `process`. On shutdown, if the counter is > 0, that means the workflow engine is processing message(s).
> * Incrementing or decrementing the counter should be thread safe so need to thread lock before updating the counter.
> * On shutdown, the workflow engine should initiate shutdown and stop receiving any more messages. Then it waits until the counter becomes zero.
> * After the workflow engine stopped processing, if there are no other workflow engine running, then acquire a distributed lock (use coordinator), query the database for any active/running workflow execution, and pause the workflows. The start up sequence of workflow engine should use the same distributed lock when trying to resume paused workflows. This is to prevent workflow engines stepping over each other.
> 
> Please review and let me know if need more clarification.

@m4dcoder Updated the MR with new implementation. One issue with the above flow is during last engine shutdown, workflows which have active tasks could transition only to pausing state instead of paused state. To address this, I've added a delay (https://github.com/StackStorm/st2/pull/5463/files#diff-e4e1e16f3a5aefdf069b0af9bd510c1fde29a1b8df10796b7e74cb183820973cR79) in the engine startup resume sequence to ensure workflows stuck in pausing state transition to paused state before we start resuming.

Please review.

st2actions/st2actions/workflows/workflows.py
```diff
@@ -40,10 +47,15 @@
     queues.WORKFLOW_ACTION_EXECUTION_UPDATE_QUEUE,
 ]
 
+WORKFLOW_ENGINE = "workflow_engine"
+SHUTDOWN_ROUTINE = "shutdown_routine"
```
m4dcoder, line 51: Since this name is used for locking during startup and shutdown, please rename this to `WORKFLOW_ENGINE_START_STOP_SEQ = "workflow_engine_start_stop_seq"`

st2actions/st2actions/workflows/workflows.py
```diff
@@ -78,13 +94,61 @@ def process(self, message):
...
             # the garbage collector will find and cancel these workflow executions.
             self.fail_workflow_execution(message, e)
+        finally:
+            with self._semaphore:
+                self._active_messages -= 1
+
+    def shutdown(self):
+        super(WorkflowExecutionHandler, self).shutdown()
+        while self._active_messages > 0:
+            concurrency.sleep(2)
```
m4dcoder, line 113: Should probably handle a timeout here to prevent an infinite while.

st2actions/st2actions/workflows/workflows.py
```diff
@@ -78,13 +94,61 @@ def process(self, message):
...
+
+        coordinator = coordination.get_coordinator()
+        member_ids = []
+        with coordinator.get_lock(SHUTDOWN_ROUTINE):
+            try:
+                member_ids = list(
+                    coordinator.get_members(WORKFLOW_ENGINE.encode("utf-8")).get()
```
m4dcoder, line 120: Instead of doing `WORKFLOW_ENGINE.encode("utf-8")`, can we add a method in the coordination service that get_group_id with the logic @ https://github.com/StackStorm/st2/blob/master/st2common/st2common/service_setup.py#L324-L327. That way, you can call `coordinator.get_group_id(WORKFLOW_ENGINE)`. This is a safer way to get group id so to shield from any changes to group_id naming/formatting.

st2actions/st2actions/workflows/workflows.py
```diff
@@ -40,10 +47,15 @@
     queues.WORKFLOW_ACTION_EXECUTION_UPDATE_QUEUE,
 ]
 
+WORKFLOW_ENGINE = "workflow_engine"
```
m4dcoder, line 50: There is already a definition at https://github.com/StackStorm/st2/blob/master/st2actions/st2actions/cmd/workflow_engine.py#L40. Can we keep this here and modify workflow_engine.py to reuse this constant?

st2common/st2common/transport/consumers.py
```diff
@@ -43,6 +43,7 @@ def __init__(self, connection, queues, handler):
         self._handler = handler
 
     def shutdown(self):
+        self.should_stop = True
```
m4dcoder, line 46: What is this for?

m4dcoder (changes requested):
This is going in the right direction. Thanks!

st2actions/st2actions/workflows/workflows.py
```diff
@@ -62,6 +74,10 @@ def handle_action_execution_with_instrumentation(ac_ex_db):
             ex_db_models.ActionExecutionDB: handle_action_execution_with_instrumentation,
         }
 
+        # This is required to ensure workflows stuck in pausing state after shutdown transition to paused state after engine startup.
+        self._delay = 30
```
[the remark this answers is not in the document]
khushboobhatia01: This is to handle cases where the workflows are stuck in PAUSING state. 
Consider a scenario where the last workflow engine is shutting down and it request pause on the running workflow. If the workflow has any running task, it'll transition PAUSING state. The workflow can transition to PAUSED state only when the running task completes and this completion should be handled by the workflow engine. Now that we stop accepting requests on WFE shutdown, there will be no WFE to handle this completion and the workflow will stay in PAUSING state. The message in queue will not be picked up unless a new WFE comes up and processes message from the queue.
This delay will ensure when a new WFE comes up it takes care of processing these messages in the queue and transition workflows stuck in PAUSING state to PAUSED state. Hope that makes sense.

st2common/st2common/transport/consumers.py
khushboobhatia01: This will shutdown the WFE queue consumer, so that the WFE doesn't pick new requests from the queue. This was added for actionrunner sometime back. https://github.com/celery/kombu/blob/master/kombu/mixins.py#L131

st2actions/st2actions/workflows/workflows.py
khushboobhatia01: [coordination.get_members()](https://github.com/openstack/tooz/blob/master/tooz/coordination.py#L551) returns an async`CoordinatorResult` which has a [get()](https://github.com/openstack/tooz/blob/master/tooz/coordination.py#L642) method that retrieves the result of the corresponding asynchronous call.

st2actions/st2actions/workflows/workflows.py
khushboobhatia01: Addressed https://github.com/StackStorm/st2/pull/5463/files#diff-8570f80aaad5c1850cb5d35d826b860901e4d95b97b3db631c486fb6ad6901e1R40

st2actions/tests/unit/test_workflow_engine.py
```diff
@@ -271,3 +271,60 @@ def test_process_error_handling_has_error(self, mock_get_lock):
         # Assert workflow execution is cleaned up and canceled.
         lv_ac_db = lv_db_access.LiveAction.get_by_id(str(lv_ac_db.id))
         self.assertEqual(lv_ac_db.status, action_constants.LIVEACTION_STATUS_CANCELED)
+
```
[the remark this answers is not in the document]
khushboobhatia01: Addressed.

st2actions/st2actions/workflows/workflows.py
khushboobhatia01: Addressed

st2actions/st2actions/workflows/workflows.py
khushboobhatia01: Addressed

st2actions/st2actions/workflows/workflows.py
```diff
@@ -78,13 +92,69 @@ def process(self, message):
...
+            with self._semaphore:
+                self._active_messages -= 1
+
+    def start(self, wait):
+        spawn_after(self._delay, self._resume_workflows_paused_during_shutdown)
+        super(WorkflowExecutionHandler, self).start(wait=wait)
+
+    def shutdown(self):
+        super(WorkflowExecutionHandler, self).shutdown()
+        exit_timeout = cfg.CONF.workflow_engine.exit_still_active_check
```
m4dcoder, line 114: Can we rename the conf option to `cfg.CONF.workflow_engine.exit_timeout` to be consistent?

st2actions/st2actions/workflows/workflows.py
```diff
@@ -78,13 +92,69 @@ def process(self, message):
...
+        sleep_delay = cfg.CONF.workflow_engine.still_active_check_interval
```
m4dcoder, line 115: Can we rename `still_active_check_interval` to `exit_check_interval`?  And rename `sleep_delay` to the same for consistency?

st2actions/tests/unit/test_workflow_engine.py
```diff
@@ -271,3 +271,201 @@ def test_process_error_handling_has_error(self, mock_get_lock):
...
+        wf_ex_db = wf_db_access.WorkflowExecution.query(
+            action_execution=str(ac_ex_db.id)
+        )[0]
+        self.assertEqual(wf_ex_db.status, action_constants.LIVEACTION_STATUS_RUNNING)
+        workflow_engine = workflows.get_engine()
+
+        workflow_engine._delay = 0
+        # Start and stop WFE simultaneously.
+        eventlet.spawn_after(1, workflow_engine.shutdown)
+        eventlet.spawn_after(1, workflow_engine.start, True)
```
m4dcoder, line 436: Can we split this into two test cases? Can we control by spawn_after?
1. Shutdown sequence acquire lock first (run workflow_engine.shutdown with no delay and the spawn workflow engine start after 1 seconds)
2. Startup sequence acquire lock first (vice versa)

m4dcoder (changes requested):
Additional follow up. We are almost there. Great job!

st2actions/st2actions/workflows/workflows.py
```diff
@@ -40,10 +47,17 @@
...
+WORKFLOW_ENGINE_START_STOP_SEQ = "workflow_engine_start_stop_seq"
+
 
 class WorkflowExecutionHandler(consumers.VariableMessageHandler):
     def __init__(self, connection, queues):
         super(WorkflowExecutionHandler, self).__init__(connection, queues)
+        self._active_messages = 0
+        self._semaphore = Semaphore()
+        # This is required to ensure workflows stuck in pausing state after shutdown transition to paused state after engine startup.
+        self._delay = 30
```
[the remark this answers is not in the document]
khushboobhatia01: Answers inline.
1. Do you know what happens if no new WFE comes up? The action execution completion message will stay in rabbitMQ?
-> Right. Since resume flow happens only once during WFE startup, until the message is consumed by a new WFE it'll stay in the queue.

2. What if this is a complete shutdown and the rabbitMQ is rebooted as well?
-> Queues declared in StackStorm are durable and the messages are sent with delivery mode "persistent" (default setting). This means that unless someone purges the data manually, all the messages persist across rabbitMQ shutdown.

3. If the messages in rabbitMQ is lost, how should we handle workflows in PAUSING?
-> The only way a message will get lost here is when someone purges data. Having said that, I think we still need some way to handle workflows in PAUSING. Do you think instead of resuming the workflows paused by shutdown routine only during WFE startup, we could check and resume such workflows at defined intervals?

st2actions/st2actions/workflows/workflows.py
khushboobhatia01: We have same config option defined for action runner graceful shutdown as well that's why I reused the same name here.

st2actions/st2actions/workflows/workflows.py
khushboobhatia01: We have same config option defined for action runner graceful shutdown as well that's why I reused the same name here.

PR #5467, opened 2021-11-30
Fix multiple file support in linux.file_watch.line + black + fstring

nzlosh:
When multiple file_watch rules are defined, the last defined file reference is used for all files being watched.  This causes trigger-instances to fail rule enforcement.

Adding the reference to the logging shows `test1.log` has the reference ending with `8c505`
```
2021-11-30 18:50:40,434 140243179888112 INFO file_watch_sensor [-] Added file "/var/log/test1.log" with reference linux.7e55ad75-b10c-44db-b53e-95164a18c505
2021-11-30 18:50:41,459 140243179888112 INFO file_watch_sensor [-] Added file "/var/log/test2.log" with reference linux.590de8c1-c578-4125-9082-2cee03b030a9
```

When the file contents are updated a trigger is emitted by the sensor using the reference of `test2.log` ending in `b030a9`
```
root@u1804:~# st2 trigger-instance get 61a6649f164625c2d94dccb8 -y
id: 61a6649f164625c2d94dccb8
occurrence_time: '2021-11-30T17:51:27.294000Z'
payload:
    file_name: test1.log
    file_path: /var/log/test1.log
    line: Tue Nov 30 18:51:27 CET 2021 dhcp
status: processed
trigger: linux.590de8c1-c578-4125-9082-2cee03b030a9
```

This PR consists of adding a dictionary that is used to track the `path_name` and `reference` pair and looks up the reference for the file that was altered when creating the trigger.

The code is formatted with black and updated to use fstrings since all instances will be using Python 3.6+


cognifloyd (approved):

cognifloyd:
I'm approving/merging this to improve the status quo action.

But I really hope that the file watch stuff will be replaced by merging #5096 (which depends on #5257 which is waiting for gunicorn to do a release. gunicorn is waiting to release because their CI is broken. :shrug: )

PR #5468, opened 2021-11-30
Enable setting ttl for MockDatastoreService

### LINKED ISSUE #4880: An exception is raised when calling MockDatastoreService.set_value() with a ttl
## SUMMARY

When using `MockDatastoreService.set_value()` for unit testing (e.g. via `BaseActionTestCase`), a `ValueError` exception will be raised if `ttl` argument is specified.

### STACKSTORM VERSION

`st2 3.1.0, on Python 3.6.8`

Custom install, but not relevant to this case, see below.

See [code](https://github.com/StackStorm/st2/blob/6d1809a4bb577e117baa00f249757284db9c6e76/st2tests/st2tests/mocks/datastore.py#L108) for `MockDatastoreService.set_value()`. If `ttl` argument is specified, `ValueError` will explicitly be raised.

```python
class MockDatastoreService(BaseDatastoreService):
    # ...
    def set_value(self, name, value, ttl=None, local=True, scope=SYSTEM_SCOPE, encrypt=False):
        """
        Store a value in a dictionary which is local to this class.
        """
        if ttl:
            raise ValueError('MockDatastoreService.set_value doesn\'t support "ttl" argument')
        # ...
```

## Expected Results

Both `action_service.set_value()` and  `sensor_service.set_value()` support `ttl` argument. Although I understand that this argument cannot be honored in a mock situation, `MockDatastoreService`  should be consistent with the behavior of the class it is mocking - perhaps fire a warning instead of an Exception?

## Actual Results

An exception is returned while running tests:
```python
ValueError: MockDatastoreService.set_value doesn't support "ttl" argument
```
It is not possible to test actions or sensors that use `xxx.set_value()` with a `ttl` argument.

ytjohn:
This allows people to test their code if they are already setting a ttl. This resolves #4880

I think we would also be fine with not issuing a log warning, but put that in for completeness.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,7 +8,7 @@ Added
 ~~~~~
 
 * Enable setting ttl for MockDatastoreService. #5468
- 
+
   Contributed by @ytjohn
 
 * Added st2 API and CLI command for actions clone operation.
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,10 @@ in development
 Added
 ~~~~~
 
+* Enable setting ttl for MockDatastoreServic
+ 
+  Contributed by @ytjohn
+
 * Added service degerestration on shutdown of a service. #5396
 
   Contributed by @khushboobhatia01
--- a/st2tests/st2tests/mocks/datastore.py
+++ b/st2tests/st2tests/mocks/datastore.py
@@ -35,7 +35,7 @@ class MockDatastoreService(BaseDatastoreService):
         self._class_name = class_name
         self._username = api_username or "admin"
         self._logger = logger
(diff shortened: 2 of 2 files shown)


amanda11 (approved):
This looks good to me, and thanks for the contribution.
Just awaiting the pipelines to run.

ytjohn:
@amanda11 Thanks for the approval. I fixed the formatting and updated changelog.  I'm not yet seeing how this change would have impacted the build, so not sure if that's just an ephemeral issue or something that was expecting ttl to be rejected.. I'm now working on setting up a full local development environment for the st2 codebase itself.

armab:
Merged, thanks @ytjohn!

Issue #5483, opened 2021-12-07
Executions can not be proceeded with errors in st2actionrunner and st2scheduler


yypptest:
## SUMMARY

We picked up stackstorm v3.6.0 and deployed on kubernetes, the pods can started normally but execution always stuck at requested status. The st2scheduler pod showed below log messages, and st2actionrunner also showed similar error exceptions. 

```
2021-12-06 06:42:44,326 DEBUG [-] Using cached coordinator instance: <tooz.drivers.etcd.EtcdDriver object at 0x7f68890b6b70>
2021-12-06 06:42:44,327 ERROR [-] Traceback (most recent call last):

2021-12-06 06:42:44,327 ERROR [-]
2021-12-06 06:42:44,328 ERROR [-]   File "/opt/stackstorm/st2/lib/python3.6/site-packages/eventlet/hubs/poll.py", line 111, in wait
    listener.cb(fileno)

2021-12-06 06:42:44,328 ERROR [-]
2021-12-06 06:42:44,328 ERROR [-]   File "/opt/stackstorm/st2/lib/python3.6/site-packages/eventlet/greenthread.py", line 221, in main
    result = function(*args, **kwargs)

2021-12-06 06:42:44,328 ERROR [-]
2021-12-06 06:42:44,328 ERROR [-]   File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2common/metrics/base.py", line 216, in wrapper
    return func(*args, **kw)

2021-12-06 06:42:44,328 ERROR [-]
2021-12-06 06:42:44,328 ERROR [-]   File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2actions/scheduler/handler.py", line 314, in _handle_execution
    self._schedule(liveaction_db, execution_queue_item_db)

2021-12-06 06:42:44,328 ERROR [-]
2021-12-06 06:42:44,328 ERROR [-]   File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2actions/scheduler/handler.py", line 422, in _schedule
    self._update_to_scheduled(liveaction_db, execution_queue_item_db)

2021-12-06 06:42:44,328 ERROR [-]
2021-12-06 06:42:44,328 ERROR [-]   File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2actions/scheduler/handler.py", line 481, in _update_to_scheduled
    publish=False,

2021-12-06 06:42:44,328 ERROR [-]
[...]

PR #5484, opened 2021-12-07
Fix etcd lock error

khushboobhatia01:
Fixes https://github.com/StackStorm/st2/issues/5483

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -28,10 +28,6 @@ Fixed
 Added
 ~~~~~
 
-* Added service degerestration on shutdown of a service. #5396
-
-  Contributed by @khushboobhatia01
-
 * Added possibility to add new values to the KV store via CLI without leaking them to the shell history. #5164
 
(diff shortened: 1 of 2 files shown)


armab (commented):
Thanks, @khushboobhatia01 for the fix! :+1: 

Could you include a Changelog record for this regression bugfix?

CHANGELOG.rst
```diff
@@ -11,6 +11,10 @@ in development
 Added
 ~~~~~
 
+* Use string lock name instead of object id. #5484
+
+  Contributed by @khushboobhatia01
+
```
armab, line 17: The changelog would need to go into `development` section as `v3.6.0` was already cut.
The record best fit for the `Bugfix` section. Please also mention that's it's a fix for a regression, referencing the #5358.
This way we'll keep the issues connected.

armab (approved):
Thanks for the hotfix! :+1:

cognifloyd (approved):
Thanks for fixing the Changelog!
The fix makes sense.

Issue #5497, opened 2021-12-09
no_proxy is not respected during index searches


minsis:
## SUMMARY

This isn't exactly a bug (kinda?) but maybe a solution to consider in how implement this.

Basically no_proxy is ignored during pack index searches.
https://github.com/StackStorm/st2/blob/15dcd3b3642e3a6af0dc82f2b746d98fa46467fa/st2common/st2common/services/packs.py#L74-L84

This makes obvious sense because if you're behind a proxy you want to be able to search the st2 Exchange Index. The issue comes in when you have a local index like I do. The local index is hosted on our controller node and the proxy doesn't have access to this URL due to FW rules. Proxy isn't local to the datacenter either. So because no_proxy isn't respected in an index search my custom index is unreachable.

To further complicate things, I also search the Exchange Index, meaning I have specified both internet based and a local based index. The internet based obviously needs to go though the proxy, but the local based one can't.

The only real way to by pass this at this point is to specify our internal git URL directly and the pack will install since its not searching the index.

Not sure if anyone else has this issue, it might be extremely specific to me with our locked down environment.

### STACKSTORM VERSION

Paste the output of ``st2 --version``: 
st2 3.6.0, on Python 3.6.8

##### OS, environment, install method

3 node HA cluster: 2 nodes, 1 controller all three behind a load balancer

minsis:
#5500

PR #5500, opened 2021-12-13
Add no_proxy to index searches

minsis:
Adding no_proxy option for index searches.

### DIFF
--- a/st2common/st2common/services/packs.py
+++ b/st2common/st2common/services/packs.py
@@ -83,6 +83,7 @@ def _fetch_and_compile_index(index_urls, logger=None, proxy_config=None):
     if proxy_config:
         https_proxy = proxy_config.get("https_proxy", None)
         http_proxy = proxy_config.get("http_proxy", None)
+        no_proxy = proxy_config.get("no_proxy", None)
         ca_bundle_path = proxy_config.get("proxy_ca_bundle_path", None)
 
         if https_proxy:
@@ -91,6 +92,9 @@ def _fetch_and_compile_index(index_urls, logger=None, proxy_config=None):
 
         if http_proxy:
             proxies_dict["http"] = http_proxy
+
+        if no_proxy:
+            proxies_dict["no"] = no_proxy
 
     for index_url in index_urls:
         index_status = {


minsis:
Resolves #5497

armab (commented):
Thanks!

Can you also add the Changelog record for this enhancement?

minsis:
@armab Updated

st2common/st2common/services/packs.py
```diff
@@ -92,6 +93,9 @@ def _fetch_and_compile_index(index_urls, logger=None, proxy_config=None):
         if http_proxy:
             proxies_dict["http"] = http_proxy
 
+        if no_proxy:
+            proxies_dict["no"] = no_proxy
```
armab, line 97: @minsis Did you test it BTW and confirm no_proxy setting works this way via `no` key?

minsis: I swear this was working, but I guess not.

I'm pretty sure the `no` key is correct. I looked in the requests library and when building the proxy info it does a split. Hence the reason the other kyes are `http` and `https`.

Looking at the requests docs they never mention no_proxy, but its mentioned in a release note saying its supposed to be `no_proxy`.

Regardless `no` and `no_proxy` doesn't even seem to work at all. I tested this directly using the requests library.

minsis: https://github.com/psf/requests/issues/4871

Looks like maybe this has been broken for quite some time. There's a PR to fix it, but its still open for 15 months.

armab: > I swear this was working, but I guess not.

Yeah, I'm just asking.

I'm good to merge it if you say if the code change was tested manually and works in your environment.
If you can confirm this is :+1: and fixes your case, - just let me know in the PR as the requests doc is a bit confusing indeed.

minsis: Add `no` or `no_proxy` doesn't work unfortunately. So I guess back to square one.

My thought now is to maybe use the same logic the PR is using. Basically if it detects the URL being used is in `no_proxy` it will remove the proxy info from the request.

minsis:
@armab I've done another commit to do similar to what the bug fix is supposed to provide from the requests library. 

Could you also do me a favor and just double check my upstream merge. I get confused on dealing with merge conflicts and such.

st2common/st2common/services/packs.py
```diff
@@ -106,7 +114,11 @@ def _fetch_and_compile_index(index_urls, logger=None, proxy_config=None):
         index_json = None
 
         try:
-            request = requests.get(index_url, proxies=proxies_dict, verify=verify)
+            request = requests.get(
+                index_url,
+                proxies=proxies_dict if not bypass_proxy else None,
+                verify=verify if not bypass_proxy else None
```
armab: I understand the need of overriding the proxies, but in this case, why it's needed to disable the `verify` flag?

Wouldn't it downgrade the security for the users?
 https://github.com/StackStorm/st2/blob/8ad5e5dcc4f7fa14c6ab5f1aae5faf240f389d07/st2common/st2common/services/packs.py#L88-L92
it's better to allow them to pass `proxy_ca_bundle_path` instead of disabling the cert verification.

armab (changes requested):

st2common/st2common/services/packs.py
minsis: I'll have to look at this again. Verify accepts a bool or a string to the path of the ca bundle to use. I was thinking more in terms of bypassing the CA cert bundle string than the bool. I'll look at it later today and fix it. Thanks for pointing it out.

st2common/st2common/services/packs.py
```diff
@@ -102,7 +114,11 @@ def _fetch_and_compile_index(index_urls, logger=None, proxy_config=None):
         index_json = None
 
         try:
-            request = requests.get(index_url, proxies=proxies_dict, verify=verify)
+            request = requests.get(
+                index_url,
+                proxies=proxies_dict if not bypass_proxy else None,
+                verify=verify if not bypass_proxy else True,
```
armab, line 120:
```suggestion
                verify=verify,
```

I still don't understand why adjusting the `verify` logic is needed in this situation.
I thought we were trying to fix `no_proxy`.

Could you perhaps provide more context or an example?

minsis: https://github.com/StackStorm/st2/blob/224e0e29310a8ebdfba522fb28e59585b8cfc940/st2common/st2common/services/packs.py#L88-L92

Because verify is overwritten with the path to your proxy's ca bundle if `proxy_ca_bundle_path` environment variable is set. Since its outside the loop there could be a chance where you are specifying a ca bundle path but needing to bypass the proxy.

An example might be (without `ca_bundle_path or True`):
```python
proxy_config = {
    "https_proxy": "http://proxy.company.local",
    "http_proxy":  "http://proxy.company.local",
    "no_proxy": "company.local,localhost,my.other.domain",
    "proxy_ca_bundle_path": "/path/to/proxy/ca_bundle"
}

proxies_dict = {
    "https": "http://proxy.company.local",
    "http":  "http://proxy.company.local",
    "no": "company.local,localhost,my.other.domain",
}

# verify is now overwritten from True to '/path/to/proxy/ca_bundle'
verify="/path/to/proxy/ca_bundle"
```

Now we enter loop:
```python
# Iteration 1
verify="/path/to/proxy/ca_bundle" # permanently set

index_url="https://index.stackstorm.com"
bypass_proxy = False

requests.get(
    "https://index.stackstorm.com",
    proxies={
        "https": "http://proxy.company.local",
        "http":  "http://proxy.company.local",
        "no": "company.local,localhost,my.other.domain",
    },
    verify="/path/to/proxy/ca_bundle",
)
```

```python
# Iteration 2
verify="/path/to/proxy/ca_bundle" # permanently set

index_url="https://index.mycompany.local"
bypass_proxy = True

requests.get(
    "https://index.mycompany.local",
    proxies=None
    verify="/path/to/proxy/ca_bundle",
)
```

So in Iteration 2 we would might end up failing because we're still trying to use our proxy's ca bundle, but not using the proxy.

armab: I see now, thanks for more context :+1:

armab (approved):
Thanks for the contribution :+1: 
The change is marked for the upcoming `v3.7.0` release.

Before that, if you can add a unit test for the new logic in `packs.py`, - that would be helpful as proxy behavior is highly inconsistent between the different tools.

minsis:
I thought about writing a test earlier, but I'm not sure how I would write a test for this. I'm still learning on writing good unit and/or integration testing. This one is particular difficult in my mind because of the looping mechanism. Any suggestions?

cognifloyd (approved):
Tests for this would be awesome. However, there are no unit tests for things that retrieve the pack index. Writing tests for all of these index functions would be much more involved than merely a test for the proxy config.

I'm going to merge this now. I'm satisfied with the manual testing for this fix.

PR #5524, opened 2021-12-29
install wheel to fix check-python-packages-nightly CI

cognifloyd:
A follow-up for #5417 
Nightly checks are still failing, so we need to explicitly install wheel in the venv.

I'll test a bit first by modifying the GHA workflow used in PRs, and then revert the debug bits to leave only the tested fix.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -253,6 +253,7 @@ check-python-packages:
 	@echo "================== CHECK PYTHON PACKAGES ===================="
 	@echo ""
 	test -f $(VIRTUALENV_COMPONENTS_DIR)/bin/activate || $(PYTHON_VERSION) -m venv $(VIRTUALENV_COMPONENTS_DIR) --system-site-packages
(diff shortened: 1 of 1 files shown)


cognifloyd:
OK. I modified the `check-python-packages` target to do the `bdist_wheel` setup bit similar to `check-python-packages-nightly`. You can see the failed CI run for 632a6f504e5d8143ed24a9f01390be852c8554ba here (I cancelled the rest of the tests once I had the expected failure output):
https://github.com/StackStorm/st2/runs/4655764424?check_suite_focus=true#step:11:1227

You on the CI run for 783043eb031b95af79e20e44eb21a9a7720c2aa6 you can see the `check-python-packages` task succeeding.
https://github.com/StackStorm/st2/runs/4655850345?check_suite_focus=true#step:11:1227

Finally, I removed the debug changes to `check-python-packages` so this PR only affects the `check-python-packages-nightly` task.

armab (approved):

Issue #5387, opened 2021-10-12
Coordinator errors when installing packs and using Consul backend


mjtice:
## SUMMARY

Using Consul as a coordinator backend, installing packs fails with errors.

### STACKSTORM VERSION

```
# st2 --version
st2 3.5.0, on Python 3.6.8
```

##### OS, environment, install method

RockyLinux 8.4 installed per https://docs.stackstorm.com/install/rhel8.html
Consul v1.5.2 (I've also tried the latest 1.10.3 with the same errors).
python-consul2-0.1.5 (python-consul is no longer maintained).

## Steps to reproduce the problem

* Start a Consul dev server
```
consul agent -dev
```
* Modify `st2.conf` to specify Consul as the coordinator
```
[coordination]
url = consul://127.0.0.1:8500
```
* Start ST2
```
st2ctl start
```
* Install a pack
```
st2 pack install st2
```

## Expected Results

The pack should install

## Actual Results

The install fails with errors:
```
# st2 pack install st2

For the "st2" pack, the following content will be registered:

actions   |  24
rules     |  0
sensors   |  0
aliases   |  12
triggers  |  0

Installation may take a while for packs with many items.

    [ succeeded ] init_task

id: 6165b63260a18b88e15ec783
action.ref: packs.install
parameters:
  packs:
  - st2
status: failed
start_timestamp: Tue, 12 Oct 2021 16:22:10 UTC
end_timestamp: Tue, 12 Oct 2021 16:22:11 UTC
log:
  - status: requested
    timestamp: '2021-10-12T16:22:10.930000Z'
  - status: scheduled
    timestamp: '2021-10-12T16:22:11.051000Z'
  - status: running
    timestamp: '2021-10-12T16:22:11.105000Z'
  - status: failed
    timestamp: '2021-10-12T16:22:11.905000Z'
result:
  errors:
  - message: Execution failed. See result for details.
    task_id: init_task
    type: error
  - message: 'TypeError: <lambda>() got an unexpected keyword argument ''retry_state'''
    route: 0
    task_id: init_task
    type: error
  - message: 'YaqlEvaluationException: Unable to resolve key ''packs_list'' in expression ''<% ctx().packs_list %>'' from context.'
    type: error
  - message: 'YaqlEvaluationException: Unable to resolve key ''message'' in expression ''<% ctx().message %>'' from context.'
    type: error
[...]

PR #5528, opened 2022-01-06
Add required dependency to support etcd3gw driver.

khushboobhatia01:
- Added etcd3gw dependency required by tooz.
- Downgraded tenacity to 6.3.1 as tooz dependency on tenacity has always been < 7.0.0 and etcd3gw fails with higher versions. https://github.com/openstack/tooz/blob/174065f8750ff374e140fae956758b0b5ec2c473/requirements.txt#L9


arm4b (commented):
Thanks for the PR!

Per https://docs.stackstorm.com/coordination.html, we explicitly state that some backends might need a pip dependency which st2 doesn't provide and only `redis` is currently shipped by default.

Are you using `etcd` for coordination in your prod?

khushboobhatia01:
Yes. We're using etcd right now, but we would want to enable service registry which requires etcd3gw.

khushboobhatia01:
@armab etcd3 driver doesn't work with StackStorm because of https://review.opendev.org/c/openstack/tooz/+/466098/. I've confirmed this locally ST2 services keep crashing. However, I've verified etcd3gw driver and everything seems to work fine.

arms11:
> In the past, we thought to make it the default driver, but eventually, switched to Redis as recommended default which provided the best experience. I've found #4608 and https://github.com/openstack/tooz/compare/master...StackStorm:etcd3_driver_fixes?expand=1 from the history and we experienced etcd3gw driver was broken in many ways, including limited functionality around the group membership which is required for [service registry](https://github.com/StackStorm/st2/pull/4548) in st2.
> 
> @Kami if you remember more context about the etcd3gw.
> 
> @arms11 Are you folks using etcd as well as a st2 coordination backend or it's Redis?

arms11:
@khushboobhatia01 - I am sorry for the dumb click and closing this by mistake. 
@armab To your question, we have been using `Redis` since I believe helm chart version `0.50`. The last time I remember we found a bug in `etcd` driver related to `keys` api in StackStorm and older implementation was anyways not being maintained at which time we decided to move to `Redis`. 

Just to understand, now since `Orquesta` workflow engine requires backend coordination and `Redis` is by default the chosen driver, does it make sense to go with `etcd`? Just curious.

khushboobhatia01:
> @khushboobhatia01 - I am sorry for the dumb click and closing this by mistake. 
> 
> @armab To your question, we have been using `Redis` since I believe helm chart version `0.50`. The last time I remember we found a bug in `etcd` driver related to `keys` api in StackStorm and older implementation was anyways not being maintained at which time we decided to move to `Redis`. 
> 
> 
> 
> Just to understand, now since `Orquesta` workflow engine requires backend coordination and `Redis` is by default the chosen driver, does it make sense to go with `etcd`? Just curious.

@arms11 @armab We have been using etcd as coordination driver since we started using StackStorm about 3 years back. Since then we have stabilized our infrastructure to support a large number of executions. Currently we saw about 3M+ executions. 

Our fleet and user base is going to increase more and I don't think we'll be moving to redis anytime soon.

armab:
We're bundling `Redis` with the st2 core at this moment as a default coordination backend, which is shipped with all the official installation methods and e2e tested. We then recommend installing manually a pip dependency for other backends, per user needs.
https://docs.stackstorm.com/coordination.html

The problem of bundling multiple dependencies for backends is that there are many of them: zookeeper, consul, Memcached, etc, etc. Once we start including more than a single default one, it would be hard to stop from accepting other dependency backends, if proposed. It'll also increase the number of dependencies/package size/etc.

Let's collaborate on how to proceed here.

@StackStorm/tsc are we good with adding an additional coordination driver dependency in the st2 core?
That'll add additional `etcd` (apart from default Redis) as a pip dependency bundled in stackstorm. Though, Redis will remain officially e2e tested.

cognifloyd (dismissed):
With how we're creating the virtualenvs right now, I think this is the best approach. So 👍 from me

nzlosh:
@armab bundling multiple tooz backend dependencies is a no-win situation.  This PR demonstrates the situation by the fact that redis end to end testing works with tenacity 7.x, yet it breaks with the etcd3gw module.  Since we only test the tooz redis backend, we can only guarantee production readiness specifically for this backend and no others.

Attempting to find common ground between 11 tooz backends and their python module dependencies is a road leading to dependency hell.  This is especially true when factoring the module compatibility with a specific software versions multiplied by the number of different versions across all the St2 user base.

I'm normally pro-choice, in this particular case, I don't see how we can reasonably support bundling more than 1 backend.  I also suggest we add a very strong warning in the coordinator documentation stating that only `redis` is officially supported for production use and the onus of testing/patching other backends rests completely with the user.

amanda11:
I'm also concerned about having dependencies in that we don't test. If we find in future that tenacity > x is needed for redis, then we are going to have problems, as we might get into incompatabilities between version of packages needed for etcd3gw and redis.

guzzijones:
I am a no on this as well

armab:
I hear that including other coordination backend dependencies like `etcd3gw` in st2 requirements.txt is a no-go at this moment.
Let's leave this open. Maybe we'll see more community requests in the future.

---
Speaking of `tenacity` which is already part of the st2 and is pulled by some other dependency ([not Redis](https://github.com/redis/redis-py/blob/master/requirements.txt)).

We have another bug report https://github.com/StackStorm/st2/issues/5387, where Consul backend also fails with the new `v7` tenacity version, similar to `etcd3gw`.

So considering we don't need to install any new packages, perhaps we could make it easier for the community and pin it?
If we'll find that some core st2 dependency needs a newer version, - we'll update it as the first priority without a notice or support promises.
But as long as there is no conflict here and it doesn't break anything, - pinning should be fine.

Is that OK?

guzzijones:
Wfm

armab:
@khushboobhatia01 Could you please create a new dedicated PR for pinning the `tenacity`, as per `tooz` requirements so we could merge that fix?
https://github.com/openstack/tooz/blob/master/requirements.txt#L9

arm4b (changes requested):
Per discussion here, we'll need a PR that pins the tenacity dependency only.

armab:
@khushboobhatia01 As a follow-up to this,
could you please create a PR to pin the tenacity dependency? I'd think it will be helpful for everyone.

khushboobhatia01:
> @khushboobhatia01 As a follow-up to this,
> 
> could you please create a PR to pin the tenacity dependency? I'd think it will be helpful for everyone.

Yes, @armab

PR #5529, opened 2022-01-06
Use byte type lock name which is supported by all tooz drivers

khushboobhatia01:
String lock names are not supported by all the tooz drivers because of unsafe encode/decode and concatenation which results in error with string types.

Example:
- etcd3gw => https://opendev.org/openstack/tooz/src/branch/master/tooz/drivers/etcd3gw.py#L72 `TypeError: can't concat bytes to str`

- memcached => https://opendev.org/openstack/tooz/src/branch/master/tooz/drivers/memcached.py#L79 `TypeError: can't concat bytes to str`

- etcd3 => https://opendev.org/openstack/tooz/src/branch/master/tooz/drivers/etcd3.py#L70 `AttributeError: 'str' object has no attribute 'decode'`

Remaining drivers support both str and byte type
- consul => https://opendev.org/openstack/tooz/src/branch/master/tooz/drivers/consul.py#L295 `Safe decode`

- etcd => https://opendev.org/openstack/tooz/src/branch/master/tooz/drivers/etcd.py#L255 `Safe encode`

- redis => https://opendev.org/openstack/tooz/src/branch/master/tooz/drivers/redis.py#L81 `Converts to string`

- zookeeper => https://opendev.org/openstack/tooz/src/branch/master/tooz/drivers/zookeeper.py#L447 `Safe decode`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -39,6 +39,10 @@ Fixed
 
 * Fix ``st2-self-check`` script reporting falsey success when the nested workflows runs failed. #5487
 
+* Use byte type lock name which is supported by all tooz drivers. #5529
+
+  Contributed by @khushboobhatia01
+
 3.6.0 - October 29, 2021
 ------------------------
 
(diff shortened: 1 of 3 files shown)


armab (approved):
Thanks for the research and fix that supports all the drivers :+1:

PR #5534, opened 2022-01-12
Fix issue with pack option not working when running policy list cli

momokuri-3:
This error occurs when run ```policy list``` with st2-client.
If the ```--pack``` option is given, it should return only the resources belonging to the provided pack.
However, there was no change in the output when the ```--pack``` option was specified or not.

Don't specify the pack option
```
st2 policy list

+---------------------------------+---------------------+-------------------------+---------+
| ref                             | resource_ref        | policy_type             | enabled |
+---------------------------------+---------------------+-------------------------+---------+
| testpack1.test.concurrency      | testpack1.test      | action.concurrency      | True    |
| testpack2.test.concurrency      | testpack2.test      | action.concurrency      | True    |
+---------------------------------+---------------------+-------------------------+---------+
```

Specify the pack option
```
st2 policy list --pack testpack1

+---------------------------------+---------------------+-------------------------+---------+
| ref                             | resource_ref        | policy_type             | enabled |
+---------------------------------+---------------------+-------------------------+---------+
| testpack1.test.concurrency      | testpack1.test      | action.concurrency      | True    |
| testpack2.test.concurrency      | testpack2.test      | action.concurrency      | True    |
+---------------------------------+---------------------+-------------------------+---------+
```

The code for the case where the pack option was specified was missing, so it will be corrected.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,10 @@ Fixed
 
   Contributed by @blackstrip
 
+* Fix issue with pack option not working when running policy list cli #5534
+
+  Contributed by @momokuri-3
+
 Added
 ~~~~~
 
--- a/st2client/st2client/commands/policy.py
+++ b/st2client/st2client/commands/policy.py
@@ -109,20 +109,19 @@ class PolicyListCommand(resource.ContentPackResourceListCommand):
 
     @resource.add_auth_token_to_kwargs_from_cli
     def run(self, args, **kwargs):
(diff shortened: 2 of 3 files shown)


armab (changes requested):
Thanks for the contribution!

Could you please include a Unit test case for this bug?
We'll also need a Changelog update.

armab (approved):
Looks good, thanks for the contribution! :+1:

PR #5513, opened 2021-12-20
Router fix to handle url encoded form payload

sravs-dev:
Fix for the issue https://github.com/StackStorm/st2/issues/5512

When there is a POST request with payload type "application/x-www-form-urlencoded" , Input body is of type bytes. We need to convert it to string before processing

Issue #4853, opened 2020-01-28
URL Encoded Webhook  - Python3 Encoding Failure


jamesdreid:
## SUMMARY

Provide a quick summary of your bug report.

### STACKSTORM VERSION

Paste the output of ``st2 --version``:

st2 3.1.0, on Python 3.6.8

##### OS, environment, install method

Post what OS you are running this on, along with any other relevant information/

- e.g. Docker, Vagrant, Kubernetes, etc. Describe how you installed ST2
- e.g. one-line install, custom install, etc -->

Ubuntu 18.04.3 LTS - Manual install

## Steps to reproduce the problem

When running the system with Python3, configuring a webhook as a trigger to receive data from a remote source.  Configure the remote source to send text data that is URL encoded in the body of the message, as shown in sample curl output below:

curl --location --request POST 'https://127.0.0.1/api/v1/webhooks/test_hook' \
--header 'St2-Api-Key: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx' \
--header 'Content-Type: application/x-www-form-urlencoded' \
--data-raw 'MyVariableOne=ValueOne&MyVariableTwo=ValueTwo'

## Expected Results

What did you expect to happen when running the steps above?

The system should properly process the URL encoded body data and should return it a proper JSON

{
    "MyVariableOne": [
        "ValueOne"
    ],
    "MyVariableTwo": [
        "ValueTwo\n"
    ]
}

## Actual Results

What happened? What output did you get?

The process fails with the following debug output:

```
[2020-01-27 11:49:57 +0000] [10770] [DEBUG] POST /v1/webhooks/valero_cpe
2020-01-27 11:49:57,579 140064340756784 DEBUG (unknown file) [-] Match path: /v1/webhooks/valero_cpe
2020-01-27 11:49:57,580 140064340756784 INFO (unknown file) [-] 684b74a5-5ea6-4602-9006-2a9a0033710c - POST /v1/webhooks/valero_cpe with query={} (method='POST',path='/v1/webhooks/valero_cpe',remote_addr='127.0.0.1',query={},request_id='684b74a5-5ea6-4602-9006-2a9a0033710c')
[...]

Issue #5512, opened 2021-12-20
st2 POST API with content type= "application/x-www-form-urlencoded" not working


sravs-dev:
## SUMMARY

Provide a quick summary of your bug report.

### STACKSTORM VERSION
3.5

##### OS, environment, install method

Post what OS you are running this on, along with any other relevant information/

Running on Mac OS, installed using Docker. Issue is related to the rest api handling, nothing to do with host environment.

## Steps to reproduce the problem

Make a post api to /sso/callback with a payload body of type "application/x-www-form-urlencoded"

## Expected Results

What did you expect to happen when running the steps above?
st2 api should be able to parse the payload body into string and process it

## Actual Results

What happened? What output did you get?

API failed to load binary input body to a dictionary, hence returned 500 with the following stacktrace. 
Fix needed: Binary input should be converted to string before processing.
 
`2021-12-20 18:04:00,275 139680124778496 DEBUG router [-] Missing x-api-model definition for st2auth.controllers.v1.sso:idp_callback_controller.post, using generic Body model.
2021-12-20 18:04:00,275 139680124778496 ERROR error_handling [-] API call failed: __init__() keywords must be strings
Traceback (most recent call last):
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2common/middleware/error_handling.py", line 49, in __call__
    return self.app(environ, start_response)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2common/router.py", line 735, in as_wsgi
    resp = self(req)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2common/router.py", line 560, in __call__
    instance = self._get_model_instance(model_cls=model, data=data)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2common/router.py", line 750, in _get_model_instance
    raise e
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2common/router.py", line 740, in _get_model_instance
    instance = model_cls(**data)
TypeError: __init__() keywords must be strings (_exception_class='TypeError',_exception_message='__init__() keywords must be strings',_exception_data={})
2021-12-20 18:04:00,275 139680124778496 DEBUG router [-] Match path: /sso/callback
[...]

PR #5536, opened 2022-01-12
Deserialize payload to str when content type is www-form-urlencoded

sravs-dev:
…oded

Opened in favor of https://github.com/StackStorm/st2/pull/5513

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,7 +7,7 @@ in development
 Fixed
 ~~~~~
 
-* Fix deserialization bug for url encoded payloads. #5513
+* Fix deserialization bug for url encoded payloads. #5536
 
   Contributed by @sravs-dev
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,10 @@ in development
 Fixed
 ~~~~~
 
+* Fix deserialization bug for url encoded payloads. #5513
+
+  Contributed by @sravs-dev
+
 * Fix Type error for ``time_diff`` critera comparison. convert the timediff value as float to match
   ``timedelta.total_seconds()`` return. #5462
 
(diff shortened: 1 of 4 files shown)


sravs-dev:
@m4dcoder closed https://github.com/StackStorm/st2/pull/5513 because of CLA issues and opened this one. Could you please review. 
Thanks!

armab (approved):
LGTM.
Thanks for the contribution and recreating the clean PR :+1:

armab:
@m4dcoder looks your previous review comments in https://github.com/StackStorm/st2/pull/5513 were addressed, but please comment if you think we need something else here before merging it tomorrow.

jamesdreid:
I have not dug too deeply into this PR, but at face value this may fix Issue #4853 as well.

m4dcoder (changes requested):
@sravs-dev Is there a unit test that shows an example for unsupported content type for x-www-form-urlencoded and how failure is handled?

sravs-dev:
> @sravs-dev Is there a unit test that shows an example for unsupported content type for x-www-form-urlencoded and how failure is handled?

@m4coder I dont see any existing unit test for that. Do you have any sample invalid data in mind for this content type? Since its key value pair, I am wondering what could be used as an invalid data.

m4dcoder:
@sravs-dev What I meant on unsupported content type for x-www-form-urlencoded is a use case where six.ensure_str returns an error and how that error is handled? Is this possible to test?

sravs-dev:
> six.ensure_str

@m4dcoder Thanks for your reply.  six.ensure_str takes bytes input, I dont think we can simulate this input as its sent by the API internally. We can only send the payload as key value pair for x-www-form-urlencoded content type.
If six. ensure_str throws error, I guess we would get 500 error similar to what I faced in this [issue](https://github.com/StackStorm/st2/issues/5512) . 
I will go ahead and revert the test case that's passing invalid data to the API.

m4dcoder (approved):

sravs-dev:
@armab @m4dcoder On the latest commit, all CI Checks have passed except Python 3.8 chunk 1 which is failing in other PRs as well, so it's not related to my changes.

PR #5538, opened 2022-01-14
WinRM parameter passing fails for larger scripts

ashwini-orchestral:
When winrm-ps-script runner type was used, found that we were unable to pass parameters successfully from Stackstorm to the PowerShell script. We noticed inconsistent parameter passing to the PowerShell scripts, seemingly being able to pass inputs to smaller scripts but not to larger ones.

### DIFF
--- a/contrib/runners/winrm_runner/winrm_runner/winrm_base.py
+++ b/contrib/runners/winrm_runner/winrm_runner/winrm_base.py
@@ -403,7 +403,7 @@ Add-Content -value $data -encoding byte -path $filePath
             # the following wraps the script (from the file) in a script block ( {} )
             # executes it, passing in the parameters built above
             # https://docs.microsoft.com/en-us/powershell/scripting/core-powershell/console/powershell.exe-command-line-help
-            ps = "& {%s}" % (tmp_script)
+            ps = tmp_script
             if params:
                 ps += " " + params
             return self._run_ps(ps)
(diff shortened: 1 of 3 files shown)


contrib/runners/winrm_runner/winrm_runner/winrm_base.py
```diff
@@ -401,9 +401,9 @@ def _run_ps_script(self, script, params=None):
         # handle deletion of the temporary file on exit of the with block
         with self._tmp_script(tmp_dir, script) as tmp_script:
             # the following wraps the script (from the file) in a script block ( {} )
-            # executes it, passing in the parameters built above
+            # executes it, passing in the parameters built above.
             # https://docs.microsoft.com/en-us/powershell/scripting/core-powershell/console/powershell.exe-command-line-help
-            ps = "& {%s}" % (tmp_script)
```
amanda11, line 406: Do we know why it was failing with the call operator and scriptblock syntax? I'm not a PowerShell expert, but I'm just wondering if there was a reason for using the & and {} syntax in the first-place?
e.g. in case there was quotes in anything?

ankita-orchestral: Wrapping the braces around the script path is unnecessary and causes the arguments to be lost.

Working with the winrm-ps-script runner type, we have found that we are unable to pass parameters successfully from StackStorm to the PowerShell script. We noticed inconsistent parameter passing to the PowerShell scripts, seemingly being able to pass inputs to smaller scripts but not to larger ones. 
This is a known issue.

We think there was no use of & and {} syntax in the first place.

amanda11: Thanks for the explanation.

amanda11 (approved):
LGTM - thanks for contribution.

PR #5552, opened 2022-01-24
Minor updates for Rocky

amanda11:
Part of https://github.com/StackStorm/community/issues/88

### DIFF
--- a/contrib/linux/actions/service.py
+++ b/contrib/linux/actions/service.py
@@ -70,6 +70,7 @@ elif (
     re.search(distro, "Redhat")
     or re.search(distro, "Fedora")
     or re.search(distro, "CentOS Linux")
+    or re.search(distro, "Rocky Linux")
 ):
     cmd_args = ["systemctl", args["act"], args["service"]]
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,6 +15,9 @@ Fixed
 Added
 ~~~~~
 
+* Minor updates for RockyLinux. 
+  Contributed by Amanda McGuinness (@amanda11 intive)
+
 * Added st2 API get action parameters by ref. #5509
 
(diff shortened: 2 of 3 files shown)


cognifloyd (approved):

contrib/linux/actions/service.py
```diff
@@ -70,6 +70,7 @@ def get_linux_distribution():
     re.search(distro, "Redhat")
     or re.search(distro, "Fedora")
     or re.search(distro, "CentOS Linux")
```
rush-skills, line 72: Not exactly related to this PR, but can we modify this to say `CentOS` instead of `CentOS Linux`? The stream releases won't match the current version, since `platform.linux_distribution()` for stream returns 
```
platform.linux_distribution()
>> ('CentOS Stream', '8', '')
```
and the `lsb_release` way returns
```
→ lsb_release -i -s
CentOSStream
```
On CentOS 7, I guess this will still fail as well with Python >= 3.8
```
→ lsb_release -i -s
CentOS
```

cognifloyd: I would be happy with that. Could you file a new PR?

armab, line 72:
```suggestion
    or re.search(distro, "CentOS")
```
I guess that's just a diff we could include here

For the core `linux` pack changes, I think we'll need to bump the pack version as well?

.circleci/config.yml
```diff
@@ -129,7 +129,7 @@ jobs:
           exclusive: true   # default - true
           # Temporary workaround for Circle CI issue
           # https://discuss.circleci.com/t/setup-remote-docker-connection-failures/26434
```
armab, line 131: If it'll work, I think we could remove these lines to avoid confusion in the future

armab (approved):
Thanks @amanda11 ! :100: 
Looks we're good now!

just added one more minor commit to sync up the CircleCI config with the upstream st2-packages version.

cognifloyd (approved):

PR #5556, opened 2022-01-26
check if value is actually jinja before adding as template

guzzijones:
check if the value is actually jinja and not a string like `{"someurl": "http://test?value={{test"}`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,10 @@ Fixed
 
   Contributed by @blackstrip
 
+* Fix exception thrown if action parameter contains {{ or {% and no closing jinja characters. #5556
+
+  contributed by @guzzijones12
+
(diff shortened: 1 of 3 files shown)


cognifloyd (approved):

armab (changes requested):
Please include a Changelog for this fix.

Did you get an exception somewhere, how the error looked like?
A more specific example with steps to reproduce and log in the first message would help for the future history and others hitting the same issue.

guzzijones:
If your parameter value has a `{{` or `{%` in it and it is not closed you will see the following exception in the st2workflowengine.log file: `TemplateSyntaxError: expected token 'end of print statement', got 'string'`

armab (approved):
Thanks for Changelog and additional context! :+1:

PR #5564, opened 2022-02-02
Add a Changelog Checker

armab:
The Changelog is important so the community can track what changes they might expect in the upcoming releases or developers could pinpoint the changes to a particular PR and diff.

We could naturally forget or new contributors might not know about updating Changelog. Additionally, we don't want to go to every PR and ask for a Changelog addition.

This GH Action (https://github.com/marketplace/actions/changelog-checker) will verify if `CHANGELOG.rst` was updated for every PR:
image

<details>
  <summary>GH Actions view</summary>
  
  image
</details>

For rare cases when it's needed to skip the check, assign `no changelog` label for the PR.

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -1,0 +1,25 @@
+name: Checks
+
+on:
+  pull_request:
+    types: [assigned, opened, synchronize, reopened, labeled, unlabeled]
+    branches:
+      - master
+      - v[0-9]+.[0-9]+
+
+jobs:
+  # Changelog checker will verify if CHANGELOG.rst was updated for every PR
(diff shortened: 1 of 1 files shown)


cognifloyd (approved):
This is ok because it provides a way to skip the changelog when needed.

armab:
We'll need something similar for StackStorm-Exchange to notify users about the Changelog change + Pack version bump as well. The majority of the PR reviews there are about asking: bump pack version + Changelog update.

Created https://github.com/StackStorm-Exchange/ci/issues/134 as part of the StackStorm-Exchange Improvements project: https://github.com/orgs/StackStorm-Exchange/projects/1

PR #5581, opened 2022-02-28
Fix build failures due to newer  MarkUpSafe

amanda11:
Fix build errors due to markupsafe 2.1.0 removing soft_unicode, which were breaking the ST2 package tests.

Also, updated the GHA workflow as problems found after requirements.txt was altered. Altered so that it only looks for a cache on exact match of key instead of partial, so that doesn't try and pick up cache from previous version of requirements.txt.


armab (approved):
Thanks a lot! :green_apple:

PR #5596, opened 2022-03-28
[WIP] Add garbage collection for trace and rule_enforcement models

amanda11:
Add garbage collection for trace and rule_enforcement models.

### DIFF
--- a/st2common/setup.py
+++ b/st2common/setup.py
@@ -53,6 +53,8 @@ setup(
         "bin/st2-register-content",
         "bin/st2-purge-executions",
         "bin/st2-purge-trigger-instances",
+        "bin/st2-purge-trace",
+        "bin/st2-purge-rule-enforcement",
         "bin/st2-run-pack-tests",
         "bin/st2ctl",
         "bin/st2-generate-symmetric-crypto-key",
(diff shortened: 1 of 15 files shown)


cognifloyd (approved):

st2common/setup.py
```diff
@@ -53,6 +53,8 @@
         "bin/st2-register-content",
         "bin/st2-purge-executions",
         "bin/st2-purge-trigger-instances",
+        "bin/st2-purge-trace",
```
Kami, line 56: For consistency it would be good to use plural form of the name here as well - e.g. purge-traces and purge-rule-enforcements.

st2tests/st2tests/config.py
```diff
@@ -484,6 +484,16 @@ def _register_garbage_collector_opts():
             default=None,
             help="Trigger instances older than this value (days) will be automatically deleted.",
         ),
+        cfg.IntOpt(
+            "rule_enforcement_ttl",
+            default=None,
+            help="Rule enforcements older than this value (days) will be automatically deleted.",
```
Kami, line 490: May be good to add something along the lines  of "Defaults to None aka disabled by default." / "If set to None, it's disabled." to the end of the help strings.

amanda11:
@Kami @cognifloyd i will aim to push a pr to address kami's comments tomorrow, as they were raised after merge.
But will do before I do corresponding doc change.

PR #5606, opened 2022-04-02
Bump black to v22.3.0 and update format

cognifloyd:
black v22.3.0 fixes an incompatibility with click, which is a transitive dep.
see: https://github.com/psf/black/issues/2964

I've had to update other projects recently for the same issue.

I just saw an error in CI that hit this issue, so let's update before anything else bites us.
see: https://github.com/StackStorm/st2/runs/5796530459?check_suite_focus=true

Also includes a minor fixture update (test_content_pack submodule) because I'm tired of Make complaining. There is no functional difference with the update. The test pack has newer commits because of CI updates on the exchange.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -187,6 +187,11 @@ Fixed
 
   Contributed by @nzlosh
 
+Changed
+~~~~~~~
+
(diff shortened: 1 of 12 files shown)


nzlosh (approved):
LGTM

PR #5638, opened 2022-05-04
pin typing-extensions<4.2 to fix st2client install

cognifloyd:
st2client requires importlib-metadata which requires typing-extensions.
But typing-extensions v4.2.0 dropped support for python3.6, so pin it.

Our CI has been broken on st2 master since April 17: https://github.com/StackStorm/st2/actions/workflows/ci.yaml?page=2
Turns out it’s because of typing-extensions which dropped support for python 3.6 on April 17 with their 4.2.0 release: https://github.com/python/typing/blob/master/typing_extensions/CHANGELOG.md

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -202,6 +202,9 @@ Changed
 
 * Bump black to v22.3.0 - This is  used internally to reformat our python code. #5606
 
+* Pin ``typing-extensions<4.2`` (used indirectly by st2client) to maintain python 3.6 support. #5638
+
+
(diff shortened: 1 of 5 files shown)


nzlosh (approved):

armab (approved):

ytjohn:
Do we still need to support Python 3.6?  This is mainly an issue  with st2sensorcontainer. Many of our sensors use pydantic which has dropped support for typing-extensions<4.2.



https://github.com/pydantic/pydantic/issues/4885

Issue #5337, opened 2021-08-22
SSL Error with st2 sensor connection redis


shellHell:
Getting these errors while connecting to redis with SSL. A normal action runs fine, sensor seems to be having problem. I did try to use the fixes mentioned in https://github.com/eventlet/eventlet/issues/371 but none of the different combinations of library versions seem to work form my case. 
```
021-08-17 14:32:25,839 WARNING [-] Sensor "XYZSensor" run method raised an exception: Error while reading from socket: ('timed out',).
Traceback (most recent call last):
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/redis/connection.py", line 198, in _read_from_socket
    data = recv(self._sock, socket_read_size)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/redis/_compat.py", line 72, in recv
    return sock.recv(*args, **kwargs)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/eventlet/green/ssl.py", line 231, in recv
    return self._base_recv(buflen, flags, into=False)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/eventlet/green/ssl.py", line 258, in _base_recv
    read = self.read(nbytes)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/eventlet/green/ssl.py", line 176, in read
    super(GreenSSLSocket, self).read, *args, **kwargs)
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/eventlet/green/ssl.py", line 156, in _call_trampolining
    timeout_exc=timeout_exc('timed out'))
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/eventlet/hubs/__init__.py", line 159, in trampoline
    return hub.switch()
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/eventlet/hubs/hub.py", line 298, in switch
    return self.greenlet.switch()
ssl.SSLError: ('timed out',)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/opt/stackstorm/st2/lib/python3.6/site-packages/st2reactor/container/sensor_wrapper.py", line 229, in run
    self._sensor_instance.run()
  File "/opt/stackstorm/packs/st2sensors/sensors/vmc_event_notification_sensor.py", line 56, in run
    self.pubsub.subscribe('XYZEvents')
[...]

PR #5660, opened 2022-06-21
Fix problems connecting to redis sentinel with SSL

amanda11:
When connecting to redis SSL via sentinel then 2 problems are encountered:

1. redis 3.5.3 did not support ssl parameters for sentinel, see https://github.com/redis/redis-py/pull/1419. So upping the version of redis library
2. When using eventlet and redis SSL then there are connection problems, due to the fact the redis library has expected timeouts, but monkey patched SSL library is throwing different exception. See https://github.com/eventlet/eventlet/issues/692. This causes problems on connecting to redis SSL via sentinel. It also is likely to be the cause of error: https://github.com/StackStorm/st2/issues/5337 as the stack trace encountered is the same.
Therefore when monkey patching SSL library, keep the exception raised on timeout to be socket.timeout.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,11 @@ Changelog
 in development
 --------------
 
+Fixed
+~~~~~
+
+* Fix redis SSL problems with sentinel #5660
+
 
 3.7.0 - May 05, 2022
 --------------------
--- a/fixed-requirements.txt
+++ b/fixed-requirements.txt
@@ -50,7 +50,7 @@ python-keyczar==0.716
 pytz==2021.1
 pywinrm==0.4.1
 pyyaml==5.4.1
-redis==3.5.3
+redis==4.1.4
 requests[security]==2.25.1
 retrying==1.3.3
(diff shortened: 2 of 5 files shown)


raviorch:
Amanda, Please let us know which version of Redis you used to fix the Redis and Sentinel issues for TLS. Redis officially supports TLS from their 6.0 version onwards. Text from redis.io : "SSL/TLS is supported by Redis starting with version 6 as an optional feature that needs to be enabled at compile time." Are you using the latest Redis version 7.0 or 6.2.2?

amanda11:
I used 6.2.7 and compiled at run-time to enable SSL.

raviorch:
Amanda, Thanks for the update, I was wondering if the ssl fix that you have put in also has the fix for providing an username password for Redis. Please let us know.

amanda11:
As far as I know there is no issue with username/password with redis. I've used StackStorm connected to redis secured by username/password before quite happily, on many StackStorm versions including 3.7.0

PR #5667, opened 2022-07-14
Fixed a bug where key value pair was not returning data correctly when get_by_name function is called

bharath-orchestral:
Bug: Calling 'get_by_name' on client for getting key details was not returning any results despite key being stored. Following is the sample code for replicating the bug
```
from st2client.client import Client

class actionContext(Action):
    def run(self):

        client = Client(base_url='http://localhost')
        result = client.keys.get_by_name(name='test_key',scope='user')
        self.logger.debug(result)
        return result
```
Resolution: 'keys' property in Client is of type 'ResourceManager'. We created a new class 'KeyValuePairResourceManager' which inherits from 'ResourceManager'. We assigned 'KeyValuePairResourceManager' to 'keys' property. We have created get_by_name in this class. KeyValuePairResourceManager's get_by_name calls "/keys/{keyname}" url while the old function used to call '/keys' url.
We looked into test cases for classes which inherit from 'ResourceManager' but did not find any. There are test cases
for functions in 'ResourceManager' but not for classes inheriting from it.
But we wrote a unit test case which tests KeyValuePairResourceManager's get_by_name function


amanda11:
Many thanks for the contribution, it looks good. Could you please sign the contributor agreement and address the lint checks please?

cognifloyd:
Looked good.
Beyond signing the CLA:
Please add a changelog entry and reformat the changed files with black. Plus I left one little nitpick.

bharath-orchestral:
Closing this PR due to commit issues. Will submit new PR with correct commits

PR #5670, opened 2022-07-16
use the runner.yaml constant

cognifloyd:
A very minor refactor.

I noticed that we had a constant and weren't using it. So, let's use it.

https://github.com/StackStorm/st2/blob/8496bb2407b969f0937431992172b98b545f6756/st2common/st2common/constants/runners.py#L40-L41


rush-skills (approved):
Good catch 👍🏼

amanda11 (approved):
LGTM

PR #5671, opened 2022-07-16
delete pointless py header in empty __init__.py

cognifloyd:
`__init__.py` files only need a copyright header if there's actually something in the file.

Plus, `pants` can skip introspecting files when they are empty. So having a comment-only python file is wasteful.


rush-skills (approved):

amanda11 (approved):
LGTM

PR #5673, opened 2022-07-17
Use PEP 440 style requirements

cognifloyd:
This adds support for PEP 440 requirements instead of using the pip-specific VCS format.
see: https://peps.python.org/pep-0440/#direct-references

This facilitates using newer tooling that does not support using the legacy pip VCS format.

Summary of changes:
- `*requirements.txt`: reformat 6 requirements repeated across 11 files (this is the primary part of the PR, everything else supports it)
- Python: add 6 lines to 19 copies of `dist_utils.py`
- Python: add 8 lines and change 1 line in `scripts/fixate-requirements.py`

## `*requirements.txt`

So, this change moves [from this legacy pip VCS](https://github.com/StackStorm/st2/pull/5673/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552L23-L25) format:

https://github.com/StackStorm/st2/blob/be437e97c8d1491eba4f28edfd65c7a790f21656/requirements.txt#L23-L25

[to this PEP 440 direct reference](https://github.com/StackStorm/st2/pull/5673/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552R69-R71) format:

https://github.com/StackStorm/st2/blob/4d46a28424c0798caf3506f821c5954927aa3f7c/requirements.txt#L69-L71

`logshipper` is one of the VCS requirements that we convert to PEP 440. It is linux-only and cannot be installed on MacOS X. So, this also adds `platform_system=="Linux"` to the logshipper dep (which was already done for logshipper's pyinotify dep) to allow building a virtualenv on MacOS X. That looks like this:

https://github.com/StackStorm/st2/blob/4d46a28424c0798caf3506f821c5954927aa3f7c/requirements.txt#L31

And it standardizes the markers format (space before and after `;`) of the `pyinotify` dep which is required by logshipper.

## `dist_utils.py`

I adjusted the `dist_utils.py` functions to maintain the same functionality. Note however that I plan to use pants to replace all the python packaging bits, including dist_utils.py. So the dist_utils.py changes are a stop-gap to keep things working until we get the rest of the pants changes in. I added [these lines](https://github.com/StackStorm/st2/pull/5673/files#diff-da36b51a6eb1b4d8f139381743fcdb3e7b8ed326e80e1321d62e0bb219646cb1R125-R130) to 20 copies of the `dist_utils.py` file:

https://github.com/StackStorm/st2/blob/2b0cd753699b45789e6b5c1fa99b889f71ce4170/scripts/dist_utils.py#L125-L130

## `fixate-requirements.py`

I also had to make a slight adjustment to `fixate-requirements.py`. This codepath is only used on CircleCI during package build when rpmbuild runs a make target that runs fixate-requirements. It does not run normally because we use pip 20, but rpmbuild runs the script before pipis updated from 19 to 20. So, this makes that corner case in our circleci workflow work like everywhere else we run fixate-requirements. Here is [the new code](https://github.com/StackStorm/st2/pull/5673/files#diff-2636adacf325215342f11387a5a05c94a1f24cd4f9f76c038641dc09dee691c6R228-R235) in `fixate-requirements.py`:

https://github.com/StackStorm/st2/blob/97dd34fa29cabb1130d1855efeb2a3de83920e81/scripts/fixate-requirements.py#L228-L235

<hr>

This depends on https://github.com/StackStorm/st2-packages/pull/722


amanda11 (approved):
LGTM

PR #5674, opened 2022-07-18
Add support for PyCharm Pro's Remote Debug Server

cognifloyd:
Allow debugging st2 services in PyCharm Pro's Remote Debug Server.

To enable debugging, before starting the st2 process(es), you need to:

  - install pydevd-pycharm (use the version specific to your pycharm install) in StackStorm's virtualenv.
  - set the ``ST2_PYCHARM_DEBUG`` environment var to something non-empty like "true".
  - (optional) set environment vars ``ST2_PYCHARM_DEBUG_HOST`` and ``ST2_PYCHARM_DEBUG_PORT`` if pycharm is listening somewhere other than ``localhost:5000``.

see: https://www.jetbrains.com/help/pycharm/remote-debugging-with-product.html#remote-debug-config

### DIFF
--- a/st2common/st2common/util/monkey_patch.py
+++ b/st2common/st2common/util/monkey_patch.py
@@ -19,6 +19,7 @@ Module for performing eventlet and other monkey patching.
 
 from __future__ import absolute_import
 
+import os
 import sys
 
 __all__ = [
@@ -43,6 +44,23 @@ def monkey_patch(patch_thread=None):
                          patched unless debugger is used.
     :type patch_thread: ``bool``
     """
+    if os.environ.get("ST2_PYCHARM_DEBUG", False):
+
+        # pydevd_pycharm uses this to save copies of stdlib modules before
+        # eventlet monkey patches them, or the debugger will not work.
+        # see: https://intellij-support.jetbrains.com/hc/en-us/community/posts/360000333980
+        os.environ["GEVENT_SUPPORT"] = "True"
+
+        import pydevd_pycharm
+
+        pydevd_pycharm.settrace(
+            os.environ.get("ST2_PYCHARM_DEBUG_HOST", "localhost"),
+            port=os.environ.get("ST2_PYCHARM_DEBUG_PORT", 5000),
+            stdoutToServer=True,
+            stderrToServer=True,
+            patch_multiprocessing=True
+        )
+
     # Eventlet when patched doesn't throw the standard ssl error on timeout, which can break
     # some third-party libraries including redis SSL.
     # See: https://github.com/eventlet/eventlet/issues/692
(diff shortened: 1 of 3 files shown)


cognifloyd:
@guzzijones JetBrains is based in the [Czech Republic](https://en.wikipedia.org/wiki/Czech_Republic) which became a state in 1993. It's prdecessor. Czechoslovakia escaped Soviet control in 1989.

Also, JetBrains has suspended operations in Russia due to the invasion of Ukraine: https://blog.jetbrains.com/blog/2022/03/11/jetbrains-statement-on-ukraine/

armab (changes requested):
I don't think it's a good idea to integrate additional logic related to any IDE into the st2 core.

ST2_PYCHARM_DEBUG, ST2_PYCHARM_DEBUG_HOST, ST2_PYCHARM_DEBUG_PORT or anything else shouldn't be present in the st2common.

st2common/st2common/util/monkey_patch.py
```diff
@@ -117,4 +135,8 @@ def is_use_debugger_flag_provided():
         if arg.startswith(PARENT_ARGS_FLAG) and USE_DEBUGGER_FLAG in arg:
             return True
 
+    # 3. Check for ST2_PYCHARM_DEBUG env var
+    if os.environ.get("ST2_PYCHARM_DEBUG", False):
+        return True
```
cognifloyd, line 140: This part is not pycharm specific - it allows enabling debug mode via env var instead of changing the cli args.

```suggestion
    # 3. Check for ST2_DEBUG env var
    if os.environ.get("ST2_DEBUG", False):
        return True
```

st2common/st2common/util/monkey_patch.py
```diff
@@ -43,6 +44,23 @@ def monkey_patch(patch_thread=None):
...
+    if os.environ.get("ST2_PYCHARM_DEBUG", False):
+
+        # pydevd_pycharm uses this to save copies of stdlib modules before
+        # eventlet monkey patches them, or the debugger will not work.
+        # see: https://intellij-support.jetbrains.com/hc/en-us/community/posts/360000333980
+        os.environ["GEVENT_SUPPORT"] = "True"
+
+        import pydevd_pycharm
+
+        pydevd_pycharm.settrace(
```
cognifloyd, line 56:
```suggestion
    if os.environ.get("ST2_DEBUG", False):

        # pydevd_pycharm uses this to save copies of stdlib modules before
        # eventlet monkey patches them, or the debugger will not work.
        # see: https://intellij-support.jetbrains.com/hc/en-us/community/posts/360000333980
        os.environ["GEVENT_SUPPORT"] = "True"

        import pydevd_pycharm

        pydevd_pycharm.settrace(
```

@armab Do you have a recommendation on how we can make this extensible so that people can inject whatever debugger before monkey_patching occurs?

cognifloyd (commented):
I can see how including something IDE-specific is distasteful, but I believe it is warranted as yet-another-cost of our reliance on eventlet.

It is impossible afaik to hook the debugger into eventlet-based code without changing the sources (adding a call before `eventlet.monkey_patch()`). For normal python processes, there are several other ways to attach a debugger to the process, but this is the best way I've found with st2 code so far.

I've isolated these codepaths so that it is as minimal as possible, but if there's another way to enable using a debugger, I'm open to that.

guzzijones:
Dang. All this time we been told we cannot use pycharm due to its Russian
connection.

PR #5675, opened 2022-07-18
add ST2_USE_DEBUGGER env var to mirror --use-debugger arg

cognifloyd:
Extracted the st2-specific bits from #5674.

This allows setting `ST2_USE_DEBUGGER` as an env var instead of passing `--use-debugger` as a command line arg. An env var is more convenient in some cases, such as when using containerized/docker st2.


rush-skills (approved):
I am okay with this

amanda11 (approved):

PR #5676, opened 2022-07-18
Drop st2exporter

### LINKED ISSUE #5295: Remove st2exporter
I think we should delete all of st2exporter. According to StackStorm/st2-packages#367 it was abandoned years ago.
See also: StackStorm/st2-packages#84 StackStorm/st2packaging-dockerfiles#30

Is there any reason to keep anything in the st2exporter directory?

cognifloyd:
Removed st2exporter service. It is unmaintained and does not get installed. It was originally meant to help with analytics by exporting executions as json files that could be imported into something like elasticsearch. Our code is now instrumented to make a wider variety of stats available to metrics drivers.

Requires StackStorm/st2-packages#723
Closes #5295
Closes StackStorm/st2-packages#367


rush-skills (approved):
Good work!
I had a question wrt DB changes - will all existing st2exporter-related DB entries (marker, etc.) be automatically purged with these changes? Or is that something we make st2garbagecollector handle?

cognifloyd:
@rush-skills Good call. I added a migration util that can drop the marker collections.
I expect most people will not have the collections because we have not shipped st2exporter in years, and the collection doesn't get created automatically. My prod instance, which has been around since the early days of st2, does not have these collections. But, if someone does, here's a script to delete it.

armab (approved):

PR #5677, opened 2022-07-19
Fixed a bug where key value pair was not returning data correctly when get_by_name function is called.

bharath-orchestral:
Bug: Calling 'get_by_name' on client for getting key details was not returning any results despite key being stored. Following is the sample code for replicating the bug
```
from st2client.client import Client

class actionContext(Action):
    def run(self):

        client = Client(base_url='http://localhost')
        result = client.keys.get_by_name(name='test_key',scope='user')
        self.logger.debug(result)
        return result
```        
Resolution: 'keys' property in Client is of type 'ResourceManager'. We created a new class 'KeyValuePairResourceManager' which inherits from 'ResourceManager'. We assigned 'KeyValuePairResourceManager' to 'keys' property. We have created get_by_name in this class. KeyValuePairResourceManager's get_by_name calls "/keys/{keyname}" url while the old function used to call '/keys' url.
We looked into test cases for classes which inherit from 'ResourceManager' but did not find any. There are test cases
for functions in 'ResourceManager' but not for classes inheriting from it.
But we wrote a unit test case which tests KeyValuePairResourceManager's get_by_name function


bharath-orchestral:
Resubmitted the PR with correct credentials.

bharath-orchestral:
Can anyone help me on how to add changes to CHANGELOG.rst? Do I need to run some command like `make lint`?

cognifloyd:
@bharath-orchestral Thanks for working on this!

Here is the changelog: https://github.com/StackStorm/st2/blob/master/CHANGELOG.rst
There isn't a command that auto-generates a changelog entry (though I wish we had that).
Here is an example of a adding a changelog entry: https://github.com/StackStorm/st2/commit/8e908a5435a5a70a80d7fb4839e2fb8a663b1a6c

For this PR, you will need to put your changelog entry under `In Development` (at the top of the file) at the end of the list of `Fixed` issues.

cognifloyd:
To run the `black` formatter you can use `make black`.

st2client/st2client/models/core.py
```diff
@@ -879,3 +879,37 @@ def list(self, group_id, **kwargs):
...
+        else:
+            response = self.client.get(url)
+        
+        if response.status_code == http_client.NOT_FOUND:
+            # for query and query_with_count
+            return []
+        if response.status_code != http_client.OK:
+            self.handle_error(response)
+
+        return self.resource.deserialize(parse_api_response(response))
```
cognifloyd, line 915: needs new line at EOF

st2client/st2client/models/core.py
```diff
@@ -879,3 +879,37 @@ def list(self, group_id, **kwargs):
...
+        for k, v in six.iteritems(kwargs):
+            # Note: That's a special case to support api_key and token kwargs
+            if k not in ["token", "api_key", "params"]:
+                params[k] = v
+
+        url = "/%s/%s/?%s" % (
+            self.resource.get_url_path_name(),
+            name,
+            urllib.parse.urlencode(params),
+        )
```
cognifloyd, line 900: So this `get_by_name` is essentially a copy of `_query_details` with but it adds name in the path, right?

I wonder if there's a way to extend `_query_details` instead of duplicating all of this auth logic. maybe we could add a special dunder arg like `__path__` that gets added right after the url. So, in `_query_params()`, this would look like:

```python
for k, v in six.iteritems(kwargs):
    # ...
    if k not in ["token", "api_key", "params", "__path__"]:
        params[k] = v

url = "/%s/%s?%s" % (
    self.resource.get_url_path_name(),
    kwargs.get("__path__", ""),
    urllib.parse.urlencode(params),
)
```

So that this method would call 
```python
instances, _ = _query_details(__path__=f"{name}/", **kwargs)
return instances[0]
```

st2client/tests/unit/test_models.py
```diff
@@ -471,3 +471,19 @@ def test_resource_clone_failed(self):
...
+            return_value=base.FakeResponse(json.dumps(base.RESOURCES[0]), 200, "OK")
+        ),
+    )
+    def test_resource_get_by_name(self):
+        mgr = models.KeyValuePairResourceManager(base.FakeResource, base.FAKE_ENDPOINT)
+        # No X-Total-Count
+        resource = mgr.get_by_name("abc")
+        actual = resource.serialize()
+        expected = json.loads(json.dumps(base.RESOURCES[0]))
+        self.assertEqual(actual, expected)
```
cognifloyd, line 489: new line at EOF

st2client/st2client/models/core.py
bharath-orchestral: Fair enough. As far as  I understand just extending _query_details is not enough, we need to extend `query` function as well. Please let me know if it makes sense and if I should extend both `query`, `_query_details` functions or should I go ahead with suggested changes(extending `get_by_name` function).

cognifloyd: Let's keep the new `KeyValuePairResourceManager`, but simplify the implementation of the get by name method. That method does not need to use `query`, it can just directly use `_query_details`.

bharath-orchestral: The problem with using _query_details is that it expects multiple results and calls deserialise on each item. In our case '/v1/keys/{keyname}' is called and returns a single object.  This following line fails in our case
```
instances = [self.resource.deserialize(item) for item in items]
```

For this to work we need change _query_details so that it can handle both single, multiple objects. That's the reason I extended get_by_name directly and changed it. Do you want me to extend _query_details and change it such that it can handle single, multiple results or go ahead with my original implementation?

I personally wouldn’t want to modify query_details as this is used by a lot of functions and I don’t know the codebase enough to make these changes confidently without making sure it won’t have any adverse affect

st2client/tests/unit/test_models.py
bharath-orchestral: Added

st2client/st2client/models/core.py
bharath-orchestral: Added

st2client/st2client/models/core.py
cognifloyd: Bummer. I guess the code duplication is fine in this case.

cognifloyd (approved):

cognifloyd:
Deleting that __init__.py file also removed the folders it was in, so now there are failures in some unit tests. Could you restore that please? Or a .gitkeep file would probably work

bharath-orchestral:
I added that __init__.py in that folder but for some reason orquesta CI/ Integration Tests are failing. @cognifloyd  Do you know the reason for this?

st2tests/st2tests/fixtures/packs/dummy_pack_23/actions/workflows/__init__.py
```diff
@@ -0,0 +1,14 @@
+# Copyright 2020 The StackStorm Authors.
```
cognifloyd, line 1: `__init__.py` should be empty unless there is logic in the file. Please delete all these headers.

cognifloyd (commented):
Flaky test. Oh, sand I just noticed the header in that empty `__init__.py`. please don't add headers to empty `__init__.py` files like that.

st2tests/st2tests/fixtures/packs/dummy_pack_23/actions/workflows/__init__.py
bharath-orchestral: Deleted all the headers. Now __init__.py is empty

CHANGELOG.rst
```diff
@@ -19,6 +19,10 @@ Fixed
 
   Contributed by @S-T-A-R-L-O-R-D
 
+* Fixed a bug where calling 'get_by_name' on client for getting key details was not returning any results despite key being stored
```
amanda11, line 22: Minor comment, it is normal for changelog entries to include the PR number, can you add a #5677 to the changelog comment. This allows people to look up the PR later to see the changes made for a fix.

amanda11 (changes requested):
Looks good to me, just one really small comment on the changelog

m4dcoder (approved):

PR #5678, opened 2022-07-19
fix use correct os environ to check HTTPS_PROXY

wfgydbu:
Currently, st2client uses ``http_proxy`` to check os environment variable ``HTTPS_PROXY``, which is not correct.

This PR fixes this by using the correct variable ``https_proxy``.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,6 +19,11 @@ Fixed
 
   Contributed by @S-T-A-R-L-O-R-D
 
+
+* Fixed ``st2client/st2client/base.py`` file to use ``https_proxy``(not ``http_proxy``) to check HTTPS_PROXY environment variables.
+
+  Contributed by @wfgydbu
+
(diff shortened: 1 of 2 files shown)


rush-skills (approved):
👍🏼

rush-skills:
@wfgydbu Before we can merge this, we will need you to click on the link from the CLA bot and sign the StackStorm CLA

wfgydbu:
@rush-skills already signed, but still in checking. Maybe need some time to refresh.

cognifloyd (approved):

cognifloyd:
The CLA check uses the email address on each commit to decide who has signed.
@wfgydbu It looks like the email on your commit is attached to @xorder701. So, you'll need to fix the commit author, or sign the CLA with that account as well.

wfgydbu:
@cognifloyd Thanks for the explanation, it should be fine now.

PR #5679, opened 2022-07-19
Add purging of old tokens

amanda11:
Update housekeeping so that it can delete tokens that have expired.


CHANGELOG.rst
```diff
@@ -28,6 +28,9 @@ Added
 * Add ``ST2_USE_DEBUGGER`` env var as alternative to the ``--use-debugger`` cli flag. #5675
   Contributed by @cognifloyd
 
+* Added purging of old tokens. #56791
+  Contributed by Amanda McGuinness (@amanda11 intive)
```
rush-skills, line 32: Are we using the `(@username <orgname>)` anywhere else in the changelogs?

rush-skills (approved):
Looks good to me

CHANGELOG.rst
cognifloyd: I think there are a few examples. It's really up to the person who wrote the PR. Some people opt not to include a "Contributed by" bit in the changelog. Others need to credit their employer (intive in the case of Amanda).

PR #5686, opened 2022-07-25
Add pytest config section to pyproject.toml

cognifloyd:

So, this is the first of several PRs that improves our compatibility with pytest. It adds a pytest config section to `pyproject.toml`. I also included `-l` (display local vars on test failure) because that is extremely helpful when I encounter a test failure via pytest.


amanda11 (approved):
LGTM

rush-skills (approved):
LGTM

PR #5689, opened 2022-07-28
Adjust monkey_patching to work under pytest

cognifloyd:
### Background

This is another of several PRs that improves our compatibility with pytest. Others include: #5686, #5690, #5691

### Overview

Tests hang (with endless cycles between green threads that can't figure out what to do) when mongoengine gets imported before the monkey patching. So, this adjusts our monkey_patching code and the tests that use mongoengine to make our tests work better under pytest.

We need this because pytest can load test files in any order, or even load only a single test file. Since eventlet monkey_patching is very sensitive to when the monkey_patching happens, we need to make these adjustments or the tests won't work under pytest.


amanda11 (approved):
LGTM

armab (approved):
LGTM

PR #5690, opened 2022-07-29
Fix running st2client unit tests under pytest

cognifloyd:
### Background

This is another of several PRs that improves our compatibility with pytest. Others include: #5686, #5689, #5691

### Overview

Capturing console output uses `capsys` under pytest.
So modify our homegrown console redirection so that it uses capsys when running under pytest.


amanda11 (approved):
LGTM

armab (approved):
:+1:

PR #5691, opened 2022-07-29
Make DBTestCase compatible with pytest

cognifloyd:
### Background

This is another of several PRs that improves our compatibility with pytest. Others include: #5686, #5689, #5690

### Overview

DBTestCase was relying on some unittest/nosetest implmentation details that are not available when running under pytest.

The purpose of those adjustments was: Keep the DB after a test failure (do not drop it) to facilitate debugging.

However, pytest does not make results availble to fixtures. There are some complex workarounds, but there are issues with blindly copying the previous logic. For example, pytest can continue running tests after a failure or error unlike nosetest which we have configured to stop on failures. In order to safely skip dropping the DB, we need to make the DB available in a clean state after the test(s) using it. So, re-implementing this test feature requires something that renames a DB to preserve data for debugging, so that subsequent tests have a clean DB to work with.

I left a `TODO` note about this for whoever ends up trying to implement this. For now, this just avoids accessing attributes on an TestResult object that pytest does not provide.

### DIFF
--- a/st2tests/st2tests/base.py
+++ b/st2tests/st2tests/base.py
@@ -16,6 +16,8 @@
 from __future__ import absolute_import
 from __future__ import print_function
 
+from unittest.result import TestResult
+
 # NOTE: We need to perform monkeypatch before importing ssl module otherwise tests will fail.
 # See https://github.com/StackStorm/st2/pull/4834 for details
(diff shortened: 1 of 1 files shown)


amanda11 (approved):
LGTM

armab (approved):

Issue #5346, opened 2021-08-28
Usage of udatetime causing CI failures for Exchange packs


blag:
## SUMMARY

The weekend CI tests for all 100+ StackStorm Exchange packs failed again this weekend:

```python
#!/bin/bash
~/ci/.circle/test ; ~/ci/.circle/exit_on_py3_checks $?

Traceback (most recent call last):
  File "/home/circleci/ci/.circle/validate.py", line 21, in <module>
    from st2common.models.api.pack import PackAPI
  File "/tmp/st2/st2common/st2common/models/api/pack.py", line 24, in <module>
    from st2common import log as logging
  File "/tmp/st2/st2common/st2common/log.py", line 32, in <module>
    from st2common.logging.handlers import FormatNamedFileHandler
  File "/tmp/st2/st2common/st2common/logging/handlers.py", line 24, in <module>
    from st2common.util import date as date_utils
  File "/tmp/st2/st2common/st2common/util/date.py", line 24, in <module>
    import udatetime
  File "/home/circleci/virtualenv/lib/python3.6/site-packages/udatetime/__init__.py", line 20, in <module>
    from udatetime.rfc3339 import (
ImportError: /home/circleci/virtualenv/lib/python3.6/site-packages/udatetime/rfc3339.cpython-36m-x86_64-linux-gnu.so: undefined symbol: __exp_finite
Unable to retrieve pack name.
Original script exit code: 1

Exited with code exit status 1

CircleCI received exit code 1
```

I believe that this time it is related to [this Debian issue](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=954305) where the udatetime library calls out to a C library, which does not properly link to the `libm` library, as detailed in freach/udatetime#30 and fixed in freach/udatetime#31, which is a miniscule, easy fix yet has been unmerged for over a year.

The udatetime package, introduced to StackStorm [back in March](https://github.com/StackStorm/st2/pull/4846) in an attempt to optimize storage performance, has not been touched [since 2018](https://github.com/freach/udatetime/commit/75a07891426364f8bf0b44305b00bb1dd90534ae) and I believe that it's been abandoned. The author's [blog](https://aboutsimon.com/blog/) has also not seen an update since 2018, and he hasn't posted anything to [Twitter](https://twitter.com/aboutsimon) since 2018 either. His [Xing profile](https://www.xing.com/profile/Simon_Pirschel2) does indicate that he has changed jobs multiple times since 2018 though.

[...]

PR #5692, opened 2022-07-29
Move to ciso8601 for datetime

amanda11:
Problems using udatetime on python3.9 and it isn't supported.
It was proposed that ciso8601 is used instead.

Relates to https://github.com/StackStorm/st2/issues/5346

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -36,7 +36,7 @@ Added
 * Add ``ST2_USE_DEBUGGER`` env var as alternative to the ``--use-debugger`` cli flag. #5675
   Contributed by @cognifloyd
 
-* Added purging of old tokens. #56791
+* Added purging of old tokens. #5679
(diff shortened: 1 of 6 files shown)


amanda11:
I've brought up a python 3.8 RockyLinux ST2 using this change, and added some debugging to check that it didn't fall through to the catchall of using standard datetime.
I also did a quick check that I could install ciso8601 on Rocky 8 under python3.9 - and then perform a ciso8601.parse_datetime to check it also worked on python3.9.

I haven't done any benchmarking but the benchmarking of ciso8601 compared to udatetime compared to native python is available on other sites. It generally is much better than native python but not quite as quick as udatetime.

cognifloyd (approved):
Awesome

rush-skills (approved):

punkrokk:
Just to be clear - you mean ISO8601, correct?

cognifloyd:
ISO8601 is the standard ciso8601 is the prob library

PR #5699, opened 2022-08-03
Use python imports to identify fixtures (part 1)

cognifloyd:
### Background

### Overview

This PR does the following:
- adjust utils in `st2tests/st2tests/fixturesloader.py` so that we can use those utilities to identify every fixture.

This PR focuses on only 2 fixture packs: `dummy_pack_1` and `dummy_pack_10`.

I'm leaving a comment on one of the `fixturesloader.py` utils, as we could simplify the interface by using `inspect`. I'd like feedback on preferences about how much magic we're okay with here.

### Commits

- Ensure that pack fixture paths are absolute
- Import PATH/NAME for each fixture pack
- Minimize pack fixture boilerplate code
- Rename fixtureloader func to be more generic

### DIFF
--- a/st2common/tests/integration/test_register_content_script.py
+++ b/st2common/tests/integration/test_register_content_script.py
@@ -23,6 +23,7 @@ from st2tests.base import IntegrationTestCase
 from st2common.util.shell import run_command
 from st2tests import config as test_config
 from st2tests.fixturesloader import get_fixtures_packs_base_path
+from st2tests.fixtures.packs.dummy_pack_1.fixture import PACK_PATH as DUMMY_PACK_1_PATH
 
 
 BASE_DIR = os.path.dirname(os.path.abspath(__file__))
@@ -43,7 +44,7 @@ class ContentRegisterScriptTestCase(IntegrationTestCase):
         test_config.parse_args()
 
     def test_register_from_pack_success(self):
-        pack_dir = os.path.join(get_fixtures_packs_base_path(), "dummy_pack_1")
+        pack_dir = DUMMY_PACK_1_PATH
         runner_dirs = os.path.join(get_fixtures_packs_base_path(), "runners")
 
         opts = [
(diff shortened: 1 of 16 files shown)


st2common/tests/integration/test_register_content_script.py
```diff
@@ -43,7 +46,7 @@ def setUp(self):
         test_config.parse_args()
 
     def test_register_from_pack_success(self):
-        pack_dir = os.path.join(get_fixtures_packs_base_path(), "dummy_pack_1")
```
cognifloyd, line 46: In most tests, we actually calculate the fixture path every time it's used. By adding the boilerplate inside each fixture, that calculation only needs to be done in one place (DRY). Instead, we just need to use the imported `DUMMY_PACK_1_PATH` constant here.

st2common/tests/unit/services/test_policy.py
```diff
@@ -30,11 +30,10 @@
 from st2common.services import policies as policy_service
 
 import st2tests
+from st2tests.fixtures.generic.fixture import PACK_NAME as PACK
```
cognifloyd, line 33: Many tests already used a constant to define the PACK name, so all I had to do was import the constant from the fixture instead of hard-coding it in the test files.

cognifloyd (commented):
There is a lot of boilerplate in this PR - moving fixture path calculation from tests into the fixtures themselves - but it should be fairly easy to review as that boilerplate is very repetitive.

st2tests/st2tests/fixturesloader.py
```diff
@@ -158,15 +158,21 @@
...
 
 def get_resources_base_path():
-    return os.path.join(os.path.dirname(__file__), "resources")
+    return os.path.abspath(os.path.join(os.path.dirname(__file__), "resources"))
+
+
+def get_fixture_name_and_path(fixture_file):
+    pack_name = os.path.basename(os.path.dirname(fixture_file))
+    pack_path = os.path.join(get_fixtures_packs_base_path(), pack_name)
+    return pack_name, pack_path
```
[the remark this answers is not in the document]
rush-skills: I think passing file makes it more easier to understand what this piece of code is doing. `inspect` method here seems rather complicated.

rush-skills (approved):

nzlosh (approved):

PR #5702, opened 2022-08-10
Use python imports to identify fixtures (part 2/6)

cognifloyd:
### Background

### Overview

This PR builds on, and follows the patterns established in, #5699.

This PR does the following:

This PR updates all fixture packs, skipping any fixtures that need something slightly different (like `dummy_pack_7` and `dummy_pack_9`).


cognifloyd:
Stats for this PR:

**32** pack fixtures got 62 new files (a `fixture.py` in each; a `__init__.py` in all but 2 pack fixtures which already had it)

All 32 copies of `fixture.py` have the same 16 lines, 13 of which are our copyright/license header.

The remaining 38 files (=100-62) are test files that use the vars from the pack fixtures.
Lines changed in test files:
- `-298` lines
- `+238` lines _= 750 lines - 512 lines (512 = 16 copied lines * 32 `fixture.py` files)_

contrib/core/tests/test_action_sendmail.py
```diff
@@ -38,6 +39,10 @@
 MOCK_EXECUTION.id = "598dbf0c0640fd54bffc688b"
```
rush-skills, line 39: Not relevant to your changes, but do we need this hardcoded id here?

rush-skills (approved):
LGTM overall

contrib/core/tests/test_action_sendmail.py
cognifloyd: I don't see a problem with creating a mock using hard-coded values.

khushboobhatia01 (approved):

PR #5703, opened 2022-08-11
Use python imports to identify fixtures (part 3/6)

cognifloyd:
### Background

### Overview

This PR does the following:
- adjust util in `st2tests/st2tests/fixturesloader.py` so that we can use those utilities to identify pack fixtures in directories outside of `st2tests/st2tests/fixtures/packs`.

This PR focuses on only fixture packs that are in: `st2tests/st2tests/fixtures/packs_1/`, `st2tests/st2tests/fixtures/packs_invalid/` and `st2tests/st2tests/fixtures/`.

### DIFF
--- a/st2api/tests/unit/controllers/v1/test_executions.py
+++ b/st2api/tests/unit/controllers/v1/test_executions.py
@@ -333,12 +333,6 @@ LIVE_ACTION_DEFAULT_ENCRYPT_SECRET_PARAM = {
     "action": "starterpack.st2.dummy.default_encrypted_value_secret_param",
 }
 
-FIXTURES_PACK = "generic"
-TEST_FIXTURES = {
-    "runners": ["testrunner1.yaml"],
-    "actions": ["action1.yaml", "local.yaml"],
-}
-
 
 @mock.patch.object(
     content_utils, "get_pack_base_path", mock.MagicMock(return_value="/tmp/test")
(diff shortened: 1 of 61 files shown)


rush-skills (approved):
LGTM

st2api/tests/unit/controllers/v1/test_executions.py
```diff
@@ -333,12 +333,6 @@
     "action": "starterpack.st2.dummy.default_encrypted_value_secret_param",
 }
 
-FIXTURES_PACK = "generic"
-TEST_FIXTURES = {
-    "runners": ["testrunner1.yaml"],
-    "actions": ["action1.yaml", "local.yaml"],
-}
-
```
cognifloyd, line 341: These vars are unused. So I removed them.

st2tests/st2tests/fixturesloader.py
```diff
@@ -170,9 +170,9 @@ def get_resources_base_path():
 
 
 def get_fixture_name_and_path(fixture_file):
-    pack_name = os.path.basename(os.path.dirname(fixture_file))
-    pack_path = os.path.join(get_fixtures_packs_base_path(), pack_name)
-    return pack_name, pack_path
+    fixture_path = os.path.dirname(fixture_file)
+    fixture_name = os.path.basename(fixture_path)
+    return fixture_name, fixture_path
```
cognifloyd, line 175: I adjusted this method to make it more generic, so we can use it for pack fixtures that are not in `st2tests/st2tests/fixtures/packs`.

nzlosh (approved):

PR #5706, opened 2022-08-11
Use python imports to identify fixtures (part 6/6)

cognifloyd:
### Background

### Overview

This PR does the following:

This PR focuses on the last couple fixtures.

### Statistics

Changes Summary (mostly in `st2reactor` with a little in the `winrm_runner`):
* 1 line updated in changelog
* 5 copies of `fixture.py` (each has 13 lines of copyright/license header): `16*5=+80 lines`
    * some of these are actually `__init__.py` when that made more sense for a directory of fixtures
* `-11 +17` lines changed in test files to use the new fixture vars.
We may find need to address additional fixtures in the future, but this is the end of this series of PRs.

### DIFF
--- a/contrib/runners/winrm_runner/tests/unit/test_winrm_ps_script_runner.py
+++ b/contrib/runners/winrm_runner/tests/unit/test_winrm_ps_script_runner.py
@@ -22,7 +22,7 @@ from st2tests.base import RunnerTestCase
 from winrm_runner import winrm_ps_script_runner
 from winrm_runner.winrm_base import WinRmBaseRunner
 
-FIXTURES_PATH = os.path.join(os.path.dirname(__file__), "fixtures")
+from .fixtures import FIXTURES_PATH
 
 POWERSHELL_SCRIPT_PATH = os.path.join(FIXTURES_PATH, "TestScript.ps1")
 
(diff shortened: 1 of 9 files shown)


contrib/runners/winrm_runner/tests/unit/test_winrm_ps_script_runner.py
```diff
@@ -22,7 +22,7 @@
 from winrm_runner import winrm_ps_script_runner
 from winrm_runner.winrm_base import WinRmBaseRunner
 
-FIXTURES_PATH = os.path.join(os.path.dirname(__file__), "fixtures")
+from .fixtures import FIXTURES_PATH
```
rush-skills, line 25: is the `.` necessary here?

rush-skills (approved):
LGTM

contrib/runners/winrm_runner/tests/unit/test_winrm_ps_script_runner.py
cognifloyd: Alternatively we might be able to use this:
```suggestion
from tests.unit.fixtures import FIXTURES_PATH
```

Relative import was simpler. But I'm fine either way.

armab (approved):

PR #5707, opened 2022-08-12
make mongo tests safer to run in parallel

cognifloyd:
### Background

pants runs each test file separately in pytest. So, if two files try to modify the same database at the same time, we'll get strange results.

### Overview

So, modify the db name for each parallel test to ensure that they can run safely in parallel.

Once we integrate pants, we will configure (in pants.toml) `[pytest].execution_slot_var = "ST2TESTS_PARALLEL_SLOT"` to take advantage of the env var introduced here.
https://www.pantsbuild.org/docs/reference-pytest#section-execution-slot-var

### DIFF
--- a/st2common/tests/unit/test_db.py
+++ b/st2common/tests/unit/test_db.py
@@ -126,6 +126,7 @@ class DbConnectionTestCase(DbTestCase):
     def test_network_level_compression(self):
         disconnect()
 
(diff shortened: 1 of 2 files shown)


nzlosh (approved):
LGTM

winem (approved):
LGTM

PR #5708, opened 2022-08-12
Refactor `st2-generate-schemas` to be importable

cognifloyd:
### Background

I'm working towards introducing [`pants`](https://www.pantsbuild.org/docs). One of the first things we will use it for is linting / reformatting (it will run black, flake8, etc). Schema generation should happen more regularly, so we'll hook that up so pants runs it when it runs fmt and other tools.

When pants runs these tools, however, it (or rather PEX, which pants uses), it needs to import the python code. Our script, however, is a kebab-case-file, so python cannot import from it. (Eventually, this will not be a requirement. PEX has been updated to handle this, so pants just needs to have the new option plumbed in before we can use it. For now, though, pants can only run importable python files.)

### Overview

Currently schema generation is often missed, so the schemas often get out of date. We have some CI to catch this, but it doesn't error in every case, so there's one API schema that actually needs to be regenerated.

So, this PR does the following:
- refactor st2-generate-schemas so that logic is in an importable module
- run st2-generate-schemas to regenerate that outdated API schema

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -94,6 +94,9 @@ Changed
 * Refactor tests to use python imports to identify test fixtures. #5699 #5702 #5703 #5704 #5705 #5706
   Contributed by @cognifloyd
 
+* Refactor ``st2-generate-schemas`` so that logic is in an importable module. #5708
+  Contributed by @cognifloyd
+
 Removed
 ~~~~~~~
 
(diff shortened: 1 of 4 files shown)


armab (approved):

nzlosh (approved):
LGTM

Issue #3788, opened 2017-10-16
Autogenerate openapi.yaml.j2 from model code


nmaludy:
Currently `st2common/st2common/openapi.yaml.j2` needs to be updated manually when a change is made to the API. This makes the update process error prone and potentially the openapi schema may not match the actual API implementation (see: https://github.com/StackStorm/st2/pull/3522, https://github.com/StackStorm/st2/issues/3783).

I'm proposing to autogenerate the `openapi.yaml.j2` from source.

It would actually be great if the CLI, API and API documentation all were generated from a single "source of truth" to avoid errors and inconsistencies in the future.

Kami:
I agree, openapi entity definitions should be generated from actual API models.

We had a huge discussion about that in the past - I was always against current approach (copy and paste) since it's very error prone and hard to maintain (even now we are missing definitions for a bunch of models inside openapi file, some are incomplete / out of date, etc.).

Initially, some people were for dropping API models all together, but I think that's a bad idea because we would lose a lot of "static like typing" functionality which we get for free with linters, etc. once we have model class definitions. And in Python in general, working with dicts instead of models (class instances) with static / fixed list of attributes makes it very painful and error prone for the developer.

Also, I believe that's a duplicate of https://github.com/StackStorm/st2/issues/3575. Please confirm, so I can close and we can continue there if that's indeed the case.

nmaludy:
@Kami Yes this is a duplicate of #3575 except i'm also proposing auto-generating the CLI based on the API definitions. This could be considered an extra feature, however i think it would fall under the same "update in one place" umbrella.

emptywee:
I second that. Moreover, we need to allow end users override some defaults, e.g. number of items fetched by st2 client from the datastore and so forth.

Kami:
I would consider that as a separate issue / task since it's quite a big one. I also assume that with CLI you mean http client / API SDK.

But yeah, long term, low-level api client should be auto-generated for sure. I think "low-level" part is important. Auto-generated stuff usually only provides basic low level client functionality (simple 1 to 1 mapping to each API endpoint) and if you want to make it more language specific and user-friendly you need to add some wrappers around low-level auto-generated client.

nmaludy:
By CLI i was referring to the `st2` command that essentially is a command line wrapper for the API.

Kami:
[...]

PR #5709, opened 2022-08-12
Fix and reenable prance-based openapi spec validation

cognifloyd:
### Background

I'm working towards introducing [`pants`](https://www.pantsbuild.org/docs). One of the first things we will use it for is linting / reformatting (it will run black, flake8, etc). Validation of our openapi schema should happen more regularly, so we'll hook that up so pants runs it when it runs lint tools

While hooking that up, I noticed parts of the validation commented out. So, I searched through the git history to figure out when and why.
Basically, it's just tech debt that no one had time to address. So, this is a step in that direction.

### Overview

Fix and reenable prance-based openapi spec validation, but make our custom `x-api-model` validation optional as the spec is out-of-date.

This also fixes issues identified by the now-enabled spec validation.

Related: #3575 #3788

### DIFF
--- a/st2common/st2common/cmd/validate_api_spec.py
+++ b/st2common/st2common/cmd/validate_api_spec.py
@@ -45,6 +45,14 @@ cfg.CONF.register_cli_opt(
     )
 )
 
+# When disabled, only load the spec in prance to validate. Otherwise check for x-api-model as well.
+# validate-defs is disabled by default until these are resolved:
+#   https://github.com/StackStorm/st2/issues/3575
+#   https://github.com/StackStorm/st2/issues/3788
+cfg.CONF.register_cli_opt(
+    cfg.BoolOpt("validate-defs", short="-d", required=False, default=False)
+)
+
 cfg.CONF.register_cli_opt(
     cfg.BoolOpt("generate", short="-c", required=False, default=False)
 )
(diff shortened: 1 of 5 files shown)


armab (approved):

st2common/st2common/cmd/validate_api_spec.py
```diff
@@ -22,8 +22,8 @@
 from __future__ import absolute_import
 import os
 
+import prance
```
nzlosh, line 25: I understand the help `prance` offers with openapi, but I'm a little concerned with the activity of the project.  It claims to be tested against py36-py38 and hasn't received an update since 26 Nov 2021.  Perhaps this is a sign of its code maturity, but it would be unfortunate to fall into the same trap we did with `udatetime` where it became more of a hindrance than a help.  If you're of the opinion that the project is alive and active, I'm OK with adding it as a dependency.

nzlosh: OK, ignore me, it's not an addition, it's already a dependency, sorry for the noise.  But my concern still stands, we may need to look for an active project if `prance` has been abandoned.

nzlosh (approved):
LGTM

st2common/st2common/cmd/validate_api_spec.py
cognifloyd: Agreed. There are a bunch of things I'd like to improve around the openapi spec. Getting rid of `prance` sounds great to me. We're also using openapi/swagger 2, which means we can't use a lot of newer tooling that only supports openapi 3+. So much to do...

PR #5710, opened 2022-08-12
Fix and consolidate generation of st2.conf.sample

cognifloyd:
### Background

I'm working towards introducing [`pants`](https://www.pantsbuild.org/docs). One of the first things we will use it for is linting / reformatting (it will run black, flake8, etc). Updating sample conf must happen whenever the source python files change. Pants will know which file changes require regenerating the sample conf, so it is advantageous to let it handle that. So, we'll hook that up so pants runs it when it runs fmt and other tools.

### Overview

Pants will not use the Makefile to generate `st2.conf.sample`, it will use `tools/config_gen.py` directly. So, this moves a header comment from the Makefile into `tools/config_gen.py`, with the rest of the generation logic.

It also fixes a long standing issue where our `st2.conf.sample` file was actually invalid. `[sensorcontainer].partition_provider` should not have a python dict literal, but that's how it got serialized. It should have space separted `key:value` pairs for `partition_provider` config. So, this also adjust `tools/config_gen.py` to fix that issue.


tools/config_gen.py
```diff
@@ -77,6 +77,11 @@
     "webui": {"webui_base_url": "https://localhost"},
 }
 
+INIT_COMMENT = """
+# Sample config which contains all the available options which the corresponding descriptions
+# Note: This file is automatically generated using tools/config_gen.py - DO NOT UPDATE MANUALLY
+""".strip()
```
armab, line 83: that's definitely much better place for the config autogenerated note :+1:

armab (approved):
Thanks for the fix, good catch! :+1:

conf/st2.conf.sample
```diff
@@ -278,7 +278,7 @@ version = 4
 # location of the logging.conf file
 logging = /etc/st2/logging.sensorcontainer.conf
 # Provider of sensor node partition config.
-partition_provider = {'name': 'default'}
+partition_provider = name:default
```
nzlosh, line 281: Is this really the format for a valid dictionary variable?

cognifloyd: Yes. I've tripped on this several times. It's how oslo_config parses it. The k8s helm chart is the most recent place I tripped. The chart auto generates this setting based on how many sensor partitions a user requested. I've got tests in place there to make sure it uses this format because it does not work otherwise.

cognifloyd: It shows this format in the docs: https://docs.stackstorm.com/reference/sensor_partitioning.html
And here is where we generate that config in the helm chart: https://github.com/StackStorm/stackstorm-k8s/blob/master/templates/deployments.yaml#L1149-L1154

nzlosh: OK, so it's a oslo_config format.  I'd be curious to know how white space is handled for multiple kv pairs and if this would become a issue for corner cases, but that's not what this PR is about.  Thanks for the explanation :+1:

nzlosh (approved):
LGTM

PR #5713, opened 2022-08-19
Initialize pantsbuild

cognifloyd:
### Background

This first part of introducing [`pants`](https://www.pantsbuild.org/docs), as discussed in the TSC Meetings on [12 July 2022](https://github.com/StackStorm/community/issues/105) and [02 Aug 2022](https://github.com/StackStorm/community/issues/107). Pants has fine-grained per-file caching of results for lint, fmt (like black), test, etc. It also has lockfiles that work well for monorepos that have multiple python packages. With these lockfiles CI should not break when any of our dependencies or our transitive dependencies release new versions, because CI will continue to use the locked version until we explicitly relock with updates.

- teach TSC members about `pants` step-by-step.

So, I want as many TSC members as possible to review each of the pants PRs. Please ask questions so I can help explain what each part does.

### Overview of this PR

This walks through a basic pants installation using:
- https://www.pantsbuild.org/docs/manual-installation
    - record `[GLOBAL].pants_version` in the root `pants.toml` file
	- Download the `./pants` entrypoint script from https://static.pantsbuild.org/setup/pants
- https://www.pantsbuild.org/docs/initial-configuration (steps 1-4)
    1. add `[GLOBAL].pants_version`
	2. fill in `[source].root_patterns` to tell pants where to find our source code
	3. enable backends in `[GLOBAL].backend_packages`
	4. include pants-specific cache/build directories to `.gitignore`

#### `./pants` launch script

`./pants` is the primary entry point for accessing pants functionality. Updating this script does not happen very often, and it is designed to be checked into git.

> Pants has a launch script (called `./pants`) that handles downloading, bootstrapping, and upgrading Pants, which you need to save at the root of your repository.

https://www.pantsbuild.org/docs/installation

Examples of usage (`::` tells pants to select everything - you can also pass a path to a file or directory instead.):
- to run linters like flake8 and shellcheck we will run `./pants lint ::`
- to run formatters like black we will run `./pants fmt ::`
- to run tests via pytest (once that works) we will run `./pants test ::`

Note that this PR does not actually hook up linters/formatters/etc, so these will not do anything yet. That will happen in follow-up PRs.

#### `pants.toml` config file

Several important things get configured in the `pants.toml` config file.

> Pants configuration lives in a file called `pants.toml` in the root of the repo.

https://www.pantsbuild.org/docs/initial-configuration#create-pantstoml

This PR configures 3 settings:
- `[GLOBAL].pants_version`
- `[GLOBAL].backend_packages`
- `[source].root_patterns`

##### `[GLOBAL].pants_version`

As the name suggests, we use `[GLOBAL].pants_version` to switch to a newer version of pants. When someone updates, make sure to do one minor version at a time (ie to go from 2.9 to 2.11, you would do 2.9-2.10 and then 2.10-2.11) so that pants can handle migrating any config or metadata. Pants provides tooling to handle most of the required changes to config/metadata, and for everything else, they make sure to issue warnings when something is deprecated and how to update it.

##### `[GLOBAL].backend_packages`

The pants docs describe `[GLOBLA].backend_packages` as follows:

> Most Pants functionality is opt-in by adding the relevant backend to the [GLOBAL].backend_packages option in pants.toml. For example:

https://www.pantsbuild.org/docs/enabling-backends

So, this is how we `opt-in` to different funcitonality. In this PR, we tell pants that we have python and shell files in this repo. We'll add other backends (like flake8 and black) in follow-up PRs.

##### `[source].root_patterns`

Essentially, `[source].root_patterns` tells pants where to find python modules (eg to infer dependencies between them) and how to configure `PYTHONPATH`.

> Some project layouts use top-level folders for namespace purposes, but have the code live underneath. However, the code's imports will ignore these top-level folders, thanks to mechanisms like the `$PYTHONPATH` and the JVM classpath. Source roots are a generic equivalent of these concepts.
> Run `./pants roots` to see what Pants is using:

https://www.pantsbuild.org/docs/source-roots

With the config in this PR, running `./pants roots` returns:

```
.
contrib/chatops
contrib/core
contrib/default
contrib/examples
contrib/examples/lib
contrib/hello_st2
contrib/linux
contrib/packs
contrib/runners/action_chain_runner
contrib/runners/announcement_runner
contrib/runners/http_runner
contrib/runners/inquirer_runner
contrib/runners/local_runner
contrib/runners/noop_runner
contrib/runners/orquesta_runner
contrib/runners/python_runner
contrib/runners/remote_runner
contrib/runners/winrm_runner
pylint_plugins
scripts
st2actions
st2api
st2auth
st2client
st2common
st2common/benchmarks/micro
st2reactor
st2stream
st2tests
st2tests/testpacks/checks
st2tests/testpacks/errorcheck
tools
```


pants
```diff
@@ -0,0 +1,401 @@
+#!/usr/bin/env bash
+# Copyright 2020 Pants project contributors (see CONTRIBUTORS.md).
+# Licensed under the Apache License, Version 2.0 (see LICENSE).
```
cognifloyd, line 3: This file is copied from https://static.pantsbuild.org/setup/pants
I did not change it at all. Any future updates (which should be infrequent) will just replace this file with a downloaded update. We should not edit this file.

cognifloyd:
I'm not sure if this should be in the 3.8.0 milestone or not.

CHANGELOG.rst
```diff
@@ -47,6 +47,11 @@ Added
 * Added purging of old tokens. #5679
   Contributed by Amanda McGuinness (@amanda11 intive)
 
+* Begin introducing `pants <https://www.pantsbuild.org/docs>`_ to improve DX (Developer Experience)
+  working on StackStorm, improve our security posture, and improe CI reliability thanks in part
```
Eric-Arellano, line 51:
```suggestion
  working on StackStorm, improve our security posture, and improve CI reliability thanks in part
```

Eric-Arellano (commented):
Hey, Pants maintainer here. Sounds like y'all have talked to @benjyw already - but I want to echo that we're happy to help y'all with adoption and eager for any feedback :)

CHANGELOG.rst
cognifloyd: fixed. Thanks!

pants
winem: I guess you don't see there any risk by these "unattended" updates? I guess it's actually a good way to reduce technical debt and "force" us to stay up-to-date.

winem (approved):

nzlosh (approved):
LGTM

cognifloyd:
@StackStorm/tsc We have the minimum 2 reviews, but I want to get more reviews before I merge this:

> So, I want as many TSC members as possible to review each of the pants PRs. Please ask questions so I can help explain what each part does.

pants
cognifloyd: Updates to pants are not unattended. The version used is pinned in `pants.toml`:
https://github.com/StackStorm/st2/blob/92f08ebafc3882573227ca388f36a9b69e242f89/pants.toml#L2

So, we need a PR + reviews whenever we update the version of pants we're using.

If we need to update this pants launch script, then we also have to manually download + commit the update.

m4dcoder (approved):

pants
winem: Ah ok, that sounds good, thanks!

amanda11 (approved):

mickmcgrath13 (approved):

mickmcgrath13:
@cognifloyd This looks good to me (approved the PR).

One clarification question: how does caching work across builds/pipeline executions (or does it?)?  Is a file (or files) checked back in?  Is there a hosted service somewhere?

cognifloyd:
@mickmcgrath13 Pants caching has two parts locally: There is an lmdb file where pants caches results, and a file cache for downloaded wheels and virtualenvs that pants creates/manages. So far I've only tested preserving those file caches in github actions, but there is also a service that we might be able to use to make things even faster. I'll look into that caching-as-a-service option once the rest of my current pants work has been integrated.

Eric-Arellano:
> but there is also a service that we might be able to use to make things even faster. 

This is called "remote caching", which uses the open source Remote Execution API also used by Bazel. https://www.pantsbuild.org/docs/remote-caching. tl;dr is that Pants has very fine-grained caching, including caching individual runs for each individual test file. If none of the dependencies for that specific file have changed, we can use the cache. 

Normally, that cache is kept locally, but remote caching lets you store it in the cloud. Unlike storing caches in GitHub Actions, where you pessimistically download the whole cache at the start of a CI job, remote caching only pulls down what is exactly needed for the current run.

armab (approved):
Great work on organizing the review here from as many as possible TSC members :+1: 

One question as we add the `./pants`, are there any documentation or development guidelines that should be updated at this stage?

cognifloyd:
> One question as we add the `./pants`, are there any documentation or development guidelines that should be updated at this stage?

At least these pages will need to be updated:
- https://github.com/StackStorm/st2/blob/master/CONTRIBUTING.rst
- [docs site](https://docs.stackstorm.com/development/index.html) | https://github.com/StackStorm/st2docs/blob/master/docs/source/development/index.rst
- orphan page | https://github.com/StackStorm/st2docs/blob/master/docs/source/development/sources.rst
- any other page under: https://github.com/StackStorm/st2docs/tree/master/docs/source/development

However, I don't think we should update those yet because we'll need several more PRs before we can use `pants`. But we could probably:
- create a WIP PR on st2docs
    - slowly update the docs PR throughout the process of integrating pantsbuild
    - each PR to the StackStorm/st2 can include a small update to the WIP docs PR
    - merge the docs PR once we are ready for people to use it more generally

I imagine we'll have CI using `pants` before the docs are ready for everyone to use it more generally.

cognifloyd:
Here is a stub PR for the docs repo: https://github.com/StackStorm/st2docs/pull/1143
We can update the development docs in the `pantsbuild` branch as we merge PRs here.

PR #5726, opened 2022-09-09
Adjust .gitignore to work better with pantsbuild

cognifloyd:
### Background

Other pants PRs include:
- https://github.com/StackStorm/st2/pull/5727

### Overview of this PR

This PR makes two adjustments to our `.gitignore` file.
1. Remove `build` from .gitignore
2. Add more vim temp file patterns

pants requires `BUILD` files to describe metadata about our code (this directory has python and it has these dependencies that can't be inferred from the imports, etc). `.gitignore` is not case sensitive, so `build` prevents adding `BUILD` files. This is required to use pants.

For more about BUILD files, see: https://www.pantsbuild.org/docs/targets#build-files

The vim temp file patterns is not required for pants but it makes it nicer to work with it. If files change while pants is running a process, it can pause and re-run. When I use `vi` to look at a file while pants is running tests, or formatters, or whatever, then the tmp files keep triggering an inspection of the filesystem (pants needs to make sure it is including all required files in the given operation). But, pants ignores all files in `.gitignore`, so, adding the `vim` temp files there prevents pants from getting false-positive filesystem changes. We might end up adding other temp files for people that use other editors to make their experience with pants nicer as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,5 +67,7 @@ benchmark_histograms/
 # Editor Saves
 *~
 \#*\#
-[._]*.sw[a-p]
-[._]sw[a-p]
+[._]*.sw[a-px]
+[._]sw[a-px]
+[._]*.sw[a-p]x
+[._]sw[a-p]x
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,6 @@
 *.egg
 *.egg-info
 dist
-build
 .venv
 eggs
 parts
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -53,7 +53,7 @@ Added
 
 * Begin introducing `pants <https://www.pantsbuild.org/docs>`_ to improve DX (Developer Experience)
   working on StackStorm, improve our security posture, and improve CI reliability thanks in part
-  to pants' use of PEX lockfiles. This is not a user-facing addition. #5713 #5724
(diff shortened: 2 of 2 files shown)


armab (approved):

.gitignore
```diff
@@ -10,7 +10,6 @@
 *.egg
 *.egg-info
 dist
-build
```
Eric-Arellano, line 13: FYI you can name the files `BUILD.pants`! Many people prefer that for clarity. It allows you to avoid making this change also.

cognifloyd: Interesting thought.
We don't have anything that creates `build` files or directories, so keeping it in `.gitignore` is kind of pointless.

But if we did change from `BUILD` to `BUILD.pants`, we would need this config in pants.toml:
```
[tailor]
build_file_name = "BUILD.pants"
```
https://www.pantsbuild.org/docs/reference-tailor#build_file_name

I thought we would also need to adjust `[GLOBAL].build_patterns`, but the default is `["BUILD", "BUILD.*"]`, so pants already recognizes `BUILD.pants`:
https://www.pantsbuild.org/docs/reference-global#build_patterns

I've done everything with `BUILD` files in my PoC. It'll be less work for me to keep using `BUILD` because that's less I have to change when I cherry-pick commits. If people would prefer `BUILD.pants` then that is possible.

amanda11 (approved):
Lgtm

PR #5730, opened 2022-09-16
Add python3.9 to GHA

amanda11:
Investigate if any changes needed for unit or integration tests to work with python 3.9.


amanda11:
The lint fails seem to do with astroid, there is mention of fixes in 2.6 for astroid for python3.9 but st2 fixes to 2.5.6. 

However if I use latest astroid/pylint combinations on python3.8 then the astroid parse on the pylint_plugins copy fails to find the property added after copy. (astroid 2.12.9).

Rolled back down to 2.6.6 astroid and still failed (python 3.8 as well).

Rolled python 3.8 back down to astroid 2.5.6 and then it passed again on python3.8 but failed on python3.9.

nzlosh:
I hit the similar issues with py3.10 and astroid 2.15.5.
```
============================================================================== short test summary info ==============================================================================
FAILED pylint_plugins/api_models_test.py::test_copied_schema - AssertionError: assert 'default_files' in {'__module__': [<Const.str l.7 at 0x7f118ae0fb20>], '__qualname__': [<Const.str l.7 at 0x7f118ae0f4c0>], 'action': [<AssignName.action...
FAILED pylint_plugins/api_models_test.py::test_copied_imported_schema - AssertionError: assert 'default_files' in {'__module__': [<Const.str l.5 at 0x7f118ae5feb0>], '__qualname__': [<Const.str l.5 at 0x7f118ae4ffa0>], 'description': [<AssignName.d...
FAILED pylint_plugins/api_models_test.py::test_indirect_copied_schema - AttributeError: 'str' object has no attribute 'value'
FAILED pylint_plugins/api_models_test.py::TestTypeChecker::test_finds_no_member_on_api_model_when_property_not_in_schema - AttributeError: module 'pylint.testutils' has no attribute 'Message'
============================================================================ 4 failed, 5 passed in 0.51s ============================================================================
```

copart-jafloyd:
Bingo: py3.9 changed the AST, dropping an intermediate `Slice` type from something like `schema["properties"]`, so I just pushed a commit that fixes that.

Oddly, the pants-based run did not fail with 3.9 even though it should have. I suspect that pants might be using the wrong python for our tests. I will figure out + address that when I update pants in another PR.

cognifloyd:
Bingo: py3.9 changed the AST, dropping an intermediate `Slice` type from something like `schema["properties"]`, so I just pushed a commit that fixes that.

Oddly, the pants-based run did not fail with 3.9 even though it should have. I suspect that pants might be using the wrong python for our tests. I will figure out + address that when I update pants in another PR.

nzlosh:
@cognifloyd Great find :100:

cognifloyd (approved):

nzlosh (approved):

armab (commented):
Thanks a lot everyone!
let’s merge it

PR #5737, opened 2022-09-20
Explicitly disable pantsbuild opt-in anonymous telemetry

cognifloyd:
### Background

This is another part of introducing [`pants`](https://www.pantsbuild.org/docs), as discussed in the TSC Meetings on [12 July 2022](https://github.com/StackStorm/community/issues/105), [02 Aug 2022](https://github.com/StackStorm/community/issues/107) and [06 Sept 2022](https://github.com/StackStorm/community/issues/108). Pants has fine-grained per-file caching of results for lint, fmt (like black), test, etc. It also has lockfiles that work well for monorepos that have multiple python packages. With these lockfiles CI should not break when any of our dependencies or our transitive dependencies release new versions, because CI will continue to use the locked version until we explicitly relock with updates.

To keep PRs as manageable/reviewable as possible, introducing pants will take a series of PRs. I do not know yet how many PRs; I will break this up into logical steps with these goals:
- introduce `pants` to the st2 repo, and
- teach some (but not all) TSC members about `pants` step-by-step.

Other pants PRs include:
- https://github.com/StackStorm/st2/pull/5713
- https://github.com/StackStorm/st2/pull/5724
- https://github.com/StackStorm/st2/pull/5725
- https://github.com/StackStorm/st2/pull/5726
- https://github.com/StackStorm/st2/pull/5732
- https://github.com/StackStorm/st2/pull/5733
- https://github.com/StackStorm/st2/pull/5738

### Overview of this PR

This PR is independent of other pantsbuild PRs.

This PR configures pants' anonymous telemetry so that it is explicitly disabled (it is already opt-in by default, but we explicitly disable it). It is also configured so that individual developers can enable it if they wish, but--again--it requires an explicit opt-in.

Documentation about anonymous telemetry is available here:
https://www.pantsbuild.org/docs/anonymous-telemetry

We need to either set the `repo_id` or explicitly disable telemetry to get rid of a non-fatal error pants prints on startup (shown below). I opted to set both so that I can still provide telemetry on my machine, but make the default state (opt-in) explicit to avoid any concerns in our community about telemetry.

```
$ ./pants --version
Bootstrapping Pants using /opt/local/bin/python3.9
Installing pantsbuild.pants==2.14.0rc1 into a virtual environment at /.../.cache/pants/setup/bootstrap-.../2.14.0rc1_py39
New virtual environment successfully created at /.../.cache/pants/setup/bootstrap-.../2.14.0rc1_py39.
13:56:52.14 [INFO] Initializing scheduler...
[...]

PR #5738, opened 2022-09-20
Create pantsbuild `BUILD` files with `./pants tailor ::`

cognifloyd:
### Background

Other pants PRs include:
- https://github.com/StackStorm/st2/pull/5733
- https://github.com/StackStorm/st2/pull/5737

### Overview of this PR

This PR does 3 things:
- Run `./pants tailor ::` ([step 5 of the initial config docs](https://www.pantsbuild.org/docs/initial-configuration#generate-build-files))
- Enable the workflow added in #5732
- Fix an edge case with our `BUILD` metadata to work around our use of git submodules (the `test_content_version` fixture pack)

_I will describe the generated BUILD file contents after I have briefly discussed each of them (the 3 things this PR does)._

#### `./pants tailor ::`

All but one of the `BUILD` files in this PR were created by running `./pants tailor ::` which is step 5 of the [initial config docs](https://www.pantsbuild.org/docs/initial-configuration#generate-build-files).

As described in #5732, BUILD files are an important part of helping pants understand our codebase. As [the docs](https://www.pantsbuild.org/docs/initial-configuration#generate-build-files) say:

> [BUILD](https://www.pantsbuild.org/docs/targets) files provide metadata about your code (the timeout of a test, any dependencies which cannot be inferred, etc). BUILD files are typically located in the same directory as the code they describe. Unlike many other systems, Pants BUILD files are usually very succinct, as most metadata is either inferred from static analysis, assumed from sensible defaults, or generated for you.

In follow-up PRs, we will customize the BUILD metadata further.

#### GHA workflow

This PR also enables the GHA workflow added in #5732 (it also fixes a typo in the workflow. oops). This workflow will ensure that we have all the BUILD files that pants needs. This is very quick, and serves as an indicator on PRs that people might need to run `./pants tailor ::`.

#### git submodule: `test_content_version` fixture pack

`./pants tailor` tried to add BUILD files in the git submodule. But, we cannot guarantee that git submodule is checked out when we run `./pants`. So, I added some targets in the parent directory's BUILD file that covers everything in the git submodule. This way, even if the submodule is not checked out, the complete set of metadata is available for pants.

_I wrote a plugin for pants that makes the developer UX even better. It checks to make sure the git submodule is checked out, and if not, it provides an error message with instructions on how to fix it. I will submit that in a later PR._

### `BUILD` file metadata

BUILD files contain metadata that guide pants through our source code. That metadata consists of "targets" with "fields".

From the docs: https://www.pantsbuild.org/docs/targets

> _Targets_ are an _addressable_ set of metadata describing your code.

> Each target type has different _fields_, or individual metadata values.

One of the nice things about `BUILD` files, is that they are essentially just python files. In fact, `./pants update-build-files ::` actually uses `black` to reformat them!

So, when looking at BUILD files, defining a target looks like a function call, and the fields are args and/or kwargs in that function call.

There are some restrictions on what python we can use in BUILD files. For example: there are no imports (Pants actually adds all of the required symbols based on which `backends` are enabled in `pants.toml`). When we need to add custom targets, we can do that with macros and pants plugins, both of which are also written in python. I have written both macros and pants plugins to add in follow-up PRs.

#### Targets

These are the targets that `./pants tailor ::` added in this PR (linked to the relevant docs page):

- [python_requirements](https://www.pantsbuild.org/docs/reference-python_requirements) (see also: [python_requirement](https://www.pantsbuild.org/docs/reference-python_requirement))
- [python_sources](https://www.pantsbuild.org/docs/reference-python_sources) (see also: [python_source](https://www.pantsbuild.org/docs/reference-python_source))
- [python_test_utils](https://www.pantsbuild.org/docs/reference-python_test_utils)
- [python_tests](https://www.pantsbuild.org/docs/reference-python_tests) (see also: [python_test](https://www.pantsbuild.org/docs/reference-python_tests))
- [shell_sources](https://www.pantsbuild.org/docs/reference-shell_sources) (see also: [shell_source](https://www.pantsbuild.org/docs/reference-shell_sources))

I also added some of these:
- [resources](https://www.pantsbuild.org/docs/reference-resources) (see also: [shell_source](https://www.pantsbuild.org/docs/reference-resource))

Tailor cannot add `resources` or [`files`](https://www.pantsbuild.org/docs/reference-files) targets, so those will be added manually in follow-up PRs.

### DIFF
--- a/.github/workflows/pants.yaml
+++ b/.github/workflows/pants.yaml
@@ -2,23 +2,21 @@
 name: Validate Pants Metadata
 
 on:
-  # temporarily only allow manual runs until we have BUILD files and lockfiles
-  workflow_dispatch:
-  #push:
-  #  branches:
-  #    # only on merges to master branch
-  #    - master
-  #    # and version branches, which only include minor versions (eg: v3.4)
-  #    - v[0-9]+.[0-9]+
-  #  tags:
-  #    # also version tags, which include bugfix releases (eg: v3.4.0)
-  #    - v[0-9]+.[0-9]+.[0-9]+
-  #pull_request:
-  #  type: [opened, reopened, edited]
-  #  branches:
-  #    # Only for PRs targeting those branches
-  #    - master
-  #    - v[0-9]+.[0-9]+
+  push:
+    branches:
+      # only on merges to master branch
+      - master
+      # and version branches, which only include minor versions (eg: v3.4)
+      - v[0-9]+.[0-9]+
+    tags:
+      # also version tags, which include bugfix releases (eg: v3.4.0)
+      - v[0-9]+.[0-9]+.[0-9]+
+  pull_request:
+    type: [opened, reopened, edited]
+    branches:
+      # Only for PRs targeting those branches
+      - master
+      - v[0-9]+.[0-9]+
 
 jobs:
   pants-tailor:
@@ -42,11 +40,12 @@ jobs:
         #   Otherwise, we may need an additional workflow or job to delete old caches
         #   if they are not expiring fast enough, and we hit the GHA 10GB per repo max.
         with:
+          base-branch: master
           # To ignore a bad cache, bump the cache* integer.
           gha-cache-key: cache0-BUILD
           # This hash should include all of our lockfiles so that the pip/pex caches
           # get invalidated on any transitive dependency update.
-          named-caches-hash: ${{ hashFiles('requirements.txt' }}
+          named-caches-hash: ${{ hashFiles('requirements.txt') }}
           # enable the optional lmdb_store cache since we're not using remote caching.
           cache-lmdb-store: 'true'
 
--- a/BUILD
+++ b/BUILD
@@ -1,0 +1,7 @@
+python_requirements(
+    name="root",
+)
+
+python_test_utils(
+    name="test_utils0",
+)
--- a/st2api/tests/unit/controllers/v1/test_packs_views.py
+++ b/st2api/tests/unit/controllers/v1/test_packs_views.py
@@ -133,7 +133,7 @@ class PacksViewsControllerTestCase(FunctionalTest):
         # Ref is not equal to the name, controller should still work
         resp = self.app.get("/v1/packs/views/files/dummy_pack_16")
         self.assertEqual(resp.status_int, http_client.OK)
-        self.assertEqual(len(resp.json), 3)
+        self.assertEqual(len(resp.json), 4)
         self.assertIn("pack.yaml", [f["file_path"] for f in resp.json])
 
         resp = self.app.get("/v1/packs/views/file/dummy_pack_16/pack.yaml")
--- a/st2tests/st2tests/fixtures/packs/BUILD
+++ b/st2tests/st2tests/fixtures/packs/BUILD
@@ -1,0 +1,30 @@
+# The files in test_content_version* targets are in ./test_content_version
+# which is a git submodule.
+# The test_content_version* targets are dependencies of ./test_content_version_fixture
+
+resources(
+    name="test_content_version_metadata",
+    sources=[
+        "test_content_version/pack.yaml",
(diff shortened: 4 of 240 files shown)


.github/workflows/pants.yaml
```diff
@@ -42,11 +40,12 @@ jobs:
         #   Otherwise, we may need an additional workflow or job to delete old caches
         #   if they are not expiring fast enough, and we hit the GHA 10GB per repo max.
         with:
+          base-branch: master
```
cognifloyd, line 43: The action defaults to using `main` as the base-branch (which I forgot, oops).

.github/workflows/pants.yaml
```diff
@@ -42,11 +40,12 @@ jobs:
...
           # To ignore a bad cache, bump the cache* integer.
           gha-cache-key: cache0-BUILD
           # This hash should include all of our lockfiles so that the pip/pex caches
           # get invalidated on any transitive dependency update.
-          named-caches-hash: ${{ hashFiles('requirements.txt' }}
+          named-caches-hash: ${{ hashFiles('requirements.txt') }}
```
cognifloyd, line 48: Doh! Typo!

BUILD
```diff
@@ -0,0 +1,7 @@
+python_requirements(
+    name="root",
+)
+
+python_test_utils(
+    name="test_utils0",
+)
```
cognifloyd, line 7: This is for `conftest.py`. I will rename the targets in this BUILD file (and others) in a follow-up PR.

st2tests/st2tests/fixtures/packs/BUILD
```diff
@@ -0,0 +1,31 @@
...
+python_sources(
+    name="test_content_version",
+    dependencies=[
+        ":test_content_version_metadata",
+        ":test_content_version_shell",
+    ],
+    sources=[
+        "test_content_version/**/*.py",
+    ],
+)
```
cognifloyd, line 31: These are the targets that cover the `test_content_version` git submodule.

st2tests/st2tests/fixtures/packs/test_content_version_fixture/BUILD
```diff
@@ -0,0 +1,3 @@
+python_sources(
+    dependencies=["st2tests/st2tests/fixtures/packs:test_content_version"],
+)
```
cognifloyd, line 3: This manually declared dependency tells pants that `fixture.py` in this directory actually needs the files from `test_content_version` (the git submodule).

st2api/tests/unit/controllers/v1/test_packs_views.py
```diff
@@ -133,7 +133,7 @@ def test_get_pack_files_and_pack_file_ref_doesnt_equal_pack_name(self):
         # Ref is not equal to the name, controller should still work
         resp = self.app.get("/v1/packs/views/files/dummy_pack_16")
         self.assertEqual(resp.status_int, http_client.OK)
-        self.assertEqual(len(resp.json), 3)
+        self.assertEqual(len(resp.json), 4)
```
cognifloyd, line 136: I had to bump the file count in the pack because this PR added a `BUILD` file in the pack.

Eric-Arellano (approved):
FYI from our docs about why we recommend doing things this way with one BUILD file per directory:

> We've found that it usually scales much better to use a single BUILD file per directory. Even if you start with using the defaults for everything, projects usually need to change some metadata over time, like adding a timeout to a test file or adding dependencies on resources.
>
> It's useful for metadata to be as fine-grained as feasible, such as by using the overrides field to only change the files you need to. Fine-grained metadata is key to having smaller cache keys (resulting in more cache hits), and allows you to more accurately reflect the status of your project. We have found that using one BUILD file per directory encourages fine-grained metadata by defining the metadata adjacent to where the code lives.

winem (approved):
Another great job!

PR #5744, opened 2022-09-23
add secret to runner overridable parameters list

jonathangull:
Add secret to the list of overridable parameters to avoid the error -  attribute “secret” for the runner parameter “cmd” in action “core.local” cannot be overridden.

Why is this required -

We want to create action core.winrm_ps_cmd_secure which is same as core.winrm_ps_cmd where the input parameter cmd is hidden. We have a use case where we need to supply credentials inside the cmd section so it's not visible in stack storm UI after execution.

### DIFF
--- a/st2common/st2common/util/schema/__init__.py
+++ b/st2common/st2common/util/schema/__init__.py
@@ -76,6 +76,7 @@ RUNNER_PARAM_OVERRIDABLE_ATTRS = [
     "enum",
     "immutable",
     "required",
+    "secret",
 ]
 
 


cognifloyd:
I think this makes sense, but I would like to see something a little more intelligent for the `secret` attr.
For example, if a runner sets `password` to `secret: true`, then the action shouldn't be able to set `secret: false`. But, adding `secret: true` to additional parameters seems like a completely valid thing to do.

Would you look at how we can prevent people from setting `secret: false` in an action when the runner schema specifies `secret: true`?

jonathangull:
@cognifloyd sorry I do not know how to do this, can you elaborate or give me some guidance on how to do it.

PR #5749, opened 2022-10-03
Encryption and decryption of secrets at rest

neeraj-orch:


guzzijones:
Don't we already encrypt secrets in the datastore if you pass `--encrypt`

fdrab:
If I'm reading the commit correctly, this would encrypt the values you pass into fields marked as secret, not the datastore --encrypt (those already are encrypted)

PR #5751, opened 2022-10-05
Use shellcheck (via pantsbuild)

cognifloyd:
### Background

This is another part of introducing [`pants`](https://www.pantsbuild.org/docs), as discussed in the TSC Meetings on [12 July 2022](https://github.com/StackStorm/community/issues/105), [02 Aug 2022](https://github.com/StackStorm/community/issues/107), [06 Sept 2022](https://github.com/StackStorm/community/issues/108), and [04 Oct 2022](https://github.com/StackStorm/community/issues/111). Pants has fine-grained per-file caching of results for lint, fmt (like black), test, etc. It also has lockfiles that work well for monorepos that have multiple python packages. With these lockfiles CI should not break when any of our dependencies or our transitive dependencies release new versions, because CI will continue to use the locked version until we explicitly relock with updates.

To keep PRs as manageable/reviewable as possible, introducing pants will take a series of PRs. I do not know yet how many PRs; I will break this up into logical steps with these goals:
- introduce `pants` to the st2 repo, and
- teach some (but not all) TSC members about `pants` step-by-step.

Other pants PRs include:
- https://github.com/StackStorm/st2/pull/5713
- https://github.com/StackStorm/st2/pull/5724
- https://github.com/StackStorm/st2/pull/5725
- https://github.com/StackStorm/st2/pull/5726
- https://github.com/StackStorm/st2/pull/5732
- https://github.com/StackStorm/st2/pull/5733
- https://github.com/StackStorm/st2/pull/5737
- https://github.com/StackStorm/st2/pull/5738
- https://github.com/StackStorm/st2/pull/5758

### Overview of this PR

It is helpful to review each commit in this PR individually.

- 6b2f9932dea938c71385f448f08086b3cb3bdae1 Rename shell targets to be a bit more consistent.
- d442fd5ae66875411487ac7a6e0cd54440f8cebe Add/adjust targets for executable files/scripts
    - In #5738, tailor added a lot of metadata, but it could not identify some of our extension-less scripts as shell or python.
    - So, we add metadata for all of those scripts (basically anything in the `bin/` directories).
    - This way, pants can run shellcheck on all of our shell files, not just `*.sh`.
- 854873c558f6f5ff39a6aa49731ba2a3e6b3b58b Add shellcheck (_skipping anything that does not pass_)
    - Enable the shellcheck backend in pants (pants.toml: `[GLOBAL].backend_packages`).
        - We enabled language backends in #5725 which included the `shell_sources` and `shell_source` targets.
[...]

PR #5758, opened 2022-10-05
Add GHA workflow to lint with pantsbuild

cognifloyd:
### Background

This is another part of introducing [`pants`](https://www.pantsbuild.org/docs), as discussed in the TSC Meetings on [12 July 2022](https://github.com/StackStorm/community/issues/105), [02 Aug 2022](https://github.com/StackStorm/community/issues/107), [06 Sept 2022](https://github.com/StackStorm/community/issues/108), and [04 Oct 2022](https://github.com/StackStorm/community/issues/111). Pants has fine-grained per-file caching of results for lint, fmt (like black), test, etc. It also has lockfiles that work well for monorepos that have multiple python packages. With these lockfiles CI should not break when any of our dependencies or our transitive dependencies release new versions, because CI will continue to use the locked version until we explicitly relock with updates.

Other pants PRs include:
- https://github.com/StackStorm/st2/pull/5733
- https://github.com/StackStorm/st2/pull/5737
- https://github.com/StackStorm/st2/pull/5738
- https://github.com/StackStorm/st2/pull/5751

### Overview of this PR

We begin to add linters in #5751. This adds a GHA workflow to automatically run the linters. This workflow copies a few pants tasks from the pants validation workflow added in #5732 and the apt tasks from our current CI workflow.

Until one of the linter backends is registered, `./pants lint ::` will just return with success because there is nothing to do.

When we add python linters (black, flake8, etc) we will need to expand this workflow to handle a matrix of the various python versions. For now, this is sufficient for running shellcheck (in #5751) and a few other linters.

- [`./pants lint` goal](https://www.pantsbuild.org/v2.14/docs/reference-lint)

### Things you can do with pantsbuild

You can run `./pants lint ::` but it doesn't do anything yet. So, output looks like this:

```
$ ./pants lint ::
13:00:22.28 [INFO] Initializing scheduler...
13:00:22.57 [INFO] Scheduler initialized.

$ echo $?
0
$ ./pants lint ::

$ echo $?
0
```

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -1,0 +1,88 @@
+---
+# This Lint workflow uses pants
+name: Lint
+
+on:
+  push:
+    branches:
+      # only on merges to master branch
+      - master
+      # and version branches, which only include minor versions (eg: v3.4)
+      - v[0-9]+.[0-9]+
+    tags:
+      # also version tags, which include bugfix releases (eg: v3.4.0)
+      - v[0-9]+.[0-9]+.[0-9]+
+  pull_request:
+    type: [opened, reopened, edited]
+    branches:
+      # Only for PRs targeting those branches
+      - master
+      - v[0-9]+.[0-9]+
+  schedule:
+    # run every night at midnight
+    - cron:  '0 0 * * *'
+
+jobs:
+  # Lint checks which don't depend on any service containes, etc. to be running.
+  lint-checks:
+    name: 'Lint Checks (pants runs: shellcheck)'
+    runs-on: ubuntu-latest
+
+    env:
+      COLUMNS: '120'
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          # a test uses a submodule, and pants needs access to it to calculate deps.
+          submodules: 'true'
+
+      #- name: Cache APT Dependencies
+      #  id: cache-apt-deps
+      #  uses: actions/cache@v2
+      #  with:
+      #    path: |
+      #      ~/apt_cache
+      #    key: ${{ runner.os }}-apt-v7-${{ hashFiles('scripts/github/apt-packages.txt') }}
+      #    restore-keys: |
+      #      ${{ runner.os }}-apt-v7-
+      - name: Install APT Depedencies
+        env:
+          CACHE_HIT: 'false' # cache doesn't work
+          #CACHE_HIT: ${{steps.cache-apt-deps.outputs.cache-hit}}
+        run: |
+          # install dev dependencies for Python YAML and LDAP packages
+          # https://github.com/StackStorm/st2-auth-ldap
+          ./scripts/github/install-apt-packages-use-cache.sh
+
+      - name: Initialize Pants and its GHA caches
+        uses: pantsbuild/actions/init-pants@c0ce05ee4ba288bb2a729a2b77294e9cb6ab66f7
+        # This action adds an env var to make pants use both pants.ci.toml & pants.toml.
+        # This action also creates 3 GHA caches (1 is optional).
+        # - `pants-setup` has the bootsrapped pants install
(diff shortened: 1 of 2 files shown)


.github/workflows/lint.yaml
```diff
@@ -0,0 +1,88 @@
...
+      - v[0-9]+.[0-9]+.[0-9]+
+  pull_request:
+    type: [opened, reopened, edited]
+    branches:
+      # Only for PRs targeting those branches
+      - master
+      - v[0-9]+.[0-9]+
+  schedule:
+    # run every night at midnight
+    - cron:  '0 0 * * *'
```
Eric-Arellano, line 23: Why this?

Eric-Arellano (approved):

.github/workflows/lint.yaml
cognifloyd: I copied that from our other workflows which we run every day. We did that in part to catch issues with transitive deps, which will not be a significant issue once we've got the lockfiles in. So, I suppose it is unlikely for linters to fail from one day to another, so maybe we don't have to be quite so aggressive here. 🤔

What do others think? Drop this? Comment it out for now?

Eric-Arellano: > which will not be a significant issue once we've got the lockfiles in.

Note that you already are using "tool lockfiles". Pants ships them by default. So your installation of tools like Flake8 and Pylint will be deterministic. It is only installs of your own user code that need to manually set up lockfiles, since we can't ship them without knowing before what your code is.

So running this is cron is likely a waste of compute resources (and trees)

rush-skills (approved):
Lgtm overall, added some comments below

amanda11 (approved):
LGTM - once the other queries on the PR have been addressed.

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,88 @@
...
+        env:
+          CACHE_HIT: 'false' # cache doesn't work
+          #CACHE_HIT: ${{steps.cache-apt-deps.outputs.cache-hit}}
+        run: |
+          # install dev dependencies for Python YAML and LDAP packages
+          # https://github.com/StackStorm/st2-auth-ldap
+          ./scripts/github/install-apt-packages-use-cache.sh
+
+      - name: Initialize Pants and its GHA caches
+        uses: pantsbuild/actions/init-pants@c0ce05ee4ba288bb2a729a2b77294e9cb6ab66f7
```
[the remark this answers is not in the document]
cognifloyd: This repo does not yet have tags we can point to. If there are changes, we might need to adjust how we use the action, app pinning to a particular version seemed the best way to go.

When we choose another approach (eg when there are tags available), then we'll need to change the pants BUILD validation workflow add well:
https://github.com/StackStorm/st2/blob/989e4ec5eaf3f458feb2b49c31715ec672a07689/.github/workflows/pants.yaml#L33

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,88 @@
...
+        #   If we ever use a remote cache, then we can drop this.
+        #   Otherwise, we may need an additional workflow or job to delete old caches
+        #   if they are not expiring fast enough, and we hit the GHA 10GB per repo max.
+        with:
+          base-branch: master
+          # To ignore a bad cache, bump the cache* integer.
+          gha-cache-key: cache0
+          # This hash should include all of our lockfiles so that the pip/pex caches
+          # get invalidated on any transitive dependency update.
+          named-caches-hash: ${{ hashFiles('requirements.txt') }}
```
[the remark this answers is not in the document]
cognifloyd: Not entirely. I plan to change this to use the lockfiles instead once the lockfiles are available.

.github/workflows/lint.yaml
cognifloyd: I commented it out.

PR #5765, opened 2022-10-10
Ignore MemberAlreadyExist exception during startup.

khushboobhatia01:
During service startup, if service registry is enabled we register the service. 
If the service member ID already exists in the registry, the service start up fails with an exception _MemberAlreadyExist_.
This change will ignore this exception and continue with service startup.

**Why do we need this change**?

Consider a scenario where pod with a Stackstorm service is running.

1. During startup this pod would have been registered in the service registry with hostname = pod name.
2. Let's say the service container inside the pod is OOMKilled. 
3. Now, because of OOM kill the de-reregistration of service from the registry might not complete.
4. After some backoff, when the container is able to start up again in the same pod with the same pod name, the service startup fails with _MemberAlreadyExist_ exception because the de-reregistration process did not complete.


cognifloyd (commented):
Looks good. Needs a changelog entry, and one of your commits is using an email address that is not attached to your github account, so you'll need to rewrite the commits to fix that (or the CLA bot will keep complaining.

Also, I wonder if there's a decent way to add a test for this. Can you look into that?

cognifloyd:
@khushboobhatia01 I rebased and fixed the author email. Now the CLA bot is not complaining.

st2common/st2common/service_setup.py
```diff
@@ -341,7 +342,10 @@ def register_service_in_service_registry(service, capabilities=None, start_heart
         'Joining service registry group "%s" as member_id "%s" with capabilities "%s"'
         % (group_id, member_id, capabilities)
     )
-    return coordinator.join_group(group_id, capabilities=capabilities).get()
+    try:
+        return coordinator.join_group(group_id, capabilities=capabilities).get()
+    except MemberAlreadyExist:
+        pass
```
amanda11, line 348: More comment on legacy, but in the try we return whatever coordinator.join_group returns, but if memberalreadyexists we don't return anything...
I think in fact the join_group doesn't return anything - so probably just worth changing it so that we change the try to remove the "return" on the call to coordinator.join_group. Or if its expected to return something then we need to adjust the except clause...

PR #5775, opened 2022-10-14
Add option to dig so we can specify type

### LINKED ISSUE #5772: Add query type to linux.dig action
## SUMMARY

I would like the ability to query TXT records and noticed there is no way to specify a query type to the dig action. 

### STACKSTORM VERSION

`st2 3.6.0, on Python 3.6.8`

I attempted a few ways to add "TXT" to the query by adding to queryopts or try appending to the string hostname. Upon looking at the code I realized nothing like that would work.

## Expected Results

Get a list returned of TXT records

## Some sample code to add it

```
class DigAction(Action):
    def run(self, rand, count, nameserver, hostname, queryopts, querytype): # Add querytype parameter
        opt_list = []
        output = []

        cmd_args = ["dig"]
        if nameserver:
            nameserver = "@" + nameserver
            cmd_args.append(nameserver)

        if isinstance(queryopts, str) and "," in queryopts:
            opt_list = queryopts.split(",")
        else:
            opt_list.append(queryopts)

        cmd_args.extend(["+" + option for option in opt_list])

        cmd_args.append(hostname)
        cmd_args.append(querytype) # append query type (Default is set to "A" in dig.yaml)

        try:
            raw_result = subprocess.Popen(
                cmd_args, stderr=subprocess.PIPE, stdout=subprocess.PIPE
            ).communicate()[0]

            if sys.version_info >= (3,):
                # This function might call getpreferred encoding unless we pass
                # do_setlocale=False.
                encoding = locale.getpreferredencoding(do_setlocale=False)
                result_list_str = raw_result.decode(encoding)
            else:
                result_list_str = str(raw_result)

            if querytype.lower() == "txt":   # improve the output formatting result of TXT records
                result_list_str = result_list_str.replace('"', '')   # strip quotes so we don't see \" wrapped around output
            result_list = list(filter(None, result_list_str.split("\n")))
```

I only spent a few minutes on this code to test making it work for me. It could be improved on to make sure works for other types as well. I added inline comments to show the only lines I added

AmbiguousYeoman:
Closes #5772

Specify the type as a parameter to dig (e.g. NS, TXT, MX) and return a list, similar to how it currently returns multiple A records.

Currently the action can only return A records since that is the default to dig, this PR adds a new parameter and sets the default to "A".


AmbiguousYeoman:
@armab wonder if you can approve the workflows next time you are online for this PR. Issue: https://github.com/StackStorm/st2/issues/5772
Sorry to bug you, but wasn't sure if I need to reach out for visibility that these are queued or not.

armab (changes requested):
Nice addition!
Additional thanks for the tests :+1: 

Please also bump the linux pack version here to `1.2.0`: 
https://github.com/StackStorm/st2/blob/4099a09b082b6e3da94ae13819c40514671b2885/contrib/linux/pack.yaml#L19

Besides that, looks good.

armab:
@cognifloyd looks like some CI checks are failing on the new pants stuff. Are they expected to fail?

cognifloyd:
No it should not be failing. I'll look into what's wrong.

cognifloyd:
The pants CI fix is in #5782

armab (approved):

rush-skills (approved):
LGTM

PR #5782, opened 2022-10-22
Update GHA pants-init action to fix build

cognifloyd:
Github updated the default minor version of python 3.9.x The cache key did not include the minor version which broke things for us.

Luckily, this was already fixed in pantsbuild/actions#6. The fix requires an update to the `./pants` entrypoint/bootstrap script which was updated in pantsbuild/setup#128.

Based on [pants docs](https://www.pantsbuild.org/v2.14/docs/manual-installation#manual-installation), I updated `./pants` with:
```
curl -L -O https://static.pantsbuild.org/setup/pants && chmod +x ./pants
```

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -57,7 +57,7 @@ jobs:
           ./scripts/github/install-apt-packages-use-cache.sh
 
       - name: Initialize Pants and its GHA caches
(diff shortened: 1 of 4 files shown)


armab (approved):
Thanks!

cognifloyd:
The version of python installed for pants (inside this pants-init action) is unrelated to the versions we use for our code.

cognifloyd:
If we do want to change the version of python used by pants, there is an action input parameter that handles that.

amanda11 (approved):

PR #5783, opened 2022-10-23
Add backward compatibility to secret masking for output schema

m4dcoder:
Add backward compatibility to secret masking introduced in #5319 to prevent security-relative issues. Migration to the new schema is required to take advantage of the full output schema validation.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -41,7 +41,7 @@ Fixed
 
 * A new output schema using full JSON schema was introduced and secrets previously masked using
   the legacy output schema are now being displayed as plain text. To prevent security relative
-  issues, add backward compatibility to secret masking. Full output schema validattion will need
+  issues, add backward compatibility to secret masking. Full output schema validation will need
   to be migrated to the new schema.
 
   Contributed by @m4dcoder
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -23,7 +23,6 @@ Fixed
 
   Contributed by @bharath-orchestral
 
-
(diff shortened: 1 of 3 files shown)


cognifloyd (commented):
Can you adjust the name of the test (legacy is no longer "noop") and clean up the comment that says nothing will be masked?

CHANGELOG.rst
```diff
@@ -39,6 +39,13 @@ Fixed
 * Fixed generation of `st2.conf.sample` to show correct syntax for `[sensorcontainer].partition_provider` (space separated `key:value` pairs). #5710
   Contributed by @cognifloyd
 
+* A new output schema using full JSON schema was introduced and secrets previously masked using
+  the legacy output schema are now being displayed as plain text. To prevent security relative
+  issues, add backward compatibility to secret masking. Full output schema validation will need
+  to be migrated to the new schema.
```
armab, line 45: Can we refactor the messaging here a bit? 
```suggestion
* Add backward compatibility to secret masking introduced in #5319 to prevent security-relative issues. Full output schema validation will need
  to be migrated to the new schema. #5783
```
It sounds like there was an issue in the previous version of stackstorm.


In fact, this patch is added on top of #5319, which wasn't released yet and there was no harm made to the production versions yet.

It's great to have this issue caught early though!

m4dcoder:
ok. I'll update the changelog and description here.

CHANGELOG.rst
m4dcoder: Done

cognifloyd (approved):
I cleaned up the test name.

armab (approved):
Thanks for the changes!

PR #5795, opened 2022-11-04
Inform Pants about our target python versions (3.6 and 3.8)

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

Our 3rd party dependencies will be in a lockfile that looks like [`lockfiles/st2.lock`](https://github.com/st2sandbox/st2/blob/pants/lockfiles/st2.lock). But, before pants can generate that, we need to add several pieces of metadata including dependencies and python version constraints.

This PR focuses on registering our python version constraints (in `pants.toml`). Other PRs will add the dependencies (see #5789) and other metadata so that we can actually generate the lockfile.

This PR also regenerates the lockfiles for bandit, black, and flake8 to use the new interpreter constraints. (most of the changed lines are part of this)

- [Python Interpreter Compatibility](https://www.pantsbuild.org/docs/python-interpreter-compatibility)
    - [Setting the default Python version](https://www.pantsbuild.org/docs/python-interpreter-compatibility#setting-the-default-python-version)

### Python Version Constraints

We support python 3.6 and python 3.8; We do not officially support python 3.7.

I opted for a broader version constraint:
```
CPython>=3.6,<3.9
```
Instead of excluding python 3.7 like this:
```
CPython>=3.6,!=3.7.*,<3.9
```

NOTE: the black wheels for 3.6 require `>=3.6.2`, so I had to use that or pants+pex fails to resolve the lockfile.

I went with the broader constraint because:
Python 3.6 is EOL. But some people still use old (probably EOL) distros where 3.8 is not an easy option. So, 3.7 might be a decent option, even though it won't be officially "supported". All of our official efforts can go towards adding python 3.9 and later support. But, if for some reason someone goes to the effort to test and use python3.7 instead, good for them; Let's not put technical barriers in their path unless we have known issues on particular versions of python 3.7.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,7 +10,7 @@ Added
 * Continue introducing `pants <https://www.pantsbuild.org/docs>`_ to improve DX (Developer Experience)
   working on StackStorm, improve our security posture, and improve CI reliability thanks in part
   to pants' use of PEX lockfiles. This is not a user-facing addition.
-  #5778 #5789 #5817
+  #5778 #5789 #5817 #5795
   Contributed by @cognifloyd
 
 
--- a/lockfiles/bandit.lock
+++ b/lockfiles/bandit.lock
@@ -6,7 +6,7 @@
 // {
 //   "version": 3,
 //   "valid_for_interpreter_constraints": [
-//     "CPython<4,>=3.7"
+//     "CPython<3.9,>=3.6"
 //   ],
 //   "generated_with_requirements": [
 //     "bandit==1.7.0",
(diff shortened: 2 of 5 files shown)


cognifloyd:
I'm trying to configure pants to match our current setup, with python 3.6 and 3.8. I haven't tried to use newer python versions because I'm trying to minimize what it takes to get pants into the repo.

Pants+PEX do not support the legacy pip resolver any more. Our current set of manually pinned requirements includes a few conflicting requirements that legacy pip only warns about, and the newer pip resolver fails with. So, to get the benefits of the lockfile (instead of manually pinning everything), we already have to make changes to our requirements. Including more python versions will probably require additional reqs changes.

When we update to newer python versions, pants will be helpful and should make that process easier. But with so much changing already, I would prefer to update python support separately from introducing pants.

cognifloyd:
#5817 should be merged before this one as it only deals with the pants 2.14 upgrade instead of throwing it into this PR.

rush-skills (approved):

cognifloyd:
I rebased this on master to drop the commit that was covered by #5817 (now merged). This needs one more review before it can be merged. Then, I'll work on preparing the next batch of pants PRs.

Eric-Arellano (approved):

amanda11 (approved):
LGTM

PR #5817, opened 2022-11-24
Update to pants 2.14 (final)

cognifloyd:
This updates to pants 2.14. We were on release candidates before.

pants 2.14 changed a default setting from `ignore` to `warning`, but most of the warnings are not actionable for us. Eventually, we should be able to address the warnings cleanly. For now, let's just ignore the warnings.

Alternatively, we could add `# pants: no-infer-dep` comments all over the place, which is not ideal.

We can revisit this setting in pants 2.16. A new feature under development for 2.16 should make cleaning up these warnings much easier, and therefore worth the effort.


Eric-Arellano (approved):

nzlosh (approved):
LGTM

rush-skills (approved):
LGTM

PR #5832, opened 2022-12-02
pin GHA runners to ubuntu-20.04

cognifloyd:
GHA just updated the ubuntu VM so that ubuntu-latest now points to 22.04.1 instead of 20.04.5

https://github.blog/changelog/2022-12-01-github-actions-larger-runners-using-ubuntu-latest-label-will-now-use-ubuntu-22-04/
 https://github.com/actions/runner-images/issues/6399

We have pinned python 3.6.13 and 3.8.10 in our workflows. There are NO 3.6.* versions available on ubuntu 22.04 in GHA. The oldest 3.8 version available is 3.8.12 on ubuntu 22.04 in GHA. So, we need to use `ubuntu-20.04` instead of `ubuntu-latest` for our workflows at least until we can drop python 3.6 support in CI.

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -12,7 +12,7 @@ jobs:
   # See: https://keepachangelog.com/en/1.0.0/
   changelog-checker:
     name: Add CHANGELOG.rst
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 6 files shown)


rush-skills (approved):

armab (approved):
Thanks for taking care of :green_circle: !

PR #5833, opened 2022-12-03
pants: disambiguate test file dependencies

cognifloyd:
### Background

This is another part of introducing [pants](https://www.pantsbuild.org/docs), as discussed in various TSC meetings.

Related PRs can be found in:
- [pantsbuild](https://github.com/StackStorm/st2/milestone/47) milestone (which includes PRs since #5713)
- https://github.com/StackStorm/st2/labels/pantsbuild label (which also includes preparatory PRs that came before adding pantsbuild)

### Overview of this PR

Pants uses dependency inference, which the [docs](https://www.pantsbuild.org/docs/how-does-pants-work#dependency-inference) describe as:

> Pants analyzes your code's import statements to determine files' dependencies automatically. Dependency information is required for precise change detection and cache invalidation, but inference means that you don't need to declare dependencies manually (and hermetic execution guarantees that they are always accurate)!

However, sometimes that inference gets confused. For example, in several places our test files import `tests.base`, but we have multiple `**/tests/base.py` files, so pants doesn't know which one we're talking about, even though python itself understands exactly what we're talking about.

So, this PR resolves that ambiguity by explicitly defining dependencies in the BUILD files. In a few cases, we can add that dependency to the one file that needs it. But in other cases, lots of test files have the "ambiguous" import on `tests.base` or `tests.unit.base`. Adding the dep to each of those files would be painful to maintain, so I added it to all of the test files in the affected directories.

_Aside: In each case, I left a comment about why the dep was "ambiguous". I hope that a future version of pants will give us some knobs to tell the inference system how to resolve dependencies it thinks are ambiguous instead of adding it explicitly to more files than necessary. Including "ambiguous" in the BUILD file comment will make it easier to grep and figure out where we can adjust things once pants includes a feature like that._

#### Relevant Pants documentation

- [How Does Pants Work? | Dependency inference](https://www.pantsbuild.org/docs/how-does-pants-work#dependency-inference)
- [Troubleshooting | Import errors and missing dependencies](https://www.pantsbuild.org/docs/troubleshooting#import-errors-and-missing-dependencies)
- [Dependency Inference (blog article)](https://blog.pantsbuild.org/dependency-inference/)

PR #5830, opened 2022-12-02
pantsbuild: Generate PEX lockfile for st2

cognifloyd:
### Background

This is another part of introducing [pants](https://www.pantsbuild.org/docs), as discussed in various TSC meetings.

Related PRs can be found in:
- [pantsbuild](https://github.com/StackStorm/st2/milestone/47) milestone (which includes PRs since #5713)
- https://github.com/StackStorm/st2/labels/pantsbuild label (which also includes preparatory PRs that came before adding pantsbuild)

### Overview of this PR

This introduces a lockfile for our 3rd party dependencies (based on dependencies/constraints from #5789 and python constraints from #5795).

To create this lockfile, we need to register a "resolve" (called "st2") in pants, which represents the wold of direct and transitive dependencies required by our code. We also register the dependency constraints file added in #5789.

Finally, we create the lockfile by running `./pants generate-lockfiles --resolve=st2`.

Note that the lockfile has 4840 lines. Beyond that this PR only:
- adds 11 lines to `pants.toml`
- updates 1 line in `CHANGELOG.rst`

#### Relevant Pants documentation

- [Python 3rd Party Dependencies](https://www.pantsbuild.org/docs/python-third-party-dependencies)
    - [Python Lockfiles](https://www.pantsbuild.org/docs/python-third-party-dependencies#lockfiles)
    - [Python dependency constraints](https://www.pantsbuild.org/docs/python-third-party-dependencies#constraints-files)
    - [Manually Generating Lockfiles](https://www.pantsbuild.org/docs/python-third-party-dependencies#manually-generating-lockfiles)
- [Python Interpreter Compatibility](https://www.pantsbuild.org/docs/python-interpreter-compatibility)
- Goals:
    - [generate-lockfiles](https://www.pantsbuild.org/docs/reference-generate-lockfiles)
- `pants.toml` options:
    - [`[python].enable_resolves = <bool>`](https://www.pantsbuild.org/docs/reference-python#enable_resolves)
    - [`[python.resolves].<resolve name> = "<lockfile path>"`](https://www.pantsbuild.org/docs/reference-python#resolves)
    - [`[python].default_resolve = "<resolve name>"`](https://www.pantsbuild.org/docs/reference-python#default_resolve)
    - [`[python.resolves_to_constraints_file].<resolve name> = "<constraints file path>"`](https://www.pantsbuild.org/docs/reference-python#resolves_to_constraints_file)

### Resolves and Lockfiles

[...]

PR #5834, opened 2022-12-03
pants: register requirements for contrib/examples pack

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

In #5733, I configured pants to ignore several requirements files, including `contrib/examples/requirements.txt`:

https://github.com/StackStorm/st2/blob/f25df1f939ef894f87d4bb064ccaadd25c0bd321/pants.toml#L28-L29

One of the reasons I excluded that file, beyond what I described in #5733, was because it includes `requests` which is already included in the main requirements files:
https://github.com/StackStorm/st2/blob/f25df1f939ef894f87d4bb064ccaadd25c0bd321/contrib/examples/requirements.txt#L1
https://github.com/StackStorm/st2/blob/f25df1f939ef894f87d4bb064ccaadd25c0bd321/requirements-pants.txt#L58

If we included both of these, then pants would start complaining that `import requests` is ambiguous all over the place. Compare that to #5833 where I had to disambiguate some other dependencies, but those affected a more isolated set of files. We do not want to register explicit dependencies on `requests` all over the place as that would be quite painful. So, excluding the requirements file was the best course of action.

However, there are two more dependencies that are NOT in the main requirements file:
https://github.com/StackStorm/st2/blob/f25df1f939ef894f87d4bb064ccaadd25c0bd321/contrib/examples/requirements.txt#L2-L3

These deps, `beautifulsoup4` and `flask`, are only needed for the examples pack. Including those in the main requirements file would be a signal for future contributors that they can use those in other parts of the codebase, but we don't want anything other than the examples pack to depend on them.

So, this PR:
- registers these deps locally in the `contrib/examples/BUILD` file, and
- regenerates `lockfiles/st2.lock` (introduced in #5830).

This PR also shows an example of how github displays the lockfile diff when we add new dependencies.

_Aside: Once we upgrade to pants 2.16 (which is in development), a new "dependency rules" feature will allow us to enforce that nothing else can import these deps. That will make this even safer._

- [How Does Pants Work? | Dependency inference](https://www.pantsbuild.org/docs/how-does-pants-work#dependency-inference)
- [Troubleshooting | Import errors and missing dependencies](https://www.pantsbuild.org/docs/troubleshooting#import-errors-and-missing-dependencies)
- [Dependency Inference (blog article)](https://blog.pantsbuild.org/dependency-inference/)
- [Python 3rd Party Dependencies](https://www.pantsbuild.org/docs/python-third-party-dependencies)
    - [Python Lockfiles](https://www.pantsbuild.org/docs/python-third-party-dependencies#lockfiles)
    - [Manually Generating Lockfiles](https://www.pantsbuild.org/docs/python-third-party-dependencies#manually-generating-lockfiles)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,7 +10,7 @@ Added
 * Continue introducing `pants <https://www.pantsbuild.org/docs>`_ to improve DX (Developer Experience)
   working on StackStorm, improve our security posture, and improve CI reliability thanks in part
   to pants' use of PEX lockfiles. This is not a user-facing addition.
-  #5778 #5789 #5817 #5795 #5830
+  #5778 #5789 #5817 #5795 #5830 #5834
   Contributed by @cognifloyd
 
 
--- a/contrib/examples/BUILD
+++ b/contrib/examples/BUILD
@@ -1,0 +1,14 @@
+# Using `python_requirements()` here results in:
+# ">1 target exports this module, so it is ambiguous which to use"
+# This error refers to the "requests" module.
+# So, we explicitly list deps instead.
+
+python_requirement(
+    name="beautifulsoup4",
+    requirements=["beautifulsoup4"],
+)
+
+python_requirement(
+    name="flask",
+    requirements=["flask"],
+)
--- a/lockfiles/st2.lock
+++ b/lockfiles/st2.lock
@@ -13,9 +13,11 @@
 //     "RandomWords",
 //     "apscheduler",
 //     "argcomplete",
+//     "beautifulsoup4",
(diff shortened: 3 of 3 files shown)


cognifloyd:
Note: This PR needs to be merged before I can add pylint. Otherwise, pants will run into these ambiguous deps whenever it runs pylint, and pylint will subsequently fail when pants doesn't add all of the required deps to the sandbox it runs pylint in.

amanda11 (approved):
LGTM

rush-skills (approved):
LGTM

PR #5837, opened 2022-12-06
Use pylint (via pantsbuild)

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

This configures pants to use `pylint` when running the `lint` goal. Configuring pylint is multi-faceted. We have to add config in `pants.toml` (like the pylint version to use and adding a new `pylint_plugins` resolve), generating lockfiles for pylint and pylint_plugins, and adding `skip_pylint=True` to a lot of `BUILD` files.

I recommend reviewing this PR by stepping through the commits.

This adds a new resolve, `pylint_plugins`, for the code under `pylint_plugins/`. This is separate from our default `st2` resolve. I would just re-use the `pylint` resolve and lockfile (for the `pylint` tool), but pants does not support reusing a tool's config like that yet (_There is discussion in the pants community on how to make this work better. We'll see how that goes._). So, we have to duplicate the version requirements in two places, and have two lockfiles that have practically the same contents.

- [`./pants lint` goal](https://www.pantsbuild.org/v2.14/docs/reference-lint)
- [`./pants generate-lockfiles` goal](https://www.pantsbuild.org/v2.14/docs/reference-generate-lockfiles)
- [Python Overview](https://www.pantsbuild.org/v2.14/docs/python)
    - [Python Linters and Formatters](https://www.pantsbuild.org/v2.14/docs/python-linters-and-formatters)
    - [pylint backend](https://www.pantsbuild.org/v2.14/docs/reference-pylint)
- [Python 3rd Party Dependencies](https://www.pantsbuild.org/docs/python-third-party-dependencies)
    - [Python Lockfiles](https://www.pantsbuild.org/docs/python-third-party-dependencies#lockfiles)
    - [Manually Generating Lockfiles](https://www.pantsbuild.org/docs/python-third-party-dependencies#manually-generating-lockfiles)

### 68f19214255e60760163066a87ece5f399896d70 record pylint version requirements

First, this commit records the pylint version in `pants.toml`. This version is for the `pylint` resolve, which is used whenever pants is running `pylint`. It also adds the `setuptools` dep, because `pylint` fails to report that it uses `pkg_resources` from setuptools, and setuptools is NOT installed by default with pants+PEX.

https://github.com/StackStorm/st2/blob/68f19214255e60760163066a87ece5f399896d70/pants.toml#L147-L151

Second, this commit records the pylint version in `pylint_plugins/BUILD`. This version is for the `pylint_plugins` resolve, which is used whenever pants is running linters/tests/etc on the code in `pylint_plugins/`. This mirrors `pants.toml` by also including the same setuptools requirement.

https://github.com/StackStorm/st2/blob/68f19214255e60760163066a87ece5f399896d70/pylint_plugins/BUILD#L7-L15

Third, this commit adds an `astroid` dependency in `pylint_plugins/BUILD`, because that is actually the only third party dependency of our `pylint_plugins/` code. But, we do not define a version for astroid; instead, the version of `astroid` gets constrained because we also require `pylint`, specifying the same as in `pants.toml`. `pylint` has narrow support for which versions of astroid it can use, so trying to maintain a version of both pylint and astroid proved irritating. Instead, we rely on the lockfile to keep things consistent in-between upgrading pylint.

https://github.com/StackStorm/st2/blob/68f19214255e60760163066a87ece5f399896d70/pylint_plugins/BUILD#L17-L21

### 395c869a2848a857d41234c4480e16ac90fea566 add new resolve for pylint_plugins/ deps

First, this commit defines the `pylint_plugins` resolve by setting the path to the lockfile (which will be generated in a later commit).

https://github.com/StackStorm/st2/blob/395c869a2848a857d41234c4480e16ac90fea566/pants.toml#L97-L99

Second, this commit changes the `__defaults__` for the `pylint_plugins/` dir so that everything in that dir uses the new resolve:

https://github.com/StackStorm/st2/blob/395c869a2848a857d41234c4480e16ac90fea566/pylint_plugins/BUILD#L1-L5

Third, this commit drops an implicit dependency on the root `conftest.py` file. Code in the `pylint_plugins` resolve cannot import code in the `st2` resolve. pants will error if something tries to do that. (Plus, this renames the target name from `test_utils0` to `test_utils` for clarity (the original name was generated by `./pants tailor`).

https://github.com/StackStorm/st2/blob/395c869a2848a857d41234c4480e16ac90fea566/pylint_plugins/BUILD#L9-L14

### 9ac83651ba51e9d7094168891e567da290f30882 enable pylint via pants

This registers the pylint backend in `pants.toml`, and tells pants where our `.pylintrc` file is, and where our pylint plugins are.

Effectively this covers how the Makefile runs `pylint` with args: `--rcfile=./lint-configs/python/.pylintrc --load-plugins=pylint_plugins.api_models --load-plugins=pylint_plugins.db_models`

https://github.com/StackStorm/st2/blob/9ac83651ba51e9d7094168891e567da290f30882/Makefile#L346

### e085f522b576a0ef93d5be8ede88de36bdbbd34d Generate pylint* lockfiles

This generates both lockfiles: `lockfiles/pylint.lock` and `lockfiles/pylint_plugins.lock`. These files are exactly the same, except for the name of the resolve (`pylint` vs `pylint_plugins`). Each lockfile is 385 lines long. So, 770 of the lines in this PR are auto-generated.

### add `skip_pylint=True` metadata to BUILD files

The next three commits add `skip_pylint=True` in a variety of contexts where running pylint doesn't work well. This runs pylint on a lot more of our code than the Makefile does, but skips files as needed. Maybe in the future we can enable more of our code.

In particular this skips running pylint on:
- [tests](https://github.com/StackStorm/st2/pull/5837/commits/4fd2f06b86a337facfb24cef9d15be6be7338380)
- [intentionally invalid python file](https://github.com/StackStorm/st2/pull/5837/commits/e80977ffacef9d945e74f86b80bbded8579247bb)
- [`st2*/bin` dirs](https://github.com/StackStorm/st2/pull/5837/commits/0be4ef83a39b3823ba3d02a36b186a9da25755a5) (because the script names in bin ended up shadowing the modules they were importing and the `-` in other script names was problematic. see: https://github.com/PyCQA/pylint/issues/2095)

There are two ways we add this metadata. We either add `skip_pylint=True` to individual targets, or we can skip entire directories with `__defaults__`:
https://github.com/StackStorm/st2/blob/4fd2f06b86a337facfb24cef9d15be6be7338380/st2actions/tests/BUILD#L1-L5

### f4cb99f5eb2940f256923f00080ba21175f602c6 fix pants dep inference for pylint_plugins

It turns out, we have to use different pylint args when pants runs `pylint` vs when the legacy `Makefile` runs it. `pants` adds source_roots to PYTHONPATH, so `pylint_plugins/` is on the PYTHONPATH which makes `api_plugins` and `api_models` top-level modules. When it is not on the PYTHONPATH, as in the Makefile, `pylint_plugins` is the top-level module. So, we had to just an import in the python test, and the args in `pants.toml.

### 4c7888d4b14af72ced50482841661c473a3e267c use pylint -E under pants

The Makefile uses `-E` when running pylint:

https://github.com/StackStorm/st2/blob/4c7888d4b14af72ced50482841661c473a3e267c/Makefile#L346

I had to look up what that meant. Running pylint without that generates a lot of warnings. It turns out `-E` is short for `--errors-only`. So, I used `--errors-only` instead of `-E` to remind myself of what it does in the future. Hopefully we can drop this at some point.

https://github.com/StackStorm/st2/blob/4c7888d4b14af72ced50482841661c473a3e267c/pants.toml#L161-L162


Eric-Arellano (approved):
Exciting!

amanda11 (approved):
LGTM

rush-skills (approved):
LGTM

PR #5838, opened 2022-12-07
Generate lockfile for using pytest

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

This configures pants so it can use `pytest` when running the `test` goal. However, our test suite does not completely support pytest. Getting this lockfile in is a prerequisite to being able to running tests with it.

#### pytest version

I opted to go with the version of pytest that pants has pinned by default.

https://github.com/StackStorm/st2/blob/43e2a1d2f44790676728c24c073f2e86a26cac99/pants.toml#L147-L149

#### pytest plugins

By default pants includes `pytest-cov` and `pytest-xdist` plugins for pytest. We get those too because I'm using `extra_requirements.add`; if I used `extra_requirements` it would overwrite those.

https://github.com/StackStorm/st2/blob/43e2a1d2f44790676728c24c073f2e86a26cac99/pants.toml#L150

We already install `pytest-benchmark[histogram]` for our tests in `st2common/benchmarks`, so I copied that pinned version as is:

https://github.com/StackStorm/st2/blob/43e2a1d2f44790676728c24c073f2e86a26cac99/test-requirements.txt#L41
https://github.com/StackStorm/st2/blob/43e2a1d2f44790676728c24c073f2e86a26cac99/pants.toml#L151

Then I added several other plugins, some of which I left commented (we can check them out later when we do more with pytest):

https://github.com/StackStorm/st2/blob/43e2a1d2f44790676728c24c073f2e86a26cac99/pants.toml#L152-L158

This is partially based on recommendations in the pants docs. See: https://www.pantsbuild.org/docs/python-test-goal
These plugins have made what I've done with pytest so far, more enjoyable.

#### pytest args

As recommended at https://www.pantsbuild.org/docs/python-test-goal#controlling-output, I dropped the pytest header. Pants runs each test file separately to help with caching, so that header becomes very annoying and repetitive.

https://github.com/StackStorm/st2/blob/43e2a1d2f44790676728c24c073f2e86a26cac99/pants.toml#L164-L166

#### parallel tests with mongo

This line of the config builds on #5707, so that pants can safely run tests that use mongo in parallel.

https://github.com/StackStorm/st2/blob/43e2a1d2f44790676728c24c073f2e86a26cac99/pants.toml#L167

see:
- https://www.pantsbuild.org/docs/reference-pytest#execution_slot_var
- https://www.pantsbuild.org/docs/troubleshooting#controlling-test-parallelism


armab (approved):

pants.toml
```diff
@@ -144,5 +144,27 @@ extra_requirements = [
...
+  "pytest-icdiff",  # make diff output easier to read
+  "pygments",  # highlight code in tracebacks
+
+  # other possible plugins
+  #"pytest-timeout",  # time limit on tests
+  #"pytest-mock",  # more convenient mocking
+
+  # included by default with pants
+  #"pytest-cov",  # coverage
+  #"pytest-xdist",  # parallel test runs (pants uses this if [pytest].xdist_enabled)
```
Eric-Arellano, line 162: You won't be able to use some Pants features with these commented out

Eric-Arellano (approved):

pants.toml
cognifloyd: This serves as documentation because I'm using `extra_requirements.add`

nzlosh (approved):

PR #5840, opened 2022-12-07
Generate lockfile for using setuptools

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

This configures pants and generates a lockfile so it can use `setuptools` when running the `package` goal, which includes building wheels.

We will still need more config before we can use pants to build wheels, but this is a prerequisite for that. To build wheels, we also need to add metadata to BUILD files. To simplify the metadata that has to be repeated for each wheel future PRs will include at least: a pants macro and a pants plugin.

Here is the `pants.toml` config this PR adds:

https://github.com/StackStorm/st2/blob/ec46508c10105c0d93978c0ad2a611e4f41f5b8b/pants.toml#L150-L153

Note: _Many other resolves also include setuptools. This resolve is for when pants has to call setuptools directly, not for when it is used indirectly._

#### setuptools version

I added a range of versions that we can use. I tried using newer versions, but they didn't work with python 3.6, so I capped that.

#### lockfiles/setuptools.lock

At 120 lines this is the smallest lockfile we have.


Eric-Arellano (approved):

amanda11 (approved):
LGTM

nzlosh (approved):

PR #5841, opened 2022-12-07
Generate lockfile for using twine

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

This configures pants and generates a lockfile so it can use `twine` when running the `publish` goal to publish wheels to pypi.

We will still need more config before we can use pants to just build wheels, let alone publish them. But, this is the last lockfile I know of that we need. Getting this in now reduces the diff with my PoC branch so it is easier to plan the next PRs.

Here is the `pants.toml` config this PR adds:

https://github.com/StackStorm/st2/blob/2eb18db83d3025219d71c97e8519c37905b9d1ae/pants.toml#L150-L152

#### twine version

I copied the pants' default version for twine: https://www.pantsbuild.org/v2.14/docs/reference-twine#version

#### lockfiles/twine.lock

This lockfile defers from the one that comes with pants because we have different python interpreter constraints.


Eric-Arellano (approved):

amanda11 (approved):
LGTM

nzlosh (approved):

PR #5843, opened 2022-12-08
[WIP] Investigate if ST2 runs with python 3.10

amanda11:


amanda11:
Bumping eventlet will require gunicorn update - see https://github.com/StackStorm/st2/pull/5257

Kami:
While at it, it may also make sense to research / test if it works under Python 3.11 since there aren't that many differences between 3.10 and 3.11.

3.11 includes specializing adaptive interpreter (https://peps.python.org/pep-0659/) which operates on Python byte code level which offers nice performance in some situations on pure Python code.

Granted, StackStorm depends on a lot of libraries with C extensions, but there is still plenty of pure Python code and who knows, it may also bring nice performance improvements to StackStorm.

I've done some micro and end to end benchmarks on a similar Python project (majority is Python code, but also a lot of libraries with C extensions for performance critical code) and we've seen some nice improvements in terms of performance and lower CPU usage (even in the places we haven't really expected it due to the code mostly calling library with C extension).

amanda11:
We need to move away from nose which is not maintained and not py 3.10 compatible. It is cause of this error:
```
  File "/home/runner/work/st2/st2/virtualenv/lib/python3.10/site-packages/nose/suite.py", line 106, in _set_tests
    if isinstance(tests, collections.Callable) and not is_suite:
AttributeError: module 'collections' has no attribute 'Callable'
```
We are moving to using pytest as part of the pants work, so ignore the nose errors, as we will be moving tests to pytest..

amanda11:
Also ignoring the fact that some of the libraries will break python 3.6. With pants we should be able to have different libraries for python 3.6 to python 3.10 -> so aim of PR is to see what we can use to get python 3.10 to work...

nzlosh:
Thanks for the work, closing as py3.10 is now merged master.

PR #5845, opened 2022-12-11
clone packs into user_home/.st2packs directory

murthysrd:
Openshift does not allow packs to be cloned to home directory as they are mounted at `/`. Proposing to use a directory `.st2packs` in home directory.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,7 @@ in development
 
 Added
 ~~~~~
+* Move `git clone` to `user_home/.st2packs` #5845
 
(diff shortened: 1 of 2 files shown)


rush-skills:
@cognifloyd @armab Will this change effect any of our build pipelines?

cognifloyd (approved):
I don't see any problems with this, and I don't think it will affect any other tooling or scripts. I didn't realize we were directly using the home directory, so this feels much better.

Please use double quotes on the string (or run `black` to reformat which will do the same).

rush-skills (approved):

PR #5846, opened 2022-12-12
pants: register assets (like conf files) and dependencies on them

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

So far, we have told pants about our python and our shell sources. `./pants tailor ::` made that much easier. But we have more than just `*.py` and `*.sh` files; we have conf, ini, yaml, json, and many other kinds of files. In pants, these miscellaneous files are all ["assets"](https://www.pantsbuild.org/docs/assets), and `tailor` does not know what to do with them. So, we need to register the assets manually.

This PR registers most of our assets with one of these targets: `file`, `files`, `resource`, or `resources`.

This PR also registers dependencies on these assets. I think I captured all the dependencies, but we may have to tweak them if I missed something.

I recommend stepping through the commits so that you can quickly see what depends on the assets added in that commit.

- [Assets and archives](https://www.pantsbuild.org/docs/assets)
    - [When to use each asset target type](https://www.pantsbuild.org/docs/assets#when-to-use-each-asset-target-type)
- individual source targets:
    - [`file`](https://www.pantsbuild.org/docs/reference-file)
    - [`resource`](https://www.pantsbuild.org/docs/reference-resource)
- multiple sources targets (aka target generators):
    - [`files`](https://www.pantsbuild.org/docs/reference-files)
    - [`resources`](https://www.pantsbuild.org/docs/reference-resources)

### Files vs Resources

It can be confusing to figure out whether we need to register the assets as files or resources. In my PoC branch, I started with files and then gradually switched things over to resources when needed (like for building wheels). We will probably have to tweak `file`/`resource` on some of these in the future.

_aside: There have been some discussions in the pantsbuild community about how `file` and `resource` could be combined into an `asset` target. Whether or not that happens, registering assets in pants will hopefully improve with time. For now, just think of "assets" as a helpful term to refer to both "files" and "resources"._

### Future PRs

- Follow-up PRs will introduce some plugins that add specific handling to some of the resources/files in this PR. This way, pants will be able to auto-generate or lint things like the api spec. I will make those once this PR is merged.

- I will register the `LICENSE` file in conjunction with the metadata that we need to build wheels.

- There is also a large set of assets that I have intentionally skipped in this PR: pack metadata (including yaml files for packs, actions, workflows, rules, etc, and a few other misc files that get included in packs). I wrote a pants plugin to handle registering these assets because there are a lot of them. The plugin teaches `./pants tailor ::` how to handle those files, automatically adding a custom `pack_metadata` target.

### DIFF
--- a/conf/BUILD
+++ b/conf/BUILD
@@ -1,0 +1,56 @@
+file(
+    name="st2client_sample_config",
+    source="st2rc.sample.ini",
+)
+
+file(
+    name="st2.conf.sample",
+    source="st2.conf.sample",
+)
+
+file(
+    name="logrotate",
+    source="logrotate.conf",
+)
+
+file(
+    name="nginx_sample_config",
+    source="nginx/st2.conf",
+)
+
+file(
+    name="st2_kvstore_demo_crypto_key",
+    source="st2_kvstore_demo.crypto.key.json",
+)
+
+files(
+    name="st2_tests_conf",
+    sources=[
+        "st2.tests*.conf",
+    ],
+    dependencies=[
+        "st2auth/conf:htpasswd",
+        "st2actions/conf:logging",
+        "st2reactor/conf:logging",
+        "st2tests/conf:other-logging-conf",
+    ],
+)
+
+file(
+    name="st2_dev_conf",
+    source="st2.dev.conf",
+    dependencies=[
+        ":st2_kvstore_demo_crypto_key",
+        "st2auth/conf:htpasswd",
+        "st2actions/conf:logging",
+        "st2api/conf:logging",
+        "st2auth/conf:logging",
+        "st2reactor/conf:logging",
+        "st2stream/conf:logging",
+    ],
+)
+
+file(
+    name="st2_package_conf",
+    source="st2.package.conf",
+)
--- a/conf/BUILD
+++ b/conf/BUILD
@@ -32,7 +32,7 @@ files(
         "st2auth/conf:htpasswd",
         "st2actions/conf:logging",
         "st2reactor/conf:logging",
-        "st2tests/conf:other-logging-conf",
+        "st2tests/conf:other_logging_conf",
     ],
 )
 
--- a/st2common/st2common/BUILD
+++ b/st2common/st2common/BUILD
@@ -1,1 +1,11 @@
-python_sources()
+python_sources(
+    dependencies=[
+        ":openapi_spec",
+    ]
+)
+
+# These may be loaded with st2common.util.spec_loader
+resources(
+    name="openapi_spec",
+    sources=["openapi.yaml", "openapi.yaml.j2"],
+)
--- a/st2common/st2common/policies/meta/BUILD
+++ b/st2common/st2common/policies/meta/BUILD
@@ -1,0 +1,5 @@
+# These are loaded by st2common.bootstrap.policiesregistrar
+resources(
+    name="policies_meta",
+    sources=["*.yaml"],
+)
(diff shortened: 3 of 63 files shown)


conf/BUILD
```diff
@@ -0,0 +1,60 @@
+files(
+    name="st2client_sample_config",
+    sources=["st2rc.sample.ini"],
+)
+
+file(
+    name="st2.conf.sample",
+    source="st2.conf.sample",
+)
```
cognifloyd, line 9: A future PR will add `pants-plugins/sample_conf` which will wire this up so that it gets regenerated whenever someone runs the `fmt` goal with something like: `./pants fmt conf/st2.conf.sample`. That will take advantage of the fine-grained caching so that this always gets updated when the files used to generate it change.

st2common/st2common/BUILD
```diff
@@ -1 +1,11 @@
...
+    dependencies=[
+        ":openapi_spec",
+    ]
+)
+
+# These may be loaded with st2common.util.spec_loader
+resources(
+    name="openapi_spec",
+    sources=["openapi.yaml", "openapi.yaml.j2"],
+)
```
cognifloyd, line 11: A future PR will add `pants-plugins/api_spec` which will wire this up so that it gets regenerated whenever someone runs the `fmt` goal with something like: `./pants fmt st2common/st2comon/openapi.yaml`. That will take advantage of the fine-grained caching so that `openapi.yaml` always gets updated when the files used to generate it change.

st2common/st2common/policies/meta/BUILD
```diff
@@ -0,0 +1,5 @@
+# These are loaded by st2common.bootstrap.policiesregistrar
+resources(
+    name="policies_meta",
+    sources=["*.yaml"],
+)
```
cognifloyd, line 5: Adding this BUILD file broke a test that was checking directory contents. So, I assisted the test to account for it.

st2tests/st2tests/policies/meta/BUILD
```diff
@@ -0,0 +1,4 @@
+resources(
+    name="policies_meta",
+    sources=["*.yaml"],
+)
```
cognifloyd, line 4: Adding this BUILD file broke a test that was checking directory contents. So, I adjusted the test to account for it.

st2common/st2common/util/schema/BUILD
```diff
@@ -1 +1,10 @@
...
+python_sources(
+    dependencies=[
+        ":jsonschema",
+    ]
+)
+
+resources(
+    name="jsonschema",
+    sources=["*.json"],
+)
```
cognifloyd, line 10: This is required before I can submit the pants plugin that wires up pants to manage building the schemas in `contrib/schemas`.

conf/BUILD
```diff
@@ -0,0 +1,60 @@
+files(
+    name="st2client_sample_config",
+    sources=["st2rc.sample.ini"],
+)
+
```
amanda11, line 5: I don't understand the difference between file and files here?
Can you add a comment as unclear why file used in some cases and files in others, when the source seems to be a singleton in both cases?

cognifloyd: Because I started with the plural `files` form everywhere, and then switched some of them to the singular `file`.

If it's likely we will add more files later, then I would want to keep the plural form. But here, a simple `file` should work just fine. I'll review and switch some more of them.

cognifloyd: OK. I rebased on master and updated the commits to prefer the singular where possible.

nzlosh (approved):

rush-skills (approved):
LGTM

PR #5848, opened 2022-12-14
GHA workflow to run (some) tests via pantsbuild

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

This adds a `Test` GHA workflow that uses pants to run tests. Our test suite does not support running via pytest (which is what pants uses), so this workflow only runs tests on `pylint_plugins/` and `pants-plugins/`. We can slowly expand this as we get more of our tests runnable under pytest.

The reason I'm doing this now: I want to add tests for plugins that will go in `pants-plugins/`, and I want to use pants to run those tests. `pylint_plugins/` serves as a starting point to show that testing is doing something before we add those pants plugins and related tests.

#5849 was split off of this PR to make review easier. #5849 is required before this PR can be rebased and then merged.

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -57,7 +57,7 @@ jobs:
           ./scripts/github/install-apt-packages-use-cache.sh
 
       - name: Initialize Pants and its GHA caches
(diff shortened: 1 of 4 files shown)


nzlosh (approved):

amanda11 (commented):
LGTM

amanda11 (approved):

PR #5849, opened 2022-12-16
use a fixture for pylint_plugins test

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

In #5848, I will add a GHA workflow to run some tests via pants, starting with tests in `pylint_plugins/`. But, the current tests in pylint_plugins import from `st2common` because that was a quick way to throw together the tests. This is a problem because `pylint_plugins/` and `st2*` each use different resolves (and therefore separate lockfiles). That means that pants puts together the sandbox for running the pylint_plugins tests, it will not include the st2* modules.

Note that running pylint ON our source code works just fine. In that case, pants knows we need pylint + any source plugins for pylint + the source code that pylint needs to inspect. For tests, however, everything is based on dependencies between targets. I could use a pants feature that puts our code in both the `st2` and the `pylint_plugins` resolve, but that makes the lockfile and BUILD file management more complex.

So, I copied a few bits from `st2common` into a test fixture for the pylint_plugins tests. The Makefile runs the pylint_plugins tests just before it runs pylint, so look in the "CI / Compile" jobs to see the test results (which pass). I've also tested with pants in #5848 to make sure this change works for both ways to test this.

### DIFF
--- a/pylint_plugins/api_models_test.py
+++ b/pylint_plugins/api_models_test.py
@@ -25,9 +25,15 @@ import pylint.testutils
 try:
     # TODO: remove this once we remove the Makefile
     from . import api_models
+
+    FIXTURE_MODULE_ACTION = "pylint_plugins.fixtures.api_models"
+    FIXTURE_MODULE_TRIGGER = "pylint_plugins.fixtures.api_models"
 except ImportError:
     # pylint_plugins is on PYTHONPATH
     import api_models
+
+    FIXTURE_MODULE_ACTION = "fixtures.api_models"
+    FIXTURE_MODULE_TRIGGER = "fixtures.api_models"
 
 
 def test_skiplist_class_gets_skipped():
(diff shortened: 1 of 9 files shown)


pylint_plugins/api_models_test.py
```diff
@@ -135,12 +141,13 @@ class ActionCreateAPI(object):
 def test_copied_imported_schema():
     code = """
     import copy
-    from st2common.models.api.action import ActionAPI
+    from %s import ActionAPI
```
rush-skills, line 144: What does this `%s` do/how does that work?

rush-skills: Stupid question, saw the string replacement below

pylint_plugins/fixtures/api_models.py
```diff
@@ -0,0 +1,159 @@
...
+    },
+    "additionalProperties": False,
+}
+
+
+def get_schema(**kwargs):
+    return {}
+
+
+# copied from st2common.models.api.base
```
amanda11, line 46: If we change the base classes in st2 do we need to change these?
If so, can we add a comment on the st2 files themselves.
But I am assuming we don't need to.

amanda11 (approved):
See comment mentioned.

pylint_plugins/fixtures/api_models.py
cognifloyd: Yeah a comment would probably make sense.

It's ok if they drift (the precise options don't matter for the test). But if we do any more major refactoring, then we need to also update this so our test works.

cognifloyd: I added comments about this.

nzlosh (approved):

PR #5853, opened 2022-12-21
pants: use `__defaults__` to propogate unit,integration,benchmarks tags on tests

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

This adjusts the BUILD metadata for all of our `python_tests` targets so that they are tagged with `unit`, `integration`, or `benchmarks`. We can add other tags when/if we find something that will be generally useful.

Once we are able to run tests via pants, pants can use this to select a subset of tests to run. For example:

```
./pants --tag=benchmarks test
```

Check out the docs for more powerful examples of how to use tags on the command line.

#### Relevant pants docs

- [Tags: annotating targets](https://www.pantsbuild.org/docs/advanced-target-selection#tags-annotating-targets)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,7 +14,7 @@ Added
   working on StackStorm, improve our security posture, and improve CI reliability thanks in part
   to pants' use of PEX lockfiles. This is not a user-facing addition.
   #5778 #5789 #5817 #5795 #5830 #5833 #5834 #5841 #5840 #5838 #5842 #5837 #5849 #5850
-  #5846
+  #5846 #5853
   Contributed by @cognifloyd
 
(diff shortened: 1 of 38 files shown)


Eric-Arellano (approved):
FYI there's a gotcha that if you have any `python_test` targets rather than `python_tests`, they won't have the tags. That's probably unlikely you'll have that happen, but could. You can avoid the risk by using the key `(python_test, python_tests)`

cognifloyd:
> FYI there's a gotcha that if you have any `python_test` targets rather than `python_tests`, they won't have the tags. That's probably unlikely you'll have that happen, but could. You can avoid the risk by using the key `(python_test, python_tests)`

Ah yes. Thanks for the pointer. Resolved.

nzlosh (approved):

PR #5857, opened 2023-01-05
Add `pants-plugins/api_spec` to streamline regenerating `openapi.yaml`

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

This PR improves the DX (developer experience) by wiring up:
- the `fmt` goal to generate `st2common/st2common/openapi.yaml` if needed (effectively runs `st2-generate-api-spec`), and
- the `lint` goal to validate `st2common/st2common/openapi.yaml` to make sure the generated schema is valid (effectively runs `st2-validate-api-spec`)

This includes creating a `api_spec` plugin for pants that adds a `api_spec` target.

### Developer Experience

Today, if someone changes one of the files/scripts used to generate our openapi spec, then CI might complain as part of the `ci-checks` Makefile target, with instructions to `Please run "make generate-api-spec"`. This also gets regenerated any time someone runs the `lint` or `tests` Makefile targets, which might be surprising (tests generally shouldn't handle codegen).

In any case, this wires up pants to run `st2-generate-api-spec` and `st2-validate-api-spec` at times that are analogous to when the Makefile runs them.

With this PR, we add api spec generation into the `fmt` goal and validation to the `lint` goal. Anything that runs under `fmt` _also_ runs under the `lint` goal, so we will get two signals about the api spec: (1) whether or not it needs to be regenerated, and (2) whether or not there are validation errors in the spec.

Our api spec is generated from a jinja template, which only does minimal interpolation (populating the values for some of the enums). There are actually quite a few inconsistencies between our models and the api spec, so we will have to refactor the spec, the spec generation, and possibly the models in the future. That refactoring work, however, is out-of-scope for this PR. After we get pants in, it would be excellent to revisit these issues.

In working on this plugin, I discovered that our api spec validation has been broken for awhile, which I fixed in #5709. Part of our custom validation is still disabled because there are so many things in the spec that are out of date. This PR also does not fix that, but it does leave a comment in the pants plugin on how to enable that later.

After someone runs `./pants fmt ::` once, they will benefit from the pants' results cache. For the api_spec plugin, here's what that means:

First, the `api_spec` target in `st2common/st2common` depends on the generate/validate binaries in `st2common/bin`:

https://github.com/StackStorm/st2/blob/1856f6108219ebb1a6886afb798fd0f6c97b184c/st2common/st2common/BUILD#L12-L21

If either script (`st2-generate-api-spec` or `st2-validate-api-spec`) changes, then pants now knows that `openapi.yaml` need to be regenerated. Or, in other words, the `st2-generate-api-spec` file is part of the cache key for `st2common/st2common/openapi.yaml`. If that file changes, then pants will re-run the generator.

`st2-generate-api-spec` is a python script, and we use pants' python dependency inference. So pants (effectively) also adds the python files imported by `st2-generate-api-spec` to the cache key. Stepping through the code, we can see what pants would infer:

https://github.com/StackStorm/st2/blob/1856f6108219ebb1a6886afb798fd0f6c97b184c/st2common/bin/st2-generate-api-spec#L18
https://github.com/StackStorm/st2/blob/1856f6108219ebb1a6886afb798fd0f6c97b184c/st2common/st2common/cmd/generate_api_spec.py#L22-L26
https://github.com/StackStorm/st2/blob/1856f6108219ebb1a6886afb798fd0f6c97b184c/st2common/st2common/util/spec_loader.py#L21-L26

Here we reach the few things that get interpolated into the Jinja Template:

https://github.com/StackStorm/st2/blob/1856f6108219ebb1a6886afb798fd0f6c97b184c/st2common/st2common/util/spec_loader.py#L30-L35

Now, if anyone changes any of these files, or any of the python bits that get imported to help define things in them, then the next time someone runs the `lint` goal (or in CI) they will find out that they need to run the `fmt` goal.

### Developing pants plugins

Pants has extensive documentation on the plugin API, including how to create targets, how to write rules, and there are even mini tutorials about how to add a linter or a formatter (adding formatters/linters is a very common thing plugins to do).

- [Add a linter](https://www.pantsbuild.org/docs/plugins-lint-goal)

### DIFF
--- a/pants-plugins/api_spec/rules.py
+++ b/pants-plugins/api_spec/rules.py
@@ -12,19 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from dataclasses import dataclass
-from textwrap import dedent
 
 from pants.backend.python.target_types import EntryPoint
 from pants.backend.python.util_rules.pex import (
-    PexRequest,
     VenvPex,
     VenvPexProcess,
 )
 from pants.backend.python.util_rules.pex_from_targets import PexFromTargetsRequest
-from pants.backend.python.util_rules.python_sources import (
-    PythonSourceFiles,
-    PythonSourceFilesRequest,
-)
 from pants.core.goals.fmt import FmtResult, FmtRequest
 from pants.core.goals.lint import LintResult, LintResults, LintTargetsRequest
 from pants.core.target_types import FileSourceField, ResourceSourceField
@@ -48,10 +42,7 @@ from pants.engine.target import (
 from pants.engine.unions import UnionRule
 from pants.util.logging import LogLevel
 
-from api_spec.target_types import (
-    APISpecSourceField,
-    APISpec,
-)
+from api_spec.target_types import APISpecSourceField
 
 
 GENERATE_SCRIPT = "generate_api_spec"
@@ -159,7 +150,7 @@ async def generate_api_spec_via_fmt(
                 "conf/st2.dev.conf",
             ),
             input_digest=input_digest,
-            description=f"Regenerating openapi.yaml api spec",
+            description="Regenerating openapi.yaml api spec",
             level=LogLevel.DEBUG,
         ),
     )
@@ -250,7 +241,7 @@ async def validate_api_spec(
                 # "--verbose",  # show definitions on failure
             ),
             input_digest=input_digest,
-            description=f"Validating openapi.yaml api spec",
+            description="Validating openapi.yaml api spec",
             level=LogLevel.DEBUG,
         ),
     )
--- a/pants-plugins/api_spec/rules.py
+++ b/pants-plugins/api_spec/rules.py
@@ -246,6 +246,8 @@ async def validate_api_spec(
             argv=(
                 "--config-file",
                 "conf/st2.dev.conf",
+                # "--validate-defs",  # check for x-api-model in definitions
+                # "--verbose",  # show definitions on failure
             ),
             input_digest=input_digest,
             description=f"Validating openapi.yaml api spec",
--- a/pants-plugins/api_spec/rules.py
+++ b/pants-plugins/api_spec/rules.py
@@ -1,0 +1,265 @@
+# Copyright 2021 The StackStorm Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from dataclasses import dataclass
+from textwrap import dedent
+
+from pants.backend.python.target_types import EntryPoint
+from pants.backend.python.util_rules.pex import (
+    PexRequest,
+    VenvPex,
+    VenvPexProcess,
+)
+from pants.backend.python.util_rules.pex_from_targets import PexFromTargetsRequest
+from pants.backend.python.util_rules.python_sources import (
+    PythonSourceFiles,
+    PythonSourceFilesRequest,
+)
+from pants.core.goals.fmt import FmtResult, FmtRequest
+from pants.core.goals.lint import LintResult, LintResults, LintTargetsRequest
+from pants.core.target_types import FileSourceField, ResourceSourceField
+from pants.core.util_rules.source_files import SourceFiles, SourceFilesRequest
+from pants.engine.addresses import Address
+from pants.engine.fs import (
+    CreateDigest,
+    Digest,
+    FileContent,
+    MergeDigests,
+    Snapshot,
+)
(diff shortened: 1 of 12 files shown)


pants-plugins/api_spec/rules.py
```diff
@@ -0,0 +1,259 @@
+# Copyright 2021 The StackStorm Authors.
```
amanda11, line 1: Was this really created in 2021?

pants-plugins/api_spec/rules.py
```diff
@@ -0,0 +1,259 @@
...
+    )
+
+    dependency_files_get = Get(
+        SourceFiles,
+        SourceFilesRequest(
+            sources_fields=[
+                tgt.get(SourcesField) for tgt in transitive_targets.dependencies
+            ],
+            for_sources_types=(FileSourceField, ResourceSourceField),
+            # enable_codegen=True,
```
amanda11, line 100: Wasn't sure why we had this commented out line?

pants-plugins/api_spec/rules.py
```diff
@@ -0,0 +1,259 @@
...
```
amanda11, line 196: Similar to above, why do we have this commented out line?

pants-plugins/api_spec/rules.py
```diff
@@ -0,0 +1,259 @@
...
+    )
+
+    process_result = await Get(
+        FallibleProcessResult,
+        VenvPexProcess(
+            pex,
+            argv=(
+                "--config-file",
+                "conf/st2.dev.conf",
+                # "--validate-defs",  # check for x-api-model in definitions
```
amanda11, line 239: Do we want to use the commented out values? Or are these just to show extra options we could use?

pants-plugins/api_spec/rules_test.py
```diff
@@ -0,0 +1,267 @@
...
+    api_spec_dir: str,
+    api_spec_file: str,
+    contents: str,
+    rc: int,
+    rule_runner: RuleRunner,
+) -> None:
+    files = {
+        f"{api_spec_dir}/{api_spec_file}": contents,
+        f"{api_spec_dir}/BUILD": f"api_spec(name='t', source='{api_spec_file}')",
+        # add in the target that's hard-coded in the generate_api_spec_via_fmt rue
```
amanda11, line 219: typo "rue" -> "rule"?

pants-plugins/api_spec/rules.py
cognifloyd: Yeah. I wrote this plugin in 2021. I can update the header though because it will be merged in 2023

pants-plugins/api_spec/rules.py
cognifloyd: Oh. I'll add an explanatory comment. These are things we might want to add in the future. We can't use `--validate-defs` yet, because there are a lot (hundreds?) of failures right now. The Makefile also does not run this yet.

cognifloyd:
@amanda11 Thanks for the feedback. I think I've addressed your comments.

amanda11 (approved):
LGTM

pants-plugins/api_spec/rules.py
```diff
@@ -0,0 +1,259 @@
...
+            [
+                Address(
+                    CMD_DIR,
+                    target_name="cmd",
+                    relative_file_path=f"{GENERATE_CMD}.py",
+                ),
+            ],
+            output_filename=f"{GENERATE_CMD}.pex",
+            internal_only=True,
+            main=EntryPoint.parse(f"{CMD_MODULE}.{GENERATE_CMD}:main"),
```
cognifloyd, line 122: We are using `st2common.cmd.generate_api_spec:main` directly instead of using the `st2-generate-api-spec` script because pants can't run python with a `-` in the name (because pants relies on importing from the target script, and python cannot import from files with a `-` in the name).

Eventually I'll get around to plumbing a change in pants to make that possible (it already works with the underlying pex, but pants hasn't learned about that arg yet). In the meantime, we just use the entrypoint itself instead of the script.

pants-plugins/api_spec/rules.py
```diff
@@ -0,0 +1,259 @@
...
+            [
+                Address(
+                    CMD_DIR,
+                    target_name="cmd",
+                    relative_file_path=f"{VALIDATE_CMD}.py",
+                ),
+            ],
+            output_filename=f"{VALIDATE_CMD}.pex",
+            internal_only=True,
+            main=EntryPoint.parse(f"{CMD_MODULE}.{VALIDATE_CMD}:main"),
```
cognifloyd, line 217: The same thing applies here. We are using `st2common.cmd.validate_api_spec:main` directly instead of using the `st2-validate-api-spec` script directly.

rush-skills (approved):
LGTM

pants-plugins/api_spec/rules.py
cognifloyd: Having a command with `-` is fairly common, so I don't think that's a problem.

Pex (used by pants) already had everything it needed to read files without importing them, but that didn't have a user facing API. So, when I asked about it, John (pex maintainer) quickly added an `--executable` option, so now pex supports it. But, pants hasn't learned about that pex option yet, so it doesn't use it.

I started adding the bits here, but haven't had time to dive farther than this: https://github.com/pantsbuild/pants/compare/main...cognifloyd:pants:pex-exe-arg

PR #5858, opened 2023-01-05
Refactor `DEFAULT_LOGGING_CONF_PATH` calculation to allow fine-grained pants dep inference

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

This is primarily a follow-up to:
- #5846

Unlike the simple BUILD config added in #5846, this PR refactors one constant so that pants can better use python imports to precisely detect who needs `st2common/st2common/conf/base.logging.conf`.

Before this PR, anything that uses any of the constants had a hard-coded dependency on `base.logging.conf`:

https://github.com/StackStorm/st2/blob/63f66b9595bf5b4e3974e7c5f327fa59f864ba3d/st2common/st2common/constants/BUILD#L1-L3

After this PR, each file will only get an inferred dependency if it actually imports the `DEFAULT_LOGGING_CONF_PATH` constant that refers to that file.

This also has the benefit of moving the constant calculation closer to the file it references. But, it still maintains backwards compatibility by also making the constant available in the old location.

### DIFF
--- a/st2common/st2common/conf/__init__.py
+++ b/st2common/st2common/conf/__init__.py
@@ -1,5 +1,4 @@
-# Copyright 2020 The StackStorm Authors.
-# Copyright 2019 Extreme Networks, Inc.
+# Copyright 2021 The StackStorm Authors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


st2common/st2common/conf/__init__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright 2021 The StackStorm Authors.
```
amanda11, line 1: Was this file really created in 2021?

cognifloyd: Yes I originally committed this on Jun 12 2021. But, since it's getting merged in 2023, I've updated the copyright.

amanda11 (approved):
LGTM

rush-skills (approved):

PR #5860, opened 2023-01-05
Add `pants-plugins/sample_conf` to streamline regenerating `conf/st2.conf.sample`

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

This PR improves the DX (developer experience) by wiring up the `fmt` and `lint` goals to generate `conf/st2.conf.sample` if needed (or for lint, warn that it needs to be done).

This includes creating a `sample_conf` plugin for pants that adds a `sample_conf` target.

### Developer Experience

Today, if someone changes one of the files used to generate `conf/st2.conf.sample`, they will probably forget to run the Makefile target the regenerates them: `make configgen`.

I've missed running that several times. CI complains as part of the `ci-checks` Makefile target, with instructions to `Please run "make configgen"`. This also gets regenerated any time someone runs the `lint` or `tests` Makefile targets, which might be surprising (tests generally shouldn't handle codegen).

With this PR, we add `st2.conf.sample` generation into the `fmt` goal. This also adds it to the `lint` goal which will tell people if `fmt` needs to run to update any code (or in this case, regenerate the sample conf file). Then, we only need simple instructions that say something like:

> Please run `./pants fmt lint ::` before submitting your PR. This will reformat code, if needed, and warn you about any other errors.

After someone runs `./pants fmt ::` once, they will benefit from the pants' results cache. For the sample_conf plugin, here's what that means:

First, the `sample_conf` target in `conf/` depends on `tools/config_gen.py`:

https://github.com/StackStorm/st2/blob/1faea8f17a9b6b55c21218df8f7d999700338826/conf/BUILD#L6-L9

If the `config_gen` script changes, then pants now knows that the sample conf file need to be regenerated. Or, in other words, the `config_gen.py` file is part of the cache key for `conf/st2.conf.sample`. If that file changes, then pants will re-run the generator.

For many of our scripts, we can depend on pants' python dependency inference to grab the other files. But, this script uses importlib instead of standard imports, so we need to add the dependencies manually (_aside: there is a feature that tries to use strings for dep inference, but it is not enabled by default, and I have not tested to see how many false positives it will find across our repo_). In the future, hopefully we can remove or simplify this.

https://github.com/StackStorm/st2/blob/1faea8f17a9b6b55c21218df8f7d999700338826/tools/BUILD#L1-L24

Please note the explicit dependency on `//:auth_backends`. Nothing in our code base directly imports from the auth backends. But, we ship with a default set of backends, so we need to make sure that their settings are included in the sample conf file's auth section. This is defined here:

https://github.com/StackStorm/st2/blob/1faea8f17a9b6b55c21218df8f7d999700338826/BUILD#L43-L49

Then, I ran into one more dependency issue. `st2-auth-ldap` does not declare a dependency on our `st2auth` module, in part because that is an implicit dep for auth modules. We should probably make that explicit at some point, but that might require publishing more wheels on pypi. That's messy, but pants meets us where we are in this mess. 😄 So, we can tell pants that any time we grab the `st2-auth-ldap` package, it requires something in the `st2auth` module, like this:

https://github.com/StackStorm/st2/blob/1faea8f17a9b6b55c21218df8f7d999700338826/BUILD#L34-L39

Then, once all of those deps that pants can't infer (for now) are recorded, pants is able to invalidate the st2.conf.sample cache if any of the dependency python source files (including all that get imported directly or transitively by `tools/config_gen.py`).

This PR actually only wires up the `fmt` goal and took advantage of the fact that pants also runs formatters whenever it runs linters, which are side-effect free. And because pants runs the formatters in a sandbox, it doesn't have to materialize any changes during lint.

Continuing the example, since pants cached the results when running the `lint` goal, running `fmt` will be very fast. Pants will just materialize the already-generated schemas from its cache.

### Developing pants plugins

### DIFF
--- a/pants-plugins/sample_conf/rules.py
+++ b/pants-plugins/sample_conf/rules.py
@@ -85,7 +85,7 @@ async def generate_sample_conf_via_fmt(
 
     # actually generate it with an external script.
     # Generation cannot be inlined here because it needs to import the st2 code.
-    pex_get = Get(
+    pex = await Get(
         VenvPex,
         PexFromTargetsRequest(
             [Address("tools", target_name="tools", relative_file_path=f"{SCRIPT}.py")],
@@ -94,11 +94,6 @@ async def generate_sample_conf_via_fmt(
             main=EntryPoint(SCRIPT),
         ),
     )
-    sources_get = Get(
-        PythonSourceFiles,
-        PythonSourceFilesRequest(transitive_targets.closure, include_files=True),
-    )
-    pex, sources = await MultiGet(pex_get, sources_get)
 
     result = await Get(
         FallibleProcessResult,
--- a/pants-plugins/sample_conf/rules.py
+++ b/pants-plugins/sample_conf/rules.py
@@ -13,9 +13,8 @@
 # limitations under the License.
 from dataclasses import dataclass
 
-from pants.backend.python.target_types import EntryPoint, PythonSourceField
+from pants.backend.python.target_types import EntryPoint
 from pants.backend.python.util_rules.pex import (
-    Pex,
     PexRequest,
     VenvPex,
     VenvPexProcess,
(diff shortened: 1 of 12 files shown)


pants-plugins/sample_conf/rules.py
```diff
@@ -0,0 +1,113 @@
+# Copyright 2021 The StackStorm Authors.
```
amanda11, line 1: Did you really write this file in 2021? Or should copyright be 2022 or 2023?

cognifloyd: Hmm. Yeah, that was copy pasta. Looks like I wrote this plugin in June of 2022. I've updated it to 2023 because that's when it will be merged.

amanda11 (approved):
LGTM

PR #5864, opened 2023-01-09
Add `pants-plugins/uses_services` to check before running tests for required services (like mongo)

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

This PR improves the DX (developer experience) by failing as early as possible if the development (or CI) environment does not have the required services.

This PR only checks for `mongo`, but I have draft branches to also add [`rabbitmq`](https://github.com/StackStorm/st2/tree/pants-plugins-uses_services-rabbitmq) and [`redis`](https://github.com/StackStorm/st2/tree/pants-plugins-uses_services-redis).

### Developer Experience

I do local development on my linux and mac laptops. Others use vagrant. Wherever you have your st2 sources, you must have mongo running in order to run some of our tests. I don't normally run mongo, so I often forget to start it before running tests. I would rather the tests bail out if the service is not running instead of giving me tons of logs to sift through to realize that mongo isn't running.

So, this PR focuses on failing as early as possible if the required services aren't running. On failure, the error message should be as actionable as possible to help guide new contributors to do the right thing without having to search our docs or ask for help in slack.

This does not run over and over. Pants has a daemon, `pantsd`, that runs in the background to handle results caching of the various test runs. The `is_mongo_running` check will re-run whenever pantsd restarts.

### Developing pants plugins

## Architecture of the `uses_services` plugin

contents in this section:
- `uses` field
- raise `ServiceMissingError` when the service is not running
- `Platform` dataclass and the rule and `inspect_platform.py` script that creates it
- `is_mongo_running.py` script and the rule that runs it
- wiring up the `test` goal so it runs `is_mongo_running` when `pytest` runs on a target with the `uses` field.
- testing the rules and scripts

### `uses` field

In BUILD files, we can add the `uses` field like this to say that these tests "use" mongo (so mongo should be running):
https://github.com/StackStorm/st2/blob/89b6960ff35f5d86a630da9310494704ced66a37/st2common/tests/unit/services/BUILD#L1-L4

This field is defined here:
https://github.com/StackStorm/st2/blob/89b6960ff35f5d86a630da9310494704ced66a37/pants-plugins/uses_services/target_types.py#L20-L23

And gets added to the `python_test` and `python_tests` targets here:
https://github.com/StackStorm/st2/blob/89b6960ff35f5d86a630da9310494704ced66a37/pants-plugins/uses_services/register.py#L25-L26

### raise `ServiceMissingError` when the service is not running

When a target has the `uses` field, some rules (described below) handle checking for running services like mongo. When the service is not running, the rule raises a `ServiceMissingError` which is defined here:

https://github.com/StackStorm/st2/blob/89b6960ff35f5d86a630da9310494704ced66a37/pants-plugins/uses_services/exceptions.py#L38-L51

Wherever possible, the exception should include general instructions to help guide the developer towards getting the service running. Those instructions are platform-specific (eg development on Linux vs on Mac, or on Ubuntu vs on Rocky). So, the error needs to now the `service` name (eg `mongo`), a `Platform` descriptor dataclass (described below), and the instructions to show.

As I wrote the instructions for `mongo`, `rabbitmq`, and `redis`, I found that I was basically copy/pasting the same message, which meant that if I wanted to reword something, I had to change 3 copies of it, which is not pleasant. So, I wrote a little util, `ServiceMissingError.generate()` that generates an error with standardized instructions. The `platform` argument is the same as what the `ServiceMissingError` requires (described below) and the `messages` argument takes a dataclass with all of the strings that need to be interpolated in those standard instructions. That class is defined here:

https://github.com/StackStorm/st2/blob/89b6960ff35f5d86a630da9310494704ced66a37/pants-plugins/uses_services/exceptions.py#L22-L35

Here is what raising the error looks like for `mongo`:

https://github.com/StackStorm/st2/blob/89b6960ff35f5d86a630da9310494704ced66a37/pants-plugins/uses_services/mongo_rules.py#L161-L196

Please review the messages that these strings get interpolated into in [`exceptions.py`](https://github.com/StackStorm/st2/pull/5864/files#diff-30a80288bf0cf04d5c342a17a72d233da7c444af7918577dcc52c8158147f0cd). These are effectively documentation to help developers have a better experience.

### `Platform` dataclass and the rule  script that creates it

Here is the platform dataclass:

https://github.com/StackStorm/st2/blob/89b6960ff35f5d86a630da9310494704ced66a37/pants-plugins/uses_services/scripts/inspect_platform.py#L21-L32

[`inspect_platform.py`](https://github.com/StackStorm/st2/pull/5864/files#diff-0b20e2d88b7dbfe97b0db4206767fb185fbf38f88d0eea1f83d11e2defe42d01) is a simple script that collects the data about the running arch, os, and distro.

The rule that runs this script to collect the `Platform` details is defined in [`platform_rules.py`](https://github.com/StackStorm/st2/pull/5864/files#diff-e306be478edba8cea113b9b367c724b5111b50b67a5257b8b2275876bf39928b)

Normally, direct file access is a bad idea in pants rules. However, after experimentation, that turned out to be the cleanest/quickest way to handle loading a python file that is part of the plugin. In most rules, pants keeps track of which files that rule needs so it can watch for changes and restart the rule if that file changes. Because the file in question is part of the plugin that needs to use it, pants is already watching the file, so using a pants `Get` request is superfluous. So the rule grabs the contents of the script like this:
https://github.com/StackStorm/st2/blob/89b6960ff35f5d86a630da9310494704ced66a37/pants-plugins/uses_services/platform_rules.py#L44-L47

Then it runs that script in a pex, which gets cached once per session (which is as long as `pantsd` is still running) according to this:
https://github.com/StackStorm/st2/blob/89b6960ff35f5d86a630da9310494704ced66a37/pants-plugins/uses_services/platform_rules.py#L71-L72

And then creates the `Platform` here:
https://github.com/StackStorm/st2/blob/89b6960ff35f5d86a630da9310494704ced66a37/pants-plugins/uses_services/platform_rules.py#L76-L77

### `is_mongo_running.py` script and the rule that runs it

Finally, [`is_mongo_running.py`](https://github.com/StackStorm/st2/pull/5864/files#diff-bd6fb6155a7e5955f228270307ef8f98feb2e8e077f18157da25aacf98abc88a) is the part that checks to see if mongo is running.

The pants plugin cannot import anything from our other `st2*` code, so `is_mongo_running.py` is a minimal self-contained script that mirrors how st2 connects to mongo. It should be as minimal as possible so that keeping it up-to-date with the core st2 code is not onerous.

The `is_mongo_running.py` rule gets opened and run in a pex with the same rule logic as `inspect_platform.py` (see above).

I avoided handling any kind of mongo auth; instead I had it use a mongo command that does not require auth, but still guarantees that mongo is actually running and accessible. For now, all of the tests hard-code the credentials and mongo connection settings like ssl. I do not expect this script to support auth in the future, but we might need to add TLS and other connection settings so we can support whatever mongo instance people want to use to run tests when developing locally. For now, it only supports the database connection settings hard-coded in the tests. Here is where the rule defines the default connection settings:

https://github.com/StackStorm/st2/blob/89b6960ff35f5d86a630da9310494704ced66a37/pants-plugins/uses_services/mongo_rules.py#L46-L68

Here is the definition of the rule that runs `is_mongo_running.py`:
https://github.com/StackStorm/st2/blob/89b6960ff35f5d86a630da9310494704ced66a37/pants-plugins/uses_services/mongo_rules.py#L116-L118

So, when another rule Gets `MongoIsRunning` with a `UsesMongoRequest`, pants will also run the rule that generates `Platform` (described above), and then it will run this `is_mongo_running` rule.

The `is_mongo_running` rule either returns `MongoIsRunning()` if it is running, or raises `ServiceMissingError` if it is not. By raising an error, this actually breaks a convention for pants rules. Exceptions stop everything and get shown to the user right away, and for most goals, pants wants to see some dataclass returned that has something like a `succeeded` boolean instead. But, we want to error as early as possible, so this breaks that convention on purpose.

### wiring up the `test` goal so it runs `is_mongo_running` when `pytest` runs on a target with the `uses` field.

The last piece that ties this all together is a rule that makes sure the `is_mongo_running` rule runs before `pytest` runs (if pants is running it on a target with the `uses` field). Here is the definition of the `mongo_is_running_for_pytest` rule:

https://github.com/StackStorm/st2/blob/89b6960ff35f5d86a630da9310494704ced66a37/pants-plugins/uses_services/mongo_rules.py#L89-L91

This rule is very simple. It just triggers running the `is_mongo_running` rule:
https://github.com/StackStorm/st2/blob/89b6960ff35f5d86a630da9310494704ced66a37/pants-plugins/uses_services/mongo_rules.py#L106-L109

This rule needs the `PytestUsesMongoRequest` which selects targets that have the `uses` field.

https://github.com/StackStorm/st2/blob/89b6960ff35f5d86a630da9310494704ced66a37/pants-plugins/uses_services/mongo_rules.py#L76-L82

https://github.com/StackStorm/st2/blob/89b6960ff35f5d86a630da9310494704ced66a37/pants-plugins/uses_services/mongo_rules.py#L202

### testing the rules and scripts

I collected a set of `Platform` samples from a variety of machines, including the 4 OSes we officially support, for use in tests. We will need them for several tests, so I put them in [`data_fixtures.py`](https://github.com/StackStorm/st2/pull/5864/files#diff-c1bc5e0f229648d7ac19a0e5ba9542d0032e34cbc7ee48c16d2fe337d16db487)

Then there are some tests that test running the scripts as part of testing the rules that run them.

There is at least one test for every new pants rule.


rush-skills:
@cognifloyd Is this ready for review? One test seems to be still failing

pants-plugins/uses_services/exceptions.py
```diff
@@ -0,0 +1,30 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from uses_services.platform_rules import Platform
+
+
+class ServiceMissingError(Exception):
+    """Error raised when a test uses a service but that service is missing."""
+
+    # TODO add special platform handling to DRY instructions across services
```
cognifloyd, line 20: Before I mark this ready for review, I'm looking into finishing this TODO.

pants-plugins/uses_services/mongo_rules.py
```diff
@@ -0,0 +1,283 @@
...
+
+    if is_running:
+        return MongoIsRunning()
+
+    # mongo is not running, so raise an error with instructions.
+
+    if platform.distro in ["centos", "rhel"] or "rhel" in platform.distro_like:
+        instructions = dedent(
+            """\
+            If mongo is installed, but not running try:
```
cognifloyd, line 166: All of the instructions here will need more careful analysis, as they are essentially documentation.

pants-plugins/uses_services/mongo_rules_test.py
```diff
@@ -0,0 +1,170 @@
...
+        # platform(
+        #    arch="x86_64",
+        #    os="Linux",
+        #    distro="",
+        #    distro_name="",
+        #    distro_codename="",
+        #    distro_like="",
+        #    distro_major_version="",
+        #    distro_version="",
+        # ),
```
cognifloyd, line 112: I need more samples. I want to grab at least one from each of our target platforms.

cognifloyd (commented):
This needs some more work. There are a few bits I could use some help on (see the comment about getting platform samples).

This plugin is not necessary to release wheels. It is strictly to improve Developer Experience. So, I'm working on a `pack_metadata` plugin which is needed for that. I'll come back to this plugin later.

Also, I had a lot going on this past week so I wasn't able to work on the pants integration much. Hopefully I can wrap up the branches for the next 2 plugins this week.

pants-plugins/uses_services/BUILD
```diff
@@ -0,0 +1,16 @@
...
+    sources=["data_fixtures.py"],
+)
+
+python_tests(
+    name="tests",
+    overrides={
+        # We use this plugin to validate we can run the tests for it.
+        "mongo_rules_test.py": {"uses": ["mongo"]},
+    },
+)
```
cognifloyd, line 16: It's kind of circular, but I declared that the `is_mongo_running` test uses mongo. That means that the plugin is used before the plugin is tested. So, if the plugin is broken, we won't even get the test results.

I'm leaning towards removing or commenting this out. Thoughts?

amanda11: I don't like idea that if it fails we won't know why..

cognifloyd: Thanks for validating my suspicion. I've commented this out with a warning about why it should not be used.

cognifloyd:
rebased

amanda11 (approved):

rush-skills (approved):

PR #5871, opened 2023-01-23
Pants pack metadata

cognifloyd:
### Background

This is another part of introducing [pants](https://www.pantsbuild.org/docs), as discussed in various TSC meetings.

Related PRs can be found in:
- [pantsbuild](https://github.com/StackStorm/st2/milestone/47) milestone (which includes PRs since #5713)
- https://github.com/StackStorm/st2/labels/pantsbuild label (which also includes preparatory PRs that came before adding pantsbuild)

### Overview of this PR

This PR adds `pack_metadata` targets, introduced in #5868 to our BUILD files.

If someone runs `./pants tailor ::`, it will automatically add `pack_metadata` targets in any directory with a `pack.yaml` file, so when we add new fixtures, it should not be much work to keep the metadata up-to-date.

There are several fixture packs that don't have a `pack.yaml` file, so I added those targets manually. Also, for the fixtures, I made the python fixture depend on the pack metadata, so that pants will include all of a pack's metadata files if a test depends on that fixture.

I also added one `pack_metadata_in_git_submodule` target, which I described in #5868 as:
> One of our packs is in a git submodule (`st2tests/st2tests/fixtures/packs/test_content_version`), so [this PR] also adds a `pack_metadata_in_git_submodule` target to cleanly handle a case where someone has not checked out the git submodules. This improves the Developer Experience by raising an error as early as possible to let people know that they need to checkout the git submodule and how to do that. Otherwise, they won't get an error until they run tests where we have some code to give the same instructions for the tests that need it.

Here is that target (sources have to be explicitly defined because they target is defined in a parent directory, not alongside the `pack.yaml` file):
https://github.com/StackStorm/st2/blob/26132d118546ee4df5bf0c819cb19e661616cf78/st2tests/st2tests/fixtures/packs/BUILD#L6-L14

And here is what that error message looks like in practice:
```
jafloyd@MB-C02XKFC0JG5H st2 % ./pants tailor update-build-files ::     
14:08:25.96 [INFO] Initializing scheduler...
14:08:26.30 [INFO] Scheduler initialized.
14:08:26.82 [WARN] Unmatched glob from st2tests/st2tests/fixtures/packs:test_content_version_shell's `sources` field: "st2tests/st2tests/fixtures/packs/test_content_version/**/*.sh"

[...]

PR #5874, opened 2023-01-27
Tell pants about tests that use globs to test with many pack fixtures

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

This PR:
- adds a `packs_glob` target (in `pants-plugins/pack_metadata`) with a rule that "infers" dependencies (like a `*` glob) on all of the subdirectories (unless ignored in the `dependencies` field).
- creates a python fixture file in `st2tests/st2tests/fixtures/packs/all_packs_glob` for tests that use globs to discover a bunch of packs. This fixture depends on the `packs_glob(name="all_packs")` target to facilitate the glob behavior.

This PR is basically a follow-up to the "Use python imports to identify fixtures" PRs: #5699, #5702, #5703, #5704, #5705, and #5706. But, instead of including these fixtures in that series of PRs, I waited to add this until after the `pack_metadata` targets were added in #5871.


amanda11 (commented):
I am worried that we are going to forget to do this when we add a new test pack.
I am just wondering if we can either discover these or if we need to have a comment in the current test packs that exist that they are registered here. So that if someone copies an existing pack they will see the message that they also have to add here.

cognifloyd:
Yes. I agree. I already stumbled over that when packs were added after I added this. So, I want a better solution than just comments.

Dependencies are not allowed to use globs because things like `import *` are generally a bad idea and make dependency inference much more complicated.

But, I can probably add a new pants plugin, or perhaps extend the `pack_metadata` plugin, with a `packs_glob` target. Then I would inject dependencies on all the globbed pack directory. 🤔


Yeah. I'll add it to `pack_metadata` and see how that works.

cognifloyd:
Upon closer inspection, I realized I don't understand the one test that would use the `child_packs` fixture, and it's not really using a glob like the `all_packs` one, so I removed it.

Also, I updated `target(name="all_packs")` to use be a new `packs_glob` target instead. This way we don't have to manually register all of the packs. As soon as I document that in the PR description, I will mark this ready for review again.

cognifloyd:
This is ready for review now.

st2tests/st2tests/fixtures/packs_1/BUILD
```diff
@@ -0,0 +1,3 @@
+packs_glob(
+    name="all_packs",
+)
```
cognifloyd, line 3: Nothing depends on this (and the one in packs_invalid) directly right now. We will need both to include all the packs in the wheel.

rush-skills (approved):

amanda11 (approved):

PR #5884, opened 2023-02-03
Update `pants-plugins/uses_services` to support checking for rabbitmq

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

This PR builds on #5864 to further improve the DX (developer experience) by failing as early as possible if the development (or CI) environment does not have the required services.

This PR adds checks for `rabbitmq`, like #5864 added `mongo` checks. I also have draft branch to add [`redis`](https://github.com/StackStorm/st2/tree/pants-plugins-uses_services-redis) checks.

Please see #5864 for a more detailed description of `pants-plugins/uses_services`.

### `is_rabbitmq_running.py` script and the rule that runs it

[`is_rabbitmq_running.py`](TODO) is the part that checks to see if rabbitmq is running.

The pants plugin cannot import anything from our other `st2*` code, so `is_rabbitmq_running.py` is a minimal self-contained script that mirrors how st2 connects to rabbitmq. It should be as minimal as possible so that keeping it up-to-date with the core st2 code is not onerous.

The `is_rabbitmq_running.py` rule gets opened and run in a pex with the same rule logic as `is_mongo_running.py` and `inspect_platform.py` (see #5864).

Rabbitmq embeds all auth stuff in the connection url. So, there are not as many settings to worry about as there were with mongo. For now, this only supports the url hard-coded in the tests. Here is where the rule defines the default url:
 
https://github.com/StackStorm/st2/blob/adb9c4b08e50b943f65a925c0f93ceff0d907798/pants-plugins/uses_services/rabbitmq_rules.py#L47-L65

Here is the definition of the rule that runs `is_rabbitmq_running.py`:
https://github.com/StackStorm/st2/blob/adb9c4b08e50b943f65a925c0f93ceff0d907798/pants-plugins/uses_services/rabbitmq_rules.py#L99-L101

So, when another rule Gets `RabbitMQIsRunning` with a `UsesRabbitMQRequest`, pants will also run the rule that generates `Platform` (described in #5864), and then it will run this `is_rabbitmq_running` rule.

The `is_rabbbitmq_running` rule either returns `RabbitMQIsRunning()` if it is running, or raises `ServiceMissingError` if it is not. By raising an error, this actually breaks a convention for pants rules. Exceptions stop everything and get shown to the user right away, and for most goals, pants wants to see some dataclass returned that has something like a `succeeded` boolean instead. But, we want to error as early as possible, so this breaks that convention on purpose.

### wiring up the `test` goal so it runs `is_rabbitmq_running` when `pytest` runs on a target with the `uses` field.

The last piece that ties this all together is a rule that makes sure the `is_rabbitmq_running` rule runs before `pytest` runs (if pants is running it on a target with the `uses` field). Here is the definition of the `rabbitmq_is_running_for_pytest` rule:

https://github.com/StackStorm/st2/blob/adb9c4b08e50b943f65a925c0f93ceff0d907798/pants-plugins/uses_services/rabbitmq_rules.py#L86-L88

This rule is very simple. It just triggers running the `is_rabbitmq_running` rule:
https://github.com/StackStorm/st2/blob/adb9c4b08e50b943f65a925c0f93ceff0d907798/pants-plugins/uses_services/rabbitmq_rules.py#L189-L92

This rule needs the `PytestUsesRabbitMQRequest` which selects targets that have the `uses` field.

https://github.com/StackStorm/st2/blob/adb9c4b08e50b943f65a925c0f93ceff0d907798/pants-plugins/uses_services/rabbitmq_rules.py#L73-L79

https://github.com/StackStorm/st2/blob/adb9c4b08e50b943f65a925c0f93ceff0d907798/pants-plugins/uses_services/rabbitmq_rules.py#L176

### DIFF
--- a/pants-plugins/uses_services/rabbitmq_rules.py
+++ b/pants-plugins/uses_services/rabbitmq_rules.py
@@ -62,7 +62,7 @@ class UsesRabbitMQRequest:
 
     #   with our version of oslo.config (newer are slower) we can't directly override opts w/ environment variables.
 
-    mq_urls: list[str] = ["amqp://guest:guest@127.0.0.1:5672//"]
+    mq_urls: tuple[str] = ("amqp://guest:guest@127.0.0.1:5672//",)
 
 
 @dataclass(frozen=True)
@@ -113,7 +113,7 @@ async def rabbitmq_is_running(
             PexRequest(
                 output_filename="kombu.pex",
                 internal_only=True,
-                requirements=PexRequirements({"kombu",}),
+                requirements=PexRequirements({"kombu"}),
             ),
         ),
     )
@@ -124,7 +124,7 @@ async def rabbitmq_is_running(
             kombu_pex,
             argv=(
                 script_path,
-                "amqp://guest:guest@127.0.0.1:5672/",
+                *request.mq_urls,
             ),
             input_digest=script_digest,
             description="Checking to see if RabbitMQ is up and accessible.",
@@ -139,9 +139,9 @@ async def rabbitmq_is_running(
         return RabbitMQIsRunning()
 
     # rabbitmq is not running, so raise an error with instructions.
-    raise ServiceMissingError(
-        platform,
-        ServiceSpecificMessages(
+    raise ServiceMissingError.generate(
+        platform=platform,
+        messages=ServiceSpecificMessages(
             service="rabbitmq",
             service_start_cmd_el_7="service rabbitmq-server start",
             service_start_cmd_el="systemctl start rabbitmq-server",
--- a/pants-plugins/uses_services/rabbitmq_rules.py
+++ b/pants-plugins/uses_services/rabbitmq_rules.py
@@ -150,7 +150,10 @@ async def rabbitmq_is_running(
                 curl -sL https://packagecloud.io/install/repositories/rabbitmq/erlang/script.rpm.sh | sudo bash
                 curl -sL https://packagecloud.io/install/repositories/rabbitmq/rabbitmq-server/script.rpm.sh | sudo bash
                 sudo yum makecache -y --disablerepo='*' --enablerepo='rabbitmq_rabbitmq-server'
-                # Install erlang and RabbitMQ
+                # Check for any required version constraints in our docs:
+                # https://docs.stackstorm.com/latest/install/rhel{platform.distro_major_version}.html
+
+                # Install erlang and RabbitMQ (and possibly constrain the version)
                 sudo yum -y install erlang{'' if platform.distro_major_version == "7" else '-*'}
                 sudo yum -y install rabbitmq-server
                 # Don't forget to start rabbitmq-server.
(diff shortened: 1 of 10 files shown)


pants-plugins/uses_services/rabbitmq_rules.py
```diff
@@ -0,0 +1,178 @@
...
+            service="rabbitmq",
+            service_start_cmd_el_7="service rabbitmq-server start",
+            service_start_cmd_el="systemctl start rabbitmq-server",
+            not_installed_clause_el="this is one way to install it:",
+            install_instructions_el=dedent(
+                """\
+                # Add key and repo for erlang and RabbitMQ
+                curl -sL https://packagecloud.io/install/repositories/rabbitmq/erlang/script.rpm.sh | sudo bash
+                curl -sL https://packagecloud.io/install/repositories/rabbitmq/rabbitmq-server/script.rpm.sh | sudo bash
+                sudo yum makecache -y --disablerepo='*' --enablerepo='rabbitmq_rabbitmq-server'
```
amanda11, line 154: Given the fun we have with RabbitMQ and the fact that we often have to choose specific versions, I wonder if in these commands we should put a comment to say to refer to the appropriate StackStorm manual installation instructions for your release on ho to install RabbitMQ locally.
That way if we change we don't need to update the instructions, we can just let them to read our manual instructions.

cognifloyd: I looked at that, but the scripts in our docs combine installing all of the dependent services: Mongo, rabbitmq, and redis ... So, it's hard to call out just what's required for any one service.

For deb (below), the instructions were so complex (both upstream and in our manual install), I opted to just link to the upstream docs. Plus, the upstream had a copy/pastable quickstart script in the docs that makes the install fairly simple.

Sadly, they don't provide an equivalent quickstart script for rpm. But we could still link to the section of the docs: https://www.rabbitmq.com/install-rpm.html#cloudsmith

So, this is basically the instructions extracted from our docs, filtered down to only deal with rabbitmq. Is there a cleaner place to link to instead of embedding this here?

amanda11: YEah you are right we don't have the dependencies separated. Maybe a comment stating to check our documentation in case we put any constraints on rabbitmq or erlang versioning.

e.g. if we move back to needing to constrain the erlang version again, then we're going to remember to change the st2docs and the automated installs, but not this file.

pants-plugins/uses_services/rabbitmq_rules.py
```diff
@@ -0,0 +1,176 @@
...
+                # Install erlang and RabbitMQ
```
cognifloyd, line 153:
```suggestion
                # Check for any required version constraints in our docs:
                # https://docs.stackstorm.com/latest/install/rhel{platform.distro_major_version}.html

                # Install erlang and RabbitMQ (and possibly constrain the version)
```

@amanda11 wdyt? Would this work for you?

I used `latest/` in the doc string because these instructions are for st2 development.

amanda11: @cognifloyd Looked good thanks, have committed the suggestion.

amanda11 (approved):

rush-skills (approved):
LGTM.

PR #5890, opened 2023-02-06
pants: Add `st2_shell_sources_and_resources()` macro to `pants-plugins/macros.py`

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

This PR adds our first pants macro: `st2_shell_sources_and_resources()`

I'm planning to add 3 more macros to help with building our `python_distribution` targets (for buildling wheels) in future PRs. This just introduces the concept of macros.

### Macros

> Macros are useful to reduce boilerplate in BUILD files.
> ...
> Macros are defined in Python files that act like a normal BUILD file. They have access to all the symbols you normally have registered in a BUILD file, such as all of your target types.
>
> Macros cannot import other modules, just like BUILD files cannot have import statements.

https://www.pantsbuild.org/docs/macros

Our macro file is `pants-plugins/macros.py`, and we tell pants about it with `[GLOBAL].build_file_prelude_globs` in `pants.toml`:

https://github.com/StackStorm/st2/blob/eb6902f341f1cda9a4dc83be442ccabb38a4c2a1/pants.toml#L11

This PR only introduces one macro. The key parts of this macro function are `shell_sources(...)` and `resources(...)` - ie, that's where our macro adds the normal pants targets using the kwargs passed to it.

https://github.com/StackStorm/st2/blob/eb6902f341f1cda9a4dc83be442ccabb38a4c2a1/pants-plugins/macros.py#L16-L28

And then we use the macro just like a regular target in a BUILD file:

https://github.com/StackStorm/st2/blob/eb6902f341f1cda9a4dc83be442ccabb38a4c2a1/st2actions/bin/BUILD#L8-L12

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,7 @@ Added
   to pants' use of PEX lockfiles. This is not a user-facing addition.
   #5778 #5789 #5817 #5795 #5830 #5833 #5834 #5841 #5840 #5838 #5842 #5837 #5849 #5850
   #5846 #5853 #5848 #5847 #5858 #5857 #5860 #5868 #5871 #5864 #5874 #5884 #5893 #5891
+  #5890
   Contributed by @cognifloyd
 
 * Added a joint index to solve the problem of slow mongo queries for scheduled executions. #5805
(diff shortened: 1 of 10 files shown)


rush-skills (approved):

cognifloyd:
Rebased and added `pants-plugins/README.md` entry.

nzlosh (approved):

PR #5891, opened 2023-02-06
Add `pants-plugins/release` to handle wheel metadata generation

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

Generally, wheels have a `setup.py` file in them. Right now, we have hard-coded `setup.py` and a copy of `dist_utils.py`. These are spread out, and no one likes maintaining them. But, using pants, this becomes much easier.

Pants can auto-generate a `setup.py` for us whenever it builds a `python_distribution()` (ie a wheel or an sdist). And, pants makes it easy to centralize all of the common bits of metadata that need to go in all the wheels (like `author="StackStorm"` or our `project_urls`). This PR adds `pants-plugins/release` to take advantage of those features.

#### `python_distribution()` targets

This PR does not actually create the `python_distribution()` targets yet--that will happen in a follow-up PR. But, here is what a python_distribution looks like with some common options we will use:

```python
python_distribution(
    name="myrunner",
    dependencies=[...],
    provides=python_artifact(
        name="myrunner",
        version="1.2.3",
        ...
        scripts=[...],  # for our pre-made scripts
    ),
    entry_points={
        "st2common.runners.runner": {...},
        "console_scripts": {...},  # for auto-generated entry point scripts
    }
)
```

#### Pants generates `setup.py`

Everything in `provides=python_artifact(...)` becomes the keyword args that pants puts in the call to `setup(...)` in the generated `setup.py` file. `setup(...)` also gets kwargs from a few other places including the `entry_points=` kwarg of python_distribution, and a special plugin rule:

https://github.com/StackStorm/st2/blob/2266086654f8c06186f317d393443bd752d06da3/pants-plugins/release/rules.py#L88-L89

_NB: Only one plugin can provide the `SetupKwargs`._

To wire up this rule, I created a `StackStormSetupKwargsRequest` class, and wired it up with a `UnionRule`:

https://github.com/StackStorm/st2/blob/2266086654f8c06186f317d393443bd752d06da3/pants-plugins/release/rules.py#L165

That way, whenever pants goes to generate `setup.py`, it will use our plugins to inject additional kwargs in the `setup(...)` call of `setup.py`.

### Dynamically retrieving `version` and `long_description` kwargs

In our current `setup.py` + `dist_utils.py` setup, we retrieve the `version` kwarg from an `__init__.py` file like this (or for st2client we actually just import `__version__` from the `__init__.py` file):
https://github.com/StackStorm/st2/blob/2266086654f8c06186f317d393443bd752d06da3/st2common/setup.py#L36

The `get_version_string` function is defined in `dist_utils.py`:
https://github.com/StackStorm/st2/blob/2266086654f8c06186f317d393443bd752d06da3/st2common/dist_utils.py#L163-L174

So, we do something very similar in this plugin. The BUILD file specifies a `version_file` arg on `python_artifact(...)`, to say which `__init__.py` file to use. Then the plugin gets the contents of the file and extracts the version string, just like the `get_version_string` function from our old `dist_utils.py` file.

_NB: The `glob_match_error_behavior=GlobMatchErrorBehavior.error` bit means that if the version_file does not exist, an error will be raised when attempting to build the wheel._

https://github.com/StackStorm/st2/blob/2266086654f8c06186f317d393443bd752d06da3/pants-plugins/release/rules.py#L100-L108
https://github.com/StackStorm/st2/blob/2266086654f8c06186f317d393443bd752d06da3/pants-plugins/release/rules.py#L118-L125

In our current `setup.py` for the `st2client` wheel, we also include a `README.rst` file in the `long_description` kwarg. So, we do the same thing in this plugin as well:

https://github.com/StackStorm/st2/blob/2266086654f8c06186f317d393443bd752d06da3/pants-plugins/release/rules.py#L109-L115
https://github.com/StackStorm/st2/blob/2266086654f8c06186f317d393443bd752d06da3/pants-plugins/release/rules.py#L132-L137

### Centralized Wheel Metadata

And finally, we also define the metadata that is common to all of our wheels here:

https://github.com/StackStorm/st2/blob/2266086654f8c06186f317d393443bd752d06da3/pants-plugins/release/rules.py#L37-L66

Note that the classifiers get combined like this:

https://github.com/StackStorm/st2/blob/2266086654f8c06186f317d393443bd752d06da3/pants-plugins/release/rules.py#L151-L157

That means that the python version classifiers are also included. As the TODO says, we still need to pull out the versions from the `interpreter_constraints`--until then, we've just got them hard-coded.

### Developing pants plugins

- [`python_distribution()` target](https://www.pantsbuild.org/docs/reference-python_distribution)
- [Custom `python_artifact()` kwargs](https://www.pantsbuild.org/docs/plugins-setup-py)


st2client/setup.py
```diff
@@ -72,7 +72,7 @@
         "Repository": "https://github.com/StackStorm/st2",
         "Documentation": "https://docs.stackstorm.com",
         "Community": "https://stackstorm.com/community-signup",
-        "Questions": "https://forum.stackstorm.com/",
+        "Questions": "https://github.com/StackStorm/st2/discussions",
```
cognifloyd, line 75: I also noticed that this URL in our old `st2client/setup.py` file was out-of date. Ideally we won't be using this setup.py file, but I went ahead and updated it here as well.

rush-skills (approved):

amanda11 (approved):

PR #5893, opened 2023-02-07
Update `pants-plugins/uses_services` to support checking for redis

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

This PR builds on #5864 and #5884 to further improve the DX (developer experience) by failing as early as possible if the development (or CI) environment does not have the required services.

This PR adds checks for `redis`, like #5864 added `mongo` checks and #5884 added `rabbitmq` checks.

Please see #5864 for a more detailed description of `pants-plugins/uses_services`.

### `is_redis_running.py` script and the rule that runs it

[`is_redis_running.py`](https://github.com/StackStorm/st2/pull/5893/files#diff-ba1edf54790a4a79ed0a370092f709cc15bf05d65ce0b62473d4d82211ba0839) is the part that checks to see if redis is running.

The pants plugin cannot import anything from our other `st2*` code, so `is_redis_running.py` is a minimal self-contained script that mirrors how st2 connects to redis. It should be as minimal as possible so that keeping it up-to-date with the core st2 code is not onerous.

The `is_redis_running.py` rule gets opened and run in a pex with the same rule logic as `is_mongo_running.py` and `inspect_platform.py` (see #5864).

Redis, like rabbitmq, embeds all auth stuff in the connection url. So, there are not as many settings to worry about as there were with mongo. For now, this only supports the url hard-coded in the tests. Here is where the rule defines the default url:
 
https://github.com/StackStorm/st2/blob/06aedfa0a1efa28d17affe6f114a7d6bac631b28/pants-plugins/uses_services/redis_rules.py#L47-L62

Here is the definition of the rule that runs `is_redis_running.py`:
https://github.com/StackStorm/st2/blob/06aedfa0a1efa28d17affe6f114a7d6bac631b28/pants-plugins/uses_services/redis_rules.py#L96-L98

So, when another rule Gets `RedisIsRunning` with a `UsesRedisRequest`, pants will also run the rule that generates `Platform` (described in #5864), and then it will run this `is_redis_running` rule.

The `is_redis_running` rule either returns `RedisIsRunning()` if it is running, or raises `ServiceMissingError` if it is not. By raising an error, this actually breaks a convention for pants rules. Exceptions stop everything and get shown to the user right away, and for most goals, pants wants to see some dataclass returned that has something like a `succeeded` boolean instead. But, we want to error as early as possible, so this breaks that convention on purpose.

### wiring up the `test` goal so it runs `is_redis_running` when `pytest` runs on a target with the `uses` field.

The last piece that ties this all together is a rule that makes sure the `is_redis_running` rule runs before `pytest` runs (if pants is running it on a target with the `uses` field). Here is the definition of the `redis_is_running_for_pytest` rule:

https://github.com/StackStorm/st2/blob/06aedfa0a1efa28d17affe6f114a7d6bac631b28/pants-plugins/uses_services/redis_rules.py#L83-L85

This rule is very simple. It just triggers running the `is_redis_running` rule:
https://github.com/StackStorm/st2/blob/06aedfa0a1efa28d17affe6f114a7d6bac631b28/pants-plugins/uses_services/redis_rules.py#L86-L89

This rule needs the `PytestUsesRedisRequest` which selects targets that have the `uses` field.

https://github.com/StackStorm/st2/blob/06aedfa0a1efa28d17affe6f114a7d6bac631b28/pants-plugins/uses_services/redis_rules.py#L70-L76

https://github.com/StackStorm/st2/blob/06aedfa0a1efa28d17affe6f114a7d6bac631b28/pants-plugins/uses_services/redis_rules.py#L168

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -54,6 +54,19 @@ jobs:
           - 5672:5672/tcp   # AMQP standard port
           - 15672:15672/tcp # Management: HTTP, CLI
 
+      redis:
+         # Docker Hub image
+         image: redis
+         # Set health checks to wait until redis has started
+         options: >-
+           --name "redis"
+           --health-cmd "redis-cli ping"
+           --health-interval 10s
+           --health-timeout 5s
+           --health-retries 5
+         ports:
+           - 6379:6379/tcp
+
     env:
       COLUMNS: '120'
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,7 +15,7 @@ Added
   working on StackStorm, improve our security posture, and improve CI reliability thanks in part
   to pants' use of PEX lockfiles. This is not a user-facing addition.
   #5778 #5789 #5817 #5795 #5830 #5833 #5834 #5841 #5840 #5838 #5842 #5837 #5849 #5850
-  #5846 #5853 #5848 #5847 #5858 #5857 #5860 #5868 #5871 #5864 #5874 #5884
+  #5846 #5853 #5848 #5847 #5858 #5857 #5860 #5868 #5871 #5864 #5874 #5884 #5893
   Contributed by @cognifloyd
 
 * Added a joint index to solve the problem of slow mongo queries for scheduled executions. #5805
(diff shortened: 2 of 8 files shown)


rush-skills (approved):
LGTM. Though is redis only required for `orquestra` tests and runner builds>

cognifloyd:
I don't know. They are needed at least for the orquesta tests. When we find other tests that need any of the services (mongo, rabbitmq, redis), then we can add `uses=` to the BUILD metadata for those tests too.

amanda11 (approved):

PR #5899, opened 2023-02-10
Upgrade pants from 2.14 to 2.15

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

This updates to pants to `2.15.0` which requires:
- regen the pants-plugins lockfile
- several changes to our `fmt` and `lint` plugins in pants-plugins.

The required plugin updates are described in the pants [Plugin upgrade guide](https://www.pantsbuild.org/v2.15/docs/plugin-upgrade-guide).

#### Notes about updating to pants 2.15.0

Here are some links to learn more about pants v2.15:
- https://blog.pantsbuild.org/pants-2-15/
- https://github.com/pantsbuild/pants/blob/main/src/python/pants/notes/2.15.x.md

NB: We must do updates 1 minor version of pants at a time so that pants `update-build-files` goal can handle any required updates before continuing. We want to go to 2.16, but the 2.15 update was significant enough that it deserves its own PR before jumping to 2.16.

_Aside: As soon as the `pants 2.16.0rc0` comes out, I want to switch to that one as we need some of the features in 2.16 like the new `stevedore` backend, and improvements to exporting virtualenvs based on lockfiles. That does not look like it will require extensive plugin upgrades like the 2.15 upgrade does._

### The Fmt/Lint plugin API changed in pants 2.15

It now requires a Subsystem with a SkipOption. I took advantage of this to consolidate the constants and all the address and pex request definitions in the Subsystem subclass.

It now requires 2 rules instead of 1: to partition, then fmt/lint. That means the structure of the rule requests had to change slightly. The tests also had to account for the modified rules.

_Refer to the [Plugin upgrade guide](https://www.pantsbuild.org/v2.15/docs/plugin-upgrade-guide) for more details._

### DIFF
--- a/pants-plugins/api_spec/rules.py
+++ b/pants-plugins/api_spec/rules.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 from dataclasses import dataclass
 
-from pants.backend.python.target_types import EntryPoint
 from pants.backend.python.util_rules import pex, pex_from_targets
 from pants.backend.python.util_rules.pex import (
     VenvPex,
@@ -21,10 +20,10 @@ from pants.backend.python.util_rules.pex import (
 )
 from pants.backend.python.util_rules.pex_from_targets import PexFromTargetsRequest
 from pants.core.goals.fmt import FmtResult, FmtTargetsRequest
-from pants.core.goals.lint import LintResult, LintResults, LintTargetsRequest
-from pants.core.target_types import FileSourceField, ResourceSourceField
+from pants.core.goals.lint import LintResult, LintTargetsRequest
+from pants.core.util_rules.config_files import ConfigFiles, ConfigFilesRequest
+from pants.core.util_rules.partitions import PartitionerType
 from pants.core.util_rules.source_files import SourceFiles, SourceFilesRequest
-from pants.engine.addresses import Address
 from pants.engine.fs import (
     CreateDigest,
     Digest,
@@ -34,24 +33,12 @@ from pants.engine.fs import (
 )
 from pants.engine.process import FallibleProcessResult, ProcessResult
 from pants.engine.rules import Get, MultiGet, collect_rules, rule
-from pants.engine.target import (
-    FieldSet,
-    SourcesField,
-    TransitiveTargets,
-    TransitiveTargetsRequest,
-)
-from pants.engine.unions import UnionRule
+from pants.engine.target import FieldSet
 from pants.util.logging import LogLevel
+from pants.util.strutil import strip_v2_chroot_path
 
+from api_spec.subsystem import GenerateApiSpec, ValidateApiSpec
 from api_spec.target_types import APISpecSourceField
-
-
-# these constants are also used in the tests
-CMD_SOURCE_ROOT = "st2common"
-CMD_DIR = "st2common/st2common/cmd"
-CMD_MODULE = "st2common.cmd"
-GENERATE_CMD = "generate_api_spec"
-VALIDATE_CMD = "validate_api_spec"
 
 
 @dataclass(frozen=True)
@@ -63,12 +50,14 @@ class APISpecFieldSet(FieldSet):
 
 class GenerateAPISpecViaFmtTargetsRequest(FmtTargetsRequest):
     field_set_type = APISpecFieldSet
-    name = GENERATE_CMD
+    tool_subsystem = GenerateApiSpec
+    partitioner_type = PartitionerType.DEFAULT_SINGLE_PARTITION
 
 
 class ValidateAPISpecRequest(LintTargetsRequest):
     field_set_type = APISpecFieldSet
-    name = VALIDATE_CMD
+    tool_subsystem = ValidateApiSpec
+    partitioner_type = PartitionerType.DEFAULT_SINGLE_PARTITION
 
 
 @rule(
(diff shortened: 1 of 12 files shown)


cognifloyd:
Aargh. Something isn't working in my update. Tests pass, but actually running the linters gets caught in some weird loop.

pants-plugins/api_spec/rules.py
```diff
@@ -63,79 +50,39 @@ class APISpecFieldSet(FieldSet):
...
     desc="Update openapi.yaml with st2-generate-api-spec",
     level=LogLevel.DEBUG,
 )
 async def generate_api_spec_via_fmt(
-    request: GenerateAPISpecViaFmtTargetsRequest,
+    request: GenerateAPISpecViaFmtTargetsRequest.Batch,
+    subsystem: GenerateApiSpec,
 ) -> FmtResult:
     # There will only be one target+field_set, but we iterate
     # to satisfy how fmt expects that there could be more than one.
```
amanda11, line 72: Does this comment make sense now we have removed this chunk of code?

pants-plugins/api_spec/rules.py
```diff
@@ -63,79 +50,39 @@ class APISpecFieldSet(FieldSet):
...
-    )
-
-    source_files_get = Get(
-        SourceFiles,
-        SourceFilesRequest(field_set.source for field_set in request.field_sets),
-    )
+    config_files_get = Get(ConfigFiles, ConfigFilesRequest, subsystem.config_request())
 
     # actually generate it with an external script.
     # Generation cannot be inlined here because it needs to import the st2 code.
```
amanda11, line 78: Does this comment make sense now we have removed this chunk of code, and so we no-longer directly call GENERATE_CMD python script here?

pants-plugins/api_spec/rules.py
```diff
@@ -144,87 +91,59 @@ async def generate_api_spec_via_fmt(
...
     desc="Validate openapi.yaml with st2-validate-api-spec",
     level=LogLevel.DEBUG,
 )
 async def validate_api_spec(
-    request: ValidateAPISpecRequest,
-) -> LintResults:
+    request: ValidateAPISpecRequest.Batch,
+    subsystem: ValidateApiSpec,
+) -> LintResult:
     # There will only be one target+field_set, but we iterate
```
amanda11, line 125: Does this comment make sense now we have removed this chunk of code?

pants-plugins/api_spec/rules.py
```diff
@@ -144,87 +91,59 @@ async def generate_api_spec_via_fmt(
...
-
     source_files_get = Get(
         SourceFiles,
-        SourceFilesRequest(field_set.source for field_set in request.field_sets),
+        SourceFilesRequest(field_set.source for field_set in request.elements),
     )
 
+    config_files_get = Get(ConfigFiles, ConfigFilesRequest, subsystem.config_request())
+
     # actually validate it with an external script.
```
amanda11, line 136: Does this comment make sense now we have removed this chunk of code, as we nolonger have code here to call VALIDATE_CMD python module

pants-plugins/sample_conf/rules.py
```diff
@@ -51,37 +46,25 @@ class GenerateSampleConfFieldSet(FieldSet):
...
-    request: GenerateSampleConfViaFmtTargetsRequest,
+    request: GenerateSampleConfViaFmtTargetsRequest.Batch,
+    subsystem: ConfigGen,
 ) -> FmtResult:
     # There will only be one target+field_set, but we iterate
     # to satisfy how fmt expects that there could be more than one.
     # If there is more than one, they will all get the same contents.
 
     # actually generate it with an external script.
     # Generation cannot be inlined here because it needs to import the st2 code.
```
amanda11, line 66: Does this comment make sense now we have removed this chunk of code?

pants-plugins/schemas/rules.py
```diff
@@ -49,37 +42,24 @@ class GenerateSchemasFieldSet(FieldSet):
...
     desc="Update contrib/schemas/*.json with st2-generate-schemas",
     level=LogLevel.DEBUG,
 )
 async def generate_schemas_via_fmt(
-    request: GenerateSchemasViaFmtTargetsRequest,
+    request: GenerateSchemasViaFmtTargetsRequest.Batch,
+    subsystem: GenerateSchemas,
 ) -> FmtResult:
     # We use a pex to actually generate the schemas with an external script.
     # Generation cannot be inlined here because it needs to import the st2 code.
```
amanda11, line 58: Does this comment make sense now we have removed this chunk of code, as the external script isnt' directly referenced here anymore?

pants-plugins/api_spec/rules.py
cognifloyd: Hmm. Yup. This is out-of-date. I'll drop the comment.

pants-plugins/api_spec/rules.py
cognifloyd: Hmm. Yup. This is out-of-date. I'll drop the comment.

pants-plugins/sample_conf/rules.py
cognifloyd: The "only one target..." bit is out of date. I dropped it.

This code block was moved, not removed. The PexFromTargetsRequest (including all of the random constants used to build it) are now defined on the subsystem. I added a comment to help clarify that.

pants-plugins/schemas/rules.py
cognifloyd: This code block was moved, not removed. The PexFromTargetsRequest (including all of the random constants used to build it) are now defined on the subsystem. I added a comment to help clarify that.

pants-plugins/api_spec/rules.py
cognifloyd: This code block was moved, not removed. The PexFromTargetsRequest (including all of the random constants used to build it) are now defined on the subsystem. I added a comment to help clarify that.

pants-plugins/api_spec/rules.py
cognifloyd: This code block was moved, not removed. The PexFromTargetsRequest (including all of the random constants used to build it) are now defined on the subsystem. I added a comment to help clarify that.

amanda11 (approved):
Thanks for updating the comments

PR #5901, opened 2023-02-14
pants: Add `st2_*_python_distribution()` macros to `pants-plugins/macros.py`

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

This PR adds pants macros that reduce boilerplate around defining `python_distribution()`:
- `st2_license()`
- `st2_component_python_distribution()`
- `st2_runner_python_distribution()`

The next PR will actually add those macros to the relevant BUILD files.

See #5890 for a description of pants macros and what is involved in using them in pants.

### `st2_license()` macro

We need to add a LICENSE file to all of our wheels. When building wheels, we don't have to add any special metadata to the setup.py `setup()` call (or `pyproject.toml` or similar); we only need to put the LICENSE file in the directory that will be the wheel.

https://github.com/StackStorm/st2/blob/7de8f3c3c47392a8841653c590bd209db6546a23/pants-plugins/macros.py#L16-L21

### `st2_*_python_distribution()` macros

These macros create:
- a copy of the LICENSE file using the `st2_license` macro.
- a `python_artifact` object with the details needed to generate the setup.py kwargs. For example, it includes the `version_file` arg which is unique to `pants-plugins/release` which uses this to inject the version based on that file.
- a `python_distribution` target with the metadata for creating the wheel.

Using this macro is like creating a `python_distribution` like this in each of the BUILD files:

```python
python_distribution(
    provides=python_artifact(
        name="...",  # the wheel name
        description="...",
        version="3.9dev",
        zip_safe=...,
        scripts=[...],
    ),
    dependencies=[
        "./....",  # the primary python module
        ":license",
    ],
    repositories=[
        "@pypi",  # we plan to publish this on pypi
    ],
)
```

#### `st2_component_python_distribution()` macro

This is meant for for our primary components: `st2common`, `st2auth`, `st2api`, `st2client`, etc.

For example, in the next PR this is how we'll define the `st2api` distribution:
https://github.com/StackStorm/st2/blob/7032faf17e6a6b1b8bf229e23b7724536e67bc55/st2api/BUILD#L1-L4

#### `st2_runner_python_distribution()` macro

This is meant for our runners: `contrib/runners/*_runner`.

For example, in the next PR this is how we'll define the `noop` runner's distribution:
https://github.com/StackStorm/st2/blob/7032faf17e6a6b1b8bf229e23b7724536e67bc55/contrib/runners/noop_runner/BUILD#L1-L7

### DIFF
--- a/pants-plugins/macros.py
+++ b/pants-plugins/macros.py
@@ -13,6 +13,79 @@
 # limitations under the License.
 
 
+def st2_runner_python_distribution(**kwargs):
+    runner_name = kwargs.pop("runner_name")
+    description = kwargs.pop("description")
+
+    kwargs["provides"] = python_artifact(  # noqa: F821
+        name=f"stackstorm-runner-{runner_name.replace('_', '-')}",
+        description=description,
+        version_file=f"{runner_name}_runner/__init__.py",  # custom for our release plugin
+        # test_suite="tests",
+        zip_safe=kwargs.pop(
+            "zip_safe", True
+        ),  # most runners are safe to run from a zipapp
+    )
+
+    dependencies = kwargs.pop("dependencies", [])
+    for dep in [f"./{runner_name}_runner"]:
+        if dep not in dependencies:
+            dependencies.append(dep)
+    kwargs["dependencies"] = dependencies
+
+    repositories = kwargs.pop("repositories", [])
+    for repo in ["@pypi"]:
+        if repo not in repositories:
+            repositories.append(repo)
+    kwargs["repositories"] = repositories
+
+    python_distribution(**kwargs)  # noqa: F821
+
+
+def st2_component_python_distribution(**kwargs):
+    st2_component = kwargs.pop("component_name")
+
+    description = (
+        f"{st2_component} StackStorm event-driven automation platform component"
+    )
+
+    scripts = kwargs.pop("scripts", [])
+
+    kwargs["provides"] = python_artifact(  # noqa: F821
+        name=st2_component,
+        description=description,
+        scripts=[
+            script[:-6] if script.endswith(":shell") else script for script in scripts
+        ],
+        version_file=f"{st2_component}/__init__.py",  # custom for our release plugin
+        # test_suite=st2_component,
+        zip_safe=False,  # We rely on __file__ to load many things, so st2 should not run from a zipapp
+    )
+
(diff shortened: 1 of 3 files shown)


pants-plugins/macros.py
```diff
@@ -13,6 +13,99 @@
...
+    )
+
+    dependencies = kwargs.pop("dependencies", [])
+    for dep in [f"./{runner_name}_runner", ":license"]:
+        if dep not in dependencies:
+            dependencies.append(dep)
+    kwargs["dependencies"] = dependencies
+
+    repositories = kwargs.pop("repositories", [])
+    for repo in ["@pypi"]:
```
rush-skills, line 55: Why the hardcoded "@pypi" and the loop?

rush-skills (approved):

pants-plugins/macros.py
cognifloyd: Oh. Yes. I never revised that.

The idea here is that we might want to publish wheels to `@testpypi` in addition to `@pypi`, but I never actually included that.

The [`repositories`](https://www.pantsbuild.org/docs/reference-python_distribution#coderepositoriescode) kwarg is:

> List of URL addresses or Twine repository aliases where to publish the Python package.
>
> Twine is used for publishing Python packages, so the address to any kind of repository that Twine supports may be used here.
>
> Aliases are prefixed with `@` to refer to a config section in your Twine configuration, such as a `.pypirc` file. Use `@pypi` to upload to the public PyPi repository, which is the default when using Twine directly.

So, aliases come from [`.pypirc`](https://packaging.python.org/en/latest/specifications/pypirc/#common-configurations) or from the `TWINE_` environment vars. By default, twine defines `@pypi` and `@testpypi`.

Should we simplify this to just check for `@pypi`? Or add `@testpypi` like I originally planned but forgot about?

cognifloyd: OK. I did some more research. I'll push a commit cleaning this up shortly including a TODO for when we switch to pants 2.16. There's a new `env()` object that will allow switch the repo we publish to using an env var.

Thanks for asking so that I remembered to look into this more.

cognifloyd: Cleaned up.

amanda11 (approved):

Issue #4928, opened 2020-04-29
Sensor token has expired.


guzzijones:
I created a sensor that uses a datastore key as a queue.
I have a sensor that runs every 10 seconds and makes a call to the datastore service to check a key for values.
As new values are added this sensor will remove them as they age.

I setup an alert to notify myself when the sensor fails and I received this today:
```
2020-04-29 15:42:39,191 139660794638416 AUDIT access [-] Access granted to "sensor_service" with the token set to expire at "2020-04-30T15:42:39.189539Z". (username='sensor_service',token_expiration='202
2020-04-29 15:42:39,224 139660794638416 INFO intel_alien_sensor [-] last pull time: 2020-04-29T03:42:38.460618+00:00
2020-04-29 15:42:39,294 139660794638416 INFO intel_alien_sensor [-] count:3347
2020-04-29 15:42:40,094 139660794638416 INFO intel_alien_sensor [-] lmintel.intelAlienSensor no events found
2020-04-29 15:42:43,081 140467950239824 INFO rate_limit_queue [-] checking queues
2020-04-29 15:42:43,081 140467950239824 INFO keyvalue [-] checking rate limit queues
2020-04-29 15:42:43,090 140467950239824 WARNING sensor_wrapper [-] Sensor "RateLimitQueue" run method raised an exception: 401 Client Error: Unauthorized
MESSAGE: Unauthorized - Token has expired. for url: http://127.0.0.1:9101/v1/keys/?prefix=ratelimit2731795c-e948-11e9-81b4-2a2ae2dbcce4%3Atodo.
Traceback (most recent call last):
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2reactor/container/sensor_wrapper.py", line 229, in run
    self._sensor_instance.run()
  File "/opt/stackstorm/st2/lib/python2.7/site-packages/st2reactor/sensor/base.py", line 119, in run
    self.poll()
  File "/opt/stackstorm/packs/lmutil27/sensors/rate_limit_queue.py", line 41, in poll
    self._client.check_todo()
  File "/opt/stackstorm/virtualenvs/lmutil27/lib/python2.7/site-packages/stackrlclient/keyvalue.py", line 314, in check_todo
    for i in self.all_todo():
  File "/opt/stackstorm/virtualenvs/lmutil27/lib/python2.7/site-packages/stackrlclient/keyvalue.py", line 271, in all_todo
    return  self._all_keyvals(RLClient.TODO)
[...]

PR #5902, opened 2023-02-14
WIP: Enhancement to process_container.py

luckenbach:
# Why
#4928
# How
Per @guzzijones's recommendation logic needed to be added to cycle the sensor if the token is expired based of the TTL of the token

### DIFF
--- a/st2reactor/st2reactor/container/process_container.py
+++ b/st2reactor/st2reactor/container/process_container.py
@@ -214,6 +214,18 @@ class ProcessSensorContainer(object):
                     # Sensor has been successfully running more than threshold seconds, clear the
                     # respawn counter so we can try to restart the sensor if it dies later on
                     self._sensor_respawn_counts[sensor_id] = 0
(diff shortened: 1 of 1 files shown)


amanda11:
As per comments in #4928 I'd like to see why the existing code to refresh the token isn't working. 
There is logic to refresh the token as long as we access via the sensor_service, which in turn calls the get_api_key in the datastore.py  (which is where the refresh logic is).
In the example of where this fails, are we trying to access the api client without going through the sensor_service methods? If so, why? Do we need to add more methods to sensor_service instead, so that you can benefit from the refresh key code in there.
OR have we got a bug in that refresh code.

guzzijones:
The fact that we restart the service to create a new token is not ideal.  yes we 'refresh' the token, but the way we do it isn't the cleanest.

guzzijones (approved):

guzzijones:
closing as [user error per this comment](https://github.com/StackStorm/st2/issues/4928#issuecomment-1502048720)

PR #5913, opened 2023-02-24
Update sensor wrapper to make `list_values()` method congruent with regular datastore method.

TimothyDJones:
Specifically, support use of `limit` and `offset` parameters.


fdrab:
this was implemented in https://github.com/StackStorm/st2/pull/6218

PR #5928, opened 2023-03-09
Pants: add dependencies to the st2* `python_distributions` (except st2common/st2tests)

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

Like #5925, which adds dependencies for the `python_distribution` for `st2common`, this PR adds the remaining dependencies for the other `st2*` component `python_distribution` targets. It also skips `st2tests` which requires quite a few more changes in #5929.

It also adds a note to `pants.toml` to make it clear that pants will generate a `setup.py` file for us.

### Pants documentation

- [Building distributions](https://www.pantsbuild.org/docs/python-distributions)
    - see also: [`python_distribution` target](https://www.pantsbuild.org/docs/reference-python_distribution)

### DIFF
--- a/st2api/st2api/BUILD
+++ b/st2api/st2api/BUILD
@@ -1,1 +1,12 @@
-python_sources()
+python_sources(
+    overrides={
+        "app.py": dict(
+            dependencies=[
+                # controller routes are wired up via st2common/st2common/openapi.yaml
+                "./controllers",
+                "./controllers/v1",
+                # "./controllers/exp",
+            ],
+        ),
+    },
+)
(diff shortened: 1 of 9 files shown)


st2api/st2api/BUILD
```diff
@@ -1 +1,12 @@
-python_sources()
+python_sources(
+    overrides={
+        "app.py": dict(
+            dependencies=[
+                # controller routes are wired up via st2common/st2common/openapi.yaml
+                "./controllers",
+                "./controllers/v1",
+                # "./controllers/exp",
```
cognifloyd, line 8: No APIs are in the experimental folder right now. So, I left this commented out.

nzlosh (approved):

rush-skills (approved):

PR #5931, opened 2023-03-10
Pants: reclassify test utils in BUILD metadata (part 2)

cognifloyd:
### Background

This is another part of introducing [pants](https://www.pantsbuild.org/docs), as discussed in various TSC meetings.

Related PRs can be found in:
- [pantsbuild](https://github.com/StackStorm/st2/milestone/47) milestone (which includes PRs since #5713)
- https://github.com/StackStorm/st2/labels/pantsbuild label (which also includes preparatory PRs that came before adding pantsbuild)

### Overview of this PR

I missed 3 BUILD files in #5926, so this reclassifies those sources as well.

> This replaces a [3] `python_sources` targets with `python_test_utils`. I excluded the `test_*.py` files in `tests/unit` directories since those are already covered by `python_tests` targets.
>
> This PR does not make any functional changes - to Pants, `python_sources` and `python_test_utils` are functionally equivalent. So, this is a very minor/small PR.

I also noticed a typo (paths weren't relative but should have been) in `st2actions/BUILD`, so I included that as well.

### Pants documentation

- [`python_test_utils` target](https://www.pantsbuild.org/docs/reference-python_test_utils)
- [`python_sources` target](https://www.pantsbuild.org/docs/reference-python_sources)

PR #5932, opened 2023-03-11
Upgrade pants from 2.15 to 2.16.0a0

cognifloyd:
### Background

Related PRs can be found in:

### Overview of this PR

_I recommend stepping through the commits to review this PR. Ignore the "[fix typo in st2actions/BUILD](https://github.com/StackStorm/st2/pull/5932/commits/8ba8b3a4d6a5cf005bb211fd5c1a8bd2597f4147)" commit, as it is part of #5931, which should be merged first._

This updates to pants to `2.16.0a0` which only requires:
- regenerate the pants-plugins lockfile
- regenerate the st2 lockfile

Unlike the upgrade to 2.15 (#5899), we do not need to update our pants-plugins because they don't use anything that changed (see the [Plugin upgrade guide](https://www.pantsbuild.org/v2.16/docs/plugin-upgrade-guide)).

Upgrading to 2.16 gives us access to some new features and allows us to delete some of our BUILD metadata:

- Several `module_mapping` definitions were upstreamed, so we no longer need to define them locally.
- `pants.backend.experimental.python.framework.stevedore`: The plugin I wrote to tell pants how to manage including runners and other stevedore plugins when the tests need them to be available.
- Uncomment `env("ST2_PUBLISH_REPO")` in `pants-plugins/macros.py` (added in #5901) so that we can use the `ST2_PUBLISH_REPO` env var to determine where we want to publish the wheels (eg to use `@testpypi` instead of `@pypi`).
- The `python-dump-source-analysis` goal makes debugging dependency inference much easier. And, it was the final piece we needed to be able to use `[python-infer].string_imports` so that pants can infer python dependencies from strings, not just the actual imports. That made it possible to delete some of our BUILD metadata.
- The `[python-infer].ambiguity_resolution = "by_source_root"` option made dependency inference a little smarter. This way, when something imports `tests.base`, it will use the closest `tests.base` instead of complaining about all of the different directories that define that. That made it possible to delete even more of the BUILD metadata, as those deps are no longer ambiguous.
- Turn on `diff` output when using `generate-lockfiles` (does not display when our git deps change, but it helps for everything else). This is what it looked like when I regenerated the `st2` lockfile (see the screenshots in https://github.com/pantsbuild/pants/pull/17347 as well):
  image

In a follow-up I will add support for:
- `pants.backend.experimental.visibility` which gives us access to `__dependency_rules__` and `__dependents_rules__`. This is a much more intelligent way to avoid circular imports (eg st2common does not import st2api) than the regex-lint based approach. _This needs a few more bug fixes before we can use it._
    - [Validating dependencies](https://www.pantsbuild.org/v2.16/docs/validating-dependencies) covers `__dependencies_rules__` and `__dependents_rules__`

### Pants documentation

- [Plugin upgrade guide](https://www.pantsbuild.org/v2.16/docs/plugin-upgrade-guide)
- [2.16 Changelog](https://github.com/pantsbuild/pants/blob/main/src/python/pants/notes/2.16.x.md)
- [`python-dump-source-analysis` goal](https://www.pantsbuild.org/v2.16/docs/reference-python-dump-source-analysis)
- [`python-infer` subsystem](https://www.pantsbuild.org/v2.16/docs/reference-python-infer)
    - [`string_imports` option](https://www.pantsbuild.org/v2.16/docs/reference-python-infer#string_imports)
    - [`ambiguity_resolution` option](https://www.pantsbuild.org/v2.16/docs/reference-python-infer#ambiguity_resolution)


cognifloyd:
Marking as draft for now because regenerating the lockfile with 2.16.0a0 seems to be broken.

cognifloyd:
blocked by https://github.com/StackStorm/st2-rbac-backend/pull/71 without which we can't regenerate the lockfile

rush-skills (approved):

amanda11 (approved):

Issue #5934, opened 2023-03-13
Unable to render KV store params in actions when RBAC is enabled


jschoewe:
## SUMMARY

After enabling RBAC in version 3.8 we are not able to run actions that look up parameters from the KV store. The issue appears to occur when there's a period in the key name such as `test.key1` and using keys with underscores instead (like `test_key1`) still works. This also appears to be an issue when loading values from pack configs.

```
ERROR: 400 Client Error: Bad Request
MESSAGE: Failed to render parameter "message": 'NoneType' object has no attribute 'scope' for url: http://127.0.0.1:9101/v1/executions
```

### STACKSTORM VERSION

st2 3.8.0, on Python 3.8.13

##### OS, environment, install method

RHEL8, one-line install

## Steps to reproduce the problem

Install stackstorm, enable RBAC, create action that looks up parameter from KV store

Test Action:
`/opt/stackstorm/packs/core/actions/echo_jinja.yaml`
```
---
description: Action that executes the Linux echo command on the localhost.
runner_type: "local-shell-cmd"
enabled: true
entry_point: ''
name: echo_jinja
parameters:
  message:
    description: The message that the command will echo.
    type: string
    required: true
    default: "{{ st2kv.system.test.key1 }}"
  cmd:
    description: Arbitrary Linux command to be executed on the local host.
    required: true
    type: string
    default: 'echo "{{ message }}"'
    immutable: true
```
`st2 key set test.key1 'HELLO'`
`st2 pack register core`
`st2 run core.echo_jinja`

## Actual Results

After running the test action above I get the following error:
```
ERROR: 400 Client Error: Bad Request
MESSAGE: Failed to render parameter "message": 'NoneType' object has no attribute 'scope' for url: http://127.0.0.1:9101/v1/executions
```

According to the `st2api.log` it doesn't seem to be looking for the whole key:
```
2023-03-13 11:27:27,121 140238292154672 DEBUG keyvalues [-] Lookup system kv: scope: st2kv.system and key: test
```

Also, when I add another key with `st2 key add test '0'`, the action does run successsfully after looking up both the `test` and `test.key1` keys:
```
2023-03-13 09:55:50,616 140707953123744 DEBUG keyvalues [-] Lookup system kv: scope: st2kv.system and key: test
2023-03-13 09:55:50,621 140707953123744 DEBUG keyvalues [-] Got value 0 from datastore.
[...]

PR #5936, opened 2023-03-14
bugfix/ISSUE-5934-fix-rbac-kv-value-render

jschoewe:
Moved permission checks into if statement because we don't need to run checks if no key was found.

This is a fix for the following issue: 
Closes https://github.com/StackStorm/st2/issues/5934


amanda11:
@jschoewe Please could you sign the CLA, see message above.

guzzijones:
I also just hit this bug.  Knowing that I need to add the key store value helped me resolve it. Thanks.

guzzijones (approved):
approved. we should not check rbac if no key value is found.

nzlosh (approved):

nzlosh:
@jschoewe Would you be able to rebase your branch against master so this PR can be merged please?

armab (approved):

armab:
Tagging this for the upcoming `v3.8.1` patch release.

@jschoewe Could you please update the branch with the latest st2 upstream so we can merge the PR?
I couldn't push to your fork.

Issue #5814, opened 2022-11-23
Have st2 pack install checkout git submodules when cloning a pack repo


jk464:
When a Pack is installed on Stackstorm using `pack install` - stackstorm performs a clone on the Repo ([Code](https://github.com/StackStorm/st2/blob/dfab25aab0468f330a635db55f01286523217135/st2common/st2common/util/pack_management.py#L205)) - however this results in any submodules present in the repo not being checked out and only being an empty directory. 

It would be useful if the `clone_repo` function also checked out any submodules that are present in the repo, as it would be helpful to add dependencies maintained in seperate repos to the pack as a submodule, while still being maintained externally.

In my use case a series of ansible playbooks are maintained in a seperate repo and I would like to also make them available as a pack to st2, and as I see it the best way to do that is via git submodules.

armab:
Thanks for the feature request. It would be a helpful addition.

This sounds like a `good first issue` if someone is willing to contribute the enhancement.

jamesdreid:
I know it is does not address the feature enhancement directly, but this function can be addressed with a workflow and rule that is triggered whenever the pack is loaded.  The workflow task would look something like:
![image](https://user-images.githubusercontent.com/6311944/204353457-a60c7c71-f258-4144-93a4-c98035e2dc61.png)

Then add a rule to the pack that will run this workflow any time the pack is installed and/or expose the workflow as an alias and allow it to be run directly via Chat if the sub-module needs updating outside of the pack install.

jk464:
Hi I've open a PR that implements this - could someone review this PR? Thanks :)

PR #5963, opened 2023-04-21
StackStorm Packs - Support for Git Submodules

jk464:
StackStorm Packs are implemented as Git Repos. Currently when `pack install` is run, submodules contained within packs are not handled.

This PR implements the `checkout_submodules` option on `pack install` (and `pack download` etc) so that when `pack install` is ran a user can optionally set this option to `True` to checkout submodules as part of the install process.

The option defaults to `False` to maintain current behaviour of the `pack install` process and thus backwards complatibity with pre-exisiting processes.

Implements #5814

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -28,6 +28,8 @@ Added
 * Expose environment variable ST2_ACTION_DEBUG to all StackStorm actions.
   Contributed by @maxfactor1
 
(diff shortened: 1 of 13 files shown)


guzzijones:
please sign the cla

guzzijones:
oh looks like this just needs a review and a milestone

PR #5992, opened 2023-06-20
Futureproofing: Removing `distutils` dependencies

AndroxxTraxxon:
Python has been warning that distutils will be removed for a while now, and the retirement is official in Python3.12, so this is my attempt at removing/replacing distutils functionality.

I opted to not replace the pip version check in the `st2client` package because Pip 6.1.0 (April 2015) was released before Python 3.6 (Dec. 2016, already EOL) existed, and we specify the pip version in the Makefile. I'm sure we can find a way to get `packaging` installed for the version check, or to replace it with a custom-rolled replacement comparison if the community deems it necessary.

I could also see the argument for removing that check from the `fixate-requirements.py` script as well, but I wanted to show the most commonly recommended replacement for distutils.version.

This change is similar to the one made for the Orquesta package here, basically just removing functionality that's been deprecated for a while and is officially removed in an upcoming version of Python: https://github.com/StackStorm/orquesta/pull/253

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -14,7 +14,7 @@ Added
 ~~~~~
 * Remove `distutils` dependencies across the project. #5992
   Contributed by @AndroxxTraxxon
-  
+
 * Move `git clone` to `user_home/.st2packs` #5845
 
 * Error on `st2ctl status` when running in Kubernetes. #5851
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,8 +15,6 @@ Fixed
 
 Added
 ~~~~~
-* Remove `distutils` dependencies across the project. #5992
-  Contributed by @AndroxxTraxxon
 
 * Move `git clone` to `user_home/.st2packs` #5845
 
@@ -39,6 +37,11 @@ Added
 
 * Expose environment variable ST2_ACTION_DEBUG to all StackStorm actions.
   Contributed by @maxfactor1
+
+Changed
+~~~~~~~
+* Remove `distutils` dependencies across the project. #5992
+  Contributed by @AndroxxTraxxon
 
 3.8.0 - November 18, 2022
 -------------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,9 @@ in development
 
 Added
 ~~~~~
+* Remove `distutils` dependencies across the project. #5992
+  Contributed by @AndroxxTraxxon
+  
 * Move `git clone` to `user_home/.st2packs` #5845
 
 * Error on `st2ctl status` when running in Kubernetes. #5851
--- a/Makefile
+++ b/Makefile
@@ -661,7 +661,7 @@ distclean: clean
 
 .PHONY: .requirements
 .requirements: virtualenv
-	$(VIRTUALENV_DIR)/bin/pip install --upgrade "pip==$(PIP_VERSION)" 
+	$(VIRTUALENV_DIR)/bin/pip install --upgrade "pip==$(PIP_VERSION)"
 	# Print out pip version
 	$(VIRTUALENV_DIR)/bin/pip --version
 	# Generate all requirements to support current CI pipeline.
--- a/Makefile
+++ b/Makefile
@@ -661,7 +661,7 @@ distclean: clean
 
 .PHONY: .requirements
 .requirements: virtualenv
-	$(VIRTUALENV_DIR)/bin/pip install --upgrade "pip==$(PIP_VERSION)"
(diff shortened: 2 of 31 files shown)


AndroxxTraxxon:
I had to remove the Pip version check in its entirety in order to remain compatible to Python 3.6. It does make me wonder if it should still be supported, since it's already past EOL.

Makefile
```diff
@@ -661,7 +661,7 @@ distclean: clean
 
 .PHONY: .requirements
 .requirements: virtualenv
-	$(VIRTUALENV_DIR)/bin/pip install --upgrade "pip==$(PIP_VERSION)"
+	$(VIRTUALENV_DIR)/bin/pip install --upgrade "pip==$(PIP_VERSION)"
```
nzlosh, line 664: Is the trailing slash intentional here?

guzzijones:
Give this error a look please: [unit test error](https://github.com/StackStorm/st2/actions/runs/5357974963/jobs/9719615644#step:15:6209)

AndroxxTraxxon:
@guzzijones 
Those tests failed with a warning about some race conditions being possible:
```
st2.st2common.services.coordination: WARNING: Coordination backend is not configured. Code paths which use coordination service will use best effort approach and race conditions are possible.
 ```
 I'm not entirely sure what causes this, but I don't think this change caused those failures. You can see that these tests are the same tests that were cancelled in the [last merge to master](https://github.com/StackStorm/st2/actions/runs/5903104626/job/16012347214)

**Comment by AndroxxTraxxon** [comment]
> Overall the change looks good to me. I do wonder if the pip checks should disabled rather than completely removed as they may still be useful for other versions of python. These checks will probably end up being performed by pants, so maybe not such a big deal to remove them.

The comparison was being done to ensure an appropriate version of pip was being used in the installation, but in the [Makefile](https://github.com/AndroxxTraxxon/st2/blob/distutils-removal/Makefile#L59), we pin that pip version to `20.3.3` during the installation, which is well past the `6.0.0` minimum that was being enforced. If these kinds of checks are going to additionally be performed by the pants system, this is all the more reason to remove it now.

**Comment by AndroxxTraxxon** [comment]
The last[ test failure is another race condition](https://github.com/StackStorm/st2/actions/runs/6167697127/job/16739043666?pr=5992#step:15:1603) for a test that was not modified. You can see in previous commits that it passed just fine
```
st2.st2common.services.coordination: WARNING: Coordination backend is not configured. Code paths which use coordination service will use best effort approach and race conditions are possible.
```

File: st2common/st2common/util/sandboxing.py [hunk 2]
```diff
@@ -23,14 +23,20 @@
...
 from oslo_config import cfg
 
 from st2common.constants.action import LIBS_DIR as ACTION_LIBS_DIR
 from st2common.constants.pack import SYSTEM_PACK_NAMES
 from st2common.content.utils import get_pack_base_path
 
+
+def get_python_lib():
+    """Replacement for distutil.sysconfig.get_python_lib, returns a string with the python platform lib path (to site-packages)"""
+    return get_path("platlib")
```
armab, line 37: Just for reference: https://bugs.python.org/issue41282#msg393018 how it's a replacement for `distutil.sysconfig.get_python_lib`.

CHANGELOG.rst
```diff
@@ -15,6 +15,9 @@ Fixed
 
 Added
 ~~~~~
+* Remove `distutils` dependencies across the project. #5992
+  Contributed by @AndroxxTraxxon
```
armab, line 19: This better fits the `Changed` changelog subcategory, instead of `Added`.
You can create one, like here https://github.com/StackStorm/st2/pull/5992/files#diff-2c623f3c6a917be56c59d43279244996836262cb1e12d9d0786c9c49eef6b43cR111

armab (approved):
Thanks, very helpful in the context of difficulties with the StackStorm's Python versions support https://github.com/StackStorm/community/issues/103

I guess @cognifloyd 👀 you might be interested to take look too, as it's build/packaging related.

CHANGELOG.rst
AndroxxTraxxon: Fair point. I'll move it over.

guzzijones:
> The last[ test failure is another race condition](https://github.com/StackStorm/st2/actions/runs/6167697127/job/16739043666?pr=5992#step:15:1603) for a test that was not modified. You can see in previous commits that it passed just fine
> 
> ```
> st2.st2common.services.coordination: WARNING: Coordination backend is not configured. Code paths which use coordination service will use best effort approach and race conditions are possible.
> ```

Yes,  sadly the unit tests run without the coordinator turned on.  A redis container is linked to the unit tests, but it is not used. Also, it looks like some unit tests patch the NoCoordinator.  I noticed this on our own fork.

guzzijones (approved):
looks good to me.

PR #5994, opened 2023-06-21
Added get-dependencies option to install_pack.py

jk464:
Enhances `install_pack.py` so that you can pass an optional flag to it (`--get-dependencies`) that will for a given pack when it is installed will also install any other packs the parent pack depdends on it.

Useful when using https://github.com/StackStorm/st2packs-dockerfiles/ to build a `st2packs` image, the user only has to provide the list of packs they want, without having to determine if they have any further packs as depdencies.


guzzijones:
needs a changelog entry

jk464:
@guzzijones I've added a Changelog entry now

st2common/st2common/cmd/install_pack.py
```diff
@@ -53,28 +56,37 @@ def _register_cli_opts():
...
-    common_setup(
-        config=config,
-        setup_db=False,
-        register_mq_exchanges=False,
-        register_internal_trigger_types=False,
-    )
+    try:
+        pack_metadata = get_pack_metadata(pack_dir=pack_path)
+        result = pack_metadata.get("dependencies", None)
+    except Exception:
```
guzzijones, line 74: Please catch IOError instead of the generic exception.

st2common/st2common/cmd/install_pack.py
```diff
@@ -53,28 +56,37 @@ def _register_cli_opts():
...
+        LOG.error("Could not open pack.yaml at location %s" % pack_path)
+        result = None
+    finally:
+        if result:
```
guzzijones, line 78: move this under line 73 and remove the finally block altogether.

st2common/st2common/cmd/install_pack.py
```diff
@@ -53,28 +56,37 @@ def _register_cli_opts():
...
 
-    packs = cfg.CONF.pack
-    verify_ssl = cfg.CONF.verify_ssl
-    force = cfg.CONF.force
-
-    proxy_config = get_and_set_proxy_config()
 
+def download_packs(packs, verify_ssl, force, dependencies, proxy_config):
     for pack in packs:
+        if pack in listdir("/opt/stackstorm/packs"):
```
guzzijones, line 86: is this path available as a config parameter?  I would hope we wouldn't have to hard code this.

guzzijones (changes requested):
just a couple small adjustments.  Thanks for your contribution.

guzzijones (approved):
looks good to me now

PR #6017, opened 2023-08-23
Fix Cert-based SSH Auth Issues in Paramiko

jk464:
Paramiko 2.10.1 is effect by:

https://github.com/paramiko/paramiko/issues/2017

That break cert-based SSH auth when using Paramiko, both are fixed in 2.10.5

This PR bumps Paramiko to 2.10.5 to fix this issue

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,9 @@ Fixed
 * Fix CI usses #6015
   Contributed by Amanda McGuinness (@amanda11 intive)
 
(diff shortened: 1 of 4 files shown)


nzlosh:
Thanks for creating this PR.  Perhaps remove https://github.com/paramiko/paramiko/issues/1963 from the documentation as it only concerns paramiko 2.9 branch where st2 is already pinned the 2.10 branch.

nzlosh (approved):

guzzijones (approved):

PR #6030, opened 2023-10-08
Fix #4676

### LINKED ISSUE #4676: st2 run --inherit-env does not work for actions run with python-script runner
##### SUMMARY

The `--inherit-env` option supposedly supports the `python-script` runner:

```
  -e, --inherit-env     Pass all the environment variables which are
                        accessible to the CLI as "env" parameter to the
                        action. Note: Only works with python, local and remote
                        runners.
```

But it doesn't actually pass the environment variables to the Python script. I suspect that this has been broken for some time.

The `core.local` and `core.remote` actions do pass their environment variables correctly.

##### ISSUE TYPE
Pick one below and delete the rest:
 - Bug Report

##### STACKSTORM VERSION

`st2 3.1dev (95d7cc3), on Python 2.7.12`

Ubuntu 16.04, installed as part of `setup_e2e_tests` in st2ci pack.

##### STEPS TO REPRODUCE

`actions/test_inherit_env_python_script_runner.yaml`:
```yaml
---
name: test_inherit_env_python_script_runner
description: Test the --inherit-env option to the st2 command for an action run with the python-script runner
enabled: true
runner_type: python-script
entry_point: test_inherit_env.py
parameters: {}
```

`actions/test_inherit_env.py`:
```python
import os

class InheritEnvTestAction(Action):
    def run(self, *args, **kwargs):
        return (True, os.environ)
```

##### EXPECTED RESULTS

```bash
env | grep SLACK_; st2 run --inherit-env tests.test_inherit_env_python_script_runner
```

```
SLACK_USER_API_TOKEN=xoxb-...
SLACK_CHANNEL=chatops_e2e_tests_u16
SLACK_BOT_USERNAME=chatops_test_bot_u16
SLACK_USER_USERNAME=chatops_test_user_u16
.
id: ...
status: succeeded
parameters: None
result:
  exit_code: 0
  result: '{
    ''SLACK_USER_API_TOKEN'': ''xoxb-...'',
    ''SLACK_CHANNEL'': ''chatops_e2e_tests_u16'',
    ''SLACK_BOT_USERNAME'': ''chatops_test_bot_u16'',
    ''SLACK_USER_USERNAME'': ''chatops_test_user_u16'',
    ''ST2_ACTION_EXECUTION_ID'': ''...',
    ''DAEMON_ARGS'': ''--config-file /etc/st2/st2.conf'',
    ''ST2_ACTION_API_URL'': ''http://127.0.0.1:9101/v1'',
    ''ST2_ACTION_AUTH_TOKEN'': ''...'',
    ''PYTHONPATH'': ''/opt/stackstorm/packs/tests/lib:/opt/stackstorm/st2/lib/python2.7/site-packages'',
    ''ST2_AUTH_TOKEN'': ''...'',
    ''SHELL'': ''/bin/sh'',
    ''LOGNAME'': ''root'',
    ''ST2_API_URL'': ''http://127.0.0.1:9101/v1'',
    ''WORKERID'': ''10'',
    ''PATH'': ''/opt/stackstorm/virtualenvs/tests/bin/:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin'',
    ''HOME'': ''/root'',
    ''USER'': ''root''
  }'
  stderr: ''
  stdout: ''
```

Note how the `SLACK_` environment variables show up in the output from st2.

##### ACTUAL RESULTS

```bash
env | grep SLACK_; st2 run --inherit-env tests.test_inherit_env_python_script_runner
```

```
SLACK_USER_API_TOKEN=xoxb-...
SLACK_CHANNEL=chatops_e2e_tests_u16
SLACK_BOT_USERNAME=chatops_test_bot_u16
SLACK_USER_USERNAME=chatops_test_user_u16
.
id: ...
status: succeeded
parameters: None
result:
  exit_code: 0
  result: '{
    ''ST2_ACTION_PACK_NAME'': ''tests'',
    ''ST2_ACTION_EXECUTION_ID'': ''...',
    ''DAEMON_ARGS'': ''--config-file /etc/st2/st2.conf'',
    ''ST2_ACTION_API_URL'': ''http://127.0.0.1:9101/v1'',
    ''ST2_ACTION_AUTH_TOKEN'': ''...'',
    ''PYTHONPATH'': ''/opt/stackstorm/packs/tests/lib:/opt/stackstorm/st2/lib/python2.7/site-packages'',
    ''ST2_AUTH_TOKEN'': ''...'',
    ''SHELL'': ''/bin/sh'',
    ''LOGNAME'': ''root'',
    ''ST2_API_URL'': ''http://127.0.0.1:9101/v1'',
    ''WORKERID'': ''10'',
    ''PATH'': ''/opt/stackstorm/virtualenvs/tests/bin/:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin'',
    ''HOME'': ''/root'',
    ''USER'': ''root''
  }'
  stderr: ''
  stdout: ''
```
Note how the `SLACK_` environment variables do not show up in the output from st2.

rebrowning:
execute populating the inherited environment variables whenever the parameter is set. Currently the doesn't work for shell commands either, it relies on those commands having default parameters that would then populate the 'parameters' variable to pass that check

Looking at fixing: https://github.com/StackStorm/st2/issues/4676

Replicating with the steps in the issue above. It looks like if `args.parameters` is not set (an empty list as well) then the empty result is returned, however this happens even if `args.inherit_env` is set. Moved `args.inherit_env` before the check so that even when we're running a job with no parameters, we'll still populate the environment variables.

Let me know if this makes sense, or any edge cases we'd want to test here. This was working for the shell runners because they all have a "cmd" parameter, which was then passing the `args.parameters` check.

This seems straight forward but I haven't been in the code base long enough to see if the original implementation is the expected approach.


guzzijones:
add to change log please.  Also can you add a unit test to cover this edge case?

m4dcoder (commented):
@rebrowning Can you add new or point to an existing unit test where the action parameters are set and inherit environment is true?

m4dcoder (approved):

rebrowning:
@guzzijones Any additional requests for this PR?

armab (approved):

st2client/tests/unit/test_command_actionrun.py
```diff
@@ -261,3 +261,135 @@ def test_get_params_from_args_with_multiple_declarations(self):
 
         # set auto_dict back to default
         mockarg.auto_dict = False
+
+    def test_correctly_process_inherit_env_when_no_parameters_set(self):
+        """test_correctly_process_inherit_env_when_no_parameters_set
+
+        This tests that we still correctly pass through the environment variables
+        when --inherit-env is set and we run a job that does not have parameters
```
cognifloyd, line 269:
```suggestion
        when --inherit-env is set and we run an action that does not have parameters
```

st2client/tests/unit/test_command_actionrun.py
```diff
@@ -261,3 +261,135 @@ def test_get_params_from_args_with_multiple_declarations(self):
...
+        self.assertIn(k1, env_params)
+        self.assertIn(k2, env_params)
+        self.assertEqual(v1, env_params[k1])
+        self.assertEqual(v2, env_params[k2])
+
+    def test_correctly_process_inherit_env_when_parameters_set(self):
+        """test_correctly_process_inherit_env_when_parameters_set
+
+        This tests that we still correctly pass through the environment variables
+        when --inherit-env is set and we run a job that has action parameters set
```
cognifloyd, line 309:
```suggestion
        when --inherit-env is set and we run an action that has action parameters set
```

cognifloyd (approved):
Minor nit: st2 does not have anything called a "job".

Otherwise, looks like a clean fix.

Issue #5766, opened 2022-10-12
Stackstorm Authentication using proxy mode in HA


ManuGeodis:
I have configured an ha infrastructure in stackstorm as mentioned in the docs. To set proxy auth it is said that CGI variable needs to be set. REMOTE_ADDR and REMOTE_USER. Can you suggest how to set these variables to set proxy auth?

PR #6041, opened 2023-10-22
Allow st2web proxy auth mode to work in HA environments

floatingstatic:
It appears that proxy auth mode only works with `REMOTE_USER` set as a cgi environment variable. This does not appear to work for those of us using the [HA helm chart](https://github.com/StackStorm/stackstorm-k8s). I saw #5766 but it does not appear this has been fixed or addressed yet. 

I had previously worked around this by modifying st2auth with a custom standalone auth module that always returns true without inspecting the username but it seems it would be beneficial to the wider community to get this working with the built-in proxy mode in st2auth. My use case is to continue to use Google IAP (identify aware proxy) in GKE which can forward remote user info to st2web (nginx) without having to roll patched version of st2auth and st2web to support this.


armab (commented):
Great stuff, thanks a lot @floatingstatic for the contribution!
Are there any Unit Tests that you could add to support this PR?

floatingstatic:
@armab good call. Added a unit test to cover this. Looks like there is already a test case that would cover lack of header and no remote user env var. Let me know if you think we need any other cases.

floatingstatic:
Theres one test that failed unrelated to my change: https://github.com/StackStorm/st2/actions/runs/6604759729/job/17939241233?pr=6041

Perhaps something flaky with that particular test? It seems it passed everywhere else.

armab (approved):
LGTM. Thanks for the fix!

cognifloyd (approved):

rameessahlu:
If it's part of the 3.9.0 milestone, why do I see these changes in the 3.8 HA release?

nzlosh:
Thanks for pointing that out.

PR #6036, opened 2023-10-16
Use codecov-cli over python-codecov

FileMagic:
This is the PR related to Issue #6035

PR #6046, opened 2023-10-24
Fix Orquesta Integration Test Codecov Upload

FileMagic:
I forgot these changes in the last PR, and it was pointed out here:

https://github.com/StackStorm/st2/pull/6036#issuecomment-1775662190

Thanks for the patience


nzlosh (approved):

armab (approved):
Thanks!

PR #6049, opened 2023-10-25
Fix st2 cli client auth in st2auth proxy mode

floatingstatic:
Adding on to work in #6041 that fixed proxy auth mode for users connecting via st2web, it appears that the `st2` client (`st2client` container in the HA helm chart) connects directly to `st2auth` instead of connecting through nginx. As a result it is unable to authenticate with `st2auth` is in proxy mode as headers required for this to work are not present. We see `401 Client Error: Unauthorized` in this scenario. The `st2` client does however use basic auth and we can extract the username from that in this scenario to resolve the issue.

This change modifies the proxy auth handler to use the username from the `Authorization` header only in the case where proxy auth mode is enabled and when `remote_user` is not passed to the proxy auth handler function.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,7 @@ in development
 
 Fixed
 ~~~~~
-* Additional fixes for st2 client auth when proxy auth mode enabled
+* Additional fixes for st2 client auth when proxy auth mode enabled #6049
   Contributed by @floatingstatic
 
 * Fix issue with linux pack actions failed to run remotely due to incorrect python shebang. #5983 #6042
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,9 @@ in development
 
 Fixed
 ~~~~~
+* Additional fixes for st2 client auth when proxy auth mode enabled
+  Contributed by @floatingstatic
+
 * Fix issue with linux pack actions failed to run remotely due to incorrect python shebang. #5983 #6042
   Contributed by Ronnie Hoffmann (@ZoeLeah Schwarz IT KG)
 
--- a/st2auth/st2auth/handlers.py
+++ b/st2auth/st2auth/handlers.py
@@ -130,6 +130,25 @@ class ProxyAuthHandler(AuthHandlerBase):
         remote_addr = headers.get("x-forwarded-for", remote_addr)
         extra = {"remote_addr": remote_addr}
 
+        # Needed to support st2client which does not connect via st2web
+        if authorization and not remote_user:
+            try:
+                auth_value = base64.b64decode(authorization[1])
+            except Exception:
+                LOG.audit("Invalid authorization header", extra=extra)
+                abort_request()
+                return
+
+            split = auth_value.split(b":", 1)
+            if len(split) != 2:
+                LOG.audit("Invalid authorization header", extra=extra)
+                abort_request()
+                return
+
+            remote_user = split[0]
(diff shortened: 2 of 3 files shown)


CHANGELOG.rst
```diff
@@ -6,6 +6,9 @@ in development
 
 Fixed
 ~~~~~
+* Additional fixes for st2 client auth when proxy auth mode enabled
```
armab, line 9:
```suggestion
* Additional fixes for st2 client auth when proxy auth mode enabled #6049
```

armab (approved):
Thanks!

st2auth/st2auth/handlers.py
```diff
@@ -130,6 +130,25 @@ def handle_auth(
...
+                return
+
+            split = auth_value.split(b":", 1)
+            if len(split) != 2:
+                LOG.audit("Invalid authorization header", extra=extra)
+                abort_request()
+                return
+
+            remote_user = split[0]
+            if six.PY3 and isinstance(remote_user, six.binary_type):
```
amanda11, line 149: Do we need to have checks on whether its python3 any more?

amanda11 (approved):
Happy to merge, but just added comment as not sure we need to keep having six.PY3 checks any more.

st2auth/st2auth/handlers.py
floatingstatic: Probably not but i was just keeping consistent with what is used in other handlers, for example: https://github.com/StackStorm/st2/blob/32a243a2ed54c5233e10f08b69f0cb21e2f10cb7/st2auth/st2auth/handlers.py#L212-L216

`decode("utf-8")` is needed even without the conditional in my testing.

hamsterdolphin:
@floatingstatic would you mind sharing your helm chart which has proxy auth enabled? I added the following to the [st2 config section](https://github.com/StackStorm/stackstorm-k8s/blob/master/values.yaml#L87) in `values.yaml`:

```
[auth]
mode=proxy
 ```

However, I get the following pod error:
`Failed to authenticate with credentials provided in the config.
ERROR: 401 Client Error: Unauthorized
MESSAGE: Invalid or missing credentials for url: http://stackstorm-chart-st2auth:9100/tokens`

Seeing a full working helm chart with proxy auth mode would be very helpful to the community :)

**Comment by floatingstatic** [comment]
@hamsterdolphin I cannot share my chart as its a fork of upstream with a bunch of things I have changed for my own use case, however the only real change you need to enable proxy auth would be here:

https://github.com/StackStorm/stackstorm-k8s/blob/87425e0e3f88d9e89ff6a356162fd0fc87a52cec/values.yaml#L87-L93

```
st2:
  config: |
    [api]
    allow_origin = '*'
    # fixes no replicaset found bug;
    [database]
    # Connection and server selection timeout (in ms).
    connection_timeout = 5000
    [auth]
    mode = proxy
```

It seems like you did that already so I suspect what you are hitting is actually the fact that you need a few patches on the upstream containers you are probably pulling in for version 3.8.0. Internal to my org we pull in upstream images for 3.8.0 then build a custom image with some patches on top of the `st2auth` and `st2web` containers.

The `st2auth` patches are basically just the diffs from #6049 and #6041 the dockerfile looks something like:

image

Patches are basically just what you see in those two PRs

Then we build and push the image to an internal registry and point to those images in our helm chart values instead of the default upstream container images.

The `st2web` patches are basically what I outlined in previous comments in this PR. Hope this helps.

PR #6050, opened 2023-10-25
Update Orquesta to v1.6.0

armab:
Update Orquesta to v1.6.0 to be included in the next st2 release.
Orquesta v1.6.0 comes with some updated dependencies.

### DIFF
--- a/fixed-requirements.txt
+++ b/fixed-requirements.txt
@@ -27,11 +27,12 @@ lockfile==0.12.2
 MarkupSafe<2.1.0,>=0.23
 mongoengine==0.23.0
 # networkx v2.6 does not support Python3.6.  Update networkx to match orquesta
-networkx>=2.5.1,<2.6
+networkx>=2.5.1,<2.6; python_version < '3.7'
+networkx>=2.6<3; python_version >= '3.7'
 # networkx requires decorator>=4.3,<5 which should resolve to version 4.4.2
 # but the wheel on pypi does not say it supports python3.8, so pip gets
 # confused. For now, pin decorator to work around pip's confusion.
-decorator==4.4.2
+decorator==4.4.2; python_version < '3.7'
 # NOTE: Recent version substantially affect the performance and add big import time overhead
 # See https://github.com/StackStorm/st2/issues/4160#issuecomment-394386433 for details
 oslo.config>=1.12.1,<1.13
--- a/fixed-requirements.txt
+++ b/fixed-requirements.txt
@@ -26,14 +26,6 @@ lockfile==0.12.2
 # >=0.23 was from jinja2
 MarkupSafe<2.1.0,>=0.23
 mongoengine==0.23.0
-# networkx v2.6 does not support Python3.6.  Update networkx to match orquesta
-networkx>=2.5.1,<2.6; python_version < '3.7'
-# use patched v2.6 networkx for Python3.8 to match orquesta 
-networkx>=2.6<3; python_version >= '3.7'
-# networkx requires decorator>=4.3,<5 which should resolve to version 4.4.2
-# but the wheel on pypi does not say it supports python3.8, so pip gets
-# confused. For now, pin decorator to work around pip's confusion.
-decorator==4.4.2; python_version < '3.7'
 # NOTE: Recent version substantially affect the performance and add big import time overhead
 # See https://github.com/StackStorm/st2/issues/4160#issuecomment-394386433 for details
 oslo.config>=1.12.1,<1.13
--- a/fixed-requirements.txt
+++ b/fixed-requirements.txt
@@ -28,6 +28,7 @@ MarkupSafe<2.1.0,>=0.23
 mongoengine==0.23.0
 # networkx v2.6 does not support Python3.6.  Update networkx to match orquesta
 networkx>=2.5.1,<2.6; python_version < '3.7'
+# use patched v2.6 networkx for Python3.8 to match orquesta 
 networkx>=2.6<3; python_version >= '3.7'
 # networkx requires decorator>=4.3,<5 which should resolve to version 4.4.2
 # but the wheel on pypi does not say it supports python3.8, so pip gets
--- a/fixed-requirements.txt
+++ b/fixed-requirements.txt
@@ -26,8 +26,12 @@ lockfile==0.12.2
 # >=0.23 was from jinja2
 MarkupSafe<2.1.0,>=0.23
 mongoengine==0.23.0
-# required by orquesta
+# required by orquesta (networkx<2.6 for py3.6, networkx<3 for py3.8)
 networkx<3
+# networkx requires decorator>=4.3,<5 which should resolve to version 4.4.2
+# but the wheel on pypi does not say it supports python3.8, so pip gets
+# confused. For now, pin decorator to work around pip's confusion.
+decorator==4.4.2
 # NOTE: Recent version substantially affect the performance and add big import time overhead
 # See https://github.com/StackStorm/st2/issues/4160#issuecomment-394386433 for details
 oslo.config>=1.12.1,<1.13
--- a/fixed-requirements.txt
+++ b/fixed-requirements.txt
@@ -26,6 +26,8 @@ lockfile==0.12.2
 # >=0.23 was from jinja2
 MarkupSafe<2.1.0,>=0.23
 mongoengine==0.23.0
+# required by orquesta
+networkx<3
 # NOTE: Recent version substantially affect the performance and add big import time overhead
 # See https://github.com/StackStorm/st2/issues/4160#issuecomment-394386433 for details
 oslo.config>=1.12.1,<1.13
--- a/requirements-pants.txt
+++ b/requirements-pants.txt
@@ -29,7 +29,7 @@ mock
 mongoengine
 # Note: networkx v2.6 dropped support for Python3.6
 # networkx version is constrained in orquesta.
-networkx
+networkx<3
 orjson
 orquesta @ git+https://github.com/StackStorm/orquesta.git@v1.6.0
 # NOTE: Recent version substantially affect the performance and add big import time overhead
--- a/requirements-pants.txt
+++ b/requirements-pants.txt
@@ -31,7 +31,7 @@ mongoengine
 # networkx version is constrained in orquesta.
 networkx
 orjson
-orquesta @ git+https://github.com/StackStorm/orquesta.git@v1.5.0
+orquesta @ git+https://github.com/StackStorm/orquesta.git@v1.6.0
 # NOTE: Recent version substantially affect the performance and add big import time overhead
 # See https://github.com/StackStorm/st2/issues/4160#issuecomment-394386433 for details
 oslo.config>=1.12.1,<1.13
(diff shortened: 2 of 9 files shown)


fixed-requirements.txt
```diff
@@ -27,11 +27,12 @@ lockfile==0.12.2
 MarkupSafe<2.1.0,>=0.23
 mongoengine==0.23.0
 # networkx v2.6 does not support Python3.6.  Update networkx to match orquesta
-networkx>=2.5.1,<2.6
+networkx>=2.5.1,<2.6; python_version < '3.7'
+networkx>=2.6<3; python_version >= '3.7'
```
armab, line 31: The python version condition in requirements comes from `orquesta` here (https://github.com/StackStorm/orquesta/pull/255/files) and will be a difficulty of this PR which fails due to duplicate `networkx` records: https://app.circleci.com/pipelines/github/StackStorm/st2/4153/workflows/0b3b950e-38de-4056-be6c-d1e4a05ff2d8/jobs/15935

```
[bdist: st2api] [20:19:54]      /usr/bin/python3 ../scripts/fixate-requirements.py -s in-requirements.txt -f ../fixed-requirements.txt
[bdist: st2api] [20:19:55]      Traceback (most recent call last):
[bdist: st2api] [20:19:55]        File "../scripts/fixate-requirements.py", line 284, in <module>
[bdist: st2api] [20:19:55]          skip=skip,
[bdist: st2api] [20:19:55]        File "../scripts/fixate-requirements.py", line 195, in write_requirements
[bdist: st2api] [20:19:55]          'Duplicate definition for dependency "%s"' % (project_name)
[bdist: st2api] [20:19:55]      ValueError: Duplicate definition for dependency "networkx"
[bdist: st2api] [20:19:55]      Makefile:74: recipe for target '.stamp-requirements' failed
[bdist: st2api] [20:19:55]      make: *** [.stamp-requirements] Error 1
```

Looks like we don't support it
https://github.com/StackStorm/st2/blob/8f6cb468ff50a827ce50c9be6ba78b4ac90da263/scripts/fixate-requirements.py#L193-L198 and relying on a dict for dependency names additionally checking for duplicates.

@cognifloyd @nzlosh Thinking how to make it work 🤔 Do you know any hacks with the @git-like links or non-standard formats in requirements we can possibly use?

fixed-requirements.txt
```diff
@@ -27,11 +27,13 @@ lockfile==0.12.2
 MarkupSafe<2.1.0,>=0.23
 mongoengine==0.23.0
 # networkx v2.6 does not support Python3.6.  Update networkx to match orquesta
-networkx>=2.5.1,<2.6
+networkx>=2.5.1,<2.6; python_version < '3.7'
+# use patched v2.6 networkx for Python3.8 to match orquesta 
+networkx>=2.6<3; python_version >= '3.7'
```
cognifloyd, line 32:
```suggestion
networkx>=2.6,<3; python_version >= '3.7'
```

cognifloyd (commented):
I like unpinning much better.

armab:
I tried to go with updating `fixate-requirements.txt` script to support duplicate requirements with markers (for different python versions added in orquestra dependencies).

But went alternative way by loosen the `networkx` in the st2 requirements, because the specific version is requested and installed by the `orquesta` dependencies here:  https://github.com/Stealthii/orquesta/blob/7dadd4cfa0e01a2fd65a545fbcdaad363bee4c68/requirements.txt#L5-L7
```
networkx>=2.5.1,<2.6; python_version < '3.7'
networkx>=2.6,<3; python_version >= '3.7'
```

I'm getting want I want during the build stage:
- U18: [networkx==2.5.1](https://app.circleci.com/pipelines/github/StackStorm/st2/4158/workflows/11838dac-5adc-4554-9a87-f7bf506bedb6/jobs/15940/parallel-runs/0?filterBy=ALL&invite=true#step-109-806306_65)
- U20: [networkx==2.8.8](https://app.circleci.com/pipelines/github/StackStorm/st2/4158/workflows/11838dac-5adc-4554-9a87-f7bf506bedb6/jobs/15940/parallel-runs/1?filterBy=ALL&invite=true#step-109-806220_61)
- EL8: [networkx==2.8.8](https://app.circleci.com/pipelines/github/StackStorm/st2/4158/workflows/11838dac-5adc-4554-9a87-f7bf506bedb6/jobs/15940/parallel-runs/1?filterBy=ALL&invite=true#step-109-806220_61)

Now the only blocker is 🔴  `EL7` (py3.6) which is getting the `python_version` markers wrong for some reason taking [this](https://github.com/Stealthii/orquesta/blob/7dadd4cfa0e01a2fd65a545fbcdaad363bee4c68/requirements.txt#L7) requirement for `py3.8` instead:
https://app.circleci.com/pipelines/github/StackStorm/st2/4158/workflows/11838dac-5adc-4554-9a87-f7bf506bedb6/jobs/15940?invite=true#step-109-929748_119
```
[package: st2] [18:59:09]      ERROR: pip's legacy dependency resolver does not consider dependency conflicts when selecting packages. This behaviour is the source of the following dependency conflicts.
[package: st2] [18:59:09]      networkx 2.5.1 requires decorator<5,>=4.3, but you'll have decorator 5.1.1 which is incompatible.
WRONG ----> [package: st2] [18:59:09]      orquesta 1.6.0 requires networkx<3,>=2.6, but you'll have networkx 2.5.1 which is incompatible.
```


I'm guessing something is outdated related to the build env for EL7 that doesn't quite work with dependency markers.
@cognifloyd @nzlosh @amanda11  Any ideas to try?

armab:
Figured it out in https://github.com/StackStorm/st2-packages/pull/728, - which is RFR now.
Need to fix e2e tests first before merging this, but this PR is RFR too.

requirements-pants.txt
```diff
@@ -31,7 +31,7 @@ mongoengine
 # networkx version is constrained in orquesta.
 networkx
 orjson
-orquesta @ git+https://github.com/StackStorm/orquesta.git@v1.5.0
+orquesta @ git+https://github.com/StackStorm/orquesta.git@v1.6.0
```
armab, line 34: For pants `./pants generate-lockfiles --resolve=st2` I've got the following error: 
```
ERROR: Cannot install networkx and orquesta==1.6.0 because these package versions have conflicting dependencies.
ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
 
 The conflict is caused by:
     The user requested networkx
     orquesta 1.6.0 depends on networkx<3 and >=2.6
 
 To fix this you could try to:
 1. loosen the range of package versions you've specified
 2. remove package versions to allow pip attempt to solve the dependency conflict
```

armab: If I remove `networkx` from `requirements-pants.txt` I'm getting the following for pants lock generation:
```
ERROR: Could not find a version that satisfies the requirement networkx<3,>=2.6 (from orquesta)
ERROR: No matching distribution found for networkx<3,>=2.6
```
🤔

nzlosh (approved):

fixed-requirements.txt
```diff
@@ -26,8 +26,8 @@ lockfile==0.12.2
 # >=0.23 was from jinja2
 MarkupSafe<2.1.0,>=0.23
 mongoengine==0.23.0
-# networkx v2.6 does not support Python3.6.  Update networkx to match orquesta
-networkx>=2.5.1,<2.6
+# required by orquesta (networkx<2.6 for py3.6, networkx<3 for py3.8)
+networkx<3
```
amanda11, line 30: @armab what happens if you change networkx here to match orquesta, e.g. >=2.6 & < 3
and then re-generate the requirements files
and then re-generate pants.

Wondering if that is why it complains that it can't rectifiy.

amanda11: Though also orquesta looks a bit more complicated...
requirements.txt:networkx>=2.5.1,<2.6; python_version < '3.7'
requirements.txt:networkx>=2.6,<3; python_version >= '3.7'

So we have different dependencies for different python versions, I wonder if the pants generate lockfiles copes with that?

armab: @amanda11  Yeah, I tried that in a previous step.
See the build failures here: https://github.com/StackStorm/st2/pull/6050#discussion_r1372367633

Outside of build issues I've found later that pip under EL7 didn't work as expected with the environment markers:
https://github.com/StackStorm/st2/pull/6050#issuecomment-1781768549

After all the tries in `st2-packages` and `st2` the current approach is the way of less resistance and at least I got the packages build working (with the https://github.com/StackStorm/st2/pull/6050#issuecomment-1782933536 fix)

armab: > I wonder if the pants generate lockfiles copes with that?

On the pants lockfile side, I saw https://github.com/StackStorm/st2/blob/8f6cb468ff50a827ce50c9be6ba78b4ac90da263/lockfiles/st2.lock#L134
so technically it may support things, but I experienced issues with pip too, so 🤷‍♂️

cognifloyd: pants does not support using the legacy resolver. I'm not sure why the legacy resolver is required in this case, as it seems clear to me. I'll have to play with it, hopefully this week.

armab: @cognifloyd Thanks for looking into it 👍  Can't wait and I'm looking forward to merge this PR so I can continue updating the dependencies for the `v3.8.1` patch.

---
Some context on that.
For some unknown reason, the new pip resolver went into an infinite loop resolving the dependencies: https://app.circleci.com/pipelines/github/StackStorm/st2/4158/workflows/11838dac-5adc-4554-9a87-f7bf506bedb6/jobs/15940?invite=true#step-109-978547_123 and couldn't process that `networkx` with conditional `python_version` env markers failing the EL7 build:

```
[package: st2] [18:59:15]      ERROR: Cannot install orquesta==1.6.0 and st2==3.9.dev0 because these package versions have conflicting dependencies.
[package: st2] [18:59:15]      
[package: st2] [18:59:15]      The conflict is caused by:
[package: st2] [18:59:15]          st2 3.9.dev0 depends on networkx
[package: st2] [18:59:15]          orquesta 1.6.0 depends on networkx<3 and >=2.6
```

In this ^^ case `orquesta 1.6.0 depends on networkx<3 and >=2.6` is a wrong condition chosen, as `EL7` runs on py3.6 and should choose a different networkx for dependency. So that might be a bug in pip.

I've tried to use [the latest pip version available for py3.6](https://github.com/StackStorm/st2-packages/pull/728/commits/28ac7f79677accc4e2209702fc44a2da1cff7364), but it didn't help.
From the pip changelog there might be a relevant fix to try: ([When checking for conflicts in the build environment, correctly skip requirements containing markers that do not match the current environment.](https://pip.pypa.io/en/stable/news/#id210)) available in pip `22.1`, but we can't use it because py3.6 support was dropped in pip `22.0`.

---

My thinking that's the worst case, for the upcoming st2 `v3.8.1` patch we'll live with it and take what worked in this PR  that works with `st2-packages`. The new pants build is not going to go live for this patch anyway.
In st2 `v3.9.0` we can deprecate python 3.6, switch to a new pip dependency resolver, latest pip and remove python env markers alltogether. That'll be quite an exercism that should potentially fix some bugs and technical debt 😈

amanda11 (approved):

PR #6055, opened 2023-11-02
Update cryptography and pyOpenSSL (security)

armab:
Update cryptography as well as compatible with it version of pyOpenSSL and paramiko.
* https://cryptography.io/en/latest/changelog/#v39-0-1
* https://www.pyopenssl.org/en/23.1.0/changelog.html
* https://www.paramiko.org/changelog.html#2.11.0


armab:
Yay, RFR!
That was quite a rabbit hole 🐰

cognifloyd (approved):

PR #6071, opened 2023-11-26
Fix ST2 Client for Windows Clients

philipphomberger:
We have try to rollout the st2client on the Windows 10/11 clients of our StackStorm users. But because PWD only exist on Unix Based Operation System it failed.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,9 @@ in development
 
 Fixed
 ~~~~~
+* ‎Fix ST2 Client for Windows Clients. PWD is a Unix only Libary. 
+  Contributed by (@philipphomberger Schwarz IT KG)
+
 * ‎Fix Snyk Security Finding Cross-site Scripting (XSS) in contrib/examples/sensors/echo_flask_app.py
   Contributed by (@philipphomberger Schwarz IT KG)
 
--- a/st2client/st2client/base.py
+++ b/st2client/st2client/base.py
@@ -37,11 +37,16 @@ from st2client.client import Client
 from st2client.config import get_config
 from st2client.utils.date import parse as parse_isotime
 from st2client.utils.misc import merge_dicts
+import platform
 
 __all__ = ["BaseCLIApp"]
 
 # Fix for "os.getlogin()) OSError: [Errno 2] No such file or directory"
-os.getlogin = lambda: pwd.getpwuid(os.getuid())[0]
+# Add Plattform Check to fix the Issue that PWD not exist on Windows and so the CLI not working.
+if platform.system() == "Windows":
+    os.getlogin = lambda: os.environ.get("USERNAME")
+else:
+    os.getlogin = lambda: pwd.getpwuid(os.getuid())[0]
 
 # How many seconds before the token actual expiration date we should consider the token as
 # expired. This is used to prevent the operation from failing durig the API request because the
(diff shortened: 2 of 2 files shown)


CHANGELOG.rst
```diff
@@ -6,6 +6,10 @@ in development
 
 Fixed
 ~~~~~
+
+* ‎Fix ST2 Client for Windows Clients. PWD is a Unix only Libary. 
+  Contributed by (@philipphomberger Schwarz IT KG)
+
 * ‎Fix Snyk Security Finding Cross-site Scripting (XSS) in contrib/examples/sensors/echo_flask_app.py
   Contributed by (@philipphomberger Schwarz IT KG)
```
armab, line 14: You probably have a trailing whitespace which fails linting.

Please also include the # PR numbers for the both Changelog records

armab:
For the record, what's the error you're getting on Windows systems?

philipphomberger:
> For the record, what's the error you're getting on Windows systems?

image

nzlosh:
Based on the proposed fix `os.getlogin()` returns the correct information on Windows.   I don't know the context around the patch, but the workaround for `getlogin` was added 11/07/2017 so I question it's relevance.  I checked on the currently supported OS' and a couple of others.  They all function as expected, so I'd be inclined to use the standard `os.getlogin()` function to have cross-platform support provided by Python, rather than maintaining these exceptions in the StackStorm code base.

**CentOS Linux 8**
```
>>> sys.version
'3.6.8 (default, Sep 10 2021, 09:13:53) \n[GCC 8.5.0 20210514 (Red Hat 8.5.0-3)]'

>>> os.getlogin()
'root'
```

**Ubuntu 20.04.6 LTS**

```
>>> sys.version
'3.8.10 (default, May 26 2023, 14:05:08) \n[GCC 9.4.0]'

>>> os.getlogin()
'root'
```

**Debian GNU/Linux 10 (buster)**

```
>>> sys.version
'3.7.3 (default, Jun 29 2023, 18:03:57) \n[GCC 8.3.0]'

>>> os.getlogin()
'root'
```

**Rocky Linux 9.2 (Blue Onyx)**
```
>>> sys.version
'3.9.16 (main, May 29 2023, 00:00:00) \n[GCC 11.3.1 20221121 (Red Hat 11.3.1-4)]'

>>> os.getlogin()
'root'
```

**Ubuntu 18.04.6 LTS**

```
>>> sys.version
'3.6.9 (default, Mar 10 2023, 16:46:00) \n[GCC 8.4.0]'

>>> os.getlogin()
'root'
```

**Ubuntu 22.04.3 LTS**

```
>>> sys.version
'3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0]'

>>> os.getlogin()
'root'
```

philipphomberger:
Hi Carlos, thank you for your Feedback.
Unfortunately, I also don't know why exactly the original fix was installed.
I can only comprehend the Problem in my Github Codespace Linux.
grafik
But I think this is a Microsoft special Linux. 
Since the problem is in the supported versions, it would be better to use the integrated Python function as you said instead of having the special lock in the code.

nzlosh:
I suspect you might be hitting this issue with your environment https://github.com/python/cpython/issues/84998.  As per the recommendation in the issue, perhaps `getpass.getuser()` is a more robust solution?  Does it work on windows 10/11?

armab (approved):
Looks good,
Thanks for the fix!

nzlosh (approved):

PR #6073, opened 2023-11-27
For "local-shell-script" runner, on readonly filesystems, don't attempt to run chmod +x on script_action.

### LINKED ISSUE #5591: "local-shell-script" does not work on read only file systems
## SUMMARY

While using the action runner type "local-shell-script", the action always try to set execute permission for the action script before the execution. 

For example, action runner logs for the core.sendmail action 

```
2022-03-09 07:02:06,866 INFO [-] Executing action via LocalRunner: dcc6c28d-5674-4cc2-9174-ca6c1e7b23a3
2022-03-09 07:02:06,866 INFO [-] [Action info] name: sendmail, Id: 622850eead44b087ebe77063, command: chmod +x /opt/stackstorm/packs/core/actions/send_mail/send_mail ; sudo -E -H -u stanley -- bash -c '/opt/stackstorm/packs/core/actions/send_mail/send_mail None stanley test@testmail.com '"'"'validate st2 local-shell-script runner'"'"' 1 text/html '"'"'validate st2 local-shell-script runner content'"'"' '"'"''"'"'', user: stanley, sudo: False
```
This "chmod +x " operation cause failure when you are using stackstorm-ha helm chart based deployment with an immutable file system for packs. 

https://docs.stackstorm.com/install/k8s_ha.html#method-1-st2packs-images-the-default 

For example, my custom pack deployed via the st2pack image method fails to run the action since the "chmod +x" instruction cannot be executed on a read-only file system. 

```
{
  "failed": true,
  "succeeded": false,
  "return_code": 1,
  "stdout": "",
  "stderr": "chmod: changing permissions of '/opt/stackstorm/packs/workflow/actions/verify_env.sh': Read-only file system"
}
```
**Logs:**

```
2022-03-08 13:06:39,736 INFO [-] Executing action via LocalRunner: a2bdfba5-682b-47b2-9485-4241b94501cf
2022-03-08 13:06:39,737 INFO [-] [Action info] name: verify_env, Id: 622754dfb56ccf38de50a292, command: chmod +x /opt/stackstorm/packs/workflow/actions/verify_env.sh ; sudo -E -H -u stanley -- bash -c '/opt/stackstorm/packs/workflow/actions/verify_env.sh '"'"'MY_ARGUMENTS_HERE'"'"'', user: stanley, sudo: False
2022-03-08 13:06:39,748 INFO [-] [622754de749089f0a983f6c3] Action execution "622754dfb56ccf38de50a293" for task "verify_env" is updated and in "running" state.
2022-03-08 13:06:39,848 INFO [-] [622754de749089f0a983f6c3] Action execution "622754dfb56ccf38de50a293" for task "verify_env" is updated and in "failed" state.
2022-03-08 13:06:39,855 INFO [-] Found 0 rules defined for trigger core.st2.generic.actiontrigger
2022-03-08 13:06:39,856 INFO [-] No matching rules found for trigger instance 622754dfe3850ebc7fd8d8c8.
```

I guess it's failing because of the condition mentioned here below for the local shell script runner. 

https://github.com/StackStorm/st2/blob/master/contrib/runners/local_runner/local_runner/base.py#L121

### STACKSTORM VERSION

```
root@m1-staging-st2actionrunner-6664d699c-55z5t:/opt/stackstorm# st2 --version
st2 3.6.0, on Python 3.6.9
```

stackstorm-ha helm deployed k8s environment running on EKS cluster 

### Steps to reproduce the problem

Follow the method mentioned here to deploy a custom st2 pack image and try to execute the action with runner type as "local-shell-script"

https://docs.stackstorm.com/install/k8s_ha.html#method-1-st2packs-images-the-default

### Expected Results

The action executes fine without a "chmod: changing permissions of '/opt/stackstorm/packs/custom-pack/actions/custom-script.sh': Read-only file system" failure and return the results.

### Actual Results

```
{
  "failed": true,
  "succeeded": false,
  "return_code": 1,
  "stdout": "",
  "stderr": "chmod: changing permissions of '/opt/stackstorm/packs/workflow/actions/verify_env.sh': Read-only file system"
}
```

Thanks!

jk464:
Fixes #5591.

For `local-shell-script` runners, the runner tries to ensure the `script_action` is executable by running: 

```
        if script_action:
            script_local_path_abs = self.entry_point
            args = "chmod +x %s ; %s" % (script_local_path_abs, args)
            sanitized_args = "chmod +x %s ; %s" % (
                script_local_path_abs,
                sanitized_args,
            )
```

On readonly filesystems, such as when StackStorm is deployed in HA on Kubernetes - this `chmod +x` command fails:

```
stderr: "chmod: changing permissions of '/opt/stackstorm/packs/ansible/actions/ansible.py': Read-only file system
```

(Note since this command isn't `&&` with the actual action, that still runs fine - there's just an error logged in the action output)

This is fixed, by only running this block if the filesystem is read-writeable - by adding the following to the conditional check:

```
bool(os.statvfs(self.entry_point).f_flag & os.ST_RDONLY)
```

Which returns `True` if the filesystem that `self.entry_point` (which is the script that will attemped to be `chmod +x`'ed) is on is readonly, and `False` otherwise.

This fix maintains functionally where there this worked previously (read-writeable fs), while fixing the bug on readonly fs. As per my comment on #5591, I think we should just drop this block all together - but that could be a change in `3.9.0` - for now this fixes the bug without changing anything else. :)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -44,6 +44,9 @@ Fixed
 * Update importlib-metadata from 3.10.1 to 4.8.3 for py3.6 and to 4.10.1 for py3.8 (security). #6072
   Contributed by @jk464
 
+* For "local-shell-script" runner, on readonly filesystems, don't attempt to run chmod +x on script_action. Fixes #5591
+  Contributed by @jk464
+
 Added
 ~~~~~
 * Move `git clone` to `user_home/.st2packs` #5845
--- a/contrib/runners/local_runner/local_runner/base.py
+++ b/contrib/runners/local_runner/local_runner/base.py
@@ -116,7 +116,10 @@ class BaseLocalShellRunner(ActionRunner, ShellRunnerMixin):
         sanitized_args = action.get_sanitized_full_command_string()
 
         # For consistency with the old Fabric based runner, make sure the file is executable
-        if script_action:
+        # Also check to ensure not Read-only file system
+        if script_action and not bool(
+            os.statvfs(self.entry_point).f_flag & os.ST_RDONLY
+        ):
             script_local_path_abs = self.entry_point
             args = "chmod +x %s ; %s" % (script_local_path_abs, args)
             sanitized_args = "chmod +x %s ; %s" % (


armab (approved):
Looks good, thanks for the fix!

nzlosh (approved):

PR #6079, opened 2023-12-01
Add Support for MongoDB 5, 6, 7 PYMONGO 4.6.1  Drop Python 3.6

philipphomberger:
This is a breaking Change because it drop Python 3.6. But I think that is the plan for Update 3.9. 

2. Update PYMONGO to 4.6.1
3. Update mongoengine to 0.27.0
4. Remove Some Tets and Code because ssl_match_hostname parameter not exist in newer Versions. 

I have tested it on my dev servers with 3 Different Mongo DB Setups:
1) mongodb flex Replika Set on Stackit Cloud Service (With SSL / User / Password)
2) mongo db single Stackit Cloud Service (With SSL / User / Password)
3) mongodb 7 local installation on devserver (Without SSL, Without User, Without Password)


armab (changes requested):
Good effort!
Please split changes into a two dedicated PRs.

One is about removing Python3.6 and another one about MongoDB (+pymongo) changes.
That'll help keeping things organized, good for history, testing and separates concerns.

nzlosh:
According to the migration documentation `ssl_match_hostname` functionality was not dropped but replaced by `tlsAllowInvalidHostnames` https://pymongo.readthedocs.io/en/stable/migrate-to-pymongo4.html#renamed-uri-options

philipphomberger:
> According to the migration documentation `ssl_match_hostname` functionality was not dropped but replaced by `tlsAllowInvalidHostnames` https://pymongo.readthedocs.io/en/stable/migrate-to-pymongo4.html#renamed-uri-options

okey. so id would be better to add this tlsAllowInvalidHostnames to replace the old ssl_match_hostname right ?

nzlosh:
I think the work for adding support for mongodb 5+ will require working through the official `migrate-to-pymongo4` document, checking the code base across the St2 repos (st2, orquesta, rbac, ldap, perhaps others) and update functions/methods/classes that reference the old v3 API.

The document recommends using pymongo 3.12 and turn on deprecation warnings to easily spot of the deprecated code.  There's also the fact that it should not be tested against python 3.6, but rather python3.8, which is to be our lowest supported version for the 3.9.0 release.

We have the choice of keeping `ssl_*` configuration names and transpose them to `tls` equivalent so that users don't need to adjust configuration files or update to use `tls` to reflect the correct name of the underlying technology being used.  Perhaps keeping ssl and tls versions for st2 v3.9.0 and with deprecation warnings of ssl version being removed in st2 v3.10

philipphomberger:
> I think the work for adding support for mongodb 5+ will require working through the official `migrate-to-pymongo4` document, checking the code base across the St2 repos (st2, orquesta, rbac, ldap, perhaps others) and update functions/methods/classes that reference the old v3 API.
> 
> The document recommends using pymongo 3.12 and turn on deprecation warnings to easily spot of the deprecated code. There's also the fact that it should not be tested against python 3.6, but rather python3.8, which is to be our lowest supported version for the 3.9.0 release.
> 
> We have the choice of keeping `ssl_*` configuration names and transpose them to `tls` equivalent so that users don't need to adjust configuration files or update to use `tls` to reflect the correct name of the underlying technology being used. Perhaps keeping ssl and tls versions for st2 v3.9.0 and with deprecation warnings of ssl version being removed in st2 v3.10

Thanks for the explanation. I thought it was too easy to just pull up the version.

cognifloyd:
I haven't read the PR, so this is a drive by comment after our TSC meeting.

Our codebase currently supports:
```
# mongoengine 0.24.0 has breaking changes to support pymongo 4.0
mongoengine>=0.21.0,<0.24.0
# pymongo 3.13 has backports of APIs from pymongo 4 to help w/ migration
pymongo>=3.11.0,<3.13.0
```

So, to update mongoengine/pymongo we need to migrate our code base based on:
https://docs.mongoengine.org/changelog.html#changes-in-0-24-0
https://pymongo.readthedocs.io/en/stable/migrate-to-pymongo4.html
https://pymongo.readthedocs.io/en/stable/changelog.html#changes-in-version-4-0

nzlosh:
Thank you @philipphomberger for proposing this PR, it provided valuable insight into challenges to support MongoDB >4.4 as well as dropping Python3.6.  These features have been integrated into the master branch so this PR can be closed.

PR #6081, opened 2023-12-04
Draft PR for Update all Dependencys (after Drop of Python 3.6)

philipphomberger:
Draft PR for Update all Dependencys (after Drop of Python 3.6)

Target: Update Dependencys that can update without change in the code.
Write down which Dependencys make problems to create special prs for them.


guzzijones:
we have implemented this in another pr.

PR #6101, opened 2023-12-21
pass without exception if jinja parameter isn't found

guzzijones:
If a jinja parameter doesn't exist as an input parameter then do not throw an exception and leave the value as a string parameter.  
We have some input paramters like:
```
{ "test": "{{some_text_here}}"}
```
These cause stackstorm to attempt to resolve `some_text_here` as a variable since it has matching {{ and }}.

This pr would allow those through as normal strings if they cannot be resolved.


armab:
Hey @guzzijones, please check Slack as your participation is needed here:
https://stackstorm-community.slack.com/archives/C020U70RZRR/p1702583037270339

cognifloyd (commented):
Maybe we could:
- collect a list of nodes (vars) that aren't valid as template vars
- if ANY of the vars DO exist, throw an error for the missing nodes.
- if NONE of the vars exist, then assume it is a string and ignore.

Otherwise, I'm concerned this could hide misspellings.

guzzijones:
I don't understand: 
> collect a list of nodes (vars) that aren't valid as template vars

guzzijones:
Most, if not all, my actions already have a jinja template parameter for {{config_context.var}}.  So I don't think check if ALL are parameters would work in my case.  

We have a log of log files with `{{...}}` in the json as a string.  These are unable to be processed by Stackstorm currently.  
I actually don't understand the use case for putting `{{...}}` templating parameter evaluation inside stackstorm except for the case of config context or secrets from the KV store.  

One can always just utilize a parameter in their python script.

nzlosh:
I'd rather see a more deterministic method to processing (or not) template strings.  For example, adding a new data type `raw_string`, which will not apply attempt to process the input variable and treat it at as raw string until it is handed off to the workflow/action.  `string` would remain unchanged and process strings against template detection.

st2common/st2common/util/param.py
```diff
@@ -174,13 +174,22 @@ def _validate(G):
     """
     Validates dependency graph to ensure it has no missing or cyclic dependencies
     """
+    g_copy = G.copy()
     for name in G.nodes:
```
cognifloyd, line 178: These are the nodes I'm talking about. Each "node" in the graph represents a named var in a string.

Within a single string, either all of the variables (nodes) should be valid, or none of them. If none of them are valid, then pass the string as-is.

You mentioned parameters, but I'm only talking about a single string, like the default value of a parameter. If that string does not have any valid cars, then assume it is not actually a Jinja template.

guzzijones: It makes sense now. Thanks for the explanation.

guzzijones: dependencies are added as nodes when it does `meta.find_undeclared_variables` and `G.add_edge(dependency, name)`.  At this point the nodes are all in the graph.

guzzijones: each root node should represent an original parameter and the edges should point to jinja template variable nodes

guzzijones: each node represents all vars including parameters.  not just named vars in strings.

guzzijones: In order to do what you ask the `_resolve_dependencies` would need to instead perform the removal of the invalid edges and nodes if all 'predecessors' or 'children' are invalid. (yes it is backwards)


@nzlosh the problem with raw_string option is it doesn't solve objects that have embedded jinja in them.

guzzijones: or I could intelligently process the root nodes first then iterate through children with recursion in `_validate`.  currently it just iterates through nodes without any order.

guzzijones: I did it, but I don't really like it.  I ended up having to add a lot of complexity.

guzzijones: I still need to add a unit test for the raised exception in the case of some but not all dependent variables exist.  Also, I have no idea if black or pylint will complain.

st2common/st2common/util/param.py
```diff
@@ -170,17 +170,80 @@ def _process_defaults(G, schemas):
...
+    for name in nodes:
+        if "value" not in G.nodes[name] and "template" not in G.nodes[name]:
+            # this is a string not a jinja template;  embedded {{sometext}}
+            check_any_bad.append(False)
+        else:
+            check_any_bad.append(True)
+        children = [i for i in G.predecessors(name)]
+        if name not in tracked_parents:
+            tracked_parents.extend(nodes)
+            _check_any_bad(G, children, check_any_bad, tracked_parents)
```
cognifloyd, line 193:
```suggestion
        if name not in tracked_parents:
            tracked_parents.extend(nodes)
            children = [i for i in G.predecessors(name)]
            _check_any_bad(G, children, check_any_bad, tracked_parents)
```
Is there a reason to calculate children on every iteration? Or can we move it into this execute-once block?

st2common/st2common/util/param.py
```diff
@@ -170,17 +170,80 @@ def _process_defaults(G, schemas):
...
+    :param check_any_bad: list[str]
+    :param tracked_parents: list[str]
+    """
+
+    if tracked_parents is None:
+        tracked_parents = [parent]
+    else:
+        tracked_parents.append(parent)
+    for i in nodes:
+        g_copy.nodes[parent]["value"] = g_copy.nodes[parent].pop("template")
```
cognifloyd, line 210: Is there a case where value would be populated but template would not?

guzzijones: yes, for nodes that are not jinja  templates.  They only have a value.

st2common/st2common/util/param.py
guzzijones: I am only processing the direct children here so it needs to happen for each iteration.  Also, the function is recursive to make the code simpler.

st2common/st2common/util/param.py
guzzijones: I added a check for the "templates" key.  I actually just bumped into this on my own instance running the fork with this fix.

st2common/st2common/util/param.py
guzzijones: oh i see what you meant. yes we could do it that way.

guzzijones:
It would be nice to get this merged in.  We have a lot of logs with `{{` strings in them that we feed into st2.

guzzijones:
Can we get some eyes on this?

st2common/st2common/util/param.py
```diff
@@ -170,17 +170,81 @@ def _process_defaults(G, schemas):
...
+    """
+
+    if tracked_parents is None:
+        tracked_parents = [parent]
+    else:
+        tracked_parents.append(parent)
+    for i in nodes:
+        if "template" in g_copy.nodes[parent].keys():
+            g_copy.nodes[parent]["value"] = g_copy.nodes[parent].pop("template")
+        if i in g_copy.nodes:
```
cognifloyd, line 212: `i` is reused here a couple of times. Could you rename this one to something like `node` to make its purpose clearer?

st2common/st2common/util/param.py
```diff
@@ -170,17 +170,81 @@ def _process_defaults(G, schemas):
...
+            children = [i for i in g_copy.predecessors(i)]
+            if children:
+                if i not in tracked_parents:
+                    _remove_bad(g_copy, i, children)
```
cognifloyd, line 216: Does this line need to pass in `tracked_parents`? If not, why not?

st2common/st2common/util/param.py
```diff
@@ -170,17 +170,81 @@ def _process_defaults(G, schemas):
...
+        tracked_parents = [parent]
+    else:
+        tracked_parents.append(parent)
+    for i in nodes:
+        if "template" in g_copy.nodes[parent].keys():
+            g_copy.nodes[parent]["value"] = g_copy.nodes[parent].pop("template")
+        if i in g_copy.nodes:
+            children = [i for i in g_copy.predecessors(i)]
+            if children:
+                if i not in tracked_parents:
```
cognifloyd, line 215: I don't see anywhere that adds the children to tracked_parents while walking the tree which means the `if i not in tracked_parents:` condition will always be true.

cognifloyd (commented):
Tree traversal logic is confusing for me. 😅 Here's a few more questions as I try to grok this logic.

cognifloyd:
I can follow the logic in this PR, and it seems like it would be functional. So, I give a technical approval of the implementation.

I'm still hesitant on this change. I'm worried about backwards incompatibility and silently ignoring actual errors. I also see how st2 is currently raising false positive errors for strings that are not Jinja.

If other members of the TSC agree with this change, then I'll approve it.

nzlosh:
I'd still prefer a deterministic solution to applying expression rendering to a parameter because I think this change is going to produce more confusion and loss of time for the general use case of typos or missing variables that silently pass through the input validation process.

Feeding raw logs into an action seems like an edge use case and comes with less control than human input.  My initial thought was to suggest log entries are sanitised before calling the action but this suffers from altering logs which would prevent or hinder further processing.

Since a new data type `raw_string` is an insufficient solution, I suggest we should add a new parameter flag that enables/disables the expression renderer for all data types.  The `renderer` flag would control input processing behaviour in a similar way to `immutable` or `secret` and gives users explicit control of rendering for any input variable data type.  The flag would be of type `boolean` and default to `true` to maintain backward compatibility.

This would allow ingest of raw logs without matching jinja expressions and still provide correct input validation processing with jinja expression rendering for the general use case.  The only use case that would not be handled by this would be when someone attempts to send a mixture of rendered and non-rendered jinja expressions in the same input.  In the case of mixed expression patterns, post-input processing would be required to tease out the parts needed to be rendered verses non-rendered.

Thoughts?

guzzijones:
I actually like the idea of the render flag.

guzzijones:
This PR is basically dead.  I am going to close this and do a 'render flag' PR.  @nzlosh @cognifloyd sound good?

PR #6102, opened 2023-12-22
Add output schemas for runners that just have a generic response

rebrowning:
add a generic output schema to the local, remote, and winrm runners. St2 already appears to return the data in this format, this is simply formalizing the response so that validations can be added for commands leveraging these runners to better allow users of st2 to verify the integrity of the data their pipelines are processing

I'm curious about any issues people are aware of this would cause. Initially the thought was this would cause regressions in existing jobs, but the results are identical between the new and old changes, so those existing jobs would be unaffected. As I'm new to the st2 code base and the community in general I'm curious if there are any other potential issues that would prevent this change from moving forwards.

Also if anyone knows where the appropriate location to add tests for these cases is, please lmk. I did an initial dig through the code but didn't find where the serialization was happening... I'm beginning to think it's happening in the DB itself, but can't confirm that's the case.

I've provided 4 examples below with the core.local, which uses the local-shell-cmd runner. The first two examples are what currently happens, and the second two examples are with the change. You'll notice they're equivalent, but with the change it allows for validations to be executed.

### core.local -> `echo "hello world"`
result:
```
{
  "failed": false,
  "succeeded": true,
  "return_code": 0,
  "stdout": "hello world",
  "stderr": ""
}
```
### core.local -> `curl <http endpoint that returns json>`
the simple http endpoint in this example returns the following object:
```
{
  "name": "billy",
  "age": 15,
  "addr": {
      "street": "abc",
      "number": 123,
      "state": "zz"
  }
}
```
result in st2:
```
{
  "failed": false,
  "succeeded": true,
  "return_code": 0,
  "stdout": {
    "name": "billy",
    "age": 15,
    "addr": {
      "street": "abc",
      "number": 123,
      "state": "zz"
    }
  },
  "stderr": "  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current\n                                 Dload  Upload   Total   Spent    Left  Speed\n\r  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0\r100    85  100    85    0     0    123      0 --:--:-- --:--:-- --:--:--   123"
}
```

### core.local -> `echo "hello world"`
```
{
  "failed": false,
  "succeeded": true,
  "return_code": 0,
  "stdout": "hello world",
  "stderr": ""
}
```
### core.local -> `curl <same http endpoint as before>`
result:
```
{
  "failed": false,
  "succeeded": true,
  "return_code": 0,
  "stdout": {
    "name": "billy",
    "age": 15,
    "addr": {
      "street": "abc",
      "number": 123,
      "state": "zz"
    }
  },
  "stderr": "  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current\n                                 Dload  Upload   Total   Spent    Left  Speed\n\r  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0\r100    85  100    85    0     0    223      0 --:--:-- --:--:-- --:--:--   223"
}
```

## Json object inside of full result
core.local -> `echo "hello world" && curl https://sstor74x3lez4ohd4x4vszvpm40gbitd.lambda-url.us-east-1.on.aws/`  
This will not parse to json, so the resulting stdout will just be a string.
```
{
  "failed": false,
  "succeeded": true,
  "return_code": 0,
  "stdout": "hello world\n{\"name\": \"billy\", \"age\": 15, \"addr\": {\"street\": \"abc\", \"number\": 123, \"state\": \"zz\"}}",
  "stderr": "  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current\n                                 Dload  Upload   Total   Spent    Left  Speed\n\r  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0\r100    85  100    85    0     0    223      0 --:--:-- --:--:-- --:--:--   223"
}
```


contrib/runners/local_runner/local_runner/runner.yaml
```diff
@@ -33,6 +33,24 @@
...
+      succeeded:
+        type: boolean
+      failed:
+        type: boolean
+      stdout:
+        anyOf:
+          - type: "object"
+          - type: "string"
+          - type: "array"
+          - type: "number"
```
cognifloyd, line 49: Can this be any of the simple json types? Why these 4?
https://github.com/StackStorm/st2/blob/master/st2common/st2common/util/schema/action_params.json#L19

I guess we need to look at how stdout is converted from json to see if there are any restrictions.

PR #6120, opened 2024-01-30
Upgrade Pantsbuild to 2.18

cognifloyd:
Updating to pants 2.18 needed to wait for these which were released as part of 2.18.3:
- https://github.com/pantsbuild/pants/pull/20472 
- https://github.com/pantsbuild/pants/pull/20477

There is a new rust-based parser for python dependency inference, but in 2.18.2 it was spewing some confusing and inaccurate warnings with our code base. https://github.com/pantsbuild/pants/pull/20472 resolved those issues. We need to switch to the rust parser when we move to 2.18 or we risk a bigger transition with the move to 2.19, when the old python parser gets removed.

Now that those are merged, we get to use that new parser, and it feels much faster.


nzlosh (approved):

mamercad (approved):
:rubber-stamp:

PR #5995, opened 2023-06-22
zstandard compression for parameters and results

guzzijones:
# why
mitigate (as much as possible) 'document too large' mongo exceptions.  16 mb is the limit for mongodb document size.

# done
1.  zstandard compression for parameter and results
2.  remove embedded liveaction doc in executions db and replace with liveaction id string. This change will  potentially save 1/2 the time to write the liveaction to the database as the liveaction won't be written 2 times.
3.  added traceback info to exception handling for action execution engine and workflow engine.  This should make it easier to debug for someone when the workflow fails for some reason.  invalid orjson or unterminated yaql strings cause these exceptions in my experience.
4. This will require a data migration for the `executionDB` model for the `liveaction` paramter as it is now a string as opposed to an embedded document
5. I added a config option to turn off zstandard compression for parameters and results in the execution db and the liveaction db

PR #6121, opened 2024-02-05
Split out zstandard compression (split from #5995)

rebrowning:
splitting out the zstandard work done by @guzzijones in https://github.com/StackStorm/st2/pull/5995

If we would rather just have the original PR go through, lmk and I can close this. But I saw a comment on his PR that it'd be nice to split the two out.

All credit for work on this goes to @guzzijones , I'm splitting it out with the hope of moving it through so we can mitigate the document size errors.


guzzijones:
The history should not be deleted so we can preserve the git blame in the future, imo.

guzzijones:
Also, thanks, I know this is a lot of work.

rebrowning:
@guzzijones by that do you mean you'd rather me fork from your repo/branch and then add a commit removing the liveaction id change? If so I can look at tackling that this week.

PR #6138, opened 2024-02-16
Renew certs in ssl_certs fixture

### LINKED ISSUE #6137: Renew test SSL CA + Cert
Our test SSL CA+cert just expired. We need to renew it and document how to do so.

https://github.com/StackStorm/st2/tree/master/st2tests/st2tests/fixtures/ssl_certs

Since this is for testing, I think we could do something like a 15 year duration.

cognifloyd:
Fixes #6137

I wrote a script to renew all of the certs in the fixture. I focused on figuring out how to each of the files could have been initially created so that I could reproduce the same method during renewal.

I used a renewal period of 15 years so we don't have to deal with this very often.


nzlosh (approved):
LGTM, :+1:

guzzijones (approved):

cognifloyd:
I'm temporarily skipping the circlieci requirement so we can get this CI fix in. We can work on the CircleCI issue in a separate PR.

PR #6170, opened 2024-03-25
Add a Secret Masking to Reule Enfocement API View Output

philipphomberger:
Hi Community, 

From My View of perspective Secrets should not show in cleartext in the Rule Enforcement View in the UI and the API. So I add a loop that auto mask the secret to *****.
I'am not 100% happy with the way I do it for get one function because of problem I had with access the keys of the object So I convert the Object to a str and than with eval to a dict. 
Maybe someone here have a better Idea.

I wait for your feedback.
Thank you all.

### DIFF
--- a/st2api/st2api/controllers/v1/rule_enforcement_views.py
+++ b/st2api/st2api/controllers/v1/rule_enforcement_views.py
@@ -26,6 +26,9 @@ from st2common.rbac.types import PermissionType
 
 from st2api.controllers.resource import ResourceController
 
+import json
+
+
 __all__ = ["RuleEnforcementViewController"]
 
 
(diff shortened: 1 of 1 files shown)


nzlosh:
The idea seems good but the implementation needs a better architectural solution.   Code already exists to handle masking secrets, would it be possible to call that?

https://github.com/StackStorm/st2/blob/afedd35c4ec623aae5494eb3d3a3c935f261eda3/st2common/st2common/models/db/execution.py#L108

I've not tested this, but perhaps it can be called like below so secrets are masked?
```
        execution_dbs_by_id = {}
        for execution_db in execution_dbs:
            execution_dbs_by_id[str(execution_db.id)] = execution_db.mask_secrets(execution_db)
```

philipphomberger:
> The idea seems good but the implementation needs a better architectural solution. Code already exists to handle masking secrets, would it be possible to call that?
> 
> https://github.com/StackStorm/st2/blob/afedd35c4ec623aae5494eb3d3a3c935f261eda3/st2common/st2common/models/db/execution.py#L108
> 
> I've not tested this, but perhaps it can be called like below so secrets are masked?
> 
> ```
>         execution_dbs_by_id = {}
>         for execution_db in execution_dbs:
>             execution_dbs_by_id[str(execution_db.id)] = execution_db.mask_secrets(execution_db)
> ```

Thanks @nzlosh  I will try it. :)

philipphomberger:
@nzlosh I have tryed it:
But the functions have a strange behavior than I use them like that:

`        for rule_enforcement_api in rule_enforcement_apis:
            if "parameters" in str(rule_enforcement_apis[i]):
                secrets = get_secret_parameters(rule_enforcement_api['execution']['action']['parameters'])
                print(f"Parameters before masking: {rule_enforcement_api['execution']['parameters']}")
                rule_enforcement_api['execution']['parameters'] = mask_secret_parameters(parameters=rule_enforcement_api['execution']['action']['parameters'], secret_parameters=secrets)
                print(f"Parameters after masking: {rule_enforcement_api['execution']['parameters']}")
                rule_enforcement_apis[i] = rule_enforcement_api
        return rule_enforcement_apis`

Output:
2024-03-25 13:12:02 Parameters before masking: {'key1': 'test1234', 'key2': ’testsecret’}
2024-03-25 13:12:02 Parameters after masking: {'key1': {'type': 'string', 'default': '{{ st2kv.system.customer2_key1 }}'}, 'key2': '********'}

So Iooks like I can not 1:1 use that function. The Secret Masking it self working but the output for the none Secret Parameters get lost.

